Issue #423, opened 2019-11-21
Investigate and resolve mdns dial errors


ryanchristo:
## Expected Behavior

We should not have to disable mdns in our p2p tests.

## Current Behavior

Without disabling mdns in our p2p tests (ie, adding `NoMdns: true` to the configuration for each node that we start), we are receiving `failed dial` and `connection refused` in our debug logs.

## Possible Solution

This might be an issue with the generated addresses we are dialing, how we are configuring libp2p within the service, or an issue in the go-libp2p package that we need to investigate.

## Steps to Reproduce

Remove the `NoMdns: true` configuration property in any of the `p2p_test.go` tests and observe the `failed dial` and `connection refused` in the debug logs.

PR #427, opened 2019-11-25
p2p: add foundation for status message exchange

### LINKED ISSUE #325: Send Status Message on Connection to Peer
When we connect to a new peer we need to exchange status messages. If this fails we should drop the peer.

See spec 4.4.2 (and network messages section).

ryanchristo:
## Changes

This pull request adds the foundation for the status message exchange between host and connected peer that will determine if the two nodes are compatible or if the peer should be dropped.

This pull request also includes some formatting, updated comments, and some added logging.

## Tests
```
go test ./p2p -v
```

## Issues

closes #325

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -19,7 +19,6 @@ require (
 	github.com/libp2p/go-libp2p v0.4.1
 	github.com/libp2p/go-libp2p-core v0.2.5
 	github.com/libp2p/go-libp2p-kad-dht v0.3.0
-	github.com/libp2p/go-libp2p-peerstore v0.1.4
 	github.com/mattn/go-isatty v0.0.9 // indirect
 	github.com/multiformats/go-multiaddr v0.1.1
 	github.com/naoina/go-stringutil v0.1.0 // indirect
--- a/go.mod
+++ b/go.mod
@@ -19,6 +19,7 @@ require (
 	github.com/libp2p/go-libp2p v0.4.1
 	github.com/libp2p/go-libp2p-core v0.2.5
 	github.com/libp2p/go-libp2p-kad-dht v0.3.0
+	github.com/libp2p/go-libp2p-peerstore v0.1.4
 	github.com/mattn/go-isatty v0.0.9 // indirect
 	github.com/multiformats/go-multiaddr v0.1.1
 	github.com/naoina/go-stringutil v0.1.0 // indirect
--- a/p2p/config.go
+++ b/p2p/config.go
@@ -28,12 +28,9 @@ import (
 	"path/filepath"
 
 	log "github.com/ChainSafe/log15"
-
 	"github.com/libp2p/go-libp2p"
-
 	"github.com/libp2p/go-libp2p-core/crypto"
 	"github.com/libp2p/go-libp2p-core/peer"
-
 	ma "github.com/multiformats/go-multiaddr"
 )
 
--- a/p2p/config.go
+++ b/p2p/config.go
@@ -28,9 +28,12 @@ import (
 	"path/filepath"
 
 	log "github.com/ChainSafe/log15"
-	libp2p "github.com/libp2p/go-libp2p"
+
+	"github.com/libp2p/go-libp2p"
+
 	"github.com/libp2p/go-libp2p-core/crypto"
 	"github.com/libp2p/go-libp2p-core/peer"
+
 	ma "github.com/multiformats/go-multiaddr"
 )
 
@@ -78,16 +81,17 @@ func (c *Config) buildOpts() ([]libp2p.Option, error) {
 		return nil, err
 	}
 
-	connMgr := ConnManager{}
+	connmgr := &ConnManager{}
 
-	return []libp2p.Option{
-		libp2p.ListenAddrs(addr),
-		libp2p.DisableRelay(),
-		libp2p.Identity(c.privateKey),
+	options := []libp2p.Option{
 		libp2p.NATPortMap(),
-		libp2p.Ping(true),
-		libp2p.ConnectionManager(connMgr),
-	}, nil
+		libp2p.DisableRelay(),
+		libp2p.ListenAddrs(addr),
+		libp2p.Identity(c.privateKey),
+		libp2p.ConnectionManager(connmgr),
+	}
+
+	return options, nil
 }
 
 // setupPrivKey will attempt to load the nodes private key, if that fails it will create one
--- a/p2p/config.go
+++ b/p2p/config.go
@@ -49,7 +49,7 @@ type Config struct {
 	Port uint32
 	// If 0, random host ID will be generated; If non-0, deterministic ID will be produced, keys will not be loaded from data dir
 	RandSeed int64
-	// Disable bootstrapping altogether. BootstrapNodes has no effect over this.
+	// Disables bootstrapping
 	NoBootstrap bool
 	// Disables gossiping
 	NoGossip bool
--- a/p2p/config.go
+++ b/p2p/config.go
@@ -51,6 +51,8 @@ type Config struct {
 	RandSeed int64
 	// Disable bootstrapping altogether. BootstrapNodes has no effect over this.
 	NoBootstrap bool
+	// Disables gossiping
+	NoGossip bool
 	// Disables MDNS discovery
 	NoMdns bool
 	// Global data directory
--- a/p2p/config.go
+++ b/p2p/config.go
@@ -84,10 +84,11 @@ func (c *Config) buildOpts() ([]libp2p.Option, error) {
 	connmgr := &ConnManager{}
 
 	options := []libp2p.Option{
-		libp2p.NATPortMap(),
-		libp2p.DisableRelay(),
 		libp2p.ListenAddrs(addr),
+		libp2p.DisableRelay(),
 		libp2p.Identity(c.privateKey),
+		libp2p.NATPortMap(),
+		libp2p.Ping(true),
 		libp2p.ConnectionManager(connmgr),
 	}
 
--- a/p2p/connmgr.go
+++ b/p2p/connmgr.go
@@ -95,7 +95,7 @@ func Disconnected(n network.Network, c network.Conn) {
 func OpenedStream(n network.Network, s network.Stream) {
 	protocol := s.Protocol()
 	if protocol != "" {
-		log.Debug(
+		log.Trace(
 			"opened stream",
 			"host", s.Conn().LocalPeer(),
 			"peer", s.Conn().RemotePeer(),
@@ -108,7 +108,7 @@ func OpenedStream(n network.Network, s network.Stream) {
 func ClosedStream(n network.Network, s network.Stream) {
 	protocol := s.Protocol()
 	if protocol != "" {
-		log.Debug(
+		log.Trace(
 			"closed stream",
 			"host", s.Conn().LocalPeer(),
 			"peer", s.Conn().RemotePeer(),
--- a/p2p/connmgr.go
+++ b/p2p/connmgr.go
@@ -19,12 +19,13 @@ package p2p
 import (
 	"context"
 
-	"github.com/ChainSafe/gossamer/common"
 	log "github.com/ChainSafe/log15"
 
 	"github.com/libp2p/go-libp2p-core/connmgr"
-	net "github.com/libp2p/go-libp2p-core/network"
-	peer "github.com/libp2p/go-libp2p-core/peer"
+	"github.com/libp2p/go-libp2p-core/network"
+	"github.com/libp2p/go-libp2p-core/peer"
+
+	ma "github.com/multiformats/go-multiaddr"
 )
 
 // ConnManager implement connmgr.ConnManager
@@ -32,12 +33,16 @@ import (
 type ConnManager struct{}
 
 // Notifee is used to monitor changes to a connection
-func (cm ConnManager) Notifee() net.Notifiee {
-	nb := new(net.NotifyBundle)
+func (cm ConnManager) Notifee() network.Notifiee {
+	nb := new(network.NotifyBundle)
+
+	nb.ListenF = Listen
+	nb.ListenCloseF = ListenClose
 	nb.ConnectedF = Connected
+	nb.DisconnectedF = Disconnected
 	nb.OpenedStreamF = OpenedStream
 	nb.ClosedStreamF = ClosedStream
-	nb.DisconnectedF = Disconnected
+
 	return nb
 }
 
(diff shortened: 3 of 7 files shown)


p2p/host.go
```diff
@@ -82,112 +113,172 @@ func newHost(ctx context.Context, cfg *Config) (*host, error) {
...
-		}
-	}
+// `getStream` attempts to get an existing stream (using `getExistingStream`)
+// and opens a new stream if an existing stream does not exist.
+func (h *host) getStream(p peer.ID) (stream net.Stream, err error) {
+	// stream = h.getExistingStream(p)
 
-	return nil
+	// if stream == nil {
+	stream, err = h.h.NewStream(h.ctx, p, h.protocolId)
```
noot: this no longer returns an existing stream

ryanchristo: That is true. Using an existing stream was preventing both nodes from receiving the status message (both would write to the stream but only one would receive the status message).

I commented this out rather than deleting it because I think we would want to use an existing stream if available but we might need a condition to determine when. I will take another look and either figure out what condition we need or open an issue and update the comment.

p2p/p2p.go
```diff
@@ -19,77 +19,96 @@ package p2p
...
+	"github.com/ChainSafe/gossamer/common"
 	log "github.com/ChainSafe/log15"
 
+	"github.com/libp2p/go-libp2p-core/network"
 	net "github.com/libp2p/go-libp2p-core/network"
 )
 
-var _ services.Service = &Service{}
+// var _ module.P2pApi = &Service{}
+// var _ services.Service = &Service{}
```
ansermino: You can remove the first. The second is optional but should probably stay to ensure we don't change the service's compatibility with the interface.

If you're not familiar with these: they just assert that you can create an instance of `Service` and store it as a `services.Service` (ie. it adheres to the interface). It never actually gets saved due to the `_`, but the compiler will complain if its not possible.

p2p/p2p_test.go
```diff
@@ -181,26 +196,76 @@ func TestSendRequest(t *testing.T) {
...
 	}
 
-	err = nodeB.host.connect(*addrInfoA)
+	err = nodeA.host.connect(*addrInfoB)
 	if err != nil {
 		t.Fatal(err)
 	}
 
+	// Wait for status exchange
+	time.Sleep(5 * time.Second)
```
ansermino: Lets chat some more about how to remove this

p2p/host.go
ryanchristo: Ok, updated implementation in `send` method and opened #437

p2p/host.go
```diff
@@ -6,28 +22,29 @@ import (
...
 	ma "github.com/multiformats/go-multiaddr"
-
-	"github.com/libp2p/go-libp2p/p2p/discovery"
 )
 
 const DefaultProtocolId = protocol.ID("/gossamer/dot/0")
 const mdnsPeriod = time.Minute
 
-// host is a wrapper around libp2p's host.host
+// `host` describes a wrapper for the libp2p host, dht, and mdns services
```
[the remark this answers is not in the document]
ryanchristo: Fixed. Old habit from the land of generated javascript documentation.

p2p/p2p.go
```diff
@@ -99,25 +118,67 @@ func (s *Service) Stop() error {
...
 
-// MsgRecPoll starts polling the msgRec channel for any blocks
-func (s *Service) MsgRecPoll(e chan error) {
+// `sendStatusMessages` starts a loop that sends the current network state as
+// a status message to each connected peer every 5 seconds.
+func (s *Service) sendStatusMessages() {
 	for {
+
+		// Send status messages every 5 seconds
+		time.Sleep(5 * time.Second)
```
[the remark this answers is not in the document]
ryanchristo: Done. Set to 5 minutes as `StatusMessageInterval` constant.

p2p/p2p.go
ryanchristo: Thanks for the explanation! Deleted first and uncommented second.

p2p/p2p_test.go
ryanchristo: Yes, please. Definitely not ideal for CI testing.

p2p/host.go
```diff
@@ -82,123 +112,177 @@ func newHost(ctx context.Context, cfg *Config) (*host, error) {
...
-		stream, err = h.h.NewStream(h.ctx, peer.ID, h.protocolId)
-		log.Debug("opening new stream ", "to", peer.ID)
-		if err != nil {
-			log.Error("failed to open stream", "error", err)
-			return err
-		}
-	} else {
-		log.Debug("using existing stream", "to", peer.ID)
-	}
+	// TODO: investigate existing stream breaking status exchange
```
ryanchristo: opened #437 to address this

p2p/p2p.go
```diff
@@ -19,77 +19,92 @@ package p2p
...
+// Start starts the service
 func (s *Service) Start() error {
 	s.host.startMdns()
 	s.host.bootstrap()
-	s.host.logAddrs()
+	s.host.printHostAddresses()
 
-	log.Info("Listening for connections...")
+	// TODO: decide whether we need errors to surface here
+	// e := make(chan error)
```
ryanchristo: Worth pointing out. David mentioned the error channel might not be necessary here.

p2p/p2p.go
```diff
@@ -99,86 +114,193 @@ func (s *Service) Stop() error {
...
+func (s *Service) broadcastStatusMessages() {
+
+	// Wait for node to finish starting
+	time.Sleep(time.Second)
+
+	// TODO: use generated message
+	msg := statusMessage
+
+	// send message to each connected peer
+	s.host.broadcast(msg)
```
noot: we don't broadcast status messages, we send point-to-point them upon connection to a peer

noot: i think doing it like this may cause us not to be able to connect to peers. since it's 5 mins between status message sends, the peers will definitely disconnect from us, since they are expecting it within a smaller time period (<1 min most likely, will need to confirm)

p2p/p2p.go
```diff
@@ -99,86 +114,193 @@ func (s *Service) Stop() error {
...
+			"peer", stream.Conn().RemotePeer(),
+			"protocol", stream.Protocol(),
+			"message", msg,
+		)
+		return
+	}
 
-// ID returns the host's ID
+	// check and store message, returns true if valid new message
+	if !s.verifyNewMessage(msg) {
```
noot: naming may be a bit confusing, the message could still be valid but seen already. maybe change to `shouldBroadcast` or something?

p2p/p2p.go
ryanchristo: Correct, status messages need be sent whenever a peer is connected. This only send status messages to connected peers when a new service is started and then every 5 minutes thereafter.

p2p/p2p.go
ryanchristo: Good point. I can update.

p2p/host.go
```diff
@@ -51,154 +72,225 @@ func newHost(ctx context.Context, cfg *Config) (*host, error) {
...
+	for _, addr := range h.h.Addrs() {
+		fmt.Println(addr.Encapsulate(h.address).String())
 	}
 }
 
-func (h *host) registerStreamHandler(handler func(net.Stream)) {
+// registerConnHandler registers the connection handler (see handleConn)
+func (h *host) registerConnHandler(handler func(network.Conn)) {
+	h.h.Network().SetConnHandler(handler)
+}
```
ryanchristo: This ended up being much easier than the approach with the connection manager.

p2p/host.go
```diff
@@ -51,154 +73,245 @@ func newHost(ctx context.Context, cfg *Config) (*host, error) {
...
-		stream, err = h.h.NewStream(h.ctx, peer.ID, h.protocolId)
-		log.Debug("opening new stream ", "to", peer.ID)
-		if err != nil {
-			log.Error("failed to open stream", "error", err)
-			return err
-		}
-	} else {
-		log.Debug("using existing stream", "to", peer.ID)
-	}
+	// TODO: investigate get existing stream breaking tests
```
ryanchristo: addressed in #437

p2p/p2p.go
```diff
@@ -19,30 +19,46 @@ package p2p
...
+	host         *host
+	msgRec       <-chan Message
+	msgSend      chan<- Message
+	blockAnnRec  map[string]bool
+	blockReqRec  map[string]bool
+	blockRespRec map[string]bool
+	txMessageRec map[string]bool
+}
+
+// TODO: use generated status message
```
ryanchristo: addresses in #443

p2p/config.go
```diff
@@ -48,6 +51,8 @@ type Config struct {
 	RandSeed int64
 	// Disable bootstrapping altogether. BootstrapNodes has no effect over this.
 	NoBootstrap bool
+	// Disables gossiping
+	NoGossip bool
```
ryanchristo: This is a solution for testing that also seemed useful as a configuration option.

p2p/host.go
```diff
@@ -6,41 +22,47 @@ import (
...
 	dht         *kaddht.IpfsDHT
+	mdns        discovery.Service
 	bootnodes   []peer.AddrInfo
 	noBootstrap bool
+	noGossip    bool
 	noMdns      bool
-	mdns        discovery.Service
+	address     ma.Multiaddr
 	protocolId  protocol.ID
+	peerStatus  map[peer.ID]bool // TODO: investigate peer status storage options
```
ryanchristo: addressed in #444

p2p/p2p.go
```diff
@@ -51,134 +67,253 @@ func NewService(conf *Config, msgSend chan<- Message, msgRec <-chan Message) (*S
...
+
+// handleStreamStatus handles status messages written to the stream
+func (s *Service) handleStreamStatus(stream network.Stream, msg Message) {
+
+	// TODO: use generated status message
+	hostStatus := statusMessage
+
+	switch {
+
+	// TODO: implement status message validation
```
ryanchristo: addressed in #444

p2p/p2p_test.go
```diff
@@ -45,18 +48,70 @@ func TestStartService(t *testing.T) {
...
 	node := startNewService(t, config, nil, nil)
 	node.Stop()
 }
 
+func TestConnect(t *testing.T) {
+	configA := &Config{
+		Port:        7001,
+		RandSeed:    1,
+		NoBootstrap: true,
+		NoMdns:      true, // TODO: investigate failed dials, disable for now
```
ryanchristo: addressed in #423

p2p/host.go
```diff
@@ -209,31 +322,11 @@ func (h *host) peerCount() int {
 	return len(peers)
 }
 
-// close shuts down the host and its components
-func (h *host) close() error {
```
noot: was this removed?

ryanchristo: Nope just moved [here](https://github.com/ChainSafe/gossamer/pull/427/files#diff-697b58a8ad3efb7d1ebbd09bcf39efb0R123)

p2p/connmgr.go
```diff
@@ -19,63 +19,99 @@ package p2p
...
+		"disconnected",
+		"host", c.LocalPeer(),
+		"peer", c.RemotePeer(),
+	)
+}
+
+// OpenedStream is called when a stream opened
+func (cm *ConnManager) OpenedStream(n network.Network, s network.Stream) {
+	protocol := s.Protocol()
+	if protocol != "" {
```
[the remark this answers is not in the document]
ryanchristo: There are a number of streams with an empty protocol id being logged. This will only log opened streams with a protocol id. Essentially this cleans up the logs for debugging purposes.

p2p/host.go
```diff
@@ -51,154 +72,225 @@ func newHost(ctx context.Context, cfg *Config) (*host, error) {
...
+		log.Trace(
+			"bootstrap",
+			"host", h.id(),
+			"peer", peerInfo.ID,
+		)
+		// connect to each peer
+		err := h.connect(peerInfo)
 		if err != nil {
-			log.Debug("failed to dial bootstrap peer", "err", err)
+			log.Error("connect", "error", err)
```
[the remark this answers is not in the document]
ryanchristo: Hmm... I can revert but `err` will provide the error message for why `connect` failed. Also, will `connect` throw any other errors rather than just a failed dial error?

p2p/host.go
```diff
@@ -51,154 +72,225 @@ func newHost(ctx context.Context, cfg *Config) (*host, error) {
...
+		// create new mdns service
+		mdns, err := discovery.NewMdnsService(
+			h.ctx,
+			h.h,
+			mdnsPeriod,
+			string(h.protocolId),
+		)
 		if err != nil {
-			log.Error("error starting MDNS", "err", err)
+			log.Error("start mdns", "error", err)
```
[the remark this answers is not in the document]
ryanchristo: I can revert just thought it might be helpful to be consistent. No need to include "error" as the first word for the message being passed to `log.Error`, which is labeled error in the logs.

p2p/p2p.go
```diff
@@ -51,134 +67,253 @@ func NewService(conf *Config, msgSend chan<- Message, msgRec <-chan Message) (*S
...
+		log.Debug(
+			"received message",
+			"host", s.host.id(),
+			"type", msg.GetType(),
+		)
+
+		// check if message should be broadcasted
+		if !s.shouldBroadcast(msg) {
+			log.Error(
+				"message ignored",
```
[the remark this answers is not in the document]
ryanchristo: good catch

go.mod
```diff
@@ -19,6 +19,7 @@ require (
 	github.com/libp2p/go-libp2p v0.4.1
 	github.com/libp2p/go-libp2p-core v0.2.5
 	github.com/libp2p/go-libp2p-kad-dht v0.3.0
+	github.com/libp2p/go-libp2p-peerstore v0.1.4
```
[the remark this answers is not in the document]
ryanchristo: fixed

p2p/p2p.go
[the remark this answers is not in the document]
ryanchristo: fixed

p2p/connmgr.go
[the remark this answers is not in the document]
ansermino: I don't think that's sufficient reason not to log them. You are hiding information this might be relevant to debugging.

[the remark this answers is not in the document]
ryanchristo: This is true. I think maybe just something I was using at the time. Fixed.

p2p/host.go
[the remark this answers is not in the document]
ryanchristo: Yea, still getting the hang of how to use logging. Fixed here and elsewhere.

p2p/host.go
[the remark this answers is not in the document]
ryanchristo: Fixed.

p2p/p2p.go
```diff
@@ -51,134 +66,248 @@ func NewService(conf *Config, msgSend chan<- Message, msgRec <-chan Message) (*S
...
+	// broadcast to all connected peers if gossip enabled
+	if !s.host.noGossip {
 
-// ID returns the host's ID
+		log.Trace(
+			"gossiping",
+			"host", stream.Conn().LocalPeer(),
+			"type", msg.GetType(),
+		)
+
```
ansermino: Also seems unnecessary, once gossiping is migrated this might have more relevance

p2p/p2p.go
```diff
@@ -198,50 +327,28 @@ func (s *Service) NoBootstrapping() bool {
...
 	}
 
-	// read entire message
-	rawMsg, err := rw.Reader.Peek(int(length))
+	// check message type
+	_, err = rw.Reader.Peek(1)
 	if err != nil {
-		log.Error("failed to read message", "err", err)
-		return nil, nil, err
+		return nil, err
```
ansermino: I don't think this is needed any more. `Peek` shouldn't change anything, iirc this was to check what type of message it was, which is no longer necessary.

p2p/p2p.go
```diff
@@ -51,134 +66,248 @@ func NewService(conf *Config, msgSend chan<- Message, msgRec <-chan Message) (*S
...
+
+	// check if message should be broadcasted
+	if !s.shouldBroadcast(msg) {
+		log.Trace(
+			"message ignored",
+			"host", s.host.id(),
+			"type", msg.GetType(),
+		)
+		return
+	}
```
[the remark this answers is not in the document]
ryanchristo: It was useful for debugging when writing the tests to know if a non-status message was sent and ignored before a status message was sent and confirmed. Would be nice to keep as `Trace` level logging in my opinion but I can remove.

p2p/p2p.go
ryanchristo: Once again, it was helpful at the time when writing tests but I can remove.

p2p/p2p.go
ryanchristo: Removed.

PR #433, opened 2019-11-27
crypto, runtime: ext_ public keys, sign

### LINKED ISSUE #279: ext_sr25519_public_keys

### LINKED ISSUE #281: ext_sr25519_sign

noot:
## Changes
- implement ext_ed25519_public_keys, ext_sr25519_public_keys, ext_ed25519_sign, ext_sr25519_sign
- implement functions to get public keys/keypairs from keystore
- add constants to crypto package for pubkey length, privkey length, seed length, signature length

## Tests:
```
go test ./crypto/ ./runtime/
```

### Issues:
closes #281 closes #279

### DIFF
--- a/crypto/ed25519.go
+++ b/crypto/ed25519.go
@@ -9,6 +9,11 @@ import (
 	"github.com/ChainSafe/gossamer/common"
 )
 
+const Ed25519PublicKeyLength int = 32
+const Ed25519SeedLength int = 32
+const Ed25519PrivateKeyLength int = 64
+const Ed25519SignatureLength int = 64
+
 // Ed25519Keypair is a ed25519 public-private keypair
 type Ed25519Keypair struct {
 	public  *Ed25519PublicKey
@@ -29,7 +34,7 @@ func NewEd25519Keypair(priv ed25519.PrivateKey) *Ed25519Keypair {
 }
 
 func NewEd25519KeypairFromSeed(seed []byte) (*Ed25519Keypair, error) {
-	if len(seed) != SeedLength {
+	if len(seed) != Ed25519SeedLength {
 		return nil, fmt.Errorf("cannot generate key from seed: seed is not 32 bytes long")
 	}
 	edpriv := ed25519.NewKeyFromSeed(seed)
@@ -38,7 +43,7 @@ func NewEd25519KeypairFromSeed(seed []byte) (*Ed25519Keypair, error) {
 
 // GenerateEd25519Keypair returns a new ed25519 keypair
 func GenerateEd25519Keypair() (*Ed25519Keypair, error) {
-	buf := make([]byte, SeedLength)
+	buf := make([]byte, Sr25519SeedLength)
 	_, err := rand.Read(buf)
 	if err != nil {
 		return nil, err
@@ -52,7 +57,7 @@ func GenerateEd25519Keypair() (*Ed25519Keypair, error) {
 // NewEd25519PublicKey returns an ed25519 public key that consists of the input bytes
 // Input length must be 32 bytes
 func NewEd25519PublicKey(in []byte) (*Ed25519PublicKey, error) {
-	if len(in) != PublicKeyLength {
+	if len(in) != Ed25519PublicKeyLength {
 		return nil, fmt.Errorf("cannot create public key: input is not 32 bytes")
 	}
 
@@ -73,7 +78,7 @@ func NewEd25519PrivateKey(in []byte) (*Ed25519PrivateKey, error) {
 
 // Verify returns true if the signature is valid for the given message and public key, false otherwise
 func Ed25519Verify(pub *Ed25519PublicKey, msg, sig []byte) bool {
-	if len(sig) != SignatureLength {
+	if len(sig) != Ed25519SignatureLength {
 		return false
 	}
 
@@ -124,7 +129,7 @@ func (k *Ed25519PrivateKey) Decode(in []byte) error {
 
 // Verify checks that Ed25519PublicKey was used to create the signature for the message
 func (k *Ed25519PublicKey) Verify(msg, sig []byte) bool {
-	if len(sig) != SignatureLength {
+	if len(sig) != Ed25519SignatureLength {
 		return false
 	}
 	return ed25519.Verify(ed25519.PublicKey(*k), msg, sig)
--- a/crypto/ed25519.go
+++ b/crypto/ed25519.go
@@ -43,7 +43,7 @@ func NewEd25519KeypairFromSeed(seed []byte) (*Ed25519Keypair, error) {
 
 // GenerateEd25519Keypair returns a new ed25519 keypair
 func GenerateEd25519Keypair() (*Ed25519Keypair, error) {
-	buf := make([]byte, Sr25519SeedLength)
+	buf := make([]byte, Ed25519SeedLength)
 	_, err := rand.Read(buf)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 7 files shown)


crypto/keypair.go
```diff
@@ -13,6 +13,12 @@ type KeyType = string
 const Ed25519Type KeyType = "ed25519"
 const Sr25519Type KeyType = "sr25519"
 
+const PublicKeyLength int = 32
+const SeedLength int = 32
+const Ed25519PrivateKeyLength int = 64
+const Sr25519PrivateKeyLength int = 32
+const SignatureLength int = 64
+
```
ansermino: These don't seem to have anything to do with the abstract keypair (ie. these don't seem to be enforced as part of the interface). It might be preferential to move these in the implementations, providing specific values for each type (eg. `Ed25519PublickeyLength`).

ansermino: You could also move them to subpackages to reduce naming conflicts

runtime/imports.go
```diff
@@ -485,28 +485,132 @@ func ext_sr25519_generate(context unsafe.Pointer, idData, seed, seedLen, out int
...
+	runtimeCtx := registry[*(*int)(instanceContext.Data())]
+	mutex.RUnlock()
+
+	keys := runtimeCtx.keystore.Ed25519Keypairs()
+
+	pubkey := memory[pubkeyData : pubkeyData+32]
+	var signingKey crypto.Keypair
+	for _, key := range keys {
+		if bytes.Equal(key.Public().Encode(), pubkey) {
+			signingKey = key
```
ansermino: `break`, also below

ansermino: this might also be better as a function of keystore (ie. `GetKeypair(key PublicKey))`

runtime/wasmer.go
```diff
@@ -127,7 +127,9 @@ func (r *Runtime) StorageRoot() (common.Hash, error) {
 }
 
 func (r *Runtime) Store(data []byte, location int32) {
+	r.mutex.Lock()
 	mem := r.vm.Memory.Data()
+	r.mutex.Unlock()
```
ansermino: This should come after the copy. We should also update `Load`.

Please also make sure to use these functions where possible.

runtime/imports.go
noot: good move, I will update that

runtime/wasmer.go
noot: when I moved it to after the copy, it causes a deadlock in one of the tests :/ I think we should remove this for now, and I'll open an issue to look into it

crypto/keypair.go
noot: fair, I'll move them to the respective files

crypto/ed25519.go
```diff
@@ -38,7 +43,7 @@ func NewEd25519KeypairFromSeed(seed []byte) (*Ed25519Keypair, error) {
 
 // GenerateEd25519Keypair returns a new ed25519 keypair
 func GenerateEd25519Keypair() (*Ed25519Keypair, error) {
-	buf := make([]byte, 32)
+	buf := make([]byte, Sr25519SeedLength)
```
ryanchristo: This should be `Ed25519SeedLength`.

runtime/imports.go
```diff
@@ -485,28 +485,129 @@ func ext_sr25519_generate(context unsafe.Pointer, idData, seed, seedLen, out int
...
-	return 0
+	instanceContext := wasm.IntoInstanceContext(context)
+	memory := instanceContext.Memory().Data()
+
+	mutex.RLock()
+	runtimeCtx := registry[*(*int)(instanceContext.Data())]
+	mutex.RUnlock()
+
+	keys := runtimeCtx.keystore.Ed25519PublicKeys()
+	// TODO: when do deallocate?
```
ryanchristo: Can you explain?

runtime/imports.go
```diff
@@ -485,28 +485,129 @@ func ext_sr25519_generate(context unsafe.Pointer, idData, seed, seedLen, out int
...
+	offset, err := runtimeCtx.allocator.Allocate(uint32(len(keys) * 32))
+	if err != nil {
+		log.Error("[ext_sr25519_public_keys]", "error", err)
```
ryanchristo: This should be `ext_ed25519_public_keys`.

runtime/wasmer.go
ansermino: Yeah you might want to checkout how the lock is actually used...

runtime/imports.go
ansermino: Please open issue or resolve

noot: we malloc some space in the memory so we can write the keys, but the memory never gets freed which is bad and could cause us to run out of memory. it needs to be freed at some point, but I'm not sure where to add that in. it should be freed once the runtime call is done, but there's no function that is called when the call finished

runtime/imports.go
noot: fixed

crypto/ed25519.go
noot: fixed

runtime/imports.go
noot: I will open an issue

ansermino:
Please move the key interfaces into `keystore`. Since `keystore` is the one operating on the interfaces this makes more sense idiomatically, it will also make it easier to re-use in other applications and remove the need to import crypto always.

noot:
@ansermino I can't do that since it causes a circular dependency (keystore depends on crypto, but then crypto will depend on keystore for the keypair interface)

runtime/wasmer.go
noot: okay I don't think I should change this right now, because it would mean moving the locks in Exec, or rewriting Exec not to use Store. Exec should be an atomic operation (which it is right now), so I don't think it should be changed

keystore/keystore.go
```diff
@@ -32,3 +33,52 @@ func (ks *Keystore) Get(pub common.Address) crypto.Keypair {
...
+	for _, key := range ks.keys {
+		if _, ok := key.(*crypto.Sr25519Keypair); ok {
+			srkeys = append(srkeys, key.Public())
+		}
+	}
+	return srkeys
+}
+
+func (ks *Keystore) Sr25519Keypairs() []crypto.Keypair {
+	edkeys := []crypto.Keypair{}
```
ryanchristo: This should be `srkeys`

ansermino:
> @ansermino I can't do that since it causes a circular dependency (keystore depends on crypto, but then crypto will depend on keystore for the keypair interface)

That shouldn't be the case. Keystore will depend on crypto, but crypto shouldn't depend on keystore.

noot:
@ansermino it doesn't right now, but if I moved the interfaces to keystore, then it would, since the interfaces are things like:
```
type PrivateKey interface {
	Public() PublicKey
}
```

If I try to change the key functions to return their specific key types:
```
func (k *Ed25519PrivateKey) Public() (*Ed25519PublicKey, error) {
     //stuff
}
```

 I get a bunch of errors like:
```
crypto/keypair.go:39:13: cannot assign *Ed25519PrivateKey to priv (type PrivateKey) in multiple assignment:
	*Ed25519PrivateKey does not implement PrivateKey (wrong type for Public method)
		have Public() (*Ed25519PublicKey, error)
		want Public() (PublicKey, error)
```

I would rather just move the specific key implementations to their own subpackages and deal with this interface stuff in a separate PR, would be nice to get this merged

Issue #400, opened 2019-11-12
Refactor State


ansermino:
Once finalized, we will be consolidating the state components of the system into a single service.

This new service will be called `StateService`. Inside this service will be: blocktree, trie, respective DBs, message pool, and peer state.

This service will expose an API. This should be a separate structure (ie. its own type) as we don't want to pass the full service to all dependants. This API has been drafted in a separate document but is not quite complete, stay tuned.

ansermino:
Reads and writes will be concurrent, use protection!

ansermino:
I've started this here, not sure how far I'll get, anyone is welcome to pick it up: https://github.com/ChainSafe/gossamer/tree/david/state-refactor

PR #440, opened 2019-11-29
state: State refactor

dziabko:
## Changes
- Refactored state services to a few services
- Implemented block, network, and state services

## Tests:
```

```

### Issues:
#400


state/block.go
```diff
@@ -0,0 +1,97 @@
...
+func (bs *blockState) GetHeader(hash common.Hash) types.BlockHeader {
+	var result types.BlockHeader
+
+	data, err := bs.bt.Db.Db.Get(headerKey(hash))
+	check(err)
+
+	err = json.Unmarshal(data, &result)
+	check(err)
+
+	return result
```
ansermino: This isn't a great approach. What happens if we lookup something that doesn't exist? It would be okay to change this to return an error.

noot:
@dziabko please fix CI errors
```
# github.com/ChainSafe/gossamer/state
state/block.go:49:51: undefined: types.BlockHeader
state/block.go:50:13: undefined: types.BlockHeader
state/block.go:73:40: undefined: types.BlockHeader
state/block.go:75:9: undefined: types.BlockHeader
state/block.go:91:40: undefined: types.BlockHeader
state/interface.go:30:30: undefined: types.BlockHeader
state/interface.go:32:19: undefined: types.BlockHeader
state/interface.go:39:19: undefined: types.BlockHeader
state/interface.go:40:40: undefined: types.BlockHeader
```

state/block.go
```diff
@@ -0,0 +1,98 @@
...
+	return types.Block{Header: header, Body: *blockBody}, nil
+}
+
+func (bs *blockState) GetBlockByNumber(n *big.Int) types.Block {
+	// Can't do yet
+	return types.Block{}
+}
+
+func (bs *blockState) SetHeader(header types.BlockHeaderWithHash) {
+	rawdb.SetHeader(bs.db.Db, &header)
```
ansermino: Please don't use rawdb, rather copy the function here and we will remove rawdb later

state/block.go
```diff
@@ -0,0 +1,113 @@
...
+
+// blockDataKey = blockDataPrefix + hash
+func blockDataKey(hash common.Hash) []byte {
+	return append(blockDataPrefix, hash.ToBytes()...)
+}
+
+func (bs *blockState) GetHeader(hash common.Hash) (types.BlockHeaderWithHash, error) {
+	var result types.BlockHeaderWithHash
+
+	data, err := bs.bt.Db.Db.Get(headerKey(hash))
```
ryanchristo: Why we are using `bs.bt.Db.Db` for `Get` but `bs.db.Db` for `Put`? If this is how it should be, maybe comments to explain would be helpful here and elsewhere.

state/block.go
```diff
@@ -0,0 +1,113 @@
...
+
+	err = json.Unmarshal(data, &result)
+
+	return result, err
+}
+
+func (bs *blockState) GetBlockData(hash common.Hash) (types.BlockData, error) {
+	var result types.BlockData
+
+	data, err := bs.bt.Db.Db.Get(blockDataKey(hash))
```
ryanchristo: See comment above.

state/block.go
```diff
@@ -0,0 +1,113 @@
...
+func (bs *blockState) SetHeader(header types.BlockHeaderWithHash) error {
+	hash := header.Hash
+
+	// Write the encoded header
+	bh, err := json.Marshal(header)
+	if err != nil {
+		return err
+	}
+
+	err = bs.db.Db.Put(headerKey(hash), bh)
```
ryanchristo: See comment above.

state/block.go
```diff
@@ -0,0 +1,113 @@
...
+}
+
+func (bs *blockState) SetBlockData(hash common.Hash, blockData types.BlockData) error {
+	// Write the encoded header
+	bh, err := json.Marshal(blockData)
+	if err != nil {
+		return err
+	}
+
+	err = bs.db.Db.Put(blockDataKey(hash), bh)
```
ryanchristo: See comment above.

state/block.go
ansermino: Good catch! I think this should be `bs.db.Db.Get()`, below is correct.

PR #446, opened 2019-12-02
cmd, core, consensus/babe: add configs for core.Service, babe.Session, add keystore

noot:
## Changes
- cleanup core and babe a bit, have their constructors take a config struct instead of a bunch of params. will make it easier to update config options in the future, rather than having to update function args everywhere
- add keystore to core/babe
- replace babe key stubs with key types

## Tests:
```
go test ./...
```

### DIFF
--- a/cmd/gossamer/configcmd.go
+++ b/cmd/gossamer/configcmd.go
@@ -91,7 +91,14 @@ func makeNode(ctx *cli.Context) (*dot.Dot, *cfg.Config, error) {
 	srvcs = append(srvcs, p2pSrvc)
 
 	// core.Service
-	coreSrvc, err := core.NewService(r, msgSend, msgRec)
+	coreCfg := &core.ServiceConfig{
+		Keystore: ks,
+		Runtime:  r,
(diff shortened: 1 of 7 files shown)


noot:
is there any way to ignore the structcheck in the linter? the struct fields are actually used (babe/runtime.go config decoding uses them) it's just not direct

ansermino:
> is there any way to ignore the structcheck in the linter? the struct fields are actually used (babe/runtime.go config decoding uses them) it's just not direct

They will be ignored when decoding, thus are unused.

noot:
@ansermino  I thought structs within structs was ok for decoding :thinking_face: slightly confused

ansermino:
@noot unexpoted fields are always ignored. A package cannot access unexported fields of a struct defined in another package. As such, `codec` can never operate on unexported fields, encoding or decoding.

noot:
@ansermino true, I forgot about the fields inside AuthorityDataRaw. I have updated them to be exported, since they should be decoded from the config

PR #456, opened 2019-12-04
p2p: migrate gossip logic

### LINKED ISSUE #411: Migrate Gossip Logic
We should move gossip logic to its own sub-package. Currently, it mainly consists of a set of maps in `Service`. We ideally want a `Send` method exposed that accepts any message, this will be quite similar to `Service.Broadcast`. After this update, `Broadcast` should do only one thing: broadcast.

ryanchristo:
## Changes

This pull request migrates and refactors the gossip logic.
 
## Tests:
```
./p2p -v
```

### Issues:
closes #411

### DIFF
--- a/p2p/config.go
+++ b/p2p/config.go
@@ -52,6 +52,8 @@ type Config struct {
 	NoGossip bool
 	// Disables MDNS discovery
 	NoMdns bool
+	// Disables status exchange
+	NoStatus bool
 	// Global data directory
 	DataDir string
 	// Identity key for node
--- a/p2p/config.go
+++ b/p2p/config.go
@@ -52,8 +52,6 @@ type Config struct {
 	NoGossip bool
 	// Disables MDNS discovery
 	NoMdns bool
-	// Disables status exchange
-	NoStatus bool
 	// Global data directory
 	DataDir string
 	// Identity key for node
--- a/p2p/gossip.go
+++ b/p2p/gossip.go
@@ -23,14 +23,14 @@ import (
 
 // gossip submodule
 type gossip struct {
-	host         *host
+	host        *host
 	hasGossiped map[string]bool
 }
 
 // newGossip creates a new gossip instance from the host
 func newGossip(host *host) (g *gossip, err error) {
 	g = &gossip{
-		host:         host,
+		host:        host,
 		hasGossiped: make(map[string]bool),
 	}
 	return g, err
--- a/p2p/gossip.go
+++ b/p2p/gossip.go
@@ -21,24 +21,23 @@ import (
 	"github.com/libp2p/go-libp2p-core/network"
 )
 
-// Gossip submodule
-type Gossip struct {
+// gossip submodule
+type gossip struct {
 	host         *host
 	hasGossipped map[string]bool
 }
 
 // newGossip creates a new gossip instance from the host
-func newGossip(host *host) (gossip *Gossip, err error) {
-	gossip = &Gossip{
+func newGossip(host *host) (g *gossip, err error) {
+	g = &gossip{
 		host:         host,
 		hasGossipped: make(map[string]bool),
 	}
-
-	return gossip, err
+	return g, err
 }
 
 // handleMessage gossips messages that have not already been gossipped
-func (g *Gossip) handleMessage(stream network.Stream, msg Message) {
+func (g *gossip) handleMessage(stream network.Stream, msg Message) {
 
 	// check if message has been gossipped
 	if !g.hasGossipped[msg.Id()] {
@@ -53,7 +52,7 @@ func (g *Gossip) handleMessage(stream network.Stream, msg Message) {
 }
 
 // sendMessage broadcasts the message to connected peers
-func (g *Gossip) sendMessage(stream network.Stream, msg Message) {
+func (g *gossip) sendMessage(stream network.Stream, msg Message) {
 
 	// loop through connected peers
 	for _, peer := range g.host.peers() {
--- a/p2p/gossip.go
+++ b/p2p/gossip.go
@@ -18,58 +18,50 @@ package p2p
 
 import (
 	log "github.com/ChainSafe/log15"
+	"github.com/libp2p/go-libp2p-core/network"
 )
 
-// Gossip describes the gossip submodule
+// Gossip submodule
 type Gossip struct {
-	host        *host
-	msgReceived map[string]bool
+	host         *host
+	hasGossipped map[string]bool
 }
 
 // newGossip creates a new gossip instance from the host
 func newGossip(host *host) (gossip *Gossip, err error) {
 	gossip = &Gossip{
-		host:        host,
-		msgReceived: make(map[string]bool),
+		host:         host,
+		hasGossipped: make(map[string]bool),
 	}
 
 	return gossip, err
 }
 
-// gossip broadcasts the message to all connected peers if gossip is enabled,
-// message is valid message type, and message has not already been gossipped
-func (g *Gossip) handleMessage(msg Message) {
+// handleMessage gossips messages that have not already been gossipped
+func (g *Gossip) handleMessage(stream network.Stream, msg Message) {
 
-	// exit if gossip is disabled
-	if g.host.noGossip {
-		return // exit
-	}
+	// check if message has been gossipped
+	if !g.hasGossipped[msg.Id()] {
 
-	// check if valid message type and message has not already been gossipped
-	if g.shouldGossip(msg) {
+		// broadcast message to peers if message has not been gossipped
+		g.sendMessage(stream, msg)
 
-		// loop through connected peers
-		for _, peer := range g.host.peers() {
+		// update message to gossipped
+		g.hasGossipped[msg.Id()] = true
 
-			// send message to each connected peer
-			err := g.host.send(peer, msg)
-			if err != nil {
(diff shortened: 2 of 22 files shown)


p2p/gossip.go
```diff
@@ -0,0 +1,75 @@
...
+	// exit if gossip is disabled
+	if g.host.noGossip {
+		return // exit
+	}
+
+	// check if valid message type and message has not already been gossipped
+	if g.shouldGossip(msg) {
+
+		// loop through connected peers
+		for _, peer := range g.host.peers() {
```
ryanchristo: This could be replaced with `broadcast` but I gossip will likely have more functionality in the future so I kept it separated but it currently uses the same code as `broadcast`.

p2p/gossip.go
```diff
@@ -0,0 +1,75 @@
...
+package p2p
+
+import (
+	log "github.com/ChainSafe/log15"
+)
+
+// Gossip describes the gossip submodule
+type Gossip struct {
+	host        *host
+	msgReceived map[string]bool
```
ryanchristo: This does not migrate the message mappings for each message type. Should we?

p2p/gossip.go
```diff
@@ -0,0 +1,75 @@
...
+// shouldGossip checks if message has already been gossipped
+func (g *Gossip) shouldGossip(msg Message) bool {
+
+	// check if message stored in received message mapping
+	if g.msgReceived[msg.Id()] {
+		return false
+	}
+
+	// update message in received message mapping
+	g.msgReceived[msg.Id()] = true
```
ryanchristo: Is this where we want to set the message as received? I was not sure where to put it.

p2p/p2p.go
```diff
@@ -36,14 +36,11 @@ const SendStatusInterval = 5 * time.Minute
...
-	host         *host
-	msgRec       <-chan Message
-	msgSend      chan<- Message
-	blockAnnRec  map[string]bool
-	blockReqRec  map[string]bool
-	blockRespRec map[string]bool
-	txMessageRec map[string]bool
+	ctx     context.Context
+	host    *host
+	gossip  *Gossip
```
ryanchristo: Should `Gossip` be `gossip` or `host` be `Host` or is this right?

p2p/p2p.go
```diff
@@ -139,98 +142,66 @@ func (s *Service) sendStatusMessages(peer peer.ID) {
...
+func (s *Service) handleMessage(stream network.Stream, msg Message) {
+
+	// check if status exchange disabled or if peer status confirmed
+	if s.host.noStatus || s.host.peerStatus[stream.Conn().RemotePeer()] {
+
+		// send all non-status messages to core service
+		s.msgSend <- msg
+
+		// handle all non-status messages with gossip protocol
+		s.gossip.handleMessage(msg)
```
ryanchristo: We should probably pass the stream along to gossip as well. I will fix.

p2p/p2p_test.go
```diff
@@ -205,38 +224,15 @@ func TestExchangeStatus(t *testing.T) {
...
-			t.Error(
-				"node B received unexpected message from node A",
-				"\nexpected:", statusMessage,
-				"\nreceived:", msg,
-			)
-		}
-	case <-time.After(TestMessageTimeout):
-		t.Error("node B timeout waiting for message")
-	}
+	time.Sleep(TestStatusTimeout)
```
ryanchristo: This is not ideal but this is the only one... for now. :sweat_smile:

p2p/config.go
```diff
@@ -52,6 +52,8 @@ type Config struct {
 	NoGossip bool
 	// Disables MDNS discovery
 	NoMdns bool
+	// Disables status exchange
+	NoStatus bool
```
ansermino: I don't think this should be exported. It should be set to true inside tests but otherwise always false.

p2p/p2p.go
ansermino: should be `gossip` as we don't plan to export it

p2p/config.go
ryanchristo: I am not sure how to otherwise conveniently disable status messages from the tests but I will take another look.

p2p/p2p.go
ryanchristo: fixed

p2p/gossip.go
```diff
@@ -0,0 +1,66 @@
...
+
+import (
+	log "github.com/ChainSafe/log15"
+	"github.com/libp2p/go-libp2p-core/network"
+)
+
+// gossip submodule
+type gossip struct {
+	host         *host
+	hasGossipped map[string]bool
```
noot: being finicky here but i think gossiping has only 1 p? :p

ryanchristo: fixed

p2p/config.go
ryanchristo: fixed

p2p/gossip.go
```diff
@@ -0,0 +1,66 @@
...
+	// loop through connected peers
+	for _, peer := range g.host.peers() {
+
+		// send message to each connected peer
+		err := g.host.send(peer, msg)
+		if err != nil {
+			log.Error("Failed to send message during gossip", "err", err)
+		}
+	}
+}
```
ansermino: I think this can replaced by `host.Broadcast`

p2p/gossip.go
```diff
@@ -0,0 +1,66 @@
...
+// newGossip creates a new gossip instance from the host
+func newGossip(host *host) (g *gossip, err error) {
+	g = &gossip{
+		host:        host,
+		hasGossiped: make(map[string]bool),
+	}
+	return g, err
+}
+
+// handleMessage gossips messages that have not already been gossiped
```
ansermino: "handleMessage broadcasts messages that have not already been seen"?

p2p/p2p.go
```diff
@@ -139,98 +142,70 @@ func (s *Service) sendStatusMessages(peer peer.ID) {
 	}
 }
 
-// broadcastReceivedMessages starts polling the msgRec channel for messages
-// from the core service and broadcasts the new messages to connected peers
-func (s *Service) broadcastReceivedMessages() {
+// broadcastMessages broadcasts messages from the core service
+func (s *Service) broadcastMessages() {
```
ansermino: not a great name, I think "broadcast" should be dropped from the name, maybe "receiveCoreMessages"?

p2p/p2p.go
```diff
@@ -139,98 +142,70 @@ func (s *Service) sendStatusMessages(peer peer.ID) {
...
-	s.msgSend <- msg
 }
 
-// handleStreamStatus handles status messages written to the stream
-func (s *Service) handleStreamStatus(stream network.Stream, msg Message) {
+// handleMessage handles non-status messages written to the stream
+func (s *Service) handleMessage(stream network.Stream, msg Message) {
+
+	// check if status exchange is disabled or peer status is confirmed
+	if s.host.noStatus || s.host.peerStatus[stream.Conn().RemotePeer()] {
```
ansermino: I think this clause is incorrect

p2p/gossip.go
ryanchristo: Yup, it can. I was not sure whether I should keep it separate given gossip might evolve.

ryanchristo: Should we keep it separate it preparation? It's also nice to have the error include "gossip" so that it is more specific for debugging and the user.

p2p/gossip.go
ryanchristo: Not using "broadcast" and "seen" were also in line with keeping gossip separate as its own specific thing and "gossiped" because `hasGossiped`.

I can change `hasGossiped` to `hasBeenSeen` but we should probably also move where we set it to true since right now it is being set after the message has been gossiped.

p2p/p2p.go
ryanchristo: True. Not a great name. Haha. I like `receiveCoreMessages`. :+1:

p2p/p2p.go
ryanchristo: how so?

p2p/p2p.go
ryanchristo: fixed

p2p/gossip.go
ryanchristo: I think this was also a result of the reference to a send method in the issue.

ryanchristo: Ok, updated. Simplified to what the `gossip` submodule breaks down to at the moment. We can add when we know need to.

p2p/gossip.go
ryanchristo: Ok, updated and simplified.

p2p/p2p.go
ryanchristo: If status exchange is disabled (we do not care if we have confirmed the peer status) or the peer status has been confirmed (we care if we have confirmed the peer status and we are checking to make sure it has been confirmed) then we will send the message to the core service.

p2p/gossip.go
ansermino: I don't see why variable names need to match comments word for word. Broadcast is a word that concisely describes what is happening here in terms of propagation. Additionally, we are tracking the messages we have seen not the messages we have sent.

ryanchristo: Using broadcast here makes sense. I think that's just me being very literal with comments while still figuring out what exactly we need to what things we should keep separate.

PR #457, opened 2019-12-04
p2p: migrate mdns logic and update tests

### LINKED ISSUE #423: Investigate and resolve mdns dial errors

ryanchristo:
## Changes

This pull request adds `TestDiscovery` to `p2p_test.go` and updates each test to use `connect` as a standard if we are not testing bootstrap or discovery.

This pull request also resolves #423. The issue with failed dials was attempting to connect manually or via bootstrap and using mdns discovery. With mdns enabled, nodes will connect automatically.

## Tests:
```
go test ./p2p -v
```

### Issues:
closes #423


p2p/discovery.go
```diff
@@ -0,0 +1,117 @@
...
+const mdnsPeriod = time.Minute
+
+// See https://godoc.org/github.com/libp2p/go-libp2p/p2p/discovery#Notifee
+type Notifee struct {
+	ctx  context.Context
+	host libp2phost.Host
+}
+
+// discovery submodule
+type disc struct {
```
ansermino: pls name `discovery`

p2p/discovery.go
```diff
@@ -0,0 +1,117 @@
...
+		// close mDNS discovery service
+		err := d.mdns.Close()
+		if err != nil {
+			return err
+		}
+
+	}
+
+	return nil
+}
```
ansermino: This seems unnecessary. Either use `d.mdns.Close()` instead or rename this `close` and add the same logic for any future components.

PR #470, opened 2019-12-06
crypto, keypair: move keys to subpackages

noot:
## Changes
- moved ed25510 and sr25519 key implementations to their own subpackages
- fixed keystore and crypto packages, are more decoupled now

## Tests:
```
go test ./crypto/... ./keystore/
```


cmd/gossamer/account.go
```diff
@@ -164,13 +166,13 @@ func generateKeypair(keytype, datadir string, password []byte) (string, error) {
 	var err error
 	if keytype == crypto.Sr25519Type {
 		// generate sr25519 keys
-		kp, err = crypto.GenerateSr25519Keypair()
+		kp, err = sr25519.GenerateKeypair()
```
ansermino: We should probably update `generateKeypair` to take ` crypto.KeyType`

runtime/wasmer_test.go
```diff
@@ -31,9 +31,11 @@ import (
 
 	"github.com/ChainSafe/gossamer/common"
 	"github.com/ChainSafe/gossamer/crypto"
+	"github.com/ChainSafe/gossamer/crypto/ed25519"
+	"github.com/ChainSafe/gossamer/crypto/sr25519"
 	"github.com/ChainSafe/gossamer/keystore"
 	"github.com/ChainSafe/gossamer/trie"
-	"golang.org/x/crypto/ed25519"
+	//"golang.org/x/crypto/ed25519"
```
ansermino: remove

cmd/gossamer/account.go
noot: why? it's now part of the subpackage so it's specific to each key implementation

runtime/wasmer_test.go
noot: done

PR #472, opened 2019-12-07
crypto: implement secp256k1 keys

### LINKED ISSUE #468: Implement Secp256k1 Key Type
We should add Secp256k1 support as this will likely be useful in the future.

noot:
## Changes
- implement secp256k1 keys and signing
- was branched off my other PR #470  so really only the crypto/secp256k1 files are new

## Tests:
```
go test ./crypto/...
```

### Issues:
closes #468


crypto/secp256k1/secp256k1.go
```diff
@@ -0,0 +1,116 @@
...
+	pub, err := secp256k1.DecompressPubkey(in)
+	if err != nil {
+		return err
+	}
+	k.key = *pub
+	return nil
+}
+
+func (k *PublicKey) Address() common.Address {
+	return crypto.PublicKeyToAddress(k)
```
ansermino: Do we maybe want a separate method for this? Like `SS58Address()`?

ansermino:
Probably also need to add CLI support, or open an issue for that

ansermino:
Also -- regarding hashing messages for sigs: we should enfore the usage of common.Hash but remove all hashing from the crypto implementations. This allows us to choose which hash algo we want while enforcing sigs are cryptographically secure

crypto/secp256k1/secp256k1.go
```diff
@@ -0,0 +1,116 @@
...
+func (pk *PrivateKey) Sign(msg []byte) ([]byte, error) {
+	hash, err := common.Blake2bHash(msg)
+	if err != nil {
+		return nil, err
+	}
+	return secp256k1.Sign(hash[:], &pk.key)
+}
+
+func (pk *PrivateKey) Public() (crypto.PublicKey, error) {
+	return pk.Public()
```
ansermino: this is problematic, recursive call

ansermino: I think this works:
```go
func (pk *PrivateKey) Public() (crypto.PublicKey, error) {
	kp := NewKeypair(pk.key)
	return kp.Public(), nil
}
```

crypto/secp256k1/secp256k1.go
noot: you mean like `(k *PublicKey) SS58Address()`? or another crypto function

also what other address types would we have?

crypto/secp256k1/secp256k1.go
noot: woops I meant `pk.key.Public()`, fixed

runtime/wasmer_test.go
```diff
@@ -31,9 +31,11 @@ import (
 
 	"github.com/ChainSafe/gossamer/common"
 	"github.com/ChainSafe/gossamer/crypto"
+	"github.com/ChainSafe/gossamer/crypto/ed25519"
+	"github.com/ChainSafe/gossamer/crypto/sr25519"
 	"github.com/ChainSafe/gossamer/keystore"
 	"github.com/ChainSafe/gossamer/trie"
-	"golang.org/x/crypto/ed25519"
+	//"golang.org/x/crypto/ed25519"
```
ryanchristo: why keep it commented out?

noot: removed

runtime/imports.go
```diff
@@ -652,16 +653,16 @@ func ext_sr25519_verify(context unsafe.Pointer, msgData, msgLen, sigData, pubkey
...
 
-	pub, err := crypto.NewSr25519PublicKey(memory[pubkeyData : pubkeyData+32])
+	pub, err := sr25519.NewPublicKey(memory[pubkeyData : pubkeyData+32])
 	if err != nil {
 		return 1
 	}
 
-	if pub.Verify(msg, sig) {
-		return 0
+	if ok, err := pub.Verify(msg, sig); err != nil || !ok {
```
ryanchristo: I wonder if we can make this cleaner. Maybe update `Verify` to only return ok if no error.

runtime/imports.go
```diff
@@ -699,16 +700,16 @@ func ext_ed25519_verify(context unsafe.Pointer, msgData, msgLen, sigData, pubkey
...
 	sig := memory[sigData : sigData+64]
-	pubkey, err := crypto.NewEd25519PublicKey(memory[pubkeyData : pubkeyData+32])
+	pubkey, err := ed25519.NewPublicKey(memory[pubkeyData : pubkeyData+32])
 	if err != nil {
 		return 1
 	}
 
-	if crypto.Ed25519Verify(pubkey, msg, sig) {
-		return 0
+	if ok, err := pubkey.Verify(msg, sig); err != nil || !ok {
```
ryanchristo: Same as above.

keystore/encrypt.go
```diff
@@ -13,6 +13,8 @@ import (
 	"path/filepath"
 
 	"github.com/ChainSafe/gossamer/crypto"
+	"github.com/ChainSafe/gossamer/crypto/ed25519"
+	"github.com/ChainSafe/gossamer/crypto/sr25519"
```
ryanchristo: Should we include secp in keystore encrypt?

keystore/encrypt_test.go
```diff
@@ -8,7 +8,8 @@ import (
 	"reflect"
 	"testing"
 
-	"github.com/ChainSafe/gossamer/crypto"
+	"github.com/ChainSafe/gossamer/crypto/ed25519"
+	"github.com/ChainSafe/gossamer/crypto/sr25519"
```
ryanchristo: Should we include secp in keystore encrypt test?

keystore/keystore.go
```diff
@@ -6,6 +6,8 @@ import (
 
 	"github.com/ChainSafe/gossamer/common"
 	"github.com/ChainSafe/gossamer/crypto"
+	"github.com/ChainSafe/gossamer/crypto/ed25519"
+	"github.com/ChainSafe/gossamer/crypto/sr25519"
```
ryanchristo: Should we include secp in keystore?

keystore/keystore_test.go
```diff
@@ -4,13 +4,13 @@ import (
 	"reflect"
 	"testing"
 
-	"github.com/ChainSafe/gossamer/crypto"
+	"github.com/ChainSafe/gossamer/crypto/sr25519"
```
ryanchristo: Should we include ed and secp in keystore test?

runtime/imports.go
```diff
@@ -64,7 +64,8 @@ import (
 	"unsafe"
 
 	common "github.com/ChainSafe/gossamer/common"
-	"github.com/ChainSafe/gossamer/crypto"
+	"github.com/ChainSafe/gossamer/crypto/ed25519"
+	"github.com/ChainSafe/gossamer/crypto/sr25519"
```
ryanchristo: Should we include secp in runtime imports?

runtime/wasmer_test.go
```diff
@@ -31,9 +31,10 @@ import (
 
 	"github.com/ChainSafe/gossamer/common"
 	"github.com/ChainSafe/gossamer/crypto"
+	"github.com/ChainSafe/gossamer/crypto/ed25519"
+	"github.com/ChainSafe/gossamer/crypto/sr25519"
```
ryanchristo: Should we include secp in wasmer test?

runtime/imports.go
noot: I could change it to this:
```
	ok, err := pub.Verify(msg, sig)
	if err != nil || !ok {
		return 1
	}
```

Verify has to return 1 if the call fails for any reason, 0 otherwise as per spec, so not sure how much I can change this

keystore/encrypt.go
noot: added

keystore/encrypt_test.go
noot: done

keystore/keystore.go
noot: done

keystore/keystore_test.go
noot: done

runtime/imports.go
noot: we currently use the go-ethereum secp256k1 library in the imports, I am ok with leaving it for now

runtime/wasmer_test.go
noot: same as above

noot:
@ansermino I can't change the param types in verify and sign to common.Hash since that would mess up the keypair interface, but I can do a check to make sure the input is 32 bytes.

PR #476, opened 2019-12-09
state, cmd, polkadb, dot: Move DBs to State Service

### LINKED ISSUE #462: Move DBs to State Service
As an initial step in this migration we should get rid of `DbService` and move its logic into `StateService`. This should have minimal impact on the system, ie. it's expected everything will still function as it was.

edwardmack:
## Changes
- Move database logic into `state.Service`
- Change references from `polkadb.DbService` to `state.Service`
- Update tests to reflect changes
- Remove `polkadb.DbService`

## Tests:
```
go test ./...
```

### Issues:
closes #462


cmd/gossamer/configcmd.go
```diff
@@ -53,11 +54,8 @@ func makeNode(ctx *cli.Context) (*dot.Dot, *cfg.Config, error) {
 
 	var srvcs []services.Service
 
-	// DB: Create database dir and initialize stateDB and blockDB
-	dbSrv, err := polkadb.NewDbService(fig.Global.DataDir)
-	if err != nil {
-		return nil, nil, fmt.Errorf("cannot create db service: %s", err)
-	}
+	// DB: Create service, initialize stateDB and blockDB
+	dbSrv := state.NewService(fig.Global.DataDir)
```
noot: perhaps rename this to `stateSrv` for clarity?

cmd/gossamer/genesis.go
```diff
@@ -30,11 +31,8 @@ func loadGenesis(ctx *cli.Context) error {
 
 	log.Info("🕸\t Initializing node", "name", gen.Name, "id", gen.Id, "protocolID", gen.ProtocolId, "bootnodes", common.BytesToStringArray(gen.Bootnodes))
 
-	// DB: Create database dir and initialize stateDB and blockDB
-	dbSrv, err := polkadb.NewDbService(fig.Global.DataDir)
-	if err != nil {
-		return err
-	}
+	// DB: Create service, initialize stateDB and blockDB
+	dbSrv := state.NewService(fig.Global.DataDir)
```
noot: same as above

cmd/gossamer/configcmd.go
edwardmack: good idea

PR #478, opened 2019-12-10
state: getBlockByNumber

### LINKED ISSUE #460: GetBlockByNumber
Current;ly no way to lookup block by number. Until we have an immediate need we should leave this open.

dziabko:
Added mapping of blockNumber to blockHash. Implemented getBlockByNumber using the mapping & getBlockByHash.

## Changes
- Added mapping of block number to block hash in the DB
- Used mapping & getBlockByHash to implement getBlockByNumber

## Tests:
```

```

### Issues:
closes #460


state/block.go
```diff
@@ -24,15 +25,28 @@ func NewBlockState() *blockState {
...
-	headerPrefix    = []byte("hdr") // headerPrefix + hash -> header
-	blockDataPrefix = []byte("hsh") // blockDataPrefix + hash -> blockData
+	headerPrefix     = []byte("hdr") // headerPrefix + hash -> header
+	blockDataPrefix  = []byte("bld") // blockDataPrefix + hash -> blockData
+	headerHashPrefix = []byte("hsh")
 )
 
+// encodeBlockNumber encodes a block number as big endian uint64
+func encodeBlockNumber(number uint64) []byte {
+	enc := make([]byte, 8)
```
ryanchristo: Maybe a comment about why we are using `8` here.

state/block.go
```diff
@@ -24,15 +25,28 @@ func NewBlockState() *blockState {
 
 var (
 	// Data prefixes
-	headerPrefix    = []byte("hdr") // headerPrefix + hash -> header
-	blockDataPrefix = []byte("hsh") // blockDataPrefix + hash -> blockData
+	headerPrefix     = []byte("hdr") // headerPrefix + hash -> header
+	blockDataPrefix  = []byte("bld") // blockDataPrefix + hash -> blockData
+	headerHashPrefix = []byte("hsh")
```
ryanchristo: Maybe a comment for `headerHashPrefix`.

state/block.go
noot: it's because in the line after we are doing PutUint64, which takes 8 bytes to put the uint64 into

ryanchristo:
Let's add a test before merging.

ryanchristo:
cc @dziabko can you please add tests for the following after updating the branch and then include a test for each state method added in future pull requests.

- `AddBlock`
- `GetBlockByHash`
- `GetBlockByNumber`

state/block_test.go
```diff
@@ -0,0 +1,75 @@
...
+	"os"
+	"reflect"
+	"testing"
+
+	"github.com/ChainSafe/gossamer/common"
+	"github.com/ChainSafe/gossamer/core/types"
+)
+
+func TestGetBlockByNumber(t *testing.T) {
+	dataDir := "../test_data/block"
```
ansermino: Please clean up any test artifacts. Check out the top `cmd/gossamer/configcmd_test.go` for some copy pasta

state/block.go
```diff
@@ -12,8 +12,8 @@ import (
 )
 
 type blockState struct {
-	bt          *blocktree.BlockTree
-	db          *polkadb.BlockDB
+	Bt          *blocktree.BlockTree
+	Db          *polkadb.BlockDB
```
ansermino: Do these need to be exported?

state/block_test.go
```diff
@@ -0,0 +1,75 @@
...
+		t.Fatal(err)
+	}
+	//Close DB & erase data dir contents
+	defer func() {
+		err = blockState.Db.Db.Close()
+		if err != nil {
+			t.Fatal("BlockDB close err: ", err)
+		}
+		if err = os.RemoveAll(dataDir); err != nil {
+			fmt.Println("removal of temp directory test_data failed")
```
ryanchristo: should we `t.Fatal(err)` here rather than print?

state/block_test.go
```diff
@@ -0,0 +1,75 @@
...
+		}
+	}()
+
+	// Create a header & blockdata
+	blockHash := common.NewHash([]byte{0, 1, 2})
+	header := types.BlockHeaderWithHash{
```
ryanchristo: should we name `blockHeader` to be consistent?

state/block_test.go
```diff
@@ -0,0 +1,75 @@
...
+	dataDir := "../test_data/block"
+
+	//Create a new blockState
+	blockState, err := NewBlockState(dataDir)
+	if err != nil {
+		t.Fatal(err)
+	}
+	//Close DB & erase data dir contents
+	defer func() {
+		err = blockState.Db.Db.Close()
```
ryanchristo: Should we create a `Stop` method on `blockState` that includes `Db.Db.Close` and `os.RemoveAll` to clean this up and also to make it reusable in multiple tests?

ryanchristo:
cc @dziabko What's the status of this pull request? Do you have any questions or need any help?

ryanchristo:
cc @dziabko what's the status of this pull request?

state/block_test.go
thomasmodeneis: this same test is in the #503 , I've added comments there with snippets for the fix, could this be applied there instead ?

state/block_test.go
thomasmodeneis: this same test is in the #503 , I've added comments there with snippets for the fix, could this be applied there instead ?

state/block_test.go
thomasmodeneis: this could be removed completely and use ioutils intead, see

state/block_test.go
```diff
@@ -0,0 +1,70 @@
...
+	dataDir := "../test_data/block"
+
+	// Create & start a new State service
+	stateService := NewService(dataDir)
+	stateService.Start()
+
+	// Close the service, and remove dataDir once test is done
+	defer stateService.Stop()
+	defer func() {
+		if err := os.RemoveAll("../test_data"); err != nil {
```
ryanchristo: Should we use the `dataDir` variable for this? ie, `os.RemoveAll(dataDir)`

state/block_test.go
ryanchristo: It would be nice to apply here but I'm ok with it being addressed in #503.

state/block.go
```diff
@@ -102,19 +115,28 @@ func (bs *blockState) GetLatestBlock() types.BlockHeader {
...
+func (bs *blockState) GetBlockByNumber(n *big.Int) (types.Block, error) {
+	// First retrieve the block hash based on the block number from the database
+	hash, err := bs.db.Db.Get(headerHashKey(n.Uint64()))
+
+	if err != nil {
+		return types.Block{}, err
+	}
+
+	// Then find the block based on the hash
+	endHash := common.NewHash(hash)
```
ryanchristo: Is `endHash` a typo?

ryanchristo: Maybe rename this `hash`, then rename the current `hash` to `data` or `bytes`.

state/block_test.go
ryanchristo: updated to `t.Fatal(err)`, let's address the latter in #503

state/block_test.go
ryanchristo: looks like this has been addressed

state/block_test.go
ryanchristo: this has not been addressed but let's add it in #503 when we add multiple tests

state/block_test.go
thomasmodeneis: done

state/block_test.go
thomasmodeneis: done

state/block.go
thomasmodeneis: done

thomasmodeneis:
PR comments accepted, code updated, ready to merge.

PR #479, opened 2019-12-10
cmd: simplify TestMakeNode

### LINKED ISSUE #360: Tests Modify Home Dir
Some tests are modifying the default datadir (`$HOME/.gossamer`). We should find some way to prevent this, so we don't use/modify existing data.

edwardmack:
## Changes
- Removed one instance of default node test in an effort to fix BadgerDB lock issue when testing on TravisCI. 

## Tests:
```
cd cmd/gossamer/
go test -v -run ^TestMakeNode$
```

### Issues:
closes #360


noot:
@edwardmack same as Ryan

edwardmack:
Yes, this should allow us to remove the `Skip` for the tests.

Issue #455, opened 2019-12-04
Providing bootnodes should disable default bootnodes


ansermino:
## Expected Behaviour

Providing bootnodes via `--bootnodes` should cause only the provided bootnodes to be used (should not include the defaults).

## Current Behaviour

Provided bootnodes and default bootnodes are combined.

## Possible Solution

Change this:
 https://github.com/ChainSafe/gossamer/blob/77da59b1409178e14474224ac5230d91d9663693/cmd/gossamer/configcmd.go#L196

PR #486, opened 2019-12-12
cmd: Providing bootnodes should disable default bootnodes

dziabko:
## Changes
- Overwrote p2pService bootnodes if there is a config flag passed 

## Tests:
```

```

### Issues:
#455


cmd/gossamer/configcmd.go
```diff
@@ -193,10 +193,17 @@ func setP2pConfig(ctx *cli.Context, fig *cfg.P2pCfg) {
 
 // createP2PService creates a p2p service from the command configuration and genesis data
 func createP2PService(fig *cfg.Config, gendata *genesis.GenesisData) (*p2p.Service, chan p2p.Message, chan p2p.Message) {
+	// Default bootnodes are from genesis.json
```
ansermino: too specific, please change to just "from genesis"

PR #487, opened 2019-12-13
api, p2p, rpc, state: network information

### LINKED ISSUE #399: Implement Peer State
The peer state is responsible for storing state information about peers. As peer events occur, the state should update itself.

The state will be queried by the RPC module, so we can use that to form some requirements. The required API is (see types [here](https://github.com/polkadot-js/api/blob/bed3f9f4c577fbe49ccfa67d5f86a82519e4cf3a/packages/types/src/interfaces/rpc/definitions.ts)):
```
networkState(): NetworkState {
    peerId // your peerId
}
peers(): []PeerInfo{
    PeerId,
    Roles,
    ProtocolVersion
    BestHash,
    BestNumber.
}
```

It seems that an event stream would work nicely for this. We can define a few events, such as `NewPeer`, `ExistingPeer`, `PeerDropped`, which contain the necessary information to update the state to reflect the actual network. The p2p layer (specifically the connection manager) can push these events into a channel which the pool polls.

ryanchristo:
## Changes

This pull request updates the public p2p service methods to align with information needed for the rpc server. This pull request also updates the network state in preparation for migrating the rpc server from using the api service to using the state service (see #488).

The information needed for the rpc server is up-to-date host and peer information that can be made available though public p2p service methods (`Health()`, `NetworkState()` and `Peers()`). Following the implementation of the status module, we do not need to store a separate peer state (see #399).

## Tests:
```
go test ./...
```

### Issues:
closes #399 and opens #488

### DIFF
--- a/rpc/modules/system.go
+++ b/rpc/modules/system.go
@@ -31,7 +31,7 @@ type SystemModule struct {
 // EmptyRequest represents an RPC request with no fields
 type EmptyRequest struct{}
 
-type NotYetImplemented string
+type StringResponse string
 
 type SystemHealthResponse struct {
 	Health p2p.Health `json:"health"`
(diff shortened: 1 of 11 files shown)


rpc/modules/system.go
```diff
@@ -63,52 +51,54 @@ type SystemPropertiesResponse struct {
 	TokenSymbol   string `json:"tokenSymbol"`
 }
 
-// NewSystemModule creates a new net API instance.
+// NewSystemModule creates a new API instance
 func NewSystemModule(api *api.Api) *SystemModule {
 	return &SystemModule{
-		api: api,
+		api: api, // TODO: migrate to network state
```
ryanchristo: addressed in #488

state/network.go
```diff
@@ -3,29 +3,27 @@ package state
...
 
 type networkState struct {
-	peer *p2p.Service
+	p2p *p2p.Service
 }
 
 func NewNetworkState() *networkState {
 	return &networkState{
-		peer: &p2p.Service{},
+		// TODO: pass p2p service instance to network state
```
ryanchristo: addressed in #488

rpc/modules/system.go
```diff
@@ -17,16 +17,15 @@
...
-	"github.com/ChainSafe/gossamer/common"
 	"github.com/ChainSafe/gossamer/internal/api"
+	"github.com/ChainSafe/gossamer/p2p"
 )
 
-// SystemModule is an RPC module providing access to core API points.
+// SystemModule is an RPC module providing access to core API points
 type SystemModule struct {
-	api *api.Api
+	api *api.Api // TODO: migrate to network state
```
ryanchristo: addressed in #488

state/network.go
ryanchristo: After a discussion with David, passing an instance of p2p service to network state might not be the route we want to take with network state. Updated issue #488.

PR #490, opened 2019-12-16
p2p: update configuration and host

ryanchristo:
## Changes

This pull request updates the p2p service configuration and updates host startup and methods.

## Tests:
```
go test ./p2p -v
```

### DIFF
--- a/p2p/host.go
+++ b/p2p/host.go
@@ -82,13 +82,13 @@ func newHost(ctx context.Context, cfg *Config) (*host, error) {
 	h = rhost.Wrap(h, dht)
 
 	// format bootnodes
-	bns, err := stringsToAddrInfos(cfg.BootstrapNodes)
+	bns, err := cfg.bootnodes()
 	if err != nil {
 		return nil, err
 	}
 
 	// format protocol id
-	pid := protocol.ID(cfg.ProtocolId)
+	pid := cfg.protocolId()
 
 	return &host{
 		ctx:        ctx,
--- a/p2p/host.go
+++ b/p2p/host.go
@@ -23,7 +23,7 @@ import (
 	"github.com/ChainSafe/gossamer/common"
 	log "github.com/ChainSafe/log15"
 	ds "github.com/ipfs/go-datastore"
-	dsync "github.com/ipfs/go-datastore/sync"
+	"github.com/ipfs/go-datastore/sync"
 	"github.com/libp2p/go-libp2p"
 	libp2phost "github.com/libp2p/go-libp2p-core/host"
 	"github.com/libp2p/go-libp2p-core/network"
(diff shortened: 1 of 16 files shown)


p2p/host.go
```diff
@@ -34,88 +34,85 @@ import (
...
-	dht := kaddht.NewDHT(ctx, h, dstore)
+	// create DHT service
+	dht := kaddht.NewDHT(ctx, h, sync.MutexWrap(ds.NewMapDatastore()))
 
-	// wrap host and DHT with routed host so that we can look up peers in DHT
+	// wrap host and DHT service with routed host
 	h = rhost.Wrap(h, dht)
 
-	// use "p2p" for multiaddress format
-	ma.SwapToP2pMultiaddrs()
```
ansermino: Don't we need this?

p2p/host.go
```diff
@@ -166,46 +152,36 @@ func (h *host) connect(addrInfo peer.AddrInfo) (err error) {
...
+// send sends a non-status message to a specific peer
+func (h *host) send(p peer.ID, msg Message) (err error) {
 
 	// create new stream with host protocol id
-	stream, err := h.h.NewStream(h.ctx, p, h.protocolId)
+	s, err := h.h.NewStream(h.ctx, p, h.protocolId)
 	if err != nil {
-		return nil, err
+		log.Debug("Failed to create new stream", "err", err)
+		return err
```
ansermino: Avoid this, you're logging and returning an error. Where ever this is called probably catches that error and logs it too

ansermino: This should also be `log.Error`

p2p/host.go
ryanchristo: yes, just moved it to the top of `newHost`

p2p/host.go
ryanchristo: ok, will fix

ryanchristo: Ok, so this was actually intentional. This and the other uses of `log.Debug` in the `send` method return an error because the error is logged with `log.Error` when `send` returns an error, so these `Debug` logs are used to debug `send` method errors.

For example, if we attempted to send a message in the status submodule and `send` failed on `NewStream`, then we would get an `Error` log that says "Failed to send status message to peer" and a `Debug` log that says "Failed to create new stream".

ansermino: I think that is an interesting idea that perhaps we should explore more. 

Is it necessary to include the error in the log at every level?

Also, my expectation is that the error should be sufficient to provide context. "Failed to create new stream" as an error propagated to `status.sendMessages` seems sufficient given you are in the context of `sendMessages`. Would better error message perhaps solve this issue? Are there any cases of excessive propagation

ryanchristo: Is it necessary? I think so.

Without verbose mode the user would only see the error log. With verbose mode and in testing we see the error log and the debug log. Maybe I am misunderstanding how we use logs?

I think the context is and will become more important for debugging. Knowing that a status message failed to be sent (and what peer it was intended to be sent to) versus a broadcast message failed to be sent (and what peer it was intended to be sent to) would be helpful.

I think one level is not quite excessive but if two levels down would be.

p2p/host.go
```diff
@@ -275,30 +244,27 @@ func (h *host) peerCount() int {
...
 		if err != nil {
 			continue
 		}
-		maddrs = append(maddrs, maddr)
-	}
-	return maddrs
-}
-
-// addrInfo returns the libp2p AddrInfo of the host
-func (h *host) addrInfo() (addrInfo *peer.AddrInfo, err error) {
```
noot: we don't need this anymore?

p2p/mdns.go
```diff
@@ -22,89 +22,88 @@ import (
 
 	log "github.com/ChainSafe/log15"
 	"github.com/libp2p/go-libp2p-core/peer"
-	libp2pdiscovery "github.com/libp2p/go-libp2p/p2p/discovery"
+	discovery "github.com/libp2p/go-libp2p/p2p/discovery"
 )
 
-const mdnsPeriod = time.Minute
+const MdnsPeriod = time.Minute
```
noot: not sure if this needs to be exported, not a huge deal tho

p2p/host.go
ryanchristo: We are getting `AddrInfo` for each address using `addrInfos()`. The only place the single address was being used was in the tests. which now uses `addrInfo[0]` (the first `AddrInfo`).

p2p/mdns.go
ryanchristo: Yea, I was not sure either.

p2p/host.go
ryanchristo: Will append information from lower level errors and pass up following discussion with David.

ryanchristo:
Will append information from lower level errors and pass up following discussion with David.

p2p/host.go
ryanchristo: Ok, removed debug logs for now. Error will be logged if `send` fails.

PR #491, opened 2019-12-16
runtime, state, core, consensus, cmd: Migrate runtime calls

### LINKED ISSUE #458: Migrate Runtime Calls
We need to move the majority of logic from the storage funcs in `runtime/imports.go` into our storage API. The storage API should be called from the imports in place of the existing logic.  

To support this we should instantiate a Trie inside the state service. For the time being lets leave the existing Trie alone.

By the end of this the runtime should depend on the storage API and not the Trie. We may beed the leave the existing Trie in there for now so we don't break tests (TBD).

edwardmack:
## Changes
- Refactor to update references to `trie.Trie` to `state.StorageState`

## Tests:
```
go test -v ./...
```

### Issues:
closes #458

### DIFF
--- a/runtime/interface.go
+++ b/runtime/interface.go
@@ -20,7 +20,7 @@ import (
 	"github.com/ChainSafe/gossamer/trie"
 )
 
-type RuntimeStorage interface {
+type Storage interface {
 	SetStorage(key []byte, value []byte) error
 	GetStorage(key []byte) ([]byte, error)
 	StorageRoot() (common.Hash, error)
--- a/runtime/interface.go
+++ b/runtime/interface.go
@@ -1,0 +1,32 @@
+// Copyright 2019 ChainSafe Systems (ON) Corp.
+// This file is part of gossamer.
+//
+// The gossamer library is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Lesser General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// The gossamer library is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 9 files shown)


state/storage.go
```diff
@@ -2,53 +2,92 @@ package state
...
 
-func (s *storageState) StorageRoot() (common.Hash, error) {
+func (s *StorageState) StorageRoot() (common.Hash, error) {
 	return s.trie.Hash()
 }
 
-func (s *storageState) EnumeratedTrieRoot(values [][]byte) {
+func (s *StorageState) EnumeratedTrieRoot(values [][]byte) {
 	//TODO
 }
```
ansermino: please add a `panic("not implemented")` for this and `ClearPrefix`

runtime/wasmer.go
```diff
@@ -98,7 +98,7 @@ func (r *Runtime) Stop() {
 }
 
 func (r *Runtime) StorageRoot() (common.Hash, error) {
-	return r.trie.Hash()
+	return r.storage.StorageRoot()
 }
```
ansermino: This should be removed once `core` is refactored, please can you add a `TODO`

edwardmack:
Created interface `RuntimeStorage` in `runtime/interface.go` and updated code to use new interface, so runtime no longer relies on `state` package.

runtime/interface.go
```diff
@@ -0,0 +1,32 @@
...
+// You should have received a copy of the GNU Lesser General Public License
+// along with the gossamer library. If not, see <http://www.gnu.org/licenses/>.
+package runtime
+
+import (
+	"github.com/ChainSafe/gossamer/common"
+	"github.com/ChainSafe/gossamer/trie"
+)
+
+type RuntimeStorage interface {
```
ansermino: Please rename to `Storage`, cardinal sin to include package name :wink:

PR #492, opened 2019-12-17
main: Fixed retrieval of genesisflag from utils.

### LINKED ISSUE #485: Fix genesis flag
When running gossamer init with a --genesis flag, and pass a different genesis file, it seems to always default to genesis.json. Also when removing the genesis.json file from the project root directory, it seems to find genesis.json file on its own.

1. Run 'build/bin/gossamer init --genesis README.md' (can include any file besides with genesis flag)

## Your Environment
* Version used:
* Commit hash:
* Go Version:

dziabko:
## Changes
- Getting local genesis string flag instead of the global string 

## Tests:
```

```

### Issues:
closes #485


noot:
@ryanchristo yeah we want Name, if we look at the flag definitions in utils/flags.go "usage" is the usage description of the flag. we look up what the flag is set to by the flag's name

PR #493, opened 2019-12-18
consensus/babe, state: Babe header type

### LINKED ISSUE #395: Introduce Babe Header Type
There is data specific to Babe that should be stored in a separate, distinct header type (`BabeHeader`).

See spec definition 5.10. These should be included in the BlockDB where the key for an entry is (epoch || slot) , since the Babe header is for a slot, not for a block.

Additionally, when we run the lottery for BABE, we will be able to create the headers at that time for the slots where we are validators. Thus, when we are running the lottery, we should be saving those headers (and possibly initializing the other headers)

edwardmack:
Note, this is currently a draft because:
- I'd like some comments regarding how this is being stored.  Are the babe header keys being generated correctly?
- Does BabeHeader struct look correct?
- Any other comments?

## Changes
- Create struct `BabeHeader` to store information an defined in definition 5.10 in PRE spec
- Added code in `block.go` to store BabeHeaders

## Tests:
```
go test -v ./...
```

### Issues:
closes #395


consensus/babe/types.go
```diff
@@ -42,3 +42,12 @@ type AuthorityData struct {
 	id     *sr25519.PublicKey
 	weight uint64
 }
+
+// BabeHeader as defined in Polkadot RE Spec, definition 4.10 in section 4.1.4
```
ryanchristo: definition 5.10 in section 5.1.4

state/block.go
```diff
@@ -43,6 +47,17 @@ func blockDataKey(hash common.Hash) []byte {
...
 }
 
+// babeHeaderKey = babeHeaderPrefix || epoch || slice
+func babeHeaderKey(epoch uint64, slice uint64) []byte {
+	epochBytes := make([]byte, 8)
+	binary.LittleEndian.PutUint64(epochBytes, epoch)
+	sliceBytes := make([]byte, 8)
+	binary.LittleEndian.PutUint64(sliceBytes, slice)
+	combined := append(epochBytes, sliceBytes...)
+	binary.LittleEndian.PutUint64(sliceBytes, slice)
```
ryanchristo: should this be in there twice?

ryanchristo:
This is looking good to me so far. Might want @ansermino to look at the `BabeHeader` struct.

consensus/babe/types.go
```diff
@@ -42,3 +42,12 @@ type AuthorityData struct {
...
+type BabeHeader struct {
+	Slot          uint64
+	VRFOutput     [32]byte
+	VRFProof      [32]byte
+	BlockProducer *sr25519.PublicKey
+	Signature     []byte
```
noot: I don't think there is a signature field?

consensus/babe/types.go
```diff
@@ -42,3 +42,12 @@ type AuthorityData struct {
 	id     *sr25519.PublicKey
 	weight uint64
 }
+
+// BabeHeader as defined in Polkadot RE Spec, definition 4.10 in section 4.1.4
+type BabeHeader struct {
+	Slot          uint64
+	VRFOutput     [32]byte
+	VRFProof      [32]byte
+	BlockProducer *sr25519.PublicKey
```
noot: the spec says this is the validator index, not the public key. practically I'm not sure it matters so much, will think about it

consensus/babe/types.go
noot: nvm, I see it now

consensus/babe/types.go
noot: actually, looks like it should be the index, so uint64 (probably), since it's included in the block digest

consensus/babe/types.go
edwardmack: It's definition 4.10 in 4.1.4, I had it wrong in the description.

consensus/babe/types.go
edwardmack: Isn't the symbol Sb for a signature of Block Header?

consensus/babe/types.go
edwardmack: I thought this was pk: is the SR25519 session public key associated with the block producer?

consensus/babe/types.go
noot: yes, but what I'm seeing in the spec is that the signature isn't part of the BABE header, it's part of the block digest, are you on the most recent version of the spec? it's definition 5.10 for me

edwardmack: Oh my, no I was looking at the wrong spec, sorry everyone!  Browser bookmarked to github.com/w3f/polkadot-re-spec which now I see is deprecated.  I'll look at the current on at github.com/w3f/polkadot-spec.

edwardmack:
Ok, I'm glad this was a draft...  I based my work on an old version of the spec :-( 
I've updated it should be better now.

consensus/babe/types.go
noot: no worries :)

ryanchristo:
@edwardmack is this pull request still a draft or ready for review?

state/block.go
```diff
@@ -43,6 +47,16 @@ func blockDataKey(hash common.Hash) []byte {
 	return append(blockDataPrefix, hash.ToBytes()...)
 }
 
+// babeHeaderKey = babeHeaderPrefix || epoch || slice
+func babeHeaderKey(epoch uint64, slice uint64) []byte {
```
ryanchristo: Should this be `slot` not `slice`?

state/block.go
```diff
@@ -116,6 +143,17 @@ func (bs *blockState) SetBlockData(hash common.Hash, blockData types.BlockData)
 	return err
 }
 
+func (bs *blockState) SetBabeHeader(epoch uint64, slice uint64, blockData babe.BabeHeader) error {
```
ryanchristo: same here

state/block.go
```diff
@@ -69,6 +83,19 @@ func (bs *blockState) GetBlockData(hash common.Hash) (types.BlockData, error) {
 	return result, err
 }
 
+func (bs *blockState) GetBabeHeader(epoch uint64, slice uint64) (babe.BabeHeader, error) {
```
ryanchristo: same here

Issue #484, opened 2019-12-12
Implement LEB128 variable-length encoding


ryanchristo:
## Expected Behavior

We have LEB123 variable-length encoding implemented. 

## Current Behavior

We have an unused `leb128ToUint64` method that decodes a byte array to uint64 using LEB128 variable-length encoding. We previously thought we needed this to send and receive messages with Alexander. We need to investigate whether we still need this and implement it if we do.

PR #495, opened 2019-12-18
p2p: send messages with existing stream

### LINKED ISSUE #437: Use existing stream to send messages
## Expected Behavior

We can use an existing stream to send messages if an existing stream available.

## Current Behavior

Nodes are not receiving send messages from other nodes on an existing stream. If allow nodes to use an existing stream, the status message exchange fails.

## Possible Solution

Maybe check whether a stream has been written to before writing to it. We might need to store some metadata associated with the stream including a timestamp and the host id.

Maybe review libp2p documentation to determine if there is built-in solution.

## Steps to Reproduce

Uncomment the `getExistingStream` method and its implementation in the `send` method.

```
go test ./p2p -v
```

ryanchristo:
## Changes

Use an existing stream to send messages if an existing stream is available.

See specifcation 4.4.2

> For the purposes of communicating Polkadot messages, the dailer of the connection opens a unique substream. Optionally, the node can keep a unique substream alive for this purpose.

## Tests:
```
go test ./p2p --run TestExistingStream -v
```

### Issues:
closes #437

### DIFF
--- a/p2p/host.go
+++ b/p2p/host.go
@@ -160,8 +160,7 @@ func (h *host) ping(peer peer.ID) error {
 }
 
 // send writes the given message to the outbound message stream for the given
-// peer (gets the already opened stream or opens a new stream). Each host uses
-// the same outbound message stream until otherwise closed or reset.
+// peer (gets the already opened outbound message stream or opens a new one).
 func (h *host) send(p peer.ID, msg Message) (err error) {
 
 	// get outbound stream for given peer
@@ -219,10 +218,9 @@ func (h *host) broadcast(msg Message) {
 	}
 }
 
-// getStream returns the outbound message stream for the given peer. Each p2p
-// service messsage is being written to the outbound message stream that uses
-// the same protocol id and that each host opens for each peer. Each host uses
-// the same outbound message stream until otherwise closed or reset.
+// getStream returns the outbound message stream for the given peer or returns
+// nil if no outbound message stream exists. For each peer, each host opens an
+// outbound message stream and writes to the same stream until closed or reset.
 func (h *host) getStream(p peer.ID) (stream network.Stream) {
 	conns := h.h.Network().ConnsToPeer(p)
 
--- a/p2p/host.go
+++ b/p2p/host.go
@@ -188,11 +188,6 @@ func (h *host) send(p peer.ID, msg Message) (err error) {
 		return err
 	}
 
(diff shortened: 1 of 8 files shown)


p2p/utils.go
```diff
@@ -130,3 +134,88 @@ func saveKey(priv crypto.PrivKey, fp string) (err error) {
 	}
 	return f.Close()
 }
+
+// TODO: implement LEB128 variable-length encoding
```
ryanchristo: addressed in #484

p2p/host.go
```diff
@@ -160,32 +159,40 @@ func (h *host) ping(peer peer.ID) error {
...
-	_, err = s.Write(common.Uint16ToBytes(uint16(len(encMsg)))[0:1])
+	encMsg, err := msg.Encode()
 	if err != nil {
 		return err
 	}
 
+	// _, err = s.Write(common.Uint16ToBytes(uint16(len(encMsg)))[0:1])
+	// if err != nil {
+	// 	return err
+	// }
```
ansermino: Can this be removed?

p2p/service.go
```diff
@@ -175,18 +168,49 @@ func (s *Service) handleConn(conn network.Conn) {
...
-// handleStream parses the message written to the data stream and calls the
-// associated message handler (non-status or status) based on message type
+// handleStream starts reading from the inbound message stream (substream with
+// a matching protocol id that was opened by the connected peer) and continues
+// reading until the inbound message stream is either closed or reset.
 func (s *Service) handleStream(stream network.Stream) {
+	ctx := context.Background()
+
+	// read from inbound stream
+	go s.readStream(ctx, stream)
```
ansermino: are you sure these needs its own goroutine? My expectation would be that this is called from a goroutine (as there are many simultaneous connections potentially happening). Removing it doesn't break tests, lmk what you think

p2p/host.go
ryanchristo: I was going to remove it when taking a look at #484 but I can remove now.

p2p/service.go
ryanchristo: I am not sure. I think we can remove... Let me take another look.

p2p/host.go
ryanchristo: done

p2p/service.go
ryanchristo: Yea, seems to be fine without it. Removed.

PR #503, opened 2019-12-20
state: AddBlock

### LINKED ISSUE #466: Add ability to add blocks
We need to provide an `AddBlock` method in our block state. This should just simply place the block in the DB.

dziabko:
## Changes
- AddBlock() implemented in blockState
- Added TestAddBlock

## Tests:
```
go test -v ./state -run TestAddBlock
```

### Issues:
closes #466


state/block.go
```diff
@@ -116,12 +123,21 @@ func (bs *blockState) SetBlockData(hash common.Hash, blockData types.BlockData)
 	return err
 }
 
-func (bs *blockState) AddBlock(block types.BlockHeaderWithHash) error {
+func (bs *blockState) AddBlock(block types.Block) error {
```
ansermino: Please comment

block/KEYREGISTRY
```diff
@@ -0,0 +1 @@
+�"�O��{1W1�v�$�Hello Badger
```
ryanchristo: What's going on here?

ryanchristo:
cc @dziabko what's the status of this pull request?

state/block_test.go
```diff
@@ -0,0 +1,99 @@
...
+	"math/big"
+	"os"
+	"reflect"
+	"testing"
+
+	"github.com/ChainSafe/gossamer/core/types"
+)
+
+func TestAddBlock(t *testing.T) {
+	dataDir := "../test_data/block"
```
thomasmodeneis: Could we use ioutils tempdir instead and replace this part till line 31 including the defer statement, for this instead: 

```
	dataDir, err := ioutil.TempDir("", "block_test")
	if err != nil {
		t.Error("Failed to create temp folder for block_test", "err", err)
		return
	}

	//Create a new blockState
	blockState, err := NewBlockState(dataDir)
	if err != nil {
		t.Fatal(err)
	}
```

that would be a bit simpler, what do you think ?

state/block_test.go
```diff
@@ -0,0 +1,99 @@
...
+	blockBody1 := types.BlockBody{}
+	blockBody1 = append(blockBody1, []byte{0, 1, 2, 3, 4, 5, 6, 7, 8, 9}...)
+
+	block1 := types.Block{
+		Header: header1,
+		Body:   &blockBody1,
+	}
+
+	// Add the block1 to the DB
+	blockState.AddBlock(block1)
```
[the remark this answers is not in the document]
thomasmodeneis: done

state/block_test.go
```diff
@@ -0,0 +1,99 @@
...
+	blockBody0 := types.BlockBody{}
+	blockBody0 = append(blockBody0, []byte{0, 1, 2, 3, 4, 5, 6, 7, 8, 9}...)
+
+	block0 := types.Block{
+		Header: header0,
+		Body:   &blockBody0,
+	}
+
+	// Add the block0 to the DB
+	blockState.AddBlock(block0)
```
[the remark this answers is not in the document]
thomasmodeneis: done

state/block_test.go
thomasmodeneis: done

thomasmodeneis:
PR comments accepted, code updated, ready for new review.

state/block.go
```diff
@@ -194,13 +195,30 @@ func (bs *blockState) SetBabeHeader(epoch uint64, slot uint64, blockData babe.Ba
...
+// AddBlock will set the latestBlock in blockState DB
+func (bs *blockState) AddBlock(block types.Block) error {
+	blockHeader := *block.Header
+
 	// Set the latest block
-	if block.Number.Cmp(bs.latestBlock.Number) == 1 {
-		bs.latestBlock = block
+	if reflect.DeepEqual(bs.latestBlock, types.BlockHeader{}) {
+		bs.latestBlock = blockHeader
+	} else if blockHeader.Number != nil && blockHeader.Number.Cmp(bs.latestBlock.Number) == 1 { // If the new block has a number greater than current latestBlock
```
noot: if blockHeader.Number == nil we probably want to return an error, since that isn't a valid block

state/block_test.go
```diff
@@ -66,3 +68,118 @@ func TestGetBlockByNumber(t *testing.T) {
...
+	// Create header
+	header0 := &types.BlockHeader{
+		Number: big.NewInt(0),
+	}
+	// Create blockHash
+	blockHash0 := header0.Hash()
+
+	// BlockBody with fake extrinsics
+	blockBody0 := types.BlockBody{}
+	blockBody0 = append(blockBody0, []byte{0, 1, 2, 3, 4, 5, 6, 7, 8, 9}...)
```
noot: this can probably be simplified into one line: `blockBody0 := types.BlockBody{0, 1, 2, 3, 4, 5, 6, 7, 8, 9}`

state/block.go
```diff
@@ -194,13 +195,30 @@ func (bs *blockState) SetBabeHeader(epoch uint64, slot uint64, blockData babe.Ba
...
-// AddBlock will add a block into the DB
-func (bs *blockState) AddBlock(block types.BlockHeader) error {
+// AddBlock will set the latestBlock in blockState DB
+func (bs *blockState) AddBlock(block types.Block) error {
+	blockHeader := *block.Header
+
 	// Set the latest block
-	if block.Number.Cmp(bs.latestBlock.Number) == 1 {
-		bs.latestBlock = block
+	if reflect.DeepEqual(bs.latestBlock, types.BlockHeader{}) {
```
noot: maybe we can change blockState.latestBlock into a pointer? comparing to nil seems cleaner

state/block_test.go
```diff
@@ -66,3 +68,118 @@ func TestGetBlockByNumber(t *testing.T) {
...
+
+	// Create header & blockData for block 1
+	header1 := &types.BlockHeader{
+		Number: big.NewInt(1),
+	}
+	blockHash1 := header1.Hash()
+
+	// Create Block with fake extrinsics
+	blockBody1 := types.BlockBody{}
+	blockBody1 = append(blockBody1, []byte{0, 1, 2, 3, 4, 5, 6, 7, 8, 9}...)
```
noot: same here

state/block_test.go
thomasmodeneis: yep, good point

state/block_test.go
thomasmodeneis: sure

state/block.go
thomasmodeneis: It will be cleaner, but returning a pointer may cause side effects later on when we modify the header for consensus changes.

thomasmodeneis: I'll make some changes here, your comment also made me see the latestBlock is actually not a block, but a header, and its confusing so I will update it a bit.

state/block.go
thomasmodeneis: yep

state/block.go
```diff
@@ -112,9 +111,9 @@ func (bs *blockState) GetBabeHeader(epoch uint64, slot uint64) (babe.BabeHeader,
 	return result, err
 }
 
-// GetLatestBlock returns the latest block available on blockState
-func (bs *blockState) GetLatestBlock() types.BlockHeader {
-	return bs.latestBlock
+// Header returns the latest block available on blockState
+func (bs *blockState) Header() *types.Header {
```
ansermino: I think this should be renamed LatestHeader() to be more descriptive

core/types/types.go
```diff
@@ -73,10 +72,27 @@ func NewBlockHeader(parentHash common.Hash, number *big.Int, stateRoot common.Ha
...
+	}
+	//copy digest byte array
+	if len(header.Digest) > 0 {
+		safeCopyHeader.Digest = make([][]byte, len(header.Digest))
+		copy(safeCopyHeader.Digest, header.Digest)
+	}
+
+	return &safeCopyHeader
+}
+
```
ansermino: I think a block builder pattern could avoid things like this. If we treat every block/header as immutable by using getters for all the fields instead of exporting them, we can limit allocations to only occur when we fetch from the DB. Please let me know your thoughts.

state/block.go
thomasmodeneis: Well, its not returning it from the DB, it returns a pointer; Latest implies a behavior that its not there.

core/types/types.go
thomasmodeneis: Block and Header will be used also by the RPC and perhaps other instances. Doing a safe copy will avoid struggles as I can see looking to other implementations. We can always remove it later if it turns out to be useless.

state/block.go
thomasmodeneis: Updated as requested.

state/block_test.go
```diff
@@ -40,29 +38,111 @@ func TestGetBlockByNumber(t *testing.T) {
...
+func TestAddBlock(t *testing.T) {
+	dataDir, err := ioutil.TempDir("", "TestAddBlock")
+	require.Nil(t, err)
+
+	//Create a new blockState
+	blockState, err := NewBlockState(dataDir)
+	require.Nil(t, err)
+
+	//Close DB
+	defer func() {
```
ryanchristo: Should we name this function and make it reusable? It would be nice to have a `blockState.Stop()` method that we use whenever starting the state service and then update this to `defer blockState.Stop()`. The error would then be logged in the `Stop` method and tested in its own `StartStateService` test. Thoughts?

core/types/types.go
```diff
@@ -18,19 +18,18 @@ package types
 
 import (
 	"errors"
-	"math/big"
-
 	scale "github.com/ChainSafe/gossamer/codec"
 	"github.com/ChainSafe/gossamer/common"
+	"math/big"
```
noot: i think this will cause a goimports error, could you move this back under "errors"

core/types/types.go
noot: also, maybe change this to a receiver function? `func (h *Header) DeepCopy() *Header`

state/block.go
```diff
@@ -3,20 +3,19 @@ package state
...
 )
 
+// blockState defines fields for manipulating the state of blocks, such as BlockTree, BlockDB and Header
 type blockState struct {
-	bt          *blocktree.BlockTree
-	db          *polkadb.BlockDB
-	latestBlock types.BlockHeader
+	bt     *blocktree.BlockTree
+	db     *polkadb.BlockDB
+	header *types.Header
```
noot: perhaps name latestHeader? this is unclear

state/block.go
```diff
@@ -194,13 +194,29 @@ func (bs *blockState) SetBabeHeader(epoch uint64, slot uint64, blockData babe.Ba
...
+func (bs *blockState) AddBlock(block types.Block) error {
+
 	// Set the latest block
-	if block.Number.Cmp(bs.latestBlock.Number) == 1 {
-		bs.latestBlock = block
+	if bs.header == nil {
+		bs.header = types.SafeCopyHeader(block.Header)
+	} else if block.Header.Number != nil && block.Header.Number.Cmp(bs.header.Number) == 1 { // If the new block has a number greater than current latestBlock
+		bs.header = types.SafeCopyHeader(block.Header)
+	}
```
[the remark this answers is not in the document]
thomasmodeneis: I think it is both correct, comment and code. By the way this is not my code, but it does what it says.

state/block_test.go
```diff
@@ -40,29 +38,111 @@ func TestGetBlockByNumber(t *testing.T) {
 
 	// Set the block's header & blockData in the blockState
 	// SetHeader also sets mapping [blockNumber : hash] in DB
-	err = stateService.Block.SetHeader(*blockHeader)
-	if err != nil {
-		t.Fatal(err)
-	}
+	err = stateService.Block.SetHeader(blockHeader)
+	require.Nil(t, err)
```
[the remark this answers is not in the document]
thomasmodeneis: Well, using require or not using is a matter of taste on tests, if people like it, it will be more adopted. There is no reason to change everything for it just yet IMHO.

state/block_test.go
thomasmodeneis: Sure but that will go a bit outside of the scope of this ticket.

state/block.go
thomasmodeneis: sure, I still dont understand how making it latestHeader when the type is Header any more clear. But if you insist I can change.

core/types/types.go
thomasmodeneis: Why changing it into a receiver going to be any better than it is now apart from your opinion ?

state/block.go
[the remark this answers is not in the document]
ryanchristo: The two outcomes do the same thing. I guess I was asking if it should be one condition using the `||` operator? It works but duplicates `bs.header = types.SafeCopyHeader(block.Header)`.

The comment says "If the new block has a number greater than current latestBlock" and the condition says `block.Header.Number.Cmp(bs.header.Number) == 1` (if the number equals 1). Maybe I am misunderstanding something here?

core/types/types.go
noot: if it is a receiver, we don't need to import `types` in other code that might use the Header type and want to copy it, we can just do `header.DeepCopy()`

state/block_test.go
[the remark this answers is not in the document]
ryanchristo: Ok, I don't think we should break consistency with how we have been writing it throughout the codebase. We should open and issue and switch format if we want to adopt this throughout the codebase but changing it here is not necessary and breaks consistency.

state/block_test.go
ryanchristo: Adding a second test to the state service provides the opportunity to reuse some code in the tests. This was requested and carried over from the previous pull request. If you want to open a separate ticket for this that works too. This pattern is consistent with out other services and we should implement it at some point.

state/block_test.go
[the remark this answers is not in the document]
thomasmodeneis: Its test, no consistency is required. This is just how we can express the assertions, there are other ways to do it as well. If everytime we need to change everything to introduce a small improvement we will never be anywhere, always refactoring.

state/block_test.go
thomasmodeneis: We ofcourse should always think on re-using, I'll suggest we tackle this as part of the refactor on https://github.com/ChainSafe/gossamer/issues/530

core/types/types.go
thomasmodeneis: Well, if someone is going to call this method, its already importing types, otherwise the parameter of the function could not be instantiated/passed. 
I see your point and it makes sense, I can refactor.

state/block.go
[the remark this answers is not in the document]
thomasmodeneis: updated, please re-check.

core/types/types.go
thomasmodeneis: updated, please re-check.

state/block_test.go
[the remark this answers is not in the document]
edwardmack: I'm reluctant to comment here, since I see this conversation has been resolved.  I like the use of `require` (I was not aware of this package until now).  I agree it's a matter of taste, and it seems like a good idea.  I also agree that we should strive for consistency, it make the code base more readable/manageable.  However I don't think it's necessary to refactor existing working (and manageable), these are test cases, and both are readable (just that one way contains more lines).  Just IMHO.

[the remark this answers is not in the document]
ryanchristo: We should be consistent with the format of our tests. Not sure why "Its test, no consistency is required" applies here or where that rule should apply. If this is an improvement, we should make note to implement the improvement in all of our tests. The more consistent we are, the easier our code will be to read and maintain.

This may be an obvious change to you but it is less obvious to developers with less experience in Go and/or with our codebase.

state/block_test.go
ryanchristo: Can you update #530, make note in it, or open a new issue to address this?

state/block.go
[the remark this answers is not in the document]
ryanchristo: Aha! Thanks! I was misunderstanding `Cmp` on the latter.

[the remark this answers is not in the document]
thomasmodeneis: no worries

state/block_test.go
thomasmodeneis: sure

Issue #504, opened 2019-12-21
BABE: determine when block is full


noot:
we finalize a block either when the block is full or the slot is finished. Need to determine method for deciding when the block is "full". see substrate: https://github.com/paritytech/substrate/blob/a98625501be68cc3084e666497c16b111741dded/client/basic-authorship/src/basic_authorship.rs#L208
https://github.com/paritytech/substrate/blob/8e98643384e28dfa0a88079161330309a6ee408a/client/block-builder/src/lib.rs#L156

PR #507, opened 2019-12-30
consensus/babe, core, runtime: apply extrinsics

### LINKED ISSUE #441: runtime: when to use malloc, free in wasm calls
in the runtime, every time memory is malloc'd, it needs to also be free'd. however, if an ext_ function mallocs some memory, it should not be freed until the entire runtime call is finished (since at that point we know the data is no longer needed), not when the ext_ function is finished. we need to investigate how to do this

### LINKED ISSUE #501: Apply block extrinsics
For each extrinsic to be included in a block in build-block, we need to call `apply_extrinsic` to add it. Right now there is some strange storage trie issue happening when we go from `initialize_block` to `apply_extrinsic`, need to investigate.

### LINKED ISSUE #506: Change types.BlockHeader to unexport hash field and add getter/setter
currently, types.Block uses types.BlockHeaderWithHash. it should be using types.BlockHeader so that scale decoding from BABE works, however, the block tree requires a hash field. we should change types.BlockHeader to have an unexported hash field and add a Hash() function on it that sets the hash if nil, otherwise it gets the hash.

noot:
## Changes
- implement apply extrinsics loop in build-block
- also use runtime allocator to allocate space in babe runtime calls.

## Tests:
```
go test ./consensus/...
```

### Issues:
closes #441  closes #501


consensus/babe/babe.go
```diff
@@ -215,9 +215,45 @@ func (b *Session) buildBlock(parent *types.BlockHeaderWithHash, slot Slot) (*typ
...
+	if err != nil {
+		return nil, err
+	}
+
+	// for each extrinsic in queue, add it to the block, until the slot ends or the block is full.
+	// TODO: check when block is full
+	extrinsic := b.nextReadyExtrinsic()
+	for !endOfSlot(slot) && extrinsic != nil {
+		log.Debug("build_block", "applying extrinsic", extrinsic)
+		_, err = b.applyExtrinsic(*extrinsic)
```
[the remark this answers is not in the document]
noot: that definitely makes sense, I'll update accordingly!

runtime/imports.go
```diff
@@ -288,6 +306,20 @@ func ext_get_allocated_storage(context unsafe.Pointer, keyData, keyLen, writtenO
...
+	actual := make([]byte, len(val))
+	copy(actual, val)
+
+	if bytes.Equal(key, []byte(":extrinsic_index")) {
+		val[0] &= 0
+		val[1] &= 0
+		val[2] &= 0
+		val[3] &= 0
+	}
+
```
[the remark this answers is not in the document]
noot: yeah it was meant to be a temporary fix for sure, gonna remove it now and work on what you mentioned

consensus/babe/types.go
```diff
@@ -43,10 +49,63 @@ type AuthorityData struct {
...
+
+func NewInherentsData() *InherentsData {
+	return &InherentsData{
+		data: make(map[[8]byte]([]byte)),
+	}
+}
+
+func (d *InherentsData) SetInt64Inherent(key []byte, data uint64) error {
+	if len(key) != 8 {
+		return errors.New("inherent key must be 8 bytes")
```
ryanchristo: Should we update the `key` input type instead of checking its length?

consensus/babe/runtime.go
```diff
@@ -46,34 +46,55 @@ func (b *Session) configurationFromRuntime() error {
...
+func (b *Session) applyExtrinsic(data types.Extrinsic) ([]byte, error) { //nolint:unused
+	ptr, err := b.rt.Malloc(uint32(len(data)))
+	if err != nil {
+		return nil, err
+	}
+
+	b.rt.Store(data, int32(ptr))
 
-	return b.rt.Exec(runtime.BlockBuilderApplyExtrinsic, loc, e)
+	ret, err := b.rt.Exec(runtime.BlockBuilderApplyExtrinsic, int32(ptr), data)
```
ryanchristo: Is `ret` short for return? We should be more descriptive here and elsewhere (ie, `rawdata`).

consensus/babe/runtime.go
```diff
@@ -46,34 +46,55 @@ func (b *Session) configurationFromRuntime() error {
...
+func (b *Session) inherentExtrinsics(data []byte) ([]byte, error) { //nolint:unused
+	ptr, err := b.rt.Malloc(uint32(len(data)))
+	if err != nil {
+		return nil, err
+	}
 
-	return b.rt.Exec(runtime.BlockBuilderInherentExtrinsics, loc, blockInherentData)
+	b.rt.Store(data, int32(ptr))
+
+	ret, err := b.rt.Exec(runtime.BlockBuilderInherentExtrinsics, int32(ptr), data)
```
ryanchristo: Is `ret` short for return? We should be more descriptive here and elsewhere (ie, `rawdata`).

consensus/babe/babe_test.go
```diff
@@ -508,22 +509,27 @@ func TestBuildBlock(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	// the hash of an empty trie
-	emptyRootHash, err := common.HexToHash("0x03170a2e7597b7b7e3d84c05391d139a62b157e78786d8c082f29dcf4c111314")
+	// hash of parent header
+	parentHash, err := common.HexToHash("0x01d43593c715fdd31c61141abd04a99fd6822c8558854ccde39a5684e7a56da2")
```
ryanchristo: Where are the example hashes coming from?

consensus/babe/runtime.go
```diff
@@ -93,12 +114,15 @@ func (b *Session) finalizeBlock() (*types.Block, error) {
...
+func (b *Session) validateTransaction(data types.Extrinsic) (*tx.Validity, error) {
+	ptr, err := b.rt.Malloc(uint32(len(data)))
+	if err != nil {
+		return nil, err
+	}
+
+	b.rt.Store(data, int32(ptr))
 
-	ret, err := b.rt.Exec(runtime.TaggedTransactionQueueValidateTransaction, loc, e)
+	ret, err := b.rt.Exec(runtime.TaggedTransactionQueueValidateTransaction, int32(ptr), data)
```
ryanchristo: Is `ret` short for return? We should be more descriptive here and elsewhere (ie, `rawdata`).

consensus/babe/runtime.go
noot: yeah it is, I could call it `data` or something.

consensus/babe/babe_test.go
noot: we are kinda assuming the runtime is creating the correct hash for `stateRoot` and `extrinsicsRoot`, and the parent hash is the hash of the parent block

consensus/babe/types.go
noot: the issue is we declare the keys as global vars, and we can't turn a string into a fixed byte array:
eg. right now we have this, which is fine: `var Timstap0 = []byte("timstap0")`
but this doesn't work: `var Timstap0 = [8]byte("timstap0")`

consensus/babe/babe.go
```diff
@@ -221,16 +222,64 @@ func (b *Session) buildBlock(parent *types.BlockHeader, slot Slot) (*types.Block
...
 }
 
+// buildBlockExtrinsics applies extrinsics to the block. it returns an array of included extrinsics.
+// for each extrinsic in queue, add it to the block, until the slot ends or the block is full.
+// if any extrinsic fails, it returns an empty array and an error.
+func (b *Session) buildBlockExtrinsics(slot Slot) ([]*tx.ValidTransaction, error) {
+	extrinsic := b.nextReadyExtrinsic()
+	included := []*tx.ValidTransaction{}
+
+	// TODO: check when block is full
```
ryanchristo: Is there an issue open for this?

consensus/babe/babe.go
```diff
@@ -221,16 +222,64 @@ func (b *Session) buildBlock(parent *types.BlockHeader, slot Slot) (*types.Block
...
+	for !hasSlotEnded(slot) && extrinsic != nil {
+		log.Trace("build_block", "applying extrinsic", extrinsic)
+		ret, err := b.applyExtrinsic(*extrinsic)
+		if err != nil {
+			return nil, err
+		}
+
+		// if ret == 0x0001, there is a dispatch error; if ret == 0x01, there is an apply error
+		if ret[0] == 1 || bytes.Equal(ret[:2], []byte{0, 1}) {
+			// TODO: specific error code checking
```
ryanchristo: Is there an issue open for this?

consensus/babe/babe.go
noot: https://github.com/ChainSafe/gossamer/issues/504

consensus/babe/babe.go
noot: https://github.com/ChainSafe/gossamer/issues/514

PR #508, opened 2019-12-31
consensus/babe, crypto/sr25519, core: add vrf functionality

### LINKED ISSUE #210: Implement VRF
See [#247      (ChainSafe/gossamer#194)](https://zube.io/chainsafe-2/gossamer/c/247)  & [ChainSafe/gossamer#207 (ChainSafe/gossamer#171)](https://zube.io/chainsafe-2/gossamer/c/207)

### LINKED ISSUE #293: VRF

noot:
## Changes
- add vrf sign and verify to crypto/sr25519
- add vrf sign to babe lottery

## Tests:
```
go test ./consensus/... ./crypto/...
```

### Issues:
closes #210 closes #293


consensus/babe/babe.go
```diff
@@ -17,48 +17,52 @@
...
-	keystore       *keystore.Keystore
+	keypair        *sr25519.Keypair
 	rt             *runtime.Runtime
 	config         *BabeConfiguration
 	authorityIndex uint64
 	authorityData  []AuthorityData
 	epochThreshold *big.Int // validator threshold for this epoch
 	txQueue        *tx.PriorityQueue
-	isProducer     map[uint64]bool    // whether we are a block producer at a slot
+	isProducer     map[uint64][]byte  // whether we are a block producer at a slot
```
ryanchristo: Why `[]byte` instead of `bool` here?

noot: we want to store the proof that we're a block producer if we are, so I've changed it to store the proof in this mapping (the byte array will be nil if we're not a producer)

ryanchristo: That makes sense. We should probably update the comment and maybe the name as well.

noot: I will update the comment, not sure what would be a good name, like `slotToProof` or something?

ryanchristo: Yea, that works. `isProducer` implies a bool so keeping it could be a bit confusing but not totally necessary I guess.

noot: ok, renamed

core/service.go
```diff
@@ -54,10 +55,24 @@ type Config struct {
...
+			return nil, err
+		}
+		cfg.Keystore.Insert(kp)
+		keys = cfg.Keystore.Sr25519Keypairs()
+	}
 
 	// BABE session configuration
 	bsConfig := &babe.SessionConfig{
-		Keystore:  cfg.Keystore,
+		Keypair:   keys[0].(*sr25519.Keypair),
```
ansermino: This seems very unsafe

noot: yeah.. I mean if the keystore is empty, we generate a new keypair, so keys[0] should always exist. there probably is a better way to specify what key we want, but we might want to use different keys for different babe sessions, which makes it a bit more complex.

crypto/sr25519/sr25519.go
```diff
@@ -123,6 +123,16 @@ func (kp *Keypair) Private() crypto.PrivateKey {
...
+// VrfSign creates a VRF output and proof from a message and private key
+func (kp *Keypair) VrfSign(msg []byte) (*sr25519.VrfOutput, *sr25519.VrfProof, error) {
+	return kp.private.VrfSign(msg)
+}
+
+// VrfVerify confirms that the output and proof are valid given a message and public key
+func (kp *Keypair) VrfVerify(msg []byte, output *sr25519.VrfOutput, proof *sr25519.VrfProof) (bool, error) {
+	return kp.public.VrfVerify(msg, output, proof)
+}
+
```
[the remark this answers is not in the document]
noot: ok removed

PR #509, opened 2019-12-31
core/blocktree, core/rawdb, core/types, state: change types.BlockHeader to unexport hash field and add getter/setter

### LINKED ISSUE #506: Change types.BlockHeader to unexport hash field and add getter/setter

noot:
# Changes
- remove types.BlockHeaderWithHash and types.BlockWithHash finally >:|
- add internal hash field that has a getter/setter. note we probably only want to use `Hash()` (and `GetHash()` once Hash() has been called) in most circumstances, however we need `Set()` for the blocktree.
- update other packages accordingly

## Tests:
```
go test ./...
```

### Issues:
closes #506

### DIFF
--- a/core/blocktree/blocktree_test.go
+++ b/core/blocktree/blocktree_test.go
@@ -167,7 +167,6 @@ func TestBlockTree_LongestPath(t *testing.T) {
 		Header: &types.BlockHeader{
 			ParentHash: zeroHash,
 			Number:     big.NewInt(1),
-			//Hash:       common.Hash{0xAB},
 		},
 		Body: &types.BlockBody{},
 	}
--- a/core/blocktree/blocktree_test.go
+++ b/core/blocktree/blocktree_test.go
@@ -18,7 +18,6 @@ package blocktree
 
 import (
 	"math/big"
-	"strconv"
 	"testing"
 
 	"github.com/ChainSafe/gossamer/core/types"
@@ -42,18 +41,6 @@ func createGenesisBlock() types.Block {
 	return b
 }
 
-func intToHashable(in int) string {
-	if in < 0 {
-		return ""
-	}
-
-	out := strconv.Itoa(in)
-	if len(out)%2 != 0 {
-		out = "0" + out
-	}
-	return "0x" + out
-}
-
 func createFlatTree(t *testing.T, depth int) (*BlockTree, []common.Hash) {
 	d := &db.BlockDB{
 		Db: db.NewMemDatabase(),
@@ -69,8 +56,7 @@ func createFlatTree(t *testing.T, depth int) (*BlockTree, []common.Hash) {
 		block := types.Block{
 			Header: &types.BlockHeader{
 				ParentHash: previousHash,
-				//Hash:       hash,
-				Number: big.NewInt(int64(i)),
+				Number:     big.NewInt(int64(i)),
 			},
 			Body: &types.BlockBody{},
 		}
--- a/core/blocktree/blocktree_test.go
+++ b/core/blocktree/blocktree_test.go
@@ -75,7 +75,6 @@ func createFlatTree(t *testing.T, depth int) (*BlockTree, []common.Hash) {
 			Body: &types.BlockBody{},
 		}
 
-		//block.Header.SetHash(hash)
 		hash, err := block.Header.Hash()
 		if err != nil {
 			t.Fatal(err)
--- a/core/blocktree/blocktree_test.go
+++ b/core/blocktree/blocktree_test.go
@@ -25,7 +25,6 @@ import (
 
 	"github.com/ChainSafe/gossamer/common"
 	db "github.com/ChainSafe/gossamer/polkadb"
-	log "github.com/ChainSafe/log15"
 )
 
 var zeroHash, _ = common.HexToHash("0x00")
@@ -38,7 +37,7 @@ func createGenesisBlock() types.Block {
 		},
 		Body: &types.BlockBody{},
 	}
-	b.Header.SetHash(common.Hash{0x00})
+	_, _ = b.Header.Hash()
 	b.SetBlockArrivalTime(uint64(0))
 	return b
 }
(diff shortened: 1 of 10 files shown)


core/types/types.go
```diff
@@ -56,57 +50,35 @@ type BlockHeader struct {
...
-func (bh *BlockHeader) Hash() (common.Hash, error) {
-	enc, err := scale.Encode(bh)
-	if err != nil {
-		return [32]byte{}, err
-	}
+// SetHash sets a block header's hash. Note, you would probably want to use Hash() instead,
+// unless you know the hash has already been set
+func (bh *BlockHeader) SetHash(h common.Hash) {
+	bh.hash = h
+}
```
ansermino: Why do we need this? Seems rather unsafe

core/types/types.go
```diff
@@ -56,57 +50,35 @@ type BlockHeader struct {
...
-	ParentHash     common.Hash `json:"parentHash"`
-	Number         *big.Int    `json:"number"`
-	StateRoot      common.Hash `json:"stateRoot"`
-	ExtrinsicsRoot common.Hash `json:"extrinsicsRoot"`
-	Digest         []byte      `json:"digest"` // any additional block info eg. logs, seal
-	// TODO: Not part of spec, can potentially remove
-	Hash common.Hash `json:"hash"`
+	bh.hash = hash
+	return hash, err
 }
```
ansermino: We don't need both of these. IMO the cleaner solution would be to always call `Hash` and we can implement caching later.

This might be a good time to think about how we are constructing blocks. The builder pattern seems to be rather common for this, using it might remove the need to track whether a hash is still valid as it would only be produced once the builder is done constructing the block. (maybe there's another way to enforce this?)

core/types/types.go
noot: yeah i thought the block tree needed it but upon further inspection I removed it, just had to reqrite the test a bit

core/types/types.go
noot: yeah you're right, I have removed GetHash and added MustHash instead, also added NewBlockHeader which hashes the header for us. using NewBlockHeader only would be preferred then

core/blocktree/blocktree_test.go
```diff
@@ -159,71 +139,65 @@ func TestNode_isDecendantOf(t *testing.T) {
...
 	}
+	bt.AddBlock(extraBlock)
 
-	subChain := bt.SubChain(common.Hash{0x01}, common.Hash{0x03})
+	subChain := bt.SubChain(hashes[1], hashes[3])
 
 	for i, n := range subChain {
-		if n.hash != expectedPath[i].hash {
-			t.Errorf("expected Hash: 0x%X got: 0x%X\n", expectedPath[i].hash, n.hash)
+		if n.hash != expectedPath[i] {
```
ryanchristo: Should we use `DeepEqual` here and update the name of `expectedPath`?

core/types/types.go
```diff
@@ -56,57 +50,67 @@ type BlockHeader struct {
...
-	if err != nil {
-		return nil, err
+	bh.hash = hash
+	return bh, nil
+}
+
+// MustHash returns the hash of the block header
+// If the internal hash field is nil, it hashes the block and sets the hash field.
+// If hashing the header errors, this will panic.
+func (bh *BlockHeader) MustHash() common.Hash {
```
ryanchristo: Why do we need both `Hash` and `MustHash`?

core/blocktree/blocktree_test.go
noot: I don't think we can use DeepEqual here, since we're iterating through a list of nodes then comparing node.hash. I think the name is fine, do you have any suggestions?

core/types/types.go
noot: it's so we can use it directly as a hash, without needing a multiple-liner to check err. eg. see where MustHash is used. I could remove it, it's mostly for convenience I suppose.

core/blocktree/blocktree_test.go
ryanchristo: Ah, ok. I think I just misread the first time.

core/types/types.go
ansermino: I don't think `MustHash` makes sense here. As we disccused yesterday it's maybe not the best choice to panic, also it doesn't solve the issue that you were trying to fix here (avoiding re-calculating the hash)

core/blocktree/blocktree_test.go
```diff
@@ -38,24 +36,12 @@ func createGenesisBlock() types.Block {
 		},
 		Body: &types.BlockBody{},
 	}
-	b.Header.SetHash(common.Hash{0x00})
+	_, _ = b.Header.Hash()
```
ansermino: Why call this here?

core/types/types.go
ansermino: Looking at the code I can see why it's so important to have a oneliner for getting the block hash. Maybe we can just call it `Hash` for now, and remove the panic later on. As I mentioned yesterday, I think a builder pattern would solve the issue of updating the hash correctly.

core/blocktree/blocktree_test.go
noot: you mean the underscore part? I can remove

core/types/types.go
noot: okay, so should I leave the panic? alternatively we could just return an empty/zero hash if there's an error

ansermino: >  IMO we should just merge Hash and MustHash for now, and remove the use of panic later on when we can guarantee that the block is totally formed.

noot: okay, I have done that

PR #511, opened 2020-01-02
runtime: fix enumerated trie root

edwardmack:
## Changes
This is a fix for an issue I introduced when I was doing storage refactoring.  In ext_blake2_256_enumerated_trie_root I inadvertently changed the reference for the trie to the storge trie.  But in this function we need to use a fresh trie to calculate root hash.  This fix corrects this so that a fresh trie is used.

## Tests:
```
cd runtime/
go test -v -run TestExt_blake2_256_enumerated_trie_root
```

### Issues:

### DIFF
--- a/runtime/imports.go
+++ b/runtime/imports.go
@@ -356,8 +356,6 @@ func ext_blake2_256_enumerated_trie_root(context unsafe.Pointer, valuesData, len
 	instanceContext := wasm.IntoInstanceContext(context)
 	memory := instanceContext.Memory().Data()
 
-	//runtimeCtx := instanceContext.Data().(*RuntimeCtx)
-	//s := runtimeCtx.storage
 	t := &trie.Trie{}
 	var i int32
 	var pos int32 = 0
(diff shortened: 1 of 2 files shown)


runtime/imports.go
```diff
@@ -354,8 +356,7 @@ func ext_blake2_256_enumerated_trie_root(context unsafe.Pointer, valuesData, len
 	instanceContext := wasm.IntoInstanceContext(context)
 	memory := instanceContext.Memory().Data()
 
-	runtimeCtx := instanceContext.Data().(*RuntimeCtx)
-	s := runtimeCtx.storage
+	t := &trie.Trie{}
```
ansermino: I am concerned with this usage. We havent really made it possible to use the trie like this, it's missing DB and children trie map. Ideally we would have a simple version of the trie that is purely in memory. There might be little/no harm in merging this, but we should definitely address this in the near future.

This will be exacerbated when we start trying to cache entries of the trie.

noot: I think this is fine, we could do `trie.NewEmptyTrie` alternatively to initialize the child stuff, but I don't think it would ever be needed in this case. the enumerated trie is supposed to be temporarily constructed.

PR #515, opened 2020-01-06
consensus/babe, core/types: implement pre-digest and seal for build block

### LINKED ISSUE #510: babe: add block digests after building block
see spec section 5.14. need to add BABE header, block signature, and block seal to block digest after build-block is finished.

noot:
# Changes
- implement Digest types
- add pre-digest and seal digests to block
- build BABE header to add to pre-digest

## Tests:
```
go test ./consensus/...
```

### Issues:
closes #510


core/types/digest.go
```diff
@@ -0,0 +1,100 @@
...
+}
+
+func (d *PreRuntimeDigest) Encode() []byte {
+	enc := []byte{PreRuntimeDigestType}
+	enc = append(enc, d.ConsensusEngineId[:]...)
+	return append(enc, d.Data...)
+}
+
+// ConsensusDigest contains messages from the runtime to the consensus engine.
+type ConsensusDigest struct {
```
ryanchristo: should we name `ConsensusMessageDigest` to be specific?

noot: it's not a ConsensusMessage, in substrate the name of the digest is just `Consensus`. https://github.com/paritytech/substrate/blob/40a16efefc070faf5a25442bc3ae1d0ea2478eee/primitives/runtime/src/generic/digest.rs#L94
it's not a message in the same sense as our other messages :p

ryanchristo: Ah, ok. I was looking at Table 3.2 in the specification:

*Type name*
Changes trie root
Pre-Runtime
Consensus Message
Seal

noot: hm... I'm gonna go off the rust code for now, I think it's a bit clearer if it's not message

consensus/babe/babe.go
```diff
@@ -43,7 +43,7 @@ type Session struct {
 	authorityData  []AuthorityData
 	epochThreshold *big.Int // validator threshold for this epoch
 	txQueue        *tx.PriorityQueue
-	slotToProof    map[uint64][]byte  // for slots where we are a producer, store the vrf output+proof
+	slotToProof    map[uint64][]byte  // for slots where we are a producer, store the vrf output (bytes 0-32) + proof (bytes 32-96)
```
ansermino: I think we could make our lives easier by using a struct instead of bytes. Even just using an alias for `[]byte` and defining `GetProof()`, `GetOutput()` methods would be great.

core/types/digest.go
```diff
@@ -0,0 +1,100 @@
...
+var PreRuntimeDigestType = byte(1)
+var ConsensusDigestType = byte(2)
+var SealDigestType = byte(4)
+
+// DigestItem can be of one of four types of digest: ChangesTrieRootDigest, PreRuntimeDigest, ConsensusDigest, or SealDigest.
+// see https://github.com/paritytech/substrate/blob/f548309478da3935f72567c2abc2eceec3978e9f/primitives/runtime/src/generic/digest.rs#L77
+type DigestItem interface {
+	Type() byte
+	Encode() []byte
+	//Decode([]byte)
```
ansermino: What's blocking this? Do you have plan to implement?

consensus/babe/types.go
```diff
@@ -45,10 +47,22 @@ type AuthorityData struct {
...
-	VRFOutput          [32]byte
-	VRFProof           [32]byte
+	VrfOutput          [32]byte
+	VrfProof           [64]byte
 	BlockProducerIndex uint64
-	Slot               uint64
+	SlotNumber         uint64
+}
+
+func (bh *BabeHeader) Encode() []byte {
```
ansermino: What encoding is this? Please add comment.

consensus/babe/babe.go
ansermino: If scale is the issue here, perhaps we can open an issue and update this once Ed refactors the codec.

thomasmodeneis: Good point, I commented something similar on on the #522

core/types/digest.go
noot: nothing, I just didn't need it for this PR. will do in subsequent verification PRs

consensus/babe/types.go
noot: scale but in a custom func since it doesn't work with known size byte arrays :/

consensus/babe/babe.go
noot: okay i'll update it

core/types/digest.go
ansermino: Please open an issue to track (unless it's already WIP...)

noot: okay https://github.com/ChainSafe/gossamer/issues/523

Issue #377, opened 2019-11-05
Consensus Message


ansermino:
Need to implement the `Consensus Message` defined in the spec under Appendix E: Network Messages. See `p2p/message.go`.

thomasmodeneis:
done

PR #519, opened 2020-01-13
p2p, Consensus Message - #377

thomasmodeneis:
Sorry for the delay, this development was already ready for some time but due to holiday season I forgot to open the PR.

fixes Consensus Message #377 
## Changes
- E.1.6. Consensus Message

## Tests:
```

```

### Issues: #377


p2p/message_test.go
```diff
@@ -601,3 +602,31 @@ func TestDecodeTransactionMessageTwoExtrinsics(t *testing.T) {
...
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	ID := uint32(8)
+	encId := make([]byte, 8)
+	binary.LittleEndian.PutUint32(encId, ID)
+	expected := &ConsensusMessage{
+		ID:   ID,
+		Data: []byte{0, 0, 0, 0, 0},
```
ansermino: Perhaps use values other than zero, I think it would create a more substantial test case

thomasmodeneis: cool, thanks for the hint, could you please check again after the update ?

p2p/message.go
```diff
@@ -54,6 +55,40 @@ type Message interface {
 	Id() string
 }
 
+// DecodeMessage accepts a raw message including the type indicator byte and decodes it to its specific message type
+func DecodeMessage(r io.Reader) (m Message, err error) {
```
noot: is this used somewhere?

ryanchristo: It is not. The method that previously lived in this file named `decodeMessage` moved to `p2p/utils.go`. We no longer need this method here or we need to update the method there.

thomasmodeneis: sure

p2p/message.go
```diff
@@ -419,3 +419,64 @@ func (tm *TransactionMessage) Id() string {
 	}
 	return hash.String()
 }
+
+// ConsensusMessage is mostly opaque to us
+type ConsensusMessage struct {
+	// Identifies consensus engine.
+	ID uint32
```
noot: could you change this to `types.ConsensusEngineId`? from core/types

p2p/message_test.go
```diff
@@ -601,3 +603,53 @@ func TestDecodeTransactionMessageTwoExtrinsics(t *testing.T) {
 		t.Fatalf("Fail: got: %v expected %v", *decodedMessage, expected)
 	}
 }
+
+func TestDecodeConsensusMessage(t *testing.T) {
+	ID := binary.BigEndian.Uint32([]byte("BABE")) // 1111573061
+	encId := make([]byte, 4)
```
noot: can use `core/types.BabeEngineId` here instead

p2p/message.go
thomasmodeneis: You are right, it makes more sense to re-use the type types.ConsensusEngineId, `ConsensusEngineId` is also the name on substrate code, however in `go` the recommendation is to use uppercase for the Initialisms, check: https://github.com/golang/go/wiki/CodeReviewComments#initialisms
so I updated it to `types.ConsensusEngineID` , hope you agree with the standard, I also saw that the BabeEngineId was not upper so I updated it to upper as well.

p2p/message_test.go
thomasmodeneis: Just updated the tests, please check again.

consensus/babe/babe.go
```diff
@@ -344,7 +344,7 @@ func (b *Session) buildBlockExtrinsics(slot Slot) ([]*tx.ValidTransaction, error
 			// remove invalid extrinsic from queue
 			b.txQueue.Pop()
 
-			// readd previously popped extrinsics back to queue
+			// read previously popped extrinsics back to queue
```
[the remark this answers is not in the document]
thomasmodeneis: oh, ops

PR #524, opened 2020-01-14
codec: Update scale decoding API

### LINKED ISSUE #376: Update SCALE Decoding API
We should update this to more closely reflect something like JSON decoding. Presently you have to pass in an instance of the return type to tell SCALE what type we are decoding, and then it returns a different instance containing the result. This call should only require one instance, which must be a pointer, to both define the return type and where to store the return vale.

edwardmack:
## Changes
- Added function to `codec.docode.go` called `DecodePtr` (let me know if there is a better name) which decodes []byte similar to `Decode` except that `DecodePtr` writes the decode results to the pointer that was passed in as the object definition. It return an error if an error occurs.  Signature `DecodePtr(in []byte, t interface{}) error`  Where `in` is the byte array to be decoded, and `t` is a pointer to the object that defines how the bytes should be decoded and location decoded results will be written to.
- Replicated all existing decoding tests to test new DecodePtr functionality.
- Kept existing `Decode` function so existing code is not affected until we decide if we want to use the new function. 

## Tests:
```
go test -v ./codec/...
```

### Issues:
closes #376


codec/decode.go
```diff
@@ -169,6 +208,73 @@ func (sd *Decoder) DecodeFixedWidthInt(t interface{}) (o interface{}, err error)
 	return o, err
 }
 
+// DecodeFixedWidthInt decodes integers < 2**32 by reading the bytes in little endian
```
thomasmodeneis: update DecodeFixedWidthInt to DecodePtrFixedWidthInt

codec/decode.go
```diff
@@ -541,6 +721,25 @@ func (sd *Decoder) DecodeIntArray() ([]int, error) {
 	return o, nil
 }
 
+// DecodeIntArray decodes a byte array to an array of ints
```
thomasmodeneis: Update DecodeIntArray to DecodePtrIntArray

ansermino:
Opened this as a next step: https://github.com/ChainSafe/gossamer/issues/525

edwardmack:
Codeclimate is complaining that I have a function with more than 50 lines, `DecodePtrFixedWidthInt` is 61 lines, but I don't know how I can make it smaller.  Any suggestions?

thomasmodeneis:
> Codeclimate is complaining that I have a function with more than 50 lines, `DecodePtrFixedWidthInt` is 61 lines, but I don't know how I can make it smaller. Any suggestions?

is there a way to tell codeclimate to ignore this one special case so we can have the pr merge unblocked ?

ansermino:
> > Codeclimate is complaining that I have a function with more than 50 lines, `DecodePtrFixedWidthInt` is 61 lines, but I don't know how I can make it smaller. Any suggestions?
> 
> is there a way to tell codeclimate to ignore this one special case so we can have the pr merge unblocked ?

These are really just suggestions for better code, we usually don't enforce them. As repo admin I can (and will) dismiss them.

PR #534, opened 2020-01-17
consensus/babe: verify authorship right

### LINKED ISSUE #447: babe: verify authorship right
see spec section 5.1.6. upon receiving a block, need to verify that the authority was authorized to produce the block.

### LINKED ISSUE #523: core/types: decoding of digest items
implement decode for the various digest items.

noot:
# Changes
- implement functionality to verify authorship right of block author (check seal and pre-runtime digest)
- implement decoding of BabeHeader and digest types

## Tests:
```
go test ./consensus/...
```

### Issues:
closes #447  closes #523

### DIFF
--- a/consensus/babe/verify.go
+++ b/consensus/babe/verify.go
@@ -21,7 +21,7 @@ func (b *Session) verifySlotWinner(slot uint64, header *BabeHeader) (bool, error
 	return pub.VrfVerify(vrfInput, header.VrfOutput[:], header.VrfProof[:])
 }
 
-func (b *Session) verifyAuthorshipRight(slot uint64, header *types.BlockHeader) (bool, error) {
+func (b *Session) verifyAuthorshipRight(slot uint64, header *types.Header) (bool, error) {
 	// header should have 2 digest items (possibly more in the future)
 	// first item should be pre-digest, second should be seal
 	if len(header.Digest) < 2 {
--- a/consensus/babe/verify.go
+++ b/consensus/babe/verify.go
@@ -5,20 +5,22 @@ import (
 	"fmt"
 
 	"github.com/ChainSafe/gossamer/core/types"
+	"github.com/ChainSafe/gossamer/crypto/sr25519"
 )
 
-func (b *Session) verifySlotWinner(slot uint64, header *BabeHeader) (bool, error) {
+func (b *Session) verifySlotWinner(slot uint64, header *BabeHeader, pub *sr25519.PublicKey) (bool, error) {
 	if len(b.authorityData) <= int(header.BlockProducerIndex) {
 		return false, fmt.Errorf("no authority data for index %d", header.BlockProducerIndex)
 	}
 
-	pub := b.authorityData[header.BlockProducerIndex].id
+	//pub := b.authorityData[header.BlockProducerIndex].id
+
+	//fmt.Println(b.authorityData[0].id.Encode())
+	fmt.Println(pub.Encode())
 
 	slotBytes := make([]byte, 8)
 	binary.LittleEndian.PutUint64(slotBytes, slot)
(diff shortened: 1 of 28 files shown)


consensus/babe/verify.go
```diff
@@ -18,3 +20,76 @@ func (b *Session) verifySlotWinner(slot uint64, header *BabeHeader) (bool, error
...
+func (b *Session) verifyAuthorshipRight(slot uint64, header *types.BlockHeader) (bool, error) {
+	// header should have 2 digest items (possibly more in the future)
+	// first item should be pre-digest, second should be seal
+	if len(header.Digest) < 2 {
+		return false, fmt.Errorf("block header is missing digest items")
+	}
+
+	// check for valid seal by verifying signature
+	preDigestBytes := header.Digest[0]
+	sealBytes := header.Digest[len(header.Digest)-1]
```
ryanchristo: Is there a reason this is not `sealBytes := header.Digest[1]`?

consensus/babe/verify.go
```diff
@@ -18,3 +20,76 @@ func (b *Session) verifySlotWinner(slot uint64, header *BabeHeader) (bool, error
...
+	}
+
+	digestItem, err = types.DecodeDigestItem(sealBytes)
+	if err != nil {
+		return false, err
+	}
+
+	seal, ok := digestItem.(*types.SealDigest)
+	if !ok {
+		return false, fmt.Errorf("last digest item is not seal")
```
ryanchristo: Is there a reason this is not "second digest item"?

consensus/babe/verify.go
```diff
@@ -18,3 +20,76 @@ func (b *Session) verifySlotWinner(slot uint64, header *BabeHeader) (bool, error
...
+	}
+
+	babeHeader := new(BabeHeader)
+	err = babeHeader.Decode(preDigest.Data)
+	if err != nil {
+		return false, fmt.Errorf("cannot decode babe header from pre-digest: %s", err)
+	}
+
+	if len(b.authorityData) <= int(babeHeader.BlockProducerIndex) {
```
ryanchristo: If the length of authority data is less than or equal to the block producer index. I am a bit confused by the use of `int(babeHeader.BlockProducerIndex)` here. Maybe a comment to explain?

consensus/babe/verify.go
```diff
@@ -18,3 +20,76 @@ func (b *Session) verifySlotWinner(slot uint64, header *BabeHeader) (bool, error
...
+	ok, err = authorPub.Verify(encHeader, seal.Data)
+	if err != nil {
+		return false, err
+	}
+
+	if !ok {
+		return false, fmt.Errorf("could not verify signature")
+	}
+
+	// TODO: check if the producer has equivocated, ie. have they produced a conflicting block?
```
ryanchristo: Should we open an issue to address this?

consensus/babe/verify.go
noot: the seal is always the last digest item, not necessarily the second. there might be other digests in between digest[0] and digest[len-1].

consensus/babe/verify.go
noot: same as above, seal is always last.

consensus/babe/verify.go
noot: the BlockProducerIndex identifies which authority in the authorityData list created the block. if the BlockProducerIndex is out of the range of the authorityData list (perhaps due to an invalid block), there will be a much worse error :p the code will panic

consensus/babe/verify.go
noot: yes I'll open an issue

noot: #553

Issue #443, opened 2019-12-01
Generate status message


ryanchristo:
##  Expected Behavior

A status message is generated with the current state of the chain and network.

## Current Behavior

The status message is a currently global variable in the p2p package.

```
// TODO: use generated status message
var statusMessage = &StatusMessage{
	ProtocolVersion:     0,
	MinSupportedVersion: 0,
	Roles:               0,
	BestBlockNumber:     0,
	BestBlockHash:       common.Hash{0x00},
	GenesisHash:         common.Hash{0x00},
	ChainStatus:         []byte{0},
}
```

## Possible Solution

Generate a status message in the core service that retrieves the current state of the chain. The status message needs to be generated before the p2p service starts connecting to peers and sending status messages.

PR #551, opened 2020-01-19
cmd, p2p: update p2p config, add cmd protocol flag

### LINKED ISSUE #467: Connect and exchange messages on kusama
## Expected Behavior

We should be able to connect to Kusama, exchange status messages, send a block request message and receive a block response message.

## Current Behavior

We previously attempted to connect to Alexander. See #227 for more details.

## Related Issues

- [ ] #443
- [ ] #521
- [x] #552 / #538
- [ ] #554
- [ ] #563

### LINKED ISSUE #484: Implement LEB128 variable-length encoding

ryanchristo:
## Changes

Implement variable length encoding for network messages.

Update network identity to use ed25519 and gossamer configuration to connect to kusama. "Similar to other decentralized networks, each Polkadot RE node possesses a network private key and a network public key representing an ED25519 key pair" (4.1). 

Implement `--protocol` flag to specify a protocol id for the host to use.

## Tests:

```
make gossamer
```

successfully connect and receive "host status message not set error" (#443):
```
./build/bin/gossamer init
./build/bin/gossamer --verbosity trace
```

successfully connect using custom protocol and receive "host status message not set error" (#443):
```
./build/bin/gossamer init
./build/bin/gossamer --protocol /hello/world --verbosity trace
```

receive error "json: cannot unmarshal array" (#552):
```
./build/bin/gossamer init --genesis config/ksmcc3.json
```

receive error "connected to wrong peer" (#554):
```
./build/bin/gossamer init
./build/bin/gossamer --bootnodes /dns4/p2p.cc3-0.kusama.network/tcp/30100/p2p/QmeCit3Nif4VfNqrEJsdYHZGcKzRCnZvGxg6hha1iNj4mk --verbosity trace
```

successfully connect and receive "host status message not set error" (#443):
```
./build/bin/gossamer init
./build/bin/gossamer --bootnodes /dns4/p2p.cc3-0.kusama.network/tcp/30100/p2p/12D3KooWDgtynm4S9M3m6ZZhXYu2RrWKdvkCSScc25xKDVSg1Sjd --verbosity trace
```

### Issues:

https://hackmd.io/ksLK42jcRJGjKQuSNAvAWw?view#Connection-Examples

related to #467, closes #484


p2p/utils.go
```diff
@@ -71,7 +72,8 @@ func generateKey(seed int64, fp string) (crypto.PrivKey, error) {
 	} else {
 		r = mrand.New(mrand.NewSource(seed))
 	}
-	key, _, err := crypto.GenerateECDSAKeyPair(r)
+	// TODO: secp256k1 in polkadot / ed25519 in specification
```
ryanchristo: addressed in #554

p2p/utils.go
```diff
@@ -101,7 +103,8 @@ func loadKey(fp string) (crypto.PrivKey, error) {
 	if err != nil {
 		return nil, err
 	}
-	return crypto.UnmarshalECDSAPrivateKey(dec)
+	// TODO: secp256k1 in polkadot / ed25519 in specification
```
ryanchristo: addressed in #554

p2p/host.go
```diff
@@ -64,7 +65,6 @@ func newHost(ctx context.Context, cfg *Config) (*host, error) {
 		libp2p.DisableRelay(),
 		libp2p.Identity(cfg.privateKey),
 		libp2p.NATPortMap(),
-		libp2p.Ping(true),
```
ryanchristo: enabled by default

ryanchristo:
Also, the error received when running `--genesis config/ksmcc3.json` is addressed in #552

p2p/host.go
```diff
@@ -188,6 +188,12 @@ func (h *host) send(p peer.ID, msg Message) (err error) {
 		return err
 	}
 
+	// variable length encoding
+	_, err = s.Write(common.Uint16ToBytes(uint16(len(encMsg)))[0:1])
```
noot: is this LEB128?

noot:
also i get a SIGABRT when I run the first set of instructions :/

p2p/host.go
ryanchristo: I am not actually sure if this is officially LEB128 encoding. This is what was here before. It was commented out at some point, then removed per clean up request.

What I do know is that this is what currently works with polkadot and is the solution to the read buffer error I mentioned at some point, `b"": Not enough data to fill buffer`.

ryanchristo:
@noot I updated the instructions and I am not able to reproduce. Hmm...

ryanchristo:
Yes it should, I will fix.

ryanchristo:
cc @edwardmack @noot I have updated the tests, using `ed25519` for the network identity properly now, and I have updated the examples. @noot Let me know if you are still getting the abort error.

cmd/gossamer/configcmd.go
```diff
@@ -84,7 +84,7 @@ func makeNode(ctx *cli.Context) (*dot.Dot, *cfg.Config, error) {
 		return nil, nil, err
 	}
 
-	log.Info("🕸\t Configuring node...", "datadir", fig.Global.DataDir, "protocolID", string(gendata.ProtocolId), "bootnodes", fig.P2p.BootstrapNodes)
+	log.Info("🕸\t Configuring node...", "datadir", fig.Global.DataDir, "protocol", fig.P2p.ProtocolId, "bootnodes", fig.P2p.BootstrapNodes)
```
thomasmodeneis: should be protocolID:, raised https://github.com/ChainSafe/gossamer/issues/559 to have lint reporting things like this next time.

ryanchristo: It looks like you update this in #555 already... should it be addressed there?

ryanchristo: It looks like you update this in #555 already... should it be addressed there?

thomasmodeneis: I will resolve that in the ticket mentioned after this is merged.

noot:
@ryanchristo i'm still getting a SIGABRT when trying to run gossamer :/ if no one else is getting this i can just do a fresh clone or something

noot:
@ryanchristo on development I don't get a SIGABRT, I'm gonna look and see if I can tell what's causing it

p2p/host.go
noot: if you use the LEB encoding function from before that would probably be preferred, for longer length messages the encoding might be >2 bytes, so this wouldn't work

ryanchristo:
cc @noot ready for another look when you get a chance

p2p/utils.go
```diff
@@ -135,12 +135,13 @@ func saveKey(priv crypto.PrivKey, fp string) (err error) {
...
 
-// TODO: implement LEB128 variable-length encoding
-
-// Decodes a byte array to uint64 using LEB128 variable-length encoding
-// func leb128ToUint64(in []byte) uint64 {
-// 	return leb128.ToUInt64(in)
-// }
+// encodeMessageLEB128 applies leb128 variable length encoding
+func encodeMessageLEB128(encMsg []byte) []byte {
+	n := leb128.ToUInt64(encMsg)
```
noot: I don't think this is correct?
`ToUint64` decodes a LEB128 encoded byte array
I think you want to use `out := leb128.FromUint64(uint64(len(encMsg))); return append(out, encMsg...)`
https://godoc.org/github.com/filecoin-project/go-leb128

ryanchristo: oof, that makes more sense

ryanchristo: fixed!

PR #555, opened 2020-01-21
cmd,trie,polkadb,p2p,config - Cannot handle home dir #496

### LINKED ISSUE #496: Cannot handle home dir
## Expected Behavior

Should expand `~` to home dir.

## Current Behavior

Creates directory `./~` when starting gossamer

1. `./build/bin/gossamer init --genesis genesis.json`
2. `gossamer --config ./config.toml `
3. `ls -la `

config.toml:
```toml
[global]
data-dir = "~/.gossamer"
```

thomasmodeneis:
Fixes #496 

## Changes
- Update cli Actions to support reversed usage
- Update var names to be more legible
- Update from id to ID in multiple places
- Update touched testcases to require funcs
- Add err log on main func
- Update weird discrepancy on log for Close func on database.go and table.go
- Update config.toml to use tilde
- 
- 

## Tests:
```

```

### Issues: #496


cmd/utils/flags.go
```diff
@@ -123,3 +123,15 @@ var (
 		Usage: "Specify account type as secp256k1",
 	}
 )
+
+// FixFlagsOder allow us to use various flag order formats, eg: (gossamer init --config config.toml and  gossamer --config config.toml init)
+func FixFlagsOder(f func(ctx *cli.Context) error) func(*cli.Context) error {
```
noot: should be `FixFlagOrder`?

cmd/gossamer/configcmd.go
```diff
@@ -46,15 +48,17 @@ import (
 
 // makeNode sets up node; opening badgerDB instance and returning the Dot container
 func makeNode(ctx *cli.Context) (*dot.Dot, *cfg.Config, error) {
-	fig, err := getConfig(ctx)
+	defaultConfig, err := getConfig(ctx)
```
noot: is this the defaultConfig? I don't think it's always the default

cmd/utils/flags.go
thomasmodeneis: Sure, I mean it fixes many flags, so that why the plural there, but can change.

cmd/gossamer/configcmd.go
thomasmodeneis: cfg.DefaultConfig is the struct that is used to marshal the config into, so that is why I called it defaultConfig everywhere its used, its already better than fig IMHO.

cmd/utils/flags.go
```diff
@@ -123,3 +123,17 @@ var (
...
+		for _, flagName := range ctx.FlagNames() {
+			if ctx.IsSet(flagName) {
+				if err := ctx.GlobalSet(flagName, ctx.String(flagName)); err != nil {
+					log.Error("Error when fixing flag", "flagName", flagName)
+				}
+			}
+		}
+		return f(ctx)
+	}
+}
```
ansermino: This is not how flags should be used. Global flags apply to `gossamer`, local flags should be applied to sub-commands (eg. `accounts`). This seems to be standard for most CLIs

cmd/gossamer/configcmd.go
ansermino: `cfg.DefaultConfig()` returns a copy of the config with all defaults applied, we then apply all specified options, so at this point it is not a default config

cmd/gossamer/configcmd.go
```diff
@@ -328,3 +338,24 @@ var tomlSettings = toml.Config{
...
+// home will return the HOME ENV DIR OR the current user dir
+func home() string {
+	if homeDir := os.Getenv("HOME"); homeDir != "" {
+		return homeDir
+	}
+	if usrDir, err := user.Current(); err == nil {
+		return usrDir.HomeDir
+	}
+	return ""
+}
```
ansermino: Same as `cfg.homeDir()`? Maybe it makes sense for this logic to live there, up to you

cmd/gossamer/configcmd.go
```diff
@@ -131,21 +135,26 @@ func loadStateAndRuntime(ss *state.StorageState, ks *keystore.Keystore) (*runtim
...
+	} else {
+		log.Debug("Config File is not set")
 	}
 
 	// Parse CLI flags
-	setGlobalConfig(ctx, &fig.Global)
-	setP2pConfig(ctx, &fig.P2p)
-	setRpcConfig(ctx, &fig.Rpc)
-	return fig, nil
+	log.Debug("Parse CLI flags")
```
ansermino: Doesn't really add any context

cmd/gossamer/configcmd.go
```diff
@@ -328,3 +338,24 @@ var tomlSettings = toml.Config{
 		return fmt.Errorf("field '%s' is not defined in %s%s", field, rt.String(), link)
 	},
 }
+
+// tilde will expand a tilde prefix path to full home path
+func tilde(targetPath string) string {
```
ansermino: I think something like `expandTilde` would be more descriptive

cmd/utils/flags.go
thomasmodeneis: Well, the issue was reported as `./build/bin/gossamer init --genesis genesis.json` does not work.
Indeed, it does not work because its order is improper. It would work if used like: `./build/bin/gossamer --genesis genesis.json init`
So, the fix is needed to avoid issues like this.

cmd/gossamer/configcmd.go
thomasmodeneis: `defaultConfig` maybe not the best name here, but its the name of the struct that is being marshaled into, but `fig` as a var name was also not good either. Do you have a better suggestion ?

cmd/gossamer/configcmd.go
thomasmodeneis: what do you mean ? you want to remove it ?

cmd/gossamer/configcmd.go
thomasmodeneis: sure

cmd/gossamer/configcmd.go
thomasmodeneis: ok, how about calling it currentConfig ?

cmd/utils/flags.go
ansermino: `--genesis` only applies to the `init` command, `gossamer --genesis genesis.json` shouldn't be allowed as the flag isn't relevant.

cmd/gossamer/configcmd.go
ansermino: Is this resolved by the comment above?

cmd/gossamer/configcmd.go
ansermino: I don't think this provides any additional information. That is what I'm suggesting

cmd/gossamer/configcmd_test.go
```diff
@@ -197,14 +185,11 @@ func TestGetConfig(t *testing.T) {
...
 		context := cli.NewContext(app, set, nil)
 
-		fig, err := getConfig(context)
-		if err != nil {
-			t.Fatalf("failed to set fig %v", err)
-		}
+		defaultConfig, err := getConfig(context)
+		require.Nil(t, err)
+
+		require.Equal(t, defaultConfig, c.expected)
```
edwardmack: shouldn't this be `require.Equal(t, c.expected, defaultConfig)`? I think the parameters for Equal are `t TestingT, expected interface{}, actual interface{}, msgAndArgs ...interface{}` Minor, but could be confusing reading the Error Trace if this test fails.

cmd/gossamer/configcmd_test.go
```diff
@@ -229,30 +214,24 @@ func TestSetGlobalConfig(t *testing.T) {
...
+			require.Nil(t, err)
 
 			tCfg := &cfg.GlobalConfig{}
 
 			setGlobalConfig(context, tCfg)
 
-			if !reflect.DeepEqual(*tCfg, c.expected) {
-				t.Errorf("\ngot: %+v \nexpected: %+v", tCfg, c.expected)
-			}
+			require.Equal(t, *tCfg, c.expected)
```
edwardmack: see above, switch actual/expected

cmd/gossamer/configcmd_test.go
```diff
@@ -310,17 +289,13 @@ func TestSetP2pConfig(t *testing.T) {
...
+			require.Nil(t, err)
 
 			input := cfg.DefaultConfig()
 			// Must call global setup to set data dir
 			setP2pConfig(context, &input.P2p)
 
-			if !reflect.DeepEqual(input.P2p, c.expected) {
-				t.Fatalf("\ngot %+v\nexpected %+v", input.P2p, c.expected)
-			}
+			require.Equal(t, input.P2p, c.expected)
```
edwardmack: switch actual/expected

cmd/gossamer/configcmd_test.go
```diff
@@ -368,16 +343,12 @@ func TestSetRpcConfig(t *testing.T) {
...
-			}
+			require.Nil(t, err)
 
 			input := cfg.DefaultConfig()
 			setRpcConfig(context, &input.Rpc)
 
-			if !reflect.DeepEqual(input.Rpc, c.expected) {
-				t.Fatalf("\ngot %+v\nexpected %+v", input.Rpc, c.expected)
-			}
+			require.Equal(t, input.Rpc, c.expected)
```
edwardmack: switch actual/expected

cmd/gossamer/configcmd_test.go
thomasmodeneis: Oh, good one!

cmd/utils/flags.go
thomasmodeneis: Well, the issue is like I said: 
`gossamer init --genesis genesis.json` does not parse the flags correctly, and values are nil.
When I run `gossamer --genesis genesis.json init` all works well and flags are processed.

cmd/gossamer/configcmd_test.go
thomasmodeneis: thanks.

cmd/gossamer/configcmd_test.go
thomasmodeneis: thanks.

cmd/gossamer/configcmd_test.go
thomasmodeneis: thanks.

cmd/gossamer/configcmd.go
ryanchristo: I agree. If everything is working as expected and there is no new information, the debug is probably not necessary.

cmd/gossamer/configcmd.go
ryanchristo: I agree, and keeping this consistent with the name used in `configcmd.go` make sense

cmd/gossamer/configcmd.go
ryanchristo: I agree, `currentConfig` would be probably be better or keeping it as it was.

cmd/utils/flags.go
ansermino: I believe this issue lies in the usage of `context.GlobalBool` and `context.Bool`. Afaik, the first will only return flags on the main command (`gossamer`) and the other on the subcommand (`init`)

ansermino: (bool is just an example, applies for all flag types)

thomasmodeneis:
> Looks good, minor comments mentioned in code.
> Regarding #496, when I do steps to reproduce I still have an issue. After running steps, a dir `~/.gossamer` is created in dir I'm running from. So, I now have:
> 
> * `~/.gossamer` that is `/home/emack/.gossamer` which was created from init
> * `/home/emack/projects/chainsafe/gossamer/~/.gossamer` from which `~/.gossamer` was created after running `gossamer --config ./config.toml`

I found that the culprit was `setGlobalConfig` messing up things silently, please check it again, thanks!

cmd/utils/flags.go
thomasmodeneis: Right, IMHO having the method FixFlagOrder executed on all cmds reduces the complexity and avoid us having to fight with how the flags all the time. AFAIK It does not add any collateral effects as well.

cmd/gossamer/configcmd_test.go
```diff
@@ -229,30 +213,24 @@ func TestSetGlobalConfig(t *testing.T) {
...
-			}
+			require.Equal(t, c.expected, *tCfg)
 		})
 	}
 }
 
 func TestCreateP2PService(t *testing.T) {
 	gendata := &genesis.GenesisData{
-		ProtocolId: "gossamer",
+		ProtocolID: "gossamer",
```
ryanchristo: Can we please change this to `/gossamer/test` to match the protocol id in other tests?

cmd/gossamer/genesis.go
```diff
@@ -15,25 +15,25 @@ import (
...
+	log.Debug("Loading genesis", "genesisPath", genesisPath, "dataDir", dataDir)
 
-	gen, err := genesis.LoadGenesisData(fp)
+	gen, err := genesis.LoadGenesisData(genesisPath)
 	if err != nil {
 		return err
 	}
 
-	log.Info("🕸\t Initializing node", "name", gen.Name, "id", gen.Id, "protocolID", gen.ProtocolId, "bootnodes", common.BytesToStringArray(gen.Bootnodes))
+	log.Info("🕸\t Initializing node", "Name", gen.Name, "ID", gen.ID, "protocolID", gen.ProtocolID, "Bootnodes", common.BytesToStringArray(gen.Bootnodes))
```
ryanchristo: I am not opposed to capitalizing values but this is inconsistent with all other logs and `protocolID` is the only one not capitalized. Maybe keep the key values lowercase for now and we can look at log formatting in an issue on its own to make sure we are looking at how we are logging as a whole.

cmd/gossamer/configcmd_test.go
```diff
@@ -383,16 +357,12 @@ func TestSetRpcConfig(t *testing.T) {
...
-			}
+			require.Nil(t, err)
 
 			input := cfg.DefaultConfig()
 			setRpcConfig(context, &input.Rpc)
 
-			if !reflect.DeepEqual(input.Rpc, c.expected) {
-				t.Fatalf("\ngot %+v\nexpected %+v", input.Rpc, c.expected)
-			}
+			require.Equal(t, c.expected, input.Rpc)
```
ryanchristo: This is nice. Does it accomplish the same level of depth as `reflect.DeepEqual`?

cmd/gossamer/configcmd.go
```diff
@@ -45,15 +46,18 @@ import (
...
-	fig, err := getConfig(ctx)
+	currentConfig, err := getConfig(ctx)
 	if err != nil {
 		return nil, nil, err
 	}
 
 	var srvcs []services.Service
 
+	dataDir := currentConfig.Global.DataDir
+	log.Debug("Will create service, initialize stateDB and blockDB", "dataDir", dataDir, "currentConfig.Global.DataDir", currentConfig.Global.DataDir)
```
noot: don't think this log is necessary

cmd/gossamer/configcmd.go
```diff
@@ -132,21 +136,30 @@ func loadStateAndRuntime(ss *state.StorageState, ks *keystore.Keystore) (*runtim
 
 // getConfig checks for config.toml if --config flag is specified and sets CLI flags
 func getConfig(ctx *cli.Context) (*cfg.Config, error) {
-	fig := cfg.DefaultConfig()
+	currentConfig := cfg.DefaultConfig()
 	// Load config file.
 	if file := ctx.GlobalString(utils.ConfigFileFlag.Name); file != "" {
-		err := loadConfig(file, fig)
+		configFile := ctx.GlobalString(utils.ConfigFileFlag.Name)
+		log.Debug("Loading config file", "configFile", configFile)
```
noot: same with this

cmd/gossamer/configcmd.go
```diff
@@ -132,21 +136,30 @@ func loadStateAndRuntime(ss *state.StorageState, ks *keystore.Keystore) (*runtim
...
+	//expand tilde or dot
+	newDataDir := expandTildeOrDot(currentConfig.Global.DataDir)
+	currentConfig.Global.DataDir = newDataDir
+
 	// Parse CLI flags
-	setGlobalConfig(ctx, &fig.Global)
-	setP2pConfig(ctx, &fig.P2p)
-	setRpcConfig(ctx, &fig.Rpc)
-	return fig, nil
+	log.Debug("Going set CLI flags (will override config file if set) ", "currentConfig", currentConfig)
```
noot: same with this

cmd/gossamer/configcmd.go
```diff
@@ -163,14 +176,16 @@ func loadConfig(file string, config *cfg.Config) error {
 	if err = tomlSettings.NewDecoder(f).Decode(&config); err != nil {
 		return err
 	}
+	log.Debug("Loaded configuration", "config", config)
```
noot: same with this

cmd/gossamer/configcmd_test.go
thomasmodeneis: yep

cmd/gossamer/genesis.go
thomasmodeneis: Well the basic rule is you log the field name with the exact same name as the variable. protocolID is not missing, I am not sure what you mean.

cmd/gossamer/configcmd.go
thomasmodeneis: sure

cmd/gossamer/configcmd.go
thomasmodeneis: ok

cmd/gossamer/configcmd.go
thomasmodeneis: ok

cmd/gossamer/configcmd.go
thomasmodeneis: ok

cmd/gossamer/configcmd_test.go
thomasmodeneis: I dont understand your request

ryanchristo:
```
ProtocolID: "/gossamer/test"
```

cmd/gossamer/genesis.go
ryanchristo: `"protocolID"` should be `"ProtocolID"` if you are following that rule

ryanchristo:
cc @thomasmodeneis can you please re-review when you get a chance?

If I change `Id` and `ProtocolId` where suggested, I would be required to change it everywhere, which is already addressed in #555 and therefore will be addressed here if #555 is merged first

cmd/gossamer/genesis.go
thomasmodeneis: oh you mean the log string, sure, good point :)

cmd/gossamer/configcmd_test.go
thomasmodeneis: sure

Makefile
```diff
@@ -45,6 +45,10 @@ build:
 	@echo "  >  \033[32mBuilding binary...\033[0m "
 	GOBIN=$(PWD)/build/bin go run scripts/ci.go install
 
+# init: Init the gossamer folder using  gssmr0.json genesis file and default config file
+init:
+	./build/bin/gossamer init --genesis config/gssmr0.json --verbosity debug --config ./config.toml
+
```
ryanchristo: `--genesis config/gssmr0.json` is the default, is it needed here?

thomasmodeneis: yes, the obvious should be mentioned to allow for change in case you want to do it differently. That is the whole point of adding the task to the Makefile.

PR #557, opened 2020-01-21
cmd, core, consensus/babe: add state interface

### LINKED ISSUE #461: Remove DB Service et al.
- [x] remove DB service
- [x] remove usages of Trie, Blocktree, Network Service
- [ ] remove rawDb
- [ ] remove `runtime.StorageRoot()` (replace `core.Service.StorageRoot()` with state service)

### LINKED ISSUE #465: Add State interfaces
- [ ] add state interface for each service 
- [ ] update services to take interface in constructor, pass in respective state service sub-component
- [ ] remove interfaces in state service

noot:
## Changes
- add state interfaces to core, consensus/babe
- add state to core and babe configs, add to their respective service structs
- pass in stateSrv.Block and stateSrv.Storage to core in configcmd
- remove state/interface.go
- remove rawDB
- update core.StorageRoot to use storage interfaces; remove runtime.StorageRoot

## Tests:
```
go test ./... 
```

### Issues:
closes #465 closes #461


core/service.go
```diff
@@ -38,23 +38,28 @@ var _ services.Service = &Service{}
...
-	Runtime  *runtime.Runtime
-	MsgRec   <-chan p2p.Message
-	MsgSend  chan<- p2p.Message
+	BlockState   BlockState
+	StorageState StorageState
+	Keystore     *keystore.Keystore
+	Runtime      *runtime.Runtime
+	MsgRec       <-chan p2p.Message
+	MsgSend      chan<- p2p.Message
+	NewBlocks    chan types.Block // only used for testing purposes
```
ansermino: Why do we need this?

core/service.go
```diff
@@ -38,23 +38,28 @@ var _ services.Service = &Service{}
...
-	rt      *runtime.Runtime
-	bs      *babe.Session
-	blkRec  <-chan types.Block // receive blocks from BABE session
-	msgRec  <-chan p2p.Message // receive messages from p2p service
-	msgSend chan<- p2p.Message // send messages to p2p service
+	blockState   BlockState
+	storageState StorageState
+	rt           *runtime.Runtime
+	bs           *babe.Session
+	blkRec       <-chan types.Block // receive blocks from BABE session
```
ansermino: Do we still need this? If we have this, can we avoid passing state into the session?

core/service.go
noot: there's a test that needs to pass in the NewBlocks channel so that it can pass in a message (core TestAnnounceBlock)

core/service.go
noot: core needs the StorageState for other reasons, and I think it also needs blkRec still so that it can propagate blocks to p2p. otherwise, it would need to poll the state

Issue #530, opened 2020-01-17
Move Shared Test Helpers to Package


ansermino:
Currently, we duplicate some code between packages. We should refactor this so that all these are within their own package, and imported to the tests that need them.

Some examples:
- Fetching test wasm blobs (eg. `runtime/wasmer_test.go`)
- `Exists(file)`
- `newRuntime()`/`newTestRuntime()`
https://github.com/ChainSafe/gossamer/pull/503#discussion_r368065174

thomasmodeneis:
done

PR #558, opened 2020-01-22
consensus/babe, core, cmd/gossamer, runtime, runtime/allocator - Move Shared Test Helpers to Package #530

thomasmodeneis:
## Changes
- update wasmer_test.go assertions and funcs
- create test package for re-usable wasm funcs and structs and update consensus/babe, core, cmd/gossamer, runtime, runtime/allocator

## Tests:
```

```

### Issues: #530


thomasmodeneis:
> This is a nice improvement! Looks good to me. You added [this item](https://github.com/ChainSafe/gossamer/pull/503#discussion_r368065174) to the issue from your previous pull request and did not address it here. Should this be addressed here?

I looked into it, but its not clear yet if the suggestion will actually be a desired improvement. 

This is the piece of code you would like to add to a func and re-use right ?
```
//Close DB
	defer func() {
		err = blockState.db.Db.Close()
		require.Nil(t, err, "BlockDB close err: ", err)
	}()
```

How do you plan to make this re-usable and how its going to help on the testcase itself ? having in mind that is a 4 lines code already.

PR #570, opened 2020-01-24
cmd, config: update genesis fields to raw data array

### LINKED ISSUE #552: error="json: cannot unmarshal array into Go struct field"
## Expected Behavior

Gossamer should be able to parse [the latest kusama genesis file](https://raw.githubusercontent.com/paritytech/polkadot/master/service/res/kusama.json).

## Current Behavior

The `raw` field within the `genesis` field is now an array/slice of type map[string]string.

```
EROR[01-19|18:02:32] error loading genesis state error="json: cannot unmarshal array into Go struct field GenesisFields.Genesis.Raw of type map[string]string"
```

ryanchristo:
## Changes

This pull request update genesis fields to raw data array fixing #552. This also updates our default genesis file (soon to be gossamer testnet) to use the same format.

## Tests:

should get "error loading state and runtime" instead of #552 
```
make gossamer
./build/bin/gossamer init --genesis config/ksmcc3.json
./build/bin/gossamer
```

### Issues:

closes #552


config/genesis/genesis.go
```diff
@@ -17,17 +17,17 @@ type Genesis struct {
...
+	Name       string
+	Id         string
+	Bootnodes  [][]byte
+	ProtocolId string
 }
 
+// GenesisFields stores genesis raw data
 type GenesisFields struct {
-	Raw map[string]string
+	Raw [2]map[string]string
```
noot: is this necessarily 2, or can it just be a slice?

ryanchristo: I originally had a slice but I changed it to `2`. I am not quite sure how it will be used but it looks like the two mappings break down to`top` and `children` in substrate ([see substrate code](https://github.com/paritytech/substrate/blob/b443ddad65f2ffed68db4973b7b78231c6fc3674/client/chain-spec/src/chain_spec.rs#L77)).

noot: okay cool, makes sense to me

cmd/gossamer/genesis.go
```diff
@@ -20,20 +19,22 @@ func loadGenesis(ctx *cli.Context) error {
...
 	dataDir := fig.Global.DataDir
 	if ctx.String(utils.DataDirFlag.Name) != "" {
 		dataDir = ctx.String(utils.DataDirFlag.Name)
 	}
+
 	log.Debug("Loading genesis", "genesisfile", fp, "datadir", dataDir)
 
-	gen, err := genesis.LoadGenesisData(fp)
+	// read genesis configuration file
+	gen, err := genesis.LoadGenesisJsonFile(fp)
```
thomasmodeneis: should be JSON, so LoadGenesisJSONFile

config/genesis/genesis.go
```diff
@@ -17,17 +17,17 @@ type Genesis struct {
...
-// Genesis stores the data parsed from the genesis configuration file
+// GenesisData defines the genesis file data formatted for trie storage
 type GenesisData struct {
-	Name          string
-	Id            string
-	Bootnodes     [][]byte
-	ProtocolId    string
-	genesisFields GenesisFields
+	Name       string
+	Id         string
```
thomasmodeneis: should be ID and ProtocolID

config/genesis/genesis_test.go
```diff
@@ -9,22 +10,30 @@ import (
...
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer os.Remove(file.Name())
+
 	expected := &Genesis{
 		Name:       "gossamer",
 		Id:         "gossamer",
 		Bootnodes:  []string{"/ip4/104.211.54.233/tcp/30363/p2p/16Uiu2HAmFWPUx45xYYeCpAryQbvU3dY8PWGdMwS2tLm1dB1CsmCj"},
 		ProtocolId: "gossamer",
```
thomasmodeneis: ID and ProtocolID here as well.

config/genesis/genesis.go
```diff
@@ -47,21 +47,17 @@ func LoadGenesisJsonFile(file string) (*Genesis, error) {
...
+func (g *Genesis) GenesisData() *GenesisData {
 	return &GenesisData{
-		Name:          g.Name,
-		Id:            g.Id,
-		Bootnodes:     common.StringArrayToBytes(g.Bootnodes),
-		ProtocolId:    g.ProtocolId,
-		genesisFields: g.Genesis,
-	}, nil
+		Name:       g.Name,
+		Id:         g.Id,
```
thomasmodeneis: ID and ProtocolID

config/genesis/genesis.go
ryanchristo: this is already addressed in #555

config/genesis/genesis_test.go
ryanchristo: this is already addressed in #555 and this code was not modified

config/genesis/genesis.go
ryanchristo: this is already addressed in #555

ryanchristo:
cc @thomasmodeneis can you please re-review when you get a chance?

If I change `Id` and `ProtocolId` where suggested, I would be required to change it everywhere, which is already addressed in #555 and therefore will be addressed here if #555 is merged first

config/genesis/genesis.go
thomasmodeneis: ok

config/genesis/genesis.go
thomasmodeneis: ok

Issue #575, opened 2020-01-28
BABE: retrieve and set authority data 


noot:
right now, the authority data consists of only ourselves. we need to retrieve and set the authority data before each epoch using the `Grandpa_authorities` runtime call.

also, after each block, we may need to update the authority data based on the results of the block's Consensus digest. see spec 5.1.2

noot:
code relevant for consensus digest format:

https://github.com/paritytech/substrate/blob/8974349874588de655b7350737bba45032bb2548/frame/babe/src/tests.rs#L112

https://github.com/paritytech/substrate/blob/6e0a4877afbe2309e106b84333bb70200531b604/primitives/consensus/babe/src/digest.rs#L199

noot:
https://github.com/paritytech/substrate/blob/7bd0dbf39e4605da1d24b0f1a4ec77c843cd49d8/client/consensus/babe/src/lib.rs#L987

PR #576, opened 2020-01-28
consensus/babe, core: initialize new babe session for each epoch

### LINKED ISSUE #565: Babe Session Initialization
Inside of `core` we need to create new session on startup (possibly after sync is complete), and continue creating sessions after it completes.

noot:
# Changes
- add done channel between babe and core, when epoch complete, done is signalled and core creates the next epoch's babe session
- slightly modify build block logic to conform to spec (we are supposed to pass in a new header to initialize_block, not the parent header)

## Tests:
```
go test ./consensus/...
```

running a 1-node block producer:
```
make gossamer
./build/bin/gossamer init
./build/bin/gossamer --verbosity debug
```

### Issues:
closes #565

### DIFF
--- a/consensus/babe/babe.go
+++ b/consensus/babe/babe.go
@@ -26,6 +26,7 @@ import (
 	"time"
 
 	scale "github.com/ChainSafe/gossamer/codec"
+	"github.com/ChainSafe/gossamer/common"
 	tx "github.com/ChainSafe/gossamer/common/transaction"
 	babetypes "github.com/ChainSafe/gossamer/consensus/babe/types"
 	"github.com/ChainSafe/gossamer/core/types"
@@ -46,6 +47,7 @@ type Session struct {
 	txQueue        *tx.PriorityQueue
 	slotToProof    map[uint64]*VrfOutputAndProof // for slots where we are a producer, store the vrf output (bytes 0-32) + proof (bytes 32-96)
 	newBlocks      chan<- types.Block            // send blocks to core service
+	done           chan<- struct{}               // lets core know when the epoch is done
 }
 
 type SessionConfig struct {
@@ -56,6 +58,7 @@ type SessionConfig struct {
 	AuthorityIndex uint64
 	AuthData       []*AuthorityData
 	EpochThreshold *big.Int // should only be used for testing
+	Done           chan<- struct{}
 }
 
 // NewSession returns a new Babe session using the provided VRF keys and runtime
@@ -74,6 +77,7 @@ func NewSession(cfg *SessionConfig) (*Session, error) {
 		authorityIndex: cfg.AuthorityIndex,
 		authorityData:  cfg.AuthData,
 		epochThreshold: cfg.EpochThreshold,
+		done:           cfg.Done,
 	}
 
 	err := babeSession.configurationFromRuntime()
@@ -148,6 +152,7 @@ func (b *Session) invokeBlockAuthoring() {
 			} else {
 				hash := block.Header.Hash()
 				log.Info("BABE", "built block", hash.String(), "number", block.Header.Number)
+				log.Debug("BABE built block", "header", block.Header, "body", block.Body)
 				b.newBlocks <- *block
 				err = b.blockState.AddBlock(*block)
 				if err != nil {
@@ -157,6 +162,14 @@ func (b *Session) invokeBlockAuthoring() {
 		}
 
 		time.Sleep(time.Millisecond * time.Duration(b.config.SlotDuration))
+	}
+
+	if b.newBlocks != nil {
+		close(b.newBlocks)
+	}
+
+	if b.done != nil {
+		close(b.done)
 	}
 }
 
@@ -271,8 +284,14 @@ func (b *Session) buildBlock(parent *types.Header, slot Slot) (*types.Block, err
 		return nil, err
 	}
 
-	// initialize block
-	encodedHeader, err := scale.Encode(parent)
+	// create new block header
+	header, err := types.NewHeader(parent.Hash(), big.NewInt(0).Add(parent.Number, big.NewInt(1)), common.Hash{}, common.Hash{}, [][]byte{})
+	if err != nil {
+		return nil, err
+	}
+
+	// initialize block header
+	encodedHeader, err := scale.Encode(header)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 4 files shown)


consensus/babe/babe.go
```diff
@@ -46,6 +47,7 @@ type Session struct {
 	txQueue        *tx.PriorityQueue
 	slotToProof    map[uint64]*VrfOutputAndProof // for slots where we are a producer, store the vrf output (bytes 0-32) + proof (bytes 32-96)
 	newBlocks      chan<- types.Block            // send blocks to core service
+	done           chan<- struct{}               // lets core know when the epoch is done
```
ryanchristo: Should this be it's own type? Why an empty struct?

consensus/babe/babe.go
```diff
@@ -148,6 +152,7 @@ func (b *Session) invokeBlockAuthoring() {
 			} else {
 				hash := block.Header.Hash()
 				log.Info("BABE", "built block", hash.String(), "number", block.Header.Number)
+				log.Debug("BABE built block", "header", block.Header, "body", block.Body)
```
ryanchristo: Seems a bit excessive to have a log info and log debug stacked like this... Should we consolidate into one?

consensus/babe/babe.go
```diff
@@ -271,8 +284,14 @@ func (b *Session) buildBlock(parent *types.Header, slot Slot) (*types.Block, err
 		return nil, err
 	}
 
-	// initialize block
-	encodedHeader, err := scale.Encode(parent)
+	// create new block header
+	header, err := types.NewHeader(parent.Hash(), big.NewInt(0).Add(parent.Number, big.NewInt(1)), common.Hash{}, common.Hash{}, [][]byte{})
```
ryanchristo: I feel like we should break this up a bit for readability.

consensus/babe/babe_test.go
```diff
@@ -544,7 +544,7 @@ func TestBuildBlock_ok(t *testing.T) {
 	block, slot := createTestBlock(babesession, t)
 
 	// hash of parent header
-	parentHash, err := common.HexToHash("0x03106e6f6f740140676f7373616d65725f69735f636f6f6c6c00000000000000")
+	parentHash, err := common.HexToHash("0xdcdd89927d8a348e00257e1ecc8617f45edb5118efff3ea2f9961b2ad9b7690a")
```
ryanchristo: Should this be a constant we can reuse for all tests? What is different about the new hash?

core/service.go
```diff
@@ -151,18 +162,59 @@ func (s *Service) StorageRoot() (common.Hash, error) {
 	return s.storageState.StorageRoot()
 }
 
+func (s *Service) handleBabeSession() {
+	for {
+		<-s.epochDone
+		log.Trace("CORE: BABE epoch complete, initializing new session")
```
ryanchristo: I like the specification of the package here in the logs. I feel like we have discussed this and we should be doing this everywhere. I am not sure about the all caps `CORE` for formatting though and formatting in general for this log looks a bit out of place alongside the other logs.

core/service.go
```diff
@@ -151,18 +162,59 @@ func (s *Service) StorageRoot() (common.Hash, error) {
...
+		epochDone := make(chan struct{})
+		s.epochDone = epochDone
+
+		// BABE session configuration
+		bsConfig := &babe.SessionConfig{
+			Keypair:        s.keys[0].(*sr25519.Keypair),
+			Runtime:        s.rt,
+			NewBlocks:      newBlocks, // becomes block send channel in BABE session
+			BlockState:     s.blockState,
+			AuthorityIndex: 0, // TODO: where do we get the BABE authority data?
```
ryanchristo: issue number?

core/service.go
```diff
@@ -151,18 +162,59 @@ func (s *Service) StorageRoot() (common.Hash, error) {
...
+			AuthData:       []*babe.AuthorityData{babe.NewAuthorityData(s.keys[0].Public().(*sr25519.PublicKey), 1)},
+			EpochThreshold: big.NewInt(0),
+			Done:           epochDone,
+		}
+
+		// create a new BABE session
+		bs, err := babe.NewSession(bsConfig)
+		if err != nil {
+			log.Error("CORE could not initialize BABE", "error", err)
```
ryanchristo: Again, not sure `CORE` should be capitalized as it is not an acronym like BABE.

consensus/babe/babe.go
noot: I don't need to see the full block if I'm at info, but I wanted to see it at debug.. not sure how to consolidate

consensus/babe/babe.go
noot: we don't need to send anything over the channel, just signal. empty struct is generally used in Go for this purpose since it's the smallest type we can have: sizeof(struct{}) == 0 since nothing is actually allocated)

consensus/babe/babe.go
noot: which part? the block number?

consensus/babe/babe.go
ryanchristo: That makes sense. It's fine as is, just noticed this has not been done before.

consensus/babe/babe_test.go
noot: it's not used in any other tests right now, before we were calculating the new block header incorrectly :p this is the actual hash of the parent block

consensus/babe/babe.go
ryanchristo: sweet, thanks for explaining

core/service.go
noot: yeah I'm not totally sure how to format logs like this either, in substrate the packages names are capitalized in the logs so I went off that, but I'm fine with changing it

core/service.go
noot: https://github.com/ChainSafe/gossamer/issues/575

consensus/babe/babe.go
ryanchristo: yes, block number would be nice to extract

core/service.go
noot: can change if you would like, maybe we could discuss later what we should do for consistency?

ryanchristo: I am ok with it as is if this is the format we want to adopt but maybe we should double check and agree on a format. `CORE` uppercase seems odd to me but might just be me.

consensus/babe/babe.go
noot: done

core/service.go
noot: I've changed it to lowercase for now, I'm personally fine either way as long as it's consistent

ryanchristo: me too

PR #577, opened 2020-01-28
trie: fix failing trie tests

### LINKED ISSUE #201: Trie tests occasionally fail on CI
## Expected Behavior
Trie tests should not fail

## Current Behavior
It seems that sometimes the tests fail and `TestFailingTests` gets run, which is supposed to run the same test case as the test that failed in `TestPutAndGet`.  However, it doesn't run the same test case.

## Possible Solution
There is a bug in TestFailingTests. Maybe it is not running the correct test cases.

`go test ./trie/ -v`

It will need to be run multiple times until the fail happens.  Since the tests are random, we don't really know what case causes it to fail, or why it appears that the wrong test case gets run.

### LINKED ISSUE #547: Refactor Trie Tests
Currently there are a lot statements that are printed to the log for the trie tests.

Refactor the subtests into groups to avoid log pollution.

ryanchristo:
## Changes

The random number generator was producing an empty value and when that test was being  written to the failing trie tests, the empty value would cause all tests to be offset and therefore we would received a very large number of errors caused by a single generated test with an empty value.

This pull request also reduces the number of delete tests generated to `100`.

## Tests:
```
go test ./trie -v
```

### Issues:
closes #201, related to #547

### DIFF
--- a/trie/trie_test.go
+++ b/trie/trie_test.go
@@ -147,7 +147,8 @@ func generateRandomTest(kv map[string][]byte) trieTest {
 	test := trieTest{}
 
 	for {
-		size := r.Intn(510) + 2
+		n := 2 // arbitrary positive number
+		size := r.Intn(510) + n
 		buf := make([]byte, size)
 		r.Read(buf)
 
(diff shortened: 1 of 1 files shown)


noot:
idk if this closes #547 but trie tests seem good now!!! thank you so much

ryanchristo:
cc @thomasmodeneis sure, only two are related but I have updated

PR #585, opened 2020-01-30
core, p2p: network state and host status message

### LINKED ISSUE #443: Generate status message

### LINKED ISSUE #488: Implement Network State
## Expected Behavior

The core service needs access to state to generate the host status message. The rpc and runtime need access to network information available in the core service and/or the p2p service.

We need to implement a state interface in the core service and/or the p2p service and migrate the rpc servcer from the api service to the state service.

ryanchristo:
## Changes

This pull request add state interface to p2p service and updates host status message on connect.

## Tests:
```
go test ./p2p -v
```

### Issues:
closes #488, closes #443


p2p/service.go
```diff
@@ -133,36 +133,39 @@ func (s *Service) receiveCoreMessages() {
...
+			Roles:               s.cfg.Roles,
+			BestBlockNumber:     latestBlock.Number.Uint64(),
+			BestBlockHash:       latestBlock.Hash(),
+			GenesisHash:         latestBlock.StateRoot,
+			ChainStatus:         []byte{0}, // TODO
+		}
+
+		// update host status message
+		s.status.setHostMessage(msg)
+
```
ryanchristo: Does it make sense to load a fresh status message here on connect?

p2p/config.go
```diff
@@ -89,47 +160,51 @@ func (c *Config) build() error {
...
 	}
 
-	// set private key
-	c.privateKey = key
+	if c.Roles == 0 {
+		log.Warn("Roles 0 is not supported, using DefaultRoles",
+			"DefaultRoles", DefaultRoles,
+		)
+		c.Roles = DefaultRoles
+	}
```
ryanchristo: Roles 0 is no network according to the specification. I'm assuming no need to support it in the network service.

common/types.go
```diff
@@ -16,35 +16,26 @@
...
-	copy(res[:], in)
-	return res
-}
-
-// ToBytes turns a hash to a byte array
-func (h Hash) ToBytes() []byte {
-	b := [32]byte(h)
-	return b[:]
+// Health is network information about host needed for the rpc server
+type Health struct {
```
noot: wondering why this is moved here? is there an import cycle or something?

p2p/config.go
```diff
@@ -17,59 +17,130 @@
...
-// KeyFile key
-const KeyFile = "node.key"
-
-// DefaultProtocolID ID
+const DefaultKeyFile = "node.key"
+const DefaultDataDir = "~/.gossamer"
+const DefaultPort = 7000
+const DefaultRandSeed = 0 // random key
 const DefaultProtocolID = "/gossamer/dot/0"
+const DefaultRoles = 1 // fulle node
```
noot: full

p2p/config.go
noot: i think if roles is 0, we don't turn on networking at all. maybe we should be checking it before starting up the p2p service? is roles something we can configure in the config?

p2p/service.go
```diff
@@ -133,36 +133,39 @@ func (s *Service) receiveCoreMessages() {
...
 }
 
 // handleConn starts processes that manage the connection
 func (s *Service) handleConn(conn network.Conn) {
-
 	// check if status is enabled
 	if !s.noStatus {
 
+		// get latest block header from block state
+		latestBlock := s.cfg.BlockState.LatestHeader()
```
noot: would it be preferred to create a field in the service itself for BlockState and StorageState so we don't need to keep the whole config?

p2p/service.go
noot: I think only reloading it upon connecting makes sense, however, there's a pretty high chance we might connect to multiple peers without our state ever changing. maybe there should be a way to check if the status message should change from previously (eg. if there's a new block), then if so, update it

noot: we could open another issue for that, since this PR is already  somewhat large

p2p/state.go
```diff
@@ -0,0 +1,68 @@
...
+	StorageRoot() (common.Hash, error)
+}
+
+type MockBlockState struct{}
+
+func (mbs *MockBlockState) AddBlock(types.Block) error {
+	return nil
+}
+
+func (mbs *MockBlockState) LatestHeader() *types.Header {
```
noot: maybe put the MockBlockState in the tests?

state/network.go
```diff
@@ -1,29 +1,21 @@
...
-		p2p: &p2p.Service{},
-	}
+	return &networkState{}
 }
 
-func (ns *networkState) Health() p2p.Health {
-	// TODO: return Health() of p2p service
-	return p2p.Health{}
+func (ns *networkState) Health() common.Health {
+	return common.Health{}
```
noot: leave the TODOs?

cmd/gossamer/configcmd_test.go
```diff
@@ -197,10 +197,12 @@ func TestSetGlobalConfig(t *testing.T) {
...
 func TestCreateP2PService(t *testing.T) {
-	srv, _, _ := createP2PService(cfg.DefaultConfig(), &genesis.GenesisData{})
-	if srv == nil {
-		t.Fatalf("failed to create p2p service")
+	stateSrv := state.NewService(TestDataDir)
+	gendata := &genesis.GenesisData{
+		ProtocolID: TestProtocolID,
 	}
+	srv, _, _ := createP2PService(cfg.DefaultConfig(), gendata, stateSrv)
+	require.NotNil(t, srv, "failed to create p2p service")
```
thomasmodeneis: cool, using require! :+1:

p2p/config_test.go
```diff
@@ -77,18 +81,20 @@ func TestBuildConfig(t *testing.T) {
...
+		RandSeed:     DefaultRandSeed,
+		Bootnodes:    DefaultBootnodes,
+		ProtocolID:   DefaultProtocolID,
+		NoBootstrap:  false,
+		NoMdns:       false,
+		privateKey:   testKey,
 	}
 
-	if reflect.DeepEqual(cfg, testCfg) {
+	if reflect.DeepEqual(cfg, expected) {
```
thomasmodeneis: you can just use require.Equal here as well if you want.

p2p/state.go
thomasmodeneis: yes, good idea, prod code mixed with mocks is not a good practice.

state/network.go
thomasmodeneis: perhaps open issues with the todos and refer it in the TODO would also be a good idea...

common/types.go
ryanchristo: correct, there was an import cycle with network and state

p2p/gossip_test.go
```diff
@@ -17,15 +17,20 @@
...
+	"os"
+	"path"
 	"testing"
 	"time"
 )
 
 // test gossip messages to connected peers
 func TestGossip(t *testing.T) {
+	dataDirA := path.Join(os.TempDir(), "gossamer-test", "nodeA")
+	defer os.RemoveAll(dataDirA)
```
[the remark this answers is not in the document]
ryanchristo: A temp directory created with ioutil still needs to be cleaned according to the docs.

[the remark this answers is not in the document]
thomasmodeneis: well, temp dir will be cleaned once computer restarts. Standard practice is to leave it there in case you want to backtrace to what happened.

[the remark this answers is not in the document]
ryanchristo: I think we need to leave it as is because we are initializing a new node each time.

p2p/service.go
ryanchristo: I feel like I did this at some point and then reverted but I don't remember why. Will check again.

p2p/service.go
ryanchristo: Yea, I guess it might not need to be updated so frequently. I will open an issue to investigate and address. Might also be nice to add a test or make note when we eventual write and run network simulations.

p2p/state.go
ryanchristo: will do

state/network.go
ryanchristo: this is no longer correct, we do not want (and are not able) to import the p2p service here in state

ryanchristo: I should have addressed this here, I will update

p2p/config.go
ryanchristo: Yea, probably should be checked before starting network service. Will fix.

p2p/service.go
ryanchristo: So we already need several items from the config (`ProtocolVersion`, `MinSupportedVersion`, `Roles`) in addition to needing `BlockState`. Should I move them all to the service struct?

cmd/gossamer/configcmd.go
```diff
@@ -85,10 +88,10 @@ func makeNode(ctx *cli.Context) (*dot.Dot, *cfg.Config, error) {
 		return nil, nil, err
 	}
 
-	log.Info("🕸\t Configuring node...", "datadir", currentConfig.Global.DataDir, "protocol", currentConfig.P2p.ProtocolID, "bootnodes", currentConfig.P2p.Bootnodes)
+	// TODO: check Roles to see if we need p2p service (Roles 0 = no network)
```
ryanchristo: #601

p2p/config.go
ryanchristo: #601

p2p/service.go
ryanchristo: #602

ryanchristo:
cc @noot @thomasmodeneis when you get a chance to rereview

cmd/gossamer/configcmd.go
thomasmodeneis: would be better if you could add the raised ticket to the TODO, so it allows good link between the code and the reported issue:
```
// TODO: check Roles to see if we need p2p service (Roles 0 = no network) #601 
```

p2p/gossip_test.go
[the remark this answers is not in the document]
thomasmodeneis: right, ok.

p2p/service.go
noot: i see, in that case this makes sense

PR #587, opened 2020-01-31
scale: allocate slice elements when decoding

### LINKED ISSUE #584: scale: properly handle arrays when decoding
currently, scale cannot create new fields of an array when decoding, it can only assign to the object that's passed in. however, we don't always know how many items are in the array before decoding it. scale should be able to handle this by creating new fields.
eg. bootnodes, digests

noot:
## Changes
- update DecodeArray to allocate a slice instead of assigning to the interface that's passed in, allows for arrays to be decoded without knowing how many elements are in it
- update DecodeTuple to handle [][]byte

## Tests:
```
go test ./codec/...
```

### Issues:
closes #584

### DIFF
--- a/consensus/babe/runtime.go
+++ b/consensus/babe/runtime.go
@@ -64,11 +64,12 @@ func (b *Session) finalizeBlock() (*types.Block, error) {
 		return nil, err
 	}
 
-	bh := &types.Block{
-		Header: new(types.Header),
-		Body:   new(types.Body),
-	}
-
+	// TODO: finalize block actually returns a header, not a block. need to update this function
+	// as well as buildBlock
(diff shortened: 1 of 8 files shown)


consensus/babe/runtime.go
```diff
@@ -64,11 +64,12 @@ func (b *Session) finalizeBlock() (*types.Block, error) {
 		return nil, err
 	}
 
-	bh := &types.Block{
-		Header: new(types.Header),
-		Body:   new(types.Body),
-	}
-
+	// TODO: finalize block actually returns a header, not a block. need to update this function
+	// as well as buildBlock
```
noot: #588

Issue #563, opened 2020-01-23
Choose Polkadot commit and update to it


ansermino:
We need to select a commit of Polkadot to target moving forward. We should establish what updates to the runtime we need to make (and possibly elsewhere). 

We will need a genesis file to match. Doesn't necessarily need to match Kusama if we are testing on a local net.

noot:
when we update the runtime, will need to change the AuraApi_authorites call to Grandpa_authorities

PR #595, opened 2020-02-06
consensus/babe, core: read BABE authority updates from block consensus digests

### LINKED ISSUE #575: BABE: retrieve and set authority data

noot:
# Changes
- add NextEpochDescriptor which is what's put in the consensus digest at the first block of each epoch
- add helper functions for AuthorityData
- add Grandpa_authorities runtime call to retrieve authorities initially
- add handleConsensusDigest function which is called after a block is imported to see if there is a NextEpochDescriptor

## Tests:
```
go test ./consensus/... ./core/...
```

### Issues:
closes #575


core/service.go
```diff
@@ -88,14 +88,34 @@ func NewService(cfg *Config) (*Service, error) {
...
+		epochDone:    epochDone,
+	}
+
+	authData, err := srv.retrieveAuthorityData()
+	if err != nil {
+		return nil, err
+	}
+
+	// TODO: our authority index should be in authData, if it isn't, we aren't authorities
+	// need to add our authority data to storage
```
ryanchristo: Do we have an issue open to address this?

core/service.go
```diff
@@ -164,6 +176,11 @@ func (s *Service) StorageRoot() (common.Hash, error) {
 	return s.storageState.StorageRoot()
 }
 
+func (s *Service) retrieveAuthorityData() ([]*babe.AuthorityData, error) {
+	// TODO: when we update to a new runtime, will need to pass in the latest block number
```
ryanchristo: Looks like we should link these to an issue and update that runtime.

core/service.go
```diff
@@ -377,3 +406,36 @@ func (s *Service) ProcessTransactionMessage(msg p2p.Message) error {
...
+		if err != nil {
+			return err
+		}
+
+		if item.Type() == types.ConsensusDigestType {
+			break
+		}
+	}
+
+	// TODO: if this block is the first in the epoch and it doesn't have a consensus digest, this is an error
```
ryanchristo: Why not return an error? Not sure I understand the todo item here.

consensus/babe/types.go
```diff
@@ -64,3 +122,49 @@ type Slot struct {
 	duration uint64
 	number   uint64
 }
+
+// NextEpochDescriptor contains information about the next epoch.
+// It is broadcast as part of the consensus digest in the first block of the epoch.
+type NextEpochDescriptor struct {
+	Authorities []*AuthorityData
+	Randomness  [sr25519.VrfOutputLength]byte // TODO: discrepancy between current BabeConfiguration from runtime and this
```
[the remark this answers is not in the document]
noot: it's due to the discrepency between our runtime and the current one, our runtime has randomness as a byte but NextEpochDescriptor (and the updated runtime) have it as a byte array

core/runtime.go
```diff
@@ -54,3 +54,16 @@ func (s *Service) executeBlock(b []byte) error {
 
 	return nil
 }
+
+// TODO: this seems to be out-of-date, the call is now named Grandpa_authorities and takes a block number.
```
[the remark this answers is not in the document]
noot: I'll post a comment in the update runtime issue, since we haven't updated yet not sure if it warrants its own issue

consensus/babe/types.go
```diff
@@ -28,7 +31,7 @@ type BabeConfiguration struct {
 	C1                 uint64 // (1-(c1/c2)) is the probability of a slot being empty
 	C2                 uint64
 	GenesisAuthorities []AuthorityDataRaw
-	Randomness         byte
+	Randomness         byte // TODO: change to [VrfOutputLength]byte when updating to new runtime
```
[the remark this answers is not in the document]
noot: this is more in general when we update it

core/service.go
noot: I'll open one

noot: #598

core/service.go
noot: we don't have a way to check what slot in the epoch we're at from core, I can open an issue for it

core/service.go
noot: will add to comment

noot: https://github.com/ChainSafe/gossamer/issues/563 should cover everything

noot:
@edwardmack it seems like the main change from the runtime we have to the current runtime is the length of the randomness, otherwise, it appears to be the same

core/service.go
noot: #599

PR #597, opened 2020-02-06
codec; Enable custom scale methods

### LINKED ISSUE #525: Enable Custom Scale Methods
Due to inconsistencies between how Rust and Go handle types, SCALE sometimes requires custom encoding definitions. It would be quite simple for us to enable types to have a custom encode/decode method defined, while using the standard encode/decode for all types that do not.

For every type we encode/decode, we can use the `reflect` package to check if a method with the correct name exists:
```
someType := reflect.TypeOf(interface)
_, ok := someType.MethodByName("EncodeScale")
if ok {
    return someType.EncodeScale()
} else {
    return Encode(someType)
}
```

edwardmack:
This PR introduces functionality for handling custom Encode/Decode function for types when these functions have been defined.

## Changes
- EncodeCustom function added which checks if interface has Encode function defined, if so it uses that for encoding, if not it uses the standard scale encoding.
- DecodeCustom function added which checks if interface has Decode function defined, if so it uses that for decoding, if not it uses the standard  scale decoding.
- Added test cases for testing this.

## Tests:
```
 go test ./codec/... 
```

### Issues:
closes #525


go.mod
```diff
@@ -11,8 +11,8 @@ require (
 	github.com/disiqueira/gotree v1.0.0
 	github.com/ethereum/go-ethereum v1.9.6
 	github.com/filecoin-project/go-leb128 v0.0.0-20190212224330-8d79a5489543
-	github.com/golang/protobuf v1.3.2 // indirect
```
thomasmodeneis: I have raised https://github.com/ChainSafe/gossamer/wiki/GoMOD#pr-597 to track the gomod changes for this PR. Lets try to keep it updated for the next changes.

PR #603, opened 2020-02-09
consensus/babe: load BABE authorities from genesis

### LINKED ISSUE #598: babe, gossamer testnet: set authorities in genesis file
add our public keys to here: https://github.com/paritytech/substrate/blob/ae9f8d04dcf39ded1e0f3ddb6477cd02c37f8621/node/cli/src/chain_spec.rs#L59

compile to genesis json file with instructions here: https://github.com/paritytech/substrate/tree/ae9f8d04dcf39ded1e0f3ddb6477cd02c37f8621#trying-out-substrate-node

noot:
## Changes
- update DecodeCustom to work inside of structs
- fix decode of BabeConfiguration
- set genesis authority data when spawning babe session from core
- update gssmr0.json to have some test authority keys, maybe we should think about testnet keypairs soon? or a test keyring like substrate has

depends on #597 #595 

## Tests:
```
go test ./... -short
```

### Issues:
closes #598


codec/decode.go
```diff
@@ -544,10 +549,12 @@ func (sd *Decoder) DecodeTuple(t interface{}) (interface{}, error) {
 				ptr := fieldValue.(*[]string)
 				*ptr = o.([]string)
 			default:
-				_, err = sd.Decode(v.Field(i).Interface())
+				o, err = sd.Decode(v.Field(i).Interface())
 				if err != nil {
 					break
 				}
+				// TODO: clean up this function, can use field.Set everywhere (remove switch case?)
```
ryanchristo: Did you mean to address in this pull request or is there an issue open for this?

consensus/babe/types.go
```diff
@@ -148,13 +169,19 @@ func (n *NextEpochDescriptor) Encode() []byte {
 }
 
 // Decode sets the NextEpochDescriptor to the SCALE decoded input.
+// TODO: change to io.Reader
```
ryanchristo: Can this be addressed here or should we open a separate issue?

codec/decode.go
noot: yes #611

consensus/babe/types.go
noot: #611

core/service.go
```diff
@@ -41,26 +40,28 @@ var _ services.Service = &Service{}
...
+	blockState    BlockState
+	storageState  StorageState
+	rt            *runtime.Runtime
+	bs            *babe.Session
+	keys          []crypto.Keypair
+	blkRec        <-chan types.Block // receive blocks from BABE session
+	msgRec        <-chan p2p.Message // receive messages from p2p service
+	epochDone     <-chan struct{}    // receive from this channel when BABE epoch changes
+	msgSend       chan<- p2p.Message // send messages to p2p service
+	babeAuthority bool
```
ryanchristo: Maybe `isAuthority` ?

noot: hmm, we might have an option for grandpaAuthority in the future, so I wanted to be a bit more specific. maybe `isBabeAuthority`?

ryanchristo: Gotcha. Yea, maybe `isBabeAuthority` to imply a bool...? Might not be necessary though.

thomasmodeneis: agree with the pattern is before the bool.

noot: will update

PR #604, opened 2020-02-09
update pending golint issues - #586

### LINKED ISSUE #586: Fix remaining issues reported by golint
After #559 is closed and #574 is merged, the following linter skip steps should be removed and a new PR should fix all the reported issues.

Special case for `"which can be annoying to use"`, this one may require a refactor on exported functions that are returning a non exported type `node`, the golint recommendation is to avoid doing it. The other special case `"and that stutters; consider calling this"` would also require special attention and decision.

```   
    - linters:
        - golint
      text: "which can be annoying to use"

    - linters:
        - golint
      text: "and that stutters; consider calling this"

    - linters:
        - golint
      text: "should have comment or be unexported"

    - linters:
        - golint
      text: "or a comment on this block"

    - linters:
        - golint
      text: "error strings should not be capitalized or end with punctuation or a newline"
```

1.
2.
3.
4.

## Your Environment
* Version used:
* Commit hash:
* Go Version:

thomasmodeneis:
## Changes
* Removed lint ingnores:
- should have comment or be unexported
- or a comment on this block
- error strings should not be capitalized or end with punctuation or a newline"
-  stutter Warnings, warns for current names used by other packages:
```
config/genesis/genesis.go:21:6: type name will be used as genesis.GenesisData by other packages, and that stutters; consider calling this Data (golint)
type GenesisData struct {
     ^
config/genesis/genesis.go:29:6: type name will be used as genesis.GenesisFields by other packages, and that stutters; consider calling this Fields (golint)
type GenesisFields struct {
     ^
runtime/runtime.go:33:6: type name will be used as runtime.RuntimeCtx by other packages, and that stutters; consider calling this Ctx (golint)
type RuntimeCtx struct {
     ^
state/storage.go:13:6: type name will be used as state.StateDB by other packages, and that stutters; consider calling this DB (golint)
type StateDB struct {
     ^
consensus/babe/types.go:25:6: type name will be used as babe.BabeConfiguration by other packages, and that stutters; consider calling this Configuration (golint)
type BabeConfiguration struct {
```
- annoying to use lint issues:
```
core/blocktree/blocktree.go:97:38: exported method GetNode returns unexported type *github.com/ChainSafe/gossamer/core/blocktree.node, which can be annoying to use (golint)
func (bt *BlockTree) GetNode(h Hash) *node {
                                     ^
core/blocktree/blocktree.go:155:36: exported method LongestPath returns unexported type []*github.com/ChainSafe/gossamer/core/blocktree.node, which can be annoying to use (golint)
func (bt *BlockTree) LongestPath() []*node {
                                   ^
core/blocktree/blocktree.go:167:53: exported method SubChain returns unexported type []*github.com/ChainSafe/gossamer/core/blocktree.node, which can be annoying to use (golint)
func (bt *BlockTree) SubChain(start Hash, end Hash) []*node {
                                                    ^
core/blocktree/blocktree.go:187:36: exported method DeepestLeaf returns unexported type *github.com/ChainSafe/gossamer/core/blocktree.node, which can be annoying to use (golint)
func (bt *BlockTree) DeepestLeaf() *node {
                                   ^
trie/node.go:197:27: exported func Decode returns unexported type github.com/ChainSafe/gossamer/trie.node, which can be annoying to use (golint)
func Decode(r io.Reader) (node, error) {
                          ^
trie/trie.go:57:23: exported method Root returns unexported type github.com/ChainSafe/gossamer/trie.node, which can be annoying to use (golint)
func (t *Trie) Root() node {
                      ^
state/block.go:44:61: exported func NewBlockState returns unexported type *github.com/ChainSafe/gossamer/state.blockState, which can be annoying to use (golint)
func NewBlockState(dataDir string, latestHash common.Hash) (*blockState, error) {
                                                            ^
state/block.go:65:70: exported func NewBlockStateFromGenesis returns unexported type *github.com/ChainSafe/gossamer/state.blockState, which can be annoying to use (golint)
func NewBlockStateFromGenesis(dataDir string, header *types.Header) (*blockState, error) {
                                                                     ^
state/network.go:10:24: exported func NewNetworkState returns unexported type *github.com/ChainSafe/gossamer/state.networkState, which can be annoying to use (golint)
func NewNetworkState() *networkState {
```

## unused (unparam)
```
core/blocktree/blocktree_test.go:43:21: `createFlatTree` - `t` is unused (unparam)
func createFlatTree(t *testing.T, depth int) (*BlockTree, []common.Hash) {
                    ^
trie/hash_test.go:32:19: `generateRand` - `size` always receives `100` (unparam)
func generateRand(size int) [][]byte {
                  ^
trie/trie.go:136:61: (*Trie).insert - result ok is never used (unparam)
func (t *Trie) insert(parent Node, key []byte, value Node) (ok bool, n Node, err error) {
                                                            ^
trie/trie.go:338:49: (*Trie).delete - result ok is never used (unparam)
func (t *Trie) delete(parent Node, key []byte) (ok bool, n Node, err error) {
                                                ^
trie/trie.go:373:71: `handleDeletion` - result `err` is always `nil` (unparam)
func handleDeletion(p *branch, n Node, key []byte) (ok bool, nn Node, err error) {
                                                                      ^
core/service.go:255:59: `(*Service).handleReceivedBlock` - result `err` is always `nil` (unparam)
func (s *Service) handleReceivedBlock(block types.Block) (err error) {
                                                          ^
cmd/gossamer/configcmd.go:48:44: makeNode - result 1 (*github.com/ChainSafe/gossamer/config.Config) is never used (unparam)
func makeNode(ctx *cli.Context) (*dot.Dot, *cfg.Config, error) {
```

## Tests:
```

```

### Issues:

closes #586


core/blocktree/node.go
```diff
@@ -25,37 +25,37 @@ import (
 	"github.com/disiqueira/gotree"
 )
 
-// node is an element in the BlockTree
-type node struct {
+// Node is an element in the BlockTree
+type Node struct {
```
noot: I don't think this needs to be exported?

p2p/config.go
```diff
@@ -28,7 +28,8 @@ const KeyFile = "node.key"
 // DefaultProtocolID ID
 const DefaultProtocolID = "/gossamer/dot/0"
 
-var DefaultBootnodes = []string{}
+// DefaultBootnodes holds the default bootnodes the client will always have
+var DefaultBootnodes []string
```
noot: this sets DefaultBootnodes to nil whereas before it was an empty array, might cause an issue?

trie/node.go
```diff
@@ -48,7 +48,8 @@ import (
 	"github.com/ChainSafe/gossamer/common"
 )
 
-type node interface {
+// Node is the interface for trie methods
+type Node interface {
```
noot: this doesn't need to be exported

core/blocktree/node.go
ryanchristo: Yea, was wondering the same thing. I'm not seeing a difference running lint with it as it was before.

p2p/config.go
ryanchristo: https://github.com/ChainSafe/gossamer/pull/585/files#diff-ee95e4ec6d917a2cfefc71b49e21a771R35

core/blocktree/node.go
thomasmodeneis: The error is:
`trie/node.go:197:27: exported func Decode returns unexported type github.com/ChainSafe/gossamer/trie.node, which can be annoying to use (golint)`

And it makes sense, there are few exported methods that are returning a private struct. We should either make the struct public or the public methods private. The easiest was to make it public, so let me know what is your opinion.

trie/node.go
thomasmodeneis: the issue is that the type is being used as return value for public exported methods. We should either make it public or methods private.

core/blocktree/node.go
noot: we should make those functions on `node` unexported in that case, since we shouldn't be using them outside of the `trie` package

trie/node.go
noot: let's make the methods private

p2p/config.go
thomasmodeneis: good point, sorry for that. I updated with develop and used the value that was committed there.

core/blocktree/node.go
thomasmodeneis: right, sure!

trie/node.go
thomasmodeneis: roger

core/blocktree/node.go
thomasmodeneis: Well, I tried. But there are several fields with the same name, eg:
```
// Root returns the root of the trie
func (t *Trie) root() node {
	return t.root
}
```
and function decode for ex already exist unexported in the same package, so after renaming I cant proceed.
```
func decode(r io.Reader) (node, error) {
func decode(r io.Reader, prev node) error {
```

I understand the concern, the having the methods unexported will be a much harder refactor, are you sure about it ?

trie/trie.go
```diff
@@ -116,16 +119,20 @@ func (t *Trie) Put(key, value []byte) error {
...
 	} else {
-		_, n, err = t.delete(t.root, k)
+		ok, n, err = t.delete(t.root, k)
 	}
 
 	if err != nil {
 		return err
 	}
+	if !ok {
+		log.Warn("tryPut returned not ok for operation insert/delete")
```
noot: I don't think we need to be logging this, it'll probably clog up the logs too much. if it's !ok, it's not an issue, if there was an issue there would be an error

trie/trie.go
```diff
@@ -217,7 +224,9 @@ func (t *Trie) updateBranch(p *branch, key []byte, value node) (ok bool, n node,
 
 		switch c := p.children[key[length]].(type) {
 		case *branch, *leaf:
-			_, n, err = t.insert(c, key[length+1:], value)
+			if ok, n, err = t.insert(c, key[length+1:], value); !ok {
+				log.Warn("updateBranch returned not ok for operation 1 insert")
```
noot: same here

trie/trie.go
```diff
@@ -235,15 +244,19 @@ func (t *Trie) updateBranch(p *branch, key []byte, value node) (ok bool, n node,
 	br := &branch{key: key[:length], dirty: true}
 
 	parentIndex := p.key[length]
-	_, br.children[parentIndex], err = t.insert(nil, p.key[length+1:], p)
+	if ok, br.children[parentIndex], err = t.insert(nil, p.key[length+1:], p); !ok {
+		log.Warn("updateBranch returned not ok for operation 2 insert")
```
noot: same

trie/trie.go
```diff
@@ -235,15 +244,19 @@ func (t *Trie) updateBranch(p *branch, key []byte, value node) (ok bool, n node,
...
 	if err != nil {
 		return false, nil, err
 	}
 
 	if len(key) <= length {
 		br.value = value.(*leaf).value
 	} else {
-		_, br.children[key[length]], err = t.insert(nil, key[length+1:], value)
+		if ok, br.children[key[length]], err = t.insert(nil, key[length+1:], value); !ok {
+			log.Warn("updateBranch returned not ok for operation insert")
```
noot: same

trie/trie.go
thomasmodeneis: well, the issue is that the ok is never used anywhere and lint complained about it. So I made use of it by adding a log. The other option is to delete the instruction all together from the method and stop returning the ok since its not used for anything. What is the preferred option ?

trie/trie.go
thomasmodeneis: same as above

trie/trie.go
thomasmodeneis: same as above

trie/trie.go
noot: let's just remove `ok`, if it's not ok, then there is an error returned so it is redundant

thomasmodeneis: yep, oks

thomasmodeneis: I've raised https://github.com/ChainSafe/gossamer/issues/617 in order to have this fixed, since after few changes the risk and the nature of the fix will add extra logic that is not related to the ticket, possibly creating a scope creep here.
If you are ok I would like to merge like this and remove the ok in another ticket.

noot: sure, are you planning to work on #617 ?

consensus/babe/types.go
```diff
@@ -23,9 +23,9 @@ import (
 	"github.com/ChainSafe/gossamer/crypto/sr25519"
 )
 
-// BabeConfiguration contains the starting data needed for Babe
+// Configuration contains the starting data needed for Babe
 // see: https://github.com/paritytech/substrate/blob/426c26b8bddfcdbaf8d29f45b128e0864b57de1c/core/consensus/babe/primitives/src/lib.rs#L132
-type BabeConfiguration struct {
+type Configuration struct {
```
ryanchristo: Maybe `Config` to be consistent with the format elsewhere?

trie/trie.go
thomasmodeneis: yes

thomasmodeneis:
> I'm not sure I understand the "which can be annoying to use" linting rule that is preventing us from not exporting `node` within `blocktree`. Is it necessary? Can you explain?
Sure, the issue is that is quite annoying to have a exported method that returns a struct that you cant use. Try using the exported method from another package and you will see how annoying it will be. The rule is, if you are exporting a method, then the struct returned should also be exported, otherwise why are you exporting the method in the first place ?

> Also, running `make lint` updates `go.mod` and adds `bin/golint-cli`... Should these be included in the commit history or ignored? Should be one or the other though.
The update on go.mod is a issue that goes beyond the scope of this ticket, in the sense that the updates depend on the libs you have available on your system. We have a wiki page for tracking the changes, would you like to have the updated go.mod changes in the PR as well ?

consensus/babe/types.go
thomasmodeneis: Well, IDK, the naming is not of my creation. Maybe @noot could en-light on this one ?

noot: sure, `Config` is fine

thomasmodeneis: oks

PR #606, opened 2020-02-10
Request Blocks if needed after Status message exchange #581

### LINKED ISSUE #581: Request Blocks if needed after Status message exchange
This is essentially a fall-back mechanism, i.e. way to catch up when a node first joins the network or when a node falls behind.

thomasmodeneis:
## Changes
- Updated `ProcessBlockAnnounceMessage` func and adding condition to send a `BlockRequestMessage` when condition is true `if latestBlockNum >= messageBlockNum`
- Adding new test file `block_msg_test.go` that tests a `BlockRequestMsgType` and a `BlockAnnounceMsgType` is received based on different behaviors 

## Tests:
```

```

### Issues:

closes #581


core/service.go
```diff
@@ -295,21 +303,57 @@ func (s *Service) handleReceivedMessage(msg network.Message) (err error) {
...
+			ParentHash:     blockAnnounceMessage.ParentHash,
+			Number:         messageBlockNum,
+			StateRoot:      blockAnnounceMessage.StateRoot,
+			ExtrinsicsRoot: blockAnnounceMessage.ExtrinsicsRoot,
+			Digest:         blockAnnounceMessage.Digest,
+		}
+
+		// send blockAnnounce message to network service
+		log.Debug("send blockAnnounce message to network service")
+		s.msgSend <- blockAnnounce
```
noot: we shouldn't be sending a BlockAnnounce message to the network, we are the ones receiving BlockAnnounce messages from the network. you can remove this whole part and just leave the `else` section

core/service.go
```diff
@@ -295,21 +303,57 @@ func (s *Service) handleReceivedMessage(msg network.Message) (err error) {
...
+
+		//generate random ID
+		s1 := rand.NewSource(uint64(time.Now().UnixNano()))
+		seed := rand.New(s1).Uint64()
+		randomID := mrand.New(mrand.NewSource(int64(seed))).Uint64()
+
+		blockRequest := &network.BlockRequestMessage{
+			ID:            randomID, // random
+			RequestedData: 2,
+			StartingBlock: []byte{},
```
noot: starting block should be `byte(0)` + block hash or `byte(1)` + block Number

core/service.go
```diff
@@ -295,21 +303,57 @@ func (s *Service) handleReceivedMessage(msg network.Message) (err error) {
...
+			StateRoot:      blockAnnounceMessage.StateRoot,
+			ExtrinsicsRoot: blockAnnounceMessage.ExtrinsicsRoot,
+			Digest:         blockAnnounceMessage.Digest,
+		}
+
+		// send blockAnnounce message to network service
+		log.Debug("send blockAnnounce message to network service")
+		s.msgSend <- blockAnnounce
+
+	} else {
```
noot: can change this to `if messageBlockNum - latestBlockNum > 2` then we send a BlockRequest for the missing blocks

core/service.go
```diff
@@ -295,21 +303,57 @@ func (s *Service) handleReceivedMessage(msg network.Message) (err error) {
...
+		//generate random ID
+		s1 := rand.NewSource(uint64(time.Now().UnixNano()))
+		seed := rand.New(s1).Uint64()
+		randomID := mrand.New(mrand.NewSource(int64(seed))).Uint64()
+
+		blockRequest := &network.BlockRequestMessage{
+			ID:            randomID, // random
+			RequestedData: 2,
+			StartingBlock: []byte{},
+			EndBlockHash:  optional.NewHash(true, common.Hash{}),
```
noot: need to update this also, starting block to ending block should be the blocks we're missing

network/status.go
```diff
@@ -102,6 +102,10 @@ func (status *status) handleMessage(peer peer.ID, msg *StatusMessage) {
 		// update peer status message
 		status.peerMessage[peer] = msg
 
+		// check if message arrived is a block bigger then our latest one
+		// if so, fallback
```
noot: should be if >1 blocks farther

core/service.go
```diff
@@ -41,15 +46,16 @@ var _ services.Service = &Service{}
...
+	blockState        BlockState
+	storageState      StorageState
+	rt                *runtime.Runtime
+	bs                *babe.Session
+	keys              []crypto.Keypair
+	blkRec            <-chan types.Block     // receive blocks from BABE session
+	msgRec            <-chan network.Message // receive messages from network service
+	epochDone         <-chan struct{}        // receive from this channel when BABE epoch changes
+	msgSend           chan<- network.Message // send messages to network service
+	requestedBlockIDs map[uint64]bool        // track requested block id messages
```
ryanchristo: Should we import and store this in network state? Would it be useful within core?

ryanchristo: cc @noot This would be a reason to keep network state and I am now imaging other possible scenarios where we might want to use network state and have network state separate from other state.

core/service.go
ryanchristo: correct, no need to send a block announce... the network service will be the one receiving the announce message (and gossiping the message upon receiving it). The only time we send an announce message to the network service is when we produce a new block.

core/service.go
thomasmodeneis: sure

core/service.go
thomasmodeneis: right, this is what I understood after our call. but will remove.

core/service.go
thomasmodeneis: oks

core/service.go
thomasmodeneis: right, any ideas where can I get the info in this regards ? I did not add because I could not find where to get it from easily.

network/status.go
thomasmodeneis: sure

core/service.go
thomasmodeneis: do you mean the requestedBlockIDs ?

ryanchristo: Yes, I am asking about storing the requestedBlockIDs

noot: yeah it could be useful to have in the network state as opposed to core

core/service.go
noot: we have `blockAnnounceMessage` which tells us the ending block and  `s.blockState.LatestHeader()` which is the latest block that we have ie. the starting block

core/service.go
thomasmodeneis: ok

core/service.go
thomasmodeneis: Well they are both big.Int, so I did similar to what you mentioned: 
`if latestBlockNum.Cmp(messageBlockNum) == -1`

core/service.go
thomasmodeneis: right, blockAnnounceMessage has the field Number, this is what you mean by ending block right ? We can consider that as source of truth, ok.

network/status.go
thomasmodeneis: Not sure how to interpret your comment here. 
Could you please give a hint on how to obtain our the actual block number inside this func ? 
I need to have something to compare against msg.BestBlockNumber, is this a correct assumption ?

core/service.go
```diff
@@ -323,21 +331,47 @@ func (s *Service) handleReceivedMessage(msg network.Message) (err error) {
 // announce messages (block announce messages include the header but the full
 // block is required to execute `core_execute_block`).
 func (s *Service) ProcessBlockAnnounceMessage(msg network.Message) error {
+	blockAnnounceMessage, ok := msg.(*network.BlockAnnounceMessage)
+	if !ok {
+		return errors.New("could not cast p2p.Message to BlockAnnounceMessage")
```
ryanchristo: should be "network.Message"

core/service.go
```diff
@@ -323,21 +331,47 @@ func (s *Service) handleReceivedMessage(msg network.Message) (err error) {
...
+		seed := rand.New(s1).Uint64()
+		randomID := mrand.New(mrand.NewSource(int64(seed))).Uint64()
+
+		currentHash, _ := common.HexToBytes(s.blockState.LatestHeader().Hash().String())
+
+		blockAnnounceStateRootHash, _ := common.HexToHash(blockAnnounceMessage.StateRoot.String())
+
+		blockRequest := &network.BlockRequestMessage{
+			ID:            randomID, // random
+			RequestedData: 2,
```
ryanchristo: maybe add `// block body` here so we know what `2` represents

network/status.go
ryanchristo: This should probably not be done in the status submodule. Anything to do with syncing should be handled at the root level of the network service and anything in the status submodule should only be concerned with the status message exchange.

core/service.go
thomasmodeneis: good point

core/service.go
thomasmodeneis: sure

network/status.go
noot: we should probably just add an interface function into status that'll allow us to request blocks if needed, not sure how that'll work but let's open an issue for it. I think it's out of scope for this

core/service.go
noot: no, StateRoot isn't the block hash, you'll need to cast the blockAnnounceMessage to a `*types.Block` and do `.Hash()` on it to get the block hash

core/service.go
noot: checking `== -1` checks if latestBlockNum is less than messageBlockNum, not more than 2 less. you'll need to do a bit more big int math, you can do `messageBlockNum.Sub(messageBlockNum, big.NewInt(1))` then compare and see if latestBlockNum is still less

network/status.go
ryanchristo: We should not be adding anything to the status submodule. `status.handleMessage` should only be handling status messages, not block request messages.

core/service.go
thomasmodeneis: Hum, I tried doing: msg.(*types.Block) but that is a  impossible type assertion,
casting blockAnnounceMessage.(*types.Block) is also not possible as `Invalid type assertion: blockAnnounceMessage.(*types.Block) (non-interface type *network.BlockAnnounceMessage on left)`

Any ideas ?
`./service.go:356:39: impossible type assertion:
	*"github.com/ChainSafe/gossamer/core/types".Block does not implement "github.com/ChainSafe/gossamer/network".Message (wrong type for Decode method)
		have Decode([]byte) error
		want Decode(io.Reader) error`

core/service.go
thomasmodeneis: Yes right, the sub will make the trick!

core/service.go
noot: make a new block using the fields from the message:
```
header := types.NewHeader(msg.ParentHash, msg.Number, msg.StateRoot, msg.ExtrinsicsRoot, msg.Digest)

block := &types.Block{
    Header: header,
   Body: nil,
}
```

noot: make a new block using the fields from the message:
```
header := types.NewHeader(msg.ParentHash, msg.Number, msg.StateRoot, msg.ExtrinsicsRoot, msg.Digest)

block := &types.Block{
    Header: header,
   Body: nil,
}
```

thomasmodeneis: great, updated!

core/service.go
```diff
@@ -324,25 +334,55 @@ func (s *Service) handleReceivedMessage(msg network.Message) (err error) {
...
+
+	// check if we should send block request message
+	if latestBlockNum.Cmp(messageBlockNumMinusOne) == -1 {
+
+		//generate random ID
+		s1 := rand.NewSource(uint64(time.Now().UnixNano()))
+		seed := rand.New(s1).Uint64()
+		randomID := mrand.New(mrand.NewSource(int64(seed))).Uint64()
+
+		currentHash, _ := common.HexToBytes(s.blockState.LatestHeader().Hash().String())
```
noot: the `.String()` and `HexToBytes` are redundant, .Hash() is already `[32]byte`, to get a `[32]byte` as a slice you can do `[:]` at the end

core/service.go
```diff
@@ -324,25 +334,55 @@ func (s *Service) handleReceivedMessage(msg network.Message) (err error) {
...
-	log.Info("[core] imported block", "number", header.Number, "hash", header.Hash())
+		endBlock := &types.Block{
+			Header: header,
+		}
+
+		blockRequest := &network.BlockRequestMessage{
+			ID:            randomID, // random
+			RequestedData: 2,        // block body
+			StartingBlock: append([]byte{0}, currentHash...),
+			EndBlockHash:  optional.NewHash(true, endBlock.Header.Hash()),
```
noot: we can directly do header.Hash() here

core/service.go
noot: eg `currentHash := s.blockState.LatestHeader().Hash()`

core/service.go
```diff
@@ -324,25 +334,55 @@ func (s *Service) handleReceivedMessage(msg network.Message) (err error) {
...
 
-	log.Info("[core] imported block", "number", header.Number, "hash", header.Hash())
+		endBlock := &types.Block{
+			Header: header,
+		}
+
+		blockRequest := &network.BlockRequestMessage{
+			ID:            randomID, // random
+			RequestedData: 2,        // block body
+			StartingBlock: append([]byte{0}, currentHash...),
```
noot: here: `StartingBlock: append([]byte{0}, currentHash[:]...),`

core/service.go
thomasmodeneis: yes, thanks.

core/service.go
thomasmodeneis: yes

network/status.go
thomasmodeneis: ok I raised #637 in order to try to address this comments, would be a good idea if we can jump into a call and get more details about the issue as I am not understanding exactly the changes requested.

PR #610, opened 2020-02-11
core, consensus/babe, state: block persistence

### LINKED ISSUE #589: state: block persistence
when running a node, we can see some blocks getting produced; however, when we restart the node, the blocks start from 0 again. upon startup, the state service should load the previous state. it should get/set the latest block hash in the DB at `common.LatestHeaderHashKey` and use this on startup.

noot:
## Changes
- add functions in StorageState to update/get LatestHeaderHashKey which stores the hash of the most recent block
- update babe to update this every time a new block is built
- update core, babe state interfaces

## Tests:
```
make gossamer
./build/bin/gossamer
```

let the node build some blocks, then stop and restart. the block building should restart where it was stopped.

### Issues:
closes #589


consensus/babe/state.go
```diff
@@ -28,17 +28,9 @@ type BlockState interface {
...
-// NetworkState interface for network state methods
-type NetworkState interface {
-	GetHealth() (*common.Health, error)
-	SetHealth(*common.Health) error
-	GetNetworkState() (*common.NetworkState, error)
-	SetNetworkState(*common.NetworkState) error
-	GetPeers() (*[]common.PeerInfo, error)
-	SetPeers(*[]common.PeerInfo) error
-}
-
```
ryanchristo: Why remove network state interface methods? Should we keep all state interfaces consistent?

noot: the state interfaces should only contain the functions that are used by that package. otherwise if we change something in state we have to change all the interfaces. also if it's unused it's kinda unnecessary

PR #614, opened 2020-02-12
cmd, core: configure whether node is authority

### LINKED ISSUE #600: cmd, config: configure node type (block producer, grandpa authority, none, etc)
in the config, we should be able to configure whether we want to run a block producer (babe)/authority (grandpa) node or not.

noot:
# Changes
- add authority option to config
- add authority flag
- update makeNode to set core config based on these

## Tests:
```
make gossamer
./build/bin/gossamer --config config.toml 
```
authority should be true

```
./build/bin/gossamer --config config.toml  --authority=false
```
authority should be false

update config authority = false 
```
./build/bin/gossamer --config config.toml  --authority=true
```
authority should be true

### Issues:
closes #600


noot:
@edwardmack I've updated it, previously it was if the authority flag wasn't set, it was equal to false. I've added a check to see if the flag is set

PR #616, opened 2020-02-13
cmd, rpc: remove internal/api, add state service

### LINKED ISSUE #542: Deprecate internal/api
Once the State Service (i.e. State Refactor) is completed, we do not need `internal/api` anymore.

noot:
## Changes
- remove internal/api and its usages from RPC
- add state service to RPC module 
- start RPC server in node startup

## Tests:
```
go test ./rpc/...
```

### Issues:
closes #542


cmd/gossamer/configcmd.go
```diff
@@ -300,20 +296,22 @@ func setRPCConfig(ctx *cli.Context, fig *cfg.RPCCfg) {
...
 }
 
-func startRPC(ctx *cli.Context, fig cfg.RPCCfg, apiSrvc *api.Service) *rpc.HTTPServer {
-	if ctx.GlobalBool(RPCEnabledFlag.Name) {
-		return rpc.NewHTTPServer(apiSrvc.API, &json2.Codec{}, fig.Host, fig.Port, fig.Modules)
+func setupRPC(ctx *cli.Context, fig cfg.RPCCfg, stateSrv *state.Service, networkSrvc *network.Service) *rpc.HTTPServer {
+	cfg := &rpc.HTTPServerConfig{
+		BlockApi:   stateSrv.Block,
+		StorageApi: stateSrv.Storage,
+		NetworkApi: networkSrvc,
```
ryanchristo: I think we should maybe keep network state. See comments in #606.

ryanchristo:
Also, I'm getting a sigabort error running gossamer with the updates...

cmd/gossamer/configcmd.go
noot: this PR doesn't remove it, it just gets the API needed for rpc directly from network. we could  swap it out later for network service

noot:
@ryanchristo please try again, should be fixed now

cmd/gossamer/configcmd.go
```diff
@@ -300,20 +296,22 @@ func setRPCConfig(ctx *cli.Context, fig *cfg.RPCCfg) {
...
 }
 
-func startRPC(ctx *cli.Context, fig cfg.RPCCfg, apiSrvc *api.Service) *rpc.HTTPServer {
-	if ctx.GlobalBool(RPCEnabledFlag.Name) {
-		return rpc.NewHTTPServer(apiSrvc.API, &json2.Codec{}, fig.Host, fig.Port, fig.Modules)
+func setupRPC(ctx *cli.Context, fig cfg.RPCCfg, stateSrv *state.Service, networkSrvc *network.Service) *rpc.HTTPServer {
+	cfg := &rpc.HTTPServerConfig{
+		BlockAPI:   stateSrv.Block,
+		StorageAPI: stateSrv.Storage,
+		NetworkAPI: networkSrvc,
```
ryanchristo: We might need both `stateSrv.Network` and `networkSrvc` in the future. Can we combine the two interfaces before passing to `NetworkAPI`? If not, maybe we should add another property to `HTTPServerConfig` and pass `networkSrvc` to something other than `NetworkAPI` here?

network/config.go
```diff
@@ -109,15 +107,10 @@ func (c *Config) build() error {
 }
 
 func (c *Config) checkState() (err error) {
-	if c.BlockState == nil {
+	if c.BlockState == nil && !c.NoStatus {
```
ryanchristo: Why was `&& !c.NoStatus` added here?

rpc/modules/author.go
```diff
@@ -67,33 +66,33 @@ type ExtrinsicStatus struct {
...
-func NewAuthorModule(api *api.API) *AuthorModule {
+func NewAuthorModule(api CoreAPI) *AuthorModule {
 	return &AuthorModule{
-		api: api,
+		coreApi: api,
 	}
 }
 
-// InsertKey Insert a key into the keystore
+// InsertKey snserts a key into the keystore
```
ryanchristo: inserts

rpc/modules/system.go
```diff
@@ -92,21 +91,24 @@ func (sm *SystemModule) Version(r *http.Request, req *EmptyRequest, res *StringR
 
 // Health returns the information about the health of the network
 func (sm *SystemModule) Health(r *http.Request, req *EmptyRequest, res *SystemHealthResponse) error {
-	// TODO: migrate from api to network state
-	res.Health = sm.api.NetworkModule.Health()
+	// TODO: rename to not have `Get` to match API
```
ryanchristo: let's open an issue to address this

rpc/server_test.go
```diff
@@ -185,19 +185,4 @@ func TestServeHTTP(t *testing.T) {
...
-	}
-	r.Header.Set("Content-Type", "www-url-encoded")
-	w = NewMockResponseWriter()
-	s.ServeHTTP(w, r)
-	if w.Status != 415 {
-		t.Errorf("unexpected status. got: %d expected: %d", w.Status, 415)
-	}
-	if w.Body != strconv.Itoa(10) {
-		t.Errorf("unexpected body content. got: %s expected %s", w.Body, strconv.Itoa(10))
-	}
```
ryanchristo: what are we removing here?

noot: duplicated test, it you look above it was the exact same thing

cmd/gossamer/configcmd.go
noot: not sure I understand, NetworkAPI is an interface and stateSrv.Network and networkSrvc are both structs that implement NetworkAPI. we don't need both right now

network/config.go
noot: it's an easy way to configure whether you need a BlockState or not in the tests

rpc/modules/system.go
noot: this is actually not relevant anymore, I've removed it

cmd/gossamer/configcmd.go
ryanchristo: We don't need it now but I thought we might prepare. Fine as is, just a heads up we might need to make `stateSrv.Netwwork` part of `NetworkAPI` in the future.

network/config.go
ryanchristo: maybe a comment to explain that, otherwise it looks a bit out of place

rpc/server_test.go
ryanchristo: gotcha, makes sense :+1:

network/state.go
```diff
@@ -17,28 +17,10 @@
...
-
-// NetworkState interface for network state methods
-type NetworkState interface {
-	GetHealth() (*common.Health, error)
-	SetHealth(*common.Health) error
-	GetNetworkState() (*common.NetworkState, error)
-	SetNetworkState(*common.NetworkState) error
-	GetPeers() (*[]common.PeerInfo, error)
-	SetPeers(*[]common.PeerInfo) error
-}
```
ryanchristo: should we keep `NetworkState` as an empty struct since it will be needed soon for requested block ids?

network/state_test.go
```diff
@@ -59,44 +49,3 @@ func (mbs *MockBlockState) LatestHeader() *types.Header {
 		Digest:         [][]byte{{}},
 	}
 }
-
-// MockNetworkState
-type MockNetworkState struct{}
```
ryanchristo: we should probably keep `MockNetworkState` as well since we will need it soon

network/config.go
noot: sure thing

network/state.go
noot: I'm not sure why the network package would need network state- the network package already knowns about its own state.

ryanchristo: the network service will need network state to set network state

noot: okay, either way we should add it in when we need it

noot: i can add it back in but i really don't like having unused code.... if a PR needs it, it should add it in. also even if it's deleted it's still always there because git

cmd/gossamer/configcmd.go
noot: yeah however I think the RPC only needs those 3 methods from network, so this probably won't need to be updated

network/state_test.go
noot: I have re-added it

rpc/modules/api.go
```diff
@@ -0,0 +1,21 @@
+package modules
+
+import (
+	"github.com/ChainSafe/gossamer/common"
+)
+
+// StorageAPI ...
```
thomasmodeneis: Can we not add (...) as a valid description for public exported structs ? would be nice to have a proper explanation of why this is useful and what it does. Same for the others you have added in this PR.

core/service.go
```diff
@@ -125,7 +125,7 @@ func NewService(cfg *Config) (*Service, error) {
 		if err != nil {
 			log.Error("[core] could not start babe session", "error", err)
 			srv.isBabeAuthority = false
-			return nil, nil
+			return srv, nil
```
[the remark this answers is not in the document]
noot: this function gets called from configcmd.go makeNode, so if the error was returned here, the entire node would shutdown if the babe session fails to start.

it depends on what we want to node functionality to do, if we run a node but can't start babe, should the node shutdown or should it just log and error?

rpc/modules/api.go
noot: done

cmd/gossamer/configcmd.go
```diff
@@ -300,20 +297,22 @@ func setRPCConfig(ctx *cli.Context, fig *cfg.RPCCfg) {
...
-func strToMods(strs []string) []api.Module {
-	var res []api.Module
-	for _, str := range strs {
-		res = append(res, api.Module(str))
+	if ctx.GlobalBool(RPCEnabledFlag.Name) {
+		return rpc.NewHTTPServer(cfg)
 	}
-	return res
+
+	return nil
```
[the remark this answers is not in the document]
noot: done

core/service.go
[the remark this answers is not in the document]
ryanchristo: logging the error and continuing to start the node would be ok with me for now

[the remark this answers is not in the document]
thomasmodeneis: well if the node is configured as a authority node and it fails to start, that is a severe and should be the error should be propagated. Otherwise its inconsistent.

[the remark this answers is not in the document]
noot: ok, upon further thought, I agree and I think it should probably exit. I will update it in my other PR

PR #618, opened 2020-02-15
core, network: clean up for syncing

noot:
## Changes
- add imported blocks to state service
- cleanup core tests
- remove LEB128 encoding for now

## Tests:

terminal 1:
```
make gossamer
./build/bin/gossamer --config config.toml init
./build/bin/gossamer --config config.toml 
```

in another terminal: make another config file, change to `authority = false` and change port
```
make gossamer
./build/bin/gossamer --config config2.toml init
./build/bin/gossamer --config config2.toml 
```


core/service.go
```diff
@@ -321,21 +329,26 @@ func (s *Service) handleReceivedMessage(msg network.Message) (err error) {
...
+		err = s.blockState.SetHeader(header)
+		if err != nil {
+			return err
+		}
+	} else {
+		return err
 	}
 
-	// send block request message to network service
-	s.msgSend <- blockRequest
```
ryanchristo: This is part of `TODO: check if we should send block request message`. Would be ok to delete it but might make be nice to just comment out since we will need it soon.

core/service_test.go
```diff
@@ -30,73 +31,105 @@ import (
...
+	}
 
-func TestValidateBlock(t *testing.T) {
-	rt := runtime.NewTestRuntime(t, tests.POLKADOT_RUNTIME)
+	// defer func() {
+	// 	err = dbSrv.Stop()
+	// 	if err != nil {
+	// 		t.Fatal(err)
+	// 	}
+	// }()
```
ryanchristo: Why is this commented out?

network/host.go
```diff
@@ -187,9 +187,7 @@ func (h *host) send(p peer.ID, msg Message) (err error) {
 		return err
 	}
 
-	// append leb128 variable-length encoding
-	encMsg = encodeMessageLEB128(encMsg)
-
+	// TODO: append leb128 variable-length encoding
```
ryanchristo: reopened #484

network/service.go
```diff
@@ -182,12 +182,7 @@ func (s *Service) handleStream(stream libp2pnetwork.Stream) {
 	r := bufio.NewReader(stream)
 
 	for {
-		// read leb128 variable-length encoding
-		_, err := common.ReadByte(r)
-		if err != nil {
-			log.Error("Failed to read message encoding", "peer", peer, "err", err)
-			return // exit
-		}
+		// TODO: re-add leb128 variable-length encoding
```
ryanchristo: reopened #484

core/service_test.go
noot: shouldn't be here, this function shouldn't stop the db service, but the tests that use it should

core/service.go
noot: we shouldn't be leaving commented out code, would rather just delete it

core/service.go
```diff
@@ -323,21 +331,25 @@ func (s *Service) handleReceivedMessage(msg network.Message) (err error) {
...
 
-	// TODO: update message properties and use generated id
-	blockRequest := &network.BlockRequestMessage{
-		ID:            1,
-		RequestedData: 2,
-		StartingBlock: []byte{},
-		EndBlockHash:  optional.NewHash(true, common.Hash{}),
-		Direction:     1,
-		Max:           optional.NewUint32(false, 0),
+	// TODO: check if we should send block request message
```
ryanchristo: Can we resolve this here? We just need to see if the announce message is later than the latest block in state then send the request message to the network service.

core/service.go
ryanchristo: rather than deleting it and adding it back in, can we just resolve it here? see above

core/service.go
noot: isn't that what Thomas is working on? I could do it but I think it's out of scope

ryanchristo: right... I was confused by the issue and the title of the pull request, which is related to sending a block request based on the status message, not the block announce message.

The two pull requests overlap quite a bit. Might be best to wait with the pull request.

noot: okay it would be good to get this one in because this allows for non authority nodes to sync a block producer

core/service.go
```diff
@@ -281,6 +280,15 @@ func (s *Service) receiveMessages() {
 //TODO: remove nolint after `(*Service).handleReceivedBlock` - result `err` is no longer always `nil`
```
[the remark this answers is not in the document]
noot: yep!

core/service.go
```diff
@@ -123,9 +122,9 @@ func NewService(cfg *Config) (*Service, error) {
 		// create a new BABE session
 		bs, err := babe.NewSession(bsConfig)
 		if err != nil {
-			log.Error("[core] could not start babe session", "error", err)
 			srv.isBabeAuthority = false
-			return nil, nil
+			log.Error("[core] could not start babe session", "error", err)
+			return srv, nil
```
thomasmodeneis: return nil err after babe.NewSession fails, why is that ?

network/host.go
thomasmodeneis: pls add the issue number to the comment so it can be tracked back.

network/service.go
thomasmodeneis: pls add the issue number to the comment so it can be tracked back.

core/service.go
```diff
@@ -323,21 +329,25 @@ func (s *Service) handleReceivedMessage(msg network.Message) (err error) {
...
-	blockRequest := &network.BlockRequestMessage{
-		ID:            1,
-		RequestedData: 2,
-		StartingBlock: []byte{},
-		EndBlockHash:  optional.NewHash(true, common.Hash{}),
-		Direction:     1,
-		Max:           optional.NewUint32(false, 0),
+	// TODO: check if we should send block request message
+	_, err = s.blockState.GetHeader(header.Hash())
+	if err != nil && err.Error() == "Key not found" {
```
thomasmodeneis: This looks like a "magic" string, shall it be a const and shared among GetHeader and places where the check is required ?

core/service.go
noot: if we return an error here, then it gets propogated up to configcmd.go `makeNode` which will cause the node to shutdown. it's not necessarily a fatal error to not be able to start BABE, since the other parts of the node can still run. 

it depends on what we want the behaviour of the node to look like, if we are trying to run an authority node and we can't start BABE, should the node exit completely or should it just log an error?

core/service.go
noot: the error actually comes from here: https://github.com/dgraph-io/badger/blob/f5b63211d7f3e2f5f8b698893313b2a54e4df7de/errors.go#L36

I don't think we are using it anywhere else at the moment, if I wanted to do an error check I would need to import badger into this package. I think it should be imported somewhere else in gossamer, perhaps state, if we are going to use this often

core/service.go
ryanchristo: logging the error and continuing to start the node seems ok for now

noot: cool, I'll leave it as is then, if that's good with everyone!

thomasmodeneis: well, I would vote to then check if babe was requested, if so, return the err.

noot: I've updated it

noot:
@edwardmack did you change the bootnodes array of the second node to include the p2p address of the first node? 

also, I think my instructions aren't complete, you will need to have one of the keypairs in the genesis file in your .gossamer/keystore directory, and then run `./build/bin/gossamer --config config.toml --unlock 0`. I can send it to you on slack for now but this will be fixed in #624

cmd/gossamer/configcmd.go
```diff
@@ -273,7 +273,8 @@ func createNetworkService(fig *cfg.Config, gendata *genesis.Data, stateService *
 func createCoreService(coreConfig *core.Config) *core.Service {
 	coreService, err := core.NewService(coreConfig)
 	if err != nil {
-		log.Error("Failed to create new core service", "err", err)
+		log.Crit("Failed to create new core service", "err", err)
+		os.Exit(0)
```
thomasmodeneis: would that not be exit(1) since it was a fatal err ?

thomasmodeneis: would that not be exit(1) since it was a fatal err ?

thomasmodeneis: well, but that was a fatal system error, since the core was not able to start and that is a critical error. If we exit with 0, it looks like it was a clean exit.

noot: yes, it did exit cleanly, because the program behaved as expected- we couldn't start a service so we exited. the error was "caught". this is not the same as a segfault or other memory access error, where the program is forced to exit by the kernel.

thomasmodeneis: the program failed to createCoreService because create it could not instantiate a new BABE session, that is not behaving as expected. for example: BDD: `Given a user set cfg.IsBabeAuthority to true, When a service fails to be instantiated, Then I  expect a exit with non zero `. This behavior is already present if you consider that main.go already exit with 1 when a err is returned. This is there so the ops knows something is wrong.

Since you wrote a critical log there and then assuming this is a critical error, I believe the logic thing to do is to propagate it back to the root (func createCoreService->func makeNode->func gossamer->exit(1) ) and that will exit the program with the propagated err.
https://github.com/ChainSafe/gossamer/blob/development/cmd/gossamer/main.go#L131

PR #621, opened 2020-02-19
rpc, state: add tx queue to state, implement author_SubmitExtrinsic and author_PendingExtrinsics

### LINKED ISSUE #263: Implement Author RPC Calls

noot:
## Changes
- implement author_SubmitExtrinsic and author_PendingExtrinsics
- add transaction queue to state, would like feedback on this; does it make sense to have it instantiated in state rather than core? this way, core, babe and rpc would push/pull from it

## Tests:
```
go test ./...
```

### Issues:
partially closes #263


rpc/http.go
```diff
@@ -37,6 +37,7 @@ type HTTPServerConfig struct {
 	StorageAPI modules.StorageAPI
 	NetworkAPI modules.NetworkAPI
 	CoreAPI    modules.CoreAPI
+	TxQueueAPI modules.TransactionQueueAPI
```
ryanchristo: Again, I would probably write out the full `TransactionQueueAPI` given its a configuration struct.

rpc/modules/author.go
```diff
@@ -20,11 +20,15 @@ import (
 	"net/http"
 
 	"github.com/ChainSafe/gossamer/common"
+	tx "github.com/ChainSafe/gossamer/common/transaction"
```
ryanchristo: this file is a good example of when you might want to save `tx` for variables

rpc/server.go
```diff
@@ -45,6 +45,7 @@ type ServerConfig struct {
 	StorageAPI modules.StorageAPI
 	NetworkAPI modules.NetworkAPI
 	CoreAPI    modules.CoreAPI
+	TxQueueAPI modules.TransactionQueueAPI
```
ryanchristo: full name should probably be used in config struct here too

rpc/modules/author.go
```diff
@@ -77,33 +81,79 @@ type ExtrinsicStatus struct {
...
+		return err
+	}
+
+	log.Trace("[rpc]", "extrinsic", extBytes)
+
+	// TODO: validate transaction before submitting to tx queue
+
+	ext := types.Extrinsic(extBytes)
+
+	// TODO: form valid transaction by decoding tx bytes
```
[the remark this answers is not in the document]
noot: we need some sort of gossamer transaction utils to do this, however transaction/extrinsic format is dependent on runtime, so maybe we can do this after we decide on a fixed runtime version, I can open an issue for it in the meantime

rpc/modules/author.go
```diff
@@ -77,33 +81,79 @@ type ExtrinsicStatus struct {
...
+// SubmitExtrinsic Submit a fully formatted extrinsic for block inclusion
+func (cm *AuthorModule) SubmitExtrinsic(r *http.Request, req *Extrinsic, res *ExtrinsicHashResponse) error {
+	extBytes, err := common.HexToBytes(string(*req))
+	if err != nil {
+		return err
+	}
+
+	log.Trace("[rpc]", "extrinsic", extBytes)
+
+	// TODO: validate transaction before submitting to tx queue
```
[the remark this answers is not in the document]
noot: same as below

cmd/gossamer/configcmd.go
```diff
@@ -105,6 +105,7 @@ func makeNode(ctx *cli.Context) (*dot.Dot, *cfg.Config, error) {
 	coreConfig := &core.Config{
 		BlockState:      stateSrv.Block,
 		StorageState:    stateSrv.Storage,
+		TxQueue:         stateSrv.TransactionQueue,
```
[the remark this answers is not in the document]
noot: I agree, I'll update it

rpc/http.go
noot: updated

rpc/server.go
noot: updated

rpc/modules/author.go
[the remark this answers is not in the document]
noot: #641

PR #625, opened 2020-02-21
core, state: store block data after BlockResponseMessage

### LINKED ISSUE #561: Validate and store received blocks
Upon receiving a block from the network, validate the block, and if valid add the block to the DB.

Acceptance Criteria
- A valid block received from the network is able to be retrieved from the db.
- An invalid block is discarded (i.e. it will not end up in the db)

Questions
- How do we deal with valid blocks that we received that have already been stored?

noot:
# Changes
- store BlockData in state after BlockResponseMessage
- store header in state after BlockResponseMessage

## Tests:
```
go test ./... -short
```

### Issues:
closes #561


core/service.go
```diff
@@ -397,14 +386,68 @@ func (s *Service) ProcessBlockResponseMessage(msg network.Message) error {
...
+		return fmt.Errorf("no blockState")
+	}
+
+	existingData, err := s.blockState.GetBlockData(bd.Hash)
+	if err != nil {
+		// no block data exists, ok
+		return s.blockState.SetBlockData(bd)
+	}
+
+	if existingData == nil {
```
ryanchristo: this can be combined with the above

Issue #615, opened 2020-02-13
Improve network package tests


ryanchristo:
## Expected Behavior

When the network tests encounter a failed dial error or a failed attempt to load the temporary data directory, the network tests should not fail (this might be out of our control) or the failed test should attempt a retry to limit the number of failed CI tests.

## Current Behavior

Intermittent failed dial errors and errors failing to load temporary data directory. It seems that when one small hiccup occurs, subsequent tests are affected as well.

## Possible Solution

If error equals failed dial error, rerun test.

## Steps to Reproduce

Wait for CI to fail on failed dials or inability to load or cleanup temporary directory

ryanchristo:
```
=== RUN   TestDot_Start
t=2020-02-12T11:35:20+0100 lvl=warn msg="Generating p2p identity from seed" DataDir=/tmp/gossamer-test RandSeed=1
t=2020-02-12T11:35:20+0100 lvl=warn msg="ProtocolID not defined, using DefaultProtocolID" DefaultProtocolID=/gossamer/dot/0
t=2020-02-12T11:35:20+0100 lvl=warn msg="Bootstrap is enabled and no bootstrap nodes are defined"
t=2020-02-12T11:35:20+0100 lvl=trce msg="Started listening" host=12D3KooWHHzSeKaY8xuZVzkLbKFfvNgPPeKhFBGrMbNzbm5akpqu address=/ip4/0.0.0.0/tcp/7000
--- FAIL: TestDot_Start (0.02s)
    dot_test.go:59: Error while writing key registry.: Error while renaming file in WriteKeyRegistry: rename /tmp/gossamer-test/state/REWRITE-KEYREGISTRY /tmp/gossamer-test/state/KEYREGISTRY: no such file or directory
```

```
=== RUN   TestMdns
t=2020-02-12T11:35:20+0100 lvl=warn msg="Generating p2p identity from seed" DataDir=/tmp/gossamer-test/nodeA RandSeed=1
t=2020-02-12T11:35:20+0100 lvl=trce msg="Started listening" host=12D3KooWHHzSeKaY8xuZVzkLbKFfvNgPPeKhFBGrMbNzbm5akpqu address=/ip4/0.0.0.0/tcp/7001
Listening on the following addresses...
/ip4/127.0.0.1/tcp/7001/p2p/12D3KooWHHzSeKaY8xuZVzkLbKFfvNgPPeKhFBGrMbNzbm5akpqu
[...]

PR #628, opened 2020-02-23
scale: move initialization of large byte arrays

noot:
## Changes
- I was occassionally getting out of memory errors when running tests due to the large byte arrays being initialized, so I moved them inside the block that's executed when not -short

## Tests:
```
go test ./... -short
```


thomasmodeneis:
Is this going to eliminate the actual failures ? I mean, how better can we expect it to be now after this is merged ?

noot:
@thomasmodeneis fixing the network tests is not in the scope of this PR. I was seeing runtime: out of memory errors on my other PRs running in travis, also I get it locally when I try to run all the tests. this fixes that error.

ryanchristo:
issue already opened for network tests that we can add to #615

Issue #626, opened 2020-02-22
utilize blocktree in block production


noot:
the current head of the chain should come from the blocktree as per spec algorithm 5.3. the newly built block should be placed on the longest chain.

Issue #562, opened 2020-01-23
Utilize blocktree in syncing


ansermino:
Upon validating a block we need to add it to our block tree.

Note: Blocktree should move to `core` so grandpa can use it too.

Issue #567, opened 2020-01-23
Test Median Algorithm


ansermino:
Once we can establish a small network we can evaluate if the median algorithm works. If a node joins an existing network, it should establish the same slot times as the existing participants.

amerameen:
Ensure to do this upon startup

PR #629, opened 2020-02-23
core/blocktree, cmd, core, state, consensus/babe: add blocktree to BlockState, persist blocktree upon restarts

### LINKED ISSUE #627: block tree initialization and persistence
the block tree should be initialized with the genesis block when `gossamer init` is executed.  also, upon node shutdown/startup, we should save/load the blocktree structure from the database. note: all the blocks in the blocktree should already be stored in the DB, however, we still need to store the structure using block hashes.

also need to modify blockState to get head of the chain from blocktree.

noot:
## Changes
- add blocktree to BlockState; core and babe shouldn't be interfacing directly with the blocktree, but it should be managed by the BlockState
- add blocktree store/load to state so that we reload the blocktree when we restart our node

## Tests:
```
go test ./... -short
```

### Issues:
closes #627 #626 #562


dot/state/block.go
```diff
@@ -277,31 +270,52 @@ func (bs *BlockState) SetBlockData(blockData *types.BlockData) error {
...
+	return bs.GetHeader(bs.BestBlockHash())
+}
+
+// SubChain returns the sub-blockchain between the starting hash and the ending hash using the block tree
+func (bs *BlockState) SubChain(start, end common.Hash) []common.Hash {
+	return bs.bt.SubBlockchain(start, end)
+}
+
+// ComputeSlotForBlock returns the slot number for a given block
+// TODO: can move this out of the blocktree into BABE
```
noot: will address in #567

lib/babe/babe_test.go
```diff
@@ -238,7 +239,97 @@ func TestSlotOffset(t *testing.T) {
...
+
+		bt.AddBlock(block)
+		previousHash = hash
+		previousAT = block.GetBlockArrivalTime()
+		blockState.AddBlock(block)
+	}
+}
+
+func TestSlotTime(t *testing.T) {
+	t.Skip()
```
noot: will fix this in #567

dot/core/service.go
```diff
@@ -350,11 +350,29 @@ func (s *Service) ProcessBlockAnnounceMessage(msg network.Message) error {
...
+	if err != nil && err.Error() == "Key not found" {
+		err = s.blockState.SetHeader(header)
+		if err != nil {
+			return err
+		}
+	} else {
+		return err
+	}
+
+	log.Info("[core] imported block", "number", header.Number, "hash", header.Hash())
```
ryanchristo: We only imported block if we set the header, correct? Should this be in the same block of code as `SetHeader`?

noot: good point, will update!

noot:
@ryanchristo yep I'm working on some tests right now, I think part of the coverage being so low is cause I put the slotTime function back in but skipped the test cause it's broken :/ but yeah coverage should defs be higher

dot/core/service.go
```diff
@@ -137,8 +137,8 @@ func NewService(cfg *Config) (*Service, error) {
 		// create a new BABE session
 		bs, err := babe.NewSession(bsConfig)
 		if err != nil {
-			log.Error("[core] could not start babe session", "error", err)
 			srv.isBabeAuthority = false
+			log.Error("[core] could not start babe session", "error", err)
```
thomasmodeneis: is this the new default prefix to use now for any method inside core ? [core] ?

dot/network/mdns.go
```diff
@@ -65,6 +65,7 @@ func (m *mdns) start() {
 	)
 	if err != nil {
 		log.Error("Failed to start mDNS discovery service", "err", err)
+		return
```
thomasmodeneis: well, if the discovery can fail and return a error, shall we just log it silently here ? is this the desired behavior ?

dot/network/service.go
```diff
@@ -153,7 +153,10 @@ func (s *Service) handleConn(conn network.Conn) {
 	if !s.noStatus {
 
 		// get latest block header from block state
-		latestBlock := s.cfg.BlockState.LatestHeader()
+		latestBlock, err := s.cfg.BlockState.BestBlockHeader()
+		if err != nil {
+			log.Error("[network] could not get chain head", "err", err)
```
thomasmodeneis: since its returning a error, if we ignore it here, is there any changes that calling latestBlock would return a nillpointer ?

dot/network/state_test.go
```diff
@@ -27,18 +27,18 @@ import (
 type MockBlockState struct{}
 
 // LatestHeader for MockBlockState
```
thomasmodeneis: update method comment

lib/babe/babe.go
```diff
@@ -190,38 +190,39 @@ func (b *Session) invokeBlockAuthoring() {
 }
 
 func (b *Session) handleSlot(slotNum uint64) {
-	parentHeader := b.blockState.LatestHeader()
+	parentHeader, err := b.blockState.BestBlockHeader()
+	if err != nil {
+		log.Error("BABE block authoring", "error", "parent header is nil")
+		return
```
thomasmodeneis: if we just return silently, how can the caller method know there was a failure and act on it ?

dot/core/service.go
noot: yeah, i think [package name] seems pretty standard for logging. there wasn't much consistency before, so i think if everyone is okay with this logging style, we should update the logs.

dot/network/mdns.go
noot: I believe this is running inside a goroutine, if we want to check for errors we would need to create an error channel. mDNS isn't a critical component so I'm okay with just logging it

dot/network/service.go
noot: will return if error

lib/babe/babe.go
noot: the caller method is `invokeBlockAuthoring` which is called in a goroutine inside `Start()`. the caller method only handles errors by logging them, it doesn't return anything to `Start()`. 

also, it's fine if BABE can't produce a block for one slot.

dot/network/mdns.go
noot: @ryanchristo thoughts?

noot:
@ryanchristo added some more tests for blocktree persistence, ready for re-review

lib/babe/babe.go
thomasmodeneis: oh right, make sense

dot/core/service.go
thomasmodeneis: yes but only for core or all packages ? if eg: core/types the log would be [types] or [core] ?

noot: good point, i think we would have to put [core/types] since there are multiple types packages.  let's discuss this on our next call

Issue #417, opened 2019-11-20
Catch unrecognized subcommands


ryanchristo:
## Expected Behavior

When running the `gossamer` command with an incorrect subcommand, the help screen stating "Incorrect usage" should appear.

## Current Behavior

When running the `gossamer` command with an incorrect subcommand, gossamer starts.

## Steps to Reproduce

```
gossamer potato
```

PR #631, opened 2020-02-25
cmd/gossamer: fix gossamer invalid arguments issue #417

thomasmodeneis:
## Changes
-  Fixes bug where any command is accepted and starts gossamer
- Add test case for the issue: Should start gossamer with invalid command and observe the stderr output and fail if does not contain the stderr we expect to see when running gossamer with invalid command 
- Fix lint issues:
`cmd/gossamer/configcmd_test.go:144:2: shadow: declaration of "app" shadows declaration at line 29 (govet)
        app := cli.NewApp()
        ^
cmd/gossamer/configcmd_test.go:170:2: shadow: declaration of "app" shadows declaration at line 29 (govet)
        app := cli.NewApp()
        ^
cmd/gossamer/configcmd_test.go:213:2: shadow: declaration of "app" shadows declaration at line 29 (govet)
        app := cli.NewApp()
        ^
cmd/gossamer/configcmd_test.go:295:2: shadow: declaration of "app" shadows declaration at line 29 (govet)
        app := cli.NewApp()
        ^
cmd/gossamer/configcmd_test.go:353:2: shadow: declaration of "app" shadows declaration at line 29 (govet)
        app := cli.NewApp()
        ^
cmd/gossamer/configcmd_test.go:405:3: shadow: declaration of "app" shadows declaration at line 29 (govet)
                app := cli.NewApp()
                ^`

## Tests:
```
TestGossamerInvalidCommand
```

### Issues: #417


cmd/gossamer/configcmd_test.go
```diff
@@ -141,8 +141,8 @@ func TestGetConfig(t *testing.T) {
 	cfgClone.Global.DataDir, err = filepath.Abs(cfgClone.Global.DataDir)
 	require.Nil(t, err)
 
-	app := cli.NewApp()
-	app.Writer = ioutil.Discard
+	thisapp := cli.NewApp()
```
noot: I don't see where the previous declaration was?

thomasmodeneis: Its:
`cmd/gossamer/main.go:29` `app       = cli.NewApp()`

thomasmodeneis:
> looks good, one suggestion would be to make the error from an invalid command include the valid commands, eg:
> 
> ```
> elizabeth@elizabeth:~/go/src/github.com/ChainSafe/gossamer$ ./build/bin/gossamer noot
> Invalid command "noot"
> Usage: gossamer account | init 
> ```
> 
> or something like that

Right, maybe, but its quite of a thing to make it happen after the error also because it will mess with the err output and the test. 
The default behavior for CLI is to exit with cmd not found when typing wrong cmd. Its common knowledge that a user should run --help to find out what the correct cmd is.

noot:
sure, that's fine then

cmd/gossamer/gossamer_cmd_test.go
```diff
@@ -0,0 +1,184 @@
+package main
```
edwardmack: Minor:  Add copyright notice.

thomasmodeneis: good point

PR #633, opened 2020-02-25
core/blocktree, consensus/babe: cleanup blocktree, add encode and decode methods

### LINKED ISSUE #627: block tree initialization and persistence

noot:
## Changes
- cleanup blocktree package- remove unused functions, unexport funcs/types that should be unexported
- add encode/decode, store/load functions for blocktree (so we can store it in the db for persistence)

## Tests:
```
go test ./lib/blocktree/
```

### Issues:
related to #627 

note: this breaks the slotTime function, but I have fixes + blocktree integration with BlockState in here https://github.com/ChainSafe/gossamer/pull/629 but broke it up into 2 PRs. but if you want to look ahead you can look at that one as well


lib/blocktree/database.go
```diff
@@ -0,0 +1,138 @@
...
+		depth:       big.NewInt(0),
+		arrivalTime: arrivalTime,
+	}
+
+	bt.leaves = leafMap{bt.head.hash: bt.head}
+
+	return bt.decode(r, bt.head)
+}
+
+func (bt *BlockTree) decode(r io.Reader, parent *node) error {
```
thomasmodeneis: do we really have to have two methods with almost the same name ? decode and Decode ?

lib/common/db_keys.go
```diff
@@ -16,11 +16,14 @@
 
 package common
 
+//nolint
```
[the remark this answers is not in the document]
noot: good point, I think it showed up when I merged development. will remove

lib/blocktree/node.go
```diff
@@ -20,43 +20,40 @@ import (
...
 // addChild appends Node to n's list of children
-func (n *Node) addChild(node *Node) {
+func (n *node) addChild(node *node) {
 	n.children = append(n.children, node)
 }
 
 // String returns stringified hash and depth of node
-func (n *Node) String() string {
-	return fmt.Sprintf("{h: %s, d: %s}", n.hash.String(), n.depth)
+func (n *node) String() string {
```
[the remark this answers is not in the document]
noot: good point, will update

lib/babe/babe_test.go
```diff
@@ -283,21 +279,21 @@ func createFlatBlockTree(t *testing.T, depth int) *blocktree.BlockTree {
 	return bt
 }
 
-func TestSlotTime(t *testing.T) {
-	bt := createFlatBlockTree(t, 100)
-	babesession := createTestSession(t, nil)
+// func TestSlotTime(t *testing.T) {
+// 	bt := createFlatBlockTree(t, 100)
```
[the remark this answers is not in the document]
noot: unfortunately it doesn't compile, I'm going to just remove it for now since I have the updated version in #629

lib/blocktree/database.go
noot: I'm fine with it, I think it's clear that one is used internally and one is used externally. what would you suggest?

lib/blocktree/blocktree.go
```diff
@@ -19,187 +19,168 @@ package blocktree
...
-		number:      genesis.Header.Number,
+func NewBlockTreeFromGenesis(genesis *types.Header, db database.Database) *BlockTree {
+	head := &node{
+		hash:        genesis.Hash(),
 		parent:      nil,
-		children:    []*Node{},
+		children:    []*node{},
 		depth:       big.NewInt(0),
-		arrivalTime: genesis.GetBlockArrivalTime(),
+		arrivalTime: uint64(time.Now().Unix()), // TODO: genesis block doesn't need an arrival time, it isn't used in median algo
```
ryanchristo: issue number? can you explain?

lib/blocktree/blocktree.go
```diff
@@ -19,187 +19,168 @@ package blocktree
...
 
 // ComputeSlotForBlock computes the slot for a block from genesis
-// helper for now, there's a better way to do this
 func (bt *BlockTree) ComputeSlotForBlock(b *types.Block, sd uint64) uint64 {
+	return bt.computeSlotForNode(&node{
+		arrivalTime: b.GetBlockArrivalTime(),
+	}, sd)
+}
+
+// TODOge: not sure how correct this is, there isn't necessarily one block per slot
```
ryanchristo: typo and issue number?

lib/blocktree/database.go
```diff
@@ -0,0 +1,138 @@
...
+	return bt.Decode(enc)
+}
+
+// Encode recursively encodes the block tree
+// enc(node) = [32B block hash + 8B arrival time + 8B num children n] | enc(children[0]) | ... | enc(children[n-1])
+func (bt *BlockTree) Encode() ([]byte, error) {
+	return encode(bt.head, []byte{})
+}
+
+func encode(n *node, enc []byte) ([]byte, error) {
```
ryanchristo: missing method comment, would be nice to clearly differentiate between `Encode` and `encode`

lib/blocktree/database.go
ryanchristo: I am ok with it for now unless someone has a better idea but a comment here is needed either way

lib/blocktree/blocktree.go
noot: will address in #567 

arrivalTime is only relevant for blocks that we receive from other nodes on the network (the purpose of arrivalTime is to allow us to calculate what slot other nodes are on) since the genesis block comes from the genesis file, not the network, it's not relevant to the slot time calculation

lib/blocktree/blocktree.go
noot: will fix, and #567

lib/blocktree/database.go
noot: will add

lib/babe/babe_test.go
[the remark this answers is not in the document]
thomasmodeneis: oks

lib/blocktree/database.go
thomasmodeneis: well, they have different arguments, the internal one could be called decodeToParent since it has a parent pointer as argument. 
The same thing applies for  `(bt *BlockTree) Encode()` and `func encode`

noot: okay, I'll rename it `decodeRecursive` and `encodeRecursive`

PR #636, opened 2020-02-25
cmd, dot, node: use subdirectories for each node

ryanchristo:
## Changes

This pull request updates the default datadir to account for multiple node implementation.

## Tests

```
./bin/gossamer init --datadir ~/.gossamer/ksmcc
```

```
./bin/gossamer --datadir ~/.gossamer/ksmcc
```

### DIFF
--- a/cmd/gossamer/account.go
+++ b/cmd/gossamer/account.go
@@ -247,20 +247,17 @@ func generateKeypair(keytype, datadir string, password []byte) (string, error) {
 	return fp, nil
 }
 
-// keystoreDir returnns the absolute filepath of the keystore directory given gossamer's datadir
-// by default, it is ~/.gossamer/keystore/
-// otherwise, it is datadir/keystore/
+// keystoreDir returns the absolute filepath of the keystore directory
 func keystoreDir(datadir string) (keystorepath string, err error) {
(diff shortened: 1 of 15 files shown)


ryanchristo:
cc @noot

At this point, all you can really do is:

```
./bin/gossamer init --datadir ~/.gossamer/ksmcc
```

```
./bin/gossamer --datadir ~/.gossamer/ksmcc
```

I was planning to add a `--node` flag so that you could do something like a followup pull request:

```
./bin/gossamer init --node ksmcc
```

```
./bin/gossamer --node ksmcc
```

noot:
@ryanchristo that would be cool! so if i do `--datadir .gossamer` it'll still put everything in `.gossamer`?

ryanchristo:
correct, but you probably would not want to do that because the default is `~/.gossamer/gssmr` now so you would be throwing in node specifc files into the default "root" directory `~/.gossamer`

ryanchristo:
I was thinking about adding another directory to distinguish between "root" and "node" directories but I also wanted to keep this pull request small. I was going to explore that in a later pull request.

noot:
@ryanchristo okay, cool. sounds good to me

ryanchristo:
opened #645

PR #640, opened 2020-02-26
network: send block request based on confirmed peer status

### LINKED ISSUE #637: core: refactor required after merge #606
In order to better address comments on PR #606, I decided to raise this ticket and get it clarified here.

@ryanchristo said:
* Anything to do with syncing should be handled at the root level of the network service and anything in the status submodule should only be concerned with the status message exchange.

* The comments in status.go and status_test.go could be cleaned up with the comments related to syncing being moved to service.go but otherwise ok.

@noot said:
* we should probably just add an interface function into status that'll allow us to request blocks if needed, not sure how that'll work but let's open an issue for it. I think it's out of scope for this

@ryanchristo replied:
* We should not be adding anything to the status submodule. status.handleMessage should only be handling status messages, not block request messages.

1.
2.
3.
4.

## Your Environment
* Version used:
* Commit hash:
* Go Version:

ryanchristo:
## Changes

This pull request sends a block request message if needed based on the best block number for the latest status message from each confirmed peer.

## Tests:
```
go test ./network -v
```

### Issues:
closes #637

### DIFF
--- a/dot/network/service.go
+++ b/dot/network/service.go
@@ -232,7 +232,27 @@ func (s *Service) handleMessage(peer peer.ID, msg Message) {
 
 			// handle status message from peer with status submodule
 			s.status.handleMessage(peer, msg.(*StatusMessage))
+
 		}
+
+		// check if peer status confirmed
+		if s.status.confirmed(peer) {
+
+			// get latest block header from block state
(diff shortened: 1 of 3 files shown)


ryanchristo:
closing in favor of #646

PR #643, opened 2020-02-26
add keyring package; update gssmr genesis

### LINKED ISSUE #624: Keypairs for Gossamer Testnet
[ ] A set of keypairs (8-10) which can be used to spin up authority nodes for our testnet.
[ ] Add to the cli so that you can select a keypair easily

noot:
## Changes
- create keyring package which contains 8 constant sr25519 keypairs
- update gssmr/genesis.json to set these keys as authorities
- update core to use Alice as the default key if none is specified

## Tests:
```
go test ./lib/keyring/
```

to test running an authority node: (probably need to remove any existing datadir)
```
make gossamer
./bin/gossamer init
./bin/gossamer 
```

### Issues:
closes #624


dot/core/service.go
```diff
@@ -82,13 +83,15 @@ func NewService(cfg *Config) (*Service, error) {
 
 	keys := cfg.Keystore.Sr25519Keypairs()
 
-	// no validator keypair found, generate a new one
+	// TODO: move this to cli, allow user to specify which test key to use
```
noot: #644

PR #646, opened 2020-02-27
network: send block request based on confirmed peer status #637

thomasmodeneis:
## Changes
- send block request based on confirmed peer status #637
- 

## Tests:
```

```


dot/network/host_test.go
```diff
@@ -54,7 +56,7 @@ func TestConnect(t *testing.T) {
 		NoMdns:      true,
 	}
 
-	nodeB, _, _ := createTestService(t, configB)
+	nodeB, _, _ := createTestService(t, configB, blockState)
```
ryanchristo: two nodes should not be sharing the same blockState

dot/network/host_test.go
```diff
@@ -37,7 +38,8 @@ func TestConnect(t *testing.T) {
 		NoMdns:      true,
 	}
 
-	nodeA, _, _ := createTestService(t, configA)
+	blockState := newMockBlockState(big.NewInt(1))
```
ryanchristo: can we include making a new mock block state in the `createTestService` helper method instead of having to do this for each test within the network package?

dot/network/service.go
```diff
@@ -232,6 +237,51 @@ func (s *Service) handleMessage(peer peer.ID, msg Message) {
...
+					err := s.host.send(peer, blockRequest)
+					if err != nil {
+						log.Error("(*Service).handleMessage, failed to send blockRequest message to peer", "err", err)
+					} else {
+						log.Info("(*Service).handleMessage, sent blockRequest message to peer")
+					}
+				} else {
+					log.Debug("(*Service).handleMessage, nothing to do here", "latestHeader.Number", latestHeader.Number, "bestBlockNum", bestBlockNum)
+				}
+			}
```
ryanchristo: I was experimenting in my pull request with extracting this into its own method or submodule, which might be nice.

dot/network/service_test.go
```diff
@@ -44,12 +45,12 @@ var TestMessage = &BlockRequestMessage{
 var TestMessageTimeout = 3 * time.Second
 
 // helper method to create and start a new network service
-func createTestService(t *testing.T, cfg *Config) (node *Service, msgSend chan Message, msgRec chan Message) {
+func createTestService(t *testing.T, cfg *Config, blockState *MockBlockState) (node *Service, msgSend chan Message, msgRec chan Message) {
```
ryanchristo: as mentioned above, I think it would make more sense to not pass block state here but rather to create the mock block state within the `createTestService` method so that you do not have to duplicate it for every node and each node should have its own state, not share the same block state

ryanchristo:
I have now reassigned #637 to you as well.

ryanchristo:
tagging #615 here as well, which this does not close but would be the third related issue

dot/network/service_test.go
thomasmodeneis: correct, however I need to have a reference to the blockstate for the testcase and have the blockstate being created pragmatically, not always the same with same dummy values. So passing it as a argument seemed to be the most reasonable option, given we have to customize it anyways.

dot/network/host_test.go
thomasmodeneis: sure

dot/network/host_test.go
thomasmodeneis: sure, I created another method for the custom state then.

dot/network/service.go
thomasmodeneis: right

dot/network/service.go
```diff
@@ -232,6 +232,9 @@ func (s *Service) handleMessage(peer peer.ID, msg Message) {
 
 			// handle status message from peer with status submodule
 			s.status.handleMessage(peer, msg.(*StatusMessage))
+
+			s.sendBlockRequestMessage(peer, msg)
```
ryanchristo: maybe a comment here to explain `sendBlockRequestMessage`

dot/network/status.go
```diff
@@ -61,28 +61,28 @@ func (status *status) setHostMessage(msg Message) {
 // handleConn starts status processes upon connection
 func (status *status) handleConn(conn network.Conn) {
 	ctx := context.Background()
-	peer := conn.RemotePeer()
+	thispeer := conn.RemotePeer()
```
ryanchristo: is this necessary? I would prefer if we didn't change this but maybe I am misunderstanding something herre?

dot/network/status_test.go
```diff
@@ -109,5 +113,141 @@ func TestStatus(t *testing.T) {
...
 }
 
-//have a peer send a message status with a block ahead
-// check that the message returned is the expected
+// helper method to create and start a new network service
+func createTestServiceWithBlockState(t *testing.T, cfg *Config, blockState *MockBlockState) (node *Service, msgSend chan Message, msgRec chan Message) {
+	msgRec = make(chan Message)
+	msgSend = make(chan Message)
+
+	// same for all network tests use the createTestService helper method
```
ryanchristo: needs revisions after copy pasta

dot/network/status_test.go
```diff
@@ -109,5 +113,141 @@ func TestStatus(t *testing.T) {
...
+
+	configA := &Config{
+		DataDir:     dataDirA,
+		Port:        7001,
+		RandSeed:    1,
+		NoBootstrap: true,
+		NoMdns:      true,
+	}
+
+	blockState := newMockBlockState(big.NewInt(3))
```
ryanchristo: can we use `blockStateA` and `blockStateB` in this test to be consistent and clear we are not using the same state?

dot/network/status_test.go
```diff
@@ -109,5 +113,141 @@ func TestStatus(t *testing.T) {
...
+
+	configB := &Config{
+		DataDir:     dataDirB,
+		Port:        7002,
+		RandSeed:    2,
+		NoBootstrap: true,
+		NoMdns:      true,
+	}
+
+	blockState = newMockBlockState(big.NewInt(1))
```
ryanchristo: can we use `blockStateA` and `blockStateB` in this test to be consistent and clear we are not using the same state?

dot/network/status_test.go
```diff
@@ -109,5 +113,141 @@ func TestStatus(t *testing.T) {
...
+	if !nodeB.status.confirmed(nodeA.host.h.ID()) {
+		t.Error("node B did not confirm status of node A")
+	}
+
+	// get latest block header from block state
+	latestHeader := blockState.LatestHeader()
+	currentHash := blockState.LatestHeader().Hash()
+
+	// expected block request message
+	var ThisTestMessage = &BlockRequestMessage{
```
ryanchristo: cane we remove `var` and name `expected` or `expectedMessage`?

dot/network/status_test.go
```diff
@@ -109,5 +113,141 @@ func TestStatus(t *testing.T) {
...
+		EndBlockHash:  optional.NewHash(true, latestHeader.Hash()),
+		Direction:     1,
+		Max:           optional.NewUint32(false, 0),
+	}
+
+	select {
+	case msg := <-msgSendA:
+		require.NotNil(t, msg)
+
+		//assert correct cast
```
ryanchristo: spacing

dot/network/status_test.go
```diff
@@ -109,5 +113,141 @@ func TestStatus(t *testing.T) {
...
+		blockRequest, ok := msg.(*BlockRequestMessage)
+		require.True(t, ok)
+		require.NotNil(t, blockRequest)
+
+		//assign ID since its random
```
ryanchristo: spacing

dot/network/service.go
thomasmodeneis: sure

dot/network/status.go
thomasmodeneis: yes that is required, we cant have a variable that shadows the import with the same name.

dot/network/status_test.go
thomasmodeneis: yes

dot/network/status_test.go
thomasmodeneis: yes

dot/network/status_test.go
thomasmodeneis: yes

ryanchristo:
also, as mentioned in the call, rename `blockrequest.go` to sync.go`

dot/network/status_test.go
thomasmodeneis: roger

dot/network/status_test.go
thomasmodeneis: roger

dot/network/status_test.go
thomasmodeneis: thanks

dot/network/status.go
noot: maybe `remotePeer`? not a huge fan of `this`

dot/network/sync.go
```diff
@@ -0,0 +1,77 @@
...
+
+			// send block request message
+			err := s.host.send(peer, blockRequest)
+			if err != nil {
+				log.Error("(*Service).sendBlockRequestMessage, failed to send blockRequest message to peer", "err", err)
+			} else {
+				log.Trace("(*Service).sendBlockRequestMessage, sent blockRequest message to peer")
+			}
+		} else {
+			log.Debug("(*Service).sendBlockRequestMessage, nothing to do here", "latestHeader.Number", latestHeader.Number, "bestBlockNum", bestBlockNum)
```
noot: can remove this else statement

dot/network/sync.go
```diff
@@ -0,0 +1,77 @@
...
+
+		bestBlockNum := big.NewInt(int64(statusMessage.BestBlockNumber))
+
+		// check if peer block number is greater than host block number
+		if latestHeader.Number.Cmp(bestBlockNum) == -1 {
+
+			//generate random ID
+			s1 := rand.NewSource(uint64(time.Now().UnixNano()))
+			seed := rand.New(s1).Uint64()
+			randomID := mrand.New(mrand.NewSource(int64(seed))).Uint64()
```
noot: let's keep track of the IDs in network state

dot/network/sync.go
```diff
@@ -0,0 +1,77 @@
...
+
+			currentHash := s.cfg.BlockState.LatestHeader().Hash()
+
+			blockRequest := &BlockRequestMessage{
+				ID:            randomID, // random
+				RequestedData: 2,        // block body
```
noot: we also want the header as well as the body - change this to 3 (0010 & 0001) for header and body

dot/network/sync.go
```diff
@@ -0,0 +1,77 @@
...
+				StartingBlock: append([]byte{0}, currentHash[:]...),
```
noot: this doesn't need to be changed here, but we should create a package similar to `optional` for variadic types. I will open an issue for it

dot/network/sync.go
thomasmodeneis: ok

dot/network/status.go
thomasmodeneis: sure

dot/network/sync.go
thomasmodeneis: sure

dot/network/service.go
```diff
@@ -33,18 +33,19 @@ var _ services.Service = &Service{}
...
+	mdns              *mdns
+	status            *status
+	gossip            *gossip
+	msgRec            <-chan Message
+	msgSend           chan<- Message
+	noBootstrap       bool
+	noMdns            bool
+	noStatus          bool            // internal option
+	noGossip          bool            // internal option
+	requestedBlockIDs map[uint64]bool // track requested block id messages
```
noot: fine for now, but let's consolidate this with the requestedBlockIDs in core soon.

thomasmodeneis: sure, any ideas ?

noot: yes, move this to state.Network and add functions for reading/writing from the mapping to the NetworkState interface

ryanchristo: #653

ryanchristo:
cc @thomasmodeneis looks like some new failing tests after merge?

dot/network/service.go
thomasmodeneis: great

PR #647, opened 2020-02-27
state: fix persistence

noot:
## Changes
- persistence was broken probably by updating the DB, this fixes it by storing the latest block hash at a known key in the db

## Tests:
```
make gossamer
./bin/gossamer
```

stop and restart the node, block production should start off where it left off

### DIFF
--- a/dot/state/block.go
+++ b/dot/state/block.go
@@ -290,7 +290,12 @@ func (bs *BlockState) AddBlock(newBlock *types.Block) error {
 	if err != nil {
 		return err
 	}
+
 	hash := newBlock.Header.Hash()
+	err = bs.setLatestHeaderKey(hash)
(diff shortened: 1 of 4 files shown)


dot/state/block.go
```diff
@@ -302,6 +307,15 @@ func (bs *BlockState) AddBlock(newBlock *types.Block) error {
 	return err
 }
 
+func (bs *BlockState) setLatestHeaderKey(hash common.Hash) error {
+	err := bs.db.db.Put(common.LatestHeaderHashKey, hash[:])
+	if err != nil {
```
thomasmodeneis: well, if err is nil, return err will also return nil. so the check is redundant;

noot: updated

PR #648, opened 2020-02-27
cmd, core: add --key flag to specify test keypair

### LINKED ISSUE #644: add ability to use test keypairs from cli
since we have a test keyring, we should be able to specify a test key to use, eg `gossamer --key alice` or `gossamer --key bob` etc. this will load the keyring when the core service starts up and use it as a BABE key.

noot:
## Changes
- add --key flag to specify which keypair should be used
- update cmd to load keystore using keyring
- update core to error if it's an authority node but no keys are provided

## Tests:
```
make gossamer
./bin/gossamer --key bob --verbosity=debug
./bin/gossamer --key charlie --verbosity=debug
```
etc.
 
you will be able to see that the key used by BABE changes

### Issues:
closes #644


cmd/gossamer/configcmd.go
```diff
@@ -274,14 +322,14 @@ func createNetworkService(fig *dot.Config, gendata *genesis.Data, stateService *
 }
 
 // createCoreService creates the core service from the provided core configuration
-func createCoreService(coreConfig *core.Config) *core.Service {
+func createCoreService(coreConfig *core.Config) (*core.Service, error) {
 	coreService, err := core.NewService(coreConfig)
 	if err != nil {
 		log.Crit("Failed to create new core service", "err", err)
-		os.Exit(0)
+		return nil, err
```
thomasmodeneis: nice err return! :)

cmd/gossamer/configcmd.go
```diff
@@ -114,7 +114,12 @@ func makeNode(ctx *cli.Context) (*dot.Node, *dot.Config, error) {
 		MsgSend:          networkMsgRec,  // message channel from core service to network service
 		IsBabeAuthority:  currentConfig.Global.Authority,
 	}
-	coreSrvc := createCoreService(coreConfig)
+
+	coreSrvc, err := createCoreService(coreConfig)
+	if err != nil {
+		return nil, nil, fmt.Errorf("error creating core service: %s", err)
```
thomasmodeneis: nice err return! :)

Issue #582, opened 2020-01-30
Block Response messages 


amerameen:
When a block request message is received, a block response message is sent if you have the requested blocks.

Issue #596, opened 2020-02-06
state: block data fields should be stored separately


noot:
right now, BlockData is stored on its own. however this causes a few problems:
1. duplication of block header and body in DB
2. the fields are optionals; having a BlockData for a hash in the DB doesn't necessarily mean all the fields exist. when receiving a BlockResponseMessage, we might only want to set one field (eg. Justification) but we need to retrieve the entire BlockData and confirm that it isn't set already, when we could just directly do `blockState.SetJustification(hash, justification)`. this over complicates the code.

solution: create separate functions in the db for storing Receipt, MessageQueue, and Justification. (Header and Body are already covered with Get/SetBlock, however we might want to split this out into Header and Body so that Header isn't duplicated)

PR #650, opened 2020-02-28
core: Block Response messages  #582 

thomasmodeneis:
## Changes
- When a block request message is received, a block response message is sent if you have the requested blocks.
- 
- 

## Tests:
```

```

### Issues: #582


dot/core/service.go
```diff
@@ -365,8 +367,6 @@ func (s *Service) ProcessBlockAnnounceMessage(msg network.Message) error {
 		seed := rand.New(s1).Uint64()
 		randomID := mrand.New(mrand.NewSource(int64(seed))).Uint64()
 
-		currentHash := s.blockState.LatestHeader().Hash()
```
noot: not sure why this doesn't work any more? not a big deal though, as long as we have number or hash it's fine

dot/core/service.go
```diff
@@ -394,6 +394,29 @@ func (s *Service) ProcessBlockAnnounceMessage(msg network.Message) error {
 	return nil
 }
 
+// ProcessBlockRequestMessage processes a block request message, returning a block response message
+func (s *Service) ProcessBlockRequestMessage(msg network.Message) error {
+	blockRequest := msg.(*network.BlockRequestMessage)
+
+	latestBlockNum := s.blockState.LatestHeader().Number
+
+	if blockRequest.StartingBlock[0] == 1 {
```
noot: we need to also take in to account the `requestedData` of the block request, they could be asking for header, body, message queue, justification, etc. you should use the `blockState.GetBlockData()` function for retrieving the requested data.

also, right now the block data is stored all at one key in the db, which might make it more complicated to get one specific field. #596 might be relevant here

dot/core/service.go
```diff
@@ -380,7 +382,7 @@ func (s *Service) ProcessBlockAnnounceMessage(msg network.Message) error {
 		blockRequest := &network.BlockRequestMessage{
 			ID:            randomID, // random
 			RequestedData: 2,        // block body
-			StartingBlock: append([]byte{0}, currentHash[:]...),
+			StartingBlock: append([]byte{1}, s.blockState.LatestHeader().Number.Bytes()...),
```
noot: the number needs to be in 8-byte little endian format, so convert the number to int64 -> uint64 -> bytes by doing
```
buf := make([]byte, 8)
binary.LittleEndian.PutUint64(buf, uint64(number.Int64()))
```

why can't we use hash here?

dot/core/service.go
```diff
@@ -398,6 +400,59 @@ func (s *Service) ProcessBlockAnnounceMessage(msg network.Message) error {
...
+			return err
+		}
+	} else {
+		err = fmt.Errorf("(*Service).ProcessBlockRequestMessage blockRequest.StartingBlock not recognized")
+		log.Trace(err.Error())
+		return err
+	}
+
+	// for now we only care about body requests
+	if blockRequest.RequestedData == 1 && latestBlockNum.Cmp(msgStartingBlockNum) >= 0 {
```
noot: this should use a bitmap comparison to check the requested data, for example, 2, 6, 10, etc. are all requests for the block body

you should do `blockRequest.RequestedData & 1` to check for header request, blockRequest.RequestedData & 2` for body request, etc

dot/core/service.go
```diff
@@ -398,6 +400,59 @@ func (s *Service) ProcessBlockAnnounceMessage(msg network.Message) error {
...
+			log.Debug("(*Service).ProcessBlockRequestMessage failed to GetBlockData with StartingBlock", "StartingBlock", common.NewHash(blockRequest.StartingBlock[1:]))
+			return err
+		}
+	} else {
+		err = fmt.Errorf("(*Service).ProcessBlockRequestMessage blockRequest.StartingBlock not recognized")
+		log.Trace(err.Error())
+		return err
+	}
+
+	// for now we only care about body requests
```
noot: we should also care about block header requests

dot/core/service.go
```diff
@@ -398,6 +400,59 @@ func (s *Service) ProcessBlockAnnounceMessage(msg network.Message) error {
...
+
+	msgStartingBlockPrefix := blockRequest.StartingBlock[0]
+	msgStartingBlockBody := blockRequest.StartingBlock[1:]
+
+	var msgStartingBlockNum *big.Int
+	var msgBlockData *types.BlockData
+	var err error
+
+	if msgStartingBlockPrefix == 1 {
+		msgStartingBlockNum = big.NewInt(0).SetBytes(msgStartingBlockBody)
```
noot: same here, need to use binary.LittleEndian to convert the bytes to uint64 then to big.int

dot/core/service.go
```diff
@@ -398,6 +400,59 @@ func (s *Service) ProcessBlockAnnounceMessage(msg network.Message) error {
...
+
+	// for now we only care about body requests
+	if blockRequest.RequestedData == 1 && latestBlockNum.Cmp(msgStartingBlockNum) >= 0 {
+		// we have the block
+		// lets send it then
+
+		//generate random ID
+		s1 := rand.NewSource(uint64(time.Now().UnixNano()))
+		seed := rand.New(s1).Uint64()
+		randomID := mrand.New(mrand.NewSource(int64(seed))).Uint64()
```
noot: we are using the same ID from the request msg

dot/core/service.go
```diff
@@ -398,6 +400,59 @@ func (s *Service) ProcessBlockAnnounceMessage(msg network.Message) error {
...
+	} else if msgStartingBlockPrefix == 1 {
+
+		msgBlockData, err = s.blockState.GetBlockData(common.NewHash(msgStartingBlockBody))
+		if err != nil {
+			log.Debug("(*Service).ProcessBlockRequestMessage failed to GetBlockData with StartingBlock", "StartingBlock", common.NewHash(blockRequest.StartingBlock[1:]))
+			return err
+		}
+	} else {
+		err = fmt.Errorf("(*Service).ProcessBlockRequestMessage blockRequest.StartingBlock not recognized")
+		log.Trace(err.Error())
```
noot: logs should be prefixed with package name then variable,value pairs

PR #652, opened 2020-02-28
babe: refactor into multiple files

noot:
## Changes
- rename babe_slots.go to median.go, move slot and median related tests to median_test.go
- move block building functions into build.go, move related tests to build_test.go

## Tests:
```
go test ./lib/babe/
```

### Issues:
closes #540

PR #654, opened 2020-03-02
babe: re-add slot time algorithm, move arrivalTime out of block and into BlockState

noot:
# Changes
- fix slotTime algorithm to work with blocktree updates
- move arrivalTime out of block struct, save it in the BlockState when AddBlock is called
- note: this branches off of #652 

## Tests:
```
go test ./... -short
```

### Issues:
working towards #567 , but computeSlotForBlock isn't correct. will open an issue for that


dot/state/block.go
```diff
@@ -270,10 +298,19 @@ func (bs *BlockState) SetBlockData(blockData *types.BlockData) error {
 	return err
 }
 
-// AddBlock adds a block to the blocktree and the DB
 func (bs *BlockState) AddBlock(block *types.Block) error {
+	return bs.AddBlockWithArrivalTime(block, uint64(time.Now().Unix()))
+}
+
+// AddBlock adds a block to the blocktree and the DB
```
thomasmodeneis: should be AddBlockWithArrivalTime

dot/state/block.go
```diff
@@ -23,13 +23,15 @@ import (
...
 	"sync"
+	"time"
 
 	"github.com/ChainSafe/gossamer/dot/core/types"
 	babetypes "github.com/ChainSafe/gossamer/lib/babe/types"
 	"github.com/ChainSafe/gossamer/lib/blocktree"
 	"github.com/ChainSafe/gossamer/lib/common"
 	"github.com/ChainSafe/gossamer/lib/common/optional"
 	"github.com/ChainSafe/gossamer/lib/database"
+	// log "github.com/ChainSafe/log15"
```
thomasmodeneis: can remove comment ?

noot: done

dot/state/block.go
noot: fixed

PR #657, opened 2020-03-03
cmd: update gossamer logs and errors

ryanchristo:
## Changes
- update gossamer logs and errors

## Tests:
```
make gossamer
```
```
./bin/gossamer init
```
```
./bin/gossamer
```


cmd/gossamer/genesis.go
```diff
@@ -27,39 +27,31 @@ import (
...
 		return err
 	}
 
 	// read genesis file
 	genesisPath := getGenesisPath(ctx)
 
-	dataDir := utils.ExpandDir(currentConfig.Global.DataDir)
-	if ctx.String(DataDirFlag.Name) != "" {
-		dataDir = utils.ExpandDir(ctx.String(DataDirFlag.Name))
-	}
```
ryanchristo: already done in `getConfig`

cmd/gossamer/configcmd.go
```diff
@@ -247,18 +247,13 @@ func loadConfig(file string, config *dot.Config) error {
 	if err != nil {
 		return err
 	}
-
-	log.Debug("[gossamer] Loading toml configuration", "path", filepath.Clean(fp))
-
```
ryanchristo: duplicate

cmd/gossamer/configcmd_test.go
```diff
@@ -34,7 +34,6 @@ import (
 	"github.com/ChainSafe/gossamer/node/gssmr"
 	"github.com/ChainSafe/gossamer/tests"
 
-	log "github.com/ChainSafe/log15"
```
edwardmack: Why use fmt instead of log in this file?

cmd/gossamer/genesis.go
```diff
@@ -27,100 +27,114 @@ import (
...
+	log.Info(
+		"[gossamer] Initializing node...",
+		"datadir", datadir,
+		"genesis", genPath,
+	)
 
-	// read genesis configuration file
-	gen, err := genesis.LoadGenesisFromJSON(genesisPath)
+	// load Genesis from genesis configuration file
+	gen, err := genesis.LoadGenesisFromJSON(genPath)
```
edwardmack: genPath abbreviated

cmd/gossamer/configcmd_test.go
```diff
@@ -371,7 +370,7 @@ func TestMakeNode(t *testing.T) {
 			context, err := createCliContext(c.name, c.flags, c.values)
 			require.Nil(t, err)
 
-			err = loadGenesis(context)
+			err = initializeNode(context)
```
edwardmack: context not abbreviated

cmd/gossamer/genesis_test.go
```diff
@@ -117,12 +117,12 @@ func TestGenesisStateLoading(t *testing.T) {
 	set.String("config", tempFile.Name(), "TOML configuration file")
 	set.String("genesis", genesisPath, "path to genesis file")
 	set.Bool("authority", false, "")
-	context := cli.NewContext(nil, set, nil)
+	ctx := cli.NewContext(nil, set, nil)
 
-	err = loadGenesis(context)
+	err = initializeNode(ctx)
```
edwardmack: ctx abbreviated

cmd/gossamer/configcmd_test.go
ryanchristo: I am pretty sure we want to avoid using logs in tests. @noot pointed this out to me when I was attempting to update the format of these logs.

noot: yeah we should be using t.Log or fmt, doesn't really make sense to use the logger in tests

cmd/gossamer/configcmd_test.go
ryanchristo: will fix

cmd/gossamer/genesis.go
ryanchristo: my thinking was that we have `gen` already, so we should use `genPath` instead of `genesisPath`

ryanchristo: also, I like to keep variables as short and consistent as possible, saving the full name (ie, `genesis`) for the package name (ie, `genesis.Genesis`).

PR #663, opened 2020-03-03
lib/babe, dot/core: implement slot time estimation, start babe session from latest known slot time

### LINKED ISSUE #665: babe: keep track of slots, begin session with last known slot
since slots count up consecutively and don't reset at each epoch, we need to keep track of the current slot outside of babe, so we can begin the next epoch session at the correct slot.

noot:
## Changes
- add getSlotForBlock which gets the slot for a block from the babe header
- add estimateCurrentSlot which estimates the current slot based of the chain head, this will be used as input to the median algorithm 
- add startSlot to babe session which is the slot that the session should start at, since slots continue to increase and don't reset at 0 for each epoch

## Tests:
```
go test ./lib/babe/...
```

try getting nodes to sync slot times:
```
make gossamer
./bin/gossamer --key alice init
./bin/gossamer --key alice
```
let it run for a bit, create bob.toml with different port,datadir, then:
```
./bin/gossamer --key bob --config bob.toml init
./bin/gossamer --key bob --config bob.toml
```
after a few blocks, the slot numbers should sync up.

### Issues:
related to #567 closes #665


dot/state/block.go
```diff
@@ -350,11 +377,53 @@ func (bs *BlockState) BestBlockHeader() (*types.Header, error) {
...
+	return header.Number, nil
+}
+
 // BestBlock returns the current head of the chain
 func (bs *BlockState) BestBlock() (*types.Block, error) {
 	return bs.GetBlockByHash(bs.BestBlockHash())
 }
 
+// GetSlotForBlock returns the slot for a block
+func (bs *BlockState) GetSlotForBlock(hash common.Hash) (uint64, error) {
```
ryanchristo: can we reduce the number of return statements in `GetSlotForBlock`? Maybe extract decode logic?

lib/babe/babe.go
```diff
@@ -171,7 +171,25 @@ func (b *Session) invokeBlockAuthoring() {
...
-	for ; slotNum < b.config.EpochLength; slotNum++ {
+	slotNum := b.startSlot
+
+	bestNum := b.blockState.HighestBlockNumber()
+	log.Debug("[babe]", "highest block num", bestNum)
+
+	var err error
+	// check if we are starting at genesis, if not, need to calculate slot
+	if bestNum.Cmp(big.NewInt(0)) == 1 && slotNum == 0 {
+		// TODO: change this to getCurrentSlot, once BlockResponse messages are implemented
```
ryanchristo: blocked by #650, addressed in #665

lib/babe/babe.go
```diff
@@ -203,12 +221,16 @@ func (b *Session) handleSlot(slotNum uint64) {
 		number:   slotNum,
 	}
 
+	// TODO: block authorization check
```
ryanchristo: issue number?

lib/babe/babe.go
```diff
@@ -203,12 +221,16 @@ func (b *Session) handleSlot(slotNum uint64) {
...
+
 	block, err := b.buildBlock(parentHeader, currentSlot)
 	if err != nil {
 		log.Error("BABE block authoring", "error", err)
 	} else {
+		// TODO: loop until slot is done, attempt to produce multiple blocks
```
ryanchristo: issue number?

lib/babe/median.go
```diff
@@ -21,12 +21,66 @@ import (
...
 	"sort"
-
-	"github.com/ChainSafe/gossamer/lib/common"
+	"time"
 )
 
-// slotTime calculates the slot time in the form of miliseconds since the unix epoch
-// for a given slot in miliseconds, returns 0 and an error if it can't be calculated
+// slotTail is the number of blocks needed for us to run the median algorithm. in the spec, it's arbitrarily set to 1200.
+// TODO: will need to update this once finished simple slot time algo testing
```
ryanchristo: issue number?

lib/babe/median.go
```diff
@@ -21,12 +21,66 @@ import (
...
+var slotTail = uint64(12)
+
+// returns the estimated current slot number, without median algorithm
+func (b *Session) estimateCurrentSlot() (uint64, error) {
+	// estimate slot of highest block we've received
+	// TODO: change this to BestBlockHash. however this requires us to have at least `slotTail` number of blocks in our state
```
ryanchristo: issue number?

lib/babe/median_test.go
```diff
@@ -67,76 +65,171 @@ func TestSlotOffset(t *testing.T) {
...
 			t.Fatal(err)
 		}
 	}()
 
-	cfg := &SessionConfig{
-		BlockState:   dbSrv.Block,
-		StorageState: dbSrv.Storage,
-	}
+	// 100 blocks / 1000 ms/s
+	// TODO: use time.Duration
```
ryanchristo: addressed in #666

lib/babe/median_test.go
```diff
@@ -67,76 +65,171 @@ func TestSlotOffset(t *testing.T) {
...
-	addBlocksToState(t, 100, dbSrv.Block)
-
-	res, err := babesession.slotTime(103, 20)
+	res, err := babesession.getCurrentSlot()
 	if err != nil {
 		t.Fatal(err)
 	}
 
-	expected := uint64(104000)
+	expected := uint64(102)
```
ryanchristo: see review comment

ryanchristo:
cc @noot your test examples include a config file not in the pull request... can you share `bob.toml`?

lib/babe/babe.go
noot: #670

lib/babe/babe.go
noot: #670

lib/babe/median.go
noot: #664

lib/babe/median.go
noot: #664

dot/state/block.go
noot: will take a look!

noot:
@ryanchristo bob.toml is the same as the default config file but with a different port and datadir, I'd rather not commit it since it's not used anywhere

will take a look at the test, I noticed it's failing on travis too. it passes about 75% of the time for me. the issue with that test is that it's based off of the current time, and depending how long execution takes, it might return a different slot (+- 1 from expected)

dot/state/block.go
noot: I think it's okay as it is, there aren't any other functions in the state that would be using that functionality so I'm not sure it makes sense to break it out.

noot:
@ryanchristo I just added a 1/2 slot duration delay to the test and it seems to be working better now. please try again and let me know!

noot:
@edwardmack I was having the same results as you locally, wondering if I should change the test to accept 102 +- 1, what does everyone think?

ryanchristo:
@noot I think 103 +/- 1 would be ok but we might want to open an issue to investigate a better solution in the future but would prefer not to have to rerun tests now

Also, no need to commit `bob.toml`, I would prefer that not be committed as well, was just looking for you to paste in a comment but I can also figure out on my own.

noot:
@ryanchristo  I've updated the test to accept 101 or 102, since it might be either depending on execution time

Issue #546, opened 2020-01-18
Use SCALE instead of JSON in state package


amerameen:
Currently we use JSON to encode/decode data that we are storing/retrieving from the database. Update the Set/Get functions in the state package to use SCALE instead.

PR #668, opened 2020-03-04
dot/network, dot/state: use SCALE instead of JSON for encoding/decoding 

edwardmack:
This PR updates code to replace JSON marshal/unmarshal with SCALE encode/decode.

## Changes
- Remove code that uses JSON marshal/unmashal and replace with SCALE encode/decode
- Added test for encoding/decoding `*[]common.PeerInfo`
- 

## Tests:
```
go test ./dot/state/... -v
go test ./dot/network/... -v
```

### Issues:
closes: #546


dot/state/network_test.go
```diff
@@ -29,7 +31,9 @@ import (
 
 var testHealth = &common.Health{}
 var testNetworkState = &common.NetworkState{}
-var testPeers = &[]common.PeerInfo{}
+var testPeers = &[]common.PeerInfo{{PeerID: "ID1", BestHash: common.Hash{}, BestNumber: 1, ProtocolVersion: 2, Roles: 0x03},
+	{PeerID: "ID40", BestHash: common.Hash{}, BestNumber: 50, ProtocolVersion: 60, Roles: 0x70},
+}
```
ryanchristo: maybe update `PeerID` for each peer to use the peer ids of `alice` and `bob` but not necessary

lib/scale/decode_ptr.go
```diff
@@ -70,6 +72,7 @@ func (sd *Decoder) DecodeCustom(t interface{}) (interface{}, error) {
 			return nil, err.(error)
 		}
 		t = res[0].Interface()
+		log.Trace("[DecodeCustom]", "decoded", t)
```
ryanchristo: I think this should be `[scale]` here for the log

dot/state/block.go
```diff
@@ -468,19 +479,19 @@ func (bs *BlockState) GetBabeHeader(epoch uint64, slot uint64) (*babetypes.BabeH
...
 func (bs *BlockState) SetBabeHeader(epoch uint64, slot uint64, bh *babetypes.BabeHeader) error {
 	// Write the encoded header
-	enc, err := json.Marshal(bh)
-	if err != nil {
-		return err
-	}
+	enc := bh.Encode()
 
-	err = bs.db.Put(babeHeaderKey(epoch, slot), enc)
+	err := bs.db.Put(babeHeaderKey(epoch, slot), enc)
```
noot: could turn this all into one line

dot/state/block.go
```diff
@@ -468,19 +479,19 @@ func (bs *BlockState) GetBabeHeader(epoch uint64, slot uint64) (*babetypes.BabeH
 		return nil, err
 	}
 
-	err = json.Unmarshal(data, result)
+	err = result.Decode(data)
+	if err != nil {
+		return nil, err
+	}
 
 	return result, err
```
noot: don't need to return err here anymore

lib/scale/decode.go
```diff
@@ -342,6 +342,12 @@ func (sd *Decoder) DecodeArray(t interface{}) (interface{}, error) {
 				copy(arr[:], buf)
 				*ptr = arr
 			}
+		case *common.PeerInfo:
```
noot: was this the one way to do this? I guess it's blocked on the "support arrays of structs" issue?

lib/scale/decode_ptr.go
```diff
@@ -70,6 +72,7 @@ func (sd *Decoder) DecodeCustom(t interface{}) (interface{}, error) {
 			return nil, err.(error)
 		}
 		t = res[0].Interface()
+		log.Trace("[scale]", "decoded", t)
```
noot: not sure about this log

lib/scale/encode.go
```diff
@@ -329,6 +329,14 @@ func (se *Encoder) encodeArray(t interface{}) (bytesEncoded int, err error) {
 			n, err = se.encodeByteArray([]byte(elem))
 			bytesEncoded += n
 		}
+	case []common.PeerInfo:
```
noot: same for here, I guess there isn't encoding support for arrays of structs either?

ryanchristo:
I agree, if there is a quick fix here, let's do it, otherwise we should probably open a separate issue to address using a common type in scale

lib/scale/decode.go
edwardmack: right, I'll start working on that issue.

edwardmack:
I agree, I'll work on the decode arrays of structs issue.

PR #677, opened 2020-03-06
network: format logging and errors

ryanchristo:
## Changes

- format logging and errors throughout the network package

## Tests:

```
make gossamer
./bin/gossamer --key alice --datadir tmp/alice --verbosity trace
```

### DIFF
--- a/cmd/gossamer/main.go
+++ b/cmd/gossamer/main.go
@@ -138,6 +138,8 @@ func gossamer(ctx *cli.Context) error {
 	// check if node has been initialized
 	if !nodeInitialized(ctx) {
 
+		log.Warn("[gossamer] Node has not been initialized, initializing new node...")
+
 		// initialize node (initialize databases and load genesis data)
 		err = initializeNode(ctx)
 		if err != nil {
--- a/cmd/gossamer/main.go
+++ b/cmd/gossamer/main.go
@@ -174,7 +174,11 @@ func nodeInitialized(ctx *cli.Context) bool {
 	registry := path.Join(cfg.Global.DataDir, "KEYREGISTRY")
 	_, err = os.Stat(registry)
 	if os.IsNotExist(err) {
-		log.Error("[gossamer] Failed to locate KEYREGISTRY", "error", err)
+		log.Warn(
+			"[gossamer] Node has not been initialized",
+			"datadir", cfg.Global.DataDir,
+			"error", "failed to locate KEYREGISTRY file in data directory",
+		)
 		return false
 	}
 
@@ -182,7 +186,11 @@ func nodeInitialized(ctx *cli.Context) bool {
 	manifest := path.Join(cfg.Global.DataDir, "MANIFEST")
 	_, err = os.Stat(manifest)
 	if os.IsNotExist(err) {
-		log.Error("[gossamer] Failed to locate MANIFEST", "error", err)
+		log.Warn(
+			"[gossamer] Node has not been initialized",
+			"datadir", cfg.Global.DataDir,
+			"error", "failed to locate MANIFEST file in data directory",
+		)
 		return false
 	}
 
--- a/dot/network/config.go
+++ b/dot/network/config.go
@@ -71,7 +71,7 @@ type Config struct {
 	// Disables bootstrapping
 	NoBootstrap bool
 	// Disables MDNS discovery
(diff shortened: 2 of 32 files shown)


dot/network/host.go
```diff
@@ -128,14 +128,6 @@ func (h *host) registerStreamHandler(handler func(libp2pnetwork.Stream)) {
 	h.h.SetStreamHandler(h.protocolID, handler)
 }
 
-// printHostAddresses prints host multiaddresses to console
-func (h *host) printHostAddresses() {
-	fmt.Println("Listening on the following addresses...")
-	for _, addr := range h.multiaddrs() {
-		fmt.Println(addr)
-	}
-}
```
ryanchristo: updated to info log in `network.Start()`

cmd/gossamer/main.go
```diff
@@ -174,15 +174,23 @@ func nodeInitialized(ctx *cli.Context) bool {
 	registry := path.Join(cfg.Global.DataDir, "KEYREGISTRY")
 	_, err = os.Stat(registry)
 	if os.IsNotExist(err) {
-		log.Error("[gossamer] Failed to locate KEYREGISTRY", "error", err)
+		log.Warn(
+			"[gossamer] Node has not been initialized",
+			"datadir", cfg.Global.DataDir,
+			"error", "failed to locate KEYREGISTRY file in data directory",
+		)
```
edwardmack: Shouldn't we also log the value of err in the log?

ryanchristo: We could but the error is "stat [datadir]" and this was cleaner. I was thinking that maybe since its a warning message for the user it would be nice to clean it up.

dot/network/status.go
```diff
@@ -123,13 +123,13 @@ func (status *status) validMessage(msg *StatusMessage) bool {
...
-		log.Error("Failed to validate status message", "err", "genesis hash")
+		log.Error("[network] Failed to validate status message", "error", "genesis hash")
 		return false
 	case msg.ProtocolVersion < status.hostMessage.MinSupportedVersion:
-		log.Error("Failed to validate status message", "err", "protocol version")
+		log.Error("[network] Failed to validate status message", "error", "protocol version")
 		return false
 	case msg.MinSupportedVersion > status.hostMessage.ProtocolVersion:
-		log.Error("Failed to validate status message", "err", "protocol version")
+		log.Error("[network] Failed to validate status message", "error", "protocol version")
```
edwardmack: Shouldn't there be a different error message for each condition?

ryanchristo: yea, you're probably right, will fix

cmd/gossamer/main.go
edwardmack: makes sense, I wasn't sure if the `err`  contained more context regarding the error.

cmd/gossamer/main.go
```diff
@@ -174,15 +174,23 @@ func nodeInitialized(ctx *cli.Context) bool {
...
 		return false
 	}
 
 	// check if manifest exists
 	manifest := path.Join(cfg.Global.DataDir, "MANIFEST")
 	_, err = os.Stat(manifest)
 	if os.IsNotExist(err) {
-		log.Error("[gossamer] Failed to locate MANIFEST", "error", err)
+		log.Warn(
```
noot: should we be warning if the node isn't initialized? or rather, if the node isn't initialized and we warn, we should also log that we initialized it for them

dot/network/config.go
```diff
@@ -18,63 +18,67 @@ package network
...
 const DefaultRoles = byte(1)
 
-// DefaultBootnodes holds the default bootnodes the client will always have
+// DefaultBootnodes the default value for Config.Bootnodes
 var DefaultBootnodes = []string(nil)
 
 // Config is used to configure a network service
 type Config struct {
-	// BlockState interface
+	// BlockState the block state interface
```
noot: adding some `is`'s to these comments would be good

dot/network/status.go
noot: agree, these should be more descriptive like `protocol version mismatch` or something

cmd/gossamer/main.go
ryanchristo: the warning is warning them that we are initializing a node for them, I can clarify in warning

PR #682, opened 2020-03-07
core, babe, state: refactor tests to use in-memory db

### LINKED ISSUE #639: refactor tests to use in-memory database
we should refactor tests to use an in-memory database when possible to reduce badger DB memory consumption overhead.

noot:
## Changes
- update state, core, babe tests to use in-memory db

## Tests:
```
go test ./... -short
```

### Issues:
closes #639


lib/database/memoryDB.go
```diff
@@ -61,7 +61,7 @@ func (db *MemDatabase) Get(k []byte) ([]byte, error) {
 	if v, ok := db.db[string(k)]; ok {
 		return v, nil
 	}
-	return nil, errors.New("not found")
+	return nil, errors.New("Key not found")
```
ryanchristo: standard error format starts with lowercase :eyes:

lib/babe/median_test.go
```diff
@@ -207,25 +191,18 @@ func TestEstimateCurrentSlot(t *testing.T) {
...
-	addBlocksToState(t, babesession, 100, dbSrv.Block, uint64(time.Now().Unix())-(babesession.config.SlotDuration/10))
+	addBlocksToState(t, babesession, 100, babesession.blockState, uint64(time.Now().Unix())-(babesession.config.SlotDuration/10))
 
 	res, err := babesession.getCurrentSlot()
 	if err != nil {
 		t.Fatal(err)
 	}
 
-	expected := uint64(101)
+	expected := uint64(100)
```
ryanchristo: so the memory database resolved this as well? nice!

lib/babe/babe_test.go
```diff
@@ -73,76 +72,29 @@ func createTestSession(t *testing.T, cfg *SessionConfig) *Session {
...
-		cfg.Runtime = rt
-	}
-
-	var err error
-	if cfg.Keypair == nil {
-		cfg.Keypair, err = sr25519.GenerateKeypair()
+	if cfg.BlockState == nil || cfg.StorageState == nil {
+		dbSrv := state.NewService("")
+		dbSrv.UseMemDB()
+		err = dbSrv.Initialize(genesisHeader, trie.NewEmptyTrie(nil))
```
ryanchristo: this is a really nice improvement :100:

dot/state/service.go
```diff
@@ -112,23 +132,22 @@ func initializeBlockTree(db database.Database, genesisHeader *types.Header) erro
 
 // Start initializes the Storage database and the Block database.
 func (s *Service) Start() error {
-	if s.Storage != nil || s.Block != nil || s.Network != nil {
-		return nil
-	}
+	db := s.db
+	if !s.isMemDB {
```
ryanchristo: is this being used for testing purposes only at the moment? should we make note in a comment here?

dot/state/service.go
```diff
@@ -44,25 +45,38 @@ func NewService(path string) *Service {
...
-	if err != nil {
-		return err
-	}
-
-	// initialize database
-	db, err := database.NewBadgerDB(datadir)
-	if err != nil {
-		return err
+	var db database.Database
+	if s.isMemDB {
```
ryanchristo: is this being used for testing purposes only at the moment? should we make note in a comment here?

lib/database/memoryDB.go
noot: this is done to match the badger DB error that's returned

lib/babe/median_test.go
noot: yeah, it sped up the gets from the DB

dot/state/service.go
noot: yes, it's only for testing purposes. I will make a note

lib/database/memoryDB.go
ryanchristo: maybe a comment so someone is not tempted to update it in the future

Issue #653, opened 2020-02-28
track our BlockRequestMessage IDs in NetworkState


noot:
currently, the message IDs for BlockRequestMessages we create are being tracked in core as well as network; these should all be stored inside the NetworkState.

PR #683, opened 2020-03-10
core, babe, network: safe channel handling

### LINKED ISSUE #679: politely and safely close channels
when services get stopped, (core, babe, network), the channels they use get closed. however, this may cause a `panic: send to close channel` if the goroutines using the channel aren't notified.

we should use a mutex to safely close the channels. see: https://go101.org/article/channel-closing.html

the mutex is locked before the channel is closed, we set `service.stopped` to true, then unlock. inside the goroutine that uses the channel, we should lock, check if service.stopped is true, and if so, unlock and return. if it's not stopped, send to the channel, then unlock.

noot:
## Changes
- add lock to service structs and lock before a channel is closed, or before a channel that could possibly be closed is written to, unlock after

## Tests:
```
go test ./... -short
```

### Issues:
closes #679


dot/core/service.go
```diff
@@ -47,23 +48,36 @@ var _ services.Service = &Service{}
...
+	// Channels for inter-process communication
+	msgRec    <-chan network.Message // receive messages from network service
+	msgSend   chan<- network.Message // send messages to network service
+	blkRec    <-chan types.Block     // receive blocks from BABE session
+	epochDone <-chan struct{}        // receive from this channel when BABE epoch changes
+	babeKill  chan<- struct{}        // close this channel to kill current BABE session
+	lock      sync.Mutex
+	closed    bool
+
+	// TODO: add to network state
```
ryanchristo: #653

dot/core/service_test.go
```diff
@@ -256,12 +256,6 @@ func TestProcessBlockResponseMessage(t *testing.T) {
 		}
 	}()
 
-	err = s.Start()
-	if err != nil {
-		t.Fatal(err)
-	}
-	defer s.Stop()
-
```
ryanchristo: the safe message send really cleans up the tests :+1:

dot/network/service.go
```diff
@@ -33,18 +35,26 @@ var _ services.Service = &Service{}
...
+	mdns   *mdns
+	status *status
+	gossip *gossip
+
+	// Channels for inter-process communication
+	// as well as a lock for safe channel closures
+	msgRec  <-chan Message
+	msgSend chan<- Message
+	lock    sync.Mutex
+	closed  bool
```
ryanchristo: I wonder if we can clean up the properties of each service using safe message send and receive (`core` and `network`) packagees within a separate lib submodule that both each service uses rather than the safe send and receive being a method appended to each service...

noot:
@ryanchristo not totally sure what you mean, the `safe` functions are channel-specific, also, if we moved the channels to a different package, that kind of defeats the purpose of having them, since we would need another channel/interface to push the information to that package. maybe I'm not understanding what you mean?

noot:
@ryanchristo ok lol no worries maybe we can discuss it later, if we end up adding more channels it might be a good idea to figure something out

PR #692, opened 2020-03-11
babe, state: use slotTime algorithm

### LINKED ISSUE #567: Test Median Algorithm

### LINKED ISSUE #664: babe: wait until enough blocks are synced before calculating slot time
need to have at least slotTail blocks in our blocktree before calculating slot

noot:
## Changes
- update babe to use getCurrentSlot which runs the actual slotTime algorithm
- update state logic to save block by number only for blocks in the current canonical chain

## Tests:
```
go test ./... -short
```

### Issues:
closes #664 closes #567


dot/state/block_test.go
```diff
@@ -72,45 +72,31 @@ func TestSetAndGetHeader(t *testing.T) {
...
+	// AddBlock also sets mapping [blockNumber : hash] in DB
+	err := bs.AddBlock(block)
 	require.Nil(t, err)
 
-	retBlock.Header.Hash()
+	retBlock, err := bs.GetBlockByNumber(blockHeader.Number)
+	require.Nil(t, err)
 
-	require.Equal(t, expectedBlock, retBlock, "Could not validate returned retBlock as expected")
+	require.Equal(t, block, retBlock, "Could not validate returned retBlock as expected")
```
ryanchristo: having expected in the name here is sometimes nice given the param is `expected` but not necessary to revert

PR #693, opened 2020-03-11
core, network: store requested block ids

### LINKED ISSUE #653: track our BlockRequestMessage IDs in NetworkState

ryanchristo:
## Changes

- store requested block ids in `syncer` submodule and handle storage in `network` service

## Tests:

```
go test ./dot/network
```

### Issues:

closes #653

### DIFF
--- a/dot/core/service.go
+++ b/dot/core/service.go
@@ -417,6 +417,8 @@ func (s *Service) handleReceivedMessage(msg network.Message) (err error) {
 // announce messages (block announce messages include the header but the full
 // block is required to execute `core_execute_block`).
 func (s *Service) ProcessBlockAnnounceMessage(msg network.Message) error {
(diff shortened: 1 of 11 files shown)


dot/state/network.go
```diff
@@ -133,3 +133,20 @@ func (ns *NetworkState) SetPeers(peers *[]common.PeerInfo) error {
 	err = ns.db.Put(peersKey, enc)
 	return err
 }
+
+// GetRequestedBlock retrieves network state from the database
+func (ns *NetworkState) GetRequestedBlock(blockId string) (bool, error) {
```
noot: this isn't something we need to save in the database, it's fine if it gets cleared out upon restarts. what do you think?

PR #695, opened 2020-03-13
network, state: remove network state db

### LINKED ISSUE #619: state: update network state to pull from network package
right now, the network state sets and gets Health, Peers, NetworkState from the DB. however, this requires the network package to write to the DB every time there is an update for any of these fields. instead it should be pulling the information it needs from the network package whenever one of its Get functions is called. 
the only package that uses this info is RPC, so there's no reason to store it in the DB.

ryanchristo:
## Changes
- remove use of database from network state
- update network state every minute (or other set time)

## Tests:

```
go test dot/network -v
```

### Issues:
closes #619

### DIFF
--- a/dot/network/service.go
+++ b/dot/network/service.go
@@ -21,6 +21,7 @@ import (
 	"context"
 	"errors"
 	"sync"
+	"time"
 
 	"github.com/ChainSafe/gossamer/lib/common"
 	"github.com/ChainSafe/gossamer/lib/services"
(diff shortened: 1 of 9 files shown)


ryanchristo:
I agree. I'm still not completely sure of the usefulness of network state but I don't think we'll know until we pick work back up on the rpc service. This just removes the database because we know we won't need that, or at least we know we want to avoid it for now. :+1:

noot:
@ryanchristo sounds good to me :)

dot/network/service.go
```diff
@@ -321,17 +342,17 @@ func (s *Service) handleMessage(peer peer.ID, msg Message) {
 }
 
 // Health returns information about host needed for the rpc server
-func (s *Service) Health() *common.Health {
-	return &common.Health{
+func (s *Service) Health() common.Health {
+	return common.Health{
 		Peers:           s.host.peerCount(),
 		IsSyncing:       false, // TODO
```
ryanchristo: #700

PR #696, opened 2020-03-13
core, common: implement variadic types package #651

### LINKED ISSUE #651: implement variadic types package
similar to the optional package, we should implement a variadic package to deal with variadic types.
this will be needed for the StartingBlock field in BlockRequestMessage which can be a 32-byte hash or 8 byte integer.

it should implement Encode and Decode for its respective types

thomasmodeneis:
## Changes

- implement variadic types package

## Tests:

```
lib/common/variadic/startingBlock_test.go
```

### Issues:

closes #651


lib/common/variadic/startingBlock.go
```diff
@@ -0,0 +1,55 @@
...
+package variadic
+
+import (
+	"encoding/binary"
+	"errors"
+	"github.com/ChainSafe/gossamer/lib/common"
+)
+
+// StartingBlock represents an optional interface type (int,hash).
+type StartingBlock struct {
```
ryanchristo: Does this need to be specific to "starting" block or can this be more generic?

lib/common/variadic/startingBlock.go
```diff
@@ -0,0 +1,55 @@
...
+	value interface{}
+}
+
+// NewStartingBlock returns a new optional.Uint32
+func NewStartingBlock(startingBlockMsg []byte) (*StartingBlock, error) {
```
ryanchristo: we should probably avoid using `Msg` here because this is within the generic `lib/common` package and this is taking `[]byte` not a network message type

lib/common/variadic/startingBlock_test.go
```diff
@@ -0,0 +1,67 @@
...
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with the gossamer library. If not, see <http://www.gnu.org/licenses/>.
+
+package variadic
+
+import (
+	"encoding/binary"
+	"github.com/ChainSafe/gossamer/dot/network"
+	"github.com/ChainSafe/gossamer/lib/common"
```
ryanchristo: I'm wondering if this test or part of this test belongs in `dot/core` given the imports you have here. The `lib/common` package should not be importing anything from `dot/...`.

lib/common/variadic/startingBlock_test.go
```diff
@@ -0,0 +1,67 @@
...
+	"github.com/stretchr/testify/require"
+	"testing"
+)
+
+func TestBlockRequestStartingBlock(t *testing.T) {
```
ryanchristo: same as above, does this need to be specific to "starting" block and does part of this test belong in `dot/core`?

lib/common/variadic/startingBlock.go
noot: I think this should be named `Uint64OrHash` or something along those lines

lib/common/variadic/startingBlock.go
noot: agree, could just call it `in` or `data` or something, not necessarily a msg

lib/common/variadic/startingBlock_test.go
```diff
@@ -0,0 +1,67 @@
...
+		},
+		{
+			description:     "block request with Block Number int type 1",
+			targetHash:      buf,
+			targetFirstByte: 1,
+			expectedType:    (uint64)(0),
+		},
+	} {
+		t.Run(x.description, func(t *testing.T) {
+			msg := network.BlockRequestMessage{
```
noot: don't need to use `network.Message` here, just test it with a byte array

lib/common/variadic/startingBlock_test.go
noot: yeah this shouldn't import dot/network

lib/common/variadic/startingBlock_test.go
thomasmodeneis: done

lib/common/variadic/startingBlock_test.go
thomasmodeneis: done

lib/common/variadic/startingBlock_test.go
thomasmodeneis: done

lib/common/variadic/startingBlock.go
thomasmodeneis: done

lib/common/variadic/startingBlock.go
thomasmodeneis: done

thomasmodeneis:
> could you also change BlockRequestMessage.StartingBlock to have type `variadic.NewUint64OrHash`, and also implement `Encode() ([]byte, error)` and `Decode(r io.Reader) error` for `variadic.NewUint64OrHash`? (so that SCALE will work on BlockRequestMessages) otherwise, looks great!

done

PR #697, opened 2020-03-13
babe: descriptive error codes

### LINKED ISSUE #504: BABE: determine when block is full

### LINKED ISSUE #514: BABE: descriptive error codes
building a block may fail when doing apply extrinsic with various error codes. we should log what the error is. see spec G.2.8

edwardmack:
## Changes
- Added function to build descriptive error string for errors generated during BlockBuilder_apply_extrinsic.  String includes error types as defined in PRE Spec G.2.8
- 
- 

## Tests:
```
go test ./lib/babe
```

### Issues:
closes #514, closes #504


lib/babe/build_test.go
```diff
@@ -236,6 +237,8 @@ func TestBuildBlock_failing(t *testing.T) {
 	if err == nil {
 		t.Fatal("should error when attempting to include invalid tx")
 	}
+	require.Equal(t, "cannot build extrinsics: Error during apply extrinsic: Apply error, type: Payment",
+		err.Error(), "Did not recieve expected error text")
```
edwardmack: Should I try to create more tests for failing apply extrinsic to test for different failure types?

lib/babe/build.go
```diff
@@ -246,3 +248,63 @@ func extrinsicsToBody(txs []*transaction.ValidTransaction) (*types.Body, error)
 
 	return types.NewBodyFromExtrinsics(extrinsics)
 }
+
+func determineError(res []byte) string {
```
edwardmack: I'm just building an error text string to include the error type in error messages.  Do you think I should create a struct for these errors to include error type, error description text, and module/error codes for dispatch errors?

lib/babe/build.go
```diff
@@ -246,3 +248,63 @@ func extrinsicsToBody(txs []*transaction.ValidTransaction) (*types.Body, error)
...
+		}
+		if bytes.Equal(res[1:], []byte{2, 0, 4}) {
+			errTxt.WriteString("BadProof")
+		}
+		if bytes.Equal(res[1:], []byte{2, 0, 5}) {
+			errTxt.WriteString("AncientBirthBlock")
+		}
+		if bytes.Equal(res[1:], []byte{2, 0, 6}) {
+			errTxt.WriteString("ExhaustsResources")
+		}
```
edwardmack: Is this check enough to close #504?  I've determined that the block is full, but do I need to do more, or is it enough just to include this information in error text?

lib/babe/build.go
```diff
@@ -246,3 +248,63 @@ func extrinsicsToBody(txs []*transaction.ValidTransaction) (*types.Body, error)
...
+		if bytes.Equal(res[1:], []byte{2, 1, 0}) {
+			errTxt.WriteString("CannotLookup")
+		}
+		if bytes.Equal(res[1:], []byte{2, 1, 1}) {
+			errTxt.WriteString("NoUnsignedValidator")
+		}
+		if bytes.Equal(res[1:], []byte{2, 1, 2}) {
+			errTxt.WriteString("Custom")
+		}
```
edwardmack: I wanted to do all this in a switch, but I was getting `invalid case []byte literal in switch` errors, so I'm using if statements, is this too ugly?  Any suggestions?

lib/babe/build.go
```diff
@@ -166,19 +167,20 @@ func (b *Session) buildBlockExtrinsics(slot Slot) ([]*transaction.ValidTransacti
 			return nil, err
 		}
 
-		// if ret == 0x0001, there is a dispatch error; if ret == 0x01, there is an apply error
-		if ret[0] == 1 || bytes.Equal(ret[:2], []byte{0, 1}) {
-			// TODO: specific error code checking
-			log.Error("[babe] build block apply extrinsic", "error", ret, "extrinsic", extrinsic)
+		// if ret != 0 there was an error
+		if ret[0] != byte(0) {
```
noot: does this still work for the  ret == 0x0001 case?

lib/babe/build.go
noot: seems fine to me, I'm good with the error message + code

lib/babe/build.go
noot: I think this is good enough

lib/babe/build.go
noot: I don't mind it, if a switch isn't working there isn't much you can do

ryanchristo: same as above, this works for me as is but if you see a potential optimization, wouldn't hurt to open an issue

lib/babe/build_test.go
ryanchristo: it wouldn't hurt to throw a couple more in :+1:

lib/babe/build.go
edwardmack: the 0x0001 case is handled at line 307.

noot: but it won't enter this statement and `determineError` unless the first byte != 0, but for the 0x0001 case, the first byte is 0, so it wouldn't enter here

edwardmack: Ok, right, I've updated this.

PR #699, opened 2020-03-13
core, network, babe: add syncer, babe locking

### LINKED ISSUE #690: don't build blocks while syncing
while syncing blocks initially via BlockRequest/BlockResponse, a node shouldn't be running a babe session and building blocks.

noot:
## Changes
- add Syncer to core, has a channel from network/core to watch for peers with higher block numbers
- if a higher block number than our current head is seen, we begin syncing; hold lock so that BABE stalls, send block request, unlock once peer's highest block number matches our highest block number

## Tests:
```
go test ./... -short
```

try it out with 2 authority nodes:
```
make gossamer
./bin/gossamer --key alice  init
./bin/gossamer --key alice 
```
another terminal:
```
./bin/gossamer --key bob --datadir ./bob/ init
./bin/gossamer --key bob --datadir ./bob/ --port 7002 --authority=true init
```

### Issues:
closes #690


dot/core/sync.go
```diff
@@ -0,0 +1,143 @@
...
+		return err
+	}
+
+	//generate random ID
+	s1 := rand.NewSource(uint64(time.Now().UnixNano()))
+	seed := rand.New(s1).Uint64()
+	randomID := mrand.New(mrand.NewSource(int64(seed))).Uint64()
+
+	// TODO: can't request from /our/ best block number, need to start requesting from the best block num we have of /theirs/
+	// otherwise there's a chance we might build a block, then miss a block of theirs, causing error="cannot find parent block in blocktree"
```
noot: #706

lib/blocktree/blocktree_test.go
```diff
@@ -169,28 +169,28 @@ func TestBlockTree_Subchain(t *testing.T) {
 
 // TODO: Need to define leftmost (see BlockTree.LongestPath)
```
noot: #704

dot/core/sync.go
```diff
@@ -0,0 +1,143 @@
...
+}
+
+// Start begins the syncer
+func (s *Syncer) Start() {
+	go s.watchForBlocks()
+}
+
+func (s *Syncer) watchForBlocks() {
+	for {
+		peerNum := <-s.blockNumberIn
```
ryanchristo: `peerNum` ?

dot/network/service.go
```diff
@@ -97,6 +106,7 @@ func NewService(cfg *Config, msgSend chan<- Message, msgRec <-chan Message) (*Se
 		noBootstrap:  cfg.NoBootstrap,
 		noMDNS:       cfg.NoMDNS,
 		noStatus:     cfg.NoStatus,
+		syncChan:     cfg.SyncChan,
```
ryanchristo: does this need to be appended here if we can already use `syncer.synChan` ?

dot/node.go
```diff
@@ -156,10 +158,12 @@ func NewNode(cfg *Config, ks *keystore.Keystore) (*Node, error) {
 
 	// TODO: Configure node based on Roles #601
 
+	syncChan := make(chan *big.Int, 128)
```
ryanchristo: should `syncChan`, `networkMsgSend`, and `networkMsgRec` all be appended to configuration to reduce the number of params going into `createNetworkService` and `createCoreService` or should all channels be created at the node level and passed in as parameters? See https://github.com/ChainSafe/gossamer/pull/701#discussion_r393297573

noot: I think it would be nicer to put them into the config so we don't need to change function sigs

ryanchristo: Either way works for me, if you add `syncChan` to config here then I will add `coreMsgs` and `networkMsgs` to config in #701.

noot: will do!

noot: it's a bit of a different situation here than in NewService though, this isn't passing it into NewService, it's passing it into a function that creates the core config

noot: and the config that createCoreService takes is the cli config not the core config

ryanchristo: which is why it makes more sense to me to declare all three channels at the same time when we are creating a new node and then to pass them as channels, which would mean keeping it the way you have here and the way I have in #701... I might be confused?

ryanchristo: ok, I figured it out, this is fine as is, I will fix in #701

dot/network/service.go
noot: hmm.. yeah I think I can remove it

dot/core/sync.go
noot: it's the peer's block num, will update it to something better

PR #701, opened 2020-03-14
dot: configure services based on roles

### LINKED ISSUE #601: Configure node based on roles
## Expected Behavior

A gossamer node should only run services it needs based on the value of `Roles`. See Table E.2 in the specification for more information about `Roles`.

## Current Behavior

If `Roles` is set to `0`, the network service starts.

ryanchristo:
## Changes

- add enabled helpers for network and rpc service to determine if the services should be started and appended to node services during the initialization of a new node

- moved the initialization of message channels outside network service and within initialization of a new node (using `coreMsgs` and `networkMsgs` at the node level, and `msgRec` and `msgSend` within each service relative to that service)

- update dot, core, and network test format and helper methods

## Tests:

```
go test ./... -short
```

### Issues:

closes #601


cmd/gossamer/config.go
```diff
@@ -127,10 +138,33 @@ func setDotAccountConfig(ctx *cli.Context, cfg *dot.AccountConfig) {
...
+			cfg.Authority = true
+		} else {
+			// if roles byte is not 4, do not act as an authority (see Table D.2)
+			log.Debug("[cmd] Authority disabled", "roles", byte(b))
+			cfg.Authority = false
+		}
+	}
+
 	// check --authority flag and update node configuration
 	if authority := ctx.GlobalBool(AuthorityFlag.Name); authority {
```
noot: we can probably just get rid of the authority flag

dot/core/service.go
```diff
@@ -79,15 +79,13 @@ type Config struct {
...
-	MsgRec           <-chan network.Message
-	MsgSend          chan<- network.Message
 	NewBlocks        chan types.Block // only used for testing purposes
 	IsBabeAuthority  bool
 }
 
 // NewService returns a new core service that connects the runtime, BABE
 // session, and network service.
-func NewService(cfg *Config) (*Service, error) {
+func NewService(cfg *Config, msgRec <-chan network.Message, msgSend chan<- network.Message) (*Service, error) {
```
noot: why is this being changed? the reason we have a config is so we don't need to change the NewService function signature

ryanchristo: I was not sure whether to include the channels in the network service configuration or remove them from the core service configuration but they should be consistent. I chose to include the channels as params because they are created at the dot node level, same level as services, and they are channels between services.

I would not be strongly opposed to reverting and adding the channels to network config as an alternative but they should be consistent.

cmd/gossamer/config.go
ryanchristo: was not sure if we were attached to it

noot: nah, feel free to remove

dot/core/service.go
noot: okay, it would be preferred to add the network channels to their config as well

ryanchristo: see https://github.com/ChainSafe/gossamer/pull/699#discussion_r393318315

noot: `NewService` is not the same function as `createCoreService`, `NewService` should only take a Config

ryanchristo: I misread this the first time and will fix.

noot: no worries! sounds good

dot/core/service_test.go
```diff
@@ -35,19 +35,22 @@ import (
 	"github.com/ChainSafe/gossamer/lib/transaction"
 	"github.com/ChainSafe/gossamer/lib/trie"
 	"github.com/ChainSafe/gossamer/tests"
+	"github.com/stretchr/testify/require"
 )
 
-var TestMessageTimeout = 5 * time.Second
+// TestMessageTimeout is the wait time for messages to be exchanged
+var TestMessageTimeout = time.Second
```
edwardmack: Why has this changed from 5 seconds to 1?

node/ksmcc/config.toml
```diff
@@ -4,6 +4,7 @@ id = "ksmcc"
 config = "./node/ksmcc/config.toml"
 genesis = "./node/ksmcc/genesis.json"
 datadir = "~/.gossamer/ksmcc"
+roles = 1
```
edwardmack: Why does ksmcc roles default to 1 while gssmr roles default to 4?

ryanchristo: no reason in particular

authority was set to false before so I set roles to 1 but not sure if there was a reason authority was set to false before

I figure we can update the ksmcc configuration when we implement the remaining external runtime methods and give more attention to interop

dot/core/service_test.go
ryanchristo: the messages are exchanged within milliseconds and sometime when we wait 5 seconds nodes connect and disconnect several times and exchange messages more than we need

cmd/gossamer/config.go
```diff
@@ -127,11 +138,21 @@ func setDotAccountConfig(ctx *cli.Context, cfg *dot.AccountConfig) {
...
+		// convert string to byte
+		b, err := strconv.Atoi(roles)
+		if err != nil {
+			log.Error("[cmd] Failed to convert Roles to byte", "error", err)
+		} else if byte(b) == 4 {
+			// if roles byte is 4, act as an authority (see Table D.2)
+			log.Debug("[cmd] Authority enabled", "roles", 4)
+			cfg.Authority = true
+		} else {
+			// if roles byte is not 4, do not act as an authority (see Table D.2)
```
noot: we should have some sort of check for invalid input eg. 3, 5

dot/core/block_msg_test.go
```diff
@@ -1,92 +0,0 @@
...
-	"time"
-
-	"github.com/ChainSafe/gossamer/dot/core/types"
-	"github.com/ChainSafe/gossamer/dot/network"
-	"github.com/ChainSafe/gossamer/lib/common"
-	"github.com/ChainSafe/gossamer/lib/keystore"
-	"github.com/stretchr/testify/require"
-)
-
-func TestProcessBlockRequest(t *testing.T) {
```
noot: where did this test go?

dot/network/service.go
```diff
@@ -95,8 +95,8 @@ func NewService(cfg *Config, msgSend chan<- Message, msgRec <-chan Message) (*Se
 		syncer:       newSyncer(host, cfg.BlockState, cfg.SyncChan),
 		blockState:   cfg.BlockState,
 		networkState: cfg.NetworkState,
-		msgRec:       msgRec,
-		msgSend:      msgSend,
+		msgRec:       cfg.MsgRec,
+		msgSend:      cfg.MsgSend,
```
noot: an error check in case these are nil would be cool

dot/node.go
```diff
@@ -147,6 +148,11 @@ func NewNode(cfg *Config, ks *keystore.Keystore) (*Node, error) {
 
 	var nodeSrvcs []services.Service
 
+	// Message Channels (send and receive messages between services)
+
+	coreMsgs := make(chan network.Message)
+	networkMsgs := make(chan network.Message)
```
noot: could you add a comment explaining these? like which one sends to network, which one sends to core, etc

cmd/gossamer/config.go
ryanchristo: will add error log :+1:

dot/core/block_msg_test.go
ryanchristo: just moved to service_test with the other process message tests

dot/network/service.go
ryanchristo: will add :+1:

dot/node.go
ryanchristo: coreMsgs are messages sent from core service to network service
networkMsgs are message sent from network service to core service

will add :+1:

Issue #662, opened 2020-03-03
NextEpochDescriptor is included in first block of an epoch


noot:
The NextEpochDescriptor includes info for the next epoch. it should be included in the first block of an epoch (to describe the next epoch)

noot:
closing for now since this is no longer needed

PR #703, opened 2020-03-16
dot/core: first block in epoch and core refactor

### LINKED ISSUE #599: core, babe: determine if block is first in epoch
needed for `NextEpochDescriptor` in consensus digests. it should be included in first block of an epoch, and only the first block. it's an error otherwise

ryanchristo:
## Changes

- [x] determine if block is first in epoch using `handleConsensusDigest`
- [x] ensure the consensus digest is included in first block of each epoch
- [x] clean up core service methods and tests / add new files to core

## Notes

- determine what epoch a block was produced at with `(slot - GENESIS_SLOT) / epoch_len`

## Tests:

all tests should pass:

```
go test dot/core -v
```

running the gossamer node...

```
make gossamer

./bin/gossamer --key alice
```

should produce the following error related to #662:

```
EROR[03-27|10:32:22] [core] failed to verify first block error="failed to verify first block: first block never set"
```

### Issues:

closes #599

### DIFF
--- a/dot/core/babe.go
+++ b/dot/core/babe.go
@@ -1,0 +1,132 @@
+// Copyright 2019 ChainSafe Systems (ON) Corp.
+// This file is part of gossamer.
+//
+// The gossamer library is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Lesser General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// The gossamer library is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with the gossamer library. If not, see <http://www.gnu.org/licenses/>.
+
+package core
+
+import (
+	"fmt"
+
+	"github.com/ChainSafe/gossamer/dot/core/types"
+	"github.com/ChainSafe/gossamer/lib/babe"
+	"github.com/ChainSafe/gossamer/lib/crypto/sr25519"
+	log "github.com/ChainSafe/log15"
+)
+
+// finalizeBabeSession finalizes the BABE session by ensuring the first block
+// was set and first block and epoch number are reset for the next epoch
+func (s *Service) finalizeBabeSession() error {
+
+	// check if first block was set for current epoch
+	if s.firstBlock == nil {
+
+		// TODO: NextEpochDescriptor is included in first block of an epoch #662
+		// return fmt.Errorf("first block not set for current epoch")
+
+		log.Error("[core] first block not set for current epoch")
+	}
+
+	// verify best block is from current epoch
+	bestHash := s.blockState.BestBlockHash()
+	currentEpoch, err := s.blockFromCurrentEpoch(bestHash)
+	if err != nil {
+		return fmt.Errorf("failed to check block from current epoch: %s", err)
+	}
+
+	if !currentEpoch {
+		return fmt.Errorf("best block is not from current epoch: %s", err)
+	}
+
+	// get best epoch number from best header
+	bestEpoch, err := s.getBlockEpoch(bestHash)
+	if err != nil {
+		return fmt.Errorf("failed to get epoch number for best block: %s", err)
+	}
+
+	// verify current epoch number matches best epoch number
+	if s.epochNumber != bestEpoch {
+		return fmt.Errorf("block epoch does not match current epoch")
+	}
+
+	// set next epoch number
+	s.epochNumber = bestEpoch + 1
+
+	// reset first block number
+	s.firstBlock = nil
+
+	return nil
+}
+
+// initializeBabeSession creates a new BABE session
+func (s *Service) initializeBabeSession() (*babe.Session, error) {
+	log.Debug(
+		"[core] initializing BABE session...",
+		"epoch", s.epochNumber,
+	)
+
+	// AuthorityData comes from NextEpochDescriptor within the ConsensusDigest
+	// of the block Digest, which is included in the first block of each epoch
+	authData := s.bs.AuthorityData()
+	if len(authData) == 0 {
+		return nil, fmt.Errorf("authority data not set")
+	}
+
+	newBlocks := make(chan types.Block)
+	s.blkRec = newBlocks
+
+	epochDone := make(chan struct{})
+	s.epochDone = epochDone
+
+	babeKill := make(chan struct{})
+	s.babeKill = babeKill
+
+	keys := s.keys.Sr25519Keypairs()
+
+	// get best slot to determine next start slot
+	bestHash := s.blockState.BestBlockHash()
+	bestSlot, err := s.blockState.GetSlotForBlock(bestHash)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get slot for latest block: %s", err)
+	}
+
+	// BABE session configuration
+	bsConfig := &babe.SessionConfig{
+		Keypair:          keys[0].(*sr25519.Keypair),
+		Runtime:          s.rt,
+		NewBlocks:        newBlocks, // becomes block send channel in BABE session
+		BlockState:       s.blockState,
+		StorageState:     s.storageState,
+		TransactionQueue: s.transactionQueue,
+		AuthData:         authData,
+		Done:             epochDone,
+		Kill:             babeKill,
+		StartSlot:        bestSlot + 1,
+		SyncLock:         s.syncLock,
+	}
+
+	// create new BABE session
+	bs, err := babe.NewSession(bsConfig)
+	if err != nil {
+		log.Error("[core] failed to initialize BABE session", "error", err)
+		return nil, err
+	}
+
+	log.Debug(
+		"[core] BABE session initialized",
+		"epoch", s.epochNumber,
+	)
+
+	return bs, nil
+}
--- a/dot/core/babe.go
+++ b/dot/core/babe.go
@@ -22,6 +22,7 @@ import (
 	"github.com/ChainSafe/gossamer/dot/core/types"
 	"github.com/ChainSafe/gossamer/lib/babe"
 	"github.com/ChainSafe/gossamer/lib/crypto/sr25519"
+
 	log "github.com/ChainSafe/log15"
 )
 
@@ -35,16 +36,17 @@ func (s *Service) finalizeBabeSession() error {
 		// TODO: NextEpochDescriptor is included in first block of an epoch #662
 		// return fmt.Errorf("first block not set for current epoch")
 
-		log.Error("[core] first block not set for current epoch")
+		log.Error("[core] first block not set for current epoch") // TODO: remove
 	}
 
-	// verify best block is from current epoch
+	// get epoch number for best block
 	bestHash := s.blockState.BestBlockHash()
 	currentEpoch, err := s.blockFromCurrentEpoch(bestHash)
 	if err != nil {
-		return fmt.Errorf("failed to check block from current epoch: %s", err)
+		return fmt.Errorf("failed to check best block from current epoch: %s", err)
 	}
 
+	// verify best block is from current epoch
 	if !currentEpoch {
 		return fmt.Errorf("best block is not from current epoch: %s", err)
 	}
--- a/dot/core/babe_test.go
+++ b/dot/core/babe_test.go
@@ -1,0 +1,19 @@
+// Copyright 2019 ChainSafe Systems (ON) Corp.
+// This file is part of gossamer.
+//
+// The gossamer library is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Lesser General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// The gossamer library is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with the gossamer library. If not, see <http://www.gnu.org/licenses/>.
+
+package core
+
+// TODO: test NextEpochDescriptor is included in first block of an epoch #662
--- a/dot/core/babe_test.go
+++ b/dot/core/babe_test.go
@@ -16,4 +16,60 @@
 
 package core
 
-// TODO: test NextEpochDescriptor is included in first block of an epoch #662
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+// test finalizeBabeSession
+func TestFinalizeBabeSession(t *testing.T) {
+	s := newTestServiceWithFirstBlock(t)
+
+	number, err := s.blockState.BestBlockNumber()
+	require.Nil(t, err)
+
+	header, err := s.blockState.BestBlockHeader()
+	require.Nil(t, err)
+
+	err = s.handleBlockDigest(header)
+	require.Nil(t, err)
+
+	require.Equal(t, number, s.firstBlock)
+
+	err = s.finalizeBabeSession()
+	require.Nil(t, err)
+
+	require.Nil(t, s.firstBlock)
+}
+
+// test initializeBabeSession
+func TestInitializeBabeSession(t *testing.T) {
+	s := newTestServiceWithFirstBlock(t)
+
+	number, err := s.blockState.BestBlockNumber()
+	require.Nil(t, err)
+
+	header, err := s.blockState.BestBlockHeader()
+	require.Nil(t, err)
+
+	err = s.handleBlockDigest(header)
+	require.Nil(t, err)
+
+	require.Equal(t, number, s.firstBlock)
+
+	epochNumber := s.epochNumber
+
+	err = s.finalizeBabeSession()
+	require.Nil(t, err)
+
+	require.Nil(t, s.firstBlock)
+
+	bs, err := s.initializeBabeSession()
+	require.Nil(t, err)
+
+	require.Equal(t, epochNumber+1, s.epochNumber)
+	require.Nil(t, s.firstBlock)
+
+	require.NotNil(t, bs)
+}
--- a/dot/core/digest.go
+++ b/dot/core/digest.go
@@ -22,6 +22,7 @@ import (
 
 	"github.com/ChainSafe/gossamer/dot/core/types"
 	"github.com/ChainSafe/gossamer/lib/babe"
+
 	log "github.com/ChainSafe/log15"
 )
 
--- a/dot/core/digest.go
+++ b/dot/core/digest.go
@@ -81,13 +81,15 @@ func (s *Service) handleConsensusDigest(header *types.Header, digest *types.Cons
 	// check if first block has been set for current epoch
 	if s.firstBlock != nil {
 
-		// check if first block has lower block number than block header
-		if s.firstBlock.Cmp(header.Number) != -1 {
+		// check if block header has lower block number than current first block
+		if header.Number.Cmp(s.firstBlock) >= 0 {
+
+			// error if block header does not have lower block number
 			return fmt.Errorf("first block already set for current epoch")
 		}
 
-		// log error if BABE produced two first blocks or received false first block from connected peer
-		log.Error("[core] received first block header with lower block number than current first block")
+		// either BABE produced two first blocks or we received invalid first block from connected peer
+		log.Warn("[core] received first block header with lower block number than current first block")
 	}
 
 	err = s.setNextEpochDescriptor(digest.Data)
--- a/dot/core/digest.go
+++ b/dot/core/digest.go
@@ -1,0 +1,118 @@
+// Copyright 2019 ChainSafe Systems (ON) Corp.
+// This file is part of gossamer.
+//
+// The gossamer library is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Lesser General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// The gossamer library is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with the gossamer library. If not, see <http://www.gnu.org/licenses/>.
+
+package core
+
+import (
+	"errors"
+	"fmt"
+
+	"github.com/ChainSafe/gossamer/dot/core/types"
+	"github.com/ChainSafe/gossamer/lib/babe"
+	log "github.com/ChainSafe/log15"
+)
+
+// handleBlockDigest checks if the provided header is the block header for
+// the first block of the current epoch, finds and decodes the consensus digest
+// item from the block digest, and then sets the epoch data for the next epoch
+func (s *Service) handleBlockDigest(header *types.Header) (err error) {
+
+	// check if block header digest items exist
+	if header.Digest == nil || len(header.Digest) == 0 {
+		return fmt.Errorf("header digest is not set")
+	}
+
+	// declare digest item
+	var item types.DigestItem
+
+	// decode each digest item and check its type
+	for _, digest := range header.Digest {
+		item, err = types.DecodeDigestItem(digest)
+		if err != nil {
+			return err
+		}
+
+		// check if digest item is consensus digest type
+		if item.Type() == types.ConsensusDigestType {
+
+			// cast decoded consensus digest item to conensus digest type
+			consensusDigest, ok := item.(*types.ConsensusDigest)
+			if !ok {
+				return errors.New("failed to cast DigestItem to ConsensusDigest")
+			}
+
+			err = s.handleConsensusDigest(header, consensusDigest)
+			if err != nil {
+				return err
+			}
+		}
+	}
+
+	return nil
+}
+
+// handleConsensusDigest checks if the first block has been set for the current
+// epoch, if first block has not been set or the block header has a lower block
+// number, sets epoch data for the next epoch using data in the ConsensusDigest
+func (s *Service) handleConsensusDigest(header *types.Header, digest *types.ConsensusDigest) error {
+
+	// check if block header is from current epoch
+	currentEpoch, err := s.blockFromCurrentEpoch(header.Hash())
+	if err != nil {
+		return fmt.Errorf("failed to check if block header is from current epoch: %s", err)
+	} else if !currentEpoch {
+		return fmt.Errorf("block header is not from current epoch: %s", err)
+	}
+
+	// check if first block has been set for current epoch
+	if s.firstBlock != nil {
+
+		// check if first block has lower block number than block header
+		if s.firstBlock.Cmp(header.Number) != -1 {
+			return fmt.Errorf("first block already set for current epoch")
+		}
+
+		// log error if BABE produced two first blocks or received false first block from connected peer
+		log.Error("[core] received first block header with lower block number than current first block")
+	}
+
+	err = s.setNextEpochDescriptor(digest.Data)
+	if err != nil {
+		return fmt.Errorf("failed to set next epoch descriptor: %s", err)
+	}
+
+	// set first block in current epoch
+	s.firstBlock = header.Number
+
+	return nil
+}
+
+// setNextEpochDescriptor sets the epoch data for the next epoch from the data
+// included in the ConsensusDigest of the first block of each epoch
+func (s *Service) setNextEpochDescriptor(data []byte) error {
+
+	// initialize epoch data interface for next epoch
+	nextEpochData := new(babe.NextEpochDescriptor)
+
+	// decode consensus digest data for next epoch
+	err := nextEpochData.Decode(data)
+	if err != nil {
+		return err
+	}
+
+	// set epoch data for next epoch
+	return s.bs.SetEpochData(nextEpochData)
+}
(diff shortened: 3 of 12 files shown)


dot/core/consensus.go
```diff
@@ -0,0 +1,177 @@
...
+
+	"github.com/ChainSafe/gossamer/dot/core/types"
+	"github.com/ChainSafe/gossamer/lib/babe"
+	"github.com/ChainSafe/gossamer/lib/crypto/sr25519"
+
+	log "github.com/ChainSafe/log15"
+)
+
+// newBabeSession creates a new BABE session
+func (s *Service) newBabeSession() (*babe.Session, error) {
```
ryanchristo: This method was code extracted from `handleBabeSession`, which has been updated below in `service.go`. This method and the other methods within this file is the bulk of new code within this pull request. I will comment where other items have simply moved from one file to another.

dot/core/consensus_test.go
```diff
@@ -0,0 +1,19 @@
...
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with the gossamer library. If not, see <http://www.gnu.org/licenses/>.
+
+package core
+
+// TODO
```
ryanchristo: we need `NextEpochDescriptor` to be included in first block before we can test, maybe we should add this to #662

dot/core/process.go
```diff
@@ -0,0 +1,292 @@
...
+	"github.com/ChainSafe/gossamer/lib/common/optional"
+	"github.com/ChainSafe/gossamer/lib/transaction"
+
+	log "github.com/ChainSafe/log15"
+)
+
+// BlockRequestMessage 1
+
+// ProcessBlockRequestMessage processes a block request message, returning a block response message
+func (s *Service) ProcessBlockRequestMessage(msg network.Message) error {
```
ryanchristo: this method did not change, just grouped with the other message processing methods

dot/core/process.go
```diff
@@ -0,0 +1,292 @@
...
+	blockResponse, err := s.createBlockResponse(blockRequest)
+	if err != nil {
+		return err
+	}
+
+	return s.safeMsgSend(blockResponse)
+}
+
+// createBlockResponse create a block response message from a block request message
+func (s *Service) createBlockResponse(blockRequest *network.BlockRequestMessage) (*network.BlockResponseMessage, error) {
```
ryanchristo: this is only called in `ProcessBlockRequestMessage`, nothing changed here either, only moved

dot/core/process.go
```diff
@@ -0,0 +1,292 @@
...
+		BlockData: responseData,
+	}, nil
+}
+
+// BlockResponseMessage 2
+
+// ProcessBlockResponseMessage attempts to validate and add the block to the
+// chain by calling `core_execute_block`. Valid blocks are stored in the block
+// database to become part of the canonical chain.
+func (s *Service) ProcessBlockResponseMessage(msg network.Message) error {
```
ryanchristo: I had to update this method to loop through block data to check for first block

dot/core/process.go
```diff
@@ -0,0 +1,292 @@
...
+
+	return nil
+}
+
+// BlockAnnounceMessage 3
+
+// ProcessBlockAnnounceMessage creates a block request message from the block
+// announce messages (block announce messages include the header but the full
+// block is required to execute `core_execute_block`).
+func (s *Service) ProcessBlockAnnounceMessage(msg network.Message) error {
```
ryanchristo: same as above, I had to update this method to check for first block

dot/core/process.go
```diff
@@ -0,0 +1,292 @@
...
+	}
+
+	return nil
+}
+
+// TransactionMessage 4
+
+// ProcessTransactionMessage validates each transaction in the message and
+// adds valid transactions to the transaction queue of the BABE session
+func (s *Service) ProcessTransactionMessage(msg network.Message) error {
```
ryanchristo: this method did not change

dot/core/process_test.go
```diff
@@ -0,0 +1,413 @@
...
+	"github.com/ChainSafe/gossamer/lib/trie"
+	"github.com/ChainSafe/gossamer/tests"
+
+	"github.com/stretchr/testify/require"
+)
+
+// BlockRequestMessage 1
+
+// tests the ProcessBlockRequestMessage method
+func TestService_ProcessBlockRequestMessage(t *testing.T) {
```
ryanchristo: none of the test methods within this file were changed, all were just moved

dot/core/runtime.go
```diff
@@ -17,18 +17,28 @@
...
+	"github.com/ChainSafe/gossamer/lib/common"
 	"github.com/ChainSafe/gossamer/lib/runtime"
 	"github.com/ChainSafe/gossamer/lib/scale"
 	"github.com/ChainSafe/gossamer/lib/transaction"
 
 	log "github.com/ChainSafe/log15"
 )
 
+// StorageRoot returns the hash of the runtime storage root
+func (s *Service) StorageRoot() (common.Hash, error) {
```
ryanchristo: this was just moved to `runtime` since it relates to runtime

dot/core/runtime.go
```diff
@@ -91,3 +101,35 @@ func (s *Service) grandpaAuthorities() ([]*babe.AuthorityData, error) {
 
 	return auths, err
 }
+
+// checkForRuntimeChanges checks if changes to the runtime code have occurred; if so, load the new runtime
+func (s *Service) checkForRuntimeChanges() error {
```
ryanchristo: same as above

dot/core/runtime_test.go
```diff
@@ -46,7 +51,7 @@ func TestRetrieveAuthorityData(t *testing.T) {
 		rt: rt,
 	}
 
-	auths, err := s.retrieveAuthorityData()
+	auths, err := s.grandpaAuthorities()
```
ryanchristo: `retrieveAuthorityData` called `grandpaAuthorities`, I removed `retrieveAuthorityData` for now

dot/core/runtime_test.go
```diff
@@ -66,3 +71,72 @@ func TestRetrieveAuthorityData(t *testing.T) {
 		t.Fatalf("Fail: got %v expected %v", auths, expected)
 	}
 }
+
+func TestValidateBlock(t *testing.T) {
```
ryanchristo: none of the tests within runtime changed, just runtime related tsts moved to runtime

dot/core/service.go
```diff
@@ -55,8 +52,10 @@ type Service struct {
 	codeHash common.Hash
 
 	// Current BABE session
-	bs              *babe.Session
-	isBabeAuthority bool
+	bs          *babe.Session
+	isAuthority bool
+	epochNumber uint64   // epoch number of current epoch
+	firstBlock  *big.Int // block number of first block in current epoch
```
ryanchristo: added two new params to track epoch and first block in core service

dot/core/service.go
```diff
@@ -138,14 +137,18 @@ func NewService(cfg *Config) (*Service, error) {
...
 		if cfg.Keystore.NumSr25519Keys() == 0 {
 			return nil, fmt.Errorf("no keys provided for authority node")
 		}
 
 		epochDone := make(chan struct{})
 		babeKill := make(chan struct{})
 
+		// non-persistent epoch tracking state stored in core service
+		epochNumber := uint64(0)
+		firstBlock := big.NewInt(0)
```
ryanchristo: should these be set to something else?

dot/core/service.go
```diff
@@ -454,258 +382,37 @@ func (s *Service) handleReceivedMessage(msg network.Message) (err error) {
...
-
-		s.rt.Stop()
+func (s *Service) handleBabeSession() {
+	for {
+		// wait for BABE epoch to complete
+		<-s.epochDone
 
-		s.rt, err = runtime.NewRuntime(code, s.storageState, s.keys)
+		// verify first block was set during current epoch
+		err := s.verifyFirstBlock()
```
ryanchristo: this is what is throwing the error in the logs and needs #662 in order to be resolved

dot/core/consensus.go
ryanchristo: I also think this method in particular could probably be refactored at a later point and included in babe as a reinitialization method that gets passed a couple params if need be.

dot/core/consensus.go
```diff
@@ -0,0 +1,177 @@
...
+			// initialize epoch data interface for next epoch
+			nextEpochData := new(babe.NextEpochDescriptor)
+
+			// decode consensus digest data for next epoch
+			err = nextEpochData.Decode(consensusDigest.Data)
+			if err != nil {
+				return err
+			}
+
+			// set epcoh data for next epoch
```
noot: epoch

dot/core/consensus.go
```diff
@@ -0,0 +1,177 @@
...
+		item, err = types.DecodeDigestItem(digest)
+		if err != nil {
+			return err
+		}
+
+		// check if digest item is consensus digest type
+		if item.Type() == types.ConsensusDigestType {
+
+			// cast decoded consensus digest item to conensus digest type
+			consensusDigest := item.(*types.ConsensusDigest)
```
noot: maybe add `ok` check here just do be safe?

dot/core/consensus.go
```diff
@@ -0,0 +1,177 @@
...
+			}
+
+			// set epcoh data for next epoch
+			err = s.bs.SetEpochData(nextEpochData)
+			if err != nil {
+				return err
+			}
+
+			// set first block in current epoch
+			s.firstBlock = header.Number
```
noot: not sure about this, this assumes that the block we have received is honest/valid, but it might notactually be the first block and this will overwrite it

dot/core/consensus.go
```diff
@@ -0,0 +1,177 @@
...
+func (s *Service) verifyCurrentEpoch(header *types.Header) error {
+
+	// get slot number to determine epoch number
+	slot, err := s.blockState.GetSlotForBlock(header.Hash())
+	if err != nil {
+		return fmt.Errorf("failed to get slot for block: %s", err)
+	}
+
+	// epoch number = (slot - genesis slot) / epoch length
+	epoch := (slot - 0) / 6 // TODO: set epoch length in core config
```
noot: add a helper method to babe so that the epoch length can be accessed with `s.bs.EpochLength()` or something similar?

noot: also `slot - 0` ? can you remove the ` -0`

dot/core/consensus.go
ryanchristo: right, I added the check first block method when finalizing an epoch and starting a new babe session and meant to add the check here as well. I was going to assume the first block to claim it was the first block is the first block and error otherwise. Not sure how else to address.

dot/core/consensus.go
ryanchristo: that would be nice, was not sure how to do this and did not want to extend this pull request beyond the core service. I can create an issue for this.

dot/core/consensus_test.go
noot: NextEpochDescriptor is implemented here https://github.com/ChainSafe/gossamer/blob/f4d80d67b7e84c73a8a2213297516c3050cfd9ef/lib/babe/types.go#L150 so I think you could write a test with it?

ryanchristo: `NextEpochDesciptor` is #662, so it might make more sense to hold off here and have a babe session running that includes all the necessary data in the first block before testing here.

ryanchristo: I was having trouble mocking the babe session before as well, which is why I am hoping it will be easier to write a test once #662 is implemented but maybe I was overcomplicating it

dot/core/consensus.go
ryanchristo: I kept `slot - 0` because `slot - genesis slot`, which I think we should keep and replace with a constant or param but was not sure if we were storing the genesis slot somewhere. I think it makes sense to keep `slot - 0` given we are assuming genesis slot is `0`.

dot/core/process.go
```diff
@@ -0,0 +1,292 @@
...
+	"github.com/ChainSafe/gossamer/dot/core/types"
+	"github.com/ChainSafe/gossamer/dot/network"
+	"github.com/ChainSafe/gossamer/lib/common"
+	"github.com/ChainSafe/gossamer/lib/common/optional"
+	"github.com/ChainSafe/gossamer/lib/transaction"
+
+	log "github.com/ChainSafe/log15"
+)
+
+// BlockRequestMessage 1
```
noot: remove?

ryanchristo: these methods were out of order and mixed with other methods, I would prefer to make it clear we are going through the process message methods in order based on they type

dot/core/process.go
```diff
@@ -0,0 +1,292 @@
...
+			}
+
+			// check authority
+			if s.isAuthority {
+
+				// check if block is from current epoch
+				err = s.verifyCurrentEpoch(header)
+				if err != nil {
+					log.Trace("[core] block is not from current epoch", "error", err)
+				}
```
noot: would this work if we are syncing? eg. the chain is at epoch 100 and we are still syncing past epochs for example

dot/core/process.go
```diff
@@ -0,0 +1,292 @@
...
+		if err != nil {
+			log.Trace("[core] block is not from current epoch", "error", err)
+		}
+
+		// check if first block and set epoch data for next epoch
+		err = s.handleConsensusDigest(header)
+		if err != nil {
+			log.Trace("[core] block is not first block in epoch", "error", err)
+		}
+	}
```
noot: since this code is the same as above, maybe move it to another function that can be used by ProcessBlockAnnounceMessage and ProcessBlockResponseMessage

dot/core/process_test.go
```diff
@@ -0,0 +1,413 @@
...
+					require.Equal(t, test.expectedMsgValue.BlockData[0].Justification, resp.(*network.BlockResponseMessage).BlockData[1].Justification)
+				}
+			case <-time.After(testMessageTimeout):
+				t.Error("timeout waiting for message")
+			}
+		})
+	}
+}
+
+// BlockResponseMessage 2
```
noot: remove

dot/core/process.go
ryanchristo: agreed, will fix

dot/core/process_test.go
```diff
@@ -0,0 +1,413 @@
...
+	select {
+	case resp := <-s.syncer.respIn:
+		msgType := resp.GetType()
+		require.Equal(t, network.BlockResponseMsgType, msgType)
+	case <-time.After(testMessageTimeout):
+		t.Error("timeout waiting for message")
+	}
+}
+
+// BlockAnnounceMessage 3
```
noot: same

dot/core/process_test.go
```diff
@@ -0,0 +1,413 @@
...
+	case msg := <-msgSend:
+		msgType := msg.GetType()
+		require.Equal(t, network.BlockAnnounceMsgType, msgType)
+		require.Equal(t, expected, msg)
+	case <-time.After(testMessageTimeout):
+		t.Error("timeout waiting for message")
+	}
+}
+
+// TransactionMessage 4
```
noot: same

dot/core/runtime.go
noot: this isn't really related to runtime, not sure if it belongs here

dot/core/process_test.go
ryanchristo: would prefer to keep

dot/core/process_test.go
ryanchristo: same

dot/core/process_test.go
ryanchristo: same

dot/core/runtime.go
ryanchristo: "returns the hash of the runtime storage root" ?

noot: yeah that's not really correct it should just say storage root

ryanchristo: ok, will move back to `service.go` and update comment

dot/core/service.go
```diff
@@ -454,258 +382,37 @@ func (s *Service) handleReceivedMessage(msg network.Message) (err error) {
...
-// adds valid transactions to the transaction queue of the BABE session
-func (s *Service) ProcessTransactionMessage(msg network.Message) error {
-
-	// get transactions from message extrinsics
-	txs := msg.(*network.TransactionMessage).Extrinsics
-
-	for _, tx := range txs {
-		tx := tx // pin
+		// update current epoch number
+		s.epochNumber = s.epochNumber + 1
```
noot: does this work with syncing?

dot/core/consensus.go
noot: okay, sure that seems fair. yeah just make sure this can't get overwritten by subsequent blocks

dot/core/consensus.go
noot: the first slot should be slot 1 I believe

dot/core/service.go
ryanchristo: related to https://github.com/ChainSafe/gossamer/pull/703#discussion_r398772061

dot/core/process.go
ryanchristo: related to https://github.com/ChainSafe/gossamer/pull/703#discussion_r398783470

dot/core/consensus.go
ryanchristo: do you mean genesis slot? should it be a constant? should it be configurable?

noot: it should always be 1 so this line should be `(slot - 1) / 6`

then slots 1-6 are in epoch 0, 7-12 in epoch 1. etc

dot/core/consensus.go
ryanchristo: added `verifyFirstBlock` if `ConsensusDigestType`

dot/core/process.go
```diff
@@ -0,0 +1,281 @@
...
+				bd.Header.Value().ExtrinsicsRoot,
+				bd.Header.Value().Digest,
+			)
+			if err != nil {
+				return err
+			}
+
+			// if authority, check if first block and set epoch data for next epoch
+			if s.isAuthority {
+				err = s.handleConsensusDigest(header)
```
noot: what about the case when firstBlock is already set?

dot/core/process.go
```diff
@@ -0,0 +1,281 @@
...
+		msg.StateRoot,
+		msg.ExtrinsicsRoot,
+		msg.Digest,
+	)
+	if err != nil {
+		return err
+	}
+
+	// if authority, check if first block and set epoch data for next epoch
+	if s.isAuthority {
```
noot: same here

dot/core/service.go
noot: I'm wondering if we should have a map of epoch number -> first block, how would this work when syncing lots of epochs? when syncing, babe sessions don't get regenerated every epoch, it just waits until syncing is done, so I'm concerned about this being incorrect for syncing

dot/core/process.go
noot: could you explain here? the comments just link to each other lol

dot/core/process.go
noot: I think you should change this statement to only execute is firstBlock for the epoch is not already set, *or* if a block is received with a lower number that the current firstBlock but is still in the epoch

ryanchristo: it is inside `handleConsensusDigest` because you request that this be consolidated

dot/core/process.go
ryanchristo: it is inside `handleConsensusDigest` because you request that this be consolidated

dot/core/process.go
ryanchristo: I was still addressing this, revised in latest commit

dot/core/service.go
ryanchristo: revised in latest commit

dot/core/consensus.go
```diff
@@ -0,0 +1,213 @@
...
+	return bs, nil
+}
+
+// handleConsensusDigest checks if the provided header is the block header for
+// the first block of the current epoch, finds and decodes the consensus digest
+// item from the block digest, and then sets the epoch data for the next epoch
+func (s *Service) handleConsensusDigest(header *types.Header) (err error) {
+
+	// check if block is from current epoch
+	err = s.verifyCurrentEpoch(header)
```
ryanchristo: see here @noot

dot/core/consensus.go
```diff
@@ -0,0 +1,213 @@
...
+		item, err = types.DecodeDigestItem(digest)
+		if err != nil {
+			return err
+		}
+
+		// check if digest item is consensus digest type
+		if item.Type() == types.ConsensusDigestType {
+
+			// check if first block already set (consensus digest already set)
+			err = s.verifyFirstBlock()
```
ryanchristo: see here @noot

dot/core/service.go
ryanchristo: Can you create a separate issue for this?

dot/core/service.go
```diff
@@ -439,290 +370,141 @@ func (s *Service) handleReceivedMessage(msg network.Message) (err error) {
...
-
-	_, err = s.blockState.GetBlockBody(header.Hash())
-	if err != nil && err.Error() == "Key not found" {
-		// send block request message
-		log.Debug("[core] sending new block to syncer", "number", msg.Number)
-		s.blockNumOut <- msg.Number
-	} else if err != nil {
-		return err
-	}
+			// TODO: NextEpochDescriptor is included in first block of an epoch #662
```
ryanchristo: #662

dot/core/consensus_test.go
```diff
@@ -0,0 +1,19 @@
...
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with the gossamer library. If not, see <http://www.gnu.org/licenses/>.
+
+package core
+
+// TODO: test handleConsensusDigest once NextEpochDescriptor is included in first block #662
```
ryanchristo: #662

dot/core/consensus.go
```diff
@@ -0,0 +1,213 @@
...
+			if err == nil {
+				return errors.New("failed to set first block: first block already set")
+			}
```
noot: this is confusing, verifyFirstBlock doesn't error if the first block is already set, thenn you check for the absense of an error? I think it should be the other way around, just check if firstBlock != nil (and set it to nil when the epoch changes)

noot: this also doesn't check for the case where first block is already set, and we receive another block with a lower number that is also in the current epoch

ryanchristo: fixed

ryanchristo:
I am having trouble writing test cases without a babe session that produces valid first blocks with `NextEpochDescriptor` included in first block but maybe I can write tests for the helper methods I created.

noot:
@ryanchristo  you don't need to use BABE, just make a block, then make a NextEpochDescriptor, then encode it and put it in a consensus digest

dot/core/consensus.go
```diff
@@ -0,0 +1,241 @@
...
+	return nil
+}
+
+// handleConsensusDigest checks if the first block has been set for the current
+// epoch, if first block has not been set, assumes first block and sets epoch
+// data for the next epoch using NextEpochDescriptor in the ConsensusDigest
+func (s *Service) handleConsensusDigest(header *types.Header, digest *types.ConsensusDigest) error {
+
+	// check if first block has been set for current epoch
+	if s.firstBlock != nil {
```
noot: also check if `header` has a lower number than the current first block and is also in this epoch?

dot/core/consensus.go
```diff
@@ -0,0 +1,241 @@
...
+	if slot != 0 {
+		// epoch number = (slot - genesis slot) / epoch length
+		epoch = (slot - 1) / 6 // TODO: use epoch length from babe or core config
+	}
+
+	return epoch, nil
+}
+
+// verifyCurrentEpoch verifies the provided block hash is from current epoch
+func (s *Service) verifyCurrentEpoch(hash common.Hash) error {
```
noot: this function is a bit confusing, it doesn't verify the current epoch, it checks if a block is in the current epoch. could you rename it? 

also, the function should return a `bool` to say whether it is in the epoch or not. `error` should be reserved for something actually going wrong

dot/core/consensus.go
```diff
@@ -0,0 +1,241 @@
...
+
+	// get epoch number of block header
+	epoch, err := s.getBlockEpoch(hash)
+	if err != nil {
+		return fmt.Errorf("failed to get epoch from block header: %s", err)
+	}
+
+	// check if block epoch number matches current epoch number
+	if epoch != s.epochNumber {
+		return fmt.Errorf("block epoch does not match current epoch")
```
noot: return `false, nil`

dot/core/consensus.go
```diff
@@ -0,0 +1,241 @@
...
+	}
+
+	return nil
```
noot: return `true, nil`

dot/core/consensus.go
ryanchristo: Not sure what the check for a lower number would do if the fist block is reset for each epoch during finalize and initialize babe session.

Are you saying if their are two blocks claiming to be the first block and their epoch number matches the current epoch, we should favor the first block with the lower number...?

Also, the check for current epoch is 5 lines down from this check for first block. Should this be moved? Can you clarify?

dot/core/consensus.go
ryanchristo: can do, will fix

dot/core/consensus.go
ryanchristo: :+1:

dot/core/consensus.go
ryanchristo: :+1:

ryanchristo:
cc @noot comments addressed with questions about the first. ready for another look.

dot/core/consensus.go
noot: yes, if you receive multiple blocks that are both in the current epoch and contain the consensus digest, you should favour the one with the lower number. the one with the higher number is definitely not the first block in that epoch if there's another block also in that epoch with a lower number

dot/core/babe.go
```diff
@@ -0,0 +1,132 @@
...
+)
+
+// finalizeBabeSession finalizes the BABE session by ensuring the first block
+// was set and first block and epoch number are reset for the next epoch
+func (s *Service) finalizeBabeSession() error {
+
+	// check if first block was set for current epoch
+	if s.firstBlock == nil {
+
+		// TODO: NextEpochDescriptor is included in first block of an epoch #662
```
ryanchristo, line 36: #662

dot/core/babe_test.go
```diff
@@ -0,0 +1,19 @@
...
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with the gossamer library. If not, see <http://www.gnu.org/licenses/>.
+
+package core
+
+// TODO: test NextEpochDescriptor is included in first block of an epoch #662
```
ryanchristo, line 19: #662

dot/core/digest.go
```diff
@@ -0,0 +1,118 @@
...
+
+	return nil
+}
+
+// handleConsensusDigest checks if the first block has been set for the current
+// epoch, if first block has not been set or the block header has a lower block
+// number, sets epoch data for the next epoch using data in the ConsensusDigest
+func (s *Service) handleConsensusDigest(header *types.Header, digest *types.ConsensusDigest) error {
+
+	// check if block header is from current epoch
```
ryanchristo, line 73: 1. check if block header is from current epoch

dot/core/digest.go
```diff
@@ -0,0 +1,118 @@
...
+	currentEpoch, err := s.blockFromCurrentEpoch(header.Hash())
+	if err != nil {
+		return fmt.Errorf("failed to check if block header is from current epoch: %s", err)
+	} else if !currentEpoch {
+		return fmt.Errorf("block header is not from current epoch: %s", err)
+	}
+
+	// check if first block has been set for current epoch
```
ryanchristo, line 81: 2. check if first block has been set for current epoch

dot/core/digest.go
```diff
@@ -0,0 +1,118 @@
...
+	if s.firstBlock != nil {
+
+		// check if first block has lower block number than block header
```
ryanchristo, line 84: 3. check if first block has lower block number than block header

dot/core/digest.go
```diff
@@ -0,0 +1,118 @@
...
+		if s.firstBlock.Cmp(header.Number) != -1 {
+			return fmt.Errorf("first block already set for current epoch")
+		}
+
+		// log error if BABE produced two first blocks or received false first block from connected peer
+		log.Error("[core] received first block header with lower block number than current first block")
```
ryanchristo, line 90: if header has lower block number than first block, log error and replace current first block with header

ryanchristo:
cc @noot I think everything is addressed, let me know if not :crossed_fingers:

dot/core/digest.go
noot: maybe change this to just `Warn` or `Debug`?

dot/core/digest.go
```diff
@@ -0,0 +1,119 @@
...
+		return fmt.Errorf("block header is not from current epoch: %s", err)
+	}
+
+	// check if first block has been set for current epoch
+	if s.firstBlock != nil {
+
+		// check if first block has lower block number than block header
+		if s.firstBlock.Cmp(header.Number) != -1 {
+			return fmt.Errorf("first block already set for current epoch")
+		}
```
noot, line 87: could you add a test case for this? I think this says if `s.firstBlock` *is not* less than `header.Number` then error but I think it should be if `s.firstBlock` *is* less than `header.Number`?

ryanchristo: done :+1:

dot/core/digest.go
ryanchristo: done :+1:

PR #707, opened 2020-03-17
state: block data fields should be stored separately 

### LINKED ISSUE #596: state: block data fields should be stored separately

thomasmodeneis:
## Changes
- Adding initial impl for Receipt, MessageQueue, and Justification get/set methods
- Adding fixes for nilpointer on optional/types.go funcs after testing the effect of removing header field from SetBlockBody func
- removing GetBlockData and SetBlockData and implementing logic to have block data fields stored separately
- 

## Tests:
```

```

### Issues:
closes #596


dot/state/block.go
```diff
@@ -291,28 +249,74 @@ func (bs *BlockState) SetHeader(header *types.Header) error {
...
 	}
-	return bs.SetBlockData(blockData)
+
+	return result, nil
 }
 
-// SetBlockData will set the block data using given hash and blockData into DB
-func (bs *BlockState) SetBlockData(blockData *types.BlockData) error {
+// GetBlockDataAllFields will return what GetBlockDataSimple plus all other optional fields
+func (bs *BlockState) GetBlockDataAllFields(hash common.Hash) (*types.BlockData, error) {
```
noot: we shouldn't have any BlockData related methods in the state anymore, at least not something for getting the entire block data. if you want to move the bitmap parsing to here, and return BlockData that's in the bitmap, then that is okay, but it's unnecessary to do gets for all the fields when we might not need them

dot/state/block.go
```diff
@@ -291,28 +249,74 @@ func (bs *BlockState) SetHeader(header *types.Header) error {
...
+
+}
+
+// SetBlockDataSimple will add a block body and the hash to the db
+func (bs *BlockState) SetBlockDataSimple(hash common.Hash, body *optional.Body) error {
 	bs.lock.Lock()
 	defer bs.lock.Unlock()
 
-	// Write the encoded header
+	blockData := &types.BlockData{
```
noot: we don't need to store the body as BlockData, we can just store it as hash->body

dot/state/block.go
```diff
@@ -322,6 +326,37 @@ func (bs *BlockState) SetBlockData(blockData *types.BlockData) error {
 	return err
 }
 
+// SetBlockDataAllFields call SetBlockDataSimple and then set all optional fields to the db
+func (bs *BlockState) SetBlockDataAllFields(blockData *types.BlockData) error {
```
noot: we should be setting all the fields separately

dot/state/block_funcs.go
```diff
@@ -0,0 +1,116 @@
...
+
+import (
+	"bytes"
+
+	"github.com/ChainSafe/gossamer/lib/common"
+	"github.com/ChainSafe/gossamer/lib/common/optional"
+)
+
+// SetReceipt sets a Receipt in the database
+func (bs *BlockState) SetReceipt(hash common.Hash, data []byte) error {
```
noot: these are good!

dot/core/service.go
```diff
@@ -526,7 +526,7 @@ func (s *Service) ProcessBlockRequestMessage(msg network.Message) error {
 	responseData := []*types.BlockData{}
 
 	for _, hash := range subchain {
-		data, err := s.blockState.GetBlockData(hash)
+		data, err := s.blockState.GetBlockDataAllFields(hash)
```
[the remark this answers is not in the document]
noot: looking at the rest of the PR, if you want to move the bitmap parsing to a function in the state that would be fine with me. it would help clean up the state interface also.

for example:
```
func (b *BlockState) GetBlockData(bitmap byte) (*types.BlockData) {
    ret = new(types.BlockData)
    if (bitmap&1) == 1 // get header and set ret.Header
    if (bitmap&2)>>1 == 1 // get body and set ret.Body
    // etc for rest of fields
}
```

dot/state/block.go
noot: we already have CompareAndSetBlockData: https://github.com/ChainSafe/gossamer/pull/709/files#diff-f3d5a6e9d6e7c92e678068b8316babd5R327

I think if we have that already we don't need this

dot/core/service.go
```diff
@@ -526,47 +526,64 @@ func (s *Service) ProcessBlockRequestMessage(msg network.Message) error {
 	responseData := []*types.BlockData{}
 
 	for _, hash := range subchain {
-		data, err := s.blockState.GetBlockData(hash)
+		dataBody, err := s.blockState.GetBlockBody(hash)
```
noot: remove this, since block body is being retrieved at line 552

dot/core/service.go
```diff
@@ -526,47 +526,64 @@ func (s *Service) ProcessBlockRequestMessage(msg network.Message) error {
...
 		blockData := new(types.BlockData)
 		blockData.Hash = hash
+		blockData.Body = dataBody
 
 		// TODO: checks for the existence of the following fields should be implemented once #596 is addressed.
 
 		// header
 		if blockRequest.RequestedData&1 == 1 {
-			blockData.Header = data.Header
+			retData, thisErr := s.blockState.GetHeader(hash)
```
noot: don't really like `thisErr`...

dot/core/service.go
```diff
@@ -526,47 +526,64 @@ func (s *Service) ProcessBlockRequestMessage(msg network.Message) error {
...
 		}
 
 		// receipt
 		if (blockRequest.RequestedData&4)>>2 == 1 {
-			blockData.Receipt = data.Receipt
+			blockData.Receipt, err = s.blockState.GetReceipt(hash)
+			if err != nil {
+				log.Error("[core] could not get Receipt for hash from db")
+			}
+
```
noot: you need to set `blockData.Receipt` as well here if err == nil, same for Justification and MessageQueue below

also, you aren't logging for Header and Body, consistency would be good

dot/core/service_test.go
```diff
@@ -257,9 +245,121 @@ func TestProcessBlockResponseMessage(t *testing.T) {
...
+		Header:        optional.NewHeader(false, nil),
+		Body:          optional.NewBody(true, body),
+		Receipt:       optional.NewBytes(true, []byte("asdf")),
+		MessageQueue:  optional.NewBytes(true, []byte("ghjkl")),
+		Justification: optional.NewBytes(true, []byte("qwerty")),
+	}}
+
+	for _, blockdata := range bds {
+
+		err := s.blockState.CompareAndSetBlockData(blockdata)
```
noot: why is this test here? it should be in `BlockState` since we are testing `BlockState` functions

dot/state/block.go
```diff
@@ -146,9 +145,9 @@ func headerHashKey(number uint64) []byte {
 	return append(headerHashPrefix, encodeBlockNumber(number)...)
 }
 
-// blockDataKey = blockDataPrefix + hash
-func blockDataKey(hash common.Hash) []byte {
-	return append(blockDataPrefix, hash.ToBytes()...)
+// blockBodyKey = blockDataPrefix + hash
```
noot: `blockBodyPrefix`

dot/state/block.go
```diff
@@ -291,35 +247,73 @@ func (bs *BlockState) SetHeader(header *types.Header) error {
 	return nil
 }
 
-// SetBlock will add a block to the DB
-func (bs *BlockState) SetBlock(block *types.Block) error {
-	// Add the blockHeader to the DB
-	err := bs.SetHeader(block.Header)
+// GetBlockBody will return Body for a given hash
+func (bs *BlockState) GetBlockBody(hash common.Hash) (*optional.Body, error) {
```
noot: don't return an optional here, if it doesn't exist then an error will be returned

dot/state/block.go
```diff
@@ -291,35 +247,73 @@ func (bs *BlockState) SetHeader(header *types.Header) error {
...
-		Body:   block.Body.AsOptional(),
-	}
-	return bs.SetBlockData(blockData)
+	return &optional.Body{Exists: true, Value: data}, nil
 }
 
-// SetBlockData will set the block data using given hash and blockData into DB
-func (bs *BlockState) SetBlockData(blockData *types.BlockData) error {
+// SetBlockBody will add a block body to the db
+func (bs *BlockState) SetBlockBody(hash common.Hash, body *optional.Body) error {
```
noot: same here, there is no point in putting an empty optional in the DB. argument should be `types.Body`

dot/state/block_funcs.go
```diff
@@ -0,0 +1,116 @@
...
+	err := bs.db.Put(prefixHash(hash, receiptPrefix), data)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// GetReceipt retrieves a Receipt from the database
+func (bs *BlockState) GetReceipt(hash common.Hash) (*optional.Bytes, error) {
```
noot: don't need to return optional here since it will error if it doesn't exist, same for MessageQueue and Justification

dot/state/block_funcs.go
```diff
@@ -0,0 +1,116 @@
...
+	data, err := bs.db.Get(prefixHash(hash, receiptPrefix))
+	if err != nil {
+		return nil, err
+	}
+
+	r := &bytes.Buffer{}
+	_, err = r.Write(data)
+	if err != nil {
+		return nil, err
+	}
```
noot: what is this? `r` isn't used anywhere

dot/state/block_funcs.go
```diff
@@ -0,0 +1,116 @@
...
+	data, err := bs.db.Get(prefixHash(hash, messageQueuePrefix))
+	if err != nil {
+		return nil, err
+	}
+
+	r := &bytes.Buffer{}
+	_, err = r.Write(data)
+	if err != nil {
+		return nil, err
+	}
```
noot: same

dot/state/block_funcs.go
```diff
@@ -0,0 +1,116 @@
...
+
```
noot: same

lib/babe/state.go
```diff
@@ -33,10 +35,24 @@ type BlockState interface {
...
+	GetHeader(common.Hash) (*types.Header, error)
+
+	GetReceipt(common.Hash) (*optional.Bytes, error)
+	SetReceipt(common.Hash, []byte) error
+
+	GetMessageQueue(common.Hash) (*optional.Bytes, error)
+	SetMessageQueue(common.Hash, []byte) error
+
+	GetJustification(common.Hash) (*optional.Bytes, error)
+	SetJustification(common.Hash, []byte) error
```
noot: remove these, `BABE` doesn't use them

dot/core/service.go
thomasmodeneis: you mean set on the db ? its already setting...
`blockData.Receipt, err = s.blockState.GetReceipt(hash)`

dot/core/service_test.go
thomasmodeneis: The test is there because its based on a test that was inside this file. I am not sure I understand your recommendation. Can you be more specific ?

dot/state/block.go
thomasmodeneis: That was the previous logic in the code, I left it like that to avoid changing things out of the scope in the ticket. But if you want I can get rid of the optionals, no problem.

dot/state/block_funcs.go
thomasmodeneis: some useless garbage it seems, will remove.

dot/core/service_test.go
ryanchristo: I think she means this particular loop only tests block state methods so its not necessary to include in the core service and should be moved to its own test within the state service.

for example, the `// test Receipt` message below checks if the block data has a receipt `Receipt.Exists` using the provided block data, then gets the receipt `blockState.GetReceipt` using block state, then checks the original block data receipt with the receipt retrieved.

This loop should be its own test in block state, which we could provide various block data cases to or leave as is with one block data case for now.

dot/core/service.go
thomasmodeneis: updated.

dot/core/service.go
thomasmodeneis: updated.

dot/core/service.go
thomasmodeneis: updated.

dot/state/block.go
thomasmodeneis: updated.

dot/state/block.go
thomasmodeneis: updated.

dot/state/block.go
thomasmodeneis: updated.

dot/state/block_funcs.go
thomasmodeneis: updated.

dot/state/block_funcs.go
thomasmodeneis: updated.

dot/state/block_funcs.go
thomasmodeneis: updated.

dot/state/block_funcs.go
thomasmodeneis: updated.

lib/babe/state.go
thomasmodeneis: updated.

dot/core/service.go
```diff
@@ -526,47 +526,64 @@ func (s *Service) ProcessBlockRequestMessage(msg network.Message) error {
...
+			}
 		} else {
 			blockData.Header = optional.NewHeader(false, nil)
 		}
 
 		// body
 		if (blockRequest.RequestedData&2)>>1 == 1 {
-			blockData.Body = data.Body
+			retData, thisErr := s.blockState.GetBlockBody(hash)
+			if thisErr == nil && retData != nil {
```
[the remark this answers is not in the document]
thomasmodeneis: updated.

dot/core/service.go
```diff
@@ -482,14 +482,9 @@ func (s *Service) ProcessBlockAnnounceMessage(msg network.Message) error {
...
 	}
 
-	_, err = s.blockState.GetBlockData(header.Hash())
-	if err != nil && err.Error() == "Key not found" {
-		// send block request message
-		log.Debug("[core] sending new block to syncer", "number", blockAnnounceMessage.Number)
-		s.blockNumOut <- blockAnnounceMessage.Number
-	} else if err != nil {
-		return err
-	}
```
noot: why is this logic removed? we should keep it, but change `GetBlockData` to `GetBlock`

dot/core/service.go
```diff
@@ -544,56 +539,63 @@ func (s *Service) createBlockResponse(msg *network.BlockRequestMessage) (*networ
...
+		blockData.MessageQueue = optional.NewBytes(false, nil)
+		blockData.Justification = optional.NewBytes(false, nil)
 
 		// header
-		if msg.RequestedData&1 == 1 {
-			blockData.Header = data.Header
-		} else {
-			blockData.Header = optional.NewHeader(false, nil)
+		if blockRequest.RequestedData&1 == 1 {
+			log.Debug("msg.RequestedData & 1, GetHeader")
```
noot: change to trace, same for others

dot/core/service.go
```diff
@@ -544,56 +539,63 @@ func (s *Service) createBlockResponse(msg *network.BlockRequestMessage) (*networ
...
+				blockData.Header = retData.AsOptional()
+			}
 		}
-
 		// body
-		if (msg.RequestedData&2)>>1 == 1 {
-			blockData.Body = data.Body
-		} else {
-			blockData.Body = optional.NewBody(false, nil)
+		if blockRequest.RequestedData&2>>1 == 1 {
```
noot: hmm, no brackets around the `blockRequest.RequestedData&2`? let's have consistency with all the statements, don't think it makes a logic difference but still nice

dot/core/service_test.go
```diff
@@ -391,40 +369,197 @@ func TestService_ProcessBlockRequest(t *testing.T) {
...
+				BlockData: []*types.BlockData{
+					{
+						Hash:          optional.NewHash(true, bestHash).Value(),
+						Header:        optional.NewHeader(false, nil),
+						Body:          optional.NewBody(false, nil),
+						Justification: bds.Justification,
+					},
+				},
+			},
+		},
```
noot: nice tests :D

dot/core/state.go
```diff
@@ -34,13 +34,26 @@ type BlockState interface {
...
+	SetBlockBody(common.Hash, *types.Body) error
+	GetBlockBody(common.Hash) (*types.Body, error)
+
+	CompareAndSetBlockData(bd *types.BlockData) error
+
 	SetHeader(*types.Header) error
 	GetHeader(common.Hash) (*types.Header, error)
+
+	GetReceipt(common.Hash) ([]byte, error)
+	SetReceipt(common.Hash, []byte) error
```
noot: I don't see the `SetReceipt`, `SetMessageQueue` or `SetJustification` functions used anywhere in core, please remove if unused

dot/state/block_funcs_test.go
```diff
@@ -0,0 +1,182 @@
...
+	ks.Insert(kp)
+
+	cfg := &core.Config{
+		Runtime:         rt,
+		Keystore:        ks,
+		IsBabeAuthority: false,
+	}
+
+	s := newTestCoreService(t, cfg)
+	require.NotNil(t, s)
```
noot: s isn't used anywhere past this line, also state shouldn't be importing core, I don't see why core is needed at all here. you can make a blockState with `NewBlockState` or just `&BlockState{}`

dot/state/block_funcs_test.go
```diff
@@ -0,0 +1,182 @@
...
+)
+
+// testGenesisHeader is a test block header
+var testGenesisHeader = &types.Header{
+	Number:    big.NewInt(0),
+	StateRoot: trie.EmptyHash,
+}
+
+// newTestCoreService creates a new test core service
+func newTestCoreService(t *testing.T, cfg *core.Config) *core.Service {
```
noot: remove

lib/babe/state.go
```diff
@@ -33,10 +33,12 @@ type BlockState interface {
 	SubChain(start, end common.Hash) ([]common.Hash, error)
 	AddBlock(*types.Block) error
 	AddBlockWithArrivalTime(*types.Block, uint64) error
-	SetBlock(*types.Block) error
+
+	SetHeader(*types.Header) error
+	GetHeader(common.Hash) (*types.Header, error)
```
noot: BABE isn't using these, remove

dot/core/service_test.go
thomasmodeneis: done

dot/core/service.go
thomasmodeneis: done

dot/core/state.go
thomasmodeneis: done

dot/core/service.go
```diff
@@ -491,6 +491,10 @@ func (s *Service) ProcessBlockAnnounceMessage(msg network.Message) error {
 		return err
 	}
 
+	// send block request message
+	log.Debug("[core] sending new block to syncer", "number", blockAnnounceMessage.Number)
+	s.blockNumOut <- blockAnnounceMessage.Number
+
```
noot: remove this, this is duplicate code of line 488

lib/babe/state.go
thomasmodeneis: SetHeader is not used, but GetHeader is in use.

dot/state/block_funcs_test.go
thomasmodeneis: correct, that avoids all the boiler plate code, thanks for the hint! :)

dot/state/block_funcs_test.go
thomasmodeneis: yes

dot/core/service.go
thomasmodeneis: done

dot/core/service.go
thomasmodeneis: done

dot/state/block_funcs_test.go
```diff
@@ -0,0 +1,113 @@
...
+
+	kp, err := sr25519.GenerateKeypair()
+	require.Nil(t, err)
+
+	pubkey := kp.Public().Encode()
+	err = tt.Put(tests.AuthorityDataKey, append([]byte{4}, pubkey...))
+	require.Nil(t, err)
+
+	ks := keystore.NewKeystore()
+	ks.Insert(kp)
```
noot: remove this, not used anywhere

dot/core/service.go
```diff
@@ -544,56 +544,58 @@ func (s *Service) createBlockResponse(msg *network.BlockRequestMessage) (*networ
...
+		blockData.Receipt = optional.NewBytes(false, nil)
+		blockData.MessageQueue = optional.NewBytes(false, nil)
+		blockData.Justification = optional.NewBytes(false, nil)
 
 		// header
-		if msg.RequestedData&1 == 1 {
-			blockData.Header = data.Header
-		} else {
-			blockData.Header = optional.NewHeader(false, nil)
+		if (blockRequest.RequestedData & 1) == 1 {
```
noot: remove spaces for consistency, same with below

dot/core/service.go
```diff
@@ -544,56 +544,58 @@ func (s *Service) createBlockResponse(msg *network.BlockRequestMessage) (*networ
...
+				blockData.Header = retData.AsOptional()
+			}
 		}
-
 		// body
-		if (msg.RequestedData&2)>>1 == 1 {
-			blockData.Body = data.Body
-		} else {
-			blockData.Body = optional.NewBody(false, nil)
+		if (blockRequest.RequestedData & 2 >> 1) == 1 {
```
noot: brackets around the `blockRequest.RequestedData&2` only, like the ones below

dot/state/block_funcs_test.go
thomasmodeneis: good point, done

dot/core/service.go
thomasmodeneis: well, goimports things otherwise: `dot/core/service.go:559: File is not `goimports`-ed (goimports)` so I have to leave it like this.

dot/core/service.go
thomasmodeneis: done

dot/core/service.go
noot: okay that's weird :/ not sure why it's ok with the other ones then

PR #708, opened 2020-03-18
core: move block response handling to syncer; update request logic

### LINKED ISSUE #702: continuously request blocks while syncing
currently, blocks are only requested upon receiving a block announce or status message. however, we should be continuously requesting blocks until syncing is complete, instead of waiting for a prompt (ie. another message) should be easy to do once #690 is done.

### LINKED ISSUE #706: sync: don't request blocks starting at our best block
currently, we are starting our block requests at our best block number. however this causes problems since we may miss blocks.
eg:
Alice builds some blocks
Bob syncs with Alice, then builds a block, then needs to sync up to Alice again
however Bob will request start from the number of the block he just built, causing Alice's block with the same number to be missed

to fix this, we could:
 - keep track of last block received from each peer (probably not a great idea, adds lots of complexity)
- keep track of how many blocks we built since syncing, then requesting starting from our best number - blocks built since sync (how will we persist this?)
- proceed as normal, but if we get a "cannot find parent block in blocktree" error, send request again with a lower starting block number (adds network overhead, may slow syncing)

noot:
## Changes
- move block response handling to syncer
- start block response watcher in syncer, this goroutine will continuously send block requests until synced
- update syncer to keep track of where to start block requests from

## Tests:
```
go test ./dot/core
```

### Issues:
closes #702 closes #706


dot/core/service.go
```diff
@@ -200,10 +207,12 @@ func NewService(cfg *Config) (*Service, error) {
 			storageState:     cfg.StorageState,
 			transactionQueue: cfg.TransactionQueue,
 			isBabeAuthority:  false,
+			lock:             chanLock,
 			closed:           false,
 			syncer:           syncer,
 			syncLock:         syncerLock,
 			syncChan:         cfg.SyncChan,
+			respChan:         respChan,
```
ryanchristo: does this need to be appended to both service and and syncer submodule?

dot/core/service.go
```diff
@@ -574,145 +592,20 @@ func (s *Service) ProcessBlockRequestMessage(msg network.Message) error {
...
-
-	if existingData.MessageQueue == nil || (!existingData.MessageQueue.Exists() && bd.MessageQueue.Exists()) {
-		existingData.MessageQueue = bd.MessageQueue
-	}
-
-	if existingData.Justification == nil || (!existingData.Justification.Exists() && bd.Justification.Exists()) {
-		existingData.Justification = bd.Justification
-	}
+	log.Debug("[core] received BlockResponseMessage")
+	s.respChan <- msg.(*network.BlockResponseMessage)
```
ryanchristo: if only appended to syncer, we can use `s.syncer.respChan` herre, which might help make it more clear the `respChan` is specific to the `syncer` submodule

dot/node.go
```diff
@@ -150,8 +150,8 @@ func NewNode(cfg *Config, ks *keystore.Keystore) (*Node, error) {
 
 	// Message Channels (send and receive messages between services)
 
-	coreMsgs := make(chan network.Message)    // message channel from core service to network service
-	networkMsgs := make(chan network.Message) // message channel from network service to core service
+	coreMsgs := make(chan network.Message, 128)    // message channel from core service to network service
```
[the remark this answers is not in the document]
noot: not really, it's just a nice number

dot/core/service.go
noot: that wouldn't work since syncer.respChan is read only and service.respChan is write only

dot/core/service.go
noot: see comment below

dot/core/service.go
ryanchristo: right, maybe `syncer.respIn` and `syncer.respOut`? they are both specific to the syncer, correct?

noot: hmm like rename the Service one to `respOut`?

ryanchristo: Yea, I think the naming is confusing me a bit.

Can we use the syncer's write only channel rather than appending a new channel to the service called `respChan` (using the syncer more like a submodule that provides both read only and write only channels to the services which the `syncer` is appended to).

We already have `syncer.msgIn` (which receives messages within syncer from core service and network service) and `syncer.msgOut` (which sends messages within syncer to network service) and the syncer is being used by both the core service and the network service like a submodule.

It would be nice if we only needed to append `syncer` to each service, which in itself (like a submodule) provided all in and out channels specific to the syncer. Is that possible?

ryanchristo: thinking out loud a bit so might not make sense...

noot: okay I see what you mean.. I could change it to a read and write channel in syncer, so that core could write to it then syncer could read from it.. or I could separate them within syncer. I just didn't do that since having a write only channel within syncer sort of suggests that syncer writes to it. but I can change it

ryanchristo: I think it would make more sense to have the separation. At the least, we should probably add comments wherever we use channels explaining the ins and outs of each.

dot/core/sync.go
```diff
@@ -50,94 +61,261 @@ func NewSyncer(cfg *SyncerConfig) (*Syncer, error) {
...
 	// send block request message to network service
-	s.msgOut <- blockRequest
+	err = s.safeMsgSend(blockRequest)
+	if err != nil {
+		log.Error("[sync] Failed to send block request", "error", err)
+	}
+}
 
-	return nil
+func (s *Syncer) handleBlockResponse(msg *network.BlockResponseMessage) (int64, error) {
```
thomasmodeneis: This will conflict with #707

noot: yes, the contents of this function are nearly the same as `core.ProcessBlockResponse` from before

PR #714, opened 2020-03-20
dot/state, lib/trie, lib/common: move database helper functions into subpackage, remove trie.Database

### LINKED ISSUE #667: storage state, trie: move trie DB functions to storage state
currently in the trie package, there's database.go which contains various database functions. however, these should all be moved to state/storage.go, since that's where the DB functionality related to the trie should be.

also, once this is done we should be able to remove the `trie.Database` type and the DB field within `trie.Trie`. can also remove any `trie.SetDB` usages.

noot:
# Changes
- remove trie.Database
- move database storage functions for well known keys to lib/common/database (let me know if it makes sense, they should be in their own package but not sure about naming)
- cleaned up state accordingly

## Tests:
```
go test ./... -short
```

### Issues:
closes #667


dot/genesis.go
```diff
@@ -23,6 +23,7 @@ import (
 	"github.com/ChainSafe/gossamer/dot/core/types"
 	"github.com/ChainSafe/gossamer/dot/state"
 	"github.com/ChainSafe/gossamer/lib/common"
+	dbutils "github.com/ChainSafe/gossamer/lib/common/database"
```
ryanchristo: I don't think it makes sense to add a database directory to `lib/common` and then to have to modify the import because we have two packages named `database`.

noot:
@ryanchristo yeah I can rename it and put it somewhere else. I just put it there cause the db keys were there previously. what would you suggest? its own package or inside something else?

I don't think the functions can be moved to state. I tried and it just doesn't work because they are used in places where the state service doesn't exist yet. eg dot/genesis.go `storeGenesisData` and state/service.go `Initialize`. I could try moving them to state, but they can't have state.Service as a receiver

ryanchristo:
Hmmm.... I would want to make it some part of state given the names of the methods and the imports of the package. If they are pre-initialization, I would almost want to add them to the `dot` package.

noot:
okay I don't think they can do in dot since it will cause a circular dependency, I can try moving them to state

noot:
@ryanchristo I have moved the functions to state, and the keys back to common/db_keys.go where they were previously (need to keep them there at the moment because otherwise blocktree has a circular import)  it makes a lot more sense now lol

Issue #681, opened 2020-03-07
test block execution of built blocks


noot:
we should make sure that `Core_execute_block` succeeds on blocks that we build. there should be a check once `buildBlock` is finished.

PR #709, opened 2020-03-19
core: clean up methods and tests

ryanchristo:
## Changes

- migrate `CompareAndSetBlockData` method to block state
- organize core service methods and tests within core service

## Tests:
```
go test ./... -short
```

PR #715, opened 2020-03-21
dot/core: call core execute block

edwardmack:
## Changes
- Added call to Core_executeBlock in sync.go to handleBlockResponse

## Tests:
```
 go test ./dot/core/... 
```

### Issues:
closes: #681


dot/core/service.go
```diff
@@ -602,6 +603,142 @@ func (s *Service) createBlockResponse(msg *network.BlockRequestMessage) (*networ
...
+
+			res, err := s.executeBlock(bdEnc)
+			if err != nil {
+				log.Error("[core] failed to validate block", "err", err)
+				return err
+			}
+			// if executeBlock return a non-empty byte array something when wrong
+			if !reflect.DeepEqual(res, []byte{}) {
+				log.Error("[core] execute block call failed", "err", res)
+			}
```
noot: looks good, when you merge with development the logic for this function is moved to syncer.handleBlockResponse so you can move this there

dot/core/service.go
```diff
@@ -602,6 +603,142 @@ func (s *Service) createBlockResponse(msg *network.BlockRequestMessage) (*networ
...
+
+			res, err := s.executeBlock(bdEnc)
+			if err != nil {
+				log.Error("[core] failed to validate block", "err", err)
+				return err
+			}
+			// if executeBlock return a non-empty byte array something when wrong
+			if !reflect.DeepEqual(res, []byte{}) {
+				log.Error("[core] execute block call failed", "err", res)
```
noot: we want to return if this fails

lib/babe/build.go
```diff
@@ -99,6 +100,34 @@ func (b *Session) buildBlock(parent *types.Header, slot Slot) (*types.Block, err
...
+	// execute block
+	res, err := b.executeBlock(bdEnc)
+	if err != nil {
+		return nil, err
+	}
+	// if execute block return a non-empty byte array, something when wrong
+	if !reflect.DeepEqual(res, []byte{}) {
+		return nil, errors.New("[babe] execute block failed")
+	}
+	log.Trace("[babe] execute block")
```
noot: can remove this, we don't need to execute it after building. however if you could add this to the build block tests that would be good!

lib/runtime/imports.go
```diff
@@ -248,7 +248,7 @@ func ext_storage_root(context unsafe.Pointer, resultPtr int32) {
 
 //export ext_storage_changes_root
 func ext_storage_changes_root(context unsafe.Pointer, a, b, c int32) int32 {
-	log.Trace("[ext_storage_changes_root] executing...")
+	log.Trace("[ext_storage_changes_root] executing...", "a", a, "b", b, "c", c)
```
noot: don't need the paramters for now

dot/core/service.go
edwardmack: Any suggestion how I can get reference to runtime from within syncer.handleBlockResponse to call executeBlock?  Should I add reference to runtime in the syncer struct?  Would that be ugly because than syncer would require runtime?  Suggestions?

noot: I think adding runtime to the syncer for now is okay, it would be nice to have all the runtime functions in one place but we can figure that out later

edwardmack: ok

dot/core/service.go
edwardmack: ok

lib/babe/build.go
edwardmack: removed this, and added test to build_test.go

lib/runtime/imports.go
edwardmack: since this is a Trace log, I thought it would be useful to add information about values passed into the function.

dot/core/service.go
ryanchristo: This was mention in #709 as well, moving it to syncer would be ok but all the runtime methods should be in one place. I will address in #709 or open a new issue if you do not address here.

edwardmack:
I think I've address all comments, let me know. Thanks.

dot/core/sync.go
```diff
@@ -257,7 +263,33 @@ func (s *Syncer) handleBlockResponse(msg *network.BlockResponseMessage) (int64,
 				Body:   body,
 			}
 
-			// TODO: execute block and verify authorship right
+			// prepare block for sending to core_executeBlock,
+			//  core_executeBlock fails if Digest and Body data are sent
```
noot: it seems strange that it fails if the Body is sent, it needs the extrinsics to validate the block?

dot/core/sync.go
```diff
@@ -257,7 +263,33 @@ func (s *Syncer) handleBlockResponse(msg *network.BlockResponseMessage) (int64,
...
+					StateRoot:      header.StateRoot,
+					ExtrinsicsRoot: header.ExtrinsicsRoot,
+				},
+				Body: types.NewBody([]byte{}),
+			}
+
+			bdEnc, err := blockData.Encode()
+			if err != nil {
+				return 0, err
+			}
```
noot: could you move this logic to the `executeBlock` function, and have `executeBlock` take a block instead?

lib/runtime/imports.go
noot: okay that's fine

dot/core/sync.go
edwardmack: I agree, should I try to determine why it doesn't work with extrinsics before merging this?

dot/core/sync.go
edwardmack: done.

dot/core/sync.go
noot: that's probably a good idea, what extrinsics are you using to test? because some of them (eg. transfer) won't work because the account has no balance

edwardmack: I think I'm probably not using extrinsic correctly, I was using the array at line 124 in build_test, but I'm not using in correctly.  Do you have an example of an extrinsic I can try?

edwardmack: I found my issue, I should have been passing encoded `types.BlockData` into execute block (I was using  `types.Block`).

edwardmack:
@noot Please re-review, I think I've addressed all issues and now have 60% diff-coverage.  I addressed issue when calling executeBlock (I was passing in Block, when I should have been using BlockData).

edwardmack:
@noot The only way I can get core execute block to work is when I pass an encoded block that does not contain header digest or body.  I know this doesn't seem right, but the only combination of bytes that I could find to work was encode hash, encoded number, encode hash, encoded hash.  Add anything else seems to cause an error.

I wonder if we'll see different behavior when we update runtime?  Not sure if we should commit this, but it's passing tests now and two clients do sync without crashing.

dot/core/sync_test.go
```diff
@@ -49,6 +59,11 @@ func newTestSyncer(t *testing.T, cfg *SyncerConfig) *Syncer {
 		cfg.MsgOut = make(chan network.Message)
 	}
 
+	if cfg.Runtime == nil {
+		cfg.Runtime = runtime.NewTestRuntime(t, tests.POLKADOT_RUNTIME)
+		//cfg.Runtime = runtime.NewTestRuntime(t, tests.ED_RUNTIME)
```
noot: remove

dot/core/sync_test.go
```diff
@@ -371,3 +386,90 @@ func TestWatchForResponses_MissingBlocks(t *testing.T) {
...
+func TestHandleBlockResponse_NoBlockData(t *testing.T) {
+	syncer := newTestSyncer(t, nil)
+	msg := &network.BlockResponseMessage{
+		ID:        0,
+		BlockData: nil,
+	}
+	res, err := syncer.handleBlockResponse(msg)
+	require.Nil(t, err)
+
+	t.Logf("res %v", res)
```
noot: remove

Issue #717, opened 2020-03-23
state: set number->hash mapping for canonical chain only


noot:
right now, `AddBlock` in state updates the number->blockhash mapping every time it imports a block, however this is wrong and will mess up BlockResponses and other features that rely on getting a block on the canonical chain by number. can be done once #716 is merged

PR #719, opened 2020-03-24
dot/network, lib/blocktree: use sync.Map in network, update blocktree subchain error handling

### LINKED ISSUE #718: ignore BlockResponses that aren't requested by us
in the network state, we shouldn't send BlockResponses to core that aren't requested by us

also, remove IDs that we've received a response for, and update the mapping to use sync.Map to avoid `fatal error: concurrent map writes`

noot:
## Changes
- update gossip, status, syncer in network to use sync.Map to prevent panic: concurrent map writes
- update network message handler to only send BlockResponses to core if it's a response to one of our requests
- update blocktree subchain error handling (was causing sigabort before when erroring)
- slightly improve core.syncer logic

## Tests:
```
go test ./... -short
```

as well, the usual running multiple nodes and syncing:
```
make gossamer
./bin/gossamer --key alice 
```
etc for other nodes. please try and make sure nothing goes wrong for >2 nodes

### Issues:
closes #718


dot/state/block.go
```diff
@@ -347,13 +347,13 @@ func (bs *BlockState) AddBlockWithArrivalTime(block *types.Block, arrivalTime ui
...
 
-	// TODO: update to use leftmost path
-
+	// TODO: update to use canonical path
 	err = bs.setBestBlockHashKey(hash)
 	if err != nil {
 		return err
 	}
 
+	// TODO: only set number->hash mapping for our canonical chain, otherwise, this messes up BlockResponses
```
noot: #717 same with above todo

dot/core/service.go
```diff
@@ -512,8 +515,12 @@ func (s *Service) createBlockResponse(msg *network.BlockRequestMessage) (*networ
 
 	switch c := msg.StartingBlock.Value().(type) {
 	case uint64:
+		if c == 0 {
+			c = 1
+		}
+
 		block, err := s.blockState.GetBlockByNumber(big.NewInt(0).SetUint64(c))
-		if err != nil {
+		if err != nil || block == nil {
```
thomasmodeneis: do you think that s.blockState.GetBlockByNumber returning a nil block and no error is a acceptable logic for the method ? shall we be doing double checks like this or update the logic in that method to avoid having to double check it ?

noot: yeah you're right I don't think it will ever return nil, will remove the `block == nil` part

dot/core/service.go
```diff
@@ -496,7 +496,10 @@ func (s *Service) ProcessBlockAnnounceMessage(msg network.Message) error {
 
 // ProcessBlockRequestMessage processes a block request message, returning a block response message
 func (s *Service) ProcessBlockRequestMessage(msg network.Message) error {
-	blockRequest := msg.(*network.BlockRequestMessage)
+	blockRequest, ok := msg.(*network.BlockRequestMessage)
```
thomasmodeneis: given the method is ProcessBlockRequestMessage, would be appropriate that the logic that validates the input is a helper placed outside the func, and the func signature is changed to accept only BlockRequestMessage. Would you agree ?

noot: yes, I agree, I'll update handleReceivedMessage to cast it beforehand

dot/core/service.go
```diff
@@ -512,6 +525,10 @@ func (s *Service) createBlockResponse(blockRequest *network.BlockRequestMessage)
 
 	switch c := blockRequest.StartingBlock.Value().(type) {
 	case uint64:
+		if c == 0 {
```
thomasmodeneis: I am curious to know why when someone does a blockrequest with starting block 0 we silently change it to 1 and move on ?

dot/core/sync.go
```diff
@@ -260,10 +262,12 @@ func (s *Syncer) handleBlockResponse(msg *network.BlockResponseMessage) (int64,
 
 			err = s.blockState.AddBlock(block)
 			if err != nil {
-				if err.Error() == "cannot find parent block in blocktree" {
+				if err.Error() == "cannot find parent block in blocktree" && header.Number.Cmp(big.NewInt(0)) != 0 {
 					return 0, err
 				} else if strings.Contains(err.Error(), "cannot add block to blocktree that already exists") {
 					// this is fine
+				} else if header.Number.Cmp(big.NewInt(0)) == 0 {
```
thomasmodeneis: would be nice to not have empty statements, imho.

dot/network/gossip_test.go
```diff
@@ -155,26 +155,26 @@ func TestGossip(t *testing.T) {
 		t.Error("node A timeout waiting for message")
 	}
 
-	hasSeenB := nodeB.gossip.hasSeen[TestMessage.IDString()]
-	if hasSeenB == false {
+	hasSeenB, ok := nodeB.gossip.hasSeen.Load(TestMessage.IDString())
+	if !ok || hasSeenB.(bool) == false {
```
thomasmodeneis: statement would look better if using a if condition with two expressions: `if hasSeenB, ok := nodeB.gossip.hasSeen.Load(TestMessage.IDString()); !ok || hasSeenB.(bool) == false {`

dot/network/gossip_test.go
```diff
@@ -155,26 +155,26 @@ func TestGossip(t *testing.T) {
...
 		t.Error(
 			"node B did not receive block request message from node A",
 			"\nreceived:", hasSeenB,
 			"\nexpected:", true,
 		)
 	}
 
-	hasSeenC := nodeC.gossip.hasSeen[TestMessage.IDString()]
-	if hasSeenC == false {
+	hasSeenC, ok := nodeC.gossip.hasSeen.Load(TestMessage.IDString())
```
thomasmodeneis: same here

dot/network/gossip_test.go
```diff
@@ -155,26 +155,26 @@ func TestGossip(t *testing.T) {
...
 		t.Error(
 			"node C did not receive block request message from node B",
 			"\nreceived:", hasSeenC,
 			"\nexpected:", true,
 		)
 	}
 
-	hasSeenA := nodeA.gossip.hasSeen[TestMessage.IDString()]
-	if hasSeenA == false {
+	hasSeenA, ok := nodeA.gossip.hasSeen.Load(TestMessage.IDString())
```
thomasmodeneis: same here

dot/network/service.go
```diff
@@ -192,8 +192,8 @@ func (s *Service) updateNetworkState() {
 func (s *Service) receiveCoreMessages() {
 	for {
 		// receive message from core service
-		msg := <-s.msgRec
-		if msg == nil {
+		msg, ok := <-s.msgRec
```
thomasmodeneis: same here

dot/network/service.go
```diff
@@ -321,9 +321,22 @@ func (s *Service) handleMessage(peer peer.ID, msg Message) {
...
 		if s.noStatus || s.status.confirmed(peer) {
 
-			err := s.safeMsgSend(msg)
-			if err != nil {
-				log.Error("[network] Failed to send message", "error", err)
+			if resp, ok := msg.(*BlockResponseMessage); ok {
+				if s.syncer.hasRequestedBlockID(resp.ID) {
+					err := s.safeMsgSend(msg)
+					if err != nil {
+						log.Error("[network] Failed to send message", "error", err)
```
thomasmodeneis: maybe nice to log the message that could not be delivered or at least ID so it could potentially be tracked ?

dot/network/service.go
```diff
@@ -321,9 +321,22 @@ func (s *Service) handleMessage(peer peer.ID, msg Message) {
...
+					}
+
+					s.syncer.removeRequestedBlockID(resp.ID)
+				} else {
+					// ignore for now, but eventually we want to re-gossip once gossip is improved
+				}
+			} else {
+				err := s.safeMsgSend(msg)
+				if err != nil {
+					log.Error("[network] Failed to send message", "error", err)
```
thomasmodeneis: maybe nice to log the message that could not be delivered or at least ID so it could potentially be tracked ?

dot/network/service.go
```diff
@@ -375,14 +388,17 @@ func (s *Service) Peers() []common.PeerInfo {
...
-			msg := s.status.peerMessage[p]
-			peers = append(peers, common.PeerInfo{
-				PeerID:          p.String(),
-				Roles:           msg.Roles,
-				ProtocolVersion: msg.ProtocolVersion,
-				BestHash:        msg.BestBlockHash,
-				BestNumber:      msg.BestBlockNumber,
-			})
+			m, ok := s.status.peerMessage.Load(p)
+			msg := m.(*StatusMessage)
```
thomasmodeneis: you are doing the cast before checking the ok, this could also be simplified.

dot/network/status.go
```diff
@@ -36,22 +37,27 @@ const ExpireStatusInterval = SendStatusInterval + time.Minute
...
+		peerConfirmed: &sync.Map{},
+		peerMessage:   &sync.Map{},
 	}
 }
 
 // confirmed returns true if peer is confirmed
 func (status *status) confirmed(peer peer.ID) bool {
-	return !status.peerConfirmed[peer].IsZero()
+	t, ok := status.peerConfirmed.Load(peer)
+	if !ok {
```
thomasmodeneis: this could be simplified in a oneliner.

dot/network/status.go
```diff
@@ -193,10 +199,13 @@ func (status *status) expireStatus(ctx context.Context, peer peer.ID) {
 	time.Sleep(ExpireStatusInterval)
 
 	// get time of last confirmed status message
-	lastConfirmed := status.peerConfirmed[peer]
+	lastConfirmed, ok := status.peerConfirmed.Load(peer)
```
thomasmodeneis: this could be simplified in a oneliner.

dot/network/syncer.go
```diff
@@ -38,28 +39,31 @@ func newSyncer(host *host, blockState BlockState, syncChan chan<- *big.Int) *syn
...
 func (s *syncer) addRequestedBlockID(blockID uint64) {
 	log.Trace("[network] Adding block to network syncer...", "block", blockID)
-	s.requestedBlockIDs[blockID] = true
+	s.requestedBlockIDs.Store(blockID, true)
 }
 
 // hasRequestedBlockID returns true if the block id has been requested
 func (s *syncer) hasRequestedBlockID(blockID uint64) bool {
-	requested := s.requestedBlockIDs[blockID]
+	requested, ok := s.requestedBlockIDs.Load(blockID)
```
thomasmodeneis: this could be simplified in a oneliner.

dot/core/service.go
noot: block 0 is the genesis block, so it has no parent, so it will error if we attempt to add it to the block tree. it's handled on the receiver side but also better not to send it in the first place, since we already know that the peer has the same genesis block as us

dot/network/service.go
noot: msg needs to be defined in a separate line so it can be used at line 202 and onwards

dot/network/service.go
noot: added block request ID to log

dot/network/service.go
noot: this statement is for all other message types, and ID of the message isn't really relevant to those types

dot/network/status.go
noot: not really, since t is used again at line 60, it needs to be defined outside of the if statement

dot/network/status.go
noot: same as above, `lastConfirmed` is used again at line 208

PR #725, opened 2020-03-26
dot/state, lib/blocktree: set block number->hash mapping for current chain only

### LINKED ISSUE #717: state: set number->hash mapping for canonical chain only

noot:
## Changes
- addisBlockOnCurrentChain function to block state to check if block we are adding is on our current chain, if so, update number->hash mapping in db, otherwise don't 
- fix blocktree deepestLeaf function, wasn't correct before

## Tests:
```
go test ./... -short
```

### Issues:
closes #717

### DIFF
--- a/dot/state/block.go
+++ b/dot/state/block.go
@@ -336,15 +336,14 @@ func (bs *BlockState) AddBlockWithArrivalTime(block *types.Block, arrivalTime ui
 	hash := block.Header.Hash()
 
 	// set best block key if this is the highest block we've seen
-	if block.Header.Number.Cmp(bs.highestBlockHeader.Number) == 1 || hash == bs.BestBlockHash() {
+	if hash == bs.BestBlockHash() {
 		err = bs.setBestBlockHashKey(hash)
 		if err != nil {
 			return err
 		}
 	}
 
-	// TODO: only set number->hash mapping for our canonical chain, otherwise, this messes up BlockResponses
-	// store number to hash
+	// Tonly set number->hash mapping for our current chain
 	if ok, err := bs.isBlockOnCurrentChain(block.Header); ok && err == nil {
 		err = bs.db.Put(headerHashKey(block.Header.Number.Uint64()), hash.ToBytes())
 		if err != nil {
--- a/dot/state/block.go
+++ b/dot/state/block.go
@@ -344,7 +344,8 @@ func (bs *BlockState) AddBlockWithArrivalTime(block *types.Block, arrivalTime ui
 	}
 
 	// only set number->hash mapping for our current chain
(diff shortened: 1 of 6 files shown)


dot/state/block_test.go
```diff
@@ -214,3 +210,94 @@ func TestGetSlotForBlock(t *testing.T) {
...
+	genesisHeader := &types.Header{
+		Number:    big.NewInt(0),
+		StateRoot: trie.EmptyHash,
+	}
+
+	bs := newTestBlockState(genesisHeader)
+	currChain, branchChains := addBlocksToState(bs, 8)
+
+	for _, header := range currChain {
+		ok, err := bs.isBlockOnCurrentChain(header)
```
ryanchristo: same as above

dot/state/block_test.go
```diff
@@ -214,3 +210,94 @@ func TestGetSlotForBlock(t *testing.T) {
...
+		if err != nil {
+			t.Fatal(err)
+		}
+		if !ok {
+			t.Fatalf("Fail: expected block %s to be on current chain", header.Hash())
+		}
+	}
+
+	for _, header := range branchChains {
+		ok, err := bs.isBlockOnCurrentChain(header)
```
ryanchristo: same as above

dot/state/block.go
```diff
@@ -336,19 +336,20 @@ func (bs *BlockState) AddBlockWithArrivalTime(block *types.Block, arrivalTime ui
...
 		}
 	}
 
-	// TODO: only set number->hash mapping for our canonical chain, otherwise, this messes up BlockResponses
-	// store number to hash
-	err = bs.db.Put(headerHashKey(block.Header.Number.Uint64()), hash.ToBytes())
-	if err != nil {
-		return err
+	// only set number->hash mapping for our current chain
+	var ok bool
```
ryanchristo: this can also be declared on the line below since it is not used outside the condition block whether you keep `ok` or change it to something more descriptive

dot/state/block.go
```diff
@@ -336,19 +336,20 @@ func (bs *BlockState) AddBlockWithArrivalTime(block *types.Block, arrivalTime ui
...
+	if ok, err = bs.isBlockOnCurrentChain(block.Header); ok && err == nil {
```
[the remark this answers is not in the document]
noot: yeah good point, will change

dot/state/block.go
noot: it's defined outside because otherwise there's a lint error saying definition of err is being shadowed

PR #726, opened 2020-03-27
dot/core, lib/blocktree, lib/database: create error definitions

### LINKED ISSUE #305: Error Definitions
If we are going to be using custom error messages for specific packages we need to begin defining them in a meaningful way. For instance, the `p2p` package should have pre-defined error messages that can be referenced and easily understood rather than utilizing `errors.New`. We can then include the meaning of these errors in our godocs through regular comments. 

## Possible Solution

Include custom errors at the top of the file or include a `errors.go` in the package

`// messageErr...`
`const messageErr = errors.New("[P2P:: ] unsupported message type")`

noot:
## Changes
- use defined errors in core, blocktree, database (I just did these ones for now since these are the ones that were doing error comparisons)

## Tests:
```
go test ./... -short
```

### Issues:
partially closes #305

### DIFF
--- a/dot/core/errors.go
+++ b/dot/core/errors.go
@@ -5,21 +5,35 @@ import (
 	"fmt"
 )
 
+// ErrNilBlockState is returned when BlockState is nik
 var ErrNilBlockState = errors.New("cannot have nil BlockState")
+
+// ErrNilStorageState is returned when StorageState is nik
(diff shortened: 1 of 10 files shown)


noot:
@ryanchristo it's go standard to start constant errors with `Err`

and yeah, if the same error is being used in many places it might be nice to have them in common or something

PR #728, opened 2020-03-28
lib/babe: check whether block producer has equivocated

### LINKED ISSUE #553: BABE: check whether block producer has equivocated
in `verifyAuthorshipRight` in BABE, we need to check whether the block producer has equivocated for that block (ie. produced multiple blocks with the same number, or blocks that otherwise conflict) 
will need to connect BABE to the blocktree for this.

thomasmodeneis:
## Changes
-  fixed other minors on related tests and other small code review
- init impl for VerifyHeaderValid 
- 

## Tests:
```
go test ./lib/babe/... -test.run ^TestVerifyAuthorshipRight$
```

### Issues:
closes #553


lib/blocktree/blocktree.go
```diff
@@ -68,6 +69,28 @@ func (bt *BlockTree) GenesisHash() Hash {
...
+		return err
+	}
+
+	// Check if it already exists
+	n := bt.getNode(header.Hash())
+	if n != nil {
+		return fmt.Errorf("cannot add block to blocktree that already exists: hash=%s", n.hash)
+	}
+
+	return nil
```
noot: this seems to be a duplicate of blocktree.AddNode? 

also, you shouldn't add the block to the blocktree when verifying, and it's not actually possible to complete the verification from the blocktree, since you need the block digests. change this to a function that calculates the depth of the block you want to add with `bt.getNode(parentHash).depth` and return the hashes of all other nodes with the same depth

lib/blocktree/leaves.go
```diff
@@ -40,7 +40,7 @@ func (ls leafMap) deepestLeaf() *node {
 		if max.Cmp(n.depth) < 0 {
 			max = n.depth
 			dLeaf = n
-		} else if max.Cmp(n.depth) == 0 && n.arrivalTime > dLeaf.arrivalTime {
+		} else if max.Cmp(n.depth) == 0 && n.arrivalTime > dLeaf.arrivalTime { // TODO: arrivalTime may lead to nil pointer ?
```
noot: the nodes in leafMap should never be nil

lib/blocktree/blocktree.go
thomasmodeneis: I addressed the point mentioned, still I am not sure if this is correct. Could you please review it again ? :)

lib/babe/verify.go
```diff
@@ -109,6 +110,23 @@ func (b *Session) verifyAuthorshipRight(slot uint64, header *types.Header) (bool
...
+	}
+
+	for key, v := range hashes {
+		//get header
+		h, err := b.blockState.GetHeader(key)
+		if err != nil {
+			return false, fmt.Errorf("could not verify digests, key %s, depth, %d", key, v)
+		}
+		//compare digest
+		if reflect.DeepEqual(h.Digest, header.Digest) {
```
noot: you need to look at the Seal digest specifically, which should be the last item in the Digest array

lib/babe/verify.go
```diff
@@ -109,6 +110,23 @@ func (b *Session) verifyAuthorshipRight(slot uint64, header *types.Header) (bool
...
+			return false, fmt.Errorf("duplicated digest")
```
noot: don't need to return an error, just `false`

lib/blocktree/blocktree.go
```diff
@@ -68,6 +70,50 @@ func (bt *BlockTree) GenesisHash() Hash {
 	return bt.head.hash
 }
 
+// GetAllHashesForParentDepth will get all hashes with the same dept for a given header
+func (bt *BlockTree) GetAllHashesForParentDepth(header *types.Header) (map[common.Hash]*big.Int, error) {
```
noot: the name of this is a confusing, you should be getting all the hashes at the same depth as the header you pass in

lib/blocktree/blocktree.go
```diff
@@ -68,6 +70,50 @@ func (bt *BlockTree) GenesisHash() Hash {
...
+	hashes := map[common.Hash]*big.Int{}
+
+	parent := bt.getNode(header.ParentHash)
+	if parent == nil {
+		err := fmt.Errorf("cannot find parent block in blocktree")
+		log.Error("Could not getNode for ParentHash", "header.ParentHash", header.ParentHash, "err", err)
+		return hashes, err
+	}
+
+	for _, child := range bt.head.children {
```
noot: how does this work? it isn't recursive so it won't go past depth 1 of the tree

lib/blocktree/blocktree.go
```diff
@@ -68,6 +70,50 @@ func (bt *BlockTree) GenesisHash() Hash {
...
+		if child.depth.Cmp(parent.depth) == -1 {
+			log.Debug("breaking range bt.head.children", "child.hash", child.hash.String(), "child.depth", child.depth)
```
noot: this should never be the case, child depth should always be greater than parent

lib/blocktree/blocktree.go
```diff
@@ -68,6 +70,50 @@ func (bt *BlockTree) GenesisHash() Hash {
...
+			break
+		} else if child.depth.Cmp(parent.depth) == 0 {
```
noot: same with this case, child depth should always be greater than parent. this is also comparing `header`'s parent to the children of the blocktree's head(root node) which doesn't really make sense

lib/blocktree/blocktree.go
```diff
@@ -68,6 +70,50 @@ func (bt *BlockTree) GenesisHash() Hash {
...
+				hashes[child.hash] = parent.depth
+			}
+		}
+
+		hashes = bt.getHashes(child, hashes)
+	}
+	return hashes, nil
+}
+
+func (bt *BlockTree) getHashes(n *node, hashes map[common.Hash]*big.Int) map[common.Hash]*big.Int {
```
noot: I'm not sure what this is doing, but you don't need to return a map, an array of hashes is all you need. since all the nodes should be at the same depth

dot/core/state.go
```diff
@@ -33,6 +33,7 @@ type BlockState interface {
 	BestBlock() (*types.Block, error)
 	SubChain(start, end common.Hash) ([]common.Hash, error)
 	AddBlock(*types.Block) error
+	GetAllHashesForParentDepth(*types.Header) (map[common.Hash]*big.Int, error)
```
noot, line 36: name is unclear, we are getting nodes at the same depth as the provided header.

should also return just an array of hashes, not a map

lib/babe/verify.go
thomasmodeneis: outdated

lib/babe/verify.go
thomasmodeneis: well, if I dont return a error how can I test that it happened?

lib/blocktree/blocktree.go
thomasmodeneis: outdated

lib/blocktree/blocktree.go
thomasmodeneis: outdated

lib/blocktree/blocktree.go
thomasmodeneis: outdated

dot/core/state.go
thomasmodeneis: updated

lib/blocktree/blocktree.go
thomasmodeneis: is it ? I was thinking that the bt.head was the genesis, and all its children are then all the blocks following it. Do you have any other suggestions ?

noot: no, the children aren't the blocks that follow it sequentially, the children are the blocks that all have the block as the parent hash. the head isn't necessarily the genesis, right now it is only because we don't have finality. I will take a look at everything and see how this works

lib/babe/build_test.go
```diff
@@ -61,36 +61,43 @@ func TestSeal(t *testing.T) {
...
+		if outAndProof == nil {
+			t.Fatal("proof was nil when over threshold")
+		}
+
+		babesession.slotToProof[slotNumber] = outAndProof
 	}
 
-	babesession.slotToProof[slotNumber] = outAndProof
+	//t.Log(babesession.authorityData[0].ID.Encode())
+	//t.Log(babesession.keypair.Public().Encode())
```
noot, line 84: remove

lib/babe/verify.go
```diff
@@ -109,6 +112,54 @@ func (b *Session) verifyAuthorshipRight(slot uint64, header *types.Header) (bool
...
 	}
 
-	// TODO: check if the producer has equivocated, ie. have they produced a conflicting block?
+	// #553 check if the producer has equivocated, ie. have they produced a conflicting block?
+	hashes, err := b.blockState.GetAllHashesForParentDepth(header)
+	if err != nil {
+		return false, err
+	}
+
+	log.Debug("Got AllHashesForParentDepth", "len(hashes)", len(hashes))
```
noot, line 121: remove

lib/babe/verify.go
```diff
@@ -109,6 +112,54 @@ func (b *Session) verifyAuthorshipRight(slot uint64, header *types.Header) (bool
...
+
+	for key, v := range hashes {
+		//get header
+		currentHeader, err := b.blockState.GetHeader(key)
+		if err != nil {
+			return false, fmt.Errorf("could not verify digests, key %s, depth, %d", key, v)
```
noot, line 127: maybe don't error but just `continue` instead.

lib/babe/verify.go
```diff
@@ -109,6 +112,54 @@ func (b *Session) verifyAuthorshipRight(slot uint64, header *types.Header) (bool
...
+		}
+
+		//compare digest
+		currentBlockProducerIndex := getBlockProducerIndex(currentHeader)
+		existingBlockProducerIndex := babeHeader.BlockProducerIndex
+
+		log.Debug("compare Digest", "currentBlockProducerIndex", currentBlockProducerIndex,
```
noot, line 134: remove or change to trace

lib/babe/verify.go
```diff
@@ -109,6 +112,54 @@ func (b *Session) verifyAuthorshipRight(slot uint64, header *types.Header) (bool
...
+			"existingBlockProducerIndex", existingBlockProducerIndex)
+
+		if reflect.DeepEqual(currentBlockProducerIndex, existingBlockProducerIndex) {
+			return false, fmt.Errorf("duplicated SealDigest")
```
noot, line 138: I don't think reflect is needed for the comparison, but it doesn't hurt. also, the error message should say "block producer equivocated" or something like that

lib/babe/verify.go
```diff
@@ -109,6 +112,54 @@ func (b *Session) verifyAuthorshipRight(slot uint64, header *types.Header) (bool
...
+
 	return true, nil
 }
+
+func getBlockProducerIndex(header *types.Header) (blockProducerIndex uint64) {
+	preDigestBytes := header.Digest[0]
+
+	digestItem, err := types.DecodeDigestItem(preDigestBytes)
+	if err != nil {
+		return
```
noot, line 150: return error in this function, otherwise if there is an error it will return 0 which looks correct

lib/babe/verify_test.go
```diff
@@ -84,23 +92,79 @@ func TestVerifySlotWinner(t *testing.T) {
...
-		t.Fatal(err)
+	testsCases := []struct {
+		description                     string
+		parentHeader                    *types.Header
+		expectedErr                     error
+		authorshipRight                 bool
+		expectedErrAfterAuthorshipRight error
+	}{
+		{
+			description:                     "test verify block with existing parent",
```
noot, line 103: I'm not sure what this description means? we don't need to test the `AddBlock` function here

lib/blocktree/blocktree.go
noot: this name of this doesn't correctly explain what this function should do, also it doesn't need to return a map, just an array of hashes

lib/blocktree/blocktree.go
```diff
@@ -69,6 +71,37 @@ func (bt *BlockTree) GenesisHash() Hash {
...
+			"bt.head.children", len(bt.head.children))
+
+		if child.depth.Cmp(parentDepthPlusOne) == 0 {
+			if hashes[child.hash] == nil {
+				log.Debug("adding child.hash", "child.hash", child.hash.String(), "child.depth", child.depth)
+				hashes[child.hash] = child.depth
+			}
+		}
+	}
+	return hashes, nil
```
noot, line 102: I don't think this function returns all nodes with the same depth. there might be multiple branches of the tree with nodes with the same depth, the nodes that are children of the parent arent necessarily all the nodes at that depth.

also, I'm not sure why you need to return a map. I think an array of hashes is sufficient

lib/babe/build_test.go
thomasmodeneis: done

lib/babe/verify.go
thomasmodeneis: done

lib/babe/verify.go
ryanchristo: would be nice to log error as debug if you continue as noot suggests

thomasmodeneis: ok

lib/babe/verify.go
thomasmodeneis: ops, reflect was required before but now just a simple `==` would work as well. Also updated err to `block producer equivocated` as suggested.

lib/babe/verify.go
thomasmodeneis: ok

lib/babe/verify.go
thomasmodeneis: ok

lib/babe/verify_test.go
thomasmodeneis: oh, well when I started this PR, the test was using a block parent that was not persisted into the database. 
There are then two tests now, the old one: `test verify block with not existing parent` that should fail with `cannot find parent block in blocktree` and then the new one for the new testcase: `test verify block with existing parent` that does not fail with that error on that step, but should fail further on with `block producer equivocated`.

lib/blocktree/blocktree.go
thomasmodeneis: I am returning a map to avoid duplicated results, since using a array would potentially cause duplicated results...

lib/blocktree/blocktree.go
thomasmodeneis: true, removed

lib/blocktree/blocktree.go
thomasmodeneis: right, that makes more sense...you are right, I updated it a bit to reflect this line of thinking, could you please check again ? :)

lib/blocktree/blocktree.go
noot: there shouldn't be any duplicated results, since you can't add a block to the blocktree if a block with that hash is already in it

lib/babe/verify_test.go
noot: okay, I don't think it's really necessary to test but it's ok to leave it in

lib/babe/verify.go
```diff
@@ -24,6 +24,8 @@ import (
 	babetypes "github.com/ChainSafe/gossamer/lib/babe/types"
 )
 
+var ErrProducerEquivocated = fmt.Errorf("block producer equivocated")
```
ryanchristo, line 27: should we start an `errors.go` file for the babe package?

lib/babe/verify_test.go
```diff
@@ -104,3 +97,61 @@ func TestVerifyAuthorshipRight(t *testing.T) {
...
+
+	err = babesession.blockState.AddBlock(block2)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	ok, err = babesession.verifyAuthorshipRight(slotNumber, block2.Header)
+	require.NotNil(t, err)
+	require.False(t, ok)
+	require.Equal(t, ErrProducerEquivocated, err)
```
ryanchristo, line 156: this is nice, we should definitely try to adopt this throughout our tests

lib/blocktree/blocktree_test.go
```diff
@@ -158,7 +163,9 @@ func TestBlockTree_Subchain(t *testing.T) {
 	}
 
 	extraBlock.Header.Hash()
-	bt.AddBlock(extraBlock, 0)
+	err := bt.AddBlock(extraBlock, 0)
+	//TODO are we really expecting a err here ?
```
ryanchristo, line 167: ?

noot: will remove

lib/babe/verify.go
noot: probably, I can move the errors from this file into an errors.go

lib/blocktree/blocktree.go
```diff
@@ -133,6 +100,39 @@ func (bt *BlockTree) AddBlock(block *types.Block, arrivalTime uint64) error {
...
+	if bt.head.depth.Cmp(depth) == 0 {
+		hashes = append(hashes, bt.head.hash)
+		return hashes
+	}
+
+	return bt.head.getNodesWithDepth(depth, hashes)
+}
+
+// getNodesWithDepth returns all descendent nodes with the desired depth
+func (n *node) getNodesWithDepth(depth *big.Int, hashes []common.Hash) []common.Hash {
```
thomasmodeneis: are you sure that a func for  (n *node) should be together with others for (bt *BlockTree) in the same file ?

noot: it's been moved

thomasmodeneis: oh cool

PR #733, opened 2020-03-31
cmd, dot: overwrite config with genesis

### LINKED ISSUE #713: load name, id, protocol id, bootnodes from genesis upon startup
currently, these are being set in config.toml, however they should be retrieved from the genesis file (with the LoadGenesisData function). we can add additional bootnodes in the config, or override the protocol ID in the config, but they should be loaded from the genesis data initially.

ryanchristo:
## Changes

- load genesis configuration values (`Name`, `ID`, `Bootnodes`, `ProtocolID`) from genesis file while setting the node configuration and overwrite configuration if configuration values do not match.

- add `name` flag for telemetry

## Tests:
```
go test ./cmd/gossamer
go test ./dot
```

### Issues:
related to #713, no longer closes - ***updated***

### DIFF
--- a/cmd/gossamer/config.go
+++ b/cmd/gossamer/config.go
@@ -22,13 +22,14 @@ import (
 	"strings"
 
 	"github.com/ChainSafe/gossamer/dot"
+	"github.com/ChainSafe/gossamer/lib/genesis"
 
 	log "github.com/ChainSafe/log15"
 	"github.com/urfave/cli"
 )
 
-// loadConfigFile loads a default config file if --node is specified, a specific config if --config is specified,
-// or the default gossamer config otherwise.
+// loadConfigFile loads a default config file if --node is specified, a specific
+// config if --config is specified, or the default gossamer config otherwise.
 func loadConfigFile(ctx *cli.Context) (cfg *dot.Config, err error) {
 	// check --node flag and load node configuration from defaults.go
 	if name := ctx.GlobalString(NodeFlag.Name); name != "" {
@@ -71,6 +72,26 @@ func loadConfigFile(ctx *cli.Context) (cfg *dot.Config, err error) {
 	return cfg, nil
 }
 
+// createInitConfig creates the configuration required to initialize a dot node
+func createInitConfig(ctx *cli.Context) (cfg *dot.Config, err error) {
+	cfg, err = loadConfigFile(ctx)
+	if err != nil {
+		log.Error("[cmd] Failed to load toml configuration", "error", err)
+		return nil, err
+	}
+
+	// set global configuration values
+	setDotGlobalConfig(ctx, &cfg.Global)
+
+	// set init configuration values
+	setDotInitConfig(ctx, &cfg.Init)
+
+	// ensure configuration values match genesis and overwrite with genesis
+	updateDotConfigFromGenesis(ctx, cfg)
+
+	return cfg, nil
+}
+
 // createDotConfig creates a new dot configuration from the provided flag values
 func createDotConfig(ctx *cli.Context) (cfg *dot.Config, err error) {
 	cfg, err = loadConfigFile(ctx)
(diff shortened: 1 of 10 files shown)


cmd/gossamer/config.go
```diff
@@ -240,10 +251,57 @@ func setDotRPCConfig(ctx *cli.Context, cfg *dot.RPCConfig) {
...
+
+// updateFromGenesis updates the configuration based on the genesis file values
+func updateFromGenesis(cfg *dot.Config) {
+
+	// load Genesis from genesis configuration file
+	gen, err := genesis.LoadGenesisFromJSON(cfg.Global.Genesis)
+	if err != nil {
+		log.Error("[cmd] failed to load genesis from file", "error", err)
+		return // exit
+	}
```
noot, line 270: unless you are doing `init`, the genesis file isn't relevant anymore, not sure why it's in the config. if you are running gossamer when it's already initialized, you should be retrieving the genesis data using `state.LoadGenesisData`

cmd/gossamer/main.go
```diff
@@ -102,6 +102,9 @@ func main() {
 // configuration, loads the keystore, initializes the node if not initialized,
 // then creates and starts the node and node services
 func gossamerAction(ctx *cli.Context) error {
+	if ctx.IsSet(GenesisFlag.Name) {
+		return fmt.Errorf("genesis flag not allowed with start command")
+	}
```
ryanchristo, line 107: cc @noot I think this solves your concerns about the genesis flag being passed with the root action

ryanchristo: I could probably move this below the arguments check but otherwise I think this is all we need for now to prevent genesis from being used with the root action

noot: you should just remove it from the GlobalFlags in flags.go
create InitFlags, add GenesisFlag there, and then update initCommand Flags in main.go to initFlags

noot: I just realized export, init, and account are all set to `AllFlags`, this is wrong. the flags are separated in flags.go for a reason, specifically `AccountFlags` which are only relevant for `gossamer account`.

ryanchristo: separate issue please, this pull request addresses loading the properties from the genesis file and making sure they match and are being set in the configuration when initializing a node

ryanchristo: I'm also not sure it is "wrong" but maybe it is being approached from a different angle

noot: it's definitely a bug, you should not be allowed to use the genesis flag without `init`.

ryanchristo: you should be able to use with export as well and it should have no affect if you are using it with other commands so I am still not sure if it is wrong or a bug but we can take a look

noot: I also don't think this fixes the issue where you `init` with genesisA.json, then run `gossamer --config configB.json` where configB genesis set to genesisB.json.

cmd/gossamer/config.go
ryanchristo: This is just loading the genesis properties so that they can be added to the configuration. Let me know if this is still confusing or if it still needs to be fixed.

noot: if this is being called from `initAction` this is ok. if this is being called from `gossamerAction` it should be reading the genesis data from the database

ryanchristo: ok, with the added check if the genesis flag is being passed to `gossamerAction`, sounds like this is ok as is then. :+1:

cmd/gossamer/main.go
ryanchristo: I am not sure this is related to this issue, can you open a new issue?

noot: https://github.com/ChainSafe/gossamer/issues/736

cmd/gossamer/config.go
noot: not really, since this still allows incorrect genesis data to be read (ie. not what the node was initialized with)

except for the `init` case, the genesis data should *only* be read from the database

noot: the issue explicity says that the genesis data should be loaded with `LoadGenesisData`

ryanchristo:
cc @noot ready for another review, this pull request no longer closes #713

ryanchristo:
also, this pull request is now based of `noot/flags` and is blocked until #742 is merged

ryanchristo:
The way I see it, the command and setting the configuration values is separate from the initialization process. All we are doing in this pull request is checking the genesis file against the configuration and logging a warning so that the user knows the genesis file and config mismatch.

This pull request does not close #742 because it does not load genesis data from state, it only checks if the genesis file and configuration file match and warns the user if they do not. I would prefer to address loading genesis data in a separate pull request.

noot:
okay I'm not sure I see the point of this then, since we would only load genesis data from the file during init, in which case the config values being overwritten doesn't matter since genesis would be the source of truth. 

if you call the function only from `setDotInitConfig` then it's okay. otherwise I don't think it's necessary

ryanchristo:
The point of this is to check if the genesis configuration values do not match the config values (either the default values from the node defaults file or a passed in config file) and warns the user if they do not match. If you still do not see the point and it does no harm, can we merge and revisit later so that we can continue to move forward?

noot:
move the updateDotConfigFromGenesis call to setDotInitConfig only, then I think it's fine. we shouldnt be reading from the genesis file unless initializing

ryanchristo:
cc @noot I have removed it per request even though I believe the check was helpful to understand how configuration values are being read and set, not just when calling `init`, and it was doing no harm, but it would be easy enough to reinclude while debugging if necessary.

PR #734, opened 2020-04-01
docs: add docs configuration and structure

### LINKED ISSUE #528: Add configuration for static docs site
We should include documentation within the repository to support versioned documentation.

### LINKED ISSUE #529: Add wiki pages to static docs site
## Expected Behavior

Once we are ready to release an initial version of gossamer, all wiki pages need to be transferred to the wiki pages directory within the repository so that we can maintain versioned documentation.

## Checklist

- [ ] #867
- [ ] #868
- [ ] #869

ryanchristo:
## Changes

Once we release an official version of gossamer, we will need to version our documentation, and the easiest way to do this is to include it within the repository.

This pull request is experimenting with a static docs site using Jekyll, which can then be used with GitHub pages (GitHub will automatically deploy and host for free). 

## Tests:
```
cd wiki

bundle exec jekyll serve
```

### Issues:
closes #528


thomasmodeneis:
I dont like the idea of adding the wiki docs into the same repo. I think this should be in another repo, otherwise we will be having tons of .md that are not really a deliverable or part of the project itself.

README.md
```diff
@@ -25,86 +25,13 @@
...
-
-### Run Gossamer Node
-
-```
-./bin/gossamer init
-```
-
-```
-./bin/gossamer
-```
```
noot, line 107: I think we should leave quick instructions like this in the readme

ryanchristo: agreed, I think I meant to just snip out the architecture stuff

ryanchristo:
cc @thomasmodeneis @noot 

I think we will find there a lot more benefits to having the documentation within the repository than there would be issues and I would advocate including documentation within the repository until we feel like we need to create a separate repository. Let's not put the cart before the horse.

I have done this with several projects before and it has proven useful to have both documentation and code in the same repository and versioned with the same tags and version numbers so that users can load documentation locally for the version they are using within the repository.

The wiki will include links to godocs and godocs will be versioned based on the code; it would be nice to keep godoc versioning and wiki versioning the same rather than having the wiki separate using separate version numbers.

When we update code, documentation should be updated in the same pull request if the code being updated makes the documentation invalid. For example, we change a package name, we then need to change the package name in the documentation, we can update the documentation and the code in the same pull request rather than opening a pull request in another repository.

Also, this is a common practice and often implemented as a standard, which is why GitHub includes it as a feature within the repository settings.

Last but not least, documentation is a deliverable for the grant. We should include it in the repo.

README.md
ryanchristo: done :+1:

PR #737, opened 2020-04-01
dot/rpc: Introduce Gorilla RPC package

edwardmack:
## Changes
- Added use of github.com/gorilla/rpc package for handling RPC server calls
- Removed code that was replaced by using gorilla RPC
- 

## Tests:
Start gossamer with rpc flag
```
make gossamer
./bin/gossamer --key alice --rpc
```
Interact with node via RPC (in another terminal)
```
curl -H "Content-Type: application/json" -d '{"id":1, "jsonrpc":"2.0", "method": "system_name", "params": []}' http://localhost:8545/
```

### Issues:


dot/rpc/server_test.go
```diff
@@ -1,188 +0,0 @@
...
-		w.Status = 200
-	}
-	return len(p), nil
-}
-
-func (w *MockResponseWriter) WriteHeader(status int) {
-	w.Status = status
-}
-
-func TestServeHTTP(t *testing.T) {
```
noot, line 125: maybe add a test similar to this?

dot/rpc/dotUpCodec.go
```diff
@@ -0,0 +1,85 @@
+// Copyright 2019 ChainSafe Systems (ON) Corp.
```
noot, line 1: bit of a nitpick but maybe rename this file to snake case? or just codec.go?

noot:
I wrote some code to test the rpc server,
https://github.com/noot/gossamer-rpc-tests/blob/master/main.go

running `./bin/gossamer --rpc --roles 1`, then running main.go I get 
```
{"result":null,"error":"rpc: service/method request ill-formed: \"system_Health\"","id":1}
{"result":null,"error":"rpc: service/method request ill-formed: \"author_SubmitExtrinsic\"","id":1}
```

but on development I get:
```
{"jsonrpc":"2.0","result":{"health":{"Peers":0,"IsSyncing":false,"ShouldHavePeers":true}},"error":null,"id":1}

{"jsonrpc":"2.0","result":[107,59,94,91,205,240,250,186,15,117,72,111,101,217,192,75,24,110,228,255,81,100,181,185,104,108,125,155,238,167,206,82],"error":null,"id":1}
```

maybe check if the underscore encoding is working properly? otherwise, looks good!

dot/rpc/server_test.go
edwardmack: Good idea, I'll add this...

dot/rpc/dotUpCodec.go
edwardmack: Ok

edwardmack:
@noot 
I believe the tests are failing because the request is ill-formed, for system_health the request generated by test is: 
```
{"jsonrpc":"2.0","method":"system_Health","params":{},"id":1}
```
- The method should be `system_health` (lower case h)
- params for request should be an array not an object
Using this request works:
```
{"jsonrpc":"2.0","method":"system_health","params":[],"id":1}
```
results:
```
{"result":{"health":{"Peers":0,"IsSyncing":false,"ShouldHavePeers":true}},"error":null,"id":1}
```

for author_submitExtrinsic
test request:
```
{"jsonrpc":"2.0","method":"author_SubmitExtrinsic","params":"0x03106e6f6f740140676f7373616d65725f69735f636f6f6c","id":1}
```
- Method should be `author_submitExtrinsic`
- params should be in array
Using this request works:
```
{"jsonrpc":"2.0","method":"author_submitExtrinsic","params":["0x03106e6f6f740140676f7373616d65725f69735f636f6f6c"],"id":1}
```
results:
```
{"result":[107,59,94,91,205,240,250,186,15,117,72,111,101,217,192,75,24,110,228,255,81,100,181,185,104,108,125,155,238,167,206,82],"error":null,"id":1}
```
Note, I'm not sure result for is correct in that it should be a hash represnted as a string, not byte array, but server is returning response.
Also, if I submit this request a second time I get (re-booting node fixes, but can still only call once):
```
Post http://localhost:8545: EOF
```
so it seems there is an issue with author_submitExtrinsic, but the RPC in handling and dispatching the calls correctly (I think).

I'm basing how these requests should be formed based on this: https://github.com/w3f/PSPs/blob/psp-rpc-api/psp-002.md

Let me know your thoughts.

noot:
@edwardmack nice, yeah I updated the format and it works! nice work :)

dot/rpc/dotUpCodec.go
```diff
@@ -0,0 +1,85 @@
...
+// on to the calling rpc server.
+func (c *DotUpCodecRequest) Method() (string, error) {
+	m, err := c.CodecRequest.Method()
+	if len(m) > 1 && err == nil {
+		parts := strings.Split(m, "_")
+		service, method := parts[0], parts[1]
+		r, n := utf8.DecodeRuneInString(method) // get the first rune, and it's length
+		if unicode.IsLower(r) {
+			upMethod := service + "." + string(unicode.ToUpper(r)) + method[n:]
+			log.Printf("lowercase method %s requested: treated as %s\n", m, upMethod)
```
noot, line 80: remove or change to trace

edwardmack:
This is ready for another look.

PR #742, opened 2020-04-03
cmd: use subcommand specific flags

### LINKED ISSUE #735: cmd: separate flags for subcommands
right now, all subcommands use `AllFlags.` this is wrong, `gossamer init` and `gossamer account` and `gossamer` all have a different set of valid flags.

eg. --generate, --import are revelant to `account` only, --genesis is relevant to `init` only

### LINKED ISSUE #736: config file should not contain genesis, config
the config path shouldn't be in the config file, since the node already knows about where the config is if it's reading the file. in dot/config.go, `GlobalConfig.Config` should have toml tags such that it isn't valid to be included in the toml, *or* there should be a `Config` type that contains only fields that are valid to be included in the toml.

similarly for `GlobalConfig.Genesis`, this is *not* valid as a global config option. This is only valid when doing `gossamer init`, thus it should either:
1. be removed from the config entirely (set toml tags so that it isn't valid to be included in the toml file)
2. create a separate InitConfig type, where the contents of `InitConfig` are only used upon `gossamer init`.  eg. config.toml:

```
[init]
genesis = "./genesis.json"
```

noot:
## Changes
- separate flags so that they are subcommand specific
- fix broken account subcommand
- remove config field from config file
- move genesis field to init section, which is only used with init subcommand

## Tests:
```
go test ./... -short
```

please try with subcommands, usual node flags as well

### Issues:
closes #736 closes #735


noot:
@edwardmack should be fixed now, please try again!

PR #743, opened 2020-04-03
dot/rpc, tests: Add Integration Tests for RPC #741

### LINKED ISSUE #741: Automated Tested of Syncing/Integration with a script
Create a bash script to: 

- Spin up several (3) nodes 
- Connect them to one another
- Send some transactions via RPC queries
- Verify that blocks contain the transactions that we've sent via RPC queries
- Verify that all the nodes have the same blocks via RPC queries

thomasmodeneis:
## Changes
- add integration tests step to travis
- add initial rpc package for hosting the new integration tests for the RPC modules
- add initial script that starts gossamer, run tests, check if they worked and terminate instance
- add it-stable test step to Makefile

## Tests:
```
make it-stable
```

### Issues:
closes #741


scripts/integration-test-all.sh
```diff
@@ -0,0 +1,88 @@
...
+
+# Compile gossamer
+echo "compiling gossamer"
+make build
+
+
+# Run node with static blockchain database
+echo "starting gossamer node in background with http listener on $HOST_RPC"
+
+$PWD/bin/gossamer --port=$PORT --key=$KEY --datadir=$DATA_DIR \
```
noot, line 54: can you suppress the gossamer logs that get printed to stdout? would be nice to pipe them into another file, it's a bit confusing to run the tests and also see info from the gossamer node

tests/rpc/integration_test.go
```diff
@@ -0,0 +1,146 @@
...
+		"GOSSAMER_NODE_HOST", GOSSAMER_NODE_HOST)
+
+	testsCases := []struct {
+		description string
+		method      string
+		expected    interface{}
+	}{
+		{
+			description: "test system_Health",
+			method:      "system_Health",
```
noot, line 74: i think this gets updated to `system_health` with Ed's new PR

scripts/integration-test-all.sh
thomasmodeneis: hmmm, see the logs is what we want to have when tests break. Pipe it to another file would just make it quite complex to debug results on travis buids as well...

thomasmodeneis:
> overall looks good, just one small comment. also, are you planning to add more RPC calls as they are implemented, or should whoever implements them add them to here?

the next steps are to have the gossamer node to start with a fixed database file that we know the outcomes and then hit the endpoints asserting the results.

Issue #724, opened 2020-03-25
migrate test utils to respective packages


ryanchristo:
## Expected Behavior

The utilities within the root level `tests` directory live within their respective packages.

## Current Behavior

We have a `tests` directory with utilities that are specific to the `babe` and `runtime` packages.

PR #747, opened 2020-04-03
lib/runtime: isolate runtime updates

noot:
## Changes
- update NewRuntime to accept a `func() (*wasm.Imports, error)` that will provide the imports for that runtime
- move runtime test helpers in tests/ to runtime package
- update runtime.POLKADOT_RUNTIME constant to have commit info
- move allocator into runtime package

## Tests:
```
go test ./... -short
```

### Issues:
related to #724 and #563


lib/runtime/test_helpers.go
```diff
@@ -14,24 +14,54 @@
...
+}
+
 //nolint
 const (
-	POLKADOT_RUNTIME     = "polkadot_runtime"
-	POLKADOT_RUNTIME_FP  = "substrate_test_runtime.compact.wasm"
-	POLKADOT_RUNTIME_URL = "https://github.com/noot/substrate/blob/add-blob/core/test-runtime/wasm/wasm32-unknown-unknown/release/wbuild/substrate-test-runtime/substrate_test_runtime.compact.wasm?raw=true"
+	POLKADOT_RUNTIME_c768a7e4c70e     = "polkadot_runtime"
+	POLKADOT_RUNTIME_FP_c768a7e4c70e  = "substrate_test_runtime.compact.wasm"
+	POLKADOT_RUNTIME_URL_c768a7e4c70e = "https://github.com/noot/substrate/blob/add-blob/core/test-runtime/wasm/wasm32-unknown-unknown/release/wbuild/substrate-test-runtime/substrate_test_runtime.compact.wasm?raw=true"
```
ryanchristo, line 64: Is there a git tag for this commit that we can use rather than the commit hash (`c768a7e4c70e`) ?

lib/runtime/imports.go
```diff
@@ -749,6 +749,11 @@ func ext_local_storage_set(context unsafe.Pointer, kind, key, keyLen, value, val
 	log.Warn("[ext_local_storage_set] Not yet implemented.")
 }
 
+// RegisterImports registers the wasm imports for the old version of the substrate test runtime.
+func RegisterImports() (*wasm.Imports, error) {
```
ryanchristo, line 753: Do we need both `RegisterImports` and `registerImports` ? If so, maybe add comments to explain and a comment for `registerImports`, or maybe rename one of them?

lib/runtime/test_helpers.go
noot: hmm, not really. I could add a tag to my fork though, that might be a better idea

lib/runtime/imports.go
noot: yeah I'll just rename them

noot: hmm, actually I'm thinking of just exposing `NewTestRuntime` but changing it to `NewRuntimeVersion` or something like that, where you can specify the runtime you want as a string, and it will use the correct imports func for you. what do you think?

PR #752, opened 2020-04-06
dot/rpc; implement RPC chain_getBlockHash

edwardmack:
## Changes
- Implement RPC function chain_getBlockHash
- Create alias chain_getHead for chain_getBlockHash
- Created tests for these functions

## Tests:
```
go test ./dot/rpc/...
```

### Issues:


dot/rpc/modules/chain_test.go
```diff
@@ -23,17 +23,71 @@ func newChainService(t *testing.T) *state.Service {
...
 	}
+
+	err = loadTestBlocks(genesisHeader.Hash(), stateSrvc.Block)
+	if err != nil {
+		t.Fatal(err)
+	}
 	return stateSrvc
 }
 
+func loadTestBlocks(gh common.Hash, bs *state.BlockState) error {
```
noot, line 45: there's another test function in state called `addBlocksToState` where you can specify the depth (the branching part of that function probably isn't too relevant here) that might be useful in the future for adding more than 2 blocks

edwardmack: excellent, I thought there may be something, but I didn't look deeply enough.  Will come in handy.

dot/rpc/modules/chain_test.go
```diff
@@ -56,10 +110,10 @@ func TestChainGetHeader_Latest(t *testing.T) {
 	chain := newChainService(t)
 	svc := NewChainModule(chain.Block)
 	expected := &ChainBlockHeaderResponse{
-		ParentHash:     "0x0000000000000000000000000000000000000000000000000000000000000000",
-		Number:         big.NewInt(0),
-		StateRoot:      "0x03170a2e7597b7b7e3d84c05391d139a62b157e78786d8c082f29dcf4c111314",
-		ExtrinsicsRoot: "0x03170a2e7597b7b7e3d84c05391d139a62b157e78786d8c082f29dcf4c111314",
+		ParentHash:     "0xdbfdd87392d9ee52f499610582737daceecf83dc3ad7946fcadeb01c86e1ef75",
```
ryanchristo, line 113: maybe just a comment to explain where this and the other test hashes are coming from?

PR #755, opened 2020-04-07
dot/core: fix node restart

noot:
## Changes
- previously, if a node was stopped and restarted, it would always error with "first block not set for current epoch"
- also cleans up some goroutine loops so they don't stop if they error but continue to the next iteration

## Tests:
```
go test ./... -short
```


dot/network/status.go
```diff
@@ -195,6 +195,8 @@ func (status *status) sendNextMessage(ctx context.Context, peer peer.ID) {
 func (status *status) expireStatus(ctx context.Context, peer peer.ID) {
 
 	// wait to check for expired status message
+	// TODO: this wakes up every 6 minutes, while sendNextMessage wakes up every 5 minutes.
```
ryanchristo, line 198: issue number?

noot: #756

ryanchristo:
would also be nice to get the diff coverage above the threshold :eyes:

PR #761, opened 2020-04-08
cmd, dot: load genesis data for initialized node

### LINKED ISSUE #713: load name, id, protocol id, bootnodes from genesis upon startup

ryanchristo:
## Changes
- load genesis data from database if node already initialized

## Tests:

run tests...
```
go test ./cmd/...
```

run gossamer command...
```
make gossamer

./bin/gossamer --key alice --datadir tmp/alice --name TESTNODE --verbosity trace init
./bin/gossamer --key alice --datadir tmp/alice --name TESTNODE --verbosity trace
```

...then update name in `gssmr/genesis.config` to `ANOTHERNODE` and rerun commands.

(1) you should see a warning that the name within the genesis file and the config file do not match and the config name value is being overwritten with the genesis name value

(2) you should see that we load and update using the genesis data for an initialized node and the config is overwritten a second time using the genesis data before being passed to `NewNode`

### Issues:
closes #713

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,4 @@
 *.wasm
 
 test_data
+tmp
--- a/cmd/gossamer/config.go
+++ b/cmd/gossamer/config.go
@@ -22,6 +22,9 @@ import (
 	"strings"
 
 	"github.com/ChainSafe/gossamer/dot"
+	"github.com/ChainSafe/gossamer/dot/state"
+	"github.com/ChainSafe/gossamer/lib/common"
+	"github.com/ChainSafe/gossamer/lib/database"
 	"github.com/ChainSafe/gossamer/lib/genesis"
 
 	log "github.com/ChainSafe/log15"
@@ -84,7 +87,7 @@ func createDotConfig(ctx *cli.Context) (cfg *dot.Config, err error) {
 	setDotGlobalConfig(ctx, &cfg.Global)
 
 	// ensure configuration values match genesis and overwrite with genesis
-	updateDotConfigFromGenesis(ctx, cfg)
+	updateDotConfigFromGenesisJSON(ctx, cfg)
 
 	// set remaining cli configuration values
 	setDotAccountConfig(ctx, &cfg.Account)
@@ -110,7 +113,7 @@ func createInitConfig(ctx *cli.Context) (cfg *dot.Config, err error) {
 	setDotInitConfig(ctx, &cfg.Init)
 
 	// ensure configuration values match genesis and overwrite with genesis
-	updateDotConfigFromGenesis(ctx, cfg)
+	updateDotConfigFromGenesisJSON(ctx, cfg)
 
 	return cfg, nil
 }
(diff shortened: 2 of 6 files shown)


.gitignore
```diff
@@ -15,3 +15,4 @@
 *.wasm
 
 test_data
+tmp
```
noot, line 18: ?

cmd/gossamer/config.go
```diff
@@ -84,7 +87,7 @@ func createDotConfig(ctx *cli.Context) (cfg *dot.Config, err error) {
 	setDotGlobalConfig(ctx, &cfg.Global)
 
 	// ensure configuration values match genesis and overwrite with genesis
-	updateDotConfigFromGenesis(ctx, cfg)
+	updateDotConfigFromGenesisJSON(ctx, cfg)
```
noot, line 90: this can be removed since you're loading the genesis from the db now

.gitignore
ryanchristo: ignore all tmp directories... I prefer to use `--datadir tmp/alice` when testing, would it be ok if we added this to the gitignore?

cmd/gossamer/config.go
ryanchristo: this is providing a warning message to the user that I think would be valuable to keep and it will allow us to track the updates to the config, from the genesis file, and from the genesis data

we could remove it but I see more value in keeping it for now.... if it does no harm and I think it provides some value, would it be ok to keep it in for now?

ryanchristo: we will also need it when we implement export

thomasmodeneis: agree with @ryanchristo ; more log is better than less log...

.gitignore
thomasmodeneis: why not, if that helps you :)

cmd/gossamer/config.go
noot: okay, it's just that the genesis json in the config might not be the same as the genesis json that the node was loaded with

ryanchristo: correct, but it will be if the node was not initialized and is auto-initialized on start

in both cases it will provide a warning log that the genesis file does not match the config even though we load genesis data from state if the node has already been initialized, which is also shown but in a debug log

in both cases, the user is warned that the genesis file does not match the config so that they can update the config to match the genesis file or simply be aware of the fact

noot: okay, I just don't think there is a reason to load the genesis file unless you're initializing the node, otherwise, you are loading genesis information twice

the [init] section in the config shouldn't be accessed unless initialization is happening

ryanchristo: Hmm... this is where I am confused.

The init section in the config is not accessed unless we are initializing a node. Yes, we are loading the genesis file if the node has already been initialized, but we are loading to add an extra check which might not be necessary if the node is already initialized but it is necessary information if the node has not been initialized and I think it would still be helpful information for users to know if the genesis file and the config file values do not match even if the node has already been initialized and the genesis data will be different (and logged correctly)

noot: I thought when you loaded the genesis data you checked it there? can you move the comparison to there? 

also, the init field is being accessed outside of initialization, since this is called from NewNode outside the initialization statement

ryanchristo: there is most likely a better way to do this but I think this works well for now...?

ryanchristo: We set the genesis data from the config when we initialize the state service, the configuration are being updated in the command package before being passed to the dot node, which allows the dot node to accept a single parameter (a dot config).

Both the dot package and the command package utilize the genesis package independently as a modular package and setting the configuration from the flags and genesis file is the responsibility of the command package and using a configuration and accessing the state to initialize and start a node is the focus of the dot package.

I will take another look but I am still a bit confused about the requested change if everything works as expected and the information regarding how configuration values is being logged at each step with a warning for users if the config and genesis files do not match.

PR #765, opened 2020-04-09
dot/rpc: Implement RPC author_insertKey

edwardmack:
This PR introduces implementation of RPC call author_insertKey, which inserts the given key into the core keystore.  Defined in [https://github.com/w3f/PSPs/blob/psp-rpc-api/psp-002.md#author_insertKey](https://github.com/w3f/PSPs/blob/psp-rpc-api/psp-002.md#author_insertKey)

## Changes
- Implement RPC method author_insertKey
- Create tests for author_insertKey

## Tests:
```
go test ./dot/rpc/...
``` 
AND

```
curl -H "Content-Type: application/json" -X POST http://127.0.0.1:8545 -d '{"id":1, "jsonrpc":"2.0", "method": "author_insertKey", "params":["babe", "0xb7e9185065667390d2ad952a5324e8c365c9bf503dcf97c67a5ce861afe97309", "0x6246ddf254e0b4b4e7dffefc8adf69d212b98ac2b579c362b473fec8c40b4c0a"]}'
```

### Issues:


noot:
@edwardmack I think just putting them into the keystore as you're doing right now is okay, and yeah they should be sr25519 keys. in the future we should probably have separate keystores for babe, grandpa, etc. let's make an issue for that

dot/rpc/modules/author_test.go
```diff
@@ -2,16 +2,87 @@ package modules
...
 var testExt = []byte{3, 16, 110, 111, 111, 116, 1, 64, 103, 111, 115, 115, 97, 109, 101, 114, 95, 105, 115, 95, 99, 111, 111, 108}
 
+// testGenesisHeader is a test block header
+var testGenesisHeader = &types.Header{
+	Number:    big.NewInt(0),
+	StateRoot: trie.EmptyHash,
+}
+
+// newTestService creates a new test core service
+func newTestService(t *testing.T, cfg *core.Config) *core.Service {
```
noot, line 32: maybe we could move this to dot/core/test_helpers.go and export it? just so we don't need to duplicate code

edwardmack: Good idea!

PR #767, opened 2020-04-10
cmd/gossamer: confirm reinitializing a node

### LINKED ISSUE #760: cmd/gossamer: confirm reinitialize with prompt
## Expected Behavior

The user should receive a [Y/n] prompt to reinitialize a node if a node has already been initialized.

## Current Behavior

We are warning the user that the node is being reinitialized without a confirmation prompt.

ryanchristo:
## Changes
- prompt user to confirm with `Y` when attempting to initialize an already initialized node

## Tests:
```
make gossamer

./bin/gossamer --datadir tmp/alice --key alice init

./bin/gossamer --datadir tmp/alice --key alice init
```

### Issues:
closes #760

### DIFF
--- a/cmd/gossamer/main.go
+++ b/cmd/gossamer/main.go
@@ -135,7 +135,7 @@ func gossamerAction(ctx *cli.Context) error {
 	}
 
 	// ensure configuration matches genesis data stored during node initialization
-	// and overwrite with flag values if flag values are provided
+	// but do not overwrite configuration if the corresponding flag value is set
 	err = updateDotConfigFromGenesisData(ctx, cfg)
 	if err != nil {
 		log.Error("[cmd] Failed to update config from genesis data", "error", err)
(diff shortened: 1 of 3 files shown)


noot:
@ryanchristo looks good, but there should be a away to overwrite the prompt (eg. with --force, -f) for during testing when we know we want to reinitialize

ryanchristo:
@noot ok, can we add the force flag in a seoarate pull request> I can create an issue.

ryanchristo:
nvm, will add in this pull request

ryanchristo:
@noot in #770 should we use the confirm message method and the force flag to overwrite the config file if the user specifies a config filepath that already exists? In which case it should probably not be an init falg but rather its own independent flag that can be reused for multiple purposes? See #770.

noot:
@ryanchristo okay sure, you can use flags for multiple subcommands without making them global

ryanchristo:
correct, which is how the `GenesisFlag` is being used in #770

ryanchristo:
It would be much easier to add the force flag in that pull request since I update flags there. I will make sure to address the force flag today in #770 right after this pull request if you are willing to approve the diff-coverage in this pull request.

PR #769, opened 2020-04-10
scripts, tests: Automated Tested of Syncing/Integration with a script #741

thomasmodeneis:
## Changes
- update integration-test-all.sh script to be able to start/stop multiple nodes at once
- add configurable nodes and waiting cmd args
- make sure go tests always run by disabling cache
- add assertion for system_health response fields IsSyncing, ShouldHave Peers and Peers
- run integration tests on N nodes

- help cmd output:
```
./scripts/integration-test-all.sh -h       
Usage: integration-test-all.sh
Optional command line arguments
-t <string>  -- Test to run. eg: rpc
-q <number>  -- Quantity of nodes to run. eg: 3
-s <number>  -- Sleep between operations in secs. eg: 5
```

## Tests:
```
 ./scripts/integration-test-all.sh -q 3 -o 10
```

### Issues:


scripts/integration-test-all.sh
```diff
@@ -40,46 +44,66 @@ if [[ ! "$DATA_DIR" ]]; then
...
 set +e
 
+if [[ -z $TEST || $TEST == "rpc" ]]; then
 
-if [[ -z $TEST || $TEST = "rpc" ]]; then
-
-GOSSAMER_INTEGRATION_TEST_MODE=$MODE GOSSAMER_NODE_HOST=$HOST_RPC go test ./tests/rpc/... -timeout=60s -v
+  for i in $(seq 1 "$QTD"); do
+    HOST_RPC=http://$IP_ADDR:$RPC_PORT"$i"
+    echo "going to test gossamer node $HOST_RPC ..."
```
ryanchristo, line 84: `testing gossamer node $HOST_RPC...` ?

scripts/integration-test-all.sh
```diff
@@ -40,46 +44,66 @@ if [[ ! "$DATA_DIR" ]]; then
...
 
-RPC_FAIL=$?
+    RPC_FAIL=$?
+  done
 
 fi
 
+stop_func() {
+  GOSSAMER_PID=$i
+  echo "shutting down gossamer node, pid=$GOSSAMER_PID"
```
ryanchristo, line 94: `shutting down gossamer node, pid=$GOSSAMER_PID...` ?

scripts/integration-test-all.sh
```diff
@@ -3,31 +3,35 @@
 # "stable" mode tests assume data is static
 # "live" mode tests assume data dynamic
 
-
-SCRIPT=`basename ${BASH_SOURCE[0]}`
+SCRIPT=$(basename ${BASH_SOURCE[0]})
 TEST=""
+QTD=1
+SLEEP_TIMEOUT=5
```
edwardmack, line 9: Perhaps we should increase this, I still see some of these failing with Travis when this is 5.

thomasmodeneis: this is now configurable, check the help output with -h and makefile task.

scripts/integration-test-all.sh
thomasmodeneis: whats wrong ?

scripts/integration-test-all.sh
thomasmodeneis: whats wrong ?

scripts/integration-test-all.sh
```diff
@@ -3,31 +3,35 @@
...
-SCRIPT=`basename ${BASH_SOURCE[0]}`
+SCRIPT=$(basename ${BASH_SOURCE[0]})
 TEST=""
+QTD=1
+TIMEOUT=5
 
-PORT="7001"
-RPC_PORT="8545"
+PORT="700"
+RPC_PORT="854"
```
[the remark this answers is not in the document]
thomasmodeneis: because its for testing and you dont want to conflict with actual running instances

scripts/integration-test-all.sh
ryanchristo: nothing is wrong, I was just recommending consistent formatting for echo statements

scripts/integration-test-all.sh
ryanchristo: same as above, you use `...` when something is about to start but you do not here

scripts/integration-test-all.sh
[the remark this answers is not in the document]
ryanchristo: I guess by standard I meant four digit port numbers. I am well aware that you are avoiding conflicting ports when running instances.

Certain operating systems may restrict opening lower port numbers, so it is usually a standard to use four digit port numbers. The numbers you are using also make it look like someone accidentally deleted the last digit off our default ports. It might be nice to make this more distinct and to use a more common port number or just add 1 to the last digit.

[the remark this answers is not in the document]
noot: yeah i think port numbers < 1024 are reserved for the kernel (at least depending on OS) change to 8546 or sometihng?

scripts/integration-test-all.sh
```diff
@@ -40,46 +44,66 @@ if [[ ! "$DATA_DIR" ]]; then
...
 
 fi
 
+stop_func() {
+  GOSSAMER_PID=$i
+  echo "shutting down gossamer node, pid=$GOSSAMER_PID"
 
-echo "shutting down gossamer node"
+  # Shutdown gossamer node
+  kill -9 "$GOSSAMER_PID"
```
noot, line 97: why `-9`? seems very forceful lol

scripts/integration-test-all.sh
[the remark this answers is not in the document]
thomasmodeneis: I see you guys are not familiar with how the shell scripts work, but that is the initial prefix for the ports. You can see later on that the $PORT and $RPC_PORT is being used with adding a suffix for the port.

scripts/integration-test-all.sh
thomasmodeneis: hmmm, going to test is not the same as testing...its the log before running the test...
Testing is for when the test is already running.

scripts/integration-test-all.sh
thomasmodeneis: that is how it should be, kill now... we dont want to be waiting for the process to decide if or not it wants to stop...

scripts/integration-test-all.sh
thomasmodeneis: I see

scripts/integration-test-all.sh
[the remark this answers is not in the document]
ryanchristo: Ok, I see. Please add comments in the code here to better document this. Seems like it might be an easy mistake for people to make.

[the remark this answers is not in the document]
thomasmodeneis: oks

Issue #669, opened 2020-03-04
Use consistent log format in all packages


ryanchrypto:
## Current Behavior

We are not using a consistent logging format in all of the packages within this repository.

## Expected Behavior

We use a consistent logging format in all of the packages within this repository.

## Checklist

- [ ] update contributing documentation or create a separate document that defines logging, printing, and error format to be used for all packages within this repository
- [ ] update packages to follow logging, printing, and error formatting (as much as possible)

dutterbutter:
Closing due to staleness.

PR #770, opened 2020-04-10
cmd/gossamer: export toml congfiguration file

### LINKED ISSUE #686: Export config to toml with gossamer export
## Expected Behavior

The `gossamer export` command should not only dump the configuration values but also export them to a toml configuration file at the provided `--config` as the destination:

## Current Behavior

The `gossamer export` command does not create a toml configuration file.

ryanchristo:
## Changes
- export toml congfiguration file using the `export` subcommand
- add `--force` flag option to skip confirmation prompts
- use lowercase unless proper name for cmd and dot logs
- fix `newTextContext` to include setting flag values

## Tests:

```
make gossamer

./bin/gossamer --config test.toml export
./bin/gossamer --config test.toml --name TESTNODE export

./bin/gossamer --config test.toml export --force

./bin/gossamer init
./bin/gossamer init --force
```

### Issues:
closes #686, related to #669


cmd/gossamer/export.go
```diff
@@ -33,43 +34,31 @@ func exportAction(ctx *cli.Context) error {
...
-			return err
-		}
-		defer func() {
-			err = export.Close()
-			if err != nil {
-				log.Error("[cmd] Failed to close connection", "error", err)
-			}
-		}()
-	}
+		// TODO: confirm once #767 is merged
```
ryanchristo, line 48: waiting for #767 to be merged, then will include the following code:

```
log.Warn(
	"[cmd] toml configuration file already exists",
	"config", config,
)

if confirmMessage("Would you like to overwrite the existing file? [Y/n]") {

	cfg := createExportConfig(ctx)
	
	file := dot.ExportConfig(cfg, config)
	// export config will exit and log error on error

	log.Info("[cmd] Exported toml configuration file", "path", file.Name())
}

return nil

**Comment by ryanchristo** [comment]
cc @noot this one needs a review as well :pray:

**Comment by ryanchristo** [comment]
but it can also wait for #767

File: cmd/gossamer/config.go [hunk 2]
```diff
@@ -31,6 +31,9 @@ import (
 	"github.com/urfave/cli"
 )
 
+// DefaultCfg is the default configuration
+var DefaultCfg = dot.GssmrConfig() // TODO: investigate default node other than gssmr #776
```
ryanchristo, line 35: #776

dot/utils.go
```diff
@@ -108,7 +109,7 @@ func NewTestGenesisFile(t *testing.T, cfg *Config) *os.File {
 	file, err := ioutil.TempFile(dir, "genesis-")
 	require.Nil(t, err)
 
-	fp := getGssmrGenesisPath(t)
+	fp := getGssmrGenesisPath(t) // TODO: investigate default node other than gssmr #776
```
ryanchristo, line 112: #776

dot/utils.go
```diff
@@ -163,6 +164,8 @@ func NewTestGenesisAndRuntime(t *testing.T) string {
 	return genFile.Name()
 }
 
+// TODO: investigate default node other than gssmr #776
```
ryanchristo, line 167: #776

ryanchristo:
cc @noot @edwardmack @thomasmodeneis ready for review when you get a chance

noot:
@ryanchristo  looks good, is there a way to get `./bin/gossamer init --force` to work along with `./bin/gossamer --force init`? maybe with the FixFlagOrder func?

cmd/gossamer/flags.go
```diff
@@ -171,32 +176,21 @@ var (
 	}
 )
 
+// flag sets that are shared by multiple commands
 var (
 	// GlobalFlags are flags that are valid for use with all commands
 	GlobalFlags = []cli.Flag{
 		VerbosityFlag,
+		ForceFlag,
```
noot, line 184: hmm, not sure if this is global, since it's not valid with just `./bin/gossamer`

ryanchristo: Not at the moment but I figured it was a global flag in the sense that we use it whenever there is a confirm prompt even though we are not currently using it in the root command.

ryanchristo: I can move it but it seems global to me. Where would you recommend I move it to?

ryanchristo:
cc @noot yea I was noticing the `FixFlagOrder` could be optimized, would be nice to open a separate issue but if you have a quick fix in mind, I would be happy to include here.

cmd/gossamer/flags.go
noot: I think we should only make it global when necessary, if you could move it to InitFlags and ExportFlags that'd be good

noot:
does the FixFlagOrder not work for this PR on --force? if not that's ok, a separate issue is fine

ryanchristo:
ok @noot, `--force` is no longer global therefore `init --force` is now working but since it is no longer global `--force init` will not work. Let me know if this addresses your comments.

PR #788, opened 2020-04-16
lib/babe: fix createTestBlock err

noot:
## Changes

- retry if unsuccessful in createTestBlock, should fix test errs

## Tests:

```
go test ./lib/babe
```

### DIFF
--- a/lib/babe/build_test.go
+++ b/lib/babe/build_test.go
@@ -114,7 +114,7 @@ func createTestBlock(t *testing.T, babesession *Session, exts [][]byte) (*types.
 	}
 
 	// build block
-	for i := 0; i < 3; i++ { // retry if error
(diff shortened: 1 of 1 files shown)


lib/babe/build_test.go
```diff
@@ -114,12 +114,12 @@ func createTestBlock(t *testing.T, babesession *Session, exts [][]byte) (*types.
 	}
 
 	// build block
-	block, err := babesession.buildBlock(parentHeader, slot)
-	if err != nil {
-		t.Fatal(err)
+	for {
```
thomasmodeneis, line 117: perhaps we could do a loop 5x with a delay of 100ms ? ... a infinite loop seems like a potential deadlock....

noot: okay I'll put a limit

Issue #792, opened 2020-04-18
create test framework to easily test all rpc endpoints as client (follow JSON-RPC Interfaces for Polkadot Nodes)


thomasmodeneis:
## Expected Behavior

-

## Current Behavior

-

## Possible Solution

-

## Steps to Reproduce

1.
2.
3.
4.

## Specification

- go version:
- gossamer version:
- gossamer commit tag:
- gossamer commit hash:
- operating system:
- additional links:

## Checklist

- [ ] I have read [CONTRIBUTING](CONTRIBUTING.md) and [CODE_OF_CONDUCT](CODE_OF_CONDUCT.md)
- [ ] I have provided as much information as possible and necessary
- [ ] I am planning to submit a pull request to fix this issue myself

thomasmodeneis:
should have all endpoints defined on this list:
https://github.com/w3f/PSPs/blob/psp-rpc-api/psp-002.md

and test cases against all them (skip ones not yet implemented)

PR #793, opened 2020-04-18
tests: create test framework to test rpc endpoints #792

thomasmodeneis:
## Changes

- create test framework to test rpc endpoints as client (follow JSON-RPC Interfaces for Polkadot Nodes)

## Tests:

```

```

## Checklist:

## Issues:

- related to #792


Makefile
```diff
@@ -46,6 +46,10 @@ it-stress: build
 	@echo "  >  \033[32mRunning Integration Tests stress mode...\033[0m "
 	GOSSAMER_NODE_HOST=0.0.0.0 GOSSAMER_INTEGRATION_TEST_MODE=stress go test ./tests/stress/... -timeout=5m -p 1 -short -v
 
+it-rpc-spec: build
```
noot, line 49: why rpc-spec? why not just rpc

tests/spec/empty.go
```diff
@@ -0,0 +1,17 @@
...
+//
+// The gossamer library is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with the gossamer library. If not, see <http://www.gnu.org/licenses/>.
+
+package spec
```
noot, line 17: same with this, why `spec`?

tests/spec/rpc_01-system_test.go
```diff
@@ -0,0 +1,140 @@
...
+			skip:        true,
+		},
+		{ //TODO
+			description: "test system_accountNextIndex",
+			method:      "system_accountNextIndex",
+			skip:        true,
+		},
+	}
+
+	t.Log("going to Bootstrap Gossamer node")
```
noot, line 110: `going to start gossamer`?

tests/spec/rpc_01-system_test.go
```diff
@@ -0,0 +1,140 @@
...
+				t.Skip("RPC endpoint not yet implemented")
+				return
+			}
+
+			respBody, err := rpc.PostRPC(t, test.method, "http://"+rpc.GOSSAMER_NODE_HOST+":"+currentPort, "{}")
+			require.Nil(t, err)
+
+			target := rpc.DecodeRPC(t, respBody, test.method)
+
+			require.NotNil(t, target)
```
noot, line 131: where is the check for if result == expected?

thomasmodeneis: Yes, I added the logic to it now... its for now duplicating a bit with `system_integration_test.go`. This is ok since system_integration_test will be updated soon and follow other directions for testing with stable pre-saved gold files

tests/spec/rpc_01-system_test.go
thomasmodeneis: updated

tests/spec/empty.go
thomasmodeneis: updated

Makefile
thomasmodeneis: updated

Makefile
```diff
@@ -46,6 +46,10 @@ it-stress: build
 	@echo "  >  \033[32mRunning Integration Tests stress mode...\033[0m "
 	GOSSAMER_NODE_HOST=0.0.0.0 GOSSAMER_INTEGRATION_TEST_MODE=stress go test ./tests/stress/... -timeout=5m -p 1 -short -v
 
+it-rpc: build
+	@echo "  >  \033[32mRunning Integration Tests RPC Specs mode...\033[0m "
+	GOSSAMER_NODE_HOST=0.0.0.0 GOSSAMER_INTEGRATION_TEST_MODE=rpc_spec go test ./tests/rpc/... -timeout=5m -p 1 -short -v
```
noot, line 51: `GOSSAMER_INTEGRATION_TEST_MODE=rpc_spec`

should be 

`GOSSAMER_INTEGRATION_TEST_MODE=rpc`

tests/rpc/rpc_00_test.go
```diff
@@ -0,0 +1,36 @@
...
+	"testing"
+)
+
+const (
+	currentPort = "8540"
+	rpcSpec     = "rpc_spec"
+)
+
+func TestMain(m *testing.M) {
+	_, _ = fmt.Fprintln(os.Stdout, "Going to start RPC spec test")
```
noot, line 31: remove spec

tests/rpc/rpc_01-system_test.go
```diff
@@ -0,0 +1,181 @@
...
+	"testing"
+	"time"
+
+	"github.com/ChainSafe/gossamer/dot/rpc/modules"
+	"github.com/ChainSafe/gossamer/lib/common"
+	"github.com/stretchr/testify/require"
+)
+
+func TestSystemRPC(t *testing.T) {
+	if GOSSAMER_INTEGRATION_TEST_MODE != rpcSpec {
```
noot, line 33: change rpcSpec to rpc

tests/rpc/rpc_01-system_test.go
```diff
@@ -0,0 +1,181 @@
...
+			description: "test system_properties",
+			method:      "system_properties",
+			skip:        true,
+		},
+		{
+			description: "test system_health",
+			method:      "system_health",
+			expected: modules.SystemHealthResponse{
+				Health: common.Health{
+					Peers:           2,
```
noot, line 69: how are there two peers if you are only starting one node? maybe I am missing something

tests/rpc/rpc_01-system_test.go
```diff
@@ -0,0 +1,181 @@
...
+			respBody, err := PostRPC(t, test.method, "http://"+GOSSAMER_NODE_HOST+":"+currentPort, "{}")
+			require.Nil(t, err)
+
+			target := DecodeRPC(t, respBody, test.method)
+
+			require.NotNil(t, target)
+
+			t.Log("Will start assertion for ", "target", target, "type", reflect.TypeOf(target))
+
+			switch v := target.(type) {
```
noot, line 141: shouldn't we switch on the test description? what if an rpc method returns the wrong type

tests/rpc/rpc_01-system_test.go
```diff
@@ -0,0 +1,181 @@
...
+				require.GreaterOrEqual(t, test.expected.(modules.SystemHealthResponse).Health.Peers, v.Health.Peers)
+
+			case *modules.SystemNetworkStateResponse:
+				t.Log("Will assert SystemNetworkStateResponse", "target", target)
+
+				require.NotNil(t, v.NetworkState)
+				require.NotNil(t, v.NetworkState.PeerID)
+
+			case *modules.SystemPeersResponse:
+				t.Log("Will assert SystemPeersResponse", "target", target)
```
noot, line 157: do we need logs here for everything? seems a bit excessive

thomasmodeneis: Well, this logs help me to make sure that we are actually getting into the case statement. I am going to remove this as part of the next iteration when the tests will be more robust and I am more confident on its behavior. Right now the logs are important since we are building up.

tests/rpc/rpc_01-system_test.go
thomasmodeneis: The RPC method should never return the wrong type, if that is happening we are in deep trouble.

tests/rpc/rpc_01-system_test.go
thomasmodeneis: That is not yet used, you can see that the assertion is currentlty not performed. I have a objective to run more peers and make this test pass as you can observe on: https://github.com/ChainSafe/gossamer/pull/793/files/f485eea41765dfb14c881ed0b610e2e6fc41fb3e#diff-e0ce41a76aeea782dcef103c35cec4bfR161

tests/rpc/rpc_01-system_test.go
thomasmodeneis: that is not a blocker for the merge, but it can be done, sure.

tests/rpc/rpc_00_test.go
thomasmodeneis: that is not a blocker for the merge, but it can be done, sure.

tests/rpc/rpc_01-system_test.go
noot: okay well this will panic if it returns the wrong type, is that the desired behaviour?

tests/rpc/rpc_01-system_test.go
thomasmodeneis: raised  #807 to cover it

Makefile
thomasmodeneis: So since the objective of naming it spec was to make a test suite, and you are not ok with calling it spec, but I really nice need to be something more than just rpc (since its supposed to be a test suite for the RPC endpoints that match against a test specification) I renamed the logs to RPC suite. Let me know if you have any objections.

tests/rpc/rpc_01-system_test.go
thomasmodeneis: Why would it panic ? it just not going to match any type. If something else is going on with the endpoint it will be either captured at unit tests or on DecodeRPC method

thomasmodeneis: There is also the require nil err just before the switch...if you happen to be able to prove the edge case you are mentioning I will be more than happy to enforce more rules. Right now it seems like a overkill...

tests/rpc/rpc_01-system_test.go
```diff
@@ -0,0 +1,182 @@
...
+			require.NotNil(t, target)
+
+			t.Log("Will start assertion for ", "target", target, "type", reflect.TypeOf(target))
+
+			switch v := target.(type) {
+
+			case *modules.SystemHealthResponse:
+				t.Log("Will assert SystemHealthResponse", "target", target)
+
+				require.Equal(t, test.expected.(modules.SystemHealthResponse).Health.IsSyncing, v.Health.IsSyncing)
```
noot, line 146: this line could potentially panic - for example, say system_networkState returns a `SystemHealthResponse` for some reason (there are probably better examples in the other modules but this is these are the only ones implemented so far) 
then, test.expected is `SystemNetworkStateResponse`, not `SystemHealthResponse`. then, the type conversion `test.expected.(modules.SystemHealthResponse)` will panic.

thomasmodeneis: But it should not return that, I mean, why would the system_networkState return a type that belongs to the other endpoint ? that should be captured at the unit test level. This are integration tests.

thomasmodeneis: We are assuming here, that the minimum tests have been already checked by proper unit testing. Again, if this is happening at the integration tests and we have a panic, that is a good thing to happen. At least we will know right away something is really broken.

tests/rpc/rpc_01-system_test.go
noot: looking at the DecodeRPC method, the switch statement seems very brittle and will need to be updated in case the RPC api changes. it would be better to pass in a `target interface{}` directly instead of `targetType string` and switching on it. maybe we could open an issue for that?

tests/rpc/rpc_01-system_test.go
noot: okay I just think it's good practice to always do an `a, ok := b.(mytype)` to avoid potential panics, it also seems to make more sense to switch on the test description or method name instead of the expected type, but if you prefer it this way

thomasmodeneis: sure, raised #815 to tackle this!

tests/rpc/rpc_01-system_test.go
thomasmodeneis: Well, that one should be always updated. That is why this is going to be part of the CI. 
The trick is that the `DisallowUnknownFields` will make it fail nicely if the endpoint return changes, so we will see the test failures beforehand.

Issue #776, opened 2020-04-13
Investigate default node other than gssmr


ryanchrypto:
##  Expected Behavior

If it is reasonable upon further investigation to have a default node that is separate from the gssmr node (currently being used as the default node), then the default node should be a separate node implementation with its own `node/default` configuration directory and package.

## Current Behavior

We are using the gssmr node as the default node and the foundation for creating new node toml configuration files using the export subcommand.

## Possible Solution

Create a new `node/default` directory with its own configuration files and defaults. Update the var `DefaultCfg` to use the default node rather than the gssmr node.

Maybe update the tests throughout to use the default node as the base but still test the gssmr node alongside the ksmcc node where appropriate.

noot:
closing since this is no longer relevant/needed

Issue #687, opened 2020-03-10
Improve dot package tests


ryanchrypto:
## Expected Behavior

The dot package has better test coverage.

## Current Behavior

```
// TODO: improve dot config tests
```

```
// TODO: improve dot node tests
```

```
// TODO: improve dot service tests
```

```
// TODO: improve dot utils tests
```

noot:
closing due to vagueness

PR #795, opened 2020-04-20
cmd/gossamer: load default genesis if genesis empty

### LINKED ISSUE #789: Failed to load genesis from file if genesis path empty
Resolve the following error:

```
failed to load genesis from file: open /home/ninja/node/gssmr/genesis.json: no such file or directory
```

**Commands to reproduce:**

```
./bin/gossamer init --datadir /tmp/gossamer-stress55448602000 --genesis ./node/gssmr/genesis.json --force
```

```
./bin/gossamer --port 7000 --key alice --datadir /tmp/gossamer-stress55448602000 --rpchost 0.0.0.0 --rpcport 8540 --rpcmods system,author,chain --key alice --roles 4 --rpc
```

ryanchristo:
## Changes

- use default genesis path if toml config or flag for genesis path is empty
- fix local flag set only if global flag set failed (fix for small issue introduced in #784)
- minor fixes to dot tests, additional cleanup required (see #687 and #776)

## Tests

```
go test ./cmd/gossamer/...
```

should work as expected:

```
make gossamer

./bin/gossamer init --force --datadir tmp/alice

./bin/gossamer --datadir tmp/alice init --force

./bin/gossamer init --force --datadir tmp/alice --verbosity trace

./bin/gossamer init --datadir tmp/alice --force --verbosity trace

./bin/gossamer init --datadir tmp/alice --verbosity trace --force

./bin/gossamer --datadir tmp/alice --verbosity trace init --force
```

should not work as expected (note location of `--force` before `init`):

```
./bin/gossamer --force init --datadir tmp/alice --verbosity trace

./bin/gossamer --datadir tmp/alice --verbosity trace --force init 
```

see https://github.com/ChainSafe/gossamer/wiki/commands

## Checklist

- [X] I have provided as much information as possible and necessary
- [ ] All integration tests and required coverage checks are passing ***- needs approval***

## Issues

- closes #789
- related to #784


cmd/gossamer/config.go
```diff
@@ -326,6 +326,13 @@ func setDotRPCConfig(ctx *cli.Context, cfg *dot.RPCConfig) {
 // updateDotConfigFromGenesisJSON updates the configuration based on the genesis file values
 func updateDotConfigFromGenesisJSON(ctx *cli.Context, cfg *dot.Config) {
 
+	// use default genesis file if genesis configuration not provided, for example,
+	// if we load a toml configuration file without a defined genesis init value or
+	// if we pass an empty string as the genesis init value using the --geneis flag
+	if cfg.Init.Genesis == "" {
+		cfg.Init.Genesis = DefaultCfg.Init.Genesis
```
noot, line 333: I feel like this could cause bugs, is it possible to completely get rid of the need for a genesis file with the root command? that seems like it should be the expected behaviour

cmd/gossamer/config_test.go
```diff
@@ -513,6 +513,44 @@ func TestUpdateConfigFromGenesisJSON(t *testing.T) {
 	require.Equal(t, expected, cfg)
 }
 
+// TestUpdateConfigFromGenesisJSON_Default tests updateDotConfigFromGenesisJSON
+// using the default genesis path if no genesis path is provided (ie, an empty
+// genesis value provided in the toml configuration file or with --genesis "")
+func TestUpdateConfigFromGenesisJSON_Default(t *testing.T) {
```
noot, line 519: what happens if the node is loaded with a non default genesis then ran with a config without genesis?

node/ksmcc/genesis.json
```diff
@@ -1,17 +1,7 @@
...
-    "/dns4/p2p.cc3-1.kusama.network/tcp/30100/p2p/12D3KooWNpGriWPmf621Lza9UWU9eLLBdCFaErf6d4HSK7Bcqnv4",
-    "/dns4/p2p.cc3-2.kusama.network/tcp/30100/p2p/12D3KooWLmLiB4AenmN2g2mHbhNXbUcNiGi99sAkSk1kAQedp8uE",
-    "/dns4/p2p.cc3-3.kusama.network/tcp/30100/p2p/12D3KooWEGHw84b4hfvXEfyq4XWEmWCbRGuHMHQMpby4BAtZ4xJf",
-    "/dns4/p2p.cc3-4.kusama.network/tcp/30100/p2p/12D3KooWF9KDPRMN8WpeyXhEeURZGP8Dmo7go1tDqi7hTYpxV9uW",
-    "/dns4/p2p.cc3-5.kusama.network/tcp/30100/p2p/12D3KooWDiwMeqzvgWNreS9sV1HW3pZv1PA7QGA7HUCo7FzN5gcA",
-    "/dns4/kusama-bootnode-0.paritytech.net/tcp/30333/p2p/12D3KooWSueCPH3puP2PcvqPJdNaDNF3jMZjtJtDiSy35pWrbt5h",
-    "/dns4/kusama-bootnode-0.paritytech.net/tcp/30334/ws/p2p/12D3KooWSueCPH3puP2PcvqPJdNaDNF3jMZjtJtDiSy35pWrbt5h",
-    "/dns4/kusama-bootnode-1.paritytech.net/tcp/30333/p2p/12D3KooWQKqane1SqWJNWMQkbia9qiMWXkcHtAdfW5eVF8hbwEDw"
-  ],
+  "bootNodes": [],
```
noot, line 4: no more bootnodes?

ryanchristo: They need to match the bootnodes in the config.toml file. I can add them there or delete them here but since we are not using this at the moment and it will likely need to be updated, I just deleted them from here. I can do add them to the config.toml file if preferred.

cmd/gossamer/config_test.go
ryanchristo: This just tests the added check for the `updateConfigFromGenesisJSON` method. When the method is used with a toml configuration file that has an empty genesis init value or an empty genesis value is passed using the --genesis, it should use the default genesis method.

If the configuration at the time `updateConfigFromGenesisJSON` is called includes an empty value for the genesis init value, it will use the default value, which is the same value being used if we run a node with no flags (ie, `./bin/gossamer init`).

I am not sure I completely understand your question but if you run `gossamer init --config test.toml --genesis test.json` and the config file has an empty value for the genesis init value, it will use the genesis init value provided within the flag option.

node/ksmcc/genesis.json
noot: why do they need to match? that seems off

cmd/gossamer/config.go
ryanchristo: expected behavior to have an empty genesis path in the toml configuration file?

ryanchristo: I can remove again but this has come up before. The only issue would be if someone uses a genesis file that has values they would prefer to the toml configuration values (the values that will be overwritten by the genesis values if a genesis path is provided) or if their config file has an empty genesis path value (which I think should be required and checked).

noot: I meant it's expected that you wouldn't need to provide a genesis file for the root command, when the node is already initialized. it would make more sense imo to not have it be required and checked if you're not initializing

node/ksmcc/genesis.json
ryanchristo: This is a feature not a bug. : )

We are testing the export and load config methods alongside the node implementations by actually exporting and loading the gssmr and ksmcc configuration files, so them needing to be the same means the tests are working.

cmd/gossamer/config.go
ryanchristo: removed https://github.com/ChainSafe/gossamer/pull/795/files#diff-6e4a8b33cf573effe37b34dc984c597eL93

node/ksmcc/genesis.json
noot: which tests do you mean? in the future there will need to be bootnodes here and we shouldn't require that users copy and paste bootnodes from the genesis into the config, it should be fine to have them only in genesis

ryanchristo: It is fine to only have them in genesis, it just warns them that the values do not match with the values in the toml config.

Testing the export command and the export method in the dot package both export the values, so if we overwrite the values from the genesis file when exporting, we will have mismatching config files and the continuous integration tests will fail because files have been updated.

noot: okay, could you update the tests to get around that? or use a temp genesis file in the tests? we shouldn't need to change the actual genesis files to get tests to work

ryanchristo: Ok, I reverted the changes to the bootnodes in the ksmcc file. The warning logs will now be displayed again when running `gossamer --node ksmcc` and when running tests because the genesis file and configuration file do not match. I think this is your request? As mentioned in the first comment of this thread, I can also clean up the warning logs by adding the bootnodes to the configuration file but I have reverted the clean up as requested.

noot: that seems reasonable, I think you can leave it for now, we don't need to duplicate bootnodes in genesis and config

Issue #310, opened 2019-09-25
Implement state RPC methods & error codes


ansermino:
_If this needs to be split into smaller PRs or some of the RPC calls can't be implemented now, please create another task with a list of whatever remains_

**References:** 
- https://github.com/w3f/PSPs/blob/master/PSPs/drafts/psp-6.md#111-state
- https://polkadot.js.org/api/substrate/rpc.html 

### State RPC calls
- [ ] [State RPC Error Codes](https://github.com/w3f/PSPs/blob/master/PSPs/drafts/psp-6.md#11121-state-errors)
- [ ] [`state_call`](https://github.com/w3f/PSPs/blob/master/PSPs/drafts/psp-6.md#1113-state_call)
- [x] [`getMetadata`](https://github.com/w3f/PSPs/blob/master/PSPs/drafts/psp-6.md#1119-state_getmetadata) 
- [x] [`queryStorage`](https://github.com/w3f/PSPs/blob/master/PSPs/drafts/psp-6.md#11111-state_querystorage) 
- [x] [`subscribeRuntimeVersion`](https://github.com/w3f/PSPs/blob/master/PSPs/drafts/psp-6.md#11113-state_subscriberuntimeversion-pubsub)
- [x] [`subscribeStorage`](https://github.com/w3f/PSPs/blob/master/PSPs/drafts/psp-6.md#11115-state_subscribestorage-pubsub)
- [ ] [`getReadProof`](https://github.com/w3f/PSPs/blob/master/PSPs/drafts/psp-6.md#11112-state_getreadproof)
- [x] [`getRuntimeVersion`](https://github.com/w3f/PSPs/blob/master/PSPs/drafts/psp-6.md#11110-state_getruntimeversion)

edwardmack:
- All methods need to implement handling block hash parameter (See issue #834)
- All Child calls, need understanding of Child Info and Child type, currently labeled "future-reserved", how should we deal with that?

danforbes:
@EclesioMeloJunior - can we close this since we have individual issues now?

EclesioMeloJunior:
@danforbes yes!

PR #806, opened 2020-04-22
dot/rpc: implement RPC state_getRuntimeVersion

edwardmack:
## Changes

- Implement RPC call state_getRuntimeVersion
- Created VersionAPI object to hold RuntimeVersion information and runtime API array
-

## Tests:

```
go test ./dot/rpc/... 
```

## Checklist:

## Issues:

- Related to #310


edwardmack:
This deals with state_getRuntimeVersion, but it was based on the ed/569-add-WS-support branch, so it contains code from that as well.

dot/rpc/modules/state_test.go
```diff
@@ -0,0 +1,55 @@
...
+		Apis: []interface{}{[]interface{}{"0xdf6acb689907609b", int32(2)},
+			[]interface{}{"0x37e397fc7c91f5e4", int32(1)},
+			[]interface{}{"0xd2bc9897eed08f15", int32(1)},
+			[]interface{}{"0x40fe3ad401f8959a", int32(3)},
+			[]interface{}{"0xc6e9a76309f39b09", int32(1)},
+			[]interface{}{"0xdd718d5cc53262d4", int32(1)},
+			[]interface{}{"0xcbca25e39f142387", int32(1)},
+			[]interface{}{"0xf78b278be53f454c", int32(1)},
+			[]interface{}{"0xab3c0572291feb8b", int32(1)}},
+	}
```
noot, line 40: a comment explaining where this is from would be good

lib/runtime/runtime.go
```diff
@@ -168,15 +165,3 @@ func (r *Runtime) malloc(size uint32) (uint32, error) {
 func (r *Runtime) free(ptr uint32) error {
 	return r.allocator.Deallocate(ptr)
 }
-
-func decodeToInterface(in []byte, t interface{}) (interface{}, error) {
```
noot, line 172: was this unused?

edwardmack: It's no longer used since we're using the Decode interface for that type.

dot/rpc/modules/state_test.go
edwardmack: I used the results returned from the runtime and assumed they are correct.  I don't how I can verify they are correct, so maybe this is not a good test.  Do you have suggestions how to handle situations like these?

noot: okay, that's fine, maybe just a comment then

Issue #812, opened 2020-04-27
Broadcast Extrinsics that are submitted via RPC


amerameen:
## Expected Behavior

- Extrinsics that are submitted by RPC should also be broadcasted to connected peers

## Current Behavior

- Currently, extrinsics that are submitted via RPC are only processed by the node that receives it and not broadcast (i.e sent out) to the rest of the network (until they are part of a block that is produced by the node that received it via RPC) 

## Possible Solution

- Upon receiving an extrinsic via RPC, the node should send it out via libp2p to it's peers (in addition to adding it to the transaction pool; which it already does)
## Checklist

- [x] I have read [CONTRIBUTING](CONTRIBUTING.md) and [CODE_OF_CONDUCT](CODE_OF_CONDUCT.md)
- [x] I have provided as much information as possible and necessary
- [ ] I am planning to submit a pull request to fix this issue myself

PR #831, opened 2020-04-30
rpc: Broadcast Extrinsics that are submitted via RPC #812

thomasmodeneis:
## Changes

- broadcast extrinsics received via rpc call SubmitExtrinsic
- Update all related tests
- 

## Tests:

```

```

## Checklist:

## Issues:

- closes: #812


thomasmodeneis:
> I think you should be using the core service to broadcast the message. eg. create a method on core.Service called HandleSubmittedExtrinsic that uses the core.Service method `safeMsgSend()`
> 
> this would also remove the need for you to add the NetworkAPI to the author module, and it would also provide safety in case of channel closures

sure, that makes more sense. I was thinking about it tbh but it was not specified in the ticket so I supposed it was up for interpretation. Anyways, let me know if its ok now! :+1:

dot/rpc/modules/author.go
```diff
@@ -169,16 +169,22 @@ func (cm *AuthorModule) SubmitExtrinsic(r *http.Request, req *Extrinsic, res *Ex
 	vtx := transaction.NewValidTransaction(ext, txv)
 
 	if cm.coreAPI.IsBabeAuthority() {
-		hash, err := cm.txQueueAPI.Push(vtx)
-		if err != nil {
-			log.Trace("[rpc] submitted extrinsic failed to push transaction to queue", "error", err)
+		hash, errQueue := cm.txQueueAPI.Push(vtx)
```
thomasmodeneis, line 172: fixed: 
```
dot/rpc/modules/author.go:172:9: shadow: declaration of "err" shadows declaration at line 155 (govet)
                hash, err := cm.txQueueAPI.Push(vtx)
                      ^
```

dot/rpc/modules/author_test.go
```diff
@@ -149,11 +149,12 @@ func TestAuthorModule_SubmitExtrinsic_InQueue(t *testing.T) {
...
 
-	txQueue.Push(expected)
+	_, err := txQueue.Push(expected)
+	require.Nil(t, err)
 
 	// this should cause error since transaction is already in txQueue
-	err := auth.SubmitExtrinsic(nil, &ext, res)
-	require.EqualError(t, err, "transaction is already in pool")
+	err = auth.SubmitExtrinsic(nil, &ext, res)
+	require.Nil(t, err)
```
thomasmodeneis, line 157: This test case is called: TestAuthorModule_SubmitExtrinsic_InQueue but it was expecting a err "transaction is already in pool", but that seemed to be a issue related to the `shadows declaration at line 155 (govet)` issue on `dot/rpc/modules/author.go:172`

dot/rpc/modules/author.go
noot: it might be nicer to declare `var hash common.Hash` outside of the if statement, then you can leave it as `hash, err = cm.txQueueAPI.Push(vtx)`
not a huge fan of renaming err :p

dot/rpc/modules/author_test.go
noot: I'm not sure I understand? this test should return an error

dot/rpc/modules/author.go
```diff
@@ -169,17 +169,23 @@ func (cm *AuthorModule) SubmitExtrinsic(r *http.Request, req *Extrinsic, res *Ex
...
+		if errQueue != nil {
+			log.Trace("[rpc] submitted extrinsic failed to push transaction to queue", "error", errQueue)
 			return err
```
noot, line 175: the reason the test fails is here: you are returning err but should return errQueue

PR #836, opened 2020-05-02
dot/rpc: implement chain_subscribeNewHeads

edwardmack:
## Changes
Basic implementation of WebSocket subscription to push new head data to client.

- Implements chain_subscribeNewHeads
- Hard code methods needed for Polkadot-js integration
-

## Tests:

Start node with RPC enabled:
```
./bin/gossamer --key alice --roles 4 --rpc --rpcmods chain,system,author,state,rpc
```

Make WebSocket call to ws://localhost:8546/ 
```
{
    "jsonrpc": "2.0",
    "method": "chain_subscribeNewHeads",
    "params": [],
    "id": 1
}
```
OR interact via Polkadot-js
Then browse to https://polkadot.js.org/apps/#/settings and configure for custom endpoint ws://127.0.0.1:8546/

Then interact with node through polkadot-js and note errors and issues so that we can determine next areas of focus.

## Checklist:

## Issues:

-


dot/rpc/websocket.go
```diff
@@ -95,3 +190,50 @@ func (h *HTTPServer) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+	initRes.ID = reqID
+
+	return sub, conn.WriteJSON(initRes)
+}
+
+func (h *HTTPServer) blockReceivedListener() {
+	blkRec := h.serverConfig.CoreAPI.GetBlockReceivedChannel()
+	for {
+		// receive block from BABE session
+		block, ok := <-blkRec
```
noot, line 217: I think if a value is read from a channel, it gets taken out of the channel, if there are 2 processes reading from this channel then only one of them will be able to see the value.  I think this will cause an issue since there is a core process also reading from the same channel.
maybe creating a separate channel for websockets would be better? or inside core, in the process that already reads from the block channel, you could send that block to here as well?

edwardmack: @noot I think you can have multiple go routines listening on one channel, according to this post, and the sample code included, it works. https://stackoverflow.com/questions/15715605/multiple-goroutines-listening-on-one-channel

noot: yeah you can, but according to that second comment, if one goroutine reads a value from the channel, then the other goroutines won't see that value, since it's already been read. 

I'm not totally sure though, I'm going to try it out

edwardmack: Ok, I see I misunderstood some of the detail in the sample code.  I'll refactor this to create a separate channel.  Thanks for clarifying.

edwardmack:
This is ready for another look.

dot/rpc/websocket.go
```diff
@@ -38,59 +91,147 @@ func (h *HTTPServer) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
-				log.Error("[rpc] failed to write message to buffer", "error", err)
-				return
+				log.Error("[rpc] websocket failed write message", "error", err)
+			}
+			continue
+		}
+		method := msg["method"]
+		// if method contains subscribe, then register subscription
+		if strings.Contains(fmt.Sprintf("%s", method), "subscribe") {
+			mid := msg["id"].(float64)
```
noot, line 126: is the subscription ID provided by the user?

dot/rpc/websocket.go
```diff
@@ -38,59 +91,147 @@ func (h *HTTPServer) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+		}
+	}
+
+}
+
+func (h *HTTPServer) registerSubscription(conn *websocket.Conn, reqID float64, subscriptionType int) (uint32, error) {
+	wssub := h.serverConfig.WSSubscriptions
+	if wssub == nil {
+		wssub = make(map[uint32]*WebSocketSubscription)
+	}
```
noot, line 198: this map should probably be initialized in the HTTPServer constructor

dot/rpc/websocket_test.go
```diff
@@ -18,16 +19,20 @@ var testCalls = []struct {
...
+	coreAPI := core.NewTestService(t, nil)
+	//blockAPI := NewTestChainService(t)
 	cfg := &HTTPServerConfig{
-		Modules: []string{"system"},
+		Modules: []string{"system", "chain"},
 		RPCPort: 8545,
 		WSPort:  8546,
 		RPCAPI:  NewService(),
+		CoreAPI: coreAPI,
+		//BlockAPI: blockAPI.Block,
```
noot, line 35: remove?

dot/rpc/websocket_test.go
```diff
@@ -18,16 +19,20 @@ var testCalls = []struct {
...
 	{[]byte(`{"jsonrpc":"2.0","method":"unknown","params":[],"id":1}`), []byte(`{"error":{"code":-32000,"data":null,"message":"rpc error method unknown not found"},"id":1,"jsonrpc":"2.0"}` + "\n")}, // unknown method
-	{[]byte{}, []byte(`{"error":{"code":-32700,"data":{"id":null,"jsonrpc":"","method":"","params":null},"message":"EOF"},"id":null,"jsonrpc":"2.0"}` + "\n")},                                        // empty request
+	{[]byte{}, []byte(`{"jsonrpc":"2.0","error":{"code":-32600,"message":"Invalid request"},"id":null}` + "\n")},                                                                                      // empty request
+	{[]byte(`{"jsonrpc":"2.0","method":"chain_subscribeNewHeads","params":[],"id":1}`), []byte(`{"jsonrpc":"2.0","result":1,"id":1}` + "\n")},
 }
 
 func TestNewWebSocketServer(t *testing.T) {
-
+	coreAPI := core.NewTestService(t, nil)
+	//blockAPI := NewTestChainService(t)
```
noot, line 28: remove?

dot/rpc/http.go
```diff
@@ -119,10 +130,19 @@ func (h *HTTPServer) Start() error {
 		}
 	}()
 
+	// init and start block received listener routine
+	if h.serverConfig.BlockAPI != nil {
+		h.serverConfig.BlockAddedReceiver = make(chan *types.Block)
+		h.serverConfig.BlockAddedReceiverDone = make(chan bool)
```
noot, line 136: this can be changed to `make(chan struct{})` and then in `Stop()` you can do `close(BlockAddedReceiverDone)` to notify the BlockState that it's done. it'll receive empty struct and unblock on `doneNotifying` 

it's a nitpick but I think it's usually what's used in go if you want a channel that only notifies

dot/rpc/websocket.go
edwardmack: No, the id provide by the user is the message id, so that can track that to the responses.  The subscription id is created by the service and returned to the requester.

noot: ah okay makes sense!

dot/rpc/http.go
edwardmack: Ok, updated.  I like the idea of applying patterns consistently (although I don't always do it), so thanks for pointing this out.

edwardmack:
@noot I think I address all comments, so it's ready for a re-review.

Issue #740, opened 2020-04-01
fix SIGABRT upon node shutdown


noot:
sometimes when shutting down a node, it SIGABRTs. this shouldn't happen. the issue is probably channels being closed then attempting to write to it, but could be something else.

PR #843, opened 2020-05-06
cmd: fix SIGABRT upon node shutdown #740

thomasmodeneis:
## Changes

- Add test case with the issue as evidence of the problem
- fix SIGABRT issue
- update handleBabeSession to range over s.epochDone instead of the inf inite loop and add its call to initializeBabeSession func
- update stop channel on dot/node and block until is closed 
- remove TODO from main_test.go since now there is a test for it
- add isClosed method to safely check if its closed
- update TestCheckForRuntimeChanges Config IsBabeAuthority to true in order to fully validate that checkForRuntimeChanges calls safeBabeKill and no error is returned
- increase main_test.go timeout to 60s in order to try to catch a err for a extended period
- update dot/core/service.go closed from bool to uint32 and add atomic support to change its state in order to avoid race conditions
- update dot/core/syncer.go stopped from bool to uint32 and add atomic support to change its state in order to avoid race conditions
- add WaitGroup and stateStarted as uint32 to dot/node.go and add atomic support to change its state in order to avoid race conditions
- update node_test.go to have a 300ms wait before check its started, and check stateStarted atomic uint32 field for started/stopped status

- update epochDone from chan to waitGroup and have it being passed around between core and babe packages in order to avoid race conditions with channels and be able to safely remove (for range s.epochDone) from service.handleBabeSession method
- update Session closed field from bool to atomic uint32 in order to safely avoid race conditions when dealing with closed states on lib/babe.go  
- update babe_test.go to reflect new changes

- update main_test.go to execute the test 10 times, wait 10s between each test
- update closed to started on service.go
- add return error to syncer.go Start method and update all related
- add return error to syncer.go Stop method and update all related
- add err checks for service_test.go on state.Stop since it was not being checked before and making the test more resilient
- update closed to startd on lib/babe.go 
- add return error on stop method for lib/babe.go and update checkForKill as well since its using it. - Update places where checkForKill was used to also check for err and log it since panic was not well received on PR review
- update isClosed to isStopped and remove lock since the atomic works without it
- add err check for babe.stop func call and log it if error since all the other error checks for on the same func are doing that only, it seemed like the standard
- update lib/babe_test.go to check for the babeSession.started to be 0 when killChan is closed in a retry basis since that is a eventual consistency check, 10x with 1s in between checks

## Tests:

```
go test ./cmd/... -test.v -test.run "^TestGossamerCommand$"
```

## Checklist:

## Issues:

- closes: #740


dot/node.go
```diff
@@ -299,10 +293,4 @@ func (n *Node) Stop() {
 	// stop all node services
 	n.Services.StopAll()
 
-	// close node stop channel if not already closed
-	if n.stop != nil {
-		close(n.stop)
-	}
```
noot, line 305: this can probably stay? it's only being closed in one spot and read from in one spot

noot:
@thomasmodeneis I'm still seeing the error when shutting down the node:
```
SIGABRT: abort
PC=0x7f0d57b96e97 m=20 sigcode=18446744073709551610

goroutine 153 [running]:
runtime: unknown pc 0x7f0d57b96e97
stack: frame={sp:0xc00a5c16d0, fp:0x0} stack=[0xc0001bc000,0xc0001cc000)

runtime: unknown pc 0x7f0d57b96e97
stack: frame={sp:0xc00a5c16d0, fp:0x0} stack=[0xc0001bc000,0xc0001cc000)

created by github.com/ChainSafe/gossamer/lib/babe.(*Session).Start
	/home/elizabeth/go/src/github.com/ChainSafe/gossamer/lib/babe/babe.go:159 +0x274
```

dot/node.go
noot: does this stop need to be removed?

thomasmodeneis:
> got an error shutting down at block 68 :(
> 
> ```
> ^CINFO[05-12|15:54:17] [dot] signal interrupt, shutting down... 
> INFO[05-12|15:54:17] Stopping services: [*state.Service *core.Service *network.Service] 
> WARN[05-12|15:54:17] [network] Received nil message from core service 
> SIGABRT: abort
> PC=0x7fdf60e50e97 m=15 sigcode=18446744073709551610
> 
> goroutine 406 [running]:
> runtime: unknown pc 0x7fdf60e50e97
> stack: frame={sp:0xc0001bb6d0, fp:0x0} stack=[0xc000252000,0xc000262000)
> 
> runtime: unknown pc 0x7fdf60e50e97
> stack: frame={sp:0xc0001bb6d0, fp:0x0} stack=[0xc000252000,0xc000262000)
> 
> created by github.com/ChainSafe/gossamer/lib/babe.(*Session).Start
> 	/home/elizabeth/go/src/github.com/ChainSafe/gossamer/lib/babe/babe.go:159 +0x274
> 
> goroutine 1 [select (no cases)]:
> github.com/ChainSafe/gossamer/dot.(*Node).Start(0xc00020bdd0)
> 	/home/elizabeth/go/src/github.com/ChainSafe/gossamer/dot/node.go:287 +0xab
> main.gossamerAction(0xc0001469a0, 0xc0001469a0, 0xc0002d1b9f)
> 	/home/elizabeth/go/src/github.com/ChainSafe/gossamer/cmd/gossamer/main.go:166 +0x7f0
> github.com/urfave/cli.HandleAction(0xd882c0, 0x1041fe8, 0xc0001469a0, 0xc0000c14a0, 0x0)
> 	/home/elizabeth/go/pkg/mod/github.com/urfave/cli@v1.20.0/app.go:490 +0xc8
> github.com/urfave/cli.(*App).Run(0xc0002fc820, 0xc000032180, 0x3, 0x3, 0x0, 0x0)
> 	/home/elizabeth/go/pkg/mod/github.com/urfave/cli@v1.20.0/app.go:264 +0x58c
> main.main()
> 	/home/elizabeth/go/src/github.com/ChainSafe/gossamer/cmd/gossamer/main.go:91 +0x55
> ```

Right ... I will keep searching then :)

dot/node.go
thomasmodeneis: hmmm, yes, not really. I restored it and it seems to work properly when using range loop. :+1:

cmd/gossamer/main_test.go
```diff
@@ -193,7 +227,56 @@ func TestInvalidCommand(t *testing.T) {
 	}
 }
 
-// TODO: TestGossmaerCommand test "gossamer" does not error
+func TestGossamerCommand(t *testing.T) {
+	basePort := 7000
+	currentDir, err := os.Getwd()
+	require.Nil(t, err)
+
+	genesisPath := filepath.Join(currentDir, "../..", "node/gssmr/genesis.json")
```
ryanchristo, line 235: You can use `utils.GetGssmrGenesisPath()` here within the `lib/utils` package.

dot/node.go
ryanchristo: Looks like @noot requested this not be removed and you mention restoring it - could you clarify where/how you restored it or is this no longer an issue?

dot/core/service.go
```diff
@@ -329,51 +329,50 @@ func (s *Service) safeBabeKill() error {
 }
 
 func (s *Service) handleBabeSession() {
-	for {
-		// wait for BABE epoch to complete
-		<-s.epochDone
+	// wait for BABE epoch to complete
+	for range s.epochDone {
+	}
```
noot, line 334: change to `<-s.epochDone` ? I don't think there's much need to `range` here

thomasmodeneis: the range is there in order to try to avoid race conditions that trigger a panic. Since this is one of the many options where that could happen, I added a for range in order to mitigate the issue and pinpoint the real issue. If in the future we feel more confident, we can revert it back.

thomasmodeneis:
> i got a SIGABRT when stopping the node, restarting it, and stopping it again :(
> 
> ```
> WARN[05-14|13:48:29] [network] Received nil message from core service 
> SIGABRT: abort
> PC=0x7f5b64eb7e97 m=4 sigcode=18446744073709551610
> 
> goroutine 72 [running]:
> runtime: unknown pc 0x7f5b64eb7e97
> stack: frame={sp:0xc0000976d0, fp:0x0} stack=[0xc0001f4000,0xc000204000)
> 
> runtime: unknown pc 0x7f5b64eb7e97
> stack: frame={sp:0xc0000976d0, fp:0x0} stack=[0xc0001f4000,0xc000204000)
> 
> created by github.com/ChainSafe/gossamer/lib/babe.(*Session).Start
> 	/home/elizabeth/go/src/github.com/ChainSafe/gossamer/lib/babe/babe.go:159 +0x274
> 
> goroutine 1 [select (no cases)]:
> github.com/ChainSafe/gossamer/dot.(*Node).Start(0xc000245650)
> 	/home/elizabeth/go/src/github.com/ChainSafe/gossamer/dot/node.go:287 +0xab
> main.gossamerAction(0xc000158b00, 0xc000158b00, 0xc0002e3b9f)
> 	/home/elizabeth/go/src/github.com/ChainSafe/gossamer/cmd/gossamer/main.go:166 +0x7f0
> github.com/urfave/cli.HandleAction(0xd88700, 0x1042488, 0xc000158b00, 0xc0000a7440, 0x0)
> 	/home/elizabeth/go/pkg/mod/github.com/urfave/cli@v1.20.0/app.go:490 +0xc8
> github.com/urfave/cli.(*App).Run(0xc00031a680, 0xc0000b2150, 0x3, 0x3, 0x0, 0x0)
> 	/home/elizabeth/go/pkg/mod/github.com/urfave/cli@v1.20.0/app.go:264 +0x58c
> main.main()
> 	/home/elizabeth/go/src/github.com/ChainSafe/gossamer/cmd/gossamer/main.go:91 +0x55
> ```

I updated more things in order to mitigate this even more:
```
increase main_test.go timeout to 60s in order to try to catch a err for a extended period
update dot/core/service.go closed from bool to uint32 and add atomic support to change its state in order to avoid race conditions
update dot/core/syncer.go stopped from bool to uint32 and add atomic support to change its state in order to avoid race conditions
add WaitGroup and stateStarted as uint32 to dot/node.go and add atomic support to change its state in order to avoid race conditions
update node_test.go to have a 300ms wait before check its started, and check stateStarted atomic uint32 field for started/stopped status
```

Could you please re-test ? Cheers.

dot/node.go
thomasmodeneis: I restored it with a wait group. Please check again.

cmd/gossamer/main_test.go
```diff
@@ -193,7 +227,56 @@ func TestInvalidCommand(t *testing.T) {
...
+
+	tempDir, err := ioutil.TempDir("", "gossamer-stress-")
```
noot, line 237: not stress?

cmd/gossamer/main_test.go
```diff
@@ -193,7 +227,56 @@ func TestInvalidCommand(t *testing.T) {
...
+
+	gossamer := runTestGossamer(t,
+		"init",
+		"--datadir", tempDir,
+		"--genesis", genesisPath,
+		"--force",
+	)
+
+	stdout, stderr := gossamer.GetOutput()
+	log.Println("init gossamer output, ", "stdout", string(stdout), "stderr", string(stderr))
```
noot, line 248: is there a reason `log` is used instead of `t.Log`? not really a big deal, just wondering

cmd/gossamer/main_test.go
```diff
@@ -193,7 +227,56 @@ func TestInvalidCommand(t *testing.T) {
...
+		"--port", strconv.Itoa(basePort),
+		"--key", "alice",
+		"--datadir", tempDir,
+		"--roles", "4",
+	)
+
+	time.Sleep(60 * time.Second)
+
+	stdout, stderr = gossamer.GetOutput()
+	log.Println("init gossamer output, ", "stdout", string(stdout), "stderr", string(stderr))
```
noot, line 269: not init

dot/core/service.go
```diff
@@ -153,12 +154,17 @@ func NewService(cfg *Config) (*Service, error) {
...
-			closed:           false,
 			syncLock:         syncerLock,
 			blockNumOut:      cfg.SyncChan,
 			respOut:          respChan,
 		}
 
+		// thread safe way to change closed status
+		canLock := atomic.CompareAndSwapUint32(&srv.closed, 0, 1)
+		if !canLock {
+			panic("[core] Error when trying to change Service status from stopped to started.")
```
noot, line 165: it would be preferred not to panic here and return error instead

dot/core/service.go
```diff
@@ -204,12 +210,17 @@ func NewService(cfg *Config) (*Service, error) {
...
+		}
```
noot, line 222: duplicated from above, can be moved to below the `else` statement

dot/core/service.go
```diff
@@ -285,14 +296,23 @@ func (s *Service) Stop() error {
...
 		if s.isBabeAuthority {
 			close(s.babeKill)
 		}
-		s.closed = true
+
+		defer func() {
+			// thread safe way to change closed status
+			canUnlock := atomic.CompareAndSwapUint32(&s.closed, 1, 0)
+			if !canUnlock {
+				panic("[core] Error when trying to change Service status from started to stopped.")
```
noot, line 312: return error instead of panic

dot/core/service.go
noot: could you clarify what race condition would be happening here? receiving from a channel will never panic, I'd prefer if you could change it back

dot/core/syncer.go
```diff
@@ -125,19 +125,28 @@ func NewSyncer(cfg *SyncerConfig) (*Syncer, error) {
 
 // Start begins the syncer
 func (s *Syncer) Start() {
+	// thread safe change stopped to 1
+	canLock := atomic.CompareAndSwapUint32(&s.stopped, 0, 1)
+	if !canLock {
+		panic("[sync] Error when trying to change Syncer from stopped to started.")
```
noot, line 131: return error

dot/core/syncer.go
```diff
@@ -125,19 +125,28 @@ func NewSyncer(cfg *SyncerConfig) (*Syncer, error) {
...
 	go s.watchForResponses()
 }
 
 // Stop stops the syncer
 func (s *Syncer) Stop() {
 	// stop goroutines
-	s.stopped = true
+	canUnlock := atomic.CompareAndSwapUint32(&s.stopped, 1, 0)
+	if !canUnlock {
+		panic("[sync] Error when trying to change Syncer from started to stopped.")
```
noot, line 142: return error

dot/core/service.go
thomasmodeneis: This is now completely gone since it was redundant and controlling state with channels is really not recommended. Please re-check.

cmd/gossamer/main_test.go
thomasmodeneis: sure

cmd/gossamer/main_test.go
thomasmodeneis: Hmmmm, that was because I was not seeing the logs right and tried to change it in order to see a different state. Should be updated.

cmd/gossamer/main_test.go
thomasmodeneis: sure.

dot/core/service.go
thomasmodeneis: sure

dot/core/service.go
thomasmodeneis: I updated the panic to err, but moving it to after its not really the same. Since the previous block is not the same as it will be calling `babe.NewSession(bsConfig)`, if we move it to the else and do not execute the action before that, it will not have the right logic.

dot/core/service.go
thomasmodeneis: That wont be possible since its inside a defer func....

dot/core/syncer.go
thomasmodeneis: The start method does not return error right now. Are you sure this change is required ?

dot/core/syncer.go
thomasmodeneis: The stop method does not return error right now, this will add extra logic into this PR. Are you sure ?

dot/core/syncer.go
noot: yes, the syncer is only started in one place, so I don't think it's a lot of logic to add for an err check.

dot/core/syncer.go
noot: yes, the syncer is only stopped in one place, so I don't think it's a lot of logic to add for an err check.

dot/core/service.go
noot: ok, log it then instead?

lib/babe/babe.go
```diff
@@ -110,12 +111,16 @@ func NewSession(cfg *SessionConfig) (*Session, error) {
...
-		done:             cfg.Done,
+		epochDone:        cfg.EpochDone,
 		kill:             cfg.Kill,
-		closed:           false,
 		syncLock:         cfg.SyncLock,
 	}
 
+	canLock := atomic.CompareAndSwapUint32(&babeSession.closed, 0, 1)
+	if !canLock {
+		panic("[core] Error when trying to change Service status from stopped to started.")
```
noot, line 121: return error

lib/babe/babe.go
```diff
@@ -167,10 +172,13 @@ func (b *Session) stop() {
 	b.lock.Lock()
 	defer b.lock.Unlock()
 
-	if !b.closed {
+	if atomic.LoadUint32(&b.closed) == uint32(1) {
+		canUnlock := atomic.CompareAndSwapUint32(&b.closed, 1, 0)
+		if !canUnlock {
+			panic("[babe] Error when trying to change Service status from started to stopped.")
```
noot, line 178: return error or log

dot/core/service.go
```diff
@@ -321,59 +343,59 @@ func (s *Service) safeMsgSend(msg network.Message) error {
...
 	return nil
 }
 
 func (s *Service) handleBabeSession() {
-	for {
-		// wait for BABE epoch to complete
-		<-s.epochDone
+	// wait for BABE epoch to complete
+	s.epochDone.Add(1)
+	s.epochDone.Wait()
```
noot, line 357: this seems excessively complicated since there was no issue with the channel, and there isn't really much point to using a WaitGroup for 1 process. can you show a case when receiving from a channel would cause a panic or other issue you are talking about? if not I would prefer it to be changed back

lib/babe/babe.go
noot: also, this is changing `b.closed` from 1 to 0, which doesn't really make sense since we are closing it. maybe change the variable name to `stopped` ?

lib/babe/babe.go
noot: also, you are changing `closed` to 1 which doesn't really make sense here. change the variable name to `stopped` ?

also, shouldn't this be moved to `babe.Start` instead of `NewSession`? since it's not actually being started here

lib/babe/babe.go
```diff
@@ -185,7 +193,7 @@ func (b *Session) Descriptor() *NextEpochDescriptor {
 func (b *Session) safeSend(msg types.Block) error {
 	b.lock.Lock()
 	defer b.lock.Unlock()
-	if b.closed {
+	if atomic.LoadUint32(&b.closed) == uint32(0) {
```
noot, line 196: also update here if you change the variable name: `b.stopped == uint32(1)`

lib/babe/babe.go
```diff
@@ -219,15 +227,26 @@ func (b *Session) setAuthorityIndex() error {
...
+	if isClosed(b.kill) {
+		b.stop()
+	}
 }
 
 func (b *Session) isClosed() bool {
 	b.lock.Lock()
 	defer b.lock.Unlock()
-	return b.closed
+	return atomic.LoadUint32(&b.closed) == uint32(0)
```
noot, line 249: same here

noot: also, do we need a lock still if this is atomic?

dot/core/syncer.go
noot: also, you are chaning `s.stopped` to 1 in the start method, shouldn't it be changed to `0`?

dot/core/syncer.go
noot: also, you are changing `s.stopped` to 0 in the Stop method, shouldn't it be changed to `1`?

dot/core/service.go
```diff
@@ -204,12 +211,17 @@ func NewService(cfg *Config) (*Service, error) {
...
 			isBabeAuthority:  false,
 			lock:             chanLock,
-			closed:           false,
 			syncLock:         syncerLock,
 			blockNumOut:      cfg.SyncChan,
 			respOut:          respChan,
 		}
 
+		// thread safe way to change closed status
+		canLock := atomic.CompareAndSwapUint32(&srv.closed, 0, 1)
```
noot, line 220: `srv.closed` doesn't need to be updated in `NewService`, it should be updated in `Start` and `Stop`

also, `srv.closed` should be set to 0, not 1, if starting the service

thomasmodeneis: Yes, I agree that the srv.closed should be inside the Start, but this is a design change. It was like this before I started. Besides, the following tests are going to require changes since they are then failing:

```
--- FAIL: TestService_ProcessBlockAnnounceMessage (0.90s)
    messages_test.go:345: 
        	Error Trace:	messages_test.go:345
        	Error:      	Expected nil, but got: &errors.errorString{s:"[core] Error when trying to change Service status from stopped to started"}
        	Test:       	TestService_ProcessBlockAnnounceMessage

--- FAIL: TestStartService (0.94s)
    service_test.go:123: 
        	Error Trace:	service_test.go:123
        	Error:      	Expected nil, but got: &errors.errorString{s:"[core] Error when trying to change Service status from stopped to started"}
        	Test:       	TestStartService

--- FAIL: TestAnnounceBlock (0.44s)
    service_test.go:151: 
        	Error Trace:	service_test.go:151
        	Error:      	Expected nil, but got: &errors.errorString{s:"[core] Error when trying to change Service status from stopped to started"}
        	Test:       	TestAnnounceBlock

--- FAIL: TestCheckForRuntimeChanges (0.93s)
    service_test.go:210: 
        	Error Trace:	service_test.go:210
        	Error:      	Expected nil, but got: &errors.errorString{s:"service has been stopped"}
        	Test:       	TestCheckForRuntimeChanges

--- FAIL: TestAuthorModule_SubmitExtrinsic (1.73s)
    author_test.go:75: 
        	Error Trace:	author_test.go:75
        	Error:      	Expected nil, but got: &errors.errorString{s:"service has been stopped"}
        	Test:       	TestAuthorModule_SubmitExtrinsic
```

This will cause the scope will creep even more. Are you sure you want this as part of this defect ?

dot/core/syncer.go
thomasmodeneis: I updated this to started and updated the method signature as requested, please re-check.

lib/babe/babe.go
thomasmodeneis: I updated this to started, please re-check.

lib/babe/babe.go
thomasmodeneis: I updated this to started and removed the lock, please re-check.

dot/core/service.go
thomasmodeneis: Well, that was the whole point of the PR. Using channels is not indicated, WaitGroups behave way better. I strongly recommend we do not go back to that since this is working much better as we see on the improvements.

lib/babe/babe.go
thomasmodeneis: I updated this to started, please re-check.

thomasmodeneis: Also updated the return.

lib/babe/babe.go
thomasmodeneis: I updated this to started, please re-check. Changing the lock to another method is another behavior change that will increase the scope. If this is required, we should do it as part of the same request for the core/service.go.

dot/core/syncer.go
thomasmodeneis: I updated this to started, please re-check.

dot/core/service.go
thomasmodeneis: logging it will not be as easy to identify, I would recommend to panic. But, as you wish.

cmd/gossamer/main_test.go
noot: reminder of this

dot/core/syncer.go
```diff
@@ -124,20 +125,32 @@ func NewSyncer(cfg *SyncerConfig) (*Syncer, error) {
 }
 
 // Start begins the syncer
-func (s *Syncer) Start() {
+func (s *Syncer) Start() error {
+	// thread safe change stopped to 1
+	canLock := atomic.CompareAndSwapUint32(&s.stopped, 0, 1)
```
noot, line 130: renamed to `started`

dot/core/service.go
noot: okay, it's fine as is then

dot/core/syncer.go
thomasmodeneis: sure

dot/core/service.go
```diff
@@ -153,12 +155,17 @@ func NewService(cfg *Config) (*Service, error) {
...
-			closed:           false,
 			syncLock:         syncerLock,
 			blockNumOut:      cfg.SyncChan,
 			respOut:          respChan,
 		}
 
+		// thread safe way to change closed status
+		canLock := atomic.CompareAndSwapUint32(&srv.started, 0, 1)
+		if !canLock {
+			return nil, errors.New("[core] Error when trying to change Service status from stopped to started")
```
noot, line 166: change error to `failed to change started to 1` or something along those lines, to align with error standards

dot/core/service.go
```diff
@@ -204,12 +211,17 @@ func NewService(cfg *Config) (*Service, error) {
...
```
noot, line 222: same here

dot/core/syncer.go
```diff
@@ -124,20 +125,32 @@ func NewSyncer(cfg *SyncerConfig) (*Syncer, error) {
 }
 
 // Start begins the syncer
-func (s *Syncer) Start() {
+func (s *Syncer) Start() error {
+	// thread safe change stopped to 1
+	canLock := atomic.CompareAndSwapUint32(&s.started, 0, 1)
+	if !canLock {
+		return errors.New("[sync] Error when trying to change Syncer from stopped to started")
```
noot, line 132: same here

dot/core/syncer.go
```diff
@@ -124,20 +125,32 @@ func NewSyncer(cfg *SyncerConfig) (*Syncer, error) {
...
 }
 
 // Stop stops the syncer
-func (s *Syncer) Stop() {
+func (s *Syncer) Stop() error {
 	// stop goroutines
-	s.stopped = true
+	canUnlock := atomic.CompareAndSwapUint32(&s.started, 1, 0)
+	if !canUnlock {
+		return errors.New("[sync] Error when trying to change Syncer from started to stopped")
```
noot, line 145: same here

lib/babe/babe.go
```diff
@@ -110,12 +111,16 @@ func NewSession(cfg *SessionConfig) (*Session, error) {
...
-		done:             cfg.Done,
+		epochDone:        cfg.EpochDone,
 		kill:             cfg.Kill,
-		closed:           false,
 		syncLock:         cfg.SyncLock,
 	}
 
+	canLock := atomic.CompareAndSwapUint32(&babeSession.started, 0, 1)
+	if !canLock {
+		return nil, errors.New("[core] Error when trying to change Service status from stopped to started")
```
noot, line 121: same here, this also isn't core

lib/babe/babe.go
```diff
@@ -158,20 +163,29 @@ func (b *Session) Start() error {
...
-func (b *Session) stop() {
+func (b *Session) stop() error {
 	b.lock.Lock()
 	defer b.lock.Unlock()
 
-	if !b.closed {
+	if atomic.LoadUint32(&b.started) == uint32(1) {
+		canUnlock := atomic.CompareAndSwapUint32(&b.started, 1, 0)
+		if !canUnlock {
+			return errors.New("[babe] Error when trying to change Service status from started to stopped")
```
noot, line 183: same here

dot/core/service.go
thomasmodeneis: wait, yes is true we change the status to 1, but in fact the representation of it is that we change it to started/ stopped. I would argue that the message is more meaningful than saying 0 or 1. Don't you agree ?

noot: okay, the point was to align with error standards, so remove `[core]` and change `error when trying` to `failed`

thomasmodeneis: roger

thomasmodeneis: done

dot/core/service.go
thomasmodeneis: done

dot/core/syncer.go
thomasmodeneis: done

dot/core/syncer.go
thomasmodeneis: done

lib/babe/babe.go
thomasmodeneis: done

lib/babe/babe.go
thomasmodeneis: done

Issue #731, opened 2020-03-30
Implement system RPC calls & error codes


amerameen:
Work on each of these system RPC calls one PR at a time. Check mark completed calls as completed. 

**Reference:** https://github.com/w3f/PSPs/blob/master/PSPs/drafts/psp-6.md

**Please include integration test with RPC call implementation**

### System RPC Calls
- [x] [`accountNextIndex`](https://github.com/w3f/PSPs/blob/master/PSPs/drafts/psp-6.md#1516-system_accountnextindex)
- [x] [`addReservedPeer`](https://github.com/w3f/PSPs/blob/master/PSPs/drafts/psp-6.md#1512-system_addreservedpeer)
- [x] [`localListenAddresses`](https://github.com/w3f/PSPs/blob/master/PSPs/drafts/psp-6.md#159-system_locallistenaddresses)
- [x] [`localPeerId`](https://github.com/w3f/PSPs/blob/master/PSPs/drafts/psp-6.md#158-system_localpeerid)
- [x] [`removeReservedPeer`](https://github.com/w3f/PSPs/blob/master/PSPs/drafts/psp-6.md#1513-system_removereservedpeer)
- [x] [`syncState`](https://github.com/w3f/PSPs/blob/master/PSPs/drafts/psp-6.md#1515-system_syncstate)

edwardmack:
Questions:
- system_addReserverPeer/ system_removeReservedPeer, are we currently storing this info?  Need understanding where I would add/remove these.
- system_accountNextIndex - nonce for given account, where would I find this?
- chain related data, seems to come from genesis definition, but our current test chain doesn't contain this, so need to research/test how we're going to handle these things (system_chain, system_properties)

EclesioMeloJunior:
Issue #1721 to `dryRun` RPC call created

EclesioMeloJunior:
This issue was closed by PR #1712

PR #849, opened 2020-05-11
dot/rpc: Implement RPC system_name, system_version, system_chain

edwardmack:
## Changes

Added code to implement RPC calls:
- system_name
- system_version
- system_chain
- system_properties, these need to be loaded by genesis see issue #865 

## Tests:

```
go test ./dot/rpc/...
```

## Checklist:

## Issues:

- related to #731


dot/node.go
```diff
@@ -170,7 +172,7 @@ func NodeInitialized(datadir string, expected bool) bool {
 }
 
 // NewNode creates a new dot node from a dot node configuration
-func NewNode(cfg *Config, ks *keystore.Keystore) (*Node, error) {
+func NewNode(cfg *Config, ks *keystore.Keystore, ctx *cli.Context) (*Node, error) {
```
noot, line 175: `name` and `id` can be found in the `Config`: https://github.com/ChainSafe/gossamer/blob/development/dot/config.go#L45
I think it would make more sense to put the version into the config, and then start the RPC service passing that info, rather than passing the `*cli.Context`, since the `*cli.Context` should only be used when the node is starting up

edwardmack: I was going to put system_name, system_version and system_chain (chain type) into the config, but I didn't because it seems that things in config can be over written by the .toml file, and these values should not be able to be set by config.  I agree I'd rather not keep a reference to `*cli.Context`, so I need to store to store App.name and App.version somewhere for system_name and system_version.  Should I store this in config? or add to one of the services?  Which one?  Regarding name in config, I'm using that for system_chain although I not sure that is correct.  Where can I find chain related values (system_chain, and system_properties: ss58Format, tokenDecimals, and tokenSymbol)?

edwardmack:
I've refactored this, it's ready for review.

dot/config.go
```diff
@@ -24,21 +24,22 @@ import (
...
-	Core    CoreConfig    `toml:"core"`
-	Network NetworkConfig `toml:"network"`
-	RPC     RPCConfig     `toml:"rpc"`
+	Global  GlobalConfig     `toml:"global"`
+	Init    InitConfig       `toml:"init"`
+	Account AccountConfig    `toml:"account"`
+	Core    CoreConfig       `toml:"core"`
+	Network NetworkConfig    `toml:"network"`
+	RPC     RPCConfig        `toml:"rpc"`
+	System  types.SystemInfo `toml:"-"`
```
noot, line 42: does this exclude it from the config file?

dot/system/service.go
```diff
@@ -0,0 +1,51 @@
...
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with the gossamer library. If not, see <http://www.gnu.org/licenses/>.
+
+package system
+
+import "github.com/ChainSafe/gossamer/dot/types"
+
+// Service struct to hold rpc service data
+type Service struct {
```
noot, line 22: this looks good! however, it doesn't look like it's started when the node is started in dot `NewNode`?

dot/config.go
edwardmack: Yes.

dot/system/service.go
edwardmack: I don't have this started because there is no go routine to start, so I guess it's not really a service.  It's currently a place to hold static chain and app values.  Should I implement the Service interface to add the Start and Stop functions and add call to start so it's treated like other services (for consistency, and in case we add a go routine later).  OR should I treat this differently since it's not really a service, and if so, how?

ryanchristo: Should it be part of core?

noot: maybe just rename it from `Service` to `Info` or something like that, since there isn't a need to make it into an actual service

ryanchristo: Hmmm... Not entirely sure at the moment. If you think it is going to be a service (adhere to the service interface) and should be its own service, I think it works as is.

edwardmack: I made it implement the Service interface, so it's handled like the others.  I was running into issues with cyclic references when I tried making it a part of rpc service.  This seems ok since it's following the pattern (now) that we're using for the others.

dot/node.go
```diff
@@ -235,13 +235,19 @@ func NewNode(cfg *Config, ks *keystore.Keystore) (*Node, error) {
 
 	}
 
+	// System Service
+
+	// create system service and append to node services
+	sysSrvc := createSystemService(&cfg.System)
+	nodeSrvcs = append(nodeSrvcs, sysSrvc)
```
ryanchristo, line 242: :+1:

dot/node.go
```diff
@@ -235,13 +235,19 @@ func NewNode(cfg *Config, ks *keystore.Keystore) (*Node, error) {
...
+
 	// RPC Service
 
 	// check if rpc service is enabled
 	if enabled := RPCServiceEnabled(cfg); enabled {
 
 		// create rpc service and append rpc service to node services
-		rpcSrvc := createRPCService(cfg, stateSrvc, coreSrvc, networkSrvc, rt)
+		rpcSrvc := createRPCService(cfg, stateSrvc, coreSrvc, networkSrvc, rt, sysSrvc)
```
ryanchristo, line 250: :+1:

cmd/gossamer/config.go
```diff
@@ -21,12 +21,11 @@ import (
 	"strconv"
 	"strings"
 
+	database "github.com/ChainSafe/chaindb"
```
ryanchristo, line 24: was this intentionally moved?

edwardmack: no, maybe my IDE did it when I did goimports? not sure.

Issue #450, opened 2019-12-03
grandpa: check validity of a vote


noot:
see spec 5.17. need to ensure that a vote for a certain round is valid.

PR #870, opened 2020-05-19
lib/grandpa: begin grandpa service, implement vote validation

noot:
## Changes

- create grandpa.Service which contains the information needed for the grandpa protocol to run
- add FullVote and SignedMessage types
- implement Ed25519Keyring since grandpa uses ed25519 keys
- implement ValidateMessage which validates a VoteMessage
- implement validateVote which validates a Vote

## Tests:

```
go test ./lib/grandpa
```

## Checklist:

## Issues:

- #450


dot/state/block.go
```diff
@@ -280,6 +280,17 @@ func (bs *BlockState) GetBlockBody(hash common.Hash) (*types.Body, error) {
 	return types.NewBody(data), nil
 }
 
+// GetFinalizedHead returns the latest finalized block header
+// TODO: relies on GRANDPA implementation. currently returns genesis header.
```
ryanchristo, line 284: do we have a specific issue we can tag to track this?

dot/state/test_helpers.go
```diff
@@ -0,0 +1,90 @@
...
+
+		currentChain = append(currentChain, block.Header)
+
+		hash := block.Header.Hash()
+		err := blockState.AddBlockWithArrivalTime(block, arrivalTime)
+		require.Nil(t, err)
+
+		previousHash = hash
+
+		isBranch := r.Intn(2)
```
ryanchristo, line 51: Do we know the seed will produce 50/50 results? Do we need to set the seed using `rand.Seed()`? Probably good as is just want to double check because I have not used this myself before.

lib/grandpa/types.go
```diff
@@ -17,39 +17,76 @@
...
+// this is encoded and signed and the signature is included in SignedMessage
+type FullVote struct {
+	stage byte
+	vote  *Vote
+	round uint64
+	setID uint64
+}
+
+// VoteMessage represents a network-level vote message
+// https://github.com/paritytech/substrate/blob/master/client/finality-grandpa/src/communication/gossip.rs#L336
```
ryanchristo, line 75: awesome, nice to have the link here and below

dot/state/block.go
noot: no, can add one

dot/state/test_helpers.go
noot: the seed should ideally produce 50/50 results, it seems to work when you run the test. it's not super important if it's slightly off since it's just to create random tree branching

lib/grandpa/types.go
noot: yeah I'll try to add links when I can!

dot/state/block.go
noot: #883

PR #879, opened 2020-05-20
lib/grandpa, lib/blocktree, dot/state: add functionality to determine pre-voted block

### LINKED ISSUE #875: grandpa: determine current pre-voted block
## Expected Behavior

- see spec Definition 6.22
- current pre-voted block is the block with the highest number that has observable votes >= 2/3 |voter set| 
- see definition 6.20 and 6.21 for counting observable votes

## Checklist

noot:
## Changes

- add `IsDescendantOf`, `Leaves` functions to blocktree, use in GRANDPA
- add functionality to determine pre-voted block from the current votes and equivocations set
- see spec definitions 6.20 to 6.22

## Tests:

```
go test ./lib/grandpa
```

## Checklist:

## Issues:

- closes #875


lib/grandpa/vote_message_test.go
```diff
@@ -0,0 +1,267 @@
+package grandpa
```
noot, line 1: this file contains the tests that were in grandpa.go

lib/grandpa/vote_message.go
```diff
@@ -0,0 +1,155 @@
+package grandpa
```
noot, line 1: these functions are all copied from grandpa.go

lib/grandpa/vote_message.go
```diff
@@ -0,0 +1,171 @@
...
+	})
+	if err != nil {
+		return err
+	}
+	ok, err := pk.Verify(msg, m.message.signature[:])
+	if err != nil {
+		return err
+	}
+
+	if !ok {
```
thomasmodeneis, line 166: this can be a else if since its related to the previous statement.

lib/grandpa/vote_message.go
```diff
@@ -0,0 +1,171 @@
...
+		return ErrBlockDoesNotExist
+	}
+
+	// check if the block is an eventual descendant of a previously finalized block
+	isDescendant, err := s.blockState.IsDescendantOf(s.head, v.hash)
+	if err != nil {
+		return err
+	}
+
+	if !isDescendant {
```
thomasmodeneis, line 144: this could be a else if, since is directly related to the previous statement.

lib/grandpa/vote_message.go
```diff
@@ -0,0 +1,171 @@
...
+// validateVote checks if the block that is being voted for exists, and that it is a descendant of a
+// previously finalized block.
+func (s *Service) validateVote(v *Vote) error {
+	// check if v.hash corresponds to a valid block
+	has, err := s.blockState.HasHeader(v.hash)
+	if err != nil {
+		return err
+	}
+
+	if !has {
```
thomasmodeneis, line 134: this could be a else if since its related to previous statement

lib/grandpa/types.go
```diff
@@ -62,6 +94,29 @@ func NewVoteFromHeader(h *types.Header) *Vote {
...
 }
 
+// NewVoteFromHash returns a new Vote given a hash and a blockState
+func NewVoteFromHash(hash common.Hash, blockState BlockState) (*Vote, error) {
+	has, err := blockState.HasHeader(hash)
+	if err != nil {
+		return nil, err
+	}
+
+	if !has {
```
thomasmodeneis, line 104: this could be a else if, since its related to previous statement.

lib/grandpa/types.go
```diff
@@ -40,6 +43,35 @@ type State struct {
...
+	}
+}
+
+// pubkeyToVoter returns a Voter given a public key
+func (s *State) pubkeyToVoter(pk *ed25519.PublicKey) (*Voter, error) {
+	id := uint64(2^64) - 1
+
+	for i, v := range s.voters {
+		if bytes.Equal(pk.Encode(), v.key.Encode()) {
+			id = uint64(i)
```
thomasmodeneis, line 61: you could add a break statement, since after this expression there is no reason to continue the loop...

lib/grandpa/types.go
```diff
@@ -40,6 +43,35 @@ type State struct {
...
+		}
+	}
+
+	if id == (2^64)-1 {
```
thomasmodeneis, line 65: the magic number `(2^64)-1` could be a var and have a name that defines what is this for...

lib/grandpa/grandpa.go
```diff
@@ -38,178 +50,192 @@ func NewService(blockState BlockState, voters []*Voter) (*Service, error) {
...
-	}
+		// check if the current block is a descendant of B
+		isDescendant, err := s.blockState.IsDescendantOf(hash, v.hash)
+		if err != nil {
+			return 0, err
+		}
 
-	return nil
-}
+		if !isDescendant {
```
thomasmodeneis, line 208: this could be a else if, since its related to previous statement.

lib/grandpa/grandpa.go
```diff
@@ -38,178 +50,192 @@ func NewService(blockState BlockState, voters []*Voter) (*Service, error) {
...
+		}
 
-	if !ok {
-		return ErrInvalidSignature
+		total, err := s.getTotalVotesForBlock(pred)
+		if err != nil {
+			return nil, err
+		}
+
+		if total >= uint64(2*len(s.state.voters)/3) {
```
thomasmodeneis, line 157: uint64(2*len(s.state.voters)/3) could be a var and named with something that gives a indication of why this is here for. The current if condition is somewhat vague without it.

lib/grandpa/grandpa.go
```diff
@@ -38,178 +50,192 @@ func NewService(blockState BlockState, voters []*Voter) (*Service, error) {
...
-		setID: m.setID,
-	})
-	if err != nil {
-		return err
-	}
-	ok, err := pk.Verify(msg, m.message.signature[:])
-	if err != nil {
-		return err
-	}
+		if pred == curr {
```
thomasmodeneis, line 148: this could be a else if condition since its related to the previous statement.

lib/grandpa/grandpa.go
```diff
@@ -38,178 +50,192 @@ func NewService(blockState BlockState, voters []*Voter) (*Service, error) {
...
+	blocks := make(map[common.Hash]uint64)
+
+	// check if any of them have >=2/3 votes
+	for v := range votes {
+		total, err := s.getTotalVotesForBlock(v.hash)
+		if err != nil {
+			return nil, err
+		}
+
+		if total >= uint64(2*len(s.state.voters)/3) {
```
thomasmodeneis, line 109: the condition could be more clear if the `2*len(s.state.voters)/3` would be extracted to a var that would represent the operation in a more meaningful way.

dot/state/block.go
```diff
@@ -526,6 +526,21 @@ func (bs *BlockState) IsDescendantOf(parent, child common.Hash) (bool, error) {
...
+	return bs.bt.HighestCommonAncestor(a, b)
+}
+
+// Leaves returns the leaves of the blocktree as an array
+func (bs *BlockState) Leaves() []common.Hash {
+	return bs.bt.Leaves()
+}
+
+// BlocktreeAsString returns the blocktree as a string
+func (bs *BlockState) BlocktreeAsString() string {
```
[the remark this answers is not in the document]
noot: the receiver is actually BlockState and I don't want to BlockState as a string, I want the blocktree

lib/grandpa/grandpa.go
noot: what do you mean?

lib/grandpa/grandpa.go
noot: the function comment describes it, but I can move it to a var

lib/grandpa/grandpa.go
noot: oh, I guess you mean 
```
if err != nil { 
} else if pred == curr {
}
```

I prefer not chaining `else`s onto error checks, I don't usually see that happening in go code

noot: https://golang.org/doc/effective_go.html#if

looks like if an `if` statement has a return in it, it's better not to chain the else

lib/grandpa/grandpa.go
noot: same as above

lib/grandpa/types.go
noot: same as above

lib/grandpa/vote_message.go
noot: same as above

lib/grandpa/vote_message.go
noot: same as above

lib/grandpa/vote_message.go
noot: same as above

lib/grandpa/grandpa.go
noot: changed to `s.state.threshold()`

lib/grandpa/grandpa.go
thomasmodeneis: The effective go is saying about another scenario. In this case, a else if is recommended since the var was initialized on the if condition itself, not in the next line like their example:
```
f, err := os.Open(name)
if err != nil {
    return err
}
d, err := f.Stat()
if err != nil {
    f.Close()
    return err
}
codeUsing(f, d)
```

dot/state/block.go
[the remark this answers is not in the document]
thomasmodeneis: This method could be not existing then, the only thing you need to do is add a GetBlockState and then use that to call blockstate and run String() in it. That would actually make more sense than having a wrapper for 3 funcs that are belonging to another package.

lib/grandpa/grandpa.go
noot: it also says right above that code:

> Since error cases tend to end in return statements, the resulting code needs no else statements

I think this is pretty explicit in that there shouldn't be an else statement.

dot/state/block.go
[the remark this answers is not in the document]
noot: you mean GetBlockTree()? it's preferred to not have direct access to the blocktree except through the state service

lib/grandpa/types.go
```diff
@@ -40,6 +43,42 @@ type State struct {
...
+		key: pk,
+		id:  id,
+	}, nil
+}
+
+// threshold returns the 2/3 |voters| threshold value
+func (s *State) threshold() uint64 {
+	return uint64(2 * len(s.voters) / 3)
+}
+
```
edwardmack, line 81: Should this be truncated?  or rounded?  In the case where s.voters = 6, the result is 4.  And when there are 7 voters the result is 4, but 4 is not greater than 2/3 * 7 which is 4.666, should this instead be rounded to 5 which is greater than 2/3 * 7?  Is this addressed in the spec?

noot: yeah this is a good question, I left it rounded down for now since in practice the validator set should be ~100 so it wouldn't make a huge difference. I can ask and clarify though

lib/grandpa/grandpa.go
```diff
@@ -38,178 +50,192 @@ func NewService(blockState BlockState, voters []*Voter) (*Service, error) {
...
+// getVotesForBlock returns the number of observed votes for a block B.
+// The set of all observed votes by v in the sub-round stage of round r for block B is
+// equal to all of the observed direct votes cast for block B and all of the B's descendants
+func (s *Service) getVotesForBlock(hash common.Hash) (uint64, error) {
+	votes := s.getDirectVotes()
 
-	if !has {
-		return ErrBlockDoesNotExist
-	}
+	// B will be counted as in it's own subchain, so don't need to start with B's vote count
```
ryanchristo, line 197: typo/grammar?

lib/grandpa/types.go
ryanchristo: Would be nice to add a comment/note here in the code or a todo to make sure we loop back.

lib/grandpa/types.go
```diff
@@ -62,6 +101,29 @@ func NewVoteFromHeader(h *types.Header) *Vote {
 	}
 }
 
+// NewVoteFromHash returns a new Vote given a hash and a blockState
```
ryanchristo, line 104: `...returns a new Vote given a hash and block header`

noot: it accepts a blockState?

lib/grandpa/types.go
noot: I've added a todo for now

PR #880, opened 2020-05-20
dot/rpc, dot/network: Implement RPC system_nodeRoles 

edwardmack:
## Changes

- Implement RPC function for system_nodeRoles see: https://github.com/w3f/PSPs/blob/psp-rpc-api/psp-002.md#system_nodeRoles
- Added function NodeRoles to network service

## Tests:

```
go test ./dot/rpc/...
go test ./dot/network/...
```

## Checklist:

## Issues:

- related to #731

### DIFF
--- a/dot/network/service.go
+++ b/dot/network/service.go
@@ -410,3 +410,8 @@ func (s *Service) Peers() []common.PeerInfo {
 	}
 	return peers
 }
+
+// NodeRoles Returns the roles the node is running as.
+func (s *Service) NodeRoles() byte {
+	return s.cfg.Roles
+}
--- a/dot/network/service_test.go
+++ b/dot/network/service_test.go
@@ -17,6 +17,7 @@
 package network
 
 import (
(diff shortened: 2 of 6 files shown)


dot/network/service.go
```diff
@@ -410,3 +410,8 @@ func (s *Service) Peers() []common.PeerInfo {
 	}
 	return peers
 }
+
+// NodeRoles Returns the roles the node is running as.
```
ryanchristo, line 414: I think this should be part of the core service rather than the network service? maybe? we pass this configuration value down from the dot configuration to the core and network services and the network service is less the source of truth for this sort of thing than core but probably ok as is - kinda just thinking this out loud.

edwardmack: I'd like to leave this as is for now.  I agree it makes sense to include this in core, however the RPC module that's using this doesn't have reference to coreAPI, it currently has access to networkAPI and systemAPI.

PR #939, opened 2020-06-07
Introduce framework for RPC testing nodes

edwardmack:
## Changes

- Added frmework to test package to handle starting nodes and RPC calls
- Store results in database for further analysis
- Method to check database records for expected results.

## Tests

Utilizes the TestMain function so that a reference to framework is kept between tests, this contains the db holding (JSON) results from calling RPC methods with CallRPC function.  Each RPC call is indexed in the database (starting an 0).  The call index can be used to query the database to check if the given JSON fields for two calls are equal using CheckEqual function.

See tests/sync/sync_test.gq TestCalls for example of call nodes and checking results.
```
GOSSAMER_INTEGRATION_TEST_MODE=sync go test ./tests/sync/... -v
```

## Checklist

## Issues

-


edwardmack:
@noot Yes, there is some refactoring to do to combine some of this functionality.  Currently I was adding tests to the current tests, and some are an overlap.  I plan on removing/consolidating these tests as I continue to build the framework.

PR #940, opened 2020-06-07
lib/trie: implement GetKeysWithPrefix

### LINKED ISSUE #837: implement getting all keys with prefix in trie
given a prefix, eg. given `0xabcd`, return all keys in the trie with that prefix ie if keys`0xabcd1`, `0xabcd2`, `0xabcd09876` etc. exist, return them

this is needed for rpc call state_getPairs

noot:
## Changes

- implement GetKeysWithPrefix which returns all keys in the trie with a given prefix

## Tests

```
go test ./lib/trie
```

## Checklist

## Issues

- closes #837


noot:
@ryanchristo I'm not assigning the nibbles array to 0, I'm slicing off the last nibble if it's zero:
`p = p[:len(p)-1]`

`getKeysWithPrefix` is recursive, so it takes the parent node, the current prefix up to that node, the key that it's still searching for, and then the keys that have the prefix that we're searching for

 I'm not passing empty values to `getKeysWithPrefix`, I'm passing the trie root (parent node to start with), the prefix for the root node (empty), the key we're searching for (`p`), and then an array of keys that have that key in them (empty)

there's nothing missing, it's just recursive so it looks funny. I can try to draw it out to explain it. you can also look at this article which explains the trie structure more: https://medium.com/chainsafe-systems/gossamer-update-1-merkle-patricia-trie-6320588efedd

PR #942, opened 2020-06-08
dot/core, lib/babe: refactor BABE to be a service that implements BlockProducer interface

### LINKED ISSUE #900: refactor: create BlockProducer interface for use with core
## Expected Behavior

- right now, BABE is controlled by the core service. swapping out the block production mechanism would require changes to core, which is not ideal.
- it would be nice to create a `BlockProducer` interface or something similar, which different block production libraries would be able to implement
- then, core will interact only with this interface
- alternatively, we could run block production as its own service instead of being controlled by core
- in this case, core will know about the `BlockProducer` interface only, similar to how it only interacts with state through the `BlockState` etc. interfaces

## Current Behavior

- babe directly controlled by core

## Possible Solution

- see expected behaviour, will likely need some consideration as to what is best / most modular / easiest to swap out

## Checklist

- [ ] I have read [CONTRIBUTING](CONTRIBUTING.md) and [CODE_OF_CONDUCT](CODE_OF_CONDUCT.md)

noot:
## Changes

- refactor BABE to implement Service interface, no longer needs to be controlled by core. runs continuously even when epoch ends.
- remove BABE control from core
- add BlockProducer interface that core uses to interact with BABE
- start BABE service along with other services

## Tests

```
go test ./... -short
```

## Checklist

## Issues

- closes #900


lib/babe/babe.go
```diff
@@ -176,64 +159,84 @@ func (b *Session) Start() error {
...
+func (b *Service) SetRuntime(rt *runtime.Runtime) error {
+	b.rt = rt
+
+	var err error
+	b.config, err = b.rt.BabeConfiguration()
+	if err != nil {
+		return err
 	}
 
 	return nil
```
edwardmack, line 201: Maybe just `return err` here

Issue #979, opened 2020-07-01
add config options to turn on/off babe, grandpa


noot:
## Expected Behavior

- currently there is only 1 option to run both
- would be nice for testing purposes to have an option to specify on/off for each separately
- also extra developer settings like setting BABE threshold manually

## Checklist

- [x] I have read [CODE_OF_CONDUCT](https://github.com/ChainSafe/gossamer/blob/development/.github/CODE_OF_CONDUCT.md) and [CONTRIBUTING](https://github.com/ChainSafe/gossamer/blob/development/.github/CONTRIBUTING.md) 
- [x] I have provided as much information as possible and necessary
- [x] I am planning to submit a pull request to fix this issue myself

PR #980, opened 2020-07-02
cmd, dot, dot/core: add BabeAuthority and GrandpaAuthority config options

noot:
## Changes

- add BabeAuthority and GrandpaAuthority config options so that the node can run only one or the other if specified
- update default package log settings in config so that they match the `log` option instead of defaulting to `info`

## Tests

```
go test ./... -short
```

also running gossamer with `authority = true` and `babe-authority = false` in the config to turn off babe, or `grandpa-authority = false` to turn off grandpa

## Checklist

## Issues

- clsoes #979

### DIFF
--- a/chain/gssmr/config.toml
+++ b/chain/gssmr/config.toml
@@ -5,13 +5,13 @@ basepath = "~/.gossamer/gssmr"
 log = "info"
 
 [log]
-core = "info"
-network = "info"
-rpc = "info"
-state = "info"
-runtime = "info"
-babe = "info"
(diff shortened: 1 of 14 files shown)


noot:
@edwardmack yeah this is mostly for testing, I don't see it being used by an actual node. I think we can keep the `authority` flag since that will be either true or false for BABE+grandpa. not sure about roles, it seems like a duplicate at this point. the BABE and grandpa flags are meant to overwrite `authority=true` to turn off a service if configured to

Issue #854, opened 2020-05-14
Improve Syncing Integration Tests


amerameen:
## Expected Behavior

Create more syncing tests cases. (E.g. More time, turning nodes off and on, a variety of different types of nodes (not producers), more nodes).

- [x] Adding block producers to existing tests
- [x] Measure syncing time/block (i.e. how long does it take a new node to sync up with the network)
- [x] Verify that syncing works when Turning Off/On node
- [x] Investigate Test for Network Size feasibility (i.e. how many nodes can one block producer support?)

## Current Behavior

-

## Possible Solution

-

## Steps to Reproduce

1.
2.
3.
4.

## Specification

- go version:
- gossamer version:
- gossamer commit tag:
- gossamer commit hash:
- operating system:
- additional links:

## Checklist

- [x] I have read [CONTRIBUTING](CONTRIBUTING.md) and [CODE_OF_CONDUCT](CODE_OF_CONDUCT.md)
- [x] I have provided as much information as possible and necessary
- [ ] I am planning to submit a pull request to fix this issue myself

PR #988, opened 2020-07-08
cmd, dot: add BabeThreshold config option

noot:
## Changes

- add `BabeThreshold` configuration value, can be "max", "min", or nil
- set `BabeThreshold` when starting BABE service

## Tests

```
go test ./...
```

also, can run gossamer with `BabeThreshold` in config set to "max", should not allow blocks to be built

## Checklist

## Issues

- related to #854

### DIFF
--- a/dot/config.go
+++ b/dot/config.go
@@ -86,10 +86,11 @@ type NetworkConfig struct {
 
 // CoreConfig is to marshal/unmarshal toml core config vars
 type CoreConfig struct {
-	Authority        bool `toml:"authority"`
-	Roles            byte `toml:"roles"`
-	BabeAuthority    bool `toml:"babe-authority"`
-	GrandpaAuthority bool `toml:"grandpa-authority"`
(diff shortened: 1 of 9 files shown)


dot/config.go
```diff
@@ -139,6 +140,7 @@ func GssmrConfig() *Config {
 			Roles:            gssmr.DefaultRoles,
 			BabeAuthority:    gssmr.DefaultBabeAuthority,
 			GrandpaAuthority: gssmr.DefaultGrandpaAuthority,
+			//BabeThreshold: "",
```
edwardmack, line 143: remove?

dot/config.go
```diff
@@ -178,6 +180,7 @@ func KsmccConfig() *Config {
 		Core: CoreConfig{
 			Authority: ksmcc.DefaultAuthority,
 			Roles:     ksmcc.DefaultRoles,
+			//BabeThreshold: "",
```
edwardmack, line 183: remove?

Issue #990, opened 2020-07-08
improve syncing protocol


noot:
## Expected Behavior

- currently, the syncing protocol is not very efficient. a BlockResponse contains only up to 8 blocks (due to some network size limitations, we can try to increase this)
- additionally BlockRequests should begin from the chain head+1 of the requestor, currently the syncer starts requests from the block number of the highest block it's seen, which is not necessarily the chain head
- there are likely other optimizations that can be made, need to investigate syncer thoroughly
- may also be helpful to investigate syncing protocols of substrate and other blockchains in general

TODO: define metrics we want to achieve (eg. blocks/second)

## Checklist

- [x] I have read [CODE_OF_CONDUCT](https://github.com/ChainSafe/gossamer/blob/development/.github/CODE_OF_CONDUCT.md) and [CONTRIBUTING](https://github.com/ChainSafe/gossamer/blob/development/.github/CONTRIBUTING.md) 
- [x] I have provided as much information as possible and necessary
- [x] I am planning to submit a pull request to fix this issue myself

PR #1007, opened 2020-07-16
dot/core, tests: improve benchmarker, increase maxResponseSize

noot:
## Changes

- fix benchmarker so that it doesn't overwrite the current sync when a new message arrives
- increase maxResponseSize to 12 (this seems to be the max size the networrk can handle for now)
- add integration test w/ one block producer, one syncing node, on my machine the sync speed is around 36.5 blocks/second

## Tests

```
MODE=stress go test ./tests/stress/ -run TestSync_SingleSync -v
```

## Checklist

## Issues

- related to #990

### DIFF
--- a/dot/core/syncer_benchmark.go
+++ b/dot/core/syncer_benchmark.go
@@ -1,0 +1,43 @@
+package core
+
+import (
+	"time"
+
+	log "github.com/ChainSafe/log15"
+)
+
+type benchmarker struct {
+	logger          log.Logger
+	start           time.Time
+	startBlock      uint64
+	blocksPerSecond []float64
+}
+
+func newBenchmarker(logger log.Logger) *benchmarker {
+	return &benchmarker{
+		logger:          logger.New("module", "benchmarker"),
+		blocksPerSecond: []float64{},
+	}
+}
+
+func (b *benchmarker) begin(block uint64) {
+	b.start = time.Now()
+	b.startBlock = block
+}
+
+func (b *benchmarker) end(block uint64) {
+	duration := time.Since(b.start)
+	blocks := block - b.startBlock
+	bps := float64(blocks) / duration.Seconds()
+	b.blocksPerSecond = append(b.blocksPerSecond, bps)
+
+	b.logger.Info("added sync time", "avg blocks/second", b.average())
+}
+
+func (b *benchmarker) average() float64 {
+	sum := float64(0)
+	for _, bps := range b.blocksPerSecond {
+		sum += bps
+	}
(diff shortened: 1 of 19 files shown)


tests/stress/stress_test.go
```diff
@@ -122,3 +122,38 @@ func TestSync_SingleBlockProducer(t *testing.T) {
...
+	defer os.Remove(tmpdir)
+	alice, err := utils.RunGossamer(t, 0, tmpdir, utils.GenesisDefault, utils.ConfigBABEMaxThreshold)
+	require.NoError(t, err)
+
+	time.Sleep(time.Second * 15)
+
+	// start syncing node
+	tmpdir, err = ioutil.TempDir("", "gossamer-stress-bob")
+	require.NoError(t, err)
+	defer os.Remove(tmpdir)
```
thomasmodeneis, line 142: you have os.Remove twice on defer for the tmpdir, perhaps you should then rename the vars so the defer will actually work properly. Otherwise it will defer remove the same dir twice ?

noot: if you run the test you can see that both directories get removed from `/tmp/`.

the `defer` captures the value of the args when it is reached and pushes the function call onto the stack, so the reassignment to `tmpdir` doesn't affect the first `os.Remove()`

dot/core/syncer_benchmark.go
```diff
@@ -21,17 +22,28 @@ func newBenchmarker(logger log.Logger) *benchmarker {
...
 func (b *benchmarker) end(block uint64) {
+	if !b.syncing {
+		return
+	}
+
 	duration := time.Since(b.start)
 	blocks := block - b.startBlock
+	if blocks == 0 {
+		blocks = 1
+	}
```
edwardmack, line 41: I don't understand this, it seem if no blocks were produced, than set it to one block produced?  Why?

noot: I noticed that when only one block is being synced, it starts and ends at the same number, causing blocks/second to become 0. alternatively I could change the syncer to call end with `endblock+1`, if that is preferred

edwardmack: Ok, that makes sense.  No need to change it.

noot:
my CPU is Intel i7-8650U @ 1.900GHz x 8, will be interesting to see how it works with more nodes!

tests/stress/stress_test.go
thomasmodeneis: oh, I never tested this before. Good to know, Thanks :)

PR #1010, opened 2020-07-19
lib/genesis Implement human-readable genesis file parser

### LINKED ISSUE #925: implement genesis utils that allow for readable genesis file
## Expected Behavior

- there should be a genesis file that is in human readable format, which can then be converted in the `raw` format that is actually used by the node using some command (or the node is able to use either)
- we will need some genesis utils that turn a human-readable genesis into a raw one
- I believe substrate has something like this already, see https://substrate.dev/docs/en/development/deployment/chain-spec#raw-chain-specs

## Current Behavior

- only have raw genesis

## Checklist

edwardmack:
## Changes

- Created command `build-spec` that outputs genesis JSON with human-readable keys or in raw format which is used to initialize node databases. 

Usage: 
- Output genesis JSON with human-readable values for the database in the node's default base dir:
```
./bin/gossamer build-spec > my_genesis.json
```
- Add `--raw` and the output will be in raw database format.  Use this to create a file that can be used to initialize nodes:
```
./bin/gossamer build-spec --raw > my_genesis-raw.json
```
- The  `--genesis` flag will read from given human-readable file name, so adding the`--raw` flag with that will allows you to create raw genesis files from files you've edited.
```
./bin/gossamer build-spec --raw --genesis my_genesis.json > my_genesis-raw.json 
```
- Then use this raw file to initialize the node.
```
./bin/gossamer init --genesis-raw my_genesis-raw.json --force
```
(Note the flag for init is now called `--genesis-raw` since it's working with genesis raw formatted files.)

- The format for human-readable genesis files is based on format that is used by substrate `build-spec` command.  There is an example human-readable genesis file located at `chain/gssmr/genesis.json`, this file will generate a raw file that replicates `chain/gssmr/genesis-raw.json`

Notes:
- The current raw format has different ways encoding storage keys:
  - raw bytes:  For example the key for `:code` is hex of ascii for code: `0x3a636f6465`
  - twoxHash:  For example the key for `Babe Authorities` is the twoxHash of Babe Authorities: `0x886726f904d8372fdabb7707870c2fad` 
  - The human-readable parser treats the default key encoding with twoxHash.  The parser does key encoding in `formatKey` in `lib/genesis/helpers.go`, so additional key formatting can be added there.
- The current raw format also handles stored values differently depending on type, for example:
  - grandpa authorities value is perpend with byte 1, followed by scale encoding of array of tuples 
  - system code value is not encoded, just raw value is used.
  - default values are scale encoded.
  - These conditions are handled in `formatValue` in `lib/genesis/helpers.go` so additional condition can be addressed there.
 

## Tests

```
go test ./lib/genesis/... -v
go test ./dot -run TestNewTestGenesisFile -v
```
Interact with node, as described above.

## Checklist

## Issues

- closes #925


edwardmack:
It handles human-readable genesis files!!  There are some todo's in here regarding error handling that I need to address, happy to discuss ideas for handling these (just log errors, or pass up the stack, or ignore). Looking forward to your reviews.

dot/node.go
```diff
@@ -64,8 +64,14 @@ func InitNode(cfg *Config) error {
...
 	// create genesis from configuration file
-	gen, err := genesis.NewGenesisFromJSON(cfg.Init.Genesis)
-	if err != nil {
+	var gen *genesis.Genesis
+	var lErr error
+	if cfg.Init.GenesisHR {
+		gen, lErr = genesis.NewGenesisFromJSONHR(cfg.Init.Genesis)
+	} else {
+		gen, lErr = genesis.NewGenesisFromJSON(cfg.Init.Genesis)
+	}
```
noot, line 73: not sure how I feel about this, you should only be able to use one of `--genesis` or `--genesis_hr`, not both, could you add some check that makes sure only one is set?

cmd/gossamer/flags.go
```diff
@@ -82,6 +82,11 @@ var (
 		Name:  "genesis",
 		Usage: "Path to genesis JSON file",
 	}
+	// GenesisHRFlag tells parser genesis.json contains human readable field, and to convert to raw
+	GenesisHRFlag = cli.BoolFlag{
+		Name:  "genesis_hr",
```
noot, line 87: maybe we should rename `--genesis` to `--genesis-raw` and this to `--genesis`, what do you think? I think that's clearer for people to use

dot/utils_test.go
```diff
@@ -43,3 +46,43 @@ func TestNewConfigAndFile(t *testing.T) {
...
+	require.NotNil(t, genRawFile)
+	defer os.Remove(genRawFile.Name())
+
+	genHR, err := genesis.NewGenesisFromJSONHR(genHRFile.Name())
+	require.NoError(t, err)
+	genRaw, err := genesis.NewGenesisFromJSON(genRawFile.Name())
+	require.NoError(t, err)
+
+	// values from raw genesis file should equal values generated from human readable genesis file
+	//  Note, we don't just compare Genesis.Raw because there are entries (balances) in raw genesis that are not yet handled by
```
noot, line 82: feel free to remove the balances from `genesis.json` since the keys have changed since the runtime was updated

lib/genesis/helpers.go
```diff
@@ -82,3 +214,28 @@ func NewGenesisBlockFromTrie(t *trie.Trie) (*types.Header, error) {
 
 	return header, nil
 }
+
+func twoxHash(msg []byte) []byte {
```
noot, line 218: since this is used here as well as in the runtime `imports_old.go`, could you move this to `common` and update it here and in the runtime imports?

lib/trie/trie_test.go
```diff
@@ -209,7 +209,7 @@ func TestLoadTrie(t *testing.T) {
 		if err != nil {
 			t.Fatal(err)
 		}
-		valueBytes, err = common.HexToBytes(value)
+		valueBytes, err = common.HexToBytes(fmt.Sprintf("%v", value))
```
noot, line 212: ?

cmd/gossamer/flags.go
edwardmack: Regarding naming, I agree it's not that clear.  The `--genesis_hr` flag is a boolean that tells the parser the genesis file is human-readable.  But since `--genesis` is a flag for path it strange.  So, I agree I should change the current `--genesis` flag `--genesis-raw` and change `--genesis_hr` into `--genesis` and make it so it's a path to human-readable path.  Then change flags to check that only one is set.

dot/node.go
edwardmack: updated flag names and check for duplicates.

lib/genesis/helpers.go
edwardmack: Yes.

lib/trie/trie_test.go
edwardmack: changed back, just left over from experimenting.

edwardmack:
@noot I've updated this, should be ready for another review.

cmd/gossamer/config.go
```diff
@@ -228,16 +236,27 @@ func setLogConfig(ctx *cli.Context, globalCfg *dot.GlobalConfig, logCfg *dot.Log
...
 	if genesis := ctx.String(GenesisFlag.Name); genesis != "" {
 		cfg.Genesis = genesis
+		cfg.GenesisRaw = ""
 	}
-
 	logger.Debug(
 		"init configuration",
+		"genesis-raw", cfg.GenesisRaw,
 		"genesis", cfg.Genesis,
 	)
```
noot, line 258: move this log to inside the `if` statements so it only prints the relevant field?

edwardmack:
@noot I'll dig into these errors...  Do you think it's reasonable to make INIT only accept genesis-raw.json files (to avoid file type errors, etc.), then use the spec-build command I'm building to build genesis-raw files that INIT can use?  That way INIT only deals with one file type, and spec-build just deals with building/parsing these files?  Similar to substrate build-spec tool?

noot:
@edwardmack that would work also! maybe you can change the flags back here ( sorry :( ) and do the subcommand in another PR?

lib/genesis/helpers.go
```diff
@@ -82,3 +85,199 @@ func NewGenesisBlockFromTrie(t *trie.Trie) (*types.Header, error) {
...
+func addBabeAuthoritiesValues(value []byte, gen *Genesis) {
+	if gen.Genesis.Runtime["babe"] == nil {
+		gen.Genesis.Runtime["babe"] = make(map[string]interface{})
+	}
+
+	buf := &bytes.Buffer{}
+	sd := scale.Decoder{Reader: buf}
+	_, err := buf.Write(value)
+	if err != nil {
+		fmt.Printf("errer %v\n", err)
```
devtomnl: Minor typo here, should be error instead.

lib/genesis/helpers_test.go
```diff
@@ -66,3 +68,49 @@ func TestNewGenesisFromJSON(t *testing.T) {
...
+	expRaw[0]["0x3a6772616e6470615f617574686f726974696573"] = "0x010834602b88f60513f1c805d87ef52896934baf6a662bc37414dbdbf69356b1a6910000000000000000" // raw grandpa authorities
+	expRaw[0]["0x886726f904d8372fdabb7707870c2fad"] = "0x08d43593c715fdd31c61141abd04a99fd6822c8558854ccde39a5684e7a56da27d0100000000000000"           // raw babe authorities
+	expectedGenesis.Genesis = Fields{
+		Raw: expRaw,
+	}
+
+	// Create temp file
+	file, err := ioutil.TempFile("", "genesis_hr-test")
+	if err != nil {
+		t.Fatal(err)
```
devtomnl, line 87: For consistency, would be good to use require.Nil here instead.

lib/genesis/helpers.go
```diff
@@ -82,3 +85,199 @@ func NewGenesisBlockFromTrie(t *trie.Trie) (*types.Header, error) {
...
+			}
+			return fmt.Sprintf("0x%x%v", lenEnc, kv.value), nil
+		}
+		return fmt.Sprintf("0x%x", kv.value), nil
+	}
+}
+
+func BuildFromMap(m map[string][]byte, gen *Genesis) {
+	for k, v := range m {
+		fmt.Printf("key %x vL %v\n", k, len(v))
```
devtomnl: Not sure why you are doing fmt in this file instead of log? On the other hand It seems like this is not yet finished? So, just adding it as reminder for you to not forget to cleanup. Same for other fmt I've seen on this file 👍

dot/utils_test.go
```diff
@@ -43,3 +46,38 @@ func TestNewConfigAndFile(t *testing.T) {
...
+	cfg := NewTestConfig(t)
+	require.NotNil(t, cfg)
+
+	genFile := NewTestGenesisRawFile(t, cfg)
+	require.NotNil(t, genFile)
+
+	defer utils.RemoveTestDir(t)
+
+	cfg.Init.GenesisRaw = genFile.Name()
+	fmt.Printf("FileName %v\n", genFile.Name())
```
devtomnl, line 61: t.log instead?

dot/build_spec.go
```diff
@@ -0,0 +1,121 @@
...
+func BuildFromGenesis(path string) (*BuildSpec, error) {
+	gen, err := genesis.NewGenesisFromJSON(path)
+	if err != nil {
+		return nil, err
+	}
+	bs := &BuildSpec{
+		genesis: gen,
+	}
+	return bs, nil
+}
```
devtomnl, line 71: Missing comments on exported method, I've seen others on same file with similar issue.

edwardmack:
@noot This is ready for a re-review.  It's a little chunky, sorry about that.  Not sure why it's failing travis, maybe related to the gossip test?

cmd/gossamer/config.go
```diff
@@ -233,14 +245,22 @@ func setLogConfig(ctx *cli.Context, globalCfg *dot.GlobalConfig, logCfg *dot.Log
...
 // setDotInitConfig sets dot.InitConfig using flag values from the cli context
 func setDotInitConfig(ctx *cli.Context, cfg *dot.InitConfig) {
-	// check --genesis flag and update init configuration
-	if genesis := ctx.String(GenesisFlag.Name); genesis != "" {
-		cfg.Genesis = genesis
-	}
-
+	//if ctx.String(GenesisRawFlag.Name) != "" && ctx.String(GenesisFlag.Name) != "" {
+	//	return fmt.Errorf("error can't have both genesis and genesis-raw set")
+	//}
```
noot, line 250: remove

cmd/gossamer/config.go
```diff
@@ -233,14 +245,22 @@ func setLogConfig(ctx *cli.Context, globalCfg *dot.GlobalConfig, logCfg *dot.Log
...
+	// check --genesis-raw flag and update init configuration
+	if genesis := ctx.String(GenesisRawFlag.Name); genesis != "" {
+		cfg.GenesisRaw = genesis
+	}
+
+	//// check --genesis flag
+	//if genesis := ctx.String(GenesisFlag.Name); genesis != "" {
+	//	cfg.Genesis = genesis
+	//	cfg.GenesisRaw = ""
+	//}
```
noot, line 260: remove

cmd/gossamer/flags.go
```diff
@@ -77,10 +77,22 @@ var (
...
+		Usage: "Path to raw genesis JSON file",
+	}
+)
+
+// BuildSpec-only flags
+var (
+	RawFlag = cli.BoolFlag{
+		Name:  "raw",
+		Usage: "Output as raw genesis JSON",
+	}
```
noot, line 92: what is this used for? is there another option that build-spec can output?

cmd/gossamer/main.go
```diff
@@ -224,3 +237,55 @@ func initAction(ctx *cli.Context) error {
...
+	res := []byte{}  //nolint
+	if ctx.Bool(RawFlag.Name) {
+		res, err = bs.ToJSONRaw()
+	} else {
+		res, err = bs.ToJSON()
+	}
+	if err != nil {
+		return err
+	}
+	fmt.Printf("%s", res)
```
noot, line 288: remove

cmd/gossamer/main.go
```diff
@@ -224,3 +237,55 @@ func initAction(ctx *cli.Context) error {
...
+		}
+		bs = bspec
+	}
+
+	if bs == nil {
+		return fmt.Errorf("error building genesis")
+	}
+
+	res := []byte{}  //nolint
+	if ctx.Bool(RawFlag.Name) {
```
noot, line 280: not sure i see the point of the raw flag, it should be turns it into raw every time? otherwise it's just copying the same file? maybe i'm missing something

edwardmack: when you run it with no flags it output a human-readable JSON based on the genesis that's currently in the db created from the init command.  So, this is useful for getting an human-readable template for editing later.  Then use the `--raw` with `--genesis` to generate a raw output based on the genesis you passed in.

dot/build_spec.go
```diff
@@ -0,0 +1,121 @@
...
+		return nil, err
+	}
+	gData, err := scale.Decode(gd, &genesis.Data{})
+	if err != nil {
+		return nil, err
+	}
+	tmpGen.Name = gData.(*genesis.Data).Name
+	tmpGen.ID = gData.(*genesis.Data).ID
+	// todo figure out how to assign bootnodes
+	//tmpGen.Bootnodes = gData.(*genesis.Data).Bootnodes
```
noot, line 114: remove? or make issue?

lib/genesis/genesis_test.go
```diff
@@ -1,32 +0,0 @@
-// Copyright 2019 ChainSafe Systems (ON) Corp.
```
noot, line 1: where did this go?

lib/common/common.go
```diff
@@ -194,3 +196,29 @@ func Read32Bytes(r io.Reader) ([32]byte, error) {
...
 	return h, nil
 }
+
+// TwoxHash128 computes xxHash64 twice with seeds 0 and 1 applied on given byte array
+func TwoxHash128(msg []byte) []byte {
+	// compute xxHash64 twice with seeds 0 and 1 applied on given byte array
+	h0 := xxhash.NewS64(0) // create xxHash with 0 seed
+	_, err := h0.Write(msg[0:])
+	if err != nil {
+		return nil
```
noot, line 206: probably return err?

lib/common/common.go
```diff
@@ -194,3 +196,29 @@ func Read32Bytes(r io.Reader) ([32]byte, error) {
 	copy(h[:], buf)
 	return h, nil
 }
+
+// TwoxHash128 computes xxHash64 twice with seeds 0 and 1 applied on given byte array
+func TwoxHash128(msg []byte) []byte {
```
noot, line 201: rename `Twox128Hash`?

cmd/gossamer/main.go
noot: i'm slightly confused, what db is it reading from?  it should be reading from the default genesis file if no genesis is provided

noot: what happens if there is no db? it's likely that the user will be doing `build-spec` before `init` since `init` requires the output of `build-spec`

cmd/gossamer/main.go
edwardmack: This is needed because currently build-spec just dumps the output, so than we can redirect with `>`.  I'll change this when an `--output` flag is added.

cmd/gossamer/flags.go
edwardmack: see below.

cmd/gossamer/main.go
edwardmack: Added issue #1029

dot/build_spec.go
edwardmack: added issue #1030

edwardmack:
@noot I've update this to address your comments.  I think it's in good shape.  The coverage for build_spec is 83.9%, I don't know where else to add testing.

cmd/gossamer/main.go
```diff
@@ -267,7 +267,7 @@ func buildSpecAction(ctx *cli.Context) error {
 
 		bspec, e := dot.BuildFromDB(cfg.Global.BasePath)
 		if e != nil {
-			return e
+			return fmt.Errorf("Error building spec from database, INIT must be run before build-spec or run build-spec with --genesis flag.\n Error %s\n", e)
```
noot, line 270: nitpick but can you change `INIT` to `gossamer init`?

noot: you can also wrap errors, so instead of `%s` you can do %w`

PR #1019, opened 2020-07-22
lib/scale: don't panic when decoding custom array type, support fixed length arrays

### LINKED ISSUE #947: scale: error gracefully instead of panicking if unable to decode custom type
## Expected Behavior

- don't panic, return error ideally with name of type that could not be decoded

## Current Behavior

- if scale attempts to decode a custom type that doesn't have `Decode()` implemented on it, it will often panic at some later stage of decoding, after it passes through some reflections. 
- we should try to catch this at an earlier stage (ie. in the `Decode` function) and return error the type isn't supported and doesn't have `Decode()` implemented
- this was discovered using a custom type that's an alias for a fixed-length array, eg `type myBytes [32]byte`, may not be relevant for other types

## Checklist

noot:
## Changes

- rename `DecodeArray` to `DecodeSlice` since that's actually what it does
- implement new `DecodeAray` that handles arrays of bytes or custom types
- add `TestEncodeDecodeCustom_Array` to show there are no panics when using a fixed length byte array custom type

## Tests

```
go test ./... -short
```

## Checklist

## Issues

- closes #947


lib/scale/enc_dec_test.go
```diff
@@ -168,3 +168,24 @@ func TestEncodeDecodeCustom_InsideStruct(t *testing.T) {
...
 }
+
+type myBytes [64]byte
+
+func TestEncodeDecodeCustom_Array(t *testing.T) {
+	b := myBytes([64]byte{1, 2, 3, 4})
+	enc, err := Encode(b)
+	require.NoError(t, err)
+
+	_, _ = Decode(enc, new(myBytes))
```
devtomnl, line 179: It could be useful to add a require.Nil on err instead of ignoring it. Or I'm missing something?

noot: this isn't supported, it is supposed to error. the test is making sure it doesn't panic as per the linked issue #947

Issue #1014, opened 2020-07-21
check for runtime changes when processing new blocks


noot:
## Expected Behavior

- the runtime may change throughout the course of running the chain
- when we process a block coming from a BlockResponse or a block that we produced, we should check for changes to the runtime
- if the block is on our current chain, we should update the runtime accordingly

## Current Behavior

- this is implemented for blocks we produce: `core.checkForRuntimeChanges`
- however this is not implemented for `BlockResponseMessage`

## Possible Solution

- in `sync.HandleBlockResponse` we should call `core.HandleRuntimeChanges` (expose func through an interface or move somewhere else?) after verifying a block
- if there is a runtime change and the block is in the current chain (ie. the current chain head or an ancestor of the current chain head) we should update the runtime accordingly and all blocks that come after should be processed with that runtime (ie. call `VerificationManager.SetRuntimeChangeAtBlock()`

## Checklist

- [x] I have read [CODE_OF_CONDUCT](https://github.com/ChainSafe/gossamer/blob/development/.github/CODE_OF_CONDUCT.md) and [CONTRIBUTING](https://github.com/ChainSafe/gossamer/blob/development/.github/CONTRIBUTING.md) 
- [x] I have provided as much information as possible and necessary
- [x] I am planning to submit a pull request to fix this issue myself

PR #1036, opened 2020-07-27
dot/core, lib/babe: refactor VerificationManager to track changes by block, update verifier with runtime changes

### LINKED ISSUE #762: add check for epoch length changes in checkForRuntimeChanges
currently, the `getBlockEpoch` function relies on epoch length being fixed. however, there is a chance it can change. a check for changes in epoch length should be added to checkForRuntimeChanges (by calling BabeApi_configuration). <s>also the historical data for epoch lengths should be stored somewhere accessible by `getBlockEpoch` so it can correctly calculate block epoch</s> the historical data for epoch lengths should be stored in the EpochManager

also, we should be loading epoch length from the runtime at genesis

### LINKED ISSUE #1023: block verifier should keep track of randomness, threshold for each BABE epoch
## Expected Behavior

- the VerificationManager needs to expose a function that sets the `epochDescriptors[epoch]` field to the `EpochDesciptor` for that epoch
- this should be called upon runtime changes or epoch changes
- NOTE: authorities may change between epochs, this needs to be implemented as well, see #997 
- also epoch length may change- if #762 is done before this, should be taken into account. otherwise can assume epoch length remains the same

## Current Behavior

- if an `EpochDescriptor` isn't available for an epoch, it uses the one from the current epoch

## Checklist

noot:
## Changes

- update `VerificationManager` to track changes in verification data via block (ie. by hash, so when verifying a block, it uses the data from the most recent descendant it has)
- update `core.checkForRuntimeChanges` to notify `VerificationManager` if there is a runtime change
- update dot startup as needed

## Tests

```
go test ./lib/babe ./dot/... 
```

## Checklist

## Issues

- closes #762 closes #1023

### DIFF
--- a/dot/core/service.go
+++ b/dot/core/service.go
@@ -152,6 +152,7 @@ func NewService(cfg *Config) (*Service, error) {
 		blockProducer:           cfg.BlockProducer,
 		finalityGadget:          cfg.FinalityGadget,
 		consensusMessageHandler: cfg.ConsensusMessageHandler,
+		verifier:                cfg.Verifier,
 		isFinalityAuthority:     cfg.IsFinalityAuthority,
 		lock:                    &sync.Mutex{},
 	}
@@ -279,7 +280,7 @@ func (s *Service) handleReceivedBlock(block *types.Block) (err error) {
 		return err
 	}
 
-	return s.checkForRuntimeChanges()
+	return s.HandleRuntimeChanges(block.Header)
 }
 
 // handleReceivedMessage handles messages from the network service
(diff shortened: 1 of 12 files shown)


dot/core/service.go
```diff
@@ -304,8 +309,9 @@ func (s *Service) handleReceivedMessage(msg network.Message) (err error) {
 	return err
 }
 
-// checkForRuntimeChanges checks if changes to the runtime code have occurred; if so, load the new runtime
-func (s *Service) checkForRuntimeChanges() error {
+// HandleRuntimeChanges checks if changes to the runtime code have occurred; if so, load the new runtime
+// It also updates the BABE service and block verifier with the new runtime
+func (s *Service) HandleRuntimeChanges(header *types.Header) error {
```
edwardmack, line 314: Does this need to be public?  It seems to only be called inside the service.

noot: the plan is for this to be called from the sync service as well once it imports a block (#1014) but could be changed back if that changes

Issue #1022, opened 2020-07-23
Add Developer RPC Endpoints for Authorities Change, BABE Randomness, & BABE epoch threshold


amerameen:
## Expected Behavior

- RPC endpoints that allows us to trigger authorities changes, a change in BABE randomness, or a change in BABE epoch threshold.

## Current Behavior

- This doesn't exist.

## Possible Solution

-

## Steps to Reproduce

1.
2.
3.
4.

## Specification

- go version:
- gossamer version:
- gossamer commit tag:
- gossamer commit hash:
- operating system:
- additional links:

## Checklist

- [x] I have read [CODE_OF_CONDUCT](https://github.com/ChainSafe/gossamer/blob/development/.github/CODE_OF_CONDUCT.md) and [CONTRIBUTING](https://github.com/ChainSafe/gossamer/blob/development/.github/CONTRIBUTING.md) 
- [x] I have provided as much information as possible and necessary
- [ ] I am planning to submit a pull request to fix this issue myself

PR #1045, opened 2020-08-03
dot/rpc, lib/babe: Update SetAuthorities RPC endpoint to SetBlockProducerAuthorities

edwardmack:
## Changes

- Changed RPC endpoint dev_setAuthorities to dev_setBlockProducerAuthorities
- Added check for key before setting authorities
- Added call to setAuthorityIndex inside SetBlockProducerAuthorities function
- The params for this endpoint is an array containing public addresses and weight in the form:

```
[
      ["5GrwvaEF5zXb26Fz9rcQpDWS57CtERHpNehXCPcNoHGKutQY", 1],
      ["5FHneW46xGXgs5mUiveU4sbTyGBzmstUspZC92UhjJM694ty", 1]
]
```

## Tests

```
 go test ./dot/rpc/...
go test ./lib/babe/...
```

## Checklist

## Issues

- related to #1022, still to complete handling for BABE Randomness, and BABE epoch.


lib/babe/babe.go
```diff
@@ -265,6 +265,24 @@ func (b *Service) SetAuthorities(a []*types.BABEAuthorityData) {
 	b.authorityData = a
 }
 
+// SetBlockProducerAuthorities sets the current Block Producer Authorities and sets Authority index
+func (b *Service) SetBlockProducerAuthorities(data []*types.BABEAuthorityData) error {
```
noot, line 269: i think you can rename this to `SetAuthorities` and delete the existing `SetAuthorities` function, unless that causes a lot of stuff to break?

edwardmack:
I've updated the name `SetBlockProducerAuthorities` to `SetAuthorities` however `SetAuthorities` now may return an error, so I'm not sure if I'm doing error handling the way we'd like in the digest go routines.  Please check it out and let me know.

dot/core/digest.go
```diff
@@ -180,14 +181,22 @@ func (h *DigestHandler) handleBABEChangesOnImport(header *types.Header) {
 	num := header.Number
 	resume := h.babeResume
 	if resume != nil && num.Cmp(resume.atBlock) == 0 {
-		h.babe.SetAuthorities(h.babeAuths)
+		err := h.babe.SetAuthorities(h.babeAuths)
+		if err != nil {
+			log.Error("error setting authorities", "error", err)
+			return
```
noot, line 187: I think you can change this to a `warn` and not return, since it's ok to update the authorities to a set that doesn't include our key

PR #1058, opened 2020-08-06
Update logging to include File Name and Line Numbers

### LINKED ISSUE #965: logger: add file and line numbers to warn, error, crit logs
## Expected Behavior

- error logs should include file and line numbers for easy debugging
- see log15 docs for adding file and line numbers: https://godoc.org/github.com/inconshreveable/log15#Logger

## Current Behavior

- no file or line numbers

edwardmack:
## Changes

- Updated logging to use `log.CallerFileHandler` which outputs file name and line number with logging output. 

## Tests

Run node and observe addition of file name and line number appended to logs as the caller parameter:
```
INFO[08-07|10:27:10] signal interrupt, shutting down...       pkg=dot caller=node.go:332
```
## Checklist

## Issues

- closes #965


edwardmack:
Note:
- I included new log format it `lib/common` but it should probable be added to ChainSafe/log15 format.go file.
- I added a lot a //nolint comments to ignore buffer write errors (They are also ignored in other formats in format.go).  Let me know your thoughts regarding this (//nolint, or handle errors?)

noot:
@edwardmack  this looks good! wondering what the advantages are of the new format vs this: https://godoc.org/github.com/inconshreveable/log15#hdr-Logging_File_Names_and_Line_Numbers

also, if you could PR the log format to ChainSafe/log15, I think that would make more sense, then we can just update the log formatter here.

edwardmack:
@noot, good question, actually I was not aware of `log.CallerFileHandler` which meets our requirements (I should have spent more time reading the docs).  The only difference is that the handler that I wrote only outputs file name/line number for WARN, ERROR and CRIT (which is what the issue describes).  However I think it makes more sense to use existing library functionality (`log.CallerFileHandler`) instead of writing our own.  I don't think having file name/line number for all log entries is a issue.  And therefore I no longer think we should update ChainSafe/log15.  

Thanks for the link, and let me know your thoughts.

noot:
@edwardmack yeah, I don't mind having line numbers for all logs. however it might be nice to have the stack trace only for crit and error, what do you think? either way, any line numbers in the logs is an improvement!

PR #1084, opened 2020-08-17
lib/grandpa: verify block justification when handling FinalizationMessage

### LINKED ISSUE #1051: verify block justification when handling FinalizationMessage
## Expected Behavior

- need to verify block justification when receiving a finalization message (in grandpa `MessageHandler.HandleMessage`)
- if justification is valid, mark block as final (ie call `blockState.SetFinalizedHash()`)
- since `grandpa.Justification` includes the authority public key and signature, we need to check that:
1. the signature is valid (need setID and stage to recreate signed `FullVote`, stage=precommit, use current grandpa setID, eventually can use block number -> setID map)
2. the authority was in the set indicated by the setID
3. the total # of signatures is >=2/3 the number of voters in the set

the justification verification will have overlap with #1070 

## Current Behavior

- no justification verification

## Checklist

edwardmack:
## Changes

- added function `verifyJustification` which check given FinalizationMessage and confirms:
  - Verifies signature is valid for singed FullVote.
  - Authority was in the set indicated by setID
  - The total number of signatures is >= 2/3 the number of voters is the list.

-

## Tests

```
go test ./lib/grandpa/...
```

## Checklist

## Issues

- closes #1051

### DIFF
--- a/lib/grandpa/errors.go
+++ b/lib/grandpa/errors.go
@@ -72,8 +72,8 @@ var ErrNotFinalizationMessage = errors.New("cannot get finalized hash from VoteM
 // ErrNoJustification is returned when no justification can be found for a block, ie. it has not been finalized
 var ErrNoJustification = errors.New("no justification found for block")
 
-// ErrMinVotesNotMet is returned when the number of votes is less than the required minimum
-var ErrMinVotesNotMet = errors.New("minimum number of votes not met")
+// ErrMinVotesNotMet is returned when the number of votes is less than the required minimum in a Justification
+var ErrMinVotesNotMet = errors.New("minimum number of votes not met in a Justification")
 
 // ErrInvalidCatchUpRound is returned when a catch-up message is received with an invalid round
 var ErrInvalidCatchUpRound = errors.New("catch up request is for future round")
--- a/lib/grandpa/errors.go
+++ b/lib/grandpa/errors.go
@@ -71,3 +71,6 @@ var ErrNotFinalizationMessage = errors.New("cannot get finalized hash from VoteM
 
 // ErrNoJustification is returned when no justification can be found for a block, ie. it has not been finalized
 var ErrNoJustification = errors.New("no justification found for block")
+
+// ErrMinVotesNotMet is returned when the number of votes is less than the required minimum
+var ErrMinVotesNotMet = errors.New("minimum number of votes not met")
--- a/lib/grandpa/message_handler.go
+++ b/lib/grandpa/message_handler.go
@@ -17,6 +17,8 @@
 package grandpa
 
 import (
+	"fmt"
+	"github.com/ChainSafe/gossamer/lib/crypto/ed25519"
 	"github.com/ChainSafe/gossamer/lib/scale"
 )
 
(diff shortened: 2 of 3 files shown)


lib/grandpa/errors.go
```diff
@@ -72,5 +72,8 @@ var ErrNotFinalizationMessage = errors.New("cannot get finalized hash from VoteM
 // ErrNoJustification is returned when no justification can be found for a block, ie. it has not been finalized
 var ErrNoJustification = errors.New("no justification found for block")
 
+// ErrMinVotesNotMet is returned when the number of votes is less than the required minimum
```
noot, line 75: could you add `in a Justification`?

lib/grandpa/message_handler.go
```diff
@@ -91,6 +98,57 @@ func (h *MessageHandler) handleFinalizationMessage(msg *FinalizationMessage) (*C
...
+			}
+		}
+		if !authFound {
+			return ErrVoterNotFound
+		}
+		sigCount++
+	}
+
+	// confirm total # signatures >= 2/3 of number of voters
+	if !(sigCount >= (2/3)*len(h.grandpa.state.voters)) {
```
noot, line 146: you could use `h.grandpa.state.threshold()` instead of `(2/3)*len(h.grandpa.state.voters)`

lib/grandpa/message_handler_test.go
```diff
@@ -177,6 +177,57 @@ func TestMessageHandler_FinalizationMessage_NoCatchUpRequest(t *testing.T) {
 	cm, err := fm.ToConsensusMessage()
 	require.NoError(t, err)
 
+	h := NewMessageHandler(gs, st.Block)
+	out, err := h.HandleMessage(cm)
+	require.EqualError(t, err, "signature is not valid")
```
noot, line 182: change to `require.Equal(t, err, ErrInvalidSignature)`?

lib/grandpa/message_handler_test.go
```diff
@@ -177,6 +177,57 @@ func TestMessageHandler_FinalizationMessage_NoCatchUpRequest(t *testing.T) {
...
+	require.Nil(t, out)
+
+}
+
+func TestMessageHandler_FinalizationMessage_NoCatchUpRequest_ValidSig(t *testing.T) {
+	st := newTestState(t)
+	voters := newTestVoters(t)
```
noot, line 189: hm, something in this test seems off, `newTestVoters` should return 9 votes, so a finalization message with only 1 signature shouldn't meet the threshold, maybe I am missing something?

lib/grandpa/message_handler.go
```diff
@@ -91,6 +98,57 @@ func (h *MessageHandler) handleFinalizationMessage(msg *FinalizationMessage) (*C
...
+		authFound := false
+		for _, auth := range h.grandpa.Authorities() {
+			if reflect.DeepEqual(auth.Key.AsBytes(), j.AuthorityID) {
+				authFound = true
+				break
+			}
+		}
+		if !authFound {
+			return ErrVoterNotFound
+		}
```
noot, line 141: it might also be nice to break this into a separate function that accepts `*Justification`, `stage`, `vote`, `round`, and `setID`, since it'll be used for catch up response handling as well

lib/grandpa/message_handler.go
edwardmack: good idea, this also fixes the issue you pointed out with the test.

lib/grandpa/message_handler_test.go
edwardmack: No, I fixed verify so this test fails, going to add tests...

lib/grandpa/message_handler.go
edwardmack: All set.

PR #1085, opened 2020-08-19
Implement MustEncode

monkrus:
## Changes

Implement MustEncode for messages and block components. Files affected: block.go header.go
-
-
-

## Tests

```

```

## Checklist

- [ x] I have read [CODE_OF_CONDUCT](https://github.com/ChainSafe/gossamer/blob/development/.github/CODE_OF_CONDUCT.md) and [CONTRIBUTING](https://github.com/ChainSafe/gossamer/blob/development/.github/CONTRIBUTING.md) 
- [x] I have provided as much information as possible and necessary
- [x] I have reviewed my own pull request before requesting a review
- [x] All integration tests and required coverage checks are passing

## Issues

-

PR #1090, opened 2020-09-02
dot/types: add MustEncode for Header

### LINKED ISSUE #351: Implement MustEncode for Messages and Block Components
I propose we implement `MustEncode` functionality for common types that should never fail. An implementation would look something like this:
```
func (st *SomeType) MustEncode() []byte {
    raw, err := codec.Encode(st)
    if err != nil {
        panic(err)
    }
    return raw
}
```

It seems the p2p message types, as well as block/header would be good places to implement this, there may be others.

Other uses of `Must***`:
https://godoc.org/github.com/cosmos/cosmos-sdk/codec#MarshalJSONIndent
https://github.com/ethereum/go-ethereum/blob/461291882edce0ac4a28f64c4e8725b7f57cbeae/common/math/integer.go#L73-L79

noot:
## Changes

- updates tests from #1085 
-
-

## Tests

```
go test ./dot/types
```

## Checklist

## Issues

- closes #351
- closes #1085


edwardmack (approved):
Nice work, lgtm.

Issue #905, opened 2020-05-28
runtime: implement v0.6.x ext_ functions


noot:
## Expected Behavior

imports_old.go (previous ext functions)
- [x] `ext_is_validator`
- [x] `ext_local_storage_get`
- [x] `ext_local_storage_compare_and_set`
- [x] `ext_network_state`
- [x] `ext_submit_transaction`
- [x] `ext_local_storage_set`

imports.go (new ext functions that are ready to be implemented)
- [x] `ext_kill_child_storage`
- [ ] `ext_child_storage_root`
- [x] `ext_clear_child_storage`
- [x] `ext_get_allocated_child_storage`
- [ ] `ext_secp256k1_ecdsa_recover_compressed`
- [ ] `ext_log`

New ext functions that are not specced yet (please create a new issue with the following fns in it when the above functions are implemented) 
- [ ] `ext_sandbox_memory_new`
- [ ] `ext_sandbox_memory_teardown`
- [ ] `ext_sandbox_instantiate`
- [ ] `ext_sandbox_invoke`
- [ ] `ext_sandbox_instance_teardown`
- [ ] `ext_sandbox_memory_get`
- [ ] `ext_sandbox_memory_set`

It will also be a good idea to look at the v0.7 ext_ functions (see spec Appendix E) to see what changes we will need to make to upgrade, and make the v0.6 functions easily wrappable for v0.7.

## Checklist

- [x] I have read [CONTRIBUTING](CONTRIBUTING.md) and [CODE_OF_CONDUCT](CODE_OF_CONDUCT.md)
- [x] I have provided as much information as possible and necessary
- [ ] I am planning to submit a pull request to fix this issue myself

edwardmack:
Items remaining undone on checklist because these have not been spec'd yet.

PR #1100, opened 2020-09-15
lib/runtime: implement ext_is_validator

edwardmack:
## Changes

- Implement `ext_is_validator` function, returns 1 if node's role is 4, returns 0 otherwise.
-
-

## Tests

```
go test ./lib/runtime/...
```

## Checklist

## Issues

- part of #905


lib/runtime/runtime.go
```diff
@@ -34,6 +34,7 @@ type Ctx struct {
 	storage   Storage
 	allocator *FreeingBumpHeapAllocator
 	keystore  *keystore.GenericKeystore
+	validator int32
```
noot, line 37: this might make more sense as a bool? then convert it to int32 inside ext_is_validator

lib/runtime/runtime.go
```diff
@@ -92,10 +94,16 @@ func NewRuntime(code []byte, cfg *Config) (*Runtime, error) {
 
 	memAllocator := NewAllocator(instance.Memory, 0)
 
+	validator := int32(0)
+	if (cfg.Role&byte(4))>>2 == 1 {
```
noot, line 98: can you do `cfg.Role == byte(4)`?

noot (approved):

PR #1113, opened 2020-09-30
dot/core: handle chain re-orgs by re-adding transactions into pool

### LINKED ISSUE #764: Upon A Chain Reorg, add Transactions back into the Transaction Pool
When there is a chain re-org (i.e. the canonical chain changes), there may be transactions that need to be put back into the Transaction Pool.

The logic would be something like the following: 

A) Event: Canonical chain changes
B) Keep track of which transactions that were included in blocks that were removed (i.e. blocks that were part of the previous canonical chain, but not the current canonical chain)
C) Sync the new canonical chain
D) Compare the transactions in the newly synced blocks to the transactions from Step B.
E) Add transactions from (B) that are not found in (D) into the Transaction Pool

noot:
## Changes

- implement `handleChainReorg` function which is called upon a block import
- it checks if there is a chain re-org and if so, validates then readds the transactions included on that chain back into the pool

## Tests

```
go test ./dot/core
```

## Checklist

## Issues

- closes #764

### DIFF
--- a/dot/core/service.go
+++ b/dot/core/service.go
@@ -256,7 +256,7 @@ func (s *Service) handleBlocks(ctx context.Context) {
 
 			err = s.handleChainReorg(prev, block.Header.Hash())
 			if err != nil {
-				log.Warn("failed to readd transactions to chain upon re-org", "error", err)
+				log.Warn("failed to re-add transactions to chain upon re-org", "error", err)
 			}
 
 			err = s.handleRuntimeChanges(block.Header)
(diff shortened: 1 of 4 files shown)


dot/core/service.go
```diff
@@ -251,6 +254,11 @@ func (s *Service) handleBlocks(ctx context.Context) {
 				log.Warn("failed to store storage trie in database", "error", err)
 			}
 
+			err = s.handleChainReorg(prev, block.Header.Hash())
+			if err != nil {
+				log.Warn("failed to readd transactions to chain upon re-org", "error", err)
```
edwardmack, line 259: readd -> read

edwardmack (approved):
Great work, lgtm.

dot/core/service.go
noot: this should say readd, I'll change it to re-add

PR #1128, opened 2020-10-13
dot/types: add scale encoding to data in digest

edwardmack:
## Changes

- Testing polkadot.js api to check method chain.getHeader was often generating error:

```
RPC-CORE: getBlock(hash?: BlockHash): SignedBlock:: createType(SignedBlock):: Struct: failed on 'block':: Struct: failed on 'header':: Struct: failed on 'digest':: Struct: failed on 'logs':: Bytes: required length less than remainder, expected at least 253533030, found 64
```

- It seems that the byte array data needs to be scale encoded so that the length of the data is known (since scale encoding encodes the array length is the first byte).  Adding scale encode/decoding to the data in digest fixes this issue.
-
-

## Tests

```
go test ./dot/types/...
```

## Checklist

## Issues

-


dot/types/digest.go
```diff
@@ -133,7 +135,16 @@ func (d *PreRuntimeDigest) Type() byte {
 func (d *PreRuntimeDigest) Encode() []byte {
 	enc := []byte{PreRuntimeDigestType}
 	enc = append(enc, d.ConsensusEngineID[:]...)
-	return append(enc, d.Data...)
+	// encode data
+	buffer := bytes.Buffer{}
+	se := scale.Encoder{Writer: &buffer}
+	_, err := se.Encode(d.Data)
```
noot, line 141: you can also replace this with just `scale.Encode(d.Data, []byte{})` I believe

dot/types/digest.go
```diff
@@ -133,7 +135,16 @@ func (d *PreRuntimeDigest) Type() byte {
...
+	if err != nil {
+		// todo handle this error
```
noot, line 143: please update runtime to return `[]byte, error`

noot (approved):

Issue #1115, opened 2020-10-01
tests: use export config command to create test configs


noot:
## Expected Behavior

- currently there are config files in the `tests/utils` dir that are used by the integration tests
- these should be removed and created on an as-needed basis by the tests, then exported to a file using the `export` subcommand 

## Checklist

- [x] I have read [CODE_OF_CONDUCT](https://github.com/ChainSafe/gossamer/blob/development/.github/CODE_OF_CONDUCT.md) and [CONTRIBUTING](https://github.com/ChainSafe/gossamer/blob/development/.github/CONTRIBUTING.md) 
- [x] I have provided as much information as possible and necessary
- [ ] I am planning to submit a pull request to fix this issue myself

PR #1144, opened 2020-10-20
Generate test configs from code.

arijitAD:
## Changes

- Generate test configs from code.

## Issues
- #1115


dot/utils.go
```diff
@@ -187,24 +188,32 @@ func NewTestConfigWithFile(t *testing.T) (*Config, *os.File) {
...
-		newFile *os.File
-		err     error
-		raw     []byte
-	)
+	raw, err := toml.Marshal(*cfg)
+	if  err != nil {
+		logger.Error("failed to marshal configuration", "error", err)
+		os.Exit(1)
+	}
+	return WriteConfig(raw, fp)
```
noot, line 196: I think this should convert the `Config` to `ctoml.Config` otherwise there may be some marshalling/unmarshalling errors

noot (approved):
looks good to me, please fix lint issues, looks like some of the `cmd` package tests are failing also

dot/utils.go
arijitAD: Test failures were due to a different issue. This would be `Config` and wouldn't have marshaling failures because non-primitive datatype would be ignored. This code was already there before.

Issue #1032, opened 2020-07-24
tests: use build-spec command to build test genesis files


noot:
## Expected Behavior

- currently there are various genesis files in the tests/utils package with different numbers of authorities
- since we now have a `build-spec` command we can easily use this to build test raw genesis files as needed by parsing and editing a non-raw genesis file
- should use this to create genesis files, can also now add tests with more flexible node numbers

## Checklist

- [x] I have read [CODE_OF_CONDUCT](https://github.com/ChainSafe/gossamer/blob/development/.github/CODE_OF_CONDUCT.md) and [CONTRIBUTING](https://github.com/ChainSafe/gossamer/blob/development/.github/CONTRIBUTING.md) 
- [x] I have provided as much information as possible and necessary
- [ ] I am planning to submit a pull request to fix this issue myself

PR #1147, opened 2020-10-23
Generate test genesis file.

arijitAD:
## Changes
- Generate test genesis file.

## Issues
- #1032


noot (approved):
looks good! in general I would prefer not to alter the codebase for the tests (unless there is some issue with the codebase), but it's a small change here so it's okay.

arijitAD:
> looks good! in general I would prefer not to alter the codebase for the tests (unless there is some issue with the codebase), but it's a small change here so it's okay.

Yes. That makes sense.

Issue #1125, opened 2020-10-08
dot/state: delete storage tries from database on block finalization


noot:
## Expected Behavior

-

## Current Behavior

- in `StorageState`, storage tries for each block are stored in the database so they can be loaded when needing to build off the storage at that block
- however when a chain gets finalized, we should delete the storage tries that are no longer needed from the database (ie. implement `pruneStorage` and call it on finalized block)

## Possible Solution

- all storage tries that are ancestors of the finalized block are no longer needed
- generally all storage tries on chains that aren't in the blocktree anymore should be removed. see `BlockState.SetFinalizedHash`

## Checklist

- [x] I have read [CODE_OF_CONDUCT](https://github.com/ChainSafe/gossamer/blob/development/.github/CODE_OF_CONDUCT.md) and [CONTRIBUTING](https://github.com/ChainSafe/gossamer/blob/development/.github/CONTRIBUTING.md) 
- [x] I have provided as much information as possible and necessary
- [ ] I am planning to submit a pull request to fix this issue myself

PR #1161, opened 2020-11-03
Remove pruned keys from Storage state.

arijitAD:
## Changes
- Start a background goroutine to GC prune keys from Storage state.
- Implement prune key API in Storage state.

## Checklist

## Issues
- #1125

### DIFF
--- a/dot/state/block.go
+++ b/dot/state/block.go
@@ -49,7 +49,7 @@ type BlockState struct {
 	importedLock  sync.RWMutex
 	finalizedLock sync.RWMutex
 
-	pruneKeyCh chan common.Hash
+	pruneKeyCh chan []common.Hash
 }
 
 // NewBlockState will create a new BlockState backed by the database located at basePath
@@ -64,7 +64,7 @@ func NewBlockState(db chaindb.Database, bt *blocktree.BlockTree) (*BlockState, e
 		db:         chaindb.NewTable(db, blockPrefix),
 		imported:   make(map[byte]chan<- *types.Block),
 		finalized:  make(map[byte]chan<- *types.Header),
-		pruneKeyCh: make(chan common.Hash, 1000),
+		pruneKeyCh: make(chan []common.Hash, 1000),
 	}
 
 	bs.genesisHash = bt.GenesisHash()
@@ -87,7 +87,7 @@ func NewBlockStateFromGenesis(db chaindb.Database, header *types.Header) (*Block
 		db:         chaindb.NewTable(db, blockPrefix),
 		imported:   make(map[byte]chan<- *types.Block),
 		finalized:  make(map[byte]chan<- *types.Header),
-		pruneKeyCh: make(chan common.Hash, 1000),
+		pruneKeyCh: make(chan []common.Hash, 1000),
 	}
 
 	err := bs.setArrivalTime(header.Hash(), uint64(time.Now().Unix()))
@@ -441,8 +441,8 @@ func (bs *BlockState) SetFinalizedHash(hash common.Hash, round, setID uint64) er
 		if err != nil {
 			return err
 		}
-		bs.pruneKeyCh <- rem
 	}
+	bs.pruneKeyCh <- pruned
 
 	return bs.db.Put(finalizedHashKey(round, setID), hash[:])
 }
--- a/dot/state/block.go
+++ b/dot/state/block.go
@@ -34,6 +34,8 @@ import (
 
 var blockPrefix = "block"
 
+const pruneKeyBufferSize = 1000
+
 // BlockState defines fields for manipulating the state of blocks, such as BlockTree, BlockDB and Header
 type BlockState struct {
 	bt                 *blocktree.BlockTree
@@ -49,7 +51,7 @@ type BlockState struct {
 	importedLock  sync.RWMutex
 	finalizedLock sync.RWMutex
 
-	pruneKeyCh chan []common.Hash
+	pruneKeyCh chan *types.Header
 }
 
 // NewBlockState will create a new BlockState backed by the database located at basePath
@@ -64,7 +66,7 @@ func NewBlockState(db chaindb.Database, bt *blocktree.BlockTree) (*BlockState, e
 		db:         chaindb.NewTable(db, blockPrefix),
 		imported:   make(map[byte]chan<- *types.Block),
 		finalized:  make(map[byte]chan<- *types.Header),
-		pruneKeyCh: make(chan []common.Hash, 1000),
+		pruneKeyCh: make(chan *types.Header, pruneKeyBufferSize),
 	}
 
 	bs.genesisHash = bt.GenesisHash()
@@ -87,7 +89,7 @@ func NewBlockStateFromGenesis(db chaindb.Database, header *types.Header) (*Block
 		db:         chaindb.NewTable(db, blockPrefix),
 		imported:   make(map[byte]chan<- *types.Block),
 		finalized:  make(map[byte]chan<- *types.Header),
-		pruneKeyCh: make(chan []common.Hash, 1000),
+		pruneKeyCh: make(chan *types.Header, pruneKeyBufferSize),
 	}
 
 	err := bs.setArrivalTime(header.Hash(), uint64(time.Now().Unix()))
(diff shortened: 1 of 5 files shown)


dot/state/block.go
```diff
@@ -48,6 +48,8 @@ type BlockState struct {
 	finalized     map[byte]chan<- *types.Header
 	importedLock  sync.RWMutex
 	finalizedLock sync.RWMutex
+
+	pruneKeyCh chan []common.Hash
```
noot, line 52: is it possible to change this to write-only? ie `chan<- []common.Hash`

also is there a reason it's a slice and not just a hash?

dot/state/block.go
```diff
@@ -57,11 +59,12 @@ func NewBlockState(db chaindb.Database, bt *blocktree.BlockTree) (*BlockState, e
...
-		baseDB:    db,
-		db:        chaindb.NewTable(db, blockPrefix),
-		imported:  make(map[byte]chan<- *types.Block),
-		finalized: make(map[byte]chan<- *types.Header),
+		bt:         bt,
+		baseDB:     db,
+		db:         chaindb.NewTable(db, blockPrefix),
+		imported:   make(map[byte]chan<- *types.Block),
+		finalized:  make(map[byte]chan<- *types.Header),
+		pruneKeyCh: make(chan []common.Hash, 1000),
```
noot, line 67: reason for 1000? seems very large

dot/state/block.go
```diff
@@ -79,11 +82,12 @@ func NewBlockState(db chaindb.Database, bt *blocktree.BlockTree) (*BlockState, e
...
-		baseDB:    db,
-		db:        chaindb.NewTable(db, blockPrefix),
-		imported:  make(map[byte]chan<- *types.Block),
-		finalized: make(map[byte]chan<- *types.Header),
+		bt:         blocktree.NewBlockTreeFromGenesis(header, db),
+		baseDB:     db,
+		db:         chaindb.NewTable(db, blockPrefix),
+		imported:   make(map[byte]chan<- *types.Block),
+		finalized:  make(map[byte]chan<- *types.Header),
+		pruneKeyCh: make(chan []common.Hash, 1000),
```
noot, line 90: same here, also could you make the size a constant?

dot/state/block.go
```diff
@@ -438,6 +442,7 @@ func (bs *BlockState) SetFinalizedHash(hash common.Hash, round, setID uint64) er
 			return err
 		}
 	}
+	bs.pruneKeyCh <- pruned
```
noot, line 445: I think this should be moved inside the for loop, since we only want to delete storage tries that are no longer part of the finalized chain (similarly to the discussion we had in standup for deleting the blocks)

dot/state/service.go
```diff
@@ -151,6 +153,21 @@ func (s *Service) Initialize(data *genesis.Data, header *types.Header, t *trie.T
...
+	for {
+		select {
+		case keys := <-s.Block.pruneKeyCh:
+			for _, v := range keys {
+				s.Storage.pruneKey(v)
+			}
+		case <-s.closeCh:
+			return
+		default:
+			time.Sleep(100 * time.Millisecond)
```
noot, line 166: I don't think this is needed, if there's nothing in either channel it'll block  at the `select` until there is

dot/state/service.go
```diff
@@ -151,6 +153,21 @@ func (s *Service) Initialize(data *genesis.Data, header *types.Header, t *trie.T
...
+		}
+	}
+}
```
noot, line 169: I think this function could go in `StorageState`, since `StorageState` contains the `blockState`

dot/state/service_test.go
```diff
@@ -128,3 +129,72 @@ func TestService_BlockTree(t *testing.T) {
...
 }
+func Test_ServicePruneStorage(t *testing.T) {
+	testDir := utils.NewTestDir(t)
+	defer utils.RemoveTestDir(t)
+
+	serv := NewService(testDir, log.LvlTrace)
+	serv.UseMemDB()
+
+	genesisHeader, err := types.NewHeader(common.NewHash([]byte{0}), big.NewInt(0), trie.EmptyHash, trie.EmptyHash, [][]byte{})
+	require.Nil(t, err)
```
noot, line 140: you can use `testGenesisHeader` I believe

dot/state/service_test.go
```diff
@@ -128,3 +129,72 @@ func TestService_BlockTree(t *testing.T) {
...
+	tr := trie.NewEmptyTrie()
+	err = serv.Initialize(genesisData, genesisHeader, tr, firstEpochInfo)
+	require.Nil(t, err)
+
+	err = serv.Start()
+	require.Nil(t, err)
+
+	serv.Block = newTestBlockState(t, testGenesisHeader)
+	serv.Storage = newTestStorageState(t)
+	ch := make(chan *types.Header, 3)
```
noot, line 153: `Service.Start()` creates a `BlockState` and `StorageState` so I think you can remove lines 151-152 and use `serv.Block` and `serv.Storage` directly

dot/state/service_test.go
```diff
@@ -128,3 +129,72 @@ func TestService_BlockTree(t *testing.T) {
...
+	serv.UseMemDB()
+
+	genesisHeader, err := types.NewHeader(common.NewHash([]byte{0}), big.NewInt(0), trie.EmptyHash, trie.EmptyHash, [][]byte{})
+	require.Nil(t, err)
+
+	genesisData := new(genesis.Data)
+
+	tr := trie.NewEmptyTrie()
+	err = serv.Initialize(genesisData, genesisHeader, tr, firstEpochInfo)
+	require.Nil(t, err)
```
noot, line 146: could you change all the `require.Nil` to `require.NoError` for error handling?

dot/state/service_test.go
```diff
@@ -128,3 +129,72 @@ func TestService_BlockTree(t *testing.T) {
...
+		if i == len(chain)-1 {
+			serv.Block.SetFinalizedHash(b.Hash(), 0, 0)
+		} else {
+			rootHash, err := emptyTrie.Hash()
+			require.NoError(t, err)
+			prunedArr = append(prunedArr, prunedBlock{hash: b.Hash(), dbKey: rootHash[:]})
+		}
+	}
+
+	time.Sleep(10 * time.Second)
```
noot, line 187: this seems long? is it necessary to be this long?

dot/state/storage.go
```diff
@@ -74,10 +74,18 @@ func NewStorageState(db chaindb.Database, blockState *BlockState, t *trie.Trie)
 	}, nil
 }
 
-func (s *StorageState) pruneStorage() { //nolint
-	// TODO: when a block is finalized, delete non-finalized tries from DB and mapping
-	// as well as all states before finalized block
-	// TODO: pruning options? eg archive, full, etc
+func (s *StorageState) pruneKey(hash common.Hash) {
```
noot, line 77: the hash that's passed into this function is the block hash, you need to find the `StateRoot` for that block, since `s.tries` stores tries by `StateRoot`

you can do this by getting the header:
```
	header, err := s.blockState.GetHeader(hash)
	if err != nil {
		return nil, err
	}
```

then using `s.tries[&header.StateRoot]`

dot/state/block.go
arijitAD: I designed it to be slice since I thought keys are only pruned when they are finalized and we can process all the prune keys at one go and then sleep for some time in the Go-routine. I have removed the sleep logic and update this to be chan of hash.

dot/state/block.go
arijitAD: `pruneKey()` Go-routine might take time to process the pruning in storage and I didn't want the `BlockState` to wait for the channel. I wasn't sure of how many keys could be pruned at one go. Can you suggest a fair value?

dot/state/service.go
arijitAD: Agree

dot/state/service_test.go
arijitAD: Updated it to one sec.

noot (approved):

edwardmack (approved):
Nice work, lgtm.

PR #1184, opened 2020-11-20
Implement IsSyncing in Health check function

### LINKED ISSUE #700: Add syncing condition to network health
## Expected Behavior

`Health()` in network service should return accurate information about network health. 

## Current Behavior

`Health()` in network service does not have a condition implemented for `isSyncing` and we should probably also look into a better way to determine `ShouldHavePeers`.

```
func (s *Service) Health() common.Health {
	return common.Health{
		Peers:           s.host.peerCount(),
		IsSyncing:       false, // TODO
		ShouldHavePeers: !s.noBootstrap,
	}
}
```

## Possible Solution

Use the new `syncing` submodule, add a `isSyncing` struct property to the `syncing` submodule, and update `isSyncing` to true when the best peer block id (all peers) is greater than the best host block id and false when the best peer block id is less than or equal to the best host block id.

RyRy79261:
## Changes
- Added a `GetSyncedState` to the Syncer interface
- Updated `Health()` to call `GetSyncedState`

## Tests

`dot/network/service_test.go: TestService_Health`

## Checklist

## Issues
closes #700

### DIFF
--- a/dot/sync/syncer.go
+++ b/dot/sync/syncer.go
@@ -492,3 +492,8 @@ func (s *Service) handleDigests(header *types.Header) error {
 
 	return nil
 }
+
+// GetSyncedState exposes the synced state
+func (s *Service) GetSyncedState() bool {
+	return s.synced
+}
--- a/dot/sync/syncer.go
+++ b/dot/sync/syncer.go
@@ -493,7 +493,7 @@ func (s *Service) handleDigests(header *types.Header) error {
 	return nil
 }
 
-// GetSyncedState exposes the synced state
-func (s *Service) GetSyncedState() bool {
+// IsSynced exposes the synced state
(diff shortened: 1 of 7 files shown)


edwardmack (approved):
Nice work, looks good to me.

dot/sync/syncer.go
```diff
@@ -492,3 +492,8 @@ func (s *Service) handleDigests(header *types.Header) error {
 
 	return nil
 }
+
+// GetSyncedState exposes the synced state
+func (s *Service) GetSyncedState() bool {
```
noot, line 497: perhaps rename to something like `IsSynced` ?

noot (approved):
overall looks good, perhaps a small rename for clarity!

PR #1187, opened 2020-11-23
dot/state: Fix node storing head data

### LINKED ISSUE #1142: node not storing head data for blocks 1-64
## Expected Behavior

- calling RPC chain_getHeader with hash for block 1 - 64 should return header data.
- 

## Current Behavior

- calling RPC chain_getHeader with hash for block 1 - 64 should returns "Key not found" error.
- works as it should for blocks higher than 64

## Possible Solution

-

## Steps to Reproduce
found testing #790 

1. RPC call `chain_getBlockHash` with parameter any block number less than 65.
2. use resulting hash to call `chain_getHeader`

## Specification

- go version:
- gossamer version:
- gossamer commit tag:
- gossamer commit hash:
- operating system:
- additional links:

## Checklist

edwardmack:
## Changes

- The code updates the check BlockState.IsDescendantOf so that it doesn't prune blocks that are on the current chain.
-
-

## Tests

```
go test ./dot/state/...
```

## Checklist

## Issues

- closes #1142


lib/blocktree/blocktree.go
```diff
@@ -157,10 +157,6 @@ func (bt *BlockTree) Prune(newRoot Hash) (pruned []Hash) {
 	// get pruned nodes
 	pruned = bt.head.getAllDescendantsExcluding(nil, newRoot)
 
-	// set blocktree with new root node
-	next := newBlockTreeFromNode(n, bt.db)
-	*bt = *next
-
```
noot, line 163: we actually still need to prune the blocktree at the end of `SetFinalizedHash`, so we'll have to re-add this. I think this could be done by first copying the original blocktree (will need to implement a `Copy` func) and using that to check `IsDescendantOf`

dot/state/block.go
```diff
@@ -444,6 +444,13 @@ func (bs *BlockState) SetFinalizedHash(hash common.Hash, round, setID uint64) er
 			return err
 		}
 
+		if des, err1 := bs.IsDescendantOf(rem, hash); des {
+			if err1 != nil {
+				logger.Error("pruning block error", "error", err)
+			}
+			continue
+		}
```
noot, line 452: I think you can ignore the error here, the only times an error would be returned are if `rem` or `hash` don't exist which should never be the case

noot (commented):
could you add a `Copy` function for the blocktree that deep copies it and use that to check `IsDescendantOf`? also, could you update the tests as well?

RyRy79261 (approved):

RyRy79261:
Ran the tests locally, looks good to me

lib/blocktree/blocktree.go
edwardmack: I've implemented DeepCopy for BlockTree and for node, but I don't understand where/how it should be used. Please let me know how these should be used.

noot: the current code is this, the issue is that `bt.prune` deletes the previous nodes, causing `IsDescendantOf` to fail
```
	pruned := bs.bt.Prune(hash)
	for _, rem := range pruned {
		header, err := bs.GetHeader(rem)
		if err != nil {
			return err
		}

		if des, _ := bs.IsDescendantOf(rem, hash); des {
			continue
		}

		err = bs.DeleteBlock(rem)
		if err != nil {
			return err
		}

		bs.pruneKeyCh <- header
	}
```

instead you could copy the blocktree before pruning, then use the copy to check `IsDescendantOf`
```
        copy := bs.bt.DeepCopy()
	pruned := bs.bt.Prune(hash)
	for _, rem := range pruned {
		header, err := bs.GetHeader(rem)
		if err != nil {
			return err
		}

		if des, _ := copy.IsDescendantOf(rem, hash); des {
			continue
		}

		err = bs.DeleteBlock(rem)
		if err != nil {
			return err
		}

		bs.pruneKeyCh <- header
	}
```

noot (approved):

Issue #1163, opened 2020-11-03
deprecate status messages and message identifiers


noot:
## Expected Behavior

- since `Status` messages are no longer used for nodes ^v0.8.24, we can remove all usages and logic related to `StatusMessage`
- we can continue to accept legacy substreams for now, but will remove support for them in the future
- also, message identifiers have been deprecated, the message received will be known by context

## Current Behavior

- sends/receives status message
- encodes/decodes messages based on identifier

## Checklist

- [x] I have read [CODE_OF_CONDUCT](https://github.com/ChainSafe/gossamer/blob/development/.github/CODE_OF_CONDUCT.md) and [CONTRIBUTING](https://github.com/ChainSafe/gossamer/blob/development/.github/CONTRIBUTING.md) 
- [x] I have provided as much information as possible and necessary
- [x] I am planning to submit a pull request to fix this issue myself

PR #1192, opened 2020-11-24
dot/network: deprecate status messages, remove state.NetworkState

noot:
## Changes

- remove state.NetworkState, since it's not used for anything 
- remove status message functionality from network
- remove related functionality from syncer

## Tests

```
go test ./... -short
```

## Checklist

## Issues

- related to #1163


edwardmack (approved):
Nice work, 1000 line removed!  Great stuff, lgtm.

dot/network/host.go
```diff
@@ -222,7 +217,7 @@ func (h *host) broadcast(msg Message) {
 }
 
 // broadcastExcluding sends a message to each connected peer except specified peer
-func (h *host) broadcastExcluding(msg Message, peer peer.ID) {
+func (h *host) broadcastExcluding(msg Message, peer peer.ID) { //nolint
```
RyRy79261, line 220: Would this work as an array? Excluding might be more than one peer in some cases

RyRy79261 (approved):

dot/network/host.go
noot: yeah, we don't really use this right now, since we need to implement some sort of actual gossip mechanism, even if it's just random selection. it might be deleted in the future

PR #1202, opened 2020-11-30
CMD/Account - Coverage test issue

### LINKED ISSUE #984: potentially broken tests in cmd/gossamer account
## Expected Behavior

Test of command gossamer account should confirm that it works properly under different situations(options, flags, etc.)

## Current Behavior

I checked in my device(MacBook Pro) what part of account.go file is covered and saw that some core parts are not covered, for example, there is a test scenario which supposedly tests command gossamer account --generate, however, this is what I get from coverage:
image

I think in the tests the cli.Context is not created correctly because after creation the arguments of the context (ctx.Args()) is empty.

## Possible Solution

One quick possible solution is to actually simulate by using the app:
```go
// TestAccountGenerate test "gossamer account --generate"
func TestAccountGenerate(t *testing.T) {
	testDir := utils.NewTestDir(t)
	defer utils.RemoveTestDir(t)

	directory := fmt.Sprintf("--basepath=%s", testDir)

	err := app.Run([]string{"irrelevant", "account", directory, "--generate=true", "--password=false"})
	if err != nil {
		t.Fatal(err)
	}
}
```
This indeed tests the command because that part of code is covered:
image

## Steps to Reproduce

If this is not happening only in my device, then you can reproduce it by:
1. Go to the cmd/gossamer directory
2. Run in the terminal > go test -coverprofile=coverage.out
3. Run in the terminal > go tool cover -html=coverage.out
4. When the page is opened in your browser, check the account.go file

## Specification

- go version: go1.14.4 darwin/amd64
- gossamer --version: 0.0.1
- operating system: macOS

## Checklist

RyRy79261:
## Changes

- Implemented solution to extend coverage 

## Tests

```
   cd cmd/gossamer 
   go test -coverprofile=coverage.out
   go tool cover -html=coverage.out
   //  When the page is opened in your browser, check the account.go file

```

## Checklist

## Issues

closes #984


noot (commented):
could you also do the same fix for the other tests in `cmd/gossamer/account_test.go`?

RyRy79261:
@noot I've added the solution with the relevant flags however I'm unsure what exactly the solution is doing or how to assess it's success

cmd/gossamer/account_test.go
```diff
@@ -26,6 +27,8 @@ import (
 func TestAccountGenerate(t *testing.T) {
 	testDir := utils.NewTestDir(t)
 	defer utils.RemoveTestDir(t)
+	directory := fmt.Sprintf("--basepath=%s", testDir)
+	err := app.Run([]string{"irrelevant", "account", directory, "--generate=true", "--password=false"})
```
noot, line 31: please check error, same for other tests

noot (approved):

edwardmack (approved):

cmd/gossamer/account_test.go
```diff
@@ -95,7 +108,11 @@ func TestAccountGenerateType(t *testing.T) {
 func TestAccountImport(t *testing.T) {
 	testDir := utils.NewTestDir(t)
 	defer utils.RemoveTestDir(t)
-
+	directory := fmt.Sprintf("--basepath=%s", testDir)
+	err := app.Run([]string{"irrelevant", "account", directory, "--import"})
```
noot, line 112: you need to provide a value for `--import`, previously it was `testfile`

RyRy79261:
Expanding the tests for this & hunting down some test keystores for the import tests

cmd/gossamer/account_test.go
RyRy79261: @noot Where is the testfile sourced?

arijitAD (approved):

Issue #1201, opened 2020-11-27
update genesis format to match Polkadot's


noot:
## Expected Behavior

- the genesis format ("raw") should match polkadot's (see https://raw.githubusercontent.com/paritytech/polkadot/3094e8d267b502351c0bccf3bfd1f7443d79fd20/service/res/polkadot.json)

## Current Behavior

- the `raw` field is different,  currently, we have an array, while polkadot has `top` and `childrenDefault` fields within raw.
- there are some extra fields that are not in our genesis file:
- [x] `chainType`
- [x] `properties`
- [x] `forkBlocks` 
- [x] `badBlocks`
- [x] `consensusEngine`

## Possible Solution

- update the `raw` field formatting
- add missing fields 

## Checklist

- [x] I have read [CODE_OF_CONDUCT](https://github.com/ChainSafe/gossamer/blob/development/.github/CODE_OF_CONDUCT.md) and [CONTRIBUTING](https://github.com/ChainSafe/gossamer/blob/development/.github/CONTRIBUTING.md) 
- [x] I have provided as much information as possible and necessary
- [ ] I am planning to submit a pull request to fix this issue myself

Issue #790, opened 2020-04-16
Polkadot.js Integration


amerameen:
Once our RPC calls are sufficiently implemented, add support for Polkadot.js

For this ticket, focus on connecting to Polkadot.js and based on that we can create more issues for added support.

PR #1226, opened 2020-12-09
implement system_chainType

noot:
## Changes

- add `ChainType` to the genesis file
- implement `system_chainType` rpc method

## Tests

```
go test ./dot/rpc/...
```

## Checklist

## Issues

- related to #1201 and #790


dot/rpc/modules/system.go
```diff
@@ -75,6 +75,12 @@ func (sm *SystemModule) Name(r *http.Request, req *EmptyRequest, res *string) er
 	return nil
 }
 
+// ChainType returns the chain type
+func (sm *SystemModule) ChainType(r *http.Request, req *EmptyRequest, res *string) error {
```
arijitAD, line 79: Can you add a struct for the response? I was thinking we should have a struct for all request and reponse.

arijitAD (approved):

dot/rpc/modules/system.go
noot: sure, will update!

edwardmack: @arijitAD Why? I'm not disagreeing that all request and response should be a struct, but I want to better understand why.

edwardmack (changes requested):
Nice work, please also add ChainType to Genesis struct and Data struct in lib/genesis/genesis.go so that buildSpecAction can use this.

dot/rpc/modules/system.go
noot: it seems like changing the types to a strict is what caused polkadot.js to break, so I think I will leave it as is for now

lib/genesis/genesis.go
```diff
@@ -33,6 +34,7 @@ type Genesis struct {
 type Data struct {
 	Name       string
 	ID         string
+	ChainType  string
```
noot, line 37: @edwardmack  I added it here and above, is there somewhere else I need to add it?

edwardmack: @noot no, I think this should be all set now.

edwardmack (approved):
Nice work, please also add ChainType to Genesis struct and Data struct in lib/genesis/genesis.go so that buildSpecAction can use this.

dot/rpc/modules/system.go
arijitAD: I was thinking that it might make the code cleaner and in the future, we can just modify the struct if we want to add fields. But we can leave it for responses that have a single field.

Issue #1105, opened 2020-09-17
Update runtime ext functions for v0.8


amerameen:
remaining:

**ext_sandbox [blocked, not spec'd]**
- [ ] ext_sandbox_instance_teardown_version_1
- [ ] ext_sandbox_instantiate_version_1
- [ ] ext_sandbox_invoke_version_1
- [ ] ext_sandbox_memory_get_version_1
- [ ] ext_sandbox_memory_new_version_1
- [ ] ext_sandbox_memory_set_version_1
- [ ] ext_sandbox_memory_teardown_version_1

**ext_crypto** (some should be implemented in legacy_imports.go)
- [x] ext_crypto_ed25519_generate_version_1 @arijitAD 
- [x] ext_crypto_ed25519_public_keys_version_1 @arijitAD 
- [x] ext_crypto_ed25519_sign_version_1 @arijitAD 
- [x] ext_crypto_ed25519_verify_version_1 @arijitAD 
- [x] ext_crypto_finish_batch_verify_version_1 @arijitAD 
- [x] ext_crypto_secp256k1_ecdsa_recover_version_1 @arijitAD 
- [x] ext_crypto_sr25519_generate_version_1 @arijitAD 
- [x] ext_crypto_sr25519_public_keys_version_1 @arijitAD 
- [x] ext_crypto_sr25519_sign_version_1 @arijitAD 
- [x] ext_crypto_sr25519_verify_version_1 [is this a duplicate of the following??) @arijitAD 
- [x] ext_crypto_sr25519_verify_version_2 @arijitAD 
- [x] ext_crypto_start_batch_verify_version_1 @arijitAD 

**ext_trie** 
- [x] ext_trie_blake2_256_root_version_1 @noot 
- [x] ext_trie_blake2_256_ordered_root_version_1 @noot 

**ext_misc** (some should be implemented in legacy_imports.go)
- [x] ext_misc_print_hex_version_1
- [x] ext_misc_print_num_version_1
- [x] ext_misc_runtime_version_version_1

**ext_default** (should be implemented in legacy_imports.go)
- [x] ext_default_child_storage_read_version_1 @arijitAD 
- [x] ext_default_child_storage_clear_version_1 @arijitAD 
- [x] ext_default_child_storage_clear_prefix_version_1 @arijitAD 
- [x] ext_default_child_storage_exists_version_1 @arijitAD 
- [x] ext_default_child_storage_get_version_1 @arijitAD 
[...]

PR #1242, opened 2020-12-16
Implement ext_default functions for v0.8 runtime.

arijitAD:
## Changes

- Implement ext_default functions for v0.8 runtime.

## Tests

```
make test
```

## Checklist

## Issues

-  #1105


lib/runtime/wasmer/imports.go
```diff
@@ -325,62 +325,187 @@ func ext_misc_runtime_version_version_1(context unsafe.Pointer, z C.int64_t) C.i
...
-	logger.Warn("[ext_default_child_storage_root_version_1] unimplemented")
-	return 0
+
+	instanceContext := wasm.IntoInstanceContext(context)
+	storage := instanceContext.Data().(*runtime.Context).Storage
+
+	// TODO: Should we write to DB or load from DB to state trie
+	// Commit() loads the state trie from DB
+	// WriteTrieToDB() writes the state trie to the DB
+	err := storage.Commit()
```
noot, line 455: you can remove this, the `Commit` is only used before the runtime execution actually starts

lib/runtime/interface.go
```diff
@@ -80,6 +80,10 @@ type Storage interface {
 	DeleteChildStorage(key []byte) error
 	ClearChildStorage(keyToChild, key []byte) error
 	NextKey([]byte) []byte
+	GetChildByPrefix(keyToChild, prefix []byte) [][]byte
+	GetChildNextKey(keyToChild, prefix []byte) []byte
+	GetChild(keyToChild []byte) (*trie.Trie, error)
+	Commit() error
```
noot, line 86: please remove

lib/runtime/wasmer/imports.go
```diff
@@ -325,62 +325,187 @@ func ext_misc_runtime_version_version_1(context unsafe.Pointer, z C.int64_t) C.i
...
+	copy(memory[valueBuf:valueBuf+valueLen], value[offset:])
+
+	size := uint32(len(value[offset:]))
+	sizeBuf := make([]byte, 4)
+	binary.LittleEndian.PutUint32(sizeBuf, size)
+
+	sizeSpan, err := toWasmMemoryOptional(instanceContext, sizeBuf)
+	if err != nil {
+		logger.Error("[ext_default_child_storage_read_version_1] failed to allocate", "error", err)
+		panic(err)
```
noot, line 351: instead of panicking can you return 0? same for all other functions

lib/runtime/wasmer/imports.go
```diff
@@ -325,62 +325,187 @@ func ext_misc_runtime_version_version_1(context unsafe.Pointer, z C.int64_t) C.i
...
+	}
+
+	child, err := storage.GetChild(asMemorySlice(instanceContext, childStorageKey))
+	if err != nil {
+		logger.Error("[ext_default_child_storage_root_version_1] failed to retrieve child", "error", err)
+		panic(err)
+	}
+
+	childRoot, err := child.EncodeRoot()
```
noot, line 467: please change to `child.Hash()`

noot (commented):
looks good!! could you add some tests? there are some tests in `imports_test.go` you can base them off. the ext_ function tests use the test wasm from the spec team which has a `rtm_ext_...` function corresponding to each `ext_...` function, so you can use that to call the `ext_...`. the source is here: https://github.com/w3f/polkadot-spec/blob/master/test/runtimes/hostapi/src/lib.rs

lib/runtime/wasmer/imports.go
arijitAD: In the spec, it is specially mentioned to commit. 
`Commits all existing operations and computes the resulting child storage root.`
So wouldn't this cause an issue?

lib/runtime/wasmer/imports.go
arijitAD: We are panicking at other places where we use `toWasmMemoryOptional`. Should I change them as well?

lib/runtime/wasmer/imports.go
noot: sorry, it's kind of a naming conflict, what the spec means is that all the changes have been put in the child trie, which has already happened at this point, so you can remove this

lib/runtime/wasmer/imports.go
noot: yes you can change them

noot (approved):
great work! :D

RyRy79261 (approved):

edwardmack (approved):
Nice work, lgtm.

PR #1247, opened 2020-12-17
Implement ext_misc for v0.8 runtime.

arijitAD:
## Changes

- Implement ext_misc for v0.8 runtime.

## Tests

```
make test
```

## Checklist

## Issues

- #1105


lib/runtime/wasmer/imports.go
```diff
@@ -299,15 +300,18 @@ func ext_trie_blake2_256_ordered_root_version_1(context unsafe.Pointer, data C.i
...
+
+	ptr, size := int64ToPointerAndSize(int64(data))
+	ext_print_hex(context, C.int32_t(ptr), C.int32_t(size))
 }
 
 //export ext_misc_print_num_version_1
-func ext_misc_print_num_version_1(context unsafe.Pointer, a C.int64_t) {
+func ext_misc_print_num_version_1(context unsafe.Pointer, data C.int64_t) {
 	logger.Trace("[ext_misc_print_num_version_1] executing...")
 	logger.Warn("[ext_misc_print_num_version_1] unimplemented")
```
noot, line 313: pls delete unimplemented log

lib/runtime/wasmer/imports.go
```diff
@@ -318,10 +322,31 @@ func ext_misc_print_utf8_version_1(context unsafe.Pointer, data C.int64_t) {
...
-	return 0
+
+	instanceContext := wasm.IntoInstanceContext(context)
+	data := asMemorySlice(instanceContext, dataSpan)
+
+	version := &runtime.VersionAPI{
+		RuntimeVersion: &runtime.Version{},
+		API:            nil,
+	}
+	version.Decode(data)
```
noot, line 335: it looks like this function returns the data that's returned from the `core_version` runtime call, you may need to add this to the context, and when the runtime is initialized, try to call the `instance.Version` function and then set the runtime version data in the context

lib/runtime/wasmer/imports.go
```diff
@@ -318,10 +322,31 @@ func ext_misc_print_utf8_version_1(context unsafe.Pointer, data C.int64_t) {
 }
 
 //export ext_misc_runtime_version_version_1
-func ext_misc_runtime_version_version_1(context unsafe.Pointer, z C.int64_t) C.int64_t {
+func ext_misc_runtime_version_version_1(context unsafe.Pointer, dataSpan C.int64_t) C.int64_t {
```
noot, line 325: according to the spec the input data is another wasm blob.... so this means you will need to instantiate a runtime inside this function using the input data, call `instance.Version` on it, then return that data

arijitAD: I have modified it. But what should I pass in config when creating a new instance?

arijitAD:
`ext_misc_runtime_version_version_1` and a couple of other functions are imported functions and are not exported. How do I write tests for them?

lib/runtime/wasmer/imports.go
noot: you can pass the existing `Storage` in the context, and `ImportsNodeRuntime` as the imports. the rest you can leave out, since the `Version` call won't be using them

noot:
@arijitAD in that case, don't worry about tests for these for now

noot (approved):
looks good!

edwardmack (approved):
Nice work, lgtm.

PR #1250, opened 2020-12-18
Implement ext_crypto for runtime v0.8.

arijitAD:
## Changes

- Implement ext_crypto for runtime v0.8.

## Tests

```

```

## Checklist

## Issues

- #1105


lib/runtime/wasmer/imports_test.go
```diff
@@ -249,6 +256,174 @@ func Test_ext_storage_set_version_1(t *testing.T) {
...
+	fmt.Println(pubKey)
+
+	encSign, err := scale.Encode(sigData)
+	require.NoError(t, err)
+	encMsg, err := scale.Encode(blakeHash.ToBytes())
+	require.NoError(t, err)
+
+	ret, err := inst.Exec("rtm_ext_crypto_secp256k1_ecdsa_recover_version_1", append(encSign, encMsg...))
+
+	buf := &bytes.Buffer{}
```
arijitAD, line 405: @noot Can you check this test? It's returning two extra bytes. In Wasm the length is set to 68. But returning 70 here. `buf.Write(ret[2:])` This fixes the test. But any idea about this strange behaviour?

arijitAD:
Currently, I am working on background verification of the signature which is used by function `ext_crypto_finish_batch_verify` and `ext_crypto_start_batch_verify`. Looking at the implementation in [substrate](https://github.com/paritytech/substrate) and [Kagome](https://github.com/soramitsu/kagome).

arijitAD:
Added commit that does background signature verification.

lib/runtime/test_helpers.go
```diff
@@ -302,3 +305,24 @@ func (trn *TestRuntimeNetwork) NetworkState() common.NetworkState {
...
+		require.NoError(t, err)
+
+		sign, err := key.Private().Sign(msg)
+		require.NoError(t, err)
+
+		signs[i] = &Signature{
+			PubKey:    key.Public().Encode(),
+			Sign:      sign,
+			Msg:       msg,
+			KyeTypeID: crypto.Ed25519Type,
```
noot, line 324: typo: should be `KeyTypeID`

lib/runtime/types.go
```diff
@@ -81,14 +88,146 @@ type Context struct {
...
+type SignatureVerifier struct {
+	batch   []*Signature
+	init    bool
+	inValid bool
+	sync.RWMutex
+	closeCh chan struct{}
+}
+
+// Init ...
+func (sv *SignatureVerifier) Init() {
```
noot, line 146: maybe name `NewSignatureVerifier`? to match other constructor functions

lib/runtime/types.go
```diff
@@ -81,14 +88,146 @@ type Context struct {
...
+		}
+	}
+	return nil
+}
+
+// SignatureVerifier ...
+type SignatureVerifier struct {
+	batch   []*Signature
+	init    bool
+	inValid bool
```
noot, line 140: change to `invalid`

lib/runtime/types.go
```diff
@@ -81,14 +88,146 @@ type Context struct {
...
+
+// IsStarted ...
+func (sv *SignatureVerifier) IsStarted() bool {
+	sv.RLock()
+	defer sv.RUnlock()
+	return sv.init
+}
+
+// IsInValid ...
+func (sv *SignatureVerifier) IsInValid() bool {
```
noot, line 164: same here

lib/runtime/types.go
```diff
@@ -81,14 +88,146 @@ type Context struct {
...
+			}
+
+			sv.Lock()
+			sign := sv.batch[0]
+			sv.batch = sv.batch[1:len(sv.batch)]
+			sv.Unlock()
+
+			err := sign.verify()
+			if err != nil {
+				log.Error("[ext_crypto_start_batch_verify_version_1] %s", err)
```
noot, line 207: can you update this to be `log.Error("[ext_crypto_start_batch_verify_version_1]", ,"error", err)` - the logger expects pairs of string-values to come after the initial message string

lib/runtime/types.go
```diff
@@ -81,14 +88,146 @@ type Context struct {
...
+	SigVerifier *SignatureVerifier
 	// TransactionStorageChanges is used by ext_storage_start_transaction to keep track of
 	// changes made after it's called. The next call to ext_storage_commit_transaction will
 	// commit all the changes, or if ext_storage_rollback_transaction is called, the changes
 	// will be discarded.
 	TransactionStorageChanges []*TransactionStorageChange
 }
 
+// Signature ...
+type Signature struct {
```
noot, line 100: maybe you could move all the signature related code to a new file, since there's quite a lot of code

lib/runtime/wasmer/imports.go
```diff
@@ -185,45 +188,191 @@ func ext_sandbox_memory_teardown_version_1(context unsafe.Pointer, a C.int32_t)
...
+func ext_crypto_ed25519_generate_version_1(context unsafe.Pointer, keyTypeID C.int32_t, seedSpan C.int64_t) C.int32_t {
 	logger.Trace("[ext_crypto_ed25519_generate_version_1] executing...")
-	logger.Warn("[ext_crypto_ed25519_generate_version_1] unimplemented")
-	return 0
+
+	instanceContext := wasm.IntoInstanceContext(context)
+	runtimeCtx := instanceContext.Data().(*runtime.Context)
+
+	// TODO: key types not yet implemented
+	// id := memory[idData:idData+4]
```
noot, line 198: key types are now implemented (if you look in lib/keystore/keystore.go the `Name` field is the 4-byte ID) however it'll require changing the keystore that's passed to the runtime to be the `GlobalKeystore` for the node, instead of just the `Acco` keystore that's currently passed. I'll open an issue for this and it can be done separately

lib/runtime/wasmer/imports.go
```diff
@@ -185,45 +188,191 @@ func ext_sandbox_memory_teardown_version_1(context unsafe.Pointer, a C.int32_t)
...
+	if err != nil {
+		logger.Warn("[ext_crypto_ed25519_generate_version_1] cannot generate key", "error", err)
+		return 0
+	}
+
+	var kp crypto.Keypair
+
+	if seed.Exists() {
+		kp, err = ed25519.NewKeypairFromSeed(seedBytes)
+	} else {
```
noot, line 214: this is a bit of a misnomer- according to the spec, the seed is a BIP39 mnemonic seed that's used to derive the key, whereas the function `NewKeypairFromSeed` generates a keypair using the seed as a random seed. you can leave it like this for now, since the BIP39 functionality hasn't been added to gossamer yet (but it's implemented here https://github.com/ChainSafe/go-schnorrkel/blob/master/bip39.go) I will open an issue for this as well

noot: maybe just add a comment here saying this needs to be updated to use BIP39 derivation

lib/runtime/wasmer/imports.go
```diff
@@ -234,44 +383,172 @@ func ext_crypto_secp256k1_ecdsa_recover_compressed_version_1(context unsafe.Poin
...
+
+	// TODO: key types not yet implemented
+	// id := asMemorySlice(instanceContext,keyTypeID)
+
+	seedBytes := asMemorySlice(instanceContext, seed)
+	kp, err := sr25519.NewKeypairFromSeed(seedBytes)
+	if err != nil {
+		logger.Trace("[ext_crypto_sr25519_generate_version_1] cannot generate key", "error", err)
+		return 0
+	}
```
noot, line 400: same here in regards to BIP39. also, I think the seed here is also optional?

lib/runtime/wasmer/imports.go
```diff
@@ -185,45 +188,191 @@ func ext_sandbox_memory_teardown_version_1(context unsafe.Pointer, a C.int32_t)
...
+
+	ptr, err := runtimeCtx.Allocator.Allocate(uint32(len(keys) * 32))
+	if err != nil {
+		logger.Error("[ext_crypto_ed25519_public_keys_version_1]", "error", err)
+		return 0
+	}
+
+	for i, key := range keys {
+		copy(memory[ptr+uint32(i*32):ptr+uint32((i+1)*32)], key.Encode())
+	}
```
noot, line 253: based on the spec, this is expecting a SCALE encoded slice of `[32]byte`, so maybe create a `[][32]byte` with the public keys and encode and return that

lib/runtime/wasmer/imports.go
```diff
@@ -185,45 +188,191 @@ func ext_sandbox_memory_teardown_version_1(context unsafe.Pointer, a C.int32_t)
...
+	if err != nil {
+		logger.Error("[ext_crypto_ed25519_sign_version_1]", "error", err)
+		return 0
+	}
+
+	signingKey := runtimeCtx.Keystore.GetKeypair(pubKey)
+	if signingKey == nil {
+		logger.Error("[ext_crypto_ed25519_sign_version_1] could not find key in keystore", "public key", pubKey)
+		return 0
+	}
```
noot, line 278: the spec says if the key can't be found it should return `Option(None)`, so this should still allocate and return `0`

lib/runtime/wasmer/imports.go
```diff
@@ -185,45 +188,191 @@ func ext_sandbox_memory_teardown_version_1(context unsafe.Pointer, a C.int32_t)
...
+
+	if signVerify.IsStarted() {
+		signature := runtime.Signature{
+			PubKey:    pubKey.Encode(),
+			Sign:      newSign,
+			Msg:       newMsg,
+			KyeTypeID: crypto.Ed25519Type,
+		}
+		signVerify.Add(&signature)
+		return 0
```
noot, line 321: this tells the runtime the signature is invalid, but it seems it was added to the batch? maybe I am misunderstanding this

lib/runtime/wasmer/imports.go
```diff
@@ -185,45 +188,191 @@ func ext_sandbox_memory_teardown_version_1(context unsafe.Pointer, a C.int32_t)
...
+	instanceContext := wasm.IntoInstanceContext(context)
+	memory := instanceContext.Memory().Data()
+
+	// msg must be the 32-byte hash of the message to be signed.
+	// sig must be a 65-byte compact ECDSA signature containing the
+	// recovery id as the last element
+	newMsg := memory[msg : msg+32]
+	newSig := memory[sig : sig+65]
+
+	// TODO: Verify the return in spec.
```
noot, line 362: I believe the `Result` type returns `byte(0) + key` if success, `byte(1)` otherwise, but I will confirm with the spec team

lib/runtime/wasmer/imports.go
```diff
@@ -234,44 +383,172 @@ func ext_crypto_secp256k1_ecdsa_recover_compressed_version_1(context unsafe.Poin
...
+	runtimeCtx := instanceContext.Data().(*runtime.Context)
+	memory := instanceContext.Memory().Data()
+
+	keys := runtimeCtx.Keystore.Sr25519PublicKeys()
+
+	ptr, err := runtimeCtx.Allocator.Allocate(uint32(len(keys) * 32))
+	if err != nil {
+		logger.Error("[ext_crypto_ed25519_public_keys_version_1]", "error", err)
+		return 0
+	}
```
noot, line 426: return `Option(None)` here

lib/runtime/wasmer/imports.go
```diff
@@ -234,44 +383,172 @@ func ext_crypto_secp256k1_ecdsa_recover_compressed_version_1(context unsafe.Poin
...
+
+	signingKey := runtimeCtx.Keystore.GetKeypair(pubKey)
+	if signingKey == nil {
+		logger.Error("[ext_crypto_sr25519_sign_version_1] could not find key in keystore", pubKey)
+		return 0
+	}
+
+	msgData := asMemorySlice(instanceContext, msg)
+	sig, err := signingKey.Sign(msgData)
+	fmt.Println("sig", sig)
```
noot, line 457: remove

lib/runtime/wasmer/imports.go
```diff
@@ -234,44 +383,172 @@ func ext_crypto_secp256k1_ecdsa_recover_compressed_version_1(context unsafe.Poin
...
-	logger.Warn("[ext_crypto_sr25519_sign_version_1] unimplemented")
-	return 0
+	instanceContext := wasm.IntoInstanceContext(context)
+	runtimeCtx := instanceContext.Data().(*runtime.Context)
+	memory := instanceContext.Memory().Data()
+
+	pubKey, err := sr25519.NewPublicKey(memory[key : key+32])
+	if err != nil {
+		logger.Error("[ext_crypto_sr25519_sign_version_1] failed to get public key", "error", err)
+		return 0
```
noot, line 446: return `Option(None)` here

lib/runtime/wasmer/imports.go
```diff
@@ -234,44 +383,172 @@ func ext_crypto_secp256k1_ecdsa_recover_compressed_version_1(context unsafe.Poin
...
+	}
+
+	signingKey := runtimeCtx.Keystore.GetKeypair(pubKey)
+	if signingKey == nil {
+		logger.Error("[ext_crypto_sr25519_sign_version_1] could not find key in keystore", pubKey)
+		return 0
```
noot, line 452: return `Option(None)` here

lib/runtime/wasmer/imports.go
```diff
@@ -234,44 +383,172 @@ func ext_crypto_secp256k1_ecdsa_recover_compressed_version_1(context unsafe.Poin
...
+	}
+
+	msgData := asMemorySlice(instanceContext, msg)
+	sig, err := signingKey.Sign(msgData)
+	fmt.Println("sig", sig)
+	if err != nil {
+		logger.Error("[ext_crypto_sr25519_sign_version_1] could not sign message")
+		return 0
```
noot, line 460: return `Option(None)` here

lib/runtime/wasmer/imports.go
```diff
@@ -234,44 +383,172 @@ func ext_crypto_secp256k1_ecdsa_recover_compressed_version_1(context unsafe.Poin
...
+	}
+
+	ret, err := toWasmMemoryOptional(instanceContext, sig)
+	if err != nil {
+		logger.Error("[ext_crypto_sr25519_sign_version_1] failed to allocate memory", err)
+		return 0
```
noot, line 466: return `Option(None)` here

lib/runtime/wasmer/imports.go
```diff
@@ -234,44 +383,172 @@ func ext_crypto_secp256k1_ecdsa_recover_compressed_version_1(context unsafe.Poin
...
+	if signVerify.IsStarted() {
+		signature := runtime.Signature{
+			PubKey:    pub.Encode(),
+			Sign:      newSig,
+			Msg:       newMsg,
+			KyeTypeID: crypto.Sr25519Type,
+		}
+		signVerify.Add(&signature)
+		return 0
+	}
```
noot, line 497: same comment here as the other verify func

lib/runtime/wasmer/imports.go
```diff
@@ -234,44 +383,172 @@ func ext_crypto_secp256k1_ecdsa_recover_compressed_version_1(context unsafe.Poin
...
+
+	if ok, err := pub.Verify(newMsg, newSig); err != nil || !ok {
+		return 0
+	}
+	return 1
 }
 
 //export ext_crypto_sr25519_verify_version_2
-func ext_crypto_sr25519_verify_version_2(context unsafe.Pointer, a C.int32_t, z C.int64_t, y C.int32_t) C.int32_t {
+func ext_crypto_sr25519_verify_version_2(context unsafe.Pointer, sig C.int32_t, msg C.int64_t, key C.int32_t) C.int32_t {
```
noot, line 506: it seems like version 1 and version 2 of this func use different versions of the underlying crypto library, will confirm the difference

lib/runtime/wasmer/instance.go
```diff
@@ -311,10 +312,10 @@ func (in *LegacyInstance) NetworkService() runtime.BasicNetwork {
 
 // int64ToPointerAndSize converts an int64 into a int32 pointer and a int32 length
 func int64ToPointerAndSize(in int64) (ptr int32, length int32) {
-	return int32(in), int32(in >> 32)
+	return int32(in & 0xffffffff), int32(in >> 32)
```
noot, line 315: why is this changed?

lib/runtime/wasmer/instance.go
```diff
@@ -311,10 +312,10 @@ func (in *LegacyInstance) NetworkService() runtime.BasicNetwork {
...
 }
 
 // pointerAndSizeToInt64 converts int32 pointer and size to a int64
 func pointerAndSizeToInt64(ptr, size int32) int64 {
-	return int64(ptr) + (int64(size) << 32)
+	return int64(ptr) | (int64(size) << 32)
```
noot, line 320: also here

noot: it should be the same so just curious

lib/utils/utils.go
```diff
@@ -192,3 +197,18 @@ func GetKsmccGenesisPath() string {
 
 	return fp
 }
+
+// GenerateKeyPairs public and private secp256k1 keys
+func GenerateKeyPairs() (pubKey, prvKey []byte) {
```
noot, line 202: can you use the `lib/crypto/secp256k1` lib for this?

noot (commented):
great work! some updates needed to change return types

lib/runtime/wasmer/imports.go
arijitAD: Add() check if it has encountered an invalid signature before. If so then it skips adding it to the batch since the result of the batch will be false anyway.

lib/runtime/wasmer/instance.go
arijitAD: I was experimenting with bitwise operation. Reverted it.

lib/runtime/wasmer/instance.go
arijitAD: Bitwise operation is more optimal than arithmetic operation. The result would be the same.

lib/runtime/wasmer/imports.go
noot: can you update the return type for this func to be `Result`? you might need to make a `toWasmMemoryResult` func or something like that

noot (commented):
looking good! please update return type for `ext_crypto_secp256k1_ecdsa_recover_version_1`

lib/crypto/secp256k1/secp256k1.go
```diff
@@ -147,6 +151,17 @@ func (k *PublicKey) Verify(msg, sig []byte) (bool, error) {
 	return secp256k1.VerifySignature(k.Encode(), msg, sig), nil
 }
 
+// UnmarshalPubkey converts bytes to a secp256k1 public key.
```
noot, line 154: does this take an uncompressed pubkey (65 bytes)? could you update comment to specify format of the expected input

arijitAD:
> looking good! please update return type for `ext_crypto_secp256k1_ecdsa_recover_version_1`

Updated the type to `Result`

lib/runtime/wasmer/imports.go
```diff
@@ -184,45 +188,198 @@ func ext_sandbox_memory_teardown_version_1(context unsafe.Pointer, a C.int32_t)
...
+	instanceContext := wasm.IntoInstanceContext(context)
+	memory := instanceContext.Memory().Data()
+
+	// msg must be the 32-byte hash of the message to be signed.
+	// sig must be a 65-byte compact ECDSA signature containing the
+	// recovery id as the last element
+	message := memory[msg : msg+32]
+	signature := memory[sig : sig+65]
+
+	// TODO: Verify the return in spec.
```
noot, line 369: remove comment

lib/runtime/wasmer/imports.go
```diff
@@ -184,45 +188,198 @@ func ext_sandbox_memory_teardown_version_1(context unsafe.Pointer, a C.int32_t)
...
+	pub, err := secp256k1.RecoverPublicKey(message, signature)
+	if err != nil {
+		logger.Error("[ext_crypto_secp256k1_ecdsa_recover_version_1] failed to recover public key", "error", err)
+		return 0
```
noot, line 373: return nil `Result` here

lib/common/optional/types.go
```diff
@@ -364,3 +364,72 @@ func (x *Body) Value() []byte {
...
+// Set sets the state and data fields of Result.
+func (r *Result) Set(state byte, value []byte) {
+	r.state = state
+	r.data = value
+}
+
+// Exists returns byte(0) if the state is 0, byte(1) if it is state is 1.
+func (r *Result) Exists() byte {
+	return r.state
+}
```
noot, line 430: since `Result` isn't an optional. you can change this to `IsError` or something like that

lib/common/optional/types.go
```diff
@@ -364,3 +364,72 @@ func (x *Body) Value() []byte {
 func (x *Body) Exists() bool {
 	return x.exists
 }
+
+// Result represents a Result type.
```
noot, line 368: could you move this to a different file / sub-package, since `Result` isn't an `Optional`

lib/common/optional/types.go
```diff
@@ -364,3 +364,72 @@ func (x *Body) Value() []byte {
...
+type Result struct {
+	state byte // If data exists then state stores byte(0), otherwise byte(1)
```
noot, line 370: can you rename this to `isError` or something like that?

RyRy79261 (approved):

lib/common/types/result.go
```diff
@@ -0,0 +1,82 @@
...
+import (
+	"io"
+
+	"github.com/ChainSafe/gossamer/lib/common"
+	"github.com/ChainSafe/gossamer/lib/scale"
+)
+
+// Result represents a Result type.
+type Result struct {
+	isErr byte // If data exists then isErr stores byte(0), otherwise byte(1)
```
noot, line 28: update comment to say `if non-error result then isErr stores byte(0), otherwise byte(1)`

lib/common/types/result.go
```diff
@@ -0,0 +1,82 @@
...
+		if err != nil {
+			return nil, err
+		}
+		r.data = value
+	}
+
+	return r, nil
+}
+
+// Value returns the []byte data. It returns nil if it is Result.None.
```
noot, line 79: update comment to say returns nil if `Error`

noot (approved):
great work :)

edwardmack (approved):
Nice work, lgtm.

anthdm (approved):
lgtm

PR #1252, opened 2020-12-18
remove prepended network message type bytes, decode request-response messages based on context 

### LINKED ISSUE #1163: deprecate status messages and message identifiers

### LINKED ISSUE #1222: block request/response should use a custom message decoder
## Expected Behavior

- block requests and responses are sent over the `/sync/2` subprotocol
- this is a request-response protocol where the request sender opens the stream and sends the request message; the responder then either replies with the response or closes the stream
- currently block request/response messages are decoded based on message prefixes, however message prefixes are no longer used so this should be removed
- messages should be decoded based on context, ie. if you pushed the request, you would expect a response on the stream; if you didn't open the stream, you'd expect a request
- need to implement this decoder and update `network.Service.handleSyncStream` accordingly

## Checklist

noot:
## Changes

- create `network.NotificationsMessage` interface which implements `network.Message`, this is used by messages sent over notifications protocol
- remove all prepended message type bytes from message encoding/decoding
- update `BlockRequestMessage` and `BlockResponseMessage` not to have ID anymore as per spec
- update request-response protocol messages to be decoded based on context, ie. whether you sent a request or not

## Tests

```
go test ./... -short
```

## Checklist

## Issues

- closes #1222 closes #1163


dot/network/block_announce.go
```diff
@@ -20,21 +20,80 @@ import (
...
 	"github.com/ChainSafe/gossamer/lib/scale"
 
 	"github.com/libp2p/go-libp2p-core/peer"
 )
 
+var (
+	_ NotificationsMessage = &BlockAnnounceMessage{}
+	_ NotificationsMessage = &BlockAnnounceHandshake{}
+)
+
```
edwardmack, line 35: I don't understand why these vars are declared (without a name)?  How are they used?

edwardmack (approved):
Great work.

dot/network/block_announce.go
noot: it's used to explicitly check that the message types satisfy the `NotificationsMessage` interface

arijitAD (approved):

PR #1260, opened 2020-12-22
Add git pre-push hook.

arijitAD:
## Changes

- Add git pre-push hook.
- Use `make init-repo` to add git hook.

## Tests

```

```

## Checklist

## Issues

-


.githooks/pre-push
```diff
@@ -0,0 +1,18 @@
...
+    exit 1
+fi
+
+go mod tidy
+git diff --exit-code -- go.mod go.sum >/dev/null
+if [ $? -ne 0 ]
+then
+    echo "***ERROR*** 'go.mod' file is updated. Add the file before proceeding to push."
+    exit 1
+fi
```
noot, line 18: what does this do? will it error if the `go.mod` is updated on purpose?

arijitAD: It will error. I assumed that the `go.mod` will be part of the commit If it's modified when we do `git push`. There was a cleaner way to take `md5` hash of `go.mod` and `go.sum` and then compare them. But Linux and mac have a different command to compute md5 and I thought it might get complicated. Also, This check can be skipped using `git push --no-verify`

noot (approved):

RyRy79261 (approved):

Issue #1243, opened 2020-12-16
potential context leak detected


dutterbutter:
## Expected Behavior

- Should be able to run `go vet ./...` without warnings. For a description of what `go vet` does please see [here](https://golang.org/cmd/vet/).

## Current Behavior

- Running `go vet ./...` results in the following warnings:

```
# github.com/ChainSafe/gossamer/dot/core
dot/core/digest.go:109:7: the cancel function returned by context.WithCancel should be called, not discarded, to avoid a context leak
dot/core/digest.go:111:7: the cancel function returned by context.WithCancel should be called, not discarded, to avoid a context leak
dot/core/service.go:188:7: the cancel function returned by context.WithCancel should be called, not discarded, to avoid a context leak
dot/core/service.go:192:7: the cancel function returned by context.WithCancel should be called, not discarded, to avoid a context leak
dot/core/service.go:195:7: the cancel function returned by context.WithCancel should be called, not discarded, to avoid a context leak
# github.com/ChainSafe/gossamer/dot/network
dot/network/service.go:95:2: the cancel function is not used on all paths (possible context leak)
dot/network/service.go:105:3: this return statement may be reached without using the cancel var defined on line 95
``` 

## Possible Solution

- Fix context usage as described in the warnings.

## Steps to Reproduce

1. `go vet ./...`
2.
3.
4.

## Specification

- go version:
- gossamer version:
- gossamer commit tag:
- gossamer commit hash:
- operating system:
- additional links:

## Checklist

- [x] I have read [CODE_OF_CONDUCT](https://github.com/ChainSafe/gossamer/blob/development/.github/CODE_OF_CONDUCT.md) and [CONTRIBUTING](https://github.com/ChainSafe/gossamer/blob/development/.github/CONTRIBUTING.md) 
- [x] I have provided as much information as possible and necessary
- [ ] I am planning to submit a pull request to fix this issue myself

PR #1262, opened 2021-01-04
Resolve context leak.

arijitAD:
The idiomatic way to use context in Go is to pass them in function.
`Do not store Contexts inside a struct type; instead, pass a Context explicitly to each function that needs it.`
Reference: https://www.ardanlabs.com/blog/2019/09/context-package-semantics-in-go.html
We can revisit the design. Probably use a close channel to signal close.

## Changes

- Fix context leak errors thrown by `go vet ./...`

## Tests

```

```

## Checklist

## Issues

- #1243


noot (approved):

RyRy79261 (approved):

dutterbutter:
Can you also update CI to include vet check now that it passes. See https://github.com/ChainSafe/gossamer/blob/development/.github/workflows/checks.yml#L36

PR #1294, opened 2021-01-12
[dot/config] add chainType to gssmr default SystemInfo

edwardmack:
## Changes

- Added DefaultChainType value to dot/config.go for rpc system_ChainType to use.  Polkadot.js calls system_chainType when connecting and throws error if null, so this now set to development.
- Not sure if this should be stored as System config value, or stored in chain data?  Open to suggestions since this seems like a temporary fix to get polkadot.js to stop complaining.
-

## Tests

```

```

## Checklist

## Issues

-


noot:
@edwardmack  the chainType comes from the genesis file, does it currently use the value there?

RyRy79261 (approved):

PR #1298, opened 2021-01-13
Read chainType from genesis data, not from system info

edwardmack:
## Changes

- Updated code to use data initialized from genesis.json for RPC system_ChainType call.
- Removed ChainType from SystemInfo configuration struct.
- Update tests.
-

## Tests

```
go test ./dot/rpc/...
```

## Checklist

## Issues

-

### DIFF
--- a/dot/rpc/modules/api.go
+++ b/dot/rpc/modules/api.go
@@ -7,7 +7,6 @@ import (
 	"github.com/ChainSafe/gossamer/dot/types"
 	"github.com/ChainSafe/gossamer/lib/common"
 	"github.com/ChainSafe/gossamer/lib/crypto"
-	"github.com/ChainSafe/gossamer/lib/genesis"
 	"github.com/ChainSafe/gossamer/lib/runtime"
 	"github.com/ChainSafe/gossamer/lib/transaction"
 )
@@ -20,7 +19,6 @@ type StorageAPI interface {
 	RegisterStorageChangeChannel(sub state.StorageSubscription) (byte, error)
 	UnregisterStorageChangeChannel(id byte)
 	GetStateRootFromBlock(bhash *common.Hash) (*common.Hash, error)
-	GetGenesisData() (*genesis.Data, error)
 }
 
 // BlockAPI is the interface for the block state
--- a/dot/rpc/modules/api.go
+++ b/dot/rpc/modules/api.go
@@ -90,4 +90,5 @@ type SystemAPI interface {
 	SystemVersion() string
 	NodeName() string
 	Properties() map[string]interface{}
+	ChainType() string
 }
--- a/dot/rpc/modules/api.go
+++ b/dot/rpc/modules/api.go
@@ -7,6 +7,7 @@ import (
 	"github.com/ChainSafe/gossamer/dot/types"
 	"github.com/ChainSafe/gossamer/lib/common"
 	"github.com/ChainSafe/gossamer/lib/crypto"
+	"github.com/ChainSafe/gossamer/lib/genesis"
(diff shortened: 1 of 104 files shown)


anthdm (approved):
lgtm

dot/state/storage.go
```diff
@@ -397,3 +397,8 @@ func (s *StorageState) pruneStorage(closeCh chan interface{}) {
 		}
 	}
 }
+
+// GetGenesisData retrieves current genesis data from database
+func (s *StorageState) GetGenesisData() (*genesis.Data, error) {
+	return LoadGenesisData(s.baseDB)
+}
```
noot, line 404: maybe we should have a separate `State` type for accessing things that aren't stored until the block, storage, etc. sub-databases, it would remove the need for baseDB. will open an issue

dot/system/service.go
```diff
@@ -35,11 +35,6 @@ func (s *Service) SystemName() string {
 	return s.systemInfo.SystemName
 }
 
-// ChainType returns the system's chain type
-func (s *Service) ChainType() string {
-	return s.systemInfo.ChainType
-}
-
```
noot, line 42: it might make more sense to leave this here, and in `dot` `createSystemService` pass in the database and read it from there and store it in `system.Service`. that was you wouldn't need to pass `storageState` into the rpc system module

dot/rpc/modules/system.go
```diff
@@ -73,8 +75,9 @@ func (sm *SystemModule) Name(r *http.Request, req *EmptyRequest, res *string) er
 
 // ChainType returns the chain type
 func (sm *SystemModule) ChainType(r *http.Request, req *EmptyRequest, res *string) error {
-	*res = sm.systemAPI.ChainType()
-	return nil
+	gd, err := sm.storageAPI.GetGenesisData()
+	*res = gd.ChainType
```
arijitAD, line 79: We should assign if err is `nil`

RyRy79261:
Receiving an error when running the test
image

dot/system/service.go
edwardmack: Yes, this does make more sense, done.

dot/rpc/modules/system.go
edwardmack: I moved initialization to dot createSystemService, and now I'm catching the error there.

dot/rpc/modules/api.go
```diff
@@ -19,6 +20,7 @@ type StorageAPI interface {
 	RegisterStorageChangeChannel(sub state.StorageSubscription) (byte, error)
 	UnregisterStorageChangeChannel(id byte)
 	GetStateRootFromBlock(bhash *common.Hash) (*common.Hash, error)
+	GetGenesisData() (*genesis.Data, error)
```
noot, line 23: is this still used in the rpc package?

dot/rpc/websocket_test.go
```diff
@@ -140,3 +141,6 @@ func (m *MockStorageAPI) UnregisterStorageChangeChannel(id byte) {
 func (m *MockStorageAPI) GetStateRootFromBlock(bhash *common.Hash) (*common.Hash, error) {
 	return nil, nil
 }
+func (m *MockStorageAPI) GetGenesisData() (*genesis.Data, error) {
```
noot, line 144: same here

noot (approved):
looks good! please cleanup interface

dot/rpc/modules/api.go
edwardmack: no, good catch, removed.

dot/rpc/websocket_test.go
edwardmack: removed.

Issue #1266, opened 2021-01-04
ensure execute_block works with Kusama and Polkadot blocks


noot:
## Expected Behavior

- currently execute_block is only tested with gossamer-made blocks
- need to ensure we can execute kusama/polkadot blocks
- steps: load the genesis file for the network into the state trie, get block 1 data from RPC, call execute_block with it and fix errors

related to #1210 

## Checklist

- [x] I have read [CODE_OF_CONDUCT](https://github.com/ChainSafe/gossamer/blob/development/.github/CODE_OF_CONDUCT.md) and [CONTRIBUTING](https://github.com/ChainSafe/gossamer/blob/development/.github/CONTRIBUTING.md) 
- [x] I have provided as much information as possible and necessary
- [ ] I am planning to submit a pull request to fix this issue myself

PR #1299, opened 2021-01-13
update codebase to use badger in-memory database 

noot:
## Changes

- update codebase to use badger in-memory database instead of `chaindb.MemDatabase`
- update `TrieState` to use badger in-memory db (since it is transient anyways), clean up `StorageState.StoreTrie` accordingly

## Tests

```
go test ./... -short
```

## Checklist

## Issues

- related to #1266


anthdm (approved):

dot/state/storage.go
```diff
@@ -92,7 +92,11 @@ func (s *StorageState) pruneKey(keyHeader *types.Header) {
 
 // StoreTrie stores the given trie in the StorageState and writes it to the database
 func (s *StorageState) StoreTrie(root common.Hash, ts *rtstorage.TrieState) error {
-	var err error
+	// write raw trie key-values stored in db to *trie.Trie
+	err := ts.Commit()
+	if err != nil {
+		logger.Error("failed to write TrieState db values to trie", "state root", root, "error", err)
```
arijitAD, line 98: Any reason we aren't returning errors?

go.mod
```diff
@@ -1,7 +1,7 @@
 module github.com/ChainSafe/gossamer
 
 require (
-	github.com/ChainSafe/chaindb v0.1.5-0.20210112170640-08c8f45c6f42
+	github.com/ChainSafe/chaindb v0.1.5-0.20210113205339-507b3a78c512
```
edwardmack, line 4: Should we make a tag on chaindb repo for this so we can require a specific release (instead of master)?

edwardmack (approved):
Great work, lgtm.

dot/state/storage.go
noot: it should be, will add!

go.mod
noot: hmm good idea, should I tag 0.1.0 at this commit?

edwardmack: Not sure, what's the difference between releases and tags?  I see we're at release 0.0.1, but the tags are at v0.1.4.  Maybe 0.1.5 or 0.2.0, are these updates breaking changes?

PR #1307, opened 2021-01-15
Added protected peers functionality

### LINKED ISSUE #1275: implement protected peers functionality
## Task summary

- there are peers who we wish to maintain connections to (ie. don't prune them when our peer count is over maximum)
- the ConnManager should be updated as to not disconnect from these peers (see `ConnManager.Connected`)
- a peer should be added to this list when we are syncing with them (they can be removed when syncing with them is completed)

## Specification reference

- n/a

- I believe substrate has this functionality implemented, can look there for some reference

anthdm:
Added protected peers functionality and prevent syncing peers from pruning when the maximum amount of connecting peers is exceeded.

## Changes

- Updated `ConnManager` so it can protect and unprotect peers
- Updated `ConnManager` so it will not prune protected peers
- Updated `Service` so it will use the `ConnManager` to flag and unflag syncing peers.

## Tests

```

```

## Checklist

## TODO
- Add test that checks if peers are flagged protected from the `Service`

## Issues

- closes #1275

### DIFF
--- a/dot/network/connmgr.go
+++ b/dot/network/connmgr.go
@@ -141,6 +141,7 @@ func (cm *ConnManager) unprotectedPeers(peers []peer.ID) []peer.ID {
 			unprot = append(unprot, id)
 		}
 	}
+
 	return unprot
 }
 
@@ -156,11 +157,13 @@ func (cm *ConnManager) Connected(n network.Network, c network.Conn) {
 	defer cm.Unlock()
 
 	if len(n.Peers()) > cm.max {
-		var (
-			unprotPeers = cm.unprotectedPeers(n.Peers())
-			// TODO: change to crypto/rand
-			i = rand.Intn(len(unprotPeers)) //nolint
-		)
+		unprotPeers := cm.unprotectedPeers(n.Peers())
+		if len(unprotPeers) == 0 {
+			return
+		}
+
+		// TODO: change to crypto/rand
+		i := rand.Intn(len(unprotPeers)) //nolint
 
 		logger.Trace("Over max peer count, disconnecting from random unprotected peer", "peer", unprotPeers[i])
 		err := n.ClosePeer(unprotPeers[i])
--- a/dot/network/connmgr.go
+++ b/dot/network/connmgr.go
@@ -34,13 +34,19 @@ type ConnManager struct {
 	max int // maximum number of peers
 	// closeHandlerMap contains close handler corresponding to a protocol.
 	closeHandlerMap map[protocol.ID]func(peerID peer.ID)
+
+	protectedPeerMapMu sync.RWMutex
+	// protectedPeerMap contains a list of peers that are protected from pruning
+	// when we reach the maximum numbers of peers.
+	protectedPeerMap map[peer.ID]struct{}
 	sync.Mutex
 }
 
 func newConnManager(max int) *ConnManager {
 	return &ConnManager{
-		max:             max,
-		closeHandlerMap: make(map[protocol.ID]func(peerID peer.ID)),
+		max:              max,
+		closeHandlerMap:  make(map[protocol.ID]func(peerID peer.ID)),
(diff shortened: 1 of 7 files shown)


RyRy79261 (approved):

arijitAD (commented):
.

dot/network/connmgr.go
```diff
@@ -73,18 +79,40 @@ func (*ConnManager) GetTagInfo(peer.ID) *connmgr.TagInfo { return &connmgr.TagIn
...
 
-// Unprotect peer
-func (*ConnManager) Unprotect(peer.ID, string) bool { return false }
+	cm.protectedPeerMap[id] = struct{}{}
+}
+
+// Unprotect peer will remove the given peer from prune protection.
+// returns true if we have successfully removed the peer from the
+// protectedPeerMap. False otherwise.
+func (cm *ConnManager) Unprotect(id peer.ID, tag string) bool {
```
arijitAD, line 94: Would it be good to rename `Unprotect ` to `UnprotectPeer` for more clarity?

arijitAD: Same for `Protect`

dot/network/connmgr.go
```diff
@@ -73,18 +79,40 @@ func (*ConnManager) GetTagInfo(peer.ID) *connmgr.TagInfo { return &connmgr.TagIn
...
+	cm.protectedPeerMapMu.Lock()
+	defer cm.protectedPeerMapMu.Unlock()
+
+	_, ok := cm.protectedPeerMap[id]
+	if ok {
+		delete(cm.protectedPeerMap, id)
```
arijitAD, line 100: Can you remove the return type? So we can directly call delete without checking for key first.

arijitAD (commented):
Having multiple locks in ConnManager would introduce additional complexity. Can you use the existing mutex from ConnManager?

dot/network/connmgr.go
noot: the functions have to be named what they are to satisfy the libp2p connection manager interface I believe

noot (commented):
great work!

could you also add the peer protection in `beginSyncing` and `handleBlockAnnounceMessage`? it would be nice to have both the `s.syncing` map set and `ConnMananger.Protect` in one function since they both happen together

 it would be nice to add a test that runs 2 network.Services and sends a BlockRequest message to the other via `beginSyncing` and makes sure they are protected

dot/network/connmgr.go
```diff
@@ -115,14 +154,18 @@ func (cm *ConnManager) Connected(n network.Network, c network.Conn) {
...
-		i := rand.Intn(len(n.Peers())) //nolint
-		peers := n.Peers()
-		logger.Trace("Over max peer count, disconnecting from random peer", "peer", peers[i])
-		err := n.ClosePeer(peers[i])
+		var (
+			unprotPeers = cm.unprotectedPeers(n.Peers())
+			// TODO: change to crypto/rand
+			i = rand.Intn(len(unprotPeers)) //nolint
+		)
+
```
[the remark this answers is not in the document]
anthdm: good catch

anthdm (commented):
@arijitAD If we use the existing mutex we will have a never-ending lock. I tested this before. Using a new RWMutex will let us lock and unlock it only for reading in most cases also.

anthdm:
> handleBlockAnnounceMessage

Good idea! Will update the PR with those suggestions.

anthdm:
> great work!
> 
> could you also add the peer protection in `beginSyncing` and `handleBlockAnnounceMessage`? it would be nice to have both the `s.syncing` map set and `ConnMananger.Protect` in one function since they both happen together
> 
> it would be nice to add a test that runs 2 network.Services and sends a BlockRequest message to the other via `beginSyncing` and makes sure they are protected

Good idea! I updated the PR with those suggestions.

noot (approved):
great work!

PR #1329, opened 2021-01-25
optimizations: add TODOs, make some functionality async

noot:
## Changes

- add TODO comments
- make digest handling async
- make writing state trie after executing block async
- 1 block now takes ~13s

## Tests

```
go test ./... -short
```

## Checklist

## Issues

-

### DIFF
--- a/dot/state/storage.go
+++ b/dot/state/storage.go
@@ -171,6 +171,9 @@ func (s *StorageState) StoreInDB(root common.Hash) error {
 	}
 
 	// notify subscribers of database changes
+	s.changedLock.Lock()
+	defer s.changedLock.Unlock()
+
 	for _, sub := range s.subscriptions {
 		subRes := &SubscriptionResult{
 			Hash: root,
@@ -188,7 +191,6 @@ func (s *StorageState) StoreInDB(root common.Hash) error {
 					subRes.Changes = append(subRes.Changes, *kv)
 				}
 			}
-
 		} else {
 			// filter result to include only interested keys
 			for k := range sub.Filter {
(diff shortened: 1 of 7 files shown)


edwardmack (approved):

dot/sync/syncer.go
```diff
@@ -469,14 +471,14 @@ func (s *Service) handleBlock(block *types.Block) error {
...
-		err = s.handleDigests(block.Header)
-		if err != nil {
-			return err
-		}
+		go func() {
+			err = s.handleDigests(block.Header)
+			if err != nil {
+				s.logger.Error("failed to handle block digest", "error", err)
+			}
+		}()
```
anthdm, line 481: Same here, can we just let the error do its thing?

dot/state/storage.go
```diff
@@ -114,13 +114,17 @@ func (s *StorageState) StoreTrie(root common.Hash, ts *rtstorage.TrieState) erro
...
+	go func() {
+		err = s.StoreInDB(root)
+		if err != nil {
+			logger.Error("failed to store encoded trie in database", "state root", root, "error", err)
+			return
+		}
 
-	logger.Trace("stored trie in database", "root", root)
+		logger.Trace("stored trie in database", "root", root)
+	}()
```
anthdm, line 127: Are we not depending on the possible error that `StoreInDB` could produce?

noot: we aren't really depending on it, this error generally shouldn't happen unless there's something wrong with the database, like it's been closed. if we want to handle the error we'd have to add some sort of err channel to get it out

dot/sync/syncer.go
noot: similar to below, if it errors it's not good but the node wouldn't crash or anything

RyRy79261 (approved):

PR #1350, opened 2021-02-02
dot/state: store trie nodes in database

### LINKED ISSUE #1343: dot/state: store trie nodes individually in database
## Task summary

- currently, once a block is imported, the entire state trie is encoded and stored in the database at the root hash as the key
- this is very inefficient since most of the keys don't change between blocks
- instead, the trie should be stored in the database by trie node
- key = hash of trie node; value = encoded trie node
- this way, the trie can be recursively reconstructed by looking up its root hash, decoding the trie node from there, and then looking up its children nodes by hash, and so on until the trie in reconstructed
- to update a trie value, only the node that is updated and its parents up to the root need to be updated

Note: trie node keys in database should be prefixed by some prefix (`trie` perhaps) to allow for the keys to have improved locality within the database

need to implement the following:
```
type Trie interface {
     // Store stores each trie node in the database, where the key is the hash of the encoded node and the value is the encoded node. Generally, this will only be used for the genesis trie.
     Store(db chaindb.Database) error

     // Load reconstructs the trie from the database from the given root hash. Used when restarting the node to load the current state trie.
     Load(db chaindb.Database, root common.Hash) error

     // PutInDB puts a value into the trie and writes the updates nodes the database. Since it needs to write all the nodes from the changed node up to the root, it writes these in a batch operation.
     PutInDB(db chaindb.Database, key, value []byte) error 

     // DeleteFromDB deletes a value from the trie and writes the updated nodes the database. Since it needs to write all the nodes from the changed node up to the root, it writes these in a batch operation.
     DeleteFromDB(db chaindb.Database, key []byte) error

     // ClearPrefixFromDB deletes all keys with the given prefix from the trie and writes the updated nodes the database. Since it needs to write all the nodes from the changed node up to the root, it writes these in a batch operation.
     ClearPrefixFromDB(db chaindb.Database, key []byte) error

     // GetFromDB retrieves a value from the trie using the database. It recursively descends into the trie using the database starting from the root node until it reaches the node with the given key. It then reads the value from the database.
     GetFromDB(db chaindb.Database, key []byte) ([]byte, error)
}
```

``` 
// StorageState is used to read and write storage tries from the database, as well as read storage values given a root hash.
// It should also store the current storage root at a database key. 
type StorageState struct {
    db chaindb.Database 
    currentTrie Trie // the current storage trie at the chain head
}

type StorageState interface {
    // Loads the current trie from the database by first reading the current storage root from the db, then reconstrucitng the trie. 
    LoadCurrentTrie() error 

     // Get returns a storage value given the trie root hash. It recursively descends into the trie using the database starting from the root node until it reaches the node with the given key. It then reads the value from the database.
    Get(root common.Hash, key []byte) ([]byte, error)

    // NewRuntimeStorage returns a new RuntimeStorage from the given storage root.  If the root is the current root, it uses the trie cached in the StorageState after copying it; otherwise, it (optionally) loads it from the database. (note: may not be needed to load the entire trie, could just pass in the database.)
    NewRuntimeStorage(root common.Hash) RuntimeStorage
}
```

```
type RuntimeStorage type {
      db chaindb.Database
      trie Trie
}

// RuntimeStorage is the interface used by the runtime to read and write the storage.
type RuntimeStorage interface {
      // Same as current runtime.Storage interface
}
```

see https://github.com/ChainSafe/gossamer/pull/1341 for preliminary benchmarking

noot:
## Changes

- fix trie dirty bit setting in Delete, Put, ClearPrefix
- implement `trie.Trie.Store` and `trie.Trie.Load` which store each node in the trie in the db
- implement `trie.Trie.WriteDirty` which writes only dirty nodes to db
- update `runtime/storage.TrieState` accordingly
- update `dot/state/storageState` accordingly
- update a lot of tests (sorry about chonkiness of PR)
- main changes are `lib/trie`, `lib/runtime/storage`, `dot/state`

## Tests

```
go test ./... -short
```

please see https://github.com/ChainSafe/gossamer/pull/1312 for kusama instructions, sync speed is improved to slightly less than 1s per block!

## Checklist

## Issues

- closes #1343


dot/state/service.go
```diff
@@ -149,10 +164,43 @@ func (s *Service) Initialize(data *genesis.Data, header *types.Header, t *trie.T
...
+	}
+
+	// load and store initial BABE epoch configuration
+	babeCfg, err := r.BabeConfiguration()
+	if err != nil {
+		return nil, fmt.Errorf("failed to fetch genesis babe configuration: %w", err)
+	}
+
+	r.Stop()
+
```
edwardmack, line 191: What is r.Stop() stopping?  I don't see when/where r was started?

dot/state/storage_notify_test.go
```diff
@@ -63,6 +63,7 @@ func TestStorageState_RegisterStorageChangeChannel(t *testing.T) {
 }
 
 func TestStorageState_RegisterStorageChangeChannel_Multi(t *testing.T) {
+	t.Skip()
```
edwardmack, line 66: Is this being skipped because we need to refactor how storage will notify about it's changes?  Is there an issue for this?

edwardmack (approved):
Great work, test sync with kssmc synced 100 blocks in 52 seconds!

dot/state/service.go
noot: `r` is the runtime instance, it's declared a few lines up: `r, err := wasmer.NewRuntimeFromGenesis(gen, rtCfg)`

dot/state/storage_notify_test.go
noot: woops forgot to uncomment this, at one point it wasn't working but I'll try to get it to work and re-add it. we will likely need to do a refactor of this at some point, since the trie structure and storing is likely to change

FlorianFranzen:
This seems to have broken the `ext_storage_clear_prefix_version_1` test, where keys not under the prefix are now cleared as well. This can be reproduced by running `gossamer-adapter host-api --function ext_storage_clear_prefix_version_1 --input sec,productivity,secondary,Total,Visionary` or
`gossamer-adapter host-api --function ext_storage_clear_prefix_version_1 --input disf,disintermediate,Grass-roots,policy,function`.

PR #1359, opened 2021-02-05
lib/runtime: add perlin/life support, fix runtime bug

### LINKED ISSUE #1126: lib/runtime: investigate memory corruption when using wasm interpreters
## Expected Behavior

- structures stored in memory should not be unexpectedly affected
- however, when executing a wasm call using `go-wasmer` or `go-wasmtime` (both go libraries that utilize cgo to bind to the underlying rust implementation`, the global variable `runtime.Context` gets affected, in particular, the `Storage`
- this causes runtime calls to unexpectedly fail due to invalid values
- since `Storage` is an interface, it can be backed by various implementations. both `trie.Trie` and `map[string][]byte` implements have the same problem. also mysterious is that the invalid values seem  to be the same for both these types

## Current Behavior

- however if a BadgerDB database backs `Storage`, this problem does not occur. (this is the current fix that is implemented into gossamer; see state.TrieState) however this causes inintended overhead, even when the writes are not being flushed to disk.
- this suggests that cgo is causing unintended effects on memory that should belong to the go process. since BadgerDB runs in a separate process, this may be what protects it. however this is just speculation and needs more investigation

## Possible Solution

- perahps run the `Storage` implementation in a separate process (perhaps goroutine)

## Checklist

noot:
## Changes

- add perlin/life support
- fix runtime bug (trie value wasn't copied so it was still pointing to a location in wasm memory)
- remove db usage for runtime storage backing, now we write the dirty nodes in the trie once execute_block is done
- syncing is around 10 blocks/second now!!

## Tests

```
go test ./... -short
```

## Checklist

## Issues

- closes #1126


RyRy79261:
Cant get the test to pass on my side. I ran it a couple times and ran out of ram the first few times.

image

noot:
can you run it with `go test ./... -short`? always run with `-short`

edwardmack (approved):
This looks great. it's syncing around 7 blocks a second on my machine, however it stops syncing after around 30 seconds with this: 
```
INFO[02-08|16:50:03] imported block                           pkg=sync number=254 hash=0x2de3bf2e4f49d7e5b2d0625de7a2612d6adb129daf5691d1f12790b627ef0f74 caller=syncer.go:468
EROR[02-08|16:50:03] Failed to read LEB128 encoding           pkg=network protocol=/ksmcc3/block-announces/1 error="stream reset" caller=service.go:426
EROR[02-08|16:50:03] Failed to read LEB128 encoding           pkg=network protocol=/ksmcc3/transactions/1 error="stream reset" caller=service.go:426
INFO[02-08|16:50:03] imported block                           pkg=sync number=255 hash=0xa9505fd8ff0e1fc72c5e231c3d90b1a52af6211d1c619293b2df8d6b3a92980c caller=syncer.go:468
INFO[02-08|16:50:04] imported block                           pkg=sync number=256 hash=0xe3eb7632fe5e2ef3560278d0ba76dc7ae0552058e591a82d9f029d758a8a6d6b caller=syncer.go:468
EROR[02-08|16:50:04] Failed to read LEB128 encoding           pkg=network protocol=/ksmcc3/sync/2 error="stream reset" caller=service.go:426
```

noot:
@edwardmack nice!! those issues should be fixed with #1365

lib/keystore/keyring.go
```diff
@@ -62,15 +62,18 @@ type Sr25519Keyring struct {
...
 
 // NewSr25519Keyring returns an initialized sr25519 Keyring
 func NewSr25519Keyring() (*Sr25519Keyring, error) {
 	kr := new(Sr25519Keyring)
+	kr.Keys = []*sr25519.Keypair{}
 
 	v := reflect.ValueOf(kr).Elem()
 
-	for i := 0; i < v.NumField(); i++ {
+	for i := 0; i < v.NumField()-1; i++ {
```
arijitAD, line 76: Why are we iterating one less?

lib/keystore/keyring.go
```diff
@@ -83,6 +86,8 @@ func NewSr25519Keyring() (*Sr25519Keyring, error) {
 		}
 
 		who.Set(reflect.ValueOf(kp))
+
+		kr.Keys = append(kr.Keys, kp)
```
arijitAD, line 90: We can preallocate the slice and then use the index instead of appending.
kr.Keys = make([]*sr25519.Keypair{}, v.NumField())
kr.Keys[i] = kp

lib/runtime/life/instance.go
```diff
@@ -0,0 +1,186 @@
...
+	}
+
+	// if cfg.LogLvl set to < 0, then don't change package log level
+	if cfg.LogLvl >= 0 {
+		h := log.StreamHandler(os.Stdout, log.TerminalFormat())
+		h = log.CallerFileHandler(h)
+		logger.SetHandler(log.LvlFilterHandler(cfg.LogLvl, h))
+	}
+
+	instance, err := exec.NewVirtualMachine(code, exec.VMConfig{
```
arijitAD, line 79: Is the formatting/indentation correct here?

lib/keystore/keyring.go
noot: the last field is an array of all the keys instead of `Alice`, `Bob`, etc

lib/runtime/life/instance.go
noot: yeah, i just indented the config while it's a param, i'll separate it out

lib/keystore/keyring.go
noot: updated!

RyRy79261:
Still running into issue a failed test in wasmer
image

noot:
@RyRy79261  the wasmer tests all pass for me and CI, that test is going to be deprecated anyways. can you re-run them?

lib/keystore/keyring.go
arijitAD: Wouldn't it be better to have a conditional skip?
```
if v.Field(i).Name == "Keys" {
    continue
}
```

arijitAD (approved):

PR #1373, opened 2021-02-09
Optimize Trie Snapshot and Benchmark against deep copy.

### LINKED ISSUE #1343: dot/state: store trie nodes individually in database

arijitAD:
## Changes

- Implement versioned trie and benchmark.

### Benchmark Results (Snapshot vs Deepcopy)
```
Number of iteration = 1000
```
#### Snapshot
```
=== RUN   TestDeepCopy/Snapshot
2021/02/09 20:09:25 
Alloc = 38 MB
TotalAlloc = 53 MB 
Sys = 71 MB 
NumGC = 4 
2021/02/09 20:09:25 Snapshot to trie took 31.693922ms
```
#### Deepcopy
```
=== RUN   TestDeepCopy/DeepCopy
2021/02/09 20:09:57
Alloc = 4869 MB
TotalAlloc = 4915 MB
Sys = 5093 MB
NumGC = 12
2021/02/09 20:09:57 DeepCopy to trie took 5.219532653s
```

## Tests

```
TestSnapshot
TestDeepCopyVsSnapshot
```

## Checklist

## Issues

- closes #1343


RyRy79261 (changes requested):
`TestDeepCopyVsSnapshot` times out: 
image

lib/trie/trie.go
```diff
@@ -65,6 +102,46 @@ func (t *Trie) DeepCopy() (*Trie, error) {
...
+			return bytes.Equal(tLeaf.value, trieLeaf.value)
+		}
+		return false
+	}
+	return false
+}
+
+// Equal returns true if both tries are same.
+func (t *Trie) Equal(trie *Trie) bool {
+	return isSimilar(t.root, trie.root)
```
noot, line 142: why is this needed? can we not just compare the root hashes?

lib/trie/trie.go
```diff
@@ -32,31 +32,68 @@ var EmptyHash, _ = NewEmptyTrie().Hash()
 // The zero value is an empty trie with no database.
 // Use NewTrie to create a trie that sits on top of a database.
 type Trie struct {
-	root     node
-	children map[common.Hash]*Trie
+	generation uint64
+	root       node
+	children   map[common.Hash]*Trie // Used to store the child trie.
```
noot, line 37: child tries*

edwardmack (approved):
Nice work, On my machine Snapshot is about 150 times faster than DeepCopy, but they both used the same memory:
```
=== RUN   TestDeepCopyVsSnapshot/DeepCopy
2021/02/09 17:24:27 
Alloc = 4868 MB 
TotalAlloc = 4914 MB 
Sys = 5093 MB 
NumGC = 12 

2021/02/09 17:24:27 DeepCopy to trie took 1.563547599s
    --- PASS: TestDeepCopyVsSnapshot/DeepCopy (1.56s)
=== RUN   TestDeepCopyVsSnapshot/Snapshot
2021/02/09 17:24:27 
Alloc = 4868 MB 
TotalAlloc = 4915 MB 
Sys = 5093 MB 
NumGC = 12 

2021/02/09 17:24:27 DeepCopy to trie took 9.643327ms
```

lib/trie/trie.go
```diff
@@ -225,30 +302,38 @@ func (t *Trie) Put(key, value []byte) error {
...
+		if p != nil && n != nil && n.isDirty() {
+			p.setDirty(true)
+		}
+		return n
 	case nil:
+		// We are creating new node so it will always have the latest generation.
+		// TODO: Sanity check. Can be removed later.
+		if t.generation != value.getGeneration() {
+			log.Error("node with invalid generation inserted")
+		}
```
noot, line 325: pls remove if no longer needed

lib/trie/trie.go
```diff
@@ -225,30 +302,38 @@ func (t *Trie) Put(key, value []byte) error {
...
 		switch v := value.(type) {
 		case *branch:
 			v.key = key
-			n = v
+			return v
```
noot, line 329: nice :D

dot/utils_test.go
```diff
@@ -86,3 +90,105 @@ func TestNewRuntimeFromGenesis(t *testing.T) {
 	_, err := wasmer.NewRuntimeFromGenesis(gen, &wasmer.Config{})
 	require.NoError(t, err)
 }
+
+func TestDeepCopyVsSnapshot(t *testing.T) {
```
noot, line 94: is it possible to change this into a test benchmark? eg. something like this https://golang.org/pkg/testing/#hdr-Benchmarks

noot (commented):
great work! nice cleanups :D i think you can probably remove the `trie.Equal` function - comparing root hashes should tell you if the tries are the same, since that's the point of being able to hash the trie.

lib/trie/trie.go
arijitAD: Silly me 😅 It skipped my mind that the trie is also a Merkle tree.

arijitAD:
> Nice work, On my machine Snapshot is about 150 times faster than DeepCopy, but they both used the same memory:
> 
> ```
> === RUN   TestDeepCopyVsSnapshot/DeepCopy
> 2021/02/09 17:24:27 
> Alloc = 4868 MB 
> TotalAlloc = 4914 MB 
> Sys = 5093 MB 
> NumGC = 12 
> 
> 2021/02/09 17:24:27 DeepCopy to trie took 1.563547599s
>     --- PASS: TestDeepCopyVsSnapshot/DeepCopy (1.56s)
> === RUN   TestDeepCopyVsSnapshot/Snapshot
> 2021/02/09 17:24:27 
> Alloc = 4868 MB 
> TotalAlloc = 4915 MB 
> Sys = 5093 MB 
> NumGC = 12 
> 
> 2021/02/09 17:24:27 DeepCopy to trie took 9.643327ms
> ```

I forgot to trigger garbage collection after each test. It should return the correct stats now.

dot/utils_test.go
arijitAD: We can but for loading genesis file the code is tightly coupled to use `testing.T` struct. We can modify the function to take `require.TestingT` interface as input but it will be major refactor. I can do this in a separate PR.

arijitAD:
> `TestDeepCopyVsSnapshot` times out:
> image

Screenshot 2021-02-10 at 5 03 17 PM
It works for me. Can you remove the timeout and use the above command?

lib/trie/trie.go
```diff
@@ -230,30 +260,34 @@ func (t *Trie) Put(key, value []byte) error {
...
 	k := keyToNibbles(key)
 
-	n, err := t.insert(t.root, k, &leaf{key: nil, value: value, dirty: true, valueDirty: true})
-	if err != nil {
-		return err
-	}
-
-	t.root = n
+	t.root = t.insert(t.root, k, &leaf{key: nil, value: value, dirty: true, valueDirty: true, generation: t.generation})
 	return nil
```
noot, line 264: since the error returned here is always nil, you can remove the return param

noot (approved):
awesome work! only 1 small comment, everything else looks great.

PR #1375, opened 2021-02-09
trie: fix ClearPrefix

### LINKED ISSUE #1353: trie.ClearPrefix does not correctly restructure trie after node deletion
## Task summary

- recently when testing this function I discovered a case where the deletion of nodes doesn't restructure the trie into the correct form
- the case is when deleting a node from a branch w/ only two children; a branch w/ only one child should become a leaf, merging with its only child (this happens in `trie.Delete`)
- however, `ClearPrefix` does not correctly restructure the trie after this happens; it leaves branches w/ only one child
- fix this so that the trie is structured properly in all cases after `ClearPrefix`

(runtime storage uses the `trie.Delete` function instead of `trie.ClearPrefix` right now for prefix clearing because of this)

noot:
## Changes

- fix `trie.ClearPrefix` so that is restructures trie correctly (branches cannot have 0 or 1 children after deletion)

## Tests

```
go test ./lib/trie -run TestClearPrefix
```

## Checklist

## Issues

- closes #1353


edwardmack (approved):
Nice work, lgtm.

RyRy79261 (approved):

PR #1385, opened 2021-02-11
dot/network: implement syncing queue

### LINKED ISSUE #1370: syncing: create queue of block responses to handle
## Task summary

- sub-tasks of #1134 
- currently, block requests are sent after the previous block response is handled, this creates a delay in between handling one block response to the next
- instead, since the node already knows from the start what block it needs to sync to, it should send out many block requests at the start, and queue the block responses for handling so that there is no delay between block responses

noot:
## Changes

- implement syncing queue to allow for block requests to be sent out and block responses to be queued so that there is no delay in between handling responses
- move creation of `BlockRequests` from `sync` package to `network` package so that the queue can easily create/modify them

## Tests

 run `./bin/gossamer --chain ksmcc --bootnodes /ip4/3.235.193.225/tcp/30333/p2p/12D3KooWLLvGJZWF7z6KaCw6CPPbPkccrkCHmQBch27vtfngcZki`. (re-init if needed) (turn on network debug logs if you want to see the request/response queue action)

it may take a minute for the node to start receiving messages. you should see that there is no delay in between responses being handled and that messages are being queued. (after a few minutes it seems all our peers start ignoring us, will open an issue for that)

## Checklist

## Issues

- closes #1370


RyRy79261:
Does this need to be from an authorized ip? Receiving a failed to dial error

noot:
> Does this need to be from an authorized ip? Receiving a failed to dial error

@RyRy79261 yes it does, are you using the IP you sent me before? if it doesn't work pls try with a local kusama node

edwardmack (approved):
Great work.

RyRy79261 (approved):
Imports running as expected my side,

Received a `Failed to read LEB128 encoding` from `/ksmcc3/sync/2` at the end, not sure if this is an issue though

noot:
@RyRy79261 yeah that's fine, it just means a peer closed the stream

Issue #1349, opened 2021-02-01
Implement versioned trie and benchmark it


arijitAD:
## Issue summary

- Implement Persistent trie and benchmark it

## Other information and links

- #1343

PR #1386, opened 2021-02-12
Implement delete and clear prefix for versioned trie.

arijitAD:
## Changes

- Implement delete and clear prefix for versioned trie.

## Tests

```
trie/trie_test.go
```

## Checklist

## Issues

- #1349

### DIFF
--- a/lib/trie/trie.go
+++ b/lib/trie/trie.go
@@ -435,7 +435,10 @@ func (t *Trie) getKeysWithPrefix(parent node, prefix, key []byte, keys [][]byte)
 		key = key[len(p.key):]
 		keys = t.getKeysWithPrefix(p.children[key[0]], append(append(prefix, p.key...), key[0]), key[1:], keys)
 	case *leaf:
-		keys = append(keys, nibblesToKeyLE(append(prefix, p.key...)))
+		length := lenCommonPrefix(p.key, key)
+		if bytes.Equal(p.key[:length], key) || len(key) == 0 {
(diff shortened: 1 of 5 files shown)


lib/trie/trie.go
```diff
@@ -550,73 +555,79 @@ func (t *Trie) clearPrefix(curr node, prefix []byte) (n node, wasUpdated bool) {
...
 }
 
-func (t *Trie) delete(parent node, key []byte) (n node, err error) {
+func (t *Trie) delete(parent node, key []byte) (node, bool) {
 	switch p := parent.(type) {
 	case *branch:
-		length := lenCommonPrefix(p.key, key)
+		// Store the current node and return it, if the trie is not updated.
+		currP := p
+		p = t.maybeUpdateBranchGeneration(p)
```
noot, line 585: `currP` and `p` are both pointers to the same node, so if you update data in `p` then `currP` gets updated also, does this result in the path's generation being updated even if nothing is deleted?

arijitAD: `currP` and `p` will be different if the generation is updated. So as we travel down the path we create a copy of every node but then if nothing is deleted we should discard the copy and return the original node. Hence, I am storing the original node in a temporary variable `currP`. 

If `p` and `currP`  both belong to the same generation then they will point to the same node and it will follow the old code logic. In this case the nodes won't be updated if there is nothing to delete.

noot (approved):
great work! :D feel free to update `lib/runtime/storage.go` `DeepCopy` to `Snapshot` if you'd like

RyRy79261 (approved):

Issue #1134, opened 2020-10-15
Sync Kusama Network


amerameen:
## Expected Behavior

- Connect to Kusama network and sync to the head

## Current Behavior

- N/A

## Checklist

- [x] I have read [CODE_OF_CONDUCT](https://github.com/ChainSafe/gossamer/blob/development/.github/CODE_OF_CONDUCT.md) and [CONTRIBUTING](https://github.com/ChainSafe/gossamer/blob/development/.github/CONTRIBUTING.md) 
- [x] I have provided as much information as possible and necessary
- [ ] I am planning to submit a pull request to fix this issue myself

andyhowell8:
Are there any details on what is blocking this issue? 

Id like to follow along with the development of this feature for Kusama & Polkadot networks

noot:
hey @andyhowell8 , glad to see the interest! the issues that are currently blocking this are #1264 (updating encoding for some network messages) and #1266 (ensuring we can correctly sync the state).

andyhowell8:
Great thanks for the reply, am I correct in saying that the current implementation as it stands will not connect Kusama or Polkadot p2p via the respective boot nodes?

I've done a quick test and noticed that the node never appears to get peers. I'm working on a Masters research project using go-lipp2p to connect to Kusama nodes to explore the DHT, I manage to connect successfully but never seem to exchange peers.

On researching the issue I stumbled across this project and the p2p implementation is rather similar, hence my interest in this open issue. I won't pollute this issue with more unrelated conversations by will keep an eye on the progress.

noot:
@andyhowell8 it should be able to connect, however it likely won't be able to stay connected. the peer will be expecting certain messages to be exchanged, and if they aren't exchanged, they will disconnect from us. currently this seems to happen within <10s or so with our implementation. 

I'm not sure if you've seen this substrate documentation: https://crates.parity.io/sc_network/index.html it may be helpful as it outlines the protocols that substrate nodes use.

I'm working on connecting (and staying connected) currently and added some DHT advertising and discovery code here: https://github.com/ChainSafe/gossamer/pull/1269 I tried it with kusama yesterday and it never seems to reach the point of being able to advertise or discover due to being unable to stay connected. This will be our primary focus for a while so we should have some progress on it soon. Feel free to email me if you'd like to discuss further: elizabeth (at) chainsafe (dot) io

PR #1389, opened 2021-02-12
lib/runtime: fix ext_crypto_secp256k1_ecdsa_recover_version_1

noot:
## Changes

 
- fix `ext_crypto_secp256k1_ecdsa_recover_version_1`, can now execute kusama block 3784
- comment out `ext_crypto_sr25519_verify_version_1` failure case for now, blocks kusama block 3876 (will open issue for this)

## Tests

```
go test ./lib/runtime/wasmer
```

## Checklist

## Issues

- related to #1134


RyRy79261 (approved):

FlorianFranzen:
I guess this would be a good time to implement the crypto tests in out host api testsuite for gossasmer. I will get stated on that this week.

lib/runtime/wasmer/imports.go
```diff
@@ -371,6 +354,16 @@ func ext_crypto_secp256k1_ecdsa_recover_version_1(context unsafe.Pointer, sig, m
...
 	signature := memory[sig : sig+65]
 
+	if signature[64] == 27 {
+		signature[64] = 0
+	}
+
+	if signature[64] == 28 {
+		signature[64] = 1
+	}
+
```
edwardmack, line 364: Why are we doing this?

edwardmack (approved):

lib/runtime/wasmer/imports.go
noot: the last byte of the signature can be either `0/1` or `27/28` depending on what signed it, see spec: `sig: a 32-bit pointer to the buffer containing the 65-byte signature in RSV format. V should be either 0/1 or 27/28.`

Issue #257, opened 2019-09-11
Transactions Network Message


amerameen:
Please see Appendix E, Section E.1.5.

Take a look at the implementations of Block Request and Block Response.

PR #1403, opened 2021-02-19
dot/network: update syncing based on cpu profiling 

noot:
## Changes

- main issue found when cpu profiling was that we were allocating a large chunk of data every time we wanted to read from a stream
- updated `readStream` to only allocate once when it's first called
- update `syncQueue` to syncronously write request to stream then read response from same stream (makes way more sense), `syncQueue` now has a buffer it uses to read stream data into

## Tests

run `./bin/gossamer --chain ksmcc --bootnodes /ip4/3.235.193.225/tcp/30333/p2p/12D3KooWLLvGJZWF7z6KaCw6CPPbPkccrkCHmQBch27vtfngcZki` , you should see that syncing is now faster and doesn't slow down as the node runs. 

if you run `top` while the node is running you will also see that cpu usage should be much improved since before (on my machine, was ~300% before, now is 50-70%)

## Checklist

## Issues

- related to #1134


RyRy79261:
I'm running into an error when it try loads block #257 
Is this expected?

image

noot:
@RyRy79261 yes that error is unrelated it just means some peer closed a stream with us

RyRy79261 (approved):

dot/network/sync_test.go
```diff
@@ -335,6 +262,8 @@ func TestSyncQueue_HandleBlockAnnounce(t *testing.T) {
 }
 
 func TestSyncQueue_ProcessBlockRequests(t *testing.T) {
+	//t.Skip() // TODO: fix this
+
```
edwardmack, line 266: Does this still need to be fixed?

edwardmack (approved):
Great work, mine was syncing at around 17 bk/sec, then got "stuck"  Have you seen this:
```
INFO[02-23|10:13:00] imported block                           pkg=sync number=1916 hash=0x312c872ad0df36f4bd74b3ad394a0aa27b168fb0938a64c790b68e0ed6c2b6b8 caller=syncer.go:314
INFO[02-23|10:13:00] imported block                           pkg=sync number=1917 hash=0x77f0341c8cf37bf43f925a4635b5f4a37714ae59c46c6dfb7e4dabb68699cbdc caller=syncer.go:314
CRIT[02-23|10:13:01] [ext_logging_log_version_1]              pkg=runtime module=go-wasmer target=runtime message="panicked at 'Timestamp must increment by at least <MinimumPeriod> between sequential blocks', /home/volt/.cargo/git/checkouts/substrate-7e08433d4c370a21/d9fca7e/frame/timestamp/src/lib.rs:147:4" caller=imports.go:134
WARN[02-23|10:13:01] failed to handle block data; re-adding to queue pkg=network start=0 end=0 error="failed to execute block 1425: Failed to call the `Core_execute_block` exported function." caller=sync.go:444
INFO[02-23|10:13:13] 🚣 currently syncing                   pkg=network average blocks/second=3.600 overall average=15.966 caller=sync.go:225
INFO[02-23|10:13:28] 🚣 currently syncing                   pkg=network average blocks/second=0.067 overall average=14.199 caller=sync.go:225
INFO[02-23|10:13:43] 🚣 currently syncing                   pkg=network average blocks/second=0.067 overall average=12.786 caller=sync.go:225

```

noot:
@edwardmack yeah I have seen that, it seems to happen when trying to execute a block that's already been executed. not sure exactly why it's that error🤔 but I'm gonna work on some improvements today to prevent that from happening

dot/network/sync_test.go
noot: oh that is already fixed I forgot to remove that, will remove

PR #1420, opened 2021-02-26
dot/network: remove unneeded map access, sync fixes

noot:
## Changes

- remove unneeded map access that was causing concurrent map read/write err

## Tests

```
go test ./dot/network -short
```

## Checklist

## Issues

-

### DIFF
--- a/dot/core/service.go
+++ b/dot/core/service.go
@@ -222,7 +222,7 @@ func (s *Service) StorageRoot() (common.Hash, error) {
 
 func (s *Service) handleBlocks(ctx context.Context) {
 	for {
-		prev := s.blockState.BestBlockHash()
+		//prev := s.blockState.BestBlockHash()
 
 		select {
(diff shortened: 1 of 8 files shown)


edwardmack (approved):

RyRy79261 (approved):

dot/core/service.go
```diff
@@ -222,7 +222,7 @@ func (s *Service) StorageRoot() (common.Hash, error) {
 
 func (s *Service) handleBlocks(ctx context.Context) {
 	for {
-		prev := s.blockState.BestBlockHash()
+		//prev := s.blockState.BestBlockHash()
```
arijitAD, line 225: Are the comments intentional?

arijitAD (approved):

dot/core/service.go
noot: yes this needs to be fixed and re-enabled. i will open an issue for it

Issue #1363, opened 2021-02-05
handle justifications received from block responses


noot:
## Task summary

- currently, we request block finalization justification, but we don't verify it and mark the block finalized if valid
- need to check the justification and mark block and all ancestors finalized if it's valid

PR #1428, opened 2021-03-02
dot/sync: handle block response justification 

noot:
## Changes

- handle justifications provided in block responses by setting latest finalized hash (will use grandpa to verify justification in a follow-up PR, wanted to get this in first since it allows the blocktree to be pruned, improving sync times)
- on node restart, if there was an "abnormal shutdown" ie the blocktree wasn't stored properly because the node panicked or was killed forcefully, then restore the state from the latest finalized hash
- update logging to log blocks imported every 5s (if you liked the previous logs, you can turn on `sync` package debug logs):
```
INFO[03-02|03:10:06] 🔨 finalized block                     pkg=sync number=1094832 hash=0xb7d4a3c8a51dd5e189c2aec82dffec97f91f2f09ce1736878aa985ec108eb41f caller=syncer.go:353
INFO[03-02|03:10:11] 🔗 imported blocks                     pkg=network from=1094832 to=1094918 hashes="[0xb7d4a3c8a51dd5e189c2aec82dffec97f91f2f09ce1736878aa985ec108eb41f ... 0xd7fd20dbe3646d2a52e3e82963ac00462cb6bfd95627a4f667903628f79000df]" caller=sync.go:275
INFO[03-02|03:10:11] 🚣 currently syncing                   pkg=network goal=6426917 average blocks/second=17.199 overall average=10.906 caller=sync.go:279
INFO[03-02|03:10:16] 🔗 imported blocks                     pkg=network from=1094918 to=1095008 hashes="[0xd7fd20dbe3646d2a52e3e82963ac00462cb6bfd95627a4f667903628f79000df ... 0x71192a673ef180eabf2df8c4baf8d0335e32cfbeda8f1bddaf6221aea0dc7774]" caller=sync.go:275
INFO[03-02|03:10:16] 🚣 currently syncing                   pkg=network goal=6426917 average blocks/second=17.988 overall average=11.087 caller=sync.go:279
```

## Tests

```
go test ./dot/state
go test ./dot/sync -short
```

as usual please try to run a kusama node and see that blocks are being marked as finalized

## Checklist

## Issues

- related to #1363

### DIFF
--- a/dot/network/sync.go
+++ b/dot/network/sync.go
@@ -72,7 +72,7 @@ func (s *Service) handleSyncMessage(stream libp2pnetwork.Stream, msg Message) er
 
 		resp, err := s.syncer.CreateBlockResponse(req)
 		if err != nil {
-			logger.Debug("cannot create response for request")
+			logger.Trace("cannot create response for request")
 			return nil
 		}
 
@@ -164,7 +164,7 @@ func (q *syncQueue) handleRequestQueue() {
 			continue
 		}
 
-		logger.Debug("sync request queue", "queue", q.stringifyRequestQueue())
+		logger.Trace("sync request queue", "queue", q.stringifyRequestQueue())
 		head := q.requests[0]
 		q.requests = q.requests[1:]
 		q.requestCh <- head
@@ -191,14 +191,14 @@ func (q *syncQueue) handleResponseQueue() {
 		}
 
 		if q.responses[0].Number().Int64() > head.Int64()+1 {
-			logger.Debug("response start isn't head+1, waiting", "queue start", q.responses[0].Number().Int64(), "head+1", head.Int64()+1)
+			logger.Trace("response start isn't head+1, waiting", "queue start", q.responses[0].Number().Int64(), "head+1", head.Int64()+1)
 			q.responseLock.Unlock()
 
 			q.setBlockRequests("")
 			continue
 		}
 
-		logger.Debug("sync response queue", "queue", q.stringifyResponseQueue())
+		logger.Trace("sync response queue", "queue", q.stringifyResponseQueue())
 		q.responseLock.Unlock()
 		q.responseCh <- q.responses
 		q.responses = []*types.BlockData{}
@@ -353,7 +353,7 @@ func (q *syncQueue) setBlockRequests(to peer.ID) {
 		start = head.Int64() + 1
 	}
 
-	logger.Debug("setting block request queue", "start", start, "goal", q.goal)
+	logger.Trace("setting block request queue", "start", start, "goal", q.goal)
 
 	reqs := createBlockRequests(start, q.goal)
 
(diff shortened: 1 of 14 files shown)


dot/network/sync.go
```diff
@@ -249,33 +251,46 @@ func (q *syncQueue) benchmark() {
 			return
 		}
 
-		head, err := q.s.blockState.BestBlockNumber()
+		before, err := q.s.blockState.BestBlockHeader()
 		if err != nil {
-			logger.Error("failed to get best block number", "error", err)
+			logger.Error("failed to get best block haeder", "error", err)
```
RyRy79261, line 256: Spelling error

dot/network/sync.go
```diff
@@ -249,33 +251,46 @@ func (q *syncQueue) benchmark() {
...
-		q.benchmarker.begin(head.Uint64())
-		time.Sleep(time.Second * 10)
+		q.benchmarker.begin(before.Number.Uint64())
+		time.Sleep(time.Second * 5)
 
-		head, err = q.s.blockState.BestBlockNumber()
+		after, err := q.s.blockState.BestBlockHeader()
 		if err != nil {
-			logger.Error("failed to get best block number", "error", err)
+			logger.Error("failed to get best block haeder", "error", err)
```
RyRy79261, line 269: Spelling error

RyRy79261 (approved):
Just the two spelling error but the code looks good

dot/network/sync.go
```diff
@@ -352,7 +367,7 @@ func (q *syncQueue) setBlockRequests(to peer.ID) {
 	newReqs = sortRequests(newReqs)
 	q.requests = newReqs
```
arijitAD, line 368: Acquire lock

arijitAD (commented):
Wouldn't it be simpler to use channels directly to implement `syncQueue` so we don't have to worry about locks. The current code has a couple of data races.

dot/network/sync.go
noot: there's currently no request lock

noot:
@arijitAD what do you mean use channels directly? 

btw, I would rather not modify the sync queue in this PR since this PR was for handling justifications and it might cause other issues. I can address the issues in another PR

arijitAD:
> @arijitAD what do you mean use channels directly?
> 
> btw, I would rather not modify the sync queue in this PR since this PR was for handling justifications and it might cause other issues. I can address the issues in another PR

Makes sense. We can definitely refactor this later.

```go
	requests  []*syncRequest // start block of message -> full message
	requestCh chan *syncRequest

	responses    []*types.BlockData
	responseCh   chan []*types.BlockData
```
Currently, we are adding the data to the `requests` slice and then moving it to the `requestCh` later. Can we directly add the `syncRequest` to the `requestCh`? That way we won't be needing locks since channels are threadsafe.

Also, we should have locks for `requests` since we are modifying it in `setBlockRequests` and reading it in `handleRequestQueue`.

noot:
@arijitAD I see, I guess the reason I made it so that the requests are separate from the channel are in the case we need to modify the requests while they are queued / before they are sent out.  I think it could work with a buffered channel as well though, we would just need to know what's in the buffer somehow. I think the queue could definitely be improved so I'll address this in a follow up PR!

edwardmack (approved):

PR #1432, opened 2021-03-02
dot/network: sync queue improvements, correctly set blocktree root after prune

noot:
## Changes

- improve sync queue by removing the `requests` slice, instead put requests directly into buffered channel 
- only send requests in `blockRequestSize` increments (ie. request starts must be 1, 129, 257, etc...) and use a map to track whether the request has been sent and whether response has been received or not. this way duplicate requests won't be sent out
- correctly set blocktree root after prune, node slowdown was still happening due to incorrectly keeping all the nodes from genesis

## Tests

```
go test ./dot/network -short
go test ./lib/blocktree
```

also, please run kusama node `./bin/gossamer --chain ksmcc` you should see that the syncing speed stays relatively stable now (around 15bps on my machine, 20bps on AWS)

## Checklist

## Issues

- related to #1134

### DIFF
--- a/dot/state/block_test.go
+++ b/dot/state/block_test.go
@@ -297,8 +297,18 @@ func TestLatestFinalizedRound(t *testing.T) {
 func TestFinalization_DeleteBlock(t *testing.T) {
 	bs := newTestBlockState(t, testGenesisHeader)
 	AddBlocksToState(t, bs, 5)
+
+	btBefore := bs.bt.DeepCopy()
+	t.Log(btBefore)
 	before := bs.bt.GetAllBlocks()
 	leaves := bs.Leaves()
+
+	// TODO: why isn't arrival time set?
+	// for _, n := range before {
+	// 	has, err := bs.HasArrivalTime(n)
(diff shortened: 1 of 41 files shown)


dot/state/block_test.go
```diff
@@ -297,15 +297,26 @@ func TestLatestFinalizedRound(t *testing.T) {
 func TestFinalization_DeleteBlock(t *testing.T) {
 	bs := newTestBlockState(t, testGenesisHeader)
 	AddBlocksToState(t, bs, 5)
+
+	btBefore := bs.bt.DeepCopy()
+	t.Log(btBefore)
 	before := bs.bt.GetAllBlocks()
 	leaves := bs.Leaves()
 
+	// TODO: why isn't arrival time set?
```
noot, line 306: will open an issue for this, it's currently not used anywhere so it's ok to ignore for now i think

RyRy79261 (approved):

edwardmack (approved):
Great work, I'm syncing at 18bps on my machine.

PR #1443, opened 2021-03-07
dot/state: update rewind to forcefully set blocktree root, update flag param to be block to rewind to

noot:
## Changes

- update rewind to forcefully set blocktree root (previously could only rewind to last finalized block, not before that)
- update flag param to be block to rewind to

## Tests

```
go test ./dot/state -run TestService_Rewind
```

## Checklist

## Issues

- realted to #1134


dot/state/service.go
```diff
@@ -321,13 +322,22 @@ func (s *Service) Start() error {
 	return nil
 }
 
-// Rewind rewinds the chain by the given number of blocks.
+// Rewind rewinds the chain to the given block number.
 // If the given number of blocks is greater than the chain height, it will rewind to genesis.
-func (s *Service) Rewind(numBlocks int) error {
+func (s *Service) Rewind(to int64) error {
```
arijitAD, line 327: Rename the variable `to` to `toBlockNumber` or something more meaningful.

arijitAD (approved):

dot/state/service.go
noot: renamed to `toBlock`

RyRy79261 (approved):
+1 for the rename from Arijit

PR #1449, opened 2021-03-10
fix: lib/trie: fix trie.NextKey

### LINKED ISSUE #1439: fix and enable trie.NextKey
## Task summary

- `trie.NextKey` does not correctly handle the case when the key parameter is not in the trie, it doesn't return the next key, instead returns nil
- fix this and re-enable it in `TrieState`

noot:
## Changes

- re-implement `trie.NextKey` to work properly (and handle keys not in the trie)
- utilize `trie.NextKey` in runtime storage
- tested w/ kusama blocks that previous broke and works, also greatly increases the speed

## Tests

```
go test ./lib/trie
go test ./lib/runtime/wasmer
```

## Checklist

## Issues

- closes #1439


RyRy79261 (approved):

arijitAD:
I have added a test for the next trie key which seems to fail.

[Test Commit](https://github.com/ChainSafe/gossamer/pull/1449/commits/c19b72117e7256f47af3ece9e598406be7d8a397
)

Also, the test seems to fail if I add an empty string to the test case. Not sure if it's a valid scenario.

noot:
@arijitAD I made an update so that there are no duplicate keys in the test list and looks like it's good now!

arijitAD (approved):
Approved. Nice work  👌

PR #1451, opened 2021-03-11
fix: lib/runtime: update sign_ funcs to return fixed size byte optional

noot:
## Changes

- update sign_ functs to return fixed size byte optional
- update `life` not to error if `core_version` doesn't exist (needed for host API tester)

## Tests

```
go test ./lib/runtime/wasmer/ -run Test_ext_crypto_ed25519_sign_version_1
go test ./lib/runtime/wasmer/ -run Test_ext_crypto_sr25519_sign_version_1
go test ./lib/runtime/life
```

## Checklist

## Issues

related to #1105

### DIFF
--- a/lib/common/optional/types.go
+++ b/lib/common/optional/types.go
@@ -20,6 +20,7 @@ import (
 	"encoding/binary"
 	"fmt"
 	"io"
+	"io/ioutil"
 	"math/big"
 
 	"github.com/ChainSafe/gossamer/lib/common"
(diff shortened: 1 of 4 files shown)


FlorianFranzen:
This does indeed fix the encoding issue and the returned signature pass verification. 

The life issue however still exists, as the call to a none-existing `Version()`  still causes a panic. Maybe calling `Version()` similar to how it is done on the wasmer backend (e.g. when the version field is needed but nil and not on initialization) would be the easiest solution.

RyRy79261 (approved):

arijitAD (approved):

Issue #1400, opened 2021-02-18
fix polkadot.js UI accounts


noot:
## Task summary

- currently, we are unable to make new accounts with the polkadot.js UI:

```
balances.transferKeepAlive
-32000: storage Account not found within module System: null
```

- as well, we are unable to see balances and thus send transactions from accounts defined in the genesis file (although they do show up)

edwardmack:
Closed, confirmed balances.transfer works.

PR #1460, opened 2021-03-13
fix (dot/rpc, dot/state): state_subscribeStorage to only notify for value changes

edwardmack:
## Changes

- fix state_subscribeStorage to only notify subscribers if the value has changed, previously it was notifying subscribers on every block built.
- fix, to notify subscribers when they make their request, so subscriber has current value immediately, previously subscriber had to wait until block was built to be notified of initial value, or not notified at all if node is not block builder.
-

## Tests

```
go test ./dot/state/... -short
```

## Checklist

## Issues

- related to #1400


RyRy79261 (approved):

dot/state/storage.go
```diff
@@ -172,18 +173,24 @@ func (s *StorageState) notifyStorageSubscriptions(root common.Hash) error {
 			}
 		} else {
 			// filter result to include only interested keys
-			for k := range sub.Filter {
+			for k, currentValue := range sub.Filter {
```
noot, line 176: maybe name this `cachedValue`?

dot/state/storage_notify.go
```diff
@@ -60,6 +60,16 @@ func (s *StorageState) RegisterStorageChangeChannel(sub StorageSubscription) (by
...
+	// notifyStorageSubscriptions here to send storage value of current state
+	sr, err := s.blockState.BestBlockStateRoot()
+	if err != nil {
+		logger.Debug("error registering storage change channel", "error", err)
+	}
+	go func() {
+		if err := s.notifyStorageSubscriptions(sr); err != nil {
+			logger.Warn("failed to notify storage subscriptions", "error", err)
+		}
+	}()
```
noot, line 72: why is this moved here? it seems like this would cause every subscription to be notified when a new subscription is registered, and only for the current state root?

dot/rpc/websocketListeners.go
```diff
@@ -0,0 +1,371 @@
...
+	rvl.subID = c.qtyListeners
+	c.subscriptions[rvl.subID] = rvl
+	initRes := newSubscriptionResponseJSON(rvl.subID, reqID)
+	c.safeSend(initRes)
+
+	return rvl.subID, nil
+}
+
+// Listen implementation of Listen interface to listen for runtime version changes
+func (l *RuntimeVersionListener) Listen() {
```
arijitAD, line 355: This always sends the runtime version even if the version is not updated. Is this the expected behaviour?

dot/rpc/websocketListeners.go
```diff
@@ -0,0 +1,371 @@
...
+type BlockFinalizedListener struct {
+	channel chan *types.Header
+	wsconn  *WSConn
+	chanID  byte
+	subID   int
+}
+
+func (c *WSConn) initBlockFinalizedListener(reqID float64) (int, error) {
+	bfl := &BlockFinalizedListener{
+		channel: make(chan *types.Header),
```
arijitAD, line 189: Why are all subscription channels unbuffered?

dot/state/storage_notify.go
edwardmack: Right, I put this here so that subscriptions are notified when a new subscription is created so that the requester get's the current value when they subscribe.  It would cause all subscriptions to be notified except that subscriptions were values have not changed will not notify.  I'd like to find a cleaner way to do this, but calling notifyStorageSubscriptions seems like the easiest way to do this.  Other suggestions?

dot/state/storage_notify_test.go
```diff
@@ -120,14 +119,18 @@ func TestStorageState_RegisterStorageChangeChannel_Multi_Filter(t *testing.T) {
...
 	num := 5
 	chs := make([]chan *SubscriptionResult, num)
 	ids := make([]byte, num)
-	subFilter := make(map[string]bool)
-	subFilter[common.BytesToHex(key1)] = true
 
 	for i := 0; i < num; i++ {
 		chs[i] = make(chan *SubscriptionResult)
+		subFilter := make(map[string][]byte)
+		subFilter[common.BytesToHex(key1)] = []byte{}
```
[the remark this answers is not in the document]
edwardmack: No, moving this to outside the loop causes the test to fail.  Each StorageSubscription holds it's own filter map, which holds cached values for the key.  When a value is updated, the filter map get updated to that value, if a second subscription uses the shared map, the second subscription will not notify because it's value won't be new.

dot/rpc/websocketListeners.go
edwardmack: Yes, this is to give the subscriber the current value when they first subscribe,  I need to add logic to send updates when the runtime changes.

dot/rpc/websocketListeners.go
edwardmack: No reason, I can add a buffer,  any suggested size?  256?

dot/state/storage_notify.go
```diff
@@ -60,6 +60,16 @@ func (s *StorageState) RegisterStorageChangeChannel(sub StorageSubscription) (by
 	s.changedLock.Lock()
 	s.subscriptions[id] = &sub
 	s.changedLock.Unlock()
+	// notifyStorageSubscriptions here to send storage value of current state
```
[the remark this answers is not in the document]
edwardmack: @arijitAD I agree this could be done better.  I don't understand your suggestion, I need to learn more about buffered channels.  Are you suggesting replacing the subscription map with a buffered chan of type StorageSubscription?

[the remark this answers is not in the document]
arijitAD: [Fixed-size buffer pool](https://github.com/oxtoacart/bpool/blob/master/bufferpool.go) We can use something like this.
I have created an issue for this. We can fix this later. https://github.com/ChainSafe/gossamer/issues/1477

dot/state/storage_notify_test.go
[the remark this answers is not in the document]
arijitAD: You can write the code like shown below as well.
This couples your map with the struct. No need to declare an additional variable.
```
sub := StorageSubscription{
	Filter:   map[string][]byte{
		common.BytesToHex(key1): {},
	},
	Listener: chs[i],
}
```

dot/rpc/websocketListeners.go
arijitAD: Yes. I think 256 would work fine. We need to make all the listener channels buffered. 
Unbuffered channels are usually slow due to their blocking nature. Here is a [benchmark report](https://tpaschalis.github.io/channels-limitations-speed/) for buffered vs unbuffered channels.
```
BenchmarkBufferedChannelEmptyStruct-8         23657084            49.9 ns/op
BenchmarkBufferedChannelOneByte-8             21230530            54.6 ns/op      18.31 MB/s

BenchmarkUnbufferedChannelEmptyStruct-8        6075384            177 ns/op
BenchmarkUnbufferedChannelOneByte-8            6341457            184 ns/op       5.44 MB/s
```
I have created a separate issue for this. https://github.com/ChainSafe/gossamer/issues/1478

dot/rpc/websocketListeners.go
```diff
@@ -0,0 +1,368 @@
...
+		if change == nil {
+			continue
+		}
+
+		result := make(map[string]interface{})
+		result["block"] = change.Hash.String()
+		changes := make([][]string, len(change.Changes))
+		for _, v := range change.Changes {
+			kv := []string{common.BytesToHex(v.Key), common.BytesToHex(v.Value)}
+			changes = append(changes, kv)
```
arijitAD, line 110: `append` adds an element to the end of the array. `changes` slice will be 2x size.

```
func main() {
	const arrSize = 5

	// Incorrect code: Final length will be 10.
	vals := make([]int, arrSize) // Creates [0,0,0,0,0]
	for i := 0; i < 5; i++ {
		vals = append(vals, i)
	}
	fmt.Println(len(vals))
	fmt.Println(vals)

	// Output:
	// 10
	// [0 0 0 0 0 0 1 2 3 4]

	// Solution 1 (Preferred): Pre declare length = 0 and capacity = 5. Creates []
	vals = make([]int,0, arrSize)
	for i := 0; i < 5; i++ {
		vals = append(vals, i)
	}
	fmt.Println(len(vals))
	fmt.Println(vals)

	// Output:
	// 5
	// [0 1 2 3 4]

	// Solution 2:
	vals = make([]int, arrSize) // Creates [0,0,0,0,0]
	for i := 0; i < 5; i++ {
		vals[i] = i
	}
	fmt.Println(len(vals))
	fmt.Println(vals)
	// Output:
	// 5
	// [0 1 2 3 4]
}
```

arijitAD (approved):

dot/state/storage_notify.go
noot: maybe you could pull the logic that notifies one subscription from the `notifyStorageSubscriptions` function  into another function and call that for just this subscription?

edwardmack: good idea, done.

edwardmack:
I've added sync.Mutex to Trie struct to fix data race, however now the tests (`go test ./dot/state/... -race`) fail with timeout, so I'm not sure I actually fixed the race, but created a deadlock.  I'm also not happy with the architecture I've created regarding websockets.  For example, having websocket_test.go outside subscription package doesn't feel right, but when I move it into the subscription package it breaks due to cyclic package references.  Any suggestions welcome. thanks.

noot:
@edwardmack I'll try to take a closer look sometime today!

arijitAD:
@edwardmack One issue I see is that you are trying to take the same lock twice.
`notifyStorageSubscriptions` locks `changedLock` mutex and calls `notifyStorageSubscription`.
`notifyStorageSubscription` internally tries to take lock on the same mutex and this results in deadlock.
```
func (s *StorageState) notifyStorageSubscriptions(root common.Hash) error {
	s.changedLock.Lock()
	defer s.changedLock.Unlock()
	for subKey := range s.subscriptions {
		err := s.notifyStorageSubscription(root, subKey)
		if err != nil {
			return err
		}
	}
	return nil
}

func (s *StorageState) notifyStorageSubscription(root common.Hash, subID byte) error {
	s.lock.RLock()
	t := s.tries[root]
	defer s.lock.RUnlock()

	if t == nil {
		return errTrieDoesNotExist(root)
	}

	// notify subscribers of database changes
	s.changedLock.Lock()
	defer s.changedLock.Unlock()

```

noot:
@edwardmack in terms of architecture, maybe you could move the `websockets.go` file in the `subscription` package back into the `dot/rpc` package, since it seems like the `listeners.go` file in `subscription` uses an interface for the `WSConn` already, do it doesn't need to import `dot/rpc`

dot/rpc/subscription/listeners_test.go
```diff
@@ -0,0 +1,157 @@
...
+// (at your option) any later version.
+//
+// The gossamer library is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with the gossamer library. If not, see <http://www.gnu.org/licenses/>.
+package subscription
```
arijitAD, line 16: Newline

dot/rpc/subscription/listeners.go
```diff
@@ -30,121 +30,61 @@ type Listener interface {
...
+// Update is called to notify observer of new value
+func (s *StorageObserver) Update(change *state.SubscriptionResult) {
+	if change == nil {
+		return
 	}
 
-	chanID, err := c.StorageAPI.RegisterStorageChangeChannel(*sub)
-	if err != nil {
-		return 0, err
+	result := make(map[string]interface{})
```
arijitAD, line 51: Replace this with `struct`. 
```
type Result struct {
   changes []Change
   block string
}
```

dot/rpc/subscription/listeners.go
```diff
@@ -30,121 +30,61 @@ type Listener interface {
...
+	result["block"] = change.Hash.String()
+	changes := make([][]string, 0, len(change.Changes))
```
arijitAD, line 53: Rename `Changes` to `Change` or better, since this represents a single change.

dot/state/storage_notify.go
```diff
@@ -33,55 +34,105 @@ type SubscriptionResult struct {
...
+					Key:   common.MustHexToBytes(fmt.Sprintf("0x%x", k)),
+					Value: v,
+				}
+				subRes.Changes = append(subRes.Changes, *kv)
+			}
+		}
+	} else {
+		// filter result to include only interested keys
+		for k, cachedValue := range o.GetFilter() {
+			value := t.Get(common.MustHexToBytes(k))
```
arijitAD, line 105: We would require a lock while reading from trie.

dot/state/storage_notify.go
```diff
@@ -33,55 +34,105 @@ type SubscriptionResult struct {
...
+	return nil
+}
+
+func (s *StorageState) removeFromSlice(observerList []Observer, observerToRemove Observer) []Observer {
+	s.changedLock.Lock()
+	defer s.changedLock.Unlock()
+	observerListLength := len(observerList)
+	for i, observer := range observerList {
+		if observerToRemove.GetID() == observer.GetID() {
+			observerList[observerListLength-1], observerList[i] = observerList[i], observerList[observerListLength-1]
```
arijitAD, line 133: Nice way to remove an element from the slice.

Replace the line with this. We don't care about the last element anyways.
```
observerList[i] = observerList[observerListLength-1]
```

arijitAD (changes requested):

arijitAD:
Nice refactor

dot/rpc/subscription/listeners.go
noot: I assume it needs to be in this format because it's what the JSON-RPC format is expecting? correct me if I'm wrong @edwardmack

dot/rpc/subscription/websocket.go
```diff
@@ -27,11 +27,13 @@ import (
...
 
 	"github.com/ChainSafe/gossamer/dot/rpc/modules"
+	"github.com/ChainSafe/gossamer/dot/types"
+	"github.com/ChainSafe/gossamer/lib/common"
 	log "github.com/ChainSafe/log15"
 	"github.com/gorilla/websocket"
 )
 
-var logger = log.New("pkg", "subscription")
+var logger = log.New("pkg", "rpc")
```
noot, line 36: maybe `rpc/subscription` as the package?

dot/rpc/subscription/websocket.go
```diff
@@ -161,3 +163,206 @@ func (c *WSConn) HandleComm() {
...
+	esl.importedChanID, err = c.BlockAPI.RegisterImportedChannel(esl.importedChan)
+	if err != nil {
+		return 0, err
+	}
+
+	esl.finalizedChanID, err = c.BlockAPI.RegisterFinalizedChannel(esl.finalizedChan)
+	if err != nil {
+		return 0, err
+	}
+
```
noot, line 289: just curious, done submitAndWatchExtrinsic notify both on the block being built as well as finalized?

dot/rpc/subscription/websocket.go
```diff
@@ -161,3 +163,206 @@ func (c *WSConn) HandleComm() {
...
+	c.BlockSubChannels[esl.subID] = esl.importedChanID
+
+	err = c.CoreAPI.HandleSubmittedExtrinsic(extBytes)
+	if err != nil {
+		return 0, err
+	}
+	c.safeSend(newSubscriptionResponseJSON(esl.subID, reqID))
+
+	// TODO (ed) since HandleSubmittedExtrinsic has been called we assume the extrinsic is in the tx queue
+	//  should we add a channel to tx queue so we're notified when it's in the queue
```
noot, line 302: could you open an issue for this?

dot/rpc/websocket_test.go
```diff
@@ -90,6 +94,7 @@ func TestHTTPServer_ServeHTTP(t *testing.T) {
 
 		_, message, err := c.ReadMessage()
 		require.Nil(t, err)
+		fmt.Printf("ACT MSG %s\n", message)
```
noot, line 97: remove?

dot/state/storage.go
```diff
@@ -70,11 +70,10 @@ func NewStorageState(db chaindb.Database, blockState *BlockState, t *trie.Trie)
 	tries[t.MustHash()] = t
 
 	return &StorageState{
-		blockState:    blockState,
-		tries:         tries,
-		baseDB:        db,
-		db:            chaindb.NewTable(db, storagePrefix),
-		subscriptions: make(map[byte]*StorageSubscription),
```
noot, line 77: does observerList need to be initialized? ie `[]Observer{}` I assume not since the tests work?

dot/state/storage_notify.go
noot: alternatively you can use a `TrieState` which is threadsafe, you can do `s.TrieState(root)`

dot/state/storage_notify.go
```diff
@@ -33,55 +34,105 @@ type SubscriptionResult struct {
...
+	}
 }
 
-func (s *StorageState) notifyChanged(change *SubscriptionResult) {
-	if len(s.subscriptions) == 0 {
-		return
+func (s *StorageState) notifyObserver(root common.Hash, o Observer) error {
+	s.lock.RLock()
+	t := s.tries[root]
+	defer s.lock.RUnlock()
```
noot, line 80: replace w `s.TrieState(root)`

dot/state/storage_notify.go
```diff
@@ -33,55 +34,105 @@ type SubscriptionResult struct {
...
-		}(ch.Listener)
+	return nil
+}
+
+func (s *StorageState) removeFromSlice(observerList []Observer, observerToRemove Observer) []Observer {
+	s.changedLock.Lock()
+	defer s.changedLock.Unlock()
+	observerListLength := len(observerList)
+	for i, observer := range observerList {
+		if observerToRemove.GetID() == observer.GetID() {
```
noot, line 132: are unique IDs guaranteed?

dot/state/storage_notify.go
```diff
@@ -33,55 +34,105 @@ type SubscriptionResult struct {
...
 }
 
-//StorageSubscription holds data for Subscription to Storage
-type StorageSubscription struct {
-	Filter   map[string]bool
-	Listener chan<- *SubscriptionResult
+// Observer interface defines functions needed for observers, Observer Design Pattern
+type Observer interface {
+	Update(result *SubscriptionResult)
+	GetID() int
```
noot, line 40: bit of a nitpick but maybe change int to uint?

lib/trie/trie.go
```diff
@@ -242,6 +242,7 @@ func (t *Trie) tryPut(key, value []byte) {
 	k := keyToNibbles(key)
 
 	t.root = t.insert(t.root, k, &leaf{key: nil, value: value, dirty: true, generation: t.generation})
+
```
noot, line 245: remove

noot (commented):
great work, really nice refactoring, just a few comments/questions, also I assume you are testing this w/ a real ws connection, could you share the calls you are using over ws for testing?

dot/rpc/subscription/listeners.go
arijitAD: In that case, we can use struct tags.

dot/rpc/subscription/listeners_test.go
edwardmack: Added.

dot/rpc/subscription/listeners.go
edwardmack: @noot yes, I did this for JSON-RPC formatting, and yes using a struct is more readable.  I've added that with tags.

dot/rpc/subscription/listeners.go
edwardmack: updated (using new struct).

dot/state/storage_notify.go
edwardmack: Updated to use TrieState

dot/state/storage_notify.go
edwardmack: all set.

dot/rpc/subscription/websocket.go
edwardmack: changed.

dot/rpc/subscription/websocket.go
edwardmack: Yes, it has a "inBlock" status for when the extrinsic has been included in block, and "finalized" status for when the extrinsic has been finalized by GRANDPA.

dot/rpc/subscription/websocket.go
edwardmack: created issue #1535

dot/rpc/websocket_test.go
edwardmack: removed.

dot/state/storage.go
edwardmack: add initialization

dot/state/storage_notify.go
edwardmack: done.

dot/state/storage_notify.go
edwardmack: I using a counter that is incremented each time an observer is created.

dot/state/storage_notify.go
edwardmack: updated.

lib/trie/trie.go
edwardmack: done.

dot/state/storage_notify.go
```diff
@@ -33,55 +34,106 @@ type SubscriptionResult struct {
...
+		logger.Debug("error registering storage change channel", "error", err)
 	}
-
-	var id byte
-	for {
-		id = generateID()
-		if s.subscriptions[id] == nil {
-			break
+	go func() {
+		if err := s.notifyObserver(sr, o); err != nil {
```
noot, line 54: will not returning cause an error since `sr` will be undefined?

noot (approved):
looks good!

dot/state/storage_notify.go
edwardmack: good point, I added a return.

arijitAD (approved):

PR #1466, opened 2021-03-15
fix:dot/network: fix grandpa protocol ID

noot:
## Changes

- previously all stream protocols were prefixed with the "base" protocol ID for example `/ksmcc3`, however the `grandpa` protocol is not prefixed, the protocol ID is just `/paritytech/grandpa/1`
- this will be updated in the future but for now it is not correct, see https://github.com/paritytech/substrate/issues/7252
- this PR allows notifications protocols to be registered with a protocol ID that "overwrites" ie. doesn't prepend the base protocol
- updates network functions to use the full protocol ID 

## Tests

```
go test ./dot/network -short
```

## Issues

- related to #1134


RyRy79261 (approved):

dot/network/host.go
```diff
@@ -276,32 +286,10 @@ func (h *host) writeToStream(s libp2pnetwork.Stream, msg Message) error {
...
-	for _, p := range h.peers() {
-		if p != peer {
-			err := h.send(p, "", msg)
-			if err != nil {
-				logger.Error("Failed to send message during broadcast", "peer", p, "err", err)
-			}
-		}
-	}
-}
-
```
edwardmack, line 300: I'm assuming these are being removed because they're not needed, is there no longer a need to broadcast messages?

edwardmack (approved):

dot/network/host.go
noot: yes, they weren't actually being used anywhere. notifications messages are still being "broadcast" (gossiped) in the `broadcastExcluding` function in notifications.go

PR #1470, opened 2021-03-16
chore: rename ksmcc -> kusama

noot:
## Changes

- rename ksmcc -> kusama 

## Tests

```
go test ./cmd/...
make gossamer
./bin/gossamer --chain kusama
```

## Issues

- n/a


chain/kusama/defaults.go
```diff
@@ -27,11 +27,11 @@ var (
 	// DefaultName Default node name
 	DefaultName = string("Kusama")
 	// DefaultID Default chain ID
-	DefaultID = string("ksmcc")
+	DefaultID = string("ksmcc3")
```
RyRy79261, line 30: Why the suffix of 3 here?

RyRy79261 (approved):
Only an inquisitory note

chain/kusama/defaults.go
noot: it's what's in the genesis file

edwardmack (approved):

Issue #1354, opened 2021-02-03
update StorageState to load storage value from database if trie is not cached


noot:
## Task summary

- for state rpc functions, we may wish to retrieve storage at a previous block that isn't cached in the StorageState
- currently the function only returns values that are cached in the trie map in `StorageState`
- in this case, we should fetch the value from the database using the given root hash
- eg: `StorageState.GetStorage(someStorageRoot, somekey)` should call `trie.GetFromDB(storageState.db, someStorageRoot, somekey)`
- same for other StorageState functions that may access previous states

PR #1486, opened 2021-03-22
fix(state) : Update StorageState to load storage from database.

arijitAD:
## Changes

- Update `StorageState` to load storage from the database.

## Tests

```

```

## Issues

- #1354

### DIFF
--- a/dot/state/storage.go
+++ b/dot/state/storage.go
@@ -204,19 +204,9 @@ func (s *StorageState) LoadFromDB(root common.Hash) (*trie.Trie, error) {
 
 // ExistsStorage check if the key exists in the storage trie with the given storage hash
 // If no hash is provided, the current chain head is used
-func (s *StorageState) ExistsStorage(hash *common.Hash, key []byte) (bool, error) {
-	if hash == nil {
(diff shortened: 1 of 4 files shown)


dot/state/storage.go
```diff
@@ -270,11 +260,15 @@ func (s *StorageState) StorageRoot() (common.Hash, error) {
...
+	t, ok := s.tries[sr]
+	if !ok {
+		t, err = s.LoadFromDB(sr)
+		if err != nil {
+			return common.Hash{}, errTrieDoesNotExist(sr)
+		}
 	}
 
-	return s.tries[sr].Hash()
+	return t.Hash()
```
noot, line 271: this function can be updated to just `return s.blockState.BestBlockStateRoot()` since it should be returning the current state root

noot (commented):
looks good, please update `StorageRoot` and add unit tests!

RyRy79261 (approved):

dot/state/storage.go
arijitAD: Done

noot (approved):

Issue #1476, opened 2021-03-17
improve gossamer memory usage


noot:
## Task summary

- when the node runs for a long time the memory usage gets quite high (~50% on a node that is currently syncing at 3-4mil blocks, 90+% on a node trying to sync at the head for a while!!)
- this needs to be improved, especially the performance when trying to stay synced
- profile and fix areas that are causing high memory usage

PR #1491, opened 2021-03-24
fix(dot/state,dot/network): improve memory usage when syncing

noot:
## Changes

- improve memory usage when syncing (not at head) by only keeping current state trie in memory
- improve memory usage by increasing garbage collection (setting garbage collection percentage to 10, can be overwritten with GOGC env var)

## Tests

```
todo
```

also, run node on development vs. this branch and check out memory usage difference

## Issues

- related to #1476

### DIFF
--- a/dot/network/sync.go
+++ b/dot/network/sync.go
@@ -249,7 +249,12 @@ func (q *syncQueue) handleResponseQueue() {
 // prune peers with low score and connect to new peers
 func (q *syncQueue) prunePeers() {
 	for {
-		time.Sleep(time.Second * 30)
+		select {
+		case <-time.After(time.Second * 30):
+		case <-q.ctx.Done():
+			return
+		}
+
 		logger.Debug("✂️ pruning peers w/ low score...")
 
 		peers := q.getSortedPeers()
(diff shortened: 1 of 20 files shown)


RyRy79261 (approved):

dot/network/sync.go
```diff
@@ -192,10 +197,7 @@ func (q *syncQueue) syncAtHead() {
 		prev = curr
 		start := uint64(curr.Number.Int64()) + 1
 		logger.Debug("haven't received new blocks since last check, pushing request", "start", start)
-		q.requestData.Store(start, requestData{
-			sent:     true,
-			received: false,
-		})
+		q.requestData.Delete(start)
```
edwardmack, line 200: I'm not sure I understand what's happening here.  It seems we're deleting start from the requestData map, but the pushing it back on the map?  It seems to be working, but I don't fully understand.

edwardmack (approved):

dot/network/sync.go
noot: the requestData tracks requests that were sent and received, in the case we are at the head it seems like often we will send out a request and receive an empty response if the node doesn't have info for that block yet. the request processor ignores requests that have already gotten a response, so we need to delete the info for it that says we got a request first. I noticed the node would go into a loop of trying to send a request but thinking it already got the response and thus never syncing

PR #1492, opened 2021-03-25
fix(dot/network): update notificationsProtocol handshakeData to sync.Map

noot:
## Changes

- update notificationsProtocol handshakeData to sync.Map since there was a concurrent map read and write panic w/ the map

## Tests

```
go test ./dot/network -short
```

## Issues

- related to #1134


RyRy79261 (approved):

edwardmack (approved):
Great work, I wasn't aware of sync.Map, looks very useful, I need to keep this is mind, probably other places I could replace maps with this.  Nice!

Issue #1496, opened 2021-03-30
Generate random name when --name flag is not used


edwardmack:
## Describe the bug

- Currently when starting the gossamer node, the node's name defaults to "Polkadot" if the `--name` flag is not used.  

## Expected Behavior

- When the `--name` flag is not used the node should generate a random name (so that the node can be uniquely identified).  The ParityTech Polkadot nodes generate a name by compining two random words followed by a 4 digit number. For example: `cuddly-rake-4006`

## Current Behavior

- Currently Gossamer node uses default name 'Polkadot' if the '--name' flag is not used.

## Possible Solution

- Add function to generate random name when `--name` flag is not used.

## To Reproduce

1. Start gossamer node without name flag.
2. Observe name defaults to "Polkadot"

## Log output

<details>
  <summary>Log Output</summary>
  
  ```
INFO[03-30|14:15:44] 🕸️ initializing node services...    pkg=dot name=Polkadot id=polkadot basepath=/home/emack/.gossamer/polkadot caller=node.go:180
  ```
</details> 
</br>

## Specification

- go version:
- gossamer version: 0.3.2
- gossamer commit tag:
- gossamer commit hash:
- operating system:
- additional links:

Issue #1429, opened 2021-03-02
telemetry.polkadot.io compatibility


noot:
## Task summary

- we should get gossamer nodes to show on telemetry.polkadot.io
- it seems the connection is via websockets 

## Other information and links

- for substrate, it is handled by the telemetry crate: https://github.com/paritytech/substrate/tree/master/client/telemetry
- on connection, this message is sent: https://github.com/paritytech/substrate/blob/master/client/telemetry/src/lib.rs#L100
- server side payload messages defined here: https://github.com/paritytech/substrate-telemetry/blob/master/backend/src/node/message.rs#L40
- look for the telemetry! macro throughout the substrate code base to see what is logged

PR #1497, opened 2021-03-30
feat: dot/telemetry: Implement basic telemetry connection

edwardmack:
## Changes

- Implement code for node to connect to telemetry endpoints defined in genesis.json `telemetryEndpoints` parameter.
- Implement message `system.connected` to notify telemetry endpoint node has connected.
- Implement message `block.import` to notify telemetry endpoint that node has imported a block.

## Related remaining TODO:
- Implement telemetry levels, Polkadot uses telemetry level values from 0 - 9 to define how much telemetry information the node will send, this needs to be researched and implemented.
- Implement remaining messages, there are additional telemetry messages the telemetry server understands, research and implement.
- Implement `--telemetry-url` flag to enable connecting to additional telemetry endpoints.

## Tests

Start gossamer node, use `--name` flag so that you can identify the node that connected.  Without name flag the node's default name is `Polkadot`, (see issue #1496)
```
./bin/gossamer --chain polkadot --name 'Test Gossamer Node'
```
Open browser to https://telemetry.polkadot.io and you'll see the Test Gossamer Node listed on the Polkadot tab.

## Issues

- issue: #1429


lib/genesis/genesis.go
```diff
@@ -22,29 +22,31 @@ import (
...
-	Genesis         Fields                 `json:"genesis"`
-	Properties      map[string]interface{} `json:"properties"`
-	ForkBlocks      []string               `json:"forkBlocks"`
-	BadBlocks       []string               `json:"badBlocks"`
-	ConsensusEngine string                 `json:"consensusEngine"`
+	Name               string                 `json:"name"`
+	ID                 string                 `json:"id"`
+	ChainType          string                 `json:"chainType"`
+	Bootnodes          []string               `json:"bootNodes"`
+	TelemetryEndpoints []interface{}          `json:"telemetryEndpoints"`
```
edwardmack, line 29: I'd rather not use `[]interface{}`, I'd rather make a struct like 
```
type TelemetryEndpoint struct {
	Endpoint string
	Level    float64
}
```
which I tried, but this caused an error when with `json.Unmarshal`  when decoding (from polkadot genesis-raw.json): 
```
"telemetryEndpoints": [
    [
      "wss://telemetry.polkadot.io/submit/",
      0
    ]
  ],
```
Any suggestions on how I can better handle this?

dot/node.go
```diff
@@ -298,6 +300,13 @@ func NewNode(cfg *Config, ks *keystore.GlobalKeystore, stopFunc func()) (*Node,
 		publishMetrics(cfg)
 	}
 
+	gd, err := stateSrvc.Storage.GetGenesisData()
+	if !cfg.Global.NoTelemetry && err == nil {
```
noot, line 304: add an err check case before this?

dot/rpc/modules/api.go
```diff
@@ -90,7 +90,7 @@ type RuntimeAPI interface {
 type SystemAPI interface {
 	SystemName() string
```
noot, line 91: is "SystemName" now the name of the node?

dot/telemetry/telemetry.go
```diff
@@ -0,0 +1,108 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to marshal fields to JSON, %w", err)
+	}
+	return append(serialized, '\n'), nil
+}
+
+var once sync.Once
+var handlerInstance *Handler
+
+// GetInstance signleton pattern to for accessing TelemeterHandler
```
noot, line 54: singleton*

dot/telemetry/telemetry.go
```diff
@@ -0,0 +1,108 @@
...
+//  log level, line number and timestamp
+func (f *MyJSONFormatter) Format(entry *log.Entry) ([]byte, error) {
+	serialized, err := json.Marshal(entry.Data)
+	if err != nil {
+		return nil, fmt.Errorf("failed to marshal fields to JSON, %w", err)
+	}
+	return append(serialized, '\n'), nil
+}
+
+var once sync.Once
```
noot, line 51: cool :D

dot/telemetry/telemetry.go
```diff
@@ -0,0 +1,108 @@
...
+				}
+				log.SetOutput(&handlerInstance.buf)
+				log.SetFormatter(new(MyJSONFormatter))
+			})
+	}
+	return handlerInstance
+}
+
+//AddConnections adds connections to telemetry sever
+func (h *Handler) AddConnections(conns []interface{}) {
```
noot, line 70: space before AddConnections pls

lib/genesis/genesis.go
noot: it seems like the endpoints are tuples/arrays (use `[ ]` instead of `{ }`) and I think Go expects `[ ]` to be an `[]interface{}` instead of a `struct`, so not sure if there's a better way to do this..

noot (approved):
works for me! even shows location :0 great work!!

noot:
@edwardmack oen thing is it doesn't show the peer count on telemetry. could you open an issue for that?

dot/rpc/modules/api.go
edwardmack: SystemName is from ctx.App.Name, and is used for RPC system_name
ChainName is from genesis.json 'name' parameter, and is used for RPC sysytem_chain
Node name shown is telemetry is from cfg.Global.Name, from '--name' flag or a random name (after #1496 is done)

dot/telemetry/telemetry.go
edwardmack: Fixed

lib/genesis/genesis.go
edwardmack: Right, it's working, so I guess this is the way to go.

dot/node.go
edwardmack: Do you mean `if err == nil && !cfg.Global.NoTelemetry {`  or put the err check in a separate block?  I have them together because both must be true before I can add that endpoint connection.

noot: yeah i mean add an err check case before:
```
gd, err := stateSrvc.Storage.GetGenesisData()
if err != nil {
    return nil, err
}

if cfg.Global.NoTelemetry {
    return node, nil 
}

// logic inside current block here, since we know  !cfg.Global.NoTelemetry && err == nil is true at this point
```

edwardmack:
@noot yes, created issue #1501 for implementing remaining telemetry messages which should fix this.

dot/node.go
edwardmack: Ok, yes this is more readable.

cmd/gossamer/config.go
```diff
@@ -433,6 +434,8 @@ func setDotGlobalConfig(ctx *cli.Context, tomlCfg *ctoml.Config, cfg *dot.Global
 		cfg.MetricsPort = uint32(metricsPort)
 	}
 
+	cfg.NoTelemetry = ctx.Bool("no-telemetry")
```
arijitAD, line 437: Instead of `no-telemetry` wouldn't it be better if the flag is named `telemetry`?

dot/telemetry/telemetry.go
arijitAD, line 51: Add them in single var block.
```
var (
    once sync.Once
    handlerInstance *Handler
)
```

dot/telemetry/telemetry.go
```diff
@@ -0,0 +1,108 @@
...
+		c, _, err := websocket.DefaultDialer.Dial(fmt.Sprintf("%s", v.([]interface{})[0]), nil)
+		if err != nil {
+			fmt.Printf("Error %v\n", err)
+		}
+		h.wsConn = append(h.wsConn, c)
+	}
+}
+
+// SendConnection sends connection request message to telemetry connection
+func (h *Handler) SendConnection(authority bool, chain, genesis_hash, system_name, node_name,
```
arijitAD, line 81: Ideally, keep function params size limited to 3. If there are more params group them in a struct.

dot/telemetry/telemetry.go
```diff
@@ -0,0 +1,108 @@
...
+		return nil, fmt.Errorf("failed to marshal fields to JSON, %w", err)
+	}
+	return append(serialized, '\n'), nil
+}
+
+var once sync.Once
+var handlerInstance *Handler
+
+// GetInstance singleton pattern to for accessing TelemetryHandler
+func GetInstance() *Handler {
```
arijitAD, line 55: Nice use to singelton design pattern.

dot/telemetry/telemetry.go
arijitAD: Also, param names should be camel case. `genesis_hash` -> `genesisHash`
Fix in all functions.

dot/telemetry/telemetry.go
```diff
@@ -0,0 +1,108 @@
...
+				}
+				log.SetOutput(&handlerInstance.buf)
+				log.SetFormatter(new(MyJSONFormatter))
+			})
+	}
+	return handlerInstance
+}
+
+// AddConnections adds connections to telemetry sever
+func (h *Handler) AddConnections(conns []interface{}) {
```
arijitAD, line 70: Why is the param type interface? Interface is handy when the params could be of any type.
But here the params type is fixed.

arijitAD (commented):
What is the expected behaviour for telemetry flag? By default, telemetry seems to be enabled.

cmd/gossamer/config.go
edwardmack: I'm replicating the behavior of the paritytech polkadot node, which has telemetry enabled by default, and a `no-telemetry` flag to disable this.

dot/telemetry/telemetry.go
edwardmack: all set

dot/telemetry/telemetry.go
edwardmack: I updated this to use a struct as parameter.  I was using []interface{} because this is the result from unmarshalling the genesis.json, but yes it makes more sense to use this value as a struct instead.

dot/telemetry/telemetry.go
edwardmack: I created a struct to hold the connection data information.  And changed from snake case to camel case.

lib/genesis/genesis.go
```diff
@@ -41,14 +41,20 @@ type Data struct {
 	ID                 string
 	ChainType          string
 	Bootnodes          [][]byte
-	TelemetryEndpoints []interface{}
+	TelemetryEndpoints []TelemetryEndpoint
```
arijitAD, line 44: Change this to pointer.
`TelemetryEndpoints []*TelemetryEndpoint`

lib/genesis/genesis.go
```diff
@@ -97,3 +103,28 @@ func (g *Genesis) ToRaw() error {
...
 }
+
+func interfaceToTelemetryEndpoint(endpoints []interface{}) []TelemetryEndpoint {
+	var res []TelemetryEndpoint
+	for _, v := range endpoints {
+		epi, ok := v.([]interface{})
+		if !ok {
+			continue
+		}
+		eps, ok := epi[0].(string)
```
arijitAD, line 114: Verfiy `len(epi) == 2`

lib/genesis/genesis.go
```diff
@@ -97,3 +103,28 @@ func (g *Genesis) ToRaw() error {
 	g.Genesis.Raw["top"] = res
 	return nil
 }
+
+func interfaceToTelemetryEndpoint(endpoints []interface{}) []TelemetryEndpoint {
```
arijitAD, line 107: return `[]*TelemetryEndpoint`

arijitAD (approved):
Looks Good!

lib/genesis/genesis.go
edwardmack: done.

lib/genesis/genesis.go
edwardmack: Added check.

lib/genesis/genesis.go
edwardmack: done.

Issue #738, opened 2020-04-01
network: improve gossip algorithm by ensuring duplicate messages aren't sent to peers


noot:
currently, the gossip algorithm is flooding, so sending the message to every peer except the one that it was received from. this should be improved as to reduce network bandwidth. one method of improving the gossip mechanism is to choose a number of peers to send the message to, say *n*, then randomly  select n peers out of our total peers. there are also other algorithms that could be used, will need some research and exploration. could make n configurable for now to determine *best* (ie. trade-off between bandwidth and propagation) value

p4u:
Hello.
I am following your developments on Gossamer, you have a great base code which I think the global FOSS community might use.

We have been working on a similar approach, which is a libp2p Gossip based abstraction layer protocol which full fill the needs of most of the p2p blockchain-related projects. In our case we named it "subpub" (similar to PubSub) https://gitlab.com/vocdoni/go-dvote/-/tree/master/subpub

We are thinking on how to improve it and make it more resilient, efficient and smart. My background is on Mesh networks and I have been working on Dynamic Routing protocols for such networks. The problems we try to solve here are quite similar to that. On Mesh networks we have mainly two kind of dynamic routing protocols:

- link state (every node knows about the state of the whole network)
- distance vector (every node only knows its best direct neighbour to fetch a destination)

The second kind is the one that gives better results. Babel, Bmx7 and Batman-adv are three well known, widely used, routing protocols that use such mechanism.

So I would like to import the same principles on the blockchain p2p gossip layer. The basics would be something like:

1. A node uses the libp2p Kademlia DHT for boostraping (find nodes to connect)
2. Once a number of connections have been made (i.e 10 peers), disconnect from the DHT
3. Every node sends a very small packet (named Originator Message or OGM) to its neighbours which will be broadcasted to the whole network
4. With this mechanism all nodes know the existence of all nodes in the network
5. These OGMs are sent constantly by all nodes, i.e every 500ms and have a tracking number (1, 2, 3, 4....)
6. Node A will receive the same OGM (from B with the tracking number) probably by all its neighbours
7. Using the latency and packet loss node A will determine that the best way to reach B is through her neighbour C
8. This is done by all nodes, so every node will keep a table with a very simple relation: For reaching node X my best neighbour is Y
9. Once a packet is sent, it will reach its destination by the best path because every node will make its best choice.

This is a very simplified version on how this could work. More information here:

[...]

PR #1511, opened 2021-04-07
feat(dot/network): Add cache for network message.

arijitAD:
## Changes

- Add cache to store network messages to prevent sending duplicate messages.

## Tests

```
TestMessageCache
TestBroadcastDuplicateMessage
```

## Issues

- #738

### DIFF
--- a/dot/network/host.go
+++ b/dot/network/host.go
@@ -23,6 +23,7 @@ import (
 	"path"
 	"time"
 
+	"github.com/dgraph-io/ristretto"
 	badger "github.com/ipfs/go-ds-badger2"
 	"github.com/libp2p/go-libp2p"
 	libp2phost "github.com/libp2p/go-libp2p-core/host"
@@ -49,20 +50,18 @@ var privateCIDRs = []string{
 
 // host wraps libp2p host with network host configuration and services
 type host struct {
-	ctx        context.Context
-	h          libp2phost.Host
-	dht        *dual.DHT
-	bootnodes  []peer.AddrInfo
-	protocolID protocol.ID
-	cm         *ConnManager
-	ds         *badger.Datastore
+	ctx          context.Context
+	h            libp2phost.Host
+	dht          *dual.DHT
+	bootnodes    []peer.AddrInfo
+	protocolID   protocol.ID
+	cm           *ConnManager
+	ds           *badger.Datastore
+	messageCache *messageCache
 }
 
 // newHost creates a host wrapper with a new libp2p host instance
 func newHost(ctx context.Context, cfg *Config) (*host, error) {
-	// use "p2p" for multiaddress format
-	ma.SwapToP2pMultiaddrs()
-
 	// create multiaddress (without p2p identity)
 	addr, err := ma.NewMultiaddr(fmt.Sprintf("/ip4/0.0.0.0/tcp/%d", cfg.Port))
 	if err != nil {
@@ -143,14 +142,29 @@ func newHost(ctx context.Context, cfg *Config) (*host, error) {
 	// wrap host and DHT service with routed host
 	h = rhost.Wrap(h, dht)
 
+	cacheSize := 64 << 20 // 64 MB
+	config := ristretto.Config{
+		NumCounters: int64(float64(cacheSize) * 0.05 * 2),
+		MaxCost:     int64(float64(cacheSize) * 0.95),
+		BufferItems: 64,
+		Cost: func(value interface{}) int64 {
+			return int64(1)
+		},
+	}
+	msgCache, err := NewMessageCache(config, MsgCacheTTL)
+	if err != nil {
+		return nil, err
+	}
+
 	return &host{
-		ctx:        ctx,
-		h:          h,
-		dht:        dht,
-		bootnodes:  bns,
-		protocolID: pid,
-		cm:         cm,
-		ds:         ds,
+		ctx:          ctx,
+		h:            h,
+		dht:          dht,
+		bootnodes:    bns,
+		protocolID:   pid,
+		cm:           cm,
+		ds:           ds,
+		messageCache: msgCache,
 	}, nil
 
 }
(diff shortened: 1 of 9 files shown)


dot/network/message_cache.go
```diff
@@ -0,0 +1,73 @@
...
+var MsgCacheTTL = 5 * time.Minute
+
+// MessageCache is used to detect duplicated messages per peer.
+type MessageCache struct {
+	cache *ristretto.Cache
+	ttl   time.Duration
+}
+
+// NewMessageCache creates a new MessageCache which takes config and TTL duration.
+func NewMessageCache(config ristretto.Config, ttl time.Duration) (*MessageCache, error) {
```
noot, line 21: does this need to be exported?

dot/network/host.go
```diff
@@ -143,14 +142,29 @@ func newHost(ctx context.Context, cfg *Config) (*host, error) {
 	// wrap host and DHT service with routed host
 	h = rhost.Wrap(h, dht)
 
+	cacheSize := 64 << 20 // 64 MB
+	config := ristretto.Config{
+		NumCounters: int64(float64(cacheSize) * 0.05 * 2),
+		MaxCost:     int64(float64(cacheSize) * 0.95),
+		BufferItems: 64,
```
noot, line 149: what do these values do?

dot/network/message_cache.go
```diff
@@ -0,0 +1,73 @@
...
+
+	_, ok := m.cache.Get(key)
+	return ok
+}
+
+func generateCacheKey(peer peer.ID, msg string) ([]byte, error) {
+	peerBytes, err := peer.Marshal()
+	if err != nil {
+		return nil, err
+	}
```
noot, line 65: the `peer.ID` is a `string` under the hood, so you can just do `string(peer)`

dot/network/notifications.go
```diff
@@ -247,6 +242,15 @@ func (s *Service) broadcastExcluding(info *notificationsProtocol, excluding peer
 			logger.Trace("sending handshake", "protocol", info.protocolID, "peer", peer, "message", hs)
 			err = s.host.send(peer, info.protocolID, hs)
 		} else {
+			if s.host.messageCache != nil {
+				var added bool
+				added, err = s.host.messageCache.Put(peer, msg.String())
```
noot, line 247: this should use `msg.Encode()` instead of `msg.String()`, not sure if `String` will correctly represent all the fields

dot/network/notifications.go
```diff
@@ -247,6 +242,15 @@ func (s *Service) broadcastExcluding(info *notificationsProtocol, excluding peer
...
+				if err != nil || !added {
+					logger.Error("failed to add message to cache", "peer", peer, "error", err)
```
noot, line 249: will `Put` error if the key already exists?

dot/network/notifications.go
```diff
@@ -247,6 +242,15 @@ func (s *Service) broadcastExcluding(info *notificationsProtocol, excluding peer
 			logger.Trace("sending handshake", "protocol", info.protocolID, "peer", peer, "message", hs)
 			err = s.host.send(peer, info.protocolID, hs)
 		} else {
+			if s.host.messageCache != nil {
```
noot, line 245: put this into one line?

noot (commented):
great work, just a couple small comments!

dot/network/host.go
edwardmack: Good question.

edwardmack (approved):
LGTM, tests show cache is working as designed. Nice.

dot/network/host.go
arijitAD: These params are used to keep the cache size fixed and helps in eviction.

dot/network/message_cache.go
arijitAD: Done

dot/network/message_cache.go
arijitAD: Fixed.

dot/network/notifications.go
arijitAD: Done

dot/network/notifications.go
arijitAD: No. It will return `false`.

dot/network/notifications.go
arijitAD: It may lead to code duplication.

dot/network/notifications.go
noot: in this case, it will log the error if the message is already in the cache, maybe remove the `!added` case for now

dot/network/message_cache.go
```diff
@@ -0,0 +1,68 @@
...
+
+	if ttl == 0 {
+		ttl = msgCacheTTL
+	}
+
+	return &messageCache{cache: cache, ttl: ttl}, nil
+}
+
+// Put appends peer ID and message data and stores it in cache with TTL.
+func (m *messageCache) Put(peer peer.ID, msg []byte) (bool, error) {
```
noot, line 35: this could be unexported also

dot/network/message_cache.go
```diff
@@ -0,0 +1,68 @@
...
+	if ok {
+		return false, nil
+	}
+
+	ok = m.cache.SetWithTTL(key, "", 1, m.ttl)
+	return ok, nil
+}
+
+// Exists checks if <peer ID, message data> exist in cache.
+func (m *messageCache) Exists(peer peer.ID, msg []byte) bool {
```
noot, line 51: same here

dot/network/notifications.go
noot: what do you mean?

dot/network/notifications.go
```diff
@@ -247,6 +242,22 @@ func (s *Service) broadcastExcluding(info *notificationsProtocol, excluding peer
...
+				var encMsg []byte
+				encMsg, err = msg.Encode()
+				if err != nil {
+					logger.Error("failed to encode message", "peer", peer, "error", err)
+					continue
+				}
+
+				var added bool
+				added, err = s.host.messageCache.Put(peer, encMsg)
```
noot, line 254: it might be nice to update `Put` to take `Message`, and encode the message inside `Put` so it doesn't need to be done here

noot (approved):
looks good! just a few minor comments

noot:
@arijitAD btw, `NotificationsMessage` interface has a `Hash` function already, so you can use that instead of `Encode` and then hashing

dot/network/notifications.go
arijitAD: I will do that. I was trying to make it generic so it can be used for other messages as well.

dot/network/notifications.go
arijitAD: Done

dot/network/message_cache.go
arijitAD: Done

dot/network/message_cache.go
arijitAD: Done

Issue #1498, opened 2021-03-31
receive notifications messages on block-announce, grandpa, transactions substreams


noot:
## Task summary

- currently, it seems we are not receiving messages (apart from handshakes) on notifications substreams
- this task is to fix the underlying issue and receive notifications messages on block-announce, grandpa, transactions substreams

PR #1512, opened 2021-04-07
feat(dot/network): implement persistent peers functionality

noot:
## Changes

- implement persistent peers functionality, connect with peers specified in config file at bootstrap and try to reconnect to them if we disconnect

## Tests

```
go test ./dot/network -short
```

## Issues

- related to #1498


arijitAD (approved):

dot/network/connmgr_test.go
```diff
@@ -86,3 +86,34 @@ func TestProtectUnprotectPeer(t *testing.T) {
...
+		Port:            7001,
+		RandSeed:        2,
+		NoMDNS:          true,
+		PersistentPeers: []string{addrs[0].String()},
+	}
+	nodeB := createTestService(t, configB)
+
+	// B should have connected to A during bootstrap
+	conns := nodeB.host.h.Network().ConnsToPeer(nodeA.host.id())
+	require.NotEqual(t, 0, len(conns))
```
edwardmack, line 112: I don't understand this, shouldn't len(conns) == 1, since nodeB should be connected to nodeA?

edwardmack (approved):
Code looks good, I just have a question regarding the test.

dot/network/connmgr_test.go
noot: yeah, this just checks that it's greater than 0, there might be multiple conns open potentially

PR #1519, opened 2021-04-13
chore(dot/network, lib/grandpa): update network.ConsensusMessage, add grandpa.NeighbourMessage and handle accordingly

### LINKED ISSUE #1481: ensure blocks are finalized on time near head
## Task summary

- when syncing near the head of kusama or polkadot, we will need to ensure blocks are finalized in a timely manner
- this requires receiving and processing finalization messages on the grandpa substream, which currently doesn't seem to be happening
- fix this and ensure blocks are marked finalized

noot:
## Changes

- update network.ConsensusMessage, no longer has ID field
- add grandpa.NeighbourMessage and finalize blocks based on it
- update blockState to correctly set the block number -> hash mapping by checking for re-org on `AddBlock`, previously was set to the most recently added block at that number which was incorrect. this was needed so we finalize the right block

## Tests

```
go test ./dot/network -short
go test ./lib/grandpa -short
```

also run a [kusama, pokladot] node w/ network logs in debug and you should see some logs where we receive grandpa neighbour messages:
```
DBUG[04-14|10:45:31] received message on notifications sub-protocol pkg=network protocol=/paritytech/grandpa/1 message="ConsensusMessage Data=0201b94e000000000000c101000000000000aca84600" peer=12D3KooWHsvEicXjWWraktbZ4MQBizuyADQtuEGr3NbDvtm5rFA5 caller=notifications.go:187
```

if you are at the head this should result in a block getting marked final

## Issues

- closes #1481


dot/state/block.go
```diff
@@ -366,11 +363,7 @@ func (bs *BlockState) GetBlockBody(hash common.Hash) (*types.Body, error) {
 
 // SetBlockBody will add a block body to the db
 func (bs *BlockState) SetBlockBody(hash common.Hash, body *types.Body) error {
-	bs.lock.Lock()
-	defer bs.lock.Unlock()
```
edwardmack, line 370: I'm a confused why some locks were removed (lines 333, and 369), but others added in different spots (add at line 423, 495 and 593).  Doesn't seem to have any race conditions now, so this is working.  How do we determine where locks are needed?

edwardmack (approved):

dot/state/block.go
noot: the db operations are already threadsafe (chaindb has locks for all db.Put, db.Get, etc.), and same for the blocktree, so I only added locks for functions that did multiple operations in them where some  variables need to not change until the function is done, for example AddBlock requires no other blocks to be added until it's finished

arijitAD (approved):

dot/state/block.go
```diff
@@ -322,17 +322,14 @@ func (bs *BlockState) GetBlockHash(blockNumber *big.Int) (*common.Hash, error) {
...
 	if err != nil {
-		return nil, fmt.Errorf("cannot get block %d: %s", blockNumber, err)
+		return nil, fmt.Errorf("cannot get block %d: %w", blockNumber, err)
 	}
 	hash := common.NewHash(byteHash)
 	return &hash, nil
 }
 
 // SetHeader will set the header into DB
 func (bs *BlockState) SetHeader(header *types.Header) error {
```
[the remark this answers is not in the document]
noot: no, chaindb already has locking: https://github.com/ChainSafe/chaindb/blob/master/database.go#L29

Issue #1257, opened 2020-12-21
handle properties, chainType genesis attributes with build-spec


edwardmack:
## Expected Behavior

- Adding properties attribute to genesis.json file should create properties attribute on genesis-raw.json after running `build-spec`
- `chainType` should be handled as well
- 
## Current Behavior

- properties attribute in genesis.json is not being parsed during `build-spec` so the properties attribute in genesis-raw.json is set to null.

## Possible Solution

- Update `build-spec` process to handle properties attribute.

## Steps to Reproduce

1.
2.
3.
4.

## Specification

- go version:
- gossamer version:
- gossamer commit tag:
- gossamer commit hash:
- operating system:
- additional links:

## Checklist

- [x] I have read [CODE_OF_CONDUCT](https://github.com/ChainSafe/gossamer/blob/development/.github/CODE_OF_CONDUCT.md) and [CONTRIBUTING](https://github.com/ChainSafe/gossamer/blob/development/.github/CONTRIBUTING.md) 
- [x] I have provided as much information as possible and necessary
- [x] I am planning to submit a pull request to fix this issue myself

Issue #1030, opened 2020-07-24
handle bootnodes in build-spec


edwardmack:
## Expected Behavior

- Bootnodes should be included in JSON created by build-spec.

## Current Behavior

- Currently bootnodes are not included in output generated by `build-spec`.

## Possible Solution

- Include bootnode check in build-spec parser.

## Steps to Reproduce

1.
2.
3.
4.

## Specification

- go version:
- gossamer version:
- gossamer commit tag:
- gossamer commit hash:
- operating system:
- additional links:

## Checklist

- [x] I have read [CODE_OF_CONDUCT](https://github.com/ChainSafe/gossamer/blob/development/.github/CODE_OF_CONDUCT.md) and [CONTRIBUTING](https://github.com/ChainSafe/gossamer/blob/development/.github/CONTRIBUTING.md) 
- [x] I have provided as much information as possible and necessary
- [x] I am planning to submit a pull request to fix this issue myself

Issue #1029, opened 2020-07-24
add --output flag to build-spec command


edwardmack:
## Expected Behavior

-

## Current Behavior

-  Currently the `build-spec` command writes output to the standard output (screen).  So the user usually needs to  redirect the output.

## Possible Solution

- Create a `--output` flag for build-spec to redirect it's output to the given file.

## Steps to Reproduce

1.
2.
3.
4.

## Specification

- go version:
- gossamer version:
- gossamer commit tag:
- gossamer commit hash:
- operating system:
- additional links:

## Checklist

- [x] I have read [CODE_OF_CONDUCT](https://github.com/ChainSafe/gossamer/blob/development/.github/CODE_OF_CONDUCT.md) and [CONTRIBUTING](https://github.com/ChainSafe/gossamer/blob/development/.github/CONTRIBUTING.md) 
- [x] I have provided as much information as possible and necessary
- [x] I am planning to submit a pull request to fix this issue myself

PR #1520, opened 2021-04-13
feat: Add properties and chainId on build-spec command

EclesioMeloJunior:
## Changes

1. Resolve the [properties and chain type issue](https://github.com/ChainSafe/gossamer/issues/1257) when the **build-spec** command does not parse these values correctly 

2. Includes the [feature to persist a new file using a --output ](https://github.com/ChainSafe/gossamer/issues/1029).

```
before:
gossamer build-spec --raw --genesis-spec ./my-genesis-spec.json > file.json

actual:
gossamer build-spec --raw --genesis-spec ./my-genesis-spec.json --output ./myoutput.json
```

_The first behavior was not removed_

If the file that was passed to **--output** flag exists then an error is returned: **"_file ... already exists, rename to avoid overwritten_"**. 

## Tests

```
1. I change the tests cases **TestBuildFromGenesis** to cover the new parsed fields
2. Created **TestWriteGenesisSpecFile** test case to test the creation of an output file and if the data stored is correct
3. Created **TestWriteGenesisSpecFileWhenFileAlreadyExists** test case to test if the **WriteGenesisSpecFile** returns an error when a filepath already exists
```

## Issues

closes https://github.com/ChainSafe/gossamer/issues/1029
closes https://github.com/ChainSafe/gossamer/issues/1257
closes https://github.com/ChainSafe/gossamer/issues/1030


lib/genesis/helpers.go
```diff
@@ -141,6 +141,7 @@ func NewGenesisSpecFromJSON(file string) (*Genesis, error) {
 	g := new(Genesis)
 
 	err = json.Unmarshal(data, g)
+
```
arijitAD, line 144: Remove this.

arijitAD (commented):
Looks good. Unit tests seem to be failing. Also, can you generate the spec and verify that it's working with gossamer?

EclesioMeloJunior:
> Looks good. Unit tests seem to be failing. Also, can you generate the spec and verify that it's working with gossamer?

The file was generated with success! But I notice that if the output path file `--output ~/path/to/output/output-json.json`  does not exist the command exit without error but doesn't create the file, so I changed the implementation to create the directory before the file creation if it doesn't exist. In the case that the dir exists then nothing happens.

image

noot (approved):
it works for me! great work :)

dot/build_spec.go
```diff
@@ -71,6 +78,21 @@ func BuildFromGenesis(path string, authCount int) (*BuildSpec, error) {
 	return bs, nil
 }
 
+// WriteGenesisSpecFile writes the build-spec in the output filepath
+func WriteGenesisSpecFile(data []byte, fp string) error {
+	// if file already exists then dont apply any written on it
+	if utils.PathExists(fp) {
+		return fmt.Errorf("file %s already exists, rename to avoid overwritten", fp)
```
noot, line 85: change to "rename to avoid overwriting"

arijitAD (approved):

Issue #1501, opened 2021-04-01
Implement remaining telemetry messages


edwardmack:
## Issue summary

Server side messages (https://github.com/paritytech/substrate-telemetry/blob/master/backend/src/node/message.rs#L40) that need to be implemented:
[x] 'system.connected'
[ ] 'system.interval'
[ ] 'system.network_state'
[x] 'block.import'
[ ] 'notify.finalized'
[ ] 'txpool.import'
[ ] 'afg.finalized'
[ ] 'afg.received_precommit'
[ ] 'afg.received_prevote'
[ ] 'afg.received_commit'
[ ] 'afg.authority_set'
[ ] 'afg.finalized_blocks_up_to'
[ ] 'aura.pre_sealed_block'
[ ] 'prepared_block_for_proposing'

## Other information and links

-

danforbes:
I updated the link that points to the telemetry APIs that Substrate Telemetry is expecting ✅

kishansagathiya:
@danforbes We should close this. Remaining telemetry message are not being sent.

PR #1528, opened 2021-04-15
feat (dot/telemetry): implement telemetry system.interval message

edwardmack:
## Changes

- Implement system.interval message for network data which send telemetry data for: peer count, bandwidth_upload and bandwidth_download.
- Implement system.interval message for block data which sends telemetry data for: best hash, finalized hash, finalized height, best height, txcount and used_state_cache_size.
- Note: txcount and used_state_cache_size are not currently being reported, I need to determine where to find these metrics.

## Tests

```
go test ./dot/telemetry/...
```
or start gossamer node on polkadot chain, and browse to https://telemetry.polkadot.io and note that the gossamer node is reporting data to this server.

## Issues

- related to #1501


dot/telemetry/telemetry.go
```diff
@@ -100,18 +102,55 @@ func (h *Handler) SendConnection(data *ConnectionData) {
...
+func (h *Handler) SendBlockIntervalData(data *BlockIntervalData) {
+	payload := log.Fields{"best": data.BestHash.String(), "finalized_hash": data.FinalizedHash.String(),
+		"finalized_height": data.FinalizedHeight, "height": data.BestHeight, "msg": "system.interval", "txcount": data.TXCount,
+		"used_state_cache_size": data.UsedStateCacheSize}
+	h.telemetryLogger = log.WithFields(log.Fields{"id": 1, "payload": payload, "ts": time.Now()})
+	h.telemetryLogger.Print()
+	h.sendTelemetry()
+}
+
+func (h *Handler) sendTelemetry() {
```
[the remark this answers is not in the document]
arijitAD: Also, the lock should be acquired in the functions calling `sendTelemetry` because `Handler.buf` is written before calling `sendTelemetry` but it is reset inside `sendTelemetry`

```
// SendBlockIntervalData send block data system interval information to telemetry connection
func (h *Handler) SendBlockIntervalData(data *BlockIntervalData) {
	h.Lock()
	defer h.Unlock()
	payload := log.Fields{"best": data.BestHash.String(), "finalized_hash": data.FinalizedHash.String(),
		"finalized_height": data.FinalizedHeight, "height": data.BestHeight, "msg": "system.interval", "txcount": data.TXCount,
		"used_state_cache_size": data.UsedStateCacheSize}
	h.telemetryLogger = log.WithFields(log.Fields{"id": 1, "payload": payload, "ts": time.Now()})
	h.telemetryLogger.Print()
	h.sendTelemetry()
}
```

arijitAD:
Race test will fail since `Handler.buf` and `Handler.wsConn` can be concurrently accessed.

Wrote a race test.
```
var resultCh chan []byte

func TestHandler_SendBlockImport(t *testing.T) {
	var wg sync.WaitGroup
	wg.Add(2)

	resultCh = make(chan []byte)

	go func() {
		GetInstance().SendBlockImport("hash", big.NewInt(2))
		// note, we only check the first 101 bytes because the remaining bytes are the timestamp, which we can't estimate
		wg.Done()
	}()

	go func() {
		GetInstance().SendNetworkData(&NetworkData{
			Peers:   1,
			RateIn:  2,
			RateOut: 3,
		})
		// note, we only check the first 103 bytes because the remaining bytes are the timestamp, which we can't estimate
		wg.Done()
	}()
	wg.Wait()

	expected1 := []byte(`{"id":1,"payload":{"bandwidth_download":2,"bandwidth_upload":3,"msg":"system.interval","peers":1},"ts":`)
	expected2 := []byte(`{"id":1,"payload":{"best":"hash","height":2,"msg":"block.import","origin":"NetworkInitialSync"},"ts":`)

	expected := [][]byte{expected1, expected2}
	var actual [][]byte
	for data := range resultCh {
		actual = append(actual, data)
		if len(actual) == 2 {
			break
		}
	}

	sort.Slice(actual, func(i, j int) bool {
		return bytes.Compare(actual[i], actual[j]) < 0
	})
	require.Contains(t, string(actual[0]), string(expected[0]))
	require.Contains(t, string(actual[1]), string(expected[1]))
}

func listen(w http.ResponseWriter, r *http.Request) {
	c, err := upgrader.Upgrade(w, r, nil)
	if err != nil {
		log.Printf("Error %v\n", err)
	}
	defer c.Close()
	for {
		_, msg, err := c.ReadMessage()
		if err != nil {
			log.Printf("read err %v", err)
			break
		}

		resultCh <- msg
		lastMessage = msg
	}
}
```

dot/telemetry/telemetry_test.go
```diff
@@ -61,6 +62,32 @@ func TestHandler_SendBlockImport(t *testing.T) {
...
+	GetInstance().SendBlockIntervalData(&BlockIntervalData{
+		BestHash:           common.MustHexToHash("0x07b749b6e20fd5f1159153a2e790235018621dd06072a62bcd25e8576f6ff5e6"),
+		BestHeight:         big.NewInt(32375),
+		FinalizedHash:      common.MustHexToHash("0x687197c11b4cf95374159843e7f46fbcd63558db981aaef01a8bac2a44a1d6b2"),
+		FinalizedHeight:    big.NewInt(32256),
+		TXCount:            2,
+		UsedStateCacheSize: 1886357,
+	})
+	time.Sleep(time.Millisecond)
+	require.Equal(t, expected, lastMessage[:295])
```
arijitAD, line 88: Use `require.Contains` in all testcase.
```
require.Contains(t, lastMessage, expected)
```

arijitAD (changes requested):

dot/network/host.go
```diff
@@ -311,7 +316,11 @@ func (h *host) writeToStream(s libp2pnetwork.Stream, msg Message) error {
 	lenBytes := uint64ToLEB128(msgLen)
 	encMsg = append(lenBytes, encMsg...)
 
-	_, err = s.Write(encMsg)
+	sent, err := s.Write(encMsg)
+	if err == nil {
+		// todo (ed) determine if there are other places to capture data sent
+		h.bwc.LogSentMessage(int64(sent))
+	}
 	return err
```
noot, line 324: change to `if err != nil { return err } // sent message`

noot: btw, this is the only function where outgoing messages are sent

dot/network/service.go
```diff
@@ -521,6 +535,9 @@ func (s *Service) readStream(stream libp2pnetwork.Stream, peer peer.ID, decoder
 			_ = stream.Close()
 			return
 		}
+
+		// todo (ed) determine if there are other places to capture data received
+		s.host.bwc.LogRecvMessage(int64(tot))
```
noot, line 540: all messages received by the network go through this function, and no where else, so having this just here is fine

dot/network/sync.go
```diff
@@ -356,6 +358,28 @@ func (q *syncQueue) benchmark() {
...
+		if err != nil {
+			continue
+		}
+
+		telemetry.GetInstance().SendBlockIntervalData(&telemetry.BlockIntervalData{
+			BestHash:           best.Hash(),
+			BestHeight:         best.Number,
+			FinalizedHash:      finalized.Hash(),
+			FinalizedHeight:    finalized.Number,
+			TXCount:            0, // todo (ed) determine where to get tx count
```
noot, line 377: is this the count of txs in a block, or in the queue?

dot/network/sync.go
```diff
@@ -356,6 +358,28 @@ func (q *syncQueue) benchmark() {
 	}
 }
 
+func (q *syncQueue) sentBlockIntervalTelemetry() {
```
noot, line 361: this function could be just part of the `Service`, since it doesn't use any `syncQueue` functionality

noot (approved):
looking good, only small comments!

dot/network/service.go
```diff
@@ -279,6 +281,18 @@ func (s *Service) logPeerCount() {
 	}
 }
 
+func (s *Service) publishNetworkTelemetry() {
```
timwu20, line 284:
```suggestion
// done be passed in to shut this goroutine down if we care about shutting down gracefully
func (s *Service) publishNetworkTelemetry(done chan interface{}) {
	ticker := time.NewTicker(s.PublishNetworkTelemetryDuration)
	defer ticker.Stop()
main:
	for {
		select {
		case <-done:
			break main
		// use a channel to ensure runtime can work on other goroutines
		case <-ticker.C:
			fmt.Println("do stuff")
		}
	}
}
```

dot/telemetry/telemetry.go
```diff
@@ -100,18 +102,55 @@ func (h *Handler) SendConnection(data *ConnectionData) {
...
 	payload := log.Fields{"best": bestHash, "height": height.Int64(), "msg": "block.import", "origin": "NetworkInitialSync"}
 	h.telemetryLogger = log.WithFields(log.Fields{"id": 1, "payload": payload, "ts": time.Now()})
 	h.telemetryLogger.Print()
-	h.sendTelemtry()
+	h.sendTelemetry()
 }
 
-func (h *Handler) sendTelemtry() {
+// NetworkData struct to hold network data telemetry information
+type NetworkData struct {
```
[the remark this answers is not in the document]
edwardmack: No, I've updated this.

dot/telemetry/telemetry.go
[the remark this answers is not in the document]
edwardmack: I've updated this to add locks and created sender which reads from buffer and sends messages.

dot/telemetry/telemetry_test.go
edwardmack: Updated.

dot/network/host.go
edwardmack: Updated. Ok, thanks.

dot/network/service.go
edwardmack: Excellent, thanks.

dot/network/sync.go
edwardmack: I'm not sure what they want reported.

dot/network/sync.go
edwardmack: Good idea, I've moved this.

dot/network/service.go
```diff
@@ -279,6 +281,18 @@ func (s *Service) logPeerCount() {
...
+	for {
+		o := s.host.bwc.GetBandwidthTotals()
+
+		telemetry.GetInstance().SendNetworkData(&telemetry.NetworkData{
+			Peers:   s.host.peerCount(),
+			RateIn:  o.RateIn,
+			RateOut: o.RateOut,
+		})
+		time.Sleep(time.Second * 5)
```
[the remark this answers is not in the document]
edwardmack: I've added `telemertyInterval` time.Duration to the service config so that can be modified as needed now.

dot/network/service.go
edwardmack: I modified publishNetworkTelemetry as noted above, however when I added `s.doneNetworkTelemerty <- struct{}{}` to line 398 in network/service.go the node seems to hang when shutting down, and never closes.  I see this when I run TestStartService (in service_test.go).  It seems that stop gets called twice, so the channel seems to block there.  I may not have set this up correctly, so let me know if I'm missing something.

dot/network/service.go
```diff
@@ -344,6 +394,9 @@ func (s *Service) Stop() error {
 		logger.Error("Failed to close host", "error", err)
 	}
 
+	// todo (ed) when this isn't commented out the node freezes on stop (it seems this is called twice)
+	//s.doneNetworkTelemetry <- struct{}{}
```
timwu20, line 398: `close(s.doneNetworkTelemetry)`

dot/network/service.go
```diff
@@ -89,6 +89,10 @@ type Service struct {
 	noDiscover  bool
 	noMDNS      bool
 	noGossip    bool // internal option
+
+	// telemetry
+	telemetryInterval    time.Duration
+	doneNetworkTelemetry chan struct{}
```
timwu20, line 95: make this a `chan interface{}`.

timwu20 (changes requested):

dot/network/service.go
edwardmack: changed.

dot/network/service.go
edwardmack: updated.

dot/network/service.go
```diff
@@ -344,6 +394,13 @@ func (s *Service) Stop() error {
 		logger.Error("Failed to close host", "error", err)
 	}
 
+	defer func() {
+		if r := recover(); r != nil {
+			logger.Error("failed to close telemetry service")
+		}
+	}()
+	close(s.doneNetworkTelemetry)
+
```
edwardmack, line 403: Note: I added this to catch panic that happens (during testing) because stop is called twice:  
in dot/core/test_helpers.go line 183:
```
	if net, ok := cfg.Network.(*network.Service); ok {
		net.SetTransactionHandler(s)
		_ = net.Stop()
	}
```

and in dot/core/test_helpers.go line 208:
```
	t.Cleanup(func() {
		err := srvc.Stop()
		require.NoError(t, err)
	})
```
So, the service is stopped (when created as test service) in line 183, then stop is called again during test cleanup in line 208.

Please advise if I should remove one of these and remove the recover check, or if I should keep the recover check in place?

arijitAD: You can rename `doneNetworkTelemetry` to `closeCh` to make it more generic to signal the close of service.
Using the channel you can also check if the service is closed or not and return early if it's already closed. This will avoid panic on multiple close calls.
Also alternatively you can wrap the close code in `sync.Once` which will only be called once.

arijitAD: Also, Channel names in Golang end  usually with `Ch`

dot/telemetry/telemetry.go
```diff
@@ -65,13 +66,16 @@ func GetInstance() *Handler {
...
 				log.SetFormatter(new(MyJSONFormatter))
+				go handlerInstance.sender()
 			})
 	}
 	return handlerInstance
 }
 
 // AddConnections adds connections to telemetry sever
 func (h *Handler) AddConnections(conns []*genesis.TelemetryEndpoint) {
+	h.Lock()
```
arijitAD, line 77: This lock may not be required as we are adding connections only when starting the node.

dot/telemetry/telemetry.go
```diff
@@ -65,13 +66,16 @@ func GetInstance() *Handler {
...
+	defer h.Unlock()
 	for _, v := range conns {
 		c, _, err := websocket.DefaultDialer.Dial(v.Endpoint, nil)
 		if err != nil {
```
arijitAD, line 81: We should continue adding connections on error.

dot/telemetry/telemetry.go
```diff
@@ -96,29 +100,87 @@ type ConnectionData struct {
...
+		line, err := h.buf.ReadBytes(byte(10))
+		h.RUnlock()
 		if err != nil {
-			// TODO (ed) determine how to handle this error
-			fmt.Printf("ERROR connecting to telemetry %v\n", err)
+			continue
+		}
+
+		for _, c := range h.wsConn {
+			h.Lock()
```
arijitAD, line 177: Even this lock is not required as connections are not accessed concurrently.

dot/telemetry/telemetry.go
```diff
@@ -96,29 +100,87 @@ type ConnectionData struct {
...
+		"finalized_height": data.FinalizedHeight, "height": data.BestHeight, "msg": "system.interval", "txcount": data.TXCount, // nolint
+		"used_state_cache_size": data.UsedStateCacheSize}
+	telemetryLogger := log.WithFields(log.Fields{"id": 1, "payload": payload, "ts": time.Now()})
+	telemetryLogger.Print()
+}
+
+func (h *Handler) sender() {
+	for {
+		h.RLock()
+		line, err := h.buf.ReadBytes(byte(10))
```
arijitAD, line 170: Add a comment specifying that newline delimiter is used.

arijitAD:
This is how you can check for closed channels.
```
package main

import (
    "fmt"
)

func main() {
    ch := make(chan int)
    go func() {
        close(ch)
    }()
mainLoop:
    for {
        select {
        case v, hasMore := <- ch:
		if !hasMore {
			break mainLoop
		}
        	fmt.Printf("v is %v\n", v)
	default:
		fmt.Println("Do stuff")
        }
    }
    fmt.Println("Break!")
}
```

dot/telemetry/telemetry.go
edwardmack: Removed lock.

dot/telemetry/telemetry.go
edwardmack: Removed return.

dot/telemetry/telemetry.go
edwardmack: Good point, removed lock.

dot/telemetry/telemetry.go
edwardmack: added comment.

edwardmack:
@arijitAD, @timwu20 I think this is ready for another look.  Thanks for the example code!

dot/telemetry/telemetry.go
```diff
@@ -76,7 +78,6 @@ func (h *Handler) AddConnections(conns []*genesis.TelemetryEndpoint) {
 		c, _, err := websocket.DefaultDialer.Dial(v.Endpoint, nil)
 		if err != nil {
 			fmt.Printf("Error %v\n", err)
```
arijitAD, line 80: `continue` 
Don't append a connection that returns an error.

arijitAD (approved):

dot/telemetry/telemetry.go
edwardmack: Good catch, updated.

timwu20 (approved):

PR #1529, opened 2021-04-15
feat(dot/network, lib/grandpa): request justification on receiving NeighbourMessage, verify justification on receipt

### LINKED ISSUE #1363: handle justifications received from block responses

noot:
## Changes

- request justification on receiving NeighbourMessage
- rename types to match substrate: `Justification` -> `SignedPrecommit`, `FullJustification` -> `Justification`
- update `Justification` type to match substrate
- verify justification on receipt (if setID is available, currently need to update catch up before it's fully available)

## Tests

```
go test ./lib/grandpa -short
go test ./dot/network -short
go test ./dot/sync -short
```

## Issues

- closes #1363


lib/grandpa/grandpa.go
```diff
@@ -535,7 +535,7 @@ func (s *Service) playGrandpaRound() error {
 				return false
 			}
 
-			if completable && finalizable && uint64(s.head.Number.Int64()) >= prevBfc.number {
+			if completable && finalizable && uint32(s.head.Number.Int64()) >= prevBfc.number {
```
arijitAD, line 538: `uint32(s.head.Number.Int64())` Wouldn't this overflow?

arijitAD (approved):

lib/grandpa/grandpa.go
noot: block numbers shouldn't go over uint32, in substrate they are set to uint32. but I agree this could be nicer, maybe we can update the `types.Header.Number` type to uint32 at some point

lib/grandpa/message_handler.go
```diff
@@ -98,6 +105,13 @@ func (h *MessageHandler) handleNeighbourMessage(msg *NeighbourMessage) error {
 		return nil
 	}
 
+	// TODO: make linter british
```
edwardmack, line 108: What does make linter british mean?

edwardmack (approved):

lib/grandpa/message_handler.go
noot: lol, I always got linter issues for misspellings when spelling things Canadian (ie neighbour vs neighbor), apparently we can change the longer settings to use UK English. I'll open an issue for it

Issue #1482, opened 2021-03-19
ensure transaction sent to non-authority node can be validated


noot:
## Task summary

- ensure transaction sent to non-authority node can be validated, added to the tx pool, and gossiped to the network and eventually be included in a block
- currently, the integration test sends the transaction to an authority node because it seems non-authority nodes cannot validate the tx
- fix this and update the integration test

noot:
@arijitAD this issue is probably due to this code:  https://github.com/ChainSafe/gossamer/blob/development/dot/core/messages.go#L28
```
func (s *Service) HandleTransactionMessage(msg *network.TransactionMessage) error {
	logger.Debug("received TransactionMessage")
	if !s.isBlockProducer {
		return nil
	}
       //etc
}
```

arijitAD:
Investigated this issue. The Authority node fails to execute the block.
Looking at the block execute code.
```
 725 ^[[36mDBUG^[[0m[04-12|18:15:55] processing block                         ^[[36mpkg^[[0m=sync ^[[36mhash^[[0m=0xb72077cd91f685f820c67a1076673a929f885f89f208d425c918acb783d1f18a ^[[36mcaller^[[0m=syncer.go:248
[...]

PR #1530, opened 2021-04-16
fix(babe): Fix extrinsic format in block.

arijitAD:
## Changes

- Fix extrinsic encoding in block
- Fix race in removing data from transaction queue.
- Determine correct error for apply extrinsic.

## Tests

```
TestApplyExtrinsicErrors
TestPendingExtrinsic
```

## Issues

- #1482


noot:
@arijitAD please put `closes` before the issue number in your description so that the issue gets closed when this is merged

lib/common/optional/types.go
```diff
@@ -166,6 +166,25 @@ func (x *Bytes) Decode(r io.Reader) (*Bytes, error) {
 	return x, nil
 }
 
+// DecodeBytes return an optional Bytes from scale encoded data
+func (x *Bytes) DecodeBytes(data []byte) (*Bytes, error) {
+	if len(data) == 0 || data[0] > 1 {
+		return nil, ErrInvalidOptional
+	}
+
+	x.exists = data[0] != 0
```
noot, line 175: set `x.value = nil` is `!x.exists`

tests/stress/stress_test.go
```diff
@@ -358,16 +358,24 @@ func TestSync_Restart(t *testing.T) {
...
+	// Randomize authority node selection.
+	rand.Seed(time.Now().UnixNano())
+	rand.Shuffle(len(utils.KeyList), func(i, j int) { utils.KeyList[i], utils.KeyList[j] = utils.KeyList[j], utils.KeyList[i] })
+
+	defer func() {
+		// Restore the key list.
+		sort.Slice(utils.KeyList, func(i, j int) bool {
+			return utils.KeyList[i] > utils.KeyList[j]
+		})
+	}()
```
noot, line 378: this doesn't actually submit it to a non-authority node, since it still submits it to `node` which is started with the BABE authority config

you can move starting the non-authority nodes up to when the authority node is started:
```
nodes, err := utils.InitializeAndStartNodes(t, numNodes-1, utils.GenesisDefault, utils.ConfigNoBABE)
nodes = append(nodes, node)
```

then start the `api` with a randomly chosen node index: 
```
api, err := gsrpc.NewSubstrateAPI(fmt.Sprintf("http://localhost:%s", nodes[randomIdx].RPCPort))
```

noot (commented):
looks good!! please update integration test

lib/babe/build.go
```diff
@@ -175,38 +176,44 @@ func (b *Service) buildBlockBABEPrimaryPreDigest(slot Slot) (*types.BabePrimaryP
...
+		ret, err := b.rt.ApplyExtrinsic(extrinsic)
+		if err != nil {
+			logger.Warn("failed to apply extrinsic", "error", err, "extrinsic", extrinsic)
 			continue
 		}
 
-		logger.Debug("build block applied extrinsic", "extrinsic", next)
+		ok, err := determineErr(ret)
+		if err != nil && ok {
+			logger.Warn("failed to interpret extrinsic error", "error", err, "extrinsic", extrinsic)
```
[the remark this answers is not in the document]
arijitAD: Error can occur while applying extrinsic or after dispatching extrinsic. I have fixed the error log.

lib/common/optional/types.go
arijitAD: Done

tests/stress/stress_test.go
arijitAD: @noot  I tried the above. Blocks are not getting propagated to other nodes. I was wondering if we can close this PR and I will investigate it in a separate issue?

noot (approved):
looks good!

tests/stress/stress_test.go
noot: is the issue that the non-authority node can't validate the transaction? in that case, the original issue still applies so you can leave it open when this PR is closed. otherwise please open another issue

lib/common/optional/types.go
```diff
@@ -166,6 +166,27 @@ func (x *Bytes) Decode(r io.Reader) (*Bytes, error) {
 	return x, nil
 }
 
+// DecodeBytes return an optional Bytes from scale encoded data
+func (x *Bytes) DecodeBytes(data []byte) (*Bytes, error) {
```
noot, line 170: please add unit test for this, test coverage is low

edwardmack (approved):
Great work, the handling for the extrinsic errors looks great.

lib/common/optional/types.go
arijitAD: Done

lib/babe/build.go
```diff
@@ -175,38 +176,44 @@ func (b *Service) buildBlockBABEPrimaryPreDigest(slot Slot) (*types.BabePrimaryP
...
+		logger.Trace("build block", "applying extrinsic", extrinsic)
+
+		ret, err := b.rt.ApplyExtrinsic(extrinsic)
+		if err != nil {
+			logger.Warn("failed to apply extrinsic", "error", err, "extrinsic", extrinsic)
 			continue
 		}
 
-		logger.Debug("build block applied extrinsic", "extrinsic", next)
+		ok, err := determineErr(ret)
```
timwu20, line 202: maybe introduce new error types  and type check `err` for `DispatchOutcome` or `TransactionValidityError` instead of relying on this `bool` return.

arijitAD: Done

PR #1537, opened 2021-04-22
WIP: feat(network): Reuse stream for notification message.

arijitAD:
## Changes

- Add stream to handshake data

## Tests

```

```

## Issues

-


dot/network/block_announce.go
```diff
@@ -198,7 +199,7 @@ func (s *Service) getBlockAnnounceHandshake() (Handshake, error) {
 	}, nil
 }
 
-func (s *Service) validateBlockAnnounceHandshake(peer peer.ID, hs Handshake) error {
+func (s *Service) validateBlockAnnounceHandshake(peer peer.ID, hs Handshake, stream libp2pnetwork.Stream) error {
```
noot, line 202: don't need to add `stream` in this function as that would require changing all the handler functions, it should be stored in the `createNotificationsMessageHandler` function

noot: changing this function signature would also require exposing streams to external packages such as grandpa which we don't want

dot/network/host.go
```diff
@@ -264,32 +264,28 @@ func (h *host) bootstrap() {
 
 // send writes the given message to the outbound message stream for the given
 // peer (gets the already opened outbound message stream or opens a new one).
-func (h *host) send(p peer.ID, pid protocol.ID, msg Message) (err error) {
-	// get outbound stream for given peer
-	s := h.getOutboundStream(p, pid)
+func (h *host) send(p peer.ID, pid protocol.ID, msg Message, stream libp2pnetwork.Stream) error {
```
noot, line 267: why does this function need to be changed? generally if we know about the stream we can use `writeToStream`, if we're opening it for the first time we can use `send`

noot: also this won't this panic if `stream` is nil which it seems to be in the tests? I would recommend changing it back

dot/network/notifications.go
```diff
@@ -126,9 +127,10 @@ func (s *Service) createNotificationsMessageHandler(info *notificationsProtocol,
 				info.handshakeData.Store(peer, &handshakeData{
 					validated: false,
 					received:  true,
+					stream:    stream,
 				})
 
-				err := handshakeValidator(peer, hs)
+				err := handshakeValidator(peer, hs, stream)
```
noot, line 133: instead of passing in the stream to the `handshakeValidator` you can just store it in this function instead

dot/network/notifications.go
```diff
@@ -156,15 +158,18 @@ func (s *Service) createNotificationsMessageHandler(info *notificationsProtocol,
...
 					return errCannotValidateHandshake
 				}
 
-				hsData.validated = true
-				hsData.received = true
+				info.handshakeData.Store(peer, &handshakeData{
+					validated: true,
+					received:  true,
+					stream:    stream,
+				})
```
noot, line 172: since we are storing the stream in this function (which is good) we don't need to pass it to `handshakeValidator`

dot/network/notifications.go
```diff
@@ -202,7 +207,7 @@ func (s *Service) createNotificationsMessageHandler(info *notificationsProtocol,
 		seen := s.gossip.hasSeen(msg)
 		if !seen {
 			// TODO: update this to write to stream w/ handshake established
-			s.broadcastExcluding(info, peer, msg)
+			s.broadcastExcluding(info, peer, msg, &stream)
```
noot, line 210: why are we passing the stream?

dot/network/notifications.go
```diff
@@ -241,7 +246,7 @@ func (s *Service) broadcastExcluding(info *notificationsProtocol, excluding peer
 			})
 
 			logger.Trace("sending handshake", "protocol", info.protocolID, "peer", peer, "message", hs)
-			err = s.host.send(peer, info.protocolID, hs)
+			err = s.host.send(peer, info.protocolID, hs, *stream)
```
noot, line 249: remove the stream here, the `peer` and `stream` do not correspond to the same peer, since we are iterating over all our peers

dot/network/notifications.go
```diff
@@ -258,7 +263,7 @@ func (s *Service) broadcastExcluding(info *notificationsProtocol, excluding peer
 
 			// we've already completed the handshake with the peer, send message directly
 			logger.Trace("sending message", "protocol", info.protocolID, "peer", peer, "message", msg)
-			err = s.host.send(peer, info.protocolID, msg)
+			err = s.host.send(peer, info.protocolID, msg, *stream)
```
noot, line 266: same here, remove `stream` from `send` function

noot: in general this function needs to check `notificationsProtocol.handshakeData[peer]` for the stream if it already exists for this peer, then use `writeToStream` if the stream is opened and established, otherwise it should use `send` to open a new stream and send the handshake (which is what was previously happening)

dot/network/notifications.go
```diff
@@ -211,7 +216,7 @@ func (s *Service) createNotificationsMessageHandler(info *notificationsProtocol,
 
 // gossipExcluding sends a message to each connected peer except the given peer
 // Used for notifications sub-protocols to gossip a message
-func (s *Service) broadcastExcluding(info *notificationsProtocol, excluding peer.ID, msg NotificationsMessage) {
+func (s *Service) broadcastExcluding(info *notificationsProtocol, excluding peer.ID, msg NotificationsMessage, stream *libp2pnetwork.Stream) {
```
noot, line 219: remove `stream`

dot/network/notifications.go
```diff
@@ -42,7 +42,7 @@ type (
 	HandshakeDecoder = func([]byte) (Handshake, error)
 
 	// HandshakeValidator validates a handshake. It returns an error if it is invalid
-	HandshakeValidator = func(peer.ID, Handshake) error
+	HandshakeValidator = func(peer.ID, Handshake, libp2pnetwork.Stream) error
```
noot, line 45: remove `stream`

arijitAD:
Closing this PR as the same has been implemented in https://github.com/ChainSafe/gossamer/pull/1545

Issue #1531, opened 2021-04-16
lib/grandpa: ensure catch-up logic works when interacting with substrate nodes


noot:
## Task summary

-  grandpa catch-up logic should be updated to work with substrate nodes
- currently is triggered only on receiving `FinalizationMessage` but should also happen when receiving a `NeighbourMessage`
- ensure algorithm works with substrate allowing us to catch up to grandpa

timwu20:
Let's revisit this after the grandpa integration

timwu20:
going to close this issue since this should be handled by the new GRANDPA impl.

Issue #1309, opened 2021-01-15
create BaseState to access database keys that aren't in other state db tables


noot:
## Task summary

- we should have a separate `State` type for accessing keys that aren't stored until the block, storage, etc. database tables (ie. prefixed)
- will remove the need for `baseDB` in block, storage
- these are items in the `dot/state/db.go` file that are generally needed for node startup and saving state before shutdown

PR #1540, opened 2021-04-23
chore(dot/core, dot/state, lib/grandpa): implement GrandpaState, use in DigestHandler and grandpa.Service

noot:
## Changes

- move `grandpa.Voter` type to `types.GrandpaVoter`
- implement `GrandpaState` to store set IDs and corresponding authority sets in the db
- update `core.DigestHandler` to use `GrandpaState` when handling grandpa related digests
-  update `grandpa.Service` to use `GrandpaState` to determine set ID, authority set

## Tests

```
go test ./dot/core 
go test ./dot/state
go test ./lib/grandpa -short
```

## Issues

- related to #1531


dot/state/grandpa.go
```diff
@@ -0,0 +1,217 @@
...
+
+	return big.NewInt(0).SetBytes(num), nil
+}
+
+// SetNextPause sets the next grandpa pause at the given block number
+func (s *GrandpaState) SetNextPause(number *big.Int) error {
+	return s.db.Put(pauseKey, number.Bytes())
+}
+
+// GetNextPause returns the block number of the next grandpa pause, nil if there is no upcoming pause
```
timwu20, line 187: is this called in the code?

dot/state/grandpa.go
```diff
@@ -0,0 +1,217 @@
...
+	return big.NewInt(0).SetBytes(num), nil
+}
+
+// SetNextResume sets the next grandpa resume at the given block number
+func (s *GrandpaState) SetNextResume(number *big.Int) error {
+	return s.db.Put(resumeKey, number.Bytes())
+}
+
+// GetNextResume returns the block number of the next grandpa resume, nil if there is no upcoming resume
+func (s *GrandpaState) GetNextResume() (*big.Int, error) {
```
timwu20, line 207: looks like this and `GetNextPause` can use the same codepath?

lib/grandpa/grandpa.go
```diff
@@ -35,7 +35,7 @@ import (
 )
 
 var (
-	interval = time.Second
+	interval = time.Second // TODO: make this configurable; currently 1s is same as substrate; total round length is then 2s
```
timwu20, line 38: 👀 👍

lib/scale/encode.go
```diff
@@ -103,6 +103,10 @@ func (se *Encoder) EncodeCustom(in interface{}) (int, error) {
 	if ok {
 		res := reflect.ValueOf(in).MethodByName("Encode").Call([]reflect.Value{})
 		val := res[0].Interface()
+		if len(res) < 2 {
```
timwu20, line 106: 👀

dot/core/digest.go
```diff
@@ -225,131 +203,113 @@ func (h *DigestHandler) handleBlockFinalization(ctx context.Context) {
...
-		h.grandpa.UpdateAuthorities([]*types.Authority{})
 		h.grandpaPause = nil
 	}
 
 	sc := h.grandpaScheduledChange
 	if sc != nil && num.Cmp(sc.atBlock) == 0 {
-		h.grandpa.UpdateAuthorities(sc.auths)
+		err := h.grandpaState.IncrementSetID()
+		if err != nil {
+			logger.Error("failed to increment grandpa set ID", "error", err)
```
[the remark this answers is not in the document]
noot: same, will update

dot/core/digest.go
```diff
@@ -225,131 +203,113 @@ func (h *DigestHandler) handleBlockFinalization(ctx context.Context) {
...
-			return errors.New("already have forced change scheduled")
-		}
-
-		fc := &types.GrandpaForcedChange{}
-		dec, err := scale.Decode(d.Data[1:], fc)
-		if err != nil {
-			return err
-		}
-		fc = dec.(*types.GrandpaForcedChange)
+	return h.grandpaState.SetNextChange(types.NewGrandpaVotersFromAuthorities(auths), big.NewInt(0).Add(header.Number, big.NewInt(int64(sc.Delay))))
```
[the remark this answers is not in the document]
noot: good idea, @dutterbutter is there a way to add max column size to the linter?

dot/state/grandpa.go
```diff
@@ -0,0 +1,217 @@
...
+	setIDChangePrefix = []byte("change")
+	pauseKey          = []byte("pause")
+	resumeKey         = []byte("resume")
+	currentSetIDKey   = []byte("setID")
+)
+
+// GrandpaState tracks information related to grandpa
+type GrandpaState struct {
+	baseDB chaindb.Database
+	db     chaindb.Database
```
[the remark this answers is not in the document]
noot: yes exactly, there's an issue open to clean this up: https://github.com/ChainSafe/gossamer/issues/1309 might just do this quick since it's a bit confusing atm

dot/state/grandpa.go
```diff
@@ -0,0 +1,217 @@
...
+
+// GetAuthorities returns the authorities for the given setID
+func (s *GrandpaState) GetAuthorities(setID uint64) ([]*types.GrandpaVoter, error) {
+	enc, err := s.db.Get(authoritiesKey(setID))
+	if err != nil {
+		return nil, err
+	}
+
+	r := &bytes.Buffer{}
+	_, _ = r.Write(enc)
```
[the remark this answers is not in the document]
noot: added check

dot/state/grandpa.go
noot: not currently, grandpa should be updated to check it. i can open an issue for this

dot/state/grandpa.go
noot: one gets the `pauseKey` and one gets the `resumeKey`

lib/scale/encode.go
noot: yeah had to add this since there was a type that had `Encode() []byte`.... your interface cleanup should fix this

dot/core/digest.go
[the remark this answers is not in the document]
dutterbutter: I can take a look!

lib/scale/encode.go
```diff
@@ -103,6 +103,10 @@ func (se *Encoder) EncodeCustom(in interface{}) (int, error) {
 	if ok {
```
arijitAD, line 103:
```
if !ok {
    return 0, fmt.Errorf("cannot call EncodeCustom")
}
// rest ...
```

lib/scale/encode.go
```diff
@@ -103,6 +103,10 @@ func (se *Encoder) EncodeCustom(in interface{}) (int, error) {
...
 		res := reflect.ValueOf(in).MethodByName("Encode").Call([]reflect.Value{})
 		val := res[0].Interface()
```
arijitAD, line 105: Add a check for `len(res) != 0` or this might segfault.

lib/grandpa/grandpa_test.go
```diff
@@ -64,30 +67,34 @@ func newTestState(t *testing.T) *state.Service {
...
 	return &state.Service{
-		Block: block,
+		Block:   block,
+		Grandpa: grandpa,
 	}
 }
 
 func newTestVoters() []*Voter {
-	voters := []*Voter{}
+	vs := []*Voter{}
```
arijitAD, line 80: Replace it with a nil slice.
```
var vs []*Voter
```

arijitAD (approved):

lib/grandpa/grandpa_test.go
noot: is there a practical difference between the two? I thought `var vs []*Voter` would make `[]*Voter(nil)` and `voters := []*Voter{}` makes `[]*Voter{}`, so one is nil and one is empty, but maybe they are the same?

edwardmack (approved):

PR #1545, opened 2021-04-27
fix(dot/network): Fix notification handshake and reuse stream.

### LINKED ISSUE #1524: dot/network: update notifications protocol to track specific stream w/ established handshake
## Task summary

- currently, a stream is established via handshake, however when messages are sent out it doesn't necessarily send it over the established stream
- we should track established streams and use them for pushing notifications messages
- this can be done potentially by storing the stream inside a peer's `handshakeData` or by tagging the libp2p stream itself `https://pkg.go.dev/github.com/libp2p/go-libp2p-core/network#Stat` using `Stat.Extra`

### LINKED ISSUE #1538: Non authority node fails to handle block announce message.
## Issue summary

- Non-authority node fails to read block announce message.

- Continuation of #1482

arijitAD:
## Changes

- Fix notification handshake
- Reuse stream created during the handshake.

## Tests

```
TestPendingExtrinsic
```

## Issues

- closes #1538
- closes #1524


dot/network/host.go
```diff
@@ -262,32 +262,26 @@ func (h *host) bootstrap() {
 	}
 }
 
-// send writes the given message to the outbound message stream for the given
-// peer (gets the already opened outbound message stream or opens a new one).
-func (h *host) send(p peer.ID, pid protocol.ID, msg Message) (err error) {
-	// get outbound stream for given peer
-	s := h.getOutboundStream(p, pid)
```
noot, line 269: why is this removed?

dot/network/notifications.go
```diff
@@ -256,13 +291,38 @@ func (s *Service) broadcastExcluding(info *notificationsProtocol, excluding peer
 				}
 			}
 
+			hsData, has = info.getHandshakeData(peer)
+			if !has {
+				continue
+			}
```
noot, line 297: shouldn't this be covered by the first case? `if hsData, has := info.getHandshakeData(peer); !has || !hsData.received`

dot/network/notifications.go
```diff
@@ -256,13 +291,38 @@ func (s *Service) broadcastExcluding(info *notificationsProtocol, excluding peer
...
 			logger.Trace("sending message", "protocol", info.protocolID, "peer", peer, "message", msg)
-			err = s.host.send(peer, info.protocolID, msg)
-		}
 
-		if err != nil {
-			logger.Debug("failed to send message to peer", "peer", peer, "error", err)
+			err = s.host.writeToStream(hsData.stream, msg)
+			if err != nil {
+				logger.Trace("failed to send message to peer", "peer", peer, "error", err)
+			}
```
noot, line 305: this can be moved outside the `if-else` case, since in the `if` block we exchange the handshake, we still want to send the message after exchanging the handshake

dot/network/notifications.go
```diff
@@ -256,13 +291,38 @@ func (s *Service) broadcastExcluding(info *notificationsProtocol, excluding peer
...
 	}
 }
+
+func readHandshake(stream libp2pnetwork.Stream) (Handshake, error) {
+	msgBytes := make([]byte, maxBlockResponseSize)
+	tot, err := readStream(stream, msgBytes)
+	if err == io.EOF {
+		return nil, io.EOF
+	} else if err != nil {
+		logger.Trace("failed to read handshake from stream", "peer", stream.Conn().RemotePeer(), "protocol", stream.Protocol(), "error", err)
```
noot, line 316: can remove the logs in this function and leave it to the caller of `readHandshake` to log

dot/network/notifications.go
```diff
@@ -256,13 +291,38 @@ func (s *Service) broadcastExcluding(info *notificationsProtocol, excluding peer
...
+		_ = stream.Close()
+		return nil, err
+	}
+
+	// TODO: Implement this for other messages.
+	handShake, err := decodeBlockAnnounceHandshake(msgBytes[:tot])
+	if err != nil {
+		logger.Trace("failed to decode handshake message from peer", "protocol", stream.Protocol(), "err", err)
```
noot, line 324: same here

dot/network/utils.go
```diff
@@ -204,6 +204,7 @@ func readStream(stream libp2pnetwork.Stream, buf []byte) (int, error) {
 		return 0, fmt.Errorf("message size greater than maximum: got %d", length)
 	}
 
+	// TODO: Allocate buffer after reading length. Avoid pre allocation.
```
noot, line 207: can you remove this? the reason it pre-allocates is because:
1. allocation is expensive, if we allocate each time we receive a message it increases our CPU usage a lot
2. we don't want to allocate a byte array based on a length a peer gives us, because a malicious peer can then tell us to allocate some very large buffer causing the node to crash

noot (commented):
looking good! just a few comments, but overall great work :D

dot/network/host.go
arijitAD: `send` is only called from `broadcastExcluding` when we don't have handshake data in which case we need to create a new stream.

dot/network/notifications.go
arijitAD: Done

dot/network/notifications.go
arijitAD: Done

dot/network/notifications.go
arijitAD: Done

dot/network/notifications.go
```diff
@@ -209,6 +214,59 @@ func (s *Service) createNotificationsMessageHandler(info *notificationsProtocol,
 	}
 }
 
+func (s *Service) sendData(peer peer.ID, hs Handshake, info *notificationsProtocol, msg NotificationsMessage) {
+	var err error
+	hsData, has := info.getHandshakeData(peer)
+	if !has || !hsData.received || !hsData.validated {
```
noot, line 220: if we've already received the data but it's not valid, then we should return (and probably close the stream with the peer, or disconnect from them)

dot/network/notifications.go
```diff
@@ -209,6 +214,59 @@ func (s *Service) createNotificationsMessageHandler(info *notificationsProtocol,
...
+
+		if !added {
+			return
+		}
+	}
+
+	// we've already completed the handshake with the peer, send message directly
+	logger.Trace("sending message", "protocol", info.protocolID, "peer", peer, "message", msg)
+
+	err = s.host.writeToStream(hsData.stream, msg)
```
noot, line 264: check `hsData.stream != nil`

dot/network/service.go
```diff
@@ -398,6 +398,10 @@ func (s *Service) RegisterNotificationsProtocol(sub protocol.ID,
 
 	decoder := createDecoder(info, handshakeDecoder, messageDecoder)
 	handlerWithValidate := s.createNotificationsMessageHandler(info, handshakeValidator, messageHandler)
+	streamHandler := func(stream libp2pnetwork.Stream, peerID peer.ID) {
+		s.readStream(stream, peerID, decoder, handlerWithValidate)
+	}
+	np.streamHandler = streamHandler
```
noot, line 404: nice :D

dot/network/notifications.go
```diff
@@ -55,23 +55,25 @@ type notificationsProtocol struct {
 	protocolID    protocol.ID
 	getHandshake  HandshakeGetter
 	handshakeData *sync.Map //map[peer.ID]*handshakeData
+	streamHandler func(stream libp2pnetwork.Stream, peerID peer.ID)
```
noot, line 58: maybe put `func(stream libp2pnetwork.Stream, peerID peer.ID)` as its own type?

noot (commented):
great work!! only small comments

dot/network/notifications.go
arijitAD: This is already done in the notification handler.

noot: in that case, can you remove the `|| !hsData.validated` case?

dot/network/notifications.go
arijitAD: Done

dot/network/notifications.go
arijitAD: Done

dot/network/notifications.go
arijitAD: Done

timwu20 (approved):

dot/network/notifications.go
arijitAD: Done

noot (approved):
amazing work! tested it out with polkadot and works as it should :)

PR #1547, opened 2021-04-28
chore(lib/common) implement MustHexToBigInt

edwardmack:
## Changes

- implement MustHexToBigInt which take a hex encoded string and converts to a big.Int
- Can handle odd length strings, i.e. 0xf works as well as 0x0f.
-
-

## Tests

```
go test ./lib/common -run TestMustHexToBigInt -v 
```

## Issues

-


noot (approved):
nice :D

lib/common/common.go
```diff
@@ -103,6 +104,31 @@ func MustHexToBytes(in string) []byte {
...
+func MustHexToBigInt(in string) *big.Int {
+	if len(in) < 2 {
+		panic("invalid string")
+	}
+
+	if strings.Compare(in[:2], "0x") != 0 {
+		panic(ErrNoPrefix)
+	}
+
+	// Ensure we have an even length
```
arijitAD, line 118:
```
	in = in[2:]

	// Ensure we have an even length
	if len(in)%2 != 0 {
		in = "0" + in
	}
```

arijitAD (approved):

lib/common/common.go
edwardmack: Yes, this looks better, updated.

Issue #1167, opened 2020-11-05
scale.DecodeCustom and decoder.DecodeCustom should return the same types


noot:
## Expected Behavior

currently:
```
func DecodeCustom(in []byte, t interface{}) error 
```
and
```
func (sd *Decoder) DecodeCustom(t interface{}) (interface{}, error) 
```

these two functions should both return `error` and set the decoded value to the passed in interface

## Checklist

- [x] I have read [CODE_OF_CONDUCT](https://github.com/ChainSafe/gossamer/blob/development/.github/CODE_OF_CONDUCT.md) and [CONTRIBUTING](https://github.com/ChainSafe/gossamer/blob/development/.github/CONTRIBUTING.md) 
- [x] I have provided as much information as possible and necessary
- [ ] I am planning to submit a pull request to fix this issue myself

noot:
related to #611

Issue #1072, opened 2020-08-10
Move SCALE into it's own package and add documentation


amerameen:
## Expected Behavior

- Since SCALE can be used as standalone package it's recommended that we move it into it's own repo and provide the appropriate documentation.

## Current Behavior

- SCALE is a part of the Gossamer repo.

## Checklist

- [x] I have read [CODE_OF_CONDUCT](https://github.com/ChainSafe/gossamer/blob/development/.github/CODE_OF_CONDUCT.md) and [CONTRIBUTING](https://github.com/ChainSafe/gossamer/blob/development/.github/CONTRIBUTING.md) 
- [x] I have provided as much information as possible and necessary
- [ ] I am planning to submit a pull request to fix this issue myself

Issue #611, opened 2020-02-12
integrate scale custom encoding, clean up custom type encoding methods


noot:
since SCALE now has EncodeCustom and DecodeCustom functions to check if a type has a custom encoding function, we can clean up instances of types that have types within them that can be custom encoded. 

SCALE may also need to be updated to integrate the Custom functions within EncodeTuple, EncodeArray, etc. same for Decode. additionally all custom functions should have the function sig `func (ms *MyStruct) Decode(r io. Reader) error` 

including but not limited to usages of: p2p message types, babe types, optional types

Noc2:
just fyi: https://github.com/itering/scale.go

PR #1548, opened 2021-04-29
refactor(lib/scale): refactor lib/scale into own package with idiomatic types, marshalling, and optionality

timwu20:
## Changes

- Moving common.Uint128 into scale package
- Revise public interface to more commonly used `Marshal/Unmarshal` interface
- Code cleanup
- Refactor regarding using pointers to types to denote optionality
- Implement `scale:"1"` struct tag field ordering for deterministic serialisation even if struct attributes are reordered
- Create new `VaryingDataType` type which is used to translate into a rust enum
- `NewVaryingDataType()` is used to construct a `VaryingDataType` with associated `VaryingDataTypeValue` entities.
- `NewVaryingDataTypeSlice()` is used to construct a slice of `VaryingDataType`.  A custom type for this is required for decoding, since the destination data type is required to decode.  

## Tests

```
go test ./pkg/scale
```

## Issues

- #611 
- #1072
- #1167


pkg/scale/encode.go
```diff
@@ -0,0 +1,357 @@
...
+		if err != nil {
+			return
+		}
+		for _, elem := range arr {
+			err = es.marshal(elem)
+			if err != nil {
+				return
+			}
+		}
+	case []bool:
```
EclesioMeloJunior, line 119: If the type is a slice of `[]bool, [][]byte, []int, []string` then could they be in the same case as `[]*big.Int`? Because seems there is the same logic for each of them.

pkg/scale/encode.go
```diff
@@ -0,0 +1,357 @@
...
+	switch arr := t.(type) {
+	// int is the only case that handles encoding differently.
+	// all other cases can recursively call es.marshal()
+	case []int:
+		err = es.encodeLength(len(arr))
+		if err != nil {
+			return
+		}
+		for _, elem := range arr {
+			err = es.encodeUint(uint(elem))
```
EclesioMeloJunior, line 101: if there some signed integer in `arr` then we can have an overflow, maybe we could create a new case type to `[]uint`.

EclesioMeloJunior (changes requested):

pkg/scale/encode.go
timwu20: These cases are added so we don't need to use the `reflect` library to iterate through the slice.  Check out the `default` case.

pkg/scale/encode.go
timwu20: it's not clearly defined in the spec or the knowledge base but it's assumed that all these are unsigned integers.

pkg/scale/README.md
```diff
@@ -0,0 +1,227 @@
+# go-scale Codec
+
+Go implementation of the SCALE (Simple Concatenated Aggregate Little-Endian) data format for types used in the Parity Substrate framework.
+
+SCALE is a light-weight format which allows encoding (and decoding) which makes it highly suitable for resource-constrained execution environments like blockchain runtimes and low-power, low-memory devices.
+
+It is important to note that the encoding context (knowledge of how the types and data structures look) needs to be known separately at both encoding and decoding ends. The encoded data does not include this contextual information.
+
+This codec attempts to translate the primitive Go types to the associated types in SCALE.  It also introduces a few custom types to implement SCALE primitives that have no direct translation to a Go primitive type.
```
noot, line 9: implement rust primitives?

pkg/scale/comparison_test.go
```diff
@@ -0,0 +1,141 @@
...
+		},
+	}
+	oldEncode, err := oldDigest.Encode()
+	if err != nil {
+		t.Errorf("unexpected err: %v", err)
+		return
+	}
+
+	type Digests VaryingDataType
+	err = RegisterVaryingDataType(Digests{}, ChangesTrieRootDigest{}, PreRuntimeDigest{}, ConsensusDigest{}, SealDigest{})
```
noot, line 75: very nice, I like this a lot

pkg/scale/comparison_test.go
```diff
@@ -0,0 +1,141 @@
...
+	for _, tt := range allTests {
+		dst := reflect.New(reflect.TypeOf(tt.in)).Elem().Interface()
+		if err := Unmarshal(tt.want, &dst); (err != nil) != tt.wantErr {
+			b.Errorf("decodeState.unmarshal() error = %v, wantErr %v", err, tt.wantErr)
+			return
+		}
+	}
+}
+
+// func BenchmarkDecode(b *testing.B) {
```
noot, line 127: ?

pkg/scale/decode.go
```diff
@@ -0,0 +1,539 @@
+package scale
```
noot, line 1: add license to tops of all files pls

pkg/scale/decode.go
```diff
@@ -0,0 +1,539 @@
...
+import (
+	"bytes"
+	"encoding/binary"
+	"errors"
+	"fmt"
+	"math/big"
+	"reflect"
+)
+
+func Unmarshal(data []byte, dst interface{}) (err error) {
```
noot, line 12: it would also be nice to support unmarshalling from an `io.Reader`

pkg/scale/decode.go
```diff
@@ -0,0 +1,539 @@
...
+			case true:
+				err = ds.decodeResult(dstv)
+			case false:
+				err = ds.decodeStruct(dstv)
+			}
+		case reflect.Array:
+			err = ds.decodeArray(dstv)
+		case reflect.Slice:
+			t := reflect.TypeOf(in)
+			// check if this is a convertible to VaryingDataType, if so encode using encodeVaryingDataType
```
noot, line 71: * decode using decodeVaryingDataType

pkg/scale/decode.go
```diff
@@ -0,0 +1,539 @@
...
+				err = ds.decodeSlice(dstv)
+			}
+		default:
+			_, ok := in.(VaryingDataTypeValue)
+			switch ok {
+			case true:
+				var temp reflect.Value
+				t := reflect.TypeOf(in)
+				switch t.Kind() {
+				// TODO: support more primitive types.  Do we need to support arrays and slices as well?
```
noot, line 85: is this for values within a `VaryingDataType`? if so, yes, I think the current usages for variadic types are are `common.Hash` or `uint64` type in the `lib/common/variadic` package

pkg/scale/decode.go
```diff
@@ -0,0 +1,539 @@
...
+			err = fmt.Errorf("unable to find registered custom Result.Ok for: %T", dstv.Interface())
+			return
+		}
+		newOk := reflect.New(reflect.TypeOf(okIn))
+		err = ds.unmarshal(newOk.Elem())
+		if err != nil {
+			break
+		}
+		res := reflect.New(dstt)
+		res.Elem().FieldByName("Ok").Set(newOk.Elem())
```
noot, line 139: should this do a check for the "Ok" field? or is it checked when the type is registered?

pkg/scale/decode.go
```diff
@@ -0,0 +1,539 @@
...
+		}
+		dstv.Set(tempElem)
+	default:
+		err = fmt.Errorf("unsupported Option value: %v, bytes: %v", rb, ds.Bytes())
+	}
+	return
+}
+
+func (ds *decodeState) decodeVaryingDataType(dstv reflect.Value) (err error) {
+	l, err := ds.decodeLength()
```
noot, line 186: does this work for singular varying data types? like for example, a single `DigestItem` and not a `Digests{}` slice?

pkg/scale/decode.go
```diff
@@ -0,0 +1,539 @@
...
+		err = ds.unmarshal(elem)
+		if err != nil {
+			return
+		}
+	}
+	dstv.Set(temp.Elem())
+	return
+}
+
+// decodeStruct comment about this
```
noot, line 261: ?

pkg/scale/scale.go
```diff
@@ -0,0 +1,93 @@
...
+type fieldScaleIndicesCache struct {
+	cache map[string]fieldScaleIndices
+	sync.RWMutex
+}
+
+func (fsic *fieldScaleIndicesCache) fieldScaleIndices(in interface{}) (v reflect.Value, indices fieldScaleIndices, err error) {
+	t := reflect.TypeOf(in)
+	v = reflect.ValueOf(in)
+	key := fmt.Sprintf("%s.%s", t.PkgPath(), t.Name())
+	if key != "." {
```
noot, line 33: is this case needed? is there a chance the key would be `"."`? if so add an else case

pkg/scale/scale.go
```diff
@@ -0,0 +1,93 @@
...
+				fieldIndex: i,
+			})
+		case "-":
+			// ignore this field
+			continue
+		default:
+			indices = append(indices, fieldScaleIndex{
+				fieldIndex: i,
+				scaleIndex: &tag,
+			})
```
noot, line 63: can there be a struct where some fields have scale tags and some don't? in that case, is there a chance two fields could end up with the same index? eg
```
type myStruct struct {
    a // add tag for scale index = 1
    b
}
```

both a and b would have index 1

pkg/scale/decode.go
```diff
@@ -0,0 +1,539 @@
...
+
+	b := make([]byte, length)
+	_, err = ds.Read(b)
+	if err != nil {
+		return
+	}
+
+	in := dstv.Interface()
+	inType := reflect.TypeOf(in)
+	dstv.Set(reflect.ValueOf(b).Convert(inType))
```
noot, line 377: would this panic if `b` cannot be converted to `inType`?

pkg/scale/decode.go
```diff
@@ -0,0 +1,539 @@
...
+			break
+		}
+		out = int(binary.LittleEndian.Uint64(buf))
+	case uint:
+		buf := make([]byte, 8)
+		_, err = ds.Read(buf)
+		if err != nil {
+			break
+		}
+		out = uint(binary.LittleEndian.Uint64(buf))
```
noot, line 517: are these cases needed still?

pkg/scale/decode_test.go
```diff
@@ -0,0 +1,169 @@
...
+		ptrTest := test{
+			name:    t.name,
+			in:      t.in,
+			wantErr: t.wantErr,
+			want:    t.want,
+			out:     t.out,
+		}
+		switch t.in {
+		case nil:
+			// this doesn't actually happen since none of the tests have nil value for tt.in
```
noot, line 142: we should probably test the nil case for this

pkg/scale/encode.go
```diff
@@ -0,0 +1,420 @@
...
+	"bytes"
+	"encoding/binary"
+	"fmt"
+	"math/big"
+	"reflect"
+)
+
+func Marshal(v interface{}) (b []byte, err error) {
+	es := encodeState{
+		fieldScaleIndicesCache: cache,
```
noot, line 13: is this needed? since `cache` is a package variable

pkg/scale/encode.go
```diff
@@ -0,0 +1,420 @@
...
+	case string:
+		err = es.encodeBytes([]byte(in))
+	case bool:
+		err = es.encodeBool(in)
+	default:
+		switch reflect.TypeOf(in).Kind() {
+		case reflect.Ptr:
+			// Assuming that anything that is a pointer is an Option to capture {nil, T}
+			elem := reflect.ValueOf(in).Elem()
+			switch elem.IsValid() {
```
noot, line 51: why `IsValid` instead of `IsNil`?

pkg/scale/encode.go
```diff
@@ -0,0 +1,420 @@
...
+			case false:
+				err = es.encodeSlice(in)
+			}
+		default:
+			_, ok := in.(VaryingDataTypeValue)
+			switch ok {
+			case true:
+				t := reflect.TypeOf(in)
+				switch t.Kind() {
+				// TODO: support more primitive types.  Do we need to support arrays and slices as well?
```
noot, line 90: same for decode, i think we'll need `common.Hash` which is `[32]byte` and `uint64`

pkg/scale/encode.go
noot: yes this is fine, when encoding all ints are treated as uints, the user would just need to decode it into the right type

pkg/scale/comparison_test.go
timwu20: I was going to benchmark it against the old code, but the types aren't the same and was panic'ing.  I'll remove this.

pkg/scale/decode.go
timwu20: That usually follows the `json.Decoder` interface if you want to supply an `io.Reader`. I'll create an improvement issue for this if you think it'll be worthwhile.  Right now all the data is buffered in memory.

pkg/scale/decode.go
timwu20: I'll add those types.

pkg/scale/decode.go
timwu20: `decodeResult` is called with a `reflect.Value` of type `Result`.  Technically this would panic if a `Result` isn't passed in, but that shouldn't happen, and this is a private receiver function.

pkg/scale/decode.go
timwu20: It would work yes, but it still needs to be treated like a `VaryingDataType` which is a `[]VaryingDataTypeValue`.  In this case it would just have a length of 1.

pkg/scale/decode.go
timwu20: I'll put a comment in.

pkg/scale/scale.go
timwu20: yea I don't think so.  I'll return an `err` if I can't generate a key.  I'll test with anonymous structs to see if causes a problem.

pkg/scale/scale.go
timwu20: they end up getting mapped anyways in the cache, so it will clobber each other depending on the order.  I can throw an err when registering to ensure this doesn't happen.

pkg/scale/decode.go
timwu20: yes the dst is always `[]byte`.  The big type switch is in `decodeState.unmarshal`

pkg/scale/encode.go
timwu20: I like having localized struct attributes.  May make it easier to test.

pkg/scale/README.md
```diff
@@ -0,0 +1,227 @@
...
+
+When decoding SCALE data, knowledge of the structure of the destination data type is required to decode.  Structs are encoded as a SCALE Tuple, where each struct field is encoded in the sequence of the fields.  
+
+#### Struct Tags
+
+go-scale uses a `scale` struct tag to modify the order of the field values during encoding.  This is also used when decoding attributes back to the original type.  This essentially allows you to modify struct field ordering but preserve the encoding/decoding ordering.
+
+See the [usage example](#Struct-Tag-Example).
+
+### Option
```
arijitAD, line 46: Is this implemented?

pkg/scale/encode.go
timwu20: >IsValid reports whether v represents a value. It returns false if v is the zero Value. If IsValid returns false, all other methods except String panic. Most functions and methods never return an invalid Value. If one does, its documentation states the conditions explicitly.

`IsNil` will panic if it isn't valid. 🤷

pkg/scale/decode.go
```diff
@@ -0,0 +1,539 @@
...
+			break
+		}
+		dstv.Set(tempElem)
+	default:
+		err = fmt.Errorf("unsupported Option value: %v, bytes: %v", rb, ds.Bytes())
+	}
+	return
+}
+
+func (ds *decodeState) decodeVaryingDataType(dstv reflect.Value) (err error) {
```
arijitAD, line 185: Avoid named return value 
Source: [practical-go](https://dave.cheney.net/practical-go/presentations/gophercon-israel.html#_avoid_named_return_values)

pkg/scale/encode.go
noot: hm interesting, that seems weird on part of the reflect library but okay

pkg/scale/README.md
timwu20: Yup!

pkg/scale/encode.go
```diff
@@ -0,0 +1,420 @@
...
+}
+
+// encodeArray encodes an interface where the underlying type is an array
+// it writes the encoded length of the Array to the Encoder, then encodes and writes each value in the Array
+func (es *encodeState) encodeArray(in interface{}) (err error) {
+	v := reflect.ValueOf(in)
+	for i := 0; i < v.Len(); i++ {
+		elem := v.Index(i).Interface()
+		switch elem := elem.(type) {
+		case int:
```
noot, line 250: should this case have `int` and `uint`? I guess I'm unsure of the difference between this case and the default

pkg/scale/encode.go
```diff
@@ -0,0 +1,420 @@
...
+	case uint32:
+		err = binary.Write(es, binary.LittleEndian, i)
+	case int64:
+		err = binary.Write(es, binary.LittleEndian, uint64(i))
+	case uint64:
+		err = binary.Write(es, binary.LittleEndian, i)
+	case int:
+		err = binary.Write(es, binary.LittleEndian, int64(i))
+	case uint:
+		err = binary.Write(es, binary.LittleEndian, uint64(i))
```
noot, line 343: remove? i think  `int` or `uint` won't be passed into this func anymore?

pkg/scale/encode.go
```diff
@@ -0,0 +1,420 @@
...
+	}
+	return
+}
+
+// encodeByteArray performs the following:
+// b -> [encodeInteger(len(b)) b]
+// it writes to the buffer a byte array where the first byte is the length of b encoded with SCALE, followed by the
+// byte array b itself
+func (es *encodeState) encodeBytes(b []byte) (err error) {
+	err = es.encodeUint(uint(len(b)))
```
noot, line 312:
```suggestion
	err = es.encodeLength(len(b))
```

pkg/scale/encode.go
```diff
@@ -0,0 +1,420 @@
...
+		for _, elem := range arr {
+			err = es.marshal(elem)
+			if err != nil {
+				return
+			}
+		}
+	default:
+		// use reflect package for any other cases
+		s := reflect.ValueOf(t)
+		err = es.encodeUint(uint(s.Len()))
```
noot, line 229:
```suggestion
		err = es.encodeLength(s.Len())
```

pkg/scale/result.go
```diff
@@ -0,0 +1,41 @@
...
+	case r.Ok == nil && r.Err != nil, r.Ok != nil && r.Err == nil:
+	default:
+		err = fmt.Errorf("Result is invalid: %+v", r)
+	}
+	return
+}
+
+var resCache resultCache = make(resultCache)
+
+func RegisterResult(in interface{}, inOK interface{}, inErr interface{}) (err error) {
```
noot, line 26: add comment pls

noot: also, can `inOk` or `inErr` be `nil`? so that encoding of `Ok` is just `[]byte{0}` and encoding of `Err` is just `[]byte{1}`? can you add a test case for that?

edwardmack (approved):
Great work!  This looks good to me.

pkg/scale/uint128.go
```diff
@@ -0,0 +1,148 @@
...
+		if len(bytes) < 16 {
+			bytes = padBytes(bytes, binary.BigEndian)
+		}
+		u = &Uint128{
+			Upper: binary.BigEndian.Uint64(bytes[:8]),
+			Lower: binary.BigEndian.Uint64(bytes[8:]),
+		}
+	case []byte:
+		var o binary.ByteOrder = binary.LittleEndian
+		if len(order) > 0 {
```
noot, line 73: why is order a variable length? it doesn't look like orders > 0 are used

pkg/scale/uint128.go
```diff
@@ -0,0 +1,148 @@
...
+		err = fmt.Errorf("unsupported type: %T", in)
+	}
+	return
+}
+
+// Bytes returns the Uint128 in little endian format by default.  A variadic paramter
+// order can be used to specify the binary.ByteOrder used
+func (u *Uint128) Bytes(order ...binary.ByteOrder) (b []byte) {
+	var o binary.ByteOrder = binary.LittleEndian
+	if len(order) > 0 {
```
noot, line 93: same here

pkg/scale/uint128.go
```diff
@@ -0,0 +1,148 @@
...
+		o = order[0]
+	}
+	b = make([]byte, 16)
+	switch o {
+	case binary.LittleEndian:
+		o.PutUint64(b[:8], u.Lower)
+		o.PutUint64(b[8:], u.Upper)
+		b = u.trimBytes(b, o)
```
noot, line 101: wondering if trimming excess bytes is needed, or if returning `[16]byte` is fine

pkg/scale/uint128.go
```diff
@@ -0,0 +1,148 @@
...
+// Cmp returns 1 if the receiver is greater than other, 0 if they are equal, and -1 otherwise.
+func (u *Uint128) Compare(other *Uint128) int {
+	switch {
+	case u.Upper > other.Upper, u.Lower > other.Lower:
+		return 1
+	case u.Upper < other.Upper, u.Lower < other.Lower:
+		return -1
+	default:
+		return 0
+	}
```
noot, line 119: are the cases ands or ors? I don't think this is correct, what if `u.Upper == other.Upper` but `u.Upper > other.Upper`? then `u` is greater than `other` but I don't think this checks that case?

pkg/scale/varying_data_type_test.go
```diff
@@ -0,0 +1,292 @@
...
+
+func Test_decodeState_decodeVaryingDataType(t *testing.T) {
+	for _, tt := range varyingDataTypeTests {
+		t.Run(tt.name, func(t *testing.T) {
+			dst := reflect.New(reflect.TypeOf(tt.in)).Elem().Interface()
+			if err := Unmarshal(tt.want, &dst); (err != nil) != tt.wantErr {
+				t.Errorf("decodeState.unmarshal() error = %v, wantErr %v", err, tt.wantErr)
+				return
+			}
+			diff := cmp.Diff(dst, tt.in, cmpopts.IgnoreUnexported(big.Int{}, VDTValue2{}, MyStructWithIgnore{}))
```
noot, line 286: haven't used this lib, would be cool to get a lil session on it potentially

noot (commented):
looks great, no major issues, just some questions/comments to be addressed. amazing work!

pkg/scale/decode.go
timwu20: there's a performance benefit if it falls into these cases, but technically no.  The `default` case that uses reflection will work in all these cases.

pkg/scale/comparison_test.go
```diff
@@ -0,0 +1,141 @@
...
+	if err != nil {
+		t.Errorf("unexpected err: %v", err)
+	}
+	if !reflect.DeepEqual(decoded, newDigest) {
+		t.Errorf("Unmarshal() = %v, want %v", decoded, newDigest)
+	}
+}
+
+func BenchmarkUnmarshal(b *testing.B) {
+	for _, tt := range allTests {
```
EclesioMeloJunior, line 118: You could use the N property from `testing.B` to get more info about the benchmark
```
for i := 0; i < b.N; i++ {
  for _, tt := range allTests {
	  ...
  }
}
```

Output without the `b.N` loop:

` BenchmarkUnmarshal-8   	1000000000	         0.0002203 ns/op	       0 B/op	       0 allocs/op`

Output with the `b.N` loop:

`BenchmarkUnmarshal-8   	    5090	    236443 ns/op	  360111 B/op	    2296 allocs/op`

EclesioMeloJunior (changes requested):

pkg/scale/result.go
```diff
@@ -0,0 +1,41 @@
...
+)
+
+type resultCache map[string]map[bool]interface{}
+type Result struct {
+	Ok  interface{}
+	Err interface{}
+}
+
+// Validate ensures Result is valid.  Only one of the Ok and Err attributes should be nil
+func (r Result) Validate() (err error) {
```
EclesioMeloJunior, line 15: Could this function return a boolean instead of an error? Returning a bool seems more appropriate to me since the responsibility of the function is to check some `Result` properties. The name could be `IsValid`

```
func (r Result) IsValid() bool {
  return (r.Ok == nil && r.Err != nil) || ( r.Ok != nil && r.Err == nil)
}
```

EclesioMeloJunior (changes requested):

pkg/scale/uint128.go
timwu20: order is variadic param, but it's assumed that only one order is passed in if at all.

pkg/scale/uint128.go
timwu20: based on the tests I ported over it, it's expecting it to be trimmed.  Do you think that the rust implementation doesn't trim?

pkg/scale/uint128.go
timwu20: OR cases.  I'll double check the original implementation.

timwu20:
```
func (u *Uint128) Cmp(other *Uint128) int {
	if u.Upper > other.Upper {
		return 1
	}

	if u.Upper < other.Upper {
		return -1
	}

	if u.Lower > other.Lower {
		return 1
	}

	if u.Lower < other.Lower {
		return -1
	}

	return 0
}
```

timwu20: I see what you mean here.  I'll rewrite.

pkg/scale/result.go
timwu20: I'm passing this error back to the caller, and using that.  If I change it to `IsValid` I would definitely use a `bool`, but then I'd have to create an error message after checking that `bool`.  I'm going to leave it as unless you feel really strongly about this.

pkg/scale/decode.go
timwu20: nope

pkg/scale/uint128.go
noot: if it matches the previous impl then let's leave it for now

pkg/scale/encode.go
timwu20: I ended up benchmarking this and I didn't see any discernible performance difference.  I'll remove these cases.

pkg/scale/result.go
timwu20: TODO @timwu20

pkg/scale/scale.go
timwu20: added tests for this.

pkg/scale/scale.go
timwu20: added tests for this.

pkg/scale/decode_test.go
timwu20: added new nil test cases for all test cases.  Maybe overkill, since it's testing the same type, but I'd rather do that than cherry pick tests from each subset of tests.

pkg/scale/encode.go
timwu20: removed this case, no longer needed.

pkg/scale/result.go
EclesioMeloJunior: I don't think the responsibility to create the error is up to this function. I mean the caller has more context of what the `result` is and if the `result` is not valid it could call another function or do something else or return a more specific error.

EclesioMeloJunior (changes requested):

pkg/scale/result.go
timwu20: 👍  updated

pkg/scale/result.go
noot: let me know when this is done and I can re-review, this should be the last thing to update :D

noot (approved):
`Result` looks good - let's get this in!! great work :D

EclesioMeloJunior (approved):
impressive work!! 👏

PR #1552, opened 2021-04-30
chore(docs): fix image paths for integrate docs page

dutterbutter:
## Changes

- fix image paths for integrate docs page
-
-

## Tests

```

```

## Issues

-


arijitAD (approved):

noot (approved):
doesn't look like the images work for me in the diff? but maybe that's cause the page isn't deployed

dutterbutter:
> doesn't look like the images work for me in the diff? but maybe that's cause the page isn't deployed

It works locally so it should work once deployed

PR #1554, opened 2021-05-03
feat(lib/grandpa): update grandpa catch-up process; add submodule, use NeighbourMessages

### LINKED ISSUE #1531: lib/grandpa: ensure catch-up logic works when interacting with substrate nodes

noot:
## Changes

- add `catchUp` submodule to track status of catch up and deal with sending out catch-up requests and handling responses
- update grandpa message handler to use `NeighbourMessage`s to trigger catch-up process if needed
- TODO: still need to test this w/ polkadot (seems we aren't getting responses, probably continue this is another PR)
- TODO: need to add test cases and update stress test

## Tests

```
go test ./lib/grandpa -short
```

## Issues

- closes #1531


danforbes:
It seems that the GRANDPA catch-up mechanism was implemented before request/response protocols were defined. I believe that these messages should be sent as "request notifications" and "response notifications".

PR #1567, opened 2021-05-07
chore: add polkadotjs tests to ci

dutterbutter:
## Changes

-  add polkadotjs tests to ci
-
-

## Tests

```

```

## Issues

-


.github/workflows/tests.yml
```diff
@@ -156,3 +156,27 @@ jobs:
         name: Run stress
         run: |
           docker run chainsafe/gossamer:test sh -c "make it-grandpa"
```
edwardmack, line 158: I know this wasn't part of this PR's changes, but I see that there are two tests with name `Run stress`, perhaps one should be named `Run grandpa`?

```suggestion
        name: Run grandpa
        run: |
          docker run chainsafe/gossamer:test sh -c "make it-grandpa"
```

edwardmack (approved):
Nice additional, one minor (unrelated) comment.

timwu20 (approved):

dutterbutter:
@edwardmack I cant seem to get the polkadot.js tests to pass locally or in the CI? Any suggestions how I can improve the debugging of these tests to see whats going on? Can you confirm that they are also failing?

I currently running `make it-polkadotjs` but have also uncommented the `printf` statement to see js `data` output and only receive the following info:

```
   start_polkadotjs_test.go:47: 
        	Error Trace:	start_polkadotjs_test.go:47
        	Error:      	Received unexpected error:
        	            	exit status 1
        	Test:       	TestStartGossamerAndPolkadotAPI
        	Messages:   	[10 10 32 32 84 101 115 116 105 110 103 32 112 111 108 107 97 100 111 116 46 106 115 47 97 112 105 32 99 97 108 108 115 58 10 32 32 32 32 97 112 105 32 99 111 110 115 116 97 110 116 115 10]
--- FAIL: TestStartGossamerAndPolkadotAPI (20.01s)
``` 

Any suggestions would be appreciated!

edwardmack:
@dutterbutter It seems the issue was that the `rpc` module was removed from the StartGossamer util, polkadot.js api uses this to query what rpc methods are available.  (We should probably make `rpc` module default on if rpc is enabled).  I also changed the api version in package.json to 2.8.1, we should update to 4.6.2 sometime, but this breaks tests with ` Error: Detected -Infinity in Keys, should be >= 2 and <= 11`.

Un-commenting printf did not help because the script was failing before any actual tests were run, which is why there was no useful data available.  

It's working for me now.

Issue #1518, opened 2021-04-12
network: update syncQueue to track sent/received requests by hash


noot:
## Task summary

- currently, the syncQueue tracks requests by number, not hash
- this is fine when syncing the chain, but at the head, we want to request blocks by hash since there are small chain forks
- currently we only request blocks by hash when needing to get a parent block we don't currently have
- update syncQueue to produce and track requests by hash instead of number

EclesioMeloJunior:
Hi! The requests are blocks that we send to or receive from others peers? And track means to save at database?

noot:
@EclesioMeloJunior hey, yes the requests are `BlockRequestMessage` s that we sent to other peers, we don't track requests that we receive. to track means to cache it in the `syncQueue` struct, for example, this is the current mapping used which tracks requests by starting number: https://github.com/ChainSafe/gossamer/blob/development/dot/network/sync.go#L130

EclesioMeloJunior:
@noot, what is the use case of justificationRequestData?

EclesioMeloJunior:
closed with PR #1575

PR #1570, opened 2021-05-09
chore: create map to cache request by hash

EclesioMeloJunior:
## Changes

- Create the **isRequestDataCached** function that receives _*variadic.Uint64OrHash_ as a parameter and returns _(*requestData, bool)_ if startingBlock **IsHash** then try to load from _requestDataByHash_ sync.Map otherwise check if startingBlock **IsUint64** then load from _requestData_ sync.Map, return false when cannot found the record.
- Create a **requestDataByHash** sync.Map to cache requests when using StartingBlock as a hash

## Tests

Will check if the function isRequestDataCached works properly when there is data cached
```
go test -timeout 30s -run ^TestSyncQueue_isRequestDataCached$ github.com/ChainSafe/gossamer/dot/network
```

## Issues

#1518

### DIFF
--- a/dot/network/sync.go
+++ b/dot/network/sync.go
@@ -459,6 +459,7 @@ func (q *syncQueue) pushResponse(resp *BlockResponseMessage, pid peer.ID) error 
 	}
 
 	startHash := resp.BlockData[0].Hash
+
 	if _, has := q.justificationRequestData.Load(startHash); has && !resp.BlockData[0].Header.Exists() {
 		numJustifications := 0
 		justificationResponses := []*types.BlockData{}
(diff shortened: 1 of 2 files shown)


dot/network/sync.go
```diff
@@ -599,10 +630,14 @@ func (q *syncQueue) trySync(req *syncRequest) {
...
-		q.justificationRequestData.Store(req.req.StartingBlock.Hash(), requestData{
+		startingBlockHash := req.req.StartingBlock.Hash()
+		reqdata := requestData{
 			sent:     true,
 			received: false,
-		})
+		}
+
+		q.justificationRequestData.Store(startingBlockHash, reqdata)
+		q.requestDataByHash.Store(startingBlockHash, reqdata)
```
noot, line 640: remove this, since `requestDataByHash` should store requests that ask for the block header + body

noot:
@EclesioMeloJunior code looks good! could you re-open this as a branch on the gossamer repo if possible?

EclesioMeloJunior:
I will close this PR and change to the new one #1575

Issue #1479, opened 2021-03-18
create gossamer devnet and document issues


noot:
## Task summary

- create a gossamer devnet with multiple authority nodes
- let devnet run for a couple days and document any issues

noot:
first issue: the node doesn't seem to produce the BABE NextEpochData digest needed for the next epoch in the first block of an epoch, I thought this was produced by the runtime but it doesn't look like it is. this is the same for the substrate default node runtime and the polkadot/kusama runtimes. asked for clarification as to whether this is produced by the runtime or the host

danforbes:
Closing as duplicate as #592. I have moved any remaining Issues from this Epic under that one. @noot - do we still need to track this?

> first issue: the node doesn't seem to produce the BABE NextEpochData digest needed for the next epoch in the first block of an epoch, I thought this was produced by the runtime but it doesn't look like it is. this is the same for the substrate default node runtime and the polkadot/kusama runtimes. asked for clarification as to whether this is produced by the runtime or the host

PR #1572, opened 2021-05-10
fix(dot/sync): fix creating block response, fixes node sync between gossamer nodes

noot:
## Changes

- previously nodes were failing to create a block response because it was using the `blocktree.SubChain` function to determine the blocks to add to the response, but when the blocktree gets pruned on finalization this caused the node not to find the blocks
- update `CreateBlockResponse` to search the blockState by number instead
- also handle ascending/descending direction in the request

## Tests

```
go test ./dot/sync -short
```

## Issues

- related to #1479

### DIFF
--- a/dot/sync/message.go
+++ b/dot/sync/message.go
@@ -54,7 +54,8 @@ func (s *Service) CreateBlockResponse(blockRequest *network.BlockRequestMessage)
 		if startBlock == 0 {
 			startBlock = 1
 		}
-		block, err := s.blockState.GetBlockByNumber(big.NewInt(0).SetUint64(startBlock))
+
+		block, err := s.blockState.GetBlockByNumber(big.NewInt(0).SetUint64(startBlock)) //nolint
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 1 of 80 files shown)


dot/sync/message.go
```diff
@@ -25,122 +25,154 @@ import (
...
-	}
-
-	logger.Trace("subchain", "start", subchain[0], "end", subchain[len(subchain)-1])
-
 	responseData := []*types.BlockData{}
 
-	// TODO: check ascending vs descending direction
-	for _, hash := range subchain {
+	// ascending (ie child to parent)
+	if blockRequest.Direction == 0 {
```
timwu20, line 103:
```suggestion
	switch blockRequest.Direction {
	    case 0:
	    case 1:
	}
```
nit: I like switch statements when handling finite cases.

dot/sync/message.go
```diff
@@ -25,122 +25,154 @@ import (
...
+		return nil, err
+	}
 
-		// header
-		if (blockRequest.RequestedData & network.RequestedDataHeader) == 1 {
-			retData, err := s.blockState.GetHeader(hash)
-			if err == nil && retData != nil {
-				blockData.Header = retData.AsOptional()
-			}
+	blockData := new(types.BlockData)
```
timwu20, line 134:
```suggestion
	blockData := types.BlockData{
	   Hash: hash,
	   ....
	}
```

dot/sync/message.go
```diff
@@ -25,122 +25,154 @@ import (
...
+	blockData.Hash = hash
+	blockData.Header = optional.NewHeader(false, nil)
+	blockData.Body = optional.NewBody(false, nil)
+	blockData.Receipt = optional.NewBytes(false, nil)
+	blockData.MessageQueue = optional.NewBytes(false, nil)
+	blockData.Justification = optional.NewBytes(false, nil)
+
+	if (requestedData & network.RequestedDataHeader) == 1 {
```
timwu20, line 142: Can you just return if `requestData == []byte(0)`

noot: yeah, will update

edwardmack (approved):

timwu20 (approved):

PR #1574, opened 2021-05-11
feat(rpc/subscription): implement state_unsubscribeStorage

edwardmack:
## Changes

- Implements `state_unsubscribeStorage` to unsubscribe listener from storage updates.
-
-

## Tests

```

```

## Issues

-


noot (approved):
nice, please add a test case!

dot/rpc/subscription/websocket.go
```diff
@@ -209,6 +212,12 @@ func (c *WSConn) initStorageChangeListener(reqID float64, params interface{}) (u
 	return myObs.id, nil
 }
 
+func (c *WSConn) unsubscribeStorageListener(params interface{}) {
+	id := params.([]interface{})[0].(float64)
```
EclesioMeloJunior, line 216: Would be nice to have some type checking here to ensure we don't face nil pointer or interface conversion panic?

timwu20: I agree, should return an error if the type check fails.  At least we can log it out in the unlikely case it happens.

timwu20 (changes requested):

dot/rpc/subscription/websocket.go
```diff
@@ -209,6 +212,12 @@ func (c *WSConn) initStorageChangeListener(reqID float64, params interface{}) (u
...
+	observer := c.Subscriptions[uint(id)].(state.Observer)
```
arijitAD, line 217: This may panic because there could be different implementation of `Listener` interface.
```
observer, ok := c.Subscriptions[uint(id)].(state.Observer)
if !ok {
   return
}
```
Also, we should only unsubscribe storage observers.

arijitAD (changes requested):

dot/rpc/subscription/websocket.go
```diff
@@ -212,10 +213,32 @@ func (c *WSConn) initStorageChangeListener(reqID float64, params interface{}) (u
 	return myObs.id, nil
 }
 
-func (c *WSConn) unsubscribeStorageListener(params interface{}) {
-	id := params.([]interface{})[0].(float64)
-	observer := c.Subscriptions[uint(id)].(state.Observer)
+func (c *WSConn) unsubscribeStorageListener(reqID float64, params interface{}) {
+	switch v := params.(type) {
+	case []interface{}:
+		if l := len(v); l < 1 {
```
arijitAD, line 222:
```suggestion
		if len(v) == 0  {
```

dot/rpc/subscription/websocket_test.go
```diff
@@ -113,6 +113,52 @@ func TestWSConn_HandleComm(t *testing.T) {
...
+	c.WriteMessage(websocket.TextMessage, []byte(`{
+    "jsonrpc": "2.0",
+    "method": "state_unsubscribeStorage",
+    "params": [],
+    "id": 7}`))
+	_, msg, err = c.ReadMessage()
+	require.NoError(t, err)
+	require.Equal(t, []byte(`{"jsonrpc":"2.0","error":{"code":-32600,"message":"Invalid request"},"id":7}`+"\n"), msg)
+
+	c.WriteMessage(websocket.TextMessage, []byte(`{
```
arijitAD, line 135: According to spec, this shouldn't fail.
`subscriber ID` is either `string` or `U32`
https://github.com/w3f/PSPs/blob/psp-rpc-api/PSPs/drafts/psp-6.md#11116-state_unsubscribestorage-pubsub
Any reason we have `subscriber ID` as float?

dot/rpc/subscription/websocket.go
```diff
@@ -212,10 +213,32 @@ func (c *WSConn) initStorageChangeListener(reqID float64, params interface{}) (u
...
+			c.safeSendError(reqID, big.NewInt(-32600), "Invalid request")
+			return
+		}
+	default:
+		c.safeSendError(reqID, big.NewInt(-32600), "Invalid request")
```
arijitAD, line 227: Declare this as constant
```
big.NewInt(-32600)
```

arijitAD (changes requested):

dot/rpc/subscription/websocket.go
edwardmack: updated.

dot/rpc/subscription/websocket_test.go
edwardmack: Good point, I've updated so that it can handle strings as well.  I've got params as a interface{}, so that it can accept any type, then when I parse the types it seems that all number are treated as float64 (I think this is happens when the json is unmashalled).  I'm now taking the float64 or string and casting into uint.  Let me know if there is a better approach to handling these.

dot/rpc/subscription/websocket.go
edwardmack: I can't make big.NewInt(-32600) a const, so I've made -32600 a const and I'm using that.

edwardmack:
I've update this based on comments,  I believe it's ready for a re-review.

arijitAD (approved):

dot/rpc/subscription/messages.go
```diff
@@ -59,3 +65,18 @@ func newSubscriptionResponseJSON(subID uint, reqID float64) ResponseJSON {
 		ID:      reqID,
 	}
 }
+
+// BooleanResponseJSON for responses that return boolean values
+type BooleanResponseJSON struct {
```
timwu20, line 70: nit: 
```suggestion
type BooleanResponse struct {
```

dot/rpc/subscription/messages.go
```diff
@@ -59,3 +65,18 @@ func newSubscriptionResponseJSON(subID uint, reqID float64) ResponseJSON {
...
+	Jsonrpc string  `json:"jsonrpc"`
+	Result  bool    `json:"result"`
+	ID      float64 `json:"id"`
+}
+
+func newBooleanResponseJSON(value bool, reqID float64) BooleanResponseJSON {
```
timwu20, line 76: if the type is public why not the constructor?

timwu20 (approved):

dot/rpc/subscription/messages.go
```diff
@@ -59,3 +65,18 @@ func newSubscriptionResponseJSON(subID uint, reqID float64) ResponseJSON {
 		ID:      reqID,
 	}
 }
+
+// BooleanResponseJSON for responses that return boolean values
+type BooleanResponseJSON struct {
+	Jsonrpc string  `json:"jsonrpc"`
```
timwu20, line 71:
```suggestion
	JSONRPC string  `json:"jsonrpc"`
```

EclesioMeloJunior (approved):

dot/rpc/subscription/messages.go
edwardmack: Updated.

dot/rpc/subscription/messages.go
edwardmack: No reason... I've updated this.

dot/rpc/subscription/messages.go
edwardmack: updated.

PR #1575, opened 2021-05-11
fix(dot/network) track sent/received requests by hash

EclesioMeloJunior:
## Changes

- Create the **isRequestDataCached** function that receives _*variadic.Uint64OrHash_ as a parameter and returns _(*requestData, bool)_ if startingBlock **IsHash** then try to load from _requestDataByHash_ sync.Map otherwise check if startingBlock **IsUint64** then load from _requestData_ sync.Map, return false when cannot found the record.
- Create a **requestDataByHash** sync.Map to cache requests when using StartingBlock as a hash

## Tests

Will check if the function isRequestDataCached works properly when there is data cached
```
go test -timeout 30s -run ^TestSyncQueue_isRequestDataCached$ github.com/ChainSafe/gossamer/dot/network
```

## Issues

#1518


EclesioMeloJunior:
@noot, on the **trySync** function if the `req.StartingBlock.IsHash()` and `req.req.RequestedData == (RequestedDataHeader + RequestedDataBody)` then we store the request?

noot (approved):
looks good to me - nice work :)

edwardmack (approved):
Great work.

Issue #1569, opened 2021-05-07
polkadot.js UI transaction pending bubble not leaving after transaction is done


noot:
## Describe the bug

- in the polkadot.js UI, if you send a transaction of any type, the blue pending bubble in the top right corner that shows the transaction you sent doesn't go away when the transaction has been included
- this bubble should leave after the transaction is included
- this probably indicates that the UI is not properly getting notified when a transaction is included

## Possible Solution

- maybe related to `submitAndWatchExtrinsic`?

## To Reproduce
Steps to reproduce the behaviour:

1. follow instructions here https://chainsafe.github.io/gossamer/integrate/connect-to-polkadot-js/
2.  send any sort of transaction

EclesioMeloJunior:
When we send a new transaction we save the request extrinsic bytes into the property `extrinsic` from the listener `ExtrinsicSubmitListener` ([here](https://github.com/ChainSafe/gossamer/blob/eef03c6bdcf4cb358ef597924ef82bcb45b0427f/dot/rpc/subscription/websocket.go#L276))

I have noticed that when the listener receives an imported block the value of `block.Body.AsExtrinsics()` doesn't match with the one registered on `ExtrinsicSubmitListener.extrinsic` ([here](https://github.com/ChainSafe/gossamer/blob/eef03c6bdcf4cb358ef597924ef82bcb45b0427f/dot/rpc/subscription/listeners.go#L170))

I suspected that it would be other blocks that were arriving, so I started to log the extrinsic and the hash of the block, and in fact, the transaction-related block doesn't match the extrinsic registered in the listener

1. I create a new transaction then I logged this transaction extrinsic
![init-submitExtrinsicWatch](https://user-images.githubusercontent.com/17255488/118317538-3293bf80-b4c6-11eb-8a90-04e710f0f60e.png)


2. When the listener catch the imported block I logged some infos like, extrinsic of the block and the block hash, as we can see the function `block.Body.AsExtrinsics()` returns a set of extrinsics to compare with the one stored in the listener, but no one can be matched.
![imported-block-arrives](https://user-images.githubusercontent.com/17255488/118317550-37587380-b4c6-11eb-97b0-c922b8fce0d1.png)

3. The block number 4 is the block with the transaction data
![block-number 4](https://user-images.githubusercontent.com/17255488/118317629-4ccd9d80-b4c6-11eb-9ff2-c5fd71ba1c29.png)
[...]

PR #1592, opened 2021-05-17
fix: pending bubble hidden after block included

EclesioMeloJunior:
## Changes

- Include `types.Body.FindExtrinsic` function that tries to find out an extrinsic (encoded or not) inside the body byte array.
- `ExtrinsicSubmitListener.Listen` function was updated  to calls `body.FindExtrinsic` instead of loop and compare the extrinsic directly

## Tests

```
go test -run ^TestBody_FindEncodedExtrinsic$ github.com/ChainSafe/gossamer/dot/types
go test -run ^TestDecodeExtrinsicBody$ github.com/ChainSafe/gossamer/lib/babe
```

## Issues

- #1569 

## Screenshots

- After the block was included the pending bubble disappear
success-block-included


dot/types/body.go
```diff
@@ -173,3 +175,35 @@ func decodeOptionalBody(r io.Reader) (*optional.Body, error) {
 
 	return optional.NewBody(false, nil), nil
 }
+
+// FindExtrinsic returns true if body contains target Extrisic
+// returns error when fails to encode decoded extrinsic on body
+func (b *Body) FindExtrinsic(target Extrinsic) (bool, error) {
```
noot, line 181: nit but maybe name `HasExtrinsic`?

noot (approved):
works with api v4.6, great work!

timwu20 (approved):

PR #1594, opened 2021-05-18
fix(dot/network): fix discovery between gossamer nodes

### LINKED ISSUE #1369: dot/network: reconnect to peers if peer count is below minimum
## Task summary

- if our peer count drops below our minimum desired peer count, we should try to connect to peers we've discovered (that are saved in our peerstore)
- should be done in `ConnManager.Disconnected`

### LINKED ISSUE #1586: devnet: ensure discovery works between gossamer nodes
## Task summary

- it appears DHT discovery isn't working properly between gossamer nodes on different machines, as peers only connect when directly bootstrapped to each other
- fix this and allow the nodes to discover each other via DHT

noot:
## Changes

- refactor discovery related code to be part of a `*discovery` type
- fix discovery between gossamer nodes by 
1. adding advertisement, 
2.  waiting for the node to have some peers before starting the DHT, this is an issue in how the DHT is written where the DHT bootnodes can only be set on initialization and not updated afterwards, so calling `bootstrap` on the DHT afterwards results in empty routing tabe errors. see https://github.com/libp2p/go-libp2p-kad-dht/issues/295 
 3. explicity telling the node to listen on its external address, previously it was only listening on 127.0.0.1.

## Tests

```
go test ./dot/network -short
```

alternatively spin up 3 gossamer nodes A, B, C and connect A to B, and A to C and ensure B and C can discover each other (this worked locally before, it was only really reproducable in the devnet environment) 

## Issues

- closes #1586 
- closes #1369


dot/network/discovery.go
```diff
@@ -0,0 +1,191 @@
...
+// start creates the DHT.
+func (d *discovery) start() error {
+	if len(d.bootnodes) == 0 {
+		// get all currently connected peers and use them to bootstrap the DHT
+		peers := d.h.Network().Peers()
+
+		for {
+			if len(peers) == 0 {
+				logger.Info("no peers yet, waiting to start DHT...")
+				time.Sleep(time.Second * 10) // wait for peers to connect before starting DHT
```
EclesioMeloJunior, line 66: maybe this `time.Second * 10` could be a constant?

dot/network/discovery.go
```diff
@@ -0,0 +1,191 @@
...
+	rd := libp2pdiscovery.NewRoutingDiscovery(d.dht)
+
+	err := d.dht.Bootstrap(d.ctx)
+	if err != nil {
+		return fmt.Errorf("failed to bootstrap DHT: %w", err)
+	}
+
+	// wait to connect to bootstrap peers
+	time.Sleep(time.Second)
+	peersToTry := make(map[*peer.AddrInfo]struct{})
```
EclesioMeloJunior, line 117: `peersToTry` could not be a slice of `*peer.AddrInfo`?

dot/network/discovery.go
```diff
@@ -0,0 +1,191 @@
...
+
+				// found a peer, try to connect if we need more peers
+				if len(d.h.Network().Peers()) < d.maxPeers {
+					err = d.h.Connect(d.ctx, peer)
+					if err != nil {
+						logger.Info("failed to connect to discovered peer", "peer", peer.ID, "err", err)
+					}
+				} else {
+					d.h.Peerstore().AddAddrs(peer.ID, peer.Addrs, peerstore.PermanentAddrTTL)
+					peersToTry[&peer] = struct{}{}
```
EclesioMeloJunior, line 183: maybe will we need to check if the peer that arrived is already on the list?

EclesioMeloJunior: Will be a good idea if we have a limit to add new peers on the `peersToTry` list? then we avoid a long list, and when the list is already full, we test the connection of the first item (most aged peer) and if the connection fails, then we remove it and add the new peer to the end of the list (newest peer)?

EclesioMeloJunior (changes requested):

dot/network/discovery.go
noot: yes I will update all these constants before un-drafting the PR

dot/network/discovery.go
noot: it's a map so that there's no duplicates

dot/network/discovery.go
noot: for the first comment, the map ensures no duplicates. and yes I'm not 100% sure about using the peersToTry list yet or just using the peerstore, either way I'm going to clean it up before opening the PR

dot/network/host.go
```diff
@@ -136,7 +140,7 @@ func newHost(ctx context.Context, cfg *Config) (*host, error) {
 					ok = append(ok, addr)
 				}
 			}
-			return ok
+			return append(ok, externalAddr)
```
EclesioMeloJunior, line 143: If `pubip.Get` function fails then `externalAddr` will be empty, will be a better approach to only appending it if it is not empty?

noot: good point I'll update it to append only if there's no error

EclesioMeloJunior (approved):
lgmt :tada:

dot/network/discovery.go
```diff
@@ -0,0 +1,203 @@
...
+	rd := libp2pdiscovery.NewRoutingDiscovery(d.dht)
+
+	err := d.dht.Bootstrap(d.ctx)
+	if err != nil {
+		return fmt.Errorf("failed to bootstrap DHT: %w", err)
+	}
+
+	// wait to connect to bootstrap peers
+	time.Sleep(time.Second)
+	peersToTry := make(map[*peer.AddrInfo]struct{})
```
arijitAD, line 129: Move this inside go routine.

dot/network/host.go
```diff
@@ -136,7 +140,11 @@ func newHost(ctx context.Context, cfg *Config) (*host, error) {
 					ok = append(ok, addr)
```
arijitAD, line 140: Rename `ok` to `addrs`

arijitAD (approved):

PR #1597, opened 2021-05-18
fix (dot/telemetry): refactor telemetry to reduce CPU usage

### LINKED ISSUE #1577: improve telemetry CPU usage
## Task summary

- currently telemetry uses a lot of CPU especially while syncing
- profiling shows this is due to waiting for the lock in `telemetry.Handler.sender`
- look into if there is a way to optimize this 
- also ensure the --no-telemetry flag fully turns off telemetry, since it seems right now it isn't entirely off

edwardmack:
## Changes

- Refactor telemetry to remove use of logging to report telemetry data, implemented using go channel.

## Tests

```
go test ./dot/telemetry/...
```

## Issues

- closes #1577


dot/telemetry/telemetry.go
```diff
@@ -57,126 +54,81 @@ var (
...
-		h.RLock()
-		line, err := h.buf.ReadBytes(byte(10)) // byte 10 is newline character, used as delimiter
-		h.RUnlock()
-		if err != nil {
-			continue
-		}
-
-		for _, c := range h.wsConn {
-			err := c.WriteMessage(websocket.TextMessage, line)
+		msg := <-t.msg
```
EclesioMeloJunior, line 113: In this case, is possible to have a context on `Handler` struct? like:

```
select {
case msg := <-t.msg: 
  // exec the logic here when a msg arrives

case <-t.ctx.Done():
  //break the loop
  break
}
```

dot/network/service.go
```diff
@@ -316,14 +317,11 @@ func (s *Service) sentBlockIntervalTelemetry() {
...
-		telemetry.GetInstance().SendBlockIntervalData(&telemetry.BlockIntervalData{
-			BestHash:           best.Hash(),
-			BestHeight:         best.Number,
-			FinalizedHash:      finalized.Hash(),
-			FinalizedHeight:    finalized.Number,
-			TXCount:            0, // todo (ed) determine where to get tx count
-			UsedStateCacheSize: 0, // todo (ed) determine where to get used_state_cache_size
-		})
+		telemetry.GetInstance().SendMessage(telemetry.NewTelemetryMessage(telemetry.NewKeyValue("best", best.Hash().String()),
+			telemetry.NewKeyValue("finalized_hash", finalized.Hash().String()), telemetry.NewKeyValue("finalized_height", finalized.Number), //nolint
```
noot, line 321: the block here is the best hash, not the best finalized hash

dot/telemetry/telemetry.go
noot: agree, `return` if `<-t.ctx.Done()`

dot/telemetry/telemetry.go
```diff
@@ -57,126 +54,81 @@ var (
...
+		for _, v := range t.connections {
+			err := v.wsconn.WriteMessage(websocket.TextMessage, msgToBytes(msg))
 			if err != nil {
 				// TODO (ed) determine how to handle this error
 				fmt.Printf("ERROR connecting to telemetry %v\n", err)
 			}
 		}
```
noot, line 120: put this in a goroutine so that reading/writing to/from `t.msg` doesn't block

noot (approved):
looks good, cpu profile doesn't show telemetry anymore!! great stuff :D

dot/telemetry/telemetry.go
```diff
@@ -57,126 +54,81 @@ var (
...
-
-		for _, c := range h.wsConn {
-			err := c.WriteMessage(websocket.TextMessage, line)
+		msg := <-t.msg
+		for _, v := range t.connections {
+			err := v.wsconn.WriteMessage(websocket.TextMessage, msgToBytes(msg))
 			if err != nil {
 				// TODO (ed) determine how to handle this error
 				fmt.Printf("ERROR connecting to telemetry %v\n", err)
 			}
```
noot, line 119: can you ignore this error for now - I sometimes see a lot of logs of this

dot/telemetry/telemetry.go
```diff
@@ -57,126 +54,81 @@ var (
...
 
 // GetInstance singleton pattern to for accessing TelemetryHandler
-func GetInstance() *Handler {
+func GetInstance() *Handler { //nolint
 	if handlerInstance == nil {
 		once.Do(
 			func() {
 				handlerInstance = &Handler{
-					buf: bytes.Buffer{},
+					msg: make(chan Message, 10),
```
arijitAD, line 62: Increase the buffer size to `256`. 
It may slow down the node if the buffer is full since `SendMessage` is called form `handleBlock`

dot/telemetry/telemetry.go
```diff
@@ -57,126 +54,81 @@ var (
...
+	for _, v := range values {
+		mvals[v.key] = v.value
+	}
+	return &Message{
+		values: mvals,
+	}
+}
+
+// NewKeyValue builds a key value pair for telemetry messages
+func NewKeyValue(key string, value interface{}) KeyValue { //nolint
```
arijitAD, line 82: Make this function return pointer.

dot/telemetry/telemetry.go
edwardmack: A added context to the Handler struct.

dot/network/service.go
edwardmack: @noot I'm confused, in there a difference between finalized hash and best finalized hash?  How do I obtain finalized hash?

dot/telemetry/telemetry.go
edwardmack: Removed printf.

dot/telemetry/telemetry.go
edwardmack: I tried putting this in a go routine, however then I was getting `panic: concurrent write to websocket connection`.  So then I tried protecting websocket connection with a mutex, but now the linter complains `copylocks: range var v copies lock: github.com/ChainSafe/gossamer/dot/telemetry.telemetryConnection contains sync.Mutex (govet)`  and it still panics with concurrent write to websocket.  So, I must be handling this in-correctly, any suggestions?  @noot, @arijitAD, @timwu20, @EclesioMeloJunior

dot/telemetry/telemetry.go
edwardmack: updateded to 256.

dot/telemetry/telemetry.go
edwardmack: Made function return pointer.

edwardmack:
There is a issue with the way I'm handling websocket connections.  DO NOT MERGE until this is fixed, currently panics with concurrent websocket writes.  Please advise if you can suggest a solution.

dot/telemetry/telemetry.go
edwardmack: Disregard above comment, I've moved mutex to Handler struct, where it makes more sense.

edwardmack:
I've updated to fix to concurrent write issue.  This is ready for a re-review.

dot/telemetry/telemetry.go
```diff
@@ -57,126 +57,89 @@ var (
...
-func NewNetworkData(peers int, rateIn, rateOut float64) *NetworkData {
-	return &NetworkData{
-		peers:   peers,
-		rateIn:  rateIn,
-		rateOut: rateOut,
+		tConn := telemetryConnection{
+			wsconn:    c,
+			verbosity: v.Verbosity,
+		}
+		t.connections = append(t.connections, tConn)
```
arijitAD, line 106: Take lock in `AddConnections` because `connections` is concurrently accessed.

dot/telemetry/telemetry.go
```diff
@@ -57,126 +57,89 @@ var (
...
+				t.mtx.Lock()
+				for _, v := range t.connections {
+					err := v.wsconn.WriteMessage(websocket.TextMessage, msgToBytes(msg))
+					if err != nil {
+						// TODO (ed) determine how to handle this error
+					}
+				}
+				t.mtx.Unlock()
+			}()
+		case <-t.ctx.Done():
```
arijitAD, line 129: `context` is not required since we are never canceling it. This will never be triggered.

arijitAD (changes requested):

dot/network/service.go
arijitAD:
```
telemetry.GetInstance().SendMessage(telemetry.NewTelemetryMessage(
	telemetry.NewKeyValue("best", best.Hash().String()),
	telemetry.NewKeyValue("finalized_hash", finalized.Hash().String()),
	telemetry.NewKeyValue("finalized_height", finalized.Number), //nolint
	telemetry.NewKeyValue("height", best.Number), telemetry.NewKeyValue("msg", "system.interval"),
	telemetry.NewKeyValue("txcount", 0),                // todo (ed) determine where to get tx count
	telemetry.NewKeyValue("used_state_cache_size", 0)), // todo (ed) determine where to get used_state_cache_size
)
```

Also, have a single `NewKeyValue` per line.

dot/telemetry/telemetry.go
arijitAD: You can create a slice of `tConn` since `Dial` may take time and we don't want to hold the lock. Add all the successful `tConns` to the `connections` after the loop.

dot/network/service.go
noot: yes, nevermind, I read this wrong, it's actually fine

dot/telemetry/telemetry.go
```diff
@@ -117,14 +117,14 @@ func (t *Handler) startListening() {
 		select {
 		case msg := <-t.msg:
 			go func() {
+				t.mtx.Lock()
```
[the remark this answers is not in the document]
edwardmack: I've added TestListenerConcurrency to telemetry_test.go that sends multiple SendMessage requests, trying to generate `panic: concurrent write to websocket connection`  This new test passes now for this updated PR.  

I've created a branch  `ed/tel_channelRefactorRace` where I use the code above with the new test.  When running the new test with this code, it fails with panic: concurrent write...  Also, when I do make lint on the the ed/tel_channelRefactorRace branch I get these issues:
```
emack@emack-Oryx-Pro:~/projects/chainsafe/gossamer$ make lint
./scripts/install-lint.sh
/home/emack/go/bin/golangci-lint run
dot/telemetry/telemetry.go:106:41: copylocks: call of append copies lock value: github.com/ChainSafe/gossamer/dot/telemetry.telemetryConnection (govet)
                t.connections = append(t.connections, tConn)
                                                      ^
dot/telemetry/telemetry.go:120:12: copylocks: range var v copies lock: github.com/ChainSafe/gossamer/dot/telemetry.telemetryConnection (govet)
                                for _, v := range t.connections {
                                       ^
make: *** [Makefile:24: lint] Error 1

```

dot/telemetry/telemetry.go
edwardmack: removed context

edwardmack:
I think I've addressed all comments, this is ready for a re-review.

dot/telemetry/telemetry_test.go
```diff
@@ -113,6 +113,32 @@ func TestHandler_SendMulti(t *testing.T) {
 	require.Contains(t, string(actual[3]), string(expected[3]))
 }
 
+func TestListenerConcurrency(t *testing.T) {
```
arijitAD, line 116: Thanks for the test. This helped me to debug the issue.

You can take a lock at the connection level which is optimized than taking the lock at the handler level because you can send multiple messages concurrently to different connections.
This issue is `connection` in `Handler.connections` is value and not pointer. Hence it creates multiple instances of lock for the same connection.

Add lock at the connection level but change `Handler.connections` to a pointer.
```
type telemetryConnection struct {
	wsconn    *websocket.Conn
	verbosity int
	sync.Mutex
}

// Handler struct for holding telemetry related things
type Handler struct {
	msg         chan Message
	connections []*telemetryConnection
}

func (t *Handler) startListening() {
	for {
		msg := <-t.msg
		go func() {
			for _, v := range t.connections {
				v.Lock()
				_ = v.wsconn.WriteMessage(websocket.TextMessage, msgToBytes(msg))
				v.Unlock()
			}
		}()
	}
}

```

Also, I think users should get notified in logs if they are unable to write messages to the telemetry server. So that they can take action accordingly.

arijitAD (changes requested):

dot/telemetry/telemetry.go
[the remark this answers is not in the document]
arijitAD: I think it related to copying locks as I mentioned here. https://github.com/ChainSafe/gossamer/pull/1597#discussion_r640342206

dot/telemetry/telemetry_test.go
edwardmack: Thanks for debugging this, it makes sense now.  I've updated to fix it, and added logging to catch any error and warn the user.

arijitAD (approved):

dot/telemetry/telemetry.go
```diff
@@ -57,126 +56,85 @@ var (
...
-// SendBlockImport sends block imported message to telemetry connection
-func (h *Handler) SendBlockImport(bestHash string, height *big.Int) {
-	h.Lock()
-	defer h.Unlock()
-	payload := log.Fields{"best": bestHash, "height": height.Int64(), "msg": "block.import", "origin": "NetworkInitialSync"}
-	telemetryLogger := log.WithFields(log.Fields{"id": 1, "payload": payload, "ts": time.Now()})
-	telemetryLogger.Print()
+// SendMessage sends Message to connected telemetry listeners
+func (t *Handler) SendMessage(msg *Message) {
+	t.msg <- *msg
```
timwu20, line 111: improvement: add a timeout and error return for this method.  I see that that it's a buffered channel of length 256, so this shouldn't happen, but it's a good practice so that this function call isn't blocking.

dot/telemetry/telemetry.go
```diff
@@ -57,126 +56,85 @@ var (
...
-
-// SendBlockImport sends block imported message to telemetry connection
-func (h *Handler) SendBlockImport(bestHash string, height *big.Int) {
-	h.Lock()
-	defer h.Unlock()
-	payload := log.Fields{"best": bestHash, "height": height.Int64(), "msg": "block.import", "origin": "NetworkInitialSync"}
-	telemetryLogger := log.WithFields(log.Fields{"id": 1, "payload": payload, "ts": time.Now()})
-	telemetryLogger.Print()
+// SendMessage sends Message to connected telemetry listeners
+func (t *Handler) SendMessage(msg *Message) {
```
timwu20, line 110:
```suggestion
func (h *Handler) SendMessage(msg *Message) {
```

dot/telemetry/telemetry.go
```diff
@@ -57,126 +56,85 @@ var (
...
-func NewNetworkData(peers int, rateIn, rateOut float64) *NetworkData {
-	return &NetworkData{
-		peers:   peers,
-		rateIn:  rateIn,
-		rateOut: rateOut,
+func (t *Handler) startListening() {
+	for {
+		msg := <-t.msg
+		go func() {
+			for _, v := range t.connections {
```
timwu20, line 118:
```suggestion
			for _, conn := range t.connections {
```

dot/telemetry/telemetry.go
```diff
@@ -57,126 +56,85 @@ var (
...
-// NewNetworkData creates networkData struct
-func NewNetworkData(peers int, rateIn, rateOut float64) *NetworkData {
-	return &NetworkData{
-		peers:   peers,
-		rateIn:  rateIn,
-		rateOut: rateOut,
+func (t *Handler) startListening() {
+	for {
+		msg := <-t.msg
+		go func() {
```
timwu20, line 117: do you care about sending messages in order to all the connections? It doesn't look like you're waiting for the goroutine to finish to broadcast to all connections.  This will essentially just spam it out, and it may be out of sequence.  Maybe add a `sync.WaitGroup` if you want the sequential behaviour.

dot/telemetry/telemetry.go
```diff
@@ -57,126 +56,85 @@ var (
...
-
-		for _, c := range h.wsConn {
-			err := c.WriteMessage(websocket.TextMessage, line)
-			if err != nil {
-				// TODO (ed) determine how to handle this error
-				fmt.Printf("ERROR connecting to telemetry %v\n", err)
-			}
-		}
+func msgToBytes(message Message) []byte {
+	res := make(map[string]interface{})
```
timwu20, line 131: maybe create a helper struct to construct the json response.
```
type response struct {
  ID int `json:"id"`
  ....
}
```

dot/sync/syncer.go
```diff
@@ -364,7 +364,12 @@ func (s *Service) handleBlock(block *types.Block) error {
 		}
 	} else {
 		logger.Debug("🔗 imported block", "number", block.Header.Number, "hash", block.Header.Hash())
-		telemetry.GetInstance().SendBlockImport(block.Header.Hash().String(), block.Header.Number)
+		telemetry.GetInstance().SendMessage(telemetry.NewTelemetryMessage(
+			telemetry.NewKeyValue("best", block.Header.Hash().String()),
+			telemetry.NewKeyValue("height", block.Header.Number.Uint64()),
+			telemetry.NewKeyValue("msg", "block.import"),
+			telemetry.NewKeyValue("origin", "NetworkInitialSync")))
+		// todo(ed) add timer to avoid a lot of sends
```
noot, line 372: does this cause a problem?

noot: also, is `block.import` used for all new builds (ie ones we built as well?) if so this might be better placed in `BlockState.AddBlock`

dot/telemetry/telemetry.go
```diff
@@ -57,126 +56,85 @@ var (
...
+}
+
+// AddConnections adds the given telemetry endpoint as listeners that will receive telemetry data
+func (t *Handler) AddConnections(conns []*genesis.TelemetryEndpoint) {
 	for _, v := range conns {
 		c, _, err := websocket.DefaultDialer.Dial(v.Endpoint, nil)
 		if err != nil {
-			fmt.Printf("Error %v\n", err)
+			// todo (ed) try reconnecting if there is an error connecting
+			t.log.Warn("issue adding telemetry connection", err)
```
noot, line 98: maybe change to debug

noot, line 98: logger needs even arguments (apart from the initial message string)

noot, line 98:
```suggestion
			t.log.Warn("issue adding telemetry connection", "error", err)
```

dot/telemetry/telemetry.go
```diff
@@ -57,126 +56,85 @@ var (
...
-		rateOut: rateOut,
+func (t *Handler) startListening() {
+	for {
+		msg := <-t.msg
+		go func() {
+			for _, v := range t.connections {
+				v.Lock()
+				err := v.wsconn.WriteMessage(websocket.TextMessage, msgToBytes(msg))
+				if err != nil {
+					t.log.Warn("issue while sending telemetry message", err)
```
noot, line 122:
```suggestion
					t.log.Warn("issue while sending telemetry message", "error", err)
```

noot: also please change to debug or just ignore, I can see the node getting spammed with this sometimes

dot/telemetry/telemetry.go
edwardmack: Good suggestion, I've added timeout and return error.

dot/telemetry/telemetry.go
edwardmack: Updated.

dot/telemetry/telemetry.go
edwardmack: Updated.

dot/telemetry/telemetry.go
edwardmack: Good point, I don't think order matters with these messages.

dot/telemetry/telemetry.go
edwardmack: Good idea, created.

dot/sync/syncer.go
edwardmack: Good point, it hasn't yet.

dot/telemetry/telemetry.go
edwardmack: Good catch, updated.

edwardmack: Updated.

dot/telemetry/telemetry.go
edwardmack: Updated.

PR #1600, opened 2021-05-19
chore(dot/network): use sync.Pool for network message buffers

### LINKED ISSUE #1576: dot/network: use sync.Pool for network message buffers
## Task summary

- currently in `network.Service.readStream` we allocate a buffer to read from the stream that the remote peer opened
- however based on recent CPU profiling, the `malloc`ing takes up a LOT of CPU (especially now that we handle the inbound stream properly)
- we should use `sync.Pool` to allocate stream buffers instead, so we `malloc` a `maxPeers` number of buffers on node startup and then take/return them to the pool as needed

- https://golang.org/pkg/sync/#Pool

noot:
## Changes

- use `sync.Pool` to get message buffers to read stream data from, improves CPU usage quite a bit since we no longer need to always allocate for a new stream 
- pre-allocate the pool when creating the service with `minPeers * 3` buffers

## Tests

```
go test ./dot/network -short
```

also, you can check the profiling by running the node with and without this PR (for example a polkadot node) with the `--cpuprof cpu.out` flag set and checking the output profile with `go tool pprof` - should no longer see that `readStream` takes up a lot of CPU

## Issues

- closes #1576


dot/network/service.go
```diff
@@ -550,15 +582,39 @@ func isInbound(stream libp2pnetwork.Stream) bool {
 	return stream.Stat().Direction == libp2pnetwork.DirInbound
 }
 
+func (s *Service) getMessageBuffer() [maxMessageSize]byte {
+	buf := s.bufPool.Get()
```
arijitAD, line 586: You can directly do this.
```
return s.bufPool.Get().([maxMessageSize]byte)
```
`Sync.Pool` will never return nil since you have defined `New` for it and since we are using the sync pool for specifically storing `[maxMessageSize]byte`. Interface conversion will never panic.

dot/network/service.go
```diff
@@ -72,6 +74,9 @@ type Service struct {
 	mdns      *mdns
 	gossip    *gossip
 	syncQueue *syncQueue
+	bufPool   *sync.Pool
+	bufMap    *sync.Map //map[*[maxMessageSize]byte]struct{}
+	bufCount  int
```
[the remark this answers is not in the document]
arijitAD: You can also use channels instead of `bufMap` for the keeping the reference of the buffer.
Using channel you can easily get it's length so need of `bufCount`

arijitAD:
I did some initial benchmarking to implement fixed sized buffer pool.

BenchmarkBufferPool3: Fixed size buffer pool using channels
BenchmarkBufferPool2: SyncPool, uses channel for keeping reference.
BenchmarkBufferPool1: SyncPool, uses sync.Map and atomic counter for keeping reference.

BenchmarkBufferPool3 performs the best.
BenchmarkBufferPool1 is the least performant.

```

func BenchmarkBufferPool1(b *testing.B) {
	service := new(Service)

	service.cfg = &Config{
		MinPeers: 3,
	}
	service.bufPool = new(sync.Pool)
	service.bufPool.New = func() interface{} {
		return [maxMessageSize]byte{}
	}
	service.bufMap = new(sync.Map)

	for i := 0; i < service.cfg.MinPeers*3; i++ {
		buf := [maxBlockResponseSize]byte{}
		service.bufPool.Put(buf)
		service.bufMap.Store(&buf, struct{}{})
	}
	service.bufCount = service.cfg.MinPeers * 3

	b.ResetTimer()
	for i := 0; i < b.N; i++ {
		for i := 0; i < 100; i++ {
			service.cleanupMessageBuffer(service.getMessageBuffer())
		}
	}
}

func BenchmarkBufferPool2(b *testing.B) {
	service := new(Service)

	service.cfg = &Config{
		MinPeers: 1,
		MaxPeers: 3,
	}
	service.bufPool = new(sync.Pool)
	service.bufPool.New = func() interface{} {
		return [maxMessageSize]byte{}
	}
	service.bufRefCh = make(chan *[maxMessageSize]byte, service.cfg.MinPeers*3+10)

	for i := 0; i < service.cfg.MinPeers*3; i++ {
		buf := [maxBlockResponseSize]byte{}
		service.bufPool.Put(buf) //nolint
		service.bufRefCh <- &buf
	}
	service.bufCount = service.cfg.MinPeers * 3

	b.ResetTimer()
	for i := 0; i < b.N; i++ {
		for i := 0; i < 100; i++ {
			service.cleanupMessageBuffer(service.getMessageBuffer())
		}
	}
}

func BenchmarkBufferPool3(b *testing.B) {
	service := new(Service)
	service.cfg = &Config{
		MinPeers: 1,
		MaxPeers: 3,
	}

	bpool := NewSizedBufferPool(service.cfg.MaxPeers, service.cfg.MinPeers)
	b.ResetTimer()
	for i := 0; i < b.N; i++ {
		for i := 0; i < 100; i++ {
			bpool.Put(bpool.Get())
		}
	}
}

type SizedBufferPool struct {
	c chan *[maxMessageSize]byte
}

func NewSizedBufferPool(max int, min int) (bp *SizedBufferPool) {
	bufferCh := make(chan *[maxMessageSize]byte, max)

	for i := 0; i < min; i++ {
		buf := [maxMessageSize]byte{}
		bufferCh <- &buf
	}

	return &SizedBufferPool{
		c: bufferCh,
	}
}

// Get gets a Buffer from the SizedBufferPool, or creates a new one if none are
// available in the pool. Buffers have a pre-allocated capacity.
func (bp *SizedBufferPool) Get() [maxMessageSize]byte {
	var buff *[maxMessageSize]byte
	select {
	case buff = <-bp.c:
	// reuse existing buffer
	default:
		// create new buffer
		buff = &[maxMessageSize]byte{}
	}
	return *buff
}

// Put returns the given Buffer to the SizedBufferPool.
func (bp *SizedBufferPool) Put(b [maxMessageSize]byte) {
	select {
	case bp.c <- &b:
	default: // Discard the buffer if the pool is full.
	}
}
```

arijitAD (changes requested):

noot:
@arijitAD  do you suggest I update this to use the fixed size buffer pool with channels instead? happy to update if it's more performant!

arijitAD:
> @arijitAD do you suggest I update this to use the fixed size buffer pool with channels instead? happy to update if it's more performant!

While running benchmark it was 2x faster and it is also simple code-wise. You can probably plug it in and see if the profile still has `readStream `.

`Sync.Pool` alone is highly performant but the`sync.Map` slows it down. Since we need to keep the reference of the object to prevent it from GC(`Sync.Pool` clears all objects if it is unused for two GC cycles).

dot/network/service.go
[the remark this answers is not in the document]
noot: removed in favour of `sizedBufferPool`

dot/network/service.go
noot: removed in favour of `sizedBufferPool`

dot/network/config.go
```diff
@@ -97,6 +97,8 @@ type Config struct {
 
 	// telemetryInterval how often to send telemetry metrics
 	telemetryInterval time.Duration
+
+	noPreAllocate bool // internal option
```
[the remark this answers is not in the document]
noot: we want to pre-allocate by default so if this is changed to `preAllocate` then it will be false by default, but we won't know if it was set to false or if it defaulted to false, hence why it's `noPreAllocate`

arijitAD (approved):

dot/network/service.go
```diff
@@ -130,6 +133,18 @@ func NewService(cfg *Config) (*Service, error) {
 		return nil, err
 	}
 
+	// pre-allocate pool of buffers used to read from streams.
+	// initially allocate as many buffers as minimally necessary which is the number inbound streams we will have,
+	// which should equal min peers times the number of notifications protocols, which is currently 3.
+	var bufPool *sizedBufferPool
+	if !cfg.noPreAllocate {
```
arijitAD, line 140: If you switch `if` with `else` it will be cleaner.

if cfg.noPreAllocate {
    bufPool = newSizedBufferPool(cfg.MinPeers*3, cfg.MaxPeers*3)
} else {
    bufPool = &sizedBufferPool{
	c: make(chan *[maxMessageSize]byte, cfg.MaxPeers*3),
    }
}

dot/network/pool.go
```diff
@@ -0,0 +1,57 @@
...
+	default:
+		// create new buffer
+		buff = &[maxMessageSize]byte{}
+	}
+	return *buff
+}
+
+// put returns the given buffer to the sizedBufferPool.
+func (bp *sizedBufferPool) put(b *[maxMessageSize]byte) {
+	select {
```
EclesioMeloJunior, line 53: should we clear the buffer when it comes back to the pool?

noot: no it doesn't make a difference since the buffer contents will be overwritten when we read into it

EclesioMeloJunior (approved):

Issue #1494, opened 2021-03-29
Node Optimizations


dutterbutter:
To improve the overall performance of Gossamer nodes by examining and refactoring different components. Below outlines loose target benchmarks the team is looking to achieve:

|      **Metrics**      	| **Node Type**    	| **Parameters**           	| **Target**   	| **Current** 	| **Delta** 	|
|:-----------------:	|--------------	|----------------------	|----------	|---------	|-------	|
| Memory Usage      	|              	|                      	|          	|         	|       	|
| CPU               	|              	|                      	|          	|         	|       	|
| Storage size      	| Archive Node 	| At block height 1.6M 	| 15-20 GB 	|         	|       	|
| Blocks Per Second 	|              	|                      	|          	| 15-20   	|       	|
| Sync Speed        	|              	|                      	|          	|         	|       	|

dutterbutter:
@timwu20 please review and update.

PR #1608, opened 2021-05-28
chore(dot/network): improve CPU usage again; no more mallocs

noot:
## Changes

- update `readLEB128ToUint64` to read into existing buffer 
- reduce `maxMessageSize` to 63kb since messages weren't actually using up the size allocated before (1mb)
- move connect to peers if below minimum case to its own goroutine, only find peers if below min

## Tests

```
go test ./dot/network -short
```

## Issues

- related to #1494

### DIFF
--- a/dot/network/discovery.go
+++ b/dot/network/discovery.go
@@ -24,7 +24,7 @@ import (
 	badger "github.com/ipfs/go-ds-badger2"
 	libp2phost "github.com/libp2p/go-libp2p-core/host"
 	"github.com/libp2p/go-libp2p-core/peer"
-	"github.com/libp2p/go-libp2p-core/peerstore"
+	//"github.com/libp2p/go-libp2p-core/peerstore"
 	"github.com/libp2p/go-libp2p-core/protocol"
 	libp2pdiscovery "github.com/libp2p/go-libp2p-discovery"
 	kaddht "github.com/libp2p/go-libp2p-kad-dht"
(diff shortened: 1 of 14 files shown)


EclesioMeloJunior (approved):

dot/network/discovery.go
```diff
@@ -35,13 +35,15 @@ var (
 	startDHTTimeout             = time.Second * 10
 	initialAdvertisementTimeout = time.Millisecond
 	tryAdvertiseTimeout         = time.Second * 30
-	connectToPeersTimeout       = time.Minute
+	connectToPeersTimeout       = time.Minute * 5
+	findPeersTimeout            = time.Minute
```
timwu20, line 39: nit: I prefer to put these configurable settings inside the struct, so `discovery` in this case.

dot/network/service.go
```diff
@@ -139,10 +139,10 @@ func NewService(cfg *Config) (*Service, error) {
 	var bufPool *sizedBufferPool
 	if cfg.noPreAllocate {
 		bufPool = &sizedBufferPool{
-			c: make(chan *[maxMessageSize]byte, cfg.MaxPeers*3),
+			c: make(chan *[maxMessageSize]byte, cfg.MinPeers*3),
```
timwu20, line 142: question: why *3?

timwu20 (approved):

dot/network/service.go
noot: there's currently 3 notifications protocols supported, so the number of open streams  is likely to be peercount * 3 (I think the comment explains this)

dot/network/discovery.go
noot: would you put them inside the struct then assign the value inside `newDiscovery`?

timwu20: Yea then you can set your desired timeout in the tests per `*discovery` rather than globally through the package.

Issue #1426, opened 2021-03-01
fix and re-enable handleChainReorg by filtering inherents


noot:
## Task summary

- `core.Service.handleChainReorg` needs to be fixed (currently disabled)
- on a chain re-org it attempts to re-add every extrinsic from every block on the re-orged chain, however it does not filter inherents, causing a lot of `failed to call validate_transaction` errors
- need to filter extrinsics, ie make sure the ones we are re-executing are signed extrinsics

PR #1609, opened 2021-05-28
fix(dot/core): Add only extrinsic during chain reorg.

arijitAD:
## Changes

- Add only extrinsic during chain reorg.

## Tests

```
TestHandleChainReorg_WithReorg_Trans
```

## Issues

- closes: #1426

### DIFF
--- a/dot/core/service.go
+++ b/dot/core/service.go
@@ -356,12 +356,12 @@ func (s *Service) handleChainReorg(prev, curr common.Hash) error {
 				continue
 			}
 
-			encode, err := scale.Encode(ext)
+			encExt, err := scale.Encode(ext)
 			if err != nil {
 				return err
 			}
 
-			txv, err := s.rt.ValidateTransaction(append([]byte{byte(types.TxnExternal)}, encode...))
+			txv, err := s.rt.ValidateTransaction(append([]byte{byte(types.TxnExternal)}, encExt...))
 			if err != nil {
 				logger.Debug("failed to validate transaction", "error", err, "extrinsic", ext)
 				continue
--- a/dot/core/service.go
+++ b/dot/core/service.go
@@ -367,7 +367,7 @@ func (s *Service) handleChainReorg(prev, curr common.Hash) error {
 				continue
 			}
 
-			vtx := transaction.NewValidTransaction(ext, txv)
+			vtx := transaction.NewValidTransaction(encExt, txv)
 			s.transactionState.AddToPool(vtx)
 		}
 	}
--- a/dot/core/service.go
+++ b/dot/core/service.go
@@ -26,9 +26,9 @@ import (
 	"github.com/ChainSafe/gossamer/lib/crypto"
 	"github.com/ChainSafe/gossamer/lib/keystore"
 	"github.com/ChainSafe/gossamer/lib/runtime"
+	"github.com/ChainSafe/gossamer/lib/scale"
 	"github.com/ChainSafe/gossamer/lib/services"
 	"github.com/ChainSafe/gossamer/lib/transaction"
-
 	log "github.com/ChainSafe/log15"
 )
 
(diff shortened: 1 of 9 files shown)


dot/sync/test_helpers.go
```diff
@@ -51,3 +70,201 @@ func (bp *mockBlockProducer) Resume() error {
...
+	genTrie, err := genesis.NewTrieFromGenesis(gen)
+	require.NoError(t, err)
+
+	genesisHeader, err := types.NewHeader(common.NewHash([]byte{0}), genTrie.MustHash(), trie.EmptyHash, big.NewInt(0), types.Digest{})
+	require.NoError(t, err)
+	return gen, genTrie, genesisHeader
+}
+
+// BuildBlock ...
+func BuildBlock(t *testing.T, srv *Service, parent *types.Header, ext types.Extrinsic) *types.Block {
```
noot, line 152: I don't think this needs to be in the `sync` package as it doesn't rely on anything in it, it would be better placed in `lib/utils` or something potentially. also now that `lib/babe` has a `BlockBuilder` type you could instantiate that to build a block instead

dot/sync/test_helpers.go
```diff
@@ -51,3 +70,201 @@ func (bp *mockBlockProducer) Resume() error {
...
+	res.Number = header.Number
+
+	return &types.Block{
+		Header: res,
+		Body:   body,
+	}
+}
+
+// CreateExtrinsic ...
+func CreateExtrinsic(t *testing.T, srv *Service) []byte {
```
noot, line 224: same here, this doesn't use `sync.Service,` only a runtime

noot (commented):
looks good, please remove usage of `sync` package from `core` since I don't think it's necessary

dot/sync/test_helpers.go
```diff
@@ -51,3 +70,201 @@ func (bp *mockBlockProducer) Resume() error {
...
+
+	bob, err := subtypes.NewAddressFromHexAccountID("0x90b5ab205c6974c9ea841be688864633dc9ca8a357843eeacf2314649965fe22")
+	require.NoError(t, err)
+
+	call, err := subtypes.NewCall(metaData, "Balances.transfer", bob, subtypes.NewUCompactFromUInt(123450000000000))
+	require.NoError(t, err)
+
+	// Create the extrinsic
+	ext := subtypes.NewExtrinsic(call)
+	genesisHash, err := subtypes.NewHashFromHexString("0x64597c55a052d484d9ff357266be326f62573bb4fbdbb3cd49f219396fcebf78")
```
EclesioMeloJunior, line 246: just curious about where this value comes from?

arijitAD: This is genesis hash. Updated the code to take it from blockstore.

dot/sync/test_helpers.go
arijitAD: Done

dot/sync/test_helpers.go
arijitAD: `BuildBlock ` is used in a couple of tests in `syncer_test.go`. I tried moving it on `lib/utils` but I am getting cyclic dependency.

dot/core/service.go
```diff
@@ -340,13 +345,29 @@ func (s *Service) handleChainReorg(prev, curr common.Hash) error {
...
+			if !decExt.IsSigned() {
+				continue
+			}
+
+			encExt, err := scale.Encode(ext)
+			if err != nil {
+				return err
+			}
+
+			txv, err := s.rt.ValidateTransaction(append([]byte{byte(types.TxnExternal)}, encExt...))
```
timwu20, line 364: nit: move `append([]byte{byte(types.TxnExternal)}, encExt...)` into it's own line/var

timwu20 (approved):

dot/core/service.go
arijitAD: Done

edwardmack (approved):

PR #1611, opened 2021-05-31
feat(dot/network): implement streamManager to cleanup not recently used streams

noot:
## Changes

- implement `streamManager` to cleanup not recently used streams, when a new inbound stream is opened the stream is logged with the `streamManager`. if data is received on the stream, then the time is logged with the `streamManager`. tthe `streamManager` runs a cleanup goroutine every minute to cleanup streams that we haven't received any data on in  that amount of time

## Tests

```
go test ./dot/network -short
```

## Issues

- related to #1494

### DIFF
--- a/dot/network/service.go
+++ b/dot/network/service.go
@@ -313,9 +313,15 @@ main:
 
 		case <-ticker.C:
 			o := s.host.bwc.GetBandwidthTotals()
-			telemetry.GetInstance().SendNetworkData(telemetry.NewNetworkData(s.host.peerCount(), o.RateIn, o.RateOut))
+			err := telemetry.GetInstance().SendMessage(telemetry.NewTelemetryMessage(
+				telemetry.NewKeyValue("bandwidth_download", o.RateIn),
+				telemetry.NewKeyValue("bandwidth_upload", o.RateOut),
+				telemetry.NewKeyValue("msg", "system.interval"),
+				telemetry.NewKeyValue("peers", s.host.peerCount())))
+			if err != nil {
+				logger.Debug("problem sending system.interval telemetry message", "error", err)
+			}
 		}
-
 	}
 }
 
@@ -330,14 +336,17 @@ func (s *Service) sentBlockIntervalTelemetry() {
 			continue
 		}
 
-		telemetry.GetInstance().SendBlockIntervalData(&telemetry.BlockIntervalData{
-			BestHash:           best.Hash(),
-			BestHeight:         best.Number,
-			FinalizedHash:      finalized.Hash(),
-			FinalizedHeight:    finalized.Number,
-			TXCount:            0, // todo (ed) determine where to get tx count
-			UsedStateCacheSize: 0, // todo (ed) determine where to get used_state_cache_size
-		})
+		err = telemetry.GetInstance().SendMessage(telemetry.NewTelemetryMessage(
+			telemetry.NewKeyValue("best", best.Hash().String()),
+			telemetry.NewKeyValue("finalized_hash", finalized.Hash().String()), //nolint
+			telemetry.NewKeyValue("finalized_height", finalized.Number),        //nolint
+			telemetry.NewKeyValue("height", best.Number),
+			telemetry.NewKeyValue("msg", "system.interval"),
+			telemetry.NewKeyValue("txcount", 0),                // todo (ed) determine where to get tx count
+			telemetry.NewKeyValue("used_state_cache_size", 0))) // todo (ed) determine where to get used_state_cache_size
+		if err != nil {
+			logger.Debug("problem sending system.interval telemetry message", "error", err)
+		}
 		time.Sleep(s.telemetryInterval)
 	}
 }
--- a/dot/network/service.go
+++ b/dot/network/service.go
@@ -45,7 +45,7 @@ const (
 	blockAnnounceID = "/block-announces/1"
 	transactionsID  = "/transactions/1"
 
-	maxMessageSize = 1024 * 1024 // 1mb for now
+	maxMessageSize = 1024 * 63 // 63kb for now
 )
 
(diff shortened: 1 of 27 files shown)


dot/network/stream_manager.go
```diff
@@ -0,0 +1,68 @@
...
+)
+
+var cleanupStreamInterval = time.Minute
+
+// streamManager tracks inbound streams and runs a cleanup goroutine every `cleanupStreamInterval` to close streams that
+// we haven't received any data on for the last time period. this prevents keeping stale streams open and continuously trying to
+// read from it, which takes up lots of CPU over time.
+type streamManager struct {
+	ctx                 context.Context
+	lastReceivedMessage *sync.Map //map[string]time.Time
```
arijitAD, line 18: Why are we using two maps?
We can have a single map with the following value
```
type val struct {
  s network.Stream
  t time.time.
}
```

noot: updated!

dot/network/stream_manager.go
```diff
@@ -0,0 +1,78 @@
...
+	lastReceivedMessage time.Time
+	stream              network.Stream
+}
+
+// streamManager tracks inbound streams and runs a cleanup goroutine every `cleanupStreamInterval` to close streams that
+// we haven't received any data on for the last time period. this prevents keeping stale streams open and continuously trying to
+// read from it, which takes up lots of CPU over time.
+type streamManager struct {
+	ctx        context.Context
+	streamData *sync.Map //map[string]streamData
```
arijitAD, line 23: Rename `streamData` to `streamMap` since we already have `streamData`.

dot/network/stream_manager.go
```diff
@@ -0,0 +1,78 @@
...
+			_ = stream.Close()
+			sm.streamData.Delete(id)
+		}
+
+		return true
+	})
+}
+
+func (sm *streamManager) logNewStream(stream network.Stream) {
+	data := streamData{
```
arijitAD, line 62: Change this to pointer.
`data := &streamData{}`

arijitAD (approved):

dot/network/stream_manager.go
noot: just wondering why?

dot/network/service.go
```diff
@@ -267,6 +269,7 @@ func (s *Service) Start() error {
 	go s.logPeerCount()
 	go s.publishNetworkTelemetry(s.closeCh)
 	go s.sentBlockIntervalTelemetry()
+	s.streamManager.start()
```
timwu20, line 272: any reason why this spawns it's own goroutine instead of calling it via `go s.streamManager.start()`

timwu20: I see that there's a `ctx` created on `NewService`.  Do we need both the `closeCh` and `ctx`?

dot/network/stream_manager.go
```diff
@@ -0,0 +1,78 @@
...
+	lastReceivedMessage time.Time
+	stream              network.Stream
+}
+
+// streamManager tracks inbound streams and runs a cleanup goroutine every `cleanupStreamInterval` to close streams that
+// we haven't received any data on for the last time period. this prevents keeping stale streams open and continuously trying to
+// read from it, which takes up lots of CPU over time.
+type streamManager struct {
+	ctx           context.Context
+	streamDataMap *sync.Map //map[string]*streamData
```
timwu20, line 23: never used a `sync.Map` before.  Are there multiple goroutines trying to read this map concurrently? Potentially you may get better performance from using a standard map and `sync.RWMutex`.

dot/network/stream_manager.go
```diff
@@ -0,0 +1,78 @@
...
+func (sm *streamManager) start() {
+	go func() {
+		for {
+			select {
+			case <-sm.ctx.Done():
+				return
+			case <-time.After(cleanupStreamInterval):
+				sm.cleanupStreams()
+			}
+		}
```
timwu20, line 42:
```suggestion
                 ticker := time.NewTicker(cleanupStreamInterval)
		for {		       
			select {
			case <-sm.ctx.Done():
				return
			case <-ticker.C:
				sm.cleanupStreams()
			}
		}
```

dot/network/service.go
noot: no particular reason, I like it like this since the main network service doesn't need to be concerned with what the streamManager does

dot/network/stream_manager.go
noot: yep there are multiple goroutines that'll access it, I think `sync.Map` is optimized for this us case so would be preferred over a standard map and a mutex

noot: https://golang.org/pkg/sync/#Map this case would be use case 2

dot/network/stream_manager.go
noot: very cool, can definitely use this in other places too

dot/network/stream_manager.go
```diff
@@ -0,0 +1,79 @@
...
+func newStreamManager(ctx context.Context) *streamManager {
+	return &streamManager{
+		ctx:           ctx,
+		streamDataMap: new(sync.Map),
+	}
+}
+
+func (sm *streamManager) start() {
+	go func() {
+        ticker := time.NewTicker(cleanupStreamInterval)
```
EclesioMeloJunior, line 35:
```suggestion
        ticker := time.NewTicker(cleanupStreamInterval)
        defer ticker.Stop()
```

EclesioMeloJunior (changes requested):

dot/network/stream_manager.go
noot: added!

EclesioMeloJunior (approved):

Issue #1587, opened 2021-05-13
devnet: ensure blocks are finalized 


noot:
## Task summary

- it appears blocks are not getting finalized on the devnet, this may be related to #1586 (network isn't properly formed)
- either way, need to ensure blocks are finalized within a reasonable timeframe on the devnet, should be 1 round every 2 seconds

PR #1613, opened 2021-05-31
fix(dot/state, lib/babe, lib/trie): improve syncing between gossamer authority nodes

noot:
## Changes

- pause block production when syncing, fix locking mechanism during pause/resume
- update `trie.Snapshot` to return new trie, previously it was returning old trie, so both syncer and babe were accessing the same trie (as the receiver of `Snapshot` was returned by `TrieState` in both cases)
- fix `HasFinalizedBlock` logic and db key

## Tests

```
go test ./lib/trie
go test ./lib/babe
go test ./dot/state
```

## Issues

- related to #1587


timwu20 (approved):

lib/runtime/storage/trie.go
```diff
@@ -61,7 +61,8 @@ func (s *TrieState) Snapshot() *trie.Trie {
 func (s *TrieState) BeginStorageTransaction() {
 	s.lock.Lock()
 	defer s.lock.Unlock()
-	s.oldTrie = s.t.Snapshot()
+	s.oldTrie = s.t
```
EclesioMeloJunior, line 64: `s.oldTrie`  points to the same ref as `s.t`, if I'm not wrong, any modifications in `s.t` will modify the `s.oldTrie`

edwardmack (approved):
Nice work, lgtm.

lib/runtime/storage/trie.go
noot: it does, but in the following line `s.t` gets snapshotted, so `s.oldTrie` won't be modified if `s.t` is modified when this function completes. `s.oldTrie` is effectively a pointer to the root of the previous trie, and `s.t` becomes a pointer to the root of the working trie

EclesioMeloJunior: Got it! just `s.t` changes the ref, `s.oldTrie` keeps pointing to the old, nice! Why not just keep `s.oldTrie = s.t.Snapshot` as before?

noot: `s.t.Snapshot()` now returns the new trie to be modified, previously it returned the old trie to be saved and `s.t` would continue to be modified

EclesioMeloJunior (approved):

PR #1615, opened 2021-06-01
fix(dot/network): implement a handshake timeout

### LINKED ISSUE #1550: Add timeout to notification message handshake
## Issue summary

- Currently, any peer can initiate a handshake and leave the stream open forever without responding to a handshake.
- Also, clear handshake data after a fixed timeout if the peer has responded with an invalid handshake.

- https://github.com/ChainSafe/gossamer/blob/028d25e72e5a3c32380f929d0db11ac4e874af37/dot/network/notifications.go#L124

EclesioMeloJunior:
## Changes
- the function `readHandshake` has a goroutine to read the stream bytes
- Setup the timeout inside the `sendData` function and use a `select` to wait for a timeout or a response from the other peer

## Tests

Details on how to run tests relevant to the changes within this pull request.

Created 2 hosts, one will send the handshake and the other will receive but never responds, the tests will validate:
1. If the host that sends the request stores an outbound handshake data until the timeout is reached.
2. If the stream is open before the timeout is reached.
3. After the timeout the stream is closed
4. After the timeout the outbound handshake data is deleted

```
go test -test.v -v -timeout 30s -run ^Test_HandshakeTimeout$ github.com/ChainSafe/gossamer/dot/network
```

## Issues

- closes #1550

### DIFF
--- a/dot/network/notifications.go
+++ b/dot/network/notifications.go
@@ -235,17 +235,17 @@ func (s *Service) sendData(peer peer.ID, hs Handshake, info *notificationsProtoc
 		}
 
 		hsTimer := time.NewTimer(handshakeTimeout)
-		defer hsTimer.Stop()
 
 		var hs Handshake
 		select {
 		case <-hsTimer.C:
-			logger.Warn("handshake timeout reached", "protocol", info.protocolID, "peer", peer)
+			logger.Trace("handshake timeout reached", "protocol", info.protocolID, "peer", peer)
 			_ = stream.Close()
 			info.outboundHandshakeData.Delete(peer)
 			return
 
 		case hsResponse := <-s.readHandshake(stream, decodeBlockAnnounceHandshake):
+			hsTimer.Stop()
 			if hsResponse.err != nil {
 				logger.Trace("failed to read handshake", "protocol", info.protocolID, "peer", peer, "error", err)
 				_ = stream.Close()
--- a/dot/network/notifications.go
+++ b/dot/network/notifications.go
@@ -17,7 +17,6 @@
 package network
 
 import (
-	"context"
 	"errors"
 	"sync"
 	"time"
@@ -136,7 +135,7 @@ func (s *Service) createNotificationsMessageHandler(info *notificationsProtocol,
 		}
 
 		if msg.IsHandshake() {
-			logger.Debug("received handshake on notifications sub-protocol", "protocol", info.protocolID,
+			logger.Trace("received handshake on notifications sub-protocol", "protocol", info.protocolID,
 				"message", msg,
 				"peer", stream.Conn().RemotePeer(),
 			)
@@ -151,14 +150,14 @@ func (s *Service) createNotificationsMessageHandler(info *notificationsProtocol,
 			// ie it is an inbound stream and we only send the handshake over it.
 			// we do not send any other data over this stream, we would need to open a new outbound stream.
 			if _, has := info.getHandshakeData(peer, true); !has {
-				logger.Debug("receiver: validating handshake", "protocol", info.protocolID)
+				logger.Trace("receiver: validating handshake", "protocol", info.protocolID)
 
 				hsData := newHandshakeData(true, false, stream)
 				info.inboundHandshakeData.Store(peer, hsData)
 
 				err := info.handshakeValidator(peer, hs)
 				if err != nil {
-					logger.Debug("failed to validate handshake", "protocol", info.protocolID, "peer", peer, "error", err)
+					logger.Trace("failed to validate handshake", "protocol", info.protocolID, "peer", peer, "error", err)
 					return errCannotValidateHandshake
 				}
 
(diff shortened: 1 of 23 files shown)


dot/network/notifications.go
```diff
@@ -294,19 +318,27 @@ func (s *Service) broadcastExcluding(info *notificationsProtocol, excluding peer
...
-	msgBytes := s.bufPool.get()
-	defer s.bufPool.put(&msgBytes)
+func (s *Service) readHandshake(stream libp2pnetwork.Stream, decoder HandshakeDecoder) <-chan *handshakeReader {
+	hsC := make(chan *handshakeReader, 1)
 
-	tot, err := readStream(stream, msgBytes[:])
-	if err != nil {
-		return nil, err
-	}
+	go func() {
```
timwu20, line 324: should we close `hsC` on errors?

dot/network/notifications.go
```diff
@@ -127,7 +136,7 @@ func (s *Service) createNotificationsMessageHandler(info *notificationsProtocol,
 		}
 
 		if msg.IsHandshake() {
-			logger.Trace("received handshake on notifications sub-protocol", "protocol", info.protocolID,
+			logger.Debug("received handshake on notifications sub-protocol", "protocol", info.protocolID,
```
noot, line 139: change these logs back to trace please

dot/network/notifications.go
```diff
@@ -226,26 +235,41 @@ func (s *Service) sendData(peer peer.ID, hs Handshake, info *notificationsProtoc
...
 
-		hs, err := s.readHandshake(stream, decodeBlockAnnounceHandshake)
-		if err != nil {
-			logger.Trace("failed to read handshake", "protocol", info.protocolID, "peer", peer, "error", err)
+		var hs Handshake
+		hstimeout, cancel := context.WithTimeout(s.ctx, handshakeTimeout)
+		defer cancel()
+
+		select {
+		case <-hstimeout.Done():
```
noot, line 243: this might slow down the performance of the node, since creating a timeout requires `malloc`ing and doing this for every outbound stream is potentially too often. I'll profile it and let you know

EclesioMeloJunior: Maybe a Ticker could be better in terms of performance?

dot/network/notifications.go
```diff
@@ -226,14 +234,32 @@ func (s *Service) sendData(peer peer.ID, hs Handshake, info *notificationsProtoc
 			return
 		}
 
-		hs, err := s.readHandshake(stream, decodeBlockAnnounceHandshake)
-		if err != nil {
-			logger.Trace("failed to read handshake", "protocol", info.protocolID, "peer", peer, "error", err)
+		hsTicker := time.NewTicker(handshakeTimeout)
```
arijitAD, line 237: User timer instead of ticker.
```
time.NewTimer(time.Second)
```

Timers are for when you want to do something once in the future - tickers are for when you want to do something repeatedly at regular intervals

dot/network/notifications.go
```diff
@@ -294,19 +320,30 @@ func (s *Service) broadcastExcluding(info *notificationsProtocol, excluding peer
 	}
 }
 
-func (s *Service) readHandshake(stream libp2pnetwork.Stream, decoder HandshakeDecoder) (Handshake, error) {
-	msgBytes := s.bufPool.get()
-	defer s.bufPool.put(&msgBytes)
+func (s *Service) readHandshake(stream libp2pnetwork.Stream, decoder HandshakeDecoder) <-chan *handshakeReader {
+	hsC := make(chan *handshakeReader, 1)
```
arijitAD, line 324: Curious why is this buffered channel? The code will still work if it's an unbuffered channel.

EclesioMeloJunior: As the `readHandshake` function return the bytes of the handshake response at once I just fixed this in the channel capacity, but yes it works with an unbuffered channel (a simple example with an unbuffered channel: https://play.golang.org/p/Y19oZxihFxD)

dot/network/notifications.go
```diff
@@ -234,20 +234,20 @@ func (s *Service) sendData(peer peer.ID, hs Handshake, info *notificationsProtoc
 			return
 		}
 
-		hsTicker := time.NewTicker(handshakeTimeout)
+		hsTimer := time.NewTimer(handshakeTimeout)
 		var hs Handshake
```
arijitAD, line 238: You can also do 
```
defer hsTimer.Stop()
```

EclesioMeloJunior: Done!

arijitAD (approved):

dot/network/notifications.go
```diff
@@ -226,14 +234,29 @@ func (s *Service) sendData(peer peer.ID, hs Handshake, info *notificationsProtoc
...
-		hs, err := s.readHandshake(stream, decodeBlockAnnounceHandshake)
-		if err != nil {
-			logger.Trace("failed to read handshake", "protocol", info.protocolID, "peer", peer, "error", err)
+		hsTimer := time.NewTimer(handshakeTimeout)
+		defer hsTimer.Stop()
+
+		var hs Handshake
+		select {
+		case <-hsTimer.C:
+			logger.Warn("handshake timeout reached", "protocol", info.protocolID, "peer", peer)
```
noot, line 243: change to trace

dot/network/notifications.go
```diff
@@ -226,14 +234,29 @@ func (s *Service) sendData(peer peer.ID, hs Handshake, info *notificationsProtoc
 			return
 		}
 
-		hs, err := s.readHandshake(stream, decodeBlockAnnounceHandshake)
-		if err != nil {
-			logger.Trace("failed to read handshake", "protocol", info.protocolID, "peer", peer, "error", err)
+		hsTimer := time.NewTimer(handshakeTimeout)
+		defer hsTimer.Stop()
```
noot, line 238: just wondering why this was chosen vs `time.After`? I guess since you can stop this if the response arrives first?

noot (commented):
profiling looks fine to me - please update logs, otherwise looks good!

dot/network/notifications.go
EclesioMeloJunior: 1) With `timer` we can cancel the timer if the response arrives first.
2) I noticed that we need to stop the timer only when the response arrives first, if the `<-timer.C` comes first then we don't need to use stop since we are not resetting the timer.

timwu20 (approved):

PR #1617, opened 2021-06-01
fix(dot/network): check if peer supports protocol

### LINKED ISSUE #1571: dot/network: check if peer supports protocol before sending message
## Task summary

- currently we don't check if a peer supports a given protocol before attempting to open a stream with them/send a message over that stream
- this results in an error
- we should check before doing this that the peer supports the given protocol

EclesioMeloJunior:
## Changes

- Encapsulate `peerstore.SupportsProtocols` function
- Check if the peer supports protocol before handshake on `sendData` function

## Tests

```
go run go test ^Test_PeerSupportsProtocol$ github.com/ChainSafe/gossamer/dot/network
```

## Issues

- closes #1571

### DIFF
--- a/dot/network/host.go
+++ b/dot/network/host.go
@@ -362,6 +362,17 @@ func (h *host) peers() []peer.ID {
 	return h.h.Network().Peers()
 }
 
+// getPeerProtocols returns all the supported protocols to one peer
+// returns an error if could not get peer protocols
+func (h *host) supportsProtocol(peerID peer.ID, protocol protocol.ID) (bool, error) {
+	peerProtocols, err := h.h.Network().Peerstore().SupportsProtocols(peerID, string(protocol))
+	if err != nil {
+		return false, err
+	}
+
+	return len(peerProtocols) > 0, nil
+}
+
 // peerCount returns the number of connected peers
 func (h *host) peerCount() int {
 	peers := h.h.Network().Peers()
--- a/dot/network/host.go
+++ b/dot/network/host.go
@@ -365,7 +365,7 @@ func (h *host) peers() []peer.ID {
 // supportsProtocol checks if the protocol is supported by peerID
 // returns an error if could not get peer protocols
 func (h *host) supportsProtocol(peerID peer.ID, protocol protocol.ID) (bool, error) {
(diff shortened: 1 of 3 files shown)


dot/network/notifications.go
```diff
@@ -198,6 +198,11 @@ func (s *Service) createNotificationsMessageHandler(info *notificationsProtocol,
 }
 
 func (s *Service) sendData(peer peer.ID, hs Handshake, info *notificationsProtocol, msg NotificationsMessage) {
+	if support, err := s.host.supportsProtocol(peer, info.protocolID); err != nil || !support {
+		logger.Warn("the peer does not supports the protocol", "protocol", info.protocolID, "peer", peer, "err", err)
```
arijitAD, line 202: This should be set at `debug` level as we may get this error often.

dot/network/host.go
```diff
@@ -362,6 +362,17 @@ func (h *host) peers() []peer.ID {
 	return h.h.Network().Peers()
 }
 
+// getPeerProtocols returns all the supported protocols to one peer
```
arijitAD, line 365: `getPeerProtocols` -> `supportsProtocol`

`supportsProtocol` doesn't return all the supported protocols.
`supportsProtocol` checks if `protocol` is supported by `peerID`.

arijitAD (approved):

dot/network/notifications.go
EclesioMeloJunior: Done!

dot/network/host.go
EclesioMeloJunior: Done!

dot/network/host.go
```diff
@@ -362,6 +362,17 @@ func (h *host) peers() []peer.ID {
 	return h.h.Network().Peers()
 }
 
+// supportsProtocol checks if the protocol is supported by peerID
+// returns an error if could not get peer protocols
+func (h *host) supportsProtocol(peerID peer.ID, protocol protocol.ID) (bool, error) {
+	peerProtocols, err := h.h.Network().Peerstore().SupportsProtocols(peerID, string(protocol))
```
noot, line 368: I think you can directly do `h.h.Peerstore().SupportsProtocols()`

noot (approved):
looks good!

timwu20 (approved):

PR #1618, opened 2021-06-01
feat(dot/telemetry): implement telemetry message network_state

edwardmack:
## Changes

- This implements telemetry message network_state.
- Part of network_state message contains `connectedPeers` field that contains: 
example data from polkadot node:
```
12D3KooWH4LX5TFftR1botGybbMQ8jWgUt5z1nToaxGsUdyiHARC":{
	"endpoint":{
		"dialing":"/ip4/135.181.214.98/tcp/30333"},
	"versionString":"Parity Polkadot/v0.9.3-aa3867609-x86_64-linux-gnu (SUBSTAR_io0)",
	"latestPingTime":{
		"secs":0,"nanos":335986866},
	"enabled":true,
	"open":true,
	"knownAddresses":["/ip4/135.181.214.98/tcp/30333"]},

```
however I don't know where to obtain this data (endpoint, versionString, latestPingTime, enabled, open, and knownAddresses

- Currently I'm reading data from s.host.peers (with Peers() method), which seems to give data from BlockAnnounceMsgType, is there a message type for getting peer info?

## Tests

```

```

## Issues

- related to #1501


dot/network/service.go
```diff
@@ -321,6 +327,35 @@ main:
...
+			for _, v := range s.Peers() {
+				p := &peerInfo{
+					Roles:      v.Roles,
+					BestHash:   v.BestHash.String(),
+					BestNumber: v.BestNumber,
+				}
+				peers[v.PeerID] = *p
+			}
+			netState["connectedPeers"] = peers
+			err = telemetry.GetInstance().SendMessage(telemetry.NewTelemetryMessage(
```
timwu20, line 353: After looking at the `msgToBytes` method it seems all that does is `json.Marshal()` the `map[string]interface{}`.  So there's no real need to construct these messages as `map[string]interface{}`.  You could essentially just pass in structs.  So maybe in this case: 

```
type netState struct {
    PeerID string `json:"peerId"`
    ExternalAddresses []string `json:"externalAddresses"`
    ListenedAddresses []string `json:"listenedAddresses"`  
}
```
You could create a constructor for this type as well to handle the population of the fields.

dot/telemetry/telemetry.go
```diff
@@ -124,31 +97,94 @@ func (h *Handler) startListening() {
 		go func() {
 			for _, conn := range h.connections {
 				conn.Lock()
-				err := conn.wsconn.WriteMessage(websocket.TextMessage, msgToBytes(msg))
+				defer conn.Unlock()
+				msgBytes, err := h.msgToJSON(msg)
+				if err != nil || len(msgBytes) == 0 {
+					h.log.Debug("issue decoding telemetry message", "error", err)
+					return
+				}
```
noot, line 105: move this out of the for loop

dot/network/service.go
```diff
@@ -334,19 +371,23 @@ func (s *Service) sentBlockIntervalTelemetry() {
...
 			continue
 		}
+		finalizedHash := finalized.Hash()
+
+		err = telemetry.GetInstance().SendMessage(telemetry.SystemIntervalTM{
+			BestHash:           &bestHash,
+			BestHeight:         best.Number,
+			FinalisedHash:      &finalizedHash,
+			FinalisedHeight:    finalized.Number,
+			TxCount:            big.NewInt(0), // todo (ed) determine where to get tx count
```
noot, line 387: is this regarding the tx pool? if so you can get the size via the `TransactionState`

dot/network/service.go
```diff
@@ -304,6 +305,12 @@ func (s *Service) logPeerCount() {
 	}
 }
 
+type peerInfo struct {
+	Roles      byte   `json:"roles"`
+	BestHash   string `json:"bestHash"`
+	BestNumber uint64 `json:"bestNumber"`
+}
+
 func (s *Service) publishNetworkTelemetry(done chan interface{}) {
```
noot, line 314: wondering why this uses ` done chan` instead of the service `ctx`?

dot/network/service.go
```diff
@@ -316,11 +323,41 @@ main:
...
+			for _, v := range s.host.h.Addrs() {
+				hostAddrs = append(hostAddrs, v.String())
+			}
+			netState["externalAddressess"] = hostAddrs
+			listAddrs := []string{}
+			for _, v := range s.host.h.Network().ListenAddresses() {
+				listAddrs = append(listAddrs, v.String())
+			}
+			netState["listenedAddressess"] = listAddrs
+
```
noot, line 346: is this expecting the multiaddress including peer ID? if so you will need to use something likenthis to create the listen address strings:
```
		externalAddr, err = ma.NewMultiaddr(fmt.Sprintf("/ip4/%s/tcp/%d", ip, cfg.Port))
		if err != nil {
			return nil, err
		}
```

dot/telemetry/telemetry.go
edwardmack: good idea, no need to keep rebuilding the message.

dot/telemetry/telemetry.go
```diff
@@ -108,9 +81,9 @@ func (h *Handler) AddConnections(conns []*genesis.TelemetryEndpoint) {
 }
 
 // SendMessage sends Message to connected telemetry listeners
-func (h *Handler) SendMessage(msg *Message) error {
+func (h *Handler) SendMessage(msg interface{}) error {
```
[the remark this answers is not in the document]
edwardmack: Yes, created TelemetryMessage interface.

dot/network/service.go
edwardmack: Ok, I've added interface to get Transactions Count.

dot/network/service.go
edwardmack: I was following suggestions from this thread: https://github.com/ChainSafe/gossamer/pull/1528#discussion_r617684192

dot/network/service.go
edwardmack: Updated.

noot (approved):

dot/network/service.go
```diff
@@ -316,11 +324,36 @@ main:
...
+			peers := make(map[string]interface{})
+			for _, v := range s.Peers() {
+				p := &peerInfo{
+					Roles:      v.Roles,
+					BestHash:   v.BestHash.String(),
+					BestNumber: v.BestNumber,
+				}
+				peers[v.PeerID] = *p
+			}
+			netState["connectedPeers"] = peers
```
timwu20, line 354: Can this code be moved into `NewNetworkStateTM`?

dot/telemetry/telemetry.go
```diff
@@ -122,33 +94,161 @@ func (h *Handler) startListening() {
 	for {
 		msg := <-h.msg
 		go func() {
+			msgBytes, err := h.msgToJSON(msg)
+			if err != nil || len(msgBytes) == 0 {
```
timwu20, line 98: nit: I would assume `msgToJSON` would error out if `msgBytes` is 0 length.

timwu20 (changes requested):

dot/network/service.go
edwardmack: Yes, it would make more sense there, moved.

dot/telemetry/telemetry.go
edwardmack: good point. removed.

dot/telemetry/telemetry.go
```diff
@@ -236,17 +238,47 @@ func (tm *SystemIntervalTM) messageType() string {
...
+
 // NetworkStateTM struct to hold network state telemetry messages
 type NetworkStateTM struct {
 	Msg   string                 `json:"msg"`
 	State map[string]interface{} `json:"state"`
 }
 
 // NewNetworkStateTM function to create new Network State Telemetry Message
-func NewNetworkStateTM(state map[string]interface{}) *NetworkStateTM {
+func NewNetworkStateTM(host libp2phost.Host, peerInfos []common.PeerInfo) *NetworkStateTM {
```
timwu20, line 254:
```suggestion
func NewNetworkStateTM(host libp2phost.Host, peerInfo []common.PeerInfo) *NetworkStateTM {
```

timwu20 (approved):

EclesioMeloJunior (approved):

PR #1620, opened 2021-06-02
feat(dot/sync): implement CodeSubstitutes

### LINKED ISSUE #1619: implement `codeSubstitutes`
## Task summary

- this is a map in the genesis file of block hash -> wasm code: https://raw.githubusercontent.com/paritytech/polkadot/072a8d2b8861414f306ade7013d9b6d31fd8e34f/node/service/res/polkadot.json
- when the block hash specified in the genesis is reached (imported), the wasm code should be replaced with the code specified in `codeSubstitutes`, until the next runtime update (determined via `spec_version`)
- see `syncer.Service.handleRuntimeChanges` for how current changes are checked, can use similar method to this

-  see https://github.com/paritytech/substrate/pull/8898

edwardmack:
## Changes

- Implement code to handle code substitutes defined in genesis.json.  wasm code is replaced with code contained in codeSubstitutes map when block is imported with same hash as key in codeSubstitutes map.
-
-

## Tests

TODO:
```

```

## Issues

- closes #1619

### DIFF
--- a/dot/services.go
+++ b/dot/services.go
@@ -387,6 +387,7 @@ func createSyncService(cfg *Config, st *state.Service, bp sync.BlockProducer, fg
 		Verifier:         verifier,
 		Runtime:          rt,
 		DigestHandler:    dh,
+		BaseState:        st.Base,
 	}
 
 	return sync.NewService(syncCfg)
--- a/dot/sync/syncer.go
+++ b/dot/sync/syncer.go
@@ -56,6 +56,9 @@ type Service struct {
 
 	// Consensus digest handling
 	digestHandler DigestHandler
+
+	// map of code substitutions keyed by block hash
+	codeSubstitute map[string]string
 }
 
 // Config is the configuration for the sync Service.
@@ -69,6 +72,7 @@ type Config struct {
 	Runtime          runtime.Instance
 	Verifier         Verifier
 	DigestHandler    DigestHandler
+	BaseState        BaseState
 }
 
 // NewService returns a new *sync.Service
@@ -102,6 +106,11 @@ func NewService(cfg *Config) (*Service, error) {
 		return nil, err
 	}
 
+	codeSubstitute, err := cfg.BaseState.LoadGenesisData()
+	if err != nil {
+		return nil, err
+	}
+
 	return &Service{
 		codeHash:         codeHash,
 		blockState:       cfg.BlockState,
@@ -114,6 +123,7 @@ func NewService(cfg *Config) (*Service, error) {
 		runtime:          cfg.Runtime,
 		verifier:         cfg.Verifier,
 		digestHandler:    cfg.DigestHandler,
+		codeSubstitute:   codeSubstitute.CodeSubstitutes,
 	}, nil
 }
 
@@ -379,6 +389,7 @@ func (s *Service) handleBlock(block *types.Block) error {
 		s.handleDigests(block.Header)
 	}
 
+	s.handleCodeSubstitution(block.Header.Hash())
 	return s.handleRuntimeChanges(ts)
 }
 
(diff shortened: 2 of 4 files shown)


edwardmack:
Note:  I still need to implement tests for this, I'm looking for advice on testing this without having to sync 5 million blocks.

dot/sync/syncer.go
```diff
@@ -56,6 +56,9 @@ type Service struct {
 
 	// Consensus digest handling
 	digestHandler DigestHandler
+
+	// map of code substitutions keyed by block hash
+	codeSubstitute map[string]string
```
noot, line 61: change to `map[common.Hash]string`?

dot/sync/syncer.go
```diff
@@ -434,6 +448,28 @@ func (s *Service) handleRuntimeChanges(newState *rtstorage.TrieState) error {
 	return nil
 }
 
+func (s *Service) handleCodeSubstitution(block common.Hash) error {
+	for key, value := range s.codeSubstitute {
```
noot, line 452: check if this is in map instead of iterating

dot/services.go
```diff
@@ -387,6 +387,7 @@ func createSyncService(cfg *Config, st *state.Service, bp sync.BlockProducer, fg
 		Verifier:         verifier,
 		Runtime:          rt,
 		DigestHandler:    dh,
+		BaseState:        st.Base,
```
noot, line 390: it might be nicer to pass in the code substitutes here instead of the BaseState, so the sync service doesn't need to deal with the conversions

noot (commented):
I can run this on the node to test this! in the meantime maybe a test similar to the existing `handleRuntimeChanges` test would be good

dot/sync/syncer.go
```diff
@@ -379,6 +389,10 @@ func (s *Service) handleBlock(block *types.Block) error {
 		s.handleDigests(block.Header)
 	}
 
+	e := s.handleCodeSubstitution(block.Header.Hash())
+	if e != nil {
+		return e
+	}
```
noot, line 395: can name `err`, just don't redeclare

noot (commented):
tested it on the chain and it worked :D great work!!!

dot/sync/syncer.go
edwardmack: Updated.

dot/sync/syncer.go
edwardmack: Updated.

dot/services.go
edwardmack: Updated.

dot/sync/syncer.go
edwardmack: updated.

dot/sync/syncer.go
```diff
@@ -434,6 +443,27 @@ func (s *Service) handleRuntimeChanges(newState *rtstorage.TrieState) error {
 	return nil
 }
 
+func (s *Service) handleCodeSubstitution(block common.Hash) error {
+	value := s.codeSubstitute[block]
+	if value != "" {
```
noot, line 448: nit but change to 
```
if value == "" {
    return 
}
// code currently in if case here
```

noot (commented):
looking good, passes `5203203` but fails at `5205487` with storage root mismatch - please add a check that if there's a code substitute currently, then when there's a runtime upgrade. only do the upgrade if the runtime `spec_version` changed using the `Version()`, I think that would likely fix the issue!

dot/sync/syncer.go
edwardmack: updated.

edwardmack:
I've updated this to include a check if code has been substituted, and compare runtime spec versions.  @noot I'm currently syncing polkadot, but I'm only on block 300,000, can you confirm this fixes the issue at 5205487?

noot:
@edwardmack it passes block `5205487` now!!! just syncing it for a bit longer until the next runtime update to ensure it's fully working, but looks good so far. please add logic to check if we are on a code substitute on restart so that the correct runtime gets loaded.

Issue #1606, opened 2021-05-25
update default runtime to v0.9


noot:
## Task summary

- update the `gssmr` and `dev` runtimes to v0.9 
- need to compile substrate v0.9 default node runtime, `dev` has `c = (1,1)` change

FlorianFranzen:
Also see current testsuite failure with updated runtime [here](https://github.com/w3f/polkadot-tests/runs/2518502437?check_suite_focus=true).

noot:
closed with #1625

PR #1625, opened 2021-06-07
chore: update gssmr genesis to v0.9; stub missing ext_ funcs

noot:
## Changes

- draft for now since this causes `ValidateTransaction` and `ApplyExtrinsic` to fail with a codec error

## Tests

```
go test ./lib/runtime/... -short
```

## Issues

- related to #1606


EclesioMeloJunior:
I'm noticed that in the test function `TestApplyExtrinsic` expects the return of `res, err := babeService.rt.ApplyExtrinsic(ext)` to be `[]byte{0x0, 0x0}` but is receiving `[]byte{0x1, 0x0, 0x6}`, and I found that this errors means:

> ExhaustsResources: Would exhaust the resource of the current block the transaction might be valid

am I missing something while building the extrinsic?

lib/babe/build_test.go
```diff
@@ -83,6 +85,45 @@ func addAuthorshipProof(t *testing.T, babeService *Service, slotNumber, epoch ui
 	babeService.slotToProof[slotNumber] = outAndProof
 }
 
+func createTestExtrinsic(t *testing.T, rt runtime.Instance, genHash common.Hash, nonce uint64) types.Extrinsic {
```
edwardmack, line 88: This func looks the same as func in dot/core/messages_test.go and in lib/runtime/wasmer/exports_test.go, I wonder if we can make one of them public and re-use it?

lib/runtime/wasmer/imports.go
```diff
@@ -1084,6 +1092,13 @@ func ext_default_child_storage_storage_kill_version_1(context unsafe.Pointer, ch
 	storage.DeleteChild(childStorageKey)
 }
 
+//export ext_default_child_storage_storage_kill_version_2
+func ext_default_child_storage_storage_kill_version_2(context unsafe.Pointer, a, b C.int64_t) C.int32_t {
+	logger.Debug("[ext_default_child_storage_storage_kill_version_2] executing...")
+	logger.Warn("[ext_default_child_storage_storage_kill_version_2] unimplemented")
+	return 0
```
edwardmack, line 1099: Do we need a TODO, or issue for this since it's unimplemented?

edwardmack (approved):
Great work, lgtm, just a couple questions.

lib/runtime/wasmer/imports.go
noot: I can open an issue for this

lib/babe/build_test.go
noot: I'd rather not make it public and import one package into the other but I'll move it to utils

noot: nevermind, that causes a circular import. I'll think about how to do this

EclesioMeloJunior (approved):
lgtm!

PR #1626, opened 2021-06-07
fix(dot,lib): Replacing handmade mocks to generated new ones

### LINKED ISSUE #1347: cleanup usage of mock interfaces in tests
## Task summary

- many packages use mock interfaces in tests; there duplicates of these mocks between packages that can be consolidated into a test-utils package
-  also we could use gomock https://github.com/golang/mock to simplify creation of these mock interfaces

EclesioMeloJunior:
## Changes

- Replace handmade mocks to generated new ones using mockery lib.

## Tests

```
go test -short ./dot/...
go test -short ./lib/...
```

## Issues

- Closes #1347


dot/core/digest.go
```diff
@@ -163,6 +163,7 @@ func (h *DigestHandler) HandleConsensusDigest(d *types.ConsensusDigest, header *
 		case types.NextEpochDataType:
 			return h.handleNextEpochData(d, header)
 		case types.BABEOnDisabledType:
+			logger.Info("on swit case BABEOnDisabledType", header)
```
noot, line 166: remove

EclesioMeloJunior: done!

dot/core/messages_test.go
```diff
@@ -50,28 +52,31 @@ func TestService_ProcessBlockAnnounceMessage(t *testing.T) {
...
+		ParentHash:     newBlock.Header.ParentHash,
+		Number:         newBlock.Header.Number,
+		StateRoot:      newBlock.Header.StateRoot,
+		ExtrinsicsRoot: newBlock.Header.ExtrinsicsRoot,
+		Digest:         newBlock.Header.Digest,
+		BestBlock:      true,
+	}
+
+	//setup the SendMessage function
+	net.On("SendMessage", expected)
```
noot, line 74: what does `.On()` do when there's no `.Return()`? does it just let the mocker know to expect this argument?

dot/network/discovery_test.go
```diff
@@ -40,12 +41,16 @@ func newTestDiscovery(t *testing.T, num int) []*discovery {
 			NoBootstrap: true,
 			NoMDNS:      true,
 		}
+		dstmpdir := t.TempDir()
+		ds, _ := badger.NewDatastore(path.Join(dstmpdir, "libp2p-datastore"), &badger.DefaultOptions)
```
noot, line 45: I think you can use an in memory datastore here

dot/network/gossip_test.go
```diff
@@ -42,7 +45,7 @@ func TestGossip(t *testing.T) {
 
 	nodeA := createTestService(t, configA)
 	handlerA := newTestStreamHandler(testBlockAnnounceMessageDecoder)
-	nodeA.host.registerStreamHandler("", handlerA.handleStream)
+	nodeA.host.registerStreamHandler(protocolToTest, handlerA.handleStream)
```
noot, line 48: just wondering why it needs a different protocol?

dot/network/transaction_test.go
```diff
@@ -90,7 +72,5 @@ func TestHandleTransactionMessage(t *testing.T) {
 	}
 
 	s.handleTransactionMessage(peer.ID(""), msg)
-	for _, tx := range msg.Extrinsics {
-		require.NotNil(t, handler.txs[tx.Hash()])
-	}
+	mockhandler.AssertCalled(t, "HandleTransactionMessage", msg)
```
noot, line 75: does this check that `HandleTransactionMessage` was called with `msg`?

dot/state/storage_notify_test.go
```diff
@@ -27,39 +27,26 @@ import (
...
-
 func TestStorageState_RegisterStorageObserver(t *testing.T) {
 	ss := newTestStorageState(t)
 
 	ts, err := ss.TrieState(nil)
 	require.NoError(t, err)
 
-	observer := &MockStorageObserver{}
-	ss.RegisterStorageObserver(observer)
+	//observer := &MockStorageObserver{}
```
noot, line 40: remove

edwardmack (approved):
Great work! Looks good to me.

lib/services/services.go
```diff
@@ -44,36 +43,40 @@ func NewServiceRegistry() *ServiceRegistry {
...
+		return
+	}
+
 	kind := reflect.TypeOf(service)
 	if _, exists := s.services[kind]; exists {
 		log.Warn("Tried to add service type that has already been seen", "type", kind)
 		return
 	}
 	s.services[kind] = service
-	s.serviceTypes = append(s.serviceTypes, kind)
```
noot, line 53: I think the problem with this change is that previously the services were being started and stopped in a deterministic order, but now when iterating over the map it's non-deterministic. this might mess up the start/stop logic, since some services need to be started/stopped before others, can you confirm starting and stopping the node still works without errors when changing this?

noot (commented):
looks good, just a few questions and a few changes needed with regards to code logic, but overall very nice cleanup!

dot/core/messages_test.go
EclesioMeloJunior: Yes! ust let the mocker know the function will be called with this argument

dot/network/transaction_test.go
EclesioMeloJunior: Yes, the mock checks if the method was called with the exact arguments

dot/network/discovery_test.go
EclesioMeloJunior: Done!

dot/network/gossip_test.go
EclesioMeloJunior: I've updated to a non-empty value just for test but now I put the original `""` value

dot/state/storage_notify_test.go
EclesioMeloJunior: Done!

dot/sync/syncer.go
```diff
@@ -90,7 +90,7 @@ func NewService(cfg *Config) (*Service, error) {
 	}
 
 	if cfg.BlockProducer == nil {
-		cfg.BlockProducer = newMockBlockProducer()
+		return nil, ErrNilBlockProducer
```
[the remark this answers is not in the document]
EclesioMeloJunior: Done!

lib/services/services.go
EclesioMeloJunior: I've started the node and sometime later stoped with **ctrl+c**

> ./bin/gossamer --chain dev
> ... waiting some time
> ^CINFO[06-11|12:42:46] signal interrupt, shutting down...       pkg=dot caller=node.go:432
> INFO[06-11|12:42:46] Stopping services...                     caller=services.go:74

I think the node stoped without errors but I just have rolled back the code to avoid unexpected node failures

noot: nice, I'm not 100% sure if it would cause errors if it's non deterministic but I remember previously it had to be in a certain order. maybe we can revisit this later to see if it's still an issue

dot/core/test_helpers.go
```diff
@@ -29,16 +28,16 @@ import (
...
 	"github.com/ChainSafe/gossamer/lib/runtime/wasmer"
 	"github.com/ChainSafe/gossamer/lib/trie"
 	log "github.com/ChainSafe/log15"
+	"github.com/stretchr/testify/mock"
 	"github.com/stretchr/testify/require"
-)
 
-// testMessageTimeout is the wait time for messages to be exchanged
-var testMessageTimeout = time.Second
+	// importing packagemocks
```
noot, line 38: remove comment

dot/network/gossip_test.go
```diff
@@ -29,6 +29,7 @@ import (
 func TestGossip(t *testing.T) {
 	if testing.Short() {
 		t.Skip("skipping TestGossip; currently, nothing is gossiped")
+		return
```
noot, line 32: don't think this is needed, can remove

noot (approved):
great work :D

dot/network/gossip_test.go
EclesioMeloJunior: Done!

PR #1627, opened 2021-06-08
maintainence: attempt to fix network tests on CI 

noot:
## Changes

- sleep after stopping network service so give it time to shut down, increase `host.connect` timeout
- remove `RandSeed` usage in tests

## Tests

```
go test ./dot/network -short
```

## Issues

-

### DIFF
--- a/dot/network/notifications_test.go
+++ b/dot/network/notifications_test.go
@@ -126,13 +126,11 @@ func TestCreateNotificationsMessageHandler_BlockAnnounce(t *testing.T) {
 	testPeerID := b.host.id()
 
 	// connect nodes
-	addrInfosB, err := b.host.addrInfos()
-	require.NoError(t, err)
-
-	err = s.host.connect(*addrInfosB[0])
+	addrInfoB := b.host.addrInfo()
+	err := s.host.connect(addrInfoB)
 	if failedToDial(err) {
 		time.Sleep(TestBackoffTimeout)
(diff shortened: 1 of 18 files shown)


dot/network/notifications_test.go
```diff
@@ -247,32 +234,50 @@ func TestCreateNotificationsMessageHandler_BlockAnnounceHandshake(t *testing.T)
...
-	ha := createTestService(t, config)
+
+	nodeA := createTestService(t, configA)
+	nodeA.noGossip = true
+
+	basePathB := utils.NewTestBasePath(t, "nodeB")
+	configB := &Config{
+		BasePath:    basePathB,
+		Port:        7002,
+		RandSeed:    2,
```
edwardmack, line 252: Why is RandSeed used here when it was removed elsewhere?

edwardmack (approved):

EclesioMeloJunior (approved):

dot/network/service_test.go
```diff
@@ -113,6 +111,7 @@ func createTestService(t *testing.T, cfg *Config) (srvc *Service) {
 
 	t.Cleanup(func() {
 		srvc.Stop()
+		time.Sleep(time.Second)
```
arijitAD, line 114: @noot We should remove this change if the test is passing without it. It will slow down the test suite.

PR #1633, opened 2021-06-10
Get node cache

### LINKED ISSUE #1604: Use a map to search nodes in the blocktree.
## Issue summary

- Use a map to search nodes in the block tree instead of iterating over the entire tree.
- Block tree can be redesigned to optimized operations. 

- https://github.com/ChainSafe/gossamer/blob/be3c325ac5ad3031a0ee9b89d1c32edc4a7da9bd/lib/blocktree/blocktree.go#L153

danigomez:
## Changes

- Cache each node returned by the `getNode` call
- Clean up the cache each time the tree is pruned

## Tests

```
go test -run TestBlockTree_GetNodeCache
go test -run TestBlockTree_PruneCache
```

## Issues

- closes #1604


lib/blocktree/blocktree.go
```diff
@@ -150,7 +153,19 @@ func (bt *BlockTree) GetAllBlocksAtDepth(hash common.Hash) []common.Hash {
 }
 
 // getNode finds and returns a node based on its Hash. Returns nil if not found.
-func (bt *BlockTree) getNode(h Hash) *node {
+func (bt *BlockTree) getNode(h Hash) (ret *node) {
+	defer func() {
+		if ret != nil {
+			if _, ok := bt.blockCache[ret.hash]; !ok {
+				bt.blockCache[ret.hash] = ret
```
EclesioMeloJunior, line 160: Would be nice to have a limit in the `blockCache` and when the `blockCache` reaches the limit then remove the first one index and insert the new `hash -> node`

lib/blocktree/blocktree.go
```diff
@@ -38,14 +38,16 @@ type BlockTree struct {
 	leaves *leafMap
 	db     database.Database
 	sync.RWMutex
+	blockCache map[Hash]*node
```
noot, line 41: name `nodeCache` or just `nodes`

lib/blocktree/blocktree.go
```diff
@@ -150,7 +153,19 @@ func (bt *BlockTree) GetAllBlocksAtDepth(hash common.Hash) []common.Hash {
...
 
 // getNode finds and returns a node based on its Hash. Returns nil if not found.
-func (bt *BlockTree) getNode(h Hash) *node {
+func (bt *BlockTree) getNode(h Hash) (ret *node) {
+	defer func() {
+		if ret != nil {
+			if _, ok := bt.blockCache[ret.hash]; !ok {
+				bt.blockCache[ret.hash] = ret
+			}
+		}
```
noot, line 162:
```suggestion
		if ret == nil {
		     return
		 }
		 
	if _, has := bt.blockCache[ret.hash]; !has {
		bt.blockCache[ret.hash] = ret
	}
```

nit :p

lib/blocktree/blocktree.go
```diff
@@ -350,7 +370,8 @@ func (bt *BlockTree) DeepCopy() *BlockTree {
 	defer bt.RUnlock()
 
 	btCopy := &BlockTree{
-		db: bt.db,
+		db:         bt.db,
+		blockCache: bt.blockCache,
```
noot, line 374: this doesn't perform a deep copy, since the nodes in the map still reference the nodes in `bt`. I think you can just remove this for now

noot (changes requested):
please also add node to cache in `AddBlock` and update `DeepCopy`, looks good otherwise

lib/blocktree/blocktree.go
danigomez: I'm not sure what criteria to use to remove the hashes... Would it be ok to remove randomly if the limit is reached??  Or maybe remove the least recently used adding a counter for each entry (I think that maybe this is a little overkill)??

What do you think?

lib/blocktree/blocktree.go
danigomez: Oops, i missed that.

lib/blocktree/blocktree.go
noot: I don't think you need to do this, the blocktree shouldn't ever grow too large

lib/blocktree/blocktree.go
danigomez: I added the implementation of the nodeCache copy!

EclesioMeloJunior (approved):
Looks good to me, great work!

lib/blocktree/blocktree.go
```diff
@@ -149,8 +153,24 @@ func (bt *BlockTree) GetAllBlocksAtDepth(hash common.Hash) []common.Hash {
...
+
+	if _, has := bt.nodeCache[b.hash]; !has {
+		bt.nodeCache[b.hash] = b
+	}
+}
+
 // getNode finds and returns a node based on its Hash. Returns nil if not found.
-func (bt *BlockTree) getNode(h Hash) *node {
+func (bt *BlockTree) getNode(h Hash) (ret *node) {
+	defer func() { bt.setInCache(ret) }()
```
noot, line 168:
```suggestion
	defer bt.setInCache(ret)
```

lib/blocktree/blocktree.go
```diff
@@ -368,5 +394,9 @@ func (bt *BlockTree) DeepCopy() *BlockTree {
 		}
 	}
 
+	for hash := range bt.nodeCache {
+		btCopy.nodeCache[hash] = btCopy.getNode(hash)
```
noot, line 398: this still references the node `bt`, you will need to allocate a new `*node` and copy the fields over to create a deep copy

lib/blocktree/blocktree_test.go
```diff
@@ -386,12 +411,46 @@ func TestBlockTree_Prune(t *testing.T) {
...
 
 	require.Equal(t, bt.db, btCopy.db)
+	for hash := range bt.nodeCache {
+		b, ok := btCopy.nodeCache[hash]
+		b2 := bt.nodeCache[hash]
+
+		require.True(t, ok)
+		require.True(t, b != b2)
+
+		require.True(t, equalNodeValue(b, b2))
```
noot, line 451: please test deep copy by checking that modifying a node with some hash in the copy doesn't modify the node with the same hash in the original

lib/blocktree/blocktree.go
danigomez: Aren't the nodes in `btCopy` already a deepCopy of each node in `bt`??

noot: nevermind, I missed that. this is fine then. however since `getNode` already places it in the cache, you can change this line to `_ = btCopy.getNode(hash)`

lib/blocktree/blocktree.go
danigomez: Correct me if i'm wrong but wouldn't that cause the `ret` param to be evaluated as `nil`, because it is immediatly evaluated in the `defer` call and not when the function returns??

lib/blocktree/blocktree.go
danigomez: Yep, i assigned the value just to make more explicit the intention of the code, but i can remove it if you want.

lib/blocktree/blocktree.go
noot: interesting, yeah looks like that is the case. this is good then!

noot (approved):

PR #1635, opened 2021-06-10
feat(dot/sync): implement codeSubstitutes

### LINKED ISSUE #1619: implement `codeSubstitutes`

edwardmack:
## Changes

- Implement code to handle code substitutes defined in genesis.json. wasm code is replaced with code contained in codeSubstitutes map when block is imported with same hash as key in codeSubstitutes map.
- (Note, this PR replaces PR#1620 which was closed due to issues with CLA check).
-

## Tests

```
go test ./dot/sync/... -v -run TestSyncer_H
```

## Issues

- closes #1619


EclesioMeloJunior (approved):
lgtm :D

dot/sync/syncer.go
```diff
@@ -424,6 +439,27 @@ func (s *Service) handleRuntimeChanges(newState *rtstorage.TrieState) error {
 		return ErrEmptyRuntimeCode
 	}
 
+	codeSubBlockHash, err := s.codeSubstitutedState.LoadCodeSubstitutedBlockHash()
+	if err != nil {
+		logger.Debug("problem checking code substituted block hash", "error", err)
```
noot, line 444: return err

dot/sync/syncer.go
```diff
@@ -424,6 +439,27 @@ func (s *Service) handleRuntimeChanges(newState *rtstorage.TrieState) error {
...
+	}
+
+	if !reflect.DeepEqual(codeSubBlockHash, common.Hash{}) {
+		// don't do runtime change if using code substitution and runtime change spec version are equal
+		//  (do a runtime change if code substituted and runtime spec versions are different, or code not substituted)
+		newVersion, err := s.runtime.CheckRuntimeVersion(code) // nolint
+		if err != nil {
+			logger.Debug("problem checking runtime version", "error", err)
```
noot, line 452: return err

dot/sync/syncer.go
```diff
@@ -434,6 +470,31 @@ func (s *Service) handleRuntimeChanges(newState *rtstorage.TrieState) error {
 	return nil
 }
 
+func (s *Service) handleCodeSubstitution(block common.Hash) error {
```
noot, line 473: rename to `hash`

dot/sync/syncer.go
```diff
@@ -424,6 +439,27 @@ func (s *Service) handleRuntimeChanges(newState *rtstorage.TrieState) error {
...
+		}
+
+		if previousVersion.SpecVersion() == newVersion.SpecVersion() {
+			return nil
+		}
+		logger.Info("🔄 detected runtime code change, upgrading...", "block", s.blockState.BestBlockHash(),
+			"previous code hash", s.codeHash, "new code hash", currCodeHash,
+			"previous spec version", previousVersion.SpecVersion(), "new spec version", newVersion.SpecVersion())
+	}
+
```
noot, line 462: you will also need to do `StoreCodeSubstitutedBlockHash(common.Hash{})` here to store in the db that no code substitution is currently happening, otherwise on restart the node will still think there's a code substitution

lib/runtime/wasmer/instance.go
```diff
@@ -216,6 +216,50 @@ func (in *Instance) UpdateRuntimeCode(code []byte) error {
...
+
+	// Assume imported memory is used if runtime does not export any
+	if !instance.HasMemory() {
+		instance.Memory = memory
+	}
+
+	instance.SetContextData(in.ctx)
+
+	in.vm = instance
+	checkVersion, err := in.Version()
```
noot, line 254: not sure why this is needed, you can just call `instance.Version()` directly and remove the `originalVM := in.vm` and `in.vm = originalVM` lines

lib/runtime/wasmer/instance.go
```diff
@@ -216,6 +216,50 @@ func (in *Instance) UpdateRuntimeCode(code []byte) error {
 	return nil
 }
 
+// CheckRuntimeVersion calculates runtime Version for runtime blob passed in
+func (in *Instance) CheckRuntimeVersion(code []byte) (runtime.Version, error) {
```
noot, line 220: this function has code that's mostly the same as `UpdateRuntimeVersion`, could you put the code that overlaps into a separate function and call that in both?

noot (commented):
tested with the network and it works :D functionality is working, just some code comments. also would be good not to change the gssmr genesis file, is there a way to use some test genesis instead or just modify the `genesisData` that's used to have a `codeSubstitution` field?

also please add this above `syncer.go` line 229 (inside the `if hasHeader && hasBody` case):
```
if err := s.handleCodeSubstitution(bd.Hash); err != nil {
    logger.Warn("failed to handle code substitution", "error", err)
    return i, err
}
```

this makes sure that the code substitution happens even when blocks are already stored in the db, in case the node crashed or used `--rewind`. had to add this to test out the changes since i had to rewind the chain

timwu20 (dismissed):

dot/sync/syncer.go
edwardmack: I updated codeSubstitutedState.LoadCodeSubstitutedBlockHash does not return error.

dot/sync/syncer.go
edwardmack: updated.

dot/sync/syncer.go
edwardmack: renamed.

dot/sync/syncer.go
edwardmack: Good catch, updated.

lib/runtime/wasmer/instance.go
edwardmack: I can't call `instance.Version()`  because instance is a wasmer instance which doesn't have Version() function.

lib/runtime/wasmer/instance.go
edwardmack: created `setupInstanceVM` function to handle redundant code.

dot/services.go
```diff
@@ -95,6 +96,23 @@ func createRuntime(cfg *Config, st *state.Service, ks *keystore.GlobalKeystore,
 		return nil, fmt.Errorf("failed to retrieve :code from trie: %s", err)
 	}
 
+	// check if code substitute is in use, if so replace code
+	codeSubHash, err := st.Base.LoadCodeSubstitutedBlockHash()
+	if err != nil {
+		logger.Debug("failed to retrieve substituted code block hash", "error", err)
+	}
+
+	if !reflect.DeepEqual(codeSubHash, common.Hash{}) {
```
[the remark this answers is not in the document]
edwardmack: Good call, I didn't realize common.Hash had an Equal function.  Updated.

dot/services.go
```diff
@@ -377,16 +395,26 @@ func createBlockVerifier(st *state.Service) (*babe.VerificationManager, error) {
 }
 
 func createSyncService(cfg *Config, st *state.Service, bp sync.BlockProducer, fg sync.FinalityGadget, dh *core.DigestHandler, verifier *babe.VerificationManager, rt runtime.Instance) (*sync.Service, error) {
```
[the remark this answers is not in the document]
edwardmack: updated.

edwardmack:
I think I've addressed all comments.  I've removed codeSubstitutes from gossamer.json genesis, and modified test helper NewTestSyncer to take a boolean usePolkadotGenesis flag, which when true uses polkadot genesis, when false uses gossamer genesis.

noot (approved):
great work!

Issue #1628, opened 2021-06-09
Fix block production timing logic


FlorianFranzen:
## Current Behavior

The timing logic used during block production is not optimal at the moment:

- You probably should not use [the full slot time](https://github.com/ChainSafe/gossamer/blob/development/lib/babe/build.go#L229) to collect extrinsics. For example substrate [reserves the last 1/3 of the slot time](https://github.com/paritytech/substrate/blob/3e5b4a2444c22561f5d4a995dec28d7ddc01865f/client/basic-authorship/src/basic_authorship.rs#L255) for finalization of the block.

- You should not abort collecting extrinsics [if the queue is or becomes empty](https://github.com/ChainSafe/gossamer/blob/development/lib/babe/build.go#L232), but continue to include new transaction throughout the slot to maximize block utilization.

(Also, why would an extrinsic [be empty](https://github.com/ChainSafe/gossamer/blob/development/lib/babe/build.go#L237)? Should this at least cause a warning?)

## Expected Behavior

- Extrinsics should be continued to be included into the block till about 2/3 of the slot time.

PR #1636, opened 2021-06-11
fix(lib/babe): fix timing for transition between epochs

noot:
## Changes

-  fix timing for transition between epochs, sometimes epoch was getting started slightly too early, so added a `time.Until` to wait until the actual start time
- update block production to use first 2/3 of slot duration, last 1/3 is reserved for finality as per substrate

## Tests

```
go test ./lib/babe
```

also, can run 2 nodes locally to test this out, let it run until it reaches the second epoch (epoch 1):
```
./bin/gossamer --key alice
./bin/gossamer --key bob --basepath /tmp/bob --port 7002
```

## Issues

- related to #1587 
- related to #1628


lib/babe/babe.go
```diff
@@ -226,20 +226,23 @@ func (b *Service) EpochLength() uint64 {
...
 	if b.paused {
 		return errors.New("service already paused")
 	}
 
-	b.Lock()
-	defer b.Unlock()
-
-	b.pause <- struct{}{}
 	b.paused = true
+	b.pause <- struct{}{}
```
timwu20, line 237: nit: seems odd that there are multiple two places in the code where `paused` is set to true.  Also having both `pause` and `paused` seems unnecessary.

lib/babe/babe.go
```diff
@@ -355,89 +355,109 @@ func (b *Service) initiate(epoch uint64) {
 }
 
 func (b *Service) invokeBlockAuthoring(epoch uint64) {
```
timwu20, line 357: does this service need to halt if `invokeBlockAuthoring` exits? If this function returned an `error`, appropriate actions could be taken in that case in `initiate`.

lib/babe/babe.go
```diff
@@ -355,89 +355,109 @@ func (b *Service) initiate(epoch uint64) {
...
-		logger.Error("failed to get start slot for current epoch", "epoch", epoch, "error", err)
-		return
-	}
+		// calculate current slot
+		startSlot := getCurrentSlot(b.slotDuration)
+		intoEpoch := startSlot - epochStart
 
-	intoEpoch := startSlot - epochStart
-	logger.Info("current epoch", "epoch", epoch, "slots into epoch", intoEpoch)
+		// if the calculated amount of slots "into the epoch" is greater than the epoch length,
```
timwu20, line 408: is it possible to move this into another function? This seems like something that happens infrequently and can be tested independently?

lib/babe/babe.go
noot: I'll change the other place to just call `Pause()` instead. it needs the `paused` bool in the `Resume` function to get if the service is paused already or not, is there a better way to do that?

lib/babe/babe.go
noot: yeah it should probably panic or `logger.Crit`. I can update that

lib/babe/babe.go
noot: it would happen if the node is down for over an epoch or if it's starting up for the first time on a network that's past epoch 0, which part would you put in a separate function? just this `if` check?

noot: going to end up removing this in my next PR actually

lib/babe/babe.go
noot: ended up removing the second place `paused` is set to true in my other PR

arijitAD: You can probably close the pause channel and reinitialize it on resume.
```
func (b *Service) IsPaused() {
	select {
	case: <-b.pause
		return true
	default:
		return false
	}
}
```

```
func (b *Service) Pause() {
	close(b.pause)
}
```

```
func otherFunc() {
	select {
	case <-time.After(time.Until(epochStartTime)):
	case <-b.ctx.Done():
		return nil
	case <-b.pause:
		return nil
	}
}
```

edwardmack (approved):
Great work, lgtm.
Testing running two nodes I see the transition to epoch 1:
```
INFO[06-15|10:34:40] epoch complete!                          pkg=babe completed epoch=0 upcoming epoch=1 caller=babe.go:463
INFO[06-15|10:34:40] initiating epoch                         pkg=babe number=1 first slot of epoch=541255882 caller=babe.go:385
INFO[06-15|10:34:40] current epoch                            pkg=babe epoch=1 slots into epoch=11 caller=babe.go:426
INFO[06-15|10:34:43] 💤 node waiting                        pkg=network peer count=1 head=7 finalised=0 caller=sync.go:325

```
Is it normal for transition epochs after only 7 blocks?
Is in ok that finilised = 0?

lib/babe/build.go
```diff
@@ -322,7 +322,7 @@ func (b *BlockBuilder) addToQueue(txs []*transaction.ValidTransaction) {
 }
 
 func hasSlotEnded(slot Slot) bool {
-	slotEnd := slot.start.Add(slot.duration)
+	slotEnd := slot.start.Add(slot.duration * 2 / 3) // reserve last 1/3 of slot for block finalisation
```
arijitAD, line 325: Instead of 1/3, Can we allocate a fixed time for block finalization?

noot: what do you mean exactly? 1/3 of the slot is a fixed slot duration, this is what substrate does afaik

noot:
@edwardmack yep that is normal, since the genesis file expects 6 block producers and not enough nodes are online to finalise. if you change the genesis to have only 2-3 block producers then there will be more blocks

lib/babe/build.go
arijitAD: I meant `500ms` or so. Instead of `1/3` of `slot.duration`

lib/babe/babe.go
```diff
@@ -301,13 +305,6 @@ func (b *Service) IsStopped() bool {
...
 
-// IsPaused returns if the service is paused or not (ie. producing blocks)
-func (b *Service) IsPaused() bool {
-	b.RLock()
-	defer b.RUnlock()
-	return b.paused
-}
-
 func (b *Service) safeSend(msg types.Block) error {
 	b.Lock()
```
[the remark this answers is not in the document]
arijitAD:
```
return errors.New("service has been stopped")
```
`S` -> `s`
Error starts with smallcase letters.

lib/babe/babe.go
```diff
@@ -228,13 +227,7 @@ func (b *Service) EpochLength() uint64 {
 func (b *Service) Pause() error {
 	b.Lock()
```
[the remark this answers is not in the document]
arijitAD: Remove the error

lib/babe/babe.go
```diff
@@ -243,22 +236,33 @@ func (b *Service) Resume() error {
...
```
arijitAD, line 236: Acquire lock after checking `b.IsPaused`

arijitAD:
Also, make `blockChan` buffered channel. https://github.com/ChainSafe/gossamer/issues/1642

EclesioMeloJunior (approved):

lib/babe/build.go
noot: no that would not be ideal, substrate uses 1/3 of the slot duration, you can see their implementation on the issue

lib/babe/babe.go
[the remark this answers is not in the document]
noot: there might be a chance that `Pause()` is called when the `b.pause` channel is already close, thus causing a close of closed channel panic

lib/babe/babe.go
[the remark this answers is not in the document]
noot: for your first comment, the lock is here so there isn't a send on a closed channel. maybe there is a better way to implement that? either way, I think I will be removing this section anyways and doing a refactor

lib/babe/babe.go
noot: I think this function should be atomic in the case that `Pause` or `Resume` are called concurrently by multiple processes, process A might see it's not paused, acquire the lock, then do `b.initiate()` but in between seeing it's not paused an acquiring the lock another process might also see the same thing, thus causing `b.initiate()` to be called twice which would be bad

PR #1638, opened 2021-06-14
fix(dot/state) Fix runtime upgrade

### LINKED ISSUE #1595: Determine steps for runtime upgrades using devnet

arijitAD:
## Changes

- Fix runtime upgrade.

## Tests

```
make it-polkadotjs
```

## Issues

- closes #1595


dot/services.go
```diff
@@ -90,18 +90,12 @@ func createStateService(cfg *Config) (*state.Service, error) {
 	return stateSrvc, nil
 }
 
-func createRuntime(cfg *Config, st *state.Service, ks *keystore.GlobalKeystore, net *network.Service) (runtime.Instance, error) {
+func createRuntime(cfg *Config, st *state.Service, ks *keystore.GlobalKeystore, net *network.Service, code []byte) (runtime.Instance, error) {
```
noot, line 93: you can load the code from `st.Storage.LoadCode()`

dot/state/block.go
```diff
@@ -183,6 +192,28 @@ func (bs *BlockState) GenesisHash() common.Hash {
 	return bs.genesisHash
 }
 
+// ValidateTransaction validates transaction
+func (bs *BlockState) ValidateTransaction(e types.Extrinsic) (*transaction.Validity, error) {
```
noot, line 196: where is this used? doesn't look like `core` uses it, seems weird for it to be in the block state anyways

dot/state/block.go
```diff
@@ -765,3 +798,99 @@ func (bs *BlockState) setArrivalTime(hash common.Hash, arrivalTime time.Time) er
...
+
+	rtCfg := &wasmer.Config{
+		Imports: wasmer.ImportsNodeRuntime,
+	}
+
+	localStorage, err := utils.SetupDatabase(filepath.Join(bs.dbPath, bHash.String(), "local_storage"), true)
+	if err != nil {
+		return err
+	}
+
```
noot, line 846: why does it need to setup a database and update the `localStorage`?

dot/state/block.go
```diff
@@ -765,3 +798,99 @@ func (bs *BlockState) setArrivalTime(hash common.Hash, arrivalTime time.Time) er
...
+// StoreRuntime stores the runtime for corresponding block hash.
+func (bs *BlockState) StoreRuntime(hash common.Hash, rt runtime.Instance) {
+	bs.bt.StoreRuntime(hash, rt)
+}
+
+// GetAllBlocks get all the blocks
+func (bs *BlockState) GetAllBlocks() []common.Hash {
+	return bs.bt.GetAllBlocks()
+}
```
noot, line 896: maybe name `GetAllUnfinalisedBlocks` for clarity?

lib/babe/state.go
```diff
@@ -44,12 +45,16 @@ type BlockState interface {
 	GetFinalizedHeader(uint64, uint64) (*types.Header, error)
 	IsDescendantOf(parent, child common.Hash) (bool, error)
 	NumberIsFinalised(num *big.Int) (bool, error)
+	HandleRuntimeChanges(newState *rtstorage.TrieState, in runtime.Instance, bHash common.Hash) error
```
noot, line 48: I don't think this is used in this package

lib/babe/state.go
```diff
@@ -44,12 +45,16 @@ type BlockState interface {
...
+	GetRuntime(*common.Hash) (runtime.Instance, error)
+	StoreRuntime(common.Hash, runtime.Instance)
```
noot, line 50: same with this

lib/babe/state.go
```diff
@@ -44,12 +45,16 @@ type BlockState interface {
...
 }
 
 // StorageState interface for storage state methods
 type StorageState interface {
 	TrieState(hash *common.Hash) (*rtstorage.TrieState, error)
 	StoreTrie(ts *rtstorage.TrieState, header *types.Header) error
+	LoadCodeHash(root *common.Hash) (common.Hash, error)
```
noot, line 57: same with this - can you also double check these in the `sync` package interfaces?

dot/core/service_test.go
```diff
@@ -453,20 +479,69 @@ func TestService_GetMetadata(t *testing.T) {
...
+	parentRt, err := s.blockState.GetRuntime(&hash)
+	require.NoError(t, err)
+
+	v, err := parentRt.Version()
+	require.NoError(t, err)
+	require.Equal(t, v.SpecVersion(), currSpecVersion)
+
+	bhash1 := newBlock1.Header.Hash()
+	err = s.blockState.HandleRuntimeChanges(ts, parentRt, bhash1)
+	require.NoError(t, err)
```
noot, line 519: can you check that the runtime for `newBlock1` is the same as the parent runtime?

noot (commented):
great work! no major issues, just some smaller comments. going to test this on the polkadot node to double check everything, will let you know if there are any issues

dot/core/service.go
```diff
@@ -59,10 +57,6 @@ type Service struct {
 	net              Network
 	digestHandler    DigestHandler
 
-	// Current runtime and hash of the current runtime code
-	rt       runtime.Instance
```
edwardmack, line 63: Should `Runtime` also be removed from `Config` struct since the service is no longer storing it?

dot/sync/syncer.go
```diff
@@ -46,7 +46,6 @@ type Service struct {
 	// Synchronisation variables
 	synced           bool
 	highestSeenBlock *big.Int // highest block number we have seen
-	runtime          runtime.Instance
```
edwardmack, line 49: Also remove `Runtime` from `Config`.

lib/babe/babe.go
```diff
@@ -56,9 +56,6 @@ type Service struct {
 	// BABE authority keypair
 	keypair *sr25519.Keypair // TODO: change to BABE keystore
 
-	// Current runtime
-	rt runtime.Instance
```
edwardmack, line 60: Also remove `Runtime` from `ServiceConfig` struct.

edwardmack (approved):
Great work!

dot/services.go
arijitAD: I think then we need to pass the `block hash` as params. otherwise, we can only get the code corresponding to the best block.

dot/state/block.go
arijitAD: Removed. I added for author module RPC. But it seems the author module doesn't use it.

dot/state/block.go
arijitAD: Done

lib/babe/state.go
arijitAD: Done

lib/babe/state.go
arijitAD: `GetRuntime` is used in `handleSlot`

lib/babe/state.go
arijitAD: Removed

dot/core/service.go
arijitAD: The test is using this runtime. I have created another issue to remove the runtime. https://github.com/ChainSafe/gossamer/issues/1674

dot/sync/syncer.go
arijitAD: The test is using this runtime. I have created another issue to remove the runtime. #1674

lib/babe/babe.go
arijitAD: The test is using this runtime. I have created another issue to remove the runtime. #1674

lib/babe/state.go
arijitAD: This is used in test. I have added an issue to remove this https://github.com/ChainSafe/gossamer/issues/1674

dot/services.go
noot: just realized, when the node restarts, do all the blocktree nodes get repopulated with the runtime? it doesn't look like the `blocktree.Encode` and `blocktree.Decode` functions have been updated

lib/babe/state.go
noot: if it's used in the test, I prefer casting the test state to the actual `*state.BlockState` and having the interface contain functions only used by the non-test code

timwu20 (approved):

noot:
@arijitAD just checking in on this:
> when the node restarts, do all the blocktree nodes get repopulated with the runtime? it doesn't look like the blocktree.Encode and blocktree.Decode functions have been updated

PR #1640, opened 2021-06-14
fix(lib/babe): fix setting first slot of network, fix loading BABE epoch params

### LINKED ISSUE #1624: lib/babe: first slot of network should be set on finalization of block 1
## Describe the bug

- the first slot of the network is determined by when block 1 was built
- however gossamer currently sets the first slot of the network in 2 places (if an authority node), when just after building block 1 OR when importing block 1. this potentially results in being unable to validate imported blocks or the BABE algorithm becoming un-synced from the network as the first slot is set incorrectly
- the first slot number should be set on finalization of block 1, instead of build/import
- in the case where block 1 has not yet been finalized, there should be some list of unfinalized blocks with block number 1. when verifying a block and block 1 is not yet finalized, it should first use the slot number of block 1 on the canonical chain (which is what currently happens) as the first slot, but if that errors, try using other pending block 1.

sub-issue of #1587

noot:
## Changes

- set first slot of network on finalisation, can no longer be changed after this point
- fix BABE timing logic with regards to starting an epoch, now waits until exact start time before initiating
- fix first slot setting if node is a block producer, it should check when block 1 is going to be produced to set the first slot instead of using the time the service was started
- fix loading BABE epoch params on startup, previously was re-loading genesis params even if at epoch > 0

## Tests

```
go test ./dot/state
go test ./lib/babe
```

## Issues

- closes #1624


dot/state/block_test.go
```diff
@@ -520,3 +534,88 @@ func TestAddBlockToBlockTree(t *testing.T) {
...
+	err = bs.SetFinalizedHash(header100.Hash(), 0, 0)
+	require.NoError(t, err)
+
+	fin, err = bs.NumberIsFinalised(big.NewInt(0))
+	require.NoError(t, err)
+	require.True(t, fin)
+
+	fin, err = bs.NumberIsFinalised(big.NewInt(1))
+	require.NoError(t, err)
+	require.True(t, fin)
```
EclesioMeloJunior, line 580: 🙋‍♂️: wondering why header number 01 is finalized without being called in `bs.SetFinalizedHash`? Is it related to `setFirstSlotOnFinalisation` function?

noot: block with number 100 was set as finalized, therefore a block with number 1 is also finalized since there had to be a block 1 that is an ancestor of block 100

lib/babe/babe.go
```diff
@@ -154,65 +158,83 @@ func NewService(cfg *ServiceConfig) (*Service, error) {
...
+func (b *Service) setupParameters(cfg *ServiceConfig) error {
+	var err error
+	b.epochData = &epochData{}
+
+	epochData, err := b.epochState.GetLatestEpochData()
+	if err != nil {
+		return err
+	}
+
+	fmt.Println(epochData)
```
EclesioMeloJunior, line 170: remove

EclesioMeloJunior (changes requested):

lib/babe/babe.go
noot: removed

lib/babe/babe_test.go
```diff
@@ -171,6 +177,133 @@ func TestMain(m *testing.M) {
...
+
+	s := &Service{
+		epochState: dbSrv.Epoch,
+	}
+
+	return s, dbSrv.Epoch, genCfg
+}
+
+func TestService_setupParameters_genesis(t *testing.T) {
+	s, _, genCfg := newTestService_setupParameters(t)
```
arijitAD, line 219: Don't use underscores in Go names
`newTestService_setupParameters` -> `newTestServiceSetupParameters`

lib/babe/babe.go
```diff
@@ -154,65 +158,82 @@ func NewService(cfg *ServiceConfig) (*Service, error) {
...
+	b.epochData.randomness = epochData.Randomness
+
+	configData, err := b.epochState.GetLatestConfigData()
+	if err != nil {
+		return err
 	}
 
 	// if slot duration is set via the config file, overwrite the runtime value
-	if cfg.SlotDuration > 0 {
+	if cfg.SlotDuration > 0 && cfg.IsDev { // TODO: remove this, needs to be set via runtime
```
arijitAD, line 178: Changing this to a switch case may clean up the code.

lib/babe/epoch.go
```diff
@@ -41,24 +41,18 @@ func (b *Service) initiateEpoch(epoch uint64) error {
...
 		}
 
+		data, err := b.epochState.GetEpochData(epoch)
 		if err != nil {
 			return err
 		}
 
 		idx, err := b.getAuthorityIndex(data.Authorities)
-		if err != nil && err != ErrNotAuthority {
+		if err != nil && err != ErrNotAuthority { // TODO: this should be checked in the upper function
```
arijitAD, line 55: Use `errors.Is` to check for a specific error

lib/babe/epoch.go
```diff
@@ -97,20 +91,38 @@ func (b *Service) initiateEpoch(epoch uint64) error {
...
 		}
-	}
 
-	if !b.authority {
-		return nil
+		logger.Debug("estimated first slot based on building block 1", "slot", startSlot)
+		for i := startSlot; i < startSlot+b.epochLength; i++ {
+			proof, err := b.runLottery(i, epoch) //nolint
+			if err != nil {
+				return fmt.Errorf("error running slot lottery at slot %d: error %s", i, err)
```
arijitAD, line 109: Use `%w` to format errors

lib/babe/epoch.go
```diff
@@ -131,6 +143,23 @@ func (b *Service) initiateEpoch(epoch uint64) error {
 	return nil
 }
 
+func (b *Service) getFirstSlot(epoch uint64) (uint64, error) {
+	startSlot := getCurrentSlot(b.slotDuration)
+	for i := startSlot; i < startSlot+b.epochLength; i++ {
+		proof, err := b.runLottery(i, epoch)
+		if err != nil {
+			return 0, fmt.Errorf("error running slot lottery at slot %d: error %s", i, err)
```
arijitAD, line 151: Use `%w` to format errors

dot/state/block.go
```diff
@@ -431,9 +456,24 @@ func (bs *BlockState) SetFinalizedHash(hash common.Hash, round, setID uint64) er
 		bs.pruneKeyCh <- header
 	}
 
+	bs.lastFinalised = hash
```
arijitAD, line 459: Can we set this before pruning and do the pruning in the background?

lib/babe/babe.go
noot: you mean something like:
```
switch cfg.IsDev {
case true:
   if cfg.SlotDuration > 0 { // do stuff }
case false:
   if cfg.SlotDuration > 0 { // error }
}
``` 
?

dot/state/block.go
noot: the pruning needs to be synchronous, otherwise we might build on invalid chains or end up re-adding invalid chains. it doesn't make a difference whether it's before the other code or not, since this is the only function that uses `bs.lastFinalised` and it's all within a mutex

lib/babe/epoch.go
noot: done

lib/babe/epoch.go
noot: done

lib/babe/babe_test.go
noot: done

lib/babe/epoch.go
noot: done

edwardmack (approved):
Great work, lgtm.

lib/babe/babe.go
```diff
@@ -446,6 +436,32 @@ func (b *Service) invokeBlockAuthoring(epoch uint64) error {
...
+	}
+
+	epochStartTime := getSlotStartTime(epochStart, b.slotDuration)
+	logger.Debug("checking if epoch started", "epoch start", epochStartTime, "now", time.Now())
+
+	// check if it's time to start the epoch yet. if not, wait until it is
+	if time.Since(epochStartTime) < 0 {
+		logger.Debug("waiting for epoch to start")
+		select {
+		case <-time.After(time.Until(epochStartTime)):
```
EclesioMeloJunior, line 454: Maybe could use `timer.NewTimer()`?

lib/babe/babe.go
timwu20: I don't mind nested switches as well. But a switch on `cfg.IsDev` would make it much easier to read.

lib/babe/babe.go
```diff
@@ -154,65 +158,82 @@ func NewService(cfg *ServiceConfig) (*Service, error) {
...
-	if err != nil {
-		logger.Error("failed to get current epoch", "error", err)
-		return err
+	if !b.authority {
+		return nil
 	}
 
-	go b.initiate(epoch)
+	// wait a bit to check if we need to sync before initiating
+	time.Sleep(initialWaitTime)
```
timwu20, line 234:
```suggestion
	<- time.NewTimer(initialWaitTime).C
```

lib/babe/babe.go
```diff
@@ -446,6 +436,32 @@ func (b *Service) invokeBlockAuthoring(epoch uint64) error {
...
 
+func (b *Service) waitForEpochStart(epoch uint64) (uint64, error) {
+	// get start slot for current epoch
+	epochStart, err := b.epochState.GetStartSlotForEpoch(epoch)
+	if err != nil {
+		logger.Error("failed to get start slot for current epoch", "epoch", epoch, "error", err)
+		return 0, err
+	}
+
+	epochStartTime := getSlotStartTime(epochStart, b.slotDuration)
```
timwu20, line 447: nit: you could make `getSlotStartTime` a receiver function on `Service` and can omit the `slotDuration` param.

lib/babe/babe.go
arijitAD:
```
	switch {
	case cfg.SlotDuration > 0 && cfg.IsDev:
		b.slotDuration, err = time.ParseDuration(fmt.Sprintf("%dms", cfg.SlotDuration))
	case cfg.SlotDuration > 0 && !cfg.IsDev:
		err = errors.New("slot duration modified in config for non-dev chain")
	default:
		b.slotDuration, err = b.epochState.GetSlotDuration()
	}
```

You can do this for the below code as well.

dot/state/epoch.go
```diff
@@ -139,26 +143,26 @@ func NewEpochState(db chaindb.Database) (*EpochState, error) {
...
 		skipToEpoch: skipToEpoch,
 	}, nil
 }
 
-func storeEpochLength(db chaindb.Database, l uint64) error {
-	buf := make([]byte, 8)
-	binary.LittleEndian.PutUint64(buf, l)
-	return db.Put(epochLengthKey, buf)
+// GetEpochLength returns the length of an epoch in slots
+func (s *EpochState) GetEpochLength() (uint64, error) {
```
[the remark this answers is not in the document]
noot: it's named with `Get` to match the corresponding `Set` function, is it more idiomatic to have it with no `Get`?

lib/babe/babe.go
noot: is this more efficient?

lib/babe/babe.go
noot: I kinda like it like this cause it seems more generic, but happy to change it if you think it's better as a receiver

dot/node_test.go
```diff
@@ -190,7 +190,7 @@ func TestStartNode(t *testing.T) {
 	go func() {
 		// TODO: need to wait until all services are started so that wg.Add is called, otherwise
 		// will call wg.Done before the counter is at 1
-		time.Sleep(time.Second * 15)
+		time.Sleep(time.Second * 20)
```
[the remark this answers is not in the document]
noot: updated to have a `started` channel in `*Node` that's closed when all the services are started

lib/babe/babe.go
noot: done

lib/babe/babe.go
noot: update

lib/babe/babe.go
noot: updated btw, just curious about the difference

EclesioMeloJunior: I think the `Start` function form `babe` package is not resource-intensive, `time.NewTimer` or `time.After` will fit well than `time.Sleep`. obs: in the case of `time.NewTimer`, would be a good practice calls `defer t.Stop()`

noot: this will always block until the timer fires so there isn't a point to calling `Stop` here afaik

EclesioMeloJunior (approved):

PR #1643, opened 2021-06-15
fix(dot/core): check transaction Validity.Propagate field to determine whether to propagate tx

### LINKED ISSUE #1610: check transaction `Validity.Propagate` field to determine whether to propagate tx
## Task summary

- when a tx is validated and put in the pool it has a `Validity` field with contains a `Propagate` bool
- this should be checked by the `network.Service.handleTransactionMessage` function which should return whether to propagate the tx or not

edwardmack:
## Changes

- add check to HandleTransactionMessage  to check Validity.Propagate field and determine if tx should be propagated. 
-
-

## Tests

```

```

## Issues

- closes #1610


noot (changes requested):
this doesn't implement the functionality of the issue which is to use the propagate field to check whether to propagate the tx to the rest of the network, you will need to update the `handleTransactionMessage` function in `dot/network/transaction.go` for this. the `bool` it returns is whether to propagate the message or not

dot/core/messages.go
```diff
@@ -17,29 +17,31 @@
...
 
 // HandleTransactionMessage validates each transaction in the message and
 // adds valid transactions to the transaction queue of the BABE session
-func (s *Service) HandleTransactionMessage(msg *network.TransactionMessage) error {
+func (s *Service) HandleTransactionMessage(msg *network.TransactionMessage) (bool, error) {
 	logger.Debug("received TransactionMessage")
 	if !s.isBlockProducer {
-		return nil
+		return false, nil
 	}
```
noot, line 33: you can remove these lines, we should be validating + propagating tx messages even if we aren't block producers

dot/core/messages.go
```diff
@@ -17,29 +17,31 @@
...
-
+	var toRemove []types.Extrinsic
 	for _, tx := range txs {
 		// validate each transaction
 		externalExt := types.Extrinsic(append([]byte{byte(types.TxnExternal)}, tx...))
 		val, err := s.rt.ValidateTransaction(externalExt)
 		if err != nil {
 			logger.Error("failed to validate transaction", "err", err)
-			return err
+			return false, err
```
noot, line 44: don't return here but continue, and mark this tx "do not propagate" instead

dot/core/messages.go
```diff
@@ -48,7 +50,28 @@ func (s *Service) HandleTransactionMessage(msg *network.TransactionMessage) erro
 		// push to the transaction queue of BABE session
 		hash := s.transactionState.AddToPool(vtx)
 		logger.Trace("Added transaction to queue", "hash", hash)
+
+		// find tx(s) that should not propagate
+		if !val.Propagate {
+			toRemove = append(toRemove, tx)
+		}
```
noot, line 57: it might be more straightforward to have a list of extrinsics to propagate instead of `toRemove`, and add valid txs with `Propagate == true` to it, that way you don't need to use `findAndDelete`

noot (commented):
logic on the network side looks good, just some code suggestions!

dot/core/messages.go
edwardmack: removed.

dot/core/messages.go
edwardmack: makes sense, updated.

dot/core/messages.go
edwardmack: Yes, more straightforward and readable, updated.

edwardmack:
Update code to address comments, ready for re-review.

dot/core/messages.go
```diff
@@ -24,19 +24,20 @@ import (
 
 // HandleTransactionMessage validates each transaction in the message and
 // adds valid transactions to the transaction queue of the BABE session
```
EclesioMeloJunior, line 26: Would be nice to have a description that the bool return means there is extrinsic to propagate.

EclesioMeloJunior (approved):

dot/core/messages.go
edwardmack: good point... Updated.

dot/core/messages.go
```diff
@@ -24,19 +24,21 @@ import (
...
-
+	var toPropagate []types.Extrinsic
 	for _, tx := range txs {
 		// validate each transaction
 		externalExt := types.Extrinsic(append([]byte{byte(types.TxnExternal)}, tx...))
 		val, err := s.rt.ValidateTransaction(externalExt)
 		if err != nil {
 			logger.Error("failed to validate transaction", "err", err)
-			return err
+			val.Propagate = false
```
noot, line 40: I think `val` will be `nil` if there's an error returned? you can remove this anyways, as it will move to the next tx

dot/core/messages_test.go
```diff
@@ -157,8 +157,9 @@ func TestService_HandleTransactionMessage(t *testing.T) {
 	extBytes := createExtrinsics(t, s.rt, genHash, 0)
 
 	msg := &network.TransactionMessage{Extrinsics: []types.Extrinsic{extBytes}}
-	err = s.HandleTransactionMessage(msg)
+	b, err := s.HandleTransactionMessage(msg)
```
noot, line 160: can you also add a test case for invalid extrinsics?

timwu20 (approved):

dot/core/messages.go
```diff
@@ -24,19 +24,21 @@ import (
 
 // HandleTransactionMessage validates each transaction in the message and
 // adds valid transactions to the transaction queue of the BABE session
-func (s *Service) HandleTransactionMessage(msg *network.TransactionMessage) error {
+// returns boolean for transaction propagation, true - transactions should be propagoted
```
arijitAD, line 27: `propagoted` -> `propagated`

arijitAD (approved):

dot/core/messages.go
edwardmack: updated.

dot/core/messages_test.go
edwardmack: added.

dot/core/messages.go
edwardmack: fixed.

dot/core/messages.go
```diff
@@ -24,19 +24,20 @@ import (
...
 	// get transactions from message extrinsics
 	txs := msg.Extrinsics
-
+	var toPropagate []types.Extrinsic
 	for _, tx := range txs {
 		// validate each transaction
 		externalExt := types.Extrinsic(append([]byte{byte(types.TxnExternal)}, tx...))
 		val, err := s.rt.ValidateTransaction(externalExt)
 		if err != nil {
 			logger.Error("failed to validate transaction", "err", err)
```
noot, line 39: I'd change this to debug since it's not an error

noot (approved):
looks good!

dot/core/messages.go
edwardmack: good point, updated.

Issue #1623, opened 2021-06-04
Use time.Timer instead of time.After in either-or cases to avoid memory leak


arijitAD:
## Issue summary

The docs for [time.After](https://golang.org/pkg/time/#After) says

> After waits for the duration to elapse and then sends the current time on the returned channel. It is equivalent to NewTimer(d).C. The underlying Timer is not recovered by the garbage collector until the timer fires. If efficiency is a concern, use NewTimer instead and call Timer.Stop if the timer is no longer needed.

We should replace `time.After` with `time.Timer` if we want to poll some code to avoid unnecessary allocation and deallocation.

Benchmark to compare `time.After` vs `time.Sleep` vs `time.Timer` vs `time.Ticker`
```
BenchmarkSleep-16    	   40676	     32230 ns/op	       0 B/op	       0 allocs/op
BenchmarkAfter-16    	   10000	    114656 ns/op	  200001 B/op	    3000 allocs/op
BenchmarkReset-16    	   27517	     44791 ns/op	     200 B/op	       3 allocs/op
BenchmarkTicker-16    	   31257	     35621 ns/op	     200 B/op	       3 allocs/op
```
https://play.golang.org/p/jfmFQQQv-0I

## Other information and links

- https://golang.org/src/time/sleep.go?s=5504:5538#L150

PR #1650, opened 2021-06-21
chore: replace time.After with time.NewTicker

edwardmack:
## Changes

- replace time.After with time.Ticker in either-or cases to avoid memory leak

## Tests

```
make test
```

## Issues

- closes: #1623


arijitAD:
It is fine to use `time.After` in the test and we should only replace `time.After` that does polling.
Example: dot/network/discovery.go
```
func (d *discovery) start() error {
	if len(d.bootnodes) == 0 {
		// get all currently connected peers and use them to bootstrap the DHT
		peers := d.h.Network().Peers()

		for {
			if len(peers) > 0 {
				break
			}

			select {
			case <-time.After(startDHTTimeout):
				logger.Debug("no peers yet, waiting to start DHT...")
				// wait for peers to connect before starting DHT, otherwise DHT bootstrap nodes
				// will be empty and we will fail to fill the routing table
			case <-d.ctx.Done():
				return nil
			}

			peers = d.h.Network().Peers()
		}

		for _, p := range peers {
			d.bootnodes = append(d.bootnodes, d.h.Peerstore().PeerInfo(p))
		}
	}
```

```
// start creates the DHT.
func (d *discovery) start() error {
	if len(d.bootnodes) == 0 {
		// get all currently connected peers and use them to bootstrap the DHT
		peers := d.h.Network().Peers()

		t := time.NewTicker(startDHTTimeout)
		defer t.Stop()
		for {
			if len(peers) > 0 {
				break
			}

			select {
			case <-t.C:
				logger.Debug("no peers yet, waiting to start DHT...")
				// wait for peers to connect before starting DHT, otherwise DHT bootstrap nodes
				// will be empty and we will fail to fill the routing table
			case <-d.ctx.Done():
				return nil
			}

			peers = d.h.Network().Peers()
		}
```

EclesioMeloJunior (approved):

lib/babe/babe.go
```diff
@@ -379,8 +379,10 @@ func (b *Service) invokeBlockAuthoring(epoch uint64) error {
 		// check if it's time to start the epoch yet. if not, wait until it is
 		if time.Since(epochStartTime) < 0 {
 			logger.Debug("waiting for epoch to start")
+			t := time.NewTicker(time.Until(epochStartTime))
+			defer t.Stop()
 			select {
-			case <-time.After(time.Until(epochStartTime)):
+			case <-t.C:
```
noot, line 385: this isn't in a `for` loop, it's a one-time check, so you can leave this as it

edwardmack: Good point, changed back.

arijitAD (approved):

dot/network/sync.go
```diff
@@ -214,9 +216,11 @@ func (q *syncQueue) syncAtHead() {
 }
 
 func (q *syncQueue) handleResponseQueue() {
+	t := time.NewTicker(time.Second)
```
timwu20, line 219: can we make these durations configurable to the type, so `syncQueue` in this case?  It will help with testing and we can adjust the durations in the unit tests.

edwardmack: good idea, I've added configuration variables.

PR #1651, opened 2021-06-21
refactor(lib/scale): Revise VaryingDataType interfaces, and introduce VaryingDataTypeSlice

timwu20:
## Changes

- Revise `VaryingDataType` public interface
- Introduce `VaryingDataTypeSlice` to handle array of enums currently used in `dot/types/digest.go`

## Tests

```
go test ./pkg/scale
```

## Issues

- #1548

### DIFF
--- a/pkg/scale/decode.go
+++ b/pkg/scale/decode.go
@@ -27,7 +27,7 @@ import (
 
 // indirect walks down v allocating pointers as needed,
 // until it gets to a non-pointer.
-func indirect(dstv reflect.Value) (elem reflect.Value, err error) {
+func indirect(dstv reflect.Value) (elem reflect.Value) {
 	dstv0 := dstv
 	haveAddr := false
 	for {
@@ -68,6 +68,7 @@ func indirect(dstv reflect.Value) (elem reflect.Value, err error) {
 	return
 }
 
+// Unmarshal takes data and a destination pointer to unmarshal the data to.
 func Unmarshal(data []byte, dst interface{}) (err error) {
 	dstv := reflect.ValueOf(dst)
 	if dstv.Kind() != reflect.Ptr || dstv.IsNil() {
@@ -75,7 +76,7 @@ func Unmarshal(data []byte, dst interface{}) (err error) {
 		return
 	}
 
-	elem, err := indirect(dstv)
+	elem := indirect(dstv)
 	if err != nil {
 		return
 	}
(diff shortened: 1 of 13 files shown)


pkg/scale/varying_data_type.go
```diff
@@ -18,34 +18,94 @@ package scale
...
+// NewVaryingDataTypeSlice is constructor for VaryingDataTypeSlice
+func NewVaryingDataTypeSlice(vdt VaryingDataType) (vdts VaryingDataTypeSlice) {
+	vdts.VaryingDataType = vdt
+	vdts.Types = make([]VaryingDataType, 0)
+	return
+}
+
+func mustNewVaryingDataTypeSliceAndSet(vdt VaryingDataType, values ...VaryingDataTypeValue) (vdts VaryingDataTypeSlice) {
+	vdts = NewVaryingDataTypeSlice(vdt)
+	err := vdts.Add(values...)
```
arijitAD, line 57: Use short syntax 
```
if err := vdts.Add(values...); err != nil {
    panic(err)
}
```

pkg/scale/decode.go
```diff
@@ -510,56 +573,56 @@ func (ds *decodeState) decodeFixedWidthInt(dstv reflect.Value) (err error) {
 		var b byte
 		b, err = ds.ReadByte()
 		if err != nil {
-			break
+			return
```
edwardmack, line 576: question: why do you return after these errors here?  While in `decodeCustomPrimitive` 148 - 230 you break on errors.  I'm wondering why the difference?

pkg/scale/varying_data_type.go
timwu20: done

edwardmack (approved):

EclesioMeloJunior (approved):

noot (approved):
looks good!!

pkg/scale/decode.go
timwu20: Sorry for not responding @edwardmack! It's because of line 632: 
```
dstv.Set(reflect.ValueOf(out))
```
I didn't want to `Set` on the error case, so I had to return.  Conversely, I could've put that line in every case. I dunno which is better.

edwardmack: @timwu20 no problem, I'm just asking to better understand. So, at 229 you do want `dstv.Set(temp.Elem().Convert(inType))` set on error?

timwu20: No I don't.  if `ds.ReadByte` returns an error, I'm just exiting this `decodeFixedWidthInt`.  I had it as `break` before, but then it would panic cause it would hit line 632.

Issue #1468, opened 2021-03-16
Enhance prometheues metrics.


arijitAD:
## Issue summary

- Currently, substrate publishes a lot of metrics.
- We should triage and figure out relevant metrics for us.

### Metrics published by Substrate.
```
# HELP substrate_authority_discovery_amount_external_addresses_last_published Number of external addresses published when authority discovery last published addresses.
# TYPE substrate_authority_discovery_amount_external_addresses_last_published gauge
substrate_authority_discovery_amount_external_addresses_last_published 0
# HELP substrate_authority_discovery_authority_address_requests_pending Number of pending authority address requests.
# TYPE substrate_authority_discovery_authority_address_requests_pending gauge
substrate_authority_discovery_authority_address_requests_pending 0
# HELP substrate_authority_discovery_authority_addresses_requested_total Number of times authority discovery has requested external addresses of a single authority.
# TYPE substrate_authority_discovery_authority_addresses_requested_total counter
substrate_authority_discovery_authority_addresses_requested_total 0
# HELP substrate_authority_discovery_dht_event_received Number of dht events received by authority discovery.
# TYPE substrate_authority_discovery_dht_event_received counter
substrate_authority_discovery_dht_event_received{name="value_put_failed"} 6
# HELP substrate_authority_discovery_handle_value_found_event_failure Number of times handling a dht value found event failed.
# TYPE substrate_authority_discovery_handle_value_found_event_failure counter
substrate_authority_discovery_handle_value_found_event_failure 0
# HELP substrate_authority_discovery_known_authorities_count Number of authorities known by authority discovery.
# TYPE substrate_authority_discovery_known_authorities_count gauge
substrate_authority_discovery_known_authorities_count 0
# HELP substrate_authority_discovery_times_published_total Number of times authority discovery has published external addresses.
# TYPE substrate_authority_discovery_times_published_total counter
substrate_authority_discovery_times_published_total 6
# HELP substrate_block_height Block height info of the chain
# TYPE substrate_block_height gauge
substrate_block_height{status="best"} 599
substrate_block_height{status="finalized"} 597
# HELP substrate_block_verification_and_import_time Time taken to verify and import blocks
# TYPE substrate_block_verification_and_import_time histogram
substrate_block_verification_and_import_time_bucket{le="0.005"} 0
substrate_block_verification_and_import_time_bucket{le="0.01"} 0
substrate_block_verification_and_import_time_bucket{le="0.025"} 0
[...]

PR #1655, opened 2021-06-23
feat(lib/grandpa) implement grandpa finality round metrics

EclesioMeloJunior:
## Changes

- Adds func `collectGranpdaMetrics` to **grandpa** package
- Collects the current `grandpa.state.round` to prometheus metrics

## Tests

```
go test -run ^TestFinalRoundGaugeMetric$ github.com/ChainSafe/gossamer/lib/grandpa
```

## Issues

- Related #1468


lib/transaction/pool.go
```diff
@@ -2,10 +2,15 @@ package transaction
 
 import (
 	"sync"
+	"time"
 
+	"github.com/ChainSafe/gossamer/dot/metrics"
 	"github.com/ChainSafe/gossamer/lib/common"
 )
 
+const collectTxMetricsTimeout = time.Second * 5
```
timwu20, line 11: It would be nice to have this defined in the `NewPool` constructor rather than package level const.  You could even create a private `newPool` that takes in configurable durations for this so you can adjust them in the tests.

EclesioMeloJunior: I've removed this and used the metrics default value

noot (approved):

timwu20 (approved):

PR #1656, opened 2021-06-24
feat(lib/transactions) ready transactions metrics

EclesioMeloJunior:
## Changes

- Implement ready transactions metrics

## Tests

transaction/pool
```
go test -run ^TestPoolCollectMetrics$ github.com/ChainSafe/gossamer/lib/transaction
```

transaction/priority_queue

```
go test -run ^TestPriorityQueue$ github.com/ChainSafe/gossamer/lib/transaction
```

## Issues

- Related #1468


lib/transaction/pool_test.go
```diff
@@ -50,3 +54,51 @@ func TestPool(t *testing.T) {
...
+		h[i] = p.Insert(v)
+	}
+
+	time.Sleep(collectTxMetricsTimeout + time.Second)
+	require.Equal(t, int64(len(validtx)), txmetrics.Value())
+
+	p.Remove(h[0])
+
+	time.Sleep(collectTxMetricsTimeout + time.Second)
+	fmt.Println(len(p.transactions))
```
edwardmack, line 102: maybe remove println

edwardmack (approved):
Great work, please add copyright comments to the top of new files.

lib/transaction/pool_test.go
EclesioMeloJunior: removed!

lib/transaction/pool.go
```diff
@@ -14,9 +19,17 @@ type Pool struct {
 
 // NewPool returns a new empty Pool
 func NewPool() *Pool {
-	return &Pool{
+	p := &Pool{
 		transactions: make(map[common.Hash]*ValidTransaction),
 	}
+
+	go metrics.CollectGaugeMetrics(
```
timwu20, line 26: I think it would be nicer if this was called outside of the `NewPool` constructor, and we could have some polling process to iterate over all the `GaugeCollector` implementations, rather than having multiple goroutines each polling their own gauge.

lib/transaction/priority_queue.go
```diff
@@ -94,6 +97,13 @@ func NewPriorityQueue() *PriorityQueue {
 		pq:  make(priorityQueue, 0),
 		txs: make(map[common.Hash]*Item),
 	}
+
+	go metrics.CollectGaugeMetrics(
```
timwu20, line 101: same thing with this one. May make sense to create a `GaugeHarvester` that takes in a number of `GaugeCollector` implementations.

dot/metrics/collector.go
```diff
@@ -0,0 +1,32 @@
...
+
+import (
+	"time"
+
+	ethmetrics "github.com/ethereum/go-ethereum/metrics"
+)
+
+// GaugeCollector abstracts the Update function to collectors
+// just implement the collection
+type GaugeCollector interface {
```
timwu20, line 11:
```suggestion
type GaugeMetric interface {
type GaugeMetricSource interface {
type GaugeMetricProvider interface {
```
`GaugeCollector` sounds like the one that collects over all these sources/providers/metrics.

timwu20 (changes requested):

dot/metrics/collector.go
```diff
@@ -0,0 +1,32 @@
...
+// and acquire the metrics timeout by timeout to a ethereum metrics gauge
+func CollectGaugeMetrics(timeout time.Duration, label string, c GaugeCollector) {
+	t := time.NewTicker(timeout)
+	defer t.Stop()
+
+	collectGauge(label, c)
+
+	for range t.C {
+		collectGauge(label, c)
+	}
```
noot, line 25: should there be a context here? otherwise I don't see where this is stopped

lib/transaction/pool.go
EclesioMeloJunior: Got it! I've implemented a `Collector`, that acts like a harvester, which has an `AddGauge` and a `Start` methods, then the `AddGauge` will "register" all the `GaugeMetrics` implementations so when `Start` is called (when the node starts) it triggers a goroutine that does polling in all the registereds implementations.

I think with this way we can remove all the spread goroutines and centralize in one in the `metrics` package.

dot/metrics/collector.go
EclesioMeloJunior: added context on all loops that get metrics

dot/metrics/collector.go
```diff
@@ -1,32 +1,131 @@
...
+		ctx:    ctx,
+		gauges: make([]GaugeMetrics, 0),
+	}
 }
 
-// CollectGaugeMetrics receives an timeout, label and a gauge collector
-// and acquire the metrics timeout by timeout to a ethereum metrics gauge
-func CollectGaugeMetrics(timeout time.Duration, label string, c GaugeCollector) {
-	t := time.NewTicker(timeout)
+// Start will start one goroutine to collect all the gauges registereds and
```
timwu20, line 47:
```suggestion
// Start will start one goroutine to collect all the gauges registered and
```

dot/metrics/collector.go
```diff
@@ -1,32 +1,131 @@
...
+// a separate goroutine to collect process metrics
+func (c *Collector) Start() {
```
timwu20, line 49: I like making `Start` functions blocking.  It is the callers responsibility to call `go c.Start()`. You could use a `sync.WaitGroup` to wait for these child goroutines to finish.

EclesioMeloJunior: Done!

timwu20 (approved):

noot (approved):
looks good!

PR #1660, opened 2021-06-24
fix (dot/telemetry): NoTelemetry flag stops telemetry

### LINKED ISSUE #1652: `NoTelemetry` flag doesn't stop telemetry
## Describe the bug

- the `--no-telemetry` flag doesn't stop telemetry
- the only place the `NoTelemetry` flag is checked is `dot/node.go`
```
	if cfg.Global.NoTelemetry {
		return node, nil
	}
```

this prevents `telemetry.GetInstance()` from being called in `dot/node.go`. however in packages that use telemetry, `NoTelemetry` is not persisted and they end up calling `telemetry.GetInstance()` thus creating the telemetry instance, for example `dot/network/service.go` `publishNetworkTelemetry`.

## Possible Solution

- I would recommend adding the `NoTelemetry` option to the `telemetry` package and creating a `Instance` interface that includes the functions the current `*telemetry.Handler` has. If `NoTelemetry == true`, then a `*NoopHandler`, which does nothing, can be created when `GetInstance()` is called. otherwise, the usual `*Handler` can be used.

edwardmack:
## Changes

- Implement No Op telemetry instance that stubs telemetry Instance functions so nothing is done.
- Create `KillInstance` function to use no op telemetry instance (so telemetry message calls are ignored).
-

## Tests

```
go test ./dot/telemetry -v
```

## Issues

- closes #1652


dot/node.go
```diff
@@ -341,7 +341,7 @@ func NewNode(cfg *Config, ks *keystore.GlobalKeystore, stopFunc func()) (*Node,
 	}
 
 	if cfg.Global.NoTelemetry {
-		return node, nil
+		telemetry.GetInstance().KillInstance()
```
noot, line 344: it would be nicer to have `GetInstance()` return a `NoopInstance` instead of having `KillInstance()`, you can add a global package var to the `telemetry` package and use that in `GetInstance()`, eg `telemetry.Enabled = cfg.Global.NoTelemetry`

dot/telemetry/telemetry_test.go
```diff
@@ -139,6 +139,42 @@ func TestListenerConcurrency(t *testing.T) {
...
+	}
+	wg.Wait()
+	counter := 0
+	tk := time.NewTicker(time.Second * 2)
+main:
+	for {
+		select {
+		case <-tk.C:
+			break main
+		case <-resultCh:
```
EclesioMeloJunior, line 169: When this case will be reached? The `resultsCh` is created but none data is pass through channel

dot/node.go
edwardmack: Yes, this is nicer, I've updated.

dot/telemetry/telemetry_test.go
edwardmack: TestMain starts a web server that has a HandleFunc listen which reads websocket messages and passes them to `resultCh` at line 191.  Added a fmt.Println to the select case confirms this.

dot/telemetry/telemetry.go
```diff
@@ -46,6 +46,13 @@ type Handler struct {
 	sendMessageTimeout time.Duration
 }
 
+// Instance functions that telemetry handler instance needs to implement
```
timwu20, line 49:
```suggestion
// Instance interface that telemetry handler instance needs to implement
```

dot/telemetry/telemetry.go
```diff
@@ -46,6 +46,13 @@ type Handler struct {
...
+type Instance interface {
+	AddConnections(conns []*genesis.TelemetryEndpoint)
+	SendMessage(msg *Message) error
+	startListening()
```
timwu20, line 53: why the private method?

dot/telemetry/telemetry.go
```diff
@@ -54,13 +61,16 @@ type KeyValue struct {
 
 var (
 	once            sync.Once
-	handlerInstance *Handler
+	handlerInstance Instance
+
+	// Enabled flag to determine if telemetry is enabled
+	Enabled = true // enabled by default
```
timwu20, line 67: make `Enabled` private.  Shouldn't be able to change this from outside the package.

dot/telemetry/telemetry.go
edwardmack: I'm trying to control who can start the listener, this is called when the instance is created.

dot/telemetry/telemetry.go
edwardmack: I want to disable telemetry from outside the package (i.e. in node.go line 343).  I'm following suggestion by @noot above.

noot: yeah I think this is fine, we could to create some `Initialise` function that accepts the `enabled bool` and is also a singleton and change this package var to internal. what do you think @timwu20

dot/telemetry/telemetry_test.go
```diff
@@ -139,6 +139,42 @@ func TestListenerConcurrency(t *testing.T) {
 	}
 }
 
+func TestKillInstance(t *testing.T) {
```
noot, line 142: rename this?

dot/telemetry/telemetry.go
timwu20: Yup that works for me.

edwardmack: Sounds good, I believe I've implemented it correctly now.

dot/telemetry/telemetry_test.go
edwardmack: Done.

EclesioMeloJunior (approved):

dot/telemetry/telemetry.go
```diff
@@ -65,9 +76,21 @@ func GetInstance() *Handler { //nolint
...
 	}
+	if !enabled {
+		return &NoopHandler{}
+	}
+
 	return handlerInstance
 }
 
+// Initialise function to set if telemetry is enabled
+func (h *Handler) Initialise(_enabled bool) {
```
timwu20, line 87:
```suggestion
func (h *Handler) Initialise(e bool) {
```
nit: we don't use leading underscore in go

timwu20 (approved):

PR #1662, opened 2021-06-29
fix(dot/network, lib/grandpa): fix node sync, improve devnet finality

### LINKED ISSUE #1587: devnet: ensure blocks are finalized

noot:
## Changes

- fix node sync (previously nodes were forking) by changing requests sent from `BlockAnnounceMessage` to by via hash rather than number
- fix block response direction, previously gossamer was responding with blocks in the wrong direction
- rename `SignedPrecommit` -> `SignedVote`
- update grandpa to correctly handle being the "primary" ie. chosen to broadcast initial prevote, and fix handling of receiving primary prevote, was not handling `primaryProposal` vote type before
- simplify vote storage in `*grandpa.Service`, now uses 2 `*sync.Map`s for signed prevotes and signed precommits respectively 
- fix creation of justification for `CommitMessage` in grandpa, previously it was only putting votes that were directly for the finalised block when it needed to also add votes for any ancestors of the finalised block as well
- expose `*network.Service.SendBlockReqestByHash` function which allows grandpa to send a block request for a block with a given hash if it receives a vote for a block it doesn't know about

(will do these in follow-up PR)
TODO: fix `receiveMessages` logic, too many goroutines are being spawned
TODO: if peer sends us a vote with lower round than us, send a commit message from that round *only to them* instead of broadcasting 

## Tests

```
go test ./dot/network -short
go test ./lib/grandpa -short
```

also, can run local "devnet" of 2-3 nodes after updating the gsmmr genesis-spec file to have 3 babe/grandpa authorities, you should see blocks getting finalised around every 2 seconds.

## Issues

- closes #1587


lib/grandpa/grandpa.go
```diff
@@ -448,12 +485,12 @@ func (s *Service) playGrandpaRound() error {
...
 			}
 
-			err = s.sendMessage(pv, prevote)
+			err = s.sendMessage(vm)
 			if err != nil {
 				logger.Error("could not send prevote message", "error", err)
 			}
 
-			time.Sleep(time.Second * 5)
+			time.Sleep(interval * 10)
```
EclesioMeloJunior, line 493: here we could use a timer:
```go
t := time.NewTimer(interval * 10)
defer t.Stop()
for {
  ... // code
  <-t.C
  t.Reset(interval * 10)
  logger.Trace("sent pre-vote message...", "vote", pv, "prevotes", s.prevotes)
}
```

EclesioMeloJunior (changes requested):

lib/grandpa/grandpa.go
```diff
@@ -260,59 +258,74 @@ func (s *Service) publicKeyBytes() ed25519.PublicKeyBytes {
 
 // initiate initates a GRANDPA round
 func (s *Service) initiate() error {
-	// if there is an authority change, execute it
-	err := s.updateAuthorities()
-	if err != nil {
-		return err
-	}
+	initiateRound := func() error {
```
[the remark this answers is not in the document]
noot: update to be a separate function

lib/grandpa/grandpa.go
noot: updated

lib/grandpa/grandpa.go
```diff
@@ -547,62 +545,134 @@ func (s *Service) playGrandpaRound() error {
...
 	err = s.attemptToFinalize()
 	if err != nil {
-		log.Error("failed to finalise", "error", err)
+		logger.Error("failed to finalise", "error", err)
 		return err
 	}
 
-	finalised = true
+	close(roundComplete[0])
+	close(roundComplete[1])
```
EclesioMeloJunior, line 557: as the two channels are updated at the same time we could use just one channel and share it with the `sendVoteMessage` function

EclesioMeloJunior (changes requested):

lib/grandpa/message_handler.go
```diff
@@ -278,17 +266,25 @@ func (h *MessageHandler) verifyCommitMessageJustification(fm *CommitMessage) err
 			continue
 		}
 
-		if just.Vote.hash == fm.Vote.hash && just.Vote.number == fm.Vote.number {
+		//count++
```
arijitAD, line 269: Remove this

lib/grandpa/vote_message.go
```diff
@@ -179,38 +192,32 @@ func (s *Service) validateMessage(m *VoteMessage) (*Vote, error) {
 	}
 
 	err = s.validateVote(vote)
-	if err == ErrBlockDoesNotExist {
+	if err == ErrBlockDoesNotExist || err == blocktree.ErrEndNodeNotFound {
```
arijitAD, line 195: Use `errors.Is`

arijitAD (commented):
Can you rebase the development branch and force push the PR so that the history is clean?

lib/grandpa/grandpa.go
noot: `sendVoteMessage` is called twice, once for prevotes and precommits, so there needs to be 2 channels

lib/grandpa/message_handler.go
noot: done

lib/grandpa/vote_message.go
noot: done

lib/grandpa/grandpa.go
EclesioMeloJunior: Since the condition to stop the `sendVoteMessage` is the `close(channel)` and both channels are being closed at the same time I think it's possible to create just one channel and
```go
ch := make(chan struct{})
```
start the first and second `sendVoteMessage`goroutines with the same channel 
```go
go s.sendVoteMessage(prevote, vm, ch)
...
go s.sendVoteMessage(precommit, pcm, ch)
```
and then close the channel once
```go
close(ch)
```
This way the 2 goroutines will be stoped as before but sharing just one channel

EclesioMeloJunior: Since the condition to stop the sendVoteMessage is the close(channel) and both channels are being closed at the same time I think it's possible to create just one channel and

```go 
ch := make(chan struct{})
```
start the first and second sendVoteMessagegoroutines with the same channel

```go
go s.sendVoteMessage(prevote, vm, ch)
...
go s.sendVoteMessage(precommit, pcm, ch)
```
and then close the channel once
```go
close(ch)
```
This way the 2 goroutines will be stoped as before but sharing just one channel

noot: I've updated this, I was under the impression that closing a channel would only send 1 instance of `struct{}{}` (or whatever the empty type is for the channel) through the channel, but it seems like reading from the closed channel always receives `struct{}{}`

EclesioMeloJunior (approved):
lgtm

edwardmack (approved):
Great work, lgtm!

lib/grandpa/grandpa.go
```diff
@@ -58,20 +59,19 @@ type Service struct {
...
-	pcJustifications map[common.Hash][]*SignedPrecommit // pre-commit justifications for the current round
-	pvEquivocations  map[ed25519.PublicKeyBytes][]*Vote // equivocatory votes for current pre-vote stage
-	pcEquivocations  map[ed25519.PublicKeyBytes][]*Vote // equivocatory votes for current pre-commit stage
-	tracker          *tracker                           // tracker of vote messages we may need in the future
-	head             *types.Header                      // most recently finalised block
+	state      *State    // current state
+	prevotes   *sync.Map // map[ed25519.PublicKeyBytes]*SignedVote // pre-votes for the current round
+	precommits *sync.Map // map[ed25519.PublicKeyBytes]*SignedVote // pre-commits for the current round
+	// TODO: change equivocations to map[ed25519.PublicKeyBytes][]*SignedVote
+	pvEquivocations map[ed25519.PublicKeyBytes][]*SignedVote // equivocatory votes for current pre-vote stage
```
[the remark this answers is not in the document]
noot: yep i removed it!

Issue #1621, opened 2021-06-03
Integrate scale package into internal packages


timwu20:
## Task summary

- Evaluate use of custom `Encode()`/`Decode()` functionality on struct types.  Should all be covered by new `Marshal()`/`Unmarshal()` functionality.  Remove the `Encode()`/`Decode()` receiver functions.
- Change all occurrences of `Encode()`/`Decode()` to use new `Marshal()`/`Unmarshal()` interface.
- Implement `VaryingDataType` pattern for existing structs that decode SCALE/Rust enums
- Implement `Result` pattern for existing use case in `lib/runtime/wasmer/imports.go`
- Ensure that existing encoded data stored in DB is successfully decoded

## Specification reference

-  https://github.com/w3f/polkadot-spec/releases/download/v0.1.0-beta1/polkadot-host-spec_v0.1.0-beta1.pdf

## Other information and links

- https://github.com/ChainSafe/gossamer/tree/tim/scale/pkg/scale
- https://github.com/paritytech/parity-scale-codec
- https://substrate.dev/docs/en/knowledgebase/advanced/codec

jimjbrettj:
- [ ] lib/babe
- [ ] lib/common
- [ ] lib/genesis
- [ ] lib/grandpa
- [ ] lib/runtime
- [ ] lib/transaction
- [ ] lib/trie
- [ ] dot/core
- [ ] dot/digest WIP
- [ ] dot/network WIP
- [ ] dot/rpc
- [ ] dot/state
- [ ] dot/sync
- [ ] dot/types WIP

noot:
closing as this has been completed

PR #1663, opened 2021-06-29
chore (pkg/scale): Integrate scale into lib/genesis

jimjbrettj:
## Changes

- Integrate new scale pkg into lib/genesis
- Replace AccountInfo type and integrate scale pkg whereever it is used
- Updates to encoding for Uint128

## Tests

```
go test ./lib/genesis -v

go test ./lib/runtime/wasmer -v

go test ./dot/rpc/modules -v
```

## Issues

- #1621


dot/types/account.go
```diff
@@ -13,9 +13,9 @@ type AccountInfo struct {
...
 	// The additional data that belongs to this account. Used to store the balance(s) in a lot of chains.
 	Data struct {
-		Free       common.Uint128
-		Reserved   common.Uint128
-		MiscFrozen common.Uint128
-		FreeFrozen common.Uint128
+		Free       *scale.Uint128
+		Reserved   *scale.Uint128
+		MiscFrozen *scale.Uint128
+		FreeFrozen *scale.Uint128
```
noot, line 19: are pointers optionals in the new scale? should this be `scale.Uint128` or does scale handle this as a non-option?

lib/runtime/wasmer/exports_test.go
```diff
@@ -394,16 +393,17 @@ func buildBlock(t *testing.T, instance runtime.Instance) *types.Block {
...
-		in, err := scale.Encode(ext) //nolint
-		require.NoError(t, err)
+	for _, ext := range exts {
+		in, er := scale.Marshal(ext)
+		require.NoError(t, er)
 
-		ret, err := instance.ApplyExtrinsic(append([]byte{1}, in...))
-		require.NoError(t, err, in)
+		ret, e := instance.ApplyExtrinsic(append([]byte{1}, in...))
+		require.NoError(t, e, in)
```
noot, line 406: nit but i think errors in go should always be named `err`, so you can add a `//nolint` if that causes lint issues

pkg/scale/comparison_test.go
```diff
@@ -16,160 +16,161 @@
...
-		for _, tt := range allTests {
-			// dst := reflect.New(reflect.TypeOf(tt.in)).Elem().Interface()
-			if _, err := Marshal(tt.in); (err != nil) != tt.wantErr {
-				b.Errorf("decodeState.unmarshal() error = %v, wantErr %v", err, tt.wantErr)
-				return
-			}
-		}
-	}
-}
+//
```
noot, line 19: why is this all commented out?

pkg/scale/encode.go
```diff
@@ -351,6 +351,10 @@ func (es *encodeState) encodeUint(i uint) (err error) {
 
 // encodeUint128 encodes a Uint128
 func (es *encodeState) encodeUint128(i *Uint128) (err error) {
-	err = binary.Write(es, binary.LittleEndian, i.Bytes())
+	if i == nil {
+		err = fmt.Errorf("uint128 is nil: %v", i)
+		return
+	}
```
noot, line 357: thought this would be handled higher up - @timwu20 is this needed?

jimjbrettj: It might not be. I added because when I was initially encoding with the old AccountInfo type there were cases where the fields weren't initialized and it was crashing. Shouldn't be an issue now since I changed that but I kept this anyways. Can remove if not necessary

pkg/scale/comparison_test.go
jimjbrettj: Was causing a circular dependency issue so I commented out for now and Tim said he would need to revise the package for this test

dot/types/account.go
timwu20: `**scale.Uint128` is optional.  Same for `**big.Int`.

pkg/scale/encode.go
timwu20: I don't think this should be necessary.  In most use cases when encoding, you would've have initialised a value for that pointer, instead of a zero case pointer.  Can you point me to the code that causes this panic?

jimjbrettj: Yeah dot/rpc/modules/system_test.go where the AccountInfo type was used the struct fields weren't initiated and thats what caused my initial issue. This pr changed it so now they are initialized now

lib/genesis/helpers.go
```diff
@@ -357,34 +358,40 @@ func addAuthoritiesValues(k1, k2 string, kt crypto.KeyType, value []byte, gen *G
...
 		auth := []interface{}{}
 		buf := make([]byte, 32)
-		if _, err = sd.Reader.Read(buf); err == nil {
+		if _, err = reader.Read(buf); err == nil {
 			var arr = [32]byte{}
 			copy(arr[:], buf)
-			pa, err := bytesToAddress(kt, arr[:])
-			if err != nil {
-				return err
+			pa, e := bytesToAddress(kt, arr[:])
```
EclesioMeloJunior, line 378: I think we could keep `err` as variable name here

jimjbrettj: agreed. I was getting linter issues which is why I changed it but im gonna go fix those to be consistent

chain/dev/genesis.json
```diff
@@ -3,7 +3,12 @@
     "id": "dev",
     "chainType": "Local",
     "bootNodes": [],
-    "telemetryEndpoints": null,
+    "telemetryEndpoints": [
+    [
+      "wss://telemetry.polkadot.io/submit/",
+      0
+    ]
+  ],
```
noot, line 11: don't think this is needed, can remove

chain/gssmr/genesis.json
```diff
@@ -3,6 +3,12 @@
     "id": "gssmr",
     "chainType": "Local",
     "bootNodes": [],
+    "telemetryEndpoints": [
+    [
+      "wss://telemetry.polkadot.io/submit/",
+      0
+    ]
+  ],
```
noot, line 11: same here

noot (commented):
not sure what happened but looks like there's a bunch of extra changes here, can you merge/rebase off development again? looks like something went wrong when merging

jimjbrettj:
> 
> 
> not sure what happened but looks like there's a bunch of extra changes here, can you merge/rebase off development again? looks like something went wrong when merging

PR #1669, opened 2021-07-01
chore (pkg/scale) scale integration into genesis

jimjbrettj:
## Changes

- Integrate new scale pkg into lib/genesis
- Replace AccountInfo type and integrate scale pkg whereever it is used
- Updates to encoding for Uint128

## Tests

```
go test ./lib/genesis -v

go test ./lib/runtime/wasmer -v

go test ./dot/rpc/modules -v
```

## Issues

- #1621


pkg/scale/comparison_test.go
```diff
@@ -16,160 +16,161 @@
...
-		for _, tt := range allTests {
-			// dst := reflect.New(reflect.TypeOf(tt.in)).Elem().Interface()
-			if _, err := Marshal(tt.in); (err != nil) != tt.wantErr {
-				b.Errorf("decodeState.unmarshal() error = %v, wantErr %v", err, tt.wantErr)
-				return
-			}
-		}
-	}
-}
+//
```
noot, line 19: why is this commented out?

noot (approved):
looks good, please uncomment commented out tests?

pkg/scale/comparison_test.go
jimjbrettj: @timwu20 Do you have suggestions on how to remove circular dependency this creates?

EclesioMeloJunior (approved):

lib/genesis/helpers.go
```diff
@@ -212,7 +213,7 @@ func buildRawArrayInterface(a []interface{}, kv *keyValue) {
 			kv.iVal = append(kv.iVal, tba)
 		case float64:
 			// TODO: determine how to handle this error
-			encVal, _ := scale.Encode(uint64(v2))
+			encVal, _ := scale.Marshal(uint64(v2))
```
timwu20, line 216: really should handle this error

pkg/scale/comparison_test.go
timwu20: You can remove this file for now

timwu20 (dismissed):

lib/genesis/helpers.go
jimjbrettj: @timwu20 I handled it with a panic, but can refactor to return an error if that is prefered

timwu20: That is preferred.  We should never panic in the codebase.

PR #1670, opened 2021-07-01
chore (pkg/scale) scale babe integration

jimjbrettj:
## Changes

- Integrated new scale pkg into lib/transactions
- Integrated scale into babe
- Refactored babe to utilize new scale features.

## Tests

```
go test ./lib/babe -v
go test ./lib/transaction -v

Note: There are errors thrown while testing with regards to block authoring (I don't think related to me) but wanted to check. Tests were still passing but wanted to get that checked out anyways.
```

## Issues

-  #1621


lib/babe/errors.go
```diff
@@ -81,88 +80,220 @@ func (e TransactionValidityError) Error() string {
...
-	}
-	errMsg, err := optional.NewBytes(false, nil).DecodeBytes(res[2:])
-	if err != nil {
-		return err
-	}
-	return fmt.Errorf("index: %d code: %d message: %s", res[0], res[1], errMsg.String())
+// A UnmarshalError is when unmarshalling fails
+type UnmarshalError struct {
+	msg string
 }
```
noot, line 86:
```suggestion
// UnmarshalError occurs when unmarshalling fails
type UnmarshalError struct {
	msg string
}
```

lib/babe/errors.go
```diff
@@ -81,88 +80,220 @@ func (e TransactionValidityError) Error() string {
...
+	var unknownCustom UnknownCustom
+
+	dispatchError := scale.MustNewVaryingDataType(other, CannotLookup{}, BadOrigin{}, Module{})
+	invalid := scale.MustNewVaryingDataType(Call{}, Payment{}, Future{}, Stale{}, BadProof{}, AncientBirthBlock{},
+		ExhaustsResources{}, invalidCustom, BadMandatory{}, MandatoryDispatch{})
+	unknown := scale.MustNewVaryingDataType(ValidityCannotLookup{}, NoUnsignedValidator{}, unknownCustom)
+
+	okRes := scale.NewResult(nil, dispatchError)
+	errRes := scale.NewResult(invalid, unknown)
+	result := scale.NewResult(okRes, errRes)
```
noot, line 255: I think these could be declared as package vars since they wouldn't ever change?

noot (approved):
nice work :D just a few smaller comments

lib/babe/errors.go
jimjbrettj: Yup you are right, will do

EclesioMeloJunior (approved):

timwu20 (approved):
well done figuring out that giant nested data structure.

PR #1671, opened 2021-07-02
fix(dot/node) Start websocket server only with `--ws` flag

### LINKED ISSUE #1654: websockets server doesn't start with only `--ws` flag
## Describe the bug

- the websockets server only starts if `--rpc --ws` are provided, but not only `--ws`
- specifying only `--ws` should start the websockets server

EclesioMeloJunior:
## Changes

- check the value of `cfg.RPC.WS` before start the server

## Tests

```
N/A
```

## Issues
- closes #1654

## Screenshots
image


noot (approved):

dot/node.go
```diff
@@ -308,7 +308,7 @@ func NewNode(cfg *Config, ks *keystore.GlobalKeystore, stopFunc func()) (*Node,
 	nodeSrvcs = append(nodeSrvcs, sysSrvc)
 
 	// check if rpc service is enabled
-	if enabled := cfg.RPC.Enabled; enabled {
+	if enabled := (cfg.RPC.Enabled || cfg.RPC.WS); enabled {
```
arijitAD, line 311: Parenthesis is not required.
Not idiomatic to use parenthesis in if-else in Go.

arijitAD (approved):

dot/node.go
EclesioMeloJunior: removed!

edwardmack (approved):

Issue #592, opened 2020-02-05
Gossamer Devnet 


amerameen:
Gossamer devnet test cases outlined below:

| Test Case                  | Status      | What                                                                  | How                                                                                                    | Expected behaviour                                                                             |
|----------------------------|-------------|-----------------------------------------------------------------------|--------------------------------------------------------------------------------------------------------|------------------------------------------------------------------------------------------------|
| Authority Nodes            | Ongoing     | Establish specific authority nodes and non-authority nodes            | 6 authority nodes configured, and check if they are rotating correctly, and producing/authoring blocks | Blocks produced and finalized over a long period of time                                       |
| Shut down authority nodes  | Ongoing     | Auth node finality                                                    | Shut down auth nodes in varying supply                                                                 | 6 auth and 1 goes down should continue, but if more than 2/3 rds go down finality should cease |
| Adding new authority nodes | Ongoing     | Ability to test new auth nodes with an ongoing chain                  | Add auth node after x blocks                                                                           | Added to auth node group, produce/author blocks                                                |
| Runtime upgrades           | Ongoing     | Ability to do runtime upgrades with gossamer nodes                    | Submit extrinsic indicating a runtime upgrade                                                          | Nodes runtime should be upgraded                                                               |
| Syncing                    | Not started | Ensure non-authority nodes are able to sync properly                  | Adding non-authority nodes at various block heights                                                    | Non-authority nodes will be able to sync properly                                              |
| Tooling checks             | Not started | Ensure rpc calls are available and accurate, and telemetry connection | Submit various rpc calls and ensure telemetry connection is correct                                    | Calls should go through and be accurate                                                        |
| Key management             | Not started | Ensure auth node key work correctly for staking and signing           | Utilize keys to stake and sign blocks                                                                  | Should not result in an error and correctly stakes / sign the block                            |
| Metrics and monitoring     | Not started | Monitor network traffic (gossip), and metrics gathering               | Configure grafana to capture metrics                                                                   |                                                                                                |

noot:
duplicate of #1356

P1sar:
Stale

PR #1675, opened 2021-07-05
fix(lib/scale): Handle eof in scale decoding.

arijitAD:
## Changes

- Node crashed during adding authority nodes.
- Handle eof in scale decoding.

## Tests

```
TestDecodeNilPublicKey
TestDecodeNilVote
```

## Issues

- #592


lib/crypto/ed25519/ed25519.go
```diff
@@ -71,6 +71,9 @@ func (b PublicKeyBytes) Encode() ([]byte, error) {
 // Decode returns the SCALE decoded PublicKeyBytes
 func (b PublicKeyBytes) Decode(r io.Reader) ([PublicKeyLength]byte, error) {
 	_, err := r.Read(b[:])
+	if errors.Is(err, io.EOF) {
+		return b, nil
+	}
```
noot, line 76: this seems wrong - if there's an EOF it's because there weren't enough bytes in the reader to create the public key, so there should be an error or else this will create an invalid public key

noot (commented):
I think this PR will likely fix any EOF issues you encountered https://github.com/ChainSafe/gossamer/pull/1662

we shouldn't be ignoring the EOFs as it will result in invalid votes/public keys

arijitAD:
> we shouldn't be ignoring the EOFs as it will result in invalid votes/public keys

Yes, I agree that we should return the error. But the node was panicking due to incorrect data from another peer. We should handle these errors gracefully in such cases.

noot:
@arijitAD please update this to handle the error instead of ignoring, then should be good

arijitAD:
No longer seeing this issue.

PR #1679, opened 2021-07-06
lib/babe: always use 2/3 of slot to produce block, re-add potentially valid txs to queue

### LINKED ISSUE #1628: Fix block production timing logic

noot:
## Changes

- always use 2/3 of slot to produce block even if tx queue becomes empty, since new txs might come in
- re-add potentially valid txs to queue (ie. txs with a nonce higher than expected or if the block gas limit is reached)

## Tests

```
go test ./lib/babe
```

## Issues

- closes #1628
- closes 1629


timwu20:
since @jimjbrettj PR touching these files already went in, does this code need to be updated?

noot:
@timwu20 just updated, should be fine now

EclesioMeloJunior (approved):

lib/babe/errors.go
```diff
@@ -86,13 +86,33 @@ func (e DispatchOutcomeError) Error() string {
...
 type TransactionValidityError struct {
-	msg string // description of error
+	msg transactionValidityErrorMsg // description of error
 }
 
 func (e TransactionValidityError) Error() string {
 	return fmt.Sprintf("transaction validity error: %s", e.msg)
 }
 
+type transactionValidityErrorMsg string
```
EclesioMeloJunior, line 96: why this type is necessary?

EclesioMeloJunior (changes requested):

lib/babe/errors.go
noot: so that not any string can be used as the error message, only the predefined ones. I can remove it if you think it's overkill

EclesioMeloJunior: Got it! IMO seems overkill, I think that each of the errors string var could be of the type `error` like that: 
```go
var (
  ErrunexpectedTxCallMsg  = errors.New("call of the transaction is not expected")
  ErrinvalidPaymentMsg  = errors.New("invalid payment")
  ...
)
```
then, will be easy to determine a if an error occurs:
```go
err := determineErrType(...)
if errors.Is(err, ErrunexpectedTxCallMsg) { ... }
```

lib/babe/build.go
```diff
@@ -270,6 +265,18 @@ func (b *BlockBuilder) buildBlockExtrinsics(slot Slot) []*transaction.ValidTrans
 			if _, ok := err.(*DispatchOutcomeError); !ok {
 				continue
 			}
+
+			// don't drop transactions that may be valid in a later block ie.
+			// run out of gas for this block or have a nonce that may be valid in a later block
+			if tvErr, ok := err.(*TransactionValidityError); ok {
```
arijitAD, line 271: Use `errors.As(err, &e)` https://blog.golang.org/go1.13-errors

arijitAD: You can remove indentation
```
	tvErr, ok := err.(*TransactionValidityError)
	if !ok {
		continue
	}
	switch tvErr.msg {
	case exhaustsResourcesMsg, invalidTransactionMsg:
		hash, err := b.transactionState.Push(txn)
		if err != nil {
			logger.Debug("failed to re-add transaction to queue", "tx", hash, "error", err)
		}
	}
```

arijitAD (approved):

lib/babe/build.go
noot: done

lib/babe/errors.go
noot: updated the errors!

lib/babe/build.go
```diff
@@ -270,6 +265,21 @@ func (b *BlockBuilder) buildBlockExtrinsics(slot Slot) []*transaction.ValidTrans
...
 			}
+
+			// don't drop transactions that may be valid in a later block ie.
+			// run out of gas for this block or have a nonce that may be valid in a later block
+			var e *TransactionValidityError
+			if !errors.As(err, &e) {
+				continue
+			}
+
+			err = err.(*TransactionValidityError).msg
```
EclesioMeloJunior, line 276: This cast is not necessary since we already change the `err` into `TransactionValidityError` on line `273`

lib/babe/build.go
```diff
@@ -270,6 +265,21 @@ func (b *BlockBuilder) buildBlockExtrinsics(slot Slot) []*transaction.ValidTrans
...
+			if errors.Is(err, errExhaustsResources) || errors.Is(err, errInvalidTransaction) {
```
EclesioMeloJunior, line 277:
```suggestion
			if errors.Is(e, errExhaustsResources) || errors.Is(e, errInvalidTransaction) {
```

EclesioMeloJunior (dismissed):

lib/babe/build.go
noot: it doesn't, the above lines do this:
```
			var e *TransactionValidityError
			if !errors.As(err, &e) {
				continue
			}
```

that doesn't cast it as a `*TransactionValidityError` afaik

lib/babe/build.go
noot: not sure why this is necessary?

lib/babe/build.go
EclesioMeloJunior: Sorry, the `errors.As` doesn't cast the variable `err`,  it gets the error and transform it to type `*TransactionValidityError` and sets to variable `&e`, without change the `err` variable, and returns true (or false if fails the type doesn't match). So basically it is possible to use `e.msg` without the cast on line `276`

from: https://golang.org/pkg/errors/#As
> ### func As(err error, target interface{}) bool
> **errors.As** finds the first error in err's chain that matches target, and if so, sets target to that error value and returns true. Otherwise, it returns false.

lib/babe/build.go
EclesioMeloJunior: it just changes to use the `e.msg` instead of `err` since `e` variable has the correct `msg` error

lib/babe/build.go
noot: okay makes sense, I updated it!

lib/babe/build.go
noot: updated

timwu20 (approved):

Issue #1680, opened 2021-07-06
devnet: ensure blocks close to chain head are finalized


noot:
## Task summary

- it seems that after some time, grandpa seems to continuously finalize the same block in each round, even when new blocks are produced and added to the chain
- usually happens at block 65 for some reason
- investigate this and ensure that if a higher blcok exists on each node's chain that block gets finalized

related to #1587

PR #1682, opened 2021-07-07
fix(dot/state, lib/grandpa): 

noot:
## Changes

- add getters and setters for prevotes and precommits for each round/setID to `state.GrandpaState`
- store prevotes, precommits, justification for each round separately in the db (previously was storing it all under justification, but that was incorrect)
- remove `grandpa.Service.justification` map, just load from db when justification is needed
- get latest round from db on restart and use when initiating grandpa

## Tests

```
go test ./dot/state
go test ./lib/grandpa -short
```

## Issues

- related to #1680

### DIFF
--- a/dot/state/block_finalisation.go
+++ b/dot/state/block_finalisation.go
@@ -1,0 +1,166 @@
+package state
+
+import (
+	"encoding/binary"
+	"fmt"
+	"math/big"
+
+	"github.com/ChainSafe/gossamer/dot/types"
+	"github.com/ChainSafe/gossamer/lib/common"
+)
+
+var (
+	prevotesPrefix   = []byte("pv")
+	precommitsPrefix = []byte("pc")
+)
+
+func prevotesKey(round, setID uint64) []byte {
+	k := roundSetIDKey(round, setID)
+	return append(prevotesPrefix, k...)
+}
+
+func precommitsKey(round, setID uint64) []byte {
+	k := roundSetIDKey(round, setID)
+	return append(precommitsPrefix, k...)
+}
+
+// finalizedHashKey = FinalizedBlockHashKey + round + setID (LE encoded)
+func finalizedHashKey(round, setID uint64) []byte {
+	return append(common.FinalizedBlockHashKey, roundSetIDKey(round, setID)...)
+}
+
+func roundSetIDKey(round, setID uint64) []byte {
+	buf := make([]byte, 8)
+	binary.LittleEndian.PutUint64(buf, round)
+	buf2 := make([]byte, 8)
+	binary.LittleEndian.PutUint64(buf2, setID)
+	return append(buf, buf2...)
+}
+
+// HasPrevotes returns if the db contains prevotes for the given round and set ID
+func (bs *BlockState) HasPrevotes(round, setID uint64) (bool, error) {
+	return bs.db.Has(prevotesKey(round, setID))
+}
+
+// SetPrevotes sets the prevotes for a specific round and set ID in the database
+func (bs *BlockState) SetPrevotes(round, setID uint64, data []byte) error {
+	return bs.db.Put(prevotesKey(round, setID), data)
+}
+
+// GetPrevotes retrieves the prevotes for a specific round and set ID from the database
+func (bs *BlockState) GetPrevotes(round, setID uint64) ([]byte, error) {
+	return bs.db.Get(prevotesKey(round, setID))
+}
+
+// HasPrecommits returns if the db contains precommits for the given round and set ID
+func (bs *BlockState) HasPrecommits(round, setID uint64) (bool, error) {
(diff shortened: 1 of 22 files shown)


dot/state/block_finalisation.go
```diff
@@ -0,0 +1,112 @@
...
+	// slot number of block 1, which is now being set as final
+	if bs.lastFinalised.Equal(bs.genesisHash) && !hash.Equal(bs.genesisHash) {
+		err := bs.setFirstSlotOnFinalisation()
+		if err != nil {
+			return err
+		}
+	}
+
+	if round > 0 {
+		go bs.notifyFinalized(hash, round, setID)
```
timwu20, line 77: does this need to be in a goroutine? I read the code for `notifyFinalized` and it doesn't look like it's blocking.

dot/state/block_finalisation.go
```diff
@@ -0,0 +1,112 @@
...
+			return err
+		}
+
+		err = bs.DeleteBlock(rem)
+		if err != nil {
+			return err
+		}
+
+		logger.Trace("pruned block", "hash", rem, "number", header.Number)
+		bs.pruneKeyCh <- header
```
timwu20, line 93: this could block if `storageState` is lagging behind or not running at all.  Do we need a timeout just in case?

dot/state/grandpa.go
```diff
@@ -264,3 +284,76 @@ func (s *GrandpaState) GetNextResume() (*big.Int, error) {
 
 	return big.NewInt(0).SetBytes(num), nil
 }
+
+var (
+	prevotesPrefix   = []byte("pv")
```
timwu20, line 289: const as string?

timwu20: or just define them in `prevotesKey` and `precommitsKey`

dot/types/grandpa.go
```diff
@@ -172,3 +173,83 @@ type FinalisationInfo struct {
...
+		return nil, err
+	}
+
+	v.Number, err = common.ReadUint32(r)
+	if err != nil {
+		return nil, err
+	}
+
+	return v, nil
+}
```
timwu20, line 255: hopefully the new package will take care of a lot of this. @jimjbrettj

dot/state/block_finalisation.go
```diff
@@ -0,0 +1,112 @@
...
+	return append(common.FinalizedBlockHashKey, roundSetIDKey(round, setID)...)
+}
+
+// HasFinalizedBlock returns true if there is a finalised block for a given round and setID, false otherwise
+func (bs *BlockState) HasFinalizedBlock(round, setID uint64) (bool, error) {
+	return bs.db.Has(finalizedHashKey(round, setID))
+}
+
+// NumberIsFinalised checks if a block number is finalised or not
+func (bs *BlockState) NumberIsFinalised(num *big.Int) (bool, error) {
```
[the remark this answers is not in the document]
noot: substrate uses `finalized` but we changed our linter a while back to british english so now our codebase is `finalised`....

dot/state/block_finalisation.go
noot: updated, don't think it needs to be either

dot/state/block_finalisation.go
noot: the storage state just deletes a key from the map on prune, so generally I don't think it would lag. how would you implement a timeout for this?

dot/state/grandpa.go
noot: updated

dot/state/block_finalisation.go
timwu20:
```
select {
   case <-bs.pruneKeyCh:
   case <-time.After(5 * time.Second):
}
```
May be a bit overkill for this case, but that's how I would do it.

noot: the code is writing to the channel, and your code is reading from it? is there a way to do a timeout when writing? either way I don't think it's too necessary for this case

dot/state/block_finalisation.go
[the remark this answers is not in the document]
timwu20: should we make the code consistent then?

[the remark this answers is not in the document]
noot: what do you mean? it should use `finalised` in our codebase

edwardmack (approved):

EclesioMeloJunior (approved):

noot:
@edwardmack everywhere should have already been changed from `finalised` to `finalized` or else the linter would have complained

dot/state/block_finalisation.go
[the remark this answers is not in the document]
timwu20: the who functions above use `Finalized`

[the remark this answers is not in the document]
noot: updated

PR #1684, opened 2021-07-08
chore(lib/grandpa, dot/network): send `CommitMessage` directly to peer on round mismatch; cleanup grandpa `receiveMessages` 

noot:
## Changes

- rename `network.Service.SendMessage` -> `GossipMessage`
- create `network.Service.SendMessage` to send message to specific peer
- update grandpa vote message validation function to send `CommitMessage` (if peer's round is lower than ours) directly to peer instead of gossiping
- update grandpa `receiveMessages` function, now just start goroutine at start of round and cancel goroutine when round ends. previously was spawning multiple `receiveMessages` goroutines that were not getting shut down properly
- update genesis runtime to have c=(1,2) instead of c=(1,4) to increase block production frequency

## Tests

```
go test ./lib/grandpa -short
go test ./dot/network -short
```

## Issues

- related to #592


timwu20 (approved):

lib/grandpa/vote_message.go
```diff
@@ -20,58 +20,49 @@ import (
...
-					"round", vm.Round,
-					"subround", vm.Message.Stage,
-					"prevote count", s.lenVotes(prevote),
-					"precommit count", s.lenVotes(precommit),
-					"votes needed", s.state.threshold(),
-				)
-			case <-ctx.Done():
-				logger.Trace("returning from receiveMessages")
-				return
+func (s *Service) receiveMessages(ctx context.Context /*cond func() bool*/) {
```
EclesioMeloJunior, line 37: Could remove the commented code?

jimjbrettj (approved):

lib/grandpa/vote_message.go
```diff
@@ -20,58 +20,49 @@ import (
...
+	for {
+		select {
+		case msg := <-s.in:
```
EclesioMeloJunior, line 40: Could add a check to the closed channel? 
```go
case msg, ok := <-s.in:
  if !ok { ... }
```

EclesioMeloJunior (changes requested):

lib/grandpa/vote_message.go
noot: done!

lib/grandpa/vote_message.go
noot: done!

EclesioMeloJunior (approved):

PR #1685, opened 2021-07-08
chore (pkg/scale) integrate scale into dot/network and dot/digest

jimjbrettj:
## Changes

- Integrate scale into network and digest packages
- integrate scale into network digest type

## Tests

```
go test ./dot/network -short -v
go test ./dot/digest -v
go test ./dot/types -v
```

## Issues

- #1621


jimjbrettj:
Still trying to get docker stress test to pass, but besides that should be ready for review

dot/network/light.go
```diff
@@ -40,22 +40,17 @@ func (l *LightRequest) SubProtocol() string {
 }
 
 // Encode encodes a LightRequest message using SCALE and appends the type byte to the start
+// TODO Remove encode/decode receiver functions
```
noot, line 43: woudl this require updating the `Message` interface?

dot/network/message_test.go
```diff
@@ -207,14 +210,15 @@ func TestEncodeBlockResponseMessage_WithAll(t *testing.T) {
...
+	//  SCALE option value: 0x01
 	//  ParentHash: Hash: 0x4545454545454545454545454545454545454545454545454545454545454545
 	//	Number: *big.Int // block number: 1
 	//	StateRoot:  Hash: 0xb3266de137d20a5d0ff3a6401eb57127525fd9b2693701f0bf5a8a853fa3ebe0
 	//	ExtrinsicsRoot: Hash: 0x03170a2e7597b7b7e3d84c05391d139a62b157e78786d8c082f29dcf4c111314
 	//	Digest: []byte
 
 	//                                    mtparenthash                                                      bnstateroot                                                       extrinsicsroot                                                di
-	expected, err := common.HexToBytes("0x454545454545454545454545454545454545454545454545454545454545454504b3266de137d20a5d0ff3a6401eb57127525fd9b2693701f0bf5a8a853fa3ebe003170a2e7597b7b7e3d84c05391d139a62b157e78786d8c082f29dcf4c1113140000")
+	expected, err := common.HexToBytes("0x01454545454545454545454545454545454545454545454545454545454545454504b3266de137d20a5d0ff3a6401eb57127525fd9b2693701f0bf5a8a853fa3ebe003170a2e7597b7b7e3d84c05391d139a62b157e78786d8c082f29dcf4c1113140000")
```
noot, line 221: why is this changed? it should be the same as before (in fact it must be, or interop will break)

noot (changes requested):
the encoding of the `BlockAnnounceMessage` or any message type shouldn't change, can you update it to be the same as before? likely why the stress tests are failing also

dot/network/light.go
jimjbrettj: Yes I believe so, I don't have to change it if I shouldn't. Was just thinking less receivers the better

dot/network/message_test.go
jimjbrettj: I believe it is because it was encoded as an optional so had a leading byte appended. Will look into working around this

dot/network/block_announce_test.go
```diff
@@ -35,19 +35,22 @@ func TestBlockAnnounce_Encode(t *testing.T) {
 		Number:         big.NewInt(77),
 		StateRoot:      common.Hash{2},
 		ExtrinsicsRoot: common.Hash{3},
-		Digest:         types.Digest{},
+		Digest:         types.Digest(nil),
```
jimjbrettj, line 38: @noot Will these changes from {} to nil on the blockAnnounceMessage.Digest also need to be reverted?

Issue #1474, opened 2021-03-17
state_subscribeRuntimeVersion does not notify when version changes


edwardmack:
## Issue summary

- Currently RPC call state_subscribeRuntimeVersion only replies with current version.  Need to add logic so that it also notifies when runtime version is updated.

## Other information and links

-

PR #1686, opened 2021-07-08
fix(rpc/subscription): subscribe runtime version notify when version changes

edwardmack:
## Changes

- implemented code that notifies subscribed listener when runtime version changes.

## Tests

```
go test ./dot/rpc/subscription/... -run ^TestRuntimeChannelListener_Listen$
```

## Issues

- closes: #1474


dot/core/service.go
```diff
@@ -579,3 +609,33 @@ func (s *Service) GetMetadata(bhash *common.Hash) ([]byte, error) {
...
+		if s.runtimeChanged[id] == nil {
+			break
+		}
+	}
+
+	s.runtimeChangedLock.RUnlock()
+
+	s.runtimeChangedLock.Lock()
+	s.runtimeChanged[id] = ch
+	s.runtimeChangedLock.Unlock()
```
EclesioMeloJunior, line 633: why are needed these two locks? Just `s.runtimeChangedLock.Lock()` could not be enough?

dot/core/service.go
```diff
@@ -579,3 +609,33 @@ func (s *Service) GetMetadata(bhash *common.Hash) ([]byte, error) {
 	s.rt.SetContextStorage(ts)
 	return s.rt.Metadata()
 }
+
+// RegisterRuntimeUpdatedChannel function to register chan that is notified when runtime version changes
+func (s *Service) RegisterRuntimeUpdatedChannel(ch chan<- runtime.Version) (byte, error) {
+	s.runtimeChangedLock.RLock()
+
+	if len(s.runtimeChanged) == 256 {
+		return 0, errors.New("channel limit reached")
```
EclesioMeloJunior, line 618: is necessary to unlock here

EclesioMeloJunior (changes requested):

dot/core/service.go
```diff
@@ -579,3 +609,33 @@ func (s *Service) GetMetadata(bhash *common.Hash) ([]byte, error) {
 	s.rt.SetContextStorage(ts)
 	return s.rt.Metadata()
 }
+
+// RegisterRuntimeUpdatedChannel function to register chan that is notified when runtime version changes
+func (s *Service) RegisterRuntimeUpdatedChannel(ch chan<- runtime.Version) (byte, error) {
```
EclesioMeloJunior, line 614: Would be nice if we have an Unregister method that closes the channel and remove it from the map

EclesioMeloJunior (changes requested):

dot/core/service.go
```diff
@@ -69,6 +71,9 @@ type Service struct {
 
 	// Keystore
 	keys *keystore.GlobalKeystore
+
+	runtimeChangedLock sync.RWMutex
+	runtimeChanged     map[byte]chan<- runtime.Version
```
noot, line 76: rename `runtimeUpdateSubscriptions` or some thing like that?

dot/core/service.go
```diff
@@ -308,6 +314,11 @@ func (s *Service) handleRuntimeChanges(newState *rtstorage.TrieState) error {
 		return err
 	}
 
+	ver, err := s.rt.Version()
+	if err == nil {
+		go s.notifyRuntimeUpdated(ver)
+	}
```
noot, line 320: if this errors that is bad, so should handle that error

noot (commented):
mostly looks good, but will conflict with @arijitAD 's PR #1638 . you two will have to coordinate on which PR will go in first, because the second one will need to update/move things around

dot/core/service.go
```diff
@@ -462,6 +473,25 @@ func (s *Service) handleChainReorg(prev, curr common.Hash) error {
...
+
+	if len(s.runtimeChanged) == 0 {
+		return
+	}
+
+	logger.Info("notifying runtime updated chans...", "chans", s.runtimeChanged)
+	for _, ch := range s.runtimeChanged {
+		go func(ch chan<- runtime.Version) {
+			select {
+			case ch <- version:
```
timwu20, line 488: why is the channel being written to as a select case?

dot/core/service.go
edwardmack: Right, I've updated to use just one lock (and unlock with defer).

dot/core/service.go
edwardmack: added defer unlock to handle this.

dot/rpc/subscription/websocket.go
```diff
@@ -367,6 +376,19 @@ func (c *WSConn) initRuntimeVersionListener(reqID float64, _ interface{}) (Liste
 	return rvl, nil
 }
 
+func (c *WSConn) unsubscribeRuntimeVersionListener(reqID float64, l Listener, _ interface{}) {
```
EclesioMeloJunior, line 379: You need to register this unsubscribe function to a webscoket method here [./dot/rpc/subscription/subscription.go](https://github.com/ChainSafe/gossamer/blob/bb8021bd45b472f511cf805df2fba2d81ad0293b/dot/rpc/subscription/subscription.go#L45)

EclesioMeloJunior (changes requested):

dot/core/service.go
edwardmack: added Unregister method.

dot/core/service.go
edwardmack: updated to `runtimeUpdateSubscriptions` and `runtimeUpdateSubscriptionsLock`

dot/core/service.go
edwardmack: updated to return err here.

dot/core/service.go
edwardmack: Good question, no reason, so I've removed the select case.

dot/rpc/subscription/websocket.go
edwardmack: yes, I've added it.

EclesioMeloJunior (approved):

dot/core/service.go
```diff
@@ -474,20 +475,17 @@ func (s *Service) handleChainReorg(prev, curr common.Hash) error {
...
 
-	if len(s.runtimeChanged) == 0 {
+	if len(s.runtimeUpdateSubscriptions) == 0 {
 		return
 	}
 
-	logger.Info("notifying runtime updated chans...", "chans", s.runtimeChanged)
-	for _, ch := range s.runtimeChanged {
+	logger.Info("notifying runtime updated chans...", "chans", s.runtimeUpdateSubscriptions)
+	for _, ch := range s.runtimeUpdateSubscriptions {
```
timwu20, line 486: it would be nice for this to use `sync.WaitGroup` to ensure that there are no blocked/leaked goroutines.  Right now this will just spawn `len(s.runtimeUpdateSubscriptions)`number of goroutines, that potentially could be blocked if there is no process reading from the channel.

dot/core/service.go
```diff
@@ -474,20 +475,17 @@ func (s *Service) handleChainReorg(prev, curr common.Hash) error {
...
 		go func(ch chan<- runtime.Version) {
-			select {
-			case ch <- version:
-			default:
-			}
+			ch <- version
```
timwu20, line 488: what if someone unregisters/closes the channel at the same time this happens? I would assume it panics.

timwu20 (changes requested):

dot/core/service.go
edwardmack: Any suggestions regarding how I should test this, and/or fixes?

dot/core/service.go
edwardmack: @timwu20  I've added the `sync.WaitGroup`, I'm not entirely sure I've done this the way you envision this, please take a look and let me know how I should adjust.

dot/core/service.go
timwu20: you could call `notifyRuntimeUpdated` and in parallel goroutines call `Register/UnregisterRuntimeUpdatedChannel`.  You should be able to produce a panic.  There are a few ways to fix it, but let's see if we can get it to panic first.

dot/core/service.go
timwu20: It's not completely necessary to pass in the `*sync.WaitGroup`, `wg` can be accessed in the closure.

dot/core/service.go
```diff
@@ -579,3 +611,41 @@ func (s *Service) GetMetadata(bhash *common.Hash) ([]byte, error) {
...
+		}
+	}
+
+	s.runtimeUpdateSubscriptions[id] = ch
+	return id, nil
+}
+
+// UnregisterRuntimeUpdatedChannel function to unregister runtime updated channel
+func (s *Service) UnregisterRuntimeUpdatedChannel(id byte) bool {
+	ch := s.runtimeUpdateSubscriptions[id]
```
timwu20, line 638: check if the id exists
```suggestion
	ch, ok := s.runtimeUpdateSubscriptions[id]
```

edwardmack: updated.

edwardmack: @timwu20 I created `TestService_RegisterUpRegisterPanic` to test parallel calls, and the test panics (although I'm not sure if it's for the correct reason).  Please take a look and let me know if I'm on the right track.

timwu20: What panic did you get?

timwu20: I pulled your branch and I'm not getting any panics

edwardmack: Currently there is a `t.Skip()` on the test TestService_RegisterUnRegisterPanic since it fails with a fatal error: concurrent map read and map write, so not the panics were expecting?

timwu20: oh well that should be fixed too.  Let me try and reproduce on my end.

dot/core/service.go
```diff
@@ -548,3 +574,41 @@ func (s *Service) GetMetadata(bhash *common.Hash) ([]byte, error) {
...
+func (s *Service) RegisterRuntimeUpdatedChannel(ch chan<- runtime.Version) (byte, error) {
+	s.runtimeUpdateSubscriptionsLock.Lock()
+	defer s.runtimeUpdateSubscriptionsLock.Unlock()
+
+	if len(s.runtimeUpdateSubscriptions) == 256 {
+		return 0, errors.New("channel limit reached")
+	}
+
+	var id byte
+	for {
```
timwu20, line 588: `generateID` should probably be a receiver function on the `Service` itself and you can check if there are collisions in the map so you don't have to loop here.

dot/core/service.go
```diff
@@ -548,3 +574,41 @@ func (s *Service) GetMetadata(bhash *common.Hash) ([]byte, error) {
...
+		}
+	}
+
+	s.runtimeUpdateSubscriptions[id] = ch
+	return id, nil
+}
+
+// UnregisterRuntimeUpdatedChannel function to unregister runtime updated channel
+func (s *Service) UnregisterRuntimeUpdatedChannel(id byte) bool {
+	ch, ok := s.runtimeUpdateSubscriptions[id]
```
timwu20, line 601: locking here will solve your concurrent read/write error and consequently avoid any case where the channel is closed.

dot/core/service.go
```diff
@@ -548,3 +574,41 @@ func (s *Service) GetMetadata(bhash *common.Hash) ([]byte, error) {
...
+		close(ch)
+		s.runtimeUpdateSubscriptions[id] = nil
+		return true
+	}
+	return false
+}
+
+func generateID() byte {
+	// skipcq: GSC-G404
+	id := rand.Intn(256) //nolint
```
timwu20, line 612: maybe use a uuid?

dot/core/service.go
```diff
@@ -548,3 +574,41 @@ func (s *Service) GetMetadata(bhash *common.Hash) ([]byte, error) {
...
+	s.runtimeUpdateSubscriptions[id] = ch
+	return id, nil
+}
+
+// UnregisterRuntimeUpdatedChannel function to unregister runtime updated channel
+func (s *Service) UnregisterRuntimeUpdatedChannel(id byte) bool {
+	ch, ok := s.runtimeUpdateSubscriptions[id]
+	if ok {
+		close(ch)
+		s.runtimeUpdateSubscriptions[id] = nil
```
timwu20, line 604: delete from the map instead of making it `nil`

timwu20 (changes requested):

dot/core/service.go
edwardmack: Yes, that seems better.

dot/core/service.go
edwardmack: Yes, updated and it does fix that error.

dot/core/service.go
edwardmack: Updated to use uuid, do I still need to do the loop check since this is a uuid?

dot/core/service.go
edwardmack: updated.

dot/core/service.go
timwu20: Yea you should still check even though it's extremely unlikely.

timwu20 (changes requested):
Just the final `uuid` uniqueness check.  In the current code you're not generating a new uuid.

dot/core/service.go
```diff
@@ -405,6 +411,26 @@ func (s *Service) handleChainReorg(prev, curr common.Hash) error {
...
 
+func (s *Service) notifyRuntimeUpdated(version runtime.Version) {
+	s.runtimeUpdateSubscriptionsLock.RLock()
+	defer s.runtimeUpdateSubscriptionsLock.RUnlock()
+
+	if len(s.runtimeUpdateSubscriptions) == 0 {
+		return
+	}
+
+	logger.Info("notifying runtime updated chans...", "chans", s.runtimeUpdateSubscriptions)
```
noot, line 422: change to debug or trace

dot/core/service.go
noot: yeah probably related to #1701 , you could add a per-channel lock and `closed bool` as a potential solution, @timwu20  do you know of any other good sulutions?

noot: wait nvm, you lock the whole map before closing any of the channels. I think that should fix it

dot/rpc/subscription/listeners.go
```diff
@@ -280,9 +287,31 @@ func (l *RuntimeVersionListener) Listen() {
 	ver.TransactionVersion = rtVersion.TransactionVersion()
 	ver.Apis = modules.ConvertAPIs(rtVersion.APIItems())
 
-	l.wsconn.safeSend(newSubscriptionResponse("state_runtimeVersion", l.subID, ver))
+	go l.wsconn.safeSend(newSubscriptionResponse("state_runtimeVersion", l.subID, ver))
+
+	// listen for runtime updates
+	go func() {
+		for info := range l.runtimeUpdate {
```
noot, line 294:
```suggestion
		for info, ok := range l.runtimeUpdate {
		    if !ok {
		        return
		    }
		    
		    // rest of code
```

add closed channel check

noot (commented):

dot/rpc/subscription/listeners.go
edwardmack: When I try to add ok assignment as above, I get compile error `too many variables in range`, is there another way to check this?

dot/core/service.go
edwardmack: Updated.

dot/rpc/subscription/listeners.go
noot: oh nvm I missed that this is a range, you'd have to change it to read from the channel in a for loop
```
for {
    info, ok := <-l.runtimeUpdate {
       if !ok {
           return
       }
    //rest of code
}
```

edwardmack:
> Just the final `uuid` uniqueness check. In the current code you're not generating a new uuid.

@timwu20 doesn't the call to `uuid := uuid.New()` in line 609 of dot/core/service.go create a new uuid?  If not, please let me know how to do this.

dot/core/service.go
```diff
@@ -550,3 +576,42 @@ func (s *Service) GetMetadata(bhash *common.Hash) ([]byte, error) {
...
+	if ok {
+		close(ch)
+		delete(s.runtimeUpdateSubscriptions, id)
+		return true
+	}
+	return false
+}
+
+func (s *Service) generateID() uint32 {
+	uuid := uuid.New()
```
timwu20, line 609: put this in the for loop to generate a new one each iteration.

edwardmack: Oh, of course, now I see.

dot/core/service.go
```diff
@@ -408,6 +408,26 @@ func (s *Service) handleChainReorg(prev, curr common.Hash) error {
...
+//	var wg sync.WaitGroup
+//	wg.Add(len(s.runtimeUpdateSubscriptions))
+//	for _, ch := range s.runtimeUpdateSubscriptions {
+//		go func(ch chan<- runtime.Version) {
+//			defer wg.Done()
+//			ch <- version
+//		}(ch)
+//	}
+//	wg.Wait()
+//}
```
noot, line 429: remove

dot/state/block.go
```diff
@@ -706,6 +711,7 @@ func (bs *BlockState) HandleRuntimeChanges(newState *rtstorage.TrieState, rt run
 		return fmt.Errorf("failed to update code substituted block hash: %w", err)
 	}
 
+	bs.notifyRuntimeUpdated(newVersion)
```
noot, line 714: `newVersion` is nil if the `!codeSubBlockHash.Equal(common.Hash{})` if block isn't entered, you can just move getting the version to right before here, since it's cached if it's already been fetched before

noot:
```suggestion
        newVersion, err := rt.Version()
        //err check
	bs.notifyRuntimeUpdated(newVersion)
```

dot/rpc/subscription/listeners.go
noot: just pinging on this

noot (commented):
great work, just smaller comments

dot/core/service.go
edwardmack: Removed.

dot/state/block.go
edwardmack: good point, updated.

dot/rpc/subscription/listeners.go
edwardmack: updated.

dot/state/block.go
```diff
@@ -705,6 +710,11 @@ func (bs *BlockState) HandleRuntimeChanges(newState *rtstorage.TrieState, rt run
 		return fmt.Errorf("failed to update code substituted block hash: %w", err)
 	}
 
+	newVersion, err := rt.Version()
+	if err != nil {
+		return fmt.Errorf("failed to retrieve runtime version: %w", err)
+	}
+	bs.notifyRuntimeUpdated(newVersion)
```
noot, line 717: I wonder if it's good to put this in a goroutine since it might block if some channel is full

noot (approved):
looks good :D

timwu20 (approved):

dot/state/block.go
edwardmack: good idea, added as goroutine.

PR #1687, opened 2021-07-09
fix(lib/grandpa): Fix max value for digest.

### LINKED ISSUE #1680: devnet: ensure blocks close to chain head are finalized

arijitAD:
## Changes

- Fix max value for digest.
- `2^64` does bitwise xor. Hence `2^64 = 66`

## Tests

```

```

## Issues

- #592
- closes #1680

### DIFF
--- a/dot/digest/digest.go
+++ b/dot/digest/digest.go
@@ -19,6 +19,7 @@ package digest
 import (
 	"context"
 	"errors"
+	"math"
 	"math/big"
 
 	"github.com/ChainSafe/gossamer/dot/types"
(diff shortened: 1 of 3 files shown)


noot (approved):
very nice!!!!

timwu20 (approved):

PR #1693, opened 2021-07-13
chore(lib/common) implement byte pool to improve websocket subscription efficiency

### LINKED ISSUE #1477: Improve performance of websocket subscription using buffered pool
## Issue summary

- Currently, we are generating a random number for subscriptions ID to find an empty slot for a subscription. We might end up wasting too much time generating a random number because Intn doesn't generate a purely random number. 
- Instead, we can use a buffered pool to achieve the same.
- Buffered pool implemented using channels is also thread-safe. (Locks are not required)

- We can modify this code to serve our use case https://github.com/oxtoacart/bpool/blob/master/bufferpool.go

edwardmack:
## Changes

- Issue #1477 outlines suggestions for changes to improve performance during websocket initialization.  This update introduces a byte pool that replaces the random number generator.
-
-

## Tests

```
go test ./lib/common/...
```

## Issues

- closes #1477


EclesioMeloJunior (changes requested):
I guess we should use `bs.importedBytepool.Put(byte)` in the `Unregister` function

dot/state/block.go
```diff
@@ -46,10 +46,12 @@ type BlockState struct {
...
-	imported      map[byte]chan<- *types.Block
-	finalised     map[byte]chan<- *types.FinalisationInfo
-	importedLock  sync.RWMutex
-	finalisedLock sync.RWMutex
+	imported          map[byte]chan<- *types.Block
+	finalised         map[byte]chan<- *types.FinalisationInfo
+	importedLock      sync.RWMutex
+	finalisedLock     sync.RWMutex
+	importedBytepool  *common.BytePool
+	finalisedBytepool *common.BytePool
```
noot, line 54:
```suggestion
	importedBytePool  *common.BytePool
	finalisedBytePool *common.BytePool
```

lib/common/bytepool.go
```diff
@@ -0,0 +1,66 @@
...
+	}
+	return
+}
+
+// Put returns the given Buffer to the BytePool.
+func (bp *BytePool) Put(b byte) error {
+	select {
+	case bp.c <- b:
+		return nil
+	default: // Discard the buffer if the pool is full.
```
noot, line 58: remove comment

dot/state/block.go
edwardmack: updated.

lib/common/bytepool.go
edwardmack: comment removed.

edwardmack:
> I guess we should use `bs.importedBytepool.Put(byte)` in the `Unregister` function

Good point, I've added this.

EclesioMeloJunior (approved):

lib/common/bytepool.go
```diff
@@ -0,0 +1,66 @@
...
+	select {
+	case bp.c <- b:
+		return nil
+	default:
+		return fmt.Errorf("pool is full")
+	}
+}
+
+// NumPooled returns the number of items currently pooled.
+func (bp *BytePool) NumPooled() int {
```
timwu20, line 64:
```suggestion
func (bp *BytePool) Len() int {
```

timwu20 (approved):

lib/common/bytepool.go
edwardmack: updated.

PR #1694, opened 2021-07-14
chore (pkg/scale): Integrate scale into grandpa library

jimjbrettj:
## Changes

- Integrate new scale pkg into grandpa

## Tests

```
go test ./lib/grandpa -short -v
```

## Issues

- #1621


jimjbrettj:
Waiting to open this until scale network PR is merged

lib/grandpa/network.go
```diff
@@ -225,41 +225,20 @@ func (s *Service) sendNeighbourMessage() {
...
-	)
-
-	switch msg.Data[0] {
-	case voteType:
-		r := &bytes.Buffer{}
-		_, _ = r.Write(msg.Data[1:])
-		vm := &VoteMessage{}
-		err = vm.Decode(r)
-		m = vm
+	err = scale.Unmarshal(msg.Data, &gossipMessage)
```
EclesioMeloJunior, line 228: we need to check this error, right?

EclesioMeloJunior (changes requested):

lib/grandpa/network.go
jimjbrettj: I have it below the switch, but I can move it up if that's better

EclesioMeloJunior: would be nice to check the error before the switch, will be easier to relate the error to the `Unmarshal`

lib/grandpa/grandpa.go
```diff
@@ -842,7 +843,7 @@ func (s *Service) finalise() error {
 		return err
 	}
 
-	pcj, err := newJustification(s.state.round, bfc.Hash, bfc.Number, pcs).Encode()
+	pcj, err := scale.Marshal(newJustification(s.state.round, bfc.Hash, bfc.Number, pcs))
```
timwu20, line 846: nit: create a local var for the justification on another line

lib/grandpa/message.go
```diff
@@ -105,31 +72,17 @@ type VoteMessage struct {
...
-	return err
-}
-
 // Type returns voteType
 func (v *VoteMessage) Type() byte {
 	return voteType
 }
 
+// Index Returns VDT index
+func (v *VoteMessage) Index() uint { return 0 }
```
timwu20, line 81:
```suggestion
func (vm *VoteMessage) Index() uint { return 0 }
```
Update all the other receiver functions too.

lib/grandpa/message.go
```diff
@@ -147,9 +100,12 @@ type NeighbourMessage struct {
 	Number  uint32
 }
 
+// Index Returns VDT index
+func (m *NeighbourMessage) Index() uint { return 2 }
```
timwu20, line 104:
```suggestion
func (nm *NeighbourMessage) Index() uint { return 2 }
```

lib/grandpa/message.go
```diff
@@ -202,57 +135,8 @@ type CommitMessage struct {
...
-		err = f.AuthData[i].Decode(r)
-		if err != nil {
-			return err
-		}
-	}
-
-	return nil
-}
+// Index Returns VDT index
+func (f *CommitMessage) Index() uint { return 1 }
```
timwu20, line 139:
```suggestion
func (cm *CommitMessage) Index() uint { return 1 }
```

lib/grandpa/network.go
timwu20: Yes please handle the error immediately

lib/grandpa/network.go
```diff
@@ -33,6 +31,7 @@ var (
 	grandpaID                protocol.ID = "/paritytech/grandpa/1"
 	messageID                            = network.ConsensusMsgType
 	neighbourMessageInterval             = time.Minute * 5
+	gossipMessage                        = scale.MustNewVaryingDataType(&VoteMessage{}, &CommitMessage{}, &NeighbourMessage{}, &catchUpRequest{}, &catchUpResponse{})
```
timwu20, line 34: don't use the same VaryingDataType in `decodeMessage`.  This isn't thread safe if there are parallel calls to `decodeMessage`. You will be overwriting the same address repeatedly, which could lead to undesired side effects.

timwu20 (changes requested):

PR #1699, opened 2021-07-15
chore (pkg/scale) types refactor to utilize new scale pkg

jimjbrettj:
## Changes

- This is a refactor of the types used in the dot package. Includes changes to digestItem, digest, header, block data, block, and grandpa (grandpa is minor currently). 
- This PR also includes a scale integration into the Digest and Network packages
- Note: Some of the files have a mix of old and new scale. I didn't want to get carried away with what I did in this PR, so I plan to continue with unfinished scale integrations in future PRs, unless it's suggested I do otherwise.

## Tests

```
go test ./dot/types -v
go test ./dot/network -v -short
go test ./dot/digest -v -short
go test ./pkg/scale -v -short
```

## Issues

- #1621


dot/types/consensus_digest.go
```diff
@@ -28,12 +28,12 @@ type GrandpaScheduledChange struct {
 
 // Encode returns a SCALE encoded GrandpaScheduledChange with first type byte
 func (sc *GrandpaScheduledChange) Encode() ([]byte, error) {
```
timwu20, line 30: please take this refactor a step further and remove the `Digest` interface.  I don't think we need it, since the interface is just handling the encoding of the type byte, and the actual scale encoding.

dot/digest/digest.go
```diff
@@ -395,14 +391,13 @@ func (h *Handler) handleResume(d *types.ConsensusDigest) error {
 		return err
 	}
 
-	p := &types.GrandpaResume{}
-	dec, err := scale.Decode(d.Data[1:], p)
+	var r types.GrandpaResume
+	err = scale.Unmarshal(d.Data[1:], &r)
```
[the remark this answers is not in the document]
jimjbrettj: Yeah I think there is more room for refactor. It does this because consensus digest defines the first byte as one of:
`var (
	GrandpaScheduledChangeType = byte(1)
	GrandpaForcedChangeType    = byte(2)
	GrandpaOnDisabledType      = byte(3)
	GrandpaPauseType           = byte(4)
	GrandpaResumeType          = byte(5)
)`
Maybe make these VDTs?

dot/types/consensus_digest.go
jimjbrettj: Okay sounds good. With the changes from the above comment I should be able to get rid of it

jimjbrettj: @timwu20 Does the types/consensus_digest refactor look alright to you?

dot/digest/digest.go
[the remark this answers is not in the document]
timwu20: Yes these are all rust enums and therefore should be VDTs.

dot/types/consensus_digest.go
timwu20: You don't need all the const types anymore, if you have them defined in the `Index()` receiver func that fulfills the `VaryingDataTypeValue` interface.

jimjbrettj: @timwu20 okay I removed the const types. That look okay to you?

timwu20: The only thing I would mention is maybe we shouldn't use the same variables for the VDTs.

```
var BabeConsensusDigest = scale.MustNewVaryingDataType(NextEpochData{}, BABEOnDisabled{}, NextConfigData{})
var GrandpaConsensusDigest = scale.MustNewVaryingDataType(GrandpaScheduledChange{}, GrandpaForcedChange{}, GrandpaOnDisabled{}, GrandpaPause{}, GrandpaResume{})
```

This isn't thread safe, and could product nasty side effects if multiple goroutines are trying to unmarshal.  Maybe create a constructor method that returns you a varying data type.  

```
func NewBabeConsensusDigest() scale.VaryingDataType {
   return scale.MustNewVaryingDataType(NextEpochData{}, BABEOnDisabled{}, NextConfigData{})
}
```

This will set us up for when I support the custom `scale.VaryingDataType`.

jimjbrettj: Good call. Will change this

dot/digest/digest_test.go
```diff
@@ -419,7 +460,16 @@ func TestHandler_HandleNextEpochData(t *testing.T) {
...
 	stored, err := handler.epochState.(*state.EpochState).GetEpochData(1)
 	require.NoError(t, err)
-	res, err := digest.ToEpochData()
+
+	var act types.NextEpochData
+	switch val := digest.Value().(type) {
+	case types.NextEpochData:
+		act = val
+	default:
+		fmt.Println("THIS SHOULDNT HAPPEN")
```
noot, line 469: maybe `t.Fatal()` instead? haha

dot/digest/digest_test.go
```diff
@@ -448,7 +500,15 @@ func TestHandler_HandleNextConfigData(t *testing.T) {
 	err = handler.handleConsensusDigest(d, header)
 	require.NoError(t, err)
 
+	var act types.NextConfigData
+	switch val := digest.Value().(type) {
+	case types.NextConfigData:
+		act = val
+	default:
+		fmt.Println("THIS SHOULDNT HAPPEN")
```
noot, line 508: same here

dot/network/block_announce.go
```diff
@@ -40,7 +39,7 @@ type BlockAnnounceMessage struct {
 	Number         *big.Int
 	StateRoot      common.Hash
 	ExtrinsicsRoot common.Hash
-	Digest         types.Digest
+	Digest         scale.VaryingDataTypeSlice
```
noot, line 42: curious why this is a SCALE type now instead of `types.Digest`?

dot/network/block_announce.go
```diff
@@ -254,6 +238,7 @@ func (s *Service) handleBlockAnnounceMessage(peer peer.ID, msg NotificationsMess
 		s.syncQueue.handleBlockAnnounce(an, peer)
 		err = s.syncer.HandleBlockAnnounce(an)
 		if err != nil {
+			fmt.Println(err)
```
noot, line 241: remove

dot/state/block.go
```diff
@@ -374,16 +369,16 @@ func (bs *BlockState) SetBlockBody(hash common.Hash, body *types.Body) error {
 // CompareAndSetBlockData will compare empty fields and set all elements in a block data to db
 func (bs *BlockState) CompareAndSetBlockData(bd *types.BlockData) error {
 	hasReceipt, _ := bs.HasReceipt(bd.Hash)
```
noot, line 371: check err?

dot/network/block_announce.go
jimjbrettj: This is because scale currently doesnt support custom VDTs, so in order to decode as a VDT it has to be of that type. If we add that functionality to scale, then we would be able to define a type Digest that can be treated as a VDT, but that isn't supported currently.

dot/types/babe.go
```diff
@@ -30,17 +30,18 @@ type BabeConfiguration struct {
...
-func BABEAuthorityRawToAuthority(adr []*AuthorityRaw) ([]*Authority, error) {
-	ad := make([]*Authority, len(adr))
+func BABEAuthorityRawToAuthority(adr []AuthorityRaw) ([]Authority, error) {
+	ad := make([]Authority, len(adr))
 	for i, r := range adr {
-		ad[i] = new(Authority)
-		err := ad[i].FromRawSr25519(r)
+		ad[i] = Authority{}
+		a := r
+		err := ad[i].FromRawSr25519(&a)
```
noot, line 44:
```suggestion
		err := ad[i].FromRawSr25519(&r)
```

dot/types/block_data.go
```diff
@@ -18,196 +18,64 @@ package types
...
-	Header        *optional.Header
-	Body          *optional.Body
-	Receipt       *optional.Bytes
-	MessageQueue  *optional.Bytes
-	Justification *optional.Bytes
+	Header        *Header
+	Body          *Body
+	Receipt       *[]byte
+	MessageQueue  *[]byte
+	Justification *[]byte
```
noot, line 33: would be nice to add a comment saying these are optional fields and that's why they are pointers

dot/types/block_data.go
```diff
@@ -18,196 +18,64 @@ package types
...
+// NewEmptyBlockData Creates an empty blockData struct
+func NewEmptyBlockData() *BlockData {
+	bd := &BlockData{
+		Header:        nil,
+		Body:          nil,
+		Receipt:       nil,
+		MessageQueue:  nil,
+		Justification: nil,
+	}
+	return bd
```
noot, line 45:
```suggestion
	return &BlockData{}
```

dot/network/block_announce.go
jimjbrettj: As to why I used a VDT instead of the old Digest type, its because in the new scale package the buffer used in decoding/encoding is not public, which makes incremental mashalling/unmarshalling (which was pretty common in the code before) harder to do. With a Vdt representing the digest I can just call unmarshal on the header and it takes care of it, whereas if I used the old type I would have to do it in stages (that new scale doesn't really allow) using custom encode/decode logic.

So before the encoding/decoding was done in stages using custom logic as opposed to letting scale handle it all. Now it is all handled by scale without the need for custom serialization logic

dot/types/block_data_test.go
```diff
@@ -17,292 +17,176 @@
...
+	"github.com/ChainSafe/gossamer/pkg/scale"
 	"github.com/stretchr/testify/require"
 )
 
-var testDigest = &Digest{
-	&PreRuntimeDigest{
+var digestItem = scale.MustNewVaryingDataType(ChangesTrieRootDigest{}, PreRuntimeDigest{}, ConsensusDigest{}, SealDigest{})
+var digest = scale.NewVaryingDataTypeSlice(digestItem)
+
+var testDigest = digest
```
noot, line 31:
```suggestion
var 
   ( digestItem = scale.MustNewVaryingDataType(ChangesTrieRootDigest{}, PreRuntimeDigest{}, ConsensusDigest{}, SealDigest{})
    digest = scale.NewVaryingDataTypeSlice(digestItem)

   testDigest = digest
)
```

dot/types/body.go
```diff
@@ -18,14 +18,13 @@ package types
...
 import (
 	"bytes"
-	"errors"
 	"fmt"
-	"io"
 	"math/big"
 
 	"github.com/ChainSafe/gossamer/lib/common"
 	"github.com/ChainSafe/gossamer/lib/common/optional"
 	"github.com/ChainSafe/gossamer/lib/scale"
```
noot, line 26: we are using the original scale here, did you want to change it to only use scalev2 or do that in a follow up?

lib/babe/build.go
```diff
@@ -113,7 +113,9 @@ func (b *BlockBuilder) buildBlock(parent *types.Header, slot Slot, rt runtime.In
 
 	// create new block header
 	number := big.NewInt(0).Add(parent.Number, big.NewInt(1))
-	header, err := types.NewHeader(parent.Hash(), common.Hash{}, common.Hash{}, number, types.NewDigest(preDigest))
+	digest := types.NewDigest()
+	_ = digest.Add(*preDigest)
```
noot, line 117: I noticed the error is often ignored for `digest.Add()`, is there a chance an error would occur?

dot/network/block_announce.go
noot: cool, this makes sense. I like it!!

noot (approved):
amazing work!!! no major issues, just smaller comments. tried syncing the node and works the same as before, so networking encoding/decoding is all good. great stuff!

dot/core/service.go
```diff
@@ -190,12 +198,12 @@ func (s *Service) HandleBlockProduced(block *types.Block, state *rtstorage.TrieS
 }
 
 func (s *Service) handleBlock(block *types.Block, state *rtstorage.TrieState) error {
-	if block == nil || block.Header == nil || state == nil {
+	if block == nil || state == nil {
```
timwu20, line 201: shouldn't this return an error?  Why would the block and state be `nil`? @noot

noot: yeah, it shouldn't be nil, should return an error

dot/types/babe.go
```diff
@@ -100,13 +101,17 @@ type ConfigData struct {
...
+	if len(header.Digest.Types) == 0 {
 		return 0, fmt.Errorf("chain head missing digest")
 	}
 
-	digestItem := header.Digest[0]
-	preDigest, ok := digestItem.(*PreRuntimeDigest)
-	if !ok {
+	digestItem := header.Digest.Types[0]
+
+	var preDigest *PreRuntimeDigest
```
timwu20, line 110: You can use alternate typecast syntax here if you're looking for a specific type.

dot/types/block.go
```diff
@@ -17,42 +17,40 @@
...
 		Body:   body,
 	}
 }
 
 // NewEmptyBlock returns a new Block with an initialised but empty Header and Body
-func NewEmptyBlock() *Block {
-	return &Block{
-		Header: new(Header),
-		Body:   new(Body),
+func NewEmptyBlock() Block {
```
timwu20, line 38: can we remove this function from this file.  Seems to be only used in tests.

lib/babe/build.go
timwu20: this error should be handled.  In the event the inputted param doesn't fulfil interface `VaryingDataTypeValue`, then it will error out.

lib/babe/verify.go
```diff
@@ -273,23 +274,29 @@ func newVerifier(blockState BlockState, epoch uint64, info *verifierInfo) (*veri
...
-	preDigest, ok := preDigestItem.(*types.PreRuntimeDigest)
-	if !ok {
+	preDigestItem := header.Digest.Types[0]
+	sealItem := header.Digest.Types[len(header.Digest.Types)-1]
+
+	var preDigest *types.PreRuntimeDigest
+	switch val := preDigestItem.Value().(type) {
+	case types.PreRuntimeDigest:
+		preDigest = &val
+	default:
```
timwu20, line 291: revise to typecast and `ok` check

lib/babe/verify.go
```diff
@@ -273,23 +274,29 @@ func newVerifier(blockState BlockState, epoch uint64, info *verifierInfo) (*veri
...
 		return fmt.Errorf("first digest item is not pre-digest")
 	}
 
-	seal, ok := sealItem.(*types.SealDigest)
-	if !ok {
+	var seal *types.SealDigest
+	switch val := sealItem.Value().(type) {
```
timwu20, line 296: same here

lib/babe/verify.go
```diff
@@ -436,14 +448,17 @@ func (b *verifier) verifyPrimarySlotWinner(authorityIndex uint32, slot uint64, v
...
+	if len(header.Digest.Types) == 0 {
 		return 0, fmt.Errorf("no digest provided")
 	}
 
-	digestItem := header.Digest[0]
+	digestItem := header.Digest.Types[0]
 
-	preDigest, ok := digestItem.(*types.PreRuntimeDigest)
-	if !ok {
+	var preDigest *types.PreRuntimeDigest
```
timwu20, line 457: same here

timwu20 (changes requested):

timwu20:
really good work @jimjbrettj

dot/state/block.go
```diff
@@ -236,7 +238,7 @@ func (bs *BlockState) HasHeader(hash common.Hash) (bool, error) {
 
 // GetHeader returns a BlockHeader for a given hash
 func (bs *BlockState) GetHeader(hash common.Hash) (*types.Header, error) {
-	result := new(types.Header)
+	result := types.NewEmptyHeader()
```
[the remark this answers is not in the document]
jimjbrettj: If I'm not mistaken you would need the custom VDTs for that. Since the header.Digest is a vdt slice, we still need to specify what VDT items to unmarshall into. Like I think unmarshalling into types.header would give you an index not found error since it scale doesn't know what its unmarshalling.

dot/state/block.go
```diff
@@ -251,18 +253,12 @@ func (bs *BlockState) GetHeader(hash common.Hash) (*types.Header, error) {
...
-	}
-
-	_, err = result.Decode(rw)
+	err = scale.Unmarshal(data, result)
 	if err != nil {
 		return nil, err
 	}
 
-	if reflect.DeepEqual(result, new(types.Header)) {
+	if reflect.DeepEqual(result, types.NewEmptyHeader()) {
```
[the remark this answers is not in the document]
jimjbrettj: I was thinking about this and am not totally sure how to do that. Because types.NewEmptyHeader() initializes the digest vdt slice, I don't believe it == types.Header{}, and don't think it can be nil. And I don't believe VDT slices are comparable, so I don't think I can compare to VDT slices for equality. Do you have an ideas on where i'm wrong here/how else to approach it @timwu20 @noot?

dot/types/body.go
jimjbrettj: My thought process was given the size of the PR, I would just change what I needed to in order to get everything stable with respect to the new digest changes, so I was thinking I would do that in a follow up. Of course I can add it to this PR if you see fit

dot/core/service.go
jimjbrettj: Sorry I'm a little confused how to act upon this. Are you just saying that this nil check be returning an error instead of nil? Also the new header type cannot be compared to nil like it was before. I figured the check was there since the old block header was a pointer, and now that its not I removed that check. However if there is more to this check than I thought, I would need some insight as to how to approach that @timwu20 @noot

dot/sync/syncer.go
```diff
@@ -306,7 +294,7 @@ func (s *Service) handleBody(body *types.Body) error {
 
 // handleHeader handles blocks (header+body) included in BlockResponses
 func (s *Service) handleBlock(block *types.Block) error {
-	if block == nil || block.Header == nil || block.Body == nil {
+	if block == nil {
```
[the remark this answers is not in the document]
jimjbrettj: Before these types were pointers so there was a nil check, but now they aren't pointers. I can still do this check on block.Body, but since the header contains a VDT slice it isn't comparable so I cannot do this nil check. I alluded to this in an above comment, but if there was more to this check than just making sure there isn't a nil pointer reference I would need to discuss how to handle that for the header. Could I get your thoughts on this @noot?

dot/types/babe.go
jimjbrettj: @noot This change causes a "Implicit memory aliasing in for loop" am I okay to just add nolint here?

noot: weird, ok just leave it I guess lol

dot/core/service.go
timwu20: the former.  
> Are you just saying that this nil check be returning an error instead of nil?
this should return an error

dot/state/block.go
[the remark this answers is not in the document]
timwu20: oh I see.  Yea makes sense.

dot/state/block.go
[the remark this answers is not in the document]
timwu20: implement a custom method `(func h Header) Empty() bool` that checks that the number is 0 and ensures that the Digest (VDTSlice) is length 0.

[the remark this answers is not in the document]
jimjbrettj: implemented. Added checks for the other fields as well because I ran into issues where it was all empty except for one or two of the hash fields

dot/sync/syncer.go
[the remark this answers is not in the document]
timwu20: Now that you've implemented the `Header.Empty()` receiver func you can use it here and implement a `Block.Empty()` method which checks zero case and/or 0 length `Block.Body` and `Header.Empty()`.

dot/types/header.go
```diff
@@ -75,9 +87,14 @@ func (bh *Header) DeepCopy() *Header {
...
 
-	if len(bh.Digest) > 0 {
-		cp.Digest = make([]DigestItem, len(bh.Digest))
-		copy(cp.Digest[:], bh.Digest[:])
+	if len(bh.Digest.Types) > 0 {
+		cp.Digest = NewDigest()
+		for _, d := range bh.Digest.Types {
+			err := cp.Digest.Add(d.Value())
+			if err != nil {
+				panic(err)
```
timwu20, line 95: we shouldn't have panics in the code.  Change the function sig of `DeepCopy` to return an error and handle it on the calling side.

lib/runtime/life/exports_test.go
```diff
@@ -14,6 +13,7 @@ import (
 	"github.com/ChainSafe/gossamer/lib/runtime/storage"
 	"github.com/ChainSafe/gossamer/lib/scale"
 	"github.com/ChainSafe/gossamer/lib/trie"
+	scale2 "github.com/ChainSafe/gossamer/pkg/scale"
```
timwu20, line 16: is the removal of the original `scale` in another issue?

tests/data/db/LOCK
```diff
@@ -0,0 +1 @@
+18991
```
timwu20, line 1: remove this file

pkg/scale/decode.go
```diff
@@ -397,6 +397,12 @@ func (ds *decodeState) decodeStruct(dstv reflect.Value) (err error) {
 		if !field.CanInterface() {
 			continue
 		}
+		// if the value is not a zero value, set it as non-zero value from dst.
+		// this is required for VaryingDataTypeSlice and VaryingDataType
+		inv := reflect.ValueOf(in)
+		if inv.Field(i.fieldIndex).IsValid() && !inv.Field(i.fieldIndex).IsZero() {
+			field.Set(inv.Field(i.fieldIndex))
+		}
```
timwu20, line 405: I forgot I wrote this.

timwu20 (changes requested):

lib/runtime/life/exports_test.go
jimjbrettj: Yeah that would be in the integrate scale into runtime issue

timwu20 (approved):

timwu20 (approved):
looks like you need to re-run your tests.  LGTM!

PR #1700, opened 2021-07-21
fix(dot/state): add StorageState Lock/Unlock API for usage by babe and sync 

noot:
## Changes

- add StorageState Lock/Unlock API which is called before babe/sync begin any trie writes to prevent concurrent trie writes (snapshot should prevent this but doesn't, will open an issue for this)
- update `StorageState.tries` to be `*sync.Map`
- re-add commented out code that has since been fixed in core for handling chain re-org

## Tests

```
go test ./dot/state
```

alternatively, run 2-3 nodes (with gssmr genesis-spec modified to have 3 authorities) and ensure that there are no `Storage root mismatch` or other trie-related panics due to unexpected roots

## Issues

- related to #592


dot/state/storage.go
```diff
@@ -190,13 +199,31 @@ func (s *StorageState) LoadFromDB(root common.Hash) (*trie.Trie, error) {
...
+func (s *StorageState) loadTrie(root *common.Hash) (*trie.Trie, error) {
+	if root == nil {
+		sr, err := s.blockState.BestBlockStateRoot()
+		if err != nil {
+			return nil, err
+		}
+		root = &sr
+	}
+
+	if t, has := s.tries.Load(*root); !has && t != nil {
```
edwardmack, line 215: should this be `has && t != nil` not `!has`?

edwardmack (approved):
Nice work.

dot/state/storage.go
noot: good catch, updated!

dot/core/service.go
```diff
@@ -422,12 +423,11 @@ func (s *Service) maintainTransactionPool(block *types.Block) error {
 	// re-validate transactions in the pool and move them to the queue
 	txs := s.transactionState.PendingInPool()
 	for _, tx := range txs {
-		// TODO: re-add this on update to v0.8
-
+		// TODO: re-add this
```
timwu20, line 426: does this need to be re-added before merging?

dot/sync/interface.go
```diff
@@ -54,9 +54,10 @@ type BlockState interface {
 // StorageState is the interface for the storage state
 type StorageState interface {
 	TrieState(root *common.Hash) (*rtstorage.TrieState, error)
-	StoreTrie(ts *rtstorage.TrieState, header *types.Header) error
 	LoadCodeHash(*common.Hash) (common.Hash, error)
 	SetSyncing(bool)
+	BeginModifyTrie()
+	FinishModifyTrie()
```
timwu20, line 60:
```suggestion
	Lock()
	Unlock()
```
you embedded the `sync.Mutex` so it should fulfil the interface

timwu20 (changes requested):

dot/core/service.go
noot: I'll open an issue for this, didn't want to do it here

dot/sync/interface.go
noot: nice, updated!

lib/trie/trie.go
```diff
@@ -54,11 +54,22 @@ func NewTrie(root node) *Trie {
...
+			root:        c.root,
+			deletedKeys: make([]common.Hash, 0),
+			parallel:    c.parallel,
+		}
+	}
+
 	newTrie := &Trie{
 		generation:  t.generation + 1,
 		root:        t.root,
 		childTries:  t.childTries,
```
EclesioMeloJunior, line 70: the map on line 57 must be used on line 70 at `childTries` property?

EclesioMeloJunior (changes requested):

lib/trie/trie.go
noot: updated!

EclesioMeloJunior (approved):

PR #1707, opened 2021-07-27
feat(dot/rpc/modules) implement `state_queryStorage` rpc method

EclesioMeloJunior:
## Changes

- Add RPC method `state_queryStorage` that allows query for keys from a starting block until an ending block
- Add `QueryStorage` method at `core` Service to interact with the block storage

## Tests

```
go test -run ^TestTryQueryStore_WhenThereIsDataToRetrieve$ github.com/ChainSafe/gossamer/dot/core
go test -run ^TestTryQueryStore_WhenDoesNotHaveDataToRetrieve$ github.com/ChainSafe/gossamer/dot/core
go test -run ^TestTryQueryState_WhenDoesNotHaveStateRoot$ github.com/ChainSafe/gossamer/dot/core
go test -run ^TestQueryStorate_WhenBlocksHasData$ github.com/ChainSafe/gossamer/dot/core
```

## Issues

- Related to #310

## Screenshots

1) Querying a key by just one block

image

2) Querying key using the start and end blocks

image
image


dot/core/interface.go
```diff
@@ -48,6 +48,7 @@ type BlockState interface {
 	HighestCommonAncestor(a, b common.Hash) (common.Hash, error)
 	SubChain(start, end common.Hash) ([]common.Hash, error)
 	GetBlockBody(hash common.Hash) (*types.Body, error)
+	GetHeader(hash common.Hash) (*types.Header, error)
```
noot, line 51: I don't see this used anywhere

dot/core/service.go
```diff
@@ -579,3 +579,60 @@ func (s *Service) GetMetadata(bhash *common.Hash) ([]byte, error) {
...
+
+type (
+	Changes map[string]string
+)
+
+func (s *Service) QueryStorage(from *common.Hash, to *common.Hash, keys []string) (map[common.Hash]Changes, error) {
+	var err error
+	blocksToQuery := []common.Hash{*from}
+	if to != nil {
+		if blocksToQuery, err = s.blockState.SubChain(*from, *to); err != nil {
```
noot, line 591: check if `from` is nil as well

dot/core/service.go
```diff
@@ -579,3 +579,60 @@ func (s *Service) GetMetadata(bhash *common.Hash) ([]byte, error) {
...
+	}
+
+	ts, err := s.storageState.TrieState(stateRootHash)
+	if err != nil {
+		return nil, err
+	}
+
+	changes := make(map[string]string)
+
+	for _, k := range keys {
```
noot, line 623: this should use `storageState.GetStorage()` as it'll be more efficient, `TrieState` might end up loading unneeded tries into memory

dot/core/service.go
```diff
@@ -579,3 +579,60 @@ func (s *Service) GetMetadata(bhash *common.Hash) ([]byte, error) {
...
+		keyBytes, err := common.HexToBytes(k)
+		if err != nil {
+			return nil, err
+		}
+
+		storedData := ts.Get(keyBytes)
+		if bytes.EqualFold(storedData, []byte{}) {
+			continue
+		}
```
noot, line 632: check if returned value is `nil` instead of `[]byte`, it's possible a key might have `[]byte{}` which is not the same as nil

dot/rpc/modules/state.go
```diff
@@ -387,8 +390,32 @@ func (sm *StateModule) GetStorageSize(r *http.Request, req *StateStorageSizeRequ
...
+			changes = append(changes, []string{key, value})
+		}
+
+		response = append(response, StorageChangeSetResponse{
+			Block:   &block,
+			Changes: changes,
+		})
+	}
+
+	logger.Warn("built response", "res", response)
```
noot, line 416: remove

dot/rpc/modules/state.go
```diff
@@ -387,8 +390,32 @@ func (sm *StateModule) GetStorageSize(r *http.Request, req *StateStorageSizeRequ
 }
 
 // QueryStorage isn't implemented properly yet.
-func (sm *StateModule) QueryStorage(r *http.Request, req *StateStorageQueryRangeRequest, res *StorageChangeSetResponse) error {
-	// TODO implement change storage trie so that block hash parameter works (See issue #834)
+func (sm *StateModule) QueryStorage(r *http.Request, req *StateStorageQueryRangeRequest, res *[]StorageChangeSetResponse) error {
+	logger.Warn("receiving data", "start", req.StartBlock, "end", req.EndBlock, "keys", req.Keys)
+	changesByBlock, err := sm.coreAPI.QueryStorage(req.StartBlock, req.EndBlock, req.Keys)
+	logger.Warn("after query", "err", err)
```
noot, line 396: remove logs

dot/core/interface.go
EclesioMeloJunior: Removed

dot/rpc/modules/state.go
```diff
@@ -387,8 +387,28 @@ func (sm *StateModule) GetStorageSize(r *http.Request, req *StateStorageSizeRequ
...
 // QueryStorage isn't implemented properly yet.
-func (sm *StateModule) QueryStorage(r *http.Request, req *StateStorageQueryRangeRequest, res *StorageChangeSetResponse) error {
-	// TODO implement change storage trie so that block hash parameter works (See issue #834)
+func (sm *StateModule) QueryStorage(r *http.Request, req *StateStorageQueryRangeRequest, res *[]StorageChangeSetResponse) error {
+	changesByBlock, err := sm.coreAPI.QueryStorage(req.StartBlock, req.EndBlock, req.Keys)
+	if err != nil {
+		return err
+	}
+
+	var response []StorageChangeSetResponse
```
arijitAD, line 396: Pre-allocate this slice.
```
response := make([]StorageChangeSetResponse, 0, len(changesByBlock))
```

dot/core/service.go
```diff
@@ -550,3 +551,66 @@ func (s *Service) GetMetadata(bhash *common.Hash) ([]byte, error) {
 	rt.SetContextStorage(ts)
 	return rt.Metadata()
 }
+
+type (
```
arijitAD, line 555: Can we declare this at the top or somewhere in `lib/common`?

dot/core/service.go
```diff
@@ -550,3 +551,66 @@ func (s *Service) GetMetadata(bhash *common.Hash) ([]byte, error) {
...
+}
+
+// tryQueryStorage will try to get all the `keys` inside the block's current state
+func (s *Service) tryQueryStorage(block common.Hash, keys []string) (Changes, error) {
+	stateRootHash, err := s.storageState.GetStateRootFromBlock(&block)
+	if err != nil {
+		return nil, err
+	}
+
+	changes := make(map[string]string)
```
arijitAD, line 595: Wouldn't it be better if we make this slice initially instead of a map?

dot/core/service.go
```diff
@@ -550,3 +551,66 @@ func (s *Service) GetMetadata(bhash *common.Hash) ([]byte, error) {
...
+		}
+
+		queries[hash] = changes
+	}
+
+	return queries, nil
+}
+
+// tryQueryStorage will try to get all the `keys` inside the block's current state
+func (s *Service) tryQueryStorage(block common.Hash, keys []string) (Changes, error) {
```
timwu20, line 589:
```suggestion
func (s *Service) QueryStorage(block common.Hash, keys ...string) (Changes, error) {
```
It's assumed you are making a best effort to retrieve the `Changes`. Variadic parameter would be nice for single keys.

dot/core/service.go
timwu20:
```suggestion
	changes := make(Changes)
```

dot/core/service.go
EclesioMeloJunior: why did we need to change the method's name?

dot/core/service.go
EclesioMeloJunior: as a map, this function can return the data as `key -> value` format, I don't know if a slice will be better for represent the data inside the block

dot/core/service.go
```diff
@@ -550,3 +554,61 @@ func (s *Service) GetMetadata(bhash *common.Hash) ([]byte, error) {
 	rt.SetContextStorage(ts)
 	return rt.Metadata()
 }
+
+// QueryStorage returns the key-value data by block based on `keys` params
+// on every block starting `from` until `to` block, if `to` is not nil
+func (s *Service) QueryStorage(from *common.Hash, to *common.Hash, keys ...string) (map[common.Hash]QueryKeyValueChanges, error) {
+	if from == nil {
+		return nil, errors.New("cannot query data without a starting block hash")
+	}
```
noot, line 563: why is `from` a pointer, if it can never be nil?

dot/core/service.go
```diff
@@ -550,3 +554,61 @@ func (s *Service) GetMetadata(bhash *common.Hash) ([]byte, error) {
...
+
+	var err error
+	blocksToQuery := []common.Hash{*from}
+	if to != nil {
+		if blocksToQuery, err = s.blockState.SubChain(*from, *to); err != nil {
+			return nil, err
+		}
+	}
```
noot, line 571: I think if there is no `to` provided, you should set it to the `BestBlockHash`, the spec implies if no `to` is provided then you query every block from `to` until the head of the chain. can you double check what substrate returns?

dot/core/service.go
EclesioMeloJunior: Resolved!

dot/core/service.go
EclesioMeloJunior: Done!

dot/rpc/modules/state.go
EclesioMeloJunior: Done!

dot/core/service.go
EclesioMeloJunior: I move to the top

dot/core/service.go
EclesioMeloJunior: Good question, I can update the `from` to be a value of `common.Hash` instead of a pointer, could `to` stay as a pointer?

noot: yes, that would make sense!

dot/core/service.go
EclesioMeloJunior: Ok! I will take a look at substrate node implementation

EclesioMeloJunior: Update to use the `BestBlockHash` if `to` is nil

edwardmack (approved):

timwu20 (approved):

dot/rpc/modules/state.go
```diff
@@ -85,9 +86,9 @@ type StateStorageRequest struct {
 
 // StateStorageQueryRangeRequest holds json fields
 type StateStorageQueryRangeRequest struct {
-	Keys       []*common.Hash `json:"keys" validate:"required"`
-	StartBlock *common.Hash   `json:"startBlock" validate:"required"`
-	Block      *common.Hash   `json:"block"`
+	Keys       []string     `json:"keys" validate:"required"`
+	StartBlock *common.Hash `json:"startBlock" validate:"required"`
```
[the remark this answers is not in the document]
EclesioMeloJunior: Done!

PR #1712, opened 2021-07-29
feat(dot/rpc/modules): add `system_addReservedPeer` and `system_removeReservedPeer` RPC call

EclesioMeloJunior:
## Changes

- Implement `system_addReservedPeer` and `system_removeReservedPeer` RPC call

## Tests

```
go test -run ^Test_RemoveReservedPeers$ github.com/ChainSafe/gossamer/dot/network
go test -run ^Test_AddReservedPeers$ github.com/ChainSafe/gossamer/dot/network
go test -run ^Test_ReservedPeersRPCCalls$ github.com/ChainSafe/gossamer/dot/rpc/modules
```

## Issues

- Related #731


dot/network/host.go
```diff
@@ -368,6 +368,68 @@ func (h *host) peers() []peer.ID {
...
+		}
+
+		addinfo, err := peer.AddrInfoFromP2pAddr(maddr)
+		if err != nil {
+			return err
+		}
+
+		h.h.Peerstore().AddAddrs(addinfo.ID, addinfo.Addrs, peerstore.PermanentAddrTTL)
+
+		fmt.Println(h.h.Peerstore().Addrs(addinfo.ID))
```
noot, line 390: this doesn't add the peer as "reserved" which means we try to keep persistent connections to them. you should add the peer's ID to the `ConnManager`'s `persistentPeers` map, and connect to the peer

dot/network/host.go
```diff
@@ -368,6 +368,68 @@ func (h *host) peers() []peer.ID {
...
+		if err != nil {
+			return err
+		}
+
+		addinfo, err := peer.AddrInfoFromP2pAddr(maddr)
+		if err != nil {
+			return err
+		}
+
+		h.h.Peerstore().AddAddrs(addinfo.ID, addinfo.Addrs, peerstore.PermanentAddrTTL)
```
noot, line 388:
```suggestion
		```suggestion
         // you'll need to add the `ConnManager` to the host
		h.connManager.persistentPeers.Store(addinfo.ID)
		err := h.connect(addinfo)
		// handle err
```
```

File: dot/network/host.go (hunk 1 above)
Comment (noot) [reply dot/network/host.go:390]:
```suggestion
         // you'll need to add the `ConnManager` to the host
		h.connManager.persistentPeers.Store(addinfo.ID)
		err := h.connect(addinfo)
		// handle err
```

dot/network/host.go
noot, line 388:
```suggestion
         // you'll need to add the `ConnManager` to the host
		h.connManager.persistentPeers.Store(addinfo.ID)
		err := h.connect(addinfo)
		// handle err
```

dot/network/host.go
```diff
@@ -368,6 +368,68 @@ func (h *host) peers() []peer.ID {
...
+		}
+
+		if err := h.closeProotocolsForPeer(peerID); err != nil {
+			return err
+		}
+
+		if err := h.closePeer(peerID); err != nil {
+			return err
+		}
+		h.h.Peerstore().ClearAddrs(peerID)
```
noot, line 415:
```suggestion
		peerID := peer.ID(id)
		h.connManager.persistentPeers.Delete(peerID)
		
```

dot/network/host.go
```diff
@@ -368,6 +368,68 @@ func (h *host) peers() []peer.ID {
...
+	if err != nil {
+		return err
+	}
+
+	for _, pid := range protocols {
+		h.closeStream(p, protocol.ID(pid))
+	}
+	return nil
+}
+
```
noot, line 432: I don't think you need to close all connections to a peer when you remove them from `reserved`, but I'll double check

dot/network/host.go
```diff
@@ -368,6 +369,53 @@ func (h *host) peers() []peer.ID {
...
+		h.h.ConnManager().Protect(addinfo.ID, "")
+		if err := h.connect(*addinfo); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+// removeReservedPeers will close all connections with the peer and remove it from the peerstore
```
noot, line 398:
```suggestion
// removeReservedPeers will remove the given peers from the protected peers list
```

dot/network/host.go
```diff
@@ -368,6 +369,53 @@ func (h *host) peers() []peer.ID {
 	return h.h.Network().Peers()
 }
 
+// addReservedPeers adds the peers `addrs` in the Peerstore
```
noot, line 372:
```suggestion
// addReservedPeers adds the peers `addrs` to the protected peers list and connects to them
```

noot (approved):
looks great :D

dot/network/host.go
```diff
@@ -368,6 +369,53 @@ func (h *host) peers() []peer.ID {
...
+
+	for _, id := range ids {
+		peerID, err := peer.Decode(id)
+		if err != nil {
+			return err
+		}
+
+		del := h.h.ConnManager().Unprotect(peerID, "")
+		if !del {
+			return errors.New("could not disconnect peer")
```
arijitAD, line 412: Can you confirm if `Unprotect` also disconnects? 
I think `Unprotect` removes it from the map and returns false when the peer was not present in the protected map.

EclesioMeloJunior: Yes, `Unprotect` just removes the `peerID` from the map

arijitAD (approved):

dot/network/host.go
EclesioMeloJunior: Good point, would be better to just jump to the next one instead of returning an error, I will fix that!

dot/network/host.go
```diff
@@ -368,6 +368,52 @@ func (h *host) peers() []peer.ID {
 	return h.h.Network().Peers()
 }
 
+// addReservedPeers adds the peers `addrs` to the protected peers list and connects to them
+func (h *host) addReservedPeers(addrs ...string) error {
+	if len(addrs) < 1 {
```
timwu20, line 373: unnecessary check IMO

dot/network/host.go
```diff
@@ -368,6 +368,52 @@ func (h *host) peers() []peer.ID {
...
+			return err
+		}
+	}
+
+	return nil
+}
+
+// removeReservedPeers will remove the given peers from the protected peers list
+func (h *host) removeReservedPeers(ids ...string) error {
+	if len(ids) < 1 {
```
timwu20, line 399: unnecessary check IMO

dot/rpc/modules/system.go
```diff
@@ -280,3 +280,21 @@ func (sm *SystemModule) LocalPeerId(r *http.Request, req *EmptyRequest, res *str
 	*res = base58.Encode([]byte(netstate.PeerID))
 	return nil
 }
+
+// AddReservedPeer adds a reserved peer. The string parameter should encode a p2p multiaddr.
+func (sm *SystemModule) AddReservedPeer(r *http.Request, req *StringRequest, res *[]byte) error {
+	if req.String == "" {
```
timwu20, line 286: Should you trim just in case?

timwu20 (changes requested):

dot/network/host.go
EclesioMeloJunior: Done!

dot/network/host.go
EclesioMeloJunior: Done!

dot/rpc/modules/system.go
EclesioMeloJunior: Done!

timwu20 (approved):

PR #1715, opened 2021-08-04
feat(dot/rpc) implement `author_hasSessionKeys` RPC call

EclesioMeloJunior:
## Changes

- Fix `author_InsertKey` RPC call to insert keys properly;
- Implement `author_HasSessionkeys` to check the key is in the node;
- Check if the public key bytes are all filled by zeros, in that case, we skip and check the next one.

## Tests

```
go test -timeout 30s -run ^TestAuthorModule_HasSessionKey$ github.com/ChainSafe/gossamer/dot/rpc/modules
```

## Issues

- N/A


EclesioMeloJunior:
This PR will be merged after PR #1714 was marged

noot (approved):

timwu20 (approved):

PR #1733, opened 2021-08-17
Return on EOF error while reading stream.

arijitAD:
## Changes

- Return on EOF error while reading stream.

## Tests

```
TestStreamCloseEOF
```

## Issues

-

## Primary Reviewer 

- @EclesioMeloJunior


dot/network/service.go
```diff
@@ -581,7 +581,7 @@ func (s *Service) readStream(stream libp2pnetwork.Stream, decoder messageDecoder
 	for {
 		tot, err := readStream(stream, msgBytes[:])
 		if err == io.EOF {
```
EclesioMeloJunior, line 583: would be nice to use `errors.Is(err, io.EOF)`?

EclesioMeloJunior (changes requested):

dot/network/service.go
arijitAD: Done

EclesioMeloJunior (approved):

edwardmack (approved):

PR #1734, opened 2021-08-17
feat(lib/trie) implement merkle proofs for tries

### LINKED ISSUE #213: Implement merkle proofs for trie
The trie currently calculates the merkle root but does not have any functionality for creating or verifying proofs.

To create a proof, we need the value whose inclusion we are trying to prove, along with the path needed to end up with the root hash.

EclesioMeloJunior:
## Changes
- Implement Merkle proofs for tries

## Tests

```
go test -run ^TestVerifyProof$ github.com/ChainSafe/gossamer/lib/trie
go test -run ^TestVerifyProofOneElement$ github.com/ChainSafe/gossamer/lib/trie
go test -run ^TestVerifyProof_BadProof$ github.com/ChainSafe/gossamer/lib/trie
go test -run ^TestGenerateProofMissingKey$ github.com/ChainSafe/gossamer/lib/trie
go test -run ^TestGenerateProofNoMorePathToFollow$ github.com/ChainSafe/gossamer/lib/trie
```

## Issues

- Closes #213
- Unblocks #310

## Primary Reviewer 

- @arijitAD


lib/trie/proof_test.go
```diff
@@ -0,0 +1,143 @@
...
+	}
+
+	return trie, vals
+}
+
+func randBytes(n int) []byte {
+	r := make([]byte, n)
+	crand.Read(r)
+	return r
+}
```
arijitAD, line 143: Also, added test for missing key.

lib/trie/proof_test.go
```diff
@@ -0,0 +1,143 @@
+package trie
+
+import (
+	"fmt"
+	"io/ioutil"
+	"math/rand"
+	"os"
+	"testing"
+
```
arijitAD, line 9: Remove this space. All standard packages should be group together.

lib/trie/proof_test.go
```diff
@@ -0,0 +1,143 @@
...
+	require.NoError(t, err)
+
+	db, err := chaindb.NewBadgerDB(&chaindb.Config{
+		InMemory: true,
+		DataDir:  tmpdir,
+	})
+	require.NoError(t, err)
+
+	clear := func() {
+		err := db.ClearAll()
```
arijitAD, line 30: No need to `ClearAll`. You can directly `Close` and `RemoveAll`

lib/trie/proof_test.go
```diff
@@ -0,0 +1,143 @@
...
+	return db, clear
+}
+
+func TestVerifyProof(t *testing.T) {
+	trie, entries := randomTrie(t, 500)
+	root, err := trie.Hash()
+	require.NoError(t, err)
+
+	for _, kv := range entries {
+		t.Run("", func(t *testing.T) {
```
arijitAD, line 49: We do not require subtests since we are testing the same thing. Remove this line.

lib/trie/proof_test.go
```diff
@@ -0,0 +1,143 @@
...
+	require.Equal(t, []byte("V"), val)
+}
+
+func TestVerifyProof_BadProof(t *testing.T) {
+	trie, entries := randomTrie(t, 200)
+	rootHash, err := trie.Hash()
+	require.NoError(t, err)
+
+	for _, entry := range entries {
+		t.Run("", func(t *testing.T) {
```
arijitAD, line 93: We do not require subtests since we are testing the same thing. Remove this line.

arijitAD (approved):

edwardmack (commented):
I'm not sure I fully understand merkle proofs, but I thought there should be a function for generating the merkle proof root, given the merkle-proof and the tx key.  

Also, I'm not use what fromLvl is used for in Prove.

I'm looking at https://computersciencewiki.org/index.php/Merkle_proof as an example, it's in javascript, so maybe I'm missing something in translation.

lib/trie/proof_test.go
EclesioMeloJunior: Done

lib/trie/proof_test.go
EclesioMeloJunior: Done

lib/trie/proof_test.go
EclesioMeloJunior: Update

lib/trie/proof_test.go
EclesioMeloJunior: Update

edwardmack (approved):
@EclesioMeloJunior and I connected an zoom call where he explained the differences between the trie structure used in the example code I was referencing and our gossamer's trie structure, which cleared up my confusion.  He also addressed by concerns with Proof fromLvl, and verifyProof.  

Great work, lgtm.

lib/trie/proof.go
```diff
@@ -0,0 +1,137 @@
...
+	// ErrEmptyNibbles occurs when trying to prove or valid a proof to an empty key
+	ErrEmptyNibbles = errors.New("empty nibbles provided from key")
+
+	// ErrInvalidProof occurs when the key could not be validated
+	ErrInvalidProof = errors.New("provided key is not present at trie")
+)
+
+// Prove constructs the merkle-proof for key. The result contains all encoded nodes
+// on the path to the value at key. Returns an error if could not found the key
+func (t *Trie) Prove(key []byte, fromLvl uint, db chaindb.Writer) (int, error) {
```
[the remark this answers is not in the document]
EclesioMeloJunior: update the name to `GenerateProof`

lib/trie/proof.go
```diff
@@ -0,0 +1,137 @@
...
+		if err = db.Put(hashNode, encHashNode); err != nil {
+			return len(nodes), err
+		}
+	}
+
+	return len(nodes), nil
+}
+
+// VerifyProof checks merkle proofs given an proof
+func VerifyProof(rootHash common.Hash, key []byte, db chaindb.Reader) ([]byte, error) {
```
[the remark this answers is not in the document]
EclesioMeloJunior: Changed to return a boolean instead of the byte value

lib/trie/proof.go
```diff
@@ -0,0 +1,145 @@
...
+			}
+
+			length := lenCommonPrefix(n.key, key)
+			next := n.children[key[length]]
+			if next == nil {
+				return false, nil
+			}
+
+			key = key[length+1:]
+			copy(wantedHash[:], next.getHash())
```
arijitAD, line 142: Do we need to copy this?

EclesioMeloJunior: not necessarily, do you think if we use `wantedHash = next.getHash()` would be better?

lib/trie/proof_test.go
EclesioMeloJunior: Done, added tests for paths with incompatible leaf nodes and for empty branch nodes slots

lib/trie/proof.go
arijitAD: Yes, We should ideally avoid copy wherever possible. This doesn't have a significant performance impact since it is 32 bytes but still good to avoid copies.

EclesioMeloJunior: Fine, I achieve that by transforming the `wantedHash` into a byte slice instead of a `common.Hash`, now the assignment occurs without the need of `copy`

Issue #1701, opened 2021-07-21
panic: send on closed channel in block import notifier


noot:
## Describe the bug

Occasionally when running the node I've seen this panic, causing the node to crash:

```
panic: send on closed channel

goroutine 56677 [running]:
github.com/ChainSafe/gossamer/dot/state.(*BlockState).notifyImported.func1(0xc005da4120, 0xc005f02780)
	/home/ubuntu/gossamer/dot/state/block_notify.go:107 +0x3f
created by github.com/ChainSafe/gossamer/dot/state.(*BlockState).notifyImported
	/home/ubuntu/gossamer/dot/state/block_notify.go:105 +0x1ae
```

need to make sure there are no writes to channels that are potentially closed

## To Reproduce
Steps to reproduce the behaviour:

This happens quite infrequently, so not sure how to reproduce it other than running the node a bunch of times

PR #1736, opened 2021-08-18
fix: confirm block import notifier is closed properly

edwardmack:
## Changes

- Refactor Block Import notification so that channel is handled in block_notify logic.
- Replace RegisterImportedBlockChannel with GetImportedBlockNotifierChannel which returns channel that will be signaled when a block is imported.
- Replace UnregisterImportedBlockImported with FreeImportedBlockNotifierChannel which removes channel from block imported map, and closes channel.

## Tests

```
make test
```

## Issues

- closes: #1701 

## Primary Reviewer 

- @timwu20


dot/rpc/subscription/websocket.go
```diff
@@ -213,6 +213,8 @@ func (c *WSConn) initStorageChangeListener(reqID float64, params interface{}) (L
 }
 
 func (c *WSConn) initBlockListener(reqID float64, _ interface{}) (Listener, error) {
+	// TODO (ed) here the imported Channel get unregistered and closed in http.Stop
+	//  there doesn't seem to be an un-subscribe for this
```
timwu20, line 217: where does `BlockListener.Channel` get closed? i don't see it in the `BlockListener.Stop`

dot/rpc/subscription/websocket.go
```diff
@@ -285,6 +287,7 @@ func (c *WSConn) initExtrinsicWatch(reqID float64, params interface{}) (Listener
 		return nil, err
 	}
 
+	// TODO (ed) the importedChan get deleted by unregister imported channel in defer in Listen (listeners.go line 236)
 	// listen for built blocks
 	esl := &ExtrinsicSubmitListener{
```
timwu20, line 292: can we move this code to `listeners.go` to a `NewExtrinsicSubmitListener` constructor.  That way we can see the creation and closing of channels all in the same file.

dot/rpc/subscription/websocket.go
timwu20: Can we move this code to a constructor for `BlockListener`.

edwardmack: It seems there is logic to close imported channel in dot/http.go  line 194-5, which in inside HTTPServer.Stop.  There is also within Listener.Stop a call to cancelWithTimeout that gets called when unsubscribe is called.  Should we move the imported channel close into Listener.Stop, then call listener.Stop from within HTTPServer.Stop?

dot/rpc/subscription/websocket.go
edwardmack: Moved code to new NewExtrinsicSubmitListener in listeners.go.

dot/rpc/subscription/websocket.go
timwu20: Yea we should localise closing channels (managing internal state) to receiver functions of the struct.

EclesioMeloJunior: @timwu20 the `BlockListener.Channel` channel is closed when `l.wsconn.BlockAPI.UnregisterImportedChannel(l.ChanID)` is called, we should not close the `BlockListener.Channel` directly because the `BlockAPI` is the owner of the channel. Every listener has a `Stop` method that calls the function `cancelWithTimeout` which will stop any running listener's goroutines, which in turn will call the `defer function` to unregister any active channel

EclesioMeloJunior: But, the implementation of `Unregister...` function is to just remove the channel from the map without closing it, and that is the problem. @edwardmack I think we could close the imported channel inside the `defer` function of each `Listen` method to keep things closer.

EclesioMeloJunior (changes requested):

dot/rpc/subscription/websocket.go
edwardmack: I added imported channel close to defer in Listen methods.

edwardmack:
I believe I've updated this for the comments, please re-review and let me know.

EclesioMeloJunior (approved):

dot/digest/digest.go
```diff
@@ -74,6 +74,8 @@ type resume struct {
 
 // NewHandler returns a new Handler
 func NewHandler(blockState BlockState, epochState EpochState, grandpaState GrandpaState) (*Handler, error) {
+	// TODO (ed) imported channel is unregistered and closed in handler.Stop
```
noot, line 77: is this still a TODO?

dot/rpc/subscription/listeners.go
```diff
@@ -123,12 +123,25 @@ type BlockListener struct {
 	cancelTimeout time.Duration
 }
 
+// NewBlockListener constructor for creating BlockListener
+func NewBlockListener(conn *WSConn) *BlockListener {
+	bl := &BlockListener{
+		Channel:       make(chan *types.Block, DEFAULT_BUFFER_SIZE),
```
noot, line 129: can this be renamed `blockCh` or something like that?

lib/grandpa/grandpa.go
```diff
@@ -377,6 +377,7 @@ func (s *Service) initiate() error {
 
 func (s *Service) waitForFirstBlock() error {
 	ch := make(chan *types.Block)
+	// TODO (ed) import channel is unregistered by defer, doesn't seem to be closed
```
noot, line 380: I think this function has a problem. you can update the defer func to first `close(ch)` then call `Unregister()`, I think that should fix the issue at least partially, maybe there is a case where the GC collects `ch` and closes it before the `Unregister` is finished

noot (commented):
I wonder if we should refactor the `block_notify` architecture to prevent this - just having some API like `GetNotifierChannel` and `FreeNotifierChannel` which returns a new channel that's managed by the `BlockState` and then deals with closing it as well. I think that might make more sense than what we currently have where the reader of the channel is closing it, since in general the writer should close the channel

edwardmack:
I've refactored this to make imported block notification channels to be handled by block notify logic.  This is ready for a re-review.

EclesioMeloJunior (changes requested):

dot/state/block_notify.go
```diff
@@ -26,22 +27,16 @@ import (
...
+// DEFAULT_BUFFER_SIZE buffer size for channels
+const DEFAULT_BUFFER_SIZE = 100
 
+// GetImportedBlockNotifierChannel function to retrieve a imported block notifier channel
+func (bs *BlockState) GetImportedBlockNotifierChannel() (chan *types.Block, error) {
+	ch := make(chan *types.Block, DEFAULT_BUFFER_SIZE)
 	bs.importedLock.Lock()
-	bs.imported[id] = ch
+	bs.imported[ch] = struct{}{}
 	bs.importedLock.Unlock()
```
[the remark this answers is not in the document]
edwardmack: updated.

EclesioMeloJunior (approved):

dot/core/interface.go
```diff
@@ -41,8 +41,8 @@ type BlockState interface {
 	GetSlotForBlock(common.Hash) (uint64, error)
 	GetFinalisedHeader(uint64, uint64) (*types.Header, error)
 	GetFinalisedHash(uint64, uint64) (common.Hash, error)
-	RegisterImportedChannel(ch chan<- *types.Block) (byte, error)
-	UnregisterImportedChannel(id byte)
+	GetImportedBlockNotifierChannel() (chan *types.Block, error)
+	FreeImportedBlockNotifierChannel(ch chan *types.Block)
```
noot, line 45:
```suggestion
	GetImportedBlockNotifierChannel() (<-chan *types.Block, error)
	FreeImportedBlockNotifierChannel(ch <-chan *types.Block)
```

update to receive-only channel

dot/digest/digest.go
```diff
@@ -111,9 +109,8 @@ func (h *Handler) Start() error {
 // Stop stops the Handler
 func (h *Handler) Stop() error {
 	h.cancel()
-	h.blockState.UnregisterImportedChannel(h.importedID)
+	h.blockState.FreeImportedBlockNotifierChannel(h.imported)
 	h.blockState.UnregisterFinalisedChannel(h.finalisedID)
-	close(h.imported)
 	close(h.finalised)
```
noot, line 114: it would be good to update the `finalised` notifiers with the new architecture as well, as it has the same problem. can open another issue or do this in this PR, up to you

dot/rpc/subscription/listeners.go
```diff
@@ -181,6 +194,7 @@ func (l *BlockFinalizedListener) Listen() {
 		defer func() {
 			l.wsconn.BlockAPI.UnregisterFinalisedChannel(l.chanID)
 			close(l.done)
+			close(l.channel)
```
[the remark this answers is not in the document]
noot: or this should be calling `FreeImportedBlockNotifierChannel` instead?

dot/state/block_notify.go
```diff
@@ -26,22 +26,18 @@ import (
...
-	bs.importedLock.RLock()
+// DEFAULT_BUFFER_SIZE buffer size for channels
+const DEFAULT_BUFFER_SIZE = 100
 
-	id, err := bs.importedBytePool.Get()
-	if err != nil {
-		return 0, err
-	}
+// GetImportedBlockNotifierChannel function to retrieve a imported block notifier channel
+func (bs *BlockState) GetImportedBlockNotifierChannel() (chan *types.Block, error) {
```
noot, line 33:
```suggestion
func (bs *BlockState) GetImportedBlockNotifierChannel() (<-chan *types.Block, error) {
```

dot/state/block_notify.go
```diff
@@ -62,17 +58,12 @@ func (bs *BlockState) RegisterFinalizedChannel(ch chan<- *types.FinalisationInfo
 	return id, nil
 }
 
-// UnregisterImportedChannel removes the block import notification channel with the given ID.
-// A channel must be unregistered before closing it.
-func (bs *BlockState) UnregisterImportedChannel(id byte) {
+// FreeImportedBlockNotifierChannel to free and close imported block notifier channel
+func (bs *BlockState) FreeImportedBlockNotifierChannel(ch chan *types.Block) {
```
noot, line 62:
```suggestion
func (bs *BlockState) FreeImportedBlockNotifierChannel(ch <-chan *types.Block) {
```

dot/state/block_notify.go
noot: also, these functions don't identify the channel by an ID anymore - is the pointer to the channel the same in both functons? (I'm not sure if the read-only and write-only parts of the channel point to different locations potentially)

dot/state/block_notify_test.go
```diff
@@ -33,11 +33,10 @@ var testMessageTimeout = time.Second * 3
 func TestImportChannel(t *testing.T) {
 	bs := newTestBlockState(t, testGenesisHeader)
 
-	ch := make(chan *types.Block, 3)
-	id, err := bs.RegisterImportedChannel(ch)
+	ch, err := bs.GetImportedBlockNotifierChannel()
 	require.NoError(t, err)
 
-	defer bs.UnregisterImportedChannel(id)
+	defer bs.FreeImportedBlockNotifierChannel(ch)
```
noot, line 39: can you confirm the channel is removed from the map after this?

edwardmack:
I’m struggling with refactoring GitImportedBlockNotifierChannel and FreeImportedNotifierChannel to use receive-only channels.  I’m storing these channels as keys in a map, this is working except when I convert the channel to a receive-only channel I’m not able to delete it from the map.  This is what I’m trying:
```
func (bs *BlockState) GetImportedBlockNotifierChannel() (<-chan *types.Block, error) {
	bs.importedLock.Lock()
	defer bs.importedLock.Unlock()

	ch := make(chan *types.Block, DEFAULT_BUFFER_SIZE)
	bs.imported[ch] = struct{}{}

	return ch, nil
}
```

So, this creates the new channel, stores it is the map, and returns the channel as a receive-only channel.  No issues thus far.  The problem arises in FreeImprotedBlockNotifierChannel when I try to delete the channel from the map:
```
func (bs *BlockState) FreeImportedBlockNotifierChannel(ch <-chan *types.Block) {
	bs.importedLock.Lock()
	defer bs.importedLock.Unlock()
	delete(bs.imported, ch)
}
```
I get compile error:  Cannot use 'ch' (type <-chan *types.Block) as the type chan *types.Block
It seems that in GetImportedBlockNotifierChannel, when returning the channel it turn it into a receive-only channel, then when I try to delete it from the map it doesn’t allow it because it thinks it’s a different type.  I removed the delete line, and ran with debugger and I see that map is storing pointers to the channels, so it should delete this but it doesn’t.  

So, I’ve tried wrapping the channel in a struct so that struct is stored as a key in the map, and that works, but passing this struct around doesn’t make it clear that this is a receive-only channel:
```
func (bs *BlockState) GetImportedBlockNotifierChannel() (*ImportNotifier, error) {
	bs.importedLock.Lock()
	defer bs.importedLock.Unlock()

	in := &ImportNotifier{
		ch: make(chan *types.Block, DEFAULT_BUFFER_SIZE),
	}
	bs.imported[in] = struct{}{}

	return in, nil
}

func (bs *BlockState) FreeImportedBlockNotifierChannel(ch *ImportNotifier) {
	bs.importedLock.Lock()
	defer bs.importedLock.Unlock()
	delete(bs.imported, ch)
}
```
Is there a way to define the map so that it is just storing pointers to the channels?  It’s doing that anyway, but won’t let me delete them because it thinks there different types (even tho there pointing to the same address).

Please suggest the preferred strategy for handling this.  I like the idea of storing the channels as keys in a map so that we don’t need to handle a separate key. Thoughts?

dot/digest/digest.go
edwardmack: Yes, I'll update once the architecture is set.

dot/core/interface.go
edwardmack: I'm struggling with this, see general comment...

timwu20:
> I’m struggling with refactoring GitImportedBlockNotifierChannel and FreeImportedNotifierChannel to use receive-only channels. I’m storing these channels as keys in a map, this is working except when I convert the channel to a receive-only channel I’m not able to delete it from the map. This is what I’m trying:
> 
> ```
> func (bs *BlockState) GetImportedBlockNotifierChannel() (<-chan *types.Block, error) {
> 	bs.importedLock.Lock()
> 	defer bs.importedLock.Unlock()
> 
> 	ch := make(chan *types.Block, DEFAULT_BUFFER_SIZE)
> 	bs.imported[ch] = struct{}{}
> 
> 	return ch, nil
> }
> ```
> 
> So, this creates the new channel, stores it is the map, and returns the channel as a receive-only channel. No issues thus far. The problem arises in FreeImprotedBlockNotifierChannel when I try to delete the channel from the map:
> 
> ```
> func (bs *BlockState) FreeImportedBlockNotifierChannel(ch <-chan *types.Block) {
> 	bs.importedLock.Lock()
> 	defer bs.importedLock.Unlock()
> 	delete(bs.imported, ch)
> }
> ```
> 
> I get compile error: Cannot use 'ch' (type <-chan *types.Block) as the type chan *types.Block
> It seems that in GetImportedBlockNotifierChannel, when returning the channel it turn it into a receive-only channel, then when I try to delete it from the map it doesn’t allow it because it thinks it’s a different type. I removed the delete line, and ran with debugger and I see that map is storing pointers to the channels, so it should delete this but it doesn’t.
> 
> So, I’ve tried wrapping the channel in a struct so that struct is stored as a key in the map, and that works, but passing this struct around doesn’t make it clear that this is a receive-only channel:
> 
> ```
> func (bs *BlockState) GetImportedBlockNotifierChannel() (*ImportNotifier, error) {
> 	bs.importedLock.Lock()
> 	defer bs.importedLock.Unlock()
> 
> 	in := &ImportNotifier{
> 		ch: make(chan *types.Block, DEFAULT_BUFFER_SIZE),
> 	}
> 	bs.imported[in] = struct{}{}
> 
> 	return in, nil
> }
> 
> func (bs *BlockState) FreeImportedBlockNotifierChannel(ch *ImportNotifier) {
> 	bs.importedLock.Lock()
> 	defer bs.importedLock.Unlock()
> 	delete(bs.imported, ch)
> }
> ```
> 
> Is there a way to define the map so that it is just storing pointers to the channels? It’s doing that anyway, but won’t let me delete them because it thinks there different types (even tho there pointing to the same address).
> 
> Please suggest the preferred strategy for handling this. I like the idea of storing the channels as keys in a map so that we don’t need to handle a separate key. Thoughts?

@edwardmack I created a PR at https://github.com/ChainSafe/gossamer/pull/1782/files.  Take a look. It uses a pointer, so in maybe in this case the pointer to the `WSConn` as the key, and the value is the actual channel.

edwardmack:
@timwu20 thanks for the example code, I've implemented it in this PR.  `GetImportedBlockNotifierChannel` now takes `interface{}`, I couldn't get it to work with `*interface{}` because the compiler would complain about anything I passed in was not of type `*interface{}`, changing to `interface{}` seems to fix that (I think it's a pointer anyway since passed by reference).  There are places where GetImportedBlockNotifierChannel is called without reference to `WSConn`, so I passed in other object to use as key.

edwardmack:
@timwu20 This has been updated to reflect our conversation, the interfaces are now `GetImportedBlockNotifierChannel() chan *types.Block` and `FreeImportedBlockNotifierChannel(ch chan *types.Block)`  

Ready for re-review, if this looks good I'll refactor finalised in another PR.

dot/digest/digest.go
```diff
@@ -48,7 +48,6 @@ type Handler struct {
 
 	// block notification channels
 	imported    chan *types.Block
```
noot, line 50: maybe store this as `<-chan *types.Block` so it's clear it's read-only. same for other services

dot/state/block_notify.go
```diff
@@ -26,24 +27,32 @@ import (
...
+//	defer bs.importedLock.Unlock()
+//
+//	in := &ImportNotifier{
+//		ch: make(chan *types.Block, DEFAULT_BUFFER_SIZE),
+//	}
+//	//ch := make(chan *types.Block, DEFAULT_BUFFER_SIZE)
+//	bs.imported[in] = struct{}{}
+//
+//	return in, nil
+//}
```
noot, line 54: remove

dot/state/block_notify.go
```diff
@@ -62,19 +71,21 @@ func (bs *BlockState) RegisterFinalizedChannel(ch chan<- *types.FinalisationInfo
...
+func (bs *BlockState) FreeImportedBlockNotifierChannel(ch chan *types.Block) {
 	bs.importedLock.Lock()
 	defer bs.importedLock.Unlock()
-
-	delete(bs.imported, id)
-	err := bs.importedBytePool.Put(id)
-	if err != nil {
-		logger.Error("failed to unregister imported channel", "error", err)
-	}
+	fmt.Printf("delete chan %v\n", ch)
```
noot, line 78: remove

dot/state/block_notify_test.go
noot: double checking on this

noot (commented):
nice work, just a few comments!

dot/state/block_notify.go
```diff
@@ -62,19 +71,21 @@ func (bs *BlockState) RegisterFinalizedChannel(ch chan<- *types.FinalisationInfo
 	return id, nil
 }
 
-// UnregisterImportedChannel removes the block import notification channel with the given ID.
-// A channel must be unregistered before closing it.
-func (bs *BlockState) UnregisterImportedChannel(id byte) {
+// FreeImportedBlockNotifierChannel to free and close imported block notifier channel
```
timwu20, line 74: where is the channel closed?

dot/state/block_notify.go
edwardmack: removed.

dot/digest/digest.go
edwardmack: I'm not able to store this as `<-chan *types.Block` because treating this as a receive channel breaks other parts of the code.

dot/state/block_notify.go
edwardmack: removed.

dot/state/block_notify.go
edwardmack: It's not being closed, I've updated comment.

dot/digest/digest.go
```diff
@@ -401,7 +401,7 @@ func newGrandpaChange(raw []types.GrandpaAuthoritiesRaw, delay uint32, currBlock
 	}, nil
 }
 
-func (*Handler) handleBABEOnDisabled(_ *types.ConsensusDigest, _ *types.Header) error {
+func (h *Handler) handleBABEOnDisabled(d types.BABEOnDisabled, _ *types.Header) error {
```
timwu20: what does this function do? looks like it just logs some struct and returns nil.

dot/rpc/subscription/listeners.go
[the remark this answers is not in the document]
noot: @edwardmack re-checking on this

noot (approved):
looks good - just one comment on the channel close in listeners.go

dot/rpc/subscription/listeners.go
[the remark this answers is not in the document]
edwardmack: @noot yes, I removed `close(l.channel)` from here, I'm going to implement FreeFinalisedChannel in another PR, similar to what we did here with `FreeImportedBlockNotifierChannel`

dot/state/block_notify_test.go
edwardmack: Yes, I've added a test to confirm this.

dot/digest/digest.go
edwardmack: I'm not sure why this is here...

edwardmack:
Can someone override the DeepSource requirement for this PR, it's complaining about code generated by mocks.
@timwu20 Please re-review and let me know if there are remaining issues.

EclesioMeloJunior:
@edwardmack add the below items to the end of `exclude_patterns` at .deepsource.toml to DeepSource ignore the generated mock files
```
 "**/*_test.go",
 "**/mocks/*",
 "**/mock_*"
```

timwu20 (approved):

PR #1739, opened 2021-08-19
feat: implement ext_offchain_index_set_version_1 for wasmer runtime

### LINKED ISSUE #1695: implement ext_offchain_index_set_version_1
## Task summary

- implement runtime host API function `ext_offchain_index_set_version_1`

## Specification reference

- see spec D.6.1.1
> Write a key value pair to the offchain database in a buffered fashion.

edwardmack:
## Changes

- Implement ext_offchain_index_set_version_1 for wasmer runtime library.
- The spec states write a key value pair to offchain database in buffered fashion, we've understood this to mean that data should be stored in both persistent and local storage databases. 
-

## Tests

Currently this test is skipped, it fails due to exported function not found. Is there a test runtime that has this?

```
go test -run ^Test_ext_offline_index_set_version_1$ github.com/ChainSafe/gossamer/lib/runtime/wasmer
```

## Issues

- closes #1695 

## Primary Reviewer 

- @arijitAD


arijitAD (approved):

lib/runtime/wasmer/imports_test.go
```diff
@@ -409,6 +409,27 @@ func Test_ext_storage_set_version_1(t *testing.T) {
 	require.Equal(t, testvalue, val)
 }
 
+func Test_ext_offline_index_set_version_1(t *testing.T) {
+	// TODO this currently fails with error could nat find exported function, determine how else to test this
+	t.Skip()
+	inst := NewTestInstance(t, runtime.HOST_API_TEST_RUNTIME)
```
EclesioMeloJunior, line 415: what if change from `HOST_API_TEST_RUNTIME` to `NODE_RUNTIME`?

EclesioMeloJunior (changes requested):

lib/runtime/wasmer/imports_test.go
edwardmack: I tried with `NODE_RUNTIME`, `NODE_RUNTIME_V098`, `POLKADOT_RUNTIME` and `DEV_RUNTIME` all with same results.  I think we need to build a updated test runtime that contains the "rtm_" imports that reflect to call the actual "ext" functions.

noot: @EclesioMeloJunior the `HOST_API_TEST_RUNTIME` is created by the w3f spec team, all it does is directly call the `ext_` function that is specified with the `rtm_` prefix. however it doesn't contain all the exports we need, so we might need to update it ourselves

lib/runtime/wasmer/imports.go
```diff
@@ -1348,7 +1348,23 @@ func ext_hashing_twox_64_version_1(context unsafe.Pointer, dataSpan C.int64_t) C
...
+
+	err := runtimeCtx.NodeStorage.PersistentStorage.Put(storageKey, cp)
+	if err != nil {
+		logger.Error("[ext_offchain_index_set_version_1] failed to set value in persistent storage", "error", err)
+	}
+
+	err = runtimeCtx.NodeStorage.LocalStorage.Put(storageKey, cp)
+	if err != nil {
+		logger.Error("[ext_offchain_index_set_version_1] failed to set value in local storage", "error", err)
+	}
```
noot, line 1367: it looks like neither of these are the raw database:
```
	localStorage, err := newInMemoryDB(st.DB().Path())
	if err != nil {
		return nil, err
	}

	ns := runtime.NodeStorage{
		LocalStorage:      localStorage,
		PersistentStorage: chaindb.NewTable(st.DB(), "offlinestorage"),
	}
```

I think you'll need to add a new field to the context that's the `BaseDB` (ie. non prefixed, peristent db) and then write to that

noot (changes requested):
please update to write to `BaseDB`!

lib/runtime/wasmer/imports.go
edwardmack: I added `BaseDB` to NodeStorage of runtime context, let me know if this is how we should handle this.

dot/services.go
```diff
@@ -120,9 +120,15 @@ func createRuntime(cfg *Config, st *state.Service, ks *keystore.GlobalKeystore,
 		return nil, err
 	}
 
+	baseDB, err := utils.SetupDatabase(filepath.Join(st.DB().Path(), "offline_storage"), false)
+	if err != nil {
+		return nil, err
+	}
+
```
noot, line 127: you can use `st.Base` for this

noot (commented):
mostly looks good, but you can use `state.Service.Base` as the BaseDB instead of creating a new db!

dot/core/test_helpers.go
```diff
@@ -105,6 +107,13 @@ func NewTestService(t *testing.T, cfg *Config) *Service {
 		rtCfg.CodeHash, err = cfg.StorageState.LoadCodeHash(nil)
 		require.NoError(t, err)
 
+		nodeStorage := runtime.NodeStorage{}
+
+		nodeStorage.BaseDB, err = utils.SetupDatabase(filepath.Join(testDatadirPath, "offline_storage"), false)
```
noot, line 112: would be preferred to use `stateSrvc.BaseDB` here if `stateSrvc != nil`

dot/sync/test_helpers.go
```diff
@@ -100,6 +100,11 @@ func NewTestSyncer(t *testing.T, usePolkadotGenesis bool) *Service {
 		rtCfg.CodeHash, err = cfg.StorageState.LoadCodeHash(nil)
 		require.NoError(t, err)
 
+		nodeStorage := runtime.NodeStorage{}
+		nodeStorage.BaseDB, err = utils.SetupDatabase(filepath.Join(testDatadirPath, "offline_storage"), false)
```
noot, line 104: same here

lib/babe/babe_test.go
```diff
@@ -138,6 +140,11 @@ func createTestService(t *testing.T, cfg *ServiceConfig) *Service {
 		rtCfg.CodeHash, err = storageState.LoadCodeHash(nil)
 		require.NoError(t, err)
 
+		nodeStorage := runtime.NodeStorage{}
+		nodeStorage.BaseDB, err = utils.SetupDatabase(filepath.Join(testDatadirPath, "offline_storage"), false)
```
noot, line 144: same here

lib/runtime/wasmer/imports.go
```diff
@@ -1580,9 +1591,9 @@ func storageAppend(storage runtime.Storage, key, valueToAppend []byte) error {
 	}
 
 	// append new length prefix to start of items array
-	finalVal := append(lengthEnc, valueRes...)
-	logger.Debug("[ext_storage_append_version_1]", "resulting value", fmt.Sprintf("0x%x", finalVal))
-	storage.Set(key, finalVal)
+	lengthEnc = append(lengthEnc, valueRes...)
+	logger.Debug("[ext_storage_append_version_1]", "resulting value", fmt.Sprintf("0x%x", lengthEnc))
+	storage.Set(key, lengthEnc)
```
noot, line 1596: why is this updated? I think `finalVal` is more descriptive, as this isn't an encoded length

noot (commented):
looks good, main comments are to update the test helpers to use the `stateSrvc.BaseDB` instead of spinning up a new one which is going to slow down the tests

dot/core/test_helpers.go
edwardmack: Updated.

dot/sync/test_helpers.go
edwardmack: Updated.

lib/babe/babe_test.go
edwardmack: Update.

lib/runtime/wasmer/imports.go
edwardmack: This was updated because DeepSource was complaining that append was appending to a variable with a different name.  Yes, finalVal is more descriptive, I've update to use that instead.

noot (approved):
looks good!

edwardmack:
This is ready for review.

dot/core/test_helpers.go
```diff
@@ -105,6 +107,17 @@ func NewTestService(t *testing.T, cfg *Config) *Service {
...
+		nodeStorage := runtime.NodeStorage{}
+
+		if stateSrvc != nil {
+			nodeStorage.BaseDB = stateSrvc.Base
+		} else {
+			nodeStorage.BaseDB, err = utils.SetupDatabase(filepath.Join(testDatadirPath, "offline_storage"), false)
+			require.NoError(t, err)
+		}
+
+		rtCfg.NodeStorage = nodeStorage
```
EclesioMeloJunior, line 119: lgmt! just this pice of code that repeats in some places, would be nice to create a test function helper that receives a `stateSrvc` and returns a `NodeStorage`

EclesioMeloJunior (changes requested):

dot/core/test_helpers.go
edwardmack: I created this:
```
func InitNodeStorage(baseDB runtime.BasicStorage, dataDir string) (runtime.NodeStorage, error) {
    nodeStorage := runtime.NodeStorage{}
    var err error
    if baseDB != nil {
        nodeStorage.BaseDB = baseDB
    } else {
        nodeStorage.BaseDB, err = utils.SetupDatabase(filepath.Join(dataDir, "offline_storage"), false)
    }
    return nodeStorage, err
}
```
however it seems that I could find a place to put this where it wouldn't cause a cycle import error in one of the files that used it.

EclesioMeloJunior: where did you place it? If you could not find a place to put there is no problem to keep the code as it is.

EclesioMeloJunior (approved):

dot/core/test_helpers.go
edwardmack: Tried using it in core/test_helpers.go, then sync/test_helpers.go and babe/babe_test.go

edwardmack:
Can someone override the requirement for DeepSource on this PR, it's complaining that there are unused parameters for the methods in wamer/imports.go, which is true, however when I try to remove them I got these errors when compiling: 
```
_cgo_export.c:67:61: error: conflicting types for ‘_’
   67 | void ext_transaction_index_index_version_1(void* _, int32_t _, int32_t _, int32_t _)
      |                                                     ~~~~~~~~^
_cgo_export.c:67:50: note: previous definition of ‘_’ was here
   67 | void ext_transaction_index_index_version_1(void* _, int32_t _, int32_t _, int32_t _)
      |                                            ~~~~~~^
_cgo_export.c:67:72: error: redefinition of parameter ‘_’
   67 | void ext_transaction_index_index_version_1(void* _, int32_t _, int32_t _, int32_t _)
      |                                                                ~~~~~~~~^
_cgo_export.c:67:61: note: previous definition of ‘_’ was here
   67 | void ext_transaction_index_index_version_1(void* _, int32_t _, int32_t _, int32_t _)
      |                                                     ~~~~~~~~^
_cgo_export.c:67:83: error: redefinition of parameter ‘_’
   67 | void ext_transaction_index_index_version_1(void* _, int32_t _, int32_t _, int32_t _)
      |                                                                           ~~~~~~~~^
_cgo_export.c:67:72: note: previous definition of ‘_’ was here
   67 | void ext_transaction_index_index_version_1(void* _, int32_t _, int32_t _, int32_t _)
      |                                                                ~~~~~~~~^
_cgo_export.c: In function ‘ext_transaction_index_index_version_1’:

...
```

Issue #1508, opened 2021-04-06
implement remaining offchain worker API


noot:
## Task summary

- implement remaining offchain worker API, `ext_offchain_timestamp`, `ext_offchain_sleep_until` , `ext_offchain_random_seed`
- currently have persistent and local storage, but could update this to be part of a some `OffchainWorker` interface 

## Specification reference

-  see spec D.5. Offchain

omar391:
PR is merged, hence closed!

PR #1774, opened 2021-08-31
feat(dot/rpc) implement `offchain_localStorageSet` and `offchain_localStorageGet`

### LINKED ISSUE #1762: Implement `offchain_localStorageGet` RPC call
## Task summary

- Implement `offchain_localStorageGet` RPC call

## Specification reference

- [Spec reference](https://github.com/w3f/PSPs/blob/master/PSPs/drafts/psp-6.md#1104-offchain_localstorageget)

- N/A

### LINKED ISSUE #1763: Implement `offchain_localStorageSet` RPC call
## Task summary

- Implement `offchain_localStorageSet` RPC call

## Specification reference

- [Spec reference](https://github.com/w3f/PSPs/blob/master/PSPs/drafts/psp-6.md#1103-offchain_localstorageset)

- N/A

EclesioMeloJunior:
## Changes

- Create `modules.Offchain` RPC module
-  Implement `offchain_localStorageSet` and `offchain_localStorageGet` using the `runtime.NodeStorage` implementation

## Tests

```
go test github.com/ChainSafe/gossamer/dot/rpc/modules
```

## Issues

- Closes #1762 
- Closes #1763 

## Primary Reviewer 

- @jimjbrettj


jimjbrettj (approved):
Looks good to me!

arijitAD (commented):
Looks good. I just had a question.

I read the following from spec regarding the local storage.
> Local storage is revertible and fork-aware. It means that any value set by the oﬀchain worker triggered at a certain block is reverted if that block is reverted as non-canonical. The value is NOT available for the worker that is re-run at the next or any future blocks.

How are we making sure the values of local storage are reverted for non-canonical blocks?
I think we would need to make local storage for every fork.

arijitAD (approved):

EclesioMeloJunior:
> Looks good. I just had a question.
> 
> I read the following from spec regarding the local storage.
> 
> > Local storage is revertible and fork-aware. It means that any value set by the oﬀchain worker triggered at a certain block is reverted if that block is reverted as non-canonical. The value is NOT available for the worker that is re-run at the next or any future blocks.
> 
> How are we making sure the values of local storage are reverted for non-canonical blocks?
> I think we would need to make local storage for every fork.

I can open an issue to check and, if necessary, implement the behaviour

lib/runtime/types.go
```diff
@@ -37,6 +37,26 @@ type NodeStorage struct {
...
+// SetPersistent persists a key and value into PERSISTENT node storage
+func (n *NodeStorage) SetPersistent(k, v []byte) error {
+	return n.PersistentStorage.Put(k, v)
+}
+
+// GetPersistent retrieve a key and value from PERSISTENT node storage
+func (n *NodeStorage) GetPersistent(k []byte) ([]byte, error) {
+	return n.PersistentStorage.Get(k)
+}
+
```
noot, line 59: it might be nice to refactor the offchain storage into its own API/package that both the runtime and the RPC can access, this doesn't need to be done here but maybe in the future. there's an issue for it: #1508

EclesioMeloJunior: I've outlined this in the issue #1776

cmd/gossamer/main.go
```diff
@@ -241,7 +241,7 @@ func gossamerAction(ctx *cli.Context) error {
 	cfg.Global.BasePath = utils.ExpandDir(cfg.Global.BasePath)
 
 	// check if node has not been initialised (expected true - add warning log)
```
noot, line 243: update comment

noot (approved):
looks good!

jimjbrettj (approved):
Looks good!

Issue #1745, opened 2021-08-23
Integrate scale package into dot/digest


jimjbrettj:
## Task summary

- Replace all occurrences of the old scale library with the new package
- Remove custom receiver functions concerning encoding/decoding where possible
- Integrate the new scale types from dot/types

## Specification reference

- https://github.com/w3f/polkadot-spec/releases/download/v0.1.0-beta1/polkadot-host-spec_v0.1.0-beta1.pdf

## Other information and links

- https://github.com/ChainSafe/gossamer/tree/tim/scale/pkg/scale
- https://github.com/paritytech/parity-scale-codec
- https://substrate.dev/docs/en/knowledgebase/advanced/codec

Issue #1743, opened 2021-08-23
Integrate scale package into dot types


jimjbrettj:
## Task summary

- Create new types in dot/types that are compatible with the new scale package
- Rewrite tests and methods where old types are used
- Integrate the new types into the codebase

## Specification reference

- https://github.com/w3f/polkadot-spec/releases/download/v0.1.0-beta1/polkadot-host-spec_v0.1.0-beta1.pdf

## Other information and links

- https://github.com/ChainSafe/gossamer/tree/tim/scale/pkg/scale
- https://github.com/paritytech/parity-scale-codec
- https://substrate.dev/docs/en/knowledgebase/advanced/codec

PR #1779, opened 2021-09-07
chore (pkg/scale): refactor the types package to utilize new scale package

jimjbrettj:
## Changes

- Wrote new types for digest, consensusDigest, header, blockData, and block that are compatible with the new scale package
- Integrated scale and new types into dot/digest and dot/network
- Finished the integration of the new block type, so in this PR I am looking specifically for feedback on dot/digest and dot/network integration, the new block type, as well as the other new types that are still a work in progress. Also am open to suggestions on changes to my approach to this integration
- I created a temporary branch to merge into as a staging branch before I merge into dev so that I can get more steady feedback on my work, as well as hopefully make the reviewing process a little easier given the size of this refactor

## Tests

```
- go test ./dot/types -v -short
- go test ./dot/network -v -short
- go test ./dot/digest -v -short
```

## Issues

- #1745 
- #1744 
- #1743 

## Primary Reviewer 

- @timwu20


dot/core/interface.go
```diff
@@ -58,7 +60,8 @@ type StorageState interface {
 	LoadCode(root *common.Hash) ([]byte, error)
 	LoadCodeHash(root *common.Hash) (common.Hash, error)
 	TrieState(root *common.Hash) (*rtstorage.TrieState, error)
-	StoreTrie(*rtstorage.TrieState, *types.Header) error
+	//StoreTrie(*rtstorage.TrieState, *types.Header) error
+	StoreTrieVdt(*rtstorage.TrieState, *types.HeaderVdt) error
```
timwu20, line 64: Are you going to replace `type.Header` and revert to the original function names?

jimjbrettj: Yup exactly. Anything with "new" or "vdt" appended to the name is temporary and will be changed back to the original naming after the integration.

timwu20: can you do the renames I think it'll make the diff look much nicer.

jimjbrettj: Absolutely I agree. I am actually working on this currently

PR #1780, opened 2021-09-08
fix(dot/core): Batch process transaction message.

### LINKED ISSUE #1781: Batch process transaction messages.
## Issue summary

After execution of **nth** block **(n+1)th** block throws below error `failed to find root key`.
```logs=
WARN[09-01|02:37:45] failed to load state for block           pkg=sync block=0xdb367756f79d5269fbf4443b9b3cc09599121adb911abaacc585f533eae5dfd1 error="failed to find root key=0xd9dd920475455acfb5f8a3d7396586ed5d00311e54ea4ac80f7375862ab34023: Key not found" caller=syncer.go:203
```
**Analysis:** There is parallel execution of [HandleTransactionMessage()](https://github.com/ChainSafe/gossamer/blob/development/dot/core/messages.go#L28) and [handleblock()](https://github.com/ChainSafe/gossamer/blob/development/dot/sync/syncer.go#L346) for nth block while syncing. The trie state is getting modified by [HandleTransactionMessage()](https://github.com/ChainSafe/gossamer/blob/development/dot/core/messages.go#L28) by **(n-1)th** block. HandelBlockImport() will store incorrect of data **(n-1)th** block trie in DB instead of nth block trie and import nth block successfully. On **executing (n+1)th block** throws an error in [TrieState()](https://github.com/ChainSafe/gossamer/blob/development/dot/state/storage.go#L162) due to absence of **nth** trie storage in db.

**Suggestion:** We should acquire the lock on `storageState` at all places where we are setting storage context for runtime. But this may cause heavy lock contention since transaction messages are broadcasted frequently.  
Instead, we should batch process transaction messages after every regular time interval.

-

arijitAD:
## Changes

- Batch process transaction message

## Tests

```
TestCreateNotificationsMessageHandler_HandleTransaction
```

## Issues

- closes #1781 

## Primary Reviewer 

- @noot


dot/network/notifications.go
```diff
@@ -53,8 +53,16 @@ type (
...
 	NotificationsMessageHandler = func(peer peer.ID, msg NotificationsMessage) (propagate bool, err error)
+
+	// NotificationsMessageBatchHandler is called when a (non-handshake) message is received over a notifications stream in batch processing mode.
+	NotificationsMessageBatchHandler = func(peer peer.ID, msg NotificationsMessage) (batchMsgs []*transactionBatchMessage, err error)
 )
 
+type transactionBatchMessage struct {
+	msg  NotificationsMessage
+	peer peer.ID
+}
```
noot, line 64: is this for all notifications messages, or only transactions messages?

dot/network/service.go
```diff
@@ -218,12 +219,46 @@ func (s *Service) Start() error {
...
+		nil,
 		false,
 	)
 	if err != nil {
 		logger.Warn("failed to register notifications protocol", "sub-protocol", blockAnnounceID, "error", err)
 	}
 
+	txnBatch := make(chan *transactionBatchMessage, batchSize)
+
+	txnBatchHandler := func(peer peer.ID, msg NotificationsMessage) (msgs []*transactionBatchMessage, err error) {
```
noot, line 231: I'd separate this into a separate function and put it in `transactions.go` and add some unit tests (seems like it's not covered for some reason?)

dot/network/notifications.go
noot: seems like it could be generic, so maybe rename to `batchMessage`?

dot/network/service.go
```diff
@@ -49,6 +49,7 @@ const (
 	maxMessageSize = 1024 * 63 // 63kb for now
 
 	gssmrIsMajorSyncMetric = "gossamer/network/is_major_syncing"
+	batchSize              = 100
```
noot, line 52: this is ok for now but we should probably be propagating txs within a time period that's less than the block time, so maybe we should check how many tx messages we get on average within a 6 second timespan

we also need the txs to be processed within a block time span, otherwise we might end up adding txs to our pool that have already been included, because we ended up validating them too late. maybe we can open another issue for this

noot (commented):
overall looks good, just a couple comments and thoughts! I'm also wondering if it would make more sense to have the batching in the tx handler function, since I'm not sure if the other message types would require batching, but I guess it's nice to have in case we do.

dot/network/notifications.go
arijitAD: Done

dot/network/service.go
arijitAD: Done. Previously we had a different handler for the test. I have updated the code to use the same handler.

dot/network/service.go
arijitAD: Yes, I was initially planning to do it time-based. It will require changing the code flow as we need to do the processing async. I will create a separate issue for it.

dot/network/transaction.go
```diff
@@ -119,6 +119,40 @@ func decodeTransactionHandshake(_ []byte) (Handshake, error) {
 	return &transactionHandshake{}, nil
 }
 
+func (s *Service) createBatchMessageHandler(txnBatch chan *batchMessage) NotificationsMessageBatchHandler {
```
noot, line 122: does `txnBatch` channel need to be passed in here? seems it's only used within this function, so can it be declared in this function>

arijitAD: Yes, It can be but we need to provide the channel from outside for the test cases.

arijitAD:
Deep source errors are fixed here https://github.com/ChainSafe/gossamer/pull/1789

noot (approved):
great work :)

dot/network/transaction.go
```diff
@@ -119,6 +119,40 @@ func decodeTransactionHandshake(_ []byte) (Handshake, error) {
...
+	txnBatchHandler := func(peer peer.ID, msg NotificationsMessage) (msgs []*batchMessage, err error) {
+		data := &batchMessage{
+			msg:  msg,
+			peer: peer,
+		}
+		txnBatch <- data
```
EclesioMeloJunior, line 128: could we check if the channel is closed before send data?

dot/network/transaction.go
EclesioMeloJunior: nit: we could receive a `txnBatch chan<- *batchMessage`

dot/network/transaction.go
```diff
@@ -119,6 +119,40 @@ func decodeTransactionHandshake(_ []byte) (Handshake, error) {
 	return &transactionHandshake{}, nil
 }
 
+func (s *Service) createBatchMessageHandler(txnBatch chan *batchMessage) NotificationsMessageBatchHandler {
+	txnBatchHandler := func(peer peer.ID, msg NotificationsMessage) (msgs []*batchMessage, err error) {
```
EclesioMeloJunior, line 123: could we return the function directly?

EclesioMeloJunior (changes requested):

dot/network/transaction.go
arijitAD: This channel will never be closed.

dot/network/transaction.go
arijitAD: Updated the code to return function. 
We cannot make the channel just receiver as we are also reading data out from it inside the callback.

EclesioMeloJunior (approved):

PR #1783, opened 2021-09-09
fix: fix Kusama sync; add storageState lock in core.HandleTransactionMessage

### LINKED ISSUE #1770: Kusama Node not syncing shortly after launch
## Issue summary

- When syncing to the Kusama network block syncing is paused based on the metrics and logs.  The following log message occurs before ultimately failing.  I've tried it a number of times and it's not block specific.
```
EROR[08-25|15:34:51] failed to handle block                   pkg=sync number=8939114 error="failed to find root key=0xb292a2a28d5a740f01852adc01227c6b91ca276446abec6a01c933b9b049717a: Key not found" caller=syncer.go:265
WARN[08-25|15:34:51] failed to handle block data              pkg=network failed on block=4353 error="failed to find root key=0xb292a2a28d5a740f01852adc01227c6b91ca276446abec6a01c933b9b049717a: Key not found" caller=sync.go:815
```
- Noticed this on the staging nodes for both Kusama and Polkadot.  The block sync number just stalls and the node continues to run, but never updates block sync progress.

- Run `gossamer --chain=kusama init` then `gossamer --chain=kusama --publish-metrics` and observe logs and `service_blocks_sync` metric.

-

noot:
## Changes

-  add storageState lock in core.HandleTransactionMessage to use while validating transaction
- there was also a race condition / edge case between calling `GetRuntime` and `TrieState` as the best block might change between these two calls, and if the runtime changes between the two calls this might be an issue. so it's updated to get the best block hash at the start of the func and use that
- note: this seems to somewhat improve the issue but not fully solve it, as there are still bugs in the current `syncQueue` (syncing refactor appears address this, currently running a node to confirm)

## Tests

```
make gossamer
./bin/gossamer --chain kusama init --force
./bin/gossamer --chain kusama
```

## Issues

- closes #1770 

## Primary Reviewer 

- @timwu20


arijitAD (approved):

dot/core/messages.go
```diff
@@ -32,42 +32,58 @@ func (s *Service) HandleTransactionMessage(msg *network.TransactionMessage) (boo
...
 		return false, err
 	}
 
 	for _, tx := range txs {
-		ts, err := s.storageState.TrieState(nil)
-		if err != nil {
-			return false, err
-		}
+		err = func() error {
+			s.storageState.Lock()
```
arijitAD, line 48: Should we move this lock outside of the loop?

noot: I figured to put it inside the loop since if it locks outside the loop instead of for each transaction, it might cause more lock contention

dot/core/interface.go
```diff
@@ -61,6 +61,8 @@ type StorageState interface {
 	StoreTrie(*rtstorage.TrieState, *types.Header) error
 	GetStateRootFromBlock(bhash *common.Hash) (*common.Hash, error)
 	GetStorage(root *common.Hash, key []byte) ([]byte, error)
+	Lock()
+	Unlock()
```
timwu20, line 65:
```suggestion
	sync.Locker
```

timwu20 (changes requested):

dot/core/interface.go
noot: updated

timwu20 (approved):

PR #1786, opened 2021-09-13
chore(CI): Check PR body and title.

arijitAD:
## Changes

- Check PR body and title.

## Tests

```
Test_PR_Checks
```

## Issues

- None

## Primary Reviewer 

- @noot

### DIFF
--- a/.github/PULL_REQUEST/pull_request.go
+++ b/.github/PULL_REQUEST/pull_request.go
@@ -1,0 +1,16 @@
+package main
+
+import (
+	"os"
+
+	"github.com/ChainSafe/gossamer/lib/utils"
+)
+
+func main() {
+	title := os.Getenv("RAW_TITLE")
+	body := os.Getenv("RAW_BODY")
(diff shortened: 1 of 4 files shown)


noot (approved):
looks good - potentially we could check that the title starts with one of "chore", "fix", "maintainence", "feature" etc. as well. 

I'd prefer to leave this unrequired for now as well, in case of external contributors perhaps deviating from the format (they shouldn't, but it might be confusing for them). we can discuss more on standup though!

arijitAD:
> looks good - potentially we could check that the title starts with one of "chore", "fix", "maintainence", "feature" etc. as well.
> 
> I'd prefer to leave this unrequired for now as well, in case of external contributors perhaps deviating from the format (they shouldn't, but it might be confusing for them). we can discuss more on standup though!

Yes, I feel the same. It should not be a mandatory check but it just automates the check for proper PR title and body.

EclesioMeloJunior (approved):

PR #1799, opened 2021-09-21
feat: implement ext_default_child_storage_storage_kill_version_2

### LINKED ISSUE #1677: runtime v0.9: implement ext_default_child_storage_storage_kill_version_2
## Task summary

- implement `ext_default_child_storage_storage_kill_version_2` imported function in wasmer which kills a child storage trie

## Specification reference

-  see spec D.2.5

edwardmack:
## Changes

- Implement code for `ext_default_storage_storage_kill_version_2`
- NOTE: this is a draft PR because `Test_ext_default_child_storage_storage_kill_version_2_limit_none` fails when calling ext with optional limit parameter set to false.  I'm not sure I've setup the reflect wasm function (in https://github.com/noot/polkadot-spec/pull/1) to handle optional parameters correctly.  Please advise.
-

## Tests

```
go test -run ^Test_ext_default_child_storage_storage_kill_version_2.*$ github.com/ChainSafe/gossamer/lib/runtime/wasmer/... -v
```

## Issues

- closes #1677 

## Primary Reviewer 

- @noot


lib/runtime/wasmer/imports.go
```diff
@@ -1120,20 +1120,34 @@ func ext_default_child_storage_storage_kill_version_1(context unsafe.Pointer, ch
...
+	}
+
+	_, all, err := storage.DeleteChildLimit(childStorageKey, limit)
+	if err != nil {
+		logger.Warn("[ext_default_child_storage_storage_kill_version_2] cannot get child storage", "error", err)
+	}
+
+	if all {
+		return 1
+	} else {
```
EclesioMeloJunior, line 1148: You could do `return 0` directly, this `else` is not necessary

EclesioMeloJunior (changes requested):

lib/runtime/constants.go
```diff
@@ -40,9 +40,11 @@ const (
...
 
 	// v0.8 test API wasm
-	HOST_API_TEST_RUNTIME     = "hostapi_runtime"
-	HOST_API_TEST_RUNTIME_FP  = "hostapi_runtime.compact.wasm"
-	HOST_API_TEST_RUNTIME_URL = "https://github.com/noot/polkadot-spec/blob/master/test/hostapi_runtime.compact.wasm?raw=true"
+	HOST_API_TEST_RUNTIME    = "hostapi_runtime"
+	HOST_API_TEST_RUNTIME_FP = "hostapi_runtime.compact.wasm"
+	//HOST_API_TEST_RUNTIME_URL = "https://github.com/noot/polkadot-spec/blob/master/test/hostapi_runtime.compact.wasm?raw=true"
+	// todo (ed) use above URL once ed/add_rtm_ext_default_child_storage_storage_kill_version_2 has been merged to master
+	HOST_API_TEST_RUNTIME_URL = "https://github.com/noot/polkadot-spec/blob/ed/add_rtm_ext_default_child_storage_storage_kill_version_2/test/hostapi_runtime.compact.wasm?raw=true"
```
noot, line 47: is this ready to be merged?

lib/runtime/storage/trie.go
```diff
@@ -178,6 +180,43 @@ func (s *TrieState) DeleteChild(key []byte) {
 	s.t.DeleteChild(key)
 }
 
+// DeleteChildLimit deletes up to limit of database entries by lexicographic order, return number
+//  deleted, true if all delete atherwise false
+func (s *TrieState) DeleteChildLimit(key []byte, limit *optional.Bytes) (uint32, bool, error) {
```
noot, line 185: I don't think this is correct, as per substrate this deletes up to `limit` keys from the child trie which is located at `key`: https://github.com/paritytech/substrate/blob/4c3a55e7ca5c4c85c1eb53fd82ed71029d952510/primitives/externalities/src/lib.rs#L131 so, you'll need to get a list of all the keys from the child trie at `key` then delete up to `limit` of then, and return true if the entire child was deleted, false otherwise

lib/runtime/storage/trie.go
```diff
@@ -178,6 +180,43 @@ func (s *TrieState) DeleteChild(key []byte) {
...
+	s.lock.Lock()
+	defer s.lock.Unlock()
+	tr, err := s.t.GetChild(key)
+	if err != nil {
+		return 0, false, err
+	}
+	qtyEntries := uint32(len(tr.Entries()))
+	if limit == nil || !limit.Exists() {
+		s.t.DeleteChild(key)
```
noot, line 194: I think this should be `DeleteFromChild`, after firstly getting the list of keys in the child

lib/runtime/constants.go
edwardmack: Yes, PR #2 on noot/polkadot-spec is ready to merge (if it looks good to you).

lib/runtime/wasmer/imports.go
edwardmack: updated.

lib/runtime/storage/trie.go
edwardmack: I don't understand which part if this is not correct?  First I'm checking if limit is set, then deleting all if not set.  Otherwise I'm using the limit, then getting a list of keys and deleting up up to limit.  Am I missing something between DeleteFromChild and DeleteChild?  

These tests: `Test_ext_default_child_storage_storage_kill_version_2_limit_1` and `Test_ext_default_child_storage_storage_kill_version_2_limit_none` are giving me expected results, did I set these tests up incorrectly? Or are this passing because I've miss-understood the spec?  
```
	tr, err := s.t.GetChild(key)
	if err != nil {
		return 0, false, err
	}
	qtyEntries := uint32(len(tr.Entries()))
	if limit == nil || !limit.Exists() {  // if limit nil or limit !exists then delete all entries at that key
		s.t.DeleteChild(key)
		return qtyEntries, true, nil
	}
	limitUint := binary.LittleEndian.Uint32(limit.Value())

        // limit not nil, so make list of keys (and sort)
	keys := make([]string, 0, len(tr.Entries()))
	for k := range tr.Entries() {
		keys = append(keys, k)
	}
	sort.Strings(keys)
	deleted := uint32(0)
        // delete up to limit quantity
	for _, k := range keys {
		tr.Delete([]byte(k))
		deleted++
		if deleted == limitUint {
			break
		}
	}

	if deleted == qtyEntries {
		return deleted, true, nil
	}

	return deleted, false, nil

```

noot: ok so sorry, I totally missed that you got `tr.Entries()` in a following line. ignore this comment

lib/runtime/constants.go
noot: merged!

EclesioMeloJunior (approved):

PR #1807, opened 2021-09-24
fix(dot/types): *types.Body to be of type []types.Extrinsic

### LINKED ISSUE #1775: update `*types.Body` to be of type `[]types.Extrinsic`
## Task summary

- currently the block body is of type `[]byte`, and we assume what is placed in it is a SCALE encoded array of extrinsics: ie `[]types.Extrinsic`
- however, this can cause confusion, and the type should be what it really is, which is `[]types.Extrinsic`

Ref: https://substrate.dev/docs/en/knowledgebase/advanced/codec

kishansagathiya:
## Changes
Changing Body from []byte to []Extrinsic, changes its meaning (encoded to not encoded), so the right way to do this should be as follows

- [x] Create a new Struct BodyExtrinsics with its own methods
- [x] Use BodyExtrinsics instead of Body everywhere and changes/use methods accordingly
- [x] Now that old Body struct is not being used anywhere, remove it.
- [x] Rename BodyExtrinsics to Body and accordingly rename its methods

## Tests

```

```

## Issues

- Closes #1775

## Primary Reviewer 

-


dot/types/body.go
```diff
@@ -24,16 +24,28 @@ import (
...
+// is assumed to be scale encoded.
+func NewBody(b []byte) (*Body, error) {
+	exts := [][]byte{}
+
+	if len(b) == 0 {
+		body := Body([]Extrinsic{})
+		return &body, nil
+	}
+
+	dec, err := scale.Decode(b, exts)
```
noot, line 42: please use the new scale package located at `pkg/scale`!

dot/types/body_extrinsics.go
```diff
@@ -23,33 +23,42 @@ import (
 
 	"github.com/ChainSafe/gossamer/lib/common"
 	"github.com/ChainSafe/gossamer/lib/common/optional"
-	"github.com/ChainSafe/gossamer/lib/scale"
```
arijitAD, line 26: Don't rename the file. I think `body.go` is fine.

dot/types/body_extrinsics.go
```diff
@@ -23,33 +23,42 @@ import (
...
-// Body is the encoded extrinsics inside a state block
-type Body []byte
+// Body is the extrinsics(not encoded) inside a state block.
+type Body []Extrinsic
 
-// NewBody returns a Body from a byte array
-func NewBody(b []byte) *Body {
-	body := Body(b)
+// NewBody returns a BodyExtrinsics from an Extrinsic array.
+func NewBody(e []Extrinsic) *Body {
```
arijitAD, line 33: Any reason for the body to be a pointer? Since it is already a slice.

dot/types/body_extrinsics.go
```diff
@@ -23,33 +23,42 @@ import (
...
+	body := Body(e)
 	return &body
 }
 
-// NewBodyFromBytes returns a new Body from a slice of byte slices
-func NewBodyFromBytes(exts [][]byte) (*Body, error) {
-	enc, err := scale.Encode(exts)
+// NewBodyFromBytes returns a BodyExtrinsics from a SCALE encoded byte array.
+func NewBodyFromBytes(b []byte) (*Body, error) {
+	exts := [][]byte{}
```
arijitAD, line 40: `var exts []Extrinsic`

dot/state/block.go
```diff
@@ -354,12 +355,16 @@ func (bs *BlockState) GetBlockBody(hash common.Hash) (*types.Body, error) {
...
 }
 
 // SetBlockBody will add a block body to the db
 func (bs *BlockState) SetBlockBody(hash common.Hash, body *types.Body) error {
-	return bs.db.Put(blockBodyKey(hash), body.AsOptional().Value())
+	optionalBody, err := body.AsOptional()
+	if err != nil {
+		return err
+	}
+	return bs.db.Put(blockBodyKey(hash), optionalBody.Value())
```
noot, line 367: I think this can be simplified to just store the body directly, as the value of the optional is the same as `body`:
```suggestion
	return bs.db.Put(blockBodyKey(hash), body))
```

dot/types/block.go
```diff
@@ -54,8 +54,12 @@ func (b *Block) Encode() ([]byte, error) {
 		return nil, err
 	}
 
-	// block body is already SCALE encoded
-	return append(enc, []byte(b.Body)...), nil
+	// get a SCALE encoded block body
+	encodedBody, err := b.Body.AsSCALEEncodedBody()
+	if err != nil {
+		return nil, err
+	}
```
noot, line 61: I think you can probably just do `scale.Marshal(b.Body)`

dot/types/body_extrinsics.go
```diff
@@ -23,33 +23,42 @@ import (
...
 
-// Body is the encoded extrinsics inside a state block
-type Body []byte
+// Body is the extrinsics(not encoded) inside a state block.
+type Body []Extrinsic
 
-// NewBody returns a Body from a byte array
-func NewBody(b []byte) *Body {
-	body := Body(b)
+// NewBody returns a BodyExtrinsics from an Extrinsic array.
```
noot, line 32:
```suggestion
// NewBody returns a Body from an Extrinsic array.
```

dot/types/body_extrinsics.go
```diff
@@ -86,73 +84,31 @@ func NewBodyFromExtrinsicStrings(ss []string) (*Body, error) {
...
-
-	for i, ext := range decodedExts {
-		ret[i], err = scale2.Marshal(ext)
-		if err != nil {
-			return nil, err
-		}
-	}
+func (b *Body) DeepCopy() Body {
+	newExtrinsics := make([]Extrinsic, len([]Extrinsic(*b)))
+	copy(newExtrinsics, []Extrinsic(*b))
```
noot, line 101: I m not sure if this deep copies the extrinsics, as `Extrinsic` is `[]byte` which is essentially a pointer, so it's copying the pointers from one slice to another. I think you'll probably need a for loop to copy each extrinsic directly, but maybe double check?

noot (commented):

dot/types/body_extrinsics.go
kishansagathiya: I have kept what was being used previously. 
Not using pointer would be safer in general. Changing this from pointer, would mean edit this at a lot places. But, I think doing so would be the right way. Will change it.

kishansagathiya: At most place where body is returned it is returned as a pointer.

So if passed without pointer to the slice (`Body` and not `*Body`), you can modify it, but to an extent. From what I read, you can modify, values but not length of the slice. If you try to append to it, slice won't have values that you expected. So far, I have not seen anywhere `Body` getting modified.

I also see that at places a pointer to Body and a pointer is stored for even other slices.
```
type BlockData struct {
	Hash          common.Hash
	Header        *Header
	Body          *Body
	Receipt       *[]byte
	MessageQueue  *[]byte
	Justification *[]byte
}
```
@noot is returning or storing `*Body` (and also other slices, like `MessageQueue *[]byte`) done deliberately? Should I care returning `Body` instead of `*Body`?

noot: all the fields in the BlockData should be a pointer as they are optional types. please leave this as is

noot: actually, for this functions specifically, I'm not sure if it really matters. I think it could be either

dot/types/body.go
```diff
@@ -86,73 +83,29 @@ func NewBodyFromExtrinsicStrings(ss []string) (*Body, error) {
...
-}
-
-// AsOptional returns the Body as an optional.Body
-func (b *Body) AsOptional() *optional.Body {
-	ob := optional.CoreBody([]byte(*b))
-	return optional.NewBody(true, ob)
+	return Body(newExtrinsics)
 }
 
 // HasExtrinsic returns true if body contains target Extrisic
```
arijitAD, line 103: type `Extrisic` -> `Extrinsic`

dot/rpc/modules/chain_test.go
```diff
@@ -140,7 +140,7 @@ func TestChainGetBlock_Genesis(t *testing.T) {
 	expected := &ChainBlockResponse{
 		Block: ChainBlock{
 			Header: *expectedHeader,
-			Body:   nil,
+			Body:   []string{"0x2800010203040506070809"},
```
arijitAD, line 143: Can you make this value constant and add a comment explaining this value?

dot/types/body.go
```diff
@@ -88,7 +88,7 @@ func NewBodyFromExtrinsicStrings(ss []string) (*Body, error) {
 
 // DeepCopy creates a new copy of the body.
 func (b *Body) DeepCopy() Body {
-	newExtrinsics := make([]Extrinsic, len([]Extrinsic(*b)))
+	newExtrinsics := []Extrinsic{}
```
arijitAD, line 91: You can do
```
newExtrinsics := make([]Extrinsic, 0, len([]Extrinsic(*b)))
```
This is much more optimized.

arijitAD (approved):

noot:
@kishansagathiya could you resolve the conflicts when you get a chance?

dot/types/body_test.go
```diff
@@ -17,97 +17,61 @@
...
 
-	for _, test := range tests {
-		res, err := test.body.HasExtrinsic(target)
-		require.Nil(t, err)
-		require.Equal(t, test.expect, res)
+	for _, ext := range exts {
+		extStrings = append(extStrings, common.BytesToHex(ext))
 	}
+
+	fmt.Println(extStrings)
```
noot, line 70: remove

noot (approved):
nice work :D

dot/types/body.go
```diff
@@ -174,3 +127,19 @@ func (b *Body) HasExtrinsic(target Extrinsic) (bool, error) {
...
 }
+
+// AsEncodedExtrinsics decodes the body into an array of SCALE encoded extrinsics
+func (b *Body) AsEncodedExtrinsics() ([]Extrinsic, error) {
+	decodedExts := *b
+	ret := make([]Extrinsic, len(decodedExts))
+	var err error
+
+	for i, ext := range decodedExts {
+		ret[i], err = scale.Marshal(ext)
```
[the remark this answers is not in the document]
kishansagathiya: The marshaling part is correct, but I think this function should be better named as `AsEncodedBytesArray`. That would avoid the confusion that it is causing. And the value returned by this function is being converted into `[][]byte` and `[]string` anyway, so returning `[][]byte` is better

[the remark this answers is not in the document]
jimjbrettj: Are we sure that the marshalling here is correct? On my branch jimmy/integrateScaleIntoTypes I had my own tests (https://github.com/ChainSafe/gossamer/blob/0cb69f5f2946a36c3fb7982102af274f2d2b3e50/dot/types/block_test.go#L111) for block encoding written which I compare the encoding to the expected byte string (hard coded so I don't test against itself) and when I merge these changes that test fails because the encoding doesn't equal what is expected. So there appears to be difference between the encodings, but I am not certain what the correct format is. For example, the body.AsEncodedExtrinsics() function specifies that it is scale encoded but that is not actually the case because there is not length prefix for the encoded slice. I am not positive if it should be scale encoded or if its just incorrect documentation but it does appear that the body encoding now is different than it was before. Perhaps @noot could help provide some clarification on this?

PR #1800, opened 2021-09-21
feat(dot/rpc): Implement `childstate_getKeys` rpc call

EclesioMeloJunior:
## Changes

- Add `childstate.go` file to place the `ChildStateModule`
- Implement `GetKeys` RPC method

## Tests

```
go test -run ^TestChildStateGetKeys$ github.com/ChainSafe/gossamer/dot/rpc/modules
```

## Issues

- Closes #1753

## Primary Reviewer 

- @edwardmack

PR #1810, opened 2021-09-24
feat(rpc): Implement `childstate_getStorageSize` RPC call

### LINKED ISSUE #1756: Implement `childstate_getStorageSize` RPC call
## Task summary

- Implement `childstate_getStorageSize` RPC call

## Specification reference

- [Spec reference](https://github.com/w3f/PSPs/blob/master/PSPs/drafts/psp-6.md#1124-childstate_getstoragesize)

- N/A

EclesioMeloJunior:
## Changes

- Create the `childstate.GetStorageSizer` method to get the size of a child storage entry

## Tests

```
go test -run ^TestChildStateGetStorageSize$ github.com/ChainSafe/gossamer/dot/rpc/modules
```

## Issues

- Closes #1756

## Primary Reviewer 

- @timwu20


EclesioMeloJunior:
Depends on PR #1800

dot/rpc/modules/childstate.go
```diff
@@ -68,3 +75,23 @@ func (cs *ChildStateModule) GetKeys(_ *http.Request, req *GetKeysRequest, res *[
...
+	if err != nil {
+		return err
+	}
+
+	item, err := cs.storageAPI.GetStorageFromChild(stateRoot, req.KeyChild, req.EntryKey)
+	if err != nil {
+		return err
+	}
+
+	*res = uint64(len(item))
```
noot, line 95: I think the return value here needs to be an optional, so if the `item == nil` then it returns `None` otherwise it returns `Some(len(item))`. currently if the item is nil or if the item has length=0 they'll both return 0

timwu20: @EclesioMeloJunior does this need to change?

EclesioMeloJunior: @timwu20 updated!

edwardmack (approved):

dot/rpc/modules/childstate.go
```diff
@@ -29,6 +29,13 @@ type GetKeysRequest struct {
 	Hash   common.Hash
 }
 
+// GetChildStorageRequest the request to get the entry child storage hash
+type GetChildStorageRequest struct {
+	KeyChild []byte
+	EntryKey []byte
+	Hash     common.Hash
```
noot, line 36: can this be a `*common.Hash` instead as the block hash is optional?

EclesioMeloJunior: Done!

arijitAD (approved):

PR #1811, opened 2021-09-24
chore(lib/grandpa): integrate scale into grandpa

jimjbrettj:
## Changes

- Remove the old scale library from grandpa
- Remove as much custom serialization logic as possible from grandpa

## Tests

```
go test ./dot/types -v
go test ./dot/state -v -short
go test ./lib/grandpa -v -short
```

## Issues

- #1747 

## Primary Reviewer 

- @timwu20


dot/state/grandpa_test.go
```diff
@@ -43,6 +43,10 @@ func TestNewGrandpaStateFromGenesis(t *testing.T) {
 	require.NoError(t, err)
 	require.Equal(t, genesisSetID, currSetID)
 
+	//testAuths2 := []types.GrandpaVoter{
+	//	{Key: *kr.Alice().Public().(*ed25519.PublicKey), ID: 0},
+	//}
```
noot, line 48: remove

dot/types/grandpa.go
```diff
@@ -111,19 +111,32 @@ func (v *GrandpaVoter) Decode(r io.Reader) error {
...
+	v.Key = *key
 	v.ID = id
 	return nil
 }
 
+// Encode will encode the GrandpaVoter
+func (v *GrandpaVoter) Encode() ([]byte, error) {
+	enc := []byte{}
+	b := v.Key.Encode()
+	enc = append(enc, b...)
```
noot, line 123: does the new SCALE package work with custom encodings? for example, if `scale.Marshal(*GrandpaVoter)` is called, will this function be called by SCALE?

lib/grandpa/grandpa.go
```diff
@@ -225,11 +226,12 @@ func (s *Service) Stop() error {
...
 	ad := make([]*types.Authority, len(s.state.voters))
-	for i, v := range s.state.voters {
+	for i := 0; i < len(s.state.voters); i++ {
 		ad[i] = &types.Authority{
-			Key:    v.Key,
-			Weight: v.ID,
+			Key:    &s.state.voters[i].Key,
+			Weight: s.state.voters[i].ID,
 		}
+		//ad[i] = a
```
noot, line 234: remove

lib/grandpa/message.go
```diff
@@ -18,43 +18,36 @@ package grandpa
...
 
-var (
-	voteType            byte
-	commitType          byte = 1
-	neighbourType       byte = 2
-	catchUpRequestType  byte = 3
-	catchUpResponseType byte = 4
-)
+// NewGrandpaMessage returns a new VaryingDataType to represent a GrandpaMessage
+func NewGrandpaMessage() scale.VaryingDataType {
```
noot, line 35: does this need to be exported?

lib/grandpa/network.go
```diff
@@ -225,48 +232,22 @@ func (s *Service) sendNeighbourMessage() {
...
+		m = &val
+	case CatchUpResponse:
+		m = &val
 	default:
-		return nil, ErrInvalidMessageType
-	}
-
-	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("message type not recognised")
```
noot, line 250: why not `ErrInvalidMessageType`?

lib/grandpa/network.go
```diff
@@ -225,48 +232,22 @@ func (s *Service) sendNeighbourMessage() {
...
-	case voteType:
-		r := &bytes.Buffer{}
-		_, _ = r.Write(msg.Data[1:])
-		vm := &VoteMessage{}
-		err = vm.Decode(r)
-		m = vm
+func decodeMessage(msg scale.VaryingDataType) (m GrandpaMessage, err error) {
+	switch val := msg.Value().(type) {
+	case VoteMessage:
+		m = &val
```
noot, line 238: is there a reason you changed this to no longer take a `*network.ConsensusMessage` and decode the `data` in it? it might clean up the tests too

noot (commented):
looks good!! basically just a question about the `decodeMessage` function, I think everything else is fine

dot/types/grandpa.go
jimjbrettj: No it does not. I also do not believe the GrandpaVoter is scale encoded because of the ed25519.PublicKey (Don't think the length is the leading byte of the encoded pubkey)

lib/grandpa/message.go
jimjbrettj: I don't think it does good call

lib/grandpa/network.go
jimjbrettj: Will change

lib/grandpa/network.go
jimjbrettj: Yeah I was doing the decoding before then passing it in but you're right it would make more sense to do it in this method

jimjbrettj: updated!

dot/types/grandpa.go
```diff
@@ -17,13 +17,13 @@
...
 
 import (
-	"encoding/binary"
 	"fmt"
 	"io"
 
 	"github.com/ChainSafe/gossamer/lib/common"
 	"github.com/ChainSafe/gossamer/lib/crypto/ed25519"
 	"github.com/ChainSafe/gossamer/lib/scale"
+	scale2 "github.com/ChainSafe/gossamer/pkg/scale"
```
timwu20, line 26: why do we need both scale packages?

lib/grandpa/types.go
```diff
@@ -34,43 +32,23 @@ type (
...
+type Subround byte
 
+//nolint
 var (
-	prevote         subround
-	precommit       subround = 1
-	primaryProposal subround = 2
+	Prevote         Subround
+	Precommit       Subround = 1
+	PrimaryProposal Subround = 2
```
timwu20, line 42: why are these public now?

jimjbrettj: So scale can access them when decoding/encoding. I believe this is necessary but I can change if I am wrong

dot/types/grandpa.go
jimjbrettj: Well right now encoding/decoding for the grandpaVoter is using custom logic (since its not scale encoded I do not believe) and its uses the old scale to encode/decode. Dot/types is the integration I will finish after this one is merged so I was planning to fix it then but if you want it in this pr I can do that.

dot/types/grandpa.go
noot: yeah, it's always a `[]byte` of length 32. unfortunately the ed25519 has it just as `[]byte` instead of `[32]byte` so we can't directly SCALE encode it. maybe as a next step for SCALE we can add custom type encodings

jimjbrettj: Yeah I think that could be a good idea

noot (approved):
looks good, would be nice to either add SCALE custom encoding or change our `crypto/ed25519/PublicKey` type to be `[32]byte` so that it can be used with new SCALE, maybe wanna open issues for these?

dot/types/grandpa.go
timwu20: As long as we have an issue for this that's fine.

lib/grandpa/types.go
timwu20: The values don't need to be public, only the attribute names if they are being used as an attribute in a struct.

timwu20:
> looks good, would be nice to either add SCALE custom encoding or change our `crypto/ed25519/PublicKey` type to be `[32]byte` so that it can be used with new SCALE, maybe wanna open issues for these?

I prefer the latter.  If it was a `[32]byte` that would encode/decode easily.

dot/types/grandpa.go
timwu20: this method should just be 
```
return scale.Marshal(v)
```
if you change `ed25519.PublicKey` to `[32]byte`

dot/types/grandpa.go
```diff
@@ -111,19 +111,32 @@ func (v *GrandpaVoter) Decode(r io.Reader) error {
...
 	}
 
-	v.Key = key
+	v.Key = *key
 	v.ID = id
 	return nil
 }
 
+// Encode will encode the GrandpaVoter
+func (v *GrandpaVoter) Encode() ([]byte, error) {
```
timwu20, line 120: nit: can you update all the receiver functions.
```suggestion
func (gv *GrandpaVoter) Encode() ([]byte, error) {
```

dot/types/grandpa.go
jimjbrettj: I think we could just delete it all together and just directly marshal/unmarshal. Would also allow us to get rid of the custom encode and decode functions for GrandpaVoters

jimjbrettj: I will look into this now actually. Could be a nice change to include in this PR

lib/grandpa/message.go
```diff
@@ -147,152 +96,85 @@ type NeighbourMessage struct {
...
-		AuthData:   authData,
+	return &ConsensusMessage{
+		Data: enc,
 	}, nil
 }
 
-func justificationToCompact(just []*SignedVote) ([]*Vote, []*AuthData) {
-	precommits := make([]*Vote, len(just))
-	authData := make([]*AuthData, len(just))
+func justificationToCompact(just []SignedVote) ([]Vote, []AuthData) {
```
EclesioMeloJunior, line 171: this function name does not correspond to the action IMO, I guess `splitJustification` or `fromJustificationToVoteAndAuth`

dot/types/grandpa.go
noot: if this is a huge change maybe do it in another PR, since I think the crypto/ed25519 golang lib we're using has some nice functions that we currently use (but maybe it's not a huge change)

lib/grandpa/message.go
jimjbrettj: That makes sense to me. @noot do you have thoughts on this change?

dot/types/grandpa.go
jimjbrettj: Yeah good call, I will do this in a separate PR

lib/grandpa/message.go
timwu20: I think this method should be a receiver function of a new type called `type SignedVotes []SignedVote`.

noot: I named it this way to match substrate's naming, so I think it's ok to leave as is but up to your judgement

timwu20 (approved):

Issue #1659, opened 2021-06-24
research and refactor syncing algorithm


noot:
## Task summary

- refactor syncing algorithm to not necessarily live in the `network` package, but to live in the `sync` package 
- this will require updating the API that the network package exposes, will need to add a function to send a request and receive a response over request/response streams (eg. the `sync` substream)
- the current sync algorithm uses a queue of block requests that get sent out in succession
- responses are sent to the `sync` package for processing
- there are 2 "modes" that need to be handled better: firstly, syncing an established chain, and secondly syncing while near the head of the chain. 
- this will require some research into sync algorithms used by other nodes to figure out a good algorithm

arijitAD:
I investigated the syncing and found a couple of issues:

### 1. HandleTransactionMessage() 
**Issue Description:** After execution of **nth** block **(n+1)th** block throws below error `failed to find root key`.
```logs=
WARN[09-01|02:37:45] failed to load state for block           pkg=sync block=0xdb367756f79d5269fbf4443b9b3cc09599121adb911abaacc585f533eae5dfd1 error="failed to find root key=0xd9dd920475455acfb5f8a3d7396586ed5d00311e54ea4ac80f7375862ab34023: Key not found" caller=syncer.go:203
```
**Analysis:** There is parallel execution of [HandleTransactionMessage()](https://github.com/ChainSafe/gossamer/blob/development/dot/core/messages.go#L28) and [handleblock()](https://github.com/ChainSafe/gossamer/blob/development/dot/sync/syncer.go#L346) for nth block while syncing. The trie state is getting modified by [HandleTransactionMessage()](https://github.com/ChainSafe/gossamer/blob/development/dot/core/messages.go#L28) by **(n-1)th** block. HandelBlockImport() will store incorrect of data **(n-1)th** block trie in DB instead of nth block trie and import nth block successfully. On **executing (n+1)th block** throws an error in [TrieState()](https://github.com/ChainSafe/gossamer/blob/development/dot/state/storage.go#L162) due to absence of **nth** trie storage in db.

**Suggestion:** We should acquire the lock on `storageState` at all places where we are setting storage context for runtime. But this may cause heavy lock contention since transaction messages are broadcasted frequently.  
Instead, we should batch process transaction messages after every regular time interval.

### 2. Issue due to Failed to call the `Core_execute_block` exported function at some x block. This seems to be the reason why Kusama syncing fails  #1770 
```logs=
[...]

PR #1812, opened 2021-09-25
maintenance: continue sync refactor, implement tip sync mode

noot:
## Changes

- implement `tipSyncer` logic
- fix various bugs; firstly fix db `BlockState` number->block hash mapping, now set canonically on block finalisation; secondly fix ordering when processing built block, now only gossips block after finishing processing
- 

## Tests

```
go test ./dot/sync
```

also, run 2-3 gossamer nodes locally for a while and check that they sync:
```
./bin/gossamer --key alice --basepath /tmp/alice
./bin/gossamer --key bob --port 7002 --basepath /tmp/bob
./bin/gossamer --key charlie --port 7003 --basepath /tmp/charlie
```

## Issues

- related to #1659

## Primary Reviewer 

- @arijitAD


edwardmack:
While running three nodes together, one of the nodes had the following error:
```
INFO[09-30|11:21:54] epoch complete!                          pkg=babe completed epoch=0 upcoming epoch=2 caller=babe.go:429
CRIT[09-30|11:21:54] no epoch data for next BABE epoch        pkg=babe epoch=2 caller=epoch.go:46
EROR[09-30|11:21:54] failed to initiate epoch                 pkg=babe epoch=2 error="no epoch data found for upcoming epoch" caller=babe.go:362
CRIT[09-30|11:21:54] block authoring error                    pkg=babe error="no epoch data found for upcoming epoch" caller=babe.go:348
```
Then a minute later it started having these errors: 
```
EROR[09-30|11:22:56] ready block failed                       pkg=sync hash=0x8c7e39624e77dd4e358c79d4c9eec52f9bdfe4a285bc97eeff59f01735ea454e error="failed to get parent header: Key not found" caller=chain_processor.go:96
EROR[09-30|11:22:56] failed to handle block                   pkg=sync number=49 error="failed to get parent header: Key not found" caller=chain_processor.go:184
INFO[09-30|11:28:10] 💤 node waiting                        pkg=sync peer count=2 head=42 hash=0xa4fd832935c0a08a2c381feda161a6a39f8e5ad58542cd09422adf8ec3c46e67 finalised=0 hash=0x26a30534b82025609b198c292634b7faacc95574ecc7a87f9f9b244d7d65e819 caller=chain_sync.go:360
```

These errors seemed to continue for one node, and it never synced with the other two nodes.  Are these error that are a know issue?  Just wanted to let you know what I'm seeing.

noot:
@edwardmack I haven't seen that before :/ if you are able to recreate it can you send me the logs with `sync` set to debug?

edwardmack:
> @edwardmack I haven't seen that before :/ if you are able to recreate it can you send me the logs with `sync` set to debug? are you initializing/starting the nodes with the same command in the comment? also please be sure to re-initialize each time. I haven't seen that issue for a long time (since the gssmr runtime was updated to always have the same authorities each epoch, that should prevent that)

I haven't been able to recreate the error...  I re-initialized them so that was probably my issue.

noot:
@edwardmack ok cool, glad to know! let me know if you find any other major issues

cmd/gossamer/prune_test.go
```diff
@@ -6,36 +6,35 @@ import (
 	"testing"
 
 	"github.com/dgraph-io/badger/v2"
-
 	"github.com/stretchr/testify/require"
 )
 
-func iterateDB(db *badger.DB, cb func(*badger.Item)) {
+func iterateDB(db *badger.DB, cb func(*badger.Item)) { //nolint
```
timwu20, line 12: what's the lint error?

cmd/gossamer/prune_test.go
```diff
@@ -55,6 +54,11 @@ func TestPruneState(t *testing.T) {
 			numStorageKeys++
 			return
 		}
+		if strings.HasPrefix(key, "block") {
+			fmt.Println("found key!!!!")
```
timwu20, line 58: lol, remove?

dot/network/sync.go
```diff
@@ -43,8 +43,11 @@ func (s *Service) DoBlockRequest(to peer.ID, req *BlockRequestMessage) (*BlockRe
 		return nil, err
 	}
 
-	err = s.host.writeToStream(stream, req)
-	if err != nil {
+	defer func() {
+		_ = stream.Close()
+	}()
```
timwu20, line 48:
```suggestion
	defer stream.Close()
```

dot/network/sync.go
```diff
@@ -92,25 +103,23 @@ func decodeSyncMessage(in []byte, _ peer.ID, _ bool) (Message, error) {
...
 	}
 
-	// if it's a BlockRequest, call core for processing
-	if req, ok := msg.(*BlockRequestMessage); ok {
-		defer func() {
-			_ = stream.Close()
-		}()
+	defer func() {
+		_ = stream.Close()
+	}()
```
timwu20, line 111:
```suggestion
	defer stream.Close()
```

dot/sync/chain_sync.go
```diff
@@ -46,6 +48,20 @@ const (
...
+func chainSyncStateString(mode chainSyncState) string {
+	switch mode {
+	case bootstrap:
+		return "bootstrap"
+	case tip:
+		return "tip"
+	default:
+		return "unknown"
+	}
+}
```
timwu20, line 60:
```suggestion
func (css chainSyncState) String() {
	switch css {
	case bootstrap:
		return "bootstrap"
	case tip:
		return "tip"
	default:
		return "unknown"
	}
}
```
The logger will call this method when using `%s`.

dot/sync/chain_sync.go
```diff
@@ -136,9 +154,18 @@ type chainSync struct {
...
 
 	benchmarker *syncBenchmarker
+
+	finalisedCh   <-chan *types.FinalisationInfo
+	finalisedChID byte
 }
 
-func newChainSync(bs BlockState, net Network, readyBlocks chan<- *types.BlockData) *chainSync {
+func newChainSync(bs BlockState, net Network, readyBlocks *blockQueue, pendingBlocks DisjointBlockSet) (*chainSync, error) {
+	finalisedCh := make(chan *types.FinalisationInfo, 128)
```
timwu20, line 163: make the length a constant? like `pendingBlocksLimit`

dot/sync/chain_sync.go
```diff
@@ -46,6 +48,20 @@ const (
...
+		return "bootstrap"
+	case tip:
+		return "tip"
+	default:
+		return "unknown"
+	}
+}
+
+// TODO: determine ideal limit for pending blocks set
+var pendingBlocksLimit = maxResponseSize * 32
```
timwu20, line 63: const?

dot/sync/chain_sync.go
```diff
@@ -369,25 +397,30 @@ func (cs *chainSync) sync() {
...
+			default:
+				// TODO: using these errs in the switch requires upgrading libp2p versions
+				if strings.Contains(res.err.err.Error(), "dial backoff") || res.err.err.Error() == "protocol not supported" {
+					cs.ignorePeer(res.err.who)
+					continue
+				}
+
+				// if errors.Is(res.err.err, io.EOF) {
+				// 	time.Sleep(time.Millisecond * 250)
+				// }
```
timwu20, line 423: remove this then? are there breaking changes in new versions of libp2p?

dot/sync/chain_sync.go
```diff
@@ -402,23 +435,49 @@ func (cs *chainSync) sync() {
...
+	if err != nil {
+		logger.Error("failed to get best block header", "error", err)
+		return
+	}
+
+	target := cs.getTarget()
+	if big.NewInt(0).Add(head.Number, big.NewInt(maxResponseSize)).Cmp(target) == -1 {
+		// we are 128 blocks or more behind the target, switch to bootstrap mode
+		cs.setMode(bootstrap)
+	} else if head.Number.Cmp(target) >= 0 {
```
timwu20, line 474: shouldn't this be an `else`? Is there another unaccounted for case?

dot/sync/chain_sync.go
```diff
@@ -609,19 +678,41 @@ func (cs *chainSync) doSync(req *network.BlockRequestMessage) *workerError {
...
+func handleReadyBlock(bd *types.BlockData, pendingBlocks DisjointBlockSet, readyBlocks *blockQueue) {
+	// see if there are any descendents in the pending queue that are now ready to be processed,
+	// as we have just become aware of their parent block
+
+	// if header was not requested, get it from the pending set
+	// if we're expecting headers, validate should ensure we have a header
+	if bd.Header == nil {
+		block := pendingBlocks.getBlock(bd.Hash)
+		if block.header == nil {
+			panic("block isn't ready! header is unknown!")
```
timwu20, line 696: any way to avoid having panics?

dot/sync/chain_sync.go
```diff
@@ -609,19 +678,41 @@ func (cs *chainSync) doSync(req *network.BlockRequestMessage) *workerError {
...
+	}
+
+	logger.Trace("new ready block", "hash", bd.Hash, "number", bd.Header.Number)
+
+	ready := []*types.BlockData{bd}
+	ready = pendingBlocks.getReadyDescendants(bd.Hash, ready)
+
+	for _, rb := range ready {
+		pendingBlocks.removeBlock(rb.Hash)
+		readyBlocks.push(rb)
```
timwu20, line 709: this call could be blocking right? Do we want to block `dispatchWorker` from progressing?

dot/sync/chain_sync.go
```diff
@@ -609,19 +678,41 @@ func (cs *chainSync) doSync(req *network.BlockRequestMessage) *workerError {
...
+	}
+}
+
 // determineSyncPeers returns a list of peers that likely have the blocks in the given block request.
-// TODO: implement this
 func (cs *chainSync) determineSyncPeers(_ *network.BlockRequestMessage) []peer.ID {
```
timwu20, line 714: why do we need this `network.BlockRequestMessage` param?

dot/sync/disjoint_block_set.go
```diff
@@ -58,50 +85,79 @@ type pendingBlock struct {
...
+		limit:            limit,
 	}
 }
 
-func (s *disjointBlockSet) addHashAndNumber(hash common.Hash, number *big.Int) {
+func (s *disjointBlockSet) addToParentMap(parent, child common.Hash) {
+	children, has := s.parentToChildren[parent]
+	if !has {
+		children = make(map[common.Hash]struct{})
+		s.parentToChildren[parent] = children
```
timwu20, line 110: do we need mutex on this?

cmd/gossamer/prune_test.go
noot: unused, since i skipped the test that uses it, since it's failing on missing db keys :/ we need to re-generate the db to fix it, or move this test the state package

dot/network/sync.go
noot: this causes a lint error for unchecked err :/

dot/network/sync.go
noot: same as above

dot/sync/chain_sync.go
noot: the `*` makes it not a const :(

dot/sync/chain_sync.go
noot: yeah, the `else` case is if the head is between `(target-128, target)`. in that case we don't want to switch modes

dot/sync/chain_sync.go
noot: yeah this shouldn't happen ever, i can remove it

dot/sync/chain_sync.go
noot: potentially, i think it's goo if it blocks, otherwise it'll return to the result handler which may be doing some logic to check what's in the `readyBlocks` queue, so if this is still waiting then the result handler might be incorrect

dot/sync/chain_sync.go
noot: I need to update this so it checks the `peerState`s to check which peers have the blocks in the request. can do that in a follow-up

dot/sync/disjoint_block_set.go
noot: it's only used within other `disjointBlockState` functions that already have the mutex

edwardmack (approved):
Nice work! lgtm.

dot/sync/tip_syncer.go
```diff
@@ -17,72 +17,172 @@
...
+	}
+
 	// cases for each block in pending set:
-	// 1. only hash and number are known; in this case, request the full block
+	// 1. only hash and number are known; in this case, request the full block (and ancestor chain)
 	// 2. only header is known; in this case, request the block body
 	// 3. entire block is known; in this case, check if we have become aware of the parent
+	// if we have, move it to the ready blocks queue; otherwise, request the chain of ancestors
+
+	workers := []*worker{}
```
arijitAD, line 134: `var workers []*worker`

dot/sync/tip_syncer.go
```diff
@@ -17,72 +17,172 @@
...
+				targetHash:   block.hash,
+				targetNumber: block.number,
+				requestData:  network.RequestedDataBody + network.RequestedDataJustification,
+			})
+			continue
 		}
 
 		// case 3
-		// WIP
+		has, _ := s.blockState.HasHeader(block.header.ParentHash)
```
arijitAD, line 169: Should we change the function definition of `HasHeader` to `HasHeader(hash common.Hash) bool`
I think the error is necessary.

dot/sync/chain_sync.go
```diff
@@ -610,16 +677,35 @@ func (cs *chainSync) doSync(req *network.BlockRequestMessage) *workerError {
...
 	peers := make([]peer.ID, 0, len(cs.peerState))
```
arijitAD, line 710: The `lock` should be taken at the starting of function as the length may change.

dot/sync/chain_sync.go
```diff
@@ -326,40 +360,34 @@ func (cs *chainSync) logSyncSpeed() {
...
 				"hash", finalised.Hash(),
 			)
 		}
 	}
 }
 
+func (cs *chainSync) ignorePeer(who peer.ID) {
+	if who == peer.ID("") {
+		return
+	}
```
arijitAD, line 375:
```suggestion
	if err := who.Validate(); err != nil {
		return
	}
```

dot/sync/chain_sync.go
```diff
@@ -403,23 +434,49 @@ func (cs *chainSync) sync() {
...
+			workers, err := cs.handler.handleTick()
 			if err != nil {
 				logger.Error("failed to handle tick", "error", err)
 				continue
 			}
 
-			if worker == nil {
+			if workers == nil {
 				continue
 			}
```
arijitAD, line 447: This `if` condition is not needed. It will be taken care of by loop.

PR #1813, opened 2021-09-28
chore(ci): Dockerfile improvements and final image based on Alpine

qdm12:
## Changes

- Change builder image from `ubuntu:18.04` to `golang:1.15-buster` (faster to pull the Go image and Debian Buster is smaller than Ubuntu)
- Final image is running on Alpine 3.14 (with glibc and needed C libraries), reducing image size from 3GB to 122MB (uncompressed amd64).
- This should keep retro-compatibility as there is still `bash` and the file paths are kept the same changed.

### Other minor improvements

- Cache Polkadot JS NPM dependencies (package.json + package-lock.json)
- Image versions as build arguments `ALPINE_VERSION=3.14` and `GO_VERSION=1.15-buster`
- Add `.dockerignore` file for better use of the Docker laying caching
- Compact commands together for smaller image sizes and smaller build cache size
- Use only `wget` and not `curl` in builder
- Re-order instructions in final image to maximize Docker layer caching efficiency
- Remove unused `GOBIN() string` function from `ci.go` since it's unused

## Questions for reviewers

1. Do we actually need the Polkadot JS dependencies in the image? Where do we run tests with it? That might need adjustments in the CI to run `docker build --target builder` since they are not carried over the final Alpine image.
2. What **files** does the Docker image need to function? For now it only uses:
   - The Go binary
   - The C libraries required, including `libwasmer.so` taken from the Go modules
   - `docker-entrypoint.sh`
   - The `./chain` directory
   - Does it also need for example `subkey` or other config files?
3. I haven't toyed around with it too much, so you might want to ensure it's fully functional (or I will with indications on what to test 😉). But for now I get the same logs as with the current `chainsafe/gossamer` image.

## Tests

```
docker build -t chainsafe/gossamer .
docker run -it --rm chainsafe/gossamer

+ '[' '!' -f /root/gossamer-dev/genesis_created ']'
+ /usr/local/gossamer init --genesis=/gocode/src/github.com/ChainSafe/gossamer/chain/gssmr/genesis.json
INFO[09-28|17:34:57] 🕸️ initialising node...             pkg=dot name=alley-cheap-1668 id=gssmr basepath=/root/.gossamer/gssmr genesis=/gocode/src/github.com/ChainSafe/gossamer/chain/gssmr/genesis.json caller=node.go:61
INFO[09-28|17:34:58] state                                    pkg=state genesis hash=0x26a30534b82025609b198c292634b7faacc95574ecc7a87f9f9b244d7d65e819 caller=initialize.go:123
INFO[09-28|17:34:58] node initialised                         pkg=dot name=alley-cheap-1668 id=gssmr basepath=/root/.gossamer/gssmr genesis=/gocode/src/github.com/ChainSafe/gossamer/chain/gssmr/genesis.json block=0 genesis hash=0x26a30534b82025609b198c292634b7faacc95574ecc7a87f9f9b244d7d65e819 caller=node.go:118
+ touch /root/gossamer-dev/genesis_created
+ exec /usr/local/gossamer
INFO[09-28|17:34:58] loaded package log configuration         pkg=cmd cfg="{CoreLvl:info SyncLvl:info NetworkLvl:info RPCLvl:info StateLvl:info RuntimeLvl:info BlockProducerLvl:info FinalityGadgetLvl:info}" caller=config.go:140
EROR[09-28|17:34:58] failed to create node services           pkg=cmd error="no keys provided for authority node" caller=main.go:299
no keys provided for authority node
```

## Issues

- Initially started to address #424 
- The build is still required to be on a glibc based OS (Debian here), but the runtime seems possible with Alpine + Glibc


Dockerfile
```diff
@@ -1,64 +1,58 @@
...
-RUN chmod +x $GOPATH/src/github.com/ChainSafe/gossamer/scripts/docker-entrypoint.sh
+# Install libwasmer.so
+ENV LD_LIBRARY_PATH=/lib:/usr/lib
+COPY --from=builder /go/src/github.com/ChainSafe/gossamer/libwasmer.so /lib/libwasmer.so
+
+EXPOSE 7001 8546 8540
 
-# Expose gossamer command and port
 ENTRYPOINT ["/gocode/src/github.com/ChainSafe/gossamer/scripts/docker-entrypoint.sh"]
 CMD ["/usr/local/gossamer"]
```
EclesioMeloJunior, line 52: I guess this `CMD` could be moved to `scripts/docker-entrypoint.sh`

EclesioMeloJunior (changes requested):

Dockerfile
qdm12: Well we could change a lot of changes regarding the entrypoint 😉

But I wanted to keep compatibility (and also cmd and entrypoint come as documentation when you `docker inspect chainsafe/gossamer`). If we're fine doing small compatibility breakings, then yes we can change that one. We could also have it in another shorter directory path than this one.

On a side note, I think (80% sure) relative paths for entrypoints do not work for example on `scratch` based Docker images, so we may want to keep using absolute paths. But that's just nitpicking from me 😄

Dockerfile
```diff
@@ -1,64 +1,69 @@
-FROM ubuntu:18.04 as builder
+ARG ALPINE_VERSION=3.14
+ARG GO_VERSION=1.15-buster
+
+FROM golang:${GO_VERSION} AS builder
 
-# Install GCC
 RUN apt-get update && \
     apt-get install -y \
     gcc \
```
EclesioMeloJunior, line 8: as we are using `golang:1.15-buster` do we need to install gcc?

Dockerfile
```diff
@@ -1,64 +1,69 @@
-FROM ubuntu:18.04 as builder
+ARG ALPINE_VERSION=3.14
+ARG GO_VERSION=1.15-buster
```
EclesioMeloJunior, line 2: could you update to 1.17?

EclesioMeloJunior (changes requested):

EclesioMeloJunior:
@qdm12 could we resume this PR as PR #1934 was merged?

Dockerfile
qdm12: Done through a rebase.

Dockerfile
qdm12: That's how it was before, although we may want to remove it in our development branch Dockerfile since I think as well gcc comes pre-installed (to verify).

qdm12:
@EclesioMeloJunior unfortunately, the container no longer works due to some C/Rust-linking-ish problem again, as mentioned in the description. I'll toy around with it again in the coming days see if I can fix it.

PR #1817, opened 2021-09-29
chore(dot/types): integrate new scale pkg into dot/types

### LINKED ISSUE #1743: Integrate scale package into dot types

jimjbrettj:
## Changes

- Removed old scale instances from dot/types
- Created a ScaleVoter struct that allows us to encode/decode GrandpaVoters using scale. This is needed since GrandpaVoter.Key is not scale encoded.
- Removed as much custom serialization logic as possible

## Tests

```
go test ./dot/types -v
go test ./dot/state -v -short
go test ./lib/grandpa -v -short
```

## Issues

- closes #1743 

## Primary Reviewer 

- @timwu20


dot/state/initialize.go
```diff
@@ -149,7 +149,7 @@ func loadGrandpaAuthorities(t *trie.Trie) ([]types.GrandpaVoter, error) {
 
 	r := &bytes.Buffer{}
 	_, _ = r.Write(authsRaw[1:])
-	return types.DecodeGrandpaVoters(r)
+	return types.DecodeGrandpaVoters(authsRaw[1:])
```
EclesioMeloJunior, line 152: as `DecodeGrandpaVoters` receive `authsRaw[1:]` directly so the `&bytes.Buffer` is not needed anymore and you can remove it

EclesioMeloJunior (changes requested):

dot/state/grandpa.go
```diff
@@ -111,7 +111,7 @@ func (s *GrandpaState) GetAuthorities(setID uint64) ([]types.GrandpaVoter, error
 		return nil, err
 	}
```
EclesioMeloJunior, line 112: remove these unused lines

dot/types/inherents.go
```diff
@@ -63,7 +63,7 @@ func (d *InherentsData) SetInt64Inherent(key []byte, data uint64) error {
 	val := make([]byte, 8)
 	binary.LittleEndian.PutUint64(val, data)
 
-	venc, err := scale.Encode(val)
+	venc, err := scale.Marshal(val)
```
EclesioMeloJunior, line 66: with the new SCALE code, I guess is just `scale.Marshal(data)` without that `LittleEndian.PutUint64` call, right?

EclesioMeloJunior (changes requested):

dot/types/inherents.go
jimjbrettj: Im not sure if I can make this change because the map data is a []byte and if I encode the uint64 as is it will be different than the []byte encoding

dot/state/grandpa.go
jimjbrettj: good catch thank you

dot/types/inherents.go
EclesioMeloJunior: hm, I told that because I used the `scale.Marshal(uint32(len(data)))` and the runtime method does not complain about the bytes format

jimjbrettj: I'm not entirely sure what you mean I don't think. My concern is when you encode a slice the leading byte represents the length of the slice, followed by the data and encoding a uint64 is just the data. So I think in this case the encoding would be different if I just directly encoded the uint64

dot/types/babe_digest.go
```diff
@@ -17,286 +17,179 @@
...
+// GetAuthorityIndex returns the digest's authority index
+func (d BabePrimaryPreDigest) GetAuthorityIndex() uint32 {
+	return d.AuthorityIndex
 }
 
-// SlotNumber returns the digest's slot number
-func (d *BabePrimaryPreDigest) SlotNumber() uint64 {
-	return d.slotNumber
+// GetSlotNumber returns the digest's slot number
+func (d BabePrimaryPreDigest) GetSlotNumber() uint64 {
```
timwu20, line 101: why did you add the `Get` prefix on these methods?

dot/types/body.go
```diff
@@ -97,18 +95,18 @@ func NewBodyFromExtrinsicStrings(ss []string) (*Body, error) {
 
 // AsExtrinsics decodes the body into an array of extrinsics
 func (b *Body) AsExtrinsics() ([]Extrinsic, error) {
-	exts := [][]byte{}
+	dec := [][]byte{}
```
timwu20, line 98: nit: move this after the first `if` block

dot/types/grandpa.go
```diff
@@ -78,6 +57,24 @@ func GrandpaAuthoritiesRawToAuthorities(adr []GrandpaAuthoritiesRaw) ([]Authorit
 	return ad, nil
 }
 
+// ScaleVoter represents a scale compatible GRANDPA voter
+type ScaleVoter struct {
```
timwu20, line 61: Do we need to call this `ScaleVoter`, why not just `Voter`?

lib/babe/secondary.go
```diff
@@ -68,25 +68,25 @@ func verifySecondarySlotVRF(digest *types.BabeSecondaryVRFPreDigest,
...
 		"number of authorities", numAuths,
-		"slot", digest.SlotNumber(),
+		"slot", digest.SlotNumber,
 		"epoch", epoch,
 		"randomness", randomness,
 		"expected index", expected,
 	)
 
-	if digest.AuthorityIndex() != expected {
+	if digest.GetAuthorityIndex() != expected {
```
noot, line 86: are the getters needed, if the fields are all exported now? ie. you can just do `digest.AuthorityIndex`

noot (commented):
looks good, just a few comments

dot/types/babe_digest.go
jimjbrettj: When I made the attributes public I was getting naming errors since the function was the same name as the variable. I just put get to differentiate the two, but can change to something more appropriate. Do you have any suggestion on how to handle this?

lib/babe/secondary.go
jimjbrettj: I don't think they are needed most of the time, but some instances use the BabePreRuntimeDigest interface rather than the specific struct and then I think we need the getters. If there is a way to change that though that I am missing I would be happy (and prefer lol) to remove the getters

Issue #1688, opened 2021-07-09
update host API for v0.9.8 runtime version


noot:
## Task summary

- runtime version v0.9.8 updates the host API, need to update to continue syncing past upgrade
- see https://github.com/paritytech/substrate/pull/9106
- maybe see also https://github.com/paritytech/substrate/pulls?q=is%3Apr+label%3AE4-newhostfunctions

TODO:

- [x] #1792
- [x] #1793
- [x] #1798
- [x] #1794
- [x] #1795
- [ ] #1796
- [ ] #1797

PR #1818, opened 2021-09-30
maintainence(lib/runtime/wasmer): add support for polkadot runtime v0.9.10

noot:
## Changes

- add support for polkadot runtime v0.9.10
- add `ext_crypto_secp256k1_ecdsa_recover_version_2` and `ext_crypto_secp256k1_ecdsa_recover_compressed_version_2` functions to wasmer imports
- not sure if there's a difference between version 1 and 2, substrate code appears to be the same for both https://github.com/paritytech/substrate/blob/2ab5cd2880b1ec798a7c2a72c29f995f54b45f13/primitives/io/src/lib.rs#L760

## Tests

```
go test ./lib/crypto/secp256k1
go test ./lib/runtime/wasmer/ -run TestInstance_Version_PolkadotRuntime_v0910
```

## Issues

- sort of related to #1688

## Primary Reviewer 

- @edwardmack


edwardmack (changes requested):
This lgtm, however I'm concerned with the version_2 just calling version_1.  I don't see these version_2s defined in the spec, but looking at the substrate code there is a slight difference:
https://github.com/paritytech/substrate/blob/2ab5cd2880b1ec798a7c2a72c29f995f54b45f13/primitives/io/src/lib.rs#L764 is different than: https://github.com/paritytech/substrate/blob/2ab5cd2880b1ec798a7c2a72c29f995f54b45f13/primitives/io/src/lib.rs#L789 for secp256k1_ecdsa_recover.  

And for secp256k1_ecdsa_recover_compressed:  https://github.com/paritytech/substrate/blob/2ab5cd2880b1ec798a7c2a72c29f995f54b45f13/primitives/io/src/lib.rs#L812 is different than:  https://github.com/paritytech/substrate/blob/2ab5cd2880b1ec798a7c2a72c29f995f54b45f13/primitives/io/src/lib.rs#L834

I'm not sure what the difference is between `parse_overflowing_slice` and `parse_standard_slice` is, we should determine how this difference affects behavior.

noot:
@edwardmack ok not sure how I missed that😬 I'll look into the difference between the two today and let you know when it's updated

noot:
@edwardmack looked into the difference between `parse_overflowing` and `parse_standard`, it seems that `parse_overflowing` is a non-standard way of parsing a signature that is potentially larger than the secp256k1 curve order: https://github.com/paritytech/libsecp256k1/blob/b30fdab3d323c5418264f961dbef6608bb04f15f/src/lib.rs#L498 I believe that the Go library (which uses the C Bitcoin secp256k1 implementation under the hood) is likely already using "standard" parsing, so this should be ok to do.

edwardmack (approved):

lib/crypto/secp256k1/secp256k1.go
```diff
@@ -48,11 +51,27 @@ type PublicKey struct {
 
 // RecoverPublicKey returns the 64-byte uncompressed public key that created the given signature.
 func RecoverPublicKey(msg, sig []byte) ([]byte, error) {
+	if sig[64] == 27 {
```
arijitAD, line 54: Can you add code comments to explain what this is doing?

lib/crypto/secp256k1/secp256k1.go
```diff
@@ -48,11 +51,27 @@ type PublicKey struct {
...
+		sig[64] = 0
+	}
+
+	if sig[64] == 28 {
```
arijitAD, line 58: Not needed but you can combine both conditions if it makes sense.
```
	if sig[64] == 27 || sig[64] == 28  {
		sig[64] -= 27
	}
````
or as done here
https://github.com/paritytech/substrate/blob/2ab5cd2880b1ec798a7c2a72c29f995f54b45f13/primitives/io/src/lib.rs#L767
```
        const someVal = 27
	if sig[64] >= someVal {
		sig[64] -= someVal
	}
````

arijitAD (approved):

lib/crypto/secp256k1/secp256k1.go
noot: done!

arijitAD: @noot Same can be done for `RecoverPublicKeyCompressed`

noot: also done!

PR #1824, opened 2021-10-01
chore: Integrate scale package into runtime

timwu20:
## Changes

-
-
-

## Tests

```

```

## Issues

- #1749 

## Primary Reviewer 

-


lib/runtime/life/resolver.go
```diff
@@ -424,26 +423,22 @@ func storageAppend(storage runtime.Storage, key, valueToAppend []byte) error {
...
 		if err != nil {
 			logger.Trace("[ext_storage_append_version_1] item in storage is not SCALE encoded, overwriting", "key", key)
 			storage.Set(key, append([]byte{4}, valueToAppend...))
 			return nil
 		}
 
-		// append new item
-		valueRes = append(r.Bytes(), valueToAppend...)
+		// append new item, pop off first byte now, since we're not using old scale.Decoder
+		valueRes = append(valueCurr[1:], valueToAppend...)
```
jimjbrettj, line 435: Is there a risk that since currLength is a *big.Int that it will consume more space than the leading byte?

lib/runtime/version.go
```diff
@@ -65,131 +61,81 @@ func NewLegacyVersionData(specName, implName []byte, authoringVersion, specVersi
...
-		SpecVersion      uint32
-		ImplVersion      uint32
-	}{
-		SpecName:         v.specName,
-		ImplName:         v.implName,
-		AuthoringVersion: v.authoringVersion,
-		SpecVersion:      v.specVersion,
-		ImplVersion:      v.implVersion,
-	}
+type legacyVersionData struct {
```
jimjbrettj, line 98: What is the purpose of having both this stuct and LegacyVersionData?

lib/runtime/version.go
```diff
@@ -219,138 +165,84 @@ func NewVersionData(specName, implName []byte, authoringVersion, specVersion, im
...
-	for _, apiItem := range v.apiItems {
-		enc = append(enc, apiItem.Name[:]...)
-
-		b, err = scale.Encode(apiItem.Ver)
-		if err != nil {
-			return nil, err
-		}
-		enc = append(enc, b...)
-	}
+type versionData struct {
```
jimjbrettj, line 202: I'm assuming there's a reason but I'm also curious about this struct.

dot/core/service_test.go
```diff
@@ -287,9 +287,9 @@ func TestHandleChainReorg_WithReorg_Transactions(t *testing.T) {
 	addTestBlocksToState(t, height, s.blockState.(*state.BlockState))
 
 	// create extrinsic
-	ext := extrinsic.NewIncludeDataExt([]byte("nootwashere"))
-	tx, err := ext.Encode()
+	enc, err := scale.Marshal([]byte("nootwashere"))
 	require.NoError(t, err)
+	tx := append([]byte{2}, enc...)
```
jimjbrettj, line 292: Am I right in saying that the 2 is appended because its a IncludeDataType extrinsic? If so, since we are deleting that file, will we be able to know what the leading byte represents/what values it can have in the future? I was initially confused at what this 2 was and I'm not sure how I would figure that out with the extrinsic file gone. Maybe this isn't a huge concern but I wanted to ask

noot: yeah would be good to document this

lib/runtime/life/resolver.go
noot: why is the first byte popped off? is it because of the length encoding? the length encoding isn't necessarily 1 byte

lib/runtime/version.go
noot: it looks like  this has exported fields so its used to decode with SCALE?

lib/runtime/wasmer/exports.go
```diff
@@ -51,15 +50,16 @@ func (in *Instance) Version() (runtime.Version, error) {
...
 	if err != nil {
 		return nil, err
 	}
 
+	version := &runtime.VersionData{}
 	err = version.Decode(res)
-	if err == io.EOF {
-		// kusama seems to use the legacy version format
+	// error comes from scale now, so do a string check
+	if strings.Contains(fmt.Sprintf("%s", err), "EOF") {
```
noot, line 61:
```suggestion
	if strings.Contains(err.Error(), "EOF") {
```

what do you think?

lib/runtime/wasmer/imports.go
```diff
@@ -1563,37 +1559,35 @@ func storageAppend(storage runtime.Storage, key, valueToAppend []byte) error {
...
 		if err != nil {
 			logger.Trace("[ext_storage_append_version_1] item in storage is not SCALE encoded, overwriting", "key", key)
 			storage.Set(key, append([]byte{4}, valueToAppend...))
 			return nil
 		}
 
-		// append new item
-		valueRes = append(r.Bytes(), valueToAppend...)
+		// append new item, pop off first byte now, since we're not using old scale.Decoder
+		valueRes = append(valueCurr[1:], valueToAppend...)
```
noot, line 1575: same concern here

noot (commented):
looks good, just a comment/concern about the `storageAppend` func

lib/runtime/life/resolver.go
timwu20: nice catch

lib/runtime/version.go
timwu20: Yea the public struct implements an interface with getters and setters, which conflict with the attribute names if I were to make them public.  Generics will be nice to implement this.

lib/runtime/wasmer/exports.go
timwu20: sure 👍

dot/core/service_test.go
timwu20: I'll add documentation

jimjbrettj: Are there other possible prefixes that should be documented besides the 2 case? Or can those be found elsewhere if need be?

jimjbrettj (approved):
Looks good! Just minor comment

PR #1825, opened 2021-10-01
feat(runtime): implement custom logging handler that print function name

### LINKED ISSUE #1456: Implement a custom logger for `wasmer `package that prints the function name.
## Issue summary

We can implement a custom logger for `wasmer `package that prints the function name dynamically instead of manually adding the function name to logs.
```
logger.Warn("[ext_crypto_ed25519_public_keys_version_1]", "name", id, "error", "keystore type is not ed25519", "type", ks.Type())
```

- https://play.golang.org/p/B4-xirJsSKd
- https://wycd.net/posts/2014-07-02-logging-function-names-in-go.html#:~:text=Go%201.3%20provides%20flags%20to,enough%20context%20for%20fast%20debugging.

edwardmack:
## Changes

- Implement logging handler that adds caller function name to the log context with key "func", so that function name is dynamically added to logging messages.

Example:
 `logger.Debug("executing...")` 
logs this message (including the function name): 
`DBUG[10-01|17:37:01] executing...                             pkg=runtime module=go-wasmer func=ext_default_child_storage_storage_kill_version_1 caller=imports.go:1112`

## Tests

```

```

## Issues

- @arijitAD 

## Primary Reviewer 

- closes #1456


edwardmack:
I'm not sure if we want to include function name in log ctx, or if it should be appended (or pre-pended) to log message text?

lib/runtime/wasmer/instance.go
```diff
@@ -100,6 +102,15 @@ func NewInstance(code []byte, cfg *Config) (*Instance, error) {
 	return newInstance(code, cfg)
 }
 
+// FuncCallerFileHandler returns a Handler that adds the name of the calling function to the context with key "func"
+//  and the line number and file of the calling function to the context with key "caller".
+func FuncCallerFileHandler(h log.Handler) log.Handler {
```
arijitAD, line 107: Maybe rename to `customFileHandler` and make the function private.

arijitAD (approved):
I think adding the function name as the context makes the log clean.

lib/runtime/wasmer/instance.go
noot: can we put this in the `runtime` package? since the other wasm interpreters will likely end up using the same logger

noot (approved):
looks good!

lib/runtime/wasmer/instance.go
edwardmack: Moved to `runtime` package, renamed to `CustomFileHandler`, kept public since it was moved to runtime package.

PR #1833, opened 2021-10-05
refactor(dot/state): only store finalised blocks in database

### LINKED ISSUE #1784: gossamer should only store finalized blocks in the database
## Task summary

- gossamer should only store finalized blocks in the database, unfinalized blocks should only be in-memory
- this is what substrate does
- it is simpler to do this as it prevents us from writing blocks to the database that may later be deleted
- also, when restarting the node, it's easier to re-instantiate the blocktree as you only need to load the latest finalized block
- currently we write the blocktree to the db on shutdown and reload it on startup, however if there is an abnormal shutdown and the node doesn't write it, it causes all sorts of errors on startup (eg, "failed to get runtime" for a block, "failed to get parent", etc)

noot:
## Changes

- refactor gossamer to only store finalised blocks in the database, unfinalised forks are stored in the blocktree and unfinalised `*types.Block` are stored in a map in the `*state.Block`
- update `*state.Block`getter functions to get `unfinalisedBlocks` map
- update `*state.Block.SetFinalisedHash` (which is called when a block is marked finalised) to store blocks that are an ancestor of the finalised block (or are the finalised block) in `unfinalisedBlocks` map into the database
- remove need for database within blocktree package, as the blocktree is no longer written to the db at all
- update `*state.BlockState` `NewBlockState` to create a blocktree on startup from the latest finalised block

## Tests

```
go test ./dot/state
go test ./lib/blocktree
```

## Issues

- closes #1784

## Primary Reviewer 

- @timwu20


dot/types/body.go
```diff
@@ -95,6 +95,17 @@ func NewBodyFromExtrinsicStrings(ss []string) (*Body, error) {
 	return &body, nil
 }
 
+func (b *Body) String() string {
+	fmt.Println("body.String")
+	if b == nil {
+		return "nil"
+	}
+
+	fmt.Println("body", *b)
```
EclesioMeloJunior, line 104: could this prints be removed?

dot/types/body.go
```diff
@@ -95,6 +95,17 @@ func NewBodyFromExtrinsicStrings(ss []string) (*Body, error) {
 	return &body, nil
 }
 
+func (b *Body) String() string {
+	fmt.Println("body.String")
+	if b == nil {
+		return "nil"
```
EclesioMeloJunior, line 101: why not return an empty string?

EclesioMeloJunior (changes requested):

dot/types/body.go
noot: yeah I'll update this. this was updated since I was debugging a test

dot/types/body.go
noot: yes will remove

PR #1880, opened 2021-10-11
maintenance(lib/runtime): remove wasmtime

noot:
## Changes

- as wasmtime is currently non-functional, I have removed it and placed it in a separate branch `wasmtime-integration`

## Tests

```
n/a
```

## Issues

- n/a

## Primary Reviewer 

- @edwardmack


noot:
@arijitAD good point, will do!

PR #1885, opened 2021-10-12
fix(lib/runtime): update HOST_API_TEST_RUNTIME_URL to use specific commit

noot:
## Changes

- update HOST_API_TEST_RUNTIME_URL to use specific commit, fixes tests on development

## Tests

```
go test ./lib/runtime/...
```

## Issues

- n/a

## Primary Reviewer 

- @edwardmack


noot:
@danforbes I had to fork it the spec repo at some point to add extra test functions that the spec repo didn't have yet. we can try to merge it with upstream potentially

PR #1910, opened 2021-10-19
feat(lib/babe): add check of types.ConfigData.SecondarySlots for disabling secondary verification

### LINKED ISSUE #1863: lib/babe: update BABE verifier to use `types.ConfigData.SecondarySlots`
## Task summary

- currently, the BABE verifier allows secondary slots by default
- however, it should configure this based on the babe `ConfigData.SecondarySlots` field, turning it off/on as specified

edwardmack:
## Changes

- Added check of `types.ConfigData.SecondarySlots` byte, if 0, disable secondary slot verification.
-
-

## Tests

```

```

## Issues

- closes #1863

## Primary Reviewer 

- @noot


lib/babe/verify.go
```diff
@@ -384,17 +388,24 @@ func (b *verifier) verifyPreRuntimeDigest(digest *types.PreRuntimeDigest) (types
...
 		}
 
-		ok, err = verifySecondarySlotVRF(d, pk, b.epoch, len(b.authorities), b.randomness)
+		ok, err = verifySecondarySlotVRF(d, pk, b.epoch, len(b.authorities), b.randomness) // nolint
+
 	case *types.BabeSecondaryPlainPreDigest:
 		ok = true
-		err = verifySecondarySlotPlain(d.AuthorityIndex(), d.SlotNumber(), len(b.authorities), b.randomness)
+		if b.secondarySlots {
+			err = verifySecondarySlotPlain(d.AuthorityIndex(), d.SlotNumber(), len(b.authorities), b.randomness)
```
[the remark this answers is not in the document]
edwardmack: Updated.

lib/babe/verify.go
```diff
@@ -409,15 +413,28 @@ func (b *verifier) verifyPreRuntimeDigest(digest *types.PreRuntimeDigest) (scale
...
 	case types.BabeSecondaryVRFPreDigest:
+		if !b.secondarySlots {
+			ok = false
+			break
+		}
 		pub := b.authorities[d.AuthorityIndex].Key
-		var pk *sr25519.PublicKey
-		pk, err = sr25519.NewPublicKey(pub.Encode())
+
+		pk, err := sr25519.NewPublicKey(pub.Encode()) // nolint
```
qdm12, line 422: Assuming this is because of govet's funny shadowing detection feature, use this instead:

```suggestion
		pk, err := sr25519.NewPublicKey(pub.Encode()) //nolint:govet
```

lib/babe/verify.go
```diff
@@ -409,15 +413,28 @@ func (b *verifier) verifyPreRuntimeDigest(digest *types.PreRuntimeDigest) (scale
...
+		if err != nil {
+			return nil, err
+		}
+
+		ok, err = verifySecondarySlotVRF(&d, pk, b.epoch, len(b.authorities), b.randomness) // nolint
```
qdm12, line 427: Same here, let's not use `// nolint` anymore. You can specify the linter you want to be ignored with `//nolint:lintername`. Note there is no space between `//` and `nolint`.

qdm12 (approved):

lib/babe/verify_test.go
```diff
@@ -162,6 +164,94 @@ func TestVerificationManager_VerifyBlock_Ok(t *testing.T) {
...
+	err = scale.Unmarshal(digEnc, &babeDigest)
+	require.NoError(t, err)
+	// Why isn't this type types.BabeSecondaryVRFPreDigest?
+	require.IsType(t, types.BabePrimaryPreDigest{}, babeDigest)
+
+	babePreDigest, err := types.DecodeBabePreDigest(digEnc)
+	require.NoError(t, err)
+	// Why isn't this type types.BabeSecondaryVRFPreDigest?
+	require.IsType(t, types.BabePrimaryPreDigest{}, babePreDigest)
+}
```
edwardmack, line 253: @jimjbrettj @timwu20 I had old code that was encoding with `dig.Encode()`, but that method is now gone, so I'm using scale marshal, but not getting expected results.  I'm I doing something wrong here?

timwu20: `digEnc` is scale encoded `BabeSecondaryVRFPreDigest`.  For it to unmarshal correctly into the `types.BabeDigest` which is a `scale.VaryingDataType` the first byte of the encoding needs to be the index of `BabeSecondaryVRFPreDigest`.

edwardmack: I'm trying to setup conditions to test digest type `types.BabeSecondaryVRFPreDigest` in `verifyPreRuntimeDigest` in `lib\babe\verify.go`.  In `TestVerificationManager_VerifyBlock_Secondary` I create digest with `createSecondaryVRFPreDigest`, then I would call `.Encode` on that result to encode it, add the encoded digest to block header and pass that to `VerifyBlock`.  VerifyBlock calls verifyPreRuntimeDigest and the switch in there would use case `types.BabeSecondaryVRFPreDigest`.  

However, now that there is no longer an `.Encode` function for the digest generated by `createSecondaryVRFPreDigest` I tried encoding with scale.Encode and experienced what I've described above.  

That's what I'm trying to accomplish, @timwu20 yes I'm happy to sync to discuss this.

noot: I think like TIm said you need to use the `scale.VaryingDataType` and register the digest types with the byte index, as the BABE digests are represented in Rust with an enum type, so when you encode them they're prefixed with a byte index

edwardmack: Yes, I've updated this, it's working now.

noot (approved):
looks good! only the few lint issues Quentin mentioned :)

lib/babe/verify.go
edwardmack: I removed //nolint tag completely, and linter no longer complains, has something changed?

lib/babe/verify.go
edwardmack: I removed //nolint tag completely, and linter no longer complains, has something changed?

lib/babe/verify.go
qdm12: Well yes, `golangci-lint` version wasn't pinned in the CI (see #1979), so maybe that was a false positive back months ago and now it doesn't detect it anymore.

lib/babe/verify.go
qdm12: Well yes, `golangci-lint` version wasn't pinned in the CI, so maybe that was a false positive back months ago and now it doesn't detect it anymore. We should remove these `// nolint` comments sometime soon since they could be hiding potentially bad code too.

PR #1913, opened 2021-10-21
dot/network: create blockResponseBuf pool

### LINKED ISSUE #1858: dot/network: create `blockResponseBuf` pool
## Task summary

- currently the network contains 1 `blockResponseBuf` of `maxBlockResponseSize` which is used to read block responses into
- we pre-allocate the buffer as `malloc`s are expensive and slow down syncing by a lot
- however this may slow down syncing if there are parallel workers
- need to implement a pre-allocated pool of `[maxBlockResponseSize]byte` 
- see existing `newSizedBufferPool` for inspiration (difference is only buffer size)

zjhuber:
## Changes

- Created pool for blockResponseBuf and updated message size to work for both pools.

## Tests

```
go test ./dot/network
```

## Issues

-Fixes #1858 

## Primary Reviewer 

- @noot

### DIFF
--- a/dot/network/service.go
+++ b/dot/network/service.go
@@ -46,7 +46,9 @@ const (
 	blockAnnounceID = "/block-announces/1"
 	transactionsID  = "/transactions/1"
 
-	maxMessageSize = 1024 * 63 // 63kb for now
+	// maxMessageSize       = 1024 * 63       // 63kb for now
+	maxMessageSize       = 1024 * 1024 * 4 // 4mb
+	maxBlockResponseSize = 1024 * 1024 * 4 // 4mb
 
 	gssmrIsMajorSyncMetric = "gossamer/network/is_major_syncing"
 )
@@ -98,8 +100,8 @@ type Service struct {
 	telemetryInterval time.Duration
 	closeCh           chan interface{}
 
-	blockResponseBuf   []byte
-	blockResponseBufMu sync.Mutex
+	blockResponseBufPool *sizedBufferPool
+	blockResponseBufMu   sync.Mutex
 
 	batchSize int
 }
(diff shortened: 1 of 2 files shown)


qdm12 (commented):
@noot why do we have this `sizedBufferPool`? Can't we just use a `sync.Pool`? Lock contention wise, the `sync.Pool` scales much better than using channels or mutexes.

noot:
@qdm12 cause stuff gets garbage collected from `sync.Pool` :( and we don't want to be re-malloc-ing at runtime, as it's very CPU intensive

dot/network/service.go
```diff
@@ -157,6 +157,15 @@ func NewService(cfg *Config) (*Service, error) {
...
 	}
 
+	var blockResponseBufPool *sizedBufferPool
+	if cfg.noPreAllocate {
+		blockResponseBufPool = &sizedBufferPool{
+			c: make(chan *[maxMessageSize]byte, cfg.MinPeers*3),
+		}
+	} else {
+		blockResponseBufPool = newSizedBufferPool(cfg.MinPeers*3, cfg.MaxPeers*3)
+	}
```
qdm12, line 167: nit just some small simplifications 😉 

```suggestion
	poolSize := 3 * cfg.MaxPeers
	var preAlloc int
	if !cfg.noPreAllocate {
		preAlloc = 3 * cfg.MaxPeers
	}
	blockResponseBufPool := newSizedBufferPool(preAlloc, poolSize)
```

qdm12 (commented):
@noot @zjhuber is there a point in doing this? We were using a single shared byte slice as buffer and it was working fine right? Unless the implementation was wrong before and multiple threads were accessing that buffer erroneously, then yes a sizedBufferPool makes sense. What do you thinK?

noot:
@qdm12 honestly the code works fine as it is with the 1 buffer used for block responses, so I don't know if this is necessary. it would only be needed for the case when we're syncing at the tip and have multiple requests going and want to receive them all concurrently, but I don't think that even is needed as they aren't processed concurrently

noot:
gonna close this for now as I don't actually think it's needed. sorry about that!

PR #1920, opened 2021-10-22
feat(lib/runtime): Implement `trie_blake2_256_verify_proof` host function

### LINKED ISSUE #1798: `ext_trie_blake2_256_verify_proof_version_1`
Depended on #213

EclesioMeloJunior:
## Changes
- Add  `trie_blake2_256_verify_proof` host function
- Improve the `VerifyProof` to check if `proofTrie.Get(item.Key)` retuns an nil item
- Improve the `VerifyProof` to check if `item.Value == nil || len(item.Value) == 0` then dont match the values

## Tests

```
go test -run ^Test_ext_trie_blake2_256_verify_proof_version_1$ github.com/ChainSafe/gossamer/lib/runtime/wasmer
```

## Issues

- Closes #1798

## Primary Reviewer 

- @edwardmack


lib/trie/proof.go
```diff
@@ -100,9 +102,11 @@ func VerifyProof(proof [][]byte, root []byte, items []Pair) (bool, error) {
 
 	for _, item := range items {
 		recValue := proofTrie.Get(item.Key)
-
+		if recValue == nil {
+			return false, ErrKeyNotFound
+		}
 		// here we need to compare value only if the caller pass the value
-		if item.Value != nil && !bytes.Equal(item.Value, recValue) {
+		if len(item.Value) > 0 && !bytes.Equal(item.Value, recValue) {
```
qdm12, line 109: Do we need this len check if it's a byte slice?

This most likely suffices I think:

```suggestion
		if !bytes.Equal(item.Value, recValue) {
```

lib/trie/proof.go
```diff
@@ -27,21 +27,23 @@ import (
...
 
-	// ErrDuplicateKeys not allowed to verify proof with duplicate keys
+	// ErrKeyNotFound ...
+	ErrKeyNotFound = errors.New("expected key not found in the trie")
+
+	// ErrDuplicateKeys ...
 	ErrDuplicateKeys = errors.New("duplicate keys on verify proof")
 
-	// ErrLoadFromProof occurs when there are problems with the proof slice while building the partial proof trie
+	// ErrLoadFromProof ...
```
qdm12, line 42: I think you can remove these `Err*` comments now that the linter should not complain about them.

lib/runtime/wasmer/imports_test.go
```diff
@@ -1492,3 +1494,93 @@ func Test_ext_trie_blake2_256_root_version_1(t *testing.T) {
...
+		{root: root, key: []byte("do"), proof: proof, value: nil, expect: true},
+		{root: root, key: []byte("unknow"), proof: proof, value: nil, expect: false},
+		{root: root, key: []byte("unknow"), proof: proof, value: []byte("unknow"), expect: false},
+		{root: root, key: []byte("do"), proof: [][]byte{}, value: nil, expect: false},
+	}
+
+	inst := NewTestInstance(t, runtime.HOST_API_TEST_RUNTIME)
+
+	for _, testcase := range testcases {
+		hashEnc, err := scale.Marshal(testcase.root)
```
qdm12, line 1560: You can also run each test case in parallel with something like:

```suggestion
	for _, testcase := range testcases {
		testcase := testcase
		t.Run(testcase.name, func(t *testing.T) {
			t.Parallel()
			
			hashEnc, err := scale.Marshal(testcase.root)
```

Instead of adding a field `name` to testcase, you could also have your testcases as a `map[string]struct{...}` such that the key is the test case name.

Then you might want to run the test with the `-race` flag to make sure there is no data races, but that is optional I guess.

lib/runtime/wasmer/imports_test.go
```diff
@@ -1492,3 +1494,93 @@ func Test_ext_trie_blake2_256_root_version_1(t *testing.T) {
 	expected := tt.MustHash()
 	require.Equal(t, expected[:], hash)
 }
+
+func Test_ext_trie_blake2_256_verify_proof_version_1(t *testing.T) {
+	t.Parallel()
+
+	tmp, err := ioutil.TempDir("", "*-test-trie")
```
qdm12, line 1501: nit Remove the directory with a defer call 😉

qdm12 (approved):

lib/trie/proof.go
EclesioMeloJunior: In the substrate code base, the `item.Value` (aka expected value) is an `Option`, so if the value `is_some()` then they compare otherwise they just skip the comparison. Here we just skip the comparison if the `item.Value` is empty as we don't have any expected value

qdm12: That makes sense.

I thought `bytes.Equal` would already do a len check as a first if block, but it turns out it just compares the string values of both with might not be that efficient ugh.

lib/runtime/wasmer/imports_test.go
EclesioMeloJunior: Done!

EclesioMeloJunior:
Depends on https://github.com/ChainSafe/polkadot-spec/pull/9 merge

lib/runtime/test_helpers.go
```diff
@@ -80,30 +80,24 @@ func GetAbsolutePath(targetDir string) string {
...
 // GetRuntimeBlob checks if the test wasm @testRuntimeFilePath exists and if not, it fetches it from @testRuntimeURL
-func GetRuntimeBlob(testRuntimeFilePath, testRuntimeURL string) (n int64, err error) {
+func GetRuntimeBlob(testRuntimeFilePath, testRuntimeURL string) error {
 	if utils.PathExists(testRuntimeFilePath) {
-		return 0, nil
+		return nil
 	}
 
-	out, err := os.Create(testRuntimeFilePath)
+	/* #nosec */
```
qdm12, line 88: What does this do? 🤔

lib/runtime/test_helpers.go
```diff
@@ -80,30 +80,24 @@ func GetAbsolutePath(targetDir string) string {
...
+	resp, err := http.DefaultClient.Get(testRuntimeURL)
```
qdm12, line 89: Careful the default client has no timeout AFAIK, so you might want to create your own client with a timeout (either here or pass it as argument). Using the default client is not ideal either, since it might conflict with other tests.

lib/runtime/test_helpers.go
```diff
@@ -80,30 +80,24 @@ func GetAbsolutePath(targetDir string) string {
...
-	resp, err := http.Get(testRuntimeURL)
+	respBody, err := io.ReadAll(resp.Body)
 	if err != nil {
-		return 0, err
+		return err
 	}
-	defer func() {
-		_ = resp.Body.Close()
-	}()
+	defer resp.Body.Close() //nolint
```
qdm12, line 98: Specify the linter you want disabled. My guess is

```suggestion
	defer resp.Body.Close() //nolint:errcheck
```

Although you could also close it and check the error after writing it to the file, but up to you 😉

lib/runtime/wasmer/imports_test.go
```diff
@@ -1492,3 +1494,97 @@ func Test_ext_trie_blake2_256_root_version_1(t *testing.T) {
...
+		"Unknow key, proof should be false":           {root: root, key: []byte("unknow"), proof: proof, value: nil, expect: false},
+		"Key and value unknow, proof should be false": {root: root, key: []byte("unknow"), proof: proof, value: []byte("unknow"), expect: false},
+		"Empty proof, should be false":                {root: root, key: []byte("do"), proof: [][]byte{}, value: nil, expect: false},
+	}
+
+	inst := NewTestInstance(t, runtime.HOST_API_TEST_RUNTIME)
+
+	for name, testcase := range testcases {
+		t.Run(name, func(t *testing.T) {
+			hashEnc, err := scale.Marshal(testcase.root)
```
qdm12, line 1563: I think you should be able to run each subtest with `t.Parallel()` right?

qdm12 (approved):

lib/runtime/test_helpers.go
EclesioMeloJunior: This comment is there before I just keep it haha and I think is just a comment

qdm12: Ah ok ! You can try removing it. That's probably a comment to disable linting of a previous linter we used to use.

lib/runtime/wasmer/imports_test.go
EclesioMeloJunior: Done!

lib/runtime/test_helpers.go
EclesioMeloJunior: Done!

lib/runtime/test_helpers.go
EclesioMeloJunior: Changed to `http.Get`

lib/runtime/test_helpers.go
EclesioMeloJunior: Removed!

lib/runtime/test_helpers.go
```diff
@@ -85,8 +85,7 @@ func GetRuntimeBlob(testRuntimeFilePath, testRuntimeURL string) error {
 		return nil
 	}
 
-	/* #nosec */
-	resp, err := http.DefaultClient.Get(testRuntimeURL)
+	resp, err := http.Get(testRuntimeURL)
```
qdm12, line 88: That uses the DefaultClient as well 😄 You should create an http client in your helper function e.g. `client := &http.Client{Timeout: time.Second}`. Ideally we would have a global HTTP client for tests so connections are re-used, but since `test_helpers.go` does not end with `_test.go`, its code might be incorporated in the production binary so let's not do that.

Two things regarding get requests:

- We should use a custom http client (even for tests)
- We should use the `http.NewRequestWithContext` followed by `client.Do(request)` instead to have cancelable

EclesioMeloJunior: got it! implemented

lib/runtime/test_helpers.go
```diff
@@ -85,7 +87,18 @@ func GetRuntimeBlob(testRuntimeFilePath, testRuntimeURL string) error {
 		return nil
 	}
 
-	resp, err := http.Get(testRuntimeURL)
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, testRuntimeURL, nil)
```
qdm12, line 93: nit nit I'm going to sound annoying, but you could actually use `context.WithTimeout(context.Background(), someTimeout)` and the default http client 😄  But the current implementation is good too 👍!

PR #1921, opened 2021-10-22
feature(rpc): handle extrinsic status future and ready for rpc call author_submitAndWatchExtrinsic

### LINKED ISSUE #1535: author_submitAndWatchExtrinsic: handle extrinsic status future and ready
## Task summary

- RPC [author_submitAndWatchExtrinsic](https://github.com/w3f/PSPs/blob/master/PSPs/drafts/psp-6.md#189-author_submitandwatchextrinsic-pubsub) defines status "future" and "ready" for when extrinsic is added to future and ready queue.  These status notifications need to be implemented.
- cross reference what ready status means in substrate
- check if the future queue is really a hashmap

## Specification reference

- see https://github.com/w3f/PSPs/blob/master/PSPs/drafts/psp-6.md#189-author_submitandwatchextrinsic-pubsub
- Around line 302 in dot/rpc/subscription/websocket.go

-

kishansagathiya:
## Changes

- [x] handles extrinsic status "future" and "ready" for websocket call call author_submitAndWatchExtrinsic
- [x] Added relevant tests
- [ ] Handle Broadcast status

## Tests

Run `Test_SubmitAndWatchExtrinsic` from stress tests

## Issues

- Closes #1535 

## Primary Reviewer 

- @noot


dot/state/transaction.go
```diff
@@ -58,5 +66,47 @@ func (s *TransactionState) RemoveExtrinsicFromPool(ext types.Extrinsic) {
 
 // AddToPool adds a transaction to the pool
 func (s *TransactionState) AddToPool(vt *transaction.ValidTransaction) common.Hash {
+	s.notifyStatus(transaction.StatusNotification{Ext: vt.Extrinsic, Status: transaction.Future.String()})
 	return s.pool.Insert(vt)
 }
+
+// GetStatusNotifierChannel creates and returns a status notifier channel.
+func (s *TransactionState) GetStatusNotifierChannel() chan transaction.StatusNotification {
```
noot, line 74: it might be nice to also have this accept an `[]types.Extrinsic` of extrinsics to watch, so it only gets notified when those update, as opposed to all transactions pushing a notifications

dot/rpc/subscription/listeners.go
```diff
@@ -309,22 +311,28 @@ type ExtrinsicSubmitListener struct {
...
 	cancelTimeout time.Duration
 }
 
 // NewExtrinsicSubmitListener constructor to build new ExtrinsicSubmitListener
-func NewExtrinsicSubmitListener(conn *WSConn, extBytes []byte) *ExtrinsicSubmitListener {
-	esl := &ExtrinsicSubmitListener{
+func NewExtrinsicSubmitListener(conn *WSConn, extBytes []byte, importedChan chan *types.Block, txStatusChan chan transaction.StatusNotification, finalisedChan chan *types.FinalisationInfo) *ExtrinsicSubmitListener {
+	return &ExtrinsicSubmitListener{
 		wsconn:        conn,
 		extrinsic:     types.Extrinsic(extBytes),
```
noot, line 328: it seems that `extBytes` could be multiple extrinsics, not just 1 (ie. the comment in `initExtrinsicWatch`). is this still the case? if so, you'll need to scale.Unmarshal the `extBytes` into `[]types.Extrinsic`

tests/stress/stress_test.go
```diff
@@ -470,3 +471,88 @@ func TestSync_SubmitExtrinsic(t *testing.T) {
...
+
+	message := []byte(`{"id":1, "jsonrpc":"2.0", "method": "author_submitAndWatchExtrinsic", "params":["` + extEnc + `"]}`)
+
+	err = conn.WriteMessage(websocket.TextMessage, message)
+	require.NoError(t, err)
+
+	var result []byte
+	_, result, err = conn.ReadMessage()
+	require.NoError(t, err)
+	fmt.Println("result:", result)
```
edwardmack, line 557: Is this println for debugging?  It could be useful for monitoring the results for test, since the result is json perhaps `fmt.Printf("result: %s\n", result)` so that we can read result.  

Also, since this is a subscription we expect multiple results, first with subscription ID, then other results as events happen.  So, maybe add something like:
```
 time.Sleep(time.Second)
	_, result, err = conn.ReadMessage()
	require.NoError(t, err)
	fmt.Printf("res 2 %s\n", result)
```
a few times to see the subscription messages, however this is a risk of blocking the test if something goes wrong with node and test is blocked waiting for results that may never arrive.  

Not sure of the best way to handle this, the stress tests seem useful for testing nodes and debugging, but they don't seem that good when trying to automate in CI, so I'm not sure how best to deal with this test.

edwardmack (approved):
Looks good to me, just some comments regarding the stress test.

tests/stress/stress_test.go
kishansagathiya: >  this is a risk of blocking the test 

I think this could be tackled by adding a timer.

dot/state/transaction.go
kishansagathiya: addressed in the latest commit

dot/state/transaction.go
```diff
@@ -10,18 +12,26 @@ import (
...
 
 // NewTransactionState returns a new TransactionState
 func NewTransactionState() *TransactionState {
 	return &TransactionState{
-		queue: transaction.NewPriorityQueue(),
-		pool:  transaction.NewPool(),
+		queue:            transaction.NewPriorityQueue(),
+		pool:             transaction.NewPool(),
+		notifierChannels: make(map[chan transaction.Status]string),
+		// notifierChannels: make(map[string]chan transaction.Status),
```
noot, line 28: remove

noot (approved):
looks good!

dot/state/transaction.go
kishansagathiya: addressed in the latest comment

Issue #1845, opened 2021-10-06
Update logger package


ansermino:
## Issue summary

Afaict Gossamer still uses ChainSafe/log15 as the primary logger, which is an unmaintained fork. We should seriously consider replacing it with the upstream version, or better yet a different logger package with a nicer feature set.

## Other information and links

- Fork: https://github.com/chainSafe/log15
- Upstream: https://github.com/inconshreveable/log15
- Potential Alternative: https://github.com/rs/zerolog

PR #1928, opened 2021-10-25
chore(logging): replace `log15` library

qdm12:
## Changes

💁  First and foremost, my apologies for the rather huge PR. Although it's mostly trivial changes, clearing up some logs.

- [x] Add an `internal/log` package (initially tried to use zerolog, then standard log, then just re-invented the wheel since we needed custom caller code for our runtime packages - although that was rather quick to develop)
- [x] Has features we used from the log15 logger except context key values
- [x] Can still be used with a global scope and thread safely (although we should inject it ideally)
- [x] Thread safety logging issues fixed. They now use all the same mutex 100% sure.
- [x] Adapt all code to not use the older context key-values
   - [x] Review most log instructions and check if they make sense (doing it at the same time)
- [x] Some thread safety issues fixed (was using different logs all writing to stdout)
- [x] `internal/log` tests
    - [x] caller_test.go
    - [x] level_test.go
    - [x] log_test.go
    - [x] logger_test.go
    - [x] settings, options are well tested (full coverage) by the tests above
    - [x] patch_test.go
    - [x] global_test.go
    - [x] Data races tests (to protect it if changing it further in the future)
- [x] ~Modify `runtime/wasmer` when my language server is willing to compile it. **RESOLVED, WIP**~
- [x] Change all logging functions using `fmt.Sprintf` to formatters (e.g. `Debug` to `Debugf` using `fmt.Sprintf` to delay formatting depending on level)
- [x] ~Removed coloring for now since some shows as `[35mCRIT[0m[10-28|15:12:03] [...` in some terminals which is extremely confusing~ Change colouring to be 'clever' with fatih/color.
- [x] Move log implementation in a subpackage so we can have more implementations (e.g. test oriented logger)
- [ ] **Question** ⁉️ Do we actually care about the logs in tests? Can we change all use of the loggers by tests to have the writer `io.Discard`. One can always change it to `os.Stdout` if some debugging is needed.
- [x] Self review 126/126
- [x] Create issues
    - [x] Inject the logger everywhere instead of using the global one: #1946
    - [x] ~Disable logger coloring with flags etc.~
    - [x] String methods needed: #1945 

## Tests

```sh
CGO_ENABLED=1 go test -race github.com/ChainSafe/gossamer/internal/log/...
```

## Issues

- #1845 

## Primary Reviewer 

- None for now


internal/log/interface.go
```diff
@@ -0,0 +1,41 @@
+package log
+
+var _ Interface = (*Logger)(nil)
+
+// Interface is the interface encompassing all the methods of the logger.
+type Interface interface {
```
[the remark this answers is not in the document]
qdm12: Changed it back to a single implementation, no interface. If caller wants they can use the interfaces defined in the package and compose them for their own interfaces, if they want to support different implementations or mock it.

qdm12:
**Question** Substrate uses log levels:

> Log levels (least to most verbose) are error, warn, info, debug, and trace.

Should we change ours to match these strings? For now it's `eror`, `dbug` etc. all 4 letters.

timwu20 (approved):

dot/state/block_notify.go
```diff
@@ -75,7 +75,7 @@ func (bs *BlockState) notifyImported(block *types.Block) {
 		return
 	}
 
-	logger.Trace("notifying imported block chans...", "chans", bs.imported)
+	logger.Trace("notifying imported block channels...")
```
qdm12, line 78: I removed logging channels here since it would just log out a pointer, I'm not sure this is useful.

qdm12 (commented):
89/126 files self-reviewed

noot:
@qdm12 can you also add the files + line numbers to the log? like how our current logger has?

noot:
> **Question** Substrate uses log levels:
> 
> > Log levels (least to most verbose) are error, warn, info, debug, and trace.
> 
> Should we change ours to match these strings? For now it's `eror`, `dbug` etc. all 4 letters.

yes, that would be nice imo! gossamer currently supports those also

qdm12:
> yes, that would be nice imo! gossamer currently supports those also

Yes I noticed, my last commit https://github.com/ChainSafe/gossamer/pull/1928/commits/d11b8e7641b3bfa8b4bd67966e282a9ecf365bc8 adds support for both (short and long forms). I'm just wondering if we should drop support for the shorter form, so there is no confusion of what is valid/not valid. But we can also leave it as is.

dot/sync/chain_sync.go
```diff
@@ -336,26 +335,17 @@ func (cs *chainSync) logSyncSpeed() {
...
-			logger.Info("🚣 currently syncing",
-				"peer count", len(cs.network.Peers()),
-				"target", target,
-				"average blocks/second", cs.benchmarker.mostRecentAverage(),
-				"overall average", cs.benchmarker.average(),
-				"finalised", finalised.Number,
-				"hash", finalised.Hash(),
-			)
+			logger.Infof(
+				"🔗 imported blocks from %d to %d (hashes [%s ... %s ])",
```
noot, line 339:
```suggestion
				"🔗 imported blocks from %d to %d (hashes [%s ... %s])",
```

noot (approved):
looks good! just one question, all the log timestamps seem to have `04:00` at the end of them, what does this mean exactly and why is it there? eg:
```
2021-11-04T12:42:56-04:00 INFO 🔗 imported blocks from 81242 to 81385 (hashes [0xe57344bbf1ba9ec304575dd830c8d84d4d54a30882acc701685fc6593cc87666 ... 0x37264728b31e6c7ccccfec11c332055af9f304ea891a28ce09fc2cd5e27d1619 ])	chain_sync.go:L338pkg=sync
2021-11-04T12:42:56-04:00 INFO 🚣 currently syncing, 50 peers connected, target block number 7289079, 28.60 average blocks/second, 19.26 overall average, finalised block number 77824 with hash 0x8819565a6d7cfd953aa0c6e07d55627d4dd5cac446962167ab62caab53a9c739	chain_sync.go:L342	pkg=sync
2021-11-04T12:42:56-04:00 DBUG imported block 0x1544b12c2f167360f876cf673cd6818ae6b2776c668fc4c29f9c99b636adf555 and stored state trie with root 0x58f27f558a781c5dfc0ab395567ecf5b3df42192ede6af38cea3e7a737b57f37	service.go:L227	pkg=core
2021-11-04T12:42:56-04:00 DBUG imported block 0xb07b5ab59c9e2b430fd73a958e66b4271f518d29d95c4ab52371c93943178498 and stored state trie with root 0x3b6ee95430b629fbd8751169431279b4fafd56594211ac7a605aaae11d95f767	service.go:L227	pkg=core
2021-11-04T12:42:56-04:00 DBUG imported block 0x4046351ed1ab979a40a92f0bcc877d2292fc9e5215199112308ea14a56366ce6 and stored state trie with root 0x66fc3a841cc507c33cb506f7a0b9f166839600eb7b26d07e17a1266988a6e4cb	service.go:L227	pkg=core
2021-11-04T12:42:56-04:00 DBUG imported block 0x042e55ddfd3b52bc0c3f36e6bcddd2f1c0048192db7350325b17ac9f0730c6e4 and stored state trie with root 0x5753d07c26468a3ef545ee77096cc80bbd6e6a47b4a6e75ef418578ec9442bed	service.go:L227	pkg=core
2021-11-04T12:42:56-04:00 DBUG imported block 0x80aed6f21d6df4847f4599f2594f1bbdda3e057360946899b22ca75a5384698b and stored state trie with root 0xab06466479f1b3408e8cca9c7749da9568a03e325498bc9c3b4b9777055dbfa4	service.go:L227	pkg=core
2021-11-04T12:42:56-04:00 DBUG imported block 0x751cfe4c19d97a1fdcb17fc4d2f112d9c7a969d48dcedf737649d59cf34ed5bb and stored state trie with root 0x1ff07420317ba7c803bce68f970bb1dd4b50e891c35af92ae7fb35498d35ab9b	service.go:L227	pkg=core
```

qdm12:
> what does this mean exactly and why is it there

It's the timezone, so -4 compared to Zulu time (GMT). Let me know if you want a different time representation, this one is rfc3339 but there [are a few standardised ones](https://github.com/golang/go/blob/577287728a8e8abbcbc6c96fdd49feb5a99ced0c/src/time/format.go#L96-L102) if you want to change. I could also change that one to be set as Zulu time so `-04:00` would become `Z`, although times will be GMT times, let me know 😉 Note having the timezone information is kind of critical when you don't live at the same place your program does 😄

noot:
@qdm12 oh cool, yeah no need to change it, was just curious. thanks for the explanation!

dot/sync/chain_sync.go
qdm12: How did my own eyes miss this NOOoooo 😄 👀

qdm12:
(note to self) need to add copyright notices everywhere

PR #1940, opened 2021-10-28
chore(lint): disable shadowing checks with `govet`

qdm12:
## Changes

- Disable shadowing check (disabled by default) with `govet`
- In my opinion:
    - these make the code longer since we need to pre-declare things
    - you can't have locally-scoped variables in an if or for block for example if the variable name is already used in the upper scope.

Feel free to comment & close it if you disagree with the change 😉 

## Tests

```
golangci-lint run
```

## Issues

- No issue

## Primary Reviewer

- No-one

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -31,6 +31,10 @@ run:
 
 # all available settings of specific linters
 linters-settings:
+  govet:
+    # report about shadowed variables
+    check-shadowing: true
+
   maligned:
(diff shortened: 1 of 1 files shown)


qdm12:
Shadowing is re-enabled, but shadow detection of errors (`err`) is disabled.

kishansagathiya (approved):

arijitAD (approved):

PR #1942, opened 2021-10-29
fix(dot/network): Implement time based handle transaction

### LINKED ISSUE #1788: Implement time based batch processing of transcation message.
## Issue summary

- Implement time-based batch processing of transcation messages.
- Currently, it is limit based which might cause an issue if we don't get enough notification messages. 

- https://github.com/ChainSafe/gossamer/pull/1780#discussion_r708481396

arijitAD:
## Changes

-  Implement time based handle transaction

## Tests

```
TestCreateNotificationsMessageHandler_HandleTransaction
```

## Issues

- closes #1788

## Primary Reviewer

- @EclesioMeloJunior


dot/network/transaction.go
```diff
@@ -119,36 +120,54 @@ func decodeTransactionHandshake(_ []byte) (Handshake, error) {
...
+		protocolID := s.host.protocolID + transactionsID
+		ticker := time.NewTicker(1 * time.Second)
+
+		for {
+		out:
+			select {
+			case <-s.ctx.Done():
+				return
+			case <-ticker.C:
+				innerTicker := time.NewTicker(300 * time.Millisecond)
```
noot, line 134: where did this time come from? and what's the purpose of it exactly?

dot/network/transaction.go
```diff
@@ -119,36 +120,54 @@ func decodeTransactionHandshake(_ []byte) (Handshake, error) {
 	return &transactionHandshake{}, nil
 }
 
-func (s *Service) createBatchMessageHandler(txnBatch chan *BatchMessage) NotificationsMessageBatchHandler {
-	return func(peer peer.ID, msg NotificationsMessage) (msgs []*BatchMessage, err error) {
+func (s *Service) createBatchMessageHandler(txnBatchCh chan *BatchMessage) NotificationsMessageBatchHandler {
+	go func() {
+		protocolID := s.host.protocolID + transactionsID
+		ticker := time.NewTicker(1 * time.Second)
```
noot, line 126: it would be nice to configure this using slot duration, i think 1/3 a slot duration is reasonable, that way the node has 1/3 a slot to process the transaction messages and put them in the pool if they want. what do you think?

dot/network/transaction.go
```diff
@@ -119,36 +120,54 @@ func decodeTransactionHandshake(_ []byte) (Handshake, error) {
...
-				})
-			}
-			if len(txnBatch) == 0 {
-				break
-			}
+		select {
+		case txnBatchCh <- data:
+		case <-time.After(time.Millisecond * 200):
+			logger.Debug("transaction message not included into batch")
+			return
```
EclesioMeloJunior, line 169: nit `return` not needed

dot/network/notifications.go
```diff
@@ -58,7 +58,7 @@ type (
 	NotificationsMessageHandler = func(peer peer.ID, msg NotificationsMessage) (propagate bool, err error)
 
 	// NotificationsMessageBatchHandler is called when a (non-handshake) message is received over a notifications stream in batch processing mode.
```
jimjbrettj, line 60: This is just for me to gain understanding, but what is the distinction here between a handshake and a non-handshake message when in batch processing mode, and why do they have to be handled differently?

noot: when a notifications stream is opened (see docs on the different stream types here https://docs.rs/sc-network/0.9.0/sc_network/) the first message each side sends to each other over it is a "handshake" message, which each side validates and if they don't like it they close the stream, otherwise it stays open and the rest of the messages that are sent over the stream (by the opening side only) are the normal messages for that stream protocol, so in this case `TransactionMessage`s

jimjbrettj: Ah okay that makes sense. I wasn't sure if handshake meant something different in this context, but this sounds like what I would think of as a typical handshake for establishing connection between peers (correct me if I'm misunderstanding though). And to clarify, the streams are one sided in this case?

dot/network/transaction.go
arijitAD: The external tick triggers at a regular interval of `slotDuration`. The inner tick was to trigger timeout, so that it doesn't take the entire `slotDuration`. I have updated the code to use a timer.

dot/network/transaction.go
arijitAD: Done

dot/network/transaction.go
arijitAD: Done

dot/network/notifications.go
arijitAD: Yes, After handshake the steam is unidirectional.

noot: yeah, it's not exactly the same as a handshake for estabilishing connection/public keys, but it's a more high-level stream level handshake

dot/network/transaction.go
```diff
@@ -119,36 +120,53 @@ func decodeTransactionHandshake(_ []byte) (Handshake, error) {
...
-					peer: txnData.peer,
-				})
-			}
-			if len(txnBatch) == 0 {
-				break
-			}
+		select {
+		case txnBatchCh <- data:
+		case <-time.After(time.Millisecond * 200):
+			logger.Debug("transaction message not included into batch")
```
noot, line 168: maybe add the tx that's not being included to the log?

noot (approved):
great work :D

dot/network/transaction.go
arijitAD: Done

EclesioMeloJunior (changes requested):

dot/network/transaction.go
```diff
@@ -119,36 +120,53 @@ func decodeTransactionHandshake(_ []byte) (Handshake, error) {
...
+		out:
+			select {
+			case <-s.ctx.Done():
+				return
+			case <-ticker.C:
+				timeOut := time.NewTimer(s.cfg.SlotDuration / 3)
+				for {
+					select {
+					case <-timeOut.C:
+						break out
```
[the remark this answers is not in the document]
arijitAD: We don't need to add a stop here because `timeOut` will be already triggered.

dot/network/transaction.go
```diff
@@ -119,36 +120,53 @@ func decodeTransactionHandshake(_ []byte) (Handshake, error) {
 	return &transactionHandshake{}, nil
 }
 
-func (s *Service) createBatchMessageHandler(txnBatch chan *BatchMessage) NotificationsMessageBatchHandler {
-	return func(peer peer.ID, msg NotificationsMessage) (msgs []*BatchMessage, err error) {
+func (s *Service) createBatchMessageHandler(txnBatchCh chan *BatchMessage) NotificationsMessageBatchHandler {
+	go func() {
+		protocolID := s.host.protocolID + transactionsID
+		ticker := time.NewTicker(s.cfg.SlotDuration)
```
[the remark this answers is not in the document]
arijitAD: Done

dot/network/transaction.go
```diff
@@ -119,36 +120,53 @@ func decodeTransactionHandshake(_ []byte) (Handshake, error) {
...
+			select {
+			case <-s.ctx.Done():
+				return
+			case <-ticker.C:
+				timeOut := time.NewTimer(s.cfg.SlotDuration / 3)
+				for {
+					select {
+					case <-timeOut.C:
+						break out
+					case txnMsg := <-txnBatchCh:
```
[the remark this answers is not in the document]
arijitAD: We are processing as many txns as possible within `timeOut `. If there is no data in channel then `timeOut` will be triggered.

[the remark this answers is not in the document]
EclesioMeloJunior: But I'm asking if the process of a `txnMsg` takes for more than a ticker duration, for example lets say `handleTransactionMessage` blocks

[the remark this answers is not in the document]
arijitAD: Then we don't process it. We process it in the next block.

dot/network/transaction.go
```diff
@@ -119,36 +120,54 @@ func decodeTransactionHandshake(_ []byte) (Handshake, error) {
...
+						propagate, err := s.handleTransactionMessage(txnMsg.peer, txnMsg.msg)
```
EclesioMeloJunior, line 141: My question is: the `txnMsg` arrived but the following line takes more than a ticker duration
```go
propagate, err := s.handleTransactionMessage(txnMsg.peer, txnMsg.msg)
```
is that a problem?

EclesioMeloJunior (dismissed):

dot/network/transaction.go
arijitAD: It won't be a problem as we need to process the extrinsic sequentially. We cannot spin a goroutine.

dot/network/config.go
```diff
@@ -52,6 +52,9 @@ const (
 
 	// DefaultDiscoveryInterval is the default interval for searching for DHT peers
 	DefaultDiscoveryInterval = time.Minute * 5
+
+	// defaultTxnBatchSize is the default size for the transaction batch
+	defaultTxnBatchSize = 100
```
qdm12, line 57: nit I know `tx` is widely used in the blockchain space, but let's perhaps use full words for the sake of readability / newcomers?

```suggestion
	defaultTransactionBatchSize = 100
```

dot/network/config.go
```diff
@@ -104,6 +107,11 @@ type Config struct {
 	telemetryInterval time.Duration
 
 	noPreAllocate bool // internal option
+
+	batchSize int // internal option
+
+	// Babe slot duration
+	SlotDuration time.Duration
```
qdm12, line 114: nit shouldn't the comment begin with `SlotDuration`? Also maybe rename it to `BabeSlotDuration`?

dot/network/host.go
```diff
@@ -394,3 +394,17 @@ func (h *host) protocols() []string {
...
+func (h *host) closeProtocolStream(pID protocol.ID, p peer.ID) {
+	connToPeer := h.h.Network().ConnsToPeer(p)
+	for _, c := range connToPeer {
+		for _, st := range c.GetStreams() {
+			if st.Protocol() == pID {
+				err := st.Close()
+				if err != nil {
+					logger.Tracef("Failed to close stream", "protocol", pID, "error", err)
+				}
+			}
```
qdm12, line 407: nit less indentation

```suggestion
			if st.Protocol() != pID {
				continue
			}
			err := st.Close()
			if err != nil {
				logger.Tracef("Failed to close stream", "protocol", pID, "error", err)
			}
```

dot/network/host.go
```diff
@@ -394,3 +394,17 @@ func (h *host) protocols() []string {
...
 }
+
+func (h *host) closeProtocolStream(pID protocol.ID, p peer.ID) {
+	connToPeer := h.h.Network().ConnsToPeer(p)
+	for _, c := range connToPeer {
+		for _, st := range c.GetStreams() {
+			if st.Protocol() == pID {
+				err := st.Close()
+				if err != nil {
+					logger.Tracef("Failed to close stream", "protocol", pID, "error", err)
```
qdm12, line 405: Also this will have to change when merging development with the newer log package (`internal/log`).

dot/network/notifications.go
```diff
@@ -220,46 +220,31 @@ func (s *Service) createNotificationsMessageHandler(info *notificationsProtocol,
...
-				s.broadcastExcluding(info, data.peer, data.msg)
-			}
-
-			// report peer if we get duplicate gossip message.
-			s.host.cm.peerSetHandler.ReportPeer(peerset.ReputationChange{
-				Value:  peerset.DuplicateGossipValue,
-				Reason: peerset.DuplicateGossipReason,
-			}, peer)
+		seen := s.gossip.hasSeen(msg)
+		if !seen {
```
qdm12, line 238: nit

```suggestion
		if !s.gossip.hasSeen(msg) {
```

dot/network/service_test.go
```diff
@@ -84,11 +86,13 @@ func createTestService(t *testing.T, cfg *Config) (srvc *Service) {
 
 	if cfg.TransactionHandler == nil {
 		mocktxhandler := &MockTransactionHandler{}
-		mocktxhandler.On("HandleTransactionMessage", mock.AnythingOfType("*TransactionMessage")).Return(nil)
+		mocktxhandler.On("HandleTransactionMessage", mock.AnythingOfType("*network.TransactionMessage")).Return(true, nil)
 		mocktxhandler.On("TransactionsCount").Return(0)
 		cfg.TransactionHandler = mocktxhandler
 	}
 
+	cfg.SlotDuration = time.Second
```
qdm12, line 94: Dumb question: can we have this smaller to have faster tests?

dot/network/transaction.go
```diff
@@ -119,36 +120,54 @@ func decodeTransactionHandshake(_ []byte) (Handshake, error) {
...
-func (s *Service) createBatchMessageHandler(txnBatch chan *BatchMessage) NotificationsMessageBatchHandler {
-	return func(peer peer.ID, msg NotificationsMessage) (msgs []*BatchMessage, err error) {
+func (s *Service) createBatchMessageHandler(txnBatchCh chan *BatchMessage) NotificationsMessageBatchHandler {
+	go func() {
+		protocolID := s.host.protocolID + transactionsID
+		ticker := time.NewTicker(s.cfg.SlotDuration)
+		defer ticker.Stop()
+
+		for {
+		out:
```
qdm12, line 130: Please get rid of the break + label, it adds unneeded complexity and is rather uncommon in Go code.

You can just use boolean and some `continue`/ `break` to control the flow.

dot/network/transaction.go
```diff
@@ -119,36 +120,54 @@ func decodeTransactionHandshake(_ []byte) (Handshake, error) {
...
+			select {
+			case <-s.ctx.Done():
+				return
+			case <-ticker.C:
+				timeOut := time.NewTimer(s.cfg.SlotDuration / 3)
+				for {
+					select {
+					case <-timeOut.C:
+						break out
+					case txnMsg := <-txnBatchCh:
```
qdm12, line 140: I think that will cause a memory leak. We should stop the timer and drain its channel

```go
if !timer.Stop() {
	<-timer.C
}
```

dot/network/transaction.go
qdm12: The ticker duration won't be respected but there is not much solution there. Although, I'm not super familiar with the context, but my question:

if ticker duration `s.cfg.SlotDuration` is say `10` and processing takes `15`, are we fine executing the next processing at t+15 (instead of t+20)? Since the ticker will have its (1-buffered) channel blocked and will send its tick as soon as we finish our processing at `t+15`.

dot/network/transaction.go
```diff
@@ -119,36 +120,54 @@ func decodeTransactionHandshake(_ []byte) (Handshake, error) {
...
-					msg:  txnData.msg,
-					peer: txnData.peer,
-				})
-			}
-			if len(txnBatch) == 0 {
-				break
-			}
+		select {
+		case txnBatchCh <- data:
+		case <-time.After(time.Millisecond * 200):
```
qdm12, line 168: Use a timer you stop when the other case triggers instead, this will cause a leak / GC pressure (goroutine won't stop if we get data, and its channel won't be emptied)

dot/network/transaction.go
arijitAD: Ideally, it will take very little time since we are just verifying transactions. But even if it does, it shouldn't matter since we are only adding transactions to the memory pool.

dot/network/config.go
arijitAD: It will make the variable name very long which I am trying to avoid.

dot/network/config.go
arijitAD: Done

arijitAD: I have removed `Babe` from comments since it is the slot duration to produce block but the algorithm may change in future.

dot/network/host.go
arijitAD: Done

dot/network/notifications.go
arijitAD: Done

dot/network/transaction.go
arijitAD: Done

dot/network/transaction.go
arijitAD: @qdm12 Can you provide more details why there will be a memory leak?
We only exit the loop when we hit the timeout. So every exit from the loop drains the timer channel.

dot/network/transaction.go
arijitAD: Done

dot/network/host.go
arijitAD: Done

dot/network/service_test.go
arijitAD: Reduced it to `100ms`

qdm12: yet another dumb question, is there a reason we can't have it to 1ns?

dot/network/transaction.go
```diff
@@ -106,36 +107,59 @@ func decodeTransactionHandshake(_ []byte) (Handshake, error) {
...
 		data := &BatchMessage{
 			msg:  msg,
 			peer: peer,
 		}
-		txnBatch <- data
 
-		if len(txnBatch) < s.batchSize {
-			return nil, nil
-		}
+		timeOut := time.NewTimer(time.Millisecond * 200)
```
qdm12, line 153: What's the reasoning behind the 200ms value?

Maybe name it as a constant together with a comment?

dot/services.go
```diff
@@ -247,6 +248,11 @@ func createNetworkService(cfg *Config, stateSrvc *state.Service) (*network.Servi
 		cfg.Core.Roles, cfg.Network.Port, strings.Join(cfg.Network.Bootnodes, ","), cfg.Network.ProtocolID,
 		cfg.Network.NoBootstrap, cfg.Network.NoMDNS)
 
+	slotDuration, err := stateSrvc.Epoch.GetSlotDuration()
+	if err != nil {
+		return nil, err
```
qdm12, line 253: We should add some context to the error. i.e.

```suggestion
		return nil, fmt.Errorf("cannot get slot duration: %w", err)
```

dot/network/service_test.go
qdm12: No don't worry about it. I was curious if we could set it to an arbitrary value and have it event driver rather than time driven. Since we already have so many tests using sleeps and all, it's fine to have one more I guess

dot/network/transaction.go
qdm12: I don't think it's up yet, did you push the changes?

dot/services.go
arijitAD: Done

dot/network/transaction.go
arijitAD: We wait `200ms` if the channel is full. Not sure what should be the exact value. Any thoughts?
Named it `txnBatchChTimeout`

dot/network/transaction.go
```diff
@@ -106,36 +107,59 @@ func decodeTransactionHandshake(_ []byte) (Handshake, error) {
...
-			}
-			if len(txnBatch) == 0 {
-				break
+		select {
+		case txnBatchCh <- data:
+			if !timeOut.Stop() {
+				<-timeOut.C
 			}
+		case <-timeOut.C:
+			logger.Debugf("transaction message not included into batch", "peer", peer.String(), "msg", msg.String())
```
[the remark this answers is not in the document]
arijitAD: Done

dot/network/transaction.go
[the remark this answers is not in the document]
arijitAD: Done

dot/network/transaction.go
```diff
@@ -106,36 +107,59 @@ func decodeTransactionHandshake(_ []byte) (Handshake, error) {
 	return &transactionHandshake{}, nil
 }
 
-func (s *Service) createBatchMessageHandler(txnBatch chan *BatchMessage) NotificationsMessageBatchHandler {
-	return func(peer peer.ID, msg NotificationsMessage) (msgs []*BatchMessage, err error) {
+func (s *Service) createBatchMessageHandler(txnBatchCh chan *BatchMessage) NotificationsMessageBatchHandler {
+	go func() {
```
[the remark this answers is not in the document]
arijitAD: Done

dot/network/host.go
```diff
@@ -381,3 +381,18 @@ func (h *host) protocols() []string {
...
+func (h *host) closeProtocolStream(pID protocol.ID, p peer.ID) {
+	connToPeer := h.h.Network().ConnsToPeer(p)
+	for _, c := range connToPeer {
+		for _, st := range c.GetStreams() {
+			if st.Protocol() != pID {
+				continue
+			}
+			err := st.Close()
+			if err != nil {
+				logger.Tracef("Failed to close stream", "protocol", pID, "error", err)
```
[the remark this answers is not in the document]
arijitAD: Done

dot/network/config.go
```diff
@@ -91,6 +95,11 @@ type Config struct {
 	telemetryInterval time.Duration
 
 	noPreAllocate bool // internal option
+
+	batchSize int // internal option
+
+	// SlotDuration is slot duration to produce block in milliseconds
```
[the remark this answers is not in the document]
arijitAD: Done

dot/network/config.go
arijitAD: Done

dot/network/config.go
```diff
@@ -98,7 +97,7 @@ type Config struct {
 
 	batchSize int // internal option
 
-	// SlotDuration is slot duration to produce block in milliseconds
+	// SlotDuration is slot duration to produce block
```
qdm12, line 100:
```suggestion
	// SlotDuration is the slot duration to produce a block
```

dot/network/transaction.go
```diff
@@ -107,58 +110,60 @@ func decodeTransactionHandshake(_ []byte) (Handshake, error) {
...
+	protocolID := s.host.protocolID + transactionsID
+	ticker := time.NewTicker(s.cfg.SlotDuration)
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-s.ctx.Done():
+			return
+		case <-ticker.C:
+			timeOut := time.NewTimer(s.cfg.SlotDuration / 3)
```
qdm12, line 123: nit this should be called `timer` instead of `timeOut`

dot/network/transaction.go
```diff
@@ -107,58 +110,60 @@ func decodeTransactionHandshake(_ []byte) (Handshake, error) {
...
+			var completed bool
+			for !completed {
```
qdm12, line 125: nit you could rename `completed` to `timedOut`

dot/network/transaction.go
qdm12: My bad, I got confused with the goto and label!

dot/network/transaction.go
qdm12: Should we have the batch size-based (event driven) on top of time-based?
In case there is a lot of messages coming in as a burst.
What do you think?

arijitAD: Having event-based will cause issues because in order to validate transactions we take a lock on the state trie. Block production might suffer due to this.

dot/network/config.go
arijitAD: Done

dot/network/transaction.go
arijitAD: Done

dot/network/transaction.go
arijitAD: Done

qdm12 (approved):

dot/network/mock_transaction_handler.go
```diff
@@ -12,20 +12,20 @@ type MockTransactionHandler struct {
...
+	if rf, ok := ret.Get(0).(func(*TransactionMessage) bool); ok {
+		r0 = rf(_a0)
 	} else {
 		r0 = ret.Get(0).(bool)
 	}
 
 	var r1 error
-	if rf, ok := ret.Get(1).(func(peer.ID, *TransactionMessage) error); ok {
-		r1 = rf(_a0, _a1)
+	if rf, ok := ret.Get(1).(func(*TransactionMessage) error); ok {
```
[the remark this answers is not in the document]
arijitAD: Done. Fixed it.

PR #1943, opened 2021-10-29
update(libp2p): Update libp2p to latest version

### LINKED ISSUE #1938: Update `libp2p` Dependency to Newer Version

kishansagathiya:
## Changes

- Updated libp2p to latest version, also update other related packages

## Tests

If CI passes, this should be mostly enough.

```

```

## Issues

- Closes #1938 

## Primary Reviewer

- @timwu20


arijitAD:
Looks good. Maybe we should add a task to remove `mdns_legacy`?

arijitAD (approved):

go.mod
```diff
@@ -5,7 +5,7 @@ require (
 	github.com/ChainSafe/go-schnorrkel v0.0.0-20210318173838-ccb5cd955283
 	github.com/ChainSafe/log15 v1.0.0
 	github.com/OneOfOne/xxhash v1.2.5
-	github.com/btcsuite/btcutil v1.0.2
+	github.com/btcsuite/btcutil v1.0.3-0.20201208143702-a53e38424cce
```
qdm12, line 8: Can we stay on v1.0.3 perhaps? Or is there a point using the master/main branch?

qdm12 (approved):

EclesioMeloJunior (approved):

go.mod
kishansagathiya: This is something that happens when I do, `go mod tidy` after updating libp2p and some other dependent modules.
Not sure how `go.mod` decides on which version to update to.
I am assuming that `go mod` would have some good checks for 1) when to update a version and 2) which version to update to .

For now it seems safe to just go with whatever `go mod` has done. I see that this is the only case where version is getting updated to a commit, while other updates were to a release version. So, there might be something that requires this.

qdm12: Try modifying `go.mod` and set it to `github.com/btcsuite/btcutil v1.0.3` manually, then run `go mod tidy`. If it changes it again to `github.com/btcsuite/btcutil v1.0.3-0.20201208143702-a53e38424cce` then indeed it has its reasons somehow.

qdm12: Oh their latest tag is `v1.0.2` ([tags](https://github.com/btcsuite/btcutil/tags)). Maybe pin it to v1.0.2? v1.0.3-xxx is the master branch, not totally ideal safety wise.

Issue #1802, opened 2021-09-22
Implement Offchain HTTP API


danforbes:
Implement low level HTTP client which mirrors substrate asynch communication.

Spec has details regarding this,  (@arijitAD add spec link)

EclesioMeloJunior:
- [ ] ext_offchain_http_request_start
- [ ] ext_offchain_http_request_add_header
- [ ] ext_offchain_http_request_write_body
- [ ] ext_offchain_http_response_wait
- [ ] ext_offchain_http_response_headers
- [ ] ext_offchain_http_response_read_body

PR #1947, opened 2021-10-31
feat(lib/runtime): Implement `ext_offchain_http_request_start_version_1` host function

EclesioMeloJunior:
## Changes

- Add `ext_offchain_http_request_start_version_1` host function
- Create a HTTP set using a singleton pattern to manage HTTP requests
- Depends on https://github.com/ChainSafe/polkadot-spec/pull/10

## Tests

```
go test -run ^Test_ext_offchain_http_request_start_version_1$ github.com/ChainSafe/gossamer/lib/runtime/wasmer
```

## Issues

- Related to #1802

## Primary Reviewer

- @arijitAD


lib/runtime/offchain/httpset.go
```diff
@@ -0,0 +1,62 @@
...
+	p.reqs = append(p.reqs, req)
+	return int16(len(p.reqs) - 1), nil
+}
+
+// Remove just remove a expecific request from reqs
+func (p *Set) Remove(id int) {
+	p.mtx.Lock()
+	defer p.mtx.Unlock()
+
+	p.reqs = append(p.reqs[:id], p.reqs[id+1:]...)
```
arijitAD, line 52: We should use a map instead of a slice because when we are removing we are shifting the index and if we call get again after removal we will get an incorrect index.
```
arr : [1,2,3,4,5,6]
Remove(2)
arr: [1,2,4,5,6]
Get(3) will return 5 instead of 4.
```

arijitAD: Also, the `Get` logic of length checking will fail.

lib/runtime/wasmer/imports_test.go
```diff
@@ -232,6 +232,59 @@ func Test_ext_offchain_local_storage_clear_version_1_Local(t *testing.T) {
 	require.Nil(t, val)
 }
 
+func Test_ext_offchain_http_request_start_version_1(t *testing.T) {
```
arijitAD, line 235: `Test_ext_offchain_http_request_start_version_1` is getting failed. Implement `rtm_ext_offchain_http_request_start_version_1` in hostapi and update hostApi wasm url.

lib/runtime/wasmer/imports.go
```diff
@@ -1682,6 +1684,48 @@ func ext_offchain_sleep_until_version_1(_ unsafe.Pointer, deadline C.int64_t) {
...
+func ext_offchain_http_request_start_version_1(context unsafe.Pointer, methodSpan, uriSpan, metaSpan C.int64_t) C.int64_t {
+	logger.Debug("[ext_offchain_submit_transaction_version_1] executing...")
+
+	instanceContext := wasm.IntoInstanceContext(context)
+
+	httpMethod := asMemorySlice(instanceContext, methodSpan)
+	uri := asMemorySlice(instanceContext, uriSpan)
+
+	reqID, err := offchain.HTTPSet.StartRequest(string(httpMethod), string(uri))
+	if err != nil {
```
arijitAD, line 1697: Set the result to error.
```
err = result.Set(scale.Err, nil)
```

lib/runtime/wasmer/imports_test.go
```diff
@@ -232,6 +232,59 @@ func Test_ext_offchain_local_storage_clear_version_1_Local(t *testing.T) {
...
+	inst := NewTestInstance(t, runtime.HOST_API_TEST_RUNTIME)
+
+	encMethod, err := scale.Marshal([]byte("GET"))
+	require.NoError(t, err)
+
+	encUri, err := scale.Marshal([]byte("https://chainsafe.io"))
```
arijitAD, line 241: Can you start an HTTP mock server and request that? Because this test will fail without the internet or if Chainsafe domain is down.

lib/runtime/offchain/httpset.go
```diff
@@ -0,0 +1,62 @@
+package offchain
+
+import (
+	"net/http"
+	"sync"
+)
+
+var (
```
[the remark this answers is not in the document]
noot: I think this is probably ok as we might have multiple runtime instances but only one offchain HTTP handler. but I guess, we could always just have the runtimes point to the same handler

lib/runtime/wasmer/imports.go
```diff
@@ -1682,6 +1684,48 @@ func ext_offchain_sleep_until_version_1(_ unsafe.Pointer, deadline C.int64_t) {
 	logger.Warn("unimplemented")
 }
 
+//export ext_offchain_http_request_start_version_1
+func ext_offchain_http_request_start_version_1(context unsafe.Pointer, methodSpan, uriSpan, metaSpan C.int64_t) C.int64_t {
+	logger.Debug("[ext_offchain_submit_transaction_version_1] executing...")
```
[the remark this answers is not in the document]
noot: I think we updated the logget to already log the function name, so you can probably remove the function name entirely

lib/runtime/wasmer/imports_test.go
EclesioMeloJunior: the host function does not execute any request so this is just a URI formatted string to the runtime

lib/runtime/wasmer/imports.go
EclesioMeloJunior: Done

lib/runtime/wasmer/imports.go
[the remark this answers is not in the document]
EclesioMeloJunior: Done

lib/runtime/offchain/httpset.go
```diff
@@ -0,0 +1,103 @@
...
+	errIntBufferEmpty        = errors.New("int buffer exhausted")
+	errIntBufferFull         = errors.New("int buffer is full")
+	errRequestIDNotAvailable = errors.New("request id not available")
+)
+
+// requestIDBuffer created to control the amount of available non-duplicated ids
+type requestIDBuffer chan int16
+
+// newIntBuffer creates the request id buffer starting from 1 till @buffSize (by default @buffSize is 1000)
+func newIntBuffer(buffSize int16) *requestIDBuffer {
```
noot, line 21: does this need to return a pointer? you can change the receiver methods below to be `(b requestIDBuffer)`, then you don't need to dereference it in the methods

lib/runtime/offchain/httpset.go
```diff
@@ -0,0 +1,103 @@
...
+		return nil
+	default:
+		return errIntBufferFull
+	}
+}
+
+// HTTPSet holds a pool of concurrent http request calls
+type HTTPSet struct {
+	mtx    *sync.Mutex
+	reqs   map[int16]*http.Request
```
noot, line 52: is there a chance this map will be accesses concurrently? I think potentially

EclesioMeloJunior: yes! There is mutexes locks in `StartRequest` and `Remove` and I've added locks on `Get` tho

lib/runtime/offchain/httpset.go
EclesioMeloJunior: Done!

lib/runtime/offchain/httpset.go
```diff
@@ -0,0 +1,105 @@
...
+	case b <- i:
+		return nil
+	default:
+		return errIntBufferFull
+	}
+}
+
+// HTTPSet holds a pool of concurrent http request calls
+type HTTPSet struct {
+	mtx    *sync.Mutex
```
noot, line 50: you can embed this, so you don't need to call `p.mtx.Lock()`, you can just call `p.Lock()` and same for `Unlock()`
```suggestion
type HTTPSet struct {
    sync.Mutex
    // etc.
}
```

noot (approved):
looks good - just embed the lock :)

lib/runtime/offchain/httpset.go
EclesioMeloJunior: Done!

arijitAD (approved):

Issue #1893, opened 2021-10-14
create CLI flags for package log levels


noot:
## Task summary

- currently the package log levels can only be set via the config file
- there should also be CLI flags for each package's log level
- `--log-<package-name>=level` look at substrate and match

qdm12:
Waiting on https://github.com/paritytech/substrate/issues/8025#issuecomment-956293219 to know what the possible target names are. For now I set them to:

```
log-core
log-sync
log-network
log-rpc
log-state
log-runtime
log-blockproducer
log-finalitygadget
```

PR #1953, opened 2021-11-01
feat(flags): read log levels from flags

qdm12:
## Changes

- [x] Read log levels per package from flags
- [x] Flags take precedence over toml configuration file
- [x] Refactor code for simplicity and testability
- [x] Find substrate all possible target names (waiting on https://github.com/paritytech/substrate/issues/8025) and match them - ~they may not answer anytime soon~
- [x] Rebase on development once #1940 is merged to pass linting
- [x] Apply Eclesio's suggestion

## Tests

```
go test ^Test_setLogConfig$ github.com/ChainSafe/gossamer/cmd/gossamer
go test ^Test_getLogLevel$ github.com/ChainSafe/gossamer/cmd/gossamer
```

## Issues

- #1893 

## Primary Reviewer

- @noot

### DIFF
--- a/cmd/gossamer/config.go
+++ b/cmd/gossamer/config.go
@@ -17,6 +17,7 @@
 package main
 
 import (
+	"errors"
 	"fmt"
 	"strconv"
 	"strings"
@@ -310,9 +311,14 @@ func getLogLevel(ctx getStringer, flagName, tomlValue string, defaultLevel log.L
 	return parseLogLevelString(tomlValue)
 }
 
+var ErrLogLevelIntegerOutOfRange = errors.New("log level integer can only be between 0 and 5 included")
+
 func parseLogLevelString(logLevelString string) (logLevel log.Lvl, err error) {
 	levelInt, err := strconv.Atoi(logLevelString)
 	if err == nil { // level given as an integer
+		if levelInt < 0 || levelInt > 5 {
+			return 0, fmt.Errorf("%w: log level given: %d", ErrLogLevelIntegerOutOfRange, levelInt)
+		}
 		logLevel = log.Lvl(levelInt)
 		return logLevel, nil
 	}
--- a/cmd/gossamer/config.go
+++ b/cmd/gossamer/config.go
@@ -18,7 +18,6 @@ package main
 
 import (
 	"fmt"
-	"regexp"
 	"strconv"
 	"strings"
 	"time"
@@ -311,14 +310,9 @@ func getLogLevel(ctx getStringer, flagName, tomlValue string, defaultLevel log.L
 	return parseLogLevelString(tomlValue)
 }
 
-var regexDigits = regexp.MustCompile("^[0-9]+$")
-
 func parseLogLevelString(logLevelString string) (logLevel log.Lvl, err error) {
-	if regexDigits.MatchString(logLevelString) {
-		levelInt, err := strconv.Atoi(logLevelString)
-		if err != nil { // should never happen
-			return 0, fmt.Errorf("cannot parse log level digits: %w", err)
-		}
+	levelInt, err := strconv.Atoi(logLevelString)
+	if err == nil { // level given as an integer
 		logLevel = log.Lvl(levelInt)
 		return logLevel, nil
 	}
--- a/cmd/gossamer/config.go
+++ b/cmd/gossamer/config.go
@@ -392,7 +392,7 @@ func setLogConfig(flagsKVStore stringKVStore, cfg *ctoml.Config, globalCfg *dot.
 		},
 		{
 			name:      "finality gadget",
-			flagName:  LogFinalityGadgetLevelFlag.Name,
+			flagName:  LogGrandpaLevelFlag.Name,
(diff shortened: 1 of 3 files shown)


cmd/gossamer/config.go
```diff
@@ -287,115 +288,98 @@ func createExportConfig(ctx *cli.Context) (*dot.Config, error) {
...
-			return err
+func parseLogLevelString(logLevelString string) (logLevel log.Lvl, err error) {
+	if regexDigits.MatchString(logLevelString) {
+		levelInt, err := strconv.Atoi(logLevelString)
+		if err != nil { // should never happen
+			return 0, fmt.Errorf("cannot parse log level digits: %w", err)
 		}
+		logLevel = log.Lvl(levelInt)
+		return logLevel, nil
+	}
```
noot, line 324: is the regex needed? can you just do `Atoi` and if the int is <0 or >4 then error? although, I guess the following lines achieve the same thing?

cmd/gossamer/flags.go
```diff
@@ -56,9 +56,50 @@ var (
...
+	LogRuntimeLevelFlag = cli.StringFlag{
+		Name:  "log-runtime",
+		Usage: "Runtime package log level. Supports levels crit (silent), eror, warn, info, dbug and trce (trace)",
+		Value: LogFlag.Value,
+	}
+	LogBlockProducerLevelFlag = cli.StringFlag{
+		Name:  "log-blockproducer",
+		Usage: "Block producer package log level. Supports levels crit (silent), eror, warn, info, dbug and trce (trace)",
+		Value: LogFlag.Value,
+	}
```
noot, line 96: this is sort of a long name, I'd probably just change it to `log-babe` for now

cmd/gossamer/flags.go
```diff
@@ -56,9 +56,50 @@ var (
...
+	LogFinalityGadgetLevelFlag = cli.StringFlag{
+		Name:  "log-finalitygadget",
+		Usage: "Finality Gadget package log level. Supports levels crit (silent), eror, warn, info, dbug and trce (trace)",
+		Value: LogFlag.Value,
+	}
```
noot, line 101: same here, would just change to `log-grandpa`

qdm12: done

cmd/gossamer/config.go
qdm12: I put this up there mostly for readability. Reverted with extra comment.

I also added a check for the integer value to be between `0` and `5` with unit test.

noot (approved):
nice work :D

qdm12:
For the target names, I asked on the substrate repository and their answer was

> No there is no documentation on all the possible levels. I would also not recommend to copy this over to gossamer. Just come up with your own logging targets. In rust logging targets are for example the crate names if there is no explicit target given.

So I think  we can keep it as is.

cmd/gossamer/config.go
```diff
@@ -287,115 +288,98 @@ func createExportConfig(ctx *cli.Context) (*dot.Config, error) {
...
-	if cfg.Log.RPCLvl == "" {
-		logCfg.RPCLvl = globalCfg.LogLvl
-	} else {
-		lvl, err := log.LvlFromString(cfg.Log.RPCLvl)
-		if err != nil {
-			return err
-		}
-
-		logCfg.RPCLvl = lvl
+	globalCfg.LogLvl, err = getLogLevel(ctx, LogFlag.Name, cfg.Global.LogLvl, gssmr.DefaultLvl)
```
EclesioMeloJunior, line 339: a suggestion using a list of parameters:
```suggestion
        toDefineLvls := []struct {
		lvl       *log.Lvl
		flagName  string
		tomlValue string
		def       log.Lvl
	}{
		{
			lvl:       &globalCfg.LogLvl,
			flagName:  LogFlag.Name,
			tomlValue: cfg.Global.LogLvl,
			def:       gssmr.DefaultLvl,
		},
		...
	}

	for _, l := range toDefineLvls {
		gotLvl, err := getLogLevel(ctx, l.flagName, l.tomlValue, l.def)
		if err != nil {
			return fmt.Errorf("cannot get %s: %w", l.flagName, err)
		}
		*l.lvl = gotLvl
	}
```

EclesioMeloJunior (changes requested):

edwardmack (approved):

cmd/gossamer/config.go
qdm12: Done :+1: Nice suggestion, I never thought about doing it like that!

cmd/gossamer/flags.go
qdm12: I also renamed `LogFinalityGadgetLevelFlag` to `LogGrandpaLevelFlag`

cmd/gossamer/flags.go
qdm12: I also renamed `LogBlockProducerLevelFlag` to `LogBabeLevelFlag` for consistency

EclesioMeloJunior (approved):

PR #1959, opened 2021-11-02
fix(dot/state): fix deadlock, fixes bootstrap syncing

noot:
## Changes

- deadlock was caused as `GetBlockByHash()` calls `RLock()`, `GetBlockByHash()` calls `GetBlockBody()` which previously also called `RLock()`. this was ok when there were no calls to the write lock `Lock()` simultaneously. however the deadlock scenario occured when one goroutine was calling  `GetBlockBody()` (with 1 read-lock obtained) and another called `SetFinalisedHash()` (thus obtaining the write-lock). as a read-lock can no longer be obtained if 1. write-lock was obtained (even if it was freed) and 2. another read-lock was already obtained, this caused a deadlock
- I ended up just removing the lock in `GetBlockBody()` to prevent multiple read-locks from being obtained. if anyone has a better method let me know :D

## Tests

```
make gossamer
```

node can sync (up until it hits max memory :p) no more stalling/deadlocks

## Issues

- related to #1479 

## Primary Reviewer

- @EclesioMeloJunior


dot/sync/chain_processor.go
```diff
@@ -191,6 +186,10 @@ func (s *chainProcessor) processBlockData(bd *types.BlockData) error {
 		s.handleJustification(bd.Header, *bd.Justification)
 	}
 
+	if err := s.blockState.CompareAndSetBlockData(bd); err != nil {
+		return fmt.Errorf("failed to compare and set data: %w", err)
+	}
```
qdm12, line 191: Why moving `CompareAndSetBlockData`? There must be a reason but it's not totally mentioned in the description 🤔

dot/state/block.go
```diff
@@ -360,9 +361,6 @@ func (bs *BlockState) HasBlockBody(hash common.Hash) (bool, error) {
 
 // GetBlockBody will return Body for a given hash
 func (bs *BlockState) GetBlockBody(hash common.Hash) (*types.Body, error) {
-	bs.RLock()
-	defer bs.RUnlock()
```
qdm12, line 364: Are we sure `GetBlockBody` is not used anywhere else? As in, maybe we need to add locking on other callers using this one?

qdm12 (approved):

dot/sync/chain_processor.go
noot: yes, sorry I should have mentioned. this should definitely be happening after the block checks happen, otherwise we might be storing some data in the database and the block turns out to be invalid, and then the extra block data is still stored.

dot/state/block.go
noot: it isn't used anywhere else in the `BlockState`, so I think it's probably ok for now. but in the future it might be nice to do some lock overhaul on the `BlockState` since it could be improved imo.

qdm12: Maybe unexport that method to make this clearer, if possible?

noot: ah sorry it is used in other spots in the codebase, so gotta leave it like this for now

qdm12: So are you sure it won't data race on the block state if it's accessed elsewhere in the codebase?

EclesioMeloJunior: currently, the `GetBlockBody` just access the `unfinalisedBlocks` field that is a `*sync.Map` so by default can prevent concurrent access to the data other than that I'm pretty sure we can remove the locks

EclesioMeloJunior (approved):

dot/state/block.go
noot: yeah there's an edge case potentially with the `handleFinalisedBlock` function (just pushed a fix for it) previously that function was doing `LoadAndDelete` on the unfinalized blocks, then storing it in the db. so there was a potential race condition where one process was finalizing a block, and it was deleted from memory but not stored in the db yet, while another process calls  `GetBlockBody` and isn't able to find it in memory or in the db. however my update to `handleFinalisedBlock` should fix this (it now gets from memory, stores in the db, then deletes from memory)

qdm12 (approved):

PR #1961, opened 2021-11-02
update(dot/rpc): insertKey and hasKey takes keystore type

### LINKED ISSUE #1850: rpc: update `author_insertKey` and `core.InsertKey()` to take into account keystore type
## Task summary

- currently `author_insertKey` does not correctly use the `Type` parameter, which should correspond to a keystore type (not a key type)
- need to also update `core.InsertKey()` to insert key into correct keystore

## Specification reference

- https://github.com/w3f/PSPs/blob/master/PSPs/drafts/psp-6.md#185-author_insertkey
https://github.com/w3f/PSPs/blob/master/PSPs/drafts/psp-6.md#1811-key-types

kishansagathiya:
## Changes

- This PR changes `author_insertKey`, `core.InsertKey` and
    `core.HasKey` to accept keystore type as an argument.
    
    So far, these operations were using `account` keystore, but now they
    will use the keystore associated with the type passed in the argument.

## Tests

I think unit tests and ci check should be enough

## Issues

- Closes #1850 

## Primary Reviewer

- @noot


noot (approved):
great work!

dot/core/service.go
```diff
@@ -438,15 +438,26 @@ func (s *Service) maintainTransactionPool(block *types.Block) {
...
-func (s *Service) HasKey(pubKeyStr, keyType string) (bool, error) {
-	return keystore.HasKey(pubKeyStr, keyType, s.keys.Acco)
+// are issues decoding string
+func (s *Service) HasKey(pubKeyStr, keystoreType string) (bool, error) {
+	ks, err := s.keys.GetKeystore([]byte(keystoreType))
+	if err != nil {
+		return false, err
+	}
+
+	return keystore.HasKey(pubKeyStr, keystoreType, ks)
```
qdm12, line 460: nit maybe instead pass in the keystore as an interface to `InsertKey` and `HasKey` instead of passing its string type?

Also should we use typed strings for the key store type, together with enums (aka iota)?

dot/core/service_test.go
```diff
@@ -117,6 +117,54 @@ func TestAnnounceBlock(t *testing.T) {
...
+			description:  "Test that insertKey succeeds when keystore type is valid and appropriate ",
+			keystoreType: "acco",
+			err:          nil,
+		},
+	}
+
+	for _, c := range testCases {
+		c := c
+		t.Run(c.description, func(t *testing.T) {
+			err := s.InsertKey(kr.Alice(), c.keystoreType)
```
qdm12, line 156: Can you run this in parallel? (and also the parent test)

```suggestion
			t.Parallel()

			err := s.InsertKey(kr.Alice(), c.keystoreType)
```

If not, there is no point pinning `c := c` 😉

dot/core/service_test.go
```diff
@@ -117,6 +117,54 @@ func TestAnnounceBlock(t *testing.T) {
...
+			require.True(t, errors.Is(err, c.err))
```
qdm12, line 157: nit I think there is a `require.IsError` to simplify that

dot/rpc/modules/api_mocks.go
```diff
@@ -46,7 +46,7 @@ func NewMockBlockAPI() *modulesmocks.BlockAPI {
 // NewMockCoreAPI creates and return an rpc CoreAPI interface mock
 func NewMockCoreAPI() *modulesmocks.CoreAPI {
 	m := new(modulesmocks.CoreAPI)
-	m.On("InsertKey", mock.AnythingOfType("crypto.Keypair"))
+	m.On("InsertKey", mock.AnythingOfType("crypto.Keypair"), mock.AnythingOfType("string")).Return(nil)
```
qdm12, line 49: nit & out of scope: we should avoid using `mock.AnythingOfType` I've seen it multiple times cause too-flexible tests letting bugs slip through. It looks like we use it pretty much everywhere so that's that, but we should not for future new mocks/tests.

dot/rpc/modules/author_test.go
```diff
@@ -323,7 +323,7 @@ func TestAuthorModule_PendingExtrinsics(t *testing.T) {
 
 func TestAuthorModule_InsertKey(t *testing.T) {
 	mockCoreAPI := &apimocks.CoreAPI{}
-	mockCoreAPI.On("InsertKey", mock.Anything).Return(nil)
+	mockCoreAPI.On("InsertKey", mock.Anything, mock.Anything).Return(nil)
```
qdm12, line 326: Can't you predict what should be received here? (same question for above `mock.AnythingOfType("string")`)

qdm12 (approved):

dot/core/service.go
kishansagathiya: > Also should we use typed strings for the key store type, together with enums (aka iota)?

That is the convention. I generally try to do this when possible. I tried for this PR as well, but that become a rather big change because it impact a lot of things.
Currently, it is not that bad though. There is a `Name` type.

dot/core/service_test.go
kishansagathiya: removed `c := c`, that way I don't have to worry if these tests can run in parallel, these seem like light operations anyway.

dot/core/service_test.go
kishansagathiya: There is `require.ErrorIs`, but it won't work. For that, errors have to be in chain.

dot/core/service_test.go
qdm12: We need to run tests in parallel to find data races, it's not really for performance.

kishansagathiya: done in the latest commit

Issue #1936, opened 2021-10-28
Profile memory usage and describe plan of action to address issues


timwu20:
## Issue summary

- Profile nodes on staging, gossamer devnet, locally to identify root cause of memory issues
- After a certain block height the node becomes unresponsive
- One contributing factor could be the node.deepCopy 
- Outline a plan of action to address the memory issues

## Other information and links

- some work has already been done regarding this in the parent epic https://github.com/ChainSafe/gossamer/issues/1494

arijitAD:
https://drive.google.com/drive/folders/1eh5_0Kb4zbwe8qckCqJ2we6EE3nTkr4C?usp=sharing
Investigate memory consumption in the Trie.

qdm12:
With #2009 memory consumption (`RES`) grows until it stabilizes and oscillates between 9.2 and 9.8GB, after several hours (depends on CPU speed). I'll investigate if they are obvious ways to reduce memory usage in the rest of the code in the coming days according to the plan outlined above.

qdm12:
Problem is definitely coming from https://github.com/ChainSafe/gossamer/blob/ccf12dec1344e9ff827bca43546e33c9c0476d30/lib/trie/trie.go#L72-L82

We are appending hashes of 'old generation node' to the `deletedKeys` hash slice, but these are actually still referenced by their parent branch (unless it's the root).

One way to fix this would be to keep a reference to the parent branch (or nil for the root node) in order to clear the reference to the node being from the older generation.

I am still reviewing and testing the code in `lib/trie/trie.go` to see if this is possible and to find the best way to do so without breaking another feature.

qdm12:
After full unit testing coverage and full linear refactoring of `trie.go`, the comment above is actually wrong. Although there were some misuses of the snapshotting mechanism in certain places of the code, which were causing some minor memory leaks.

Most notably, the following issues were the root of the memory usage increasing over time:

### Issues fixed during the trie refactoring

- Minor memory leak fixes due to bad logic (regarding generation mostly)

### Issues fixed by refactoring the trie node code

- Sync.Pool memory leak due to buffers not being placed back into the buffer pool

### Issues fixed in `dot/state`

- Fix prune keys being empty, to remove unfinalized tries once a trie is finalized
- Fix `sync.Map` not allowing proper garbage collection when a key is deleted, for the map from root hash to trie
- Fix removing the previous finalized trie when finalizing a trie

qdm12:
[...]

PR #1963, opened 2021-11-02
fix(dot/network): fix memory allocations with `sizedBufferPool`

qdm12:
## Changes

- Use a slice instead of an array, since returning the array was effectively copying it

## Tests

```
go test -benchmem -run=Benchmark_sizedBufferPool github.com/ChainSafe/gossamer/dot/network
```

## Issues

- Related to #1936 and found by reviewing PR #1913 

## Primary Reviewer

- Anyone is welcome

### DIFF
--- a/dot/network/service.go
+++ b/dot/network/service.go
@@ -148,12 +148,13 @@ func NewService(cfg *Config) (*Service, error) {
 	// pre-allocate pool of buffers used to read from streams.
 	// initially allocate as many buffers as liekly necessary which is the number inbound streams we will have,
 	// which should equal average number of peers times the number of notifications protocols, which is currently 3.
-	var bufPool *sizedBufferPool
-	preAllocatedInPool := 0
-	if !cfg.noPreAllocate {
-		preAllocatedInPool = cfg.MinPeers * 3
+	preAllocateInPool := cfg.MinPeers * 3
+	poolSize := cfg.MaxPeers * 3
+	if cfg.noPreAllocate { // testing
+		preAllocateInPool = 0
+		poolSize = cfg.MinPeers * 3
 	}
-	bufPool = newSizedBufferPool(preAllocatedInPool, cfg.MinPeers*3)
+	bufPool := newSizedBufferPool(preAllocateInPool, poolSize)
 
 	network := &Service{
 		ctx:                    ctx,
--- a/dot/network/service.go
+++ b/dot/network/service.go
@@ -149,13 +149,11 @@ func NewService(cfg *Config) (*Service, error) {
 	// initially allocate as many buffers as liekly necessary which is the number inbound streams we will have,
 	// which should equal average number of peers times the number of notifications protocols, which is currently 3.
 	var bufPool *sizedBufferPool
-	if cfg.noPreAllocate {
-		bufPool = &sizedBufferPool{
-			c: make(chan *[maxMessageSize]byte, cfg.MinPeers*3),
-		}
-	} else {
-		bufPool = newSizedBufferPool(cfg.MinPeers*3, cfg.MaxPeers*3)
+	preAllocatedInPool := 0
+	if !cfg.noPreAllocate {
+		preAllocatedInPool = cfg.MinPeers * 3
 	}
+	bufPool = newSizedBufferPool(preAllocatedInPool, cfg.MinPeers*3)
 
 	network := &Service{
 		ctx:                    ctx,
@@ -563,7 +561,7 @@ func (s *Service) readStream(stream libp2pnetwork.Stream, decoder messageDecoder
 
 	peer := stream.Conn().RemotePeer()
 	msgBytes := s.bufPool.get()
-	defer s.bufPool.put(&msgBytes)
+	defer s.bufPool.put(msgBytes)
 
 	for {
 		tot, err := readStream(stream, msgBytes[:])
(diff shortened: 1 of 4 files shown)


danforbes:
Is there any way to quantify the impact/improvement this has the performance of the code?

qdm12:
> Is there any way to quantify the impact/improvement this has on the performance of the code?

Not really, although it is now significantly better, see [benchmark results](https://github.com/ChainSafe/gossamer/pull/1963/files#diff-1c450d8732ec5666cc3fa5ae5d2ef3b2b15b3957c9de12d2037bd50d78c45bccR18-R21). Since that piece was used for all dot networking, it could either had a very bad impact or it would had been barely noticeable. Only way to find out is to run it 😄 (I mean it, and check memory usage over time, profiling won't help really here I think)

danforbes:
Benchmark results are super quantifiable...thanks, @qdm12 🙏🏻

qdm12:
@danforbes but let's not get too excited. Go's runtime is pretty efficient even with bad code, so there might not be any concrete difference sadly. Anyway, one can only hope 🌟 !

noot:
@qdm12 how does the CPU usage get affected by this (if any)? as the main point of the buffer pool was to prevent unnecessary malloc's which used a lot of CPU

qdm12:
@noot the previous implementation makes sense but was bugged. It would copy the array on every call to `get()` so effectively it was useless (the same as declaring a new array on every call). See the benchmark results commented below the benchmark.

I benchmarked using an array pointer and a slice (which has a pointer to the underlying data), and both gave similar results, so I decided to go with the slice for simplicity.

Side question, did we notice a performance bottleneck in this part? We should avoid such performance improvements without identifying a bottleneck and making sure there is a visible difference of performance (don't get me wrong, I love optimising 😄). That pool being bugged, it must had been slower than the original code, and we still used it; my point is we should be more careful about such optimisation changes in the future 👍

noot:
@qdm12 sorry, did you bench the CPU or just memory? yes, this was created because previously we were allocating a new buffer every time we received a network message, which was causing gossamer to use 100% CPU on every core basically immediately after starting. the CPU usage improved dramatically after adding this

qdm12:
> did you bench the CPU or just memory

It does both.

> I guess more ops in the same amount of time

Yes. You can see before it was using a lot more time (and memory) per operation (op).

> the CPU usage improved dramatically after adding this ... which was causing gossamer to use 100% CPU on every core basically immediately after starting

That doesn't make sense, since the array was copied to another memory location on every `get` request. It might had been something else resolving the 100% cpu performance issue. I also doubt the GC (or malloc) would use 100% of even a single CPU core, even under high load. Also the network is magnitudes slower than memory bandwidth.
For science's sake 🥼, and once we have something easy to deploy changes, I would be curious to try reverting it and check CPU usage 😄

noot:
@qdm12 I'm just going based off the profiling lol, it showed malloc using up all the CPU it could previously. I can try to find an old commit for you to try to profile if you'd like ;)

noot (changes requested):
[redacted] nevermind, let me test this more

dot/network/service.go
```diff
@@ -149,13 +149,11 @@ func NewService(cfg *Config) (*Service, error) {
...
-		bufPool = &sizedBufferPool{
-			c: make(chan *[maxMessageSize]byte, cfg.MinPeers*3),
-		}
-	} else {
-		bufPool = newSizedBufferPool(cfg.MinPeers*3, cfg.MaxPeers*3)
+	preAllocatedInPool := 0
+	if !cfg.noPreAllocate {
+		preAllocatedInPool = cfg.MinPeers * 3
 	}
+	bufPool = newSizedBufferPool(preAllocatedInPool, cfg.MinPeers*3)
```
noot, line 156: the maximum should be `cfg.MaxPeers*3`

qdm12: Ohhh thank you, blind mee!! I'll fix that.

qdm12: Wait that's confusing too, why would the channel capacity `make(chan *[maxMessageSize]byte, cfg.MinPeers*3),` above not be `cfg.MaxPeers*3` as well then? Pre-allocation or not, I don't see how this should influence the channel capacity? 🤔

noot: you mean in the `!cfg.noPreAllocate` case? that's only used for testing, so there isn't really a point in pre-allocating `cfg.MaxPeers*3` buffers

qdm12: No, I mean why is the maximum channel capacity (aka `max`, different from the pre-allocated buffers in the channel queue) different.

When noPreAllocated (production) is true, the capacity is `cfg.MinPeers*3`. When it's false (for testing), it's `cfg.MaxPeers*3` 🤔  That looks strange right? Didn't we want to pre-allocate `cfg.MinPeers*3` and have a max of `cfg.MinPeers*3` for testing as well?

noot: `noPreAllocated` is always false in production, as we want to pre-allocate, it's only set to true in the tests. in testing, it's fine to have a maximum channel capacity of `cfg.MinPeers*3` as we won't really be using many buffers (we could probably set it to 3 or something even). when it's false, in production, we want to be able to store `cfg.MaxPeers*3` buffers potentially as that is the maximum amount of buffers we woudl ever need (# of peers * 3 streams per peer)

noot: @qdm12 can you update this when you get a chance? everything else looks good :)

qdm12: Done :+1: Yeah I had it all wrong, it should be fixed now.

arijitAD (approved):

noot (approved):

PR #1964, opened 2021-11-02
chore(ci): remove dependency on `scripts/ci.go`

qdm12:
## Changes

- Remove unused `scripts/ci.sh`
- Dockerfile: build Go binary without `scripts/ci.go`
- Build Go binary with `-trimpath`
- Makefile:
  - remove outdated `ci.go test` from Makefile
  - build binary without `scripts/ci.go`
  - build binary without changing directory
  - `install` binary by executing `build` target and just `mv` the built binary

## Tests

Whole CI passing

## Issues

Related to a CI issue Omar had with `go install` vs `go build`.

## Primary Reviewer

- @omar391


Makefile
```diff
@@ -80,11 +79,11 @@ deps:
 ## build: Builds application binary and stores it in `./bin/gossamer`
 build:
 	@echo "  >  \033[32mBuilding binary...\033[0m "
-	GOBIN=$(PWD)/bin go run scripts/ci.go install
+	go build -trimpath -o ./bin/gossamer ./cmd/gossamer
```
omar391, line 82: @qdm12, what's the effect of `-trimpath` in our use case here?

BTW we can or maybe should also add `-race` argument to detect possible race conditions in the codebase.

qdm12: Yes we will but in the staging Dockerfile only. Maybe we can add another target to the Makefile. We can do this later in another PR.

Note running with `-race` is much slower, so we should only run it on our environments to find data races.

qdm12: Oh and `-trimpath` removes the build path from the binary for stack traces. So you won't see your local path `/home/omar/gossamer/bin/gossamer/file.go:43` for example, it will be trimmed down to just `gossamer/file.go:43`

omar391: Cool!

omar391 (approved):
lgtm!

EclesioMeloJunior (approved):

PR #1977, opened 2021-11-03
maintenance: improve `*chainSync.determineSyncPeers`, add max worker retries

noot:
## Changes

- update `*chainSync` to not  retry a worker with the same peer, instead error if it runs out of peers (ie it tried the worker with every peer it has)
- also update `determineSyncPeers` to take into account what block it's requesting and what block peers have reported they have when determining potential peers to sync with, eg. if a peer's best head is 100, and we want block 101, don't sync with them
- also create `maxWorkerRetries`, if a worker has been retried this many times then just cancel it to prevent infinite looping of workers. `maxWorkerRetries` is currently set to max peer count, so case 1 should be hit before `maxWorkerRetries` is hit

## Tests

```
go test ./dot/sync
```

## Issues

- related to #1479 

## Primary Reviewer

- @jimjbrettj


dot/sync/chain_sync.go
```diff
@@ -691,17 +712,43 @@ func handleReadyBlock(bd *types.BlockData, pendingBlocks DisjointBlockSet, ready
...
 		if _, has := cs.ignorePeers[p]; has {
 			continue
 		}
 
+		if _, has := peersTried[p]; has {
+			continue
+		}
+
+		// if peer definitely doesn't have any blocks we want in the request,
+		// don't request from them
```
EclesioMeloJunior, line 747: should this comment be after the last if statement? Also, the comment can explain the condition to remove peer that does not have the sought number

EclesioMeloJunior (approved):
lgtm! just a small comment

dot/sync/chain_sync.go
noot: i think this should be here, as it's above the case that checks the highest block the peer has. basically if a peer reports some highest number x, but we want to request  x+1 for example, we shouldn't request from that peer as they won't have it. I can add more info to it

dot/services.go
```diff
@@ -419,6 +419,7 @@ func newSyncService(cfg *Config, st *state.Service, fg sync.FinalityGadget, veri
 		BabeVerifier:       verifier,
 		BlockImportHandler: cs,
 		MinPeers:           cfg.Network.MinPeers,
+		MaxPeers:           cfg.Network.MaxPeers,
```
jimjbrettj, line 422: Is max peers to prevent someone spamming peer connection requests, or is there more reasons for this upper bound?

noot: yeah it's kinda dangerous if we have unlimited as we can get bogged some if someone spams us. we also don't need *that* many peers necessarily for a healthy network :p

dot/sync/chain_sync.go
```diff
@@ -392,13 +404,16 @@ func (cs *chainSync) sync() {
...
 			// temporarily add them to the ignore list.
 			switch {
 			case errors.Is(res.err.err, context.Canceled):
 				return
+			case errors.Is(res.err.err, errNoPeers):
+				logger.Debug("not able to sync with any peer!", "worker ID", res.id)
+				continue
 			case errors.Is(res.err.err, context.DeadlineExceeded):
 				cs.ignorePeer(res.err.who)
 			case strings.Contains(res.err.err.Error(), "dial backoff"):
```
jimjbrettj, line 419: Nit: Can we remove the empty default case here?

dot/sync/chain_sync.go
```diff
@@ -420,6 +435,24 @@ func (cs *chainSync) sync() {
 				continue
 			}
 
+			worker.retryCount = res.retryCount + 1
+			if worker.retryCount > cs.maxWorkerRetries {
```
jimjbrettj, line 439: I'm a little confused on what happens when retryCount has passed the maxWorkerRetries, do we not want to ignore the peer in this case? I'm probably missing some context but im not quite sure what this is doing

jimjbrettj: I guess what Im asking is the log says the peer is being discarded, so how/where is that happening?

dot/sync/chain_sync.go
```diff
@@ -691,17 +712,43 @@ func handleReadyBlock(bd *types.BlockData, pendingBlocks DisjointBlockSet, ready
...
+	var start uint64
+	if req.StartingBlock.IsUint64() {
+		start = req.StartingBlock.Uint64()
+	}
+
 	cs.RLock()
 	defer cs.RUnlock()
 
+	// if we're currently ignoring all our peers, clear out the list.
+	if len(cs.peerState) == len(cs.ignorePeers) {
```
jimjbrettj, line 725: Is this so we don't find the node in a state where its stuck not listening to any peers?

jimjbrettj (approved):
Just minor comments for my own understanding or NIT, but looks good to me!

dot/sync/chain_sync.go
noot: yeah probably. I kinda like it for explicit-ness but can remove :p

dot/sync/chain_sync.go
noot: the worker isn't tied to a specific peer, in this case what's happening is that the worker has tried to sync with various peers `maxWorkerRetries` times, and we want to just cancel it since it's been retried too much. I saw a lot of infinite looping without this

dot/sync/chain_sync.go
noot: yeah, not "not listening" exactly but basically ignoring every peer for syncing purposes ;p

PR #1980, opened 2021-11-03
testing (dot) create dot unit tests

### LINKED ISSUE #1952: Create dot unit tests
## Issue summary

- Write unit tests for dot package
- Use mocks for interfaces
- 70% coverage minimum

- https://github.com/ChainSafe/gossamer/pull/1666

edwardmack:
## Changes

- Create unit tests `dot` package: `build_spec_test.go`, `config_test.go`, `import_test.go`, `node_test.go`, `service_test.go` and `utils_test.go`
- 

## Tests

```
go test ./dot/ -v

```

## Issues

- closes #1952 

## Primary Reviewer

- @timwu20, @qdm12


qdm12:
I see a lot of integration tests deltas, is this expected? Did you maybe branch off your other integration tests PR?

> TestBuildSpec_ToJSON and TestBuildSpec_ToJSONRaw do not have test for errors with json.Marshal, I don't know how to create condition that will simulate an error with this.

JSON Marshal will fail if you try to marshal e.g. a function/channel. Maybe add some dummy function field to your `field` struct to trigger an error?

> TestWriteGenesisSpecFile doesn't test for errors from os.MkdirAll, I don't know how to create condition to simulate error.

Have a file at the same path 😉 Example

```
/tmp # touch test
/tmp # mkdir -p test
mkdir: can't create directory 'test': File exists
```

> since these functions build the struct I don't how to compare pointer addresses

Why would we care about the pointer address?

> Cleanup after errors

Shouldn't it cleanup even when there is no error? 🤔

### Other general comments

- For testing errors, I would suggest to use (since we already use testify around in the repo):

    ```go
    if testCase.err != nil {
    	assert.EqualError(t, err, testCase.err.Error())
    } else {
    	assert.NoError(t, err)
    }
    ```

- You can use `assert.Equal(t, expected, actual)` instead of `reflect.DeepEqual` + `t.Errorf(...)` since it does the same thing, it would accelerate test writing/shorten your test code 😉

edwardmack:
> I see a lot of integration tests deltas, is this expected? Did you maybe branch off your other integration tests PR?
> 
I did branch off my other integration tests PR, is this a bad idea?  Whats the best way to work on stuff that depends on other stuff you've done.  I've gotten into git branch trouble before, so I'm open to suggestions of best way to organize this.

> JSON Marshal will fail if you try to marshal e.g. a function/channel. Maybe add some dummy function field to your `field` struct to trigger an error?
> 
I'm no longer sure I need to test for this since the struct passed to this function insures it will not receive a function/channel.
 
> > TestWriteGenesisSpecFile doesn't test for errors from os.MkdirAll, I don't know how to create condition to simulate error.
> 
> Have a file at the same path wink Example
> 
> ```
> /tmp # touch test
> /tmp # mkdir -p test
> mkdir: can't create directory 'test': File exists
> ```

I wasn't able to create condition where `os.MkdirAll` failed.

> > since these functions build the struct I don't how to compare pointer addresses
> 
> Why would we care about the pointer address?
> 
Good point, I've updated tests to compare returned values.

> > Cleanup after errors
> 
> Shouldn't it cleanup even when there is no error? thinking
> 

Yes, good point, I've updated this.

> ### Other general comments
> * For testing errors, I would suggest to use (since we already use testify around in the repo):
>   ```go
>   if testCase.err != nil {
>   	assert.EqualError(t, err, testCase.err.Error())
>   } else {
>   	assert.NoError(t, err)
>   }
>   ```
> * You can use `assert.Equal(t, expected, actual)` instead of `reflect.DeepEqual` + `t.Errorf(...)` since it does the same thing, it would accelerate test writing/shorten your test code wink

Updated.

qdm12:
> I did branch off my other integration tests PR, is this a bad idea? Whats the best way to work on stuff that depends on other stuff you've done. I've gotten into git branch trouble before, so I'm open to suggestions of best way to organize this.

I'd suggest you point this PR to that other branch you branched out from, merge the second one (this one) first and then the other branch (the first one) in development.

And yes don't worry too much about impossible to reproduce errors 😄 I'll mess around with `os.MkdirAll` though I'm curious now!

timwu20:
Can you change the base branch of this PR to the other branch you moved the integration tests too? It'll make the diff easier to read.

dot/build_spec_test.go
```diff
@@ -16,136 +16,193 @@
...
+	}
+	tests := []struct {
+		name   string
+		fields fields
+		want   []byte
+		err    error
+	}{
+		{
+			name:   "normal conditions",
+			fields: fields{genesis: &genesis.Genesis{Name: "test"}},
```
timwu20, line 39: can you add more tests to test the other attributes of `genesis.Genesis`

timwu20:
I wouldn't worry about trying to test coverage up to 100%.  It's not really necessary to test `json.Marshal` and file read errors if you're already above 70% coverage.

dot/node.go
```diff
@@ -180,7 +180,7 @@ func NodeInitialized(basepath string) bool {
 }
 
 // LoadGlobalNodeName returns the stored global node name from database
-func LoadGlobalNodeName(basepath string) (nodename string, err error) {
+func LoadGlobalNodeName(basepath string) (string, error) {
```
qdm12, line 183: Please let's use named returned values, it's really useful for readers/documentation.

dot/services_test.go
```diff
@@ -100,54 +214,124 @@ func Test_createCoreService(t *testing.T) {
...
-				return
+			if tt.err != nil {
+				assert.EqualError(t, err, tt.err.Error())
+			} else {
+				assert.NoError(t, err)
 			}
-			if !reflect.DeepEqual(got, tt.want) {
-				t.Errorf("createCoreService() got = %v, want %v", got, tt.want)
+			if tt.want != nil {
+				assert.NotNil(t, got)
```
qdm12, line 249: Maybe assert it's nil if we expect it to be nil as well :thinking: ?

dot/services_test.go
```diff
@@ -100,54 +214,124 @@ func Test_createCoreService(t *testing.T) {
...
 			}
 		})
 	}
 }
 
 func Test_createDigestHandler(t *testing.T) {
+	cfg := NewTestConfig(t)
+	require.NotNil(t, cfg)
```
qdm12, line 257: nit I don't think this NotNil is really required since (afaik) we are asserting something from the test (`NewTestConfig`). Same for below checks as well. Although if these helper functions do use some production code, maybe do the assertion inside them since you are already passing `t` to them.

dot/services_test.go
```diff
@@ -247,22 +527,49 @@ func Test_createRuntime(t *testing.T) {
...
+			err: errors.New("failed to create runtime executor: code is empty"),
+		},
+		{
+			name: "bad code test",
+			args: args{
+				cfg:  cfg,
+				st:   stateSrvc,
+				code: []byte(`fake code`),
+			},
+			err: errors.New("failed to create runtime executor: Failed to instantiate the module:\n    compile error: Validation error \"Bad magic number\""),
```
qdm12, line 550: Is it us or the runtime returning an error with `\n`?

lib/babe/verify.go
```diff
@@ -58,12 +60,26 @@ type VerificationManager struct {
...
+	switch blockState.(type) {
+	case *state.BlockState:
+		if blockState == (*state.BlockState)(nil) {
+			return nil, errNilBlockState
+		}
+	default:
+		if reflect.ValueOf(blockState).IsNil() {
+			return nil, errNilBlockState
+		}
 	}
```
qdm12, line 72: supernit to reduce code duplication if we add more cases :wink: 

```suggestion
	var isNil bool
	switch blockState.(type) {
	case *state.BlockState:
		isNil = blockState == (*state.BlockState)(nil)
	default:
		isNil = reflect.ValueOf(blockState).IsNil()
	}
	if isNil {
		return nil, errNilBlockState
	}
```

dot/build_spec_test.go
edwardmack: added.

dot/node.go
edwardmack: updated.

dot/services_test.go
edwardmack: Good idea, added.

dot/services_test.go
edwardmack: Ok, removed.

lib/babe/verify.go
edwardmack: updated.

dot/node.go
```diff
@@ -146,29 +146,27 @@ func NodeInitialized(basepath string) bool {
...
 	db, err := utils.SetupDatabase(basepath, false)
 	if err != nil {
 		return "", err
 	}
 
 	defer func() {
-		err = db.Close()
-		if err != nil {
-			logger.Errorf("failed to close database: %s", err)
+		e := db.Close()
```
qdm12, line 157: nit, maybe instead of logging the close error:

1. Rename `e` to `closeErr` (sorry I'm more of an entire words person that single letter C style names 😄) 
2. If `err` (the returned error) is nil, set it to `closeErr`

dot/node.go
```diff
@@ -146,29 +146,27 @@ func NodeInitialized(basepath string) bool {
...
 	basestate := state.NewBaseState(db)
-	nodename, err := basestate.LoadNodeGlobalName()
+	nodename, err = basestate.LoadNodeGlobalName()
 	if err != nil {
 		logger.Warnf("failed to load global node name from base path %s: %s", basepath, err)
-		return "", err
 	}
-
-	return nodename, err
+	return
```
qdm12, line 169: No naked returns, it's really harder to read and leads to confusion

https://dave.cheney.net/practical-go/presentations/gophercon-israel.html#_avoid_naked_returns

To be fair, I have no clue why this is even part of the language...

dot/node.go
```diff
@@ -326,6 +326,7 @@ func newNodeC(cfg *Config, nn newNodeIface) (*Node, error) {
 
 	return node, nil
 }
+//go:generate mockgen -source=node.go -destination=mock_node.go -package=dot
```
qdm12, line 330: Change it to

```suggestion
//go:generate mockgen -destination=mock_node_test.go -package $GOPACKAGE . newNodeIface
```

So it's unexported to other packages (as a test file) and also using `source` isn't the best if it can be avoided. There are some corner cases where it's necessary (I don't remember which though haha), but it's better to just precise the interface name.

You should

dot/node_test.go
```diff
@@ -200,17 +201,75 @@ func TestNewNodeC(t *testing.T) {
...
+				},
+			},
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			got, err := NewNodeC(tt.args.cfg)
+			if tt.err != nil {
+				assert.EqualError(t, err, tt.err.Error())
+				utils.RemoveTestDir(t)
```
qdm12, line 221: nit maybe use `t.TestDir` instead?

dot/node_test.go
```diff
@@ -200,17 +201,75 @@ func TestNewNodeC(t *testing.T) {
...
+			} else {
+				assert.NoError(t, err)
+			}
+
+			if tt.want != nil {
+				assert.Equal(t, tt.want.Name, got.Name)
+			}
```
qdm12, line 228: Why not assert the entire Node instead of just its name?

dot/node_test.go
```diff
@@ -200,17 +201,75 @@ func TestNewNodeC(t *testing.T) {
...
+
+	m := NewMocknewNodeIface(ctrl)
+	m.EXPECT().nodeInitialised(gomock.Any()).Return(true)
+	m.EXPECT().initKeystore(gomock.Any())
+	m.EXPECT().createStateService(gomock.Any())
+	m.EXPECT().createRuntimeStorage(gomock.Any())
+	m.EXPECT().loadRuntime(gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any())
+	m.EXPECT().createBlockVerifier(gomock.Any())
+	m.EXPECT().createDigestHandler(gomock.Any())
+	m.EXPECT().createCoreService(gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any())
```
qdm12: Please try to not use `gomock.Any()` since it covers code but test depth is quite low and some bugs can sneak in.

Ideally declare both the gomock controller and mocks inside the subtests with parameters from each test case to define the mock behavior.

dot/node.go
edwardmack: When I tried running this I get:
```
emack@system76-pc:~/projects/ChainSafe/gossamer/dot$ go generate
# command-line-arguments
./prog.go:27:38: cannot refer to unexported name dot.newNodeIface

```
so I needed to use the `-source` flag to solve issue.

dot/node_test.go
edwardmack: I'm not sure I understand what you mean here?

dot/node_test.go
edwardmack: Updated.

dot/node.go
qdm12: Totally! That was the case where you needed it 😄  Although you can use `$GOPACKAGE` instead of `dot` (if you want to)

dot/node_test.go
qdm12: I'm not sure where the directory is created but you can use `t.TestDir` to create a temp directory that gets removed at the end of the test. It might be out of scope here though.

dot/node_test.go
edwardmack: updated to remove use of `gomock.Any()`.

dot/node.go
edwardmack: updated.

dot/node.go
edwardmack: updated.

dot/build_spec_test.go
```diff
@@ -4,136 +4,345 @@
...
+	}
+	//<<<<<<< HEAD
+	//=======
+	//
+	//	bs.genesis.ChainType = expectedChainType
+	//	bs.genesis.Properties = expectedProperties
+	//
+	//	require.NoError(t, err)
+	//
+	//	// confirm human-readable fields
```
noot, line 123: remove commented out block

dot/build_spec_test.go
```diff
@@ -4,136 +4,345 @@
...
+	//		require.NoError(t, err)
+	//		defer file.Close()
+	//
+	//		genesisBytes, err := io.ReadAll(file)
+	//		require.NoError(t, err)
+	//
+	//		gen := new(genesis.Genesis)
+	//		err = json.Unmarshal(genesisBytes, gen)
+	//		require.NoError(t, err)
+	//>>>>>>> development
```
noot, line 228: same here

dot/node.go
```diff
@@ -333,19 +386,52 @@ func NewNode(cfg *Config, ks *keystore.GlobalKeystore) (*Node, error) {
...
+		peerID,
 		strconv.FormatInt(time.Now().UnixNano(), 10),
 		sysSrvc.SystemVersion()))
 	if err != nil {
 		logger.Debugf("problem sending system.connected telemetry message: %s", err)
 	}
-	return node, nil
+}
+
+//InitKeystore to initialize node keystore
```
noot, line 407:
```suggestion
// InitKeystore initialised the node's keystore
```

dot/node.go
```diff
@@ -41,9 +48,87 @@ type Node struct {
 	started  chan struct{}
 }
 
+//go:generate mockgen -source=node.go -destination=mock_node_test.go -package=$GOPACKAGE
+
+type newNodeIface interface {
```
noot, line 53: nice! I like this interface a lot, maybe name it `nodeBuilder` or `nodeCreator` or something like that?

lib/babe/verify.go
```diff
@@ -47,11 +49,25 @@ type VerificationManager struct {
...
 
-	if epochState == nil {
+	var isNilEpoch bool
+	switch epochState.(type) {
+	case *state.EpochState:
+		isNilEpoch = epochState == (*state.EpochState)(nil)
+	default:
+		isNilEpoch = reflect.ValueOf(epochState).IsNil()
+	}
+	if isNilEpoch {
```
noot, line 70: do we ned this `isNilBlock = blockState == (*state.BlockState)(nil)` or does `reflect.ValueOf(blockState).IsNil()` evaluate to the same thing? I'd rather not have the `(*state.BlockState)` type assertion if possible

lib/blocktree/blocktree.go
```diff
@@ -403,7 +403,7 @@ func (bt *BlockTree) StoreRuntime(hash common.Hash, in runtime.Instance) {
 // GetBlockRuntime returns block runtime for corresponding block hash.
 func (bt *BlockTree) GetBlockRuntime(hash common.Hash) (runtime.Instance, error) {
 	ins, ok := bt.runtime.Load(hash)
-	if !ok {
+	if !ok || ins == nil {
```
noot, line 406: is `ins == nil` a case that happens? that seems concerning to me lol

noot (commented):
very nice work! just a few comments, but the tests look good to me :D

edwardmack:
closed, replaced by PR #2112

Issue #1969, opened 2021-11-03
Pprof http server


qdm12:
Setup a pprof HTTP server listening on `127.0.0.1:6062` for easier access

noot:
@qdm12 is this done?

qdm12:
Yes thanks for notifying me

PR #1991, opened 2021-11-05
feat(pprof): Pprof HTTP server service

qdm12:
## Changes

- [x] Add a pprof HTTP server service
  - [x] `internal/httpserver` package
  - [x] `internal/pprof` package
- [x] Register this service in the dot node
- [x] Configurable listening address
- [x] Configurable pprof server on/off
- [ ] Add config tests
- [x] Finish documentation
- [x] ~Add copyright notices~ will do with #1997 
- Notes:
  - this does not affect performance, except when triggered by an HTTP client
  - this listens on `localhost:6060` (localhost so it works with IPv6 local too)

## Tests

```
go test -race github.com/ChainSafe/gossamer/internal/pprof
go test -race github.com/ChainSafe/gossamer/internal/httpserver
go test github.com/ChainSafe/gossamer/dot
```

## Issues

- #1969 

## Primary Reviewer

-


dot/services.go
```diff
@@ -429,3 +431,8 @@ func newSyncService(cfg *Config, st *state.Service, fg sync.FinalityGadget, veri
 func createDigestHandler(st *state.Service) (*digest.Handler, error) {
 	return digest.NewHandler(st.Block, st.Epoch, st.Grandpa)
 }
+
+func createPprofService() (service *pprof.Service) {
+	pprofLogger := log.New("pkg", "pprof")
+	return pprof.NewService("localhost:6060", pprofLogger)
```
noot, line 437: i saw you have a todo for making the port configurable, do you want to do that in this PR?

internal/httpserver/address.go
```diff
@@ -0,0 +1,7 @@
+package httpserver
+
+// GetAddress obtains the address the HTTP server is listening on.
+func (s *Server) GetAddress() (address string) {
+	<-s.addressSet
+	return s.address
+}
```
noot, line 7: is this used anywhere?

internal/httpserver/matchers_test.go
```diff
@@ -0,0 +1,31 @@
...
+
+import (
+	"regexp"
+
+	gomock "github.com/golang/mock/gomock"
+)
+
+var _ gomock.Matcher = (*regexMatcher)(nil)
+
+type regexMatcher struct {
```
noot, line 11: is this used for comparing logs in tests? or is it used anywhere else?

qdm12: Oh yeah sorry I forgot to mention, I started using golang/gomock for mocks locally in these 2 packages, I hope you don't mind!

And that's a custom matcher used only for this package's tests, since the port number picked (at random availability) is unpredictable.

internal/httpserver/address.go
qdm12: It might be used in one of the pprof package test to find what port the server is listening on. I'll double check, please leave this unresolved.

dot/services.go
qdm12: Yes I'll 😉 My bad, I wanted to make a draft PR but my finger decided not to somehow. Thanks for the ⚡ review though!

EclesioMeloJunior (approved):
Would be nice to use `go:generate` to generate the `regex_matchers_test.go` files?

qdm12:
> Would be nice to use `go:generate` to generate the `regex_matchers_test.go` files?

These are files manually written :smile: It's just a custom matcher accepting regexes to assert the mocked function arguments. We could put it in test focused internal package later if we feel we need it for other tests. But I also like the "a bit of code duplication is better than dependency headaches" :smile:

internal/httpserver/address.go
qdm12: Yes it's used in a test in the `internal/pprof` package ([here](https://github.com/ChainSafe/gossamer/blob/8985412277b6c700b847748768f2a7ce76c872a9/internal/pprof/server_test.go#L42)), so it has to stay. It might be useful for future http servers to run tests on a randomly assigned port (aka run in parallel).

noot (approved):
love the docs!! I have one question, if someone sets `--pprofserver`, but the block rate and mutex rate are set to 0, will the server run or no?

qdm12:
@noot Yes the server will run, it's just the /debug/pprof/block and /debug/pprof/mutex routes won't be working. Note the other way around, if you set e.g. `--pprofmutexrate 1` **without** `--pprofserver`, it won't be set since the server is disabled. Let me know if I should change that/document that.

qdm12:
I'm thinking of either:

- removing the existing profiling functions (to a file) in favor of the http server; or
- moving the existing profiling functions in the pprof service with additional settings to choose if the program should write to a file and/or run an http server.

I am leaning towards the first one, but I don't want to hurt feelings or habits either :smile_cat: What do you think?

Issue #1026, opened 2020-07-24
fix skipped tests


noot:
## Expected Behavior

- tests should pass on CI if they pass locally
- add back skipped tests

## Current Behavior

some tests fail randomly on CI: (add more as needed)

- [ ] network `TestGossip`, `TestStreamManager_KeepStream`, `TestPersistentPeers`, `TestRemovePeer`, `TestSetReservedPeer`
- [ ] messages `TestService_HandleTransactionMessage`
- [ ] author `TestAuthorModule_SubmitExtrinsic`
- [ ] author `TestAuthorModule_SubmitExtrinsic_invalid`
- [ ] author `TestAuthorModule_SubmitExtrinsic_InQueue`
- [ ] rpc `TestStateModule_GetMetadata`
- [ ] build `TestBuildBlock_failing`
- [ ] extrinsics `TestAuthoritiesChangeExt_Encode`
- [ ] exports `TestInstance_ExecuteBlock_PolkadotRuntime_PolkadotBlock1`, `TestInstance_Version_NodeRuntime`
- [ ] imports `Test_ext_storage_clear_prefix_version_1`
- [ ] wasmer `TestInstance_ExecuteBlock_GossamerRuntime`, `TestInstance_ApplyExtrinsic_GossamerRuntime`, `Test_ext_offline_index_set_version_1`
- [ ] grandpa `TestGrandpa_NonAuthority`, `TestStress_Grandpa_ThreeAuthorities`, `TestStress_Grandpa_SixAuthorities`, `TestGrandpa_DifferentChains`
- [ ] stress tests `TestSync_SingleSyncingNode`, `TestSync_ManyProducers`, `TestSync_Restart`, `TestSync_MultipleEpoch`

## Possible Solution

- this may be due to some underlying randomness
- for integration tests, this is probably due to timing issues

## Checklist

- [x] I have read [CODE_OF_CONDUCT](https://github.com/ChainSafe/gossamer/blob/development/.github/CODE_OF_CONDUCT.md) and [CONTRIBUTING](https://github.com/ChainSafe/gossamer/blob/development/.github/CONTRIBUTING.md) 
- [x] I have provided as much information as possible and necessary
- [ ] I am planning to submit a pull request to fix this issue myself

kishansagathiya:
Some more context on TestGossip here https://github.com/ChainSafe/gossamer/issues/1026

kishansagathiya:
[...]

PR #1993, opened 2021-11-05
fix(lib/blocktree): fix potential nil pointer dereference in `HighestCommonAncestor`, core `handleBlocksAsync`

noot:
## Changes

- fix potential nil pointer dereference in blocktree `HighestCommonAncestor`
- same in core `handleBlocksAsync`
- also increase BABE `firstBlockTimeout` to 5 mins, 1 min wasn't long enough sometimes

## Tests

```
go test ./lib/blocktree
go test ./dot/core
```

## Issues

- related to #592 

## Primary Reviewer

- @timwu20


dot/core/service.go
```diff
@@ -361,13 +367,21 @@ func (s *Service) handleChainReorg(prev, curr common.Hash) error {
...
+	}
+
 	// for each block in the previous chain, re-add its extrinsics back into the pool
 	for _, hash := range subchain {
 		body, err := s.blockState.GetBlockBody(hash)
 		if err != nil {
 			continue
 		}
 
+		if body == nil {
```
qdm12, line 381: nit Maybe compact it with the above if with `|| body == nil`

lib/blocktree/blocktree.go
```diff
@@ -338,12 +338,20 @@ func (bt *BlockTree) HighestCommonAncestor(a, b Hash) (Hash, error) {
...
 	if bn == nil {
 		return common.Hash{}, ErrNodeNotFound
 	}
 
-	return an.highestCommonAncestor(bn).hash, nil
+	ancestor := an.highestCommonAncestor(bn)
+	if ancestor == nil {
+		// this case shouldn't happen - any two nodes in the blocktree must
+		// have a common ancestor, the lowest of which is the root node
+		return common.Hash{}, ErrNoCommonAncestor
```
qdm12, line 351: nit Can you maybe add some context with the error e.g.  which node?

qdm12 (approved):

EclesioMeloJunior (approved):

dot/network/connmgr_test.go
```diff
@@ -132,6 +129,10 @@ func TestProtectUnprotectPeer(t *testing.T) {
 }
 
 func TestPersistentPeers(t *testing.T) {
+	if testing.Short() {
+		t.Skip() // this sometimes fails on CI
```
qdm12, line 133: Maybe add a TODO with an issue number to fix the skipped tests? :thinking:

lib/blocktree/node.go
```diff
@@ -55,17 +55,24 @@ func (n *node) createTree(tree gotree.Tree) {
...
+	if n == nil {
+		return nil
+	}
+
 	if n.hash == h {
 		return n
-	} else if len(n.children) == 0 {
+	}
+
+	if len(n.children) == 0 {
```
qdm12, line 66: nit I guess you could remove that if clause, since the for loop below will just not execute below, and return `nil`.

lib/blocktree/node.go
```diff
@@ -55,17 +55,24 @@ func (n *node) createTree(tree gotree.Tree) {
 
 // getNode recursively searches for a node with a given hash
 func (n *node) getNode(h common.Hash) *node {
+	if n == nil {
+		return nil
+	}
```
qdm12, line 60: nit Does this happen? Should we check at the caller level if node is nil or not instead of calling `getNode` on it perhaps?

dot/network/connmgr_test.go
noot: yep I'll add it to here: #1026

lib/blocktree/node.go
noot: it shouldn't happen, but I'd rather have this check here just to be safe... do you know of a good method for dealing with things like this, where it *shouldn't* happen but if it does it's really bad? kinda what I thought `panic`s were for, but seems it's not good to use those

lib/blocktree/node.go
noot: done

lib/blocktree/node.go
qdm12: In my opinion, panics are when you can't break a function signature. For example if you have an exported, public `Divide(a, b int) int` and later realise you cannot divide with `b = 0`, use a `panic`.

I usually try to return an error instead of panicing. In this situation though, I think you are returning a `nil` result. If this is not case meant to be handled there, you should either return an error or panic there. If it's really critical and that would mess up everything else, then panic for sure.

The best long term solution is testing depth and coverage, but that's not an easy task!

Issue #1829, opened 2021-10-04
Create dot/rpc/modules unit tests


timwu20:
## Issue summary

- Write unit tests to test rpc/module functionality
- Use mocks for interfaces 
- 70% minimum

## Other information and links

- https://github.com/ChainSafe/gossamer/pull/1666

PR #1995, opened 2021-11-05
testing (rpc/modules): write new unit tests to increase code coverage

jimjbrettj:
## Changes

- Write unit tests for ./dot/rpc/modules

## Tests

```
go test ./dot/rpc/modules/ -short -v -coverprofile coverage.out
```

## Issues

- #1829

## Primary Reviewer

- @timwu20


dot/rpc/modules/author_test.go
```diff
@@ -33,10 +33,7 @@ func TestAuthorModule_HasSessionKey_WhenScaleDataEmptyOrNil(t *testing.T) {
 		decodeSessionKeysMock.ReturnArguments = []interface{}{dec, err}
 	})
 
-	module := &AuthorModule{
-		coreAPI: coremockapi,
-		logger:  log.New("service", "RPC", "module", "author"),
-	}
+	module := NewAuthorModule(log.New("service", "RPC", "module", "author"), coremockapi, nil)
```
qdm12, line 36: Do we care about the logs in the test here? Maybe change NewAuthorModule to accept the `log15.Logger` interface and use a mock expecting anything instead (or your own no-op test logger implementation)?

dot/rpc/modules/chain_test.go
```diff
@@ -15,3 +15,473 @@
...
+
+	apimocks "github.com/ChainSafe/gossamer/dot/rpc/modules/mocks"
+	"github.com/ChainSafe/gossamer/dot/types"
+	"github.com/ChainSafe/gossamer/lib/common"
+	"github.com/stretchr/testify/mock"
+	"github.com/stretchr/testify/require"
+)
+
+func TestChainModule_GetBlock(t *testing.T) {
+	mockedHash := common.NewHash([]byte{0x01, 0x02})
```
qdm12, line 33: nit It's not really `mocked`, it's just some hash. Mock usually means it's an implementation of some interface designed just for testing.

dot/rpc/modules/chain_test.go
```diff
@@ -15,3 +15,473 @@
...
+	"testing"
+
+	apimocks "github.com/ChainSafe/gossamer/dot/rpc/modules/mocks"
+	"github.com/ChainSafe/gossamer/dot/types"
+	"github.com/ChainSafe/gossamer/lib/common"
+	"github.com/stretchr/testify/mock"
+	"github.com/stretchr/testify/require"
+)
+
+func TestChainModule_GetBlock(t *testing.T) {
```
qdm12, line 32: Can you run this (and others) with `t.Parallel()`? Let me know if they conflict with each other :wink:

dot/rpc/modules/chain_test.go
```diff
@@ -15,3 +15,473 @@
...
+	mockedHash := common.NewHash([]byte{0x01, 0x02})
+	emptyBlock := types.NewEmptyBlock()
+
+	mockBlockAPI := new(apimocks.BlockAPI)
+	mockBlockAPI.On("GetBlockByHash", mock.AnythingOfType("common.Hash")).Return(&emptyBlock, nil)
```
qdm12, line 37: Try to avoid `mock.AnythingOfType` / `mock.Anything`, instead give it what it should expect. I guess it should be deterministic here. If it has some randomness factor, then you need to mock the randomness so it's always the same in the test.

I personally use `mock.AnythingOfType` when for example it's a timing out context, which you can't predict or mock. But it's rather rare.

dot/rpc/modules/chain_test.go
```diff
@@ -15,3 +15,473 @@
...
+		args    args
+		wantErr bool
+	}{
+		{
+			name: "GetBlock OK",
+			fields: fields{
+				chainModule.blockAPI,
+			},
+			args: args{
+				r:   nil,
```
qdm12, line 70: nit You could remove go zero value fields. It does make it a bit clearer, but it does add quite some lines in the long run.

dot/rpc/modules/chain_test.go
```diff
@@ -15,3 +15,473 @@
...
+		r   *http.Request
+		req *ChainHashRequest
+		res *ChainBlockResponse
+	}
+	tests := []struct {
+		name    string
+		fields  fields
+		args    args
+		wantErr bool
+	}{
```
qdm12, line 63: nit A tip/suggestion for the future (please don't refactor your test code, it's fine!), you can use a map[string]struct{...} where the map key is the name of the test.

dot/rpc/modules/chain_test.go
```diff
@@ -15,3 +15,473 @@
...
+	type args struct {
+		r   *http.Request
+		req *ChainHashRequest
+		res *ChainBlockResponse
+	}
+	tests := []struct {
+		name    string
+		fields  fields
+		args    args
+		wantErr bool
```
qdm12, line 62: You should have an actual error here instead. I've seen cases where the wrong error was returned so asserting the actual error message is useful.

And then you can assert it with something like

```go
if testCase.err != nil {
	assert.EqualError(t, err, testCase.err.Error())
} else {
	assert.NoError(t, err)
}
```

dot/rpc/modules/chain_test.go
```diff
@@ -15,3 +15,473 @@
...
+				r:   nil,
+				req: &ChainHashRequest{},
+				res: &res,
+			},
+			wantErr: false,
+		},
+		{
+			name: "GetBlockByHash Err",
+			fields: fields{
+				mockBlockAPIGetHashErr,
```
qdm12, line 79: You should define these mocks in the actual subtest below. Instead you can have parameters to configure your mock, here I guess the expected hash and the error to return. Otherwise they'll conflict with each other, especially more once you start running parallel subtests.

dot/rpc/modules/chain_test.go
```diff
@@ -15,3 +15,473 @@
...
+				r:   nil,
+				req: &ChainHashRequest{&mockedHash},
+				res: &res,
+			},
+			wantErr: false,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			cm := &ChainModule{
```
qdm12, line 103: To run each subtest in parallel:

```suggestion
	for _, tt := range tests {
		tt := tt
		t.Run(tt.name, func(t *testing.T) {
			t.Parallel()

			cm := &ChainModule{
```

qdm12 (commented):
Some improvement comments :wink: I'll continue later!

dot/rpc/modules/chain_test.go
timwu20: If there's not too many error cases being produced by the function I don't mind using `wantErr`.

dot/rpc/modules/chain_test.go
timwu20: this structure is code generated by goland and vscode.

dot/rpc/modules/chain_test.go
```diff
@@ -15,3 +15,448 @@
...
+				blockAPI: tt.fields.blockAPI,
+			}
+			if err := cm.GetBlock(tt.args.r, tt.args.req, tt.args.res); (err != nil) != tt.wantErr {
+				t.Errorf("GetBlock() error = %v, wantErr %v", err, tt.wantErr)
+			}
+		})
```
timwu20, line 104: we should have a `expected` attribute in each test.  Since there's no return param and this function is expecting it to write to `res` the code generation doesn't generate the `expected` attribute.  So we need to inspect the `tt.args.res` for an expected `Block`.

dot/rpc/modules/chain_test.go
```diff
@@ -15,3 +15,448 @@
...
+			},
+			wantErr: true,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			cm := &ChainModule{
+				blockAPI: tt.fields.blockAPI,
+			}
+			if err := cm.GetBlockHash(tt.args.r, tt.args.req, tt.args.res); (err != nil) != tt.wantErr {
```
timwu20, line 205: needs to check that the `res *ChainHashResponse` is what is expected.

timwu20 (changes requested):
similar feedback for all the rpc modules tests, we should be inspecting the response in these tests.

dot/rpc/modules/author_test.go
jimjbrettj: @qdm12 I'm a little confused on this one, aren't we going to be replacing the log15 library?

qdm12: Yes we are, it's just that I see sooo much logs in my tests, I'm trying to restrict them at least for new tests. Or at least make people aware, we shouldn't need logs by default in our tests :wink: 

What you could do, which is easier, is:

```go
logger := log.New("service", "RPC", "module", "author")
handler := log.StreamHandler(ioutil.Discard, log.TerminalFormat())
handler = log.CallerFileHandler(handler)
logger.SetHandler(log.LvlFilterHandler(log.LvlTrace, handler))

module := NewAuthorModule(logger, coremockapi, ...)
```

You can write a little test helper function to create a logger that discards everything by default too, if you want to re-use it across tests. You can have that helper function take a boolean to enable actual logging (swap `ioutil.Discard` with `os.Stdout`), when we need to debug a test with additional logs.

dot/rpc/modules/chain_test.go
qdm12: Makes sense, won't comment again on that!

This reminds me I got mad at Github Copilot for generating test cases as slices :smile:

dot/rpc/modules/chain_test.go
qdm12: Yes it makes sense

dot/rpc/modules/chain_test.go
jimjbrettj: So I'm running the tests in parallel now and haven't gotten conflicts, is it still best to make this change?

qdm12: Try

```sh
CGO_ENABLED=1 go test -race -run ^TestAuthorModule_HasSessionKey_WhenScaleDataEmptyOrNil$ github.com/ChainSafe/gossamer/dot/rpc/modules
```

To see if it finds any races on the mocks? If not leave it as is and my apologies!

jimjbrettj: Okay for that test specifically there weren't issues, but when I ran it on the whole pkg I had conflicts. Will make these changes!

qdm12: > Okay for that test specifically there weren't issues

Are you running the subtests below with `t.Parallel()` though? If the race detector doesn't throw up on the single test, it means all the subtests don't data race on shared variables (like the mock). That might be true as well since the mock might just be read/executed and not written to.

> but when I ran it on the whole pkg I had conflicts

That could be due to a million different things, don't invest too much time in that either.
Although fixing the data races on tests not running in parallel would be nice, so we can sometime soon setup the CI to use the race detector to detect newer data races introduced.

jimjbrettj: Yeah I have t.Parallel() in both the test and subtests. So do you think then I should make these changes, and if there are still conflicts after when I run the pkg tests with -race move on for now?

dot/rpc/modules/chain_test.go
```diff
@@ -15,3 +15,467 @@
...
+	for _, tt := range tests {
+		tt := tt
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+			cm := &ChainModule{
+				blockAPI: tt.fields.blockAPI,
+			}
+			if err := cm.GetFinalizedHead(tt.args.r, tt.args.req, tt.args.res); (err != nil) != tt.wantErr {
+				t.Errorf("GetFinalizedHead() error = %v, wantErr %v", err, tt.wantErr)
+			}
```
noot, line 274: same here - check for expected `res`

noot: same for all tests in this file i think

dot/rpc/modules/childstate_integration_test.go
```diff
@@ -293,5 +293,6 @@ func setupChildStateStorage(t *testing.T) (*ChildStateModule, common.Hash) {
 	require.NoError(t, err)
 
 	hash, _ := st.Block.GetBlockHash(b.Header.Number)
+	fmt.Println("Hash: ", hash)
```
noot, line 296: remove

noot (commented):
tests are much improved!! pls check `res` in all the tests by adding some `expected` field to the test cases

dot/rpc/modules/chain_test.go
qdm12: If you are rewriting all the tests in the package, you might want to fix the data races **before** adding `t.Parallel()` to tests.

If you find data races in existing tests running without parallelism, then it might be some critical data race in the production code.

Now when you  all your tests in parallel (with `t.Parallel()`), it will likely reveal additional data races, which are mostly test only data races (e.g. writing to a shared test global variable).

Let me know if you need help figuring it out, we can pair program tomorrow for ~30 min or so to fix them for the tests of that package.

dot/rpc/modules/chain_test.go
```diff
@@ -15,3 +15,448 @@
...
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			cm := &ChainModule{
+				blockAPI: tt.fields.blockAPI,
+			}
+			if err := cm.GetBlock(tt.args.r, tt.args.req, tt.args.res); (err != nil) != tt.wantErr {
+				t.Errorf("GetBlock() error = %v, wantErr %v", err, tt.wantErr)
+			}
```
qdm12, line 103: nit this works, but since we already have the testify dependency, you might want to use

```go
assert.ErrorIs(t, err, tt.wantErr)
```

dot/rpc/modules/chain_test.go
```diff
@@ -15,3 +15,448 @@
...
+		})
+	}
+}
+
+func TestChainModule_GetBlockHash(t *testing.T) {
+	testHash := common.NewHash([]byte{0x01, 0x02})
+	i := make([]interface{}, 1)
+	i[0] = "a"
```
qdm12, line 111: nit

```suggestion
	i := []interface{}{"a"}
```

dot/rpc/modules/chain_test.go
```diff
@@ -15,3 +15,448 @@
...
+		wantErr bool
+	}{
+		{
+			name: "GetHighestFinalisedHash OK",
+			fields: fields{
+				mockBlockAPI,
+			},
+			args: args{
+				req: &EmptyRequest{},
+				res: &res,
```
qdm12, line 243: That will likely mess up if we run each subtest in parallel later, since I would assume they all write to the same response in parallel?

Instead define it in the subtest.

As a general rule, it's usually better to define things in the subtest (in the for loop) scope instead of the parent scope, especially when it comes to pointy pointers/interfaces.

jimjbrettj: To clarify, should I just define the response in the subtest, or should I be defining the test cases in their entirety in the subtest?

dot/rpc/modules/test_data/test_metadata.go
```diff
@@ -0,0 +1,19 @@
...
...6d756d206e756d626572206f66207375622d6163636f756e747320616c6c6f77656420706572206964656e746966696564206163636f756e742e4c4d61784164646974696f6e616c4669656c64730c7533321064000000086501204d6178696d756d206e756d626572206f66206164646974696f6e616c206669656c64732074686174206d61792062652073746f72656420696e20616e2049442e204e656564656420746f20626f756e642074686520492f4fe020726571756972656420746f2061636365737320616e206964656e746974792c206275742063616e2062652070726574747920686967682e344d6178526567697374726172730c7533321014000000085101204d61786d696d756d206e756d626572206f66207265676973747261727320616c6c6f77656420696e207468652073797374656d2e204e656564656420746f20626f756e642074686520636f6d706c65786974797c206f662c20652e672e2c207570646174696e67206a756467656d656e74732e4048546f6f4d616e795375624163636f756e7473046020546f6f206d616e7920737562732d6163636f756e74732e204e6f74466f756e640454204163636f756e742069736e277420666f756e642e204e6f744e616d65640454204163636f756e742069736e2774206e616d65642e28456d707479496e646578043420456d70747920696e6465782e284665654368616e676564044020466565206973206368616e6765642e284e6f4964656e74697479044c204e6f206964656e7469747920666f756e642e3c537469636b794a756467656d656e74044820537469636b79
```
qdm12, line 19: My IDE doesn't manage to render since the string is too big. Instead have your hex string in a file e.g. `metadata.hex` and use

```suggestion
import (
	"io"
	"os"

	"github.com/ChainSafe/gossamer/lib/common"
)

func init() {
	var err error
	defer func() {
		if err != nil {
			panic(err)
		}
	}()

	var file *os.File
	file, err = os.Open("metadata.hex")
	if err != nil {
		return
	}

	var b []byte
	b, err = io.ReadAll(file)
	if err != nil {
		_ = file.Close()
		return
	}

	err = file.Close()
	if err != nil {
		return
	}

	testData = common.BytesToHex(b)
}

var testData string

// GetTestData returns the testData string.
func GetTestData() string {
	return testData
}

```

Note `testData` is unexported since it's not a constant, we don't want other packages to accidentally modify it. Use `GetTestData()` instead.

Despite hating `init()` functions, it sort of makes sense here, especially if we all `GetTestData()` in multiple places.

If the file size is bigger than a few KBs, you can even go further and compress it (e.g. with zip) and decompress it in the `init()`.

dot/rpc/modules/chain_test.go
qdm12: I'm not sure anymore since the code changed when I pulled it haha! I think it was about `res` being shared among subtests and written to.

Also, you can also write your test cases without `args` and `fields` structs and have everything at the same level of indentation, having tried both, I find it cleaner. But up to you!

dot/rpc/modules/dev_test.go
```diff
@@ -77,18 +79,28 @@ func TestDevModule_EpochLength(t *testing.T) {
...
+			var err error
+			if err = m.EpochLength(tt.args.r, tt.args.req, tt.args.res); (err != nil) != tt.wantErr {
 				t.Errorf("EpochLength() error = %v, wantErr %v", err, tt.wantErr)
+				return
+			}
+			if tt.wantErr {
+				assert.EqualError(t, err, tt.err.Error())
+			} else {
+				assert.NoError(t, err)
+				assert.Equal(t, tt.exp, *tt.args.res)
```
qdm12, line 103: You can probably move this after the `else` clause. If there is an error, it should not modify `t.args.res`, so the assertion should pass as well.

Also it's usually easier to have `tt.args.res` as a value, and pass it as pointer with `&tt.args.res` to the function `EpochLength`. I think you save a single *` character overall :exploding_head:

dot/rpc/modules/dev_test.go
```diff
@@ -122,19 +135,28 @@ func TestDevModule_SlotDuration(t *testing.T) {
...
 			args: args{
 				req: &EmptyRequest{},
-				res: &res,
 			},
+			exp: "0x1700000000000000",
 		},
 	}
 	for _, tt := range tests {
+		var res string
+		tt.args.res = &res
```
qdm12, line 144: move this inside the subtest. Once you run (in the future) parallel subtests, each run in their own goroutine so it's rather hard to ensure they don't access some shared pointer etc.

TLDR, don't put anything between the `for` and the `func(t *testing.T) {` of the subtest :wink:

dot/rpc/modules/author_test.go
```diff
@@ -101,59 +103,99 @@ func TestAuthorModule_HasSessionKey_WhenThereIsNoKeys(t *testing.T) {
...
+	})
+
+	mockHasKey2 := coremockapi2.On("HasKey", mock.AnythingOfType("string"), mock.AnythingOfType("string"))
+	mockHasKey2.Run(func(args mock.Arguments) {
+		pubKeyHex := args.Get(0).(string)
+		keyType := args.Get(1).(string)
+
+		ok, err := keystore.HasKey(pubKeyHex, keyType, globalStore.Acco)
+		mockHasKey2.ReturnArguments = []interface{}{ok, err}
+	})
```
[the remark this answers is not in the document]
jimjbrettj: I believe because HasKey is not part of the keystore interface

dot/rpc/modules/dev_test.go
```diff
@@ -1 +1,326 @@
...
+				input: uint64(18446744073709551615),
+			},
+			want: "0xffffffffffffffff",
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			if got := uint64ToHex(tt.args.input); got != tt.want {
+				t.Errorf("uint64ToHex() = %v, want %v", got, tt.want)
+			}
```
qdm12, line 50: nit use testify's `assert` lib

```suggestion
			got := uint64ToHex(tt.args.input)
			assert.Equal(t, tt.want, got)
```

Unless it's generated by your IDE then up to you

dot/rpc/modules/sync_state_test.go
```diff
@@ -1 +1,200 @@
...
+			} else {
+				assert.NoError(t, err)
+				assert.Equal(t, tt.exp, res)
+			}
+		})
+	}
+}
+
+func Test_syncState_GenSyncSpec(t *testing.T) {
+	g := &genesis.Genesis{}
```
qdm12, line 154: Define inline in your test cases

i.e.

```go
fields: fields{
    chainSpecification: &genesis.Genesis{},
},
```

For readability and to avoid subtests to eventually depend on each other

dot/rpc/modules/test_data/test_metadata.go
```diff
@@ -0,0 +1,50 @@
...
+)
+
+//go:embed test_metadata.zip
+var testMetaData []byte
+
+func init() {
+	bytesReader := bytes.NewReader(testMetaData)
+	zipReader, err := zip.NewReader(bytesReader, int64(bytesReader.Len()))
+	if err != nil {
+		return
```
qdm12, line 20: You should panic when an error occurs, otherwise it'll be hard to understand why the test is failing.

go.mod
```diff
@@ -45,8 +45,9 @@ require (
 	google.golang.org/protobuf v1.27.1
 )
 
+require github.com/ChainSafe/log15 v1.0.0
```
qdm12, line 48: Remove usages of `log15` you have in your tests by replacing with `github.com/ChainSafe/gossamer/internal/log` and then re-run `go mod tidy`

dot/rpc/modules/dev_test.go
jimjbrettj: It was generated but I do like assets better I think, so I can work on shifting

jimjbrettj: I also think there were redundant checks here for a lot of these tests, so I think this change will clean it up a bit more as well

dot/rpc/modules/sync_state_test.go
jimjbrettj: Good call. The other instance of this in the above function I wasn't able to inline, but there were only 2 instances so I just made a second variable for it

dot/rpc/modules/author_integration_test.go
```diff
@@ -35,7 +35,7 @@ var testInvalidExt = []byte{1, 212, 53, 147, 199, 21, 253, 211, 28, 97, 20, 26,
 func TestMain(m *testing.M) {
 	wasmFilePaths, err := runtime.GenerateRuntimeWasmFile()
 	if err != nil {
-		log.Error("failed to generate runtime wasm file", err)
+		log.Errorf("failed to generate runtime wasm file", err)
```
qdm12, line 38:
```suggestion
		log.Errorf("failed to generate runtime wasm file: %s", err)
```

dot/rpc/modules/author_integration_test.go
```diff
@@ -48,7 +48,7 @@ func TestMain(m *testing.M) {
 
 func TestAuthorModule_Pending(t *testing.T) {
 	txQueue := state.NewTransactionState()
-	auth := NewAuthorModule(nil, nil, txQueue)
+	auth := NewAuthorModule(log.New(), nil, txQueue)
```
qdm12, line 51:
```suggestion
	auth := NewAuthorModule(log.New(log.SetWriter(io.Discard)), nil, txQueue)
```

dot/rpc/modules/author_integration_test.go
```diff
@@ -277,5 +277,5 @@ func setupAuthModule(t *testing.T, txq *state.TransactionState) *AuthorModule {
 	t.Cleanup(func() {
 		rt.Stop()
 	})
-	return NewAuthorModule(nil, cs, txq)
+	return NewAuthorModule(log.New(), cs, txq)
```
qdm12, line 280:
```suggestion
	return NewAuthorModule(log.New(log.SetWriter(io.Discard)), cs, txq)
```

dot/rpc/modules/author_test.go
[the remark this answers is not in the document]
qdm12: Is there a reason for it not being in the interface? Should we add it :thinking:

dot/rpc/modules/author_test.go
```diff
@@ -228,23 +274,27 @@ func TestAuthorModule_SubmitExtrinsic(t *testing.T) {
...
-			if err := am.SubmitExtrinsic(tt.args.r, tt.args.req, tt.args.res); (err != nil) != tt.wantErr {
-				t.Errorf("AuthorModule.SubmitExtrinsic() error = %v, wantErr %v", err, tt.wantErr)
-			}
-			if diff := cmp.Diff(tt.wantRes, *tt.args.res); diff != "" {
-				t.Errorf("unexpected response: %s", diff)
+			err := am.SubmitExtrinsic(tt.args.r, tt.args.req, tt.args.res)
+			if tt.wantErr {
+				if assert.NotNil(t, err) {
+					assert.EqualError(t, err, tt.err.Error())
+				}
```
qdm12, line 294:
```suggestion
				require.Error(t, err) {
				assert.EqualError(t, err, tt.err.Error())
```

You can use `require` to make the test instantly fail and not continue instead. That usually reduces indentation and if/else clauses.

dot/rpc/modules/author_test.go
```diff
@@ -228,23 +274,27 @@ func TestAuthorModule_SubmitExtrinsic(t *testing.T) {
...
+			} else {
+				assert.NoError(t, err)
+				assert.Equal(t, &tt.wantRes, tt.args.res)
```
qdm12, line 297: nit maybe put this line under the `if`/`else` clauses, since ideally you should check both the error and the value returned (e.g. in case some very bad person ignores the error for whatever evil reason)

dot/rpc/modules/author_test.go
```diff
@@ -298,36 +346,55 @@ func TestAuthorModule_PendingExtrinsics(t *testing.T) {
...
-			},
 			wantRes: PendingExtrinsicsResponse{
 				common.BytesToHex([]byte("someExtrinsic")),
 				common.BytesToHex([]byte("someExtrinsic1")),
 			},
 		},
 	}
 	for _, tt := range tests {
+		res := PendingExtrinsicsResponse{}
+		tt.args.res = &res
```
qdm12, line 357: Can you put these in the subtest please? It looks strange to be above here, and might not work with future eventual parallel subtests

dot/rpc/modules/author_test.go
```diff
@@ -410,28 +480,32 @@ func TestAuthorModule_InsertKey(t *testing.T) {
...
 			}
 		})
 	}
 }
 
 func TestAuthorModule_HasKey(t *testing.T) {
+	kr, err := keystore.NewSr25519Keyring()
+	if err != nil {
+		panic(err)
+	}
```
qdm12, line 499: Don't panic in tests, use `t.Fatal` or `require.XXX` functions instead

```suggestion
	require.NoError(t, err)
```

dot/rpc/modules/state.go
```diff
@@ -226,6 +228,9 @@ func (sm *StateModule) GetKeysPaged(_ *http.Request, req *StateStorageKeyRequest
 		return err
 	}
 	keys, err := sm.storageAPI.GetKeysWithPrefix(req.Block, hPrefix)
+	if err != nil {
+		return err
```
qdm12, line 232: A bit out of scope, but we should return a context wrapping the error

```suggestion
		return fmt.Errorf("cannot get keys with prefix %s: %w", hPrefix, err)
```

dot/rpc/modules/state.go
```diff
@@ -181,11 +181,13 @@ func (sm *StateModule) GetPairs(_ *http.Request, req *StatePairRequest, res *Sta
...
 			*res = append(*res, []string{common.BytesToHex([]byte(k)), common.BytesToHex(v)})
 		}
-
 		return nil
 	}
 
-	reqBytes, _ := common.HexToBytes(*req.Prefix)
+	reqBytes, err := common.HexToBytes(*req.Prefix)
+	if err != nil {
+		return err
```
qdm12, line 189: A bit out of scope, but we should return a context wrapping the error

```suggestion
		return fmt.Errorf("cannot convert hex prefix %s to bytes: %w", *req.Prefix, err)
```

dot/rpc/modules/system_integration_test.go
```diff
@@ -33,6 +33,7 @@ import (
 	"github.com/ChainSafe/gossamer/dot/rpc/modules/mocks"
 	"github.com/ChainSafe/gossamer/dot/state"
 	"github.com/ChainSafe/gossamer/dot/types"
+	log "github.com/ChainSafe/gossamer/internal/log"
```
qdm12, line 36:
```suggestion
	"github.com/ChainSafe/gossamer/internal/log"
```

dot/rpc/modules/test_data/test_metadata.go
```diff
@@ -0,0 +1,50 @@
...
+
+//go:embed test_metadata.zip
+var testMetaData []byte
+
+func init() {
+	bytesReader := bytes.NewReader(testMetaData)
+	zipReader, err := zip.NewReader(bytesReader, int64(bytesReader.Len()))
+	if err != nil {
+		panic(err)
+	}
```
qdm12, line 21:
```suggestion
	} else if len(zipReader.File) == 0 {
		panic("no file in test_metadata.zip"	)
	}
```

dot/rpc/modules/author_test.go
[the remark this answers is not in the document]
jimjbrettj: Hmm in not sure. I think it is outside of the scope of this PR, but maybe if @noot has thoughts on this we can open a different issue for it

dot/rpc/modules/dev_test.go
jimjbrettj: Missed this one, but will make this change!

dot/rpc/modules/author_test.go
```diff
@@ -30,7 +32,7 @@ func TestAuthorModule_HasSessionKey_WhenScaleDataEmptyOrNil(t *testing.T) {
 
 	coremockapi := new(apimocks.CoreAPI)
 
-	decodeSessionKeysMock := coremockapi.On("DecodeSessionKeys", mock.AnythingOfType("[]uint8"))
+	decodeSessionKeysMock := coremockapi.On("DecodeSessionKeys", common.MustHexToBytes("0x0400"))
 	decodeSessionKeysMock.Run(func(args mock.Arguments) {
 		b := args.Get(0).([]byte)
 		dec, err := runtimeInstance.DecodeSessionKeys(b)
```
timwu20, line 38: Do we really need to instantiate a `*wasmer.Instance` to run this test?  Seems like this is the only thing slowing down this package at this point.

dot/rpc/modules/author_test.go
```diff
@@ -51,12 +53,12 @@ func TestAuthorModule_HasSessionKey_WhenScaleDataEmptyOrNil(t *testing.T) {
 	require.NoError(t, err)
 	require.False(t, bool(res))
 
-	coremockapi.AssertCalled(t, "DecodeSessionKeys", mock.AnythingOfType("[]uint8"))
+	coremockapi.AssertCalled(t, "DecodeSessionKeys", common.MustHexToBytes("0x0400"))
 }
 
 func TestAuthorModule_HasSessionKey_WhenRuntimeFails(t *testing.T) {
```
timwu20, line 59: Can we revise all the `HasSessionKey` to follow the slice of anonymous structs style of tests.

dot/rpc/modules/author_test.go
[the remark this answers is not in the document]
timwu20: From what I gather this test doesn't even need a keystore.  This test also seems like more of an integration test, testing what happens based on all these calls before calling `HasSessionKey`.  Do we even need to keep this test in the test suite? If we're testing `AuthorModule.HasSessionKey`, we should only be mocking `am.coreAPI.DecodeSessionKey` and `am.coreAPI.HasKey`.

dot/rpc/modules/chain_test.go
```diff
@@ -2,3 +2,529 @@
...
+
+	mockBlockAPIGetHashErr := new(apimocks.BlockAPI)
+	mockBlockAPIGetHashErr.On("GetBlockByHash", inputHash).Return(nil, errors.New("GetJustification error"))
+
+	bodyBlock := types.NewEmptyBlock()
+	bodyBlock.Body = types.BytesArrayToExtrinsics([][]byte{{1}})
+	mockBlockAPIWithBody := new(apimocks.BlockAPI)
+	mockBlockAPIWithBody.On("GetBlockByHash", inputHash).Return(&bodyBlock, nil)
+
+	expEmptyBlock := &ChainBlockResponse{Block: ChainBlock{
```
timwu20, line 37: nit, move these within the test structs if they aren't reused by multiple tests.

dot/rpc/modules/chain_test.go
```diff
@@ -2,3 +2,529 @@
...
+		r   *http.Request
+		req *ChainBlockNumberRequest
+		res *ChainHashResponse
+	}
+	tests := []struct {
+		name    string
+		fields  fields
+		args    args
+		wantErr bool
+		err     error
```
timwu20, line 154:
```suggestion
		expErr     error
```

dot/rpc/modules/chain_test.go
```diff
@@ -2,3 +2,529 @@
...
+			fields: fields{
+				mockBlockAPI,
+			},
+			args: args{
+				req: &EmptyRequest{},
+			},
+			exp: &expRes,
+		},
+		{
+			name: "GetHighestFinalisedHash ERR",
```
timwu20, line 277:
```suggestion
			name: "error case",
```

dot/rpc/modules/chain_test.go
```diff
@@ -2,3 +2,529 @@
...
+	tests := []struct {
+		name    string
+		fields  fields
+		args    args
+		wantErr bool
+		err     error
+		exp     *ChainHashResponse
+	}{
+		{
+			name: "GetHighestFinalisedHash OK",
```
timwu20, line 267:
```suggestion
			name: "happy path",
```

dot/rpc/modules/chain_test.go
timwu20: The test name shows that it's testing `GetHighestFinalisedHash`.

dot/rpc/modules/payment_test.go
```diff
@@ -1 +1,172 @@
...
+			args: args{
+				req: &PaymentQueryInfoRequest{
+					Ext:  "0x0000",
+					Hash: &testHash,
+				},
+			},
+			exp: PaymentQueryInfoResponse{
+				Weight:     uint64(21),
+				Class:      21,
+				PartialFee: scale.MustNewUint128(new(big.Int).SetBytes([]byte{1, 2, 3, 4, 5, 6, 7, 8, 1, 2, 3, 4, 5, 6})).String(),
```
timwu20, line 97: line length?

dot/rpc/modules/system_test.go
```diff
@@ -1,4 +1,702 @@
...
+	"github.com/ChainSafe/gossamer/dot/types"
+	"github.com/ChainSafe/gossamer/lib/common"
+	"github.com/ChainSafe/gossamer/lib/transaction"
+	"github.com/multiformats/go-multiaddr"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestSystemModule_BasicGetters(t *testing.T) {
```
timwu20, line 22: I'd prefer each receiver function have its own test.

dot/rpc/modules/system_test.go
```diff
@@ -1,4 +1,702 @@
...
+	require.NoError(t, err)
+	require.NotNil(t, res)
+
+	err = sm.Peers(nil, req, &sysPeerRes)
+	require.NoError(t, err)
+	require.NotNil(t, res)
+	require.Equal(t, SystemPeersResponse{}, sysPeerRes)
+}
+
+func TestSystemModule_TestNodeRoles(t *testing.T) {
```
timwu20, line 85:
```suggestion
func TestSystemModule_NodeRoles(t *testing.T) {
```

dot/rpc/modules/system_test.go
```diff
@@ -1,4 +1,702 @@
...
+		blockAPI   BlockAPI
+	}
+	type args struct {
+		r   *http.Request
+		req *StringRequest
+		res *U64Response
+	}
+	tests := []struct {
+		name    string
+		fields  fields
```
timwu20, line 220: you can replace fields with just `SystemModule` and use the `newSystemModule` constructor.

dot/rpc/modules/test_data/test_metadata.go
```diff
@@ -0,0 +1,50 @@
...
+
+	testData = string(b)
+}
+
+var testData string
+
+// GetTestData returns the testData string.
+func GetTestData() string {
```
timwu20, line 48: what kind of test data is this? Maybe rename to be more explicit and follow `New....` syntax.

timwu20 (changes requested):

dot/rpc/modules/author_test.go
```diff
@@ -353,9 +353,9 @@ func TestAuthorModule_PendingExtrinsics(t *testing.T) {
 		},
 	}
 	for _, tt := range tests {
-		res := PendingExtrinsicsResponse{}
-		tt.args.res = &res
 		t.Run(tt.name, func(t *testing.T) {
+			res := PendingExtrinsicsResponse{}
+			tt.args.res = &res
```
qdm12: Why not set `res: &PendingExtrinsicsResponse{}` directly in the `tests` declaration above by the way? :thinking: 

Same comment for other tests

dot/rpc/modules/childstate_test.go
```diff
@@ -138,9 +138,9 @@ func TestChildStateModule_GetKeys(t *testing.T) {
 		},
 	}
 	for _, tt := range tests {
-		res := make([]string, 0)
-		tt.args.res = &res
 		t.Run(tt.name, func(t *testing.T) {
+			res := make([]string, 0)
```
qdm12: nit the same as:

```suggestion
			res := []string{}
```

Also you could move this in your `tests` declaration by using a little helper function like

```go
stringSlicePtr := func(s []string) *[]string { return &s }
```

which you can declare inline in your test before `tests := ...`

dot/rpc/modules/author_test.go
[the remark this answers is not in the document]
jimjbrettj: @timwu20 I think you are right. This test was in the file before so I thought to keep it in but I don't think its needed. I'm thinking I will remove these tests and just add unit tests for the remaining functions not covered by the tests I wrote

dot/rpc/modules/system_test.go
jimjbrettj: I did this initially but got errors since the integration test has a function of the same name

dot/rpc/modules/author_test.go
jimjbrettj: I think this would work, will change

jimjbrettj: Actually on second thought, this will add a lot of lines so I'm not sure that its worth doing this change. If you think it is I will though

dot/rpc/modules/system_test.go
jimjbrettj: I left as is for now, but maybe we let these default names be for the unit tests, then use a different naming convention for the integration tests

dot/rpc/modules/chain_test.go
```diff
@@ -259,7 +259,7 @@ func TestChainModule_GetFinalizedHead(t *testing.T) {
 		exp     ChainHashResponse
 	}{
 		{
-			name: "GetHighestFinalisedHash OK",
+			name: "happy path",
```
qdm12, line 262: :laughing:

dot/rpc/modules/system_test.go
```diff
@@ -95,54 +95,46 @@ func TestSystemModule_TestNodeRoles(t *testing.T) {
...
+			fields: NewSystemModule(mockNetworkAPI3, nil, nil, nil, nil, nil),
 			args: args{
 				req: &EmptyRequest{},
 			},
 			exp: []interface{}{"Authority"},
 		},
 		{
 			name:   "UnknownRole",
-			fields: fields{mockNetworkAPI4, nil, nil, nil, nil, nil},
+			fields: NewSystemModule(mockNetworkAPI4, nil, nil, nil, nil, nil),
```
qdm12, line 175: Nice :+1:  Although perhaps rename `fields`? (same for below tests)

dot/rpc/modules/test_data/test_metadata.go
```diff
@@ -46,7 +46,7 @@ func init() {
 
 var testData string
 
-// GetTestData returns the testData string.
-func GetTestData() string {
+// NewTestMetadata returns the testData string.
+func NewTestMetadata() string {
```
qdm12, line 50: Sure :smile:

dot/rpc/modules/author_test.go
```diff
@@ -14,173 +14,145 @@ import (
...
-	require.False(t, bool(res))
-
-	coremockapi.AssertCalled(t, "DecodeSessionKeys", mock.AnythingOfType("[]uint8"))
-}
-
-func TestAuthorModule_HasSessionKey_WhenRuntimeFails(t *testing.T) {
-	coremockapi := new(apimocks.CoreAPI)
-	coremockapi.On("DecodeSessionKeys", mock.AnythingOfType("[]uint8")).Return(nil, errors.New("problems with runtime"))
+func TestAuthorModule_HasSessionKeys(t *testing.T) {
+	testReq := "0xd43593c715fdd31c61141abd04a99fd6822c8558854ccde39a5684e7a56da27d34309a9d2a24213896ff06895db16" +
```
qdm12, line 28: nit make stuff constant whenever possible

```suggestion
	const testReq = "0xd43593c715fdd31c61141abd04a99fd6822c8558854ccde39a5684e7a56da27d34309a9d2a24213896ff06895db16" +
```

dot/rpc/modules/author_test.go
```diff
@@ -14,173 +14,145 @@ import (
...
+	coreMockAPIOk := new(apimocks.CoreAPI)
+	coreMockAPIOk.On("DecodeSessionKeys", pkeys).Return(data, nil)
+	coreMockAPIOk.On("HasKey", mock.AnythingOfType("string"), mock.AnythingOfType("string")).Return(true, nil)
 
-func TestAuthorModule_HasSessionKey_WhenThereIsNoKeys(t *testing.T) {
-	keys := "0xd43593c715fdd31c61141abd04a99fd6822c8558854ccde39a5684e7a56da27d34309a9d2a24213896ff06895db16aade8b6502f3a71cf56374cc3852042602634309a9d2a24213896ff06895db16aade8b6502f3a71cf56374cc3852042602634309a9d2a24213896ff06895db16aade8b6502f3a71cf56374cc38520426026"
-	runtimeInstance := wasmer.NewTestInstance(t, runtime.NODE_RUNTIME)
+	coreMockAPIErr := new(apimocks.CoreAPI)
+	coreMockAPIErr.On("DecodeSessionKeys", pkeys).Return(data, nil)
+	coreMockAPIErr.On("HasKey", mock.AnythingOfType("string"), mock.AnythingOfType("string")).Return(false, errors.New("HasKey err"))
```
qdm12, line 50: Let's not have huge strings

```suggestion
	coreMockAPIErr.
		On("HasKey", mock.AnythingOfType("string"), mock.AnythingOfType("string")).
		Return(false, errors.New("HasKey err"))
```

dot/rpc/modules/author_test.go
```diff
@@ -14,173 +14,145 @@ import (
...
-
-	req := &HasSessionKeyRequest{
-		PublicKeys: keys,
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			am := tt.fields.authorModule
+			err := am.HasSessionKeys(tt.args.r, tt.args.req, &tt.args.res)
+			if tt.wantErr {
+				require.Error(t, err)
+				assert.EqualError(t, err, tt.expErr.Error())
```
qdm12, line 136: nit no need for both since `assert.EqualError` takes care of it

```suggestion
				assert.EqualError(t, err, tt.expErr.Error())
```

dot/rpc/modules/author_test.go
```diff
@@ -228,24 +201,26 @@ func TestAuthorModule_SubmitExtrinsic(t *testing.T) {
 			},
 			args: args{
 				req: &Extrinsic{fmt.Sprintf("0x%x", testExt)},
-				res: new(ExtrinsicHashResponse),
 			},
 			wantRes: ExtrinsicHashResponse(types.Extrinsic(testExt).Hash().String()),
 		},
 	}
 	for _, tt := range tests {
+		tt.args.res = ExtrinsicHashResponse("")
```
qdm12, line 209: nit don't define it here or in the test cases (unless it's generated)

Instead define it above where it's called, i.e.

```go
			response := ExtrinsicHashResponse("")
			err := am.SubmitExtrinsic(tt.args.r, tt.args.req, &response)
```

You should only have inputs and expected data in your test cases, and this one is not really a variable input it seems. Or define it for each test case, but defining things before the subtest and after the test cases looks strange imo

dot/rpc/modules/author_test.go
```diff
@@ -309,25 +279,45 @@ func TestAuthorModule_PendingExtrinsics(t *testing.T) {
...
-			if err := cm.PendingExtrinsics(tt.args.r, tt.args.req, tt.args.res); (err != nil) != tt.wantErr {
-				t.Errorf("AuthorModule.PendingExtrinsics() error = %v, wantErr %v", err, tt.wantErr)
-				return
-			}
-			if diff := cmp.Diff(tt.wantRes, *tt.args.res); diff != "" {
-				t.Errorf("unexpected response: %s", diff)
+			err := cm.PendingExtrinsics(tt.args.r, tt.args.req, &tt.args.res)
+			if tt.wantErr {
+				require.Error(t, err)
+				assert.EqualError(t, err, tt.expErr.Error())
```
qdm12, line 291: nit then you can also get rid of `wantErr`

```suggestion
			if tt.expErr != nil {
				assert.EqualError(t, err, tt.expErr.Error())
```

dot/rpc/modules/test_data/test_metadata.go
qdm12: I suggested this. We just get the test data from the test_data package. But I'm fine with `New...` as well.
And yes naming the `TestData` to something more specific would be nice.

dot/rpc/modules/author_test.go
qdm12: See https://github.com/ChainSafe/gossamer/pull/1995#discussion_r756771003

timwu20 (changes requested):
Just the `tt.args.res` thing.  Amazing work @jimjbrettj!

timwu20 (approved):

dot/rpc/modules/chain_test.go
qdm12: I actually use `err` for the expected error :smile: I guess each build their own little naming conventions here and there

dot/rpc/modules/chain_integration_test.go
```diff
@@ -36,8 +36,8 @@ import (
 	"github.com/ChainSafe/gossamer/pkg/scale"
 
 	database "github.com/ChainSafe/chaindb"
+	log "github.com/ChainSafe/gossamer/internal/log"
```
qdm12, line 39:
```suggestion
	"github.com/ChainSafe/gossamer/internal/log"
```

dot/rpc/modules/author_test.go
```diff
@@ -314,20 +272,39 @@ func TestAuthorModule_PendingExtrinsics(t *testing.T) {
...
-				return
-			}
-			if diff := cmp.Diff(tt.wantRes, *tt.args.res); diff != "" {
-				t.Errorf("unexpected response: %s", diff)
+			res := PendingExtrinsicsResponse{}
+			err := cm.PendingExtrinsics(tt.args.r, tt.args.req, &res)
+			if tt.expErr != nil {
+				assert.EqualError(t, err, tt.expErr.Error())
+			} else {
+				assert.NoError(t, err)
```
qdm12, line 280: Just a FYI (don't do it here), I came up (like a week ago or so) with a newer way to assert errors, example:

https://github.com/ChainSafe/gossamer/blob/91ed71612dd7fc72370f212e97e5e4d7fba96743/lib/trie/leaf/header_test.go#L64-L69

That's nice to assert the wrapped error is the expected one as well as to check the error message.
As in, **eventually** use that in the future. Your method (which was mine for years as well :smile:) works quite fine too so no worry.

dot/rpc/modules/chain_test.go
```diff
@@ -2,3 +2,494 @@
...
+	}
+}
+
+func TestChainModule_ErrSubscriptionTransport(t *testing.T) {
+	req := &EmptyRequest{}
+	res := &ChainBlockHeaderResponse{}
+	cm := NewChainModule(new(apimocks.BlockAPI))
+
+	err := cm.SubscribeFinalizedHeads(nil, req, res)
+	require.Equal(t, err, ErrSubscriptionTransport)
```
qdm12, line 427: nit Use `require.ErrorIs` instead? As in Equal works but ErrorIs sounds more appropriate here (in case we wrap the error later too)

Same for below calls.

dot/rpc/modules/chain_test.go
```diff
@@ -2,3 +2,494 @@
...
+	}{
+		{
+			name: "HeaderToJSON Empty Header",
+			args: args{
+				header: *emptyHeader,
+			},
+			exp: expResEmpty,
+		},
+		{
+			name: "HeaderToJSON NonEmpty Header",
```
qdm12, line 478: Just call it `not empty`. The parent test name already contains the `HeaderToJSON` string. Same for above test case as well.

dot/rpc/modules/dev_test.go
```diff
@@ -1 +1,307 @@
...
+			exp: "network service started",
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			m := &DevModule{
+				networkAPI:       tt.fields.networkAPI,
+				blockProducerAPI: tt.fields.blockProducerAPI,
+			}
+			res := ""
```
qdm12, line 297: supernit usually use `var xxx type` to initialize a variable to its zero value. More of a convention, and really supernit

```suggestion
			var res string
```

dot/rpc/modules/mocks/SyncStateAPI.go
```diff
@@ -0,0 +1,36 @@
+// Code generated by mockery v2.9.4. DO NOT EDIT.
```
qdm12, line 1: Do you have `//go:generate` comments for these new mocks?

dot/rpc/modules/offchain_test.go
```diff
@@ -1 +1,210 @@
...
+			s := &OffchainModule{
+				nodeStorage: tt.fields.nodeStorage,
+			}
+			res := StringResponse("")
+			err := s.LocalStorageSet(tt.args.in0, tt.args.req, &res)
+			if tt.expErr != nil {
+				assert.EqualError(t, err, tt.expErr.Error())
+			} else {
+				assert.NoError(t, err)
+			}
```
qdm12, line 207: assert the response perhaps?

dot/rpc/modules/state.go
qdm12: Reminder @jimjbrettj  :wink:

go.mod
qdm12: This is still there, please re-run `go mod tidy` once you got rid of `log15` in your tests.

qdm12 (changes requested):
Almost there!! :+1: (feel free to slack me to approve it)

dot/rpc/modules/mocks/SyncStateAPI.go
jimjbrettj: Could you elaborate what you mean by do I have the comments? Not really sure what you mean I just generated the mocks

dot/rpc/modules/offchain_test.go
jimjbrettj: there is no response set in this function

dot/rpc/modules/dev_test.go
```diff
@@ -2,3 +2,306 @@
 // SPDX-License-Identifier: LGPL-3.0-only
 
 package modules
+
+import (
+	"errors"
+	"net/http"
+	"testing"
+
+	apimocks "github.com/ChainSafe/gossamer/dot/rpc/modules/mocks"
```
qdm12, line 11: nit why not just call them mocks and remove the `apimocks` alias? :thinking:

dot/rpc/modules/offchain_test.go
qdm12: Isn't there that `res := StringResponse("")`? Since it's passed as pointer, even if it's not modified, you can assert it's untouched too.

dot/rpc/modules/dev_test.go
jimjbrettj: resolved!

kishansagathiya (dismissed):
Blocking this for linting and deepsource warnings.

Generally, we can ignore these, but I think this times deepsource has extra warnings. Bunch of them are good.

dot/rpc/modules/system_test.go
```diff
@@ -2,3 +2,589 @@
...
+			expErr: errors.New("getStorage error"),
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			res := U64Response(0)
+			sm := tt.sysModule
+			err := sm.AccountNextIndex(tt.args.r, tt.args.req, &res)
+			if tt.expErr != nil {
+				assert.EqualError(t, err, tt.expErr.Error())
```
kishansagathiya, line 287: Or you could keep `tt.expErr` as string so that you don't have to convert it to string every time you compare. I think this would apply to a lot of other tests.

dot/rpc/modules/mocks/BlockProducerAPI.go
```diff
@@ -0,0 +1,66 @@
+// Code generated by mockery v2.9.4. DO NOT EDIT.
```
kishansagathiya, line 1: should rename this file to block_producer_api.go

dot/rpc/modules/mocks/SyncStateAPI.go
```diff
@@ -0,0 +1,36 @@
...
+
+package mocks
+
+import (
```
kishansagathiya, line 5: should rename the file to sync_state_api.go

dot/rpc/modules/mocks/SyncStateAPI.go
kishansagathiya: He might have meant this
 
```
//go:generate mockery --name SyncStateAPI --structname SyncStateAPI --case underscore --keeptree

// SyncStateAPI is the interface to interact with sync state.
type SyncStateAPI interface {
	GenSyncSpec(raw bool) (*genesis.Genesis, error)
}
```

dot/rpc/modules/test_data/test_metadata.go
```diff
@@ -0,0 +1,52 @@
...
+
+import (
+	"archive/zip"
+	"bytes"
+	_ "embed"
+	"io"
+)
+
+//go:embed test_metadata.zip
+var testMetaData []byte
```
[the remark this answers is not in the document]
qdm12: Or even better rename it to something more precise. Careful to leave `//go:embed` above it if you go the commnent route.

dot/rpc/modules/system_test.go
qdm12: Empty string error doesn't necessarily mean nil error, so that combines a `wantErr bool` + `errMsg string` in one test field.

qdm12 (approved):
Looking pretty good :+1: 

Just make sure to go through hidden collapsed unresolved conversations in the PR conversation :wink:

jimjbrettj:
@kishansagathiya will make these changes, was planning to resolve linting issues after merging into my other PR (since this isn't into dev) so I can handle them all at once without creating conflicts

jimjbrettj:
> @kishansagathiya will make these changes, was planning to resolve linting issues after merging into my other PR (since this isn't into dev) so I can handle them all at once without creating conflicts

actually I will do this here you are right!

dot/rpc/modules/system.go
```diff
@@ -258,8 +257,8 @@ func (sm *SystemModule) LocalListenAddresses(r *http.Request, req *EmptyRequest,
 	return nil
 }
 
-// LocalPeerId Returns the base58-encoded PeerId fo the node.
-func (sm *SystemModule) LocalPeerId(r *http.Request, req *EmptyRequest, res *string) error {
+// LocalPeerID Returns the base58-encoded PeerId fo the node.
+func (sm *SystemModule) LocalPeerID(r *http.Request, req *EmptyRequest, res *string) error {
```
EclesioMeloJunior, line 261: this should stay as `LocalPeerId` to represent the RPC call `system_localPeerId` that is the name placed in the RPC specs -> https://polkadot.js.org/docs/substrate/rpc/#system

EclesioMeloJunior (changes requested):

dot/rpc/http_test.go
```diff
@@ -245,7 +245,7 @@ func TestRPCExternalEnable_UnsafeExternalNotEnabled(t *testing.T) {
 
 	safeData := []byte(fmt.Sprintf(
 		`{"jsonrpc":"2.0","method":"%s","params":[],"id":2}`,
-		"system_localPeerId"))
+		"system_localPeerID"))
```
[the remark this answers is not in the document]
jimjbrettj: The linter made me do it, so I will change and no lint it. Thank you for this catch

dot/rpc/modules/system.go
jimjbrettj: done!

EclesioMeloJunior (approved):
nice job!!

Issue #1823, opened 2021-10-01
Continuous Integration Gossamer Devnet


timwu20:
## Issue summary

- Create new cluster of containerised nodes that run Gossamer from specific branch
- Integrate prometheus to datadog metrics collection
- Integrate into Github CI on specific `gossamer-devnet` branch
- Create new datadog dashboard and alerting
- Re-use existing instances or sunset existing nodes.

## Other information and links

-

PR #2008, opened 2021-11-09
feat(devnet): Local Gossamer Devnet

timwu20:
## Changes

- docker-compose.yml  and associated docker images to setup a local devnet in a containerised environment.
- prometheus metrics are automatically piped to Datadog
- README.md for setup and running instructions

## Tests

```
go test devnet/
```

## Issues

- #1823 

## Primary Reviewer

- @noot


noot:
@timwu20  looks epic, will test this out tomorrow!! do we want to add this to CI? (or wait until it's merged?)

devnet/README.md
```diff
@@ -0,0 +1,71 @@
...
+- `docker-compose.yml`
+
+To symlink to the repository root
+```
+# cd into your Gossamer root
+cd ~/dev/gossamer
+
+ln -s devnet/alice.Dockerfile alice.Dockerfile
+ln -s devnet/bob.Dockerfile bob.Dockerfile
+ln -s docker-compose.yml docker-compose.yml
```
kishansagathiya, line 23: This doesn't seem right (creating link at the same path). Also, root already has a `docker-compose.yml`. Is this a mistake or am I missing something?

timwu20: Good catch.  Updated the README.

timwu20:
> @timwu20 looks epic, will test this out tomorrow!! do we want to add this to CI? (or wait until it's merged?)

I'm still working on the actual ECS integration using the docker-compose ECS plugin.  I'll integrate the CI then.  I thought there would be value for the team to be able to run it locally first.  The `docker-compose.yml` file with have more aws specific annotations and changes for it to deploy to ECS.

devnet/alice.Dockerfile
```diff
@@ -0,0 +1,33 @@
+FROM golang:1.17
+
+ARG DD_API_KEY=somekey
+
+ENV GO111MODULE=on
```
qdm12, line 5: I think this is enabled by default now

```suggestion
```

devnet/alice.Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+
+WORKDIR /gossamer
+
+COPY . . 
+
+RUN go run devnet/cmd/update-dd-agent-confd/main.go -n="gossamer.local.devnet" -t="key:alice" > /tmp/conf.yaml
+RUN mv /tmp/conf.yaml /etc/datadog-agent/conf.d/openmetrics.d/conf.yaml
+RUN service datadog-agent start
+
+RUN go get ./...
```
qdm12, line 20: You should cache Go dependencies with something similar to

https://github.com/ChainSafe/gossamer/blob/development/Dockerfile#L27-L35

devnet/alice.Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+RUN rm chain/gssmr/genesis-spec.json
+RUN cp devnet/chain/gssmr/genesis-spec.json chain/gssmr/genesis-spec.json
+
+RUN gossamer --key alice init
+
+# use a hardcoded key for alice, so we can determine what the peerID is for subsequent nodes
+RUN cp devnet/alice.node.key ~/.gossamer/gssmr/node.key
+
+ENTRYPOINT service datadog-agent restart && gossamer --key=alice --babe-lead --publish-metrics --rpc --rpc-external=true --pubip=10.5.0.2
+EXPOSE 7001 8545 8546 8540 9876
```
qdm12, line 33: Since `EXPOSE` is for documentation only, let's precise the protocol as well :wink: 

```suggestion
EXPOSE 7001/tcp 8545/tcp 8546/tcp 8540/tcp 9876/tcp

```

noot:
@timwu20 readme is a bit confusing, for this step:

> To symlink to the repository root
```
# cd into your Gossamer root
cd ~/dev/gossamer
```


do you mean the `--basepath`? so I need to copy the docker files to the basepath?

devnet/README.md
```diff
@@ -0,0 +1,71 @@
...
+- `docker-compose.yml`
+
+To symlink to the repository root
+```
+# cd into your Gossamer root
+cd ~/dev/gossamer
+
+ln -s devnet/alice.Dockerfile alice.Dockerfile
+ln -s devnet/bob.Dockerfile bob.Dockerfile
+ln -s devnet/docker-compose.yml docker-compose.yml
```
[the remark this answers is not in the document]
timwu20: right now alice.Dockerfile and bob.Dockerfile live in the `/devnet`, and I can't get the image to build without the codebase in the same directory.  Do you have any methods/suggestions around this?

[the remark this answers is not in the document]
timwu20: nevermind I got this working, and will remove this entire section.

devnet/alice.Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+# use modified genesis-spec.json with only 3 authority nodes
+RUN rm chain/gssmr/genesis-spec.json
+RUN cp devnet/chain/gssmr/genesis-spec.json chain/gssmr/genesis-spec.json
+
+RUN gossamer --key alice init
+
+# use a hardcoded key for alice, so we can determine what the peerID is for subsequent nodes
+RUN cp devnet/alice.node.key ~/.gossamer/gssmr/node.key
+
+ENTRYPOINT service datadog-agent restart && gossamer --key=alice --babe-lead --publish-metrics --rpc --rpc-external=true --pubip=10.5.0.2
```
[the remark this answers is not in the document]
timwu20: It is started already in the build, but I found it doesn't work without kicking it like this on container start.

cmd/gossamer/config.go
```diff
@@ -678,6 +678,11 @@ func setDotNetworkConfig(ctx *cli.Context, tomlCfg ctoml.NetworkConfig, cfg *dot
 		cfg.NoMDNS = true
 	}
 
+	// check --pubip flag and update node configuration
+	if pubip := ctx.GlobalString(PublicIPFlag.Name); pubip != "" {
+		cfg.PublicIP = pubip
```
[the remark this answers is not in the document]
timwu20: This gets parsed in `newHost`

devnet/README.md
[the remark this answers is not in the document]
qdm12: Nice :+1: Yes you can use the build context which can be different than where the Dockerfile is located.

devnet/alice.Dockerfile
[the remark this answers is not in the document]
qdm12: Maybe stop it in the build then, as soon as it gets auto-started? It feels kinda wrong to have something running in a 'frozen' image :sweat_smile:

timwu20:
> @timwu20 readme is a bit confusing, for this step:
> 
> > To symlink to the repository root
> 
> ```
> # cd into your Gossamer root
> cd ~/dev/gossamer
> ```
> 
> do you mean the `--basepath`? so I need to copy the docker files to the basepath?

I'm getting rid of this whole step.  I got it working by just running `docker-compose up` in the devnet folder.

devnet/README.md
```diff
@@ -0,0 +1,71 @@
...
+
+# destroys the devnet
+docker-compose down
+```
+
+> **_NOTE:_**  The devnet is not stateful, so subsequent runs will start from the genesis block.
+
+## Prometheus Datadog Integration
+
+All Prometheus metrics from the nodes are piped to Datadog.  You can setup your own dashboard and add additional tags by modifying the Dockerfiles.  Currently the metrics are prefixed with `gossamer.local.devnet` and are tagged with a `key` tag for `alice`, `bob`, and `charlie`.
```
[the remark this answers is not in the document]
timwu20: In Datadog they call them tags, and that's how they are defined the datadag confd.yml.

devnet/alice.Dockerfile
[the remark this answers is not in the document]
timwu20: I'll try that.

devnet/docker-compose.yml
```diff
@@ -0,0 +1,65 @@
+services: 
+  alice:
+    build:
+      context: ./..
+      dockerfile: devnet/alice.Dockerfile
+      args:
+        DD_API_KEY: ${DD_API_KEY}
+        pubip: 10.5.0.2
```
qdm12, line 8: Can't the program auto-detect it's LAN IP address?

devnet/docker-compose.yml
```diff
@@ -0,0 +1,65 @@
...
+    ports:
+      - "7001"
+      - "8545"
+      - "8546"
+      - "8540"
+      - "9876"
```
qdm12, line 14: nit no need for double quotes :wink: 

```suggestion
      - 7001
      - 8545
      - 8546
      - 8540
      - 9876
```

devnet/docker-compose.yml
```diff
@@ -0,0 +1,65 @@
...
+    networks:
+      vpcbr:
+        ipv4_address: 10.5.0.2
+
+  bob:
+    build:
+      context: ./..
+      dockerfile: devnet/bob.Dockerfile
+      args:
+        DD_API_KEY: ${DD_API_KEY}
```
qdm12, line 24: nit maybe/ideally we should load that variable from a .gitignore'd `.env` file since exporting in your shell will leave a trace in your shell history and this is an API key. Or not if we don't care, that works too :smile_cat:

devnet/docker-compose.yml
```diff
@@ -0,0 +1,65 @@
...
+      - "7001"
+      - "8545"
+      - "8546"
+      - "8540"
+      - "9876"
+    networks:
+      vpcbr:
+        ipv4_address: 10.5.0.3
+    depends_on:
+      - "alice"
```
qdm12, line 37: When do we know when Alice is ready? Maybe should we have a shell/Go entrypoint to wait for Alice to be ready for bob and charlie? Or will e.g. `bob` fail if `alice` is not up yet?

dot/network/host.go
```diff
@@ -65,15 +63,27 @@ func newHost(ctx context.Context, cfg *Config) (*host, error) {
...
 	var externalAddr ma.Multiaddr
-	ip, err := pubip.Get()
-	if err != nil {
-		logger.Errorf("failed to get public IP: %s", err)
-	} else {
-		logger.Debugf("got public IP %s", ip)
+	if cfg.PublicIP != "" {
+		ip := net.ParseIP(cfg.PublicIP)
+		if ip == nil {
+			return nil, fmt.Errorf("invalid public ip: %s", cfg.PublicIP)
```
qdm12, line 69: Define an error at global scope e.g. `var ErrPublicIPIsNotValid = errors.New("...")` and have this as

```suggestion
			return nil, fmt.Errorf("%w: %s", ErrPublicIPIsNotValid, cfg.PublicIP)
```

dot/network/host_test.go
```diff
@@ -42,6 +40,37 @@ func TestExternalAddrs(t *testing.T) {
...
+		NoBootstrap: true,
+		NoMDNS:      true,
+	}
+
+	node := createTestService(t, config)
+
+	addrInfo := node.host.addrInfo()
+	privateIPs := ma.NewFilters()
+	for _, cidr := range privateCIDRs {
+		_, ipnet, err := net.ParseCIDR(cidr) //nolint
```
qdm12, line 57: Please precise the linter you want to silence with `//nolint:lintername`

dot/network/host_test.go
```diff
@@ -42,6 +40,37 @@ func TestExternalAddrs(t *testing.T) {
...
+
+	for i, addr := range addrInfo.Addrs {
+		switch i {
+		case len(addrInfo.Addrs) - 1:
+			// would be blocked by privateIPs, but this address injected from Config.PublicIP
+			require.True(t, privateIPs.AddrBlocked(addr))
+		default:
+			require.False(t, privateIPs.AddrBlocked(addr))
+		}
+
```
qdm12, line 70: nit shorter, and clarified comment

```suggestion
		if i == len(addrInfo.Addrs) - 1 {
			// the last address is the address injected from Config.PublicIP so it's not blocked.
			require.True(t, privateIPs.AddrBlocked(addr))
			continue
		}
		require.False(t, privateIPs.AddrBlocked(addr))
```

go.mod
```diff
@@ -44,6 +44,11 @@ require (
 	google.golang.org/protobuf v1.27.1
 )
 
+require (
+	github.com/jessevdk/go-flags v1.4.0
+	gopkg.in/yaml.v2 v2.4.0
+)
```
qdm12, line 50: Can you merge this with the block above?

devnet/README.md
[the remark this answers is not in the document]
qdm12: Ah ok, Prometheus calls them labels that's why I was wondering. Maybe add `(Prometheus "labels")` after `additional tags`

noot:
@timwu20 I'm having an issue running the commands:
```
r$ docker-compose up --abort-on-exit --build 
Builds, (re)creates, starts, and attaches to containers for a service.

Unless they are already running, this command also starts any linked services.

The `docker-compose up` command aggregates the output of each container. When
the command exits, all containers are stopped. Running `docker-compose up -d`
starts the containers in the background and leaves them running.

If there are existing containers for a service, and the service's configuration
or image was changed after the container's creation, `docker-compose up` picks
up the changes by stopping and recreating the containers (preserving mounted
volumes). To prevent Compose from picking up changes, use the `--no-recreate`
flag.

If you want to force Compose to stop and recreate all containers, use the
`--force-recreate` flag.
```

seems like the command is not working for me? :(

devnet/alice.Dockerfile
```diff
@@ -0,0 +1,30 @@
+FROM golang:1.17
+
+ARG DD_API_KEY=somekey
+ENV DD_API_KEY=${DD_API_KEY}
+
+RUN DD_AGENT_MAJOR_VERSION=7 DD_INSTALL_ONLY=true DD_SITE="datadoghq.com" bash -c "$(curl -L https://s3.amazonaws.com/dd-agent/scripts/install_script.sh)"
```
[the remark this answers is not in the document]
timwu20: this was essentially copy pasted from datadog instructions.  The only reason I made DD_API_KEY an ARG cause it's user dependant.

devnet/alice.Dockerfile
```diff
@@ -0,0 +1,30 @@
+FROM golang:1.17
+
+ARG DD_API_KEY=somekey
```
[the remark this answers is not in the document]
timwu20: just at build.

devnet/cmd/update-dd-agent-confd/main.go
```diff
@@ -0,0 +1,545 @@
...
+		log.Panicf("%v", err)
+	}
+	fmt.Printf("%s", yml)
+}
+
+func marshalYAML(opts options) (yml []byte, err error) {
+	var c conf
+	err = yaml.Unmarshal([]byte(confYAML), &c)
+	if err != nil {
+		return
```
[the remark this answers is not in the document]
timwu20: I actually really like naked returns.  Check out the scale package 😛

devnet/cmd/update-dd-agent-confd/main.go
```diff
@@ -0,0 +1,545 @@
...
+	HealthServiceCheck bool     `yaml:"health_service_check"`
+	Tags               []string `yaml:"tags,omitempty"`
+}
+
+type conf struct {
+	InitConfig struct{}   `yaml:"init_config"`
+	Instances  []instance `yaml:"instances"`
+}
+
+var confYAML = `
```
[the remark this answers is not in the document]
timwu20: I've used the embed package before and found it a little cumbersome. I'll try it again.

devnet/cmd/update-dd-agent-confd/main_test.go
```diff
@@ -0,0 +1,56 @@
...
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			gotYml, err := marshalYAML(tt.args.opts)
+			if (err != nil) != tt.wantErr {
+				t.Errorf("marshalYAML() error = %v, wantErr %v", err, tt.wantErr)
+				return
+			}
+			if !reflect.DeepEqual(gotYml, tt.wantYml) {
+				t.Errorf("marshalYAML() = %s, want %s", gotYml, tt.wantYml)
+			}
```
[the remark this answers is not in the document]
timwu20: this was codegen'd from vscode.

devnet/docker-compose.yml
timwu20: We use a library called `pubip` which requests from the internet from a bunch of those "what's my ip" services.  The `pubip` key I introduced forces the node to broadcast from the specified address.  I had some code that scanned all network interfaces, but if you have multiple addresses which one would you pick?

devnet/docker-compose.yml
timwu20: I don't think it matters that much for people working locally on their computer.  I need to ensure that we don't broadcast it on the github actions logs when I deploy to ECS.

devnet/docker-compose.yml
timwu20: In @noot PR for the `--babe-lead` flag, she changed the behaviour for all other non-lead auth nodes to wait for the first block.  So we still wait for alice to start up first, but it's not completely necessary.

dot/network/host.go
timwu20: The other errors for the host are produced with `fmt.Errorf`.  I don't mind doing this, but we should be changing all the errors for the node.

timwu20:
> @timwu20 I'm having an issue running the commands:
> 
> ```
> r$ docker-compose up --abort-on-exit --build 
> Builds, (re)creates, starts, and attaches to containers for a service.
> 
> Unless they are already running, this command also starts any linked services.
> 
> The `docker-compose up` command aggregates the output of each container. When
> the command exits, all containers are stopped. Running `docker-compose up -d`
> starts the containers in the background and leaves them running.
> 
> If there are existing containers for a service, and the service's configuration
> or image was changed after the container's creation, `docker-compose up` picks
> up the changes by stopping and recreating the containers (preserving mounted
> volumes). To prevent Compose from picking up changes, use the `--no-recreate`
> flag.
> 
> If you want to force Compose to stop and recreate all containers, use the
> `--force-recreate` flag.
> ```
> 
> seems like the command is not working for me? :(

Seems odd.  What version of Docker are you running? What happens if you omit the `--build` flag?

devnet/alice.Dockerfile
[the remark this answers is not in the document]
qdm12: Ok let's not have `ENV DD_API_KEY=${DD_API_KEY}` then. `ENV` is mostly to persist it in the final image for runtime usage.

devnet/docker-compose.yml
qdm12: Awesome, let's leave those `depends_on` for documentation really then :wink:

dot/network/host.go
qdm12: Not really, just do it bit by bit I think, especially since it's new code. I don't believe we should go all in and refactor all errors returned :sweat_smile:

go.mod
timwu20: this is what `go mod tidy` does automatically and I'm not sure why.  I'll try it and see if it complains.

devnet/README.md
```diff
@@ -0,0 +1,54 @@
...
+
+The Datadog API key must be an environment variable on your own machine, which Docker Compose will pick up and inject when building the node images.
+
+```
+export $DD_API_KEY=YourKey
+```
+
+## Files
+
+### Dockefiles
```
noot, line 40:
```suggestion
### Dockerfiles
```

noot (approved):
worked for me - great work!

devnet/cmd/update-dd-agent-confd/main_test.go
```diff
@@ -0,0 +1,56 @@
...
+		wantYml []byte
+		wantErr bool
+	}{
+		{
+			args:    args{opts: options{}},
+			wantYml: expected,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
```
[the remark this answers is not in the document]
timwu20: When I add a test case, that fails, both tests show failed for some reason.  I'm going to leave out the `t.Parallel()`.

```
=== RUN   Test_marshalYAML
=== RUN   Test_marshalYAML/zero_case_options
=== PAUSE Test_marshalYAML/zero_case_options
=== RUN   Test_marshalYAML/options_with_ns_and_tags
=== PAUSE Test_marshalYAML/options_with_ns_and_tags
=== CONT  Test_marshalYAML/zero_case_options
=== CONT  Test_marshalYAML/options_with_ns_and_tags
=== CONT  Test_marshalYAML/zero_case_options
    /Users/timothywu/dev/gossamer/devnet/cmd/update-dd-agent-confd/main_test.go:82: marshalYAML() diff =   bytes.Join({
          	"init_config: {}\ninstances:\n- prometheus_url: http://127.0.0.1:98",
          	"76/metrics\n  namespace: SomeNamespace\n  metrics:\n  - gossamer",
        + 	"1",
          	"_*\n  - network_*\n  - service_*\n  - system_*\n  health_service_che",
          	"ck: true\n  tags:\n  - some\n  - tags\n",
          }, "")
=== CONT  Test_marshalYAML/options_with_ns_and_tags
    /Users/timothywu/dev/gossamer/devnet/cmd/update-dd-agent-confd/main_test.go:82: marshalYAML() diff =   bytes.Join({
          	"init_config: {}\ninstances:\n- prometheus_url: http://127.0.0.1:98",
          	"76/metrics\n  namespace: SomeNamespace\n  metrics:\n  - gossamer",
        + 	"1",
          	"_*\n  - network_*\n  - service_*\n  - system_*\n  health_service_che",
          	"ck: true\n  tags:\n  - some\n  - tags\n",
          }, "")
--- FAIL: Test_marshalYAML (0.00s)
    --- FAIL: Test_marshalYAML/zero_case_options (0.00s)
    --- FAIL: Test_marshalYAML/options_with_ns_and_tags (0.00s)
```

dot/network/host.go
timwu20: `host` is a private struct.  Not sure this is really required, since the only thing calling `newHost` is `NewService` and even that doesn't have package errors at this point.

devnet/docker-compose.yml
qdm12: For LAN ip addresses, you can `go get github.com/vishvananda/netlink` and use

```go
func Test_findLANIP(t *testing.T) {
	ip, err := findLANIP()
	require.NoError(t, err)
	t.Log(ip)
}

var (
	ErrNoLinkFound            = errors.New("no link was found")
	ErrDefaultGatewayNotFound = errors.New("default gateway not found")
)

func findLANIP() (ip net.IP, err error) {
	links, err := netlink.LinkList()
	if err != nil {
		return nil, fmt.Errorf("cannot list links: %w", err)
	} else if len(links) == 0 {
		return nil, ErrNoLinkFound
	}

	for _, link := range links {
		routes, err := netlink.RouteList(link, netlink.FAMILY_ALL)
		if err != nil {
			return nil, fmt.Errorf(
				"cannot list routes for link %d: %w",
				link.Attrs().Index, err)
		}

		for _, route := range routes {
			if route.Dst == nil {
				return route.Gw, nil
			}
		}
	}

	return nil, ErrDefaultGatewayNotFound
}
```

Although that's limited to unix systems and introduces an indirect dependency on syscall (so for windows we would need different build tags etc.)

devnet/docker-compose.yml
qdm12: I think you can also use `secrets:` in docker-compose.yml, if that may be of use. [LINK](https://docs.docker.com/compose/compose-file/compose-file-v3/#secrets)

devnet/cmd/update-dd-agent-confd/main.go
```diff
@@ -0,0 +1,61 @@
...
+
+package main
+
+import (
+	"fmt"
+	"log"
+
+	_ "embed"
+
+	"github.com/jessevdk/go-flags"
```
qdm12, line 12: Can't you use the stdlib's `flag` package for this?

devnet/bob.Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+RUN test -n "$pubip"
+
+ENV key=${key}
+ENV pubip=${pubip}
+
+RUN gossamer --key=${key} init
+
+RUN go run devnet/cmd/update-dd-agent-confd/main.go -n=gossamer.local.devnet -t=key:${key} > /etc/datadog-agent/conf.d/openmetrics.d/conf.yaml
+
+ENTRYPOINT service datadog-agent start && gossamer --key=${key} --bootnodes=/ip4/10.5.0.2/tcp/7001/p2p/12D3KooWMER5iow67nScpWeVqEiRRx59PJ3xMMAYPTACYPRQbbWU --publish-metrics --rpc --pubip=${pubip}
```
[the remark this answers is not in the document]
timwu20: I don't see a case where we do want to adjust the bootnodes for the `gssmr` devnet.

timwu20:
I don't know why I can't reply to your feedback about the lan IP stuff.  Since this will only ever be used by the devnet, I'd rather not bring in extra code when the user can just force the node to broadcast the intended address.

devnet/cmd/update-dd-agent-confd/main.go
timwu20: I prefer `go-flags` and just using a struct.  It has served me well for previous projects.

dot/network/host.go
```diff
@@ -65,15 +63,27 @@ func newHost(ctx context.Context, cfg *Config) (*host, error) {
...
+		}
+		logger.Debugf("using config PublicIP: %s", ip)
 		externalAddr, err = ma.NewMultiaddr(fmt.Sprintf("/ip4/%s/tcp/%d", ip, cfg.Port))
 		if err != nil {
 			return nil, err
 		}
+	} else {
+		ip, err := pubip.Get()
+		if err != nil {
+			logger.Errorf("failed to get public IP error: %v", err)
```
kishansagathiya, line 79: should we return error here?

If we are here, that means we are failing to set `externalAddr`. `externalAddr` is getting used in the code after this, but not getting set. Looks like we should return the error here.

dot/network/host.go
```diff
@@ -65,15 +63,27 @@ func newHost(ctx context.Context, cfg *Config) (*host, error) {
...
+		} else {
+			logger.Debugf("got public IP", "IP", ip)
+			externalAddr, err = ma.NewMultiaddr(fmt.Sprintf("/ip4/%s/tcp/%d", ip, cfg.Port))
+			if err != nil {
+				return nil, err
+			}
+		}
 	}
```
kishansagathiya, line 87: We could use if, else to just set ip and shift some common code out side those blocks

```go
	var ip net.IP
	if cfg.PublicIP != "" {
		ip = net.ParseIP(cfg.PublicIP)
		if ip == nil {
			return nil, fmt.Errorf("invalid public ip: %s", cfg.PublicIP)
		}
		logger.Debugf("using config PublicIP: %s", ip)

	} else {
		ip, err = pubip.Get()
		if err != nil {
			return nil, fmt.Errorf("failed to get public IP error: %v", err)
		}
		logger.Debugf("got public IP", "IP", ip)
	}

	externalAddr, err := ma.NewMultiaddr(fmt.Sprintf("/ip4/%s/tcp/%d", ip, cfg.Port))
	if err != nil {
		return nil, err
	}
```
(Looks like, code suggestions are not supported for more that a few lines)

devnet/bob.Dockerfile
```diff
@@ -1,33 +1,32 @@
...
 COPY . .
 
-RUN go run devnet/cmd/update-dd-agent-confd/main.go -n=gossamer.local.devnet -t=key:${key} > /etc/datadog-agent/conf.d/openmetrics.d/conf.yaml
-RUN service datadog-agent start
-
-# RUN go get ./...
 RUN go install github.com/ChainSafe/gossamer/cmd/gossamer
 
 # use modified genesis-spec.json with only 3 authority nodes
 RUN cp -f devnet/chain/gssmr/genesis-spec.json chain/gssmr/genesis-spec.json
```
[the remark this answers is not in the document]
qdm12: Yes and also we should start and stop the service here, no point letting it run I think

dot/network/host.go
timwu20: This is how the code was.  In the event we can't get a public IP via `pubip` I would assume nothing gets broadcasted.  What do you think @noot?

devnet/bob.Dockerfile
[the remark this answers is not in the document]
timwu20: I modified this in my other branch to not start it here.  I'll add it to this branch.

dot/network/host.go
timwu20: I'm adding another flag for `cfg.PublicDNS` so this is changing in the next PR, to have an `else if` case where the multiaddr is no longer `"/ip4/%s.."`, but rather `"/dns/%s..."`.

devnet/bob.Dockerfile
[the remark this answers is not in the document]
timwu20: in terms of readability, does it make sense to append multiple RUN statements together?

devnet/bob.Dockerfile
```diff
@@ -1,33 +1,32 @@
...
+ARG pubip
+RUN test -n "$pubip"
+
+ENV key=${key}
+ENV pubip=${pubip}
+
 RUN gossamer --key=${key} init
 
-ENTRYPOINT service datadog-agent restart && gossamer --key=${key} --bootnodes=/ip4/10.5.0.2/tcp/7001/p2p/12D3KooWMER5iow67nScpWeVqEiRRx59PJ3xMMAYPTACYPRQbbWU --publish-metrics --rpc --pubip=${pubip}
+RUN go run devnet/cmd/update-dd-agent-confd/main.go -n=gossamer.local.devnet -t=key:${key} > /etc/datadog-agent/conf.d/openmetrics.d/conf.yaml
```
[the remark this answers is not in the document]
timwu20: Move them up where? If I move the `key` and `pubip` args/env above the `RUN go install` doesn't that break caching for subsequent builds?

devnet/bob.Dockerfile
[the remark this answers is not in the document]
kishansagathiya: I don't think readability would worsen if we append statements. You could also indent it to make to more readable by writing each command on different lines using `\\`

devnet/bob.Dockerfile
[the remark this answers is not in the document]
kishansagathiya: I don't have a good understanding of how caching works in docker, will read up and comment again! I will resolve this for now and un-resolve if I find any issues.

devnet/bob.Dockerfile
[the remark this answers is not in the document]
timwu20: I concatenated the RUN lines together and it made no significant difference in build time.

kishansagathiya (approved):

devnet/bob.Dockerfile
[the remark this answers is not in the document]
qdm12: At least the IP address `10.5.0.2` could be changed if we change the docker network right?

qdm12:
> I don't know why I can't reply to your feedback about the lan IP stuff.

It's because it's replies to conversations above (probably hidden since they are many conversations) :wink:

dot/network/host.go
noot:
```
		libp2p.AddrsFactory(func(as []ma.Multiaddr) []ma.Multiaddr {
			addrs := []ma.Multiaddr{}
			for _, addr := range as {
				if !privateIPs.AddrBlocked(addr) {
					addrs = append(addrs, addr)
				}
			}
			if externalAddr == nil {
				return addrs
			}

			return append(addrs, externalAddr)
		}),
```

other addresses can still be advertised even if `pubip.Get` fails, so I think that's why I didn't make it return an error

devnet/bob.Dockerfile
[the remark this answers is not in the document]
timwu20: In the next PR I do actually have to change the bootnode ports, so port will be configurable on `bob.Dockerfile`.  I'm trying to hardcode the configuration for `alice.Dockerfile` since that is the lead node.  I think being explicit in this case  shows that everything is dependant on alice.

Issue #1973, opened 2021-11-03
Investigate codebase for memory problems


qdm12:
This is more of low hanging fruits that can be worth checking/fixing before investing more time in profiling and metrics, to fix the current memory issue we have (2021-11-03).

- [ ] Check usage of `bytes.Buffer`
- [ ] Check usage of pools
- [ ] Check usage of `copy`
- [ ] Check buffered channels (using regex `make\(chan .+, [1-9].+\)`) 
- [ ] Check DeepCopy functions (search for `deepcopy` without case sensitivity)
- [x] Check usage of `sizedBufferPool` - found a problem, fixed in #1963, although not a reason for OOM
- [ ] Check for arrays returned (copy, it's not a slice), use regex ` \[[a-zA-Z0-9]+\](byte|int|uint|uint32|uint64|int32|int64|string)+?`
- [ ] Check packages
  - [ ] `dot/network`
  - [ ] `lib/trie`
  - [ ] `dot/sync`

qdm12:
The current state as of today is:

- the trie works fine and does not leak (probably thanks to #2929 and #2919)
- Node running the Kusama chain does not cleanup tries in the in-memory map of tries after syncing 1.5M blocks
- Node running the Polkadot chain seems to use ~10MB per day of sync (= ~1M blocks) so that should be fine
- On the polkadot chain, leaks now come from the badgerdb implementation and the hashToRuntime map (see #2767)

PR #2009, opened 2021-11-09
fix(trie): memory leak fix in `lib/trie`

qdm12:
## Changes

- Buffer and pool usage fixed and improved 
  - Fix buffers not put back in pool
  - Write to buffer passed as arguments
  - Decouple pools for encoding, digests and hashers
- Improve `sync.Pool` usage generally
- Improve parallel encoding of branches and remove dependency on `sync/x`
- Do not copy when not needed

## Tests

```
go test -race github.com/ChainSafe/gossamer/lib/trie
```

## Issues

- #1973

### DIFF
--- a/lib/trie/hash.go
+++ b/lib/trie/hash.go
@@ -240,22 +240,36 @@ func encodeChild(child node, buffer *bytes.Buffer) (err error) {
 func encodeLeaf(l *leaf, buffer *bytes.Buffer) (err error) {
 	if !l.dirty && l.encoding != nil {
 		_, err = buffer.Write(l.encoding)
-		return err
+		if err != nil {
+			return fmt.Errorf("cannot write stored encoding to buffer: %w", err)
+		}
+		return nil
 	}
 
 	encoding, err := l.header()
 	if err != nil {
-		return err
+		return fmt.Errorf("cannot encode header: %w", err)
 	}
-	_, _ = buffer.Write(encoding)
 
-	_, _ = buffer.Write(nibblesToKeyLE(l.key))
+	_, err = buffer.Write(encoding)
+	if err != nil {
+		return fmt.Errorf("cannot write encoded header to buffer: %w", err)
+	}
+
+	_, err = buffer.Write(nibblesToKeyLE(l.key))
+	if err != nil {
+		return fmt.Errorf("cannot write LE key to buffer: %w", err)
+	}
 
 	bytes, err := scale.Marshal(l.value) // TODO scale encoder to write to buffer
 	if err != nil {
 		return err
 	}
 
-	_, _ = buffer.Write(bytes)
+	_, err = buffer.Write(bytes)
+	if err != nil {
+		return fmt.Errorf("cannot write scale encoded value to buffer: %w", err)
+	}
+
 	return nil
 }
--- a/lib/trie/hash.go
+++ b/lib/trie/hash.go
@@ -156,10 +156,10 @@ func encodeChildsInParallel(children [16]node, buffer *bytes.Buffer) (err error)
 	type result struct {
 		index  int
 		buffer *bytes.Buffer
+		err    error
 	}
 
 	resultsCh := make(chan result)
-	errorCh := make(chan error)
 
 	for i, child := range children {
 		go func(index int, child node) {
@@ -169,14 +169,11 @@ func encodeChildsInParallel(children [16]node, buffer *bytes.Buffer) (err error)
 			// data in the select block below.
 
 			err := encodeChild(child, buffer)
-			if err != nil {
-				errorCh <- err
-				return
-			}
 
 			resultsCh <- result{
 				index:  index,
 				buffer: buffer,
+				err:    err,
 			}
 		}(i, child)
 	}
@@ -184,28 +181,26 @@ func encodeChildsInParallel(children [16]node, buffer *bytes.Buffer) (err error)
 	currentIndex := 0
 	resultBuffers := make([]*bytes.Buffer, len(children))
 	for range children {
-		select {
-		case result := <-resultsCh:
-			resultBuffers[result.index] = result.buffer
+		result := <-resultsCh
+		if result.err != nil && err == nil { // only set the first error we get
+			err = result.err
+		}
 
-			// write as many completed buffers to the result buffer.
-			for currentIndex < len(children) &&
-				resultBuffers[currentIndex] != nil {
-				// note buffer.Write copies the byte slice given as argument
-				_, writeErr := buffer.Write(resultBuffers[currentIndex].Bytes())
-				if writeErr != nil && err == nil {
-					err = writeErr
-				}
+		resultBuffers[result.index] = result.buffer
 
-				encodingBufferPool.Put(resultBuffers[currentIndex])
-				resultBuffers[currentIndex] = nil
-
-				currentIndex++
+		// write as many completed buffers to the result buffer.
+		for currentIndex < len(children) &&
+			resultBuffers[currentIndex] != nil {
+			// note buffer.Write copies the byte slice given as argument
+			_, writeErr := buffer.Write(resultBuffers[currentIndex].Bytes())
+			if writeErr != nil && err == nil {
+				err = writeErr
 			}
-		case newErr := <-errorCh:
-			if err == nil { // only set the first error we get
-				err = newErr
-			}
+
+			encodingBufferPool.Put(resultBuffers[currentIndex])
+			resultBuffers[currentIndex] = nil
+
+			currentIndex++
 		}
 	}
 
--- a/lib/trie/hash.go
+++ b/lib/trie/hash.go
@@ -5,195 +5,262 @@ package trie
 
 import (
 	"bytes"
-	"context"
+	"errors"
+	"fmt"
 	"hash"
 	"sync"
 
 	"github.com/ChainSafe/gossamer/lib/common"
 	"github.com/ChainSafe/gossamer/pkg/scale"
 	"golang.org/x/crypto/blake2b"
-	"golang.org/x/sync/errgroup"
 )
 
-// Hasher is a wrapper around a hash function
-type hasher struct {
-	hash     hash.Hash
-	tmp      bytes.Buffer
-	parallel bool // Whether to use parallel threads when hashing
-}
-
-// hasherPool creates a pool of Hasher.
-var hasherPool = sync.Pool{
+var encodingBufferPool = &sync.Pool{
 	New: func() interface{} {
-		h, _ := blake2b.New256(nil)
-		var buf bytes.Buffer
-		// This allocation will be helpful for encoding keys. This is the min buffer size.
-		buf.Grow(700)
-
-		return &hasher{
-			tmp:  buf,
-			hash: h,
-		}
+		const initialBufferCapacity = 1900000 // 1.9MB, from checking capacities at runtime
+		b := make([]byte, 0, initialBufferCapacity)
+		return bytes.NewBuffer(b)
 	},
 }
 
-// NewHasher create new Hasher instance
-func newHasher(parallel bool) *hasher {
-	h := hasherPool.Get().(*hasher)
-	h.parallel = parallel
-	return h
+var digestBufferPool = &sync.Pool{
+	New: func() interface{} {
+		const bufferCapacity = 32
+		b := make([]byte, 0, bufferCapacity)
+		return bytes.NewBuffer(b)
+	},
 }
 
-func (h *hasher) returnToPool() {
-	h.tmp.Reset()
-	h.hash.Reset()
-	hasherPool.Put(h)
+var hasherPool = &sync.Pool{
+	New: func() interface{} {
+		hasher, err := blake2b.New256(nil)
+		if err != nil {
+			panic("cannot create Blake2b-256 hasher: " + err.Error())
+		}
+		return hasher
+	},
 }
 
-// Hash encodes the node and then hashes it if its encoded length is > 32 bytes
-func (h *hasher) Hash(n node) (res []byte, err error) {
-	encNode, err := h.encode(n)
+func hashNode(n node, digestBuffer *bytes.Buffer) (err error) {
+	encodingBuffer := encodingBufferPool.Get().(*bytes.Buffer)
+	encodingBuffer.Reset()
+	defer encodingBufferPool.Put(encodingBuffer)
+
+	const parallel = false
+
+	err = encodeNode(n, encodingBuffer, parallel)
 	if err != nil {
-		return nil, err
+		return fmt.Errorf("cannot encode node: %w", err)
 	}
 
 	// if length of encoded leaf is less than 32 bytes, do not hash
-	if len(encNode) < 32 {
-		return encNode, nil
+	if encodingBuffer.Len() < 32 {
+		_, err = digestBuffer.Write(encodingBuffer.Bytes())
+		return err
 	}
 
-	h.hash.Reset()
 	// otherwise, hash encoded node
-	_, err = h.hash.Write(encNode)
-	if err == nil {
-		res = h.hash.Sum(nil)
+	hasher := hasherPool.Get().(hash.Hash)
+	hasher.Reset()
+	defer hasherPool.Put(hasher)
+
+	// Note: using the sync.Pool's buffer is useful here.
+	_, err = hasher.Write(encodingBuffer.Bytes())
+	if err != nil {
+		return fmt.Errorf("cannot hash encoded node: %w", err)
 	}
 
-	return res, err
+	_, err = digestBuffer.Write(hasher.Sum(nil))
+	return err
 }
 
-// encode is the high-level function wrapping the encoding for different node types
-// encoding has the following format:
+var ErrNodeTypeUnsupported = errors.New("node type is not supported")
+
+// encodeNode writes the encoding of the node to the buffer given.
+// It is the high-level function wrapping the encoding for different
+// node types. The encoding has the following format:
 // NodeHeader | Extra partial key length | Partial Key | Value
-func (h *hasher) encode(n node) ([]byte, error) {
+func encodeNode(n node, buffer *bytes.Buffer, parallel bool) (err error) {
 	switch n := n.(type) {
 	case *branch:
-		return h.encodeBranch(n)
+		err := encodeBranch(n, buffer, parallel)
+		if err != nil {
+			return fmt.Errorf("cannot encode branch: %w", err)
+		}
+		return nil
 	case *leaf:
-		return h.encodeLeaf(n)
+		err := encodeLeaf(n, buffer)
+		if err != nil {
+			return fmt.Errorf("cannot encode leaf: %w", err)
+		}
+		n.encoding = make([]byte, buffer.Len())
+		copy(n.encoding, buffer.Bytes())
+		return nil
 	case nil:
-		return []byte{0}, nil
+		buffer.Write([]byte{0})
+		return nil
+	default:
+		return fmt.Errorf("%w: %T", ErrNodeTypeUnsupported, n)
 	}
-
-	return nil, nil
 }
 
 func encodeAndHash(n node) ([]byte, error) {
-	h := newHasher(false)
-	defer h.returnToPool()
+	buffer := digestBufferPool.Get().(*bytes.Buffer)
+	buffer.Reset()
+	defer digestBufferPool.Put(buffer)
 
-	encChild, err := h.Hash(n)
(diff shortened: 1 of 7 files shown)


lib/trie/hash_test.go
```diff
@@ -7,86 +7,69 @@ import (
...
-	h, err := hasher.Hash(n)
+
+	buffer := bytes.NewBuffer(nil)
+	const parallel = false
+	err := encodeNode(n, buffer, parallel)
+
 	if err != nil {
 		t.Errorf("did not hash leaf node: %s", err)
-	} else if h == nil {
+	} else if buffer.Len() == 0 {
```
EclesioMeloJunior, line 40: could we change to `require.NoError` and `require.Greater`?

lib/trie/hash_test.go
```diff
@@ -7,86 +7,69 @@ import (
...
-	h, err := hasher.Hash(n)
+
+	buffer := bytes.NewBuffer(nil)
+	const parallel = false
+	err := encodeNode(n, buffer, parallel)
+
 	if err != nil {
 		t.Errorf("did not hash branch node: %s", err)
-	} else if h == nil {
+	} else if buffer.Len() == 0 {
```
EclesioMeloJunior, line 55: let's change to `require` here

lib/trie/node.go
```diff
@@ -211,71 +210,91 @@ func (b *branch) setKey(key []byte) {
...
+	bufferBytes := buffer.Bytes()
+
+	l.encoding = make([]byte, len(bufferBytes))
+	copy(l.encoding, bufferBytes)
+	encoding = l.encoding // no need to copy
+
+	if len(bufferBytes) < 32 {
+		l.hash = make([]byte, len(bufferBytes))
+		copy(l.hash, bufferBytes)
+		hash = l.hash // no need to copy
```
EclesioMeloJunior, line 276: in the `(b *branch) encodeAndHash()` there is the line `hash = make([]byte, len(bufferBytes))` to alloc space on `hash` before assign `hash = l.hash`, but here this line is not present, is there any difference?

EclesioMeloJunior (changes requested):

lib/trie/hash.go
```diff
@@ -5,195 +5,262 @@ package trie
...
+	errorCh := make(chan error)
 
-		for _, v := range resBuff {
-			if v != nil {
-				h.tmp.Write(v)
+	for i, child := range children {
+		go func(index int, child node) {
+			buffer := encodingBufferPool.Get().(*bytes.Buffer)
+			buffer.Reset()
+			// buffer is put back in the pool after processing its
```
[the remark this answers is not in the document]
qdm12: Yes it is below in the for loop accumulating results [here](https://github.com/ChainSafe/gossamer/blob/0bc4a88a2271d849ddbe431989b0b11db41ada3e/lib/trie/hash.go#L200), only once its result are 'accumulated' in the main buffer.

It's a bit convoluted but that's the only way I see to do it efficiently. Feel free to suggest another way though.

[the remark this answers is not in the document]
qdm12: Totally not, thanks for noticing this! Fixed (also simplified) in b8f648ce8e46e6b1a906046fdde47b904347c5a8

lib/trie/hash_test.go
qdm12: If you don't mind I'm working on refactoring tests and existing code in another branch, to keep deltas low in this already decently large PR :smile: But definitely something to do!

qdm12: Done in https://github.com/ChainSafe/gossamer/pull/2049

lib/trie/hash_test.go
qdm12: Done in https://github.com/ChainSafe/gossamer/pull/2049

lib/trie/node.go
qdm12: Good find! That's a wasted `make` in the branch's `encodeAndHash`. We don't care about having deep copied bytes for both AFAIK.

lib/trie/hash.go
```diff
@@ -5,195 +5,262 @@ package trie
...
 	encoding, err := l.header()
-	h.tmp.Write(encoding)
 	if err != nil {
-		return nil, err
+		return err
 	}
+	_, _ = buffer.Write(encoding)
 
-	h.tmp.Write(nibblesToKeyLE(l.key))
+	_, _ = buffer.Write(nibblesToKeyLE(l.key))
```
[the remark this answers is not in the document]
qdm12: Good point, done in 423f4aa30090d80e46d19377c275792bf0a13d9b

EclesioMeloJunior (approved):

lib/trie/hash.go
```diff
@@ -5,195 +5,285 @@ package trie
...
-	h.tmp.Reset()
-	h.hash.Reset()
-	hasherPool.Put(h)
-}
+func hashNode(n node, digestBuffer *bytes.Buffer) (err error) {
+	encodingBuffer := encodingBufferPool.Get().(*bytes.Buffer)
+	encodingBuffer.Reset()
+	defer encodingBufferPool.Put(encodingBuffer)
+
+	const parallel = false
```
noot, line 49: is this always false now? can we set it to not false by default or no?

lib/trie/hash.go
```diff
@@ -5,195 +5,285 @@ package trie
...
+	if parallel {
+		return encodeChildsInParallel(b.children, buffer)
+	}
+	return encodeChildsSequentially(b.children, buffer)
+}
 
-		for _, v := range resBuff {
-			if v != nil {
-				h.tmp.Write(v)
+func encodeChildsInParallel(children [16]node, buffer *bytes.Buffer) (err error) {
```
noot, line 159:
```suggestion
func encodeChildrenInParallel(children [16]node, buffer *bytes.Buffer) (err error) {
```

lib/trie/hash.go
```diff
@@ -5,195 +5,285 @@ package trie
...
+				err:    err,
 			}
+		}(i, child)
+	}
+
+	currentIndex := 0
+	resultBuffers := make([]*bytes.Buffer, len(children))
+	for range children {
+		result := <-resultsCh
+		if result.err != nil && err == nil { // only set the first error we get
```
noot, line 189: is there ever a case `err` would not be nil here?

lib/trie/hash.go
```diff
@@ -5,195 +5,285 @@ package trie
...
+			err = result.err
```
noot, line 190: also should we just return the error if there's any err hashing/encoding the child?

lib/trie/hash.go
```diff
@@ -5,195 +5,285 @@ package trie
...
+			_, writeErr := buffer.Write(resultBuffers[currentIndex].Bytes())
+			if writeErr != nil && err == nil {
+				err = writeErr
 			}
+
+			encodingBufferPool.Put(resultBuffers[currentIndex])
+			resultBuffers[currentIndex] = nil
+
+			currentIndex++
 		}
```
noot, line 208: ok, maybe I'm just having trouble reading this code lol, but does this preserve the order that the children are written? eg. a child at index 0 should be written before one at index 1, etc

lib/trie/hash.go
```diff
@@ -5,195 +5,285 @@ package trie
...
+	}
+
+	return err
 }
 
-// encodeLeaf encodes a leaf with the encoding specified at the top of this package
-func (h *hasher) encodeLeaf(l *leaf) ([]byte, error) {
-	if !l.dirty && l.encoding != nil {
-		return l.encoding, nil
+func encodeChildsSequentially(children [16]node, buffer *bytes.Buffer) (err error) {
```
noot, line 221:
```suggestion
func encodeChildrenSequentially(children [16]node, buffer *bytes.Buffer) (err error) {
```

lib/trie/hash.go
```diff
@@ -5,195 +5,285 @@ package trie
...
+	l.encodingMu.RLock()
+	if !l.dirty && l.encoding != nil {
+		_, err = buffer.Write(l.encoding)
+		l.encodingMu.RUnlock()
+		if err != nil {
+			return fmt.Errorf("cannot write stored encoding to buffer: %w", err)
+		}
+		return nil
+	}
+	l.encodingMu.RUnlock()
```
noot, line 261:
```suggestion
	l.encodingMu.RLock()
	defer l.encodingMu.RUnlock()
	if !l.dirty && l.encoding != nil {
		_, err = buffer.Write(l.encoding)
		if err != nil {
			return fmt.Errorf("cannot write stored encoding to buffer: %w", err)
		}
		return nil
	}
```

this locking seems simpler, would it be ok?

lib/trie/node.go
```diff
@@ -83,19 +84,19 @@ func (l *leaf) setGeneration(generation uint64) {
...
 func (b *branch) copy() node {
-	b.Lock()
-	defer b.Unlock()
+	b.RLock()
+	defer b.RUnlock()
+
 	cpy := &branch{
 		key:        make([]byte, len(b.key)),
-		children:   [16]node{},
+		children:   b.children, // copy interface pointers
```
noot, line 92: if one of the copy's children changes, will that also change the child of the original node? if so seems kinda dangerous, but maybe go doesn't work like that? haha

lib/trie/node.go
```diff
@@ -211,71 +219,95 @@ func (b *branch) setKey(key []byte) {
...
-	if len(enc) < 32 {
-		b.encoding = enc
-		b.hash = enc
-		return enc, enc, nil
+	bufferBytes := buffer.Bytes()
+
+	b.encoding = make([]byte, len(bufferBytes))
+	encoding = make([]byte, len(bufferBytes))
+	copy(b.encoding, bufferBytes)
+	encoding = b.encoding // no need to copy
```
noot, line 241: is the `make` for `encoding` necessary if you're just assigning it?
```suggestion
	copy(b.encoding, bufferBytes)
	encoding := b.encoding // no need to copy
```

lib/trie/hash.go
noot: oh wait, i guess you want to unlock before the rest of the function?

lib/trie/print.go
```diff
@@ -25,38 +26,49 @@ func (t *Trie) String() string {
...
 		tree.Add(bstr)
 	default:
 		return
 	}
 }
-
-// Print prints the trie through pre-order traversal
-func (t *Trie) Print() {
-	fmt.Println(t.String())
-}
```
noot, line 62: did you move this? or does the trie already have a `String()` func somewhere else?

noot (commented):
nice work! I ran this overnight and the memory usage does still crawl up after a while, but seems improved for sure. nice work :D

lib/trie/hash.go
qdm12: I have no idea. It should work the same, but I left it as is it was for now.

Memory stack wise, I would tend to think that would end up running 100(00)s of goroutines, leading to hundreds of MBs on the stack. I would rather have it run slower for now. We could have it in the future with `cpu cores` worker goroutines instead so it would be fast and with a low memory footprint, but definitely not in this PR :wink:

lib/trie/hash.go
qdm12: Yes once it's set by the first encountered `result.err` from the encoding of one of the 16 children.

lib/trie/hash.go
qdm12: That's what it does, although we only care about the first encoding error encountered.

We can't break or return from the for loop since we don't want to return without having the 16 goroutines be done. Ideally we would have a context to cancel the processing in the unfinished goroutines to accelerate cancelation and not waste useless processing, but that would require a larger refactor! Maybe in another PR :wink:

lib/trie/hash.go
qdm12: Haha sorry I know the code is convoluted due to the use of pooled buffers and other buffers, for memory footprint.

`resultBuffers` is a slice of buffers (pointers to buffers), and we set the index in this `resultBuffers` slice using the index coming from the results channel. We then process from index 0 the buffers from `resultBuffers`, and as soon as a sequence of buffers is completed in order to release back buffers to the buffer pool asap.

lib/trie/hash.go
qdm12: Yeah I just wanted to release the lock as soon as possible, since it would have to wait for 3 encoding below it before the defer gets executed. But that might be a bit too pre-optimization, let me know I can commit your suggestion.

lib/trie/node.go
qdm12: It's an array, not a slice :wink: 

```go
arr1 := [1]int{9}
arr2 = arr1
arr2[0] = 3
fmt.Println(arr1) // prints [9] still
```

aka array are always passed by value, there is no pointer unlike slices.

lib/trie/node.go
qdm12: Definitely not, just a remain of the past :tm:

lib/trie/print.go
qdm12: Yeah it does have a `String()` method (:eyes: see the body of `Print()` :wink:). And I removed it since it's unused.

qdm12:
@noot 

> the memory usage does still crawl up after a while

After approximately how long and at how much memory usage? Mine oscillates between 9 and 10GB with

```sh
go run ./cmd/gossamer --chain polkadot --basepath ./datadir
```

I'll remove the `datadir` and try again tomorrow.

lib/trie/hash.go
qdm12: done in 99f273dc1c9e1b69d30de3995b37dd349b4769fd

lib/trie/hash.go
qdm12: done in 95abb513949f26a6bcebcae4972e5dcd66ad53d0

lib/trie/hash.go
timwu20: I'd remove `parallel` completely then.  Not sure why we have this as an option if the caller of this method can never modify it.

lib/trie/hash.go
timwu20: From my understanding of `sync.RWMutex`, it will continue to allow access to read.  If the `if` condition is true then you'll have to wait for for the buffer to write anyways.  I prefer the defer style.

lib/trie/node.go
```diff
@@ -211,71 +219,94 @@ func (b *branch) setKey(key []byte) {
 	b.key = key
 }
 
-func (b *branch) encodeAndHash() ([]byte, []byte, error) {
+func (b *branch) encodeAndHash() (encoding, hash []byte, err error) {
```
timwu20, line 222: why do we have both `encodeAndHash` functions on the package level and on the node receiver functions?

lib/trie/hash.go
noot: I kinda liked the parallel option (I think it increased the sync time, I did notice it is a bit slower on here, but would need to profile it to really make sure) but yeah either remove it or re-add it as an option I guess

lib/trie/node.go
noot: oh yes it is an array! ok cool :D

lib/trie/hash.go
noot: ok cool, I kinda understand now lol. as long as the tests pass it's good :p

noot (approved):
nice work! in terms of the memory - it appears after running for ~9h the memory is 10.7GB, so it's improved for sure but still does crawl up, I'm just running the node with `./bin/gossamer --chain polkadot`

lib/trie/hash.go
```diff
@@ -5,195 +5,310 @@ package trie
...
-		return h.encodeLeaf(n)
+		err := encodeLeaf(n, buffer)
+		if err != nil {
+			return fmt.Errorf("cannot encode leaf: %w", err)
+		}
+
+		n.encodingMu.Lock()
+		defer n.encodingMu.Unlock()
+
+		n.encoding = make([]byte, buffer.Len())
```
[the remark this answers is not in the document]
qdm12: > I don't think we should be setting attributes on the `leaf` within this function. Could we not call some `leaf.encode()` method and copy it into the buffer then?

Done in 42f83f106870fbbdff023779962e66873ded05cd. I think was trying to split out encoding from storing things inside the leaf. Maybe to do in another PR later.

> Don't we have to `setDirty(true)` as well?

I am not sure, it was not there before. It looks like receiver methods on the `Trie` are the ones using `setDirty` :thinking: I will definitely look into it while refactoring the trie package in other PRs.

lib/trie/hash.go
```diff
@@ -5,195 +5,310 @@ package trie
...
-// encode is the high-level function wrapping the encoding for different node types
-// encoding has the following format:
+var ErrNodeTypeUnsupported = errors.New("node type is not supported")
+
+// encodeNode writes the encoding of the node to the buffer given.
+// It is the high-level function wrapping the encoding for different
+// node types. The encoding has the following format:
 // NodeHeader | Extra partial key length | Partial Key | Value
-func (h *hasher) encode(n node) ([]byte, error) {
+func encodeNode(n node, buffer *bytes.Buffer, parallel bool) (err error) {
```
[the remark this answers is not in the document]
qdm12: Taking note, will do in another PR. Although:

1. we need to use `*bytes.Buffer` to use the sync.Pool (or pass in a byte slice as argument)
2. we cannot have a method on `node`. We can have an `encode` method on branch and leaf, but I would like to do this in another PR once this get merged.

As soon as this gets merged, I'll planning on having packages as

```s
/lib/trie
       |--- node
             |--- leaf
             |--- branch
```

with exported methods on each implementation.

lib/trie/hash.go
qdm12: Let's leave it for now, to keep it similar to how it was in this PR. I'll do another PR to refactor it to use N workers with N = cpu cores. That would not use more memory and would be faster.

lib/trie/hash.go
```diff
@@ -5,195 +5,310 @@ package trie
...
 
-// NewHasher create new Hasher instance
-func newHasher(parallel bool) *hasher {
-	h := hasherPool.Get().(*hasher)
-	h.parallel = parallel
-	return h
+var hasherPool = &sync.Pool{
+	New: func() interface{} {
+		hasher, err := blake2b.New256(nil)
+		if err != nil {
```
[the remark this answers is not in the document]
qdm12: Well, I agree and disagree because

1. If someone in the future decides to set a key for whatever future reason, a panic would help out quickly find why it's crashing
2. `blake2b` is still in the experimental part of the standard library

Let me know if I'm way too paranoid though :smile_cat:

lib/trie/hash.go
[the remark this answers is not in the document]
qdm12: Actually now that I remember, the aim to take this out of `encodeLeaf` is to expose the copying in order to try to remove it. I prefer to have them higher up in the call stack than at the bottom so I can eventually take them out somehow. For now, this copying defeats the point of the sync.Pool buffers in `encodeLeaf` for example. I need to investigate more how this all works together first though. I reverted it and added a comment to remove it.

qdm12:
@timwu20 our holy CI let all the required workflows pass, so *ideally* I'll write down further changes to do for other PRs and merge that one, unless it's a critical problem.

timwu20 (approved):

lib/trie/hash.go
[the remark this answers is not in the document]
timwu20: I think linking to this discussion as comment above the panic would be nice.

timwu20:
> @timwu20 our holy CI let all the required workflows pass, so _ideally_ I'll write down further changes to do for other PRs and merge that one, unless it's a critical problem.

LGTM!

lib/trie/node.go
qdm12: This should be addressed to some extent in #2077

lib/trie/hash.go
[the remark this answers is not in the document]
qdm12: This is placed inside the leaf's `Encode()` method in #2077 in order to not do type assertion and rely on the node interface.

lib/trie/hash.go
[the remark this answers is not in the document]
qdm12: This is done in #2077

lib/trie/hash.go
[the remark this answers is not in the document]
qdm12: This is done in #2077 (https://github.com/ChainSafe/gossamer/pull/2077/files#diff-2ab83e082ae5556c9b4a34bd55d1068ffeb45fa6212e848643fcc9f161cf77bbR36)

PR #2016, opened 2021-11-11
fix(log-levels): do not ignore configuration file log levels

qdm12:
## Changes

- Fix config log levels being ignored (remove default flag values)
- Use `parseLogLevelString` in other places to reduce code duplication
- Rename `cfg` to `tomlConfig` in `setLogConfig` since I got confused by my own code

## Primary Reviewer

@noot


noot (approved):
thanks for the quick fix!

PR #2029, opened 2021-11-12
maintainence(lib/trie): attempt to improve memory by reducing node key/value copies

noot:
## Changes

- I've attempted to reduce the times the trie node key/values are copied, as what was happening previously is that the nodes for an entire path up the tree were copied when one node was modified/inserted/deleted etc. however, in that case, the node key/value generally wouldn't change, only the "encoding/hash" (which depends on the hashes of its children, as it is a merkle tree) so, it was probably unnecessary to copy the keys and values each time. 

## Tests

I suggest testing this by syncing the actual chain, to make sure the trie is still compatible:

```
```

also, will need to profile this to see if it actually improves memory usage at all.

## Issues

- related to #1973 

## Primary Reviewer

- @qdm12


lib/trie/trie.go
```diff
@@ -669,7 +683,12 @@ func (t *Trie) clearPrefix(cn node, prefix []byte) (node, bool) {
 		if len(prefix) == len(c.key)+1 && length == len(prefix)-1 {
 			// found prefix at child index, delete child
 			i := prefix[len(c.key)]
-			c.children[i] = nil
+
+			cpy := [16]node{}
+			copy(cpy[:], c.children[:])
```
qdm12, line 688: nit/FYI this is an array, so you can simply do `cpy := c.children`, no need to make and copy it :wink: Not the case for slices though.

qdm12 (commented):
Nice :+1: I'll bundle it on my sketchy branch together with the sync.pool fix, see what happens! Thanks!

lib/trie/trie.go
```diff
@@ -353,14 +357,24 @@ func (t *Trie) updateBranch(p *branch, key []byte, value node) (n node) {
 		switch c := p.children[key[length]].(type) {
 		case *branch, *leaf:
 			n = t.insert(c, key[length+1:], value)
-			p.children[key[length]] = n
+
+			cpy := [16]node{}
+			copy(cpy[:], p.children[:])
+			cpy[key[length]] = n
+			p.children = cpy
```
qdm12, line 364: What's the point here if we set back `cpy` to `p.children`? :thinking: Same question for below.

noot: I don't think this is actually needed since children is an array :p

noot:
closing for now as @qdm12 has #2009 . if profiling shows that the copying is still an issue, this *could* potentially help

PR #1933, opened 2021-10-27
chore(go-version): upgrade from Go `1.15` to Go `1.17`

qdm12:
## Changes

- Document pre-requisite as Go `1.16`
- Upgrade go.mod, Dockerfiles and workflows to use Go 1.17

## Tests

- All tests

## Issues

- No issue, but blocks me slightly on #1928 

## Primary Reviewer

- @noot

PR #2030, opened 2021-11-12
chore(ioutil): move from io/ioutil to io and os packages

Juneezee:
## Changes

The `io/ioutil` package has been deprecated in Go 1.16 (See https://golang.org/doc/go1.16#ioutil). Since this project has been upgraded to Go 1.17 in #1933, this PR replaces the existing `io/ioutil` functions with their new definitions in `io` and `os` packages.

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -34,6 +34,9 @@ jobs:
     if: ${{ github.actor != 'dependabot[bot]' }}
     runs-on: ubuntu-latest
     steps:
+      - uses: actions/setup-go@v2
+        with:
+          go-version: "1.17.x"
       - uses: actions/checkout@v2
       - name: Checks PR has title and body description
         run: |
--- a/dot/state/storage_notify_test.go
+++ b/dot/state/storage_notify_test.go
@@ -7,7 +7,6 @@ import (
 	"context"
 	"fmt"
 	"log"
-	"os"
 	"sync"
 	"testing"
 	"time"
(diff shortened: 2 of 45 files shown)


.github/workflows/checks.yml
```diff
@@ -34,6 +34,9 @@ jobs:
     if: ${{ github.actor != 'dependabot[bot]' }}
     runs-on: ubuntu-latest
     steps:
+      - uses: actions/setup-go@v2
+        with:
+          go-version: "1.17.x"
```
qdm12, line 39: :pray: Thank you! That went through without us noticing!

kishansagathiya (approved):
:tada: Looks good to me!!

But, I wonder, if we can get away with using `T.TempDir` instead of `os.MkdirTemp` at places. 

`os.MkdirTemp` leaves it to us when the directory is not needed, while `T.TempDir` automatically removes the directory on test cleanup.

qdm12 (approved):
Awesome, thanks a lot :100: !

qdm12:
@kishansagathiya there is no change to the (2) uses of `t.TempDir` in the codebase in this PR :wink:

Juneezee:
@kishansagathiya I have changed all usage of `os.MkdirTemp` in tests to `t.tempDir`. I'll let GitHub Actions runs the unit tests and hopefully the tests can go smoothly.

EclesioMeloJunior (approved):

dot/state/storage_notify_test.go
```diff
@@ -173,16 +171,6 @@ func Test_Example(t *testing.T) {
...
-	dir, err := ioutil.TempDir("", "badger-test")
-	if err != nil {
-		log.Fatal(err)
-	}
-	defer func() {
-		if err = os.RemoveAll(dir); err != nil {
-			log.Fatal(err)
-		}
-	}()
-
```
kishansagathiya, line 185: @EclesioMeloJunior Is this removal harmless?

qdm12: Yes since `NewInMemoryDB` uses `t.TempDir` which removes the directory when the test finishes.

PR #2031, opened 2021-11-12
testing(lib/trie): change `lib/trie` slow and intensive benchmark-oriented tests to benchmarks

qdm12:
## Changes

- Why?
  - This must be slowing down the CI quite badly
  - This makes my CPU melt when running them and forgetting to `t.Skip()` them
  - These are really benchmarks, and not unit testing anything really as far as i know
- Change unit tests to a merged Go benchmark (you can run them with `go test -bench`)
- Use deepcopy to copy the trie instead of re-generating one, which takes time
- Use math/rand instead of crypto/rand to generate trie values faster. I don't think we care about crypto secured randomness in most unit tests, and especially not in a benchmark

## Tests

```
go test -bench -run ^Benchmark_Trie_Hash$ github.com/ChainSafe/gossamer/lib/trie
```

## Primary Reviewer

- @arijitAD

### DIFF
--- a/lib/trie/test_utils.go
+++ b/lib/trie/test_utils.go
@@ -3,7 +3,7 @@ package trie
 import (
 	"crypto/rand"
 	"encoding/binary"
-	"math/big"
+	prand "math/rand"
 	"testing"
 
 	"github.com/stretchr/testify/require"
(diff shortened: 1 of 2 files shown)


EclesioMeloJunior (approved):

noot (approved):
is there a unit test for the parallel trie hashing? we should keep a unit test for it somewhere, but this change is fine

qdm12:
> is there a unit test for the parallel trie hashing? we should keep a unit test for it somewhere, but this change is fine

I'll add one in #2009 since I changed/improved the implementation for the parallel hashing.

PR #2032, opened 2021-11-12
fix(cmd/cfg): Use  Babe Lead value from toml config

kishansagathiya:
## Changes

- Earlier, babe lead value was read only from "babe-lead" cli flag.
cfg.BABELead was being set to false in case this flag was not provided.

This commit sets cfg.BABELead from toml config and uses "babe-lead" cli
flag only when it is set.

## Tests

Run 
```
./bin/gossamer --chain dev
```

If we are lead not, we would not wait for first block, otherwise babe service would fail throwing a timeout error.
## Issues

-

## Primary Reviewer

- @arijitAD

### DIFF
--- a/cmd/gossamer/config.go
+++ b/cmd/gossamer/config.go
@@ -581,7 +581,11 @@ func setDotCoreConfig(ctx *cli.Context, tomlCfg ctoml.CoreConfig, cfg *dot.CoreC
 	cfg.BabeAuthority = tomlCfg.Roles == types.AuthorityRole
 	cfg.GrandpaAuthority = tomlCfg.Roles == types.AuthorityRole
 	cfg.GrandpaInterval = time.Second * time.Duration(tomlCfg.GrandpaInterval)
-	cfg.BABELead = ctx.GlobalBool(BABELeadFlag.Name)
+
+	cfg.BABELead = tomlCfg.BABELead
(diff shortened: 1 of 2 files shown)


noot (approved):

cmd/gossamer/config.go
```diff
@@ -581,7 +581,11 @@ func setDotCoreConfig(ctx *cli.Context, tomlCfg ctoml.CoreConfig, cfg *dot.CoreC
 	cfg.BabeAuthority = tomlCfg.Roles == types.AuthorityRole
 	cfg.GrandpaAuthority = tomlCfg.Roles == types.AuthorityRole
 	cfg.GrandpaInterval = time.Second * time.Duration(tomlCfg.GrandpaInterval)
-	cfg.BABELead = ctx.GlobalBool(BABELeadFlag.Name)
+
+	cfg.BABELead = tomlCfg.BABELead
+	if ctx.IsSet(BABELeadFlag.Name) {
+		cfg.BABELead = ctx.GlobalBool(BABELeadFlag.Name)
+	}
```
qdm12, line 588: Flag should always take precedence over the config file right? If so, all good :+1:

qdm12 (approved):

cmd/gossamer/config.go
kishansagathiya: yes

PR #2047, opened 2021-11-17
chore(lint): check each `//nolint` and improve them

qdm12:
## Changes

- [x] Remove unneeded `//nolint` comments
- [x] Fix some of the lint-ignored errors
- [x] Specify a linter that should be ignored as `//nolint:lintername`
- [x] Add `nolintlint` linter to detect future unused `//nolint` comments
- [x] Move exclude rules from golangci.yml to inlined `//nolint` so `nolintlint` can pick them up.

The idea behind those changes was to find potential code issues that could be the root of the memory issue

## Tests

```
golangci-lint run
```


.golangci.yml
```diff
@@ -125,41 +126,30 @@ issues:
       linters:
         - govet
 
-    # Exclude known linters from partially hard-vendored code,
-    # which is impossible to exclude via "nolint" comments.
-    - path: internal/hmac/
-      text: "weak cryptographic primitive"
+    - text: "`finalized` is a misspelling of `finalised`"
```
noot, line 129: does this ignore this? or what does this mean

noot (approved):

.golangci.yml
qdm12: yes that instructs it to ignore that misspelling error, since we have a bunch of `finalized` words (I guess due to external spec in american english and not UK english).

EclesioMeloJunior (approved):

PR #2049, opened 2021-11-18
chore(trie): `lib/trie/hash.go` tests

qdm12:
## Changes

- Near full coverage and deeper tests for lib/trie/hash.go
- Small production code changes

## Tests

```
go test -race github.com/ChainSafe/gossamer/lib/trie
```

## Issues

- #1973


lib/trie/hash_test.go
```diff
@@ -4,72 +4,1009 @@
...
+			} else {
+				require.NoError(t, err)
+			}
+		})
 	}
 }
 
-func TestHashBranch(t *testing.T) {
-	n := &branch{key: generateRandBytes(380), value: generateRandBytes(380)}
-	n.children[3] = &leaf{key: generateRandBytes(380), value: generateRandBytes(380)}
```
noot, line 47: should we add a test for random nodes as well?

noot (approved):
looks good to me! how do you feel about adding more tests using `generateRand` haha

edwardmack (approved):
Nice work, lgtm!

lib/trie/hash_test.go
qdm12: Why? :thinking: 
I'm a big fan of deterministic tests.
Unless we need crypto randomness in certain niche cases (i.e.generate uniform distribution etc.)?

PR #2051, opened 2021-11-19
chore(lint): limit line length to `120`

qdm12:
## Changes

- Add `lll` linter configured to allow 120 chars per line, with 1 tab = 1 character
- Zip 2 mega huge strings used in tests in `testdata` so my IDE doesn't break. Note `testdata` is ignored on the built binary so the `init()` functions won't trigger.
- Fix most lines to be < 120 characters
- Add `//nolint:lll` comments on most hex strings
- Fix `lll` exclusion for `go generate` comments

## Tests

```
golangci-lint run
```

## Issues

- Following our dev tooling yesterday. I've done that during our polkadot guild meeting since it's rather no-brain-required :smile: 

## Primary Reviewer

- @danforbes since you looked quite enthusiast about it :smile_cat:


danforbes:
Did you manually fix up the offending lines or use some tool to do so?

qdm12:
@danforbes Manually :sweat_smile: But now since it's integrated in the CI, once this is merged no line longer than 120 lines will be allowed. It will be underlined in IDEs as well.

There is [golines](https://github.com/segmentio/golines) to sort of auto-format lines. Personally, I prefer to review these manually since (unlike js) they happen rather rarely in Go, and sometimes they make sense (like with `//nolint:lll` comments).

EclesioMeloJunior (approved):

noot (approved):

PR #2055, opened 2021-11-19
maintenance(dot/sync): add clearing of pending blocks set

### LINKED ISSUE #1976: dot/sync: pending blocks set blocks should have "time to live" or be cleared if no peer is aware of them
## Task summary

the sync package contains a "pending blocks set" which is a set of blocks that the node is aware of, but does not have enough information about them to process them (ie. missing data, missing parent/ancestors). in tip mode, this set is iterated through and requests are made in an attempt to retrieve the missing data and process those blocks.
- the pending blocks in the sync package are never cleared out, unless they are made ready to process
- this is an attack vector, eg. some malicious node spams us with some blocks that they say they have, but don't actually exist
- a non-malicious node could also build some block, gossip it to us, then go offline right after, so then no one on the network actually has the full info for this block and it ends up trapped in the set
- a pending block should be cleared if: 1. we are in tip syncing mode and we haven't managed to sync this block for some amount of time, or 2. we have asked every peer for this block, and none of them have it
- personally I think 1 will be simpler to implement, although both are doable (we could have both)

noot:
## Changes

- clear pending block if we've tried to request it from every peer and none of them have it
- clear pending block if it's past its TTL (currently set to 10min, somewhat arbitrarily)

## Tests

```
go test ./dot/sync 
```

## Issues

- closes #1976

## Primary Reviewer

- @jimjbrettj


dot/sync/disjoint_block_set.go
```diff
@@ -90,6 +103,35 @@ func newDisjointBlockSet(limit int) *disjointBlockSet {
...
 }
 
+func (s *disjointBlockSet) start(ctx context.Context) {
+	timer := time.NewTimer(clearBlocksInterval)
+
+	go func() {
+		for {
+			select {
+			case <-timer.C:
+				s.clearBlocks()
```
EclesioMeloJunior, line 113: if this function should run every minute, then you should change from `time.NewTimer()` to `time.NewTicker()`

EclesioMeloJunior (changes requested):

EclesioMeloJunior (changes requested):

dot/sync/disjoint_block_set.go
noot: done!

dot/sync/disjoint_block_set.go
```diff
@@ -113,8 +156,9 @@ func (s *disjointBlockSet) addHashAndNumber(hash common.Hash, number *big.Int) e
 	}
 
 	s.blocks[hash] = &pendingBlock{
-		hash:   hash,
-		number: number,
+		hash:    hash,
+		number:  number,
+		clearAt: time.Now().Add(ttl),
```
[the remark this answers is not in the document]
noot: hmm, I guess the issue is that a pending block can be constructed from just a (hash, number) pair, a header, or a block. I can still do this if you'd like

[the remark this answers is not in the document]
noot: done!

EclesioMeloJunior (approved):

dot/sync/disjoint_block_set.go
```diff
@@ -4,14 +4,22 @@
...
 	"sync"
+	"time"
 
 	"github.com/ChainSafe/gossamer/dot/types"
 	"github.com/ChainSafe/gossamer/lib/common"
 )
 
+const (
+	// ttl is the time that a block can stay in this set before being cleared.
+	ttl                 = time.Minute * 10
```
qdm12, line 19: nit call me old school, but shouldn't we put the unit at the end? :smile_cat: 

```suggestion
	ttl                 = 10 * time.Minute
```

dot/sync/disjoint_block_set.go
```diff
@@ -44,6 +53,20 @@ type pendingBlock struct {
...
+	clearAt time.Time
+}
+
+func newPendingBlock(hash common.Hash, number *big.Int, header *types.Header, body *types.Body) *pendingBlock {
+	return &pendingBlock{
+		hash:    hash,
+		number:  number,
+		header:  header,
+		body:    body,
+		clearAt: time.Now().Add(ttl),
```
qdm12, line 68: Should we inject the current time? :thinking: For future testability purposes?

dot/sync/disjoint_block_set.go
```diff
@@ -90,6 +113,33 @@ func newDisjointBlockSet(limit int) *disjointBlockSet {
 	}
 }
 
+func (s *disjointBlockSet) start(ctx context.Context) {
```
qdm12, line 116: A suggestion for here and other 'start'-like functions.

The problem with these 'fire-and-forget' start goroutines is that we don't know when they are done.
The simplest way to change this is for `start` to return a `done <-chan struct{}` channel which is closed by a `defer close(done)` inside the goroutine launched. The caller should then wait (or ignore) when e.g. canceling the context for the goroutine to cleanly exit.

Ideally, to also reduce indentation and complexity of the function, change it to a

```go
func (s *disjointBlockSet) run(ctx context.Context, done chan<- struct{}) {
```

Which would run the for loop directly without lauching a goroutine. The caller would have to use `go dBlockSet.run(ctx, done)` to run async if he wants to. That way it's up to the caller to run it in a blocking manner or not.

Finally, you can also change `done` to be an error channel if you want to get a final error such as `ctx.Err()` or some other error.

dot/sync/disjoint_block_set.go
```diff
@@ -90,6 +113,33 @@ func newDisjointBlockSet(limit int) *disjointBlockSet {
...
+	timer := time.NewTicker(clearBlocksInterval)
+
+	go func() {
+		for {
+			select {
+			case <-timer.C:
+				s.clearBlocks()
+			case <-ctx.Done():
+				timer.Stop()
```
qdm12, line 125:
```suggestion
				if !timer.Stop() {
					<-timer.C
				}
```

dot/sync/disjoint_block_set.go
```diff
@@ -125,21 +172,17 @@ func (s *disjointBlockSet) addHeader(header *types.Header) error {
 	defer s.Unlock()
 
 	hash := header.Hash()
-	b, has := s.blocks[hash]
-	if has {
+	if b, has := s.blocks[hash]; has {
```
qdm12, line 175: Let's use `ok` unless it has a special meaning, to fit with Go's standard lib 'convention'

```suggestion
	if b, ok := s.blocks[hash]; ok {
```

dot/sync/disjoint_block_set.go
```diff
@@ -125,21 +172,17 @@ func (s *disjointBlockSet) addHeader(header *types.Header) error {
...
 		b.header = header
+		b.clearAt = time.Now().Add(ttl)
```
qdm12, line 177: Let's have a field in `disjointBlockSet` as `timeNow func() time.Time` and set it it its constructor. You can even inject `time.Now` to the constructor so even the constructor can be used in tests.

And use `b.clearAt = s.timeNow().Add(ttl)` here

dot/sync/disjoint_block_set.go
```diff
@@ -149,23 +192,18 @@ func (s *disjointBlockSet) addBlock(block *types.Block) error {
 	defer s.Unlock()
 
 	hash := block.Header.Hash()
-	b, has := s.blocks[hash]
-	if has {
+	if b, has := s.blocks[hash]; has {
```
qdm12, line 195:
```suggestion
	if b, ok := s.blocks[hash]; ok {
```

dot/sync/disjoint_block_set.go
```diff
@@ -187,6 +226,12 @@ func (s *disjointBlockSet) addJustification(hash common.Hash, just []byte) error
 func (s *disjointBlockSet) removeBlock(hash common.Hash) {
 	s.Lock()
 	defer s.Unlock()
+	s.removeBlockInner(hash)
+}
+
+// this function does not lock!!
```
qdm12, line 232: :pray: thank you haha!

dot/sync/disjoint_block_set_test.go
```diff
@@ -195,3 +196,24 @@ func TestDisjointBlockSet_getReadyDescendants_blockNotComplete(t *testing.T) {
...
+
+func TestDisjointBlockSet_ClearBlocks(t *testing.T) {
+	s := newDisjointBlockSet(pendingBlocksLimit)
+
+	testHashA := common.Hash{0}
+	testHashB := common.Hash{1}
+
+	s.blocks[testHashA] = &pendingBlock{
+		hash:    testHashA,
+		clearAt: time.Now().Add(ttl * -2),
```
qdm12, line 208: nit you could use a common `now := time.Unix(1000, 0)` for both A and B so we don't rely on the clock. I'm a bit paranoid we usage of `time.Now()` in tests although it's probably fine here I think.

qdm12 (approved):

dot/sync/disjoint_block_set.go
qdm12: I have seen in some places like in the Docker repo, where there is also a struct channel passed in that is closed right before the `select` to signal the caller the goroutine is ready and operational. I don't really do that personally but it could be nice :wink:

dot/sync/disjoint_block_set.go
noot: this is only for timers, not tickers (this is a ticker, I'll rename it) https://pkg.go.dev/time#Ticker.Stop

dot/sync/disjoint_block_set.go
noot: i like has better as it's more descriptive, we're checking if the set has the block or not, but if you feel strongly i can change it

dot/sync/disjoint_block_set.go
noot: same as above

dot/sync/disjoint_block_set.go
noot: haha, done

dot/sync/disjoint_block_set.go
noot: updated

dot/sync/disjoint_block_set.go
noot: updated!

dot/sync/disjoint_block_set.go
noot: updated

dot/sync/disjoint_block_set_test.go
noot: updated

timwu20:
@jimjbrettj you are primary on this PR

dot/sync/disjoint_block_set.go
qdm12: Ah oops, I thought it was a timer indeed!

dot/sync/chain_sync.go
```diff
@@ -192,12 +193,16 @@ func (cs *chainSync) start() {
 		time.Sleep(time.Millisecond * 100)
 	}
 
+	pendingBlockDoneCh := make(chan struct{})
+	cs.pendingBlockDoneCh = pendingBlockDoneCh
+	go cs.pendingBlocks.start(cs.ctx, pendingBlockDoneCh)
```
qdm12, line 198: nit

```suggestion
	cs.pendingBlockDoneCh = make(chan struct{})
	go cs.pendingBlocks.start(cs.ctx, cs.pendingBlockDoneCh)
```

dot/sync/chain_sync.go
```diff
@@ -192,12 +193,16 @@ func (cs *chainSync) start() {
...
 	go cs.sync()
 	go cs.logSyncSpeed()
 }
 
 func (cs *chainSync) stop() {
 	cs.cancel()
+	<-cs.pendingBlockDoneCh
```
qdm12, line 205: Awesome! :+1:

dot/sync/disjoint_block_set.go
```diff
@@ -80,13 +103,42 @@ type disjointBlockSet struct {
...
 func newDisjointBlockSet(limit int) *disjointBlockSet {
 	return &disjointBlockSet{
 		blocks:           make(map[common.Hash]*pendingBlock),
 		parentToChildren: make(map[common.Hash]map[common.Hash]struct{}),
 		limit:            limit,
+		timeNow:          time.Now,
+	}
+}
+
+func (s *disjointBlockSet) start(ctx context.Context, done chan<- struct{}) {
```
qdm12, line 119: You should call that `run`. In my opinion `start` is something launching a goroutine (only starting something then returning), `run` on the other hand is 'running' aka blocking. Feel free to correct me or introduce a new word though, I'd be happy to use something else than `run`

dot/sync/disjoint_block_set.go
```diff
@@ -80,13 +103,42 @@ type disjointBlockSet struct {
...
+	defer close(done)
+	ticker := time.NewTicker(clearBlocksInterval)
```
qdm12, line 121: nit You can defer the ticker Stop here instead of having it in the `case <-ctx.Done():` block

dot/sync/disjoint_block_set.go
```diff
@@ -80,13 +103,42 @@ type disjointBlockSet struct {
...
+		}
+	}
+}
+
+func (s *disjointBlockSet) clearBlocks() {
+	s.Lock()
+	defer s.Unlock()
+
+	for _, block := range s.blocks {
+		if time.Since(block.clearAt) > 0 {
```
qdm12, line 139: Use the `s.timeNow()` for testability. From the top of my head this should work

```suggestion
		if s.timeNow().Add(-block.clearAt) > 0 {
```

dot/sync/disjoint_block_set_test.go
```diff
@@ -15,6 +16,9 @@ import (
 
 func TestDisjointBlockSet(t *testing.T) {
 	s := newDisjointBlockSet(pendingBlocksLimit)
+	s.timeNow = func() time.Time {
+		return time.Time{}
+	}
```
qdm12, line 21: nit you could use a constant reference point, i.e.

```go
	const nowUnix = 1000
	now := time.Unix(nowUnix, 0)
	s.timeNow = func() time.Time {
		return now
	}

	// Further down you can use `now` instead of `time.Time{}`
```

That helped me out when writing tests for heavily time oriented things. Not really the case here but a good way to write time tests I think :wink:

dot/sync/disjoint_block_set_test.go
```diff
@@ -195,3 +204,24 @@ func TestDisjointBlockSet_getReadyDescendants_blockNotComplete(t *testing.T) {
...
+		clearAt: time.Unix(1000, 0),
+	}
+	s.blocks[testHashB] = &pendingBlock{
+		hash:    testHashB,
+		clearAt: time.Now().Add(ttl * 2),
+	}
+
+	s.clearBlocks()
+	require.Equal(t, 1, len(s.blocks))
+	_, has := s.blocks[testHashB]
```
qdm12, line 225: Should the returned value be asserted as well :thinking:

dot/sync/worker.go
```diff
@@ -76,6 +76,7 @@ type worker struct {
 	startNumber  *big.Int
 	targetHash   common.Hash
 	targetNumber *big.Int
+	pendingBlock *pendingBlock // if this worker is tied to a specific pending block, this field is set
```
qdm12, line 79: Less work for #2051 :smile_cat: 

```suggestion
	// if this worker is tied to a specific pending block, this field is set
	pendingBlock *pendingBlock
```

qdm12 (approved):

dot/sync/chain_sync.go
noot: `cs.pendingBlockDoneCh` is of type `<-chan struct{}` but the parameter to `start` is `chan<- struct{}` so i don't think this works

dot/sync/disjoint_block_set_test.go
noot: I think  `time.Time{}` is essentially time 0 - the problem I found with using `time.Unix` was that the other cached values inside the `time.Time` struct didn't match when doing the `Equal` check, causing tests to fail

dot/sync/disjoint_block_set_test.go
noot: we just need to check that the block with hash b exists and a doesn't, `clearBlocks` doesn't modify blocks

dot/sync/chain_sync.go
qdm12: Ah good point! My bad!

dot/sync/disjoint_block_set_test.go
qdm12: Yes it's time 0.

That's a strange problem, I never seen this. There is the location information in time.Time but it shouldn't matter. Let me know if you encounter this problem again, I'd be curious!

dot/sync/disjoint_block_set_test.go
qdm12: > doesn't modify blocks

Yes, we should check that :smile: Just in case :tm: (also in case someone modifies the code in the future)
The deeper the test the better!

qdm12 (approved):

PR #2057, opened 2021-11-20
feat(wasmer/crypto): move sig verifier to crypto pkg

### LINKED ISSUE #1930: Move `./lib/runtime/sig_verifier.go` to `./lib/crypto` package
## Issue summary

- The file `./lib/runtime/sig_verifier.go` placed at the `runtime` package does not depend on runtime stuff, and it just acts like a worker that verifies batch signatures in a concurrently way. 
- We need to move the file to `./lib/crypto` package as make more sense to that file be there.

- N/A

1garo:
## Changes

- Move sig_verifier from `runtime` to `crypto` pkg;

## Tests

```
go test ./lib/crypto
go test ./lib/runtime 
```

## Issues

Closes #1930

## Primary Reviewer

@EclesioMeloJunior


lib/crypto/sig_verifier.go
```diff
@@ -1,26 +1,24 @@
...
 	"time"
 
 	"github.com/ChainSafe/gossamer/internal/log"
-	"github.com/ChainSafe/gossamer/lib/crypto"
-	"github.com/ChainSafe/gossamer/lib/crypto/ed25519"
-	"github.com/ChainSafe/gossamer/lib/crypto/sr25519"
-	"github.com/ethereum/go-ethereum/crypto/secp256k1"
 )
 
+// SigVerifyFunc verify an signature given a pubkey and msg
```
qdm12, line 13: nit fixing up the comment :wink: 

```suggestion
// SigVerifyFunc verifies a signature given a public key and a message
```

lib/crypto/sig_verifier.go
```diff
@@ -1,26 +1,24 @@
...
+type SigVerifyFunc func(pubkey, sig, msg []byte) (bool, error)
```
qdm12, line 14: nit using named returned values for documentation purposes

```suggestion
type SigVerifyFunc func(pubkey, sig, msg []byte) (ok bool, err error)
```

lib/crypto/sig_verifier.go
```diff
@@ -66,12 +64,12 @@ func (sv *SignatureVerifier) Start() {
 			case <-sv.closeCh:
 				return
 			default:
-				sign := sv.Remove()
-				if sign == nil {
+				sig := sv.Remove()
```
qdm12, line 67: nit can you rename it to `signature` please? I think we should aim to use full words to avoid confusion :wink: Thanks!

lib/crypto/sr25519/sr25519.go
```diff
@@ -43,6 +44,20 @@ type PrivateKey struct {
 	key *sr25519.SecretKey
 }
 
+// SignVerify verify signature given pubkey and msg
```
qdm12, line 47:
```suggestion
// SigVerify verifies a signature corresponds to a message given a public key
```

lib/crypto/sr25519/sr25519.go
```diff
@@ -43,6 +44,20 @@ type PrivateKey struct {
...
+func SignVerify(pubkey, sig, msg []byte) (bool, error) {
```
qdm12, line 48:
```suggestion
func SignVerify(pubkey, sig, msg []byte) (ok bool, err error) {
```

lib/crypto/sr25519/sr25519.go
```diff
@@ -43,6 +44,20 @@ type PrivateKey struct {
...
+	pubKey, err := NewPublicKey(pubkey)
+	if err != nil {
+		return false, fmt.Errorf("failed to fetch sr25519 public key: %s", err)
```
qdm12, line 51: I know this is not new code, but let's patch it up whilst we're at it :wink: 

Using `%w` to wrap the error

```suggestion
		return false, fmt.Errorf("failed to fetch sr25519 public key: %w", err)
```

lib/crypto/sr25519/sr25519.go
```diff
@@ -43,6 +44,20 @@ type PrivateKey struct {
...
+	}
+	ok, err := pubKey.Verify(msg, sig)
+	if err != nil || !ok {
+		return false, fmt.Errorf("failed to verify sr25519 signature: %s", err)
```
qdm12, line 55:
```suggestion
		return false, fmt.Errorf("failed to verify sr25519 signature: %w", err)
```

qdm12, line 55: Please split this if condition; we want to have one checking the error and returning `return false, fmt.Errorf("failed to verify signature: %w", err)` and another condition `if !ok` returning `return false, nil`

Otherwise there is no point returning two values `ok` and `err` from this function

lib/runtime/wasmer/instance.go
```diff
@@ -15,6 +15,8 @@ import (
 	"github.com/ChainSafe/gossamer/lib/runtime/offchain"
 	"github.com/ChainSafe/gossamer/lib/trie"
 
+	"github.com/ChainSafe/gossamer/lib/crypto"
+
```
qdm12, line 19: I think you can remove the empty lines surrounding this new import line.

lib/crypto/secp256k1/secp256k1.go
```diff
@@ -36,6 +37,16 @@ type PublicKey struct {
 	key ecdsa.PublicKey
 }
 
+// SignVerify verify signature given pubkey and msg
+func SignVerify(pubkey, sig, msg []byte) (bool, error) {
+	ok := secp256k1.VerifySignature(pubkey, msg, sig)
+	if !ok {
+		return false, fmt.Errorf("failed to verify secp256k1 signature")
```
qdm12, line 44: Please define this error as a global scope variable at the top of the file

```go
var ErrSignatureVerificationFailed = errors.New("failed to verify signature")
```

and you can use it here with

```go
return false, fmt.Errorf("secp256k1: %w: for message 0x%x, signature 0x%x and public key 0x%x",
     ErrSignatureVerificationFailed, msg, sig, pubkey)
```

This applies to other uses of errors in other signature verification functions

lib/crypto/secp256k1/secp256k1.go
```diff
@@ -36,6 +37,16 @@ type PublicKey struct {
 	key ecdsa.PublicKey
 }
 
+// SignVerify verify signature given pubkey and msg
```
qdm12, line 40: nit fixing up the comment :wink: 

```suggestion
// SignVerify verifies a signature given a public key and a message
```

lib/crypto/secp256k1/secp256k1.go
```diff
@@ -36,6 +37,16 @@ type PublicKey struct {
...
+func SignVerify(pubkey, sig, msg []byte) (bool, error) {
```
qdm12, line 41: Can you change this to use `publicKey, signature, message []byte` for readability? Thanks!

qdm12, line 41:
```suggestion
func SignVerify(pubkey, sig, msg []byte) (ok bool, err error) {
```

lib/crypto/ed25519/ed25519.go
```diff
@@ -44,6 +44,20 @@ type PublicKey ed25519.PublicKey
 // PublicKeyBytes is an encoded ed25519 public key
 type PublicKeyBytes [PublicKeyLength]byte
 
+// SignVerify verify signature given pubkey and msg
```
qdm12, line 47: nit fixing up the comment :wink: 

```suggestion
// SignVerify verifies a signature given a public key and a message
```

lib/crypto/ed25519/ed25519.go
```diff
@@ -44,6 +44,20 @@ type PublicKey ed25519.PublicKey
...
+func SignVerify(pubkey, sig, msg []byte) (bool, error) {
```
qdm12, line 48:
```suggestion
func SignVerify(pubkey, sig, msg []byte) (ok bool, err error) {
```

qdm12, line 48: Can you change this to use `publicKey, signature, message []byte` for readability? Thanks!

lib/crypto/sig_verifier.go
```diff
@@ -66,12 +64,12 @@ func (sv *SignatureVerifier) Start() {
...
+				if sig == nil {
 					continue
 				}
-				err := sign.verify()
-				if err != nil {
+				ok, err := sig.VerifyFunc(sig.PubKey, sig.Sign, sig.Msg)
+				if err != nil || !ok {
```
qdm12, line 72: Please split this in two. Otherwise we might log a `nil` error if `ok` is false.

lib/runtime/test_helpers.go
```diff
@@ -12,9 +12,10 @@ import (
 	"testing"
 	"time"
 
+	"github.com/ChainSafe/gossamer/lib/crypto"
+
```
qdm12, line 16:
```suggestion
```

lib/crypto/ed25519/ed25519.go
```diff
@@ -44,6 +44,20 @@ type PublicKey ed25519.PublicKey
...
+	pubKey, err := NewPublicKey(pubkey)
+	if err != nil {
+		return false, fmt.Errorf("failed to fetch ed25519 public key: %s", err)
+	}
+	ok, err := pubKey.Verify(msg, sig)
+	if err != nil || !ok {
+		return false, fmt.Errorf("failed to verify ed25519 signature: %s", err)
```
qdm12, line 55: Let's split this in two if conditions, since error might be `nil` here.

qdm12 (commented):
Thanks! Some comments mostly regarding the older code

lib/crypto/ed25519/ed25519.go
noot:
```suggestion
func VerifySignature(pubkey, sig, msg []byte) (bool, error) {
```

lib/crypto/ed25519/ed25519.go
```diff
@@ -44,6 +44,20 @@ type PublicKey ed25519.PublicKey
...
+	}
```
noot, line 56:
```suggestion
	ok, err := pubKey.Verify(msg, sig)
	if err != nil {
		return false, fmt.Errorf("failed to verify ed25519 signature: %s", err)
	}
	
	if !ok {
	    return false, nil
	}
	
```

lib/crypto/secp256k1/secp256k1.go
noot:
```suggestion
func VerifySignature(pubkey, sig, msg []byte) (bool, error) {
```

lib/crypto/sr25519/sr25519.go
noot:
```suggestion
func VerifySignature(pubkey, sig, msg []byte) (bool, error) {
```

1garo:
I hope that everything looks fine, thank y'all for the comments! @noot @qdm12

EclesioMeloJunior:
Good job @1garo! Please create a file `sig_verifier_test.go` where you create a set of different `Signature` objects and add them to `SignatureVerifier` using the function `SignatureVerifier.Add(...)`, start the the verifier using the function `SignatureVerifier.Start()`, and checks the result using `SignatureVerifier.Finish()`

dot/network/mock_transaction_handler.go
```diff
@@ -12,20 +12,20 @@ type MockTransactionHandler struct {
...
+		r0 = rf(_a0, _a1)
 	} else {
 		r0 = ret.Get(0).(bool)
 	}
 
 	var r1 error
-	if rf, ok := ret.Get(1).(func(*TransactionMessage) error); ok {
-		r1 = rf(_a0)
+	if rf, ok := ret.Get(1).(func(peer.ID, *TransactionMessage) error); ok {
+		r1 = rf(_a0, _a1)
```
qdm12, line 28: That breaks some tests, I've updated the mocks and fixed the tests in #2062 
Either apply the same fix as in this other PR, or wait for it to be merged and then rebase on development / merge development in your branch.

lib/crypto/ed25519/ed25519_test.go
```diff
@@ -99,3 +99,19 @@ func TestNewKeypairFromMnenomic_Again(t *testing.T) {
...
 	require.Equal(t, expectedPubkey, kp.Public().Encode())
 }
+
+func TestVerifySignature(t *testing.T) {
+	keypair, err := GenerateKeypair()
+	require.NoError(t, err)
+
+	content := "Hello world!"
+
+	message := []byte(content)
```
qdm12, line 109: nit simplify

```suggestion
	message := []byte("Hello world!")
```

lib/crypto/ed25519/ed25519_test.go
```diff
@@ -99,3 +99,19 @@ func TestNewKeypairFromMnenomic_Again(t *testing.T) {
...
+	signature, err := keypair.Sign(message)
+	require.NoError(t, err)
+
+	ok, err := VerifySignature(keypair.public.Encode(), signature, message)
+	require.NoError(t, err)
+
+	require.Equal(t, ok, true)
```
qdm12, line 116: nit

```suggestion
	require.True(t, ok)
```

lib/crypto/ed25519/ed25519_test.go
```diff
@@ -99,3 +99,19 @@ func TestNewKeypairFromMnenomic_Again(t *testing.T) {
 	expectedPubkey := common.MustHexToBytes("0xf56d9231e7b7badd3f1e10ad15ef8aa08b70839723d0a2d10d7329f0ea2b8c61")
 	require.Equal(t, expectedPubkey, kp.Public().Encode())
 }
+
+func TestVerifySignature(t *testing.T) {
```
qdm12, line 103: Can you have a test with the verification failing as well please?

You could use test cases and subtests, [example](https://github.com/ChainSafe/gossamer/blob/e7fbfe5ef8241ad0d50850c8b3d7ae417f428f35/lib/trie/hash_test.go#L834-L874)

lib/crypto/secp256k1/secp256k1_test.go
```diff
@@ -154,3 +154,19 @@ func TestRecoverPublicKeyCompressed(t *testing.T) {
...
 	require.Equal(t, kp.Public(), r)
 }
+
+func TestVerifySignature(t *testing.T) {
+	keypair, err := GenerateKeypair()
+	require.NoError(t, err)
+
+	content := "a225e8c75da7da319af6335e7642d473"
+
+	message := []byte(content)
```
qdm12, line 164: nit

```suggestion
	message := []byte("a225e8c75da7da319af6335e7642d473")
```

lib/crypto/secp256k1/secp256k1_test.go
```diff
@@ -154,3 +154,19 @@ func TestRecoverPublicKeyCompressed(t *testing.T) {
...
+	signature, err := keypair.Sign(message)
+	require.NoError(t, err)
+
+	ok, err := VerifySignature(keypair.Public().Encode(), signature[:64], message)
+	require.NoError(t, err)
+
+	require.Equal(t, ok, true)
```
qdm12, line 171:
```suggestion
	require.True(t, ok)
```

lib/crypto/secp256k1/secp256k1_test.go
```diff
@@ -154,3 +154,19 @@ func TestRecoverPublicKeyCompressed(t *testing.T) {
 	require.NoError(t, err)
 	require.Equal(t, kp.Public(), r)
 }
+
+func TestVerifySignature(t *testing.T) {
```
qdm12, line 158: Same as above, please add a test case when there is an error.

lib/crypto/sig_verifier.go
```diff
@@ -1,26 +1,28 @@
...
 	"time"
 
 	"github.com/ChainSafe/gossamer/internal/log"
-	"github.com/ChainSafe/gossamer/lib/crypto"
-	"github.com/ChainSafe/gossamer/lib/crypto/ed25519"
-	"github.com/ChainSafe/gossamer/lib/crypto/sr25519"
-	"github.com/ethereum/go-ethereum/crypto/secp256k1"
 )
 
+// ErrSignatureVerificationFailed usaged when signature verification fails
```
qdm12, line 14: no need, the linter does not require it. And the name is self explanatory.

```suggestion
```

lib/crypto/sig_verifier.go
qdm12: This is not patched, please patch, thanks!

lib/crypto/sig_verifier.go
qdm12: This is not patched, please patch, thanks!

lib/crypto/sr25519/sr25519.go
```diff
@@ -43,6 +44,25 @@ type PrivateKey struct {
 	key *sr25519.SecretKey
 }
 
+// VerifySignature verifies a signature given a public key and a message
+func VerifySignature(publicKey, signature, message []byte) (bool, error) {
+	pubKey, err := NewPublicKey(publicKey)
+	if err != nil {
+		return false, fmt.Errorf("failed to fetch sr25519 public key: %w", err)
```
qdm12, line 51: Are we actually fetching a public key? Sorry I have no idea what `NewPublicKey` does

lib/crypto/sr25519/sr25519_test.go
```diff
@@ -119,3 +119,19 @@ func TestNewKeypairFromMnenomic(t *testing.T) {
...
 	require.NoError(t, err)
 }
+
+func TestVerifySignature(t *testing.T) {
+	keypair, err := GenerateKeypair()
+	require.NoError(t, err)
+
+	content := "Hello world!"
+
+	message := []byte(content)
```
qdm12, line 129: nit

```suggestion
	message := []byte("Hello world!")
```

lib/crypto/sr25519/sr25519_test.go
```diff
@@ -119,3 +119,19 @@ func TestNewKeypairFromMnenomic(t *testing.T) {
...
+	signature, err := keypair.Sign(message)
+	require.NoError(t, err)
+
+	ok, err := VerifySignature(keypair.public.Encode(), signature, message)
+	require.NoError(t, err)
+
+	require.Equal(t, ok, true)
```
qdm12, line 136: nit

```suggestion
	require.True(t, ok)
```

lib/crypto/sr25519/sr25519_test.go
```diff
@@ -119,3 +119,19 @@ func TestNewKeypairFromMnenomic(t *testing.T) {
 	_, err = NewKeypairFromMnenomic(mnemonic, "")
 	require.NoError(t, err)
 }
+
+func TestVerifySignature(t *testing.T) {
```
qdm12, line 123: Add a test case when this errors

qdm12 (changes requested):

lib/crypto/sr25519/sr25519.go
EclesioMeloJunior: `ed25519.NewPublicKey` and `sr25519.NewPublicKey` functions creates a public key given a byte slice, @1garo I think we can just return the `err`

qdm12: Yes let's return `err` only :+1: 

The `fetch` is wrong here, and the `sr25519` error context should be added by the caller. I'll add comment on the caller.

lib/crypto/sig_verifier.go
```diff
@@ -1,26 +1,28 @@
...
+var ErrSignatureVerificationFailed = errors.New("failed to verify signature")
+
+// SigVerifyFunc verify an signature given a pubkey and msg
+type SigVerifyFunc func(pubkey, sig, msg []byte) (ok bool, err error)
+
 // Signature ...
 type Signature struct {
```
qdm12, line 21: A bit out of scope, but this is certainly not a `Signature`, maybe rename it to i.e. `Signing` or `SigningVerifier`? :thinking:

lib/crypto/ed25519/ed25519.go
```diff
@@ -44,6 +44,25 @@ type PublicKey ed25519.PublicKey
...
+	if err != nil {
+		return false, fmt.Errorf("ed25519: %w: for message 0x%x, signature 0x%x and public key 0x%x",
+			crypto.ErrSignatureVerificationFailed, message, signature, publicKey)
+	}
+
+	if !ok {
+		return false, nil
+	}
+
+	return true, nil
```
qdm12, line 63: Please change the signatures of all `VerifySignature` functions to only return an error. `ok` is not needed here since we return `crypto.ErrSignatureVerificationFailed` if the verification failed.

lib/crypto/ed25519/ed25519_test.go
```diff
@@ -99,3 +102,54 @@ func TestNewKeypairFromMnenomic_Again(t *testing.T) {
...
+		},
+		"failed to fetch publicKey": {
+			publicKey: []byte(""),
+			signature: signature,
+			message:   message,
+			err:       errors.New("cannot create public key: input is not 32 bytes"),
+		},
+		"verification failed": {
+			publicKey: keypair.public.Encode(),
+			signature: []byte(""),
```
qdm12, line 133: Supernit no need for conversion string to bytes

```suggestion
			signature: []byte{},
```

lib/crypto/ed25519/ed25519_test.go
```diff
@@ -99,3 +102,54 @@ func TestNewKeypairFromMnenomic_Again(t *testing.T) {
...
+		publicKey, signature, message []byte
+		err                           error
+	}{
+		"success": {
+			publicKey: keypair.public.Encode(),
+			signature: signature,
+			message:   message,
+			err:       nil,
+		},
+		"failed to fetch publicKey": {
```
qdm12, line 125:
```suggestion
		"bad public key input": {
```

lib/crypto/ed25519/ed25519_test.go
```diff
@@ -99,3 +102,54 @@ func TestNewKeypairFromMnenomic_Again(t *testing.T) {
...
+
+	testCase := map[string]struct {
+		publicKey, signature, message []byte
+		err                           error
+	}{
+		"success": {
+			publicKey: keypair.public.Encode(),
+			signature: signature,
+			message:   message,
+			err:       nil,
```
qdm12, line 123: nit no need to set zeroed fields

```suggestion
```

lib/crypto/ed25519/ed25519.go
```diff
@@ -44,6 +44,22 @@ type PublicKey ed25519.PublicKey
...
+// VerifySignature verifies a signature given a public key and a message
+func VerifySignature(publicKey, signature, message []byte) error {
+	pubKey, err := NewPublicKey(publicKey)
+	if err != nil {
+		return err
+	}
+
+	_, err = pubKey.Verify(message, signature)
+	if err != nil {
+		return fmt.Errorf("ed25519: %w: for message 0x%x, signature 0x%x and public key 0x%x",
```
qdm12, line 56: `ed25519` should be set by the caller, not here.

lib/crypto/sig_verifier_test.go
```diff
@@ -0,0 +1,222 @@
...
+		require.NoError(t, err)
+
+		signs[i] = &crypto.Signing{
+			PubKey:     key.Public().Encode(),
+			Sign:       sign,
+			Msg:        msg,
+			VerifyFunc: ed25519.VerifySignature,
+		}
+	}
+	signVerify := crypto.NewSignatureVerifier(log.New())
```
qdm12, line 37: You'll have to import `io` as well at the top

```suggestion
	signVerify := crypto.NewSignatureVerifier(log.New(log.SetWriter(io.Discard)))
```

This applies to other `log.New()` in this file.

lib/crypto/sig_verifier_test.go
```diff
@@ -0,0 +1,222 @@
...
+	for i := 0; i < 2; i++ {
+		msg := []byte("Hello")
+		key, err := ed25519.GenerateKeypair()
+		require.NoError(t, err)
+
+		sign, err := key.Sign(msg)
+		require.NoError(t, err)
+		if i == 1 {
+			signs[i] = &crypto.Signing{
+				PubKey:     []byte(""),
```
qdm12, line 61: nit replace all the `[]byte("")` with `[]byte{}` :wink:

lib/crypto/sig_verifier_test.go
```diff
@@ -0,0 +1,222 @@
...
+	for _, sig := range signs {
+		signVerify.Add(sig)
+	}
+
+	signVerify.Start()
+	ok := signVerify.Finish()
+	require.True(t, ok)
+}
+
+func TestSigVerifierSr25519Fails(t *testing.T) {
```
qdm12, line 116: Can you perhaps merge the success and fail tests together? :thinking: It looks like quite a bit of code duplication

lib/crypto/sr25519/sr25519_test.go
```diff
@@ -119,3 +123,54 @@ func TestNewKeypairFromMnenomic(t *testing.T) {
...
```
qdm12, line 127: Run the parent test in parallel with

```suggestion
func TestVerifySignature(t *testing.T) {
	t.Parallel()
	
```

lib/runtime/life/resolver.go
```diff
@@ -933,11 +933,11 @@ func ext_crypto_ed25519_verify_version_1(vm *exec.VirtualMachine) int64 {
...
-		signature := runtime.Signature{
-			PubKey:    pubKey.Encode(),
-			Sign:      signature,
-			Msg:       message,
-			KeyTypeID: crypto.Ed25519Type,
+		signature := crypto.Signing{
+			PubKey:     pubKey.Encode(),
+			Sign:       signature,
+			Msg:        message,
+			VerifyFunc: ed25519.VerifySignature,
```
qdm12, line 940: The caller here should wrap the error with the signature type used. Since it's not aware of the signature type after this point, I suggest you use a wrapping helper function such as

```go
func verifySignatureWrapper(signatureType string, verifyFunc SigVerifyFunc) SigVerifyFunc {
	return func(publicKey, signature, message []byte) (err error) {
		err = verifyFunc(publicKey, signature, message)
		if err != nil {
			return fmt.Errorf("%s: %w", signatureType, err)
		}
		return nil
	}
}
```

which you can then use with

```suggestion
			VerifyFunc: verifySignatureWrapper("ed25519", ed25519.VerifySignature),
```

This applies to other fields in this file.

1garo: I don't see the necessity of using a wrapper here, each signature already returns it's sig type when an error occurs, for example, `ed25519`:

```go
pubKey, err := NewPublicKey(publicKey)
if err != nil {
     return err
}

_, err = pubKey.Verify(message, signature)
if err != nil {
     return fmt.Errorf("ed25519: %w: for message 0x%x, signature 0x%x and public key 0x%x",
	    crypto.ErrSignatureVerificationFailed, message, signature, publicKey)
}
```

In my mind this could be resolved just this, instead of the above:

```go
pubKey, err := NewPublicKey(publicKey)
if err != nil {
     return fmt.Errorf("ed25519: %w", err)
}
...
```

lib/crypto/sig_verifier_test.go
1garo: A had a little trouble trying to use test cases and sub tests in this case.

While doing it I had some problems to infer the keypair type, for example, to dynamically call `type.VerifySignature`, ended up looking more verbose than doing them apart.

lib/runtime/life/resolver.go
1garo: I don't see the necessity for using the wrapper here, since each signature already adds it's sign type when an error occurs, maybe i'm not seeing something

qdm12: Ok thinking about it, my suggestion makes it just stranger, so never mind!

But to explain myself, the `VerifySignature` of i.e. `ed25519` should not be responsible for wrapping the error with its signature type ideally, it should only be concerned on which parts of itself failed. You would also have to wrap every error returned inside VerifySignature with the prefix `ed25519: `.

For example I prefer

```go
func divide(a, b int) (result int, err error) {
  if b == 0 {
     return fmt.Errorf("cannot divide by zero")
  } else if b < 0 {
     return fmt.Errof("cannot divide with a negative number")
  }
  return a / b
}

func caller() error {
  result, err := divide(5, 2)
  if err != nil {
    return fmt.Errorf("divide: %w", err)
  }
}
```

compared to

```go
func divide(a, b int) (result int, err error) {
  if b == 0 {
     return fmt.Errorf("divide: cannot divide by zero")
  } else if b < 0 {
     return fmt.Errof("divide: cannot divide with a negative number")
  }
  return a / b
}

func caller() error {
  result, err := divide(5, 2)
  if err != nil {
    return err
  }
}
```

lib/crypto/sr25519/sr25519_test.go
```diff
@@ -5,8 +5,12 @@ package sr25519
 
 import (
 	"crypto/rand"
+	"errors"
+
```
qdm12, line 9: nit I don't think there should  be an empty line here. Although fine by me if the  CI allows it

```suggestion
```

lib/crypto/sr25519/sr25519.go
```diff
@@ -195,7 +212,7 @@ func (kp *Keypair) Private() crypto.PrivateKey {
 }
 
 // VrfSign creates a VRF output and proof from a message and private key
-func (kp *Keypair) VrfSign(t *merlin.Transcript) ([VRFOutputLength]byte, [VRFProofLength]byte, error) {
+func (kp *Keypair) VrfSign(t *merlin.Transcript) ([32]byte, [64]byte, error) {
```
qdm12, line 215: Why not use the constants here?

Maybe redefine them in a shared /lib/crypto/constants package?
Alternatively, please name the returns as

```suggestion
func (kp *Keypair) VrfSign(t *merlin.Transcript) (
	vfrOutput [32]byte, vfrProof [64]byte, err error) {
```

To serve as documentation, thanks!!

lib/runtime/life/resolver.go
1garo: ohhh now I get it, at first I thought that was something wrong with the code, but it was a suggestion, my bad haha 😅

lib/crypto/ed25519/ed25519.go
```diff
@@ -44,6 +44,22 @@ type PublicKey ed25519.PublicKey
...
 type PublicKeyBytes [PublicKeyLength]byte
 
+// VerifySignature verifies a signature given a public key and a message
+func VerifySignature(publicKey, signature, message []byte) error {
+	pubKey, err := NewPublicKey(publicKey)
+	if err != nil {
+		return fmt.Errorf("ed25519: %w", err)
+	}
+
+	_, err = pubKey.Verify(message, signature)
```
noot, line 54: you need to check the `bool` returned here, as the verification might fail, but there might not be an error

lib/crypto/sig_verifier.go
```diff
@@ -1,31 +1,32 @@
...
-	Sign      []byte
-	Msg       []byte
-	KeyTypeID crypto.KeyType
+var ErrSignatureVerificationFailed = errors.New("failed to verify signature")
+
+// SigVerifyFunc verifies a signature given a public key and a message
+type SigVerifyFunc func(pubkey, sig, msg []byte) (err error)
+
+// Signing ...
+type Signing struct {
```
noot, line 20: rename this `SignatureInfo` or something like that?

lib/crypto/sr25519/sr25519.go
```diff
@@ -48,6 +49,22 @@ type PrivateKey struct {
...
 }
 
+// VerifySignature verifies a signature given a public key and a message
+func VerifySignature(publicKey, signature, message []byte) error {
+	pubKey, err := NewPublicKey(publicKey)
+	if err != nil {
+		return fmt.Errorf("sr25519: %w", err)
+	}
+
+	_, err = pubKey.Verify(message, signature)
```
noot, line 59: same here - please check `bool` as the signature might be invalid, but no error is returned

lib/crypto/sr25519/sr25519.go
noot: why is this changed? please leave it the same as before

noot (changes requested):
thank you for the PR! just a few more changes please :)

lib/crypto/sr25519/sr25519.go
1garo: I always forget that we can document the code like this, it really improves understanding, appreciate. 
I'm gonna rollback to the constants, looks more clear and they are used in another files, so a shared pkg seems useful

lib/crypto/ed25519/ed25519.go
1garo: its okay to do something like this:

```go
ok, err := pubKey.Verify(message, signature)
    if err != nil || !ok {
        return fmt.Errorf("ed25519: %w: for message 0x%x, signature 0x%x and public key 0x%x",
	    crypto.ErrSignatureVerificationFailed, message, signature, publicKey)
    }
```

since we don't return a bool anymore?

lib/crypto/sr25519/sr25519.go
1garo: its okay to do something like this:

```go
ok, err := pubKey.Verify(message, signature)
    if err != nil || !ok {
        return fmt.Errorf("ed25519: %w: for message 0x%x, signature 0x%x and public key 0x%x",
	    crypto.ErrSignatureVerificationFailed, message, signature, publicKey)
    }
```

since we don't return a bool anymore?

lib/crypto/sig_verifier.go
1garo: for me `SignatureInfo` is okay

lib/crypto/sig_verifier_test.go
EclesioMeloJunior: here use a `map[string]struct{expect err, signaturesToVerify []crypto.SignatureInfo}` then you set manually a success case which has signatures and expect no err and a case error which has wrong signature info data and expects an error. To resolve the problem of keypair generation, I would suggest creating a keypair at the beginning of the function and then use them as you want while constructing the test case map

EclesioMeloJunior (changes requested):

lib/crypto/sig_verifier_test.go
1garo: thanks @EclesioMeloJunior for the hint on how to solve it, u are the best!

lib/crypto/sig_verifier_test.go
```diff
@@ -0,0 +1,132 @@
...
+	secpSign, err := secpKeypair.Sign(message)
+	require.NoError(t, err)
+
+	srKeypair, err := sr25519.GenerateKeypair()
+	require.NoError(t, err)
+	srSign, err := srKeypair.Sign(message)
+	require.NoError(t, err)
+
+	testCase := map[string]struct {
+		expect             error
```
EclesioMeloJunior, line 42: as the finish function just returns `true` or `false` so the `expect` field can be a boolean

lib/crypto/sig_verifier_test.go
```diff
@@ -0,0 +1,132 @@
...
+			signVerify := crypto.NewSignatureVerifier(log.New(log.SetWriter(io.Discard)))
+
+			for _, sig := range testCase.signaturesToVerify {
+				signVerify.Add(sig)
+			}
+
+			signVerify.Start()
+
+			ok := signVerify.Finish()
+			if testCase.expect != nil {
```
EclesioMeloJunior, line 123: and here you can do `require.Equal(t, testCase.expect, ok)`

EclesioMeloJunior (approved):
Just a small comment about the test assertion, but looks good to me! Great job @1garo !

lib/crypto/sig_verifier_test.go
1garo: done!

lib/crypto/sig_verifier_test.go
1garo: done!

lib/crypto/ed25519/ed25519.go
qdm12: Just have a single error check `err != nil` I guess, that should work.

lib/runtime/life/resolver.go
qdm12: No worry! Your code is good, it's just me being a bit of an error wrapping extremist :smile_cat:

lib/crypto/ed25519/ed25519.go
1garo: the `err != nil` verification already existed, what I did understand was that here we need to check the bool `i.e ok`, 'cause it can fail and `err` be `nil`

lib/crypto/sr25519/sr25519.go
1garo: rollback done in 7da55ab! @noot @qdm12

lib/crypto/ed25519/ed25519.go
1garo: done in da14c99

lib/crypto/secp256k1/secp256k1.go
```diff
@@ -36,6 +37,17 @@ type PublicKey struct {
...
 
+// VerifySignature verifies a signature given a public key and a message
+func VerifySignature(publicKey, signature, message []byte) error {
+	ok := secp256k1.VerifySignature(publicKey, message, signature)
+	if !ok {
+		return fmt.Errorf("secp256k1: %w: for message 0x%x, signature 0x%x and public key 0x%x",
+			crypto.ErrSignatureVerificationFailed, message, signature, publicKey)
+	}
+
+	return nil
```
qdm12, line 48: nit you could flip those around to have less indentation / line length

```suggestion
	if ok {
		return nil
	}

	return fmt.Errorf("secp256k1: %w: for message 0x%x, signature 0x%x and public key 0x%x",
		crypto.ErrSignatureVerificationFailed, message, signature, publicKey)
```

qdm12 (changes requested):
Thanks for the changes :+1: ! Just a few tiny things to modify please :pray:

lib/crypto/ed25519/ed25519.go
```diff
@@ -44,6 +44,22 @@ type PublicKey ed25519.PublicKey
...
 
+// VerifySignature verifies a signature given a public key and a message
+func VerifySignature(publicKey, signature, message []byte) error {
+	pubKey, err := NewPublicKey(publicKey)
+	if err != nil {
+		return fmt.Errorf("ed25519: %w", err)
+	}
+
+	ok, err := pubKey.Verify(message, signature)
+	if err != nil || !ok {
```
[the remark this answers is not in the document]
1garo: My suggestion would be:
```go
ok, err := pubKey.Verify(message, signature)
if err != nil {
  return fmt.Errorf("sr25519: %w: because %s", crypto.ErrSignatureVerificationFailed, err)
}

if !ok {
  return fmt.Errorf("sr25519: %w: for message 0x%x, signature 0x%x and public key 0x%x",
			crypto.ErrSignatureVerificationFailed, message, signature, publicKey)
}
```
since when `err != nil`, the cause of it will be explicit on `err`, e.g if one of those happen:
```go
if k.key == nil {
  return false, errors.New("nil public key")
}

if len(sig) != SignatureLength {
  return false, errors.New("invalid signature length")
}
```

and in the case where `err == nill && !ok`, the problem was really the the verification step, i.e:
```go
t := sr25519.NewSigningContext(SigningContext, msg)
return k.key.Verify(s, t), nil
```
The same would apply for `ed25518`.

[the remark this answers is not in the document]
qdm12: Sure, although don't wrap both errors with `crypto.ErrSignatureVerificationFailed` that looks strange. The parent calling function should wrap the error in this case. Instead you can just have

```go
ok, err := pubKey.Verify(message, signature)
if err != nil {
  return fmt.Errorf("sr25519: %w", err)
} else if !ok {
  return fmt.Errorf("sr25519: %w: for message 0x%x, signature 0x%x and public key 0x%x",
			crypto.ErrSignatureVerificationFailed, message, signature, publicKey)
}
```

?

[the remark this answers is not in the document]
1garo: looking now it really seems strange that way, changed it on 1ee5e82 4012a66

noot (approved):
looks good to me! thanks for the contribution :)

qdm12 (approved):
Awesome :medal_sports: Thanks for the contribution!!

lib/crypto/sr25519/sr25519_test.go
```diff
@@ -119,3 +122,63 @@ func TestNewKeypairFromMnenomic(t *testing.T) {
...
+				crypto.ErrSignatureVerificationFailed, []byte("a225e8c75da7da319af6335e7642d473"), signature, publicKey),
+		},
+	}
+
+	for name, value := range testCase {
+		testCase := value
+		t.Run(name, func(t *testing.T) {
+			t.Parallel()
+
+			err = VerifySignature(testCase.publicKey, testCase.signature, testCase.message)
```
kishansagathiya, line 174:
```suggestion
			err := VerifySignature(testCase.publicKey, testCase.signature, testCase.message)
```

kishansagathiya (approved):
:tada:

PR #2059, opened 2021-11-22
fix(dot/network): fix dht connection on discovery on devnet

### LINKED ISSUE #2026: fix DHT connection on discovery on devnet
## Task summary

- DHT discovery seems to have regressed on the devnet and no longer works
- previously, when running 3 nodes Alice, Bob and Charlie, Bob and Charlie would use Alice as their bootnode, and then discover each other via the DHT, resulting in each node having 2 peers
- however it seems to have regressed w/ the peer scoring update and this no longer happens (Alice has 2 peers, but Bob and Charlie only have 1, and they no longer connect to each other)
- it seems they still discover each other, but don't connect to each other, as adding in code to directly connect them after discovering results in the correct behaviour (ie. adding a line here to connect them, as can be seen in older commits)
https://github.com/ChainSafe/gossamer/blob/d7147cf4048de2ac41f5ef27462c20b10d4fe28f/dot/network/discovery.go#L212-L221

kishansagathiya:
## Changes

- We had a limit on number on ingoing and outgoing connections (maxOut and maxIn in peerset.Info). The limit on incoming connection was 2. Thus when we tried with a 3 node dev net, the first node was able to connect with both remaining nodes. Since, we were bootstrapping using address of the first node, it only received incoming connection. Since the limit there was 2 we did not see the problem.
- For remaining two nodes, they were able to connect with only one peer (first node), since they had to send outgoing connection request, limit for which was 1.
- For creating these limits we were using number of min peers and max peers from network config. This PR sets max peers in network config to 100 (I don't if that if an appropriate value). number of min peers was being used as limit for outgoing connection, I have changed that to number of max peers, so that nodes will be able to send a lot more outgoing requests. (Any arguments for why this should not be allowed?)

## Tests

Run a dev net with 3 gossamer nodes on different machines with public IP (not behind nat). Use  https://hackmd.io/@nZ-twauPRISEa6G9zg3XRw/B1RFO3uDt for that. From the logs (block import messages), you would see that all nodes have two peers.

## Issues

- Closes #2026 

## Primary Reviewer

- @noot


cmd/gossamer/config.go
```diff
@@ -121,6 +121,9 @@ func createDotConfig(ctx *cli.Context) (*dot.Config, error) {
 		return nil, err
 	}
 
+	// TODO: log this better. Show package wise log levels. Currently
+	// it prints something like `{INFO INFO INFO INFO INFO INFO INFO INFO}`
+	// which is not informative.
```
qdm12, line 126:
```suggestion
	// TODO: log this better
	// See https://github.com/ChainSafe/gossamer/issues/1945
```

cmd/gossamer/config.go
```diff
@@ -121,6 +121,9 @@ func createDotConfig(ctx *cli.Context) (*dot.Config, error) {
...
 	logger.Infof("loaded package log configuration: %v", cfg.Log)
```
qdm12, line 127: You can use for now to show the field names

```suggestion
	logger.Infof("loaded package log configuration: %#v", cfg.Log)
```

dot/network/discovery.go
```diff
@@ -150,7 +150,8 @@ func (d *discovery) advertise() {
 
 			ttl, err = d.rd.Advertise(d.ctx, string(d.pid))
 			if err != nil {
-				logger.Debugf("failed to advertise in the DHT: %s", err)
+				// TODO: This fails consistently.
```
qdm12, line 153: Please create and link up to an issue

dot/network/service.go
```diff
@@ -669,21 +669,21 @@ func (s *Service) processMessage(msg peerset.Message) {
...
-				logger.Debugf("failed to find peer id %s: %s", peerID, err)
+				logger.Errorf("failed to find peer id %s: %s", peerID, err)
 				return
 			}
 		}
 
 		err := s.host.connect(addrInfo)
 		if err != nil {
-			logger.Debugf("failed to open connection for peer %s: %s", peerID, err)
+			logger.Errorf("failed to open connection for peer %s, error: %s", peerID, err)
```
qdm12, line 679: To keep it consistent with other messages. Plus we already mention `failed ...` so it's kind of clear this didn't work :wink: 

```suggestion
			logger.Errorf("failed to open connection for peer %s: %s", peerID, err)
```

dot/peerset/peerset.go
```diff
@@ -348,6 +348,7 @@ func (ps *PeerSet) allocSlots(setIdx int) error {
 		}
 
 		if n.getReputation() < BannedThresholdValue {
+			logger.Warnf("reputation is lower than banned threshold value")
```
qdm12, line 351: nit let's add some additional context. Feel free to add more if you feel it's relevant!

```suggestion
			logger.Warnf(
				"reputation %d is lower than banned threshold value %d", 
				n.getReputation(), BannedThresholdValue)
```

dot/peerset/peerset.go
```diff
@@ -379,6 +382,7 @@ func (ps *PeerSet) allocSlots(setIdx int) error {
 		}
 
 		if err = peerState.tryOutgoing(setIdx, peerID); err != nil {
+			logger.Errorf("could not set peer as outgoing connection, peer: %s, error: %d", peerID.Pretty(), err)
```
qdm12, line 385:
```suggestion
			logger.Errorf("could not set peer %s as outgoing connection: %s", peerID.Pretty(), err)
```

dot/peerset/peerstate.go
```diff
@@ -57,8 +57,8 @@ type Info struct {
 
 // node represents state of a single node that we know about
 type node struct {
-	// list of Set the node belongs to.
-	// always has a fixed size equal to the one of PeersState Set. The various possible Set
+	// list of Set, the node belongs to.
```
qdm12, line 60: I'm a bit confused by this comment (before and after)... is `state` a list of sets the node belongs to?

dot/peerset/peerstate.go
```diff
@@ -57,8 +57,8 @@ type Info struct {
...
+	// always has a fixed size, equal to the one of PeersState Set. The various possible Set
```
qdm12, line 61: A bit out of scope, but this is confusing too

dot/peerset/peerstate.go
```diff
@@ -246,6 +246,8 @@ func (ps *PeersState) addNoSlotNode(idx int, peerID peer.ID) {
 	ps.sets[idx].noSlotNodes[peerID] = struct{}{}
 	n, err := ps.getNode(peerID)
 	if err != nil {
+		// TODO: Return the error
+		logger.Warnf("could not get node, error: %s", err)
```
qdm12, line 250: nit no need to mention it's an `error`, it should be obvious from the context preceding it

```suggestion
		logger.Warnf("could not get node: %s", err)
```

dot/peerset/peerstate.go
```diff
@@ -261,12 +263,15 @@ func (ps *PeersState) addNoSlotNode(idx int, peerID peer.ID) {
 
 func (ps *PeersState) removeNoSlotNode(idx int, peerID peer.ID) {
 	if _, ok := ps.sets[idx].noSlotNodes[peerID]; !ok {
+		logger.Debugf("peer %s already does not exist in no slot node", peerID)
```
qdm12, line 266: Why `already`? :thinking:

dot/peerset/peerstate.go
```diff
@@ -354,16 +359,13 @@ func (ps *PeersState) forgetPeer(set int, peerID peer.ID) error {
 }
 
 // tryOutgoing tries to set the peer as connected as an outgoing connection.
-// If there are enough slots available, switches the node to Connected and returns nil error. If
-// the slots are full, the node stays "not connected" and we return error.
+// If there are enough slots available, switches the node to Connected and returns
+// nil error. If the slots are full, the node stays "not connected" and we return error.
```
qdm12, line 363: out of scope, but whilst we're at it

```suggestion
// a nil error. If the slots are full, the node stays "not connected" and we return an error.
```

dot/peerset/peerstate.go
kishansagathiya: afaiu, yes

dot/peerset/peerstate.go
kishansagathiya: i think, it says the list is of fix size. we have four types of membership state, so there would be four sets.

dot/peerset/peerstate.go
kishansagathiya: we are trying to remove something, but it is already not there

dot/peerset/peerstate.go
qdm12: please then change to:

```suggestion
	// state is a list of sets the node belongs to.
```

dot/peerset/peerstate.go
qdm12: Maybe rephrase to

```suggestion
	// It always has the fixed size of the PeersState set. The various possible sets
```

dot/peerset/peerstate.go
qdm12: Do you mean it already left?
As in, `already does not exist` doesn't add up to me, it could `already exist` but not the other way around if you see what I mean? But that's just me supernitting though

dot/peerset/peerset.go
qdm12: Note the previous code contains `%d` for the error, so at least fix that part :wink:

dot/network/discovery.go
```diff
@@ -150,7 +150,8 @@ func (d *discovery) advertise() {
...
+				logger.Warnf("failed to advertise in the DHT: %s", err)
```
noot, line 154: i think we can probably remove the error log or change it do debug, as when you start the first node and it has no peers, this will always error

dot/network/service.go
```diff
@@ -669,21 +669,21 @@ func (s *Service) processMessage(msg peerset.Message) {
 			var err error
 			addrInfo, err = s.host.discovery.findPeer(peerID)
 			if err != nil {
-				logger.Debugf("failed to find peer id %s: %s", peerID, err)
+				logger.Errorf("failed to find peer id %s: %s", peerID, err)
```
noot, line 672: this would be a `Warn` level imo... or debug

dot/network/service.go
noot: would say this is a warn or debug

dot/network/service.go
```diff
@@ -669,21 +669,21 @@ func (s *Service) processMessage(msg peerset.Message) {
...
-			logger.Debugf("failed to open connection for peer %s: %s", peerID, err)
+			logger.Errorf("failed to open connection for peer %s, error: %s", peerID, err)
 			return
 		}
 		logger.Debugf("connection successful with peer %s", peerID)
 	case peerset.Drop, peerset.Reject:
 		err := s.host.closePeer(peerID)
 		if err != nil {
-			logger.Debugf("failed to close connection with peer %s: %s", peerID, err)
+			logger.Errorf("failed to close connection with peer %s: %s", peerID, err)
```
noot, line 686: same here

chain/gssmr/defaults.go
```diff
@@ -74,6 +74,8 @@ var (
 	DefaultNoMDNS = false
 	// DefaultMinPeers is the default minimum desired peer count
 	DefaultMinPeers = 1
+	// DefaultMaxPeers is the default maximum desired peer count
+	DefaultMaxPeers = 100
```
noot, line 78: change to 50 to match polkadot/kusama?

dot/network/host.go
```diff
@@ -88,7 +88,8 @@ func newHost(ctx context.Context, cfg *Config) (*host, error) {
 		return nil, err
 	}
 
-	peerCfgSet := peerset.NewConfigSet(uint32(cfg.MaxPeers-cfg.MinPeers), uint32(cfg.MinPeers), false, peerSetSlotAllocTime)
+	// TODO: What should be the right value of max in and max out?
+	peerCfgSet := peerset.NewConfigSet(uint32(cfg.MaxPeers-cfg.MinPeers), uint32(cfg.MaxPeers), false, peerSetSlotAllocTime)
```
noot, line 92: the right value would be (1/2 * max peers) I believe, as that would allow for half incoming and half outgoing

dot/peerset/peerstate.go
noot:
```suggestion
		logger.Debugf("peer %s is not in "no slot node" map", peerID)
```

what about something like this?

dot/peerset/peerstate.go
```diff
@@ -261,12 +263,15 @@ func (ps *PeersState) addNoSlotNode(idx int, peerID peer.ID) {
...
 func (ps *PeersState) removeNoSlotNode(idx int, peerID peer.ID) {
 	if _, ok := ps.sets[idx].noSlotNodes[peerID]; !ok {
+		logger.Debugf("peer %s already does not exist in no slot node", peerID)
 		return
 	}
 
 	delete(ps.sets[idx].noSlotNodes, peerID)
 	n, err := ps.getNode(peerID)
 	if err != nil {
+		// TODO: Return the error
```
noot, line 273: do you want to open an issue for this or do follow up PR to fix this?

noot (commented):
works for me :D nice work, just a few comments!

dot/peerset/peerstate.go
kishansagathiya: done

dot/network/discovery.go
qdm12: > this will always error

For some time until another node is up right? Maybe we can have it as info with something less dramatic than `failed` such as

```suggestion
				logger.Infof("cannot advertise in the DHT: %s", err)
```

dot/network/host.go
```diff
@@ -98,12 +98,15 @@ func newHost(ctx context.Context, cfg *Config) (*host, error) {
 		return nil, err
 	}
 
+	// TODO: What should be the right value of max in and max out?
 	const reservedOnly = false
 	peerCfgSet := peerset.NewConfigSet(
 		uint32(cfg.MaxPeers-cfg.MinPeers),
-		uint32(cfg.MinPeers),
+		uint32(cfg.MaxPeers/2),
```
qdm12, line 105: nit it would be nice to set this to some variables before passing them as arguments for readability. For example reading this without my IDE, I have no clue what `uint32(cfg.MaxPeers/2),` is or why it's maxpeers/2. That might apply to `uint32(cfg.MaxPeers-cfg.MinPeers),` as well.

dot/peerset/peerset.go
```diff
@@ -362,12 +364,14 @@ func (ps *PeerSet) allocSlots(setIdx int) error {
...
 		}
 	}
+
 	// nothing more to do if we're in reserved mode.
 	if ps.isReservedOnly {
 		return nil
 	}
 
 	for peerState.hasFreeOutgoingSlot(setIdx) {
+
```
qdm12, line 374: nit not really useful empty line imo

```suggestion
```

dot/peerset/peerset.go
```diff
@@ -380,6 +384,7 @@ func (ps *PeerSet) allocSlots(setIdx int) error {
 		}
 
 		if err = peerState.tryOutgoing(setIdx, peerID); err != nil {
+			logger.Errorf("could not set peer as outgoing connection, peer: %s, error: %s", peerID.Pretty(), err)
```
qdm12, line 387: nit make it short and human readable (aka left to right)

```suggestion
			logger.Errorf("could not set peer %s as outgoing connection: %s", peerID.Pretty(), err)
```

dot/peerset/peerset.go
```diff
@@ -402,7 +407,9 @@ func (ps *PeerSet) addReservedPeers(setID int, peers ...peer.ID) error {
 		}
 
 		ps.reservedNode[peerID] = struct{}{}
-		ps.peerState.addNoSlotNode(setID, peerID)
+		if err := ps.peerState.addNoSlotNode(setID, peerID); err != nil {
+			return err
```
qdm12, line 411: Feel free to add more context to the wrapping

```suggestion
			return fmt.Errorf("cannot add  no slot node: %w", err)
```

dot/peerset/peerset.go
```diff
@@ -418,7 +425,9 @@ func (ps *PeerSet) removeReservedPeers(setID int, peers ...peer.ID) error {
 		}
 
 		delete(ps.reservedNode, peerID)
-		ps.peerState.removeNoSlotNode(setID, peerID)
+		if err := ps.peerState.removeNoSlotNode(setID, peerID); err != nil {
+			return err
```
qdm12, line 429:
```suggestion
			return fmt.Errorf("cannot remove no slot node: %w", err)
```

dot/peerset/peerstate.go
```diff
@@ -258,17 +259,19 @@ func (ps *PeersState) addNoSlotNode(idx int, peerID peer.ID) {
...
 
 	ps.nodes[peerID] = n
+	return nil
 }
 
-func (ps *PeersState) removeNoSlotNode(idx int, peerID peer.ID) {
+func (ps *PeersState) removeNoSlotNode(idx int, peerID peer.ID) error {
 	if _, ok := ps.sets[idx].noSlotNodes[peerID]; !ok {
-		return
+		logger.Debugf("peer %s is not in \"no slot node\" map", peerID)
```
qdm12, line 267: nit use back ticks it looks PrEtTiEr

```suggestion
		logger.Debugf(`peer %s is not in "no slot node" map`, peerID)
```

kishansagathiya: addressed

dot/peerset/peerset.go
kishansagathiya: done

dot/peerset/peerset.go
kishansagathiya: addressed

dot/peerset/peerstate.go
kishansagathiya: it would rather keep this as it is. `Set` has a special meaning here.

dot/peerset/peerstate.go
```diff
@@ -249,7 +249,7 @@ func (ps *PeersState) addNoSlotNode(idx int, peerID peer.ID) error {
 	ps.sets[idx].noSlotNodes[peerID] = struct{}{}
 	n, err := ps.getNode(peerID)
 	if err != nil {
-		return fmt.Errorf("could not get node: %w", err)
+		return fmt.Errorf("could not get node for peer id %s: %w", peerID, err)
```
qdm12, line 252: :+1:

dot/peerset/test_helpers.go
```diff
@@ -44,7 +44,7 @@ func newTestPeerSet(t *testing.T, in, out uint32, bootNodes, reservedPeers []pee
 
 	handler.AddPeer(0, bootNodes...)
 	handler.AddReservedPeer(0, reservedPeers...)
-	time.Sleep(time.Millisecond * 4000)
+	time.Sleep(time.Millisecond * 100)
```
qdm12: are we sure lowering these sleep duration won't cause instabilities when testing with slow machines (:eyes: our CI)?

kishansagathiya: TBH, I think I have done that by mistake.
But now that I have added `discover` function in `AddReservedPeer`, we would not need a long delay.

dot/network/discovery.go
kishansagathiya: > i think we can probably remove the error log or change it do debug, as when you start the first node and it has no peers, this will always error

I just tried a dev network with 3 nodes. I saw it only in the beginning.

I will remove my comment `This fails consistently`.

We `advertise` in the DHT at regular intervals, so we shouldn't remove this log. During the first time it says that `there are no peers in the table` (so anyone looking at the logs, would know to avoid it). This warning does not show after that.

I would rather have this warning in case this fails for any other reason, than have that error hidden in the debug mode.

dot/network/host.go
kishansagathiya: added a comment

dot/peerset/peerstate.go
kishansagathiya: changed

dot/peerset/peerstate.go
kishansagathiya: changed

dot/network/discovery.go
kishansagathiya: removed

dot/peerset/peerstate.go
```diff
@@ -258,17 +260,19 @@ func (ps *PeersState) addNoSlotNode(idx int, peerID peer.ID) {
...
-		return
+		logger.Debugf("peer %s is not in no-slot node map", peerID)
+		return nil
 	}
 
 	delete(ps.sets[idx].noSlotNodes, peerID)
 	n, err := ps.getNode(peerID)
 	if err != nil {
-		return
+		return fmt.Errorf("could not get node: %w", err)
```
EclesioMeloJunior, line 275: just to keep consistent with the error in `addNoSlotNode` function 
```suggestion
		return fmt.Errorf("could not get node for peer id %s: %w", peerID, err)
```

EclesioMeloJunior (changes requested):

dot/peerset/peerstate.go
```diff
@@ -71,6 +72,7 @@ type node struct {
 }
 
 // newNode method to create a node with 0 Reputation at starting.
+// Here, n is number of sets.
```
qdm12, line 75:
```suggestion
// newNode creates a node with n number of sets and 0 reputation.
```

dot/peerset/peerstate.go
```diff
@@ -356,16 +361,13 @@ func (ps *PeersState) forgetPeer(set int, peerID peer.ID) error {
 }
 
 // tryOutgoing tries to set the peer as connected as an outgoing connection.
-// If there are enough slots available, switches the node to Connected and returns nil error. If
-// the slots are full, the node stays "not connected" and we return error.
+// If there are enough slots available, switches the node to Connected and returns nil.
+// If the slots are full, the node stays "not connected" and we return the error.
```
qdm12, line 365:
```suggestion
// If the slots are full, the node stays "not connected" and we return the error ErrOutgoingSlotsUnavailable.
```

qdm12 (approved):

dot/peerset/peerstate.go
```diff
@@ -237,17 +238,17 @@ func (ps *PeersState) hasFreeIncomingSlot(set int) bool {
 
 // addNoSlotNode adds a node to the list of nodes that don't occupy slots.
 // has no effect if the node was already in the group.
-func (ps *PeersState) addNoSlotNode(idx int, peerID peer.ID) {
+func (ps *PeersState) addNoSlotNode(idx int, peerID peer.ID) error {
```
EclesioMeloJunior, line 241: could we add a condition `if idx >= len(ps.sets)` then return an error? it could not happen but just to adopt a kind of defensive programming.

dot/peerset/peerstate.go
```diff
@@ -258,17 +259,19 @@ func (ps *PeersState) addNoSlotNode(idx int, peerID peer.ID) {
 	}
```
EclesioMeloJunior, line 259: just a question: `numIn` and `numOut` can achieve negative values? @arijitAD

dot/peerset/peerstate.go
```diff
@@ -258,17 +259,19 @@ func (ps *PeersState) addNoSlotNode(idx int, peerID peer.ID) {
...
 
 	ps.nodes[peerID] = n
+	return nil
 }
 
-func (ps *PeersState) removeNoSlotNode(idx int, peerID peer.ID) {
+func (ps *PeersState) removeNoSlotNode(idx int, peerID peer.ID) error {
 	if _, ok := ps.sets[idx].noSlotNodes[peerID]; !ok {
-		return
+		logger.Debugf("peer %s is not in no-slot node map", peerID)
```
EclesioMeloJunior, line 267: same condition to check if the `idx` is in the boundaries of `ps.sets`

EclesioMeloJunior (dismissed):

dot/peerset/peerstate.go
kishansagathiya: numIn: number of connections where other nodes connected to us
numOut: number of connections where we connected to other nodes

don't think these values could be negative

EclesioMeloJunior: ok, should we add some sort of check to avoid these values being negatives?

dot/network/discovery.go
noot: I'm fine to leave it as-is for now

noot (approved):

PR #2060, opened 2021-11-22
fix(lib/blocktree): fix blocktree bug 

### LINKED ISSUE #2050: devnet: blocktree bug causes finality to fail (and possibly other things)
## Task summary

- on the devnet, I saw the following error:
```
2021-11-17T23:08:11Z WARN failed to play grandpa round: no common ancestor between two nodes: 0xf48c9edaf2da5ec19edb7ced855158d295ad00ed2108cec4c8e22a87e2b86d33 and 0xcf2b5f5b59def5e4df6ff2237cca1fda5e5d114c20e796a90a58ace1bd91c20d	grandpa.go:L346	pkg=grandpa
```

this error comes from the blocktree function `HighestCommonAncestor` which returns the common ancestor between two blocktree nodes with the highest number. however this error should *NOT* happen, as explained by the comment above the error:

```
	ancestor := an.highestCommonAncestor(bn)
	if ancestor == nil {
		// this case shouldn't happen - any two nodes in the blocktree must
		// have a common ancestor, the lowest of which is the root node
		return common.Hash{}, fmt.Errorf("%w: %s and %s", ErrNoCommonAncestor, a, b)
	}
```

when this function returns an error it causes grandpa to exit as it's an unexpected err, then finality stops.

fix the blocktree so this does not happen!!

noot:
## Changes

- update `leafMap` locking
- remove `nodeCache` as it was buggy and basically unnecessary, as in a healthy network the blocktree shouldn't grow very deep, probably max depth ~10

## Tests

```
go test ./lib/blocktree
```

## Issues

- closes #2050

## Primary Reviewer

- @jimjbrettj


lib/blocktree/leaves.go
```diff
@@ -48,13 +49,18 @@ func (ls *leafMap) load(key Hash) (*node, error) {
 
 // Replace deletes the old node from the map and inserts the new one
 func (ls *leafMap) replace(oldNode, newNode *node) {
```
timwu20, line 51:
```suggestion
func (lm *leafMap) replace(oldNode, newNode *node) {
```

lib/blocktree/leaves.go
```diff
@@ -13,6 +13,7 @@ import (
 
 // leafMap provides quick lookup for existing leaves
 type leafMap struct {
+	sync.RWMutex
```
timwu20, line 16: why do we need this mutex if it's a `sync.Map`?

noot: there's no way to atomically delete one key from a map and then add another key, eg. in `Replace`:
```
// Replace deletes the old node from the map and inserts the new one
func (ls *leafMap) replace(oldNode, newNode *node) {
	ls.Lock()
	defer ls.Unlock()
	ls.smap.Delete(oldNode.hash)
	ls.store(newNode.hash, newNode)
}
```

it needs to atomically do the replace, otherwise when other functions read the leaf map there might not see either the old node or the new node

lib/blocktree/leaves.go
noot: done

lib/blocktree/blocktree.go
```diff
@@ -49,34 +47,25 @@ func NewBlockTreeFromRoot(root *types.Header) *BlockTree {
...
-		nodeCache: make(map[Hash]*node),
-		runtime:   &sync.Map{},
+		root:    n,
+		leaves:  newLeafMap(n),
+		runtime: &sync.Map{},
 	}
 }
 
-// GenesisHash returns the hash of the genesis block
-func (bt *BlockTree) GenesisHash() Hash {
```
jimjbrettj, line 60: Is this function just not something we need anymore?

lib/blocktree/blocktree.go
```diff
@@ -49,34 +47,25 @@ func NewBlockTreeFromRoot(root *types.Header) *BlockTree {
...
-	defer bt.RUnlock()
-	return bt.root.hash
-}
-
 // AddBlock inserts the block as child of its parent node
 // Note: Assumes block has no children
 func (bt *BlockTree) AddBlock(header *types.Header, arrivalTime time.Time) error {
-	bt.Lock()
-	defer bt.Unlock()
+	// bt.Lock()
```
jimjbrettj, line 59: Are these commented out because they were added to the replace function instead of here?

EclesioMeloJunior (changes requested):

lib/blocktree/blocktree.go
```diff
@@ -49,34 +47,25 @@ func NewBlockTreeFromRoot(root *types.Header) *BlockTree {
...
+	// defer bt.Unlock()
```
EclesioMeloJunior, line 60: should this code be removed?

EclesioMeloJunior (changes requested):

lib/blocktree/blocktree.go
```diff
@@ -87,17 +76,16 @@ func (bt *BlockTree) AddBlock(header *types.Header, arrivalTime time.Time) error
...
-	n = &node{
+	n := &node{
 		hash:        header.Hash(),
 		parent:      parent,
 		children:    []*node{},
 		number:      number,
 		arrivalTime: arrivalTime,
 	}
+
 	parent.addChild(n)
```
jimjbrettj, line 87: Want to make sure I understand here, so we are okay with concurrent calls to addChild here because whichever acquires the lock will perform the replace, then that state change to the leafMap would invalidate the other replace requests if its the same parent. Is this right? 

Sorry if that was confusing, I'm trying to reason about moving the lock from add into replace

lib/blocktree/blocktree_test.go
```diff
@@ -458,38 +453,15 @@ func TestBlockTree_Prune(t *testing.T) {
...
 func TestBlockTree_GetHashByNumber(t *testing.T) {
 	bt, _ := createTestBlockTree(t, testHeader, 8)
 	best := bt.DeepestBlockHash()
-	bn := bt.nodeCache[best]
+	bn := bt.getNode(best)
 
 	for i := int64(0); i < bn.number.Int64(); i++ {
 		hash, err := bt.GetHashByNumber(big.NewInt(i))
 		require.NoError(t, err)
-		require.Equal(t, big.NewInt(i), bt.nodeCache[hash].number)
```
jimjbrettj, line 492: So was the overhead of maintaining the cache just not worth it in this situation? I suppose this means that lookup times in the tree will be quick enough that its better to just lookup than have to maintain the cache?

lib/grandpa/message_handler_test.go
```diff
@@ -628,6 +629,56 @@ func TestMessageHandler_VerifyBlockJustification(t *testing.T) {
 	require.NoError(t, err)
 	err = gs.VerifyBlockJustification(testHash, data)
 	require.NotNil(t, err)
+	require.Equal(t, blocktree.ErrEndNodeNotFound, err)
+}
+
+func TestMessageHandler_VerifyBlockJustification_invali(t *testing.T) {
```
jimjbrettj, line 635: NIT: Should make this TestMessageHandler_VerifyBlockJustification_invalid

jimjbrettj (commented):
Looks good, just a few questions/comments!

lib/grandpa/message_handler_test.go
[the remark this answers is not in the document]
jimjbrettj: I commented same thing oops, will resolve mine. But good suggestion

lib/blocktree/blocktree.go
noot: yeah it was unused

lib/blocktree/blocktree.go
noot: omg, i didn't mean to comment this out, re-added

lib/blocktree/blocktree.go
noot: no i didnt mean to comment it

lib/blocktree/blocktree.go
noot: sorry the lock was not supposed to be removed from AddBlock lol, but in general the functions that are exported need to lock and the internal ones don't as they should only be called by functions that lock

lib/blocktree/blocktree_test.go
noot: yeah I just don't think it was necessary, the blocktree shouldn't ever grow too deep as we finalize blocks relatively quickly

lib/blocktree/blocktree.go
jimjbrettj: Ahhh okay that answers the question then lol, thanks!

lib/grandpa/message_handler_test.go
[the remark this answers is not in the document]
noot: updated!

jimjbrettj (approved):
lgtm!!

qdm12 (approved):

lib/blocktree/blocktree.go
```diff
@@ -23,17 +23,15 @@ type BlockTree struct {
...
 // NewEmptyBlockTree creates a BlockTree with a nil head
 func NewEmptyBlockTree() *BlockTree {
 	return &BlockTree{
-		root:      nil,
-		leaves:    newEmptyLeafMap(),
-		nodeCache: make(map[Hash]*node),
-		runtime:   &sync.Map{}, // map[Hash]runtime.Instance
+		root:    nil,
+		leaves:  newEmptyLeafMap(),
+		runtime: &sync.Map{}, // map[Hash]runtime.Instance
```
kishansagathiya, line 34: better to put this comment in definition of BlockTree. I am assuming that this map would always have `map[Hash]runtime.Instance` type

kishansagathiya (commented):
The code looks good.

It is not clear to me what was causing `no common ancestor between two nodes`. How can I replicate this error?

noot:
@kishansagathiya `no common ancestor between two nodes` error should never happen, the case that was causing it to happen was because the node cache was not cleared properly. to create the issue in a unit test, you could create a blocktree w/ a root node, then add some leaf node that doesn't have the root node as an ancestor. then if you call `HighestCommonAncestor` passing in those two nodes, you will see the error as there is no common ancestor between them.

PR #2067, opened 2021-11-24
maintenance(docs): update go version in readme

noot:
## Changes

- update go version in readme to 1.17

## Tests

```
n/a
```

## Issues

- n/a

## Primary Reviewer

- @qdm12

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,7 +43,7 @@ For more information about Gossamer, the Polkadot ecosystem, and how to use Goss
 
 ### Prerequisites
 
-install go version `>=1.16`
+install go version `>=1.17`
 
 ### Installation
 
--- a/docs/docs/getting-started/installation.md
+++ b/docs/docs/getting-started/installation.md
@@ -8,7 +8,7 @@ permalink: /getting-started/installation
 
 ## Prerequisites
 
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -43,7 +43,7 @@ For more information about Gossamer, the Polkadot ecosystem, and how to use Goss
 
 ### Prerequisites
 
-install go version `>=1.16`
+install go version `>=1.17`
```
qdm12, line 46: I put 1.16 on purpose, I think it should work still.
We use `io` and `embed` from 1.16, however I realised our go.mod got 1.17-updated so maybe it doesn't work with 1.16 :thinking: They sort of broke compatibility with that go.mod change when I think about it...

noot: go 1.16 doesn't work anymore as we use `time.UnixMilli` which is only in go 1.17

qdm12: Ok so actually, the go.mod still works on 1.16 but we do use `time.UnixMilli` which is only here with 1.17. APPROVED!!

qdm12 (approved):

kishansagathiya (approved):
Change the version here as well before merging

https://github.com/ChainSafe/gossamer/blob/0bc4bf1879f35fbce61d2914dbe0feda21f5651d/docs/docs/getting-started/installation.md#L11

Issue #2039, opened 2021-11-12
update peer scoring to prevent us from going over max peers, update network ConnManager `Connected` not to handle that


noot:
## Task summary

- currently in the `ConnManager.Connected()` function, which is called when a new connection is opened, it checks whether we are over our max peers, and if so randomly disconnects from a peer
- this should be handled by the peer scoring module instead (so the lowest scored peer is removed), but if `Connected()` is updated not to perform that functionality, we go over our max peer count (by a significant amount), can be seen by running a polkadot or kusama node

Issue #1886, opened 2021-10-13
Add multiple peerset config for each protocol in Peer scoring.


arijitAD:
## Issue summary

- Add multiple peerset config for each protocol in Peer scoring

## Other information and links

- https://github.com/ChainSafe/gossamer/pull/1791#discussion_r728006472

PR #2068, opened 2021-11-24
fix(dot/network): move low reputation peer removal from network ConnManager to peer scoring logic (dot/peerstate)

kishansagathiya:
## Changes

- moves low reputation peer removal from network ConnManager to peer scoring logic (dot/peerstate)
- After studying the code I realized that peer scoring logic was already taking care of not going above max peer count. But it was not working as expected.
- peerset service (doWorks goroutine) sends messages to network service
about peers to connect, drop, reject, disconnect etc. Some of those
message were being send without a peer id and set id. This was the
reason why our excess peers were not getting removed.
- To save ourselves from such a problem in future, processMessage now
checks if peer id is empty or not.
- Also change the resultMsgCh type to peerset.Message from interface{}

## Tests

```
```
check block import logs to see number of peers. Default number of max peer is 50. Expected behaviour is that, any point peers should not go above 50. Mostly it would not increase after 45. 
## Issues

- #2039 

## Primary Reviewer

- @noot


dot/peerset/peerset.go
```diff
@@ -183,6 +183,8 @@ func NewConfigSet(in, out uint32, reservedOnly bool, allocTime time.Duration) *C
 	}
 
 	return &ConfigSet{
+		// Why are we using an array of config in the set, when we are
+		// using just one config
```
noot, line 187: #1886 @arijitAD could you give more context on this?

if I remember correctly it was so we could score the different stream protocols differently (eg. sync, block-announce, etc)

dot/peerset/peerset.go
```diff
@@ -546,11 +553,24 @@ func (ps *PeerSet) incoming(setID int, peers ...peer.ID) error {
...
 		case state.tryAcceptIncoming(setID, pid) != nil:
-			ps.resultMsgCh <- Message{Status: Reject}
+			ps.resultMsgCh <- Message{
+				Status: Reject,
+				setID:  uint64(setID),
+				PeerID: pid,
+			}
 		default:
-			ps.resultMsgCh <- Message{Status: Accept}
+			logger.Infof("incoming connection accepted from peer %s", pid)
```
noot, line 568:
```suggestion
			logger.Debugf("incoming connection accepted from peer %s", pid)
```

noot (approved):
works for me! great work, just update the log level please :p

dot/peerset/peerset.go
kishansagathiya: addressed

dot/network/service.go
```diff
@@ -670,6 +669,9 @@ func (s *Service) startPeerSetHandler() {
 
 func (s *Service) processMessage(msg peerset.Message) {
 	peerID := msg.PeerID
+	if peerID == "" {
+		logger.Errorf("found empty peer id in peerset message")
```
qdm12, line 673: should we return here?

EclesioMeloJunior (approved):

dot/network/service.go
kishansagathiya: No, processMessage is used to execute network tasks. If peer id is empty, that means we cannot execute that particular task. We should continue to process other tasks. If we get an empty message, that mean we have written wrong code (somewhere we are sending mesage without peerid).

This check is a precautionary measure.

qdm12: > If peer id is empty, that means we cannot execute that particular task. We should continue to process other tasks.

What do you mean by 'particular task'? In my view, `processMessage(msg)` is that particular task, and we should return if the peer id is empty right?

kishansagathiya: Oh yes, my bad! Thanks for spotting this! Also, Fixed!

Issue #1974, opened 2021-11-03
Improve linting to detect potential memory problems


qdm12:
- [x] Check `// nolint` comments
  - Check it makes sense to ignore
  - Replace each with `//nolint:<linter>`
  - Add `nolintlint` linter so outdated `//nolint` comments are detected
- [x] Add linters `nilerr`, `nilnil` and `exportloopref`

noot:
@qdm12 is this done?

qdm12:
First part is done in #2047, I just did the second part in #2072

qdm12:
Done in #2047 and #2072

PR #2072, opened 2021-11-26
chore(lint): add `nilnil`, `nilerr` and `exportloopref` linters

qdm12:
## Changes

- Add `nilnil`, nilerr` and `exportloopref` linters
- Fix lint errors whenever possible or add `//nolint` comments when not applicable

## Tests

```
golangci-lint run
```

And CI to pass

## Issues

- #1974 

## Primary Reviewer

### DIFF
--- a/dot/state/grandpa.go
+++ b/dot/state/grandpa.go
@@ -233,13 +233,11 @@ func (s *GrandpaState) SetNextPause(number *big.Int) error {
 	return s.db.Put(pauseKey, number.Bytes())
 }
 
-// GetNextPause returns the block number of the next grandpa pause, nil if there is no upcoming pause
+// GetNextPause returns the block number of the next grandpa pause.
+// It returns an error on failure, which is contains the
+// chaindb.ErrKeyNotFound error if the key is not found.
 func (s *GrandpaState) GetNextPause() (*big.Int, error) {
 	num, err := s.db.Get(pauseKey)
-	if err == chaindb.ErrKeyNotFound {
-		return nil, nil
-	}
-
 	if err != nil {
 		return nil, err
 	}
@@ -252,12 +250,11 @@ func (s *GrandpaState) SetNextResume(number *big.Int) error {
 	return s.db.Put(resumeKey, number.Bytes())
 }
 
-// GetNextResume returns the block number of the next grandpa resume, nil if there is no upcoming resume
+// GetNextResume returns the block number of the next grandpa resume.
+// It returns an error on failure, which is contains the
+// chaindb.ErrKeyNotFound error if the key is not found.
 func (s *GrandpaState) GetNextResume() (*big.Int, error) {
 	num, err := s.db.Get(resumeKey)
-	if err == chaindb.ErrKeyNotFound {
-		return nil, nil
-	}
 	if err != nil {
 		return nil, err
 	}
--- a/dot/state/grandpa.go
+++ b/dot/state/grandpa.go
@@ -234,11 +234,14 @@ func (s *GrandpaState) SetNextPause(number *big.Int) error {
 }
 
 // GetNextPause returns the block number of the next grandpa pause.
-// It returns an error on failure, which is contains the
-// chaindb.ErrKeyNotFound error if the key is not found.
+// If the key is not found in the database, a nil block number is returned
+// to indicate there is no upcoming Grandpa pause.
+// It returns an error on failure.
 func (s *GrandpaState) GetNextPause() (*big.Int, error) {
 	num, err := s.db.Get(pauseKey)
-	if err != nil {
+	if errors.Is(err, chaindb.ErrKeyNotFound) {
+		return nil, nil //nolint:nilnil
+	} else if err != nil {
 		return nil, err
 	}
 
@@ -251,11 +254,14 @@ func (s *GrandpaState) SetNextResume(number *big.Int) error {
 }
 
 // GetNextResume returns the block number of the next grandpa resume.
-// It returns an error on failure, which is contains the
-// chaindb.ErrKeyNotFound error if the key is not found.
+// If the key is not found in the database, a nil block number is returned
+// to indicate there is no upcoming Grandpa resume.
+// It returns an error on failure.
 func (s *GrandpaState) GetNextResume() (*big.Int, error) {
 	num, err := s.db.Get(resumeKey)
-	if err != nil {
+	if errors.Is(err, chaindb.ErrKeyNotFound) {
+		return nil, nil //nolint:nilnil
+	} else if err != nil {
 		return nil, err
 	}
 
--- a/dot/sync/bootstrap_syncer.go
+++ b/dot/sync/bootstrap_syncer.go
@@ -5,6 +5,7 @@ package sync
 
 import (
 	"errors"
+	"fmt"
 	"math/big"
 
 	"github.com/ChainSafe/gossamer/dot/network"
@@ -26,6 +27,8 @@ func newBootstrapSyncer(blockState BlockState) *bootstrapSyncer {
 	}
 }
 
+var errNoWorker = errors.New("no worker")
+
 func (s *bootstrapSyncer) handleNewPeerState(ps *peerState) (*worker, error) {
 	head, err := s.blockState.BestBlockHeader()
 	if err != nil {
@@ -33,7 +36,9 @@ func (s *bootstrapSyncer) handleNewPeerState(ps *peerState) (*worker, error) {
 	}
 
 	if ps.number.Cmp(head.Number) <= 0 {
-		return nil, nil
+		return nil, fmt.Errorf(
+			"%w: for head number %s and peer state number %s",
+			errNoWorker, head.Number, ps.number)
 	}
 
 	return &worker{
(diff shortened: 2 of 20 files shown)


dot/state/grandpa.go
```diff
@@ -233,13 +233,11 @@ func (s *GrandpaState) SetNextPause(number *big.Int) error {
...
-// GetNextPause returns the block number of the next grandpa pause, nil if there is no upcoming pause
+// GetNextPause returns the block number of the next grandpa pause.
+// It returns an error on failure, which is contains the
+// chaindb.ErrKeyNotFound error if the key is not found.
 func (s *GrandpaState) GetNextPause() (*big.Int, error) {
 	num, err := s.db.Get(pauseKey)
-	if err == chaindb.ErrKeyNotFound {
-		return nil, nil
-	}
-
```
noot, line 242: we shouldn't remove this, we should ignore this error

dot/state/grandpa.go
```diff
@@ -252,12 +250,11 @@ func (s *GrandpaState) SetNextResume(number *big.Int) error {
...
 
-// GetNextResume returns the block number of the next grandpa resume, nil if there is no upcoming resume
+// GetNextResume returns the block number of the next grandpa resume.
+// It returns an error on failure, which is contains the
+// chaindb.ErrKeyNotFound error if the key is not found.
 func (s *GrandpaState) GetNextResume() (*big.Int, error) {
 	num, err := s.db.Get(resumeKey)
-	if err == chaindb.ErrKeyNotFound {
-		return nil, nil
-	}
```
noot, line 260: same here

dot/sync/bootstrap_syncer.go
```diff
@@ -26,14 +27,18 @@ func newBootstrapSyncer(blockState BlockState) *bootstrapSyncer {
...
 	head, err := s.blockState.BestBlockHeader()
 	if err != nil {
 		return nil, err
 	}
 
 	if ps.number.Cmp(head.Number) <= 0 {
-		return nil, nil
+		return nil, fmt.Errorf(
+			"%w: for head number %s and peer state number %s",
+			errNoWorker, head.Number, ps.number)
```
noot, line 41: please revert this, this case isn't an error

dot/sync/bootstrap_syncer.go
```diff
@@ -45,21 +50,22 @@ func (s *bootstrapSyncer) handleNewPeerState(ps *peerState) (*worker, error) {
 	}, nil
 }
 
-func (s *bootstrapSyncer) handleWorkerResult(res *worker) (*worker, error) {
+func (s *bootstrapSyncer) handleWorkerResult(res *worker) (
+	workerToRetry *worker, retry bool, err error) {
```
noot, line 54: whats the point of the `retry bool`? it seems if the workerToRetry isn't nil then retry is always true and same for the other way around

lib/babe/errors.go
```diff
@@ -66,6 +66,7 @@ var (
 	errNoEpochData           = errors.New("no epoch data found for upcoming epoch")
 	errFirstBlockTimeout     = errors.New("timed out waiting for first block")
 	errChannelClosed         = errors.New("block notifier channel was closed")
+	errPrimarySlotThreshold  = errors.New("failed checking primary threshold")
```
noot, line 69:
```suggestion
	errOverPrimarySlotThreshold  = errors.New("cannot claim slot, over primary threshold")
```

dot/sync/bootstrap_syncer.go
qdm12: It's just to satisfy the `nilnil` linter. It complains whenever we return an invalid result (`nil` worker) and a nil error.

I can revert it and ignore the linter for this particular case if you want though.

dot/sync/bootstrap_syncer.go
qdm12: It's just to satisfy the `nilnil` linter. It complains whenever we return an invalid result (`nil` worker) and a nil error.

I changed the caller code to check for this error instead of returning `nil, nil` here. But up to you, I can revert it as well.

dot/state/grandpa.go
qdm12: I changed the callers to check for `chaindb.ErrKeyNotFound` and adapt their behavior.
Should I still revert this and nolint this linter?

dot/state/grandpa.go
qdm12: I changed the callers to check for `chaindb.ErrKeyNotFound` and adapt their behavior.
Should I still revert this and nolint this linter?

dot/sync/bootstrap_syncer.go
qdm12: Wellll that's up for debate. We can also use `//nolint:nilnil`, but the point of the linter is also to avoid any confusion, so returning an extra var like `retry bool` can sort of clarify things (although it's technically not necessary for it to work)

dot/state/grandpa.go
qdm12: If you check the callers of this function, when they get a `nil` error and a `nil` worker, it continues in their for loops.
So techically, it's true this is not really an error in the eyes of the current callers, although it could be in a future eventual caller :thinking:

dot/sync/bootstrap_syncer.go
noot: I think it's cleaner in this case to just put a `//nolint` as it seems cleaner to me

dot/state/grandpa.go
noot: @kishansagathiya we can ignore this because if the error is `KeyNotFound`, then there is no grandpa authorities pause coming up, so we don't need to do anything.

@qdm12  ok, as long as the logic is the same, that should be fine then

dot/sync/bootstrap_syncer.go
noot: I feel like errors should only be returned when there's actually an error happening, in this case, the fact that there's no worker isn't an error, it's expected, as we don't need to do anything if receiving a peer state below our current head state. definitely up for debate though i guess lol

dot/state/grandpa.go
noot: ok, same as above, if the caller logic checks it should be ok.

dot/state/grandpa.go
qdm12: The fact @kishansagathiya asked the question of why we ignore the error is also a good point we shouldn't discard error and return i.e. `nil, nil` since it can cause confusion.

Alternatively, instead of returning the not found error, we can change the signature from

```go
GetNextPause() (blockNumber *big.Int, err error)
```

to

```go
GetNextPause() (blockNumber *big.Int, exists bool, err error)
```

? :thinking:

lib/common/types/result.go
```diff
@@ -53,8 +54,10 @@ func (r *Result) Decode(reader io.Reader) (*Result, error) {
 
 	for {
 		b, err := common.ReadByte(reader)
-		if err != nil {
+		if errors.Is(err, io.EOF) {
```
qdm12, line 57: @noot about that one, if there is no data left in the reader, should we error out or just break and leave data empty? :thinking:

tests/utils/gossamer_utils.go
```diff
@@ -285,10 +285,11 @@ func InitNodes(num int, config string) ([]*Node, error) {
 
 // StartNodes starts given array of nodes
 func StartNodes(t *testing.T, nodes []*Node) error {
-	for _, n := range nodes {
+	for i, n := range nodes {
 		err := StartGossamer(t, n, false)
 		if err != nil {
-			return nil
+			return fmt.Errorf("node %d of %d: %w",
+				i+1, len(nodes), err)
```
qdm12, line 292: I guess that's fine to return an error here if we fail to start gossamer in a test right?

lib/babe/errors.go
qdm12: Done in https://github.com/ChainSafe/gossamer/pull/2072/commits/5848eaff5e92c1431a2a601ca79bd8d06fe540b0

dot/sync/bootstrap_syncer.go
qdm12: Reverted back in https://github.com/ChainSafe/gossamer/pull/2072/commits/de7f0973d2a9492f59cc349002fc4d75835f00d9

dot/sync/bootstrap_syncer.go
qdm12: Reverted back in https://github.com/ChainSafe/gossamer/pull/2072/commits/7cba22899d39e6e99ce0161c38798c07720c3c8d

dot/state/grandpa.go
kishansagathiya: I would rather just ignore the error

noot: can we just add a comment saying `if the key is not found in the database, that's fine as it means there is no upcoming grandpa pause`

lib/common/types/result.go
noot: i think we can just break if there's no data left

tests/utils/gossamer_utils.go
noot: yeah that's fine

dot/state/grandpa.go
qdm12: Actually `GetNextResume` and `GetNextPause` are both only used in the test `TestHandler_GrandpaPauseAndResume`. Hence the caller code wasn't changed, my bad. Anyway, I reverted it to return `nil, nil` for both and added a not nil check in the test.

noot (approved):

kishansagathiya (approved):
:tada:

Issue #1939, opened 2021-10-28
Remove uint128.go from lib/common Package


danforbes:

kishansagathiya:
Can you provide more info here?
uint128.go is being used. I think we want to replace with something else. Is that the case? Is it decided, what it should be replaced with, or is it to be researched?

timwu20:
I moved this to `pkg/scale`, we just need to update the imports and adjust the syntax.

kishansagathiya:
Closed in https://github.com/ChainSafe/gossamer/pull/2074

PR #2074, opened 2021-11-26
chore(lib/common): Use Uint128 from scale instead of lib/common

kishansagathiya:
## Changes

- Removed uint128.go
- Replaced common.Uint128 with scale.Uint128

## Tests

`make gossamer` and `go test ./lib/babe` should be enough

## Issues

- #1939

## Primary Reviewer

- @timwu20


noot (approved):
looks good!

timwu20 (approved):
LGTM!

lib/babe/crypto.go
```diff
@@ -58,25 +62,28 @@ func claimPrimarySlot(randomness Randomness,
...
-) bool {
+) (bool, error) {
 	t := makeTranscript(randomness, slot, epoch)
 	inout := sr25519.AttachInput(output, pub, t)
 	res := sr25519.MakeBytes(inout, 16, babeVRFPrefix)
 
-	inoutUint := common.Uint128FromLEBytes(res)
+	inoutUint, err := scale.NewUint128(res)
+	if err != nil {
+		return false, fmt.Errorf("failed to convert bytes to Uint128, %w", err)
```
qdm12, line 74: nit I know it's a super detail, but should we agree on a common way to wrap errors? I tend to use `context: %w` which seems like the common standard out there, but I'm happy to use any wrapping, as long as we make it the same in this codebase.

qdm12 (approved):
:+1: Nice!

lib/babe/crypto.go
kishansagathiya: will change it :+1:

kishansagathiya: done

PR #2075, opened 2021-11-26
fix(lib/babe, lib/runtime/wasmer): fixes for v0.9.8+ runtime

noot:
## Changes

- update BABE timestamp inherent to be in milliseconds instead of seconds
- fix `ext_storage_clear_prefix_version_2` by checking if the limit is `None` instead of `Some`

## Tests

```
go test ./lib/babe
go test ./lib/runtime/wasmer
```

or, feel free to try the instructions here on updating the runtime using the front-end https://hackmd.io/@nZ-twauPRISEa6G9zg3XRw/BJh5NvYdF I can send you the runtime I used as well if you'd like

## Issues

- related to #1688 

## Primary Reviewer

- @EclesioMeloJunior


dot/rpc/modules/author.go
```diff
@@ -76,8 +76,8 @@ type decodedKey struct {
 type ExtrinsicStatus struct {
 	IsFuture    bool
 	IsReady     bool
-	Isfinalised bool
-	Asfinalised common.Hash
+	IsFinalized bool
+	AsFinalized common.Hash
```
qdm12, line 80: Are we certain about the non UK spelling?

lib/babe/build.go
```diff
@@ -290,7 +290,8 @@ func (b *BlockBuilder) buildBlockExtrinsics(slot Slot, rt runtime.Instance) []*t
 func (b *BlockBuilder) buildBlockInherents(slot Slot, rt runtime.Instance) ([][]byte, error) {
 	// Setup inherents: add timstap0
 	idata := types.NewInherentsData()
-	err := idata.SetInt64Inherent(types.Timstap0, uint64(time.Now().Unix()))
+	timestamp := uint64(time.Now().UnixMilli())
```
qdm12, line 293: nit Call it 

```suggestion
	timestampMilliseconds := uint64(time.Now().UnixMilli())
```

Since `timestamp` is generally in seconds

qdm12 (approved):

dot/rpc/modules/author.go
noot: yeah, I was seeing an error in the front end console due to not being able to find "finalised" and expecting "finalized', substrate uses `finalized` so for things that require interop we need to use `finalized`

qdm12: :smile: yes that's true. I think `finalized` is lint-ignored anyway!

EclesioMeloJunior (approved):
LGTM

Issue #1897, opened 2021-10-14
pkg/scale: Add `Encoder` and `NewEncoder(io.Writer)`


timwu20:
## Issue summary

- Add `Encoder` and `NewEncoder(io.Writer)` to pair with the implemented `Decoder` in #1896 
- Write unit tests

## Other information and links

-

PR #2077, opened 2021-11-29
chore(lib/trie): refactor encoding and hash related code in trie package

qdm12:
:tipping_hand_man: don't be too scared by the deltas, it is mostly test and mock code being moved in subpackages

## Changes

- [x] Split code in subpackages
- [x] Add tests
- [x] Replace existing tests with more in-depth tests
- [x] Self reviewed
- [x] Ran gossamer without error

### Notes

- This does not change behavior really, it's just moving things around to see more clearly

### Branched out PRs

- #2081
- #2082 
- #2086 

## Tests

```
go test -race ./lib/trie/...
```

## Issues

- #1936 

## Primary Reviewer


lib/trie/branch/branch.go
```diff
@@ -0,0 +1,35 @@
...
+	"sync"
+
+	"github.com/ChainSafe/gossamer/lib/common"
+	"github.com/ChainSafe/gossamer/lib/trie/node"
+)
+
+var _ node.Node = (*Branch)(nil)
+
+// Branch is a branch in the trie.
+type Branch struct {
```
timwu20, line 17: did you intend for the stutter in this package? `branch.Branch`? I assumed that it would still be under the `trie` package.

lib/trie/decode/byte.go
```diff
@@ -0,0 +1,16 @@
+// Copyright 2021 ChainSafe Systems (ON)
+// SPDX-License-Identifier: LGPL-3.0-only
+
+package decode
```
timwu20, line 4: does need to be separated out into their own packages? Seems like all the types in here are only used by the `trie` package so not really seeing the value in changing that.  Seperated into their own file makes sense, but I'm not sure we need another package for this.

lib/trie/decode/byte.go
```diff
@@ -0,0 +1,16 @@
...
+
+import "io"
+
+// ReadNextByte reads the next byte from the reader.
+func ReadNextByte(reader io.Reader) (b byte, err error) {
```
timwu20, line 9: I would've assumed this would have been some private helper function inside the `trie` package.

lib/trie/decode/key.go
```diff
@@ -0,0 +1,77 @@
...
+		return nil, fmt.Errorf("%w: read %d bytes instead of %d",
+			ErrReadKeyData, n, len(key))
+	}
+
+	return KeyLEToNibbles(key)[publicKeyLength%2:], nil
+}
+
+// KeyLEToNibbles converts a Little Endian byte slice into nibbles.
+// It assumes bytes are already in Little Endian and does not rearrange nibbles.
+func KeyLEToNibbles(in []byte) (nibbles []byte) {
```
timwu20, line 62: I feel like there should be some sort of `Key` type, and this function should just be a receiver function on that type.
`Key.Nibbles() ([]byte)`.

lib/trie/leaf/copy.go
```diff
@@ -0,0 +1,29 @@
+// Copyright 2021 ChainSafe Systems (ON)
+// SPDX-License-Identifier: LGPL-3.0-only
+
+package leaf
```
timwu20, line 4: Stutter here as well `leaf.Leaf`.

lib/trie/branch/branch.go
qdm12: There is a lot of code for each node type, so splitting them is useful. I don't like the stuttering with names either, but I haven't figured out a solution to prevent that for now. Maybe in the future once there is more code re-used, I can move them to the lib/trie/node package instead.

lib/trie/decode/byte.go
qdm12: `Key` is used both in the leaf and branch packages, and `ReadNextByte` is used by `Key`, so I can't move it up in the `trie` package. I'm also aiming at reducing code in lib/trie since it's a rather large mix of things (all related to the trie, but still).

lib/trie/decode/byte.go
qdm12: See comment https://github.com/ChainSafe/gossamer/pull/2077#discussion_r760183225

lib/trie/branch/encode.go
```diff
@@ -0,0 +1,238 @@
...
+	"github.com/ChainSafe/gossamer/lib/trie/leaf"
+	"github.com/ChainSafe/gossamer/lib/trie/node"
+	"github.com/ChainSafe/gossamer/lib/trie/pools"
+	"github.com/ChainSafe/gossamer/pkg/scale"
+)
+
+// ScaleEncodeHash hashes the node (blake2b sum on encoded value)
+// and then SCALE encodes it. This is used to encode children
+// nodes of branches.
+func (b *Branch) ScaleEncodeHash() (encoding []byte, err error) {
```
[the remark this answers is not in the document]
timwu20: I created an issue a while ago for this. https://github.com/ChainSafe/gossamer/issues/1897

lib/trie/branch/branch.go
timwu20: yea `node.Branch` and `node.Leaf` makes a lot of sense.  I don't mind having the interface and the implementations of that interface in the same pacakge.

lib/trie/decode/byte.go
qdm12: Changed in c8deadd377c4b1781e631e6e56c21066135801cc so there is no longer a helper function and it also uses sync.Pools to save those 1-byte slices allocations :wink:

noot (commented):
I don't really like the idea of separating node (branch and leaf) types into separate packages, imo they should stay unexported as they don't need to show up in the API docs. `node` should be used only internally to the trie package imo

qdm12:
> imo they should stay unexported as they don't need to show up in the API docs

I can move them in the internal directory such as internal/trie/node so they won't show in docs/as API outside this project. I think we also all agree both branch and leaf packages should be moved together in the node subpackage as well. However we do need subpackages, the trie package is way too dense, so we can't go back to have everything in a single package and have unexported things.

lib/trie/decode/byte.go
qdm12: Done in https://github.com/ChainSafe/gossamer/pull/2077/commits/30e2318a9ddfca712a319b3aecce07beaf252d85

`branch`, `leaf`, `encode` and `decode` are merged in `lib/trie/node` and it looks decent. Only key -> nibbles and nibbles -> key were moved to the new lib/trie/codec package since they're used in lib/trie as well and to avoid a confusing API.

lib/trie/leaf/copy.go
qdm12: Done in https://github.com/ChainSafe/gossamer/pull/2077/commits/30e2318a9ddfca712a319b3aecce07beaf252d85

`branch`, `leaf`, `encode` and `decode` are merged in `lib/trie/node`

qdm12:
@noot @timwu20 

1. I merged the `lib/trie/leaf` + `lib/trie/branch` + `lib/trie/encode` + `lib/trie/decode` package in `lib/trie/node`
2. I moved `lib/trie/node` in `internal/trie/node` to not expose the node API
3. I added an interface alias in `lib/trie`

     ```go
     import "github.com/ChainSafe/gossamer/internal/trie/node"
     
     // Node is node in the trie and can be a leaf or a branch.
     type Node node.Node
     ```

     So that developers importing the gossamer repository can implement the interface if they want to for whatever reason.

Let me know if that fits what you wanted :wink: Thanks!

lib/trie/decode/key.go
qdm12: I started working on it in a branch, but it's a lot more changes than I expected. I will pause it for now but taking note to do it at a later time eventually.

lib/trie/branch/branch.go
qdm12: Done, see https://github.com/ChainSafe/gossamer/pull/2077#issuecomment-987742040

noot (commented):
tbh I don't know how I feel about moving everything into separate packages, I feel like it's more confusing but if you think it's clearer then happy to review :p

qdm12:
> tbh I don't know how I feel about moving everything into separate packages, I feel like it's more confusing but if you think it's clearer then happy to review :p

It does make sense to me to split it since node/leaf/branch can be split apart from the trie / database / generations / pruning operations. I'm opposed to have many nested micro packages, but here we really had some heavy weight trie package :smile_cat: I'm working now off this branch and it feels quite nicer I think.

lib/trie/decode.go
```diff
@@ -0,0 +1,48 @@
...
+	"github.com/ChainSafe/gossamer/internal/trie/node"
+	"github.com/ChainSafe/gossamer/internal/trie/pools"
+)
+
+var (
+	ErrReadHeaderByte  = errors.New("cannot read header byte")
+	ErrUnknownNodeType = errors.New("unknown node type")
+)
+
+func decodeNode(reader io.Reader) (n Node, err error) {
```
timwu20, line 21: I think this function should be in the `node` package with public sig `Decode(reader) (n Node, err error)`.  That way all the `DecodeBranch|Leaf` functions can be private.

timwu20 (changes requested):

lib/trie/decode.go
qdm12: Totally! Done in 4be79860a78972e191b9d9aa1cc7888f9876f7a5

internal/trie/node/branch_encode_test.go
```diff
@@ -0,0 +1,514 @@
...
+
+import (
+	"testing"
+
+	"github.com/golang/mock/gomock"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func Test_Branch_Encode(t *testing.T) {
```
kishansagathiya, line 14:
```suggestion
func TestBranchEncode(t *testing.T) {
```

internal/trie/node/decode.go
```diff
@@ -0,0 +1,145 @@
...
+}
+
+// decodeBranch reads and decodes from a reader with the encoding specified in lib/trie/node/encode_doc.go.
+// Note that since the encoded branch stores the hash of the children nodes, we are not
+// reconstructing the child nodes from the encoding. This function instead stubs where the
+// children are known to be with an empty leaf. The children nodes hashes are then used to
+// find other values using the persistent database.
+func decodeBranch(reader io.Reader, header byte) (branch *Branch, err error) {
+	nodeType := header >> 6
+	if nodeType != 2 && nodeType != 3 {
```
kishansagathiya, line 65:
```suggestion
	if nodeType != BranchType && nodeType != BranchWithValueType {
```

internal/trie/node/decode.go
```diff
@@ -0,0 +1,145 @@
...
+
+	childrenBitmap := make([]byte, 2)
+	_, err = reader.Read(childrenBitmap)
+	if err != nil {
+		return nil, fmt.Errorf("%w: %s", ErrReadChildrenBitmap, err)
+	}
+
+	sd := scale.NewDecoder(reader)
+
+	if nodeType == 3 {
```
kishansagathiya, line 85:
```suggestion
	if nodeType == BranchWithValueType {
```

internal/trie/node/decode.go
```diff
@@ -0,0 +1,145 @@
...
+
+	branch.Dirty = true
+
+	return branch, nil
+}
+
+// decodeLeaf reads and decodes from a reader with the encoding specified in lib/trie/node/encode_doc.go.
+func decodeLeaf(reader io.Reader, header byte) (leaf *Leaf, err error) {
+	nodeType := header >> 6
+	if nodeType != 1 {
```
kishansagathiya, line 119:
```suggestion
	if nodeType != LeafType {
```

internal/trie/codec/nibbles.go
```diff
@@ -0,0 +1,47 @@
...
+}
+
+// KeyLEToNibbles converts a Little Endian byte slice into nibbles.
+// It assumes bytes are already in Little Endian and does not rearrange nibbles.
+func KeyLEToNibbles(in []byte) (nibbles []byte) {
+	if len(in) == 0 {
+		return []byte{}
+	} else if len(in) == 1 && in[0] == 0 {
+		return []byte{0, 0}
+	}
```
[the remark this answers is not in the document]
qdm12: I prefer `else if` to save a line :wink: But I understand your point, it also makes sense. Just a matter of supernit preference I guess

internal/trie/node/branch_encode.go
```diff
@@ -0,0 +1,231 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("cannot hash node: %w", err)
+	}
+
+	encoding, err = scale.Marshal(buffer.Bytes())
+	if err != nil {
+		return nil, fmt.Errorf("cannot scale encode hashed node: %w", err)
+	}
+
+	return encoding, nil
```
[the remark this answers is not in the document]
qdm12: I disagree, it's important to know which of the two possible errors was triggered. There would be a single error possible, then yes wrapping the error would have no point. What do you think?

On the other hand, I will change `node` to `branch` (and `leaf` for leaves) since we're handling a branch here!

internal/trie/node/branch_encode.go
```diff
@@ -0,0 +1,231 @@
...
+			return fmt.Errorf("cannot scale encode value: %w", err)
+		}
+
+		_, err = buffer.Write(bytes)
+		if err != nil {
+			return fmt.Errorf("cannot write encoded value to buffer: %w", err)
+		}
+	}
+
+	const parallel = false // TODO
```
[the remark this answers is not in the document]
qdm12: This is done in #2081 I updated the comment.

It was left disabled since this implementation would use potentially a lot of stack memory due to the high number / unbound number of goroutines.

internal/trie/node/children.go
```diff
@@ -0,0 +1,26 @@
...
+
+package node
+
+// ChildrenBitmap returns the 16 bit bitmap
+// of the children in the branch.
+func (b *Branch) ChildrenBitmap() (bitmap uint16) {
+	for i := uint(0); i < 16; i++ {
+		if b.Children[i] != nil {
+			bitmap = bitmap | 1<<i
+		}
```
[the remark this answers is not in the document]
qdm12: Good suggestion, keeping it!

internal/trie/node/copy.go
```diff
@@ -0,0 +1,54 @@
...
+package node
+
+// Copy deep copies the branch.
+func (b *Branch) Copy() Node {
+	b.RLock()
+	defer b.RUnlock()
+
+	cpy := &Branch{
+		Key:        make([]byte, len(b.Key)),
+		Children:   b.Children, // copy interface pointers
```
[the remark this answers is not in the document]
qdm12: No, AFAIK we don't want to copy the entire child nodes, just the current node and re-use the pointers to the existing nodes. Note the pointers (aka addresses) are copied since `b.Children` is an array and not a slice.

internal/trie/node/encode_test.go
```diff
@@ -0,0 +1,14 @@
+// Copyright 2021 ChainSafe Systems (ON)
+// SPDX-License-Identifier: LGPL-3.0-only
+
+package node
+
+import "errors"
+
+type writeCall struct {
+	written []byte
+	n       int
```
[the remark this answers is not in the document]
qdm12: Yes it's the `n` returned from a `n, err := writer.Write()` call. It's named as such in the standard lib so I thought I would leave it as `n` here too. I'll commit your comment, but feel free to comment if you feel we should rename that var.

lib/trie/trie.go
```diff
@@ -18,10 +21,9 @@ var EmptyHash, _ = NewEmptyTrie().Hash()
 // Use NewTrie to create a trie that sits on top of a database.
 type Trie struct {
 	generation  uint64
-	root        node
+	root        Node
 	childTries  map[common.Hash]*Trie // Used to store the child tries.
 	deletedKeys []common.Hash
-	parallel    bool
```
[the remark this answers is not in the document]
qdm12: Because it was always true in production code.
And it being true might use a lot of stack memory with the current parallel hashing implementation.

This is improved/fixed in #2081 where opportunistic hashing is always enabled.

internal/trie/node/branch_encode_test.go
qdm12: I prefer formatting my test names with underscores as `Test_Receiver_MethodName` otherwise it gets hard to read when mixing multiple-words together. In example `NetworkService` with private method `getPeer` would result in `TestNetworkServicegetPeer` which is worst than `Test_NetworkService_getPeer`. But don't get me wrong, I dislike snake_case apart from this niche usage haha

internal/trie/node/branch.go
```diff
@@ -0,0 +1,34 @@
...
+
+// Branch is a branch in the trie.
+type Branch struct {
+	Key        []byte // partial key
+	Children   [16]Node
+	Value      []byte
+	Dirty      bool
+	Hash       []byte
+	Encoding   []byte
+	Generation uint64
```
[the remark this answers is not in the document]
qdm12: Commented in https://github.com/ChainSafe/gossamer/pull/2077/commits/2feebe715089c2f5d3d28f834ddf78bb226592b3

internal/trie/node/branch.go
```diff
@@ -0,0 +1,34 @@
...
+)
+
+var _ Node = (*Branch)(nil)
+
+// Branch is a branch in the trie.
+type Branch struct {
+	Key        []byte // partial key
+	Children   [16]Node
+	Value      []byte
+	Dirty      bool
```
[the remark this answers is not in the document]
qdm12: Commented in https://github.com/ChainSafe/gossamer/pull/2077/commits/2feebe715089c2f5d3d28f834ddf78bb226592b3

kishansagathiya (commented):
This is tricky since I don't have a clear view on what has changed and what has moved. I have tried to comment wherever I see problems, but I have also ran quickly with some parts and avoided some comments since I don't think it is a good idea to add more changes in this PR.

- I don't like that some smaller parts have their own files, for example dirty, key, generation, value, copy, etc.
- To me, methods of same struct being together(in the same file) makes more than same methods of similar structs. 
- It might be possible to just merge Leaf and Branch struct. It might require some gymnastic like using an interface for methods that are different or functions as values. But if we can do that, that could make some things clearer. (Not as part of this PR)

Though I have made some smaller comments, we should (I realize that now) not make more changes in this PR to keep things simpler.

internal/trie/node/decode.go
```diff
@@ -0,0 +1,145 @@
...
+	}
+}
+
+// decodeBranch reads and decodes from a reader with the encoding specified in lib/trie/node/encode_doc.go.
+// Note that since the encoded branch stores the hash of the children nodes, we are not
+// reconstructing the child nodes from the encoding. This function instead stubs where the
+// children are known to be with an empty leaf. The children nodes hashes are then used to
+// find other values using the persistent database.
+func decodeBranch(reader io.Reader, header byte) (branch *Branch, err error) {
+	nodeType := header >> 6
```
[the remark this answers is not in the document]
qdm12: `6` is just the amount of bits to shift to get the node type, from the header byte. Should I name it as a constant? Although I have no idea how to name it really :thinking: Feel free to come up with something

internal/trie/node/types.go
```diff
@@ -0,0 +1,17 @@
+// Copyright 2021 ChainSafe Systems (ON)
+// SPDX-License-Identifier: LGPL-3.0-only
+
+package node
+
+// Type is the byte type for the node.
+type Type byte
```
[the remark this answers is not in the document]
qdm12: Good spot, `Type` got used in #2086 only. Updated in 7503f21e381245378d0c3250b84a9b2d2a1ceab0

internal/trie/node/decode_test.go
```diff
@@ -0,0 +1,309 @@
...
+			leaf, err := decodeLeaf(testCase.reader, testCase.header)
+
+			assert.ErrorIs(t, err, testCase.errWrapped)
+			if err != nil {
+				assert.EqualError(t, err, testCase.errMessage)
+			}
+			assert.Equal(t, testCase.leaf, leaf)
+		})
+	}
+}
```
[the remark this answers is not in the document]
qdm12: Comments added in 3129e796ad7d177783c12589a4305a9e22da88ab

internal/trie/node/leaf_encode.go
```diff
@@ -0,0 +1,185 @@
...
+
+	"github.com/ChainSafe/gossamer/internal/trie/codec"
+	"github.com/ChainSafe/gossamer/internal/trie/pools"
+	"github.com/ChainSafe/gossamer/lib/common"
+	"github.com/ChainSafe/gossamer/pkg/scale"
+)
+
+// SetEncodingAndHash sets the encoding and hash slices
+// given to the branch. Note it does not copy them, so beware.
+func (l *Leaf) SetEncodingAndHash(enc, hash []byte) {
```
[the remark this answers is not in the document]
qdm12: Taking note of it. There could be some further refactor indeed, I'll see what I can do.

qdm12:
@kishansagathiya 

> This is tricky since I don't have a clear view on what has changed and what has moved.

Agreed. There was unfortunately no way to split the changes really due to the file movements.
On the other hand:

- I added deeper tests with higher coverage in all the `internal/trie` packages
- Existing higher level tests (i.e. in `lib/trie`) still pass

So I am confident no bug was introduced. But I agree it's hard to review.

> avoided some comments since I don't think it is a good idea to add more changes in this PR.
> Though I have made some smaller comments, we should (I realize that now) not make more changes in this PR to keep things simpler.

Feel free to comment, I already have a few PRs pointing to this PR, adding a few more is fine too. I'll rebase them on development once this one merges.
I'm also taking note of things I can't do right now as side branches, I will eventually create an issue with each suggestion.

> I don't like that some smaller parts have their own files, for example dirty, key, generation, value, copy
> To me, methods of same struct being together(in the same file) makes more than same methods of similar structs.

Agreed; however the alternative is to have only a `leaf.go` and `branch.go` files which would be in the 1000+ lines, and that's hard to navigate in.
That is why I took the approach to split by 'action' (i.e. encoding, decoding, generation etc.) rather than by type. That would definitely be improved by the suggestion you made below:

> It might be possible to just merge Leaf and Branch struct

Definitely. I'll explore this once the memory issue is fixed. Added this to my list of suggestions.

EclesioMeloJunior (changes requested):

internal/trie/node/decode.go
```diff
@@ -0,0 +1,145 @@
...
+// find other values using the persistent database.
+func decodeBranch(reader io.Reader, header byte) (branch *Branch, err error) {
+	nodeType := Type(header >> 6)
+	if nodeType != BranchType && nodeType != BranchWithValueType {
+		return nil, fmt.Errorf("%w: %d", ErrNodeTypeIsNotABranch, nodeType)
+	}
+
+	branch = new(Branch)
+
+	keyLen := header & 0x3f
```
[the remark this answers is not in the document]
qdm12: Done in https://github.com/ChainSafe/gossamer/pull/2077/commits/e17c08e42cc377144e4d6a23b13520b8f1c7f3fc

timwu20 (approved):

internal/trie/node/branch.go
```diff
@@ -0,0 +1,50 @@
...
+	// dirty is true when the branch differs
+	// from the node stored in the database.
+	dirty      bool
+	hashDigest []byte
+	encoding   []byte
+	// generation is incremented on every trie Snapshot() call.
+	// Nodes that are part of the trie are then gradually updated
+	// to have a matching generation number as well, if they are
+	// still relevant.
+	generation uint64
```
noot, line 29: not sure what this part of the comment means, I'd just remove it haha

noot: the generation is used to track distinct "versions" of the trie, which may point to nodes from previous generations. eg, when we sync a new block, we make a snapshot, and then all new nodes that are added during the state transition process have the new generation, but the old unmodified nodes have the older generation

internal/trie/node/branch_encode.go
```diff
@@ -0,0 +1,231 @@
...
+	encodingBuffer := pools.EncodingBuffers.Get().(*bytes.Buffer)
+	encodingBuffer.Reset()
+	defer pools.EncodingBuffers.Put(encodingBuffer)
+
+	err = b.Encode(encodingBuffer)
+	if err != nil {
+		return fmt.Errorf("cannot encode leaf: %w", err)
+	}
+
+	// if length of encoded leaf is less than 32 bytes, do not hash
```
noot, line 49:
```suggestion
	// if length of encoded node is less than 32 bytes, do not hash
```

noot (approved):
looks good, just minor comments!

internal/trie/node/branch.go
qdm12: Updated in 8c3178891f3f0aeeb842482f8c9213619e04b232

internal/trie/node/branch_encode.go
qdm12: Change to `branch` in https://github.com/ChainSafe/gossamer/pull/2077/commits/865aae6d143ec3137c92d30a7ce7235b96bb41bb

internal/trie/node/branch_encode.go
```diff
@@ -0,0 +1,231 @@
...
+			return fmt.Errorf("cannot scale encode value: %w", err)
+		}
+
+		_, err = buffer.Write(bytes)
+		if err != nil {
+			return fmt.Errorf("cannot write encoded value to buffer: %w", err)
+		}
+	}
+
+	const parallel = false // TODO Done in pull request #2081
```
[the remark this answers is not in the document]
qdm12: :+1: yes it's removed in the (rebased) #2081

internal/trie/node/decode.go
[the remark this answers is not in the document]
qdm12: Updated to a constant in https://github.com/ChainSafe/gossamer/pull/2077/commits/e17c08e42cc377144e4d6a23b13520b8f1c7f3fc

PR #2081, opened 2021-11-30
chore(lib/trie): opportunist parallel hashing

qdm12:
## Changes

Benchmarks:

```
// Opportunistic parallel:	13781602 ns/op	14419488 B/op	  323575 allocs/op
// Sequentially:		24269268 ns/op	20126525 B/op	  327668 allocs/op
```

- [x] Encode in parallel only branches, opportunistically if one goroutine is available
- [x] Limit the number of simultaneous goroutines to the number of CPUs available.
- [x] Add more test cases for branch encoding (will add once reviewed)
- [x] Add benchmarks eventually for the fun of it
- [x] Only write child encoding to buffer if no error was encountered in a previous child encoding

## Tests

```
go test -race ./internal/trie/...
```

## Issues

- #1973 

## Primary Reviewer

- @timwu20

### DIFF
--- a/internal/trie/node/branch_encode.go
+++ b/internal/trie/node/branch_encode.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"hash"
 	"io"
+	"runtime"
 
 	"github.com/ChainSafe/gossamer/internal/trie/codec"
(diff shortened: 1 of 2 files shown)


EclesioMeloJunior (approved):
🚀

Issue #1968, opened 2021-11-03
Performance monitoring


qdm12:
Epic focused on performance monitoring.

This was created with the aim of solving #1936 initially.
It is meant to also facilitate in the future:

- the debugging of performance issues
- optimize performance (see epic #1494)

PR #2082, opened 2021-11-30
chore(lib/trie): `lib/trie/recorder` sub-package

qdm12:
## Changes

- See title
- Add tests for the recorder

## Tests

```
go test -race ./lib/trie/...
```

## Issues

- #1968


lib/trie/record/recorder.go
```diff
@@ -0,0 +1,38 @@
...
+}
+
+// Record appends a node to the list of visited nodes.
+func (r *Recorder) Record(hash, rawData []byte) {
+	r.nodes = append(r.nodes, Node{RawData: rawData, Hash: hash})
+}
+
+// Next returns the first node in the recorded list
+// and removes it (shift operation).
+func (r *Recorder) Next() (node *Node) {
```
EclesioMeloJunior, line 23: taking advantage of this improvements I would like to suggest changing the behavior of the `Next` function, instead of shifting the data we could act as an iterator and keep track of an internal index that will be incremented each time the `Next` function is called and in the `IsEmpty` just return if the internal index is equal the `len(data)`

EclesioMeloJunior (changes requested):

lib/trie/record/recorder.go
qdm12: That was a good idea, but I ended up making [it fast and dumb since we don't need all the fanciness](https://github.com/ChainSafe/gossamer/pull/2082/commits/ba6cc4010f8c61c535cadcfca7ee50715084dfaa)

noot (approved):

EclesioMeloJunior (approved):

lib/trie/record/recorder.go
EclesioMeloJunior: !it got even better!!

qdm12:
@EclesioMeloJunior careful when merging PRs even when approved, this was pointing to #2077 and I originally wanted to rebase it on development once the base branch #2077 would be merged, to limit deltas.

EclesioMeloJunior:
@qdm12 sorry about that, I thought everything was fine, I will be careful next times! Thanks for warning!

Issue #2085, opened 2021-11-30
lib/trie: refactor and add tests for trie and database related code


qdm12:
## Task summary

- [x] Database code fully tested; then
- [x] Database code refactored
- [x] Trie code (lib/trie/trie.go) fully tested
- [ ] Trie code refactored

PR #2086, opened 2021-11-30
chore(lib/trie): node `Type() Type` method

qdm12:
## Changes

- [x] Add node `Type() Type` method
- [x] Implement `Type() Type` method for branches (`2` or `3` type)
- [x] Implement `Type() Type` method for leaves (`1`)
- [x] Update existing code to use `Type()` on nodes
- [x] Update existing code to use node type constants (1, 2, 3)
- [x] Use `Type()` methods instead of type switches, with the aim of getting rid of type assertions in the future

## Tests

```
go test -race ./lib/trie/...
```

## Issues

- #2085 

## Primary Reviewer

### DIFF
--- a/internal/trie/node/header.go
+++ b/internal/trie/node/header.go
@@ -15,9 +15,9 @@ const (
 func (b *Branch) encodeHeader(writer io.Writer) (err error) {
 	var header byte
 	if b.Value == nil {
-		header = 2 << 6
+		header = byte(BranchType) << 6
 	} else {
-		header = 3 << 6
+		header = byte(BranchWithValueType) << 6
 	}
 
 	if len(b.Key) >= keyLenOffset {
@@ -44,7 +44,7 @@ func (b *Branch) encodeHeader(writer io.Writer) (err error) {
 
 // encodeHeader creates the encoded header for the leaf.
 func (l *Leaf) encodeHeader(writer io.Writer) (err error) {
-	var header byte = 1 << 6
+	header := byte(LeafType) << 6
 
 	if len(l.Key) < 63 {
 		header |= byte(len(l.Key))
--- a/internal/trie/node/header.go
+++ b/internal/trie/node/header.go
@@ -14,10 +14,11 @@ const (
 // encodeHeader creates the encoded header for the branch.
 func (b *Branch) encodeHeader(writer io.Writer) (err error) {
 	var header byte
+	const nodeHeaderShift = 6
 	if b.Value == nil {
-		header = byte(BranchType) << 6
+		header = byte(BranchType) << nodeHeaderShift
 	} else {
-		header = byte(BranchWithValueType) << 6
+		header = byte(BranchWithValueType) << nodeHeaderShift
 	}
 
 	if len(b.Key) >= keyLenOffset {
@@ -44,7 +45,8 @@ func (b *Branch) encodeHeader(writer io.Writer) (err error) {
 
 // encodeHeader creates the encoded header for the leaf.
 func (l *Leaf) encodeHeader(writer io.Writer) (err error) {
-	header := byte(LeafType) << 6
+	const nodeHeaderShift = 6
+	header := byte(LeafType) << nodeHeaderShift
 
 	if len(l.Key) < 63 {
 		header |= byte(len(l.Key))
--- a/internal/trie/node/header.go
+++ b/internal/trie/node/header.go
@@ -8,13 +8,13 @@ import (
 )
 
 const (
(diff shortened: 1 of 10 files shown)


lib/trie/branch/type.go
```diff
@@ -0,0 +1,12 @@
+package branch
```
EclesioMeloJunior, line 1: could we add a license here?

EclesioMeloJunior (approved):
Wondering if we could return an error instead of `panic`, I know this should be an edge case, but we could be able to handle a wrong type node, for example ignoring it.

lib/trie/branch/type.go
qdm12: done :+1: Note this is in the CI so it will ultimately not allow to merge without it :wink:

qdm12:
@EclesioMeloJunior 

> I know this should be an edge case, but we could be able to handle a wrong type node, for example ignoring it.

Code is changed simplified in an upcoming PR refactoring `lib/trie/trie.go`.
However, I decided to keep the program to panic on an unknown node type, which is really not meant to happen.

internal/trie/node/header.go
```diff
@@ -11,9 +11,9 @@ import (
 func (b *Branch) encodeHeader(writer io.Writer) (err error) {
 	var header byte
 	if b.Value == nil {
-		header = 2 << 6
+		header = byte(BranchType) << 6
```
EclesioMeloJunior, line 14:
```suggestion
                const nodeHeaderShift = 6
		header = byte(BranchType) << nodeHeaderShift
```

EclesioMeloJunior (changes requested):

internal/trie/node/header.go
qdm12: Nice name :+1: Added here and in 2 other places in that file.

internal/trie/node/header.go
```diff
@@ -44,7 +45,8 @@ func (b *Branch) encodeHeader(writer io.Writer) (err error) {
 
 // encodeHeader creates the encoded header for the leaf.
 func (l *Leaf) encodeHeader(writer io.Writer) (err error) {
-	var header byte = 1 << 6
+	const nodeHeaderShift = 6
```
EclesioMeloJunior, line 48: should we maintain 2 `const` with the same name and value? why not put in the pkg level and reuse it?

EclesioMeloJunior (changes requested):

internal/trie/node/header.go
qdm12: Totally, done! Good catch!

EclesioMeloJunior (approved):

lib/trie/trie.go
```diff
@@ -324,9 +329,9 @@ func (t *Trie) insert(parent Node, key []byte, value Node) Node {
 		}
 
 		return br
+	default:
+		panic("unknown node type: " + fmt.Sprint(newParent.Type()))
```
jimjbrettj, line 333: Why do we want to panic rather than return an error here?

jimjbrettj (approved):
Comment for my learning but looks good!

lib/trie/trie.go
qdm12: If this happens we have something terribly wrong coded :smile: We should really panic instead of trying to recover and continue things. Maybe that can be removed with generics in Go 1.18 or (as Kishan suggested) if we merge both leaf and branch in the same structure and don't use interfaces. But yeah not for now :wink:

lib/trie/database.go
```diff
@@ -363,13 +374,15 @@ func (t *Trie) writeDirty(db chaindb.Batch, n Node) error {
 			hash, err)
 	}
 
-	branch, ok := n.(*node.Branch)
-	if !ok {
-		// the node is a leaf
+	switch n.Type() {
+	case node.BranchType, node.BranchWithValueType:
+	default: // not a branch
```
noot, line 379:
```suggestion
	default: // not a branch, so is a leaf
```

lol, very minor, up to you to change this

lib/trie/trie.go
noot: yeah this case should never be hit, we just have to have it to make the compiler happy afaik

noot (approved):

lib/trie/database.go
qdm12: Oops sorry I didn't see your comment, my bad!
It could be an untyped nil as well eventually :thinking: (not super sure, my 2021 memory faded a bit)

Issue #1720, opened 2021-08-06
[RPC] Improve integration tests with the RPC layer


EclesioMeloJunior:
## Task summary

- Actually the majority of the RPC tests inside `test/rpc` folder are beign skipped or expect and empty data

## Specification reference

Would be nice to:
- Rename the files to the module (eg. `system_test.go`)
- Update the `expect` field for every test case, or create a `Test_` function for more complex cases
- Guarantee that the JSON response format is correct

## Other information and links

- N/A

timwu20:
@EclesioMeloJunior can you create sub issues for this epic, and un-assign yourself

PR #2091, opened 2021-11-30
tests(dot/rpc/modules): implement integration tests to rpc `author` module

### LINKED ISSUE #2172: [RPC] Implement `Author module` integration tests
## Task summary

- Currently, we don't have an integration test for the RPC modules
- We should be able to make sure the author module main dependencies are working
- This issue ensures the generated extrinsic works with runtimes at version 09.10 (#904)

## Specification reference

- N/A

- N/A

EclesioMeloJunior:
## Changes

- Remove `t.Skip()` from all author tests
- Build an integration test controller which instantiates the needed dependencies of `core.Service`
- Merge some tests into a unique test with a map of test cases

## Tests

```
go test github.com/ChainSafe/gossamer/dot/rpc/modules -tags=integration
```

## Issues

- related with #1720
- closes #2172

## Primary Reviewer

- @edwardmack


edwardmack (approved):
Great work!

dot/rpc/modules/author_integration_test.go
```diff
@@ -395,6 +385,108 @@ func TestAuthorModule_HasKey_Integration(t *testing.T) {
...
+		waitErr        error
+	}{
+		"public keys are in the right order, should return true": {
+			pubSessionKeys: "0x6b802349d948444d41397da09ec597fbd8ae8fdd3dfa153b2bb2bddcf020457c" + // gran
+				"d43593c715fdd31c61141abd04a99fd6822c8558854ccde39a5684e7a56da27d" + // babe
+				"d43593c715fdd31c61141abd04a99fd6822c8558854ccde39a5684e7a56da27d" + // imon
+				"d43593c715fdd31c61141abd04a99fd6822c8558854ccde39a5684e7a56da27d", // audi
+			expect: true,
+		},
+		"uknown public keys in the right order, should return false": {
```
qdm12, line 442:
```suggestion
		"unknown public keys in the right order, should return false": {
```

dot/rpc/modules/author_integration_test.go
```diff
@@ -395,6 +385,108 @@ func TestAuthorModule_HasKey_Integration(t *testing.T) {
...
+			require.NoError(t, err)
+		}
+	}
+
+	testcases := map[string]struct {
+		pubSessionKeys string
+		expect         bool
+		waitErr        error
+	}{
+		"public keys are in the right order, should return true": {
```
qdm12, line 435: nit Don't change here, just for future tests - I'm not sure how punctuation is handled for subtests, they might be removed or changed to `_`, so in the future try to have test case names without punctuation i.e. `"public keys are in the right order returns true"`. No need to say the return either I would say since it's in the test case values, but that's just my opinion :wink:

dot/rpc/modules/author_integration_test.go
```diff
@@ -395,6 +385,108 @@ func TestAuthorModule_HasKey_Integration(t *testing.T) {
...
+				"de6fa0da51c52cc117d77aeb329595b15070db444e7ed4c4adec714b291c1845" + // babe
+				"de6fa0da51c52cc117d77aeb329595b15070db444e7ed4c4adec714b291c1845" + // imon
+				"de6fa0da51c52cc117d77aeb329595b15070db444e7ed4c4adec714b291c1845", // audi
+			expect: false,
+		},
+		"public keys are not in the right order, should return false": {
+			pubSessionKeys: "0xd43593c715fdd31c61141abd04a99fd6822c8558854ccde39a5684e7a56da27d" + // gran
+				"6b802349d948444d41397da09ec597fbd8ae8fdd3dfa153b2bb2bddcf020457c" + // babe
+				"d43593c715fdd31c61141abd04a99fd6822c8558854ccde39a5684e7a56da27d" + // imon
+				"d43593c715fdd31c61141abd04a99fd6822c8558854ccde39a5684e7a56da27d", // audi
```
qdm12, line 453: Nice split :100: !

dot/rpc/modules/author_integration_test.go
```diff
@@ -395,6 +385,108 @@ func TestAuthorModule_HasKey_Integration(t *testing.T) {
...
+			waitErr:        errors.New("invalid string"),
+		},
+	}
+
+	for tname, tt := range testcases {
+		tt := tt
+		t.Run(tname, func(t *testing.T) {
+			t.Parallel()
+			var req HasSessionKeyRequest
+			req.PublicKeys = tt.pubSessionKeys
```
qdm12, line 473: nit

```suggestion
			req := HasSessionKeyRequest{
				PublicKeys: tt.pubSessionKeys
			}
```

dot/rpc/modules/author_integration_test.go
```diff
@@ -395,6 +385,108 @@ func TestAuthorModule_HasKey_Integration(t *testing.T) {
...
+			err := auth.HasSessionKeys(nil, &req, &res)
+
+			if tt.waitErr != nil {
+				require.EqualError(t, tt.waitErr, err.Error())
+				require.False(t, bool(res))
+				return
+			}
+
+			require.NoError(t, err)
+			require.Equal(t, tt.expect, bool(res))
```
qdm12, line 486: nit so we have only one assert for res

```suggestion
			} else {
				require.NoError(t, err)
			}
			require.Equal(t, tt.expect, bool(res))
```

dot/rpc/modules/chain_integration_test.go
```diff
@@ -0,0 +1,423 @@
...
+}
+
+func TestChainGetHeader_Latest(t *testing.T) {
+	state := newTestStateService(t)
+	svc := NewChainModule(state.Block)
+
+	header, err := state.Block.BestBlockHeader()
+	require.NoError(t, err)
+
+	di := types.NewDigestItem()
```
qdm12, line 81: Just a general comment, I think it would be better to use full names like `digestItem`, `di` is just plain confusing. I even prefer `request` instead of `res` for example, but that might be just my opinion / full name exaggeration. If it would be me I would rename `err` to `error` and that would blow up the Go type system so yeah I'm also a full name maniac.

So all good here, just a thing to keep in mind for future code :+1:

dot/rpc/modules/chain_integration_test.go
```diff
@@ -0,0 +1,423 @@
...
+		},
+	}
+
+	res := &ChainBlockResponse{}
+	req := &ChainHashRequest{Bhash: &hash}
+
+	err = svc.GetBlock(nil, req, res)
+	require.Nil(t, err)
+
+	require.Equal(t, expected, res)
```
qdm12, line 163: Can't we have these few tests testing GetBlock as test cases and subtests?

dot/rpc/modules/chain_integration_test.go
```diff
@@ -0,0 +1,423 @@
...
+	resString := string("")
+	res := ChainHashResponse(resString)
+	req := ChainBlockNumberRequest{"0x01"}
+
+	err := svc.GetBlockHash(nil, &req, &res)
+	require.NoError(t, err)
+
+	expected, err := state.Block.GetBlockByNumber(big.NewInt(1))
+	require.NoError(t, err)
+	require.Equal(t, expected.Header.Hash().String(), res)
```
qdm12, line 265: Same here, these could be arranged as test cases and subtests

dot/rpc/modules/chain_integration_test.go
```diff
@@ -0,0 +1,423 @@
...
+		rtCfg.NodeStorage.BaseDB, err = utils.SetupDatabase(filepath.Join(testDatadirPath, "offline_storage"), false)
+		require.NoError(t, err)
+	}
+
+	rt, err := wasmer.NewRuntimeFromGenesis(rtCfg)
+	require.NoError(t, err)
+
+	loadTestBlocks(t, genesisHeader.Hash(), stateSrvc.Block, rt)
+
+	t.Cleanup(func() {
```
qdm12, line 379: Maybe move that up as soon as `stateSrvc` is defined? In case the test fails with a `require` instruction

dot/rpc/modules/childstate.go
```diff
@@ -136,7 +136,7 @@ func (cs *ChildStateModule) GetStorageHash(_ *http.Request, req *GetStorageHash,
 }
 
 // GetStorage returns a child storage entry.
-func (cs *ChildStateModule) GetStorage(_ *http.Request, req *ChildStateStorageRequest, res *StateStorageResponse) error {
+func (cs *ChildStateModule) GetStorage(_ *http.Request, req *ChildStateStorageRequest, res *StateStorageResponse) error { //nolint:lll
```
qdm12, line 139:
```suggestion
func (cs *ChildStateModule) GetStorage(_ *http.Request, req *ChildStateStorageRequest,
	res *StateStorageResponse) error {
```

dot/rpc/modules/chain.go
```diff
@@ -128,7 +128,7 @@ func (cm *ChainModule) GetFinalizedHead(r *http.Request, req *EmptyRequest, res
 }
 
 // GetFinalizedHeadByRound returns the hash of the block finalised at the given round and setID
-func (cm *ChainModule) GetFinalizedHeadByRound(r *http.Request, req *ChainFinalizedHeadRequest, res *ChainHashResponse) error {
+func (cm *ChainModule) GetFinalizedHeadByRound(r *http.Request, req *ChainFinalizedHeadRequest, res *ChainHashResponse) error { //nolint:lll
```
qdm12, line 131:
```suggestion
func (cm *ChainModule) GetFinalizedHeadByRound(r *http.Request, req *ChainFinalizedHeadRequest,
	res *ChainHashResponse) error {
```

dot/rpc/modules/chain.go
```diff
@@ -152,7 +152,7 @@ func (cm *ChainModule) GetHeader(r *http.Request, req *ChainHashRequest, res *Ch
 
 // SubscribeFinalizedHeads handled by websocket handler, but this func should remain
 //  here so it's added to rpc_methods list
-func (cm *ChainModule) SubscribeFinalizedHeads(r *http.Request, req *EmptyRequest, res *ChainBlockHeaderResponse) error {
+func (cm *ChainModule) SubscribeFinalizedHeads(r *http.Request, req *EmptyRequest, res *ChainBlockHeaderResponse) error { //nolint:lll
```
qdm12, line 155:
```suggestion
func (cm *ChainModule) SubscribeFinalizedHeads(r *http.Request, req *EmptyRequest,
	res *ChainBlockHeaderResponse) error {
```

dot/rpc/modules/author.go
```diff
@@ -191,7 +191,7 @@ func (am *AuthorModule) PendingExtrinsics(r *http.Request, req *EmptyRequest, re
 }
 
 // RemoveExtrinsic Remove given extrinsic from the pool and temporarily ban it to prevent reimporting
-func (am *AuthorModule) RemoveExtrinsic(r *http.Request, req *ExtrinsicOrHashRequest, res *RemoveExtrinsicsResponse) error {
+func (am *AuthorModule) RemoveExtrinsic(r *http.Request, req *ExtrinsicOrHashRequest, res *RemoveExtrinsicsResponse) error { //nolint:lll
```
qdm12, line 194:
```suggestion
func (am *AuthorModule) RemoveExtrinsic(r *http.Request, req *ExtrinsicOrHashRequest,
	res *RemoveExtrinsicsResponse) error {
```

dot/rpc/modules/childstate_integration_test.go
```diff
@@ -0,0 +1,7 @@
+// Copyright 2021 ChainSafe Systems (ON)
+// SPDX-License-Identifier: LGPL-3.0-only
+
+//go:build integration
+// +build integration
+
+package modules
```
qdm12, line 7: What's the purpose of that file?

dot/rpc/modules/dev_integration_test.go
```diff
@@ -0,0 +1,136 @@
...
+)
+
+var genesisBABEConfig = &types.BabeConfiguration{
+	SlotDuration:       1000,
+	EpochLength:        200,
+	C1:                 1,
+	C2:                 4,
+	GenesisAuthorities: []types.AuthorityRaw{},
+	Randomness:         [32]byte{},
+	SecondarySlots:     0,
```
qdm12, line 35: nit

```suggestion
```

dot/rpc/modules/grandpa.go
```diff
@@ -56,15 +56,16 @@ type ProveFinalityRequest struct {
...
+// If no authorities set is provided, the current one will be attempted.
 func (gm *GrandpaModule) ProveFinality(r *http.Request, req *ProveFinalityRequest, res *ProveFinalityResponse) error {
 	blocksToCheck, err := gm.blockAPI.SubChain(req.blockHashStart, req.blockHashEnd)
 	if err != nil {
 		return err
 	}
 
 	// Leaving check in for linter
-	if req.authorityID != uint64(0) {
+	if req.authorityID != uint64(0) { // nolint:staticcheck
```
qdm12, line 68: :pray: thanks!

dot/rpc/modules/grandpa_integration_test.go
```diff
@@ -0,0 +1,114 @@
...
+
+func TestGrandpaProveFinality(t *testing.T) {
+	testStateService := newTestStateService(t)
+
+	state.AddBlocksToState(t, testStateService.Block, 3, false)
+	bestBlock, err := testStateService.Block.BestBlock()
+
+	if err != nil {
+		t.Errorf("Fail: bestblock failed")
+	}
```
qdm12, line 33:
```suggestion
	assert.NoError(t, err, "bestblock failed")
```

dot/rpc/modules/grandpa_integration_test.go
```diff
@@ -0,0 +1,114 @@
...
+
+	res := new(ProveFinalityResponse)
+	err = gmSvc.ProveFinality(nil, &ProveFinalityRequest{
+		blockHashStart: bestBlock.Header.ParentHash,
+		blockHashEnd:   bestBlock.Header.Hash(),
+	}, res)
+
+	if err != nil {
+		t.Fatal(err)
+	}
```
qdm12, line 51:
```suggestion
	require.NoError(t, err)
```

dot/rpc/modules/grandpa_integration_test.go
```diff
@@ -0,0 +1,114 @@
...
+
+	if !reflect.DeepEqual(*res, expectedResponse) {
+		t.Errorf("Fail: expected: %+v got: %+v\n", res, &expectedResponse)
+	}
```
qdm12, line 55: You're probably just moving tests around, sorry for the suggestion noise! 

```suggestion
	assert.Equal(t, expectedResponse, *res)
```

dot/core/interface.go
```diff
@@ -47,6 +47,7 @@ type BlockState interface {
 	StoreRuntime(common.Hash, runtime.Instance)
 }
 
+//go:generate mockgen -destination=./mocks/storage_state_mock.go -package mocks . StorageState
```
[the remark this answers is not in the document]
EclesioMeloJunior: Agreed with you, but these mocks have been used by almost all other packages so I think we should keep them inside the `dot/core/mocks` to be more easy to use and avoid a lot of packages with lot of mocks there

dot/core/interface.go
```diff
@@ -68,8 +69,10 @@ type TransactionState interface {
 	RemoveExtrinsic(ext types.Extrinsic)
 	RemoveExtrinsicFromPool(ext types.Extrinsic)
 	PendingInPool() []*transaction.ValidTransaction
+	InPool(ext types.Extrinsic) bool
```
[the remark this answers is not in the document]
EclesioMeloJunior: Changed!

dot/rpc/modules/author_integration_test.go
```diff
@@ -7,133 +7,169 @@
...
-		log.Error("failed to generate runtime wasm file", err)
+		log.Errorf("failed to generate runtime wasm file: %s", err)
 		os.Exit(1)
 	}
 
 	// Start all tests
 	code := m.Run()
 
-	runtime.RemoveFiles(wasmFilePaths)
+	//runtime.RemoveFiles(wasmFilePaths)
```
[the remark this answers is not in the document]
EclesioMeloJunior: Removed!

dot/rpc/modules/author_integration_test.go
EclesioMeloJunior: Got it!

dot/rpc/modules/author_integration_test.go
EclesioMeloJunior: Resolved!

dot/rpc/modules/author_integration_test.go
EclesioMeloJunior: Changed!

dot/rpc/modules/author_integration_test.go
EclesioMeloJunior: Changed!

chain/gssmr/config.toml
```diff
@@ -40,6 +40,8 @@ port = 8545
 host = "localhost"
 modules = ["system", "author", "chain", "state", "rpc", "grandpa", "offchain", "childstate", "syncstate", "payment"]
 ws-port = 8546
+ws-unsafe-external = false
+unsafe-external = false
```
noot, line 44: are these not false by default?

dot/rpc/modules/childstate_integration_test.go
EclesioMeloJunior: This is the childstate's integration test file, I will update it in another PR, this change is just to fix the license (same for other integration tests files)

dot/core/service.go
```diff
@@ -497,6 +497,10 @@ func (s *Service) HandleSubmittedExtrinsic(ext types.Extrinsic) error {
 		return nil
 	}
 
+	if s.transactionState.IsInPool(ext) {
+		return nil
+	}
```
noot, line 502: we should also check if it's in the queue (i don't think that's already checked?)

noot (commented):
looks good!! i think some core and babe tests are failing however

chain/gssmr/config.toml
EclesioMeloJunior: I put to be explicit but definitely we don't need

lib/runtime/test_helpers.go
```diff
@@ -161,3 +169,123 @@ func RemoveFiles(files []string) error {
 	}
 	return nil
 }
+
+// CreateTestExtrinsic builds a new extrinsic using centrifuge pkg
+func CreateTestExtrinsic(t *testing.T, rt Instance, genHash, blockHash common.Hash, nonce uint64, call string, args ...interface{}) string { //nolint
```
timwu20, line 174:
```suggestion
func newTestExtrinsic(t *testing.T, rt Instance, genHash, blockHash common.Hash, nonce uint64, call string, args ...interface{}) string { //nolint
```

timwu20 (changes requested):

dot/core/test_helpers.go
```diff
@@ -22,14 +24,40 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+// NewService2Test is ideal to Integration tests because creates a pointer to core.Service
+// given a configuration without doing extra checks or assign default values to nil dependencies
+func NewService2Test(ctx context.Context, t *testing.T, cfg *Config, blockAddCh chan *types.Block) *Service {
```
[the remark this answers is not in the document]
qdm12: Agreed! :+1: Also try to have everything test related with files ending with `_test.go` so the Go tooling distinguishes from production code.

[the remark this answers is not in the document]
EclesioMeloJunior: @timwu20 the `dot/core/Service` has package level fields whose the package `dot/rpc/modules` is not able to access so the function `NewService2Test` is in the `dot/core` package to access those fields and assign values to them, could I transform those fields (NewService2Test is a constructor without guards)

lib/runtime/test_helpers.go
EclesioMeloJunior: Updated to `NewTestExtrinsic` as it needs to be exported

dot/core/test_helpers.go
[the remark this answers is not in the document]
timwu20: Oh I see.  I'd still rename this function to something else.  `NewShellService` or something.

dot/core/service_test.go
```diff
@@ -38,6 +38,10 @@ func TestMain(m *testing.M) {
 		os.Exit(1)
 	}
 
+	for _, p := range wasmFilePaths {
+		fmt.Println(p)
+	}
+
```
timwu20, line 44: remove?

noot (approved):

Makefile
```diff
@@ -63,7 +63,7 @@ it-sync: build
 
 it-polkadotjs: build
 	@echo "  >  \033[32mRunning Integration Tests polkadot.js/api mode...\033[0m "
-	HOSTNAME=0.0.0.0 MODE=polkadot go test ./tests/polkadotjs_test/... -timeout=5m -v
+	HOSTNAME=0.0.0.0 MODE=polkadot go test ./tests/polkadotjs_test/... -timeout=5m -v -test.v
```
qdm12, line 66: What does `-test.v` do? :thinking:

tests/polkadotjs_test/start_polkadotjs_test.go
```diff
@@ -30,13 +30,23 @@ func TestStartGossamerAndPolkadotAPI(t *testing.T) {
...
 
-	command := "npx mocha ./test --timeout 30000"
+	node := nodes[0]
+
+	const command = "npx mocha ./test --timeout 30000"
+
+	err = os.Setenv("WSHOST", fmt.Sprintf("ws://127.0.0.1:%d", node.WSPort))
+	require.NoError(t, err)
+
+	defer os.Unsetenv("WSHOST")
```
qdm12, line 40:
```suggestion
	t.Setenv("WSHOST", fmt.Sprintf("ws://127.0.0.1:%d", node.WSPort))
```

tests/polkadotjs_test/start_polkadotjs_test.go
```diff
@@ -30,13 +30,23 @@ func TestStartGossamerAndPolkadotAPI(t *testing.T) {
...
+
 	parts := strings.Fields(command)
 	data, err := exec.Command(parts[0], parts[1:]...).Output()
-	assert.NoError(t, err, string(data))
 
+	fmt.Println(len(data))
 	//uncomment this to see log results from javascript tests
-	//fmt.Printf("%s\n", data)
+	fmt.Printf("%s\n", data)
```
qdm12, line 47: Assert or remove these?

tests/polkadotjs_test/start_polkadotjs_test.go
```diff
@@ -30,13 +30,23 @@ func TestStartGossamerAndPolkadotAPI(t *testing.T) {
...
+
+	assert.NoError(t, err, string(data))
```
qdm12, line 49: Is there a point showing the data if there is an error? I think the error should be enough information right?

```suggestion
	assert.NoError(t, err)
```

tests/polkadotjs_test/test/test-polkadot.js
```diff
@@ -11,7 +11,7 @@ describe('Testing polkadot.js/api calls:', function () {
     let done = false;
 
     before (async function () {
-        const wsProvider = new WsProvider('ws://127.0.0.1:8546');
+        const wsProvider = new WsProvider(process.env.WSHOST);
```
qdm12, line 14: Interesting. Maybe you could copy the js files to a `t.TempDir`, modify them there. That way you can run each test in parallel without depending on the shared env variable or shared files

Makefile
EclesioMeloJunior: Is similar to `-v` but I'm not sure ()https://pkg.go.dev/testing#T.Log)

tests/polkadotjs_test/start_polkadotjs_test.go
EclesioMeloJunior: the data contains information about what's going wrong on javascript test side, the err just shows: `command ... exit with non zero code`

EclesioMeloJunior:
@qdm12 I removed the code that touches on `polkadot-js` tests because is out of the scope of this PR

codecov.yml
```diff
@@ -1,7 +1,72 @@
...
+          - "lib/common"
+          - "lib/utils"
+          - "lib/genesis"
+          - "lib/keystore"
+          - "lib/genesis"
+
+comment:
+  layout: diff
+  behavior: once
+  require_changes: true
```
qdm12, line 69: This is merged into development, I'm not sure why it shows up here :thinking:

kishansagathiya (changes requested):
Mostly looks good. There are just a few parts, for which I am seeking clarification.

dot/core/service.go
```diff
@@ -497,6 +497,10 @@ func (s *Service) HandleSubmittedExtrinsic(ext types.Extrinsic) error {
 		return nil
 	}
 
+	if s.transactionState.Exists(ext) {
```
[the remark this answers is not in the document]
EclesioMeloJunior: This function is just to avoid calling multiple times the validation function to the same transaction, but you're right, there is no chance to have any duplicated in the pool!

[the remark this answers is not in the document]
kishansagathiya: Sorry, if I don't have the context here.

Do we have any logic because of which we might call this function multiple times for the same transaction? or are users of gossamers likely call this for the same transaction?

If that is not likely to happen, I would likely not have this check.

dot/rpc/modules/author_integration_test.go
```diff
@@ -179,148 +231,552 @@ func TestAuthorModule_SubmitExtrinsic_invalid_input(t *testing.T) {
...
+			ksType: "gran",
+			seed:   seed,
+			kp:     grandKp,
+		},
+
+		"invalid babe key type": {
+			ksType:  "babe",
+			seed:    seed,
+			kp:      "0x0000000000000000000000000000000000000000000000000000000000000000",
+			waitErr: errors.New("generated public key does not equal provide public key"),
```
kishansagathiya, line 327: Declare an error variable for this  in `dot/rpc/modules/author.go` and use that in tests (here and in author_test.go)

dot/rpc/modules/childstate_test.go
```diff
@@ -44,7 +44,44 @@ func createTestTrieState(t *testing.T) (*trie.Trie, common.Hash) {
...
+	expStr := []string{
+		"0x11f3ba2e1cdd6d62f2ff9b5589e7ff81ba7fb8745735dc3be2a2c61a72c39e78",
```
kishansagathiya, line 48: we have `tests/data`. I think places like that could be used to store such static data.

EclesioMeloJunior: This slice is only used by the `TestChildStateModule_GetKeys` and is not such a big data value and I think it is not worth putting it into a file just to load it in just one test function

kishansagathiya (approved):
:tada:

timwu20 (approved):

PR #2092, opened 2021-12-01
fix(dot/network): Fix missing digest in header

### LINKED ISSUE #2014: devnet: occassionally seeing error "header does not contain pre-runtime digest"
## Task summary

- occasionally on the devnet, I've seen the following error when a node tries to sync a block another node built:

```
EROR[11-10|23:29:05] ready block failed                       pkg=sync hash=0x7075f64377d6406f817055f833863af2f62c5ecf6b6cbb2b1b5261c42aba7e65 error="could not verify block: failed to get epoch for block header: header does not contain pre-runtime digest" caller=chain_processor.go:96
```

for some reason, the pre-runtime digest of the block is missing, so the node can't verify it.

there are a few possible reasons for this:
1. the block producer didn't add the pre-runtime digest. this would be really bad, and probably isn't the case, as the BABE code always adds it. however would be good to double check
2. some encoding/decoding error. potentially the block is missing the pre-runtime digest when it's being encoded on the sender side, or it's missing it when decoded on the receiver side.

arijitAD:
## Changes

- Fix missing digest in header

## Tests

```

```

## Issues

- closes #2014 

## Primary Reviewer

- @noot

### DIFF
--- a/dot/network/message.go
+++ b/dot/network/message.go
@@ -317,6 +317,8 @@ func protobufToBlockData(pbd *pb.BlockData) (*types.BlockData, error) {
 		}
 
 		bd.Header = header
+	} else {
+		bd.Header = nil
 	}
 
 	if pbd.Body != nil {
--- a/dot/network/message.go
+++ b/dot/network/message.go
@@ -8,12 +8,13 @@ import (
 	"errors"
 	"fmt"
 
+	"google.golang.org/protobuf/proto"
+
 	pb "github.com/ChainSafe/gossamer/dot/network/proto"
 	"github.com/ChainSafe/gossamer/dot/types"
 	"github.com/ChainSafe/gossamer/lib/common"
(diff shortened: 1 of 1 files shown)


dot/network/message.go
```diff
@@ -317,6 +317,8 @@ func protobufToBlockData(pbd *pb.BlockData) (*types.BlockData, error) {
 		}
 
 		bd.Header = header
+	} else {
+		bd.Header = nil
```
qdm12, line 321: Remove line `309` `Header: types.NewEmptyHeader(),` instead

qdm12 (dismissed):

EclesioMeloJunior (approved):

timwu20 (approved):
That was it? 😂

noot (approved):
wild if true haha, will test this out!!

Issue #2046, opened 2021-11-16
dot/network: close notifications streams for reading/writing when outbound/inbound respectively 


noot:
## Task summary

- after a handshake has been exchanged on  a notifications stream, the stream can be closing for reading if it's outbound or closing for writing if it's inbound
-  see https://pkg.go.dev/github.com/libp2p/go-libp2p-core@v0.11.0/mux#MuxedStream

PR #2093, opened 2021-12-01
fix(dot/network): close notifications streams

kishansagathiya:
## Changes

- close notifications streams for reading/writing when outbound/inbound
respectively

## Tests

```

```

## Issues

- https://github.com/ChainSafe/gossamer/issues/2046

## Primary Reviewer

- @noot


dot/network/notifications.go
```diff
@@ -152,9 +152,12 @@ func createDecoder(info *notificationsProtocol, handshakeDecoder HandshakeDecode
...
-func (s *Service) createNotificationsMessageHandler(info *notificationsProtocol,
+func (s *Service) createNotificationsMessageHandler(
+	info *notificationsProtocol,
 	messageHandler NotificationsMessageHandler,
-	batchHandler NotificationsMessageBatchHandler) messageHandler {
+	batchHandler NotificationsMessageBatchHandler,
+) messageHandler {
+
 	return func(stream libp2pnetwork.Stream, m Message) error {
 		if m == nil || info == nil || info.handshakeValidator == nil || messageHandler == nil {
```
EclesioMeloJunior, line 162: is it possible to add a defer to close this stream?

```go
return func(stream libp2pnetwork.Stream, m Message) error {
  defer func(){
    if err := stream.CloseWrite(); err != nil {
      logger.Tracef("failed to close stream for writing: %s", err)
    }
  }()
}
```

EclesioMeloJunior (changes requested):

dot/network/notifications.go
```diff
@@ -201,6 +204,7 @@ func (s *Service) createNotificationsMessageHandler(info *notificationsProtocol,
 				info.inboundHandshakeData.Store(peer, hsData)
 
 				// once validated, send back a handshake
+				// TODO: getHandshake creates a new handshake struct. Rename?
```
noot, line 207: `getHandshake` is just some function that satisfies the handshake getter function signature, it doesn't necessarily make a new struct, and could be different for each notifications protocol

noot (approved):
looks good!

edwardmack (approved):
Nice work, lgtm.

dot/network/notifications.go
```diff
@@ -152,9 +152,12 @@ func createDecoder(info *notificationsProtocol, handshakeDecoder HandshakeDecode
...
 
 // createNotificationsMessageHandler returns a function that is called by the handler of *inbound* streams.
-func (s *Service) createNotificationsMessageHandler(info *notificationsProtocol,
+func (s *Service) createNotificationsMessageHandler(
+	info *notificationsProtocol,
 	messageHandler NotificationsMessageHandler,
-	batchHandler NotificationsMessageBatchHandler) messageHandler {
+	batchHandler NotificationsMessageBatchHandler,
+) messageHandler {
+
```
qdm12, line 160: nit Let's get rid of the leading empty line perhaps

```suggestion
```

dot/network/notifications.go
qdm12: Do we want to close the stream in every return case :thinking: ? How about `CloseRead`?

EclesioMeloJunior: hmm, got it!

EclesioMeloJunior (approved):

dot/network/notifications.go
qdm12: @EclesioMeloJunior how about `CloseRead`?

EclesioMeloJunior: @qdm12 @kishansagathiya at line 354 from `dot/network/notifications.go` we open a stream, we could add a `defer` which will call the `CloseRead` when the function exits

kishansagathiya: `messageHandler` returned by `createNotificationsMessageHandler` is used in `readStream`.

`readStream` always resets the stream in the end using defer. When you reset the stream, the stream gets closed from both ends.

https://github.com/ChainSafe/gossamer/blob/f091a756d24663000b4ca99ba39e4aafbde95e84/dot/network/inbound.go#L14

Please unresolve, if you feel this still needs to be addressed.

Issue #1975, opened 2021-11-03
Fix flaky tests


qdm12:
Since some tests fail occasionally, it might be an indicator of bad logic which could cause memory leaks and/or data races.

These should be investigated. The following tests were found to fail occasionally:

- https://github.com/ChainSafe/gossamer/runs/3977236105?check_suite_focus=true
- https://github.com/ChainSafe/gossamer/runs/3977134279?check_suite_focus=true
- https://github.com/ChainSafe/gossamer/runs/3979841722?check_suite_focus=true
- https://github.com/ChainSafe/gossamer/runs/3980840080?check_suite_focus=true
- https://github.com/ChainSafe/gossamer/runs/4002865116?check_suite_focus=true
- https://github.com/ChainSafe/gossamer/runs/4009896511?check_suite_focus=true
- https://github.com/ChainSafe/gossamer/runs/4026380823?check_suite_focus=true
- Data races found in websocket package

EclesioMeloJunior:
I'm investigating the `TestBlockTree_GetHashByNumber`

qdm12:
Our saviour has arrived

timwu20:
@danforbes can we change this to an epic and we will will create issues for the checkboxes in the description.

PR #2094, opened 2021-12-01
fix(lib/blocktree): removes the inconsistency to choose a deepest leaf

EclesioMeloJunior:
### Quick explanation

Given the following block tree representation

```
           /-> d -> f
a -> b -> c -> d
      \-> c -> d -> f
```

considering the upper `f` and lower `f` have the same number and arrival time, the function `blocktree.leaves. deepestLeaf()` can return inconsistent output since `sync.Map` doesn't guarantee the same sequence every time.

So this inconsistency causes the problem in the test `TestBlockTree_GetHashByNumber` for the following reasons:
1. If the function `blocktree.GetHashByNumber` needs to find the block with number `c` and takes the upper `f` as its deepest leaf then the output will be the middle `c` block
2. However, the function `blocktree.IsDescendantOf` receives the lower `f` and middle `c` as parameters to check if `f` is descendant of `c` which is false because lower `f` is descendant of **lower `c`**

### Updates
- This PR adds a field called `currentDeepestLeaf` at `leafMap` struct which holds the deepest node and ignores the others one whose have the same number and arrival time

## Tests

```
WIP
```

## Issues

- Related #1975

## Primary Reviewer

- @qdm12

### DIFF
--- a/lib/blocktree/blocktree_test.go
+++ b/lib/blocktree/blocktree_test.go
@@ -494,13 +494,13 @@ func TestBlockTree_AllLeavesHasSameNumberAndArrivalTime_DeepestBlockHash_ShouldH
 	the leaves has the same number (8) the numbers in the right represents the order
 	the nodes are inserted into the blocktree.
 
-		a -> b -> c -> d -> e -> f -> g -> h (1)
-			 |    |    |    |    |    |->  h (7)
-			 |    |    |    |    |->  g -> h (6)
-		     |    |    |    |->  f -> g -> h (5)
-	         |    |    |->  e -> f -> g -> h (4)
-		     |    |->  d -> e -> f -> g -> h (3)
-		     |->  c -> d -> e -> f -> g -> h (2)
+	a -> b -> c -> d -> e -> f -> g -> h (1)
+		|    |    |    |    |    |> h (7)
+		|    |    |    |    |> g -> h (6)
+		|    |    |    |> f -> g -> h (5)
+		|    |    |> e -> f -> g -> h (4)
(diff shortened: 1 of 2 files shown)


lib/blocktree/blocktree_test.go
```diff
@@ -474,6 +477,111 @@ func TestBlockTree_GetHashByNumber(t *testing.T) {
...
+	const fixedArrivalTime = 99
+	const deep = 8
+
+	// create a base tree with a fixed amount of blocks
+	// and all block with the same arrival time
+
+	/**
+	base tree and nodes representation, all with the same arraival time and all
+	the leaves has the same number (8) the numbers in the right represents the order
+	the nodes are inserted into the blocktree.
```
1garo, line 495: typo at arraival

1garo, line 495: typo at arraival

lib/blocktree/blocktree_test.go
```diff
@@ -474,6 +477,111 @@ func TestBlockTree_GetHashByNumber(t *testing.T) {
...
+		// the last block on the base tree should not generates a branch
+		if i < deep {
+			branches = append(branches, testBranch{
+				hash:   hash,
+				number: bt.getNode(hash).number,
+			})
+		}
+	}
+
+	// create all the branch nodes with the same arraival time
```
1garo, line 529: typo here too

noot (approved):
this seems like it should work, thanks for the fix!

edwardmack (approved):
Great work, lgtm.

PR #2099, opened 2021-12-02
fix(dot/sync): add nil header checks

noot:
## Changes

- fix SIGABORT introduced by recent commit to development, add nil checks for now until the reason this is happening is fixed
-
-

## Tests

```
go test ./dot/sync
```

## Issues

- n/a

## Primary Reviewer

- @timwu20

### DIFF
--- a/dot/sync/chain_sync.go
+++ b/dot/sync/chain_sync.go
@@ -720,7 +720,7 @@ func handleReadyBlock(bd *types.BlockData, pendingBlocks DisjointBlockSet, ready
 	}
 
 	if bd.Header == nil {
-		logger.Criticalf("new ready block number (unknown) with hash %s", bd.Header.Number, bd.Hash)
+		logger.Criticalf("new ready block number (unknown) with hash %s", bd.Hash)
 		return
 	}
 
--- a/dot/sync/chain_sync.go
+++ b/dot/sync/chain_sync.go
@@ -719,7 +719,7 @@ func handleReadyBlock(bd *types.BlockData, pendingBlocks DisjointBlockSet, ready
 		bd.Header = block.header
 	}
 
-	if bd.Header != nil {
+	if bd.Header == nil {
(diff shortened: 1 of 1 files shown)


jimjbrettj (approved):
Great work!

dot/sync/chain_sync.go
```diff
@@ -711,10 +711,20 @@ func handleReadyBlock(bd *types.BlockData, pendingBlocks DisjointBlockSet, ready
 	// if we're expecting headers, validate should ensure we have a header
 	if bd.Header == nil {
 		block := pendingBlocks.getBlock(bd.Hash)
-		bd.Header = block.header
+		if block != nil {
```
EclesioMeloJunior, line 714: I would suggest to `if block == nil` then log and return, this will remove the need of `else` branch

EclesioMeloJunior (approved):

dot/sync/chain_sync.go
```diff
@@ -711,9 +711,19 @@ func handleReadyBlock(bd *types.BlockData, pendingBlocks DisjointBlockSet, ready
...
 		block := pendingBlocks.getBlock(bd.Hash)
+		if block == nil {
+			logger.Criticalf("block with unknown header is ready: hash=%s", bd.Hash)
+			return
+		}
+
 		bd.Header = block.header
 	}
 
+	if bd.Header != nil {
```
timwu20, line 722: I tried this branch on my local devnet and it wasn't finalising blocks.  After closer inspection, if `bd.Header == nil` and then the header is copied, wouldn't this if condition always be true?

timwu20 (changes requested):

dot/sync/chain_sync.go
noot: sorry i messed up the if statement here when i was updating it, should have been `== nil`, please pull again and retry!

PR #123, opened 2019-05-25
trie: improve generation of random tests

noot:
## Changes
- this improvement came to me at 6am this morning when I was half awake
- to prevent duplicate keys being randomly generated we store the previous keys created in that test in a mapping
- previously it was just a list which was slow, tests run quite a bit faster now

## Tests:
```
go test ./trie/
```

PR #2105, opened 2021-12-06
fix(dot/network): fix data race and add `t.Parallel` to all tests

### LINKED ISSUE #2044: dot/network: remove race conditions and parellelize tests
## Task summary

- there are currently race conditions in the network package as can be seen with `go test -race ./dot/network/...`
- fix these race conditions and update tests to run in parallel
- tests currently start up libp2p hosts, so there are conflicts between ports/basepaths between hosts. this will need to be resolved for tests to run in parallel

EclesioMeloJunior:
## Changes

- Removes all the race conditions at `dot/network` package

## Tests

```
go test ./dot/network -race -v -short
```

On apple silicon processors 
```
MallocNanoZone=0 go test ./dot/network -race -v -short
```

## Issues

- Closes #2044

## Primary Reviewer

- @timwu20


dot/network/utils_test.go
```diff
@@ -11,6 +11,27 @@ import (
...
+	*p <- port
+}
+
+var availablePorts portsQueue
+
+func init() {
+	availablePorts = make(portsQueue, portsAmount)
+	for port := 7001; port <= portsAmount; port++ {
+		availablePorts <- port
+	}
```
qdm12, line 32: Interesting way to do it! Although the problem is it will fail if you have anything else listening on port `7001` to `14200`.
Would there be a way to bind on port `0` and let the OS assign a free port?

For example to bind an HTTP server on a random available port and get the port back (for tests and/or logs), you can use:

```go
listener, err := net.Listen("tcp", ":0")
if err != nil {
	// ...
}
address := listener.Addr().String()

server := http.Server{Handler: myHandler}
// ...
err = server.Serve(listener)
```

This should work for all TCP listeners, and there is an equivalent API for UDP if that's what we're using. Maybe another PR to this branch would be interesting :thinking: That might also allow users to set `0` as the port to let the OS pick a random port and we can just log out the listening port.

dot/network/utils_test.go
```diff
@@ -11,6 +11,27 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+const portsAmount = 7100
+
+type portsQueue chan int
+
+func (p *portsQueue) get() int {
+	return <-*p
```
qdm12, line 19: no need to have it as a pointer receiver AFAIK since a channel is a pointer itself

dot/network/test_helpers.go
```diff
@@ -134,7 +135,10 @@ func (s *testStreamHandler) handleStream(stream libp2pnetwork.Stream) {
 func (s *testStreamHandler) handleMessage(stream libp2pnetwork.Stream, msg Message) error {
 	msgs := s.messages[stream.Conn().RemotePeer()]
 	s.messages[stream.Conn().RemotePeer()] = append(msgs, msg)
-	return s.writeToStream(stream, testBlockAnnounceHandshake)
+	announceHandshake := &BlockAnnounceHandshake{
+		BestBlockNumber: 0,
+	}
```
qdm12, line 140:
```suggestion
	announceHandshake := &BlockAnnounceHandshake{}
```

dot/network/test_helpers.go
```diff
@@ -190,12 +194,16 @@ var starting, _ = variadic.NewUint64OrHash(uint64(1))
 
 var one = uint32(1)
 
-var testBlockRequestMessage = &BlockRequestMessage{
-	RequestedData: RequestedDataHeader + RequestedDataBody + RequestedDataJustification,
-	StartingBlock: *starting,
-	EndBlockHash:  &common.Hash{},
-	Direction:     1,
-	Max:           &one,
+func NewTestBlockRequestMessage(t *testing.T) *BlockRequestMessage {
```
qdm12, line 197: Do we need to export this? Also, if not, can you rename `test_helpers.go` to `helpers_test.go` so it's treated as a test file?

dot/network/test_helpers.go
```diff
@@ -190,12 +194,16 @@ var starting, _ = variadic.NewUint64OrHash(uint64(1))
...
+	t.Helper()
+
+	return &BlockRequestMessage{
+		RequestedData: RequestedDataHeader + RequestedDataBody + RequestedDataJustification,
+		StartingBlock: *starting,
+		EndBlockHash:  &common.Hash{},
+		Direction:     1,
+		Max:           &one,
```
qdm12, line 205:
```suggestion
		Max:           uint32Ptr(1),
```

and define

```go
func uint32Ptr(n uint32) *uint32 { return &n }
```

So we avoid relying on global vars, especially for global vars pointed to

lib/utils/test_utils.go
```diff
@@ -32,25 +34,17 @@ func NewTestDir(t *testing.T) string {
...
-	}
+	t.Helper()
 
-	err = os.Mkdir(testDir, os.ModePerm)
-	if err != nil && !PathExists(testDir) {
-		fmt.Println(fmt.Errorf("failed to create test directory: %s", err))
-	}
+	basepathTmp := path.Join(t.TempDir(), t.Name(), name)
+	// create the subdirectories all the subdirectories
+	// os.MkdirAll is equivalent to mkdir -p /dir/subdir1/subdir2
```
qdm12, line 41: I don't think that's necessary, one can always check out `os.MkdirAll` documentation if they want.

```suggestion
```

dot/network/service_test.go
```diff
@@ -105,10 +108,7 @@ func createTestService(t *testing.T, cfg *Config) (srvc *Service) {
 
 	t.Cleanup(func() {
 		srvc.Stop()
```
qdm12, line 110: Check the error here? :thinking:

qdm12 (commented):
An early review, sorry I couldn't hold myself as I got excited with all this test parallelism :100: :smile:

EclesioMeloJunior:
blocked by PR #2106

dot/network/utils_test.go
EclesioMeloJunior: @qdm12 thanks for the suggestion, I changed to use `addr := net.ResolveTCPAddr("tcp", "localhost:0")` and `l := net.ListenTCP("tcp", addr)` then the OS can resolve an available port (`port := l.Addr().(*net.TCPAddr).Port`) for each call!

dot/network/utils_test.go
EclesioMeloJunior: removed!

dot/network/test_helpers.go
EclesioMeloJunior: file and function renamed!

dot/peerset/peerstate.go
```diff
@@ -108,6 +109,8 @@ type PeersState struct {
 	// configuration of each set. The size of this Info is never modified.
 	// since, single Info can also manage the flow.
 	sets []Info
+
+	mu *sync.Mutex
```
qdm12, line 113: Does this need to be a pointer to a mutex? Maybe it is, just asking :wink:

dot/peerset/peerstate.go
```diff
@@ -207,6 +211,21 @@ func (ps *PeersState) sortedPeers(idx int) peer.IDSlice {
 	return peerIDs
 }
 
+func (ps *PeersState) addReputation(pid peer.ID, change ReputationChange) (Reputation, error) {
+	ps.mu.Lock()
+	defer ps.mu.Unlock()
+
+	n, err := ps.getNode(pid)
+	if err != nil {
+		return 0, err
```
qdm12, line 220:
```suggestion
		return 0, fmt.Errorf("cannot get node for pid %s: %w"m pid, err)
```

internal/log/logger.go
```diff
@@ -34,6 +34,9 @@ func New(options ...Option) *Logger {
 // It can use a different writer, but it is expected to use the
 // same writer since it is thread safe.
 func (l *Logger) New(options ...Option) *Logger {
+	l.mutex.Lock()
+	defer l.mutex.Unlock()
```
qdm12, line 38: :pray: awesome, sorry again for that one!

test.log
```diff
@@ -0,0 +1,3257 @@
+=== RUN   TestEncodeBlockAnnounce
```
qdm12, line 1: remove here or in your other PR :wink: Up to you!

dot/network/utils_test.go
qdm12: Continued in https://github.com/ChainSafe/gossamer/pull/2105#discussion_r764713206

dot/network/utils_test.go
```diff
@@ -5,22 +5,35 @@ package network
...
+func availablePort(t *testing.T) uint16 {
+	addr, err := net.ResolveTCPAddr("tcp", "localhost:0")
+	require.NoError(t, err)
+
+	l, err := net.ListenTCP("tcp", addr)
+	require.NoError(t, err)
+	defer l.Close()
+
+	port := l.Addr().(*net.TCPAddr).Port
+	return uint16(port)
```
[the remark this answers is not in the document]
EclesioMeloJunior: got it! I was thinking about: first check if the default port is available otherwise get a free port from OS using the `*:0`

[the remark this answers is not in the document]
qdm12: You need to bind to an available port at the same time as you request it; otherwise if you release it back to the OS it might get binded by something else. Although now that I think about it, I'm not 100% sure how the OS handles port `0` for more than 2^16 parallel tests for example. Does it just block or does it panic 🤔 Anyway I doubt we would reach that many tests 😄

internal/log/logger.go
qdm12: Done in #2114 as well to accelerate the merge of the fix

dot/network/host.go
```diff
@@ -27,31 +27,43 @@ import (
...
+var (
+	privateIPs *ma.Filters
+)
+
+func init() {
+	var err error
+	privateIPs, err = newPrivateIPFilters()
+	if err != nil {
+		log.Panic(err)
+	}
```
qdm12, line 59: Please remove this and use 

```go
	return ip.IsPrivate() || ip.IsLoopback() ||
		ip.IsLinkLocalUnicast() || ip.IsLinkLocalMulticast()
```

to check if an IP is private. This is hugely faster, require no dependency and works with IPv4, IPv6 and IPv4 wrapped in IPv6.

Alternatively add missing private networks to your list.

dot/network/host.go
```diff
@@ -74,12 +85,9 @@ func newHost(ctx context.Context, cfg *Config) (*host, error) {
...
 		}
 	} else {
-		ip, err := pubip.Get()
-		if err != nil {
-			logger.Errorf("failed to get public IP error: %v", err)
-		} else {
-			logger.Debugf("got public IP", "IP", ip)
-			externalAddr, err = ma.NewMultiaddr(fmt.Sprintf("/ip4/%s/tcp/%d", ip, cfg.Port))
+		if strings.TrimSpace(publicIP) != "" {
+			logger.Debugf("got public IP", "IP", publicIP)
```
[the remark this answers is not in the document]
qdm12: Done in #2140

dot/network/helpers_test.go
```diff
@@ -18,15 +19,21 @@ import (
...
 
 type testStreamHandler struct {
+	sync.Mutex
+
 	messages map[peer.ID][]Message
 	decoder  messageDecoder
-	exit     bool
+
+	// exitChan is a notify chan that the readStream
+	// works and sucessfuly returns
```
timwu20, line 28:
```suggestion
	// works and sucessfully returns
```

dot/network/helpers_test.go
```diff
@@ -69,7 +79,7 @@ func (s *testStreamHandler) readStream(stream libp2pnetwork.Stream,
 	msgBytes := make([]byte, maxBlockResponseSize)
 
 	defer func() {
-		s.exit = true
+		close(s.exitChan)
```
timwu20, line 82:
```suggestion
defer close(s.exitChan)
```

dot/network/service.go
```diff
@@ -384,47 +386,56 @@ func (s *Service) logPeerCount() {
 	}
 }
 
-func (s *Service) publishNetworkTelemetry(done <-chan struct{}) {
+func (s *Service) publishTelemetry(done <-chan struct{}) {
```
timwu20, line 389: nit: I usually use `done` to signal when goroutines have exited.  I usually use `cancel` or `stop` to signal a goroutine to shutdown
```suggestion
func (s *Service) publishTelemetry(cancel <-chan struct{}) {
```

dot/network/service.go
```diff
@@ -384,47 +386,56 @@ func (s *Service) logPeerCount() {
...
+	o := s.host.bwc.GetBandwidthTotals()
+	s.telemetry.SendMessage(telemetry.NewBandwidth(o.RateIn, o.RateOut, s.host.peerCount()))
+}
 
-		finalised, err := s.blockState.GetHighestFinalisedHeader()
-		if err != nil {
-			continue
-		}
-		finalizedHash := finalised.Hash()
+func (s *Service) sentBlockIntervalTelemetry() (err error) {
```
timwu20, line 414:
```suggestion
func (s *Service) sendBlockIntervalTelemetry() (err error) {
```

dot/network/service_test.go
```diff
@@ -292,23 +295,84 @@ func TestBroadcastDuplicateMessage(t *testing.T) {
...
+	}
+
+	// Only one message will be sent.
 	for i := 0; i < 5; i++ {
 		nodeA.GossipMessage(announceMessage)
 		time.Sleep(time.Millisecond * 10)
 	}
-	require.Equal(t, 6, len(handler.messages[nodeA.host.id()]))
+
+	time.Sleep(time.Millisecond * 200)
```
timwu20, line 373: Is there a way to do this test without all these sleeps?

dot/peerset/handler.go
```diff
@@ -5,12 +5,15 @@ package peerset
...
 import (
 	"context"
+	"sync"
 
 	"github.com/libp2p/go-libp2p-core/peer"
 )
 
 // Handler manages peerSet.
 type Handler struct {
+	mu sync.Mutex
```
timwu20, line 15: why do we need a mutex for the channel?

dot/network/helpers_test.go
qdm12, line 28: Do you mean that the `readStream` function uses and closes when it returns? I'm a bit confused here

dot/network/host_test.go
```diff
@@ -273,24 +274,32 @@ func TestExistingStream(t *testing.T) {
 	require.NoError(t, err)
 
 	time.Sleep(TestMessageTimeout)
-	require.NotNil(t, handlerB.messages[nodeA.host.id()], "node B timeout waiting for message from node A")
+
+	messages, _ := handlerB.messagesFrom(nodeA.host.id())
+	assert.Len(t, messages, 1)
+	require.NotNil(t, messages, "node B timeout waiting for message from node A")
```
qdm12, line 280:
```suggestion
	require.NotNil(t, messages, "node B timed out waiting for message from node A")
```

qdm12, line 280: Should you instead check for the `ok` returned by `handlerB.messagesFrom(nodeA.host.id())`??

dot/network/host_test.go
```diff
@@ -273,24 +274,32 @@ func TestExistingStream(t *testing.T) {
...
 
 	// node B opens the stream to send the first message
 	stream, err = nodeB.host.send(addrInfoA.ID, nodeB.host.protocolID, testBlockReqMessage)
 	require.NoError(t, err)
 
 	time.Sleep(TestMessageTimeout)
-	require.NotNil(t, handlerA.messages[nodeB.host.id()], "node A timeout waiting for message from node B")
+	messages, _ = handlerA.messagesFrom(nodeB.host.id())
+	require.NotNil(t, messages, "node A timeout waiting for message from node B")
```
qdm12, line 295:
```suggestion
	require.NotNil(t, messages, "node A timed out waiting for message from node B")
```

dot/network/host_test.go
```diff
@@ -273,24 +274,32 @@ func TestExistingStream(t *testing.T) {
...
 
 	// node B uses the stream to send a second message
 	err = nodeB.host.writeToStream(stream, testBlockReqMessage)
 	require.NoError(t, err)
-	require.NotNil(t, handlerA.messages[nodeB.host.id()], "node A timeout waiting for message from node B")
+
+	messages, _ = handlerA.messagesFrom(nodeB.host.id())
+	require.NotNil(t, messages, "node A timeout waiting for message from node B")
```
qdm12, line 302:
```suggestion
	require.NotNil(t, messages, "node A timed out waiting for message from node B")
```

dot/network/host_test.go
```diff
@@ -548,14 +556,15 @@ func TestStreamCloseEOF(t *testing.T) {
...
 
 	err = stream.Close()
 	require.NoError(t, err)
 
-	time.Sleep(TestBackoffTimeout)
-
-	require.True(t, handler.exit)
+	select {
+	case <-time.After(time.Millisecond * 500):
+		require.Fail(t, "stream handler does not exit after stream closed")
```
qdm12, line 565:
```suggestion
		t.Fatal("stream handler does not exit after stream closed")
```

dot/network/host_test.go
```diff
@@ -548,14 +556,15 @@ func TestStreamCloseEOF(t *testing.T) {
...
-	require.False(t, handler.exit)
 
 	err = stream.Close()
 	require.NoError(t, err)
 
-	time.Sleep(TestBackoffTimeout)
-
-	require.True(t, handler.exit)
+	select {
+	case <-time.After(time.Millisecond * 500):
```
qdm12, line 564: nit not a big deal, but using a `timer.NewTimer(duration)` is better since `time.After` will leak and have its channel blocked forever if `case <-handler.exitChan:` triggers before it, since you cannot stop the underlying timer. Again this is a test so doesn't matter, but I felt like sHaRiNg 😉

dot/network/service.go
qdm12: I agree with Tim, call it `cancel` or `stop`. Or (Tim might hate on me for that), use a context 🤣

dot/network/service.go
```diff
@@ -384,47 +386,56 @@ func (s *Service) logPeerCount() {
...
+	best, err := s.blockState.BestBlockHeader()
+	if err != nil {
+		return err
```
qdm12, line 417:
```suggestion
		return fmt.Errorf("cannot get best block header: %w", err)
```

dot/network/service.go
```diff
@@ -384,47 +386,56 @@ func (s *Service) logPeerCount() {
...
-			finalised.Number,
-			big.NewInt(int64(s.transactionHandler.TransactionsCount())),
-			big.NewInt(0), // TODO: (ed) determine where to get used_state_cache_size (#1501)
-		))
+	bestHash := best.Hash()
 
-		time.Sleep(s.telemetryInterval)
+	finalised, err := s.blockState.GetHighestFinalisedHeader()
+	if err != nil {
+		return err
```
qdm12, line 424:
```suggestion
		return fmt.Errorf("cannot get highehst finalised header: %w", err)
```

dot/network/service_test.go
```diff
@@ -292,23 +295,84 @@ func TestBroadcastDuplicateMessage(t *testing.T) {
 		Digest: types.NewDigest(),
 	}
 
-	// Only one message will be sent.
+	// disable message cache before sent the messages
```
qdm12, line 298:
```suggestion
	// disable message cache before sending the messages
```

dot/network/stream_manager.go
```diff
@@ -23,20 +21,22 @@ type streamData struct {
 // This prevents keeping stale streams open and continuously trying to
 // read from it, which takes up lots of CPU over time.
 type streamManager struct {
-	ctx           context.Context
-	streamDataMap *sync.Map //map[string]*streamData
+	ctx                   context.Context
+	streamDataMap         *sync.Map //map[string]*streamData
```
qdm12, line 25: maybe a bit out of scope, but **SYNC.MAP** here

dot/peerset/handler.go
qdm12: Yes I'm curious as well. I get why for the `ps.nodes` map but not for the channel.

dot/network/helpers_test.go
noot: yeah i'd just say "readStream closes exitChan when it returns"

dot/peerset/handler.go
EclesioMeloJunior: The race condition tests were accusing a goroutine writing in the channel while another goroutine was closing the channel

dot/network/helpers_test.go
EclesioMeloJunior: I think I forgot some words 😆 🤣

dot/peerset/handler.go
timwu20: Hmmm.. that sounds like we need to refactor the code to avoid this.  A single producer (writer & closer) to the channel will alleviate this issue.

dot/network/service.go
```diff
@@ -384,47 +387,56 @@ func (s *Service) logPeerCount() {
...
-			finalised.Number,
-			big.NewInt(int64(s.transactionHandler.TransactionsCount())),
-			big.NewInt(0), // TODO: (ed) determine where to get used_state_cache_size (#1501)
-		))
+	bestHash := best.Hash()
 
-		time.Sleep(s.telemetryInterval)
+	finalised, err := s.blockState.GetHighestFinalisedHeader()
+	if err != nil {
+		return fmt.Errorf("cannot get highehst finalised header: %w", err)
```
jimjbrettj, line 455: *highest

dot/network/service_test.go
jimjbrettj: seconding this question

dot/network/stream_manager.go
jimjbrettj: SYNC.MAP??? Probably out of scope I would think. If we really are anti sync map maybe we should open an issue/epic to remove them from codebase?

dot/peerset/handler.go
```diff
@@ -30,67 +34,89 @@ func NewPeerSetHandler(cfg *ConfigSet) (*Handler, error) {
...
 
+func (h *Handler) setActionQueue(act action) {
+	go func(data action) {
+		h.writerWGMutex.Lock()
+		h.writersWG.Add(1)
+		h.writerWGMutex.Unlock()
+
+		defer h.writersWG.Done()
+
+		select {
```
jimjbrettj, line 45: I'm a bit confused what the purpose of these two selects are, why cant they just be one? Would you mind explaining just for my own learning?

dot/peerset/peerstate.go
```diff
@@ -212,15 +215,16 @@ func (ps *PeersState) sortedPeers(idx int) peer.IDSlice {
 
 func (ps *PeersState) addReputation(pid peer.ID, change ReputationChange) (
 	newReputation Reputation, err error) {
-	ps.mu.Lock()
-	defer ps.mu.Unlock()
-
 	n, err := ps.getNode(pid)
```
jimjbrettj, line 218: we don't need to lock before this call?

jimjbrettj (approved):
A few comments but overall looks good, great work! Network is a good place to remove race conditions so good stuff

dot/peerset/peerstate.go
EclesioMeloJunior: The `PeersState` causes race condition into `network` pkg because the `getNode` function access the `ps.nodes` map at the same time `addReputation` updates the map. I added `Mutex` to `PeersState` struct so `getNode` uses the `Lock/Unlock` to lock while reading and after `getNode` executes `addReputation` locks to write. It is not possible locking after `getNode` since `getNode` will starve waiting the unlock that will never be called, makes sense? 😅 

However the previous implementation is not good so I updated to use a `RWMutex` instead and remove the `getNode` call from `addReputation` which uses the `Lock/Unlock` and now `getNode` uses `RLock/RUnlock`, basically `getNode` will blocks if another goroutine calls the `addReputation` first

dot/network/host_test.go
```diff
@@ -694,6 +707,9 @@ func TestPeerReputation(t *testing.T) {
 	time.Sleep(100 * time.Millisecond)
 
 	rep, err := nodeA.host.cm.peerSetHandler.PeerReputation(addrInfoB.ID)
+
+	const zeroReputation int32 = 0
```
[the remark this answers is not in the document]
EclesioMeloJunior: honestly I think this const should not exist lol

dot/peerset/handler.go
EclesioMeloJunior: basically, the first one prevents later goroutines to write on the already closed channel, the second is when the close case is triggered while we are blocked trying to write in a full channel, but this code is a way to gracefully close a channel that is being used by other goroutines (since closing a channel while writing causes race condition).

dot/network/config.go
```diff
@@ -105,7 +105,9 @@ type Config struct {
 	SlotDuration time.Duration
 
 	Telemetry telemetry.Client
-	Metrics   metrics.IntervalConfig
+
+	MessageCacheTTL time.Duration
```
qdm12, line 109: nit I think you can have this unexported

dot/network/helpers_test.go
```diff
@@ -18,15 +19,19 @@ import (
 )
 
 type testStreamHandler struct {
+	sync.Mutex
```
qdm12, line 22: nit Maybe name the mutex so we know what it is protecting? Is it the `messages` map? Or something else?

dot/network/helpers_test.go
```diff
@@ -42,9 +47,11 @@ func (s *testStreamHandler) handleStream(stream libp2pnetwork.Stream) {
 }
 
 func (s *testStreamHandler) handleMessage(stream libp2pnetwork.Stream, msg Message) error {
+
```
qdm12, line 50:
```suggestion
```

dot/network/helpers_test.go
```diff
@@ -69,13 +76,10 @@ func (s *testStreamHandler) readStream(stream libp2pnetwork.Stream,
...
 	msgBytes := make([]byte, maxBlockResponseSize)
 
-	defer func() {
-		s.exit = true
-	}()
-
 	for {
 		tot, err := readStream(stream, msgBytes)
 		if errors.Is(err, io.EOF) {
+			s.eofCh <- struct{}{}
```
qdm12, line 82: Can you close it instead? I doubt you would get `EOF` twice right?

dot/network/host_test.go
```diff
@@ -73,16 +74,14 @@ func TestExternalAddrsPublicIP(t *testing.T) {
 		default:
 			require.False(t, privateIPs.AddrBlocked(addr))
 		}
-	}
 
-	expected := []ma.Multiaddr{
-		mustNewMultiAddr(fmt.Sprintf("/ip4/127.0.0.1/tcp/%d", port)),
-		mustNewMultiAddr(fmt.Sprintf("/ip4/10.0.5.2/tcp/%d", port)),
+		require.True(t, multiAddrRegex.MatchString(addr.String()))
```
qdm12, line 78: nit you can use `require.Regexp(t, "myregex", "mystring")`

dot/network/host_test.go
```diff
@@ -548,14 +554,20 @@ func TestStreamCloseEOF(t *testing.T) {
 
 	stream, err := nodeA.host.send(addrInfoB.ID, nodeB.host.protocolID, testBlockReqMessage)
 	require.NoError(t, err)
-	require.False(t, handler.exit)
 
 	err = stream.Close()
 	require.NoError(t, err)
 
-	time.Sleep(TestBackoffTimeout)
+	timeout := time.NewTimer(TestBackoffTimeout)
```
qdm12, line 561: nit rename to `timer`

dot/network/host_test.go
```diff
@@ -548,14 +554,20 @@ func TestStreamCloseEOF(t *testing.T) {
...
 
-	require.True(t, handler.exit)
+	select {
+	case <-timeout.C:
+		t.Fatal("stream handler does not exit after stream closed")
```
qdm12, line 565: nit

```suggestion
		t.Fatal("stream handler did not exit after stream closed")
```

dot/network/notifications.go
```diff
@@ -310,7 +311,7 @@ func (s *Service) sendData(peer peer.ID, hs Handshake, info *notificationsProtoc
 		}
 	}
 
-	logger.Tracef("successfully sent message on protocol %s to peer %s: message=", info.protocolID, peer, msg)
+	logger.Debugf("successfully sent message on protocol %s to peer %s: message=", info.protocolID, peer, msg)
```
qdm12, line 314: Why the level change here 🤔 ? (applies to above as well)

dot/network/service.go
```diff
@@ -414,47 +417,56 @@ func (s *Service) logPeerCount() {
...
-func (s *Service) publishNetworkTelemetry(done <-chan struct{}) {
+func (s *Service) publishTelemetry(cancel <-chan struct{}) {
 	ticker := time.NewTicker(s.telemetryInterval)
 	defer ticker.Stop()
 
 	for {
+		s.sentBandwidthTelemetry()
+		err := s.sendBlockIntervalTelemetry()
+		if err != nil {
+			logger.Warnf("failed to sent block interval telemetry: %s", err)
```
qdm12, line 428: nit

```suggestion
			logger.Warnf("failed to send block interval telemetry: %s", err)
```

dot/network/service.go
```diff
@@ -414,47 +417,56 @@ func (s *Service) logPeerCount() {
...
-func (s *Service) sentBlockIntervalTelemetry() {
-	for {
-		best, err := s.blockState.BestBlockHeader()
-		if err != nil {
-			continue
-		}
-		bestHash := best.Hash()
+func (s *Service) sentBandwidthTelemetry() {
+	o := s.host.bwc.GetBandwidthTotals()
+	s.telemetry.SendMessage(telemetry.NewBandwidth(o.RateIn, o.RateOut, s.host.peerCount()))
```
qdm12, line 442: nit rename `o` to something else. I guess `totals`?

```suggestion
	totals := s.host.bwc.GetBandwidthTotals()
	s.telemetry.SendMessage(telemetry.NewBandwidth(totals.RateIn, totals.RateOut, s.host.peerCount()))
```

dot/network/service.go
```diff
@@ -414,47 +417,56 @@ func (s *Service) logPeerCount() {
...
+	finalizedHash := finalised.Hash()
+
+	s.telemetry.SendMessage(telemetry.NewBlockInterval(
+		&bestHash,
+		best.Number,
+		&finalizedHash,
+		finalised.Number,
+		big.NewInt(int64(s.transactionHandler.TransactionsCount())),
+		big.NewInt(0), // TODO: (ed) determine where to get used_state_cache_size (#1501)
+	))
```
qdm12, line 467: nit it would be nice to split this in two, one instruction for `blockInterval := ...` and the other for `s.telemetry.SendMessage(blockInterval)`

dot/network/service_test.go
```diff
@@ -237,35 +237,36 @@ func TestBroadcastMessages(t *testing.T) {
 	// simulate message sent from core service
 	nodeA.GossipMessage(anounceMessage)
 	time.Sleep(time.Second * 2)
-	require.NotNil(t, handler.messages[nodeA.host.id()])
+
+	messages, _ := handler.messagesFrom(nodeA.host.id())
+	require.NotNil(t, messages)
```
qdm12, line 242: Maybe not empty? Could you also assert we get predetermined messages as well?

```suggestion
	require.NotEmpty(t, messages)
```

dot/network/service_test.go
```diff
@@ -237,35 +237,36 @@ func TestBroadcastMessages(t *testing.T) {
...
 }
 
-func TestBroadcastDuplicateMessage(t *testing.T) {
+func Test_Broadcast_Duplicate_Messages_WithDisabled_MessageCache(t *testing.T) {
```
qdm12, line 245: supernit

```suggestion
func Test_Broadcast_Duplicate_Messages_With_Disabled_MessageCache(t *testing.T) {
```

dot/network/service_test.go
```diff
@@ -237,35 +237,36 @@ func TestBroadcastMessages(t *testing.T) {
...
+		Port:            availablePort(t),
+		NoBootstrap:     true,
+		NoMDNS:          true,
+		MessageCacheTTL: 2 * time.Second,
 	}
 
 	nodeB := createTestService(t, configB)
 	nodeB.noGossip = true
 
-	// TODO: create a decoder that handles both handshakes and messages
```
qdm12, line 268: Is this done? 🤔

dot/network/service_test.go
```diff
@@ -294,26 +295,81 @@ func TestBroadcastDuplicateMessage(t *testing.T) {
 		Digest: types.NewDigest(),
 	}
 
-	delete(handler.messages, nodeA.host.id())
+	// disable message cache before sending the messages
+	nodeA.host.messageCache = nil
 
-	// Only one message will be sent.
+	// All 5 message will be sent since cache is disabled.
```
qdm12, line 301: nit

```suggestion
	// All 5 messages will be sent since the cache is disabled.
```

dot/network/service_test.go
```diff
@@ -294,26 +295,81 @@ func TestBroadcastDuplicateMessage(t *testing.T) {
...
 	for i := 0; i < 5; i++ {
 		nodeA.GossipMessage(announceMessage)
 		time.Sleep(time.Millisecond * 10)
 	}
 
-	time.Sleep(time.Millisecond * 500)
-	require.Equal(t, 2, len(handler.messages[nodeA.host.id()]))
+	messages, _ := handler.messagesFrom(nodeA.host.id())
+	require.Len(t, messages, 6)
```
qdm12, line 308: Same question here, can we assert `messages` entirely with `assert.Equal`? Can we predict the messages received?

dot/network/service_test.go
```diff
@@ -294,26 +295,81 @@ func TestBroadcastDuplicateMessage(t *testing.T) {
...
+	nodeB := createTestService(t, configB)
+	nodeB.noGossip = true
+
+	handler := newTestStreamHandler(testBlockAnnounceHandshakeDecoder)
+	nodeB.host.registerStreamHandler(nodeB.host.protocolID+blockAnnounceID, handler.handleStream)
+
+	addrInfoB := nodeB.host.addrInfo()
+	err := nodeA.host.connect(addrInfoB)
+	// retry connect if "failed to dial" error
+	if failedToDial(err) {
```
qdm12, line 342: Isn't this just a `errors.Is`?

dot/network/service_test.go
qdm12: Thirding the question

dot/network/service_test.go
```diff
@@ -294,26 +295,81 @@ func TestBroadcastDuplicateMessage(t *testing.T) {
...
+	// Only one message will be sent.
 	for i := 0; i < 5; i++ {
 		nodeA.GossipMessage(announceMessage)
 		time.Sleep(time.Millisecond * 10)
 	}
 
-	require.Equal(t, 7, len(handler.messages[nodeA.host.id()]))
+	time.Sleep(time.Millisecond * 200)
+	messages, _ := handler.messagesFrom(nodeA.host.id())
+	require.Len(t, messages, 2)
```
qdm12, line 372: Can we assert the messages entirely?

dot/peerset/handler.go
```diff
@@ -5,15 +5,19 @@ package peerset
...
 type Handler struct {
-	actionQueue chan<- action
-	peerSet     *PeerSet
-	closeCh     chan struct{}
+	actionQueue   chan<- action
+	closeCh       chan struct{}
+	writersWG     sync.WaitGroup
+	writerWGMutex sync.Mutex
+
+	peerSet *PeerSet
```
qdm12, line 20: nit just putting it back in the original order for less deltas. I'm not sure why it got changed.


```suggestion
	actionQueue   chan<- action
	peerSet     *PeerSet
	closeCh       chan struct{}
	writersWG     sync.WaitGroup
	writerWGMutex sync.Mutex
```

dot/peerset/handler.go
```diff
@@ -30,67 +34,89 @@ func NewPeerSetHandler(cfg *ConfigSet) (*Handler, error) {
...
+			return
+		default:
+		}
+
+		select {
+		case <-h.closeCh:
+		case h.actionQueue <- data:
+		}
+	}(act)
+
```
qdm12, line 56:
```suggestion
```

dot/peerset/handler.go
```diff
@@ -30,67 +34,89 @@ func NewPeerSetHandler(cfg *ConfigSet) (*Handler, error) {
 	}, nil
 }
 
+func (h *Handler) setActionQueue(act action) {
+	go func(data action) {
```
qdm12, line 38: Change this to not launch a goroutine here, and instead have the caller call `go setActionQueue(act)`

dot/peerset/handler.go
```diff
@@ -30,67 +34,89 @@ func NewPeerSetHandler(cfg *ConfigSet) (*Handler, error) {
...
+		h.writerWGMutex.Lock()
+		h.writersWG.Add(1)
+		h.writerWGMutex.Unlock()
```
qdm12, line 41: You should add to the writes wait group before launching the goroutine. And also the mutex protecting the waitgroup looks sketchy, is it really needed?

dot/peerset/handler.go
qdm12: Ooofff this bad. You don't have any guarantee `h.closeCh` won't get closed between the two select blocks (and hence closing `h.actionQueue`). Please change somehow.

dot/peerset/handler.go
```diff
@@ -131,11 +157,11 @@ func (h *Handler) Start(ctx context.Context) {
 // SortedPeers return chan for sorted connected peer in the peerSet.
 func (h *Handler) SortedPeers(setIdx int) chan peer.IDSlice {
 	resultPeersCh := make(chan peer.IDSlice)
-	h.actionQueue <- action{
+	h.setActionQueue(action{
```
qdm12, line 160: perhaps rename to `addToActionQueue`?

dot/peerset/handler.go
```diff
@@ -147,6 +173,11 @@ func (h *Handler) Stop() {
 	default:
 		h.cancelCtx()
 		close(h.closeCh)
+
+		h.writerWGMutex.Lock()
+		h.writersWG.Wait()
+		h.writerWGMutex.Unlock()
```
qdm12, line 179: A mutex for the wait group looks very strange, try to find another way 🤔 ?

dot/peerset/peerstate.go
```diff
@@ -171,6 +175,9 @@ func (ps *PeersState) peerStatus(set int, peerID peer.ID) string {
 
 // peers return the list of all the peers we know of.
 func (ps *PeersState) peers() []peer.ID {
+	ps.mu.RLock()
```
qdm12, line 178: perhaps name `mu` with a name so we know what it's protecting? I guess `peerStateMutex`??

dot/network/host_test.go
```diff
@@ -273,24 +271,32 @@ func TestExistingStream(t *testing.T) {
 	require.NoError(t, err)
 
 	time.Sleep(TestMessageTimeout)
```
timwu20, line 273: can we refactor this test to not use sleeps anymore? Now that you have `testStreamHandler`, you can write messages into an internal channel from `testStreamHandler.handleMessage` and just listen on that channel.

dot/network/host_test.go
timwu20: I don't mind this naming.

dot/network/service_test.go
```diff
@@ -237,35 +237,36 @@ func TestBroadcastMessages(t *testing.T) {
 	// simulate message sent from core service
 	nodeA.GossipMessage(anounceMessage)
 	time.Sleep(time.Second * 2)
```
timwu20, line 239: can we remove sleep?

dot/network/host_test.go
qdm12: I guess it could be a *timeout timer* as well 😄

dot/network/service.go
```diff
@@ -414,47 +417,56 @@ func (s *Service) logPeerCount() {
...
+func (s *Service) publishTelemetry(cancel <-chan struct{}) {
 	ticker := time.NewTicker(s.telemetryInterval)
 	defer ticker.Stop()
 
 	for {
+		s.sentBandwidthTelemetry()
+		err := s.sendBlockIntervalTelemetry()
+		if err != nil {
+			logger.Warnf("failed to sent block interval telemetry: %s", err)
+			continue
```
qdm12, line 429: So if we fail, we instantly try again; should we instead wait for the next tick? Or at least some backoff? I'm just thinking for example if the network goes down, this loop might get got

noot: yeah, this should just fall through to the `select` statement

dot/network/stream_manager.go
qdm12: :smile: issue is #2281 and PR for that sync.Map is #2366 , feel free to review that one too!

timwu20:
@EclesioMeloJunior can you resolve the conflicts?

EclesioMeloJunior:
> @EclesioMeloJunior can you resolve the conflicts?

Of course, once I finish PR #123 which this one depends on

timwu20:
@EclesioMeloJunior can you change this to draft if you don't intend to merge this.

PR #2106, opened 2021-12-07
fix(dot/telemetry): remove singleton and data races

EclesioMeloJunior:
## Changes

- Currently, the telemetry package uses a singleton to handle and dispatch messages. However, that approach is causing data race conditions because 2 goroutines are trying to get the global `instance` value, and while one is checking if the global `instance` value is nil the other one is trying to assign a value to the global `instance`

- This PR proposes the removal of singleton instance and use a channel to receive and inject messages into the `messager dispatch` goroutine

## Tests

```
go test github.com/ChainSafe/gossamer/dot/telemetry -v -race
```

This approach fixes the telemetry data race conditions at `dot/network` package

## Issues

- Related to PR #2105

## Primary Reviewer

- @qdm12


chain/polkadot/config.toml
```diff
@@ -5,7 +5,7 @@ metrics-port = 9876
 
 [log]
 core = ""
-network = ""
+network = "debug"
```
qdm12, line 8:
```suggestion
network = ""
```

dot/telemetry/mailer.go
```diff
@@ -0,0 +1,132 @@
...
+
+	"github.com/ChainSafe/gossamer/internal/log"
+	"github.com/ChainSafe/gossamer/lib/genesis"
+	"github.com/gorilla/websocket"
+)
+
+const (
+	defaultMessageTimeout = time.Second
+	defaultMaxRetries     = 5
+	defaultRetryDelay     = time.Second * 15
```
qdm12, line 18: Please inline this one as a constant in `BootstrapMailer` since it's only used there. You could place it right before passing it to the timer. Also rename it, it's not really a `default`.

Same comment for the other 2 constants above

dot/telemetry/mailer.go
```diff
@@ -0,0 +1,132 @@
...
+)
+
+var messageReceiver chan Message = make(chan Message, 256)
```
qdm12, line 21: Maybe rename it to messageQueue?

dot/telemetry/mailer.go
```diff
@@ -0,0 +1,132 @@
...
+
+type telemetryConnection struct {
+	wsconn    *websocket.Conn
+	verbosity int
+	sync.Mutex
+}
+
+// Handler struct for holding telemetry related things
+type mailer struct {
+	msg         chan Message
```
qdm12, line 31: Maybe rename it to `messageQueue`?

dot/telemetry/mailer.go
```diff
@@ -0,0 +1,132 @@
...
+	connections []*telemetryConnection
+	log         log.LeveledLogger
```
qdm12, line 33: This should be `logger`

dot/telemetry/mailer.go
```diff
@@ -0,0 +1,132 @@
...
+}
+
+func newMailer() *mailer {
+	return &mailer{
+		msg: messageReceiver,
+		log: log.NewFromGlobal(log.AddContext("pkg", "telemetry")),
```
qdm12, line 39: Perhaps dependency inject it to `newMailer`? You can create it with `log.NewFromGlobal` from calling layers

dot/telemetry/mailer.go
```diff
@@ -0,0 +1,132 @@
...
+// BootstrapMailer setup the mailer, the connections and start the async message shipment
+func BootstrapMailer(ctx context.Context, conns []*genesis.TelemetryEndpoint) {
+	mlr := newMailer()
+
+	for _, v := range conns {
+		for connAttempts := 0; connAttempts < defaultMaxRetries; connAttempts++ {
+			c, _, err := websocket.DefaultDialer.Dial(v.Endpoint, nil)
+			if err != nil {
+				mlr.log.Debugf("issue adding telemetry connection: %s", err)
+				time.Sleep(defaultRetryDelay)
```
qdm12, line 52: use a timer so it can be stopped if the context gets canceled

dot/telemetry/mailer.go
```diff
@@ -0,0 +1,132 @@
...
+func newMailer() *mailer {
+	return &mailer{
+		msg: messageReceiver,
+		log: log.NewFromGlobal(log.AddContext("pkg", "telemetry")),
+	}
+}
+
+// BootstrapMailer setup the mailer, the connections and start the async message shipment
+func BootstrapMailer(ctx context.Context, conns []*genesis.TelemetryEndpoint) {
+	mlr := newMailer()
```
qdm12, line 45: please rename it to `mailer`

dot/telemetry/mailer.go
```diff
@@ -0,0 +1,132 @@
...
+
+	for _, v := range conns {
+		for connAttempts := 0; connAttempts < defaultMaxRetries; connAttempts++ {
+			c, _, err := websocket.DefaultDialer.Dial(v.Endpoint, nil)
+			if err != nil {
+				mlr.log.Debugf("issue adding telemetry connection: %s", err)
```
qdm12, line 51:
```suggestion
				mlr.log.Debugf("cannot dial telemetry endpoint %s (try %d of %d): %s",
					v.Endpoint, connAttempts+1, defaultMaxRetries, err)
```

dot/telemetry/mailer.go
```diff
@@ -0,0 +1,132 @@
...
+	}
+}
+
+// BootstrapMailer setup the mailer, the connections and start the async message shipment
+func BootstrapMailer(ctx context.Context, conns []*genesis.TelemetryEndpoint) {
+	mlr := newMailer()
+
+	for _, v := range conns {
+		for connAttempts := 0; connAttempts < defaultMaxRetries; connAttempts++ {
+			c, _, err := websocket.DefaultDialer.Dial(v.Endpoint, nil)
```
qdm12, line 49: please rename `c` to `websocketConn`, or at least `conn` :wink:

dot/telemetry/mailer.go
```diff
@@ -0,0 +1,132 @@
...
+
+			mlr.connections = append(mlr.connections, &telemetryConnection{
+				wsconn:    c,
+				verbosity: v.Verbosity,
+			})
+			break
+		}
+	}
+
+	go mlr.asyncShipment(ctx)
```
qdm12, line 64: We should pass a ready channel which can be closed inside this goroutine, to signal to the caller of `BootstrapMailer` we're ready to process

dot/telemetry/mailer.go
```diff
@@ -0,0 +1,132 @@
...
+
+// SendMessage sends Message to connected telemetry listeners throught messageReceiver
+func SendMessage(msg Message) error {
+	t := time.NewTimer(defaultMessageTimeout)
+	defer t.Stop()
+
+	select {
+	case messageReceiver <- msg:
+	case <-t.C:
+		return errors.New("timeout sending message")
```
qdm12, line 75: define this error as a global exported var

dot/telemetry/mailer.go
```diff
@@ -0,0 +1,132 @@
...
+	go mlr.asyncShipment(ctx)
+}
+
+// SendMessage sends Message to connected telemetry listeners throught messageReceiver
+func SendMessage(msg Message) error {
+	t := time.NewTimer(defaultMessageTimeout)
+	defer t.Stop()
+
+	select {
+	case messageReceiver <- msg:
```
qdm12, line 73:
```suggestion
	case messageReceiver <- msg:
		if !t.Stop() {
			<-t.C
		}
```

dot/telemetry/mailer.go
```diff
@@ -0,0 +1,132 @@
...
+			break
+		}
+	}
+
+	go mlr.asyncShipment(ctx)
+}
+
+// SendMessage sends Message to connected telemetry listeners throught messageReceiver
+func SendMessage(msg Message) error {
+	t := time.NewTimer(defaultMessageTimeout)
```
qdm12, line 69: Maybe rename `t` to `timer`

dot/telemetry/mailer.go
```diff
@@ -0,0 +1,132 @@
...
+	}
+	return nil
+}
+
+func (m *mailer) asyncShipment(ctx context.Context) {
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case msg, ok := <-m.msg:
```
qdm12, line 85: Is `m.msg` ever closed by anything?

dot/telemetry/mailer.go
```diff
@@ -0,0 +1,132 @@
...
+				for _, conn := range m.connections {
+					conn.Lock()
+					defer conn.Unlock()
+
+					err = conn.wsconn.WriteMessage(websocket.TextMessage, msgBytes)
+					if err != nil {
+						m.log.Debugf("issue while sending telemetry message: %s", err)
+					}
+				}
+			}(msg)
```
qdm12, line 106: Define this outside this function

dot/telemetry/mailer.go
```diff
@@ -0,0 +1,132 @@
...
+		}
+	}
+}
+
+func (h *mailer) msgToJSON(message Message) ([]byte, error) {
```
qdm12, line 111: Instead have another structure like

```go
type messageWrapper struct {
	MessageType string `json:"msg"`
	Timestamp time.Time `json:"ts"`
	Message
}
```

Set the fields and then marshal it once.

dot/telemetry/mailer_test.go
```diff
@@ -37,7 +38,8 @@ func TestMain(m *testing.M) {
 		Endpoint:  "ws://127.0.0.1:8001/",
 		Verbosity: 0,
 	}
-	GetInstance().AddConnections(append(testEndpoints, testEndpoint1))
+
+	BootstrapMailer(context.TODO(), append(testEndpoints, testEndpoint1))
```
qdm12, line 42:
```suggestion
	BootstrapMailer(context.Background(), append(testEndpoints, testEndpoint1))
```

dot/telemetry/mailer_test.go
```diff
@@ -108,7 +110,8 @@ func TestHandler_SendMulti(t *testing.T) {
 	for _, message := range messages {
 		wg.Add(1)
 		go func(msg Message) {
-			GetInstance().SendMessage(msg)
+			//GetInstance().SendMessage(msg)
```
qdm12, line 113:
```suggestion
```

dot/telemetry/mailer_test.go
```diff
@@ -145,7 +148,9 @@ func TestListenerConcurrency(t *testing.T) {
 	for i := 0; i < qty; i++ {
 		go func() {
 			bestHash := common.Hash{}
-			GetInstance().SendMessage(NewBlockImportTM(&bestHash, big.NewInt(2), "NetworkInitialSync"))
+			//GetInstance().SendMessage(NewBlockImportTM(&bestHash, big.NewInt(2), "NetworkInitialSync"))
```
qdm12, line 151:
```suggestion
```

test.log
```diff
@@ -1,3257 +0,0 @@
-=== RUN   TestEncodeBlockAnnounce
-=== PAUSE TestEncodeBlockAnnounce
```
qdm12, line 2: :smile: WHY WAS THIS COMMITTED :laughing:

dot/telemetry/mailer_test.go
```diff
@@ -145,7 +148,9 @@ func TestListenerConcurrency(t *testing.T) {
...
+			msg := NewBlockImportTM(&bestHash, big.NewInt(2), "NetworkInitialSync")
+			SendMessage(msg)
```
qdm12, line 153: A bit out of scope, so feel free to ignore. Just some knowledge sharing!

That could be improved such that:

- all goroutines start when they are all ready
- they all run for a fixed mount of time i.e. 50ms

Example I used to test logger data races 

https://github.com/ChainSafe/gossamer/blob/dc8bbefe8acc439be5f08491d8341daa9f63812f/internal/log/race_test.go#L31-L64

test.log
EclesioMeloJunior: removed lol

dot/telemetry/mailer.go
EclesioMeloJunior: Nice!

dot/telemetry/mailer.go
EclesioMeloJunior: I don't think we need to pass a ready function because `asyncShipment` does not have any conditions to start process the channel,  and the channel is already created by calling `BootstrapMailer` so I think we must change the `BootstrapMailer` to return an error that means: _you told us to start the telemetry but we are not able to bootstrap the async shipment then the `SendMessage` function will put your message into the channel however we won't send your message_

dot/telemetry/mailer.go
EclesioMeloJunior: Why is this needed if I already have defined `defer t.Stop()` right after the timer creation?

.github/workflows/release.yml
```diff
@@ -14,7 +14,7 @@ jobs:
         with:
           fetch-depth: 0
       - name: Setup Node.js
-        uses: actions/setup-node@v2.4.1
+        uses: actions/setup-node@v2.5.0
```
qdm12, line 17: Maybe have this in another PR? :thinking:

CHANGELOG.md
```diff
@@ -0,0 +1,149 @@
+# Semantic Versioning Changelog
```
qdm12, line 1: Is this file related to the PR?

dot/node.go
```diff
@@ -318,33 +318,41 @@ func NewNode(cfg *Config, ks *keystore.GlobalKeystore) (*Node, error) {
...
-		&genesisHash,
-		sysSrvc.SystemName(),
-		cfg.Global.Name,
-		networkSrvc.NetworkState().PeerID,
-		strconv.FormatInt(time.Now().UnixNano(), 10),
-		sysSrvc.SystemVersion()))
-	if err != nil {
-		logger.Debugf("problem sending system.connected telemetry message: %s", err)
+	if !cfg.Global.NoTelemetry {
+		telemetryLogger := logger.New(log.AddContext("pkg", "telemetry"))
```
qdm12, line 332:
```suggestion
		telemetryLogger := logger.NewFromGlobal(log.AddContext("pkg", "telemetry"))
```

Otherwise it would race on os.Stdout (and/or interleave logs)

dot/node.go
```diff
@@ -318,33 +318,41 @@ func NewNode(cfg *Config, ks *keystore.GlobalKeystore) (*Node, error) {
...
+
+		err := telemetry.BootstrapMailer(context.Background(), telemetryEndpoints, telemetryLogger)
+		if err != nil {
+			return nil, err
```
qdm12, line 336:
```suggestion
			return nil, fmt.Errorf("cannot bootstrap mailer: %w", err)
```

dot/telemetry/mailer.go
```diff
@@ -0,0 +1,152 @@
...
+	"sync"
+	"time"
+
+	"github.com/ChainSafe/gossamer/internal/log"
+	"github.com/ChainSafe/gossamer/lib/genesis"
+	"github.com/gorilla/websocket"
+)
+
+var (
+	ErrInsufficientConnections = errors.New("not able to connect to any telemetry endpoints")
```
qdm12, line 19: error name and error message don't really match I think

dot/telemetry/mailer.go
```diff
@@ -0,0 +1,152 @@
...
+		messageQueue: messageQueue,
+		logger:       logger,
+	}
+}
+
+// BootstrapMailer setup the mailer, the connections and start the async message shipment
+func BootstrapMailer(ctx context.Context, conns []*genesis.TelemetryEndpoint, logger log.LeveledLogger) error {
+	const (
+		maxRetries = 5
+		retryDelay = time.Second * 15
```
qdm12, line 54: You could move this down right above `timer := time.NewTimer(retryDelay)` to make it even more local. Forgive me I'm a local-everything maniac. Also compilation wise it doesn't change since it's a constant.

dot/telemetry/mailer.go
```diff
@@ -0,0 +1,152 @@
...
+
+				timer := time.NewTimer(retryDelay)
+
+				select {
+				case <-timer.C:
+					continue
+				case <-ctx.Done():
+					mailer.logger.Debugf("bootstrap telemetry issue: %w", ctx.Err())
+
+					timer.Stop()
```
qdm12, line 74:
```suggestion
					if !timer.Stop() {
						<-timer.C
					}
```

in case the timer triggered at the 'same time' as the context got canceled.

dot/telemetry/mailer.go
```diff
@@ -0,0 +1,152 @@
...
+	go mailer.asyncShipment(ctx)
+	return nil
+}
+
+// SendMessage sends Message to connected telemetry listeners throught messageReceiver
+func SendMessage(msg Message) error {
+	const messageTimeout = time.Second
+
+	timer := time.NewTimer(messageTimeout)
+	defer timer.Stop()
```
qdm12, line 100:
```suggestion
```

dot/telemetry/mailer.go
```diff
@@ -0,0 +1,152 @@
...
+	telemetryMsg := telemetryMessage{
+		msg.messageType(),
+		time.Now(),
+		msg,
+	}
+
+	msgBytes, err := json.Marshal(telemetryMsg)
+
+	if err != nil {
+		m.logger.Debugf("issue decoding telemetry message: %s", err)
```
qdm12, line 139:
```suggestion
		m.logger.Debugf("issue encoding telemetry message: %s", err)
```

dot/telemetry/mailer_test.go
```diff
@@ -37,7 +38,9 @@ func TestMain(m *testing.M) {
 		Endpoint:  "ws://127.0.0.1:8001/",
 		Verbosity: 0,
 	}
-	GetInstance().AddConnections(append(testEndpoints, testEndpoint1))
+
+	logger := log.New(log.SetLevel(log.DoNotChange))
```
qdm12, line 42: Ehh I'm not sure if `DoNotChange` is appropriate here :thinking: It has a specific purpose which I don't exactly remember, it's not really a level.
Also why is it a global logger? Can't you have a logger local to each test i.e. `logger := log.New(log.SetWriter(io.Discard))` You could also have a mocked logger but that's for another day!

dot/telemetry/mailer_test.go
```diff
@@ -206,13 +177,14 @@ func TestInfiniteListener(t *testing.T) {
 func listen(w http.ResponseWriter, r *http.Request) {
 	c, err := upgrader.Upgrade(w, r, nil)
 	if err != nil {
-		log.Printf("Error %v\n", err)
+		fmt.Printf("Error %v\n", err)
```
qdm12, line 180:
```suggestion
		t.Logf("Error %v\n", err)
```

Shouldn't we fail the test here? Not sure where `listen` is used though.

dot/telemetry/mailer_test.go
```diff
@@ -206,13 +177,14 @@ func TestInfiniteListener(t *testing.T) {
...
 	}
+
 	defer c.Close()
 	for {
 		_, msg, err := c.ReadMessage()
 		if err != nil {
-			log.Printf("read err %v", err)
+			fmt.Printf("read err %v", err)
```
qdm12, line 187: same question here, this should fail the test and not log something (aka no-one reads test logs if it doesn't fail)

lib/crypto/sig_verifier.go
```diff
@@ -1,31 +1,32 @@
...
-	Sign      []byte
-	Msg       []byte
-	KeyTypeID crypto.KeyType
+var ErrSignatureVerificationFailed = errors.New("failed to verify signature")
+
+// SigVerifyFunc verifies a signature given a public key and a message
+type SigVerifyFunc func(pubkey, sig, msg []byte) (err error)
+
+// SignatureInfo ...
+type SignatureInfo struct {
```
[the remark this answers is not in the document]
EclesioMeloJunior: Removed!

dot/telemetry/mailer_test.go
EclesioMeloJunior: We need to inject a logger to BootstrapMailer

dot/telemetry/mailer.go
```diff
@@ -0,0 +1,152 @@
...
+
+			go m.shipTelemetryMessage(msg)
+		}
+	}
+}
+
+func (m *mailer) shipTelemetryMessage(msg Message) {
+	telemetryMsg := telemetryMessage{
+		msg.messageType(),
+		time.Now(),
```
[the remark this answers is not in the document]
EclesioMeloJunior: removed `telemetryMessage` as we need to keep the message fields in the same lvl of `ts` and `msg`. Using a struct with message and marshaling the output is: `{Message:{...}, ts: ..., msg: ...}` and this is not correct.

CHANGELOG.md
EclesioMeloJunior: Nop, removed!

.github/workflows/release.yml
EclesioMeloJunior: removed!

dot/telemetry/mailer_test.go
```diff
@@ -22,32 +23,74 @@ import (
...
+		require.NoError(t, err)
+
+		defer c.Close()
+
+		for {
+			_, msg, err := c.ReadMessage()
+			require.NoError(t, err)
+
+			resCh <- msg
+		}
```
qdm12: How does this for loop exit? Should we pass in a context to `setupMockListener` and exit on context cancellation?

dot/telemetry/mailer.go
```diff
@@ -0,0 +1,161 @@
...
+	"time"
+
+	"github.com/ChainSafe/gossamer/internal/log"
+	"github.com/ChainSafe/gossamer/lib/genesis"
+	"github.com/gorilla/websocket"
+)
+
+var (
+	// ErrInsufficientConnections occurs when BootstrapMailer is not able to connect to any endpoint
+	ErrInsufficientConnections = errors.New("insufficient connections")
```
qdm12, line 20: Maybe rename to `ErrNoEndpointFound`?

dot/telemetry/mailer.go
```diff
@@ -0,0 +1,161 @@
...
+	ErrTimoutMessageSending    = errors.New("timeout sending telemetry message")
+)
+var messageQueue chan Message = make(chan Message, 256)
```
qdm12, line 23:
```suggestion
var messageQueue = make(chan Message, 256)
```

dot/telemetry/mailer.go
```diff
@@ -0,0 +1,161 @@
...
+				mailer.logger.Debugf("cannot dial telemetry endpoint %s (try %d of %d): %s",
+					v.Endpoint, connAttempts+1, maxRetries, err)
+
+				timer := time.NewTimer(retryDelay)
+
+				select {
+				case <-timer.C:
+					continue
+				case <-ctx.Done():
+					mailer.logger.Debugf("bootstrap telemetry issue: %w", ctx.Err())
```
qdm12, line 65:
```suggestion
					mailer.logger.Debugf("bootstrap telemetry issue: %s", ctx.Err())
```

dot/telemetry/mailer.go
```diff
@@ -0,0 +1,161 @@
...
+	}
+
+	messageMap["ts"] = time.Now()
+	messageMap["msg"] = message.messageType()
+
+	fullRes, err := json.Marshal(messageMap)
+	if err != nil {
+		return nil, err
+	}
+	return fullRes, nil
```
qdm12, line 160: Try with


```go
type jsonMessage struct {
	Timestamp time.Time `json:"ts"`
	MessageType string `json:"msg"`
	*Message
}
```

That should put everything on the same level. Using `Message` instead of `*Message` doesn't properly inherit sorry I forgot

dot/telemetry/mailer_test.go
```diff
@@ -137,62 +141,59 @@ func TestHandler_SendMulti(t *testing.T) {
...
+			readyWait.Wait()
+
+			<-timerStartedCh
+
+			for ctx.Err() == nil {
+				bestHash := common.Hash{}
+				msg := NewBlockImportTM(&bestHash, big.NewInt(2), "NetworkInitialSync")
+				err := SendMessage(msg)
+				require.NoError(t, err)
+			}
```
qdm12, line 181: This so beautiful yet it hurts my brain hahaha

dot/telemetry/mailer_test.go
qdm12: We could pass a signal channel to the mock server which would send a signal when encountering an error to the calling test (close the channel or send the error encountered in an error channel). The test that launched the WS server goroutine would then error the test with the error received, stop the client and stop the server (with contexts or otherwise). That may be a bit complicated to implement here, I'm not 100% familiar with the test code here.

lib/utils/test_utils.go
```diff
@@ -36,7 +36,8 @@ func NewTestDir(t *testing.T) string {
 func NewTestBasePath(t *testing.T, name string) string {
 	t.Helper()
 
-	basepathTmp := path.Join(t.TempDir(), t.Name(), name)
+	tmpdir := t.TempDir()
+	basepathTmp := path.Join(tmpdir, t.Name(), name)
```
qdm12, line 40: Maybe just use `os.TempFile` instead?

lib/utils/utils_test.go
```diff
@@ -55,14 +53,10 @@ func TestBasePath(t *testing.T) {
...
-	homeDir := HomeDir()
-	basePath := BasePath(testDir)
+	const shouldContains = "/test/keystore"
 
 	keystoreDir, err := KeystoreDir(testDir)
 	require.Nil(t, err)
 
-	require.NotEqual(t, testDir, basePath)
-	require.Equal(t, strings.Contains(keystoreDir, homeDir), true)
+	require.Contains(t, keystoreDir, shouldContains)
```
qdm12, line 61: supernit, remove the suffix `s` from `shouldContains` :wink:

dot/telemetry/mailer.go
```diff
@@ -0,0 +1,161 @@
...
+)
+var messageQueue chan Message = make(chan Message, 256)
+
+type telemetryConnection struct {
+	wsconn    *websocket.Conn
+	verbosity int
+	sync.Mutex
+}
+
+// Handler struct for holding telemetry related things
```
noot, line 31: update comment

dot/telemetry/mailer.go
```diff
@@ -0,0 +1,161 @@
...
+	messageBytes, err := json.Marshal(message)
+	if err != nil {
+		return nil, err
+	}
+
+	messageMap := make(map[string]interface{})
+	err = json.Unmarshal(messageBytes, &messageMap)
+	if err != nil {
+		return nil, err
+	}
```
noot, line 151: maybe instead of marshalling and unmarshalling the message, you could have a method on `Message` called `toMap` or something that returns the message as a `map[string]interface{}`

noot (approved):
looks good!! just one comment/suggestion

lib/utils/test_utils.go
qdm12: Actually just change to

```suggestion
	tmpdir := t.TempDir()
	basepathTmp := path.Join(tmpdir, name)
```

the test name is already in `tmpdir`

dot/telemetry/mailer.go
qdm12: I think you can do it with https://github.com/ChainSafe/gossamer/pull/2106#discussion_r767745353 if I'm not mistaken

EclesioMeloJunior:
This PR is not needed anymore since PR #2113 was merged!

Issue #1855, opened 2021-10-08
lib/grandpa: ensure messages are stored and re-processed when needed


noot:
## Task summary

- currently, the network layer always re-broadcasts grandpa messages; this is to prevent an issue where a node may receive a vote message too early and discard it due to "round mismatch" errors or some other timing error
- however, this will cause unnecessary network traffic, and the node should be storing all grandpa messages it receives a re-processing them unless they are deemed too outdated (ie for some previously finalized round)
- the grandpa `messageTracker` attempts to do this but can be improved
- check that the tracker is reprocessing messages at the right time
- storing the messages when they should

kishansagathiya:
I have assumed that this is issue is really about reprocessing catch up requests and catch response messages.

From my understanding of the code, commit messages and vote message are getting added in tracker, getting processed and getting deleted as we would like them to be.

neighbour message seems like something that could be ignored if we fail to process it (since it gets sent at regular intervals)

We however don't have logic to send catchup requests and responses.

> however, this will cause unnecessary network traffic,

We send messages in loop, but the loop gets closed as the round gets finalized. Would it still cause traffic?
Apart from that we have neighbour message, which we are supposed to keep sending at certain intervals.

kishansagathiya:
closed in https://github.com/ChainSafe/gossamer/pull/2107

PR #2107, opened 2021-12-08
task(lib/grandpa): ensure messages are stored and re-processed when needed

kishansagathiya:
## Changes

- We are storing grandpa notification messages in cache to make sure that we send them only one. If the message was a consensus messages, we were sending it even if it was already sent.
- With this PR, we will not be sending consensus messages again, if they were already sent. The receiver keeps track of received messages that they could not process, so that they can be processed in future. (check usage of `addVote` for more context)

## Tests

- check that tests for `dot/network` and `lib/grandpa` passes and so does other relevant tests
- run grandpa rounds, check that we don't see votes that we already have

## Issues

- #1855

## Primary Reviewer

- @noot


dot/network/notifications.go
```diff
@@ -289,16 +289,15 @@ func (s *Service) sendData(peer peer.ID, hs Handshake, info *notificationsProtoc
...
 
-		// TODO: ensure grandpa stores *all* previously received votes and discards them
-		// only when they are for already finalised rounds; currently this causes issues
-		// because a vote might be received slightly too early, causing a round mismatch err,
-		// causing grandpa to discard the vote. (#1855)
-		_, isConsensusMsg := msg.(*ConsensusMessage)
-		if !added && !isConsensusMsg {
+		// if we could not add message in cache, message was already present in the
+		// cache and was sent.
+		if !added {
```
qdm12, line 294: Should we rename `added` to `alreadyInCache`?

lib/grandpa/grandpa.go
```diff
@@ -507,6 +507,8 @@ func (s *Service) playGrandpaRound() error {
 
 	logger.Debugf("sending pre-vote message %s...", pv)
 	roundComplete := make(chan struct{})
+	// <-roundComplete will receive the default value of channel's type when it gets
+	// closed, so we don't need to explicitly send a value.
```
qdm12, line 511:
```suggestion
	// roundComplete is a signal channel which is closed when the round completes
```

It's commonly called signal channels :wink:

lib/grandpa/message_tracker.go
```diff
@@ -23,18 +24,18 @@ type tracker struct {
 	mapLock        sync.Mutex
 	in             chan *types.Block // receive imported block from BlockState
 	stopped        chan struct{}
+	// round is used as key
+	catchUpResponseMessages map[uint64]*CatchUpResponse
```
qdm12, line 28: Should we make the map in `newTracker`?

lib/grandpa/message_tracker.go
```diff
@@ -71,6 +72,13 @@ func (t *tracker) addCommit(cm *CommitMessage) {
 	t.commitMessages[cm.Vote.Hash] = cm
 }
 
+func (t *tracker) addCatchUpResponse(cr *CatchUpResponse) {
+	t.mapLock.Lock()
+	defer t.mapLock.Unlock()
```
qdm12, line 77: Should we have a `catchUpResponseMessages` specific mutex instead of `mapLock`? To prevent eventual deadlocks and/or reduce locking contention

qdm12 (commented):
There are still a few comments / TODOs I'm not sure how to address, let's see what others have to say :wink:

dot/network/notifications.go
```diff
@@ -289,16 +289,15 @@ func (s *Service) sendData(peer peer.ID, hs Handshake, info *notificationsProtoc
...
 			return
 		}
 	}
 
+	// TODO: Add message to cache after we successfully wrote it to stream?
```
noot, line 299: the message was already written to the cache in line 286

noot: oh this is a good point actually - we should only write it to the cache if it was sent successfully. can you put a `has` check at line 286 instead, and move the `put` after the `writeToStream`?

lib/grandpa/grandpa.go
```diff
@@ -550,6 +552,7 @@ func (s *Service) sendVoteMessage(stage Subround, msg *VoteMessage, roundComplet
 	ticker := time.NewTicker(s.interval * 4)
 	defer ticker.Stop()
 
+	// TODO: Send the message just once?
```
noot, line 555: maybe we can leave this as-is, since the cache should prevent double sends to the same peer

lib/grandpa/message_handler.go
```diff
@@ -53,11 +53,22 @@ func (h *MessageHandler) handleMessage(from peer.ID, m GrandpaMessage) (network.
...
 		return h.handleCatchUpRequest(msg)
 	case *CatchUpResponse:
-		return nil, h.handleCatchUpResponse(msg)
+		err := h.handleCatchUpResponse(msg)
+		// TODO: Retry for which errors
+		if err != nil {
+			// TODO: If I can directly access tracker, why are we using `in` channel for
+			// networkVoteMessage
+			h.grandpa.tracker.addCatchUpResponse(msg)
+		}
```
noot, line 70: I think we only want to add the catch up response to the tracker for certain errors, such as if a block is not yet synced, eg the blocktree `ErrNodeNotFound`

lib/grandpa/message_tracker.go
```diff
@@ -93,6 +101,7 @@ func (t *tracker) handleBlock(b *types.Block) {
 	h := b.Header.Hash()
 	if vms, has := t.voteMessages[h]; has {
 		for _, v := range vms {
+			// handleMessage would never error for vote message
```
noot, line 104: maybe update the warn log? :p

lib/grandpa/vote_message.go
```diff
@@ -153,7 +153,9 @@ func (s *Service) validateMessage(from peer.ID, m *VoteMessage) (*Vote, error) {
 		return nil, err
 	}
 
-	// check that setIDs match
+	// TODO: Change in set ID means possible change in voters (authorities). That
+	// would make me think that I could avoid the message in this case. Is that so?
+	// It seems the vote is considered invalid if set ID do not match.
```
noot, line 158: yeah I think you can discard the message completely if there's a setID mismatch

noot (commented):
looks good, going to try this on the devnet to make sure there's no regressions, then i will approve :p

noot (approved):
ok, so it seems finality isnt realy working on development :/ so going to approve this, as I don't think it makes anything worse haha

dot/network/notifications.go
kishansagathiya: Actually, that is not the meaning of this variable. 

In `added, err := s.host.messageCache.put(peer, msg)`, if message is already in cache, put method does not add the message and `added` is false.  If we could add, added will be true.

So, it sort of like `!added` is similar to `alreadyInCache`. But, value of `added` does not depend just on whether the message was in cache or not. If `put` fails for some other reason, that would also impact the value of `added`.

so, `alreadyInCache := !added & err !=nil`

dot/network/notifications.go
kishansagathiya: addressed

lib/grandpa/message_tracker.go
kishansagathiya: I like the idea of preventing eventual deadlocks. Right now, we process all kinds of messages (vote and commit) in the same function and for reprocessing also we have a single function. I don't see possibility of a deadlock there.

For catch up messages, I am yet not sure when are we going to reprocess them. I am going to change this map to sync.Map and add a todo for it to revisit in https://github.com/ChainSafe/gossamer/issues/1531

lib/grandpa/message_tracker.go
kishansagathiya: not sure what you want here? Add this comment in log or remove the log?

lib/grandpa/grandpa.go
kishansagathiya: rephrased the comment with this suggestion

dot/network/notifications.go
```diff
@@ -288,21 +288,9 @@ func (s *Service) sendData(peer peer.ID, hs Handshake, info *notificationsProtoc
...
-
-		// TODO: ensure grandpa stores *all* previously received votes and discards them
-		// only when they are for already finalised rounds; currently this causes issues
-		// because a vote might be received slightly too early, causing a round mismatch err,
-		// causing grandpa to discard the vote. (#1855)
-		_, isConsensusMsg := msg.(*ConsensusMessage)
-		if !added && !isConsensusMsg {
-			return
-		}
+	if (s.host.messageCache != nil) && (s.host.messageCache.exists(peer, msg)) {
```
qdm12, line 291: I don't think you need parentheses here

```suggestion
	if s.host.messageCache != nil && s.host.messageCache.exists(peer, msg) {
```

Also a little trick you could have is have the `exists` method check if its receiver (`messageCache`) is nil and return false in this case. That way even if messageCache is nil, `messageCache.exists(...)` would return false and not panic. This is used in proto generated Go code for example.

dot/network/notifications.go
```diff
@@ -315,6 +303,11 @@ func (s *Service) sendData(peer peer.ID, hs Handshake, info *notificationsProtoc
 			closeOutboundStream(info, peer, stream)
 		}
 		return
+	} else if s.host.messageCache != nil {
+		if _, err := s.host.messageCache.put(peer, msg); err != nil {
+			logger.Errorf("failed to add message to cache for peer %s: %s", peer, err)
```
qdm12, line 308:
```suggestion
			logger.Errorf("failed to add message to cache for peer %s: %w", peer, err)
```

dot/network/notifications.go
```diff
@@ -315,6 +303,11 @@ func (s *Service) sendData(peer peer.ID, hs Handshake, info *notificationsProtoc
...
+			return
```
qdm12, line 309: Do we want to return here if we fail to put it in the cache? Or should we let it continue? :thinking:

lib/grandpa/grandpa.go
```diff
@@ -507,6 +507,9 @@ func (s *Service) playGrandpaRound() error {
 
 	logger.Debugf("sending pre-vote message %s...", pv)
 	roundComplete := make(chan struct{})
+	// roundComplete is a signal channel which is closed when the round completes
+	// (will receive the default value of channel's type), so we don't need to
+	// explicitly send a value.
```
qdm12, line 512: nit I think that would be enough of a comment

```suggestion
	// roundComplete is a signal channel which is closed when the round completes
```

lib/grandpa/grandpa.go
```diff
@@ -550,6 +553,8 @@ func (s *Service) sendVoteMessage(stage Subround, msg *VoteMessage, roundComplet
 	ticker := time.NewTicker(s.interval * 4)
 	defer ticker.Stop()
 
+	// Eventhough, this looks like we are sending messages multiple times,
```
qdm12, line 556: nit not sure what you mean by `Eventhough`?

lib/grandpa/grandpa.go
```diff
@@ -550,6 +553,8 @@ func (s *Service) sendVoteMessage(stage Subround, msg *VoteMessage, roundComplet
...
+	// caching would make sure that they are being sent only once.
```
qdm12, line 557: Is it caching then? Maybe rename it to something like recorder? :thinking:

lib/grandpa/message_handler.go
```diff
@@ -53,11 +53,20 @@ func (h *MessageHandler) handleMessage(from peer.ID, m GrandpaMessage) (network.
...
 		return nil, h.handleCommitMessage(msg)
 	case *NeighbourMessage:
+		// we can afford to not retry handling neighbour message, if it errors.
 		return nil, h.handleNeighbourMessage(msg)
 	case *CatchUpRequest:
 		return h.handleCatchUpRequest(msg)
 	case *CatchUpResponse:
-		return nil, h.handleCatchUpResponse(msg)
+		err := h.handleCatchUpResponse(msg)
+		if err == blocktree.ErrNodeNotFound {
```
qdm12, line 62:
```suggestion
		if errors.Is(err, blocktree.ErrNodeNotFound) {
```

lib/grandpa/message_handler.go
```diff
@@ -53,11 +53,20 @@ func (h *MessageHandler) handleMessage(from peer.ID, m GrandpaMessage) (network.
...
+			// TODO: we are adding these messages to reprocess them again, but we
+			// haven't added code to reprocess them. Do that.
+			// Also, revisit if we need to add these message in synchronous manner
+			// or not. If not, change catchUpResponseMessages to a normal map.  #1531
```
qdm12, line 66: Why not have it as a standard `map` with a mutex instead? It adds some type safety compared to the `sync.Map`

lib/grandpa/vote_message.go
```diff
@@ -22,8 +22,8 @@ type networkVoteMessage struct {
 	msg  *VoteMessage
 }
 
-// receiveMessages receives messages from the in channel until the specified condition is met
-func (s *Service) receiveMessages(ctx context.Context) {
+// receiveVoteMessages receives messages from the in channel until the specified condition is met
```
qdm12, line 25: nit a bit OOS, but maybe add what that `specified condition` is? :thinking:

lib/grandpa/vote_message.go
```diff
@@ -153,7 +153,7 @@ func (s *Service) validateMessage(from peer.ID, m *VoteMessage) (*Vote, error) {
 		return nil, err
 	}
 
-	// check that setIDs match
+	// vote is considered invalid if set ID do not match.
```
qdm12, line 156: nit the comment is not really needed I think

```suggestion
```

lib/grandpa/vote_message.go
```diff
@@ -229,6 +229,7 @@ func (s *Service) validateMessage(from peer.ID, m *VoteMessage) (*Vote, error) {
 
 	equivocated := s.checkForEquivocation(voter, just, m.Message.Stage)
 	if equivocated {
+		// A vote is considered invalid if it is equivocatory
```
qdm12, line 232: nit not needed

```suggestion
```

lib/grandpa/grandpa.go
kishansagathiya:
```suggestion
	// Though, this looks like we are sending messages multiple times,
```

qdm12: call me Grammarly :smile: , but starting a sentence with `Though` doesn't make much sense, use `Although` instead :wink: (and also remove that comma)

lib/grandpa/message_handler.go
kishansagathiya: That is an option. I think I have used sync.Map, since it is tidier than using a mutex for every map. This logic could have multiple maps for each kind of message. But it is best to discuss this during the catchUpResponse related issue.

Also, you can always check the type of an object received sync.Map. I don't see if any of these approaches are clearly better than other, they are just different.

qdm12: Just create a custom type with a map and a mutex as fields. You can add typed get and put methods?

I don't think it's tidier to have no type safety.

There is a point in using `sync.Map` really only if we have a lot of reads on it and we can see it as a bottleneck (like `sync.Pool` really).

EclesioMeloJunior:
@noot @kishansagathiya should we add in the `MessageHandler.handleCatchUpResponse` a code (if the handle was successful) to check if the given message was stored in the tracker and then remove it?

EclesioMeloJunior (approved):
LGTM

dot/network/notifications.go
kishansagathiya: Unfortunately, I don't have a clear answer to this.

The difference between returning and not returning is execution of ReportPeer. 
- I you don't return ReportPeer will get executed multiple times (increasing the reputation every time) until we are able to put the message in cache. 

- If we do return, ReportPeer does not get executed, until we are able to put the message in cache, increasing the reputation only once.

None of these options are good, but the second one seems less bad. Also, we were returning previously as well, so this behavior has not changed.

However, if we fail to add msg to cache, in both cases, we have a bigger problem, that we keeping sending the same message, and our reputation decreases.

lib/grandpa/grandpa.go
kishansagathiya: > Is it caching then? Maybe rename it to something like recorder?

You not wrong, but we have called this caching everywhere (outside this package as well, where we need to store messages). A lot of variable renaming.

lib/grandpa/vote_message.go
kishansagathiya: addressed

lib/grandpa/vote_message.go
kishansagathiya: addressed

lib/grandpa/message_handler.go
kishansagathiya: addressed, using mutex now

dot/network/notifications.go
kishansagathiya: I would rather have the nil check here than in exists method. I am assuming that messageCache is not supposed to be nil.

PR #2108, opened 2021-12-10
chore(lib/trie): refactor `lib/trie/database.go`

qdm12:
## Changes

This is a simple 'linear' refactor in order to see diffs in a clearer way.

- Explicit variable names
- Less indentation (avoid type switches)
- More return early
- Add named returns
- Clarify comments
- Wrap errors

## Tests

```
go test github.com/ChainSafe/gossamer/lib/trie/... github.com/ChainSafe/gossamer/internal/trie/...
```

## Issues

- #2085

## Primary Reviewer

### DIFF
--- a/lib/trie/database.go
+++ b/lib/trie/database.go
@@ -191,7 +191,7 @@ func (t *Trie) load(db chaindb.Database, n Node) error {
 
 		err = t.load(db, decodedNode)
 		if err != nil {
-			return fmt.Errorf("cannot load child with hash 0x%x: %w", hash, err)
+			return fmt.Errorf("cannot load child at index %d with hash 0x%x: %w", i, hash, err)
 		}
 	}
 
@@ -275,9 +275,6 @@ func GetFromDB(db chaindb.Database, rootHash common.Hash, key []byte) (
 // slice will modify the value of the node in the trie.
 func getFromDB(db chaindb.Database, n Node, key []byte) (
 	value []byte, err error) {
-	// if parent == nil {
-	// 	return nil, nil
-	// }
 	leaf, ok := n.(*node.Leaf)
 	if ok {
 		if bytes.Equal(leaf.Key, key) {
--- a/lib/trie/database.go
+++ b/lib/trie/database.go
@@ -15,8 +15,10 @@ import (
 	"github.com/ChainSafe/chaindb"
 )
 
-// ErrEmptyProof indicates the proof slice is empty
-var ErrEmptyProof = errors.New("proof slice empty")
+var (
+	ErrEmptyProof = errors.New("proof slice empty")
+	ErrDecodeNode = errors.New("cannot decode node")
+)
 
 // Store stores each trie node in the database,
 // where the key is the hash of the encoded node
@@ -67,10 +69,6 @@ func (t *Trie) store(db chaindb.Batch, n Node) error {
 
 	return nil
 }
-
-var (
-	ErrDecodeNode = errors.New("cannot decode node")
-)
 
 // loadFromProof create a partial trie based on the proof slice, as it only contains nodes that are in the proof afaik.
 func (t *Trie) loadFromProof(rawProof [][]byte, rootHash []byte) error {
--- a/lib/trie/database.go
+++ b/lib/trie/database.go
@@ -33,23 +33,23 @@ func (t *Trie) Store(db chaindb.Database) error {
 	return batch.Flush()
 }
 
-func (t *Trie) store(db chaindb.Batch, curr Node) error {
-	if curr == nil {
+func (t *Trie) store(db chaindb.Batch, n Node) error {
+	if n == nil {
 		return nil
 	}
 
-	enc, hash, err := curr.EncodeAndHash()
+	encoding, hash, err := n.EncodeAndHash()
 	if err != nil {
 		return err
 	}
 
-	err = db.Put(hash, enc)
+	err = db.Put(hash, encoding)
 	if err != nil {
 		return err
 	}
 
-	if c, ok := curr.(*node.Branch); ok {
-		for _, child := range c.Children {
+	if branch, ok := n.(*node.Branch); ok {
+		for _, child := range branch.Children {
 			if child == nil {
 				continue
 			}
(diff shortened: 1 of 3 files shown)


lib/trie/database.go
```diff
@@ -61,241 +61,270 @@ func (t *Trie) store(db chaindb.Batch, curr Node) error {
...
-			if err != nil {
-				return err
-			}
+		decodedNode.SetDirty(false)
+		decodedNode.SetEncodingAndHash(encodedNode, hash)
+		branch.Children[i] = decodedNode
+
+		err = t.load(db, decodedNode)
+		if err != nil {
+			return fmt.Errorf("cannot load child with hash 0x%x: %w", hash, err)
```
EclesioMeloJunior, line 194: Would be nice to log the index here
```suggestion
			return fmt.Errorf("cannot load child at index %d with hash 0x%x: %w", i, hash, err)
```

lib/trie/database.go
```diff
@@ -61,241 +61,270 @@ func (t *Trie) store(db chaindb.Batch, curr Node) error {
...
-			return nil, nil
+// getFromDB recursively searches through the trie and database
+// for the value corresponding to a key.
+// Note it does not copy the value so modifying the value bytes
+// slice will modify the value of the node in the trie.
+func getFromDB(db chaindb.Database, n Node, key []byte) (
+	value []byte, err error) {
+	// if parent == nil {
+	// 	return nil, nil
+	// }
```
EclesioMeloJunior, line 280:
```suggestion
```

lib/trie/database.go
```diff
@@ -310,94 +339,117 @@ func (t *Trie) WriteDirty(db chaindb.Database) error {
...
+// We need to compute the hash values of each newly inserted node.
+func (t *Trie) GetInsertedNodeHashes() (hashes []common.Hash, err error) {
 	return t.getInsertedNodeHashes(t.root)
 }
 
-func (t *Trie) getInsertedNodeHashes(curr Node) ([]common.Hash, error) {
-	var nodeHashes []common.Hash
-	if curr == nil || !curr.IsDirty() {
+func (t *Trie) getInsertedNodeHashes(n Node) (hashes []common.Hash, err error) {
+	// TODO pass map of hashes or slice as argument to avoid copying
```
EclesioMeloJunior, line 403: @qdm12 could I create an issue to address this TODO?

EclesioMeloJunior (changes requested):

lib/trie/database.go
qdm12: Opened #2115 for it (pointing to this PR), it was quicker than writing an issue imo :smile_cat:

lib/trie/database.go
qdm12: Yeah left it to try it out, I ran the node in production and never encountered a panic. I'll assume this never happen :wink:

noot (approved):

lib/trie/database.go
```diff
@@ -61,241 +61,267 @@ func (t *Trie) store(db chaindb.Batch, curr Node) error {
...
 
-		value, err = getFromDB(db, child, key[length+1:])
-		if err != nil {
-			return nil, err
-		}
-	case *node.Leaf:
-		if bytes.Equal(p.Key, key) {
-			return p.Value, nil
-		}
-	case nil:
```
kishansagathiya, line 294: I am guessing this would have never happened, because go does not permit use of untyped nil `use of untyped nil`. parent has to be one of node or leaf.

kishansagathiya (approved):
:tada: 
I like that we are checking `!ok` instead of `ok`
Looks good, only small comments!

lib/trie/database.go
```diff
@@ -61,241 +61,267 @@ func (t *Trie) store(db chaindb.Batch, curr Node) error {
...
 		return nil
 	}
 
-	enc, err := db.Get(root[:])
+	rootHashBytes := rootHash[:]
+
+	encodedNode, err := db.Get(rootHashBytes)
 	if err != nil {
-		return fmt.Errorf("failed to find root key=%s: %w", root, err)
+		return fmt.Errorf("failed to find root key %s: %w", rootHash, err)
```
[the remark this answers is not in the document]
qdm12: rootHash is a `common.Hash` and its `String() string` method returns hexadecimal already :wink: 
Just re-using existing code hehe, although I agree with you it's not that obvious!

Most other places in this file uses raw `[]byte` where we do need `0x%x` however.

lib/trie/database.go
```diff
@@ -61,241 +61,267 @@ func (t *Trie) store(db chaindb.Batch, curr Node) error {
...
 		return nil, nil
 	}
 
 	k := codec.KeyLEToNibbles(key)
 
-	enc, err := db.Get(root[:])
+	encodedRootNode, err := db.Get(rootHash[:])
 	if err != nil {
-		return nil, fmt.Errorf("failed to find root key=%s: %w", root, err)
+		return nil, fmt.Errorf("cannot find root hash key 0x%x: %w", rootHash, err)
```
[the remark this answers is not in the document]
qdm12: Totally, changed from `0x%x` to `%s`

lib/trie/database.go
```diff
@@ -61,241 +61,267 @@ func (t *Trie) store(db chaindb.Batch, curr Node) error {
...
-		if err != nil {
-			return nil, err
-		}
-	case *node.Leaf:
-		if bytes.Equal(p.Key, key) {
-			return p.Value, nil
-		}
-	case nil:
+	// childIndex is the nibble after the common prefix length in the key being searched.
+	childIndex := key[commonPrefixLength]
```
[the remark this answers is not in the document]
qdm12: Readability :wink: (aka smaller line length, more descriptive names for each subpart)

lib/trie/database.go
```diff
@@ -61,241 +61,267 @@ func (t *Trie) store(db chaindb.Batch, curr Node) error {
...
+	childHash := childWithHashOnly.GetHash()
+	encodedChild, err := db.Get(childHash)
+	if err != nil {
+		return nil, fmt.Errorf(
+			"cannot find child with hash 0x%x in database: %w",
+			childHash, err)
 	}
-	return value, nil
+
+	reader := bytes.NewReader(encodedChild)
```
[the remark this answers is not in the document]
qdm12: Same as above, this is for readability. I used to nest calls to save on the number of lines but I now prefer to split in on more lines since I find it way more nice to read.

lib/trie/database.go
qdm12: It can happen if the parent is `nil`, for example with an empty trie the root can be `nil` (untyped).

lib/trie/database.go
```diff
@@ -61,241 +61,267 @@ func (t *Trie) store(db chaindb.Batch, curr Node) error {
...
-		for _, child := range c.Children {
-			if child == nil {
-				continue
-			}
-
-			hash := child.GetHash()
-			keys[common.BytesToHash(hash)] = struct{}{}
+// GetNodeHashes writes hashes of each children of the node given
+// as keys to the map hashesSet.
+func (t *Trie) GetNodeHashes(n Node, hashesSet map[common.Hash]struct{}) {
```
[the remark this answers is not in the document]
qdm12: Changed to `PopulatedHashesSet`

lib/trie/database.go
```diff
@@ -61,241 +61,267 @@ func (t *Trie) store(db chaindb.Batch, curr Node) error {
...
 	}
 
 	return nil
 }
 
-// LoadFromProof create a partial trie based on the proof slice, as it only contains nodes that are in the proof afaik.
-func (t *Trie) LoadFromProof(proof [][]byte, root []byte) error {
-	if len(proof) == 0 {
+var (
+	ErrDecodeNode = errors.New("cannot decode node")
```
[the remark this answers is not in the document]
qdm12: Done :+1:

qdm12:
@kishansagathiya 

> I like that we are checking !ok instead of ok

Where do you mean this? :thinking:

kishansagathiya:
@qdm12 

> like that we are checking !ok instead of ok

Sometimes, we check for the expected condition (true) and run when true and don't do anything when false.

I think you would have checked for the unexpected condition first and run some code on that unexpected condition.

qdm12:
@kishansagathiya I tend to have the smaller code part inside the if block checking `ok` to reduce code indentation and improve  readability. Maybe do you have an example where this could be improved in this PR?

lib/trie/database.go
qdm12: I'll merge #2115 in development once this one is merged.

EclesioMeloJunior (approved):

PR #2113, opened 2021-12-13
chore(dot): transform `telemetry` into an injectable interface

### LINKED ISSUE #2111: test(dot/telemetry): improve telemetry tests and add `t.Parallel()` and remove race conditions
## Issue summary

- We need to create an instance of the test mocked WebSocket server for each test, currently we have just one shared websocket server and a global channel to all tests.
- Initialize a `logger` for each test instead of maintain a global test logger

- More context on PR #2106

EclesioMeloJunior:
## Changes

- Remove mockery as mock generator and use mockgen
- Adjust all tests with new `EXPECTs`
- Create the `telemetry.Telemetry` interface
- Update all services that uses telemetry to receive the `telemetry.Telemetry` interface as argument
- Split the `dot.createStateService` in `dot.dot.createStateService` and `dot.startStateService`. This is necessary because the `state.Start()` function needs the telemetry and the telemetry needs the state initialization of the database, so I decoupled the database initialization from the state service start function. 

## Tests

```
go test github.com/ChainSafe/gossamer/dot/network -short
go test github.com/ChainSafe/gossamer/dot/telemetry -race
```

## Issues

- Related #2105
- Closes #2111

## Primary Reviewer

-


dot/network/connmgr_test.go
```diff
@@ -216,6 +216,7 @@ func TestSetReservedPeer(t *testing.T) {
 
 	nodes := make([]*Service, 3)
 	for i := range nodes {
+
```
qdm12, line 219:
```suggestion
```

dot/network/discovery_test.go
```diff
@@ -22,6 +22,7 @@ func newTestDiscovery(t *testing.T, num int) []*discovery {
 
 	var discs []*discovery
 	for i := 0; i < num; i++ {
+
```
qdm12, line 25:
```suggestion
```

dot/network/service_test.go
```diff
@@ -72,17 +103,39 @@ func createTestService(t *testing.T, cfg *Config) (srvc *Service) {
...
-		mocktxhandler := &MockTransactionHandler{}
-		mocktxhandler.On("HandleTransactionMessage",
-			mock.AnythingOfType("peer.ID"),
-			mock.AnythingOfType("*network.TransactionMessage")).
-			Return(true, nil)
-		mocktxhandler.On("TransactionsCount").Return(0)
-		cfg.TransactionHandler = mocktxhandler
+		th := NewMockTransactionHandler(ctrl)
+		th.EXPECT().
+			HandleTransactionMessage(gomock.Any(), gomock.Any()).
```
qdm12, line 134: You can use `gomock.AssignableToTypeOf(peer.ID(""))` for example to stay a bit closer to how it was before.

qdm12 (commented):
Awesome! I wanted to do this during Christmas lol, but I guess you were :electron: faster! :+1: 

Will finish reviewing today!

dot/network/service_test.go
```diff
@@ -12,10 +12,11 @@ import (
 	"testing"
 	"time"
 
-	mock "github.com/stretchr/testify/mock"
+	gomock "github.com/golang/mock/gomock"
```
qdm12, line 15:
```suggestion
	"github.com/golang/mock/gomock"
```

dot/rpc/modules/system_integration_test.go
```diff
@@ -61,11 +62,12 @@ var (
...
 
+	ctrl := gomock.NewController(t)
 	cfg := &network.Config{
-		BlockState:         network.NewMockBlockState(nil),
+		BlockState:         network.NewMockBlockState(ctrl),
 		BasePath:           testDir,
-		Syncer:             network.NewMockSyncer(),
-		TransactionHandler: network.NewMockTransactionHandler(),
+		Syncer:             network.NewMockSyncer(ctrl),
+		TransactionHandler: network.NewMockTransactionHandler(ctrl),
```
qdm12, line 70: Do you need to initialize all this actually if they don't expect anything? Can't these be left to `nil`?

dot/network/utils_test.go
```diff
@@ -20,31 +20,13 @@ func availablePort(t *testing.T) uint16 {
...
-	return <-p
-}
-
-func (p portsQueue) put(port int) {
-	p <- port
-}
+	port := uint16(l.Addr().(*net.TCPAddr).Port)
 
-var availablePorts portsQueue
+	err = l.Close()
```
qdm12, line 26: A bit out of scope perhaps but this won't work to reserve a certain port for a certain test since it releases it back to the OS.

dot/network/transaction_test.go
```diff
@@ -31,29 +31,28 @@ func TestDecodeTransactionHandshake(t *testing.T) {
...
-		Return(true, nil)
-	mockhandler.On("TransactionsCount").Return(0)
 
 	config := &Config{
 		BasePath:           basePath,
 		Port:               availablePort(t),
 		NoBootstrap:        true,
 		NoMDNS:             true,
-		TransactionHandler: mockhandler,
+		TransactionHandler: th,
```
qdm12, line 53: nit maybe rename `th` to `transactionHandler`? :thinking:

dot/network/transaction_test.go
```diff
@@ -6,12 +6,12 @@ package network
 import (
 	"testing"
 
+	gomock "github.com/golang/mock/gomock"
```
qdm12, line 9:
```suggestion
	"github.com/golang/mock/gomock"
```

dot/network/state.go
```diff
@@ -42,7 +42,7 @@ type Syncer interface {
 	CreateBlockResponse(*BlockRequestMessage) (*BlockResponseMessage, error)
 }
 
-//go:generate mockery --name TransactionHandler --structname MockTransactionHandler --case underscore --inpackage
+//go:generate mockgen -destination=mock_transaction_handler.go -package $GOPACKAGE . TransactionHandler
```
qdm12, line 45: change to 

```suggestion
//go:generate mockgen -destination=mock_transaction_handler_test.go -package $GOPACKAGE . TransactionHandler
```

Unless it's needed by other packages. Actually ideally I think generating package local mocks in `_test.go` files is best.

If you need i.e. to mock `io.Writer` for the tests in `some/package`, then have 

```go
//go:generate mockgen -destination=mock_writer_test.go -package $GOPACKAGE io Writer
```

Where you need it (in your test file for example).

That ensures the go API is not polluted by mocks and also makes them local where they're needed. It does introduce some generated code duplication though but I think it's worth it.

dot/network/service_test.go
```diff
@@ -54,9 +55,39 @@ func createServiceHelper(t *testing.T, num int) []*Service {
...
 
+func createTestBlockResponseMessage(t *testing.T) *BlockResponseMessage {
+	t.Helper()
+
+	msg := &BlockResponseMessage{
+		BlockData: []*types.BlockData{},
+	}
+
+	const blockRequestSize uint32 = 128
+	for i := 0; i < int(blockRequestSize); i++ {
```
qdm12, line 66:
```suggestion
	const blockRequestSize = 128
	for i := 0; i < blockRequestSize; i++ {
```

dot/network/service_test.go
```diff
@@ -54,9 +55,39 @@ func createServiceHelper(t *testing.T, num int) []*Service {
...
+
+		body := types.NewBody([]types.Extrinsic{[]byte{4, 4, 2}})
+
+		msg.BlockData = append(msg.BlockData, &types.BlockData{
+			Hash:          testHeader.Hash(),
+			Header:        testHeader,
+			Body:          body,
+			MessageQueue:  nil,
+			Receipt:       nil,
+			Justification: nil,
```
qdm12, line 80:
```suggestion
```

dot/network/utils_test.go
EclesioMeloJunior: yeah, I will fix it here -> #2105

dot/network/state.go
EclesioMeloJunior: These mocks need to be used by other packages, although agreed with you I think I will leave as it is to avoid too many changes in the codebase right now, but we can create an issue under Tests epic to handle the mocks

noot (approved):
i guess we are officially moving to `gomock`?

qdm12:
> i guess we are officially moving to `gomock`?

We should although this PR is just for the telemetry package I think right?

I might do a PR during Christmas time to move to gomock/mockgen through the entire codebase.

EclesioMeloJunior:
@qdm12 is for all the network package

dot/network/state.go
qdm12: Totally, thanks!

dot/network/service_test.go
```diff
@@ -54,9 +56,36 @@ func createServiceHelper(t *testing.T, num int) []*Service {
 	return srvcs
 }
 
+func createTestBlockResponseMessage(t *testing.T) *BlockResponseMessage {
+	t.Helper()
+
+	msg := &BlockResponseMessage{
+		BlockData: []*types.BlockData{},
```
qdm12, line 63: nit not that it matters, but you could pre-allocate with `make([]*types.BlockData, blockRequestSize)` and assign directly using the index instead of using `append`. But it doesn't matter though :wink:

dot/network/service_test.go
```diff
@@ -72,17 +101,42 @@ func createTestService(t *testing.T, cfg *Config) (srvc *Service) {
...
+		blockstate := NewMockBlockState(ctrl)
+
+		blockstate.EXPECT().BestBlockHeader().Return(header, nil).AnyTimes()
+		blockstate.EXPECT().GetHighestFinalisedHeader().Return(header, nil).AnyTimes()
+		blockstate.EXPECT().GenesisHash().Return(common.NewHash([]byte{})).AnyTimes()
+		blockstate.EXPECT().BestBlockNumber().Return(big.NewInt(1), nil).AnyTimes()
+
+		blockstate.EXPECT().HasBlockBody(
+			gomock.AssignableToTypeOf(common.Hash([32]byte{}))).Return(false, nil).AnyTimes()
+		blockstate.EXPECT().GetHashByNumber(gomock.Any()).Return(common.Hash{}, nil).AnyTimes()
```
qdm12, line 125: Try to expect a concrete type if possible. Although maybe this is used everywhere and it's too complicated to do / out of scope, I'm not sure. Same goes for `AnyTimes()`, it would be better to expect a precise number of times.

dot/network/transaction_test.go
```diff
@@ -31,29 +31,28 @@ func TestDecodeTransactionHandshake(t *testing.T) {
...
 
+	ctrl := gomock.NewController(t)
+
+	expectedMsgArg := &TransactionMessage{
+		Extrinsics: []types.Extrinsic{{1, 1}, {2, 2}},
+	}
+
+	transactionHandler := NewMockTransactionHandler(ctrl)
+	transactionHandler.EXPECT().
+		HandleTransactionMessage(gomock.Any(), expectedMsgArg).
```
qdm12, line 42: expect a concrete arg here; I don't think it would be complicated in this case.

dot/network/transaction_test.go
```diff
@@ -31,29 +31,28 @@ func TestDecodeTransactionHandshake(t *testing.T) {
...
+		Return(true, nil).AnyTimes()
+	transactionHandler.EXPECT().TransactionsCount().Return(0).AnyTimes()
```
qdm12, line 44: expect a precise number of times here.

dot/node.go
```diff
@@ -75,13 +75,19 @@ func InitNode(cfg *Config) error {
 		return fmt.Errorf("failed to create genesis block from trie: %w", err)
 	}
 
+	telemetryMailer, err := setupTelemetry(cfg, nil)
+	if err != nil {
+		return err
```
qdm12, line 80:
```suggestion
		return fmt.Errorf("cannot setup telemetry mailer: %w", err)
```

dot/node.go
```diff
@@ -204,14 +210,54 @@ func NewNode(cfg *Config, ks *keystore.GlobalKeystore) (*Node, error) {
 		return nil, fmt.Errorf("failed to create state service: %s", err)
 	}
 
+	gd, err := stateSrvc.Base.LoadGenesisData()
+	if err != nil {
+		return nil, err
+	}
+
+	telemetryMailer, err := setupTelemetry(cfg, gd)
```
qdm12, line 218: Do you need to setupTelemetry for both `InitNode` and `NewNode`? :thinking:

dot/node.go
```diff
@@ -204,14 +210,54 @@ func NewNode(cfg *Config, ks *keystore.GlobalKeystore) (*Node, error) {
...
+	if err != nil {
+		return nil, err
```
qdm12, line 220:
```suggestion
		return nil, fmt.Errorf("cannot setup telemetry mailer: %w", err)
```

dot/node.go
```diff
@@ -204,14 +210,54 @@ func NewNode(cfg *Config, ks *keystore.GlobalKeystore) (*Node, error) {
...
+	}
+
+	stateSrvc.SetTelemetry(telemetryMailer)
+	err = startStateService(cfg, stateSrvc)
+	if err != nil {
+		return nil, err
```
qdm12, line 226:
```suggestion
		return nil, fmt.Errorf("cannot start state service: %w", err)
```

dot/node.go
```diff
@@ -204,14 +210,54 @@ func NewNode(cfg *Config, ks *keystore.GlobalKeystore) (*Node, error) {
...
+			&genesisHash,
+			sysSrvc.SystemName(),
+			cfg.Global.Name,
+			networkSrvc.NetworkState().PeerID,
+			strconv.FormatInt(time.Now().UnixNano(), 10),
+			sysSrvc.SystemVersion())
+
+		err = telemetryMailer.SendMessage(connectedMsg)
+		if err != nil {
+			logger.Debugf("problem sending system.connected telemetry message: %s", err)
```
qdm12, line 259: nit

```suggestion
			logger.Debugf("failed sending system.connected telemetry message: %s", err)
```

dot/node.go
```diff
@@ -313,47 +353,23 @@ func NewNode(cfg *Config, ks *keystore.GlobalKeystore) (*Node, error) {
...
 	}
 
-	gd, err := stateSrvc.Base.LoadGenesisData()
-	if err != nil {
-		return nil, err
-	}
+	return node, nil
+}
 
+func setupTelemetry(cfg *Config, genesisData *genesis.Data) (*telemetry.Mailer, error) {
```
qdm12, line 359: Please use named returns

```suggestion
func setupTelemetry(cfg *Config, genesisData *genesis.Data) (mailer *telemetry.Mailer, err error) {
```

dot/node.go
```diff
@@ -313,47 +353,23 @@ func NewNode(cfg *Config, ks *keystore.GlobalKeystore) (*Node, error) {
...
-
-		err = telemetry.SendMessage(connectedMsg)
-		if err != nil {
-			logger.Debugf("problem sending system.connected telemetry message: %s", err)
-		}
-	}
-
-	return node, nil
+	telemetryLogger := log.NewFromGlobal(log.AddContext("pkg", "telemetry"))
+	return telemetry.BootstrapMailer(context.Background(),
```
qdm12, line 371: Should you pass a context here instead of `context.Background()`?

dot/rpc/http_test.go
```diff
@@ -351,3 +365,128 @@ func externalIP() (string, error) {
...
+	cfg := &core.Config{}
+
+	cfg.DigestHandler = new(coremocks.DigestHandler)
+	cfg.DigestHandler.(*coremocks.DigestHandler).On("HandleDigests", mock.AnythingOfType("*types.Header"))
+
+	if cfg.Keystore == nil {
+		cfg.Keystore = keystore.NewGlobalKeystore()
+		kp, err := sr25519.GenerateKeypair()
+		if err != nil {
+			t.Fatal(err)
```
qdm12, line 382:
```suggestion
			require.NoError(t, err)
```

dot/rpc/http_test.go
```diff
@@ -351,3 +365,128 @@ func externalIP() (string, error) {
...
+// NewTestService creates a new test core service
+func newCoreServiceTest(t *testing.T) *core.Service {
+	t.Helper()
+
+	cfg := &core.Config{}
+
+	cfg.DigestHandler = new(coremocks.DigestHandler)
+	cfg.DigestHandler.(*coremocks.DigestHandler).On("HandleDigests", mock.AnythingOfType("*types.Header"))
+
+	if cfg.Keystore == nil {
```
qdm12, line 378: Isn't this always true?

dot/rpc/http_test.go
```diff
@@ -351,3 +365,128 @@ func externalIP() (string, error) {
...
+			Path:      testDatadirPath,
+			LogLevel:  log.Info,
+			Telemetry: telemetryMock,
+		}
+
+		stateSrvc = state.NewService(config)
+		stateSrvc.UseMemDB()
+
+		err := stateSrvc.Initialise(gen, genHeader, genTrie)
+		require.Nil(t, err)
```
qdm12, line 413:
```suggestion
		require.NoError(t, err)
```

dot/rpc/http_test.go
```diff
@@ -351,3 +365,128 @@ func externalIP() (string, error) {
...
+
+		err = stateSrvc.Start()
+		require.Nil(t, err)
```
qdm12, line 416: nit

```suggestion
		require.NoError(t, err)
```

dot/rpc/http_test.go
```diff
@@ -351,3 +365,128 @@ func externalIP() (string, error) {
...
+
+	gen, genTrie, genHeader := genesis.NewTestGenesisWithTrieAndHeader(t)
+
+	if cfg.BlockState == nil || cfg.StorageState == nil ||
+		cfg.TransactionState == nil || cfg.EpochState == nil ||
+		cfg.CodeSubstitutedState == nil {
+
+		ctrl := gomock.NewController(t)
+		telemetryMock := telemetry.NewMockTelemetry(ctrl)
+		telemetryMock.EXPECT().SendMessage(gomock.Any())
```
qdm12, line 401: expect concrete arg is possible

dot/rpc/http_test.go
```diff
@@ -351,3 +365,128 @@ func externalIP() (string, error) {
...
+
+		config := state.Config{
+			Path:      testDatadirPath,
+			LogLevel:  log.Info,
```
qdm12, line 405: Put this as a higher level to avoid spamming test with logs

dot/rpc/http_test.go
```diff
@@ -351,3 +365,128 @@ func externalIP() (string, error) {
...
+		if err != nil {
+			t.Fatal(err)
+		}
+		err = cfg.Keystore.Acco.Insert(kp)
+		require.NoError(t, err)
+	}
+
+	cfg.LogLvl = 3
+
+	var stateSrvc *state.Service
```
qdm12, line 390: if the `if` block below doesn't execute, the code will panic on a nil pointer

dot/rpc/modules/author_integration_test.go
```diff
@@ -50,7 +52,14 @@ func TestMain(m *testing.M) {
...
 
 func TestAuthorModule_Pending(t *testing.T) {
-	txQueue := state.NewTransactionState()
+	ctrl := gomock.NewController(t)
+	telemetryMock := telemetry.NewMockTelemetry(ctrl)
+	telemetryMock.
+		EXPECT().
+		SendMessage(gomock.Any()).
+		AnyTimes()
+
```
qdm12, line 61: expect concrete args and times (or expect multiple different things)

dot/network/config.go
```diff
@@ -101,6 +102,8 @@ type Config struct {
 
 	// SlotDuration is the slot duration to produce a block
 	SlotDuration time.Duration
+
+	Telemetry telemetry.Telemetry
```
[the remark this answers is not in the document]
EclesioMeloJunior: renamed to `telemetry.Client`

dot/telemetry/mailer.go
```diff
@@ -41,7 +40,7 @@ type Mailer struct {
 func newMailer(enabled bool, logger log.LeveledLogger) *Mailer {
 	return &Mailer{
 		enabled:      enabled,
-		messageQueue: messageQueue,
+		messageQueue: make(chan Message),
```
qdm12, line 43: :+1: Unbuffered channels

test.log
```diff
@@ -0,0 +1,10 @@
+=== RUN   TestHandler_SendMulti
```
qdm12, line 1: remove

dot/telemetry/mailer_test.go
```diff
@@ -212,7 +210,7 @@ func TestListenerConcurrency(t *testing.T) {
 	}
 }
 
-func pipeRequestToChannel(t *testing.T, wsUpgrader websocket.Upgrader, ch chan<- []byte) http.HandlerFunc {
+func pipeRequestToChannel(t *testing.T, wsUpgrader *websocket.Upgrader, ch chan<- []byte) http.HandlerFunc {
```
qdm12, line 213: Why changing it to a pointer etc here?

dot/build_spec.go
```diff
@@ -94,29 +96,37 @@ func BuildFromDB(path string) (*BuildSpec, error) {
...
-		LogLevel: log.Info,
+		Path:      path,
+		LogLevel:  log.Info,
+		Telemetry: disabledTelemetry,
 	}
+
 	stateSrvc := state.NewService(config)
+	err := stateSrvc.SetupBase()
+	if err != nil {
+		return nil, err
```
qdm12, line 113: wrap the error

```suggestion
		return nil, fmt.Errorf("failed to setup base: %w", err)
```

dot/network/config.go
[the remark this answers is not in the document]
qdm12: Awesome!

dot/node.go
EclesioMeloJunior: Yes, `InitNode` calls the `state.Initialise()` function that calls `NewBlockStateFromGenesis(db, header, telemetry)` and this last function uses the `telemetry.SendMessage(NewNotifyFinalizedTM)`

dot/network/service_test.go
```diff
@@ -60,11 +60,11 @@ func createServiceHelper(t *testing.T, num int) []*Service {
 func createTestBlockResponseMessage(t *testing.T) *BlockResponseMessage {
 	t.Helper()
 
+	const blockRequestSize = 128
 	msg := &BlockResponseMessage{
-		BlockData: []*types.BlockData{},
+		BlockData: make([]*types.BlockData, blockRequestSize),
```
qdm12, line 65: You need to change the code below to not `append` to the slice but use the index instead

dot/node.go
EclesioMeloJunior: @qdm12 you mean to create a context outside of that function? I'm using `context.Background()` as there is no available context provided by that function caller, however I can create but the context will be used only by this function (`setupTelemetry`)

dot/rpc/http_test.go
EclesioMeloJunior: yes!

dot/rpc/modules/author_integration_test.go
EclesioMeloJunior: This file is related to the author integration test PR, once this PR was merged I will address this there!

dot/network/host.go
```diff
@@ -111,7 +111,7 @@ func newHost(ctx context.Context, cfg *Config) (*host, error) {
 		if err != nil {
 			logger.Errorf("failed to get public IP error: %v", err)
 		} else {
-			logger.Debugf("got public IP", "IP", ip)
+			logger.Debugf("got public IP address %s", ip)
```
qdm12, line 114: nit :bulb:  this is merged in development

test.log
```diff
@@ -0,0 +1,378 @@
+=== RUN   TestEncodeBlockAnnounce
```
qdm12, line 1: remove

also why do you output test logs to files by the way? Why not just to stdout?

Maybe we should .gitignore *.log files?

dot/node.go
qdm12: Interesting. Should I create an issue to review all usage of `context.Background()` and refactor code to pass down contextes? :thinking:

test.log
EclesioMeloJunior: Yes, I will definitely ignore `*.log` file. I use it because sometimes is better to find things using `ctrl+f` 😂

dot/node.go
EclesioMeloJunior: Will be a nice issue!

test.log
qdm12: A little trick is to pipe `go test ./... | code -` so there is no file involved :wink: 
But yeah let's ignore log files too.

dot/build_spec.go
```diff
@@ -94,33 +96,44 @@ func BuildFromDB(path string) (*BuildSpec, error) {
 	tmpGen.Genesis.Raw = make(map[string]map[string]string)
 	tmpGen.Genesis.Runtime = make(map[string]map[string]interface{})
 
+	// ignore the error value because there is no telemetry endpoint to connect
+	// error might happen only if there's some problem while connecting
+	disabledTelemetry, err := telemetry.BootstrapMailer(context.Background(), nil, false, nil)
+	if err != nil {
+		panic(fmt.Sprintf("telemetry should not fail at BuildFromDB function: %s\n", err.Error()))
```
qdm12, line 103: nit simpler and no need for a trailing new line with panics

```suggestion
		panic("telemetry should not fail at BuildFromDB function: " + err.Error())
```

dot/build_spec.go
```diff
@@ -94,33 +96,44 @@ func BuildFromDB(path string) (*BuildSpec, error) {
 	tmpGen.Genesis.Raw = make(map[string]map[string]string)
 	tmpGen.Genesis.Runtime = make(map[string]map[string]interface{})
 
+	// ignore the error value because there is no telemetry endpoint to connect
+	// error might happen only if there's some problem while connecting
```
qdm12, line 100: I'm confused by those two lines... don't they contradict each other?

dot/build_spec.go
```diff
@@ -94,33 +96,44 @@ func BuildFromDB(path string) (*BuildSpec, error) {
...
+	// start state service (initialise state database)
+	err = stateSrvc.Start()
+	if err != nil {
+		return nil, fmt.Errorf("cannot start state service: %w", err)
+	}
 	// set genesis fields data
 	ent, err := stateSrvc.Storage.Entries(nil)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("fail to get storage trie entries: %w", err)
```
qdm12, line 127: nit

```suggestion
		return nil, fmt.Errorf("failed to get storage trie entries: %w", err)
```

dot/build_spec.go
```diff
@@ -94,33 +96,44 @@ func BuildFromDB(path string) (*BuildSpec, error) {
...
 	}
-
 	err = genesis.BuildFromMap(ent, tmpGen)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("fail to build from map: %w", err)
```
qdm12, line 131:
```suggestion
		return nil, fmt.Errorf("failed to build from map: %w", err)
```

dot/build_spec.go
```diff
@@ -94,33 +96,44 @@ func BuildFromDB(path string) (*BuildSpec, error) {
...
 	}
-
 	// set genesisData
 	gd, err := stateSrvc.DB().Get(common.GenesisDataKey)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("fail to retrieve genesis data: %w", err)
```
qdm12, line 136: nit

```suggestion
		return nil, fmt.Errorf("failed to retrieve genesis data: %w", err)
```

dot/core/helpers_test.go
```diff
@@ -51,10 +53,17 @@ func NewTestService(t *testing.T, cfg *Config) *Service {
 	if cfg.BlockState == nil || cfg.StorageState == nil ||
 		cfg.TransactionState == nil || cfg.EpochState == nil ||
 		cfg.CodeSubstitutedState == nil {
+
```
qdm12, line 56:
```suggestion
```

dot/core/service_test.go
```diff
@@ -27,6 +28,7 @@ import (
 	"github.com/ChainSafe/gossamer/lib/transaction"
 	"github.com/ChainSafe/gossamer/lib/trie"
 	"github.com/ChainSafe/gossamer/pkg/scale"
+	gomock "github.com/golang/mock/gomock"
```
qdm12, line 31:
```suggestion
	"github.com/golang/mock/gomock"
```

dot/network/state.go
```diff
@@ -13,7 +13,7 @@ import (
 	"github.com/ChainSafe/gossamer/lib/common"
 )
 
-//go:generate mockery --name BlockState --structname MockBlockState --case underscore --inpackage
+//go:generate mockgen -destination=mock_block_state.go -package $GOPACKAGE . BlockState
```
qdm12, line 16: Write it to a file with a `_test.go` suffix so it doesn't pollute the Go API perhaps? Same for below if you can / it's not imported by other packages.

dot/network/transaction_test.go
```diff
@@ -31,29 +32,32 @@ func TestDecodeTransactionHandshake(t *testing.T) {
...
 
+	expectedMsgArg := &TransactionMessage{
+		Extrinsics: []types.Extrinsic{{1, 1}, {2, 2}},
+	}
+
+	ctrl := gomock.NewController(t)
+	transactionHandler := NewMockTransactionHandler(ctrl)
+	transactionHandler.EXPECT().
+		HandleTransactionMessage(peer.ID(""), expectedMsgArg).
+		Return(true, nil).MaxTimes(1)
```
qdm12, line 43: No need by default it's once. You can remove all `.MaxTimes(1)`

```suggestion
		Return(true, nil)
```

dot/node.go
qdm12: Created https://github.com/ChainSafe/gossamer/issues/2145 although it's kind of a recursive refactoring until we reach back to `dot/services.go` really :thinking:

dot/build_spec.go
```diff
@@ -96,11 +96,10 @@ func BuildFromDB(path string) (*BuildSpec, error) {
 	tmpGen.Genesis.Raw = make(map[string]map[string]string)
 	tmpGen.Genesis.Runtime = make(map[string]map[string]interface{})
 
-	// ignore the error value because there is no telemetry endpoint to connect
-	// error might happen only if there's some problem while connecting
+	// BootstrapMailer should not return an error here since there is no URLs to connect
```
qdm12, line 99: meganit

```suggestion
	// BootstrapMailer should not return an error here since there is no URLs to connect to
```

dot/network/service_test.go
```diff
@@ -83,6 +83,9 @@ func createTestBlockResponseMessage(t *testing.T) *BlockResponseMessage {
 	return msg
 }
 
+//go:generate mockgen -destination=mock_block_state.go -package $GOPACKAGE . BlockState
+//go:generate mockgen -destination=mock_syncer.go -package $GOPACKAGE . Syncer
```
qdm12, line 87: Can you perhaps have these with `_test.go` suffixes? Or are these imported by other packages :thinking:

dot/network/transaction_test.go
```diff
@@ -29,6 +29,7 @@ func TestDecodeTransactionHandshake(t *testing.T) {
 	require.Equal(t, testHandshake, msg)
 }
 
+//go:generate mockgen -destination=mock_transaction_handler.go -package $GOPACKAGE . TransactionHandler
```
qdm12, line 32: Same question here, can you have it as a test file so it's not exported out of this package?

dot/telemetry/mailer_test.go
```diff
@@ -159,7 +174,8 @@ func TestListenerConcurrency(t *testing.T) {
 
 	defer cancel()
 
-	resultCh = make(chan []byte)
+	resultCh := make(chan []byte)
+	mailer := bootstrapMailer2Test(t, resultCh)
```
qdm12, line 178: Is resultsCh used at all? :thinking:

dot/telemetry/mailer_test.go
```diff
@@ -106,13 +119,13 @@ func TestHandler_SendMulti(t *testing.T) {
 			"1"),
 	}
 
-	resultCh = make(chan []byte)
-
+	resultCh := make(chan []byte)
+	mailer := bootstrapMailer2Test(t, resultCh)
```
qdm12, line 123: Maybe you could have the assertion code inside the handler instead of piping the handler data into the resultsCh?

dot/node_test.go
```diff
@@ -194,9 +196,15 @@ func TestInitNode_LoadGenesisData(t *testing.T) {
 	err := InitNode(cfg)
 	require.NoError(t, err)
 
+	telemetryNotEnabled, err := telemetry.
+		BootstrapMailer(context.Background(), nil, false, nil)
```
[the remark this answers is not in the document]
EclesioMeloJunior: Sure!

dot/state/storage_notify_test.go
```diff
@@ -45,7 +45,7 @@ func TestStorageState_RegisterStorageObserver(t *testing.T) {
 		}},
 	}
 
-	time.Sleep(time.Millisecond)
+	time.Sleep(time.Second + 1)
```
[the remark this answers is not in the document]
EclesioMeloJunior: changed to `time.Milisecond * 250`, this test calls the `RegisterStorageObserver` which execs goroutines, this sleep basically waits for those goroutines execution, I'm not sure why, and can be improved as well but is something for another PR

dot/telemetry/mailer.go
```diff
@@ -15,38 +15,45 @@ import (
...
+	enabled      bool
 }
 
-func newMailer(logger log.LeveledLogger) *mailer {
-	return &mailer{
-		messageQueue: messageQueue,
+func newMailer(enabled bool, logger log.LeveledLogger) *Mailer {
+	return &Mailer{
+		enabled:      enabled,
+		messageQueue: make(chan Message),
```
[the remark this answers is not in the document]
EclesioMeloJunior: Thinking about it will be better to have a mutex instead of a queue because the telemetry mailer is used by some services whose calls the `SendMessage` inside inner goroutines

dot/telemetry/mailer_test.go
```diff
@@ -23,32 +23,45 @@ import (
...
 
-	time.Sleep(time.Millisecond)
-	// instantiate telemetry to connect to websocket (test) server
-	var testEndpoints []*genesis.TelemetryEndpoint
+	mux := http.NewServeMux()
+	mux.HandleFunc("/", listen)
+
+	srv := httptest.NewServer(mux)
+	t.Cleanup(func() {
+		srv.Close()
```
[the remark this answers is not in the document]
EclesioMeloJunior: The `func (*httptest.Server).Close()` does not return any value

dot/telemetry/telemetry.go
```diff
@@ -23,6 +23,11 @@ const (
 	txPoolImportMsg = "txpool.import"
 )
 
+// Client is the interface required by send messages to telemetry servers
```
qdm12, line 26: nit

```suggestion
// Client is the interface to send messages to telemetry servers
```

dot/telemetry/mailer_test.go
EclesioMeloJunior: Added an assertion to the `resultCh` messages

lib/grandpa/grandpa_test.go
```diff
@@ -51,9 +50,11 @@ func NewMockDigestHandler() *mocks.DigestHandler {
 	return m
 }
 
+//go:generate mockgen -source=../../dot/telemetry/telemetry.go -destination=mock_telemetry_test.go -package $GOPACKAGE Client
```
qdm12, line 53: Don't use `-source`, just the Go module name `github.com/ChainSafe/gossamer/dot/telemetry`

```suggestion
//go:generate mockgen -destination=mock_telemetry_test.go -package $GOPACKAGE github.com/ChainSafe/gossamer/dot/telemetry Client
```

Same for other usage of `-source`

qdm12 (changes requested):

dot/telemetry/mailer_test.go
EclesioMeloJunior: yes, I will need to pass as argument an `http.HandlerFunc` instead of a channel and for each test write its own handler, it's a good suggestion, I will try to implement

dot/build_spec.go
```diff
@@ -94,33 +96,43 @@ func BuildFromDB(path string) (*BuildSpec, error) {
 	tmpGen.Genesis.Raw = make(map[string]map[string]string)
 	tmpGen.Genesis.Runtime = make(map[string]map[string]interface{})
 
+	// BootstrapMailer should not return an error here since there is no URLs to connect to
+	disabledTelemetry, err := telemetry.BootstrapMailer(context.Background(), nil, false, nil)
```
qdm12, line 100: Maybe change to `context.TODO()` :thinking:

dot/network/service.go
```diff
@@ -385,18 +388,14 @@ func (s *Service) publishNetworkTelemetry(done <-chan struct{}) {
 	ticker := time.NewTicker(s.telemetryInterval)
 	defer ticker.Stop()
 
-main:
 	for {
 		select {
 		case <-done:
-			break main
+			return
```
qdm12, line 394: :pray: awesome

dot/network/service.go
```diff
@@ -415,17 +414,15 @@ func (s *Service) sentBlockIntervalTelemetry() {
 		}
 		finalizedHash := finalised.Hash()
 
-		err = telemetry.SendMessage(telemetry.NewBlockIntervalTM(
+		s.telemetry.SendMessage(telemetry.NewBlockIntervalTM(
```
qdm12, line 417: Is it ok to ignore the error? :thinking:

dot/network/transaction_test.go
```diff
@@ -28,32 +29,36 @@ func TestDecodeTransactionHandshake(t *testing.T) {
...
+	}
+
+	ctrl := gomock.NewController(t)
+	transactionHandler := NewMockTransactionHandler(ctrl)
+	transactionHandler.EXPECT().
+		HandleTransactionMessage(peer.ID(""), expectedMsgArg).
 		Return(true, nil)
-	mockhandler.On("TransactionsCount").Return(0)
+
+	transactionHandler.EXPECT().TransactionsCount().Return(0).MaxTimes(1)
```
qdm12, line 46: nit you could also batch replace all `.MaxTimes(1)` with nothing

```suggestion
	transactionHandler.EXPECT().TransactionsCount().Return(0)
```

dot/node_test.go
```diff
@@ -194,9 +198,16 @@ func TestInitNode_LoadGenesisData(t *testing.T) {
 	err := InitNode(cfg)
 	require.NoError(t, err)
 
+	ctrl := gomock.NewController(t)
+	telemetryMock := NewMockClient(ctrl)
+	telemetryMock.EXPECT().SendMessage(gomock.AssignableToTypeOf(&telemetry.NotifyFinalizedTM{}))
```
qdm12, line 203: If there is only one message being sent, you could easily (I think) expect the actual message instead of just its type

dot/telemetry/mailer_test.go
```diff
@@ -106,42 +113,64 @@ func TestHandler_SendMulti(t *testing.T) {
...
+
+			actual[idx] = msg
+		}
+
+		// sort the actual slice in alphabetical order
+		sort.Slice(actual, func(i, j int) bool {
+			return bytes.Compare(actual[i], actual[j]) < 0
+		})
 
+		// assert
```
qdm12, line 150: nit

```suggestion
```

dot/telemetry/mailer_test.go
```diff
@@ -150,6 +179,41 @@ func TestListenerConcurrency(t *testing.T) {
...
+			counter++
+
+			assert.Contains(t,
+				string(msg),
+				expectedResult,
+			)
+
+			if counter == qty {
+				break
+			}
```
qdm12, line 213: Have `counter` as the index of this for loop

dot/network/service.go
qdm12: Never mind, didn't see `SendMessage` no longer returns an error

dot/node.go
```diff
@@ -204,14 +209,52 @@ func NewNode(cfg *Config, ks *keystore.GlobalKeystore) (*Node, error) {
...
+			cfg.Core.GrandpaAuthority,
+			sysSrvc.ChainName(),
+			&genesisHash,
+			sysSrvc.SystemName(),
+			cfg.Global.Name,
+			networkSrvc.NetworkState().PeerID,
+			fmt.Sprint(time.Now().UnixNano()),
+			sysSrvc.SystemVersion())
+
+		telemetryMailer.SendMessage(connectedMsg)
```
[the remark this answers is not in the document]
qdm12: Never mind, didn't see `SendMessage` no longer returns an error

dot/node.go
timwu20: I asked @edwardmack to do some refactoring around this.  I don't see the need for both `InitNode` and `NewNode` to be called.

dot/telemetry/mailer_test.go
```diff
@@ -20,35 +20,42 @@ import (
 	"github.com/ChainSafe/gossamer/lib/common"
 	"github.com/ChainSafe/gossamer/lib/genesis"
 	"github.com/gorilla/websocket"
+	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
 
-var upgrader = websocket.Upgrader{}
-var resultCh chan []byte
+func bootstrapMailer2Test(t *testing.T, handler http.HandlerFunc) (mailer *Mailer) {
```
timwu20, line 27:
```suggestion
func newTestMailer(t *testing.T, handler http.HandlerFunc) (mailer *Mailer) {
```

timwu20 (changes requested):

dot/build_spec.go
EclesioMeloJunior: Better suitable

dot/network/service.go
EclesioMeloJunior: `SendMessage` does not return error

dot/network/service_test.go
```diff
@@ -72,17 +106,42 @@ func createTestService(t *testing.T, cfg *Config) (srvc *Service) {
 	}
 
 	if cfg.BlockState == nil {
-		cfg.BlockState = NewMockBlockState(nil)
+		parentHash := common.MustHexToHash("0x4545454545454545454545454545454545454545454545454545454545454545")
+		stateRoot := common.MustHexToHash("0xb3266de137d20a5d0ff3a6401eb57127525fd9b2693701f0bf5a8a853fa3ebe0")
+		extrinsicsRoot := common.MustHexToHash("0x03170a2e7597b7b7e3d84c05391d139a62b157e78786d8c082f29dcf4c111314")
```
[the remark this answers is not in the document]
EclesioMeloJunior: no, they actually can be just `common.Hash{}`

dot/state/service.go
```diff
@@ -85,34 +90,46 @@ func (s *Service) DB() chaindb.Database {
...
 }
 
-// Start initialises the Storage database and the Block database.
-func (s *Service) Start() error {
-	if !s.isMemDB && (s.Storage != nil || s.Block != nil || s.Epoch != nil || s.Grandpa != nil) {
+// SetTelemetry add a telemetry impl to state.Service.Telemetry field
+func (s *Service) SetTelemetry(t telemetry.Client) {
+	s.Telemetry = t
+}
+
```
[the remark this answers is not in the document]
EclesioMeloJunior: Really we don't need, removed!

dot/telemetry/mailer_test.go
```diff
@@ -20,35 +20,42 @@ import (
...
+	mux := http.NewServeMux()
+	mux.HandleFunc("/", handler)
 
-	time.Sleep(time.Millisecond)
-	// instantiate telemetry to connect to websocket (test) server
-	var testEndpoints []*genesis.TelemetryEndpoint
+	srv := httptest.NewServer(mux)
+	t.Cleanup(func() {
+		srv.Close()
+	})
```
qdm12, line 36: nit do you need this? Maybe `httptest` handles it for you? :thinking: I have no idea, just being curious here

qdm12 (approved):

dot/telemetry/mailer_test.go
EclesioMeloJunior: Yes we need, `httptest` does not closes after test ends (https://pkg.go.dev/net/http/httptest#example-Server)

dot/network/message_test.go
```diff
@@ -105,12 +104,18 @@ func TestBlockRequestString(t *testing.T) {
 		Max:           nil,
 	}
 
-	_ = bm.String()
+	var blockRequestStringRegex = regexp.MustCompile(
+		`^\ABlockRequestMessage RequestedData=[0-9]* StartingBlock={[\[0-9(\s?)]+\]} EndBlockHash=0x[0-9]+ Direction=[0-9]* Max=[0-9]*\z$`) //nolint:lll
+
+	fmt.Println(bm.String())
```
qdm12, line 110:
```suggestion
```

dot/peerset/peerset.go
```diff
@@ -666,18 +666,26 @@ func (ps *PeerSet) disconnect(setIdx int, reason DropReason, peers ...peer.ID) e
 }
 
 // start handles all the action for the peerSet.
-func (ps *PeerSet) start(aq chan action) {
+func (ps *PeerSet) start(aq chan action, stopCh chan struct{}) {
```
qdm12, line 669: pass in a context instead of `stopCh`?

dot/peerset/peerset.go
```diff
@@ -666,18 +666,26 @@ func (ps *PeerSet) disconnect(setIdx int, reason DropReason, peers ...peer.ID) e
...
 	ps.actionQueue = aq
+
 	ps.resultMsgCh = make(chan Message, msgChanSize)
-	go ps.doWork()
+	go ps.doWork(stopCh)
 }
 
-func (ps *PeerSet) doWork() {
+func (ps *PeerSet) doWork(stopCh chan struct{}) {
+	ps.stopCh = make(chan struct{})
```
qdm12, line 677: Pass a context here otherwise it's confusing to have two stopCh

dot/peerset/peerstate.go
```diff
@@ -108,6 +109,8 @@ type PeersState struct {
 	// configuration of each set. The size of this Info is never modified.
 	// since, single Info can also manage the flow.
 	sets []Info
+
+	mu *sync.Mutex
```
qdm12, line 113: Does it have to be a pointer to a mutex? :thinking: A mutex value would do I think here

lib/utils/test_utils_test.go
```diff
@@ -22,20 +22,6 @@ func TestNewTestDir(t *testing.T) {
...
-
-	testDir := NewTestBasePath(t, basePath)
-
-	expected := path.Join(TestDir, t.Name(), basePath)
-
-	require.Equal(t, expected, testDir)
-	require.Equal(t, PathExists(testDir), true)
-
-	RemoveTestDir(t)
-}
```
qdm12, line 37: Why removing this?

qdm12 (changes requested):

dot/network/service.go
qdm12: Ooops sorry Tim I forgot it was you :smile: Feeling bad now lol!

dot/network/utils_test.go
```diff
@@ -5,26 +5,51 @@ package network
...
+	select {
+	case pq <- p:
+	default:
+	}
+}
+
+func (pq portQueue) get() (port uint16) {
+	select {
+	case port = <-pq:
+	default:
```
qdm12, line 29: Make it blocking, you don't want to return `0` here

dot/network/utils_test.go
```diff
@@ -5,26 +5,51 @@ package network
...
+// portQueue is a non-blocking port queue
+type portQueue chan uint16
 
-	l, err := net.ListenTCP("tcp", addr)
-	require.NoError(t, err)
+func (pq portQueue) put(p uint16) {
+	select {
+	case pq <- p:
+	default:
+	}
```
qdm12, line 23: Make it blocking, you don't want to not put it back here

dot/network/utils_test.go
```diff
@@ -5,26 +5,51 @@ package network
...
 
-	port := uint16(l.Addr().(*net.TCPAddr).Port)
+	return port
+}
+
+var availablePorts portQueue
+
+func init() {
+	availablePorts = make(chan uint16, portsAmount)
+	startAt := uint16(7500)
```
qdm12, line 39:
```suggestion
	const startAt = uint16(7500)
```

dot/network/utils_test.go
```diff
@@ -5,26 +5,51 @@ package network
...
+// availablePort is test helper function that gets an available port and release the same port after test ends
+func availablePort(t *testing.T) uint16 {
+	t.Helper()
+	port := availablePorts.get()
 
-	err = l.Close()
-	require.NoError(t, err, "failed to release port: %d", port)
+	t.Cleanup(func() {
+		availablePorts.put(port)
+	})
```
qdm12, line 52: Nice!

dot/peerset/peerset.go
```diff
@@ -666,15 +666,14 @@ func (ps *PeerSet) disconnect(setIdx int, reason DropReason, peers ...peer.ID) e
 }
 
 // start handles all the action for the peerSet.
-func (ps *PeerSet) start(aq chan action, stopCh chan struct{}) {
+func (ps *PeerSet) start(ctx context.Context, aq chan action) {
```
qdm12, line 669: Rename `aq` to `actionQueue` please; also maybe restrict to a read only / write only channel if you can

dot/peerset/test_helpers.go
```diff
@@ -24,7 +25,7 @@ const (
 	peer2         = peer.ID("testPeer2")
 )
 
-func newTestPeerSet(t *testing.T, in, out uint32, bootNodes, reservedPeers []peer.ID, reservedOnly bool) *Handler {
+func newTestPeerSet(t *testing.T, ctx context.Context, in, out uint32, bootNodes, reservedPeers []peer.ID, reservedOnly bool) *Handler {
```
qdm12, line 28: Why not use `context.Background()` directly in there? I doubt we would have another use in tests right? Or maybe we would, not sure.

EclesioMeloJunior: yeah, I will change to use `context.Background()` directly!

lib/utils/test_utils_test.go
EclesioMeloJunior: because this Test tests if the `NewTestBasePath` (which is a helper test) so I think this is not useful

EclesioMeloJunior: And I modified the `NewTestBasePath` to use `t.TmpDir()` instead of create a `./test_data` file in the current folder to test things

dot/peerset/peerstate.go
```diff
@@ -139,6 +142,7 @@ func NewPeerState(cfgs []*config) (*PeersState, error) {
 	peerState := &PeersState{
 		nodes: make(map[peer.ID]*node),
 		sets:  infoSet,
+		mu:    sync.Mutex{},
```
qdm12, line 145:
```suggestion
```

dot/peerset/peerstate.go
```diff
@@ -207,6 +211,21 @@ func (ps *PeersState) sortedPeers(idx int) peer.IDSlice {
 	return peerIDs
 }
 
+func (ps *PeersState) addReputation(pid peer.ID, change ReputationChange) (Reputation, error) {
```
qdm12, line 214: please name returns

```suggestion
func (ps *PeersState) addReputation(pid peer.ID, change ReputationChange) (
	newReputation Reputation, err error) {
```

qdm12 (approved):

dot/peerset/peerset.go
EclesioMeloJunior: the `PeerSet.actionQueue` already is a `<-chan action` (read only channel)

qdm12 (approved):

dot/peerset/peerset.go
qdm12: So you could restrict the argument to be `actionQueue chan<- action` too :wink:

lib/utils/test_utils_test.go
qdm12: Awesome! By the way feel free to review my start-of-the-year #2137 :smile:

timwu20 (changes requested):

dot/telemetry/mailer.go
```diff
@@ -0,0 +1,140 @@
...
+		defer conn.Unlock()
+
+		err = conn.wsconn.WriteMessage(websocket.TextMessage, msgBytes)
+		if err != nil {
+			m.logger.Debugf("issue while sending %T telemetry message: %s", msg, err)
+		}
+	}
+}
+
+func msgToJSON(message Message) ([]byte, error) {
```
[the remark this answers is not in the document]
EclesioMeloJunior: @timwu20 implemented!

dot/telemetry/mailer.go
```diff
@@ -117,24 +119,24 @@ func (m *Mailer) shipTelemetryMessage(msg Message) {
...
+
+// 	messageMap["ts"] = time.Now()
+// 	messageMap["msg"] = message.messageType()
+
+// 	fullRes, err := json.Marshal(messageMap)
+// 	if err != nil {
+// 		return nil, err
+// 	}
+// 	return fullRes, nil
+// }
```
timwu20: remove please

dot/telemetry/mailer.go
```diff
@@ -100,7 +101,8 @@ func (m *Mailer) SendMessage(msg Message) {
 }
 
 func (m *Mailer) shipTelemetryMessage(msg Message) {
-	msgBytes, err := msgToJSON(msg)
+	msgBytes, err := json.Marshal(msg)
+	fmt.Printf(">>>>>> %s\n", string(msgBytes))
```
timwu20: remove

dot/telemetry/telemetry.go
```diff
@@ -31,4 +31,5 @@ type Client interface {
 // Message interface for Message functions
 type Message interface {
 	messageType() string
```
timwu20, line 33: Does this need to be private still?

timwu20 (changes requested):

dot/telemetry/telemetry.go
EclesioMeloJunior: no, changed to `Type()`

PR #2127, opened 2021-12-17
Jimmy/babe unit tests

jimjbrettj:
## Changes

- Created new unit tests for babe/crypto.go, babe/secondary.go, and babe/verify.go
- @noot is doing a babe refactor, so will wait to redo the remaining tests

## Tests

```
go test ./lib/babe/ -v -coverprofile coverage.out
go tool cover -func coverage.out // To see more detailed test coverage
```

## Issues

- #1831 

## Primary Reviewer

- @noot


lib/babe/babe_integration_test.go
```diff
@@ -408,4 +408,3 @@ func TestService_PauseAndResume(t *testing.T) {
 	err = bs.Stop()
 	require.NoError(t, err)
 }
-
```
qdm12, line 411: What's wrong with trailing new lines? Let's think about CLI people using `cat` or other poor man's editor :smile:

lib/babe/crypto.go
```diff
@@ -84,6 +84,9 @@ func checkPrimaryThreshold(randomness Randomness,
 // equation: threshold = 2^128 * (1 - (1-c)^(1/len(authorities))
 // see https://github.com/paritytech/substrate/blob/master/client/consensus/babe/src/authorship.rs#L44
 func CalculateThreshold(C1, C2 uint64, numAuths int) (*scale.Uint128, error) {
+	if C1 == 0 || C2 == 0 {
+		return nil, errors.New("invalid input: C1 and C2 cannot be 0")
```
qdm12, line 88: make this a global exported error i.e.

```go
var ErrThresholdBothZero = errors.New("both inputs cannot be zero")
```

and have it as

```suggestion
		return nil, ErrThresholdBothZero
```

Also, unless `C1` and `C2` have an actual naming meaning, consider changing them to `c1` and `c2` and don't mention them in the error message.

lib/babe/mocks/mock_state.go
```diff
@@ -0,0 +1,839 @@
+// Code generated by MockGen. DO NOT EDIT.
+// Source: ./lib/babe/state.go
+
+// Package mock_babe is a generated GoMock package.
+package mocks
```
qdm12, line 5: `//go:generate` comment is missing I think.

Please also try to have it generated as `mock_state_test.go` in the `lib/babe` package so it doesn't pollute the Go API (tests exported things are not exported to other packages)

lib/babe/mocks/mock_state.go
```diff
@@ -0,0 +1,839 @@
...
+	big "math/big"
+	reflect "reflect"
+	time "time"
+
+	types "github.com/ChainSafe/gossamer/dot/types"
+	common "github.com/ChainSafe/gossamer/lib/common"
+	runtime "github.com/ChainSafe/gossamer/lib/runtime"
+	storage "github.com/ChainSafe/gossamer/lib/runtime/storage"
+	transaction "github.com/ChainSafe/gossamer/lib/transaction"
+	gomock "github.com/golang/mock/gomock"
```
qdm12, line 17: All these don't need to be aliased. Girlfriend is coming to get me, my review will end here for now :laughing:

lib/babe/babe_integration_test.go
jimjbrettj: ah you are so right! I think goland format did it to me

EclesioMeloJunior (changes requested):
I think `lib/babe/crypto_integration_test` can be removed because the `lib/babe/crypto.go` does not depends from others services or interfaces

lib/babe/crypto.go
noot: the error should say `numerator or denominator cannot be 0` or something like that, as C1 and C2 represent the numerator and denominator to `c` in the equation in the comment, and neither should be 0

lib/babe/crypto_test.go
```diff
@@ -1,4 +1,187 @@
...
+				numAuths: 0,
+			},
+			expErr: errors.New("invalid C1/C2: greater than 1"),
+		},
+		{
+			name: "max threshold",
+			args: args{
+				C1:       2147483647,
+				C2:       2147483647,
+				numAuths: 3,
```
noot, line 61:
```suggestion
			args: args{
				C1:       1,
				C2:       1,
				numAuths: 3,
```

can you do this (since it's the c1/c2 fraction that gets used in the equation)?

lib/babe/verify_test.go
```diff
@@ -1,4 +1,1296 @@
...
+		vm     *VerificationManager
+		epoch  uint64
+		exp    *types.ConfigData
+		expErr error
+	}{
+		{
+			name:   "cant find ConfigData",
+			vm:     vm0,
+			epoch:  0,
+			expErr: errors.New("cannot find ConfigData for epoch"),
```
noot, line 813: can you use named errors here? eg. 
`var errNoConfigData = errors.New("cannot find ConfigData for epoch")` and assert that the error is of type `errNoConfigData`, same for the others here

lib/babe/verify_test.go
```diff
@@ -1,4 +1,1296 @@
...
+		vm     *VerificationManager
+		epoch  uint64
+		exp    *verifierInfo
+		expErr error
+	}{
+		{
+			name:   "getEpochData error",
+			vm:     vm0,
+			epoch:  0,
+			expErr: errors.New("failed to get epoch data for epoch 0: cant get ConfigData"),
```
noot, line 905: same here

lib/babe/verify_test.go
```diff
@@ -1,4 +1,1296 @@
...
+		name   string
+		vm     *VerificationManager
+		header *types.Header
+		expErr error
+	}{
+		{
+			name:   "fail to check block 1 finalisation",
+			vm:     vm0,
+			header: block1Header,
+			expErr: errors.New("failed to check if block 1 is finalised: failed to check finalisation"),
```
noot, line 1058: and here

lib/babe/verify_test.go
```diff
@@ -1,4 +1,1296 @@
...
+		expErr error
+	}{
+		{
+			name: "get epoch err",
+			vm:   vm0,
+			args: args{
+				index:  0,
+				header: types.NewEmptyHeader(),
+			},
+			expErr: errors.New("get epoch error"),
```
noot, line 1230: also here

noot (commented):
basically looks good, just a few comments!

lib/babe/crypto_test.go
jimjbrettj: Yup good call much cleaner

lib/babe/state.go
```diff
@@ -15,6 +15,8 @@ import (
 	"github.com/ChainSafe/gossamer/lib/transaction"
 )
 
+//go:generate mockgen -source=./state.go -destination=./mock_state_test.go
```
EclesioMeloJunior, line 18: you don't need to use -source when the interface is in the same package, you can do instead
`go:generate mockgen -destination=./mock_state_test.go -package $GOPACKAGE . BlockState`
then it will generate using the `package babe`. However if you need to use the source flag you can use with this path format `github.com/ChainSafe/gossamer/babe`

lib/babe/verify.go
```diff
@@ -174,7 +175,7 @@ func (v *VerificationManager) VerifyBlock(header *types.Header) error {
 			// thus missing the epoch data for previous epochs.
 			skip, err2 := v.epochState.SkipVerify(header)
```
EclesioMeloJunior, line 176: could you rename to `skipErr`

lib/babe/verify_test.go
```diff
@@ -2,3 +2,1295 @@
...
+	"github.com/ChainSafe/gossamer/dot/types"
+	"github.com/ChainSafe/gossamer/lib/common"
+	"github.com/ChainSafe/gossamer/lib/crypto/sr25519"
+	"github.com/ChainSafe/gossamer/pkg/scale"
+	"github.com/golang/mock/gomock"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func createNewTestHeader(t *testing.T, digest ...scale.VaryingDataTypeValue) *types.Header {
```
EclesioMeloJunior, line 20: add `t.Helper()` here

EclesioMeloJunior: and rename to `newTestHeader` as the other helper functions has `new...`

EclesioMeloJunior (changes requested):

lib/babe/state.go
jimjbrettj: If I omit the ". BlockState" will it still generate mocks for all interfaces in the file? Or will I need to add line for each interface?

qdm12: You can have it with a comma like `. BlockState,SomeOtherInterface` if you want to have them on a single line.

I nowadays tend to write `//go:generate` comments above the test / test file needing it too since mocks are really a test only thing, so that usually removes the need to have all the go generate interfaces on a single line too. But that's just a preference, both work the same and it's unclear still which way is better :smile:

lib/babe/crypto_test.go
```diff
@@ -2,3 +2,187 @@
...
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			res, err := CalculateThreshold(tt.args.C1, tt.args.C2, tt.args.numAuths)
+			if tt.expErr != nil {
+				assert.EqualError(t, err, tt.expErr.Error())
+			} else {
+				assert.NoError(t, err)
+			}
```
qdm12, line 73: Ideally assert the error matches from a programmatic standpoint, using `assert.ErrorIs(t, err, tt.expErr)` where `tt.expErr` is the sentinel error or the wrapped sentinel error.

You can then `assert.EqualError(t, err, tt.expErrMsg)` if you want to assert the message is as expected, this is more from a logging standpoint.

You can compact both with

```go
assert.ErrorIs(t, err, tt.expErr) // takes care of nil as well
if tt.expErr != nil { // only assert message if we expect the error to not be nil so you don't have to set `tt.expErrMsg` to <nil> every time
    assert.EqualError(t, err, tt.expErrMsg)
}
```

lib/babe/crypto_test.go
```diff
@@ -2,3 +2,187 @@
...
+		exp    bool
+		expErr error
+	}{
+		{
+			name: "happy path true",
+			args: args{
+				randomness: Randomness{},
+				slot:       uint64(0),
+				epoch:      uint64(0),
+				output:     [32]byte{},
```
qdm12, line 102: Unless it's generated, otherwise remove empty fields

At least remove the unneeded conversions `uint64(0)` and change to `0`

```suggestion
```

jimjbrettj:
> looks good! do you wanna merge this into dev when it's approved perhaps?

@noot Yeah I can do that! Then I would just have his PR close out both the integration and unit test issues?

lib/babe/state.go
jimjbrettj: okay cool I can make this change

lib/babe/crypto_test.go
jimjbrettj: @qdm12 question: is it better to do error comparisons using strings vs the errors themselves? Also is it worth adding the extra error msg field to all the tests to make this change?

lib/babe/verify.go
```diff
@@ -172,9 +172,9 @@ func (v *VerificationManager) VerifyBlock(header *types.Header) error {
 			v.lock.Unlock()
 			// SkipVerify is set to true only in the case where we have imported a state at a given height,
 			// thus missing the epoch data for previous epochs.
-			skip, err2 := v.epochState.SkipVerify(header)
-			if err2 != nil {
-				return fmt.Errorf("failed to check if verification can be skipped: %w", err)
+			skip, skipErr := v.epochState.SkipVerify(header)
```
timwu20, line 175: 👍

lib/babe/verify_test.go
```diff
@@ -2,3 +2,1236 @@
...
+
+	return header
+}
+
+func signAndAddSeal(t *testing.T, kp *sr25519.Keypair, header *types.Header, data []byte) error {
+	t.Helper()
+	sig, err := kp.Sign(data)
+	assert.NoError(t, err)
+
+	return header.Digest.Add(types.SealDigest{
```
timwu20, line 37: why do you `assert.noError` for some on line 35, but not in the actual return?  Seems odd that we make sure no errors are returned here, but the function signature still returns an error, and then `assert.NoError` on the calling side.

timwu20 (changes requested):
Great work Jimmy! Just a few things.

timwu20:
Can you see what's up with the docker stress tests? I'm surprised code-cov is green.

lib/babe/verify_test.go
jimjbrettj: Well my thinking was that DIgest.Add returns an error so I should make sure no errors before that function call, then assert that the add executed correctly on the caller side. Do you think there is a better way to handle that?

lib/babe/crypto_test.go
qdm12: I'd say asserting the wrapped error is more important, since some programmatic behavior depending on it (using `errors.Is(err, someSentinelError)` might break without tests detecting it. The string assertion is just for logs really, so we can skip it really.

Although all this is extra test depth, you can leave it as it is now to accelerate merging. At least you're aware for next tests :wink:

lib/babe/verify_test.go
qdm12: Just add another assert.NoError for the error returned by `header.Digest.Add`?

Also you might want to change the kp.Sign error assertion from `assert.NoError` to `require.NoError` to stop the test execution, since I don't think consecutive assertions would be relevant here

.gitattributes
```diff
@@ -1,2 +1,4 @@
+*.go -text diff=golang
+
```
qdm12, line 2: Rebase on development I think this is already there (`git rebase development`)

lib/babe/babe_test.go
```diff
@@ -2,406 +2,3 @@
 // SPDX-License-Identifier: LGPL-3.0-only
 
 package babe
-
```
qdm12, line 5: Why not remove the file?

lib/babe/build_test.go
```diff
@@ -2,490 +2,3 @@
 // SPDX-License-Identifier: LGPL-3.0-only
 
 package babe
-
```
qdm12, line 5: why not remove the file?

qdm12 (commented):
A bunch of test files with no content, why not just remove them entirely? :thinking:

lib/babe/crypto_test.go
jimjbrettj: Okay cool that makes sense, sounds good to me!

lib/babe/verify_test.go
```diff
@@ -5,396 +5,1227 @@ package babe
...
+	t.Helper()
+	babeDigest := types.NewBabeDigest()
+	err := babeDigest.Set(value)
+	assert.NoError(t, err)
+	return scale.Marshal(babeDigest)
 }
 
-func TestVerificationManager_OnDisabled_InvalidIndex(t *testing.T) {
-	vm := newTestVerificationManager(t, nil)
+func encodeAndHashHeader(t *testing.T, header *types.Header) (common.Hash, error) {
```
timwu20, line 52: If you're going to keep this a test helper function, I don't think it should return an error.  As mentioned in @qdm12, you should `require.NoError` in these cases, and not return an error from this function.

On the flip side we we don't pass `*testing.T` to this function at all and just `require.NoError` on the calling side.

timwu20:
> I changed base branch to dev in order to streamline this being merged to unblock @noot . Only added diffs will be the existing tests being moved to integration tests

This really screws up the discussion since you did some funky merging.  I don't see how changing the base banch would streamline things at this point.  This branch should have been merged into the integration test ones, and then that branch could/should have been rebased.

timwu20 (changes requested):

lib/babe/verify_test.go
jimjbrettj: errors have been removed from helpers now and asserts are now requires

jimjbrettj:
> > I changed base branch to dev in order to streamline this being merged to unblock @noot . Only added diffs will be the existing tests being moved to integration tests
> 
> This really screws up the discussion since you did some funky merging. I don't see how changing the base banch would streamline things at this point. This branch should have been merged into the integration test ones, and then that branch could/should have been rebased.

My bad. Will not do this in the future

lib/babe/verify_test.go
```diff
@@ -41,20 +41,25 @@ func signAndAddSeal(t *testing.T, kp *sr25519.Keypair, header *types.Header, dat
...
-	assert.NoError(t, err)
+	require.NoError(t, err)
 
-	return common.Blake2bHash(encHeader)
+	hash, err := common.Blake2bHash(encHeader)
+	require.NoError(t, err)
+	return hash
 }
 
 func newTestVerifier(t *testing.T, kp *sr25519.Keypair, blockState BlockState,
```
timwu20, line 65: last one to update

jimjbrettj: Done! @timwu20

timwu20 (approved):

lib/babe/crypto_test.go
```diff
@@ -4,46 +4,175 @@
...
+		epoch      uint64
+		output     [sr25519.VRFOutputLength]byte
+		threshold  *scale.Uint128
+		pub        *sr25519.PublicKey
+	}
+	tests := []struct {
+		name   string
+		args   args
+		exp    bool
+		expErr error
```
EclesioMeloJunior, line 94: `expErr` can be removed since it is not used in the cases

lib/babe/secondary_integration_test.go
```diff
@@ -0,0 +1,73 @@
...
+)
+
+func TestVerifySecondarySlotPlain(t *testing.T) {
+	err := verifySecondarySlotPlain(0, 77, 1, Randomness{})
+	require.NoError(t, err)
+
+	err = verifySecondarySlotPlain(0, 77, 2, Randomness{})
+	require.NoError(t, err)
+
+	numAuths := 20
```
EclesioMeloJunior, line 25: nit
```suggestion
	const numAuths = 20
```

lib/babe/secondary_test.go
```diff
@@ -9,62 +9,144 @@ import (
...
+	type args struct {
+		slot       uint64
+		numAuths   int
+		randomness Randomness
+	}
+	tests := []struct {
+		name   string
+		args   args
+		exp    uint32
+		expErr error
```
EclesioMeloJunior, line 25: `expErr` can be removed

lib/babe/secondary_integration_test.go
```diff
@@ -0,0 +1,73 @@
...
+	slot := uint64(77)
+	epoch := uint64(0)
+
+	digest := createSecondaryVRFPreDigest(t, kp, 0, slot, epoch, Randomness{})
+
+	ok, err := verifySecondarySlotVRF(digest, kp.Public().(*sr25519.PublicKey), epoch, 1, Randomness{})
+	require.NoError(t, err)
+	require.True(t, ok)
+
+	numAuths := 20
```
EclesioMeloJunior, line 61: nit
```suggestion
	const numAuths = 20
```

lib/babe/state.go
```diff
@@ -15,6 +15,10 @@ import (
 	"github.com/ChainSafe/gossamer/lib/transaction"
 )
 
+////go:generate mockgen -source=./state.go -destination=./mock_state_test.go
```
EclesioMeloJunior, line 18: this command can be removed since you are using `mockgen`, right?

lib/babe/state.go
```diff
@@ -82,7 +86,7 @@ type DigestHandler interface {
 	HandleDigests(*types.Header)
 }
 
-//go:generate mockery --name BlockImportHandler --structname BlockImportHandler --case underscore --keeptree
+////go:generate mockery --name BlockImportHandler --structname BlockImportHandler --case underscore --keeptree
```
EclesioMeloJunior, line 89: this command can be removed since you are using `mockgen`, right?

EclesioMeloJunior (changes requested):

EclesioMeloJunior (approved):

PR #2136, opened 2022-01-03
chore(tests): unexport test helper functions from non `_test.go` files

qdm12:
## Changes

This is to reduce Go API noise and inter-package dependencies. Further PRs will be added to unexport these further (with other changes) and move them to `_test.go` files (not available in Go API).

- [x] Unexport test helper functions from non `_test.go` files, whenever possible (without other changes)

## Primary Reviewer


dot/utils.go
```diff
@@ -24,8 +24,8 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-// NewTestGenesis returns a test genesis instance using "gssmr" raw data
-func NewTestGenesis(t *testing.T) *genesis.Genesis {
+// newTestGenesis returns a test genesis instance using "gssmr" raw data
+func newTestGenesis(t *testing.T) *genesis.Genesis {
```
EclesioMeloJunior, line 28: is it possible to move those not exported helpers test functions into `dot/utils_test.go` file? And move the function `RandomNodeName` into `dot/node.go` and then remove the file `./dot/utils.go`?

EclesioMeloJunior (changes requested):
LGTM, just a small suggestion to remove the file `dot/utils.go`

dot/utils.go
qdm12: That's definitely the longer term goal. We should have no test helper exported functions at all (prefer test helping code duplication vs package dependency), so the goal is to (through multiple PRs):

1. Unexport what's unexportable (this PR)
2. Unexport the remaining and duplicate test helper code in packages that need it (btw same goes for mocks in my opinion, we shouldn't share mocks across packages to reduce Go API noise).
3. Rename to utils_test.go (or helpers_test.go)

That 3-steps is to have the minimal amount of deltas too

noot (approved):

dot/utils.go
EclesioMeloJunior: Got it!

EclesioMeloJunior (approved):

PR #2137, opened 2022-01-03
chore(tests): use `t.TempDir()`

qdm12:
## Changes

- [x] Remove custom temp dir creation and deletion functions
- [x] Use `t.TempDir()` whenever possible in tests
- [x] Fix/improve `lib/utils/utils_test.go` (was broken on development branch)
- [x] Use `t.TempDir()` to have a test directory to write test files to instead of `./testdata` (to avoid conflicts)

Now

- tests can be run in parallel without worrying about directory/file names (at least in that aspect)
- developers don't need NOT to forget to remove the directory with a defer call, since t.TempDir does it by itself

## Primary Reviewer

- @kishansagathiya


kishansagathiya (commented):
LGTM

Will approve after CI checks are cleared.

EclesioMeloJunior (commented):
LGTM!

lib/keystore/encrypt.go
```diff
@@ -137,8 +137,17 @@ func EncryptAndWriteToFile(file *os.File, pk crypto.PrivateKey, password []byte)
...
-	return err
+	_, err = file.Write(data)
+	if err != nil {
+		return err
+	}
+
+	_, err = file.Write([]byte{'\n'})
+	if err != nil {
+		return err
+	}
```
EclesioMeloJunior, line 148: I prefer the way that uses only one `Write`.
```suggestion
	_, err = file.Write(append(data, byte('\n')))
	if err != nil {
	    return err
	}
```

lib/keystore/encrypt.go
```diff
@@ -137,8 +137,17 @@ func EncryptAndWriteToFile(file *os.File, pk crypto.PrivateKey, password []byte)
...
+
+	return file.Close()
```
EclesioMeloJunior, line 150: the `file.Close` needs to be placed into a defer statement since we should close a file whenever an error occurs

EclesioMeloJunior (changes requested):

dot/build_spec_test.go
```diff
@@ -5,19 +5,17 @@ package dot
...
 )
 
 func TestBuildFromGenesis(t *testing.T) {
-	file, err := genesis.CreateTestGenesisJSONFile(false)
-	defer os.Remove(file)
-	require.NoError(t, err)
+	file := genesis.CreateTestGenesisJSONFile(t, false)
 	bs, err := BuildFromGenesis(file, 0)
 
 	expectedChainType := "TESTCHAINTYPE"
```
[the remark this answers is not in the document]
qdm12: Yes but out of scope, so will skip here. I'm not really refactoring the entire test here :wink:

lib/keystore/encrypt.go
qdm12: So you're happy with the changes right? Or unhappy? Just making sure :smile:

lib/keystore/encrypt.go
qdm12: Totally. In the end I decided to change `EncryptAndWriteToFile` to handle both opening the file and closing it. It didn't make much sense to just do one or the other, and it was also easy to forget to close the file (we forgot in our unit tests). So now it's more of an 'atomic' operation where it does everything in the function. It doesn't matter from the current usage we make of it as well. I also further simplified using `os.WriteFile` to handle the open+write+close for us.

EclesioMeloJunior (approved):

kishansagathiya (approved):

dot/network/config_test.go
```diff
@@ -40,9 +38,7 @@ func TestBuildIdentity(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	if !reflect.DeepEqual(configA.privateKey, configB.privateKey) {
-		t.Error("Private keys should match")
-	}
+	assert.Equal(t, configA.privateKey, configB.privateKey)
```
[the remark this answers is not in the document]
qdm12: `assert.Equal` does it :wink:

PR #2143, opened 2022-01-05
chore(pkg/scale): return error if `Result` already has an assigned value 

### LINKED ISSUE #1962: `scale.Result` cannot allow more then one `Set` call
## Issue summary

- Currently, we allow a `scale.Result` instance calls the `Set` method for `ok` and `err` values multiple times which can cause problems when decoding into a real rust `Result` type.
- We should return an error for a `scale.Result` that has already been set

-

EclesioMeloJunior:
## Changes

- Add new error `ErrResultAlreadySet`
- Inside `Set` function check if `result.mode` is not `Unset` then return the error

## Tests

```
go test -run ^TestResult_UseSetMoreThanOnce$ ./pkg/scale/...
```

## Issues

- Closes #1962

## Primary Reviewer

- @edwardmack


pkg/scale/result.go
```diff
@@ -4,10 +4,13 @@
 package scale
 
 import (
+	"errors"
 	"fmt"
 	"reflect"
 )
 
+var ErrResultAlreadySetted = errors.New("result already has an assigned value")
```
qdm12, line 12: nit rename `ErrResultAlreadySetted` to `ErrResultAlreadySet` (past of set is set)

pkg/scale/result.go
```diff
@@ -46,6 +49,11 @@ func NewResult(okIn, errIn interface{}) (res Result) {
 
 // Set takes in a mode (OK/Err) and the associated interface and sets the Result value
 func (r *Result) Set(mode ResultMode, in interface{}) (err error) {
+	if r.mode != Unset {
+		err = ErrResultAlreadySetted
+		return
```
qdm12, line 54: nit

```suggestion
		return ErrResultAlreadySetted
```

pkg/scale/result_test.go
```diff
@@ -304,3 +306,14 @@ func TestResult_Set(t *testing.T) {
...
 }
+
+func TestResult_UseSetMoreThanOnce(t *testing.T) {
+	result := NewResult(int(0), string(""))
+
+	err := result.Set(OK, 10)
+	require.NoError(t, err)
+
+	err = result.Set(Err, "something failed")
+	require.Error(t, err)
```
qdm12, line 317: `require.ErrorIs` takes care of this already

```suggestion
```

qdm12 (commented):
What's the reason to prevent a subsequent `Set()` on a result?

EclesioMeloJunior:
@qdm12 there's a slack thread about this issue but basically, if the `result.Set` is called for the first time with mode `Ok` and the next time with mode `Err` the function doesn't replace the previous value. @timwu20 suggested to:

> I would probably err on the side of throwing an error for a result that has already been set.

qdm12 (approved):

jimjbrettj:
lgtm!

jimjbrettj (approved):

PR #2144, opened 2022-01-06
chore(lib/trie): add unit tests covering `lib/trie/trie.go`

qdm12:
## Changes

- Add unit tests
- Refactor existing tests

## Tests

```
go test -race ./lib/trie/...
```

## Issues

- #2085


lib/trie/trie_test.go
```diff
@@ -62,46 +79,734 @@ const (
...
+			key:  []byte{1},
+			trie: &Trie{},
+		},
+		"nil key returns root leaf": {
+			trie: &Trie{
+				root: &node.Leaf{
+					Key: []byte{1},
+				},
+			},
+			nextKey: []byte{1}, // TODO should this be nil?
```
noot, line 638: this is correct, as the leaf is the next node after nil

lib/trie/trie_test.go
```diff
@@ -62,46 +79,734 @@ const (
...
+		},
+		"nil key returns root branch": {
+			trie: &Trie{
+				root: &node.Branch{
+					Key: []byte{1},
+				},
+			},
+			// nextKey: []byte{1}, TODO it returns nil for now
```
noot, line 646: the trie in this case is an invalid trie - a branch with no children will always become a leaf. thus, I don't think this is a valid case to test, and probably why it's returning nil

lib/trie/trie_test.go
```diff
@@ -62,46 +79,734 @@ const (
...
+			key: []byte{0x01},
+		},
+		"key smaller than root branch full key": {
+			trie: &Trie{
+				root: &node.Branch{
+					Key: []byte{0x02},
+				},
+			},
+			key: []byte{0x01},
+			// nextKey: []byte{0x02}, TODO it returns nil for now
```
noot, line 671: I suspect this is because the node key is in nibbles form, while  the input to NextKey is ib bytes form, so you're comparing `nibble(2)` with `[nibble(0), nibble(1)]`, so the next key is nil

lib/trie/trie_test.go
```diff
@@ -62,46 +79,734 @@ const (
...
+						nil, nil,
+						&node.Leaf{
+							// full key: 0x012a
+							Key: []byte{0xa},
+						},
+					},
+				},
+			},
+			key:     []byte{0x01, 0x2a},
+			nextKey: []byte{0x01, 0x2a}, // TODO should this return nil?
```
noot, line 703: I think same issue here with the nibbles - node key is in nibble form

lib/trie/trie_test.go
```diff
@@ -62,46 +79,734 @@ const (
...
+						nil, nil,
+						&node.Leaf{
+							// full key: 0x012a
+							Key: []byte{0xa},
+						},
+					},
+				},
+			},
+			key:     []byte{0x01, 0x2b},
+			nextKey: []byte{0x01, 0x2a}, // TODO should this return nil?
```
noot, line 719: same here

noot: is this test the same as above?

lib/trie/trie_test.go
```diff
@@ -62,46 +79,734 @@ const (
...
+									Value: []byte("bottom leaf"),
+								},
+							},
+						},
+					},
+				},
+			},
+			key: []byte{0xa},
+			// TODO is this fine to ignore branches without a value?
+			nextKey: []byte{0xa, 0x3b, 0x3c},
```
noot, line 770: yes, `NextKey` should only return keys that have a value associated with them

lib/trie/trie_test.go
```diff
@@ -62,46 +79,734 @@ const (
...
+									Key:   []byte{0xc},
+									Value: []byte("bottom leaf"),
+								},
+							},
+						},
+					},
+				},
+			},
+			key:     []byte{0xa, 0x3b, 0x0},
+			nextKey: []byte{0xa, 0x3b}, // TODO should this return 0xa, 0x3b, 0x0c
```
noot, line 794: is it returning nil? probably same nibble issue

lib/trie/trie_test.go
qdm12: I think it's just because the root is a branch without a value so that's an invalid trie as well. Changing the branch to a leaf fixed it. Thanks!!

lib/trie/trie_test.go
qdm12: Hmm not sure here.

The input is `[0x01, 0x2a]` :arrow_right: (nibbles) `[0x01, 0x02, 0x0a]` which should hit the leaf which has full key in nibbles `[0x01, 0x02, 0x0a]` right?

noot: the input `[0x01, 0x2a]` would become `[0x00, 0x01, 0x02, 0x0a]` in nibbles I believe (correct me if I'm wrong), so the next key is correct here

qdm12: Ah yes thanks.

I needed key `[0x12, 0xa0]` which converts to `[1, 2, a, 0]` in nibbles for it to work. I really had a hard time getting I had to add trailing `0` :smile:

internal/trie/node/copy.go
```diff
@@ -4,17 +4,29 @@
...
+			}
+			cpy.Children[i] = child.Copy(copyChildren)
+		}
+	} else {
+		cpy.Children = b.Children // copy interface pointers only
+	}
+
 	if b.Key != nil {
 		cpy.Key = make([]byte, len(b.Key))
 		copy(cpy.Key, b.Key)
```
EclesioMeloJunior, line 32: this condition executes the same code already executed at line `16` unless by the allocation before copy, maybe remove the code at line `16`

internal/trie/node/dirty.go
```diff
@@ -5,20 +5,20 @@ package node
 
 // IsDirty returns the dirty status of the branch.
 func (b *Branch) IsDirty() bool {
-	return b.dirty
+	return b.Dirty
 }
 
 // SetDirty sets the dirty status to the branch.
 func (b *Branch) SetDirty(dirty bool) {
```
EclesioMeloJunior, line 12: should we keep maintaining this method once `Dirty` is an exported value?

EclesioMeloJunior (changes requested):

internal/trie/node/copy.go
qdm12: Not sure what you mean, there is no code on line 16 really :thinking: 
Also note this is an array so no need to allocate it if that's what you mean?

internal/trie/node/dirty.go
qdm12: Yes if you want to check on the node interface without doing type assertions on branch/leaf.

Also note I exported it to write tests in `lib/trie` otherwise it's really a pain to write test code. This is not really for production code purposes.

internal/trie/node/copy.go
EclesioMeloJunior: Some lines above (before the `if copyChildren`) there is a `copy(cpy.Key, b.Key)` and here we check `if b.Key != nil` then do the copy again

qdm12: Totally what was that, anyways; fixed!

EclesioMeloJunior (approved):

internal/trie/node/copy.go
```diff
@@ -4,16 +4,27 @@
 package node
 
 // Copy deep copies the branch.
-func (b *Branch) Copy() Node {
+// Setting copyChildren to true will deep copy
+// children as well.
+func (b *Branch) Copy(copyChildren bool) Node {
```
noot, line 9: which are we using right now? not sure what the purpose is of having both atm?

internal/trie/node/leaf.go
```diff
@@ -14,13 +14,14 @@ var _ Node = (*Leaf)(nil)
...
+	// Partial key bytes in nibbles (0 to f in hexadecimal)
+	Key   []byte
 	Value []byte
 	// Dirty is true when the branch differs
 	// from the node stored in the database.
-	dirty      bool
+	Dirty      bool
 	hashDigest []byte
-	encoding   []byte
+	Encoding   []byte
```
noot, line 24: why are these exported now? i couldn't really see why. same for branch

qdm12: For testing in the `lib/trie` package, to be able to assert/expect leaves/branches and set their encoding without weird wrappers. I have it locally but wanted to have this PR first to split the work since so the next test PR will be just test code.

internal/trie/node/copy.go
qdm12: It's to deep copy the entire trie for tests. `copyChildren` is always false for production code though.

noot: hmm, I generally try not to add things to the code that's only for testing. would it make sense to just make a separate `DeepCopy` implementation?

noot: oh wait, i see there is, nevermind. this is fine i guess haha

noot (approved):

internal/trie/node/copy.go
qdm12: No problem, I agree with your point as well. This is used in the trie `DeepCopy()` method which is only used by tests in this package and in other packages. There are some private members in the node implementation so we need to have these in the production code for other tests in other packages to be able to deep copy. Not a big fan of that dependency, but it will do for now :wink:

PR #2151, opened 2022-01-07
maintainence(lib/babe): refactor code to use epochHandler for control of epoch logic

### LINKED ISSUE #2097: devnet: BABE does not restart after pausing, and pauses at wrong times
## Task summary

- on the devnet, I've seen the following logs:
```
2021-12-01T15:42:57Z DBUG checking if epoch started with epoch start 2021-12-01 15:11:27 +0000 UTC and current time 2021-12-01 15:42:57.3872532 +0000 UTC m=+70304.868444174	babe.go:L457	pkg=bab
e2021-12-01T15:42:57Z DBUG pausing BABE, need to sync since we have 630 slots (start slot 546124459) into the epoch starting at 546123829	babe.go:L373	pkg=babe  
2021-12-01T15:43:00Z INFO 💤 node waiting, 2 peers connected, head block number 6723 with hash 0x2056dcdca68c41d53e84dbb2e0903fe6abf706b12b409728d04c01c7495a106f, finalised block number 6205 with hash 
```

firstly, BABE should not be pausing here (the epoch start is less thant he current time), secondly BABE should restart if it's synced. there is probably some timing issue happening here.

noot:
## Changes

- refactor babe epoch handling logic to be separated into an `epochHandler` module, which will handle the authorship slots for that epoch

## Tests

```
go test ./lib/babe
```

## Issues

- closes #2097

## Primary Reviewer

-  @EclesioMeloJunior


dot/peerset/peerset.go
```diff
@@ -340,8 +342,9 @@ func (ps *PeerSet) reportPeer(change ReputationChange, peers ...peer.ID) error {
 					return err
 				}
 
+				// TODO: change Noop to Drop once #2098 is fixed
```
noot, line 345: let me know if i should put commenting out this ban peer stuff into another PR

lib/babe/epoch_test.go
```diff
@@ -0,0 +1,201 @@
...
+
+	"github.com/ChainSafe/gossamer/dot/types"
+	"github.com/ChainSafe/gossamer/lib/crypto/sr25519"
+	"github.com/ChainSafe/gossamer/lib/keystore"
+
+	"github.com/golang/mock/gomock"
+	"github.com/stretchr/testify/require"
+)
+
+var keyring, _ = keystore.NewSr25519Keyring() //nolint:typecheck
```
noot, line 18: when I don't declare keyring here I get an undeclared error, but I then get a lint error since it's defined in `babe_integration_test.go`. how do i fix this?

lib/babe/babe.go
```diff
@@ -336,104 +301,85 @@ func (b *Service) initiate() {
...
 	epoch, err := b.epochState.GetCurrentEpoch()
 	if err != nil {
 		logger.Errorf("failed to get current epoch: %s", err)
 		return err
 	}
 
 	for {
-		err := b.initiateEpoch(epoch)
+		ctx, cancel := context.WithCancel(b.ctx)
+		defer cancel()
```
EclesioMeloJunior, line 347: I'm not sure, but can this lead to a stack or frame stack overflow? Since the defer is only executed when the function exits then if this`for` never ends the defer calls will be accumulated until the overflow?

EclesioMeloJunior (changes requested):

lib/babe/babe.go
EclesioMeloJunior: An alternative will be call cancel in every path that `continues` or `break` the for loop without `defer` statement or create the `ctx` inside `initiateAndGetEpochHandler` method

noot: good catch, this definitely seems sus. I'll update it!

noot: I made it a closure, let me know what you think

dot/peerset/peerset.go
timwu20: Can you not just uncomment this out for now? I wasn't able to reproduce the nodes banning at this point at least locally.

lib/babe/babe.go
```diff
@@ -336,77 +301,73 @@ func (b *Service) initiate() {
 		return
 	}
 
-	err := b.invokeBlockAuthoring()
+	// we should consider better error handling for this - we should
+	// retry to run the engine at some point (maybe the next epoch) if
+	// there's an error.
+	if err := b.runEngine(); err != nil {
+		logger.Criticalf("failed to run block production engine: %s", err)
```
timwu20, line 308: shouldn't the whole node shut down in this case?

lib/babe/babe.go
```diff
@@ -336,77 +301,73 @@ func (b *Service) initiate() {
...
-		logger.Infof("current epoch %d has %d slots", epoch, intoEpoch)
-
-		// get start slot for current epoch
-		nextEpochStart, err := b.epochState.GetStartSlotForEpoch(epoch + 1)
-		if err != nil {
-			logger.Errorf("failed to get start slot for next epoch %d: %s", epoch+1, err)
-			return err
-		}
+		err = func() error {
+			ctx, cancel := context.WithCancel(b.ctx)
```
timwu20, line 347: why do we have these nested contexts? Makes it pretty difficult to read.

lib/grandpa/grandpa.go
```diff
@@ -193,8 +193,13 @@ func (s *Service) Start() error {
...
 	go func() {
-		if err := s.initiate(); err != nil {
-			logger.Criticalf("failed to initiate: %s", err)
+		for {
+			// TODO: sometimes grandpa fails to initiate due to a "Key not found"
+			// error, this shouldn't happen.
+			if err := s.initiate(); err != nil {
+				logger.Criticalf("failed to initiate: %s", err)
+			}
+			time.Sleep(s.interval)
```
timwu20, line 202: sus..

lib/babe/babe.go
kishansagathiya: Not sure what is the problem here.

To me it makes sense to have a nested context to manage a new epoch handler. This way we can manage it independently. if `epochhandler.run` times out in one loop run, we can still move on to other one.

Other option is to use the same context `b.ctx`, but I think that approach has problems. Using `b.ctx.cancel` could impact other things.
( cancel should only be called by function that creates the context. You can pass it around, but it is not recommended since it is likely to create problems.)

Let me know if I might have misunderstood what you have written.

lib/babe/babe.go
```diff
@@ -336,77 +301,73 @@ func (b *Service) initiate() {
...
+				}
+			}
 
-		for i := 0; i < int(b.epochLength-intoEpoch); i++ {
-			done := false
+			errCh := make(chan error)
+			go b.epochHandler.run(errCh)
 
 			select {
 			case <-b.ctx.Done():
```
kishansagathiya, line 373: I feel like you would want to return for `runEngine` func here, don't that's what we are doing.

(May be we should start passing `ctx` to func as well. That would make the purpose of the `ctx` more obvious)

lib/babe/babe.go
noot: yeah, I agree with Kishan, the point is to have a context per epoch that is canceled when the epoch finished

lib/babe/babe.go
```diff
@@ -415,79 +376,33 @@ func (b *Service) invokeBlockAuthoring() error {
...
 		}()
-
 		if err != nil {
-			return 0, err
+			return err
 		}
 	}
 
-	return epochStart, nil
+	return nil
```
EclesioMeloJunior, line 402:
```suggestion
```

lib/babe/babe.go
```diff
@@ -415,79 +376,33 @@ func (b *Service) invokeBlockAuthoring() error {
...
-						slotNum, epoch, slotNum-epochStartSlot)
-					continue
-				} else if err != nil {
-					logger.Warnf("failed to handle slot %d: %s", slotNum, err)
-					continue
-				}
 			case <-epochTimer.C:
-				done = true
+				// stop current epoch handler
+				cancel()
```
EclesioMeloJunior, line 381: when the `epochTimer.C` gets triggered should we just let the code flow until the next return statement to call the `defer cancel()` function?
```suggestion
```

EclesioMeloJunior (changes requested):

lib/babe/babe.go
noot: probably, so should i panic here?

lib/babe/babe.go
```diff
@@ -105,30 +108,22 @@ func NewService(cfg *ServiceConfig) (*Service, error) {
...
+		telemetry: cfg.Telemetry,
 	}
 
 	logger.Debugf(
-		"created service with epoch %d, block producer=%t, slot duration %s, epoch length (slots) %d, authorities %v, "+
-			"authority index %d, threshold %v and randomness %s",
-		epoch, cfg.Authority, babeService.slotDuration, babeService.epochLength,
-		Authorities(babeService.epochData.authorities), babeService.epochData.authorityIndex,
-		babeService.epochData.threshold, babeService.epochData.randomness)
+		"created service with block producer ID=%t, slot duration %s, epoch length (slots) %d",
```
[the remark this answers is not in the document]
noot: seems sus, will update

lib/babe/babe.go
```diff
@@ -229,12 +194,12 @@ func (b *Service) waitForFirstBlock() error {
 
 // SlotDuration returns the current service slot duration in milliseconds
 func (b *Service) SlotDuration() uint64 {
-	return uint64(b.slotDuration.Milliseconds())
+	return uint64(b.constants.slotDuration.Milliseconds())
```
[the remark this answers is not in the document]
noot: lol, i just like having them specified as constants, it's easier to pass around

lib/babe/babe.go
noot: what do you mean? I never want to return from `runEngine` unless the node is shut down

noot: I have to initialize the ctx in the function so that the cancel is called when the function returns

lib/babe/babe.go
noot: no, because we want to cancel the current epoch, then do the `incrementEpoch` call

lib/babe/epoch_test.go
noot: nvm, i just gotta put it in a non integration test file

dot/types/authority.go
```diff
@@ -93,6 +94,10 @@ type AuthorityRaw struct {
 	Weight uint64
 }
 
+func (a *AuthorityRaw) String() string {
+	return fmt.Sprintf("Key=0x%x Weight=%d", a.Key, a.Weight)
```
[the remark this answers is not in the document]
noot: done

lib/babe/babe.go
```diff
@@ -95,6 +88,16 @@ func NewService(cfg *ServiceConfig) (*Service, error) {
 
 	logger.Patch(log.SetLevel(cfg.LogLvl))
 
+	slotDuration, err := cfg.EpochState.GetSlotDuration()
+	if err != nil {
+		return nil, err
```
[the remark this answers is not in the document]
noot: done

lib/babe/babe.go
```diff
@@ -95,6 +88,16 @@ func NewService(cfg *ServiceConfig) (*Service, error) {
...
+	}
+
+	epochLength, err := cfg.EpochState.GetEpochLength()
+	if err != nil {
+		return nil, err
```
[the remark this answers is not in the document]
noot: done

lib/babe/babe.go
```diff
@@ -194,6 +150,15 @@ func (b *Service) Start() error {
 }
 
 func (b *Service) waitForFirstBlock() error {
+	head, err := b.blockState.BestBlockHeader()
+	if err != nil {
+		return err
```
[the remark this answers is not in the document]
noot: done

lib/babe/babe.go
```diff
@@ -297,7 +262,7 @@ func (b *Service) Stop() error {
 
 // Authorities returns the current BABE authorities
 func (b *Service) Authorities() []types.Authority {
-	return b.epochData.authorities
+	return b.epochHandler.epochData.authorities
```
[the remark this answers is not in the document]
noot: done

lib/babe/babe.go
```diff
@@ -336,104 +301,85 @@ func (b *Service) initiate() {
 		return
 	}
 
-	err := b.invokeBlockAuthoring()
+	// we should consider better error handling for this - we should
+	// retry to run the engine at some point (maybe the next epoch) if
+	// there's an error.
+	if err := b.runEngine(); err != nil {
+		logger.Criticalf("failed to run block production engine: %s", err)
```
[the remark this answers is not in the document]
noot: we should probably panic (or shutdown somehow), but if i panic here the integration tests fail due to the database closing, which returns an error in `runEngine` causing everything to panic

lib/babe/babe.go
```diff
@@ -336,104 +301,85 @@ func (b *Service) initiate() {
...
+	}
+}
+
+func (b *Service) initiateAndGetEpochHandler(ctx context.Context, epoch uint64) (*epochHandler, error) {
+	epochData, err := b.initiateEpoch(epoch)
 	if err != nil {
-		logger.Criticalf("block authoring error: %s", err)
+		logger.Errorf("failed to initiate epoch %d: %s", epoch, err)
+		return nil, err
```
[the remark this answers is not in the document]
noot: done

lib/babe/babe.go
noot: done

lib/babe/babe.go
kishansagathiya: > I never want to return from runEngine unless the node is shut down

`b.ctx` is context of the service. If `b.ctx` is done, that should mean we don't need the use of service anymore, hence all of its tasks should stop, which does not happen. Is there any other meaning of `b.ctx` here?

But, even if there is any other meaning of `b.ctx`, once `b.ctx` is Done, in the next iteration we end up creating a nested context from a context that is already finished (which is possible, I don't know why). And then without waiting you will do cleanup again because `b.ctx` is already done.
So, it basically becomes an infinite loop of cleanup.

noot: ah ok I understand, I've updated the inner function to return an error if `b.ctx` is cancelled, which will then cause the function to return.

lib/babe/babe.go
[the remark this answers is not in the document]
qdm12: you could recover from the panic in your integration tests. Quite ugly though.

lib/babe/babe.go
timwu20: I'm of the proponent where we should not be using `context.WithCancel()` within package boundaries.  I know we already do this in most of the `Service` implementations.  If what we're looking for is cancellation, then we can just create a channel for that purpose.  `context.Context` is meant to traverse through API boundaries and contain request scoped data.  I've seen a lot of examples of go code using them to handle long running processes and I feel like it introduces more weight than required.

In this case the `Service.ctx` is being wrapped to add cancellation to the current `epochHandler`.  So to my knowledge that the `Service.ctx` was created with `context.WithCancel`, then `ctx.Done()` would close if the original `CancelFunc` was called.  But we wrap this context again in line 347.  So does that mean that if we call the local `cancel CancelFunc` than any goroutines listening on `Service.ctx` would receive the channel close? I would assume no, but these are things that make the code more difficult to understand.  I think as a best practice you shouldn't call `context.WithCancel` or any of the other `context.WithXXX` functions on supplied contexts.

I think `epochHandler.run()` should be modified to accept a `context.Context` that's not the original one stored in  `Service`.  This `runEngine` function should listen on the `Service` one and trigger cancellation within this function for the current `epochHandler`.

lib/babe/babe.go
timwu20: I'd rather not panic.  Can we not signal the node to shutdown? This would require us to refactor the `Service` interface to return an `err` through an `errChan`, and the node could then shutdown knowing that `babe.Service` had failed.

lib/babe/epoch_handler.go
```diff
@@ -0,0 +1,164 @@
...
+
+var (
+	errEpochPast = errors.New("cannot run epoch that has already passed")
+)
+
+type epochHandler struct {
+	ctx context.Context
+
+	epochNumber uint64
+	startTime   time.Time
```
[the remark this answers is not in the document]
noot: you're right will remove!

lib/babe/babe_integration_test.go
```diff
@@ -38,7 +38,7 @@ var (
 	emptyHash         = trie.EmptyHash
 	testEpochIndex    = uint64(0)
 
-	keyring, _ = keystore.NewSr25519Keyring()
+	keyring, _ = keystore.NewSr25519Keyring() //nolint:typecheck
```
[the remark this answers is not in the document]
noot: removed

lib/babe/build_integration_test.go
```diff
@@ -211,15 +190,13 @@ func TestApplyExtrinsic(t *testing.T) {
...
-	babeService.epochData.threshold = maxThreshold
 
 	builder, _ := NewBlockBuilder(
 		babeService.keypair,
 		babeService.transactionState,
 		babeService.blockState,
-		babeService.slotToProof,
-		babeService.epochData.authorityIndex,
+		&VrfOutputAndProof{},
+		0,
```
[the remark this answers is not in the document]
noot: done

lib/babe/build.go
```diff
@@ -196,13 +195,8 @@ func (b *BlockBuilder) buildBlockSeal(header *types.Header) (*types.SealDigest,
...
-	data, err := b.buildBlockBABEPrimaryPreDigest(slot)
-	if err != nil {
-		return nil, err
-	}
-
-	err = babeHeader.Set(*data)
-	if err != nil {
+	data := b.buildBlockBABEPrimaryPreDigest(slot)
+	if err := babeHeader.Set(*data); err != nil {
 		return nil, err
```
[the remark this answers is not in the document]
noot: done

lib/babe/build.go
```diff
@@ -60,29 +59,29 @@ type BlockBuilder struct {
...
 	if ts == nil {
 		return nil, errors.New("cannot create block builder; transaction state is nil")
 	}
 	if bs == nil {
 		return nil, errors.New("cannot create block builder; block state is nil")
 	}
-	if sp == nil {
-		return nil, errors.New("cannot create block builder; slot to proff is nil")
+	if proof == nil {
+		return nil, errors.New("cannot create block builder; slot VRF proof is nil")
```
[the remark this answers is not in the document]
noot: done

kishansagathiya (approved):

lib/babe/babe.go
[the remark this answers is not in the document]
noot: sorry i missed this! yeah like panicking in the code that starts the node and all the services instead?

lib/babe/babe.go
```diff
@@ -322,12 +289,12 @@ func (b *Service) getAuthorityIndex(Authorities []types.Authority) (uint32, erro
 }
 
 func (b *Service) getSlotDuration() time.Duration {
```
EclesioMeloJunior, line 291: I noticed this function is not being unused, could you remove it?

lib/babe/babe.go
```diff
@@ -336,158 +303,105 @@ func (b *Service) initiate() {
...
+			}
 
-		for i := 0; i < int(b.epochLength-intoEpoch); i++ {
-			done := false
+			errCh := make(chan error)
+			go b.epochHandler.run(ctx, errCh)
 
 			select {
 			case <-b.ctx.Done():
+				cancel()
```
EclesioMeloJunior, line 371: as we return we don't need to call the `cancel()` already defined in the `defer` statement
```suggestion
				cancel()
```

lib/babe/epoch_handler.go
```diff
@@ -0,0 +1,156 @@
...
+			epochNumber,
+			epochData.threshold,
+			keypair,
+		)
+		if err != nil {
+			if errors.Is(err, errOverPrimarySlotThreshold) {
+				continue
+			}
+			return nil, fmt.Errorf("error running slot lottery at slot %d: error %w", i, err)
+		}
```
EclesioMeloJunior, line 65:
```suggestion
		if errors.Is(err, errOverPrimarySlotThreshold) {
				continue
	         } else if err != nil {
			return nil, fmt.Errorf("error running slot lottery at slot %d: error %w", i, err)
		}
```

EclesioMeloJunior (approved):

EclesioMeloJunior (changes requested):
Just small changes!

lib/babe/epoch_handler.go
qdm12: nit I would also suggest to remove the `error ` prefix before `%w`

lib/babe/epoch_handler_test.go
```diff
@@ -0,0 +1,92 @@
...
+
+	eh, err := newEpochHandler(context.Background(), 1, 9999, epochData, constants, testHandleSlotFunc, keypair)
+	require.NoError(t, err)
+	require.Equal(t, 200, len(eh.slotToProof))
+	require.Equal(t, uint64(1), eh.epochNumber)
+	require.Equal(t, getSlotStartTime(9999, sd), eh.startTime)
+	require.Equal(t, uint64(9999), eh.firstSlot)
+	require.Equal(t, constants, eh.constants)
+	require.Equal(t, epochData, eh.epochData)
+	require.NotNil(t, eh.handleSlot)
```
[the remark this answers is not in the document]
qdm12: Can't we assign it some predictable value in the test? Or perhaps dependency inject the source of randomness? 🤔

lib/babe/babe.go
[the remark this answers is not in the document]
qdm12: Yeah or return an error that makes the whole program stop gracefully ideally.

lib/babe/epoch_handler_test.go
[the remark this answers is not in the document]
noot: hmm it would probably require overriding the `claimPrimarySlot` function which is used to generate the proof... I can do it if you'd really like but seems like that sorta veers into territory of changing the code for tests

lib/babe/babe.go
qdm12: > I'm of the proponent where we should not be using context.Context within package boundaries

The problem is some of the std lib uses context like `http.NewRequestWithContext` so how do you do it then? There is no other way really to cancel an http request otherwise.

Also contexts are used for a few small cases whereas channels can be used for various things. In my opinion it's always clearer to use contexts to cancel long running operations. Channels are useful to return readiness or errors from long running operations though.

lib/babe/babe.go
noot: updated

lib/babe/babe.go
noot: done

lib/babe/babe.go
```diff
@@ -105,30 +108,22 @@ func NewService(cfg *ServiceConfig) (*Service, error) {
...
-	}
-
-	epoch, err := cfg.EpochState.GetCurrentEpoch()
-	if err != nil {
-		return nil, err
-	}
-
-	if err = babeService.setupParameters(cfg); err != nil {
-		return nil, err
+		constants: &constants{
```
qdm12, line 116: Why is a pointer by the way?

lib/babe/epoch_handler.go
noot: updated

lib/babe/babe.go
```diff
@@ -297,7 +262,9 @@ func (b *Service) Stop() error {
 
 // Authorities returns the current BABE authorities
 func (b *Service) Authorities() []types.Authority {
-	return b.epochData.authorities
+	auths := make([]types.Authority, len(b.epochHandler.epochData.authorities))
+	copy(auths, b.epochHandler.epochData.authorities)
+	return auths
```
qdm12, line 267: I don't think you can copy `types.Authority` directly, you would need a deep copy method I guess?

lib/babe/babe.go
qdm12: This 💯

lib/babe/babe.go
```diff
@@ -336,158 +303,105 @@ func (b *Service) initiate() {
...
+	// retry to run the engine at some point (maybe the next epoch) if
+	// there's an error.
+	if err := b.runEngine(); err != nil {
+		logger.Criticalf("failed to run block production engine: %s", err)
 	}
 }
 
-func (b *Service) invokeBlockAuthoring() error {
-	epoch, err := b.epochState.GetCurrentEpoch()
+func (b *Service) initiateAndGetEpochHandler(epoch uint64) (*epochHandler, error) {
```
qdm12, line 310: Please use named returns 😃 

```suggestion
func (b *Service) initiateAndGetEpochHandler(epoch uint64) (epochHandler *epochHandler, err error) {
```

lib/babe/babe.go
```diff
@@ -336,158 +303,105 @@ func (b *Service) initiate() {
...
+	epochData, err := b.initiateEpoch(epoch)
 	if err != nil {
-		logger.Errorf("failed to get current epoch: %s", err)
-		return err
+		return nil, fmt.Errorf("failed to initiate epoch %d: %s", epoch, err)
```
qdm12, line 313:
```suggestion
		return nil, fmt.Errorf("failed to initiate epoch %d: %w", epoch, err)
```

lib/babe/babe.go
```diff
@@ -336,158 +303,105 @@ func (b *Service) initiate() {
...
-		// resume it when ready
-		if b.epochLength <= intoEpoch && !b.dev {
-			logger.Debugf(
-				"pausing BABE, need to sync since we have %d slots (start slot %d) into the epoch starting at %d",
-				intoEpoch, startSlot, epochStartSlot)
-			return b.Pause()
-		}
+	epochStartSlot, err := b.epochState.GetStartSlotForEpoch(epoch)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get start slot for current epoch %d: %s", epoch, err)
```
qdm12, line 321:
```suggestion
		return nil, fmt.Errorf("failed to get start slot for current epoch %d: %w", epoch, err)
```

lib/babe/babe.go
```diff
@@ -336,158 +303,105 @@ func (b *Service) initiate() {
...
+		b.handleSlot,
+		b.keypair,
+	)
+}
 
-		logger.Infof("current epoch %d has %d slots", epoch, intoEpoch)
+func (b *Service) runEngine() error {
+	epoch, err := b.epochState.GetCurrentEpoch()
+	if err != nil {
+		return fmt.Errorf("failed to get current epoch: %s", err)
```
qdm12, line 336:
```suggestion
		return fmt.Errorf("failed to get current epoch: %w", err)
```

lib/babe/babe.go
```diff
@@ -336,158 +303,105 @@ func (b *Service) initiate() {
...
+	}
 
-		// get start slot for current epoch
-		nextEpochStart, err := b.epochState.GetStartSlotForEpoch(epoch + 1)
-		if err != nil {
-			logger.Errorf("failed to get start slot for next epoch %d: %s", epoch+1, err)
-			return err
-		}
+	for {
+		err = func() error {
```
qdm12, line 340: can we have this one as a separate global scope function?

lib/babe/babe.go
```diff
@@ -336,158 +303,105 @@ func (b *Service) initiate() {
...
+			ctx, cancel := context.WithCancel(context.Background())
+			defer cancel()
+			b.epochHandler, err = b.initiateAndGetEpochHandler(epoch)
+			if err != nil {
+				return err
```
qdm12, line 345:
```suggestion
				return fmt.Errorf("cannot initiate epoch handler: %w", err)
```

lib/babe/babe.go
```diff
@@ -336,158 +303,105 @@ func (b *Service) initiate() {
...
 
-		nextEpochStartTime := getSlotStartTime(nextEpochStart, b.slotDuration)
-		epochTimer := time.NewTimer(time.Until(nextEpochStartTime))
-		cleanup := func() {
-			if !epochTimer.Stop() {
-				<-epochTimer.C
+			// get start slot for current epoch
+			nextEpochStart, err := b.epochState.GetStartSlotForEpoch(epoch + 1)
+			if err != nil {
+				return fmt.Errorf("failed to get start slot for next epoch %d: %s", epoch+1, err)
```
qdm12, line 351:
```suggestion
				return fmt.Errorf("failed to get start slot for next epoch %d: %w", epoch+1, err)
```

qdm12 (dismissed):

lib/grandpa/grandpa.go
```diff
@@ -193,8 +193,13 @@ func (s *Service) Start() error {
 	s.tracker.start()
 
 	go func() {
-		if err := s.initiate(); err != nil {
-			logger.Criticalf("failed to initiate: %s", err)
+		for {
+			// TODO: sometimes grandpa fails to initiate due to a "Key not found"
+			// error, this shouldn't happen.
+			if err := s.initiate(); err != nil {
+				logger.Criticalf("failed to initiate: %s", err)
```
qdm12, line 200: Panic? Error channel? 🤔

lib/babe/verify_integration_test.go
```diff
@@ -373,32 +409,34 @@ func TestVerifyAuthorshipRight_Equivocation(t *testing.T) {
...
 	babeService := createTestService(t, cfg)
-	babeService.epochData.threshold = maxThreshold
+	epochData, err := babeService.initiateEpoch(testEpochIndex)
+	require.NoError(t, err)
 
-	babeService.epochData.authorities = make([]types.Authority, 1)
-	babeService.epochData.authorities[0] = types.Authority{
+	epochData.threshold = maxThreshold
+	epochData.authorities = make([]types.Authority, 1)
+	epochData.authorities[0] = types.Authority{
```
qdm12, line 417:
```suggestion
	epochData.authorities = []types.Authority{
```

lib/babe/verify_integration_test.go
```diff
@@ -298,22 +329,26 @@ func TestVerifyPimarySlotWinner(t *testing.T) {
...
+	epochData, err := babeService.initiateEpoch(0)
+	require.NoError(t, err)
 
 	// create proof that we can authorize this block
-	babeService.epochData.threshold = maxThreshold
-	babeService.epochData.authorityIndex = 0
+	epochData.threshold = maxThreshold
+	epochData.authorityIndex = 0
+
+	var slotNumber uint64 = 1
```
qdm12, line 339:
```suggestion
	const slotNumber = 1
```

lib/babe/verify_integration_test.go
```diff
@@ -214,14 +230,18 @@ func TestVerificationManager_VerifyBlock_Secondary(t *testing.T) {
 }
 
 func TestVerificationManager_VerifyBlock_MultipleEpochs(t *testing.T) {
+	t.Skip() // no idea why it's complaining it can't find the epoch data. fix later
```
qdm12, line 233: TODO 😉

lib/babe/verify_integration_test.go
```diff
@@ -35,19 +36,21 @@ func newTestVerificationManager(t *testing.T, genCfg *types.BabeConfiguration) *
...
 	err := dbSrv.Initialise(gen, genHeader, genTrie)
 	require.NoError(t, err)
 
 	err = dbSrv.Start()
 	require.NoError(t, err)
+
+	if genCfg == nil {
+		genCfg = genesisBABEConfig
+	}
+
```
qdm12, line 53: why is this moved?

lib/babe/epoch_test.go
```diff
@@ -0,0 +1,201 @@
...
+
+	mockEpochState2.EXPECT().GetLatestConfigData().Return(testLatestConfigData, nil)
+
+	testEpochDataEpoch0 := &types.EpochData{
+		Randomness:  [32]byte{9},
+		Authorities: []types.Authority{*authority},
+	}
+
+	mockEpochState0.EXPECT().GetLatestEpochData().Return(testEpochDataEpoch0, nil)
+	mockEpochState0.EXPECT().GetLatestConfigData().Return(testConfigData, nil)
```
qdm12, line 121: Perhaps you could configure your mocks according to test cases?

Like https://github.com/qdm12/gluetun/blob/55e609cbf43c76570c7831c5122ab121fd5c2382/internal/wireguard/rule_test.go#L84 for example.

There is also a reason where I believe having `NewController(t)` with the parent test might give some funny logs/behavior especially when running parallel subtests inside it.

lib/babe/babe.go
noot: i don't know lol. want me to change it?

lib/babe/epoch_handler_test.go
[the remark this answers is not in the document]
qdm12: As in if it's a lot of code change, don't; otherwise I'd say yes your code should be changed to be more testable 😉

lib/babe/babe.go
qdm12: Yes pleaaaase

lib/babe/babe.go
noot: why here? the variable name is the same as the type name, not sure if it adds anything

lib/grandpa/grandpa.go
noot: I think this is the same sort of issue as with BABE, we probably need to add an error channel so the higher-level service handler can either panic or restart the service. I'll make an issue for it

noot: #2235

lib/babe/epoch.go
```diff
@@ -6,137 +6,182 @@ package babe
...
+		return ed, startSlot, nil
+	}
 
-			threshold, err := CalculateThreshold(cfgData.C1, cfgData.C2, len(data.Authorities))
-			if err != nil {
-				return err
-			}
+	has, err := b.epochState.HasEpochData(epoch)
+	if err != nil {
+		return nil, 0, err
```
qdm12, line 95:
```suggestion
		return nil, 0, fmt.Errorf("cannot check for epoch data for epoch %d: %w", epoch, err)
```

lib/babe/babe.go
noot: done!

lib/babe/babe.go
noot: done

lib/babe/babe.go
noot: done (also fixed all the other error returns)

lib/babe/babe.go
noot: done!

lib/babe/epoch.go
```diff
@@ -6,137 +6,182 @@ package babe
...
-			b.epochData = &epochData{
-				randomness:     data.Randomness,
-				authorities:    data.Authorities,
-				authorityIndex: idx,
-				threshold:      b.epochData.threshold, // TODO: threshold might change if authority count changes
-			}
-		}
+	if !has {
+		logger.Criticalf("%s number=%d", errNoEpochData, epoch)
+		return nil, 0, errNoEpochData
```
qdm12, line 100:
```suggestion
		return nil, 0, fmt.Errorf("%w: for epoch %d", errNoEpochData, epoch)
```

lib/babe/verify_integration_test.go
noot: done

lib/babe/verify_integration_test.go
noot: done

lib/babe/epoch.go
```diff
@@ -6,137 +6,182 @@ package babe
...
+	}
 
-		startSlot, err = b.epochState.GetStartSlotForEpoch(epoch)
-		if err != nil {
-			return err
-		}
+	data, err := b.epochState.GetEpochData(epoch)
+	if err != nil {
+		return nil, 0, err
```
qdm12, line 105:
```suggestion
		return nil, 0, fmt.Errorf("cannot get epoch data for epoch %d: %w", epoch, err)
```

lib/babe/verify_integration_test.go
noot: if this isn't here the verifier tests panic :p

lib/babe/epoch.go
```diff
@@ -6,137 +6,182 @@ package babe
...
-	// by checking when we will be able to produce block 1.
-	// note that this assumes there will only be one producer of block 1
-	if b.blockState.BestBlockHash() == b.blockState.GenesisHash() {
-		startSlot, err = b.getFirstSlot(epoch)
-		if err != nil {
-			return err
-		}
+	idx, err := b.getAuthorityIndex(data.Authorities)
+	if err != nil {
+		return nil, 0, err
```
qdm12, line 110:
```suggestion
		return nil, 0, fmt.Errorf("cannot get authority index: %w", err)
```

lib/babe/epoch.go
```diff
@@ -6,137 +6,182 @@ package babe
...
-			_, err := b.runLottery(i, epoch)
-			if err != nil {
-				if errors.Is(err, errOverPrimarySlotThreshold) {
-					continue
-				}
-				return fmt.Errorf("error running slot lottery at slot %d: error %w", i, err)
-			}
+	has, err = b.epochState.HasConfigData(epoch)
+	if err != nil {
+		return nil, 0, err
```
qdm12, line 115:
```suggestion
		return nil, 0, fmt.Errorf("cannot check for config data for epoch %d: %w", epoch, err)
```

lib/babe/epoch_test.go
noot: yeah that would be nice, I'll work on that. gotta work on making my tests cleaner haha

lib/babe/epoch.go
```diff
@@ -6,137 +6,182 @@ package babe
...
+	}
 
-			startSlot = i
-			break
+	var cfgData *types.ConfigData
+	if has {
+		cfgData, err = b.epochState.GetConfigData(epoch)
+		if err != nil {
+			return nil, 0, err
```
qdm12, line 122:
```suggestion
			return nil, 0, fmt.Errorf("cannot get config data for epoch %d: %w", epoch, err)
```

lib/babe/epoch.go
```diff
@@ -6,137 +6,182 @@ package babe
...
 		}
-
-		// we are at genesis, set first slot by checking at which slot we will be able to produce block 1
-		err = b.epochState.SetFirstSlot(startSlot)
+	} else {
+		cfgData, err = b.epochState.GetLatestConfigData()
 		if err != nil {
-			return err
+			return nil, 0, err
```
qdm12, line 127:
```suggestion
			return nil, 0, fmt.Errorf("cannot get latest config data from epoch state: %w", err)
```

lib/babe/epoch.go
```diff
@@ -6,137 +6,182 @@ package babe
...
 		}
 	}
 
-	logger.Infof("initiating epoch %d with start slot %d", epoch, startSlot)
+	threshold, err := CalculateThreshold(cfgData.C1, cfgData.C2, len(data.Authorities))
+	if err != nil {
+		return nil, 0, err
```
qdm12, line 133:
```suggestion
		return nil, 0, fmt.Errorf("cannot calculate threshold: %w", err)
```

lib/babe/epoch.go
```diff
@@ -6,137 +6,182 @@ package babe
...
-		proof, err := b.runLottery(i, epoch)
-		if err != nil {
-			if errors.Is(err, errOverPrimarySlotThreshold) {
-				continue
-			}
-			return fmt.Errorf("error running slot lottery at slot %d: error %w", i, err)
-		}
+	startSlot, err := b.epochState.GetStartSlotForEpoch(epoch)
+	if err != nil {
+		return nil, 0, err
```
qdm12, line 145:
```suggestion
		return nil, 0, fmt.Errorf("cannot get start slot for epoch %d: %w", epoch, err)
```

lib/babe/epoch.go
```diff
@@ -6,137 +6,182 @@ package babe
...
+	}
+
+	return ed, startSlot, nil
+}
 
-		b.slotToProof[i] = proof
-		logger.Tracef("claimed slot %d, there are now %d slots into epoch", startSlot, i-startSlot)
+func (b *Service) getLatestEpochData() (*epochData, error) {
+	var err error
+	resEpochData := &epochData{}
```
qdm12, line 153: nit this would be nicer🙏 

```suggestion
func (b *Service) getLatestEpochData() (resEpochData *epochData, err error) {
	resEpochData = &epochData{}
```

lib/babe/epoch.go
```diff
@@ -6,137 +6,182 @@ package babe
...
+
+	epochData, err := b.epochState.GetLatestEpochData()
+	if err != nil {
+		return nil, err
```
qdm12, line 157:
```suggestion
		return nil, fmt.Errorf("cannot get latest epoch data: %w", err)
```

lib/babe/epoch.go
```diff
@@ -6,137 +6,182 @@ package babe
...
 	}
 
-	return nil
+	resEpochData.randomness = epochData.Randomness
+	resEpochData.authorities = epochData.Authorities
+
+	configData, err := b.epochState.GetLatestConfigData()
+	if err != nil {
+		return nil, err
```
qdm12, line 165:
```suggestion
		return nil, fmt.Errorf("cannot get epoch state latest config data: %w", err)
```

lib/babe/epoch.go
```diff
@@ -6,137 +6,182 @@ package babe
...
+	}
+
+	resEpochData.threshold, err = CalculateThreshold(configData.C1, configData.C2, len(resEpochData.authorities))
+	if err != nil {
+		return nil, err
```
qdm12, line 170:
```suggestion
		return nil, fmt.Errorf("cannot calculate threshold: %w", err)
```

lib/babe/epoch.go
```diff
@@ -6,137 +6,182 @@ package babe
...
+	}
+
+	if !b.authority {
+		return resEpochData, nil
+	}
+
+	resEpochData.authorityIndex, err = b.getAuthorityIndex(resEpochData.authorities)
+	return resEpochData, err
```
qdm12, line 178:
```suggestion
	resEpochData.authorityIndex, err = b.getAuthorityIndex(resEpochData.authorities)
	if err != nil {
		return nil, fmt.Errorf("cannot get authority index: %w", err)
	}
	
	return resEpochData, nil
```

lib/babe/epoch.go
```diff
@@ -6,137 +6,182 @@ package babe
...
 }
 
-func (b *Service) getFirstSlot(epoch uint64) (uint64, error) {
-	startSlot := getCurrentSlot(b.slotDuration)
-	for i := startSlot; i < startSlot+b.epochLength; i++ {
-		_, err := b.runLottery(i, epoch)
+func (b *Service) getFirstAuthoringSlot(epoch uint64, epochData *epochData) (uint64, error) {
```
qdm12, line 181: nit Why do we pass the epoch data by pointer? It feels it would be simpler by value 🤔

lib/babe/epoch_handler.go
```diff
@@ -0,0 +1,155 @@
...
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// The gossamer library is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with the gossamer library. If not, see <http://www.gnu.org/licenses/>.
```
qdm12, line 15: remove and run `make license` to have it shorter 😉

lib/babe/epoch_handler.go
```diff
@@ -0,0 +1,155 @@
...
+	// for slots where we are a producer, store the vrf output (bytes 0-32) + proof (bytes 32-96)
+	slotToProof map[uint64]*VrfOutputAndProof
+
+	handleSlot handleSlotFunc
+}
+
+func newEpochHandler(epochNumber, firstSlot uint64, epochData *epochData, constants *constants,
+	handleSlot handleSlotFunc, keypair *sr25519.Keypair) (*epochHandler, error) {
+	// determine which slots we'll be authoring in by pre-calculating VRF output
+	slotToProof := make(map[uint64]*VrfOutputAndProof)
```
qdm12, line 51: nit preallocating

```suggestion
	slotToProof := make(map[uint64]*VrfOutputAndProof, constants.epochLength)
```

lib/babe/epoch_handler.go
```diff
@@ -0,0 +1,155 @@
...
+		errCh <- errEpochPast
+		return
+	}
+
+	// for each slot we're handling, create a timer that will fire when it starts
+	// we create timers only for slots where we're authoring
+	authoringSlots := getAuthoringSlots(h.slotToProof)
+
+	type slotWithTimer struct {
+		timer   <-chan time.Time
```
qdm12, line 98: Use `*time.Timer` instead so you can stop them when the context gets canceled. You cannot cancel `time.After` and it will trigger even if the function has exit.

lib/babe/epoch_handler.go
```diff
@@ -0,0 +1,155 @@
...
+		slotNum uint64
+	}
+
+	slotTimeTimers := []*slotWithTimer{}
```
qdm12, line 102: nit

```suggestion
	slotTimeTimers := make([]*slotWithTimer, 0, len(authoringSlots))
```

lib/babe/epoch_handler.go
```diff
@@ -0,0 +1,155 @@
...
+	for _, authoringSlot := range authoringSlots {
+		// ignore slots already passed
+		if authoringSlot < currSlot {
+			continue
```
qdm12, line 106: nit this might be pure preference, but I feel like comment about an if block should at the start of the block or next to the if condition (on the right side). Otherwise it feels like you comment what the inner for loop does.

```suggestion
		if authoringSlot < currSlot {
			// ignore slots already passed
			continue
```

lib/babe/epoch_handler.go
```diff
@@ -0,0 +1,155 @@
...
+		logger.Debugf("waiting for next authoring slot %d", swt.slotNum)
+
+		select {
+		case <-ctx.Done():
+			return
+		case <-swt.timer:
+			if _, has := h.slotToProof[swt.slotNum]; !has {
+				// this should never happen
+				logger.Errorf("no VRF proof for authoring slot! slot=%d", swt.slotNum)
+				continue
```
qdm12, line 129: Perhaps panic instead if this is never meant to happen? Or remove?

lib/babe/epoch_handler.go
```diff
@@ -0,0 +1,155 @@
...
+			}
+
+			if err := h.handleSlot(h.epochNumber, swt.slotNum, h.epochData.authorityIndex, h.slotToProof[swt.slotNum]); err != nil { //nolint:lll
+				logger.Warnf("failed to handle slot %d: %s", swt.slotNum, err)
+				continue
+			}
```
qdm12, line 135: nit don't use the short err check form with a big line like this.

```suggestion
			err := h.handleSlot(h.epochNumber, swt.slotNum, 
				h.epochData.authorityIndex, h.slotToProof[swt.slotNum])
			if err != nil {
				logger.Warnf("failed to handle slot %d: %s", swt.slotNum, err)
				continue
			}
```

lib/babe/epoch_handler.go
```diff
@@ -0,0 +1,155 @@
...
+}
+
+// getAuthoringSlots returns an ordered slice of slot numbers where we can author blocks,
+// based on the given VRF output and proof map.
+func getAuthoringSlots(slotToProof map[uint64]*VrfOutputAndProof) []uint64 {
+	authoringSlots := make([]uint64, len(slotToProof))
+	i := 0
+	for authoringSlot := range slotToProof {
+		authoringSlots[i] = authoringSlot
+		i++
```
qdm12, line 147: nit simpler

```suggestion
	authoringSlots := make([]uint64, 0, len(slotToProof))
	for authoringSlot := range slotToProof {
		authoringSlots = append(authoringSlots, authoringSlot)
```

lib/babe/epoch_handler_test.go
```diff
@@ -0,0 +1,91 @@
...
+	eh, err := newEpochHandler(1, startSlot, epochData, constants, testHandleSlotFunc, keypair)
+	require.NoError(t, err)
+	require.Equal(t, 100, len(eh.slotToProof))
+
+	errCh := make(chan error)
+	go eh.run(ctx, errCh)
+	timer := time.After(sd * 100)
+	select {
+	case <-timer:
+		require.Equal(t, 100-(firstExecutedSlot-startSlot), callsToHandleSlot)
```
qdm12, line 87: nit Have `100` as a shared inline constant please

lib/babe/epoch_handler_test.go
```diff
@@ -0,0 +1,91 @@
...
+
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+	eh, err := newEpochHandler(1, startSlot, epochData, constants, testHandleSlotFunc, keypair)
+	require.NoError(t, err)
+	require.Equal(t, 100, len(eh.slotToProof))
+
+	errCh := make(chan error)
+	go eh.run(ctx, errCh)
+	timer := time.After(sd * 100)
```
qdm12, line 84: nit Ideally use `time.NewTimer(sd * 100)` so you can stop the timer in the error receiving select case.

lib/babe/epoch_handler_test.go
```diff
@@ -0,0 +1,91 @@
...
+	constants := &constants{
+		slotDuration: sd,
+		epochLength:  100,
+	}
+
+	keypair := keyring.Alice().(*sr25519.Keypair)
+
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+	eh, err := newEpochHandler(1, startSlot, epochData, constants, testHandleSlotFunc, keypair)
```
qdm12, line 78: nit Please rename `eh` to `epochHandler` or `handler` or any full work 😸

lib/babe/babe.go
```diff
@@ -263,7 +263,9 @@ func (b *Service) Stop() error {
 // Authorities returns the current BABE authorities
 func (b *Service) Authorities() []types.Authority {
 	auths := make([]types.Authority, len(b.epochHandler.epochData.authorities))
-	copy(auths, b.epochHandler.epochData.authorities)
+	for i, auth := range b.epochHandler.epochData.authorities {
+		auths[i] = *(auth.DeepCopy())
```
qdm12, line 267: nit do you need the parentheses? I think this would work too

```suggestion
		auths[i] = *auth.DeepCopy()
```

lib/babe/babe.go
```diff
@@ -321,181 +290,122 @@ func (b *Service) getAuthorityIndex(Authorities []types.Authority) (uint32, erro
...
-			select {
-			case <-b.ctx.Done():
-				cleanup()
-				return nil
-			case <-b.pause:
-				cleanup()
+	for {
+		if err = b.handleEpoch(epoch); err != nil {
+			if errors.Is(err, errServicePaused) || errors.Is(err, context.Canceled) {
 				return nil
```
EclesioMeloJunior, line 344: nit
```suggestion
	for {
	        err = b.handleEpoch(epoch)
		if errors.Is(err, errServicePaused) || errors.Is(err, context.Canceled) {
			return nil
		} else if err != nil {
		        return err
		}
```

lib/babe/babe.go
```diff
@@ -337,74 +339,73 @@ func (b *Service) runEngine() error {
...
-				if !epochTimer.Stop() {
-					<-epochTimer.C
-				}
-			}
+func (b *Service) handleEpoch(epoch uint64) (err error) {
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+	b.epochHandler, err = b.initiateAndGetEpochHandler(epoch)
+	if err != nil {
+		return err
```
qdm12, line 357:
```suggestion
		return fmt.Errorf("cannot initiate and get epoch handler for epoch %d: %w", epoch, err)
```

lib/babe/babe.go
```diff
@@ -337,74 +339,73 @@ func (b *Service) runEngine() error {
...
-			nextEpochStartTime := getSlotStartTime(nextEpochStart, b.constants.slotDuration)
-			epochTimer := time.NewTimer(time.Until(nextEpochStartTime))
-			cleanup := func() {
-				if !epochTimer.Stop() {
-					<-epochTimer.C
-				}
-			}
+func (b *Service) handleEpoch(epoch uint64) (err error) {
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
```
qdm12, line 354: nit perhaps move this further down before it's needed at `go b.epochHandler.run(ctx, errCh)`

lib/babe/babe.go
```diff
@@ -337,74 +339,73 @@ func (b *Service) runEngine() error {
...
+		cleanup()
+		return b.ctx.Err()
+	case <-b.pause:
+		cleanup()
+		return errServicePaused
+	case <-epochTimer.C:
+		// stop current epoch handler
+		cancel()
+	case err := <-errCh:
+		logger.Errorf("error from epochHandler: %w", err)
```
qdm12, line 388:
```suggestion
		logger.Errorf("error from epochHandler: %s", err)
```

EclesioMeloJunior (approved):

lib/babe/babe.go
```diff
@@ -337,74 +339,73 @@ func (b *Service) runEngine() error {
...
+	case <-b.ctx.Done():
+		cleanup()
+		return b.ctx.Err()
+	case <-b.pause:
+		cleanup()
+		return errServicePaused
+	case <-epochTimer.C:
+		// stop current epoch handler
+		cancel()
+	case err := <-errCh:
```
qdm12, line 387: You should also call `cleanup()` here
```suggestion
	case err := <-errCh:
		cleanup()
```

lib/babe/babe.go
```diff
@@ -336,158 +299,104 @@ func (b *Service) initiate() {
...
-
-	// check if it's time to start the epoch yet. if not, wait until it is
-	if time.Since(epochStartTime) < 0 {
-		logger.Debug("waiting for epoch to start")
-		err = func() error {
-			timer := time.NewTimer(time.Until(epochStartTime))
-			defer timer.Stop()
-			select {
-			case <-timer.C:
+			if errors.Is(err, errServicePaused) || errors.Is(err, context.Canceled) {
```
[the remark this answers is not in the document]
noot: we can? idk, the only time the context would be cancelled is if the node shutdowns afaik, so i don't think it matters to return an error

noot:
@qdm12 I think I addressed all your comments, feel free to rereview!

noot:
i think this also closes #2217 (don't see the error anymore when running this branch), will try it more to confirm though.

PR #2152, opened 2022-01-10
feat(lib/grandpa): ensure catch-up logic works

### LINKED ISSUE #1531: lib/grandpa: ensure catch-up logic works when interacting with substrate nodes

kishansagathiya:
## Changes
- ensure catch-up logic works when interacting with other gossamer nodes
- If round in the neighbour message is ahead of our current round by a
threshold, send a catch up request
- process the catch up response, if we can't process it at the moment,
store it to process later.
- [ ]  tests

## Tests

- Run gossamer devnet (https://hackmd.io/@nZ-twauPRISEa6G9zg3XRw/B1RFO3uDt). 
- I tried with 4 peers. Let them run for a few minutes, let few rounds finish. 
- Stop one of those peers (something other than babe lead). Wait for few more rounds to finish (at least 2 more rounds). 
- Start the node you stopped earlier.
- Check through logs that it sends a catch up request to other nodes and catches up to the latest round

## Issues

- #1531

## Primary Reviewer

- @noot


lib/grandpa/message_handler.go
```diff
@@ -87,18 +96,58 @@ func (h *MessageHandler) handleNeighbourMessage(msg *NeighbourMessage) error {
...
+	// if err != nil {
+	// 	return err
+	// }
+
+	// TODO: Why are we ignoring these? Isn't
+	// msg.Number likely to be higher if we are lagging behind?
+	// ignore neighbour messages that are above our head
+	// if int64(msg.Number) > head.Int64() {
+	// 	return nil
+	// }
```
kishansagathiya, line 109: kept this for some clarification.

dot/network/notifications.go
```diff
@@ -419,6 +419,7 @@ func (s *Service) broadcastExcluding(info *notificationsProtocol, excluding peer
 			continue
 		}
 
+		info.outboundHandshakeMutexes.Store(peer, new(sync.Mutex))
```
noot, line 422: don't think we need to add this here, as it's already stored in `sendHandshake`

lib/grandpa/message_handler.go
```diff
@@ -37,6 +40,8 @@ func NewMessageHandler(grandpa *Service, blockState BlockState, telemetryMailer
 	}
 }
 
+//nolint
+// TODO: NotificationMessage is used at places. But NotificationMessage we return is always nil.
```
noot, line 44: the message that's returned is the response that's sent to the peer, `handleCatchUpRequest` should return a response potentially

lib/grandpa/message_handler.go
noot, line 109: we shouldn't send a catch up request for blocks we haven't synced yet as we won't be able to process them

noot: we also receive neighbour messages each time a new block is finalized, so we get them very often

lib/grandpa/message_handler.go
```diff
@@ -87,18 +96,58 @@ func (h *MessageHandler) handleNeighbourMessage(msg *NeighbourMessage) error {
...
+	// TODO: Why are we ignoring these? Isn't
+	// msg.Number likely to be higher if we are lagging behind?
+	// ignore neighbour messages that are above our head
+	// if int64(msg.Number) > head.Int64() {
+	// 	return nil
+	// }
+
+	logger.Debugf("got neighbour message with number %d, set id %d and round %d, from: %s ",
+		msg.Number, msg.SetID, msg.Round, from)
+	// TODO: should we send a justification request here? potentially re-connect this to sync package? (#1815)
```
noot, line 113: you can delete this, the sync algorithm takes care of this now

lib/grandpa/message_handler.go
```diff
@@ -87,18 +96,58 @@ func (h *MessageHandler) handleNeighbourMessage(msg *NeighbourMessage) error {
...
+	highestRound, setID, err := h.blockState.GetHighestRoundAndSetID()
 	if err != nil {
 		return err
 	}
 
-	// ignore neighbour messages that are above our head
-	if int64(msg.Number) > head.Int64() {
-		return nil
+	// catch up only if we are behind by more than catchup threshold
+	if (int(msg.Round) - int(highestRound)) > catchupThreshold {
```
noot, line 121: if our set IDs are different, this may not work correctly. round resets to 0 when the set ID changes, so I'd recommend checking that the set IDs match first before doing this

lib/grandpa/message_handler.go
```diff
@@ -149,31 +198,46 @@ func (h *MessageHandler) handleCommitMessage(msg *CommitMessage) error {
...
+
+	cm, err := resp.ToConsensusMessage()
+	if err != nil {
+		return err
+	}
+
+	err = h.grandpa.network.SendMessage(from, cm)
+	if err != nil {
+		return err
 	}
```
noot, line 234: ah i see you are sending the message here, feel free to change the `handleMessage` return params then

lib/grandpa/message_tracker.go
```diff
@@ -110,15 +111,27 @@ func (t *tracker) handleBlock(b *types.Block) {
...
 		_, err := t.handler.handleMessage("", cm)
 		if err != nil {
 			logger.Warnf("failed to handle commit message %v: %s", cm, err)
+		} else {
+			delete(t.commitMessages, h)
 		}
+	}
 
-		delete(t.commitMessages, h)
+	// TODO: Can I use the same mapLock or do I need to use catchUpResponseLock?
```
noot, line 128: I think it's fine

noot (commented):
basically looks good, the main problem i see is that handling the response isn't thread safe, there is a very high chance we will be sending out multiple requests and getting multiple responses at handling them concurrently. you could add something to check if catch up is currently in progress already, and if so, then don't send out any more requests (see https://github.com/ChainSafe/gossamer/pull/1554/files) it would also be nice to split up the catch-up logic into its own file/module. it would also be nice to make catch up synchronous (the way block request-responses are), this would also prevent multiple requests/responses being sent out/handled at once unnecessarily

dot/network/notifications.go
kishansagathiya: I don't see any code storing mutexes in sendhandshake.

In fact, I added this line because sendHandshake was not able to load these mutexes.

https://github.com/ChainSafe/gossamer/blob/e4bc3757c2011a387d34baba64c8a5bb9aee1a81/dot/network/notifications.go#L321

lib/grandpa/message_handler.go
kishansagathiya: added set id check

lib/grandpa/message_handler.go
kishansagathiya: uncommented it

lib/grandpa/message_handler.go
kishansagathiya: change the `handleMessage` return params

kishansagathiya:
Haven't been able to test these changes because of https://github.com/ChainSafe/gossamer/issues/2237

dot/network/notifications.go
noot: sorry not here I meant in there's a function to store them in the ConnManager in `Start()`:
```

	// this handles all new connections (incoming and outgoing)
	// it creates a per-protocol mutex for sending outbound handshakes to the peer
			prtl.outboundHandshakeMutexes.Store(peerID, new(sync.Mutex))
		}
	}
```

so the mutex should already be stored, as it's created on peer connection

noot: were you seeing `errMissingHandshakeMutex` being returned?

lib/grandpa/catch-up.go
```diff
@@ -0,0 +1,289 @@
...
+	case <-c.catchUpResponseCh:
+		return nil
+	case <-timer.C:
+		return errors.New("timeout")
+	}
+}
+
+func (c *catchUp) sendCatchUpRequest(to peer.ID, req *CatchUpRequest) error {
+	if c.bestResponse.Load() != nil {
+		logger.Debugf("ignoring neighbour message since we are already processing a catch response")
```
noot, line 66:
```suggestion
		logger.Debugf("ignoring neighbour message since we are already processing a catch-up response")
```

lib/grandpa/catch-up.go
```diff
@@ -0,0 +1,289 @@
...
+		return nil
+	}
+
+	// TODO: Clean up all request sent before 5 min / (neighbour message interval)
+	c.lock.Lock()
+	_, ok := c.requestsSent[to]
+	c.lock.Unlock()
+	if ok {
+		logger.Debugf("ignoring neighbour message since we already sent a catch request to this peer: %s", to)
```
noot, line 75:
```suggestion
		logger.Debugf("ignoring neighbour message since we already sent a catch-up request to this peer: %s", to)
```

lib/grandpa/message_handler.go
```diff
@@ -51,30 +55,34 @@ func (h *MessageHandler) handleMessage(from peer.ID, m GrandpaMessage) (network.
...
+		return h.handleCatchUpRequest(msg, from)
 	case *CatchUpResponse:
-		err := h.handleCatchUpResponse(msg)
-		if errors.Is(err, blocktree.ErrNodeNotFound) {
-			// TODO: we are adding these messages to reprocess them again, but we
-			// haven't added code to reprocess them. Do that.
-			// Also, revisit if we need to add these message in synchronous manner
-			// or not. If not, change catchUpResponseMessages to a normal map.  #1531
-			h.grandpa.tracker.addCatchUpResponse(msg)
+		// err := h.handleCatchUpResponse(msg)
```
noot, line 67: remove

lib/grandpa/catch-up.go
```diff
@@ -0,0 +1,289 @@
...
+		return ErrSetIDMismatch
+	}
+
+	if msg.Round <= c.grandpa.state.round {
+		return ErrInvalidCatchUpResponseRound
+	}
+
+	if c.bestResponse.Load().(*CatchUpResponse).Round >= msg.Round {
+		logger.Debug("ignoring catch up response, since we are already processing one with a higher round")
+	}
```
noot, line 124: I'm sort of confused here, why are we checking the `msg` in most spots but then checking `c.bestResponse` here? what is `bestResponse` used for exactly?

noot (commented):
looks good, just a question about the `bestResponse` usage! can you also add/describe some steps for testing this out?

lib/grandpa/catch-up.go
```diff
@@ -0,0 +1,289 @@
...
+	c.grandpa.paused.Store(true)
+
+	timer := time.NewTimer(catchUpResponseTimeout)
+	defer timer.Stop()
+
+	select {
+	case <-c.catchUpResponseCh:
+		return nil
+	case <-timer.C:
+		return errors.New("timeout")
```
EclesioMeloJunior, line 60: use an exported error to this case, like: `var ErrCatchUpTimeout = ...`

lib/grandpa/catch-up.go
```diff
@@ -0,0 +1,289 @@
...
+	// update state and signal to grandpa we are ready to initiate
+	head, err := c.grandpa.blockState.GetHeader(msg.Hash)
+	if err != nil {
+		logger.Debugf("failed to process catch up response for round %d, storing the catch up response to retry", msg.Round)
+		return err
+	}
+
+	c.grandpa.head = head
+	c.grandpa.state.round = msg.Round
+	close(c.grandpa.resumed)
```
EclesioMeloJunior, line 161: why are you closing this channel? the reader is expecting the close?

lib/grandpa/catch-up.go
```diff
@@ -0,0 +1,289 @@
...
+
+	"github.com/ChainSafe/gossamer/lib/common"
+	"github.com/ChainSafe/gossamer/lib/crypto/ed25519"
+	"github.com/libp2p/go-libp2p-core/peer"
+)
+
+const catchUpResponseTimeout = time.Second * 5
+
+type catchUp struct {
+	lock sync.Mutex // applied on requestsSent and bestResponse
```
EclesioMeloJunior, line 20: since `bestResponse` is atomic should not need mutexes, right?
```suggestion
	lock sync.Mutex // applied on requestsSent
```

EclesioMeloJunior, line 20: I would suggest to use RWMutex to maps, use `lock.RLock()` while reading and `lock.Lock()` while writing to the map
```suggestion
	lock sync.RWMutex // applied on requestsSent and bestResponse
```

lib/grandpa/catch-up.go
```diff
@@ -0,0 +1,289 @@
...
+		catchUpResponseCh: make(chan *CatchUpResponse),
+	}
+}
+
+func (c *catchUp) do(to peer.ID, round uint64, setID uint64) error {
+	defer c.waitingOnResponse.Store(false)
+
+	if err := c.sendCatchUpRequest(
+		to, newCatchUpRequest(round, setID),
+	); err != nil {
```
EclesioMeloJunior, line 43: Separate the call and the error check one above the other

EclesioMeloJunior (changes requested):

lib/grandpa/catch-up.go
```diff
@@ -0,0 +1,289 @@
...
+
+	err = c.grandpa.network.SendMessage(to, cm)
+	if err != nil {
+		return err
+	}
+
+	c.lock.Lock()
+	c.requestsSent[to] = *req
+	c.lock.Unlock()
+	c.grandpa.paused.Store(true)
```
EclesioMeloJunior, line 94: The method `do(...)` already call the `c.grandpa.paused.Store(true)`, is it needed to be called here?

EclesioMeloJunior (changes requested):

lib/grandpa/message.go
```diff
@@ -13,9 +13,20 @@ import (
 	"github.com/ChainSafe/gossamer/pkg/scale"
 )
 
+type GrandpaMessageType int64
```
qdm12, line 16: nit uint8 should be enough

```suggestion
type GrandpaMessageType uint8
```

lib/grandpa/message_handler.go
```diff
@@ -92,13 +100,34 @@ func (h *MessageHandler) handleNeighbourMessage(msg *NeighbourMessage) error {
...
+	if err != nil {
+		return err
+	}
+
+	if msg.SetID != setID {
+		return ErrSetIDMismatch
+	}
+
+	// catch up only if we are behind by more than catchup threshold
+	if (int(msg.Round) - int(highestRound)) > catchupThreshold {
```
qdm12, line 124: Define `catchupThreshold` as an inlined constant here instead

dot/network/notifications.go
kishansagathiya: > were you seeing errMissingHandshakeMutex being returned?

Sorry, missed this comment. Yes, I have seen this being returned a lot of times!

kishansagathiya:
While, testing out a gossamer network of 4 node nodes, nodes stop with `runtime: unknown pc` I am continuing to investigate this, but posting the error here for reference if anyone else has any idea
https://gist.github.com/kishansagathiya/6d10d7823403b489481623129d3b663c

I tried eclecio's networking race condition PR  with this branch (https://github.com/ChainSafe/gossamer/pull/2105), but that did not fix the problem.

timwu20:
@kishansagathiya can you make this a draft PR if you don't intend to merge anytime soon?

kishansagathiya:
too old, closing this.

PR #2155, opened 2022-01-10
feature(cmd): add --log-digest flag

### LINKED ISSUE #1851: config: add `dot/digest` to package log level config
## Task summary

- add a config option for log level for the `digest` package to the config file
- pass this into `digest.NewHandler` and use to initialize the package logger

noot:
## Changes

- add `--log-digest` flag to set dot/digest log level from the command line

## Tests

```
go test ./cmd/...
```

## Issues

- closes #1851

## Primary Reviewer

- @jimjbrettj

### DIFF
--- a/dot/digest/digest.go
+++ b/dot/digest/digest.go
@@ -57,7 +57,8 @@ type resume struct {
 }
 
 // NewHandler returns a new Handler
-func NewHandler(lvl log.Level, blockState BlockState, epochState EpochState, grandpaState GrandpaState) (*Handler, error) {
+func NewHandler(lvl log.Level, blockState BlockState, epochState EpochState,
+	grandpaState GrandpaState) (*Handler, error) {
 	imported := blockState.GetImportedBlockNotifierChannel()
 	finalised := blockState.GetFinalisedNotifierChannel()
 
--- a/dot/digest/digest.go
+++ b/dot/digest/digest.go
@@ -18,8 +18,8 @@ import (
 var maxUint64 = uint64(math.MaxUint64)
 
 var (
-	_      services.Service  = &Handler{}
-	logger log.LeveledLogger = log.NewFromGlobal(log.AddContext("pkg", "digest")) // TODO: add to config options (#1851)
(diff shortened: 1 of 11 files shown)


dot/digest/digest.go
```diff
@@ -18,8 +18,8 @@ import (
 var maxUint64 = uint64(math.MaxUint64)
 
 var (
-	_      services.Service  = &Handler{}
-	logger log.LeveledLogger = log.NewFromGlobal(log.AddContext("pkg", "digest")) // TODO: add to config options (#1851)
+	_      services.Service = &Handler{}
+	logger                  = log.NewFromGlobal(log.AddContext("pkg", "digest")) // TODO: add to config options (#1851)
```
qdm12, line 22: You may want to keep it as an interface here to hide the actual implementation. I think `log.NewFromGlobal` returns a `*log.Logger` and not an interface. Not that it matters that much though

dot/services.go
```diff
@@ -426,8 +426,8 @@ func newSyncService(cfg *Config, st *state.Service, fg sync.FinalityGadget,
 	return sync.NewService(syncCfg)
 }
 
-func createDigestHandler(st *state.Service) (*digest.Handler, error) {
-	return digest.NewHandler(st.Block, st.Epoch, st.Grandpa)
+func createDigestHandler(cfg *Config, st *state.Service) (*digest.Handler, error) {
+	return digest.NewHandler(cfg.Log.DigestLvl, st.Block, st.Epoch, st.Grandpa)
```
qdm12, line 430: Why not get a log level argument directly instead of `cfg *Config` if we only need the log level? :thinking:

qdm12 (approved):

dot/digest/digest.go
noot: I removed it cause I got the error:
```
dot/digest/digest.go:65:8: logger.Patch undefined (type log.LeveledLogger has no field or method Patch)
```

other packages don't seem to use the interface either

PR #2188, opened 2022-01-12
fix(dot/sync): cleanup logs; don't log case where we fail to get parent while processing

### LINKED ISSUE #2154: Record failed block addition to prevent processing of child blocks in the blocktree
> basically blocks get processed in 128 block batches generally, so if one block early on fails for whatever reason, then the rest will fail after (as it's ancestor isn't in the blocktree)

Logs get spammed with

```
2022-01-10T10:49:54-05:00 WARN failed to add block with hash 0xa737f399cd30a3d7d5719932a11b3180b911394ff2e6bf71ae7370cb78d00c4a to blocktree: cannot find parent block in blocktree     chain_processor.go:L130  pkg=sync
2022-01-10T10:49:54-05:00 EROR block data processing for block with hash 0xa737f399cd30a3d7d5719932a11b3180b911394ff2e6bf71ae7370cb78d00c4a failed: cannot find parent block in blocktree       chain_processor.go:L86   pkg=sync
2022-01-10T10:49:54-05:00 WARN failed to add block with hash 0xd13ebf257460e4f6276d80b143eca75785dea68f2b707eb4031d9932671410e9 to blocktree: cannot find parent block in blocktree     chain_processor.go:L130  pkg=sync
2022-01-10T10:49:54-05:00 EROR block data processing for block with hash 0xd13ebf257460e4f6276d80b143eca75785dea68f2b707eb4031d9932671410e9 failed: cannot find parent block in blocktree       chain_processor.go:L86   pkg=sync
2022-01-10T10:49:54-05:00 WARN failed to add block with hash 0xda1411a97dff6e487a7e8e5439f48629260e661eb698c7aa56fafe931dfedc50 to blocktree: cannot find parent block in blocktree     chain_processor.go:L130  pkg=sync
2022-01-10T10:49:54-05:00 EROR block data processing for block with hash 0xda1411a97dff6e487a7e8e5439f48629260e661eb698c7aa56fafe931dfedc50 failed: cannot find parent block in blocktree       chain_processor.go:L86   pkg=sync
2022-01-10T10:49:54-05:00 WARN failed to add block with hash 0xe0a0eedf7e790f84df48a5b486597f7896781084a2c6a8d36319438139efc6a3 to blocktree: cannot find parent block in blocktree     chain_processor.go:L130  pkg=sync
2022-01-10T10:49:54-05:00 EROR block data processing for block with hash 0xe0a0eedf7e790f84df48a5b486597f7896781084a2c6a8d36319438139efc6a3 failed: cannot find parent block in blocktree       chain_processor.go:L86   pkg=sync
```

We should record failed block hash addition in a `map[common.Hash]struct{}` for example, and check the set if we cannot find the parent block before logging out an error (or at least add information to the log).

One might also want to set an event driven expiration for each failed hash so it does not stay in memory forever.

noot:
## Changes

- don't log case where we fail to get parent while processing (except as a trace log)
- this doesn't explicitly track the first block that fails (and its whole descendant chain), but I think it's ok as the processing of the block will fail almost immediately if we don't have the parent (ie line 214)

## Tests

```
```

ensure that logs don't get spammed anymore

## Issues

- closes #2154

## Primary Reviewer

- @qdm12

### DIFF
--- a/dot/sync/chain_processor.go
+++ b/dot/sync/chain_processor.go
@@ -169,7 +169,7 @@ func (s *chainProcessor) processBlockData(bd *types.BlockData) error {
 		}
 
 		if err := s.handleBlock(block); err != nil {
-			logger.Errorf("failed to handle block number %s: %s", block.Header.Number, err)
+			logger.Debugf("failed to handle block number %s: %s", block.Header.Number, err)
 			return err
 		}
 
--- a/dot/sync/chain_processor.go
+++ b/dot/sync/chain_processor.go
@@ -83,17 +83,19 @@ func (s *chainProcessor) processReadyBlocks() {
 		}
 
 		if err := s.processBlockData(bd); err != nil {
-			logger.Errorf("block data processing for block with hash %s failed: %s", bd.Hash, err)
-
 			// depending on the error, we might want to save this block for later
 			if errors.Is(err, errFailedToGetParent) {
+				logger.Tracef("block data processing for block with hash %s failed: %s", bd.Hash, err)
 				if err := s.pendingBlocks.addBlock(&types.Block{
 					Header: *bd.Header,
 					Body:   *bd.Body,
 				}); err != nil {
 					logger.Debugf("failed to re-add block to pending blocks: %s", err)
 				}
+				continue
 			}
+
+			logger.Errorf("block data processing for block with hash %s failed: %s", bd.Hash, err)
 		}
 	}
 }
--- a/dot/sync/chain_processor.go
+++ b/dot/sync/chain_processor.go
@@ -84,18 +84,18 @@ func (s *chainProcessor) processReadyBlocks() {
 
 		if err := s.processBlockData(bd); err != nil {
 			// depending on the error, we might want to save this block for later
-			if errors.Is(err, errFailedToGetParent) {
-				logger.Tracef("block data processing for block with hash %s failed: %s", bd.Hash, err)
-				if err := s.pendingBlocks.addBlock(&types.Block{
-					Header: *bd.Header,
-					Body:   *bd.Body,
-				}); err != nil {
-					logger.Debugf("failed to re-add block to pending blocks: %s", err)
-				}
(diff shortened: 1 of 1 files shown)


dot/sync/chain_processor.go
```diff
@@ -83,17 +83,19 @@ func (s *chainProcessor) processReadyBlocks() {
...
 				if err := s.pendingBlocks.addBlock(&types.Block{
 					Header: *bd.Header,
 					Body:   *bd.Body,
 				}); err != nil {
 					logger.Debugf("failed to re-add block to pending blocks: %s", err)
 				}
+				continue
 			}
+
+			logger.Errorf("block data processing for block with hash %s failed: %s", bd.Hash, err)
```
qdm12, line 98: nit Maybe swap around the condition for less indentation?

```suggestion
			if !errors.Is(err, errFailedToGetParent) {
				logger.Errorf("block data processing for block with hash %s failed: %s", bd.Hash, err)
				continue
			}

			logger.Tracef("block data processing for block with hash %s failed: %s", bd.Hash, err)
			if err := s.pendingBlocks.addBlock(&types.Block{
				Header: *bd.Header,
				Body:   *bd.Body,
			}); err != nil {
					logger.Debugf("failed to re-add block to pending blocks: %s", err)
				}
			}
```

qdm12 (approved):

qdm12:
Are you sure that's fine to just silence away? Is it fine to get this `errFailedToGetParent` in all cases? That's why recording blocks that failed to be added would be nice since we could separate the `errFailedToGetParent` that are due to a previous block error and some that might be due to something else :thinking: ?

noot:
@qdm12 yeah I think it's fine, as the sync algorithm will refetch the blocks if they're still needed. one case why this would happen is that the blocktree gets pruned as some block got finalized, and then any blocks on the pruned chain will result in this error when processing, in which case we don't want them anyways. we were also just ignoring it before anyways and it was fine haha

danforbes (approved):

PR #2191, opened 2022-01-12
fix(dot/sync): fix "block with unknown header is ready" error

### LINKED ISSUE #2189: fix "block with unknown header is ready" error
## Task summary

- when syncing on the devnet, the error "block with unknown header is ready" shows up somewhat often
- this really shouldn't be happening, as this means the `handleReadyBlock` function is being called with a block that isn't ready (ie. it's header isn't available to us)
- the error is probably coming from `tipSyncer.handleTick`, as when `handleReadyBlock` is called from `doSync`, the block data is verified to already have headers

noot:
## Changes

- fix "block with unknown header is ready" error which was caused by already having the block in the ready queue or processed, so it was deleted from the pending set already (which is why we couldn't find the header in the pending set)

## Tests

```
go test ./dot/sync
```

## Issues

- closes #2189

## Primary Reviewer

- @EclesioMeloJunior

### DIFF
--- a/dot/sync/chain_sync.go
+++ b/dot/sync/chain_sync.go
@@ -700,8 +700,10 @@ func (cs *chainSync) doSync(req *network.BlockRequestMessage, peersTried map[pee
 }
 
 func handleReadyBlock(bd *types.BlockData, pendingBlocks DisjointBlockSet, readyBlocks *blockQueue) {
-	// see if there are any descendents in the pending queue that are now ready to be processed,
-	// as we have just become aware of their parent block
+	// first check that the block isn't already in the ready queue
+	if readyBlocks.has(bd.Hash) {
+		return
+	}
 
 	// if header was not requested, get it from the pending set
 	// if we're expecting headers, validate should ensure we have a header
@@ -722,6 +724,8 @@ func handleReadyBlock(bd *types.BlockData, pendingBlocks DisjointBlockSet, ready
 
 	logger.Tracef("new ready block number %s with hash %s", bd.Header.Number, bd.Hash)
 
+	// see if there are any descendents in the pending queue that are now ready to be processed,
+	// as we have just become aware of their parent block
 	ready := []*types.BlockData{bd}
 	ready = pendingBlocks.getReadyDescendants(bd.Hash, ready)
 
--- a/dot/sync/chain_sync.go
+++ b/dot/sync/chain_sync.go
@@ -703,7 +703,7 @@ func (cs *chainSync) doSync(req *network.BlockRequestMessage, peersTried map[pee
 func (cs *chainSync) handleReadyBlock(bd *types.BlockData) {
 	// first check that the block isn't already in the ready queue
 	if cs.readyBlocks.has(bd.Hash) {
-		logger.Tracef("ignoring block %s, already in ready queue", bd.Hash)
+		logger.Tracef("ignoring block %s in response, already in ready queue", bd.Hash)
 		return
 	}
 
@@ -722,7 +722,7 @@ func (cs *chainSync) handleReadyBlock(bd *types.BlockData) {
 
 			if has {
 				// seems we already processed the block, so let's just return
-				logger.Tracef("ignoring block we already have, hash=%s", bd.Hash)
+				logger.Tracef("ignoring block we've already processed, hash=%s", bd.Hash)
 				return
 			}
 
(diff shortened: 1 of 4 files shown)


jimjbrettj (approved):
nice work!

dot/sync/chain_sync.go
```diff
@@ -692,41 +693,61 @@ func (cs *chainSync) doSync(req *network.BlockRequestMessage, peersTried map[pee
...
+			// block wasn't in the pending set!
+			// let's check the db as maybe we already processed it
+			has, err := cs.blockState.HasHeader(bd.Hash)
+			if err != nil && !errors.Is(err, chaindb.ErrKeyNotFound) {
+				logger.Debugf("failed to check if header is known for hash %s: err=%s", bd.Hash, err)
+				return
+			}
+
+			if has {
+				// seems we already processed the block, so let's just return
```
qdm12, line 723: nit unneded the log message says it already

```suggestion
```

dot/sync/chain_sync.go
```diff
@@ -692,41 +693,61 @@ func (cs *chainSync) doSync(req *network.BlockRequestMessage, peersTried map[pee
...
 	}
 
 	return nil
 }
 
-func handleReadyBlock(bd *types.BlockData, pendingBlocks DisjointBlockSet, readyBlocks *blockQueue) {
-	// see if there are any descendents in the pending queue that are now ready to be processed,
-	// as we have just become aware of their parent block
+func (cs *chainSync) handleReadyBlock(bd *types.BlockData) {
+	// first check that the block isn't already in the ready queue
```
qdm12, line 703: nit unneeded

```suggestion
```

dot/sync/chain_sync.go
```diff
@@ -692,41 +693,61 @@ func (cs *chainSync) doSync(req *network.BlockRequestMessage, peersTried map[pee
...
+			}
+
+			if has {
+				// seems we already processed the block, so let's just return
+				logger.Tracef("ignoring block we've already processed, hash=%s", bd.Hash)
+				return
+			}
+
+			// this is bad and shouldn't happen
+			logger.Errorf("block with unknown header is ready: hash=%s", bd.Hash)
```
qdm12, line 729: Would this happen if `errors.Is(err, chaindb.ErrKeyNotFound)`? :thinking:

dot/sync/tip_syncer.go
```diff
@@ -8,23 +8,29 @@ import (
 	"math/big"
 
 	"github.com/ChainSafe/gossamer/dot/network"
+	"github.com/ChainSafe/gossamer/dot/types"
 	"github.com/ChainSafe/gossamer/lib/common"
 )
 
 var _ workHandler = &tipSyncer{}
 
+type handleReadyBlockFunc = func(*types.BlockData)
```
qdm12, line 17: nit does the following work?

```suggestion
type handleReadyBlockFunc func(*types.BlockData)
```

qdm12 (approved):

dot/sync/chain_sync.go
```diff
@@ -692,41 +693,61 @@ func (cs *chainSync) doSync(req *network.BlockRequestMessage, peersTried map[pee
...
 	if bd.Header == nil {
-		block := pendingBlocks.getBlock(bd.Hash)
+		block := cs.pendingBlocks.getBlock(bd.Hash)
 		if block == nil {
-			logger.Criticalf("block with unknown header is ready: hash=%s", bd.Hash)
+			// block wasn't in the pending set!
+			// let's check the db as maybe we already processed it
+			has, err := cs.blockState.HasHeader(bd.Hash)
+			if err != nil && !errors.Is(err, chaindb.ErrKeyNotFound) {
+				logger.Debugf("failed to check if header is known for hash %s: err=%s", bd.Hash, err)
```
EclesioMeloJunior, line 718:
```suggestion
				logger.Debugf("failed to check if header is known for hash %s: %s", bd.Hash, err)
```

dot/sync/chain_sync.go
EclesioMeloJunior: this code already checks for `!errors.Is(err, chaindb.ErrKeyNotFound)` so if `has` is false then we don't need to check for this `chaindb.ErrKeyNotFound` error

EclesioMeloJunior (changes requested):

dot/sync/tip_syncer.go
noot: yes it does, is there a preference for one?

dot/sync/chain_sync.go
noot: done!

EclesioMeloJunior (approved):

Issue #1857, opened 2021-10-08
dot/rpc: update `system_syncState` to get `HighestBlock`, `StartingBlock` from dot/sync


noot:
## Task summary

- currently the functions
```
	HighestBlock() int64
	StartingBlock() int64
```
in the `NetworkAPI` are unimplemented due to the sync logic being moved to `dot/sync`
- need to update the rpc methods that use these functions to get the info from `dot/sync` instead
- ie. need to create a `SyncAPI` for `dot/state`

\

edualb:
Hello, I would like to contribute for this refactor, so, I understand that the logic of `HighestBlock` and `StartingBlock` were moved to `dot/sync` package and we need to update who is using them. However, I tried to find implementations for those functions in `dot/sync` and I couldn't. Do I understand it right? If I do, where is this implementation?

noot:
hey @edualb thanks for the interest! currently the `dot/sync` package doesn't contain functions for `HighestBlock` and `StartingBlock`, so you'll have to implement them. 

the `*chainSync` type contains `peerState   map[peer.ID]*peerState` which tracks the current states that our peers report. to implement `HighestBlock`, you will need to iterate through this map and return the highest `peerState.number` found. 

to implement `StartingBlock`, you will need to find the lowest block currently attemping to be synced. you will need to iterate through `*chainSync.workerState.workers` and return the lowest `startNumber` in the map.

btw - the spec for this is here: https://github.com/w3f/PSPs/blob/master/PSPs/drafts/psp-6.md#1515-system_syncstate

hope this helps, let me know if you have any more questions.

edualb:
Thanks @noot ! I will try to implement it after my shift, this comment was very useful 💛

PR #2195, opened 2022-01-14
fix(dot/sync, dot/rpc): implement HighestBlock

edualb:
## Changes
- dot:
	- Create createRPCServiceParams to utilize Parameter Object Pattern
		- Update node.go to utilize createRPCServiceParams
- dot/rpc/modules:
	- Create SyncAPI interface and remove HighestBlock() from NetworkAPI
	- Include SyncAPI interface in SystemModule
	- Calling SyncAPI.HighestBlock() in SystemModule.SyncState() instead of NetworkAPI.HighestBlock()
- dot/rpc:
	- Include SyncAPI interface in HTTPServerConfig
- dot/sync:
	- include getHighestBlock() in ChainSync interface
	- implement HighestBlock() in Sync.Service
	- getHighestBlock() uses strategy to include peers numbers in a slice > sort this slice > get first value.

## Tests

- dot/rpc/modules/mock:
	- dot/rpc/modules/mock: 
		- Generate mock for SyncAPI and update NetworkAPI
	- dot/sync: 
		- Create TestHighestBlock
		- Test implementaion for *chainSync.getHighestBlock

* `cd dot/sync && go test -run TestChainSync_highestBlock`
* `cd dot/sync && go test -run TestHighestBlock`
* I recommend to run others test which was modified in this PR just to connect with SyncAPI interface.

@noot this connection is working, however, I would like to check it in my dev environment. How could I do that?

What I have tried:
* build with: `make build`
* Up my dev environment with: `./bin/gossamer --chain dev init && ./bin/gossamer --chain dev`
* Exec: `curl -H "Content-Type: application/json" -d '{"id":1, "jsonrpc":"2.0", "method": "system_syncState", "params":[]}' http://localhost:8545`

Do I need to change this ID? If I do, for what? 
How could I test it? Do u have any idea?

## Issues

- #1857

## Primary Reviewer

@noot


dot/services.go
```diff
@@ -35,6 +35,19 @@ import (
...
+	cfg         *Config
+	ns          *runtime.NodeStorage
+	stateSrvc   *state.Service
+	coreSrvc    *core.Service
+	networkSrvc *network.Service
+	bp          modules.BlockProducerAPI
+	sysSrvc     *system.Service
+	finSrvc     *grandpa.Service
+	syncer      *sync.Service
+}
```
edualb, line 49: I like this pattern, but if you all wanna rollback it, I can do :)

dot/sync/chain_sync.go
```diff
@@ -932,6 +936,44 @@ func (cs *chainSync) validateJustification(bd *types.BlockData) error {
...
+		numbers = append(numbers, ps.number)
+	}
+
+	if len(numbers) <= 0 {
+		return 0, errNilBlockData
+	}
+
+	sort.SliceStable(numbers, func(i, j int) bool {
+		return numbers[i].Cmp(numbers[j]) >= 0 // numbers[i] >= numbers[j]: https://pkg.go.dev/math/big#Int.Cmp
+	})
```
edualb, line 972: Thinking better... Instead of sorting it, we can just compare inside of the map loop and get the max value from there. What you think? In this way, we wont need `numbers` anymore. 🚀

dot/sync/syncer_test.go
```diff
@@ -42,6 +44,17 @@ func TestMain(m *testing.M) {
...
+type syncAPIMock struct {
+	fnGetHighestBlock func() (int64, error)
+}
+
+func (_m syncAPIMock) start()                                                         {}
+func (_m syncAPIMock) stop()                                                          {}
+func (_m syncAPIMock) setBlockAnnounce(from peer.ID, header *types.Header) error      { return nil }
+func (_m syncAPIMock) setPeerHead(p peer.ID, hash common.Hash, number *big.Int) error { return nil }
+func (_m syncAPIMock) syncState() chainSyncState                                      { return 0 }
+func (_m syncAPIMock) getHighestBlock() (int64, error)                                { return _m.fnGetHighestBlock() }
```
EclesioMeloJunior, line 56: I would suggest to use the generate command `//go:generate mockgen -destination=./mock_chain_sync_test.go -package $GOPACKAGE . ChainSync` (execute `go generate -run mockgen ./...`) to generate the `ChainSync` interface mock ([link](https://github.com/ChainSafe/gossamer/blob/development/.github/CONTRIBUTING.md)), this command will create a new in the `./dot/sync` folder file with the function `NewMockChainSync` and at the function `TestHighestBlock`:

```go
t.Run(ts.name, func(t *testing.T) {
   chainSyncMock := NewMockChainSync(ctrl)
   chainSyncMock.EXPECT().getHighestBlock().Return(ts.in.highestBlock, ts.in.err)
   ...
```

dot/sync/chain_sync.go
```diff
@@ -932,6 +936,44 @@ func (cs *chainSync) validateJustification(bd *types.BlockData) error {
...
+	}
+
+	// improve performance returning first highestBlock
+	if len(ps) == 1 {
+		for _, p := range ps {
+			if p.number == nil {
+				return 0, errNilBlockData
+			}
+			return p.number.Int64(), nil
+		}
```
EclesioMeloJunior, line 955: I don't think the for statement is needed here since we ensure the `len(ps) == 1`
```suggestion
               p := ps[0]
               if p.number == nil {
			return 0, errNilBlockData
		}
		return p.number.Int64(), nil
```

EclesioMeloJunior (changes requested):

dot/sync/chain_sync.go
edualb: Good approach. I will apply it after my shift 😄 thanks!

dot/node.go
```diff
@@ -317,7 +317,18 @@ func NewNode(cfg *Config, ks *keystore.GlobalKeystore) (*Node, error) {
...
+			cfg,
+			ns,
+			stateSrvc,
+			coreSrvc,
+			networkSrvc,
+			bp,
+			sysSrvc,
+			fg,
+			syncer,
+		}
```
qdm12, line 330: Use named fields please.
Also please rename `createRPCServiceParams` to `rpcServiceSettings` for example

dot/services_test.go
```diff
@@ -221,7 +221,18 @@ func TestCreateRPCService(t *testing.T) {
...
+		cfg,
+		ns,
+		stateSrvc,
+		coreSrvc,
+		networkSrvc,
+		nil,
+		sysSrvc,
+		nil,
+		nil,
+	}
```
qdm12, line 234: Use named fields please, thanks! Same below

dot/sync/syncer.go
```diff
@@ -136,6 +136,15 @@ func (s *Service) IsSynced() bool {
 	return s.chainSync.syncState() == tip
 }
 
+func (s *Service) HighestBlock() int64 {
+	highestBlock, err := s.chainSync.getHighestBlock()
+	if err != nil {
+		logger.Errorf("failed to get the highest block: %s", err)
+		return 0
```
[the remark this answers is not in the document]
EclesioMeloJunior: we should not panic. the [RPC documentation](https://github.com/w3f/PSPs/blob/master/PSPs/drafts/psp-6.md#1515-system_syncstate) does not say how to handle errors so I think there is no problem returning 0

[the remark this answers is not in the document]
qdm12: Ok interesting, thanks @EclesioMeloJunior

dot/rpc/modules/api.go
```diff
@@ -153,3 +152,8 @@ type RuntimeStorageAPI interface {
 type SyncStateAPI interface {
 	GenSyncSpec(raw bool) (*genesis.Genesis, error)
 }
+
+//go:generate mockery --name SyncAPI --structname SyncAPI --case underscore --keeptree
```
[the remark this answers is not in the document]
edualb: I will update it... But in `CONTRIBUTING.md` is recommeding to use `mockery`. Maybe it needs to be updated.

dot/sync/syncer_test.go
edualb: Done in 2c5b83c

dot/rpc/modules/api.go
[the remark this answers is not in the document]
edualb: Done in 2c5b83c

dot/sync/syncer.go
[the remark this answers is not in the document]
edualb: Could we resolve this conversation?

dot/services_test.go
edualb: Done in 3de34da

dot/node.go
edualb: Done in 3de34da

dot/sync/chain_sync.go
edualb: I've changed this logic to get a better performance. I will mark you in the comment to review the new approach.

dot/sync/chain_sync.go
```diff
@@ -932,6 +936,44 @@ func (cs *chainSync) validateJustification(bd *types.BlockData) error {
...
+	}
+
+	numbers := []*big.Int{}
```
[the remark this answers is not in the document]
edualb: I've changed this logic to get a better performance. I will mark you in a comment to review the new approach.

dot/sync/chain_sync.go
edualb: @EclesioMeloJunior @qdm12 

As I have commented, I changed the logic of this function. I built this Benchmark to check which logic is better:

```golang
func BenchmarkChainSync_highestBlock(b *testing.B) {
	type input struct {
		peerState map[peer.ID]*peerState
	}
	type output struct {
		highestBlock int64
		err          error
	}
	type test struct {
		name string
		in   input
		out  output
	}
	m := make(map[peer.ID]*peerState)
	for i := 0; i < 5000; i++ {
		id := uuid.New()
		n := rand.Intn(20000)
		ps := &peerState{
			who:    "idtest#1",
			hash:   common.BytesToHash([]byte("hash#1")),
			number: big.NewInt(int64(n)),
		}
		m[peer.ID(id.String())] = ps
	}
	tests := []test{
		{
			name: "when has two peers (p1, p2) with p1.number 190 and p2.number 90 should return 190, nil",
			in: input{
				peerState: m,
			},
			out: output{
				highestBlock: int64(193),
				err:          nil,
			},
		},
	}
	cs := &chainSync{}

	for _, ts := range tests {
		b.Run(ts.name, func(b *testing.B) {
			cs.peerState = ts.in.peerState
			cs.getHighestBlock()
		})
	}
}
```

* Result:
```
# Before

BenchmarkChainSync_highestBlock/when_has_two_peers_(p1,_p2)_with_p1.number_190_and_p2.number_90_should_return_190,_nil-8         	1000000000	         0.002441 ns/op	       0 B/op	       0 allocs/op

# After

BenchmarkChainSync_highestBlock/when_has_two_peers_(p1,_p2)_with_p1.number_190_and_p2.number_90_should_return_190,_nil-8         	1000000000
```

As you can see, we can get a better performance using the "After" approach. For that, I would like a second review here: 8b40b9b

dot/sync/chain_sync.go
```diff
@@ -932,6 +936,44 @@ func (cs *chainSync) validateJustification(bd *types.BlockData) error {
 	return nil
 }
 
+func (cs *chainSync) getHighestBlock() (int64, error) {
+	cs.RLock()
+	ps := cs.peerState
+	cs.RUnlock()
+
+	if len(ps) <= 0 {
```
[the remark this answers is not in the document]
edualb: Done in 8b40b9b

dot/sync/chain_sync.go
```diff
@@ -932,6 +936,44 @@ func (cs *chainSync) validateJustification(bd *types.BlockData) error {
...
+
+	numbers := []*big.Int{}
+	for _, ps := range ps {
+		if ps.number == nil {
+			continue
+		}
+		numbers = append(numbers, ps.number)
+	}
+
+	if len(numbers) <= 0 {
```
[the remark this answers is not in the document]
edualb: Done in 8b40b9b

dot/sync/syncer.go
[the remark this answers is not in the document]
EclesioMeloJunior: yep!

dot/sync/chain_sync.go
```diff
@@ -932,6 +936,34 @@ func (cs *chainSync) validateJustification(bd *types.BlockData) error {
...
+
+	highestBlock := big.NewInt(-1)
+
+	for _, ps := range ps {
+		if ps.number == nil {
+			continue
+		}
+		if ps.number.Cmp(highestBlock) < 0 { // ps.number < highestBlock: https://pkg.go.dev/math/big#Int.Cmp
+			continue
+		}
```
EclesioMeloJunior, line 956:
```suggestion
		if ps.number == nil || ps.number.Cmp(highestBlock) < 0 {
			continue
		}
```

dot/sync/chain_sync.go
```diff
@@ -932,6 +936,34 @@ func (cs *chainSync) validateJustification(bd *types.BlockData) error {
...
+		highestBlock = ps.number
+	}
+
+	if highestBlock.Cmp(big.NewInt(-1)) == 0 { // highestBlock == -1 : https://pkg.go.dev/math/big#Int.Cmp
```
EclesioMeloJunior, line 960:
```suggestion
	if highestBlock.Cmp(big.NewInt(-1)) == 0 {
```

EclesioMeloJunior (changes requested):

dot/rpc/modules/api.go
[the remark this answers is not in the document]
qdm12: My apologies, yes we are currently transitioning from mockery to mockgen/gomock. I'll update the contributing document!

dot/rpc/modules/api.go
```diff
@@ -153,7 +153,8 @@ type SyncStateAPI interface {
 	GenSyncSpec(raw bool) (*genesis.Genesis, error)
 }
 
-//go:generate mockery --name SyncAPI --structname SyncAPI --case underscore --keeptree
+//go:generate mockgen -destination=mock_sync_api_test.go -package $GOPACKAGE . SyncAPI
+// SyncAPI is the interface to interact with sync service
```
qdm12, line 157: nit

```suggestion
// SyncAPI is the interface to interact with the sync service
```

dot/rpc/modules/api.go
```diff
@@ -153,7 +153,8 @@ type SyncStateAPI interface {
 	GenSyncSpec(raw bool) (*genesis.Genesis, error)
 }
 
-//go:generate mockery --name SyncAPI --structname SyncAPI --case underscore --keeptree
+//go:generate mockgen -destination=mock_sync_api_test.go -package $GOPACKAGE . SyncAPI
```
qdm12, line 156: nit adding a new line so linters/go tooling doesn't confuse it with the `SyncAPI` comment below.

```suggestion
//go:generate mockgen -destination=mock_sync_api_test.go -package $GOPACKAGE . SyncAPI

```

dot/services.go
```diff
@@ -36,17 +36,18 @@ import (
...
+type rpcServiceSettings struct {
+	config        *Config
+	nodeStorage   *runtime.NodeStorage
+	state         *state.Service
+	core          *core.Service
+	network       *network.Service
+	blockProducer modules.BlockProducerAPI
+	system        *system.Service
+	blockFinality *grandpa.Service
+	syncer        *sync.Service
```
qdm12, line 50: Much better naming, thank you!! :100:

dot/services.go
```diff
@@ -36,17 +36,18 @@ import (
...
-	ns          *runtime.NodeStorage
-	stateSrvc   *state.Service
-	coreSrvc    *core.Service
-	networkSrvc *network.Service
-	bp          modules.BlockProducerAPI
-	sysSrvc     *system.Service
-	finSrvc     *grandpa.Service
-	syncer      *sync.Service
+// rpcServiceSettings struct to implements the Parameter Object Pattern:
+// https://refactoring.guru/introduce-parameter-object
```
qdm12, line 40: nit I think we can remove this, the type name is quite self explanatory :thinking: 

```suggestion
```

dot/sync/chain_sync.go
```diff
@@ -80,6 +79,7 @@ type workHandler interface {
 	handleTick() ([]*worker, error)
 }
 
+//go:generate mockgen -destination=./mock_chain_sync_test.go -package $GOPACKAGE . ChainSync
```
qdm12, line 82: nit just add a new line and removed unneeded `./` prefix

```suggestion
//go:generate mockgen -destination=mock_chain_sync_test.go -package $GOPACKAGE . ChainSync

```

dot/sync/chain_sync.go
```diff
@@ -941,37 +941,27 @@ func (cs *chainSync) getHighestBlock() (int64, error) {
...
-	}
+	highestBlock := big.NewInt(-1)
 
-	numbers := []*big.Int{}
 	for _, ps := range ps {
 		if ps.number == nil {
 			continue
 		}
-		numbers = append(numbers, ps.number)
+		if ps.number.Cmp(highestBlock) < 0 { // ps.number < highestBlock: https://pkg.go.dev/math/big#Int.Cmp
```
qdm12, line 954: nit unneeded comment, one can just check the Cmp comment. Also this will be changed soon to use `uint` instead of `*big.Int` (in a few weeks) :wink: 

```suggestion
		if ps.number.Cmp(highestBlock) < 0 {
```

dot/sync/chain_sync_test.go
```diff
@@ -827,3 +827,189 @@ func TestChainSync_determineSyncPeers(t *testing.T) {
...
+					"idtest#2": {
+						who:    "idtest#2",
+						hash:   common.BytesToHash([]byte("hash#2")),
+						number: big.NewInt(90),
+					},
+				},
+			},
+			out: output{
+				highestBlock: int64(190),
+				err:          nil,
```
qdm12, line 936: remove setting empty fields (apply to other test cases as well)
```suggestion
```

dot/sync/chain_sync_test.go
```diff
@@ -827,3 +827,189 @@ func TestChainSync_determineSyncPeers(t *testing.T) {
...
+				},
+			},
+			out: output{
+				highestBlock: int64(0),
+				err:          errNilBlockData,
+			},
+		},
+	}
+
+	cs, _ := newTestChainSync(t) // just to improve tests performance
```
qdm12, line 1004: How does this improve performance? If we gain just a few milliseconds, I would prefer to have this in the subtest so that subtests can be ran independently (no surprise) and in parallel (faster)

dot/sync/chain_sync_test.go
```diff
@@ -827,3 +827,189 @@ func TestChainSync_determineSyncPeers(t *testing.T) {
...
+
+	for _, ts := range tests {
+		t.Run(ts.name, func(t *testing.T) {
+			cs.peerState = ts.in.peerState
+
+			highestBlock, err := cs.getHighestBlock()
+			require.ErrorIsf(t, err, ts.out.err, ts.name)
+			require.Equalf(t, highestBlock, ts.out.highestBlock, ts.name)
```
qdm12, line 1012: the failed test will be logged, no need to relog it.

```suggestion
			require.ErrorIs(t, err, ts.out.err)
			require.Equal(t, highestBlock, ts.out.highestBlock)
```

dot/sync/syncer_test.go
```diff
@@ -167,3 +168,64 @@ func newTestGenesisWithTrieAndHeader(t *testing.T) (*genesis.Genesis, *trie.Trie
...
+			s := newTestSyncer(t)
+
+			ctrl := gomock.NewController(t)
+			chainSync := NewMockChainSync(ctrl)
+			chainSync.EXPECT().getHighestBlock().Return(ts.in.highestBlock, ts.in.err)
+
+			s.chainSync = chainSync
+
+			result := s.HighestBlock()
+			require.Equalf(t, result, ts.out.highestBlock, ts.name)
```
qdm12, line 228:
```suggestion
			require.Equal(t, result, ts.out.highestBlock)
```

dot/sync/chain_sync.go
```diff
@@ -932,6 +936,34 @@ func (cs *chainSync) validateJustification(bd *types.BlockData) error {
 	return nil
 }
 
+func (cs *chainSync) getHighestBlock() (int64, error) {
+	cs.RLock()
+	ps := cs.peerState
+	cs.RUnlock()
```
noot, line 942:
```suggestion
	cs.RLock()
	defer cs.RUnlock()
	ps := cs.peerState
```

afaik `ps` will still point to `cs.peerState`, it's just an alias. so you'd need to unlock when you're completely done with reading the map, or there could be a race condition

noot: I'd also probably remove the `ps := cs.peerState` as well as just reference `cs.peerState` directly

dot/sync/chain_sync.go
```diff
@@ -932,6 +936,34 @@ func (cs *chainSync) validateJustification(bd *types.BlockData) error {
...
+			continue
+		}
+		if ps.number.Cmp(highestBlock) < 0 { // ps.number < highestBlock: https://pkg.go.dev/math/big#Int.Cmp
+			continue
+		}
+		highestBlock = ps.number
+	}
+
+	if highestBlock.Cmp(big.NewInt(-1)) == 0 { // highestBlock == -1 : https://pkg.go.dev/math/big#Int.Cmp
+		return 0, errNilBlockData
```
noot, line 961: you can probably just return 0 in this case, or check it in the higher up RPC function and return 0 there.

noot: ah I see you return 0 in the higher function, so that's fine

dot/sync/syncer.go
```diff
@@ -136,6 +136,16 @@ func (s *Service) IsSynced() bool {
 	return s.chainSync.syncState() == tip
 }
 
+// HighestBlock gets the highest known block number
+func (s *Service) HighestBlock() int64 {
+	highestBlock, err := s.chainSync.getHighestBlock()
+	if err != nil {
+		logger.Errorf("failed to get the highest block: %s", err)
```
noot, line 143: maybe change to `Warnf`?

noot (commented):
nice work, looks good! just one comment

edualb:
@noot @qdm12 @EclesioMeloJunior 

I didn't have time to check your comments yesterday, tonight I will, ok? :)

dot/sync/chain_sync.go
edualb: Yeah, I did some tests here and I thought that I could get a panic when the `ps.number` is `nil` but I didn't. Good catch!

Done in 5afd977

dot/sync/chain_sync.go
edualb: Done in 5afd977

dot/services.go
edualb: Thanks! 😄

dot/rpc/modules/api.go
edualb: Done in c2e3d69

dot/rpc/modules/api.go
edualb: Done in c2e3d69

dot/services.go
edualb: Done in c2e3d69

dot/sync/chain_sync.go
edualb: Done in c2e3d69

dot/sync/chain_sync.go
edualb: Done in c2e3d69

dot/sync/chain_sync_test.go
edualb: If a new developer face those tests without fields, I believe he/she could be stunned in a first moment... Although this cause more code, I think it is more clear (as a developer) when all fields filled out. Of course, there is cases that isn't good to have that but in this case I think is useful.

What you think? I can change it but I would like to know your opinion before.

dot/sync/chain_sync_test.go
edualb: I realized that is hard to measure it in my machine... Sometimes when I put it in the subtest is faster than out the subtest and vice versa... I will put it in the subtests.

edualb: Done in d09bc0b

dot/sync/chain_sync_test.go
edualb: Done in d09bc0b

dot/sync/syncer_test.go
edualb: Done in d09bc0b

dot/sync/chain_sync.go
edualb: Its true @noot. I checked it here: https://go.dev/play/p/bCsiQceROqb

I didn't notice that all values in the map has the same reference... Good catch.

dot/sync/chain_sync.go
edualb: Could we resolve this conversation? 😛

edualb: Done in 60eaaa8

dot/sync/chain_sync.go
edualb: Done in 60eaaa8

dot/sync/syncer.go
edualb: I will search more about that and put my opinion here tomorrow, ok? 🙂

edualb: Morning, I found some definitions about warn vs error and I believe that makes sense. I will change it to `Warnf` tonight.

Reasons:
* this error doesn't block all process
* this error just means that we are missing an information and the reason for that
* The application there is a chance to recover from this error later

Resources:
* https://reflectoring.io/logging-levels/
* https://stackoverflow.com/questions/2031163/when-to-use-the-different-log-levels

dot/sync/chain_sync_test.go
qdm12: Well I have to admit I am big fan of explicitness. 
On the other hand, settings all empty fields explicitly to their empty value does add a lot of code, especially for more complex test cases or a lot of test cases, making it quite unreadable.

Some like [Uber also advise to leave empty fields empty](https://github.com/uber-go/guide/blob/master/style.md#omit-zero-value-fields-in-structs).

Although again, in this particular case, up to you it doesn't matter much 😉

dot/sync/chain_sync_test.go
```diff
@@ -827,3 +827,188 @@ func TestChainSync_determineSyncPeers(t *testing.T) {
...
+	}
+
+	for _, ts := range tests {
+		t.Run(ts.name, func(t *testing.T) {
+			cs, _ := newTestChainSync(t)
+			cs.peerState = ts.in.peerState
+
+			highestBlock, err := cs.getHighestBlock()
+			require.ErrorIs(t, err, ts.out.err, ts.name)
+			require.Equal(t, highestBlock, ts.out.highestBlock, ts.name)
```
qdm12, line 1011:
```suggestion
			require.ErrorIs(t, err, ts.out.err)
			require.Equal(t, highestBlock, ts.out.highestBlock)
```

dot/sync/chain_sync.go
qdm12: Sounds good!

dot/sync/chain_sync_test.go
edualb: I've never seen this guide, cool! I can use that to defend some arguments in the future 😆 

So, in this case I would prefer keep it. These zero values could help to understand these tests, because names are written in this template: "when {{input}} should return {{output}}" and the uber guide says:

```
Include zero values where field names provide meaningful context. For example, test cases in Test Tables can benefit from names of fields even when they are zero-valued.

tests := []struct{
  give string
  want int
}{
  {give: "0", want: 0},
  // ...
}
```

I think I can remove some fields from `peerState` like: `who` and `hash` because they aren't being used in this test.

Just to finish, could we close this conversation if I do these things(?):

* Remove `who` and `hash` from `peerState` (they aren't provide any meaningful)
* Keep others meaningful zero values for tests.

qdm12: Yes that makes sense! Good catch the Uber guide bite me back 😆  I'll resolve it and let it up to you 😉

dot/sync/chain_sync_test.go
edualb: Done in dad0fc5

dot/sync/syncer.go
edualb: Done in a3b4324

EclesioMeloJunior (approved):

dot/sync/chain_sync.go
```diff
@@ -932,6 +937,30 @@ func (cs *chainSync) validateJustification(bd *types.BlockData) error {
 	return nil
 }
 
+func (cs *chainSync) getHighestBlock() (int64, error) {
+	cs.RLock()
+	defer cs.RUnlock()
+
+	if len(cs.peerState) == 0 {
+		return 0, errNoPeers
```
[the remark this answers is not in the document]
qdm12: Oh yeah right, if it doesn't cross package boundaries you can keep it unexported, my bad!

dot/sync/chain_sync_test.go
```diff
@@ -827,3 +827,140 @@ func TestChainSync_determineSyncPeers(t *testing.T) {
...
+		{
+			name: "when has two peers (p1, p2) with p1.number nil and p2.number nil should return 0, errNilBlockData",
+			in: input{
+				peerState: map[peer.ID]*peerState{
+					"idtest#1": {number: nil},
+					"idtest#2": {number: nil},
+				},
+			},
+			out: output{
+				highestBlock: int64(0),
```
qdm12, line 950: nit I think this should do (same for other test cases)

```suggestion
				highestBlock: 0,
```

edualb: totally true. Done in 5bb3b1b

qdm12 (approved):

PR #2196, opened 2022-01-14
fix(dot/state): actually prune finalized tries from memory

noot:
## Changes

- fix removal of tries older than current finalized from memory
- exchange `BlockAnnounceHandshake` with peer on connect, previously it would sometimes take many minutes for the node to start bootstrap syncing. this makes it start basically as soon as we get peers.

## Tests

```
run gossamer for many hours and see better memory usage (hopefully)?
```

## Issues

- #1973 

## Primary Reviewer

- @qdm12

### DIFF
--- a/dot/network/service.go
+++ b/dot/network/service.go
@@ -443,12 +443,12 @@ func (s *Service) handleConn(conn libp2pnetwork.Conn) {
 		return
 	}
 
-	stream, err := s.sendHandshake(conn.RemotePeer(), hs, prtl)
+	_, err = s.sendHandshake(conn.RemotePeer(), hs, prtl)
 	if err != nil {
 		return
 	}
 
-	_ = stream.Close()
+	// leave stream open if there's no error
 }
 
 // Stop closes running instances of the host and network services as well as
--- a/dot/network/service.go
+++ b/dot/network/service.go
@@ -440,11 +440,19 @@ func (s *Service) handleConn(conn libp2pnetwork.Conn) {
 
 	hs, err := prtl.getHandshake()
 	if err != nil {
+		logger.Warnf("failed to get handshake for protocol %s: %s",
+			prtl.protocolID,
+			err,
+		)
 		return
 	}
 
 	_, err = s.sendHandshake(conn.RemotePeer(), hs, prtl)
 	if err != nil {
+		logger.Debugf("failed to send handshake to peer on connection, peer=%s: %s",
+			conn.RemotePeer(),
+			err,
+		)
 		return
 	}
 
--- a/dot/network/service.go
+++ b/dot/network/service.go
@@ -449,7 +449,7 @@ func (s *Service) handleConn(conn libp2pnetwork.Conn) {
 
 	_, err = s.sendHandshake(conn.RemotePeer(), hs, prtl)
 	if err != nil {
-		logger.Debugf("failed to send handshake to peer on connection, peer=%s: %s",
+		logger.Debugf("failed to send handshake to peer %s on connection: %s",
 			conn.RemotePeer(),
 			err,
 		)
--- a/dot/network/service.go
+++ b/dot/network/service.go
@@ -430,6 +430,25 @@ func (s *Service) sentBlockIntervalTelemetry() {
 func (s *Service) handleConn(conn libp2pnetwork.Conn) {
 	// TODO: currently we only have one set so setID is 0, change this once we have more set in peerSet.
 	s.host.cm.peerSetHandler.Incoming(0, conn.RemotePeer())
+
+	// exchange BlockAnnounceHandshake with peer so we can start to
+	// sync if necessary.
+	prtl, has := s.notificationsProtocols[BlockAnnounceMsgType]
+	if !has {
+		return
+	}
+
+	hs, err := prtl.getHandshake()
+	if err != nil {
+		return
+	}
+
+	stream, err := s.sendHandshake(conn.RemotePeer(), hs, prtl)
+	if err != nil {
+		return
+	}
+
+	_ = stream.Close()
 }
 
 // Stop closes running instances of the host and network services as well as
--- a/dot/network/service_test.go
+++ b/dot/network/service_test.go
@@ -100,7 +100,7 @@ func createTestService(t *testing.T, cfg *Config) (srvc *Service) {
 			Port:         availablePort(t),
 			NoBootstrap:  true,
 			NoMDNS:       true,
-			LogLvl:       4,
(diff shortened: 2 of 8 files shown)


noot:
on this brach: 
- at block 24k and around 20min uptime: the memory usage was 11.2% of 15.4gb = 1.725gb
- at bloock 213k and ~3h uptime: 52.6% of 15.4gb = 8.1gb

on development: 
- at block 26k around 22min uptime: 12.9% of 15.5gb = 2gb
= at block 183k and ~3h uptime: 60.4% of 15.5gb = 9.362gb

seems like there's moderate improvement with these changes, but memory usage still slowly creeps up. will continue to run the nodes and report back

noot:
on this branch at 3h36min, block height 250k, usage = 63% of 15.4gb = 9.7gb
on development at 3h36min, block height 215k, usage = 76.8% of 15.5gb = 11.9gb

it seems after the 3 hour mark is when the memory usage really starts to increase, seems quite a bit better here

dot/network/service.go
```diff
@@ -430,6 +430,25 @@ func (s *Service) sentBlockIntervalTelemetry() {
 func (s *Service) handleConn(conn libp2pnetwork.Conn) {
 	// TODO: currently we only have one set so setID is 0, change this once we have more set in peerSet.
 	s.host.cm.peerSetHandler.Incoming(0, conn.RemotePeer())
+
+	// exchange BlockAnnounceHandshake with peer so we can start to
+	// sync if necessary.
+	prtl, has := s.notificationsProtocols[BlockAnnounceMsgType]
```
qdm12, line 436: Why not `protocol` :cry:

dot/network/service.go
```diff
@@ -430,6 +430,25 @@ func (s *Service) sentBlockIntervalTelemetry() {
...
+	if !has {
+		return
+	}
+
+	hs, err := prtl.getHandshake()
+	if err != nil {
+		return
```
qdm12, line 443: Maybe log it out? :thinking: Does this happen often?

Same question for `sendHandshake` below

dot/state/block_finalisation.go
```diff
@@ -235,8 +235,17 @@ func (bs *BlockState) handleFinalisedBlock(curr common.Hash) error {
...
+		block, has = bs.getAndDeleteUnfinalisedBlock(hash)
+		if !has {
+			continue
+		}
+
+		logger.Tracef("cleaned out finalised block from memory; block number %s with hash %s", block.Header.Number, hash)
+
+		go func(header *types.Header) {
+			bs.pruneKeyCh <- header
+		}(&block.Header)
```
qdm12, line 248: Do we really need to run this in a goroutine? Can this be blocking :thinking:

lib/blocktree/blocktree.go
```diff
@@ -133,20 +133,20 @@ func (bt *BlockTree) getNode(h Hash) (ret *node) {
 // Prune sets the given hash as the new blocktree root,
 // removing all nodes that are not the new root node or its descendant
 // It returns an array of hashes that have been pruned
-func (bt *BlockTree) Prune(finalised Hash) (pruned []Hash) {
+func (bt *BlockTree) Prune(finalised Hash) []Hash {
```
qdm12, line 136: nit What's wrong with the named return? :smiling_face_with_tear:

qdm12 (approved):
Improvements :100: ! Although not the thick silver bullet I have been dreaming of lately :smile:

lib/blocktree/blocktree.go
noot: nothing, i was just trying things out lol, changed it back

dot/state/block_finalisation.go
noot: no it shouldn't be blocking anymore, will update

dot/network/service.go
noot: i don't think this would ever really happen, I can add a log though

dot/network/service.go
noot: there's a package already imported at `protocol`, i can change it though if you want!

qdm12: Ah yes, one of the dumb Go feature/bug 😢 

Maybe have it as `proto`? Or rename the import to `protocolpkg` 🤔 Up to you!

dot/network/service.go
qdm12: Yeah log it out at the error level so we can see if it happens? If it does happen a lot, we can just then either handle it or log it at the trace level.

dot/network/service.go
```diff
@@ -440,15 +440,23 @@ func (s *Service) handleConn(conn libp2pnetwork.Conn) {
...
+			prtl.protocolID,
+			err,
+		)
 		return
 	}
 
-	stream, err := s.sendHandshake(conn.RemotePeer(), hs, prtl)
+	_, err = s.sendHandshake(conn.RemotePeer(), hs, prtl)
 	if err != nil {
+		logger.Debugf("failed to send handshake to peer on connection, peer=%s: %s",
```
qdm12, line 452: supernit making it more sentence oriented

```suggestion
		logger.Debugf("failed to send handshake to peer %s on connection: %s",
```

qdm12 (approved):

dot/network/service.go
```diff
@@ -440,15 +440,23 @@ func (s *Service) handleConn(conn libp2pnetwork.Conn) {
...
+			conn.RemotePeer(),
+			err,
+		)
 		return
 	}
 
-	_ = stream.Close()
+	// leave stream open if there's no error
```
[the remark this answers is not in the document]
noot: yeah, `sendHandshake` will close the stream if it returns and error

dot/network/service_test.go
```diff
@@ -100,7 +100,7 @@ func createTestService(t *testing.T, cfg *Config) (srvc *Service) {
 			Port:         availablePort(t),
 			NoBootstrap:  true,
 			NoMDNS:       true,
-			LogLvl:       4,
+			LogLvl:       5,
```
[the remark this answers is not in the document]
noot: yes will do!

dot/network/service.go
noot: done

qdm12 (changes requested):

dot/state/block_finalisation.go
noot: woops, i updated the other part but not here. done!

qdm12 (approved):

EclesioMeloJunior (approved):

Issue #2186, opened 2022-01-11
[Core] rewrite dot/core message tests


jimjbrettj:
## Issue summary

- Rewrite dot/core message tests to increase speed and code coverage

PR #2197, opened 2022-01-14
testing (dot/core): rewrite message.go unit tests

jimjbrettj:
## Changes

- Rewrite dot/core message.go unit tests to increase speed and code coverage

## Tests

```
 go test ./dot/core/ -v -coverprofile coverage.out
 go tool cover -func coverage.out // To view file code coverage
```

## Issues

- #2186

## Primary Reviewer

- @timwu20


dot/core/interface.go
```diff
@@ -18,7 +18,7 @@ import (
 	"github.com/ChainSafe/gossamer/lib/transaction"
 )
 
-//go:generate mockery --name BlockState --structname BlockState --case underscore --keeptree
+//go:generate mockgen -destination=./mock_core_test.go -package $GOPACKAGE . BlockState,StorageState,TransactionState,Network,EpochState,CodeSubstitutedState,DigestHandler
```
qdm12, line 21: meganit


```suggestion
//go:generate mockgen -destination=mock_core_test.go -package $GOPACKAGE . BlockState,StorageState,TransactionState,Network,EpochState,CodeSubstitutedState,DigestHandler
```

dot/core/messages.go
```diff
@@ -63,6 +64,8 @@ func (s *Service) HandleTransactionMessage(peerID peer.ID, msg *network.Transact
 						Reason: peerset.BadTransactionReason,
 					}, peerID)
 				}
+
+				hasInvalidTxn = true
```
qdm12, line 68: Ok this is out of scope, but it would be really nice if you could move this mega confusing `func() error` outside of this function? And you can pass things as arguments instead.

dot/core/messages_test.go
```diff
@@ -2,3 +2,265 @@
...
+
+	"github.com/golang/mock/gomock"
+	"github.com/libp2p/go-libp2p-core/peer"
+	"github.com/stretchr/testify/assert"
+)
+
+var (
+	errBestHeader = errors.New("best header error")
+	errGetRuntime = errors.New("get runtime error")
+	errTrieState  = errors.New("trie state error")
```
qdm12, line 27: Why not just have a `errTest = errors.New("dummy error")` if it's only used for tests? Otherwise use errors defined in the production code.

dot/core/messages_test.go
```diff
@@ -2,3 +2,265 @@
...
+)
+
+// TODO use new test format
```
qdm12, line 30: What do you mean here? :thinking:

dot/core/messages_test.go
```diff
@@ -2,3 +2,265 @@
...
+
+func TestService_TransactionsCount(t *testing.T) {
+	ctrl := gomock.NewController(t)
+	mockTxnStateEmpty := NewMockTransactionState(ctrl)
+	mockTxnState := NewMockTransactionState(ctrl)
+
+	txs := make([]*transaction.ValidTransaction, 2)
```
qdm12, line 37: meganit just for a bit more expliciteness since we never set it.

```suggestion
	txs := []*transaction.ValidTransaction{nil, nil}
```

dot/core/messages_test.go
```diff
@@ -2,3 +2,265 @@
...
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			s := tt.service
+			res := s.TransactionsCount()
+			assert.Equal(t, tt.exp, res)
+		})
+	}
+}
+
+func TestServiceHandleTransactionMessage(t *testing.T) {
```
qdm12, line 67: Do you think it would be possible to define these mocks in the subtests using test case parameters instead?

[Example](https://github.com/ChainSafe/gossamer/blob/e1f7f96cb8cb7e82156f7333b9a0704ab8151fd2/internal/trie/node/header_test.go#L110-L121)

That would make it easier to follow how the mock is configured by just looking up the test case fields.

qdm12 (approved):
That `test.log` though :smile: 

Looks good, just a few comments here and there

dot/core/messages_test.go
```diff
@@ -2,3 +2,265 @@
...
+
+func TestService_TransactionsCount(t *testing.T) {
+	ctrl := gomock.NewController(t)
+	mockTxnStateEmpty := NewMockTransactionState(ctrl)
+	mockTxnState := NewMockTransactionState(ctrl)
+
+	txs := make([]*transaction.ValidTransaction, 2)
+
+	mockTxnStateEmpty.EXPECT().PendingInPool().Return([]*transaction.ValidTransaction{})
+	mockTxnState.EXPECT().PendingInPool().Return(txs)
```
noot, line 40: wait why is the transaction state returning a slice with nil transactions?

noot (commented):
can you pull development into here? that should remove the test.log file completely

dot/core/messages_test.go
jimjbrettj: oops meant to delete! It was a reminder to myself to use the feedback you gave on my last PR on how to structure tests

dot/core/messages_test.go
jimjbrettj: I can certainly try!

dot/core/messages_test.go
jimjbrettj: Cuz I'm only testing the length function, so i thought it would be good to have a test case testing empty input. Can remove that case if that's preferred

dot/core/messages_test.go
jimjbrettj: The only reason I could think is if we wanted to make sure we are asserting specific errors, but I think this should be fine to do

dot/core/messages.go
jimjbrettj: I fully support this, it was a point of confusion for me while testing this lol

dot/core/messages_test.go
jimjbrettj: @qdm12 Maybe I'm not totally understanding, but I think this might be more tricky in my case since the different test cases have different mock calls. From what I can tell in this example each case calls the Write function from write mock, but if the mock methods called are variable per test case is this still possible/reasonable to do?

dot/core/messages.go
qdm12: I bet it was 😄

dot/core/messages.go
```diff
@@ -6,15 +6,57 @@ package core
...
 	"github.com/ChainSafe/gossamer/dot/peerset"
 	"github.com/ChainSafe/gossamer/dot/types"
 	"github.com/ChainSafe/gossamer/lib/runtime"
 	"github.com/ChainSafe/gossamer/lib/transaction"
+
+	"github.com/libp2p/go-libp2p-core/peer"
 )
 
+func (s *Service) validateTransaction(peerID peer.ID, head *types.Header, rt runtime.Instance,
+	tx types.Extrinsic, toPropagate *[]types.Extrinsic) (bool, error) {
```
qdm12, line 19: Use named returns, I guess the bool could be `failed`? See confusion I got below 😄 Also maybe flip it around, usually `true` means ok and `false` means not ok

```suggestion
	tx types.Extrinsic, toPropagate *[]types.Extrinsic) (failed bool, err error) {
```

dot/core/messages.go
```diff
@@ -6,15 +6,57 @@ package core
...
+	s.storageState.Lock()
+	defer s.storageState.Unlock()
+
+	ts, err := s.storageState.TrieState(&head.StateRoot)
+	if err != nil {
+		return true, err
```
qdm12, line 25:
```suggestion
		return true, fmt.Errorf("cannot get trie state from storage for root %s: %w", head.StateRoot, err)
```

qdm12, line 25: Should this return `false`? 🤔

dot/core/messages.go
```diff
@@ -6,15 +6,57 @@ package core
...
+	if err != nil {
+		if errors.Is(err, runtime.ErrInvalidTransaction) {
+			s.net.ReportPeer(peerset.ReputationChange{
+				Value:  peerset.BadTransactionValue,
+				Reason: peerset.BadTransactionReason,
+			}, peerID)
+		}
+
+		logger.Debugf("failed to validate transaction: %s", err)
+		return true, nil
```
qdm12, line 42: What's the boolean returned here?

dot/core/messages.go
```diff
@@ -6,15 +6,57 @@ package core
...
+	}
+
+	// create new valid transaction
```
qdm12, line 45:
```suggestion
```

dot/core/messages.go
```diff
@@ -6,15 +6,57 @@ package core
...
+	vtx := transaction.NewValidTransaction(tx, val)
+
+	// push to the transaction queue of BABE session
+	hash := s.transactionState.AddToPool(vtx)
+	logger.Tracef("added transaction with hash %s to pool", hash)
+
+	// find tx(s) that should propagate
+	if val.Propagate {
+		*toPropagate = append(*toPropagate, tx)
```
qdm12, line 54: do we really need to pass it as a pointer? Can we instead return it and assign on the caller side?

dot/core/messages_test.go
qdm12: You can have a `callXXX bool` field in the test case to check in the subtest to expect another mock call or not. It looks there is a lot of repetition so maybe try doing it for this test only to see how it is? 🤔

dot/core/service_integration_test.go
```diff
@@ -504,7 +385,11 @@ func TestMaintainTransactionPool_EmptyBlock(t *testing.T) {
 		},
 	}
 
-	ts := state.NewTransactionState()
+	ctrl := gomock.NewController(t)
+	telemetryMock := NewMockClient(ctrl)
+	telemetryMock.EXPECT().SendMessage(gomock.Any()).AnyTimes()
```
qdm12, line 390: Can we expect particular messages or are there too many being sent out?

dot/core/messages.go
jimjbrettj: Yeah we can do that

dot/core/service_integration_test.go
jimjbrettj: Um i am not sure I did not do telemetry tests, these are same as whats in dev. I had to manually bring them over though because of file change. But the contents of this file should (unless I messed it up somewhere) match service_test.go on dev branch

dot/core/messages_test.go
jimjbrettj: okay yeah ill give it a try and see how it goes

dot/core/messages.go
qdm12: As in it is as intuitive as the other way. It's just Go standards have disformed my brain so when something returns a boolean that is `not Something` it triggers me because I really don't want to have `!notSomething` 😄

dot/core/service_integration_test.go
qdm12: Oh yeah no worry then! Sorry I'm just brute-reviewing all this 😄

dot/core/messages.go
```diff
@@ -16,13 +17,13 @@ import (
 )
 
 func (s *Service) validateTransaction(peerID peer.ID, head *types.Header, rt runtime.Instance,
-	tx types.Extrinsic, toPropagate *[]types.Extrinsic) (bool, error) {
+	tx types.Extrinsic) (validity *transaction.Validity, failed bool, err error) {
```
qdm12, line 20: failed is meant to be `ok` or `success` here right?

```suggestion
	tx types.Extrinsic) (validity *transaction.Validity, valid bool, err error) {
```

dot/core/messages.go
```diff
@@ -83,19 +78,24 @@ func (s *Service) HandleTransactionMessage(peerID peer.ID, msg *network.Transact
...
 	}
 
-	hasInvalidTxn := false
+	isValid := true
 	for _, tx := range txs {
-		isInvalid, err := s.validateTransaction(peerID, head, rt, tx, &toPropagate)
-		if isInvalid {
-			hasInvalidTxn = true
+		validity, isValidTxn, err := s.validateTransaction(peerID, head, rt, tx)
+		if !isValidTxn {
```
qdm12, line 84: Maybe check the error first?

dot/core/messages.go
jimjbrettj: yup good call!

dot/core/messages.go
```diff
@@ -81,6 +81,10 @@ func (s *Service) HandleTransactionMessage(peerID peer.ID, msg *network.Transact
 	isValid := true
 	for _, tx := range txs {
 		validity, isValidTxn, err := s.validateTransaction(peerID, head, rt, tx)
+		if err != nil {
+			return false, err
```
qdm12, line 85: feel free to add more context to the wrapping

```suggestion
			return false, fmt.Errorf("failed validating transaction for peerID %s: %w:", peerID, err)
```

qdm12 (approved):

dot/core/messages_test.go
```diff
@@ -2,3 +2,372 @@
...
+		{
+			name: "runtime.ErrInvalidTransaction",
+			mockNetwork: &mockNetwork{
+				IsSynced: true,
+				ReportPeer: &mockReportPeer{
+					change: peerset.ReputationChange{
+						Value:  peerset.BadTransactionValue,
+						Reason: peerset.BadTransactionReason,
+					},
+					id: peer.ID("jimbo"),
```
qdm12, line 240: Leaving the author's mark 😄

qdm12 (approved):

timwu20 (approved):

PR #2206, opened 2022-01-17
fix(dot/state): change map of tries implementation to have working garbage collection

qdm12:
## Changes

- [x] Remove `syncing` field (always false) from `StorageState`
- [x] Add new type `tries` implementing a mutex locked map of root hash to trie pointers

## Tests

```
go test ./dot/state
```

## Issues

- #1973 

## Primary Reviewer

- @noot

### DIFF
--- a/dot/state/storage.go
+++ b/dot/state/storage.go
@@ -30,7 +30,7 @@ func errTrieDoesNotExist(hash common.Hash) error {
 // StorageState is the struct that holds the trie, db and lock
 type StorageState struct {
 	blockState *BlockState
-	tries      *sync.Map // map[common.Hash]*trie.Trie // map of root -> trie
+	tries      *tries
 
 	db chaindb.Database
 	sync.RWMutex
@@ -52,8 +52,7 @@ func NewStorageState(db chaindb.Database, blockState *BlockState,
 		return nil, fmt.Errorf("cannot have nil trie")
 	}
 
-	tries := new(sync.Map)
-	tries.Store(t.MustHash(), t)
+	tries := newTries(t)
 
 	storageTable := chaindb.NewTable(db, storagePrefix)
 
@@ -79,14 +78,14 @@ func NewStorageState(db chaindb.Database, blockState *BlockState,
 
 func (s *StorageState) pruneKey(keyHeader *types.Header) {
 	logger.Tracef("pruning trie, number=%d hash=%s", keyHeader.Number, keyHeader.Hash())
-	s.tries.Delete(keyHeader.StateRoot)
+	s.tries.delete(keyHeader.StateRoot)
 }
 
 // StoreTrie stores the given trie in the StorageState and writes it to the database
 func (s *StorageState) StoreTrie(ts *rtstorage.TrieState, header *types.Header) error {
 	root := ts.MustRoot()
 
-	_, _ = s.tries.LoadOrStore(root, ts.Trie())
+	s.tries.softSet(root, ts.Trie())
 
 	if _, ok := s.pruner.(*pruner.FullNode); header == nil && ok {
 		return fmt.Errorf("block cannot be empty for Full node pruner")
@@ -127,20 +126,16 @@ func (s *StorageState) TrieState(root *common.Hash) (*rtstorage.TrieState, error
 		root = &sr
 	}
 
-	st, has := s.tries.Load(*root)
-	if !has {
+	t := s.tries.get(*root)
+	if t == nil {
 		var err error
-		st, err = s.LoadFromDB(*root)
+		t, err = s.LoadFromDB(*root)
 		if err != nil {
 			return nil, err
 		}
 
-		_, _ = s.tries.LoadOrStore(*root, st)
-	}
-
-	t := st.(*trie.Trie)
-
-	if has && t.MustHash() != *root {
+		s.tries.softSet(*root, t)
+	} else if t.MustHash() != *root {
 		panic("trie does not have expected root")
 	}
 
@@ -162,7 +157,7 @@ func (s *StorageState) LoadFromDB(root common.Hash) (*trie.Trie, error) {
 		return nil, err
 	}
 
-	_, _ = s.tries.LoadOrStore(t.MustHash(), t)
+	s.tries.softSet(t.MustHash(), t)
 	return t, nil
 }
 
@@ -175,8 +170,9 @@ func (s *StorageState) loadTrie(root *common.Hash) (*trie.Trie, error) {
 		root = &sr
 	}
 
-	if t, has := s.tries.Load(*root); has && t != nil {
-		return t.(*trie.Trie), nil
+	t := s.tries.get(*root)
+	if t != nil {
+		return t, nil
 	}
 
 	tr, err := s.LoadFromDB(*root)
@@ -205,8 +201,9 @@ func (s *StorageState) GetStorage(root *common.Hash, key []byte) ([]byte, error)
 		root = &sr
 	}
 
-	if t, has := s.tries.Load(*root); has {
-		val := t.(*trie.Trie).Get(key)
+	t := s.tries.get(*root)
+	if t != nil {
+		val := t.Get(key)
 		return val, nil
 	}
 
--- a/dot/state/tries.go
+++ b/dot/state/tries.go
@@ -1,0 +1,60 @@
+// Copyright 2022 ChainSafe Systems (ON)
+// SPDX-License-Identifier: LGPL-3.0-only
+
+package state
+
+import (
+	"sync"
+
+	"github.com/ChainSafe/gossamer/lib/common"
+	"github.com/ChainSafe/gossamer/lib/trie"
+)
+
+type tries struct {
+	rootToTrie map[common.Hash]*trie.Trie
+	mapMutex   sync.RWMutex
+}
+
+func newTries(t *trie.Trie) *tries {
+	return &tries{
+		rootToTrie: map[common.Hash]*trie.Trie{
+			t.MustHash(): t,
+		},
+	}
+}
+
+// softSet sets the given trie at the given root hash
+// in the memory map only if it is not already set.
+func (t *tries) softSet(root common.Hash, trie *trie.Trie) {
+	t.mapMutex.Lock()
+	defer t.mapMutex.Unlock()
+
+	_, has := t.rootToTrie[root]
+	if has {
+		return
+	}
+
+	t.rootToTrie[root] = trie
+}
+
+func (t *tries) delete(root common.Hash) {
+	t.mapMutex.Lock()
+	defer t.mapMutex.Unlock()
+	delete(t.rootToTrie, root)
+}
+
+// get retrieves the trie corresponding to the root hash given
+// from the in-memory thread safe map.
+func (t *tries) get(root common.Hash) (tr *trie.Trie) {
+	t.mapMutex.RLock()
+	defer t.mapMutex.RUnlock()
+	return t.rootToTrie[root]
+}
+
+// len returns the current numbers of tries
+// stored in the in-memory map.
+func (t *tries) len() int {
+	t.mapMutex.RLock()
+	defer t.mapMutex.RUnlock()
+	return len(t.rootToTrie)
+}
--- a/dot/state/service_test.go
+++ b/dot/state/service_test.go
@@ -293,8 +293,8 @@ func TestService_PruneStorage(t *testing.T) {
 	time.Sleep(1 * time.Second)
 
 	for _, v := range prunedArr {
-		_, has := serv.Storage.tries.Load(v.hash)
-		require.Equal(t, false, has)
+		tr := serv.Storage.tries.get(v.hash)
+		require.Nil(t, tr)
 	}
 }
 
--- a/dot/state/storage_test.go
+++ b/dot/state/storage_test.go
@@ -5,7 +5,6 @@ package state
 
 import (
 	"math/big"
-	"sync"
 	"testing"
 	"time"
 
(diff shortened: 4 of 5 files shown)


dot/state/storage.go
```diff
@@ -71,36 +70,22 @@ func NewStorageState(db chaindb.Database, blockState *BlockState,
...
 	root := ts.MustRoot()
 
-	if s.syncing {
-		// keep only the trie at the head of the chain when syncing
-		// TODO: probably remove this when memory usage improves (#1494)
-		s.tries.Range(func(k, _ interface{}) bool {
-			s.tries.Delete(k)
-			return true
-		})
-	}
```
qdm12, line 101: this was never run and looks rather strange, so I removed it

dot/state/roottottrie.go
```diff
@@ -0,0 +1,87 @@
...
+	defer r.Unlock()
+
+	var calculatedRoot common.Hash
+	t, has := r.rootToTrie[root]
+	if has {
+		calculatedRootHash := t.MustHash()
+		if !calculatedRootHash.Equal(root) {
+			panic(fmt.Sprintf("trie does not have expected root, expected %s but got %s",
+				root, calculatedRootHash))
+		}
```
qdm12, line 69: Can I remove those panics (here and below)? Or is there a risk of a root hash mismatch here? :thinking: 

That might accelerate processing since we don't need to encode and hash the entire trie (`t.MustHash()`)

dot/node.go
```diff
@@ -177,13 +176,6 @@ func LoadGlobalNodeName(basepath string) (nodename string, err error) {
 
 // NewNode creates a new dot node from a dot node configuration
 func NewNode(cfg *Config, ks *keystore.GlobalKeystore) (*Node, error) {
-	// set garbage collection percent to 10%
-	// can be overwritten by setting the GOGC env variable, which defaults to 100
-	prev := debug.SetGCPercent(10)
-	if prev != 100 {
-		debug.SetGCPercent(prev)
-	}
-
```
noot, line 186: why is this removed? does it not affect memory usage anymore?

dot/state/roottottrie.go
noot: yeah, you can remove them, in the case where the trie gets corrupted for whatever reason and the roots don't match, the block processing higher up will fail with a state root mismatch

noot (approved):
nice cleanup!

dot/state/roottottrie.go
```diff
@@ -0,0 +1,87 @@
...
+
+	"github.com/ChainSafe/chaindb"
+	"github.com/ChainSafe/gossamer/lib/common"
+	"github.com/ChainSafe/gossamer/lib/trie"
+)
+
+type rootToTrieMap struct {
+	rootToTrie map[common.Hash]*trie.Trie
+	sync.RWMutex
+}
```
noot, line 18: I'm curious as to the advantage of this over a *sync.Map

dot/node.go
qdm12: I would like to remove it to have a clear picture. I don't really think it helps much anyway, the GC is pretty optimized. And also our memory problem is not GC related. Is there a reason we added it in the first place?

noot: the memory usage was way worse without it, but we can remove it and just set it if needed

dot/state/roottottrie.go
qdm12: Just type safety / cleaner code. That's how I found out about the empty pruned keys so I thought I would push it 😄

I also don't think it matters much since we just read and access it not in the million per second AFAIK.
If we identify it as a bottleneck later, then we can change it to a sync.Map, which is really for corner performance cases I would say. Let's not pre-pre-optimize 😉

dot/node.go
qdm12: We can always set it with `GOGC` if you need to run the node for as long as you want.
Although I guess setting to 10 will bottleneck the CPU so the program will run slower, hence the longer time perhaps? 🤔 
From my point of view, I prefer to have it run as fast as possible and crash to see if my changes fixed it or not 😄 So I guess I can also use `GOGC`, up to you I can also revert it!

dot/state/roottottrie.go
qdm12: So I removed the check when loading from memory since it's rather not possible to have it corrupt I think. I left the one when loading the trie from disk.

By the way, I ran gossamer with a print to see how often it would load the trie from disk in this function and it looks like it never does (for 30 min), is this expected?

noot: yeah, it shouldn't really ever load from disk except on start-up, so that sounds about right

noot: it's rather slow to load from disk so we don't want that when syncing

dot/node.go
noot: no, it's fine to delete it, we can just re-add it later if we want

dot/state/roottottrie.go
qdm12: Ok that makes sense. Maybe some simplification could be done, since it likely loads from disk from another code path so this code path in this function is never really run. Anyway, in another PR later maybe some day in the future.

dot/state/roottottrie.go
```diff
@@ -0,0 +1,88 @@
...
+	r.rootToTrie[root] = trie
+}
+
+func (r *rootToTrieMap) delete(root common.Hash) {
+	r.Lock()
+	defer r.Unlock()
+	delete(r.rootToTrie, root)
+}
+
+func (r *rootToTrieMap) getFromMemOrDB(root common.Hash, db chaindb.Database) (
```
timwu20, line 57: can we move this code into the caller of this function `StorageState.TrieState()`. Doesn't make a lot of sense for the in-memory `rootToTrieMap` has to worry about offline storage `chaindb.Database`.

dot/state/storage.go
```diff
@@ -30,7 +30,7 @@ func errTrieDoesNotExist(hash common.Hash) error {
 // StorageState is the struct that holds the trie, db and lock
 type StorageState struct {
 	blockState *BlockState
-	tries      *sync.Map // map[common.Hash]*trie.Trie // map of root -> trie
+	rootToTrie *rootToTrieMap
```
timwu20, line 33:
```suggestion
	tries *rootToTrieMap
```
nit: can we keep the same syntax.  I kinda like `s.tries`.

qdm12: done in d7b3f3ffe256a3aa9fa602c5ddd35a414fe177a8

dot/state/roottottrie.go
qdm12: >> Doesn't make a lot of sense...

Actually it does, I think it should be an atomic operation to get from memory + get from db + set to memory.

Otherwise we could end up with cases where

- goroutine A gets root hash XYZ from memory, it's not found
- goroutine B sets trie at root hash XYZ in memory
- goroutine A gets stored trie for root hash XYZ from disk
- goroutine A sets trie for root hash XYZ in memory

Effectively rendering the write of B erased. Probably won't happen, but better safe 😉

dot/state/roottottrie.go
qdm12: It doesn't suck at garbage collection

dot/state/roottottrie.go
timwu20: Then shouldn't `chaindb` be stored in `rootToTrieMap` instead of being passed on every call to `getFromMemOrDB`.  Also after looking at the calling code in `StorageState` it looks it already checks the map if it has the key, and will load from db in `StorageState.loadTrie`, and it only needs this atomic read in `StorageState.TrieState`.  If it were me, I'd simplify this and just add `Lock()`, `Unlock()` methods to the `rootToTrieMap`.

dot/state/roottottrie.go
```diff
@@ -0,0 +1,88 @@
...
+import (
+	"fmt"
+	"sync"
+
+	"github.com/ChainSafe/chaindb"
+	"github.com/ChainSafe/gossamer/lib/common"
+	"github.com/ChainSafe/gossamer/lib/trie"
+)
+
+type rootToTrieMap struct {
```
timwu20, line 15:
```suggestion
type trieMap struct {
```

dot/state/roottottrie.go
```diff
@@ -0,0 +1,88 @@
...
+	r.Lock()
+	defer r.Unlock()
+
+	_, exists := r.rootToTrie[root]
+	if exists {
+		return
+	}
+
+	r.rootToTrie[root] = trie
+}
```
EclesioMeloJunior, line 49: I would suggest to use the already written function because they use diferent locks for reading and writing
```suggestion
	_, exists := r.get(root)
	if exists { return }
	r.set(root, trie)
}
```

dot/state/roottottrie.go
```diff
@@ -0,0 +1,88 @@
...
+func (r *rootToTrieMap) delete(root common.Hash) {
+	r.Lock()
+	defer r.Unlock()
+	delete(r.rootToTrie, root)
+}
+
+func (r *rootToTrieMap) getFromMemOrDB(root common.Hash, db chaindb.Database) (
+	t *trie.Trie, err error) {
+	r.Lock()
+	defer r.Unlock()
```
EclesioMeloJunior, line 60:
```suggestion
```

dot/state/roottottrie.go
```diff
@@ -0,0 +1,88 @@
...
+
+	var calculatedRoot common.Hash
+	t, has := r.rootToTrie[root]
+	if has {
+		return t, nil
+	}
```
EclesioMeloJunior, line 66: use the `get` to use the inner read locks
```suggestion
	t, has := r.get(root)
	if has {
		return t, nil
	}
```

dot/state/roottottrie.go
```diff
@@ -0,0 +1,88 @@
...
+		return nil, fmt.Errorf("cannot load root %s from database: %w", root, err)
+	}
+
+	calculatedRootHash := t.MustHash()
+	if !calculatedRootHash.Equal(root) {
+		panic(fmt.Sprintf("trie does not have expected root, expected %s but got %s",
+			root, calculatedRootHash))
+	}
+
+	r.rootToTrie[calculatedRoot] = t
```
EclesioMeloJunior, line 80:
```suggestion
	r.set(calculatedRoot, t)
```

EclesioMeloJunior (changes requested):

dot/state/roottottrie.go
qdm12: It has to be atomic, this won't work (get is atomic, set is atomic but not both)

dot/state/roottottrie.go
qdm12: the whole operation has to be thread safe, this is not possible 😉

dot/state/roottottrie.go
qdm12: the whole operation has to be thread safe, this is not possible 😉

dot/state/roottottrie.go
qdm12: the whole operation has to be thread safe, this is not possible 😉

dot/state/tries.go
```diff
@@ -0,0 +1,117 @@
...
+func (t *tries) getValue(trieRoot common.Hash, key []byte) (
+	value []byte, err error) {
+	// Try to get from memory
+	t.mapMutex.RLock()
+	tr, has := t.rootToTrie[trieRoot]
+	t.mapMutex.RUnlock()
+	if has {
+		value = tr.Get(key)
+		// TODO: should we return an error if the value is not found?
+		return value, nil
```
qdm12, line 44: Question here! ❓ 

When a key is not found in the database, we return an error. However, from memory we don't, should we change that? Or return a `nil` value from the database if not found?

dot/state/roottottrie.go
qdm12: Changed it to `tries` since it now has the in memory map and a reference to the persistent database.

dot/state/roottottrie.go
qdm12: Changed pretty much a lot of code in there; now more operations are atomic and the newer `tries` structure handles the operations of memory + database.

dot/state/roottottrie.go
qdm12: I changed it to be used in more places, now the read from disk code path is indeed called. 👍 
That way I also spotted a child trie test bug with the root hash panic check when reading from disk, although I'm still working on debugging it.

dot/state/tries.go
EclesioMeloJunior: @qdm12 I was looking at the [GetFromDB](https://github.com/ChainSafe/gossamer/blob/6c64521f57f24556489cb2183bf19b1e2f02ab79/lib/trie/database.go#L280) and [getFromDB](https://github.com/ChainSafe/gossamer/blob/6c64521f57f24556489cb2183bf19b1e2f02ab79/lib/trie/database.go#L306), the functions whose retrieves the data from database, and the returned errors means problems while getting data (eg. database closed or database doesn't exists) or decoding the data (eg. incomplete or wrong encoded hashes), however, if we can retrieve and decode the hashes and even though the key is not present both functions will return nil without error, so I think we don't need to change the behavior. Please, correct me if I'm wrong

dot/state/roottottrie.go
qdm12: Moved changes regarding atomic operations in #2250

dot/state/tries.go
qdm12: Awesome thanks! Not sure why I thought it was returning an error when not found...

I also moved changes regarding atomic operations in #2250 so I'll remove that `TODO`

lib/trie/database.go
```diff
@@ -279,7 +279,7 @@ func (t *Trie) ClearPrefixFromDB(db chaindb.Database, prefix []byte) error {
 // It then reads the value from the database.
 func GetFromDB(db chaindb.Database, rootHash common.Hash, key []byte) (
 	value []byte, err error) {
-	if rootHash == EmptyHash {
+	if len(key) == 0 && rootHash == EmptyHash {
```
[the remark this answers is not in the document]
qdm12: Totally. Changed back in #2250 , not sure why I did that in the first place 🤔

dot/state/roottottrie.go
qdm12: Root hash check moved to PR #2251 (with child trie test failing for now)

EclesioMeloJunior (approved):

PR #2215, opened 2022-01-20
(tests) add integration tests into CI process

### LINKED ISSUE #2214: Create CI workflow file for integration tests
## Issue summary

- We should create a workflow test file to include the integration test execution in the CI process

- The CI file should execute the `go test` command with the `-tags integration` flag

EclesioMeloJunior:
## Changes

- Fix failures at `dot/rpc/modules` integration tests
- Create `workflows/integration-tests.yml` file

## Tests

```
go test -tags integration -v ./dot/rpc/modules/...
```

## Issues

- Closes #2214

## Primary Reviewer

- @qdm12


.github/CONTRIBUTING.md
```diff
@@ -78,6 +78,8 @@ For coding style, you may refer to the [code style](CODE_STYLE.md) document whic
 
     ⚠️ We are still using `mockery` instead of `gomock` for older mocks, so you need to run `go generate -run "mockery|mockgen" ./...` to update all the existing mocks. We are slowly migrating our mocks and test code to use `gomock`.
 
+    > To execute `//go:generate` commands that are placed at files with `// +build integration` remember to add `-tags integration` in the `go generate command` eg. `go generate -tags integration ...`
```
qdm12, line 81: The `-tags` is passed to the `go generate` command and not `mockgen`/`mockery` right?

also supernit

```suggestion
    > To execute `//go:generate` commands that are placed at files with `// +build integration` remember to add `-tags integration` in the `go generate` command eg. `go generate -tags integration ...`
```

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,76 @@
+on:
+  pull_request:
+    paths:
+      - .github/workflows/unit-tests.yml
```
qdm12:
```suggestion
      - .github/workflows/integration-tests.yml
```

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,76 @@
...
+      - "internal/**"
+      - "lib/**"
+      - "pkg/**"
+      - "tests/**"
+      - go.mod
+      - go.sum
+      - Makefile
+name: integration-tests
+env:
+  GO111MODULE: on
```
qdm12, line 18: I think we can remove these

```suggestion
```

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,76 @@
...
+
+jobs:
+  unit-tests:
```
qdm12:
```suggestion
  integration-tests:
```

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,76 @@
...
+    timeout-minutes: 60
+    strategy:
+      matrix:
+        go-version: [1.17.x]
+        platform: [ubuntu-latest]
+        packages: [
+          github.com/ChainSafe/gossamer/dot/rpc/modules
+        ]
```
qdm12, line 29: I don't think there is much point in having a matrix, what do you think? Just run the integration tests with one single `go test -tags integration ./...` command from the root?

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,76 @@
...
+          elif [ "$RUNNER_OS" == "macOS" ]; then
+            wget -P $HOME/.local/bin/ https://chainbridge.ams3.digitaloceanspaces.com/subkey-v2.0.0-macos
+            mv $HOME/.local/bin/subkey-v2.0.0-macos $HOME/.local/bin/subkey
+          else
+            echo "Subkey for $RUNNER_OS is not supported"
+            exit 1
+          fi
+
+          chmod +x $HOME/.local/bin/subkey
+          echo "$HOME/.local/bin" >> $GITHUB_PATH
```
qdm12, line 72: Do we need the subkey for integration tests?

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,76 @@
...
+
+      - name: Run unit tests
```
qdm12, line 74:
```suggestion
      - name: Run integration tests
```

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,76 @@
...
+        run: go test ${{ matrix.packages }} -timeout=30m -tags integration
```
qdm12, line 75: I'm not 100% sure but doesn't the package path have to be the last one? Maybe I'm mistaken though

dot/core/interface.go
```diff
@@ -47,6 +47,7 @@ type BlockState interface {
 	StoreRuntime(common.Hash, runtime.Instance)
 }
 
+//go:generate mockgen -destination=./mocks/storage_state_mock.go -package mocks . StorageState
 //go:generate mockery --name StorageState --structname StorageState --case underscore --keeptree
```
qdm12, line 51: Why do we have both? Also should we have the `StorageState` mock in a `_test.go` file? I'm not sure where it's used really though.

dot/rpc/modules/author_integration_test.go
```diff
@@ -176,7 +176,7 @@ func TestAuthorModule_SubmitExtrinsic_invalid_input(t *testing.T) {
 
 	res := new(ExtrinsicHashResponse)
 	err := auth.SubmitExtrinsic(nil, &ext, res)
-	require.EqualError(t, err, "could not byteify non 0x prefixed string: 0x31")
+	require.EqualError(t, err, "could not byteify non 0x prefixed string: 31")
```
qdm12, line 179: Sorry that was my fault in a previous PR 😢 Didn't check integration tests

lib/babe/build_integration_test.go
```diff
@@ -75,48 +75,9 @@ func addAuthorshipProof(t *testing.T, babeService *Service, slotNumber, epoch ui
...
-	err = ext.Sign(signature.TestKeyringPairAlice, o)
-	require.NoError(t, err)
-
-	extEnc, err := ctypes.EncodeToHexString(ext)
-	require.NoError(t, err)
-
-	return types.Extrinsic(common.MustHexToBytes(extEnc))
-}
+func createTestBlock(t *testing.T, babeService *Service, parent *types.Header, exts [][]byte,
+	slotNumber, epoch uint64) (*types.Block, Slot) {
```
qdm12, line 80: Why the leading empty line 🤔

lib/genesis/test_utils.go
```diff
@@ -100,23 +102,25 @@ func CreateTestGenesisJSONFile(t *testing.T, asRaw bool) (filename string) {
...
 // NewDevGenesisWithTrieAndHeader generates test dev genesis, genesis trie and genesis header
 func NewDevGenesisWithTrieAndHeader(t *testing.T) (*Genesis, *trie.Trie, *types.Header) {
-	gen, err := NewGenesisFromJSONRaw("../../chain/dev/genesis.json")
-	if err != nil {
-		gen, err = NewGenesisFromJSONRaw("../../../chain/dev/genesis.json")
-		require.NoError(t, err)
-	}
+	_, fullpath, _, _ := runtime.Caller(0)
+	rootDir := path.Dir(path.Dir(path.Dir(fullpath))) // same as ../../..
+	genesisPath := path.Join(rootDir, "chain/dev/genesis.json")
```
qdm12, line 120: This is risky since depending where the caller is, this won't work. Maybe implementing a search function would be nice (it's also needed when we want to move directories to `internal/` later). What I'm thinking of is to:

1. Search for directory containing `go.mod` with `module github.com/ChainSafe/gossamer` by going up everytime it's not found. That way we can find the root of the repo.
2. Search for the chain directory by walking from the root, and then use that path together with the root to have an absolute path to the desired JSON file.

This can be shared in the several exported helpers test functions in this file.

lib/runtime/test_helpers.go
```diff
@@ -161,3 +167,123 @@ func RemoveFiles(files []string) error {
 	}
 	return nil
 }
+
+// NewTestExtrinsic builds a new extrinsic using centrifuge pkg
+func NewTestExtrinsic(t *testing.T, rt Instance, genHash, blockHash common.Hash, nonce uint64, call string, args ...interface{}) string { //nolint
```
qdm12, line 172: Please split the line and remove `//nolint`.

qdm12 (changes requested):

.github/CONTRIBUTING.md
EclesioMeloJunior: right!

.github/workflows/integration-tests.yml
EclesioMeloJunior: In previous retrospectives, I remember we talked about using the Github actions matrix feature to parallelize tests as some of them take a lot of time and others just go quickly. So I don't think it's all bad to have the packages that contain integration tests being added there.

qdm12: Fair enough, although remove the platform and go-version. We will re-add them once we need them to contain multiple values. So far we never did.

.github/workflows/checks.yml
```diff
@@ -3,6 +3,7 @@ on:
     types:
       - opened
       - edited
+      - reopened
```
qdm12, line 6: Why do you precise the `types` of pull request? 🤔 Just being curious!

.github/workflows/integration-tests.yml
```diff
@@ -26,6 +26,7 @@ jobs:
         platform: [ubuntu-latest]
         packages: [
           github.com/ChainSafe/gossamer/dot/rpc/modules
+          github.com/ChainSafe/gossamer/lib/babe
```
qdm12, line 29: 💯

.github/workflows/integration-tests.yml
```diff
@@ -55,22 +56,6 @@ jobs:
...
-            echo "Subkey for $RUNNER_OS is not supported"
-            exit 1
-          fi
-
-          chmod +x $HOME/.local/bin/subkey
-          echo "$HOME/.local/bin" >> $GITHUB_PATH
-
       - name: Run unit tests
-        run: go test ${{ matrix.packages }} -timeout=30m -tags integration
+        run: go test -timeout=30m -tags integration ./...
```
qdm12, line 60: Revert this if you want to use the matrix 😉

noot (approved):

.github/workflows/checks.yml
EclesioMeloJunior: Whenever the PR is updated (e.g adjust the title) then the checks rerun again. I modify because I was in trouble with this PR title is not being updated at `Check title job`

dot/core/interface.go
EclesioMeloJunior: In fact, we don't need both, remove and kept just the mockery. Change to mockgen will need a large refactor that is not part of this PR

lib/genesis/test_utils.go
EclesioMeloJunior: `runtime.Caller(0)` with argument 0 returns the caller of `runtime.Caller` function so it will always be the current folder where the function is, however I agreed that it can be improved to be more flexibe, because right now we need to exec 3 hard coded jumps to achieve the root, I will improve given your suggestion!

timwu20 (approved):

lib/genesis/test_utils.go
```diff
@@ -100,12 +101,33 @@ func CreateTestGenesisJSONFile(t *testing.T, asRaw bool) (filename string) {
...
-func NewTestGenesisWithTrieAndHeader(t *testing.T) (*Genesis, *trie.Trie, *types.Header) {
+func getAbsolutePath(t *testing.T, p string) string {
+	t.Helper()
+
 	_, fullpath, _, _ := runtime.Caller(0)
-	rootDir := path.Dir(path.Dir(path.Dir(fullpath))) // same as ../../..
-	genesisPath := path.Join(rootDir, "chain/gssmr/genesis.json")
+	finderPath := path.Dir(fullpath)
+
+	const sarchingFor = "go.mod"
```
qdm12, line 110: typo

lib/genesis/test_utils.go
```diff
@@ -100,12 +101,33 @@ func CreateTestGenesisJSONFile(t *testing.T, asRaw bool) (filename string) {
 	return filename
 }
 
-// NewTestGenesisWithTrieAndHeader generates genesis, genesis trie and genesis header
-func NewTestGenesisWithTrieAndHeader(t *testing.T) (*Genesis, *trie.Trie, *types.Header) {
+func getAbsolutePath(t *testing.T, p string) string {
```
qdm12, line 104: maybe rename `p` I'm not sure what it is from the function signature

I guess `pathFromRepoRoot`?

qdm12 (changes requested):

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,57 @@
...
+      - "chain/**"
+      - "cmd/**"
+      - "dot/**"
+      - "internal/**"
+      - "lib/**"
+      - "pkg/**"
+      - "tests/**"
+      - go.mod
+      - go.sum
+      - Makefile
```
qdm12, line 15: I don't think the Makefile matters here.

```suggestion
```

lib/genesis/test_utils.go
```diff
@@ -98,25 +101,46 @@ func CreateTestGenesisJSONFile(t *testing.T, asRaw bool) (filename string) {
...
-		gen, err = NewGenesisFromJSONRaw("../../../chain/gssmr/genesis.json")
+func getAbsolutePath(t *testing.T, p string) string {
+	t.Helper()
+
+	_, fullpath, _, _ := runtime.Caller(0)
+	finderPath := path.Dir(fullpath)
+
+	const sarchingFor = "go.mod"
+
+	for {
```
qdm12, line 113: Limit when you reach the root perhaps? In case `go.mod` disappears on a branch or something like that.

EclesioMeloJunior: Nice, I will add a requirement at the `for` beginning
```go
const boundaries = "gossamer"

for {
    require.Contains(t, finderPath, boundaries)
```

if we reach some path outside the project then the `require` fails and exit
`"/Users/.../go/src/github.com/ChainSafe" does not contain "gossamer"`

lib/genesis/test_utils.go
```diff
@@ -98,25 +101,52 @@ func CreateTestGenesisJSONFile(t *testing.T, asRaw bool) (filename string) {
...
+
+	const searchingFor = "go.mod"
+	for {
+		filepathToCheck := path.Join(finderPath, searchingFor)
+		_, err := os.Stat(filepathToCheck)
+
+		fileNotFound := errors.Is(err, os.ErrNotExist)
+		if fileNotFound && finderPath == "/" {
+			require.FailNow(t, "cannot found project root")
+			break
```
qdm12, line 119: nit

```suggestion
			t.Fatal(t, "cannot find project root")
```

qdm12 (approved):

lib/genesis/test_utils.go
```diff
@@ -98,25 +101,51 @@ func CreateTestGenesisJSONFile(t *testing.T, asRaw bool) (filename string) {
...
+
+		fileNotFound := errors.Is(err, os.ErrNotExist)
+		if fileNotFound && finderPath == "/" {
+			t.Fatal(t, "cannot find project root")
+		}
+
+		if fileNotFound {
+			finderPath = path.Dir(finderPath)
+			continue
+		}
```
qdm12, line 124: nit in my Friday morning refactoring mood 😄 

```suggestion
		if fileNotFound {
			previousFinderPath := finderPath
			finderPath = path.Dir(finderPath)
			if finderPath == previousFinderPath {
				t.Fatal(t, "cannot find project root")
			}
			continue
		}
```

lib/genesis/test_utils.go
```diff
@@ -98,25 +101,51 @@ func CreateTestGenesisJSONFile(t *testing.T, asRaw bool) (filename string) {
...
+
 		require.NoError(t, err)
+		break
 	}
 
+	return path.Join(finderPath, pathFromRoot)
```
qdm12, line 130: Use `filepath.Join` so it works on all weird OSes

qdm12 (approved):

PR #2219, opened 2022-01-20
refactor(metrics): use default prometheus go metrics collector, and implement standard Prometheus gauges for existing metrics

timwu20:
## Changes

- Remove `dot/metrics` and the use of `ethmetrics`
- Create `internal/metrics` which starts the http server with prometheus handler
- Migrate all existing metrics to use `github.com/prometheus/client_golang/prometheus` metrics
  - All names of metrics have been revised with prometheus naming best practices applied
- Unblocks the ability to tag metrics with given labels using `prometheus.GaugeVec`, `prometheus.CounterVec`, etc

## Tests

```
go test ./...
```
Run node and ensure all go metrics and gossamer metrics are being outputted via the `/metrics` endpoint.
## Primary Reviewer

- @EclesioMeloJunior


internal/metrics/metrics.go
```diff
@@ -0,0 +1,46 @@
...
+// Start will start a dedicated metrics server at the given address.
+func Start(address string) {
+	logger.Info("Starting metrics server at http://" + address + "/metrics")
+	http.Handle("/metrics", promhttp.Handler())
+	go func() {
+		if err := http.ListenAndServe(address, nil); err != nil {
+			log.Errorf("Metrics HTTP server crashed: %s", err)
+		}
+	}()
+}
```
qdm12, line 46: You could use `internal/httpserver` which has semantic for readiness and controlled ending.

I would vouch against go routine you 'fire and forget' like this one. And also global function are not great, it's better to have a constructor with a start method on it.

You can use the [pprof server](https://github.com/ChainSafe/gossamer/blob/e4bc3757c2011a387d34baba64c8a5bb9aee1a81/internal/pprof/server.go#L27) constructor as an example:

```go
func NewServer(address string, logger httpserver.Logger) *httpserver.Server {
	handler := http.NewServeMux()
	handler.HandleFunc("/debug/pprof/", pprof.Index)
	// ...
	return httpserver.New("pprof", address, handler, logger, options...)
}
```

To use it as a service, you can use something like [this](https://github.com/ChainSafe/gossamer/blob/e4bc3757c2011a387d34baba64c8a5bb9aee1a81/internal/pprof/service.go#L25)

lib/grandpa/grandpa.go
```diff
@@ -231,13 +243,15 @@ func (s *Service) authorities() []*types.Authority {
...
-	return map[string]int64{
-		finalityGrandpaRoundMetrics: int64(s.state.round),
+func (s *Service) updateMetrics() {
+	ticker := time.NewTicker(s.Metrics.Interval)
+	for {
+		select {
+		case <-s.ctx.Done():
+			return
+		case <-ticker.C:
+			transactionPoolGauge.Set(float64(s.state.round))
```
qdm12, line 253: Shouldn't this be updated from whatever is dealing with the state round directly?

lib/grandpa/grandpa.go
```diff
@@ -231,13 +243,15 @@ func (s *Service) authorities() []*types.Authority {
...
 
-// CollectGauge returns the map between metrics label and value
-func (s *Service) CollectGauge() map[string]int64 {
-	s.roundLock.Lock()
-	defer s.roundLock.Unlock()
-
-	return map[string]int64{
-		finalityGrandpaRoundMetrics: int64(s.state.round),
+func (s *Service) updateMetrics() {
+	ticker := time.NewTicker(s.Metrics.Interval)
```
qdm12, line 247:
```suggestion
	ticker := time.NewTicker(s.Metrics.Interval)
	defer ticker.Stop()

```

tests/stress/stress_test.go
```diff
@@ -102,7 +102,7 @@ func TestSync_SingleBlockProducer(t *testing.T) {
 	require.NoError(t, err)
 	nodes = append(nodes, node)
 
-	time.Sleep(time.Second * 20)
+	time.Sleep(time.Second * 30)
```
qdm12, line 105: sus 😄

dot/state/service.go
```diff
@@ -175,9 +192,27 @@ func (s *Service) Start() error {
...
+		go s.updateMetrics()
+	}
 	return nil
 }
 
+func (s *Service) updateMetrics() {
+	ticker := time.NewTicker(s.Metrics.Interval)
+	for {
+		select {
+		case <-s.closeCh:
```
qdm12, line 206: use a context instead. Ideally pass it as argument to the function

dot/state/service.go
```diff
@@ -175,9 +192,27 @@ func (s *Service) Start() error {
...
 	go s.Storage.pruneStorage(s.closeCh)
+
+	if s.Metrics.Publish {
+		go s.updateMetrics()
+	}
 	return nil
 }
 
+func (s *Service) updateMetrics() {
+	ticker := time.NewTicker(s.Metrics.Interval)
```
qdm12, line 203:
```suggestion
	ticker := time.NewTicker(s.Metrics.Interval)
	defer ticker.Stop()

```

dot/state/service.go
```diff
@@ -175,9 +192,27 @@ func (s *Service) Start() error {
...
+	for {
+		select {
+		case <-s.closeCh:
+			return
+		case <-ticker.C:
+			transactionPoolGauge.Set(float64(s.Transaction.pool.Len()))
+			transactionQueueGauge.Set(float64(s.Transaction.queue.Len()))
+			leavesGauge.Set(float64(len(s.Block.Leaves())))
```
qdm12, line 211: Shouldn't this be set by what is dealing with these directly? I would rather dependency inject a metrics interface where it's needed that have it set values with a mutex with these ticker based goroutines.

dot/state/service.go
```diff
@@ -7,28 +7,44 @@ import (
...
 
+var (
+	transactionPoolGauge = promauto.NewGauge(prometheus.GaugeOpts{
+		Namespace: "gossamer_state_transaction",
+		Name:      "pool_total",
+		Help:      "total number of transactions in ready pool",
+	})
+	transactionQueueGauge = promauto.NewGauge(prometheus.GaugeOpts{
+		Namespace: "gossamer_state_transaction",
+		Name:      "queue_total",
```
qdm12, line 38: If I'm not mistaken namespace should be the same for the whole program no?

dot/network/service.go
```diff
@@ -289,44 +346,38 @@ func (s *Service) Start() error {
 	return nil
 }
 
-func (s *Service) collectNetworkMetrics() {
+func (s *Service) updateMetrics() {
+	ticker := time.NewTicker(s.Metrics.Interval)
```
qdm12, line 350:
```suggestion
	ticker := time.NewTicker(s.Metrics.Interval)
	defer ticker.Stop()

```

internal/metrics/metrics.go
timwu20: I agree. I pretty much copy pasta'd this code from `dot/metrics`.

dot/state/service.go
timwu20: this service uses a `closeCh` to end spawned goroutines.

dot/state/service.go
timwu20: I was migrating how it was originally done.  I'll instrument the sub types.

dot/state/service.go
timwu20: There's the auto included go metrics which are all prefixed with `go_metric_name`.  There's also system level metrics too.  I think I can add them to a registry that's already prefixed.

internal/metrics/metrics.go
```diff
@@ -0,0 +1,46 @@
...
+// Config is base config for metrics
+type Config struct {
+	Publish bool
+}
+
+// IntervalConfig for interval collection
+type IntervalConfig struct {
+	Config
+	Interval time.Duration
+}
```
EclesioMeloJunior, line 27: `interval` and `publish` looks like fields of a `Config` struct
```suggestion
type Config struct {
	Publish bool
	Interval time.Duration
}
```

dot/network/service.go
```diff
@@ -289,44 +336,24 @@ func (s *Service) Start() error {
...
-			syncedBlocks.Update(0)
-		} else {
-			syncedBlocks.Update(num.Int64())
+		select {
+		case <-s.ctx.Done():
+			return
+		case <-ticker.C:
+			peerCountGauge.Set(float64(s.host.peerCount()))
+			connectionsGauge.Set(float64(len(s.host.h.Network().Conns())))
+			nodeLatencyGauge.Set(float64(s.host.h.Peerstore().LatencyEWMA(s.host.id()).Milliseconds()))
```
qdm12, line 349: nit that's a big boy, maybe you can split it up a bit?

internal/metrics/metrics.go
qdm12: Also I'm not sure we need a constructor that much, this is just a config struct. Maybe you can add a method on the configuration like `func (c *Config) SetDefaults()` to set defaults values for unset values.

lib/grandpa/grandpa.go
```diff
@@ -1310,9 +1307,10 @@ func (s *Service) GetSetID() uint64 {
 
 // GetRound return the current round number
 func (s *Service) GetRound() uint64 {
+	// Tim: I don't think we need to lock in these case.  Reading an int will
+	// not produce a concurrent read/write panic
```
qdm12, line 1311: That depends on how `func (gm *GrandpaModule) RoundState` is called which is the only place where this function is called.

dot/sync/chain_sync.go
```diff
@@ -48,7 +50,14 @@ func (s chainSyncState) String() string {
 	}
 }
 
-var pendingBlocksLimit = maxResponseSize * 32
+var (
+	pendingBlocksLimit = maxResponseSize * 32
```
qdm12, line 54: nit a bit out of scope, but I think this could be a constant

dot/state/service.go
```diff
@@ -43,14 +38,13 @@ type Service struct {
 	Grandpa     *GrandpaState
 	closeCh     chan interface{}
 
+	PrunerCfg pruner.Config
+	Telemetry telemetry.Client
+	// Metrics   metrics.IntervalConfig
```
qdm12, line 43: nit just a reminder for that commented out metrics

dot/state/block.go
```diff
@@ -480,12 +488,16 @@ func (bs *BlockState) BestBlockHash() common.Hash {
 
 // BestBlockHeader returns the block header of the current head of the chain
 func (bs *BlockState) BestBlockHeader() (*types.Header, error) {
-	return bs.GetHeader(bs.BestBlockHash())
+	header, err := bs.GetHeader(bs.BestBlockHash())
+	if err == nil {
+		syncedBlocksGauge.Set(float64(header.Number.Int64()))
+	}
+	return header, err
```
qdm12, line 495: nit Maybe flip this around to have our most famous `err != nil` first?

And also wrapping the error

```suggestion
	if err != nil {
		return nil, fmt.Errorf("cannot get header for hash %s: %w", bs.BestBlockHash(), err)
	}

	syncedBlocksGauge.Set(float64(header.Number.Int64()))
	return header, nil
```

internal/metrics/metrics.go
timwu20: I like to keep creation of types to be centralised within the package.  I've had cases in larger codebases, where there were regressions after adding attributes to the types, but not updating every instantiation of that type outside of the package.

qdm12: I agree, although a `SetDefaults()` method would achieve the same thing. You can call it to the constructor where that config struct is passed to? As in, I guess it's the 'same difference' 😄

dot/network/service.go
```diff
@@ -289,44 +336,27 @@ func (s *Service) Start() error {
...
-		} else {
-			syncedBlocks.Update(num.Int64())
+		select {
+		case <-s.ctx.Done():
+			return
+		case <-ticker.C:
+			peerCountGauge.Set(float64(s.host.peerCount()))
+			connectionsGauge.Set(float64(len(s.host.h.Network().Conns())))
+			nodeLatencyGauge.Set(float64(
+				s.host.h.Peerstore().LatencyEWMA(s.host.id()).Milliseconds()))
```
qdm12, line 350: nit I mean to split those long lines into multiple instructions. In example:

```suggestion
			peerStore := s.host.h.Peerstore()
			latencyOfHost := peerStore.LatencyEWMA(s.host.id())
			nodeLatencyGauge.Set(float64(latencyOfHost.Milliseconds()))
```

But just nitting feel free to ignore

noot (approved):

internal/metrics/metrics.go
```diff
@@ -0,0 +1,87 @@
...
+	m.Handle("/metrics", promhttp.Handler())
+	s = &Server{
+		server: httpserver.New("metrics", address, m, logger),
+	}
+	return
+}
+
+// Start will start a dedicated metrics server at the given address.
+func (s *Server) Start(address string) (err error) {
+	logger.Info("Starting metrics server at http://" + address + "/metrics")
```
EclesioMeloJunior, line 53:
```suggestion
	logger.Infof("Starting metrics server at http://%s/metrics", address)
```

EclesioMeloJunior (approved):
just a nit

dot/network/service.go
```diff
@@ -289,44 +336,27 @@ func (s *Service) Start() error {
...
-		totalInboundStreams.Update(s.getTotalStreams(true))
-		totalOutboundStreams.Update(s.getTotalStreams(false))
-
-		num, err := s.blockState.BestBlockNumber()
-		if err != nil {
-			syncedBlocks.Update(0)
-		} else {
-			syncedBlocks.Update(num.Int64())
+		select {
+		case <-s.ctx.Done():
```
qdm12, line 344: Maybe pass the context as a first argument to that function? Even if it means calling `updateMetrics(s.ctx)`? That way we can push context 'up' in the call stack and maybe one day not have them as service struct field

dot/network/service.go
```diff
@@ -289,44 +336,27 @@ func (s *Service) Start() error {
...
+			nodeLatencyGauge.Set(float64(
+				s.host.h.Peerstore().LatencyEWMA(s.host.id()).Milliseconds()))
+			inboundBlockAnnounceStreamsGauge.Set(float64(
+				s.getNumStreams(BlockAnnounceMsgType, true)))
+			outboundBlockAnnounceStreamsGauge.Set(float64(
+				s.getNumStreams(BlockAnnounceMsgType, false)))
+			inboundGrandpaStreamsGauge.Set(float64(s.getNumStreams(ConsensusMsgType, true)))
+			outboundGrandpaStreamsGauge.Set(float64(s.getNumStreams(ConsensusMsgType, false)))
+			inboundStreamsGauge.Set(float64(s.getTotalStreams(true)))
+			outboundStreamsGauge.Set(float64(s.getTotalStreams(false)))
```
qdm12, line 358: Are these all metrics we cannot access elsewhere (due to being in another library)? 🤔

dot/node.go
```diff
@@ -341,16 +343,13 @@ func NewNode(cfg *Config, ks *keystore.GlobalKeystore) (*Node, error) {
...
-
-		go c.Start()
-
 		address := fmt.Sprintf("%s:%d", cfg.RPC.Host, cfg.Global.MetricsPort)
 		logger.Info("Enabling stand-alone metrics HTTP endpoint at address " + address)
-		metrics.PublishMetrics(address)
+		node.metricsServer = metrics.NewServer(address)
+		err := node.metricsServer.Start(address)
+		if err != nil {
+			return nil, err
```
qdm12, line 351:
```suggestion
			return nil, fmt.Errorf("cannot start metrics server: %w", err)
```

dot/node.go
```diff
@@ -424,6 +423,12 @@ func (n *Node) Stop() {
 	// stop all node services
 	n.Services.StopAll()
 	n.wg.Done()
+	if n.metricsServer != nil {
+		err := n.metricsServer.Stop()
+		if err != nil {
+			log.Errorf("%v", err)
```
qdm12, line 429: nit

```suggestion
			log.Errorf("cannot stop metrics server: %s", err)
```

dot/state/block.go
```diff
@@ -480,12 +488,17 @@ func (bs *BlockState) BestBlockHash() common.Hash {
 
 // BestBlockHeader returns the block header of the current head of the chain
 func (bs *BlockState) BestBlockHeader() (*types.Header, error) {
-	return bs.GetHeader(bs.BestBlockHash())
+	header, err := bs.GetHeader(bs.BestBlockHash())
+	if err != nil {
+		return nil, err
```
qdm12, line 493:
```suggestion
		return nil, fmt.Errorf("cannot get header of best block: %w", err)
```

internal/metrics/metrics.go
```diff
@@ -0,0 +1,88 @@
...
+	"github.com/ChainSafe/gossamer/internal/log"
+	"github.com/prometheus/client_golang/prometheus/promhttp"
+)
+
+const defaultInterval = 10 * time.Second
+
+var logger log.LeveledLogger = log.NewFromGlobal(log.AddContext("pkg", "metrics"))
+
+// IntervalConfig for interval collection
+type IntervalConfig struct {
```
qdm12, line 22: Why not call it just `Config` though?

internal/metrics/metrics.go
```diff
@@ -0,0 +1,88 @@
...
+}
+
+// NewServer is a constructor for metrics server
+func NewServer(address string) (s *Server) {
+	m := http.NewServeMux()
+	m.Handle("/metrics", promhttp.Handler())
+	s = &Server{
+		server: httpserver.New("metrics", address, m, logger),
+	}
+	return
```
qdm12, line 49: nit

```suggestion
	return &Server{
		server: httpserver.New("metrics", address, m, logger),
	}
```

internal/metrics/metrics.go
```diff
@@ -0,0 +1,88 @@
...
+
+	select {
+	case <-ready:
+		return nil
+	case err := <-s.done:
+		close(s.done)
+		if err != nil {
+			return err
+		}
+		return fmt.Errorf("metrics server exited unexpectedly")
```
qdm12, line 71: have this as a global scope exported error

internal/metrics/metrics.go
```diff
@@ -0,0 +1,88 @@
...
+func (s *Server) Stop() (err error) {
+	s.cancel()
+	select {
+	case err := <-s.done:
+		close(s.done)
+		if err != nil {
+			return err
+		}
+		return fmt.Errorf("metrics server exited unexpectedly")
+	case <-time.NewTimer(30 * time.Second).C:
```
qdm12, line 85: declare this before the `select` block and drain in the `case err := <-done` case with

```go
if !timer.Stop() {
	<-timer.C
}
```

Or at the very least stop the timer in that first case, the channel race is unlikely given the 30 wait time.

dot/network/service.go
timwu20: We have another issue to change the `Service` interface.  With that refactor, I'm going to try and remove the contexts.   I'll update this code then.

dot/network/service.go
timwu20: Yea I looked into the libp2p package and there wasn't anything in there.

internal/metrics/metrics.go
timwu20: I'd hope in most cases we wouldn't need any default metrics interval.  I'd rather the metrics be instrumented at the event.

dot/sync/chain_sync.go
timwu20: I get an `unparam` lint error.  Apparently there's a function `newDisjointBlockSet(limit int)` that only gets called with this variable.

qdm12: Update your golangci-lint? 🤔

internal/metrics/metrics.go
qdm12: Ah I get it, this is only to configure the periodic metrics scrapping, this makes sense. Thanks!

PR #2233, opened 2022-01-25
fix(dot/network): memory improvement for network buffers

qdm12:
## Changes

- [x] Use a `sync.Pool` instead of a `sizedBufferPool`. `sizedBufferPool` was using more and more memory over time for some reason. Since it was not size limited either (and we might want to avoid limiting it if our program is event-driven), a sync.Pool is more performant anyway.
- [ ] Try without sync.Pool... do we use buffers correctly??

## Primary Reviewer

- @noot


noot (approved):
works for me!! after 2.5 hours on AWS memory usage is only 4.6gb, massive improvement!

EclesioMeloJunior (approved):

timwu20 (approved):

danforbes:
Is this related to #1973?

PR #2234, opened 2022-01-25
fix(dot/sync): sync benchmark

qdm12:
## Changes

- [x] only keep 30 latest samples with a ring buffer
- [x] Inject time to benchmark functions
- [x] Remove force setting blocks to `1` when `0` so we don't have 0.2 bps when it's actually 0
- [x] Unit tests
- [x] Ran Gossamer with it

### DIFF
--- a/dot/sync/benchmark.go
+++ b/dot/sync/benchmark.go
@@ -4,19 +4,24 @@
 package sync
 
 import (
+	"container/ring"
 	"time"
 )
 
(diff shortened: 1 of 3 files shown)


noot (approved):

timwu20 (approved):

qdm12:
Do not merge, I'm changing this to use a ring buffer which is nicer memory wise.

PR #2238, opened 2022-01-26
chore(trie): improve `String()` methods for `trie.Trie` and `node.Node`

qdm12:
## Changes

Why? Because I'm going crazy trying to print a nice trie or nodes when debugging tests

- Use github.com/qdm12/gotree to have a tree string structure
- Add `StringNode` methods to node implementations
- Update node's `String` methods and `Trie`'s `String` method

## Tests

```
```

### DIFF
--- a/internal/trie/node/node.go
+++ b/internal/trie/node/node.go
@@ -3,6 +3,8 @@
 
 package node
 
+import "github.com/qdm12/gotree"
+
 // Node is a node in the trie and can be a leaf or a branch.
 type Node interface {
 	Encode(buffer Buffer) (err error) // TODO change to io.Writer
@@ -20,4 +22,5 @@ type Node interface {
 	SetGeneration(generation uint64)
 	Copy(copyChildren bool) Node
 	Type() Type
+	StringNode() (stringNode *gotree.Node)
 }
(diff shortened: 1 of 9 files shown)


internal/trie/node/node.go
```diff
@@ -20,4 +22,5 @@ type Node interface {
 	SetGeneration(generation uint64)
 	Copy(copyChildren bool) Node
 	Type() Type
+	StringNode() (stringNode *gotree.Node)
```
noot, line 25: why not return the string directly? seem like both places StringNode are used it gets turned into a string?

qdm12: So it can print the entire trie nicely. You can also see the branch node uses `StringNode()` recursively on its children nodes.

It's really a pain to otherwise re-split lines and re-assemble, prefix with the right amount of indentation and right prefixes, that's why I wrote that tiny library over Christmas 😅

An alternative would be to have it as a function taking in a node interface and do type assertions? Or keep it as a method? I'm asking since this is only used for debugging really.

noot: ok, I missed that part, seems fine to me then

noot (approved):

jimjbrettj (approved):
Cool that you did this i like it a lot, nice work

kishansagathiya (approved):

Issue #2184, opened 2022-01-11
[Core] rewrite dot/core service unit tests part1


jimjbrettj:
## Issue summary

- Rewrite dot/core service tests to increase speed and code coverage
- This is a 2 part issue, so this one will cover the first have of the function

PR #2239, opened 2022-01-26
testing(dot/core): rewrite dot/core service tests (part 1)

jimjbrettj:
## Changes

- Test half of dot/core/service.go (second half will be in separate PR). This PR includes the functions: StorageRoot, HandleBlockProduced, handleBlock, handleCodeSubstitution, handleBlockAsync, and maintainTransactionPool
- Uncommented necessary code and removed TODO from maintainTransactionPool

## Tests

```
go test ./dot/core -v -coverprofile coverage.out
```

## Issues

- #2184

## Primary Reviewer

- @noot

### DIFF
--- a/dot/core/service.go
+++ b/dot/core/service.go
@@ -5,6 +5,7 @@ package core
 
 import (
 	"context"
+	"fmt"
 	"math/big"
 	"sync"
 
@@ -310,12 +311,14 @@ func (s *Service) handleBlocksAsync() {
 			}
 
 			if err := s.handleChainReorg(prev, block.Header.Hash()); err != nil {
+				fmt.Println("reorg error")
 				logger.Warnf("failed to re-add transactions to chain upon re-org: %s", err)
 			}
 
 			s.maintainTransactionPool(block)
 		case <-s.ctx.Done():
 			return
+
 		}
 	}
 }
--- a/dot/core/service.go
+++ b/dot/core/service.go
@@ -55,7 +55,8 @@ type Service struct {
 	codeSubstitutedState CodeSubstitutedState
 
 	// Keystore
-	keys *keystore.GlobalKeystore
+	keys        *keystore.GlobalKeystore
+	newInstance newWasmerInstanceFunc
 }
 
 // Config holds the configuration for the core Service.
@@ -120,6 +121,7 @@ func NewService(cfg *Config) (*Service, error) {
 		codeSubstitute:       cfg.CodeSubstitutes,
 		codeSubstitutedState: cfg.CodeSubstitutedState,
 		digestHandler:        cfg.DigestHandler,
+		newInstance:          wasmer.NewInstance,
 	}
 
 	return srv, nil
@@ -250,12 +252,11 @@ func (s *Service) handleBlock(block *types.Block, state *rtstorage.TrieState) er
 }
 
 func (s *Service) handleCodeSubstitution(hash common.Hash, state *rtstorage.TrieState) error {
-	return s.handleCodeSubstituionWithWasmerInstance(hash, state, wasmer.NewInstance)
+	return s.handleCodeSubstituionWithWasmerInstance(hash, state)
 }
 func (s *Service) handleCodeSubstituionWithWasmerInstance(
 	hash common.Hash,
 	state *rtstorage.TrieState,
-	newInstance newWasmerInstanceFunc,
 ) error {
 	value := s.codeSubstitute[hash]
 	if value == "" {
@@ -288,7 +289,7 @@ func (s *Service) handleCodeSubstituionWithWasmerInstance(
 		cfg.Role = 4
 	}
 
-	next, err := newInstance(code, cfg)
+	next, err := s.newInstance(code, cfg)
 	if err != nil {
 		return err
 	}
--- a/dot/core/service.go
+++ b/dot/core/service.go
@@ -258,7 +258,7 @@ func (s *Service) handleCodeSubstitution(hash common.Hash, state *rtstorage.Trie
 	if len(code) == 0 {
 		return ErrEmptyRuntimeCode
 	}
-	
+
 	rt, err := s.blockState.GetRuntime(&hash)
 	if err != nil {
 		return err
--- a/dot/core/service.go
+++ b/dot/core/service.go
@@ -231,7 +231,7 @@ func (s *Service) handleBlock(block *types.Block, state *rtstorage.TrieState) er
 	}
 
 	// check if there was a runtime code substitution
-	if err := s.handleCodeSubstitution(block.Header.Hash(), state); err != nil {
+	if err := s.handleCodeSubstitution(block.Header.Hash(), state, wasmer.NewInstance); err != nil {
 		logger.Criticalf("failed to substitute runtime code: %s", err)
 		return err
 	}
@@ -249,10 +249,10 @@ func (s *Service) handleBlock(block *types.Block, state *rtstorage.TrieState) er
 	return nil
 }
 
-func (s *Service) handleCodeSubstitution(hash common.Hash, state *rtstorage.TrieState) error {
-	return s._handleCodeSubstitution(hash, state, wasmer.NewInstance)
-}
-func (s *Service) _handleCodeSubstitution(
+//func (s *Service) handleCodeSubstitution(hash common.Hash, state *rtstorage.TrieState) error {
+//	return s.handleCodeSubstitution(hash, state, wasmer.NewInstance)
+//}
+func (s *Service) handleCodeSubstitution(
 	hash common.Hash,
 	state *rtstorage.TrieState,
 	newInstance newWasmerInstanceFunc,
--- a/dot/core/service.go
+++ b/dot/core/service.go
@@ -189,7 +189,7 @@ func (s *Service) HandleBlockProduced(block *types.Block, state *rtstorage.TrieS
 
 func (s *Service) handleBlock(block *types.Block, state *rtstorage.TrieState) error {
 	if block == nil || state == nil {
-		return errNilBlockHandlerParameter
+		return ErrNilBlockHandlerParameter
 	}
 
 	// store updates state trie nodes in database
--- a/dot/core/service.go
+++ b/dot/core/service.go
@@ -410,20 +410,20 @@ func (s *Service) maintainTransactionPool(block *types.Block) {
 	// re-validate transactions in the pool and move them to the queue
 	txs := s.transactionState.PendingInPool()
 	for _, tx := range txs {
+		// get the best block corresponding runtime
 		rt, err := s.blockState.GetRuntime(nil)
 		if err != nil {
 			logger.Warnf("failed to get runtime to re-validate transactions in pool: %s", err)
 			continue
 		}
 
-		val, err := rt.ValidateTransaction(tx.Extrinsic)
+		txnValidity, err := rt.ValidateTransaction(tx.Extrinsic)
 		if err != nil {
-			// failed to validate tx, remove it from the pool or queue
 			s.transactionState.RemoveExtrinsic(tx.Extrinsic)
 			continue
 		}
 
-		tx = transaction.NewValidTransaction(tx.Extrinsic, val)
+		tx = transaction.NewValidTransaction(tx.Extrinsic, txnValidity)
 
 		// Err is only thrown if tx is already in pool, in which case it still gets removed
 		h, _ := s.transactionState.Push(tx)
--- a/dot/core/service.go
+++ b/dot/core/service.go
@@ -31,6 +31,8 @@ var (
 // QueryKeyValueChanges represents the key-value data inside a block storage
 type QueryKeyValueChanges map[string]string
 
+type newWasmerInstanceFunc func(code []byte, cfg *wasmer.Config) (instance *wasmer.Instance, err error)
+
 // Service is an overhead layer that allows communication between the runtime,
 // BABE session, and network service. It deals with the validation of transactions
 // and blocks by calling their respective validation functions in the runtime.
@@ -248,12 +250,12 @@ func (s *Service) handleBlock(block *types.Block, state *rtstorage.TrieState) er
 }
 
 func (s *Service) handleCodeSubstitution(hash common.Hash, state *rtstorage.TrieState) error {
-	return s._handleCodeSubstitution(hash, state, wasmer.NewInstance)
+	return s.handleCodeSubstituionWithWasmerInstance(hash, state, wasmer.NewInstance)
 }
-func (s *Service) _handleCodeSubstitution(
+func (s *Service) handleCodeSubstituionWithWasmerInstance(
 	hash common.Hash,
 	state *rtstorage.TrieState,
-	newInstance func(code []byte, cfg *wasmer.Config) (*wasmer.Instance, error),
+	newInstance newWasmerInstanceFunc,
 ) error {
 	value := s.codeSubstitute[hash]
 	if value == "" {
--- a/dot/core/service.go
+++ b/dot/core/service.go
@@ -250,9 +250,9 @@ func (s *Service) handleBlock(block *types.Block, state *rtstorage.TrieState) er
 }
 
 func (s *Service) handleCodeSubstitution(hash common.Hash, state *rtstorage.TrieState) error {
-	return s.handleCodeSubstitutionWithWasmerInstance(hash, state, wasmer.NewInstance)
+	return s._handleCodeSubstitution(hash, state, wasmer.NewInstance)
 }
-func (s *Service) handleCodeSubstitutionWithWasmerInstance(
+func (s *Service) _handleCodeSubstitution(
 	hash common.Hash,
 	state *rtstorage.TrieState,
 	newInstance newWasmerInstanceFunc,
--- a/dot/core/service.go
+++ b/dot/core/service.go
@@ -250,9 +250,9 @@ func (s *Service) handleBlock(block *types.Block, state *rtstorage.TrieState) er
 }
 
 func (s *Service) handleCodeSubstitution(hash common.Hash, state *rtstorage.TrieState) error {
-	return s.handleCodeSubstituionWithWasmerInstance(hash, state, wasmer.NewInstance)
+	return s.handleCodeSubstitutionWithWasmerInstance(hash, state, wasmer.NewInstance)
 }
-func (s *Service) handleCodeSubstituionWithWasmerInstance(
+func (s *Service) handleCodeSubstitutionWithWasmerInstance(
 	hash common.Hash,
 	state *rtstorage.TrieState,
 	newInstance newWasmerInstanceFunc,
--- a/dot/core/service.go
+++ b/dot/core/service.go
@@ -258,7 +258,7 @@ func (s *Service) handleCodeSubstitution(hash common.Hash, state *rtstorage.Trie
 	if len(code) == 0 {
 		return ErrEmptyRuntimeCode
 	}
-
+	
 	rt, err := s.blockState.GetRuntime(&hash)
 	if err != nil {
 		return err
--- a/dot/core/service.go
+++ b/dot/core/service.go
@@ -316,7 +316,6 @@ func (s *Service) handleBlocksAsync() {
 			s.maintainTransactionPool(block)
 		case <-s.ctx.Done():
 			return
-
 		}
 	}
 }
--- a/dot/core/service.go
+++ b/dot/core/service.go
@@ -31,7 +31,7 @@ var (
 // QueryKeyValueChanges represents the key-value data inside a block storage
 type QueryKeyValueChanges map[string]string
 
-type newWasmerInstanceFunc func(code []byte, cfg *wasmer.Config) (instance *wasmer.Instance, err error)
+type wasmerInstanceFunc func(code []byte, cfg *wasmer.Config) (instance *wasmer.Instance, err error)
 
 // Service is an overhead layer that allows communication between the runtime,
 // BABE session, and network service. It deals with the validation of transactions
@@ -249,13 +249,10 @@ func (s *Service) handleBlock(block *types.Block, state *rtstorage.TrieState) er
 	return nil
 }
 
-//func (s *Service) handleCodeSubstitution(hash common.Hash, state *rtstorage.TrieState) error {
-//	return s.handleCodeSubstitution(hash, state, wasmer.NewInstance)
-//}
 func (s *Service) handleCodeSubstitution(
 	hash common.Hash,
 	state *rtstorage.TrieState,
-	newInstance newWasmerInstanceFunc,
+	instance wasmerInstanceFunc,
 ) error {
 	value := s.codeSubstitute[hash]
 	if value == "" {
@@ -288,7 +285,7 @@ func (s *Service) handleCodeSubstitution(
 		cfg.Role = 4
 	}
 
-	next, err := newInstance(code, cfg)
+	next, err := instance(code, cfg)
 	if err != nil {
 		return err
 	}
--- a/dot/core/service.go
+++ b/dot/core/service.go
@@ -248,6 +248,13 @@ func (s *Service) handleBlock(block *types.Block, state *rtstorage.TrieState) er
 }
 
 func (s *Service) handleCodeSubstitution(hash common.Hash, state *rtstorage.TrieState) error {
+	return s._handleCodeSubstitution(hash, state, wasmer.NewInstance)
+}
+func (s *Service) _handleCodeSubstitution(
+	hash common.Hash,
+	state *rtstorage.TrieState,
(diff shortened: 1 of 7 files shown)


dot/core/service_test.go
```diff
@@ -0,0 +1,616 @@
...
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			s := tt.service
+			s.handleBlocksAsync()
+		})
+	}
+}
+
+func TestCleanup(t *testing.T) {
```
timwu20, line 613: what is this for?

jimjbrettj: The tests need to generate and remove wasm files, but cant have a TestMain function to handle it since it already exists in this packages integration tests. My attempt of a work around was to have a method to generate the wasm then this TestCleanup function to remove the generated wasm.

noot: maybe just name it `cleanup` then? since it's not really a test

noot (approved):

dot/core/service_test.go
jimjbrettj: Will it run automatically still? If not should I just call it at the end of the last test?

jimjbrettj: I can make this change, but then if any tests are added whoever adds them will have to remember to move the call to cleanup() to the last test run. @noot is this ideal or should I leave as is?

noot: I think it's better to call it at the end of every test that uses it in a t.Cleanup(), just to make it clearer when it's run

jimjbrettj: So for every test I should generate and take down the wasm? Sounds cleaner but would definitely add a decent amount of time to the tests

dot/core/service_test.go
```diff
@@ -0,0 +1,616 @@
...
+			err = s.handleCodeSubstitution(tt.args.hash, tt.args.state)
+			assert.ErrorIs(t, err, tt.expErr)
+			if tt.expErr != nil {
+				assert.EqualError(t, err, tt.expErrMsg)
+			}
+		})
+	}
+}
+
+func TestService_handleBlock(t *testing.T) {
```
qdm12, line 188: Improve readability with 2 ways:

### Mocks configuration changing a lot from a test case to another

⚠️ this is usually a code smell showing the production function being tested is too long/complex.

BUT it does happen. This is the case here.

We don't want to have huge mocks definition away from a particular test case also away from the actual subtest testing code, this hurts readability. We can't either define mocks from test cases in subtests since this would need a bunch of structs and if blocks in the subtest, making unreadable as well.

The solution is to have just subtests **without** test cases slices or for loop, so everything is locally scoped despite a bit more code duplication.

Example:

```go
t.Run("store trie error", func(t *testing.T) {
	mockStorageStateErr := NewMockStorageState(ctrl)
	mockStorageStateErr.EXPECT().StoreTrie(trieState, &block.Header).Return(errTestDummyError)
	// ...
	s := &Service{storageState: mockStorageStateErr}
	err := s.handleBlock(block, state)
	assert.ErrorIs(t, err, expErr)
	if expErr != nil {
		assert.EqualError(t, err, expErrMsg)
	}
})
```

### Mocks configuration similar from a test case to another

Then configure the mocks in the subtest, and have its expected args / return values in the test cases slice.

dot/core/service.go
```diff
@@ -411,23 +410,24 @@ func (s *Service) maintainTransactionPool(block *types.Block) {
...
+		val, err := rt.ValidateTransaction(tx.Extrinsic)
+		if err != nil {
+			// failed to validate tx, remove it from the pool or queue
+			s.transactionState.RemoveExtrinsic(tx.Extrinsic)
+			continue
+		}
+
+		tx = transaction.NewValidTransaction(tx.Extrinsic, val)
+
+		// Err is only thrown if tx is already in pool, in which case it still gets removed
```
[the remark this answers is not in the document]
jimjbrettj: This comment already existed, but I believe what was meant is that the only possible error thrown in this case would be if the txn is already in the pool, in which case the behavior is the same

jimjbrettj:
https://github.com/ChainSafe/gossamer/pull/2239#discussion_r803778798

@qdm12 It takes a few seconds, maybe 5-10 on my machine, to generate the wasm. This is why I was skeptical to construct/clean up for every test even though it is cleaner.

I'm actually not sure why we remove them, maybe just due to size? Maybe @noot or @timwu20 could provide more insight on this.

dot/core/service_test.go
jimjbrettj: I like this. I think this would be a good "unit test standard" to help out tests be more consistent across the codebase

dot/core/service_test.go
```diff
@@ -0,0 +1,651 @@
...
+var errTestDummyError = errors.New("test dummy error")
+var testWasmPaths []string
+
+func TestGenerateWasm(t *testing.T) {
+	wasmFilePaths, err := runtime.GenerateRuntimeWasmFile()
+	require.NoError(t, err)
+	testWasmPaths = wasmFilePaths
+}
+
+func Test_Service_StorageRoot(t *testing.T) {
```
timwu20, line 39: why is it when I run this test on it's own it fails?
```
go test -timeout 30s -run ^Test_Service_StorageRoot$ github.com/ChainSafe/gossamer/dot/core -v

=== RUN   Test_Service_StorageRoot
=== RUN   Test_Service_StorageRoot/nil_storage_state
=== PAUSE Test_Service_StorageRoot/nil_storage_state
=== RUN   Test_Service_StorageRoot/storage_trie_state_error
=== PAUSE Test_Service_StorageRoot/storage_trie_state_error
=== RUN   Test_Service_StorageRoot/storage_trie_state_ok
=== PAUSE Test_Service_StorageRoot/storage_trie_state_ok
=== CONT  Test_Service_StorageRoot/nil_storage_state
=== CONT  Test_Service_StorageRoot/storage_trie_state_ok
=== CONT  Test_Service_StorageRoot/storage_trie_state_error
=== CONT  Test_Service_StorageRoot/nil_storage_state
    /Users/timothywu/dev/gossamer/dot/core/service.go:148: Unexpected call to *core.MockStorageState.TrieState([<nil>]) at /Users/timothywu/dev/gossamer/dot/core/service.go:148 because:
        expected call at /Users/timothywu/dev/gossamer/dot/core/service_test.go:80 has already been called the max number of times
=== CONT  Test_Service_StorageRoot/storage_trie_state_ok
    /Users/timothywu/dev/gossamer/dot/core/controller.go:137: missing call(s) to *core.MockStorageState.TrieState(is nil) /Users/timothywu/dev/gossamer/dot/core/service_test.go:80
    /Users/timothywu/dev/gossamer/dot/core/controller.go:137: aborting test due to missing call(s)
--- FAIL: Test_Service_StorageRoot (0.00s)
    --- FAIL: Test_Service_StorageRoot/nil_storage_state (0.00s)
    --- PASS: Test_Service_StorageRoot/storage_trie_state_error (0.00s)
    --- FAIL: Test_Service_StorageRoot/storage_trie_state_ok (0.00s)
FAIL
FAIL	github.com/ChainSafe/gossamer/dot/core	0.275s
```

timwu20 (changes requested):

dot/core/service_test.go
jimjbrettj: hmm odd, this works for me as well

dot/core/service_test.go
```diff
@@ -0,0 +1,651 @@
...
+			if tt.expErr != nil {
+				assert.EqualError(t, err, tt.expErrMsg)
+			}
+			assert.Equal(t, tt.exp, res)
+		})
+	}
+}
+
+func Test_Service_handleCodeSubstitution(t *testing.T) {
+	testRuntime, err := os.ReadFile(runtime.POLKADOT_RUNTIME_FP)
```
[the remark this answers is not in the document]
jimjbrettj: @timwu20 Could you elaborate on this a bit more, I don't quite understand. Doesn't the Service interface provide a runtime through the config? The only usage of wasmer.NewInstance(code, cfg) I see is in handleCodeSubstitution() which I would think we need since its a change to the runtime. Unless you are saying there is a better way of going about that?

dot/core/service_test.go
```diff
@@ -0,0 +1,651 @@
...
+			exp: common.Hash{0x3, 0x17, 0xa, 0x2e, 0x75, 0x97, 0xb7, 0xb7, 0xe3, 0xd8, 0x4c, 0x5, 0x39, 0x1d, 0x13, 0x9a,
+				0x62, 0xb1, 0x57, 0xe7, 0x87, 0x86, 0xd8, 0xc0, 0x82, 0xf2, 0x9d, 0xcf, 0x4c, 0x11, 0x13, 0x14},
+			retTrieState: ts,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+			s := tt.service
+			if tt.name != "nil storage state" {
```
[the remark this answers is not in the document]
timwu20: IMO this is really ugly and difficult to read.  Changing the name of the test would result in test failure? I'd consider not even making this test run in parallel in this case.  The parent `Test_Service_StorageRoot` can run in parallel yes, but I don't see the harm in running the subtests consecutively.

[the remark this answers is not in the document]
jimjbrettj: @timwu20 yup I have changed it to check a parameter like quentin had suggested. Besides making the subtests no longer run in parallel, is there anything else I need to do to clean this up?

dot/core/service_test.go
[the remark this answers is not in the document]
timwu20: create a new `_handleCodeSubstitution` that takes in an extra parameter `func (code []byte, cfg *wasmer.Config) (*wasmer.Instance, error)` and have `handleCodeSubstitution` call that method instead using `wasmer.NewInstance` by default.  In your unit tests you just provide your own `*runtime.Instance`

dot/core/service_test.go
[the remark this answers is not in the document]
jimjbrettj: Maybe this will remove flakiness

dot/core/service_test.go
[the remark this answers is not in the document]
jimjbrettj: ohhh okay I misunderstood I see. Makes a lot of sense, will implement

dot/core/service_test.go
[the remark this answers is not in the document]
qdm12: I don't see how running tests in parallel or not is relevant here though?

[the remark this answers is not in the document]
timwu20: I discussed this with @jimjbrettj  over zoom.  With the introduction of gomock controllers and it's reliance on a locally scoped `t *testing.T` we may not be able to use the IDE code generation for tests if we want the subtests to run in parallel.  

Yes detecting race conditions would be better if each subtest is run in parallel, but the tradeoff is we can't use the code generated tests with predefined state if we are using gomock controllers.  In the past I just created my own mock implementations of interfaces that did not rely on `t *testing.T`.  Rather, just emit testing errors when assertions fail or through inspection of the mock implementation.

In this case specifically, it may just make more sense to avoid using the slice of anonymous test structs and just adopt the explicit way of writing out each subtest.  We should avoid repeating code over and over again, which I have discussed, and it will allow us to keep using gomock controllers and run the subtests in parallel.

dot/core/service_test.go
timwu20: this should be resolved by removing the reliance on calling `wasmer.NewInstance`.

dot/core/service_test.go
```diff
@@ -0,0 +1,627 @@
...
+
+	"github.com/golang/mock/gomock"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+var errTestDummyError = errors.New("test dummy error")
+var testWasmPaths []string
+
+func Test_Service_StorageRoot(t *testing.T) {
```
timwu20, line 31: this test can be run in parallel

dot/core/service_test.go
```diff
@@ -0,0 +1,627 @@
...
+			assert.ErrorIs(t, err, tt.expErr)
+			if tt.expErr != nil {
+				assert.EqualError(t, err, tt.expErrMsg)
+			}
+			assert.Equal(t, tt.exp, res)
+		})
+	}
+}
+
+func newTestInstance(code []byte, cfg *wasmer.Config) (*wasmer.Instance, error) {
```
timwu20, line 88: nit: move this into the `Test_Service_StorageRoot`

dot/core/service_test.go
```diff
@@ -0,0 +1,627 @@
...
+	return &wasmer.Instance{}, nil
+}
+
+func Test_Service_handleCodeSubstitution(t *testing.T) {
+	testRuntime := []byte{21}
```
timwu20, line 93: add `t.Parallel` to the parent as well

dot/core/service_test.go
```diff
@@ -0,0 +1,627 @@
...
+			codeSubstitute:       testCodeSubstitute,
+			blockState:           mockBlockState,
+			codeSubstitutedState: mockCodeSubState,
+		}
+		err := s._handleCodeSubstitution(blockHash, nil, newTestInstance)
+		assert.NoError(t, err)
+	})
+}
+
+func Test_Service_handleBlock(t *testing.T) {
```
timwu20, line 188: add `t.Parallel` to the parent as well

dot/core/service_test.go
```diff
@@ -0,0 +1,627 @@
...
+		}
+		err = s.handleBlock(&block, trieState)
+		assert.ErrorIs(t, err, expErr)
+		if expErr != nil {
+			assert.EqualError(t, err, expErrMsg)
+		}
+	})
+}
+
+func Test_Service_HandleBlockProduced(t *testing.T) {
```
timwu20, line 389: add `t.Parallel` to the parent as well

dot/core/service_test.go
jimjbrettj: yup yup it is no longer an issue

dot/core/service_test.go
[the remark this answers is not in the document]
qdm12: > we may not be able to use the IDE code generation for tests if we want the subtests to run in parallel.

It's not necessarily about running in parallel. If one subtest fails because of a mock assertion failing, it might fail the parent test since it would take `t` from the parent test, but would run within the subtest. I guess it's *alriiight* to share a mock controller  at the parent test, although it's not ideal either. For test generation, you could use Github copilot it does help quite a bit to have a good middleground between manual and automated test writing.

> Rather, just emit testing errors when assertions fail or through inspection of the mock implementation.

That's the same problem though, since it would fail the parent test if one of the subtest fails for example. This is the same as passing `t` once to a subtest-local mock controller.

> In this case specifically, it may just make more sense to avoid using the slice of anonymous test structs and just adopt the explicit way of writing out each subtest. 

Agree depending on the case. If mocks configuration is quite different from one test case to another (i.e. a mock depending on another mock, for certain test cases), then yes dumb subtests without for loop / test case slices is best. Otherwise, configure your mock within the subtest. I wrote [this comment](https://github.com/ChainSafe/gossamer/pull/2239#discussion_r803765992) which sums it up from my perspective.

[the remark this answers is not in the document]
qdm12: Created https://github.com/golang/mock/issues/625 if that interests anyone 😄
But yes, ultimately, controller construction within the subtest is pretty good for local scoping / readability / parallel tests.

dot/core/service_test.go
```diff
@@ -0,0 +1,588 @@
...
+		}
+
+		extrinsic := types.Extrinsic{21}
+		vt := transaction.NewValidTransaction(extrinsic, validity)
+
+		ctrl := gomock.NewController(t)
+		runtimeMock := new(mocksruntime.Instance)
+		runtimeMock.On("ValidateTransaction", types.Extrinsic{21}).Return(nil, errTestDummyError)
+		mockTxnState := NewMockTransactionState(ctrl)
+		mockTxnState.EXPECT().RemoveExtrinsic(types.Extrinsic{21}).MaxTimes(2)
```
qdm12, line 444: perhaps?

```suggestion
		mockTxnState.EXPECT().RemoveExtrinsic(types.Extrinsic{21}).Times(2)
```

dot/core/service.go
```diff
@@ -249,6 +248,13 @@ func (s *Service) handleBlock(block *types.Block, state *rtstorage.TrieState) er
 }
 
 func (s *Service) handleCodeSubstitution(hash common.Hash, state *rtstorage.TrieState) error {
+	return s._handleCodeSubstitution(hash, state, wasmer.NewInstance)
```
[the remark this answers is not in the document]
jimjbrettj: man I just wanted to pretend I was a solidity programmer!

dot/core/service.go
```diff
@@ -249,6 +248,13 @@ func (s *Service) handleBlock(block *types.Block, state *rtstorage.TrieState) er
...
+}
+func (s *Service) _handleCodeSubstitution(
+	hash common.Hash,
+	state *rtstorage.TrieState,
+	newInstance func(code []byte, cfg *wasmer.Config) (*wasmer.Instance, error),
```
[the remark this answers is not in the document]
jimjbrettj: I'm thinking we should leave as is so that handleCodeSubstitution uses wasmer.NewInstance by default without requiring us to set it manually as a field of Service. If you really think I should change I can, but I think since its only used in the one function and even that's just for testing that its easier to leave as is.

jimjbrettj:
updated!

dot/core/service.go
```diff
@@ -53,7 +55,8 @@ type Service struct {
 	codeSubstitutedState CodeSubstitutedState
 
 	// Keystore
-	keys *keystore.GlobalKeystore
+	keys        *keystore.GlobalKeystore
+	newInstance newWasmerInstanceFunc
```
timwu20, line 59: why did you add this as a `Service` attribute? It's not used anywhere else I believe, so I think it should be localized to `handleCodeSubstitution` as a parameter.

timwu20 (changes requested):

dot/core/service.go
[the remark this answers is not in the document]
timwu20: I disagree with storing this in `Service`. IMO doesn't make a lot of sense that this is part of the "state" of `Service`.  This is just to make `handleCodeSubsitution` more testable and it should be localised to that function.

dot/core/service.go
timwu20: I think we should change it back. I added a comment in the resolved conversation. https://github.com/ChainSafe/gossamer/pull/2239#discussion_r810232906

dot/core/service_test.go
qdm12: Added in https://github.com/ChainSafe/gossamer/pull/2311

dot/core/service_test.go
qdm12: Is it a flaky test still now?

dot/core/service_test.go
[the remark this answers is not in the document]
qdm12: Added some code style section for mocks configuration in subtests at https://github.com/ChainSafe/gossamer/pull/2311

dot/core/service.go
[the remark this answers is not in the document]
qdm12: I get your point although having 2 `handleCodeSubstitution` is kind of ugly too.

dot/core/service.go
```diff
@@ -249,6 +252,12 @@ func (s *Service) handleBlock(block *types.Block, state *rtstorage.TrieState) er
 }
 
 func (s *Service) handleCodeSubstitution(hash common.Hash, state *rtstorage.TrieState) error {
+	return s.handleCodeSubstituionWithWasmerInstance(hash, state)
+}
+func (s *Service) handleCodeSubstituionWithWasmerInstance(
```
qdm12, line 257:
```suggestion
func (s *Service) handleCodeSubstitutionWithWasmerInstance(
```

dot/core/service.go
```diff
@@ -249,6 +252,12 @@ func (s *Service) handleBlock(block *types.Block, state *rtstorage.TrieState) er
 }
 
 func (s *Service) handleCodeSubstitution(hash common.Hash, state *rtstorage.TrieState) error {
+	return s.handleCodeSubstituionWithWasmerInstance(hash, state)
```
qdm12, line 255:
```suggestion
	return s.handleCodeSubstitutionWithWasmerInstance(hash, state)
```

qdm12 (approved):

dot/core/service.go
```diff
@@ -252,11 +250,12 @@ func (s *Service) handleBlock(block *types.Block, state *rtstorage.TrieState) er
 }
 
 func (s *Service) handleCodeSubstitution(hash common.Hash, state *rtstorage.TrieState) error {
-	return s.handleCodeSubstituionWithWasmerInstance(hash, state)
+	return s.handleCodeSubstitutionWithWasmerInstance(hash, state, wasmer.NewInstance)
 }
-func (s *Service) handleCodeSubstituionWithWasmerInstance(
+func (s *Service) handleCodeSubstitutionWithWasmerInstance(
```
timwu20, line 255: can you change it back to `_handleCodeSubstitution`.  It's understood that you're doing this to make the `handleCodeSubstitution` more testable.  The new name implies it's doing something different and is called by other receiver methods.

dot/core/service.go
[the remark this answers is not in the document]
qdm12: Conclusion:

- pass it as function argument so it's more explicit and localized
- have 2 similarly-named functions which is not too pretty, but worth the point above!

dot/core/service.go
jimjbrettj: will do!

timwu20 (approved):

dot/core/service.go
```diff
@@ -249,6 +250,13 @@ func (s *Service) handleBlock(block *types.Block, state *rtstorage.TrieState) er
 }
 
 func (s *Service) handleCodeSubstitution(hash common.Hash, state *rtstorage.TrieState) error {
+	return s._handleCodeSubstitution(hash, state, wasmer.NewInstance)
+}
+func (s *Service) _handleCodeSubstitution(
```
EclesioMeloJunior, line 255: I think we don't need `_` for a helper or private functions, you could split the `_handleCodeSubstitution` into `hasCodeSubstitue` and `substitute` then you call them sequentially inside `handleCodeSubstitution` like:

```go
func (s *Service) handleCodeSubstitution ... {
    if hasCodeSubstitue(hash) {
        applySubstitute(hash, state, instance)
    }
}
```

dot/core/service_test.go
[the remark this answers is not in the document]
qdm12: Tim found a new way to do mocks with a test case slice/map + for loop and subtests. However I don't think it works when a mock needs to return a mock for example.

1. For each mock, add a mock constructor field in your test case with signature `func(ctrl *gomock.Controller) YourInterface`
2. Configure your mock in the test case by defining the function body
3. Call that function from your test case in the subtest, passing it the subtest's Gomock controller as argument

Again, if you need to i.e. return a mock from another mock, this method doesn't seem to work and in this case you need to use one of the two options mentioned in the codestyle document.

dot/core/service.go
jimjbrettj: what do you think @timwu20 @qdm12 ?

dot/core/service.go
```diff
@@ -249,6 +250,13 @@ func (s *Service) handleBlock(block *types.Block, state *rtstorage.TrieState) er
 }
 
 func (s *Service) handleCodeSubstitution(hash common.Hash, state *rtstorage.TrieState) error {
+	return s._handleCodeSubstitution(hash, state, wasmer.NewInstance)
```
noot, line 253: whats the point of this? also `_` prefixed funcs are not really a thing in go afaik

noot: oh its so you can inject the function during tests i guess?

noot:
the code logic essentially looks fine, i would double check that the extrinsic you're submitting in the stress test `TestSync_SubmitExtrinsic` is the same version and such as in the unit tests here. however it should be a valid transaction, as the `SubmitExtrinsic` function validates it, and will return error if it's invalid, so not entirely sure how your changes would effect that :/ 

also, does `Test_SubmitAndWatchExtrinsic` (other stress test) pass for you on this branch? if so, i think we can skip `TestSync_SubmitExtrinsic`, as `Test_SubmitAndWatchExtrinsic` looks to be an improved version of it

dot/core/service.go
jimjbrettj: yup yup exactly! Makes it so I dont have to generate a valid wasm instance for the mocked unit tests so it goes a lot faster (no more need to generate wasm files)

dot/core/service.go
jimjbrettj: @EclesioMeloJunior updated!

EclesioMeloJunior (approved):

Issue #2098, opened 2021-12-02
gossamer should not ban other gossamer nodes


noot:
## Task summary

- on the devnet, if you add a log in the `peerset` `reportPeer` function to check if a peer reaches the `BannedThresholdValue`, you will see that this ends up happening quite often.
- gossamer should not be banning other gossamer nodes, so either there is an issue with the scoring, or gossamer nodes are performing some "bad" behaviour that needs to be fixed

timwu20:
Was unable to reproduce this issue on current development branch.  Block production fails after the epoch transition. The actual epoch varies from run to run.  Will pick this back up after @noot finishes her refactor of BABE.

kishansagathiya:
Was able to replicate this in one of my branches. I saw some 3009 negative reputation changes, out of which 2851 were because of duplicate gossip messages. 
I believe making sure we don't send duplicate gossip message could fix this problem.

kishansagathiya:
Some common negative peer reward
```
2022-03-18T17:12:17+05:30 DBUG reporting reputation change of -1024 to peer 12D3KooWH1cgFi2PkehEH2sHtkx7eQDkyyk3JikpvVNZqTRsoWv3, reason: Request timeout	handler.go:L81	pkg=peerset
```
frequency 33
```
2022-03-18T17:12:07+05:30 DBUG reporting reputation change of -4 to peer 12D3KooWH1cgFi2PkehEH2sHtkx7eQDkyyk3JikpvVNZqTRsoWv3, reason: Duplicate gossip	handler.go:L81	pkg=peerset
```
frequency 3009
```
2022-03-18T17:12:09+05:30 DBUG reporting reputation change of -2147483648 to peer 12D3KooWFgz1t6FUAFDMYwUjHdcZn3Td1Lts19sS5SaBX9q6DLMn, reason: Unsupported protocol	handler.go:L81	pkg=peerset
```
frequency 6

The message of banned threshold is coming right after negative rewards by unsupported protocol, which makes sense. bunch such negative reward could easily get the peer below banned threshold.

kishansagathiya:
So, sometimes I am seeing that sometime we fail to check if `/paritytech/grandpa/1` protocol is supported and sometimes we find that it is not supported!

PR #2241, opened 2022-01-27
feat(dot/peerset): Instrument peerset with peer reputation and number of bans

### LINKED ISSUE #2221: Add metrics for peerset

timwu20:
## Changes

- add metrics for reputation score of peers with labelled peer ID
- add metrics for peer bannings labelled with labelled peer ID

## Tests

```
go test ./dot/peerset
```

## Issues

- #2098 
- closes #2221 

## Primary Reviewer

- @EclesioMeloJunior

### DIFF
--- a/dot/peerset/peerset.go
+++ b/dot/peerset/peerset.go
@@ -12,10 +12,19 @@ import (
 
 	"github.com/ChainSafe/gossamer/internal/log"
 	"github.com/libp2p/go-libp2p-core/peer"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
 )
 
 var (
-	logger = log.NewFromGlobal(log.AddContext("pkg", "peerset"))
+	logger        = log.NewFromGlobal(log.AddContext("pkg", "peerset"))
+	bannedCounter = promauto.NewCounterVec(prometheus.CounterOpts{
+		Namespace: "gossamer_peerset",
+		Name:      "banned_total",
+		Help:      "total number of peer bans",
+	}, []string{
+		"peer_id",
+	})
 )
 
 const (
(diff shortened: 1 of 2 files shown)


EclesioMeloJunior (approved):
just wondering if is it possible to

edwardmack (approved):

noot (approved):

dot/peerset/peerset.go
```diff
@@ -282,6 +291,7 @@ func (ps *PeerSet) updateTime() error {
 			after := reputationTick(before)
 			n.setReputation(after)
 			ps.peerState.nodes[peerID] = n
+			reputationGauge.WithLabelValues(peerID.Pretty()).Set(float64(after))
```
qdm12, line 294: That will potentially blow up  memory usage of the Prometheus server since peerID is rather random. Maybe leave this as a log or find another way to report the metric with a limited-space label.

dot/peerset/peerset.go
```diff
@@ -334,6 +344,7 @@ func (ps *PeerSet) reportPeer(change ReputationChange, peers ...peer.ID) error {
 		for i := 0; i < setLen; i++ {
 			if ps.peerState.peerStatus(i, pid) == connectedPeer {
 				// disconnect peer
+				bannedCounter.WithLabelValues(pid.Pretty()).Inc()
```
qdm12, line 347: Same problem here

qdm12 (changes requested):

dot/peerset/peerset.go
timwu20: Ideally I only want to label this on the devnets where the peers are bound to only like 3-4 nodes.  I'll refactor this PR to make this configurable.

qdm12: Maybe just have it as a periodic log? 🤔 Like our sync benchmarker?

Or you could have a mapping from [1 to 50] to peerIDs, use the [1-50] set for the metrics labels and log out the mapping at program start. Not totally ideal either, but that would work.

PR #2248, opened 2022-01-27
fix(dot/network): remove `defer cancel()` inside loop

### LINKED ISSUE #2247: remove `defer cancel()` calls inside a infinite `for` loop
## Issue summary

- At file [`dot/network/discovery.go`](https://github.com/ChainSafe/gossamer/blob/a90a6e0f518d398ec34c2e21187f2511121fa3ee/dot/network/discovery.go#L164) the function `checkPeerCount` starts a looping to discover new peers at a defined amount of time, however, a new context is created inside the looping followed by the statement `defer cancel()` which is only called once the function exits
- This context is used by the ` findPeers` as a timeout trigger, but this can be done inside the function using `time.NewTimer(...)` without the needed to use a context.

-

EclesioMeloJunior:
## Changes

- Remove unnecessary context creation inside a infinity for loop

## Tests

```
go test ./dot/network/... -v -short -timeout 5m
```

## Issues

- Closes #2247

## Primary Reviewer

- @noot

### DIFF
--- a/dot/network/discovery.go
+++ b/dot/network/discovery.go
@@ -82,8 +82,8 @@ func (d *discovery) waitForPeers() (peers []peer.AddrInfo, err error) {
 	}
 
 	peers = make([]peer.AddrInfo, len(currentPeers))
-	for idx, p := range currentPeers {
-		peers[idx] = d.h.Peerstore().PeerInfo(p)
+	for idx, peer := range currentPeers {
+		peers[idx] = d.h.Peerstore().PeerInfo(peer)
 	}
 
 	return peers, nil
@@ -151,15 +151,15 @@ func (d *discovery) advertise() {
 	ttl := initialAdvertisementTimeout
 
 	for {
-		wait := time.NewTimer(ttl)
+		timer := time.NewTimer(ttl)
 
 		select {
 		case <-d.ctx.Done():
-			if !wait.Stop() {
-				<-wait.C
+			if !timer.Stop() {
+				<-timer.C
 			}
 			return
-		case <-wait.C:
+		case <-timer.C:
 			logger.Debug("advertising ourselves in the DHT...")
 			err := d.dht.Bootstrap(d.ctx)
 			if err != nil {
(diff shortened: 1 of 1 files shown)


dot/network/discovery.go
```diff
@@ -60,34 +60,44 @@ func newDiscovery(ctx context.Context, h libp2phost.Host,
...
 		}
 
-		for _, p := range peers {
-			d.bootnodes = append(d.bootnodes, d.h.Peerstore().PeerInfo(p))
+		currentPeers = d.h.Network().Peers()
+	}
+
+	peers = make([]peer.AddrInfo, len(currentPeers))
+	for idx, p := range currentPeers {
+		peers[idx] = d.h.Peerstore().PeerInfo(p)
```
qdm12, line 86: nit why not `peer` instead of `p` 😉

dot/network/discovery.go
```diff
@@ -141,8 +151,15 @@ func (d *discovery) advertise() {
 	ttl := initialAdvertisementTimeout
 
 	for {
+		wait := time.NewTimer(ttl)
```
qdm12, line 154: nit rename to timer?

dot/network/discovery.go
```diff
@@ -173,35 +189,34 @@ func (d *discovery) checkPeerCount() {
...
-func (d *discovery) findPeers(ctx context.Context) {
+func (d *discovery) findPeers() {
 	logger.Debug("attempting to find DHT peers...")
 	peerCh, err := d.rd.FindPeers(d.ctx, string(d.pid))
 	if err != nil {
 		logger.Warnf("failed to begin finding peers via DHT: %s", err)
 		return
 	}
 
+	timeout := time.NewTimer(findPeersTimeout)
```
qdm12, line 205: nit rename to timer. Usually timeout is the duration value passed to the timer.

dot/network/discovery.go
```diff
@@ -173,35 +189,34 @@ func (d *discovery) checkPeerCount() {
...
+	defer timeout.Stop()
```
qdm12, line 206: Stop and drain the channel in the `case peer := <-peerCh:` block in case you receive the peer at the same time as the timer triggers. You can use

```go
if !timer.Stop() {
    <-timer.C
}
```

qdm12 (approved):

noot (approved):

Issue #2208, opened 2022-01-18
lib/trie: refactor `lib/trie/trie.go` code


qdm12:
## Task summary

- Linear refactor of `lib/trie/trie.go`

PR #2255, opened 2022-01-28
chore(lib/trie): refactor `trie.go` and fix additional memory problems

qdm12:
## Changes

This is a linear refactor without changing function signatures. All tests are aimed to be kept the same to ease reviewing.

**Notable changes**

- If not modified, nodes of the previous trie generation are returned, meaning newer copied nodes will be left for garbage collection. This caused some of the test cases to be changed slightly regarding generation number.
- Comments revised, reduced or added depending on the case, with the aim of 'knowledge sharing' my trie knowledge so far for the next person to dig in there.
- Split functions into smaller functions
- Slice pre-allocation before appends for better memory performance (not usage)

## Tests

```
go test -race ./lib/trie/...
```

## Issues

- #2208


lib/trie/trie.go
```diff
@@ -167,675 +177,895 @@ func (t *Trie) Entries() map[string][]byte {
...
+	branch := newParent.(*node.Branch)
 
-			c.Children[i] = nil
-			c.SetDirty(true)
-			curr = handleDeletion(c, prefix)
-			return curr, true
-		}
+	if len(prefix) == len(branch.Key)+1 &&
+		bytes.HasPrefix(branch.Key, prefix[:len(prefix)-1]) {
+		// Prefix is one the children of the branch
```
qdm12, line 876:
```suggestion
		// Prefix is one of the children of the branch
```

EclesioMeloJunior (changes requested):

lib/trie/trie.go
```diff
@@ -167,675 +177,895 @@ func (t *Trie) Entries() map[string][]byte {
...
 }
 
-// NextKey returns the next key in the trie in lexicographic order. It returns nil if there is no next key
-func (t *Trie) NextKey(key []byte) []byte {
-	k := codec.KeyLEToNibbles(key)
+// NextKey returns the next key in the trie in lexicographic order.
+// It returns nil if no next key is found.
+func (t *Trie) NextKey(keyLE []byte) (nextKeyLE []byte) {
+	prefix := []byte(nil)
+	key := codec.KeyLEToNibbles(keyLE)
```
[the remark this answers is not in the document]
qdm12: It does return the key in nibbles format.

Why I kept it as simply `key` is that all the rest of the code uses `key` for the nibbles key. Most internal/private functions only handle nibbles key as well.

I started renaming all `key` to `keyNibbles` to be explicit about its format (like with `keyLE`), but that kind of looked ugly and added a lot of deltas. So in the end I took the approach to have it as `key` for nibbles and `keyLE` for Little Endian formatted keys. What do you think?

As in, I could also rename `key` to `keyNibbles` only at the 'entrypoints' where we deal with both nibbles and little endian keys.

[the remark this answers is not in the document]
qdm12: Also key fields in nodes (branch, leaf) are just named `Key` although they are in nibbles, so I think it would be good to just keep it as 'key' for nibbles-formatted keys

lib/trie/trie.go
```diff
@@ -167,675 +177,895 @@ func (t *Trie) Entries() map[string][]byte {
...
+	const startChildIndex = 0
+	return findNextKeyChild(parentBranch.Children, startChildIndex,
+		fullKey, searchKey)
+}
+
+// findNextKeyChild searches for a next key in the children
+// given and returns a next key or nil if no next key is found.
+func findNextKeyChild(children [16]node.Node, startIndex byte,
+	fullKey, key []byte) (nextKey []byte) {
+	for i := startIndex; i < byte(len(children)); i++ {
```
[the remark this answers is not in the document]
qdm12: Nice! Done in df05849d90ca622eb1adba80d9de6e089d072e65 with constant `node.ChildrenCapacity = 16`

EclesioMeloJunior (approved):
image

lib/trie/trie.go
```diff
@@ -167,675 +177,895 @@ func (t *Trie) Entries() map[string][]byte {
...
+func entries(parent Node, prefix []byte, kv map[string][]byte) map[string][]byte {
+	if parent == nil {
+		return kv
+	}
+
+	if parent.Type() == node.LeafType {
+		parentKey := parent.GetKey()
+		fullKeyNibbles := make([]byte, 0, len(prefix)+len(parentKey))
+		fullKeyNibbles = append(fullKeyNibbles, prefix...)
+		fullKeyNibbles = append(fullKeyNibbles, parentKey...)
```
noot, line 189:
```suggestion
		fullKeyNibbles := append(prefix, parentKey...)
```

I'm curious what the difference between these two functionally is? the values aren't being copied either way, so it seems to me it's the same?

lib/trie/trie.go
```diff
@@ -167,675 +177,895 @@ func (t *Trie) Entries() map[string][]byte {
...
+	parentLeafKey := leaf.Key
+	fullKey := make([]byte, 0, len(prefix)+len(parentLeafKey))
+	fullKey = append(fullKey, prefix...)
+	fullKey = append(fullKey, parentLeafKey...)
+
+	searchKeyBigger :=
+		(len(searchKey) < len(fullKey) &&
+			bytes.Compare(searchKey, fullKey[:len(searchKey)]) == 1) ||
+			(len(searchKey) >= len(fullKey) &&
+				bytes.Compare(searchKey[:len(fullKey)], fullKey) != -1)
```
noot, line 257: this really hurts my head and is a lot less readable than before, can you break it up and put comments for the different cases?

lib/trie/trie.go
```diff
@@ -167,675 +177,895 @@ func (t *Trie) Entries() map[string][]byte {
...
+
+	searchKeyShorter := len(searchKey) < len(fullKey)
+	searchKeyLonger := len(searchKey) > len(fullKey)
+
+	searchKeyBigger :=
+		(searchKeyShorter &&
+			bytes.Compare(searchKey, fullKey[:len(searchKey)]) == 1) ||
+			(!searchKeyShorter &&
+				bytes.Compare(searchKey[:len(fullKey)], fullKey) != -1)
+
```
noot, line 283: same here lol

lib/trie/trie.go
qdm12: Super micro optimization, see [this go-guild Slack comment](https://chainsafeio.slack.com/archives/C022YPNK4HF/p1642803472018200) or benchmark it yourself with https://go.dev/play/p/ZuD_AWdqZxo 😉 

Since I found out recently, I couldn't hold myself to add it here 😆

lib/trie/trie.go
qdm12: Totally, done in d8b98631c716c908c8a43a294b98dae20cbf05d7

lib/trie/trie.go
qdm12: Totally, done in d8b98631c716c908c8a43a294b98dae20cbf05d7

lib/trie/trie_test.go
```diff
@@ -2088,8 +2090,7 @@ func Test_Trie_clearPrefixLimit(t *testing.T) {
 			limit:         1,
 			expectedLimit: 1,
 			newParent: &node.Leaf{
-				Key:        []byte{1, 2},
-				Generation: 1,
```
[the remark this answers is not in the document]
qdm12: Yes it was; in this case for example, we do try to clear {1, 3} which doesn't exist, so the original parent isn't modified. Hence we should return the original parent, and not the copied parent with generation 1.

In the previous code we would return a copied generation-updated node even when no modification was done, so this was changed (and hence less memory usage as well).

timwu20 (approved):

lib/trie/trie.go
```diff
@@ -167,675 +177,888 @@ func (t *Trie) Entries() map[string][]byte {
...
-					return next
-				}
-			}
+	return fullKey
+}
+
+func findNextKeyBranch(parentBranch *node.Branch, prefix, searchKey []byte) (nextKey []byte) {
+	fullKey := make([]byte, 0, len(prefix)+len(parentBranch.Key))
+	fullKey = append(fullKey, prefix...)
+	fullKey = append(fullKey, parentBranch.Key...)
```
kishansagathiya, line 263: Make a helper function for this, since similar logic is being used at a bunch of places. May be something like `sliceWithCapacity(capacity int, arrays ...[]byte)`

lib/trie/trie.go
```diff
@@ -167,675 +177,888 @@ func (t *Trie) Entries() map[string][]byte {
...
-func nextKey(curr Node, prefix, key []byte) []byte {
-	switch c := curr.(type) {
-	case *node.Branch:
-		fullKey := append(prefix, c.Key...)
-		var cmp int
-		if len(key) < len(fullKey) {
-			if bytes.Compare(key, fullKey[:len(key)]) == 1 { // arg key is greater than full, return nil
-				return nil
-			}
+func findNextKey(parent Node, prefix, searchKey []byte) (nextKey []byte) {
```
kishansagathiya, line 232: I prefer the way it was previously. I find switch case more readable. I also prefer unnamed returns. Also why change the function name?

lib/trie/trie.go
```diff
@@ -41,54 +39,64 @@ func NewTrie(root Node) *Trie {
...
+		childTries:  childTries,
 		deletedKeys: make(map[common.Hash]struct{}),
 	}
-
-	return newTrie
 }
 
-func (t *Trie) maybeUpdateGeneration(n Node) Node {
-	if n == nil {
-		return nil
```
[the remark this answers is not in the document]
qdm12: Since there are nil checks before for every call to `maybeUpdateGeneration`.
Before we would run `maybeUpdateGeneration` then check for nil and mostly just return nil from the caller, so this moved above; there was no point calling `maybeUpdateGeneration` when we could just check for nil and return beforehand.

lib/trie/trie.go
qdm12: Added `concatenateSlices` together with a Benchmark to explain why it's here eventually.

lib/trie/trie.go
```diff
@@ -41,54 +39,64 @@ func NewTrie(root Node) *Trie {
...
+func (t *Trie) maybeUpdateGeneration(currentNode Node) (newNode Node) {
+	if currentNode.GetGeneration() == t.generation {
```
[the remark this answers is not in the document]
qdm12: On that note, there is an optimization to be done which is to only call `updateGeneration` when modification is needed. Right now we deep copy the node for every operation, even if we don't modify it, although I was careful to change the code to return the original copied node when there was no modification so the deep copied node is ready to be garbage collected. But it's still deep copies operation wasted and more load for the GC. Anyway, my point is this suggestion will help make the transition easier I think!

lib/trie/trie.go
qdm12: Added a benchmark in https://github.com/ChainSafe/gossamer/pull/2255/commits/9757382edea4068a4463229b5f937fcf84e3b701 with a helper function to concatenate slices.

lib/trie/trie.go
```diff
@@ -41,54 +39,54 @@ func NewTrie(root Node) *Trie {
...
-		oldNodeHash := n.GetHash()
-		if len(oldNodeHash) > 0 {
-			hash := common.BytesToHash(oldNodeHash)
-			t.deletedKeys[hash] = struct{}{}
-		}
-		return newNode
+// updateGeneration is called when the currentNode is from
+// an older trie generation (snapshot) so we deep copy the
+// node and update the generation on the newer copy.
+func updateGeneration(currentNode Node, trieGeneration uint64,
```
noot, line 68: should we leave the check for if the currentNode generation matches `trieGeneration`?

lib/trie/trie_test.go
```diff
@@ -91,101 +91,6 @@ func Test_Trie_Snapshot(t *testing.T) {
 	assert.Equal(t, expectedTrie, newTrie)
 }
 
-func Test_Trie_maybeUpdateGeneration(t *testing.T) {
```
noot, line 94: why is this removed?

noot, line 94: do we want to still have a test for `updateGeneration`?

noot (approved):

lib/trie/trie.go
qdm12: So....

I tried the following:

1. Add a generation equal check and return `currentNode` if the generations are equal.. but then this is back to `maybeUpdateGeneration` 😄 so that feels wrong
2. Add a generation equal check and panic if it's equal. Since this code should be called when the generation differs only, and we really don't want to deep copy a node for no reason.

I'm planning sometime in the future to call `updateGeneration` only when the node is modified. Right now we call it (and hence deep copy) even if no modification is done downstream. That would save some time since we would only deep copy when it would be actually needed. So I will keep `2.` to help development of this enhancement, since even unit tests might not pick up the deep copying really.


I added a check with panic if the node generation and trie generation are the same, since the caller shouldn't be calling this code path really, and this is susceptible to happen with future changes since it's a nifty detail.

lib/trie/trie_test.go
qdm12: Re-added a test for `updateGeneration`, good catch 👍 !

PR #2260, opened 2022-02-01
feat(lib/babe): implement secondary slot block production

### LINKED ISSUE #2216: lib/babe: implement secondary slot block production
## Task summary

- gossamer already has secondary slot verification, we can use this to implement secondary slot production
- if the `ConfigData` for an epoch has `secondary = 1 or 2` we should enable secondary slot production
- this can be done by calling `getSecondarySlotAuthor` for each slot in the epoch and determining if we can author a block in that slot (similar to the existing lottery)
- we will also need to update pre-runtime digest when producing a block in secondary to be of type `BabeSecondaryPlainPreDigest`
- also, when is `BabeSecondaryVRFPreDigest` used? (needs info from spec team probably)
- add stress tests as well

## Specification reference

- 

- https://research.web3.foundation/en/latest/polkadot/block-production/Babe.html
- https://polkadot.network/blog/polkadot-consensus-part-3-babe/

kishansagathiya:
## Changes

- claim secondary slot, if we can't get primary by errOverPrimarySlotThreshold

### Gossamer Devnet
- Run a devnet with 3 nodes, https://hackmd.io/@nZ-twauPRISEa6G9zg3XRw/B1RFO3uDt (2 nodes work as well!)
- You should see some secondary slots getting claimed in debug logs. You should see blocks getting created and some getting finalized.
- I think we could test this with dev chain, but not sure.
### Cross-Client Devnet
- Run a cross-client devnet with 2 nodes, https://hackmd.io/LfRK6aG6T7mMCE4Uybr-Mg. You will have to edit allowed slots in genesis-spec.json file to `"allowed_slots": "PrimaryAndSecondaryVRFSlots"`. Use the modifies genesis-spec.json to create genesis.json
- You will see blocks getting imported and built. On polkadot side, you should see messages suggesting that secondary vrf slots are being verified correctly. Something like below
```
2022-02-18 16:04:26.807 DEBUG tokio-runtime-worker babe: Verifying secondary VRF block #18 at slot: 274196743    
2022-02-18 16:04:26.807 TRACE             async-io async_io::driver: main_loop: waiting on I/O    
2022-02-18 16:04:26.807 TRACE             async-io async_io::reactor: process_timers: 0 ready wakers    
2022-02-18 16:04:26.807 TRACE             async-io polling: Poller::wait(_, Some(230.265526807s))    
2022-02-18 16:04:26.807 TRACE             async-io polling::epoll: wait: epoll_fd=22, timeout=Some(230.265526807s)    
2022-02-18 16:04:26.807 TRACE             async-io polling::epoll: modify: epoll_fd=22, fd=24, ev=Event { key: 18446744073709551615, readable: true, writable: false }    
2022-02-18 16:04:26.807 TRACE tokio-runtime-worker trie: Committing trie changes to db.    
```
## Issues

- Fixes #2216 

## Primary Reviewer

- @noot


lib/babe/epoch.go
```diff
@@ -222,11 +223,26 @@ func (b *Service) incrementEpoch() (uint64, error) {
...
 		epoch,
 		epochData.threshold,
 		b.keypair,
 	)
+	if err == nil {
+		b.slotToIfPrimary[slot] = true
+		return proof, nil
+	}
+
+	if epochData.secondary != 1 && epochData.secondary != 2 {
```
noot, line 238:
```suggestion
	if epochData.secondary == 0 {
```

lib/babe/babe.go
```diff
@@ -45,6 +45,9 @@ type Service struct {
 	// BABE authority keypair
 	keypair *sr25519.Keypair // TODO: change to BABE keystore (#1864)
 
+	// for slots where we are a producer, store the vrf output (bytes 0-32) + proof (bytes 32-96)
+	slotToIfPrimary map[uint64]bool
+
```
noot, line 50: this should be stored in the `epochHandler`

lib/babe/build.go
```diff
@@ -66,7 +67,7 @@ type BlockBuilder struct {
 // NewBlockBuilder creates a new block builder.
 func NewBlockBuilder(kp *sr25519.Keypair, ts TransactionState,
 	bs BlockState, proof *VrfOutputAndProof,
-	authidx uint32) (*BlockBuilder, error) {
+	authidx uint32, slotToIfPrimary map[uint64]bool) (*BlockBuilder, error) {
```
noot, line 70: there's actually two types of secondary production (VRF and Plain) so it would be nice to make this more generic. another idea is to pass the `PreRuntimeDigest` into this function and create it at the caller level, which would make more sense since creating the pre-runtime digest doesn't actually depend on anything in the `BlockBuilder`

lib/babe/crypto.go
```diff
@@ -83,6 +84,40 @@ func checkPrimaryThreshold(randomness Randomness,
...
+			}
+
+			logger.Debugf("claimed secondary slot, for slot number: %d", slot)
+
+			return &VrfOutputAndProof{
+				output: out,
+				proof:  proof,
+			}, nil
+		}
+	}
```
noot, line 115: this is where your error is coming from - this doesn't actually check if this node is allowed to produce a secondary block. `authorities` is the list of all the current authorities, this is returning an authorship proof if anyone is allowed to produce a secondary block.

you need to pass in our `authorityIndex` into this function and check if `secondarySlotAuthor == authorityIndex`.

lib/babe/epoch.go
```diff
@@ -222,11 +223,26 @@ func (b *Service) incrementEpoch() (uint64, error) {
...
-	return claimPrimarySlot(
+	proof, err := claimPrimarySlot(
 		epochData.randomness,
 		slot,
 		epoch,
 		epochData.threshold,
 		b.keypair,
 	)
+	if err == nil {
+		b.slotToIfPrimary[slot] = true
```
noot, line 234: remove this from the `Service` - I would actually put this into two separate functions, `runPrimaryLottery` and `runSecondaryLottery`. then, in `epochHandler` you could separate the current `slotToProof` into `primarySlotToProof` and `secondarySlotToProof` and store the proofs there in separate maps. this will also make it easier in the future if we implement secondary VRF

lib/babe/epoch_handler.go
```diff
@@ -44,23 +46,34 @@ func newEpochHandler(epochNumber, firstSlot uint64, epochData *epochData, consta
...
+		}
+
+		proof, err = claimSecondarySlot(epochData.randomness, i, epochNumber, epochData.authorities, epochData.threshold, keypair)
+		if err != nil {
 			return nil, fmt.Errorf("error running slot lottery at slot %d: %w", i, err)
 		}
 
 		slotToProof[i] = proof
+		slotToIfPrimary[i] = false
 		logger.Debugf("epoch %d: claimed slot %d", epochNumber, i)
```
noot, line 66: duplicate of `runLottery` code - can you store primary and secondary proofs in separate maps potentially?

noot (changes requested):
please update `claimSecondarySlot` logic to check if we can produce a secondary slot, this should fix your error!

lib/babe/babe.go
```diff
@@ -400,7 +400,7 @@ func (b *Service) handleEpoch(epoch uint64) (next uint64, err error) {
 	return next, nil
 }
 
-func (b *Service) handleSlot(epoch, slotNum uint64, authorityIndex uint32, proof *VrfOutputAndProof) error {
+func (b *Service) handleSlot(epoch, slotNum uint64, authorityIndex uint32, proof *VrfOutputAndProof, ifPrimary bool) error {
```
noot, line 403:
```suggestion
func (b *Service) handleSlot(epoch, slotNum uint64, authorityIndex uint32, proof *VrfOutputAndProof, isPrimary bool) error {
```

same for others pls

lib/babe/build.go
```diff
@@ -26,7 +26,7 @@ const (
 
 // construct a block for this slot with the given parent
 func (b *Service) buildBlock(parent *types.Header, slot Slot, rt runtime.Instance,
-	authorityIndex uint32, proof *VrfOutputAndProof) (*types.Block, error) {
+	authorityIndex uint32, proof *VrfOutputAndProof, ifPrimary bool) (*types.Block, error) {
```
noot, line 29:
```suggestion
	authorityIndex uint32, proof *VrfOutputAndProof, secondarySlots byte) (*types.Block, error) {
```

since `types.ConfigData.SecondarySlots` is type `byte` allowing us to indicate primary, secondary plain, or secondary VRF, it would be good to pass that in here so that when we implement secondary VRF we can just check the byte

lib/babe/epoch.go
```diff
@@ -222,11 +223,25 @@ func (b *Service) incrementEpoch() (uint64, error) {
 // if it is not authorised.
 // output = return[0:32]; proof = return[32:96]
 func (b *Service) runLottery(slot, epoch uint64, epochData *epochData) (*VrfOutputAndProof, error) {
-	return claimPrimarySlot(
+	// TODO: Check if whether to store if a primary slot was claimed? How?
```
noot, line 226: return a boolean as well?

noot: acutally, it doesn't matter here, as this is only used to get the first authoring slot

lib/babe/epoch.go
kishansagathiya: using pre runtime digests now, so don't need to use maps for proofs and if primary.

lib/babe/epoch.go
```diff
@@ -221,12 +222,34 @@ func (b *Service) incrementEpoch() (uint64, error) {
...
+		if err != nil {
+			return nil, err
+		}
+		if proof != nil {
+			preRuntimeDigest := types.NewBabeSecondaryPlainPreDigest(epochData.authorityIndex, slot)
+			return preRuntimeDigest.ToPreRuntimeDigest()
+		}
+	}
+
+	return nil, err
```
noot, line 254: `claimSecondarySlot` doesn't return an error if we fail to claim a slot like `claimPrimarySlot` does, this causes the calling function `getFirstAuthoringSlot` to set the first authoring slot incorrectly as it will set it to a slot where we can't actually author as primary or secondary

lib/babe/crypto.go
```diff
@@ -83,6 +84,38 @@ func checkPrimaryThreshold(randomness Randomness,
...
+		logger.Debugf("claimed secondary slot, for slot number: %d", slot)
+
+		return &VrfOutputAndProof{
+			output: out,
+			proof:  proof,
+		}, nil
+	}
+
+	// It is not our turn to propose
+	return nil, nil
```
noot, line 116: return an error, see other comment

lib/babe/epoch_handler.go
```diff
@@ -44,23 +44,36 @@ func newEpochHandler(epochNumber, firstSlot uint64, epochData *epochData, consta
 			epochData.threshold,
 			keypair,
 		)
-		if errors.Is(err, errOverPrimarySlotThreshold) {
+		if err == nil {
+			preRuntimeDigest, _ := types.NewBabePrimaryPreDigest(epochData.authorityIndex, i, proof.output, proof.proof).ToPreRuntimeDigest()
+			slotToPreRuntimeDigest[i] = preRuntimeDigest
+			logger.Debugf("epoch %d: claimed slot %d", epochNumber, i)
```
noot, line 50:
```suggestion
			logger.Debugf("epoch %d: claimed primary slot %d", epochNumber, i)
```

lib/babe/epoch_handler.go
```diff
@@ -44,23 +44,36 @@ func newEpochHandler(epochNumber, firstSlot uint64, epochData *epochData, consta
...
+		if err != nil {
 			return nil, fmt.Errorf("error running slot lottery at slot %d: %w", i, err)
 		}
 
-		slotToProof[i] = proof
-		logger.Debugf("epoch %d: claimed slot %d", epochNumber, i)
+		if proof != nil {
+			preRuntimeDigest, _ := types.NewBabeSecondaryPlainPreDigest(epochData.authorityIndex, i).ToPreRuntimeDigest()
+			slotToPreRuntimeDigest[i] = preRuntimeDigest
+			logger.Debugf("epoch %d: claimed slot %d", epochNumber, i)
```
noot, line 65:
```suggestion
			logger.Debugf("epoch %d: claimed secondary slot %d", epochNumber, i)
```

lib/babe/epoch_handler.go
```diff
@@ -44,23 +44,36 @@ func newEpochHandler(epochNumber, firstSlot uint64, epochData *epochData, consta
...
+		if err == nil {
+			preRuntimeDigest, _ := types.NewBabePrimaryPreDigest(epochData.authorityIndex, i, proof.output, proof.proof).ToPreRuntimeDigest()
+			slotToPreRuntimeDigest[i] = preRuntimeDigest
+			logger.Debugf("epoch %d: claimed slot %d", epochNumber, i)
 			continue
-		} else if err != nil {
+		}
+		if !errors.Is(err, errOverPrimarySlotThreshold) {
+			return nil, fmt.Errorf("error running slot lottery at slot %d: %w", i, err)
+		}
```
noot, line 55:
```suggestion
		if !errors.Is(err, errOverPrimarySlotThreshold) {
			continue
		}
```

this isn't an error, this causes us to return basically immediately

lib/babe/verify_integration_test.go
```diff
@@ -338,28 +337,45 @@ func TestVerifyPimarySlotWinner(t *testing.T) {
...
+
+	// slot := Slot{
+	// 	start:    time.Now(),
+	// 	duration: duration,
+	// 	number:   slotNumber,
+	// }
+
 	// create babe header
-	babeHeader := builder.buildBlockBABEPrimaryPreDigest(slot)
+	// babeHeader := builder.buildBlockBABEPrimaryPreDigest(slot)
```
noot, line 378: remove

noot (commented):
basically looks good, one logic error. i tried this with substrate and substrate doesn't accept our secondary blocks - might be due to the issue in the comment

dot/sync/chain_processor.go
```diff
@@ -103,15 +103,24 @@ func (s *chainProcessor) processReadyBlocks() {
...
 	if bd == nil {
 		return ErrNilBlockData
 	}
 
-	hasHeader, _ := s.blockState.HasHeader(bd.Hash)
-	hasBody, _ := s.blockState.HasBlockBody(bd.Hash)
+	hasHeader, err := s.blockState.HasHeader(bd.Hash)
+	if err != nil {
+		logger.Debugf("failed to check if block state has header for hash %s: %s", bd.Hash, err)
+		return err
```
qdm12, line 116: Should we instead just wrap the error here and log the error in one of the caller? Same suggestion for below.

lib/babe/build.go
```diff
@@ -59,30 +59,27 @@ type BlockBuilder struct {
...
 	currentAuthorityIndex uint32
+	preRuntimeDigest      *types.PreRuntimeDigest
 }
 
 // NewBlockBuilder creates a new block builder.
 func NewBlockBuilder(kp *sr25519.Keypair, ts TransactionState,
-	bs BlockState, proof *VrfOutputAndProof,
-	authidx uint32) (*BlockBuilder, error) {
+	bs BlockState,
+	authidx uint32, preRuntimeDigest *types.PreRuntimeDigest) (*BlockBuilder, error) {
```
qdm12, line 69: nit equal line lenghts are nice

```suggestion
	bs BlockState, authidx uint32, 
	preRuntimeDigest *types.PreRuntimeDigest) (
	*BlockBuilder, error) {
```

lib/babe/build.go
```diff
@@ -276,7 +236,7 @@ func (b *BlockBuilder) buildBlockExtrinsics(slot Slot, rt runtime.Instance) []*t
 	return included
 }
 
-func (b *BlockBuilder) buildBlockInherents(slot Slot, rt runtime.Instance) ([][]byte, error) {
+func buildBlockInherents(slot Slot, rt runtime.Instance) ([][]byte, error) {
```
qdm12, line 239: 👍 !

lib/babe/build_integration_test.go
```diff
@@ -177,8 +169,9 @@ func TestApplyExtrinsic(t *testing.T) {
 		duration: duration,
 		number:   2,
 	}
+	testVRFOutputAndProof := &VrfOutputAndProof{}
 
-	preDigest2, err := builder.buildBlockPreDigest(slot2)
+	preDigest2, err := types.NewBabePrimaryPreDigest(authorityIndex, slot2.number, testVRFOutputAndProof.output, testVRFOutputAndProof.proof).ToPreRuntimeDigest()
```
qdm12, line 174: nit long line

```suggestion
	preDigest2, err := types.NewBabePrimaryPreDigest(authorityIndex, slot2.number,
		testVRFOutputAndProof.output, testVRFOutputAndProof.proof).
		ToPreRuntimeDigest()
```

lib/babe/build_integration_test.go
```diff
@@ -190,7 +183,7 @@ func TestApplyExtrinsic(t *testing.T) {
 	require.NoError(t, err)
 	rt.SetContextStorage(ts)
 
-	preDigest, err := builder.buildBlockPreDigest(slot)
+	preDigest, err := types.NewBabePrimaryPreDigest(authorityIndex, slot.number, testVRFOutputAndProof.output, testVRFOutputAndProof.proof).ToPreRuntimeDigest()
```
qdm12, line 186:
```suggestion
	preDigest, err := types.NewBabePrimaryPreDigest(authorityIndex, slot.number,
		testVRFOutputAndProof.output, testVRFOutputAndProof.proof).
		ToPreRuntimeDigest()
```

lib/babe/crypto.go
```diff
@@ -83,6 +84,38 @@ func checkPrimaryThreshold(randomness Randomness,
 	return inoutUint.Compare(threshold) < 0, nil
 }
 
+func claimSecondarySlot(randomness Randomness,
+	slot, epoch uint64,
+	authorities []types.Authority,
+	keypair *sr25519.Keypair,
+	authorityIndex uint32,
+) (*VrfOutputAndProof, error) {
+
```
qdm12, line 93:
```suggestion
```

lib/babe/crypto.go
```diff
@@ -83,6 +84,38 @@ func checkPrimaryThreshold(randomness Randomness,
...
+	secondarySlotAuthor, err := getSecondarySlotAuthor(slot, len(authorities), randomness)
+	if err != nil {
+		return nil, err
```
qdm12, line 96:
```suggestion
		return nil, fmt.Errorf("cannot get secondary slot author: %w", err)
```

lib/babe/crypto.go
```diff
@@ -83,6 +84,38 @@ func checkPrimaryThreshold(randomness Randomness,
...
+	}
+
+	if int(authorityIndex) == int(secondarySlotAuthor) {
+		transcript := makeTranscript(randomness, slot, epoch)
+
+		out, proof, err := keypair.VrfSign(transcript)
+		if err != nil {
+			return nil, err
```
qdm12, line 104:
```suggestion
			return nil, fmt.Errorf("cannot verify transcript: %w", err)
```

lib/babe/epoch.go
```diff
@@ -221,12 +222,34 @@ func (b *Service) incrementEpoch() (uint64, error) {
...
+	}
+
+	if epochData.secondary == 0 {
+		return nil, err
+	}
+
+	if errors.Is(err, errOverPrimarySlotThreshold) {
+		proof, err = claimSecondarySlot(epochData.randomness, slot, epoch, epochData.authorities, b.keypair, epochData.authorityIndex)
+		if err != nil {
+			return nil, err
```
qdm12, line 246:
```suggestion
			return nil, fmt.Errorf("cannot claim secondary slot: %w", err)
```

lib/babe/epoch_handler.go
```diff
@@ -26,16 +27,15 @@ type epochHandler struct {
...
+	slotToPreRuntimeDigest map[uint64]*types.PreRuntimeDigest
 
 	handleSlot handleSlotFunc
 }
 
 func newEpochHandler(epochNumber, firstSlot uint64, epochData *epochData, constants constants,
 	handleSlot handleSlotFunc, keypair *sr25519.Keypair) (*epochHandler, error) {
 	// determine which slots we'll be authoring in by pre-calculating VRF output
-	slotToProof := make(map[uint64]*VrfOutputAndProof, constants.epochLength)
+	slotToPreRuntimeDigest := make(map[uint64]*types.PreRuntimeDigest)
```
qdm12, line 38: keep the pre-allocation?

```suggestion
	slotToPreRuntimeDigest := make(map[uint64]*types.PreRuntimeDigest, constants.epochLength)
```

lib/babe/verify_integration_test.go
```diff
@@ -338,28 +337,45 @@ func TestVerifyPimarySlotWinner(t *testing.T) {
...
-	duration, err := time.ParseDuration("1s")
+	babePreDigest, err := types.DecodeBabePreDigest(preRuntimeDigest.Data)
 	require.NoError(t, err)
-
-	slot := Slot{
-		start:    time.Now(),
-		duration: duration,
-		number:   slotNumber,
+	
+	// var slotNumber uint64
```
qdm12, line 346: nit remove?

```suggestion
```

tests/stress/stress_test.go
```diff
@@ -641,3 +642,64 @@ func TestSync_SubmitExtrinsicLoad(t *testing.T) {
 		// Output: Balance transferred from Alice to Bob: 100000000000000
 	}
 }
+
+func TestSecondarySlotProducer(t *testing.T) {
+	numNodes := 4
```
qdm12, line 647: nit

```suggestion
	const numNodes = 4
```

tests/stress/stress_test.go
```diff
@@ -641,3 +642,64 @@ func TestSync_SubmitExtrinsicLoad(t *testing.T) {
...
+	utils.Logger.Patch(log.SetLevel(log.Info))
```
qdm12, line 648: Why patch the log level here?

tests/stress/stress_test.go
```diff
@@ -641,3 +642,64 @@ func TestSync_SubmitExtrinsicLoad(t *testing.T) {
...
+
+	// start block producing node first
+	node, err := utils.RunGossamer(t, numNodes-1,
+		utils.TestDir(t, utils.KeyList[numNodes-1]),
```
qdm12, line 652: Use `t.TempDir()` instead of `utils.TestDir`

tests/stress/stress_test.go
```diff
@@ -641,3 +642,64 @@ func TestSync_SubmitExtrinsicLoad(t *testing.T) {
...
+	time.Sleep(time.Second * 15)
+	nodes, err := utils.InitializeAndStartNodes(t, numNodes-1, utils.GenesisDev, utils.ConfigNotAuthority)
+	require.NoError(t, err)
+	nodes = append(nodes, node)
+
+	time.Sleep(time.Second * 30)
+
+	defer func() {
+		errList := utils.StopNodes(t, nodes)
+		require.Len(t, errList, 0)
```
qdm12, line 668: nit

```suggestion
		require.Empty(t, errList)
```

tests/stress/stress_test.go
```diff
@@ -641,3 +642,64 @@ func TestSync_SubmitExtrinsicLoad(t *testing.T) {
...
+	}()
+
+	numCmps := 10
```
qdm12, line 671: nit

```suggestion
	const numCmps = 10
```

tests/stress/stress_test.go
```diff
@@ -641,3 +642,64 @@ func TestSync_SubmitExtrinsicLoad(t *testing.T) {
...
+	secondaryCount := 0
+
+	for i := 0; i < numCmps; i++ {
+		time.Sleep(3 * time.Second)
+		t.Log("comparing...", i)
```
qdm12, line 676: nit

```suggestion
```

tests/stress/stress_test.go
```diff
@@ -641,3 +642,64 @@ func TestSync_SubmitExtrinsicLoad(t *testing.T) {
...
+	defer func() {
+		errList := utils.StopNodes(t, nodes)
+		require.Len(t, errList, 0)
+	}()
+
+	numCmps := 10
+	secondaryCount := 0
+
+	for i := 0; i < numCmps; i++ {
+		time.Sleep(3 * time.Second)
```
qdm12, line 675: Do we need these 3s sleeps?

tests/stress/stress_test.go
```diff
@@ -641,3 +642,64 @@ func TestSync_SubmitExtrinsicLoad(t *testing.T) {
...
+		t.Log("comparing...", i)
+		hashes, err := compareBlocksByNumberWithRetry(t, nodes, strconv.Itoa(i))
+		if len(hashes) > 1 || len(hashes) == 0 {
```
qdm12, line 678:
```suggestion
		if len(hashes) != 1 {
```

tests/stress/stress_test.go
```diff
@@ -641,3 +642,64 @@ func TestSync_SubmitExtrinsicLoad(t *testing.T) {
...
+				block := utils.GetBlock(t, nodeWithHash, hash)
+				vdt := block.Header.Digest.VaryingDataType.Value()
+
+				switch vdt.(type) {
+				case gosstypes.BabeSecondaryPlainPreDigest:
+					// TODO: verify the block
+					secondaryCount++
+					fmt.Println("gosstypes.BabeSecondaryPlainPreDigest")
+				case gosstypes.BabePrimaryPreDigest:
+					fmt.Println("gosstypes.BabePrimaryPreDigest")
```
qdm12, line 697:
```suggestion
				case gosstypes.BabePrimaryPreDigest:
```

tests/stress/stress_test.go
kishansagathiya: test logic was inappropriate, I have changed it

kishansagathiya:
@noot After these changes, the behaviour is better. I do see the error on substrate side. Substrate fails to verify the block produced by secondary slots. The error is the log is somewhat mis-leading. The problems is really that it fails to match pre-digest with any of the valid pre-digest.

```
2022-02-11 11:47:40.023 DEBUG tokio-runtime-worker babe: Secondary slot assignments are disabled for the current epoch.    
2022-02-11 11:47:40.023 TRACE tokio-runtime-worker sync: Verifying 7(0x7527…c4ba) from 12D3KooWG9DPxJRmV3voVjK7jhYEMvsS4Pewbbw4NG5B7ykoowWt failed: Secondary slot assignments are disabled for the current epoch.    
2022-02-11 11:47:40.023 TRACE             async-io async_io::driver: main_loop: waiting on I/O    
2022-02-11 11:47:40.023 TRACE             async-io async_io::reactor: process_timers: 0 ready wakers    
2022-02-11 11:47:40.023 TRACE             async-io polling: Poller::wait(_, Some(247.951504614s))    
2022-02-11 11:47:40.023 TRACE tokio-runtime-worker sync: Imported 0 of 1    
2022-02-11 11:47:40.023 TRACE             async-io polling::epoll: wait: epoll_fd=21, timeout=Some(247.951504614s)    
2022-02-11 11:47:40.023 TRACE             async-io polling::epoll: modify: epoll_fd=21, fd=23, ev=Event { key: 18446744073709551615, readable: true, writable: false }    
2022-02-11 11:47:40.023  WARN tokio-runtime-worker sync: 💔 Verification failed for block 0x752765f86e01b0b1cb31684dea8263988ce3cb840e215e2f8ac1b1cc67e4c4ba received from peer: 12D3KooWG9DPxJRmV3voVjK7jhYEMvsS4Pewbbw4NG5B7ykoowWt, "Secondary slot assignments are disabled for the current epoch."    
```

kishansagathiya:
One of the troubling logs that I see is 
```
2022-02-14T17:56:40+05:30 WARN attempted to start epoch that has passed: current slot=274140266, start slot of epoch=274137643	epoch_handler.go:L98	pkg=babe
```
which might be causing problems.

Sometimes, I see following critical error
```
2022-02-14T07:58:50+05:30 CRIT no epoch data found for upcoming epoch number=2	epoch.go:L99	pkg=babe
2022-02-14T07:58:50+05:30 CRIT failed to run block production engine: cannot initiate and get epoch handler for epoch 2: failed to initiate epoch 2: cannot get epoch data and start slot: no epoch data found for upcoming epoch: for epoch 2	babe.go:L308	pkg=babe
```
I suspect that might also have to do with the warning above.

During verification of imported blocks, digests are verified as `SecondarySlotVRF` instead of `SecondarySlotPlain`. I have set `"allowed_slots": "PrimaryAndSecondaryPlainSlots"` in genesis file. It's not yet clear why this is happening.

noot:
@kishansagathiya maybe we can merge this for now but turn off secondary plain as it's unclear when it's used, and create another issue for secondary VRF implementation

kishansagathiya:
> @kishansagathiya maybe we can merge this for now but turn off secondary plain as it's unclear when it's used, and create another issue for secondary VRF implementation

Works for me

lib/babe/epoch_handler.go
```diff
@@ -44,23 +44,55 @@ func newEpochHandler(epochNumber, firstSlot uint64, epochData *epochData, consta
...
+				return nil, fmt.Errorf("error converting babe primary pre-digest to pre-runtime digest: %w", err)
+			}
+			slotToPreRuntimeDigest[i] = preRuntimeDigest
+			logger.Debugf("epoch %d: claimed primary slot %d", epochNumber, i)
+			continue
+		}
+
+		if !errors.Is(err, errOverPrimarySlotThreshold) {
+			return nil, fmt.Errorf("error running slot lottery at slot %d: %w", i, err)
+		}
```
timwu20, line 60:
```suggestion
 		switch err {
		case nil:
			preRuntimeDigest, err := types.NewBabePrimaryPreDigest(
				epochData.authorityIndex, i, proof.output, proof.proof).ToPreRuntimeDigest()
			if err != nil {
				return nil, fmt.Errorf("error converting babe primary pre-digest to pre-runtime digest: %w", err)
			}
			slotToPreRuntimeDigest[i] = preRuntimeDigest
			logger.Debugf("epoch %d: claimed primary slot %d", epochNumber, i)
			continue
		default:
			if !errors.Is(err, errOverPrimarySlotThreshold) {
				return nil, fmt.Errorf("error running slot lottery at slot %d: %w", i, err)
			}
		}
```

lib/babe/build.go
```diff
@@ -276,7 +240,7 @@ func (b *BlockBuilder) buildBlockExtrinsics(slot Slot, rt runtime.Instance) []*t
...
```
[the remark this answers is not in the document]
kishansagathiya: We dont need anything from BlockBuilder in this function

lib/babe/epoch.go
kishansagathiya: Yes,

When this comment was made, `claimSecondarySlot` did not return an error on failure. Thus, it would claim the slot. I have changed the code since then. And an error gets returned as expected.

tests/stress/stress_test.go
kishansagathiya: have changed the test

tests/stress/stress_test.go
kishansagathiya: changed the test

dot/digest/digest.go
```diff
@@ -123,6 +123,8 @@ func (h *Handler) HandleDigests(header *types.Header) {
 	for i, d := range header.Digest.Types {
 		val, ok := d.Value().(types.ConsensusDigest)
 		if ok {
+			// TODO: handleConsensusDigest never gets called other than the first time. (In development as well)
```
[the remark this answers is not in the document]
kishansagathiya: I am not sure.
After going through specs, it seems like this should be called when authorities change. This seemed something suspicious in the moment. @noot  would know better

lib/babe/babe.go
```diff
@@ -461,6 +465,7 @@ func (b *Service) handleSlot(epoch, slotNum uint64, authorityIndex uint32, proof
 		),
 	)
 
+	// TODO: sometimes this failes with cannot find parent block in blocktree
```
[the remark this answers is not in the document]
kishansagathiya: Happened with substrate, but this doesn't happen anymore

lib/babe/crypto.go
```diff
@@ -83,6 +84,36 @@ func checkPrimaryThreshold(randomness Randomness,
...
+	authorities []types.Authority,
+	keypair *sr25519.Keypair,
+	authorityIndex uint32,
+) (*VrfOutputAndProof, error) {
+	secondarySlotAuthor, err := getSecondarySlotAuthor(slot, len(authorities), randomness)
+	if err != nil {
+		return nil, fmt.Errorf("cannot get secondary slot author: %w", err)
+	}
+
+	if int(authorityIndex) == int(secondarySlotAuthor) {
```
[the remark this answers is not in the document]
kishansagathiya: fixed

lib/babe/babe.go
```diff
@@ -386,6 +386,7 @@ func (b *Service) handleEpoch(epoch uint64) (next uint64, err error) {
 		// stop current epoch handler
 		cancel()
 	case err := <-errCh:
+		// TODO: errEpochPast is sent on this channel, but it doesnot get logged here
```
qdm12, line 389: Not sure what you mean here, the error is logged below no?

lib/babe/crypto.go
```diff
@@ -83,6 +84,36 @@ func checkPrimaryThreshold(randomness Randomness,
 	return inoutUint.Compare(threshold) < 0, nil
 }
 
+func claimSecondarySlot(randomness Randomness,
+	slot, epoch uint64,
+	authorities []types.Authority,
+	keypair *sr25519.Keypair,
+	authorityIndex uint32,
+) (*VrfOutputAndProof, error) {
```
qdm12, line 92: nit why not have just 2 long-ish-but-not-too-long lines of arguments (especially since `slot` and `epoch` are on the same line)?

lib/babe/epoch_handler.go
```diff
@@ -44,23 +44,57 @@ func newEpochHandler(epochNumber, firstSlot uint64, epochData *epochData, consta
 			epochData.threshold,
 			keypair,
 		)
-		if errors.Is(err, errOverPrimarySlotThreshold) {
+
+		switch err {
+		case nil:
```
qdm12, line 49: Why do we use a switch here instead of the usual `if err != nil` or `err == nil`? That would bring less indentation

lib/babe/epoch_handler.go
```diff
@@ -44,23 +44,57 @@ func newEpochHandler(epochNumber, firstSlot uint64, epochData *epochData, consta
...
+			logger.Debugf("epoch %d: claimed primary slot %d", epochNumber, i)
+			continue
+		default:
+			if !errors.Is(err, errOverPrimarySlotThreshold) {
+				return nil, fmt.Errorf("error running slot lottery at slot %d: %w", i, err)
+			}
+		}
+
+		if epochData.secondary == 0 {
+			// return nil, errSecondarySlotProductionDisabled
```
qdm12, line 65: nit Don't forget to uncomment

lib/babe/epoch_handler.go
```diff
@@ -44,23 +44,57 @@ func newEpochHandler(epochNumber, firstSlot uint64, epochData *epochData, consta
...
+			preRuntimeDigest, err := types.NewBabeSecondaryPlainPreDigest(
+				epochData.authorityIndex, i).ToPreRuntimeDigest()
+			if err != nil {
+				return nil, fmt.Errorf(
+					"failed to get preruntime digest from babe secondary plain predigest for slot %d: %w", i, err)
+			}
+			slotToPreRuntimeDigest[i] = preRuntimeDigest
+			logger.Debugf("epoch %d: claimed secondary slot %d", epochNumber, i)
+		}
+
```
qdm12, line 88:
```suggestion
```

lib/babe/epoch_handler_test.go
```diff
@@ -8,14 +8,17 @@ import (
...
 	"github.com/ChainSafe/gossamer/pkg/scale"
 
 	"github.com/stretchr/testify/require"
 )
 
 func TestNewEpochHandler(t *testing.T) {
-	testHandleSlotFunc := func(epoch, slotNum uint64, authorityIndex uint32, proof *VrfOutputAndProof) error {
+	testHandleSlotFunc := func(epoch, slotNum uint64, authorityIndex uint32,
+		preRuntimeDigest *types.PreRuntimeDigest,
+	) error {
```
qdm12, line 21: nit dis looks better imo but feel free to discard

```suggestion
		preRuntimeDigest *types.PreRuntimeDigest) error {
```

lib/babe/verify_integration_test.go
```diff
@@ -338,29 +337,24 @@ func TestVerifyPimarySlotWinner(t *testing.T) {
...
+	var vrfOutput [sr25519.VRFOutputLength]byte
+	var vrfProof [sr25519.VRFProofLength]byte
+	switch d := babePreDigest.(type) {
+	case types.BabePrimaryPreDigest:
+		authorityIndex = d.AuthorityIndex
+		vrfOutput = d.VRFOutput
+		vrfProof = d.VRFProof
+	default:
+		t.Fatal("expect babe primary pre digest")
 	}
```
qdm12, line 356:
```suggestion
	d, ok := babePreDigest.(types.BabePrimaryPreDigest)
	require.True(t, ok)
	authorityIndex := d.AuthorityIndex
	vrfOutput := d.VRFOutput
	vrfProof := d.VRFProof
```

lib/babe/verify_integration_test.go
```diff
@@ -374,9 +368,7 @@ func TestVerifyPimarySlotWinner(t *testing.T) {
 	})
 	require.NoError(t, err)
 
-	ok, err := verifier.verifyPrimarySlotWinner(
-		babeHeader.AuthorityIndex, slot.number,
-		babeHeader.VRFOutput, babeHeader.VRFProof)
+	ok, err := verifier.verifyPrimarySlotWinner(authorityIndex, slotNumber, vrfOutput, vrfProof)
```
qdm12, line 371: Why not use directly fields of `d` for args

```suggestion
	ok, err := verifier.verifyPrimarySlotWinner(d.AuthorityIndex, slotNumber, d.VRFOutput, d.VRFProof)
```

And remove the variable assignments above

tests/stress/stress_test.go
```diff
@@ -641,3 +642,44 @@ func TestSync_SubmitExtrinsicLoad(t *testing.T) {
 		// Output: Balance transferred from Alice to Bob: 100000000000000
 	}
 }
+
+func TestStress_SecondarySlotProduction(t *testing.T) {
+	numNodes := 3
```
qdm12, line 647:
```suggestion
	const numNodes = 3
```

tests/stress/stress_test.go
```diff
@@ -641,3 +642,44 @@ func TestSync_SubmitExtrinsicLoad(t *testing.T) {
...
+
+	nodes, err := utils.InitializeAndStartNodes(t, numNodes, utils.GenesisThreeAuths, utils.ConfigDefault)
+	require.NoError(t, err)
+
+	primaryCount := 0
+	secondaryCount := 0
+	otherCount := 0
+	for i := 1; i < 20; i++ {
+		fmt.Printf("%d iteration\n", i)
```
qdm12, line 656:
```suggestion
```

tests/stress/stress_test.go
```diff
@@ -641,3 +642,44 @@ func TestSync_SubmitExtrinsicLoad(t *testing.T) {
...
+
+		switch babePreDigest.(type) {
+		case gosstypes.BabePrimaryPreDigest:
+			primaryCount++
+		case gosstypes.BabeSecondaryVRFPreDigest:
+			otherCount++
+		case gosstypes.BabeSecondaryPlainPreDigest:
+			secondaryCount++
+		}
+		require.NotEqual(t, babePreDigest, nil)
```
qdm12, line 679:
```suggestion
		require.NotNil(t, babePreDigest)
```

tests/stress/stress_test.go
```diff
@@ -641,3 +642,44 @@ func TestSync_SubmitExtrinsicLoad(t *testing.T) {
...
+
+		time.Sleep(10 * time.Second)
```
qdm12, line 681: Do we need to sleep that long here 🤔

tests/stress/stress_test.go
```diff
@@ -641,3 +642,44 @@ func TestSync_SubmitExtrinsicLoad(t *testing.T) {
...
+	}
+
+	assert.Greater(t, secondaryCount, 0)
```
qdm12, line 684: Should we assert `primaryCount` and `otherCount` as well?

dot/digest/digest.go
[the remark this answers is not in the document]
noot: what do you mean other than the first time?  this is called every time there's consensus digests in blocks afaik?

dot/types/babe.go
```diff
@@ -10,6 +10,18 @@ import (
...
+
+const (
+	// PrimarySlots only allows primary slots.
+	PrimarySlots AllowedSlots = iota
+	// PrimaryAndSecondaryPlainSlots allow primary and secondary plain slots.
+	PrimaryAndSecondaryPlainSlots
+	// PrimaryAndSecondaryVRFSlots allows primary and secondary VRF slots.
+	PrimaryAndSecondaryVRFSlots
+)
+
```
noot, line 24: is this the same as substrate? can you add a link if so?

lib/babe/epoch.go
```diff
@@ -221,12 +227,35 @@ func (b *Service) incrementEpoch() (uint64, error) {
...
+
+		proof, err = claimSecondarySlot(epochData.randomness, slot, epoch,
+			epochData.authorities, b.keypair, epochData.authorityIndex)
+		if err != nil {
+			return nil, fmt.Errorf("cannot claim secondary slot: %w", err)
+		}
+		if proof != nil {
+			preRuntimeDigest := types.NewBabeSecondaryPlainPreDigest(epochData.authorityIndex, slot)
+			return preRuntimeDigest.ToPreRuntimeDigest()
+		}
```
noot, line 257: proof isn't used for secondary plain, only secondary VRF, so you don't need to return the proof from `claimSecondarySlot`, maybe you can update it to be `claimSecondarySlotPlain` and not return any proof?

lib/babe/epoch_handler.go
```diff
@@ -44,23 +44,57 @@ func newEpochHandler(epochNumber, firstSlot uint64, epochData *epochData, consta
...
 
-		slotToProof[i] = proof
-		logger.Debugf("epoch %d: claimed slot %d", epochNumber, i)
+		if proof != nil {
+			preRuntimeDigest, err := types.NewBabeSecondaryPlainPreDigest(
+				epochData.authorityIndex, i).ToPreRuntimeDigest()
+			if err != nil {
+				return nil, fmt.Errorf(
+					"failed to get preruntime digest from babe secondary plain predigest for slot %d: %w", i, err)
+			}
```
noot, line 84: same here, proof isn't used

noot (commented):
a few comments, I'm good to merge this in if secondary slot production is disabled for now, as it fully breaks interop. alternatively, you could also make another PR off this one for secondary VRF and we could merge them both at the same time.

noot:
@kishansagathiya substrate is producing `BabeSecondaryVRFPreDigest` digests, based on the instructions here. https://hackmd.io/@nZ-twauPRISEa6G9zg3XRw/ByNkNNITK I think we'll need to implement that for the secondary slots to interop.

dot/digest/digest.go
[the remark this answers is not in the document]
kishansagathiya: > this is called every time there's consensus digests in blocks afaik?

True. I was not sure if consensus digest is always supposed to be in a block. It seems like it is not supposed to.

lib/babe/babe.go
kishansagathiya: I have seen below part getting executed 
https://github.com/ChainSafe/gossamer/blob/67a9bbbc4d840858000aabb486438db2d7d435aa/lib/babe/epoch_handler.go#L73-L76

Because `errEpochPast` is being sent on the channel here, I should see `error from epochHandler` in logs because of following code
https://github.com/ChainSafe/gossamer/blob/67a9bbbc4d840858000aabb486438db2d7d435aa/lib/babe/babe.go#L388-L390

I did not see `error from epochHandler` in the logs when I was running gossamer

lib/babe/epoch_handler.go
kishansagathiya: removed it

dot/types/babe.go
kishansagathiya: done

lib/babe/epoch.go
kishansagathiya: done

lib/babe/epoch_handler.go
kishansagathiya: done

tests/stress/stress_test.go
kishansagathiya: yes, I have tried with 5 sec, which does not work. May be there is a perfect number, but it is better to keep a few seconds extra to ensure few failed attempts.

lib/babe/crypto.go
kishansagathiya: changed

tests/stress/stress_test.go
kishansagathiya: changed

dot/digest/digest.go
```diff
@@ -123,6 +123,8 @@ func (h *Handler) HandleDigests(header *types.Header) {
...
+			// i.e., header.Digest does not have consensus digest. Why?
```
noot, line 127: remove - not every block has a consensus digest, usually only first block of an epoch

dot/types/babe_digest.go
```diff
@@ -51,20 +52,6 @@ func NewBabePrimaryPreDigest(authorityIndex uint32,
...
-	err := digest.Set(*d)
-	if err != nil {
-		return nil, err
-	}
-	enc, err := scale.Marshal(digest)
-	if err != nil {
-		return nil, err
-	}
-	return NewBABEPreRuntimeDigest(enc), nil
-}
```
noot, line 66: why are these deleted? curious why you favour a standalone function over a method on the digests

lib/babe/epoch.go
```diff
@@ -215,18 +222,68 @@ func (b *Service) incrementEpoch() (uint64, error) {
 	return next, nil
 }
 
-// runLottery runs the lottery for a specific slot number.
+// claimSlot claims slot for a specific slot number.
```
noot, line 225:
```suggestion
// claimSlot attempts to claim a slot for a specific slot number.
```

lib/babe/epoch.go
```diff
@@ -215,18 +222,68 @@ func (b *Service) incrementEpoch() (uint64, error) {
...
+
+		if err != nil {
+			return nil, fmt.Errorf(
+				"failed to get preruntime digest from babe secondary plain predigest for slot %d: %w", slotNumber, err)
+		}
+		logger.Debugf("epoch %d: claimed secondary plain slot %d", epochNumber, slotNumber)
+		return preRuntimeDigest, nil
+	default:
+		// this should never occur
+		return nil, errors.New("invalid slot claiming technique")
```
noot, line 287: make this a constant

lib/babe/epoch.go
```diff
@@ -215,18 +222,68 @@ func (b *Service) incrementEpoch() (uint64, error) {
...
+		}
+		logger.Debugf("epoch %d: claimed primary slot %d", epochNumber, slotNumber)
+		return preRuntimeDigest, nil
+	} else if !errors.Is(err, errOverPrimarySlotThreshold) {
+		return nil, fmt.Errorf("error running slot lottery at slot %d: %w", slotNumber, err)
+	}
+
+	switch epochData.secondary {
+	case types.PrimarySlots:
+		return nil, errSecondarySlotProductionDisabled
```
noot, line 255: you could probably just return `errOverPrimarySlotThreshold` here

lib/babe/epoch.go
```diff
@@ -215,18 +222,68 @@ func (b *Service) incrementEpoch() (uint64, error) {
...
+	case types.PrimaryAndSecondaryVRFSlots:
+		proof, err := claimSecondarySlotVRF(
+			epochData.randomness, slotNumber, epochNumber, epochData.authorities, keypair, epochData.authorityIndex)
+		if err != nil {
+			return nil, fmt.Errorf("error claim secondary vrf slot at %d: %w", slotNumber, err)
+		}
```
noot, line 261: add spaces after the closing bracket, this is sort of hard to read

lib/babe/epoch.go
```diff
@@ -215,18 +222,68 @@ func (b *Service) incrementEpoch() (uint64, error) {
...
+		preRuntimeDigest, err := types.ToPreRuntimeDigest(*types.NewBabeSecondaryVRFPreDigest(
+			epochData.authorityIndex, slotNumber, proof.output, proof.proof))
+
```
noot, line 264:
```suggestion
```

lib/babe/epoch_handler.go
```diff
@@ -26,41 +27,36 @@ type epochHandler struct {
...
+		preRuntimeDigest, err := claimSlot(epochNumber, i, epochData, keypair)
+		if err == nil {
+			slotToPreRuntimeDigest[i] = preRuntimeDigest
+			continue
+		}
+
+		if errors.Is(err, errNotOurTurnToPropose) || errors.Is(err, errSecondarySlotProductionDisabled) {
 			continue
-		} else if err != nil {
-			return nil, fmt.Errorf("error running slot lottery at slot %d: %w", i, err)
```
noot, line 50: re add this case

noot (commented):

dot/types/babe_digest.go
kishansagathiya: All `ToPreRuntimeDigest` methods use the same code. This is only to reduce repetitive code. That change deliberately has a commit of its own, so it can be easily reverted, if you prefer that.

lib/babe/verify_integration_test.go
kishansagathiya: adressed

tests/stress/stress_test.go
kishansagathiya: this is a long test. Knowing the iteration number if useful.

lib/babe/epoch_handler.go
kishansagathiya: addressed

lib/babe/errors.go
```diff
@@ -64,27 +64,27 @@ var (
...
+	errGetEpochData             = errors.New("get epochData error")
+	errFailedFinalisation       = errors.New("failed to check finalisation")
+	errMissingDigest            = errors.New("chain head missing digest")
+	errSetFirstSlot             = errors.New("set first slot error")
+	errGetEpoch                 = errors.New("get epoch error")
+	errSkipVerify               = errors.New("skipVerify error")
+	errMissingDigestItems       = errors.New("block header is missing digest items")
+	errDescendant               = errors.New("descendant err")
+	errServicePaused            = errors.New("service paused")
+	errInvalidSlotTechnique     = errors.New("invalid slot claiming technique")
```
qdm12, line 87: I hate the Go formatter for introducing useless deltas like that one... come we just changed 2 lines!

tests/stress/stress_test.go
```diff
@@ -644,42 +644,83 @@ func TestSync_SubmitExtrinsicLoad(t *testing.T) {
...
+		// {
+		// 	description:  "with secondary plain slots enabled",
+		// 	genesis:      utils.GenesisTwoAuthsSecondaryPlain_0_9_10,
+		// 	allowedSlots: gosstypes.PrimaryAndSecondaryPlainSlots,
+		// },
+		// {
+		// 	description:  "with secondary slots disabled",
+		// 	genesis:      utils.GenesisTwoAuthsPrimary_0_9_10,
+		// 	allowedSlots: gosstypes.PrimarySlots,
+		// },
```
qdm12, line 670: nit left TODO reminder

tests/stress/stress_test.go
```diff
@@ -644,42 +644,83 @@ func TestSync_SubmitExtrinsicLoad(t *testing.T) {
...
+	}
+	const numNodes = 2
+	for _, c := range testcases {
+		c := c // bypass scopelint false positive
```
qdm12, line 674: nit remove if you don't run tests in parallel. Also we don't use scopelint I think anymore.

```suggestion
```

tests/stress/stress_test.go
```diff
@@ -644,42 +644,83 @@ func TestSync_SubmitExtrinsicLoad(t *testing.T) {
...
+		// 	genesis:      utils.GenesisTwoAuthsSecondaryPlain_0_9_10,
+		// 	allowedSlots: gosstypes.PrimaryAndSecondaryPlainSlots,
+		// },
+		// {
+		// 	description:  "with secondary slots disabled",
+		// 	genesis:      utils.GenesisTwoAuthsPrimary_0_9_10,
+		// 	allowedSlots: gosstypes.PrimarySlots,
+		// },
+	}
+	const numNodes = 2
```
qdm12, line 672: nit why not the old 3?

qdm12 (approved):

lib/babe/epoch.go
```diff
@@ -184,11 +188,14 @@ func (b *Service) getLatestEpochData() (resEpochData *epochData, error error) {
...
-		_, err := b.runLottery(i, epoch, epochData)
+		_, err := claimSlot(epoch, i, epochData, b.keypair)
 		if err != nil {
 			if errors.Is(err, errOverPrimarySlotThreshold) {
 				continue
 			}
+			if errors.Is(err, errNotOurTurnToPropose) {
+				continue
+			}
 			return 0, fmt.Errorf("error running slot lottery at slot %d: error %w", i, err)
```
EclesioMeloJunior, line 199: this nested if can be removed.

```suggestion
	if errors.Is(err, errOverPrimarySlotThreshold) {
		continue
	}
	if errors.Is(err, errNotOurTurnToPropose) {
		continue
	} else if err != nil {
			      return 0, fmt.Errorf("error running slot lottery at slot %d: error %w", i, err)
	}			
```

tests/stress/stress_test.go
kishansagathiya: I know better now. 3 is not a requirement. I think this test could even run by just one node. I have chosen to go with two.

lib/babe/epoch.go
```diff
@@ -221,41 +222,71 @@ func (b *Service) incrementEpoch() (uint64, error) {
...
+				"failed to get preruntime digest from babe secondary plain predigest for slot %d: %w", slotNumber, err)
 		}
-	}
 
-	return nil, err
+		logger.Debugf("epoch %d: claimed secondary plain slot %d", epochNumber, slotNumber)
+		return preRuntimeDigest, nil
+	default:
+		// this should never occur
+		return nil, errInvalidSlotTechnique
```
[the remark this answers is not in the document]
kishansagathiya: > Although if it's fully tested, there shouldn't be a need for this case nor the panic really

I would rather have such dead code than not, just in case! Helps, cover all cases!

lib/babe/epoch.go
```diff
@@ -138,6 +139,7 @@ func (b *Service) getEpochDataAndStartSlot(epoch uint64) (*epochData, uint64, er
 		authorities:    data.Authorities,
 		authorityIndex: idx,
 		threshold:      threshold,
+		secondary:      types.AllowedSlots(cfgData.SecondarySlots),
```
EclesioMeloJunior, line 142: I would suggest to rename `secondary` to `slotProduction` as this property enables/not enable the vrf/plain secondary block productions
```suggestion
		slotProduction:      types.AllowedSlots(cfgData.SecondarySlots),
```

EclesioMeloJunior (changes requested):

dot/types/babe_digest.go
noot: I like @qdm12 's suggestion!

kishansagathiya: Made this change

lib/babe/babe.go
noot: @kishansagathiya can you open an issue for this?

tests/stress/stress_test.go
noot: @kishansagathiya

lib/babe/epoch.go
```diff
@@ -215,18 +218,71 @@ func (b *Service) incrementEpoch() (uint64, error) {
...
+	}
+
+	switch epochData.allowedSlots {
+	case types.PrimarySlots:
+		return nil, errNotOurTurnToPropose
+	case types.PrimaryAndSecondaryVRFSlots:
+		proof, err := claimSecondarySlotVRF(
+			epochData.randomness, slotNumber, epochNumber, epochData.authorities, keypair, epochData.authorityIndex)
+		if err != nil {
+			return nil, fmt.Errorf("error claim secondary vrf slot at %d: %w", slotNumber, err)
```
noot, line 256:
```suggestion
			return nil, fmt.Errorf("cannot claim secondary vrf slot at %d: %w", slotNumber, err)
```

lib/babe/epoch.go
```diff
@@ -215,18 +218,71 @@ func (b *Service) incrementEpoch() (uint64, error) {
...
+			return nil, fmt.Errorf("error converting babe secondary vrf pre-digest to pre-runtime digest: %w", err)
+		}
+
+		logger.Debugf("epoch %d: claimed secondary vrf slot %d", epochNumber, slotNumber)
+		return preRuntimeDigest, nil
+	case types.PrimaryAndSecondaryPlainSlots:
+		err = claimSecondarySlotPlain(
+			epochData.randomness, slotNumber, epochData.authorities, epochData.authorityIndex)
+		if err != nil {
+			return nil, fmt.Errorf("error claiming secondary plain slot at %d: %w", slotNumber, err)
```
noot, line 271:
```suggestion
			return nil, fmt.Errorf("cannot claim secondary plain slot at %d: %w", slotNumber, err)
```

lib/babe/epoch.go
```diff
@@ -215,18 +218,71 @@ func (b *Service) incrementEpoch() (uint64, error) {
...
+		}
+
+		preRuntimeDigest, err := types.NewBabeSecondaryPlainPreDigest(
+			epochData.authorityIndex, slotNumber).ToPreRuntimeDigest()
+
```
noot, line 276:
```suggestion
```

noot (approved):
looks good, just minor comments!

PR #2263, opened 2022-02-02
Eclesio/substrate docker image

### LINKED ISSUE #2148: create Docker image for substrate node running `gssmr` chain
## Issue summary

Add Docker image to be added to the local and ECS devnets which will run a substrate node running the `gssmr` chain.

- Should include datadog integration to pipe the substrate telemetry endpoint to our datadog account
  - Metrics should be differentiated by adding `substrate` to the the metrics prefix.
- Ensure that the same build arguments are supported as `devnet/bob.Dockerfile`.

## Other Information
- https://github.com/ChainSafe/gossamer/pull/2008

EclesioMeloJunior:
## Changes

- Create substrate docker images using `parity/polkadot:v0.9.10` image
- Update our current genesis.json and add a genesis-raw.json

## Tests

```

```

## Issues

- Closes #2148

## Primary Reviewer

- @timwu20


devnet/docker-compose-substrate.yml
```diff
@@ -0,0 +1,54 @@
...
+  alice:
+    platform: linux/amd64
+    build:
+      context: ./..
+      dockerfile: devnet/substrate_alice.Dockerfile
+    ports:
+      - 30333
+      - 9933
+      - 9944
+      - 9615:9615
```
timwu20, line 24:
```suggestion
      - 9615
```

devnet/docker-compose.yml
```diff
@@ -11,8 +12,10 @@ services:
       - 8546
       - 8540
       - 9876
+      - 6060
```
timwu20, line 15: what is this port for?

devnet/substrate_alice.Dockerfile
```diff
@@ -0,0 +1,26 @@
...
+ARG METRICS_NAMESPACE=substrate.local.devnet
+
+RUN apt-get update
+
+RUN DD_AGENT_MAJOR_VERSION=7 DD_INSTALL_ONLY=true DD_SITE="datadoghq.com" bash -c "$(curl -L https://s3.amazonaws.com/dd-agent/scripts/install_script.sh)"
+
+COPY --from=polkadot /usr/bin/polkadot /usr/bin/polkadot
+COPY devnet/chain/gssmr/genesis-raw.json genesis-spec.json
+
+RUN go run cmd/update-dd-agent-confd/main.go -n=${METRICS_NAMESPACE} -t=key:alice > /etc/datadog-agent/conf.d/openmetrics.d/conf.yaml
```
timwu20, line 16: what port is prometheus running on? `update-dd-agent-config` is currently expecting the prometheus endpoint to be `http://127.0.0.1:9876/metrics`.

devnet/substrate_bob.Dockerfile
```diff
@@ -0,0 +1,27 @@
...
+RUN test -n "$key"
+RUN DD_AGENT_MAJOR_VERSION=7 DD_INSTALL_ONLY=true DD_SITE="datadoghq.com" bash -c "$(curl -L https://s3.amazonaws.com/dd-agent/scripts/install_script.sh)"
+
+COPY --from=polkadot /usr/bin/polkadot /usr/bin/polkadot
+COPY devnet/chain/gssmr/genesis-raw.json genesis-spec.json
+
+RUN go run cmd/update-dd-agent-confd/main.go -n=${METRICS_NAMESPACE} -t=key:${key} > /etc/datadog-agent/conf.d/openmetrics.d/conf.yaml
+
+ENTRYPOINT service datadog-agent start && /usr/bin/polkadot \
+    --bootnodes=/dns/substrate-alice/tcp/30333/p2p/12D3KooWDpJ7As7BWAwRMfu1VU2WCqNjvq387JEYKDBj4kx6nXTN \
```
timwu20, line 21: if I were to use this Dockerfile to connect to a gossamer `alice` how would this work? If I wanted to connect substrate non-lead nodes to a gossamer lead node how would that work? 

The libp2p id would have to be the same for both substrate and gossamer `alice` nodes.

timwu20 (changes requested):

devnet/docker-compose-substrate.yml
timwu20: can we make these exposed ports the same as gossamer? Or we should at least comment what the diff ports do.

timwu20:
Are we absolutely sure that the alice, bob, and charlie keys are the same for both gossamer nodes and substrate nodes? Like if we run the gossamer only devnet with these changes using the raw genesis, are we sure those are the only 3 authorities?

EclesioMeloJunior:
> Are we absolutely sure that the alice, bob, and charlie keys are the same for both gossamer nodes and substrate nodes? Like if we run the gossamer only devnet with these changes using the raw genesis, are we sure those are the only 3 authorities?

Yes, those keys are the same for both gossamer and substrate nodes, and our raw genesis contains only those 3 authorities

devnet/docker-compose-substrate.yml
EclesioMeloJunior: Yes, it is possible!

devnet/docker-compose.yml
EclesioMeloJunior: This is the gossamer node pproof server port

devnet/substrate_alice.Dockerfile
EclesioMeloJunior: Changed substrate nodes to expose Prometheus at port 9876

noot (commented):
instead of modifying the existing gossamer-only devnet can you make a new folder for the cross-client devnet? it would be nice to have both

devnet/substrate_bob.Dockerfile
EclesioMeloJunior: Now both substrate and gossamer `alice` nodes uses the same node key so if you spin up a gossamer `alice` container and spin up `bob` and `charlie` substrate nodes they will connect without modifications to the image.

devnet/docker-compose-substrate.yml
```diff
@@ -0,0 +1,58 @@
+services:
+  prometheus:
+    image: prom/prometheus
+    volumes:
+    - ./../prometheus.yml:/etc/prometheus/prometheus.yml
```
qdm12, line 5: nit maybe read only?

```suggestion
    - ./../prometheus.yml:/etc/prometheus/prometheus.yml:ro
```

devnet/docker-compose-substrate.yml
```diff
@@ -0,0 +1,58 @@
+services:
+  prometheus:
+    image: prom/prometheus
```
qdm12, line 3: nit perhaps pin the image? At least to the major version?

devnet/docker-compose.yml
qdm12: Nice addition 👍

devnet/substrate_alice.Dockerfile
```diff
@@ -0,0 +1,44 @@
+FROM parity/polkadot:v0.9.10 as polkadot
```
qdm12, line 1: nit use `AS` so it's syntax highlighted better 😉 

```suggestion
FROM parity/polkadot:v0.9.10 AS polkadot
```

devnet/substrate_alice.Dockerfile
```diff
@@ -0,0 +1,44 @@
...
+
+ARG CHAIN=cross-client
+ARG VERSION=v0.9.10
+ARG DD_API_KEY=somekey
+ARG METRICS_NAMESPACE=substrate.local.devnet
+ARG NODE_KEY="93ce444331ced4d2f7bfb8296267544e20c2591dbf310c7ea3af672f2879cf8f"
+
+ENV CHAIN=${CHAIN}
+ENV NODE_KEY=${NODE_KEY}
+ENV DD_API_KEY=${DD_API_KEY}
```
qdm12, line 12: Do we need to persist all these env in the final image? Or is it just used at build time?

devnet/substrate_alice.Dockerfile
```diff
@@ -0,0 +1,44 @@
...
+    --rpc-port 8545 \
+    --ws-port 8546 \
+    --node-key ${NODE_KEY} \ 
+    --tmp \
+    --prometheus-external \
+    --prometheus-port 9876 \
+    --unsafe-rpc-external \
+    --unsafe-ws-external
+
+EXPOSE 7001 8545 8546 9876
```
qdm12, line 44: nit add `/tcp` to each ports? 🙏  Also adding `6060/tcp` for pprof

```suggestion
EXPOSE 7001/tcp 8545/tcp 8546/tcp 9876/tcp 6060/tcp
```

devnet/substrate_bob.Dockerfile
```diff
@@ -0,0 +1,45 @@
...
+    --rpc-port 8545 \
+    --ws-port 8546 \
+    --${key} \
+    --tmp \
+    --prometheus-external \
+    --prometheus-port 9876 \
+    --unsafe-rpc-external \
+    --unsafe-ws-external
+
+EXPOSE 7001 8545 8546 9876
```
qdm12, line 45: nit add `/tcp` to each ports? 🙏  Also adding `6060/tcp` for pprof

```suggestion
EXPOSE 7001/tcp 8545/tcp 8546/tcp 9876/tcp 6060/tcp
```

devnet/substrate_bob.Dockerfile
```diff
@@ -0,0 +1,45 @@
+# This is the build stage for Substrate. Here we create the binary.
```
qdm12, line 1: This comment is wrong I think?

EclesioMeloJunior: yes, it is

devnet/substrate_bob.Dockerfile
EclesioMeloJunior: I don't think substrate nodes have a `pprof` server

devnet/substrate_alice.Dockerfile
EclesioMeloJunior: I think that `DD_API_KEY` is the only one we need to keep

devnet/substrate_alice.Dockerfile
qdm12: What port does substrate uses? Shouldn't we use the standard `9090` port for Gossamer as well 🤔

devnet/substrate_bob.Dockerfile
qdm12: Oh my bad, never mind 😄

devnet/docker-compose-substrate.yml
```diff
@@ -1,8 +1,8 @@
 services:
   prometheus:
-    image: prom/prometheus
+    image: prom/prometheus:v2.33.1
     volumes:
-    - ./../prometheus.yml:/etc/prometheus/prometheus.yml
+    - ./../prometheus.yml:/etc/prometheus/prometheus.yml:ro
     command:
     - '--config.file=/etc/prometheus/prometheus.yml'
```
qdm12, line 7: nit this is the default and we can remove it I think

```suggestion
```

devnet/substrate_bob.Dockerfile
```diff
@@ -1,14 +1,11 @@
...
 FROM golang:1.17
 
 ARG key
 ARG CHAIN=cross-client
 ARG DD_API_KEY=somekey
 ARG METRICS_NAMESPACE=substrate.local.devnet
 
-ENV key=${key}
-ENV CHAIN=${CHAIN}
 ENV DD_API_KEY=${DD_API_KEY}
```
qdm12, line 9: I might be repeating myself, but shouldn't the API key be injected at runtime and not build? Sounds risky to have a key there

devnet/substrate_alice.Dockerfile
EclesioMeloJunior: by default the port 9090 is the one Prometheus docker image [prom/prometheus](https://hub.docker.com/r/prom/prometheus) uses to bootstrap it's own server to collect metrics

devnet/substrate_bob.Dockerfile
EclesioMeloJunior: This image is stored inside a private container registry, but @timwu20 can provide more context about the need of this key inside the container environment

qdm12: ohh yeah that's true so never mind. Although it should be possible to have at build time, and then injected at runtime (and not stored in the image). Anyway for another time!

qdm12 (approved):

devnet/substrate_alice.Dockerfile
```diff
@@ -0,0 +1,40 @@
+FROM parity/polkadot:v0.9.10 AS polkadot
+FROM golang:1.17
+
+ARG CHAIN=cross-client
+ARG VERSION=v0.9.10
+ARG DD_API_KEY=somekey
+ARG METRICS_NAMESPACE=substrate.local.devnet
+
+ENV CHAIN=${CHAIN}
```
jimjbrettj, line 9: Should we have the Chain env and arg variables have different names for clarity?

jimjbrettj (approved):
looks good!

devnet/substrate_alice.Dockerfile
EclesioMeloJunior: Not really, both values are used only once in the same place, docker file uses `ARG CHAIN` only in build time and we need to use the `CHAIN` value in the `ENTRYPOINT` which is only executed after the build time. To achieve this we just assign the `ARG` to `ENV` and now we can access the variable value after build time

devnet/docker-compose-substrate.yml
```diff
@@ -0,0 +1,57 @@
+services:
+  prometheus:
```
timwu20, line 2: why are we running prometheus in this docker-compose?

devnet/substrate_alice.Dockerfile
```diff
@@ -0,0 +1,40 @@
+FROM parity/polkadot:v0.9.10 AS polkadot
+FROM golang:1.17
+
+ARG CHAIN=cross-client
```
timwu20, line 4: what is the difference between the `gssmr` chain that's within our codebase and this one?

devnet/substrate_alice.Dockerfile
```diff
@@ -0,0 +1,40 @@
...
+
+ENTRYPOINT service datadog-agent start && /usr/bin/polkadot \
+    --chain chain/$CHAIN/genesis-raw.json \
+    --alice \
+    --port 7001 \
+    --rpc-port 8545 \
+    --ws-port 8546 \
+    --node-key "93ce444331ced4d2f7bfb8296267544e20c2591dbf310c7ea3af672f2879cf8f" \ 
+    --tmp \
+    --prometheus-external \
```
timwu20, line 35: do we need to prometheus-external?

devnet/alice.Dockerfile
```diff
@@ -32,6 +35,6 @@ RUN go run cmd/update-dd-agent-confd/main.go -n=${METRICS_NAMESPACE} -t=key:alic
 
 WORKDIR /gossamer
 
-ENTRYPOINT service datadog-agent start && gossamer --key=alice --babe-lead --publish-metrics --rpc --rpc-external=true --pubdns=alice
+ENTRYPOINT service datadog-agent start && gossamer --key=alice --babe-lead --publish-metrics --rpc --rpc-external=true --pubdns=alice --port 7001
 
-EXPOSE 7001/tcp 8545/tcp 8546/tcp 8540/tcp 9876/tcp
+EXPOSE 7001/tcp 8545/tcp 8546/tcp 8540/tcp 9876/tcp 6060/tcp
```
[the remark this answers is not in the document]
qdm12: pprof server

devnet/substrate_alice.Dockerfile
qdm12: Yes that makes sense. Ultimately we just want to inject easily the `ENV CHAIN` with `--build-arg CHAIN=blabla`, and that does it 😉 Taking the liberty to resolve this.

devnet/substrate_alice.Dockerfile
qdm12: Yes but by default what's the port substrate/polkadot uses for Prometheus? We force it to `9876` here it seems. I'm thinking we should change it in Gossamer (for end users experience).

As far as I know, applications usually have their Prom registry server listening on `9090`, so maybe we want to use that instead? Although if polkadot uses 9876 for whatever weird reason, then sure we can match it as well I guess.

devnet/substrate_bob.Dockerfile
```diff
@@ -0,0 +1,42 @@
+FROM parity/polkadot:v0.9.10 AS polkadot
+FROM golang:1.17
```
qdm12, line 2: It's rather strange to have a golang based image to run polkadot.

We should have instead two builders, one `polkadot` (already one) and one for the datadog yml setup.

devnet/substrate_bob.Dockerfile
```diff
@@ -0,0 +1,42 @@
...
+COPY --from=polkadot /usr/bin/polkadot /usr/bin/polkadot
+
+WORKDIR /gossamer
+
+COPY go.mod go.sum ./
+RUN go mod download
+
+COPY . . 
+
+WORKDIR /gossamer/devnet
```
qdm12, line 25: Careful `/gossamer/devnet` uses a different `go.mod` than the root `go.mod`.

qdm12 (changes requested):

devnet/docker-compose-substrate.yml
EclesioMeloJunior: I added Prometheus while testing the exported metric data from authority nodes, I decided to keep it in the same file to easily start a Prometheus server with authorities nodes, should I remove it?

devnet/substrate_alice.Dockerfile
EclesioMeloJunior: Both have 3 authority nodes, however, the `cross-client` genesis file is accepted by the substrate node since it has `staking` and `session` json keys while `gssmr` chain has `Babe` and `Grandpa` json keys. The `cross-client` genesis file has the runtime version 09.10 that don't trigger errors like:
````
CRIT target=runtime message=panicked at 'Timestamp slot must match `CurrentSlot`'
````

File: devnet/substrate_alice.Dockerfile (hunk 17 above)
Comment (EclesioMeloJunior) [reply devnet/substrate_alice.Dockerfile:35]: yes, otherwise the substrate will not expose the prometheus metrics through all network interfaces (`0.0.0.0`)

File: devnet/substrate_bob.Dockerfile (hunk 20 above)
Comment (EclesioMeloJunior) [reply devnet/substrate_bob.Dockerfile:25]: When I do `COPY . .` after `RUN go mod download` I'm copying the `devnet` folder as well right? I'm not sure why do I need to copy `go.mod` and `go.sum` before the code?

Comment (qdm12) [reply devnet/substrate_bob.Dockerfile:25]: Oh I'm an idiot, this is a Dockerfile in the devnet directory, my bad!



> I'm not sure why do I need to copy `go.mod` and `go.sum` before the code?

For Docker layer caching to have faster rebuilds, since go.mod doesn't change much and go mod download takes some decent amount of time 😉

**Comment by timwu20** [comment]
Can we update this to polkadot v0.9.16

File: devnet/substrate_alice.Dockerfile (hunk 16 above)
Comment (kishansagathiya) [reply devnet/substrate_alice.Dockerfile:4]: > where did the original genesis-spec.json come from?

`./polkadot build-spec --disable-default-bootnode --dev > genesis-spec.json`
And then you will have to modify it according to your requirements, like adding authorities. 

And then you run something like this to get a genesis.json file
`./polkadot build-spec --chain genesis-spec.json --raw --disable-default-bootnode > genesis.json`

Will make a pr to add this to read me.

New versions have a different genesis-spec file than 0.9.10.
Also running cross-client dev net with newer version of polkadot gives runtime errors. So, yeah we need to update.

File: devnet/substrate_bob.Dockerfile (hunk 19 above)
Comment (EclesioMeloJunior) [reply devnet/substrate_bob.Dockerfile]: Nice comment. I changed to:
1. Use the golang image to run the `cmd/update-dd-agent-confd/main.go` made by @timwu20  to generate the `config.yml` file 
2. Use the `parity/polkadot:v0.9.17` to execute the polkadot binary just copying the `config.yml` from the upper golang image

File: devnet/substrate_alice.Dockerfile [hunk 21]
```diff
@@ -1,35 +1,48 @@
...
 ARG DD_API_KEY=somekey
-ARG METRICS_NAMESPACE=substrate.local.devnet
 
-ENV CHAIN=${CHAIN}
 ENV DD_API_KEY=${DD_API_KEY}
+ENV CHAIN=${CHAIN}
 
-RUN DD_AGENT_MAJOR_VERSION=7 DD_INSTALL_ONLY=true DD_SITE="datadoghq.com" bash -c "$(curl -L https://s3.amazonaws.com/dd-agent/scripts/install_script.sh)"
+USER root
+RUN apt update && apt install -y curl
```
qdm12, line 24: supernit just to have a smaller image

```suggestion
RUN apt update && apt install -y curl && rm -r /var/cache/* /var/lib/apt/lists/*
```

devnet/substrate_alice.Dockerfile
```diff
@@ -1,35 +1,48 @@
...
+RUN curl -L https://s3.amazonaws.com/dd-agent/scripts/install_script.sh --output install_script.sh && \
+    chmod +x ./install_script.sh
 
-COPY go.mod go.sum ./
-RUN go mod download
+RUN DD_AGENT_MAJOR_VERSION=7 DD_INSTALL_ONLY=true DD_SITE="datadoghq.com" ./install_script.sh
+COPY --from=openmetrics /devnet/conf.yaml /etc/datadog-agent/conf.d/openmetrics.d/
 
-COPY . .
+RUN service datadog-agent start
```
qdm12, line 34: nit should we have this in the entrypoint with `&&`?

devnet/substrate_bob.Dockerfile
```diff
@@ -1,32 +1,46 @@
...
-ENV key=${key}
-ENV CHAIN=${CHAIN}
 ENV DD_API_KEY=${DD_API_KEY}
+ENV CHAIN=${CHAIN}
+ENV key=${key}
 
-RUN test -n "${key}"
-RUN DD_AGENT_MAJOR_VERSION=7 DD_INSTALL_ONLY=true DD_SITE="datadoghq.com" bash -c "$(curl -L https://s3.amazonaws.com/dd-agent/scripts/install_script.sh)"
+USER root
+RUN apt update && apt install -y curl
```
qdm12, line 27:
```suggestion
RUN apt update && apt install -y curl && rm -r /var/cache/* /var/lib/apt/lists/*
```

qdm12 (approved):

devnet/substrate_bob.Dockerfile
qdm12: Awesome 👍 👍

EclesioMeloJunior:
@timwu20 now there is only one genesis file!

devnet/substrate_alice.Dockerfile
```diff
@@ -0,0 +1,55 @@
+# Copyright 2022 ChainSafe Systems (ON)
+# SPDX-License-Identifier: LGPL-3.0-only
+
+FROM golang:1.17 as openmetrics
+ARG METRICS_NAMESPACE=substrate.local.devnet
+
+WORKDIR /devnet
+
+COPY ./devnet/go.mod ./devnet/go.sum ./
+RUN go mod download
```
timwu20, line 10: do we need this? won't `go run ...` on line 13 just auto-download dependencies?

devnet/substrate_alice.Dockerfile
timwu20: can we rename this to something more descriptive.  `3-auth-node-0.9.10` or something.

timwu20 (changes requested):

devnet/substrate_alice.Dockerfile
qdm12: This is for Docker layer caching, so dependencies are cached and rebuilds will be faster.

devnet/substrate_alice.Dockerfile
qdm12: Also we could have `0.9.10` as a build argument at the top global scope in the Dockerfile.

i.e.

```Dockerfile
ARG POLKADOT_VERSION=v0.9.10

FROM parity/polkadot:${POLKADOT_VERSION} AS polkadot
FROM golang:1.17

ARG CHAIN=3-auth-node-${POLKADOT_VERSION}
```

EclesioMeloJunior: @qdm12 I was having problems trying to use arguments declared before `FROM`, following this issue ([link](https://github.com/moby/moby/issues/34129)) I could resolve and apply the suggestion. 

Basically, all the `ARG`'s declared/initialized before `FROM` needs to be redeclared after `FROM` to use its values otherwise will not be possible to get the value from `ARG`. I your suggestion we need to add

```
ARG POLKADOT_VERSION=v0.9.10

FROM parity/polkadot:${POLKADOT_VERSION} AS polkadot
FROM golang:1.17

+ ARG POLKADOT_VERSION
# Now its possible to use the value here
ARG CHAIN=3-auth-node-${POLKADOT_VERSION} 
```

devnet/substrate_bob.Dockerfile
```diff
@@ -13,17 +13,22 @@ RUN go mod download
...
 ARG DD_API_KEY=somekey
+ARG key
 
 ENV DD_API_KEY=${DD_API_KEY}
 ENV CHAIN=${CHAIN}
 ENV key=${key}
 
 USER root
+RUN gpg --recv-keys --keyserver hkps://keys.mailvelope.com 9D4B2B6EB8F97156D19669A9FF0812D491B96798
+RUN gpg --export 9D4B2B6EB8F97156D19669A9FF0812D491B96798 > /usr/share/keyrings/parity.gpg
```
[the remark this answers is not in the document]
EclesioMeloJunior: I was getting problem in the `apt update` because no pub keys were provided, so I follow the Polkadot README ([link](https://github.com/paritytech/polkadot#debian-based-debian-ubuntu)) installation guide that adds these gpg keys

Issue #1887, opened 2021-10-13
Check areas to add peer reputation


arijitAD:
## Issue summary

-  Check areas to add peer reputation

## Other information and links

- https://github.com/ChainSafe/gossamer/pull/1791#discussion_r725987714

timwu20:
@arijitAD can you provide a little more context?

arijitAD:
I added peer scoring in major places similar to Substrate. But I didn't explore the entire substrate code to look for all possible areas where peer scoring was used. The task is to compare our peer scoring with substrate and see if we missed out somewhere.

timwu20:
Protobuf file is missing `StateRequest`, `StateResponse` and `StateEntry` message types.  This was added into substrate via this [PR](https://github.com/paritytech/substrate/pull/8884/files) Implementation of "fast sync" and the handling these message types is not implemented.  `BAD_MESSAGE` is reported if `StateResponse` fails to decode ([line](https://github.com/paritytech/substrate/blob/master/client/network/src/protocol.rs#L1476)).  

Added `GenesisMismatch` type to match substrate `GENESIS_MISMATCH` where change is min int32 (essentially banning) that node.  Added report peer to gossamer code, PR otw.

substrate `rep::REFUSED` is not implemented in gossamer.  Need to investigate how to detect if a peer is refusing our connections.

substrate `rep::ANY_TRANSACTION` and `rep::ANY_TRANSACTION_REFUND` is not implemented.

PR #2265, opened 2022-02-02
feat(dot/network): add mismatched genesis peer reporting

timwu20:
## Changes

- Matches substrate to apply reputation change for a given peer if their genesis hash doesn't match node genesis hash.

## Tests

```
go test ./dot/network
```

## Issues

- #1887 

## Primary Reviewer

- @EclesioMeloJunior

### DIFF
--- a/dot/peerset/constants.go
+++ b/dot/peerset/constants.go
@@ -62,4 +62,9 @@ const (
 	BadJustificationValue Reputation = -(1 << 16)
 	// BadJustificationReason is used when peer send invalid justification.
 	BadJustificationReason = "Bad justification"
+
+	// GenesisMismatch is used when peer has a different genesis
+	GenesisMismatch Reputation = math.MinInt32
+	// GenesisMismatchReason used when a peer has a different genesis
+	GenesisMismatchReason = "Genesis mismatch"
 )
--- a/dot/network/block_announce.go
+++ b/dot/network/block_announce.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"math/big"
 
+	"github.com/ChainSafe/gossamer/dot/peerset"
 	"github.com/ChainSafe/gossamer/dot/types"
 	"github.com/ChainSafe/gossamer/lib/common"
 	"github.com/ChainSafe/gossamer/pkg/scale"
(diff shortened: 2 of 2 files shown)


EclesioMeloJunior (approved):

kishansagathiya (commented):
Looking at substrate code, it looks like they report only when a new peer gets connected.

Doesn't `validateBlockAnnounceHandshake` run a lot more frequently?

Also, it might be worth adding some tests for this as well (not sure how)!

noot (approved):
maybe out of scope for this PR but if a peer has a genesis mismatch we should disconnect from them and not connect to them again, as they're on a different chain

timwu20:
> maybe out of scope for this PR but if a peer has a genesis mismatch we should disconnect from them and not connect to them again, as they're on a different chain

Applying this reputation change will essentially disconnect the peer.  The score is the minimum int32, so this will be way below the banning threshold. The actual banning and disconnect happens in another goroutine.

noot:
@timwu20 ok that's good to know!

dot/peerset/constants.go
```diff
@@ -62,4 +62,9 @@ const (
 	BadJustificationValue Reputation = -(1 << 16)
 	// BadJustificationReason is used when peer send invalid justification.
 	BadJustificationReason = "Bad justification"
+
+	// GenesisMismatch is used when peer has a different genesis
+	GenesisMismatch Reputation = math.MinInt32
+	// GenesisMismatchReason used when a peer has a different genesis
```
qdm12, line 68: nit

```suggestion
	// GenesisMismatchReason is used when a peer has a different genesis
```

qdm12 (approved):

timwu20:
> Looking at substrate code, it looks like they report only when a new peer gets connected.
> 
> Doesn't `validateBlockAnnounceHandshake` run a lot more frequently?
> 
> Also, it might be worth adding some tests for this as well (not sure how)!

This is only when we receive a handshake and there's a new peer detected ([line](https://github.com/ChainSafe/gossamer/blob/tim/report-peer-additions/dot/network/notifications.go#L188)). 

I tested this manually, I'm not sure it's worth the extra code to unit test gauge reporting.

PR #2267, opened 2022-02-03
fix(dot/peerset): remove race conditions from `peerset` package

### LINKED ISSUE #2266: trying to access a nil pointer at `PeersState.lastConnectedAndDiscovered`
## Describe the bug

Inside the `./dot/peerstate/peerstate.go` file the method `lastConnectedAndDiscovered` calls the method `getNode` with the current condition:

```go
node, err := ps.getNode(peerID)
if err != nil && node.state[set] == notConnected {
    return node.lastConnected[set]
}
```

However, this condition is not correct because `ps.getNode` returns `nil, ErrPeerDoesNotExist` if the peerID is not in the map, so the expression `node.state[...` will cause a nil pointer panic

## Possible Solution

Return the error

```go
node, err := ps.getNode(peerID)
if err != nil {
   return time.Time{}, err
}

if node.state[set] == notConnected {
    return node.lastConnected[set], nil
}

return time.Now(), nil
```

### LINKED ISSUE #2268: dot/peerset: remove race conditions
## Description

- there are currently race conditions in the peerset package as can be seen with `go test -race ./dot/peerset/...`
- fix these race conditions and update tests

EclesioMeloJunior:
## Changes

- Remove race conditions from `peerset` package

## Tests

```
go test -race -timeout 1m ./dot/peerset/...
```

## Issues

- Closes #2268
- Closes #2266
- Unblocks #2105

## Primary Reviewer

- @timwu20


dot/peerset/peerstate.go
```diff
@@ -171,78 +169,111 @@ func (ps *PeersState) peerStatus(set int, peerID peer.ID) string {
...
 		peerIDs = append(peerIDs, k)
 	}
 	return peerIDs
 }
 
+type connectedPeerReputation struct {
+	peerID     peer.ID
+	reputation Reputation
+}
+
```
noot, line 186: it looks like this is only used in the `sortedPeers` function, can you define it in there?

dot/peerset/peerstate.go
```diff
@@ -110,10 +102,12 @@ type PeersState struct {
 	// since, single Info can also manage the flow.
 	sets []Info
 
-	mu sync.Mutex
+	peerStateRWMutex sync.RWMutex
```
noot, line 105: why not just imbed it? then you can just do `ps.Lock()` and `ps.Unlock()`

noot, line 105:
```suggestion
	sync.RWMutex
```

dot/peerset/peerset.go
```diff
@@ -163,22 +115,22 @@ func newReputationChange(value Reputation, reason string) ReputationChange {
 
 // PeerSet is a container for all the components of a peerSet.
 type PeerSet struct {
-	peerState *PeersState
+	peerSetMutex sync.Mutex
```
noot, line 118:
```suggestion
	sync.Mutex
```

noot (approved):
nice work!

dot/peerset/peerset_test.go
```diff
@@ -4,17 +4,78 @@
...
+
+	// follow the order we defined
+	for i := 1; i <= mock.callCounter; i++ {
+		assert, has := mock.expects[i]
+		require.True(t, has)
+		require.Equal(t, assert.expect, assert.got)
+	}
+}
+
 func TestPeerSetBanned(t *testing.T) {
```
kishansagathiya, line 72: Since, test name is does not clearly say what behaviour we are testing we might want to add a comment

dot/peerset/peerset_test.go
```diff
@@ -25,96 +86,98 @@ func TestPeerSetBanned(t *testing.T) {
...
 	// wait a bit for the node's reputation to go above the threshold.
 	time.Sleep(time.Millisecond * 1200)
 
+	mock.ExpectByCall(Message{Status: Accept, setID: 0x0, PeerID: "testPeer1"})
 	// try again. This time the node should be accepted.
 	handler.Incoming(0, peer1)
-	require.NoError(t, err)
-	checkMessageStatus(t, <-ps.resultMsgCh, Accept)
+
+	mock.Assert(t)
```
kishansagathiya, line 106: Instead of making all assertions together, it would be nicer to make assertions after each respective function.

Also, since assertions are happening in mockProcessMessage.Assert method, it does not give us a clear picture of what is being tested. I think it would be better if you can move those assertions in individual tests.

This applies to other tests in this file as well.

You could bring back `checkMessageStatus` in individual tests if possible.

kishansagathiya (changes requested):
This is great!!

For most things I only have minor comments apart from `peerset_test.go`. 
Tests is `peerset_test.go` do what they are supposed to but what they are testing is not obvious. It can probably be better structured.
Once that is fixed, I think this is good to go!

EclesioMeloJunior:
@kishansagathiya, I chose for replace the function signature property by a interface this way I can use the gomock to create and assert the calls!

dot/peerset/peerset.go
```diff
@@ -665,20 +687,14 @@ func (ps *PeerSet) disconnect(setIdx int, reason DropReason, peers ...peer.ID) e
...
-func (ps *PeerSet) start(ctx context.Context, actionQueue chan action) {
-	ps.actionQueue = actionQueue
-
-	ps.resultMsgCh = make(chan Message, msgChanSize)
+func (ps *PeerSet) start(ctx context.Context, processor MessageProcessor) {
+	ps.processor = processor
 	go ps.doWork(ctx)
 }
 
 func (ps *PeerSet) doWork(ctx context.Context) {
```
qdm12, line 695: Maybe rename `doWork` to `periodicallyAllocateSlots`?

dot/peerset/peerstate.go
```diff
@@ -88,19 +88,11 @@ func newNode(n int) *node {
 	}
 }
 
-func (n *node) getReputation() Reputation {
-	return n.rep
```
qdm12, line 92: nit rename the `rep` field to `reputation`? Now that we use it in other files, it's a bit non-intuitive of what it is.

dot/network/service.go
```diff
@@ -691,16 +690,15 @@ func (s *Service) ReportPeer(change peerset.ReputationChange, p peer.ID) {
 }
 
 func (s *Service) startPeerSetHandler() {
-	s.host.cm.peerSetHandler.Start(s.ctx)
+	s.host.cm.peerSetHandler.Start(s.ctx, s)
```
qdm12, line 693: That's mega strange :smile: We pass the service context AND the service to a sub-sub-sub field of the service... 

Fine with the context, but passing the service itself is just weird i think

dot/peerset/handler.go
```diff
@@ -30,80 +30,82 @@ func NewPeerSetHandler(cfg *ConfigSet) (*Handler, error) {
...
+
 // AddReservedPeer adds reserved peer into peerSet.
 func (h *Handler) AddReservedPeer(setID int, peers ...peer.ID) {
-	h.actionQueue <- action{
-		actionCall: addReservedPeer,
-		setID:      setID,
-		peers:      peers,
+	err := h.peerSet.addReservedPeers(setID, peers...)
+	if err != nil {
+		msg := fmt.Sprintf(logStringPattern, addReservedPeer, setID, nil, stringfyPeers(peers))
```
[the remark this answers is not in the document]
EclesioMeloJunior: reputation change is a struct that contains the value and the reason, I've created a function `String()` to `ReputationChange` struct and now instead of nil we pass a string

dot/peerset/handler.go
```diff
@@ -30,80 +30,82 @@ func NewPeerSetHandler(cfg *ConfigSet) (*Handler, error) {
...
 // AddPeer adds peer to peerSet.
 func (h *Handler) AddPeer(setID int, peers ...peer.ID) {
-	h.actionQueue <- action{
-		actionCall: addToPeerSet,
-		setID:      setID,
-		peers:      peers,
+	err := h.peerSet.addPeer(setID, peers)
+	if err != nil {
+		msg := fmt.Sprintf(logStringPattern, addToPeerSet, setID, nil, stringfyPeers(peers))
+		logger.Errorf("failed to do action %s on peerSet: %s", msg, err)
```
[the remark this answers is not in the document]
EclesioMeloJunior: I think this is out of scope here, but it could be a discussion subject and I'm not sure why we chose not to return errors.

dot/peerset/peerset.go
```diff
@@ -586,27 +589,43 @@ func (ps *PeerSet) incoming(setID int, peers ...peer.ID) error {
...
-		case p.getReputation() < BannedThresholdValue:
-			ps.resultMsgCh <- Message{
+		case nodeReputation < BannedThresholdValue:
+			rejectMessage := Message{
 				Status: Reject,
 				setID:  uint64(setID),
 				PeerID: pid,
 			}
+
+			ps.processor.Process(rejectMessage)
```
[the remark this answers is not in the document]
EclesioMeloJunior: Good call!

dot/peerset/peerset.go
```diff
@@ -647,12 +666,15 @@ func (ps *PeerSet) disconnect(setIdx int, reason DropReason, peers ...peer.ID) e
...
 		}
-		ps.resultMsgCh <- Message{
+
+		dropMessage := Message{
 			Status: Drop,
 			setID:  uint64(setIdx),
 			PeerID: pid,
 		}
 
+		ps.processor.Process(dropMessage)
```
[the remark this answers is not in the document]
EclesioMeloJunior: Yes, but is not in this PR scope. A little bit of context here: the processor's responsibility is to connect/drop/reject a peer (a.k.a the network layer) given a message from the `peerset` package, so we can implement specific methods that execute those actions without a central method `Process`! I think it is worth creating a ticket to split the `Processor.Process` into `Drop` and `Connect` methods @danforbes @noot fyi

EclesioMeloJunior:
@qdm12 changed the actual implementation:
- Before we were invoking the `PeerSetHandler.Start(ctx, s)` where `s` is the implementation of `MessageProcessor` (`network.Service`)
- Now, the `network/host` implements `MessageProcessor` and we pass the host before initializing the `peerset.Handler`
````
cm.peerSetHandler.SetMessageProcessor(host)
````

File: dot/peerset/peerset.go [hunk 13]
```diff
@@ -665,20 +687,14 @@ func (ps *PeerSet) disconnect(setIdx int, reason DropReason, peers ...peer.ID) e
 }
 
 // start handles all the action for the peerSet.
-func (ps *PeerSet) start(ctx context.Context, actionQueue chan action) {
-	ps.actionQueue = actionQueue
-
-	ps.resultMsgCh = make(chan Message, msgChanSize)
+func (ps *PeerSet) start(ctx context.Context, processor MessageProcessor) {
+	ps.processor = processor
 	go ps.doWork(ctx)
```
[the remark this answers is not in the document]
EclesioMeloJunior: Not really, do work don't need the processor

dot/network/host.go
```diff
@@ -437,3 +438,38 @@ func (h *host) closeProtocolStream(pID protocol.ID, p peer.ID) {
...
+		}
+
+		err := h.connect(addrInfo)
+		if err != nil {
+			logger.Warnf("failed to open connection for peer %s: %s", peerID, err)
+			return
+		}
+		logger.Debugf("connection successful with peer %s", peerID)
+	case peerset.Drop, peerset.Reject:
+		err := h.closePeer(peerID)
```
jimjbrettj, line 468: Does closePeer () handle the case where a peer isn't connected (imagining thats the Reject case)?

dot/network/host.go
```diff
@@ -437,3 +438,38 @@ func (h *host) closeProtocolStream(pID protocol.ID, p peer.ID) {
...
+// Process will connect, drop or reject a peer based on a peerset message
+func (h *host) Process(msg peerset.Message) {
+	peerID := msg.PeerID
+	if peerID == "" {
+		logger.Errorf("found empty peer id in peerset message")
+		return
+	}
+	switch msg.Status {
+	case peerset.Connect:
+		addrInfo := h.h.Peerstore().PeerInfo(peerID)
```
qdm12, line 451: h.h, maybe we can rename the `h` field :thinking:

qdm12, line 451: h.h, maybe we can rename the `h` field :thinking:

dot/peerset/handler.go
```diff
@@ -30,80 +30,82 @@ func NewPeerSetHandler(cfg *ConfigSet) (*Handler, error) {
...
 }
 
+// SetReservedOnlyPeer not yet implemented
+func (h *Handler) SetReservedOnlyPeer(setID int, peers ...peer.ID) {
+	// TODO: not yet implemented (#1888)
+	logger.Errorf("failed to do action %s on peerSet: not implemented yet", setReservedOnly)
+}
+
 // AddReservedPeer adds reserved peer into peerSet.
 func (h *Handler) AddReservedPeer(setID int, peers ...peer.ID) {
```
jimjbrettj, line 40: Do we not want any of these functions returning errors?

dot/peerset/handler.go
```diff
@@ -30,80 +30,82 @@ func NewPeerSetHandler(cfg *ConfigSet) (*Handler, error) {
...
-		peers:      peers,
+	err := h.peerSet.addReservedPeers(setID, peers...)
+	if err != nil {
+		msg := fmt.Sprintf(logStringPattern, addReservedPeer, setID, "", stringfyPeers(peers))
+		logger.Errorf("failed to do action %s on peerSet: %s", msg, err)
 	}
 }
 
-// RemoveReservedPeer remove reserved peer from peerSet.
+// RemoveReservedPeer removes reserved peer from peerSet.
```
qdm12, line 48: nit 
```suggestion
// RemoveReservedPeer removes the reserved peers from the peerSet of the handler.
```

dot/peerset/handler.go
```diff
@@ -30,80 +30,82 @@ func NewPeerSetHandler(cfg *ConfigSet) (*Handler, error) {
...
-		peers:      peers,
+	err := h.peerSet.removeReservedPeers(setID, peers...)
+	if err != nil {
+		msg := fmt.Sprintf(logStringPattern, removeReservedPeer, setID, "", stringfyPeers(peers))
+		logger.Errorf("failed to do action %s on peerSet: %s", msg, err)
 	}
 }
 
-// SetReservedPeer set the reserve peer into peerSet
+// SetReservedPeer sets the reserve peer into peerSet
```
qdm12, line 57: nit

```suggestion
// SetReservedPeer sets the reserved peer into the peerSet of the handler
```

dot/peerset/handler.go
```diff
@@ -5,17 +5,17 @@ package peerset
 
 import (
 	"context"
+	"fmt"
+	"strings"
 
 	"github.com/libp2p/go-libp2p-core/peer"
 )
 
+const logStringPattern = "call=%s, set-id=%d, reputation change %s, peers=[%s]"
```
jimjbrettj, line 14: nice

dot/peerset/handler.go
```diff
@@ -113,40 +115,29 @@ func (h *Handler) PeerReputation(peerID peer.ID) (Reputation, error) {
 	if err != nil {
 		return 0, err
 	}
-	return n.getReputation(), nil
+	return n.reputation, nil
+}
+
+// SetMessageProcessor set a processor to process peerset messages
```
qdm12, line 121: nit

```suggestion
// SetMessageProcessor sets the peerset processor of the handler
// to process peerset messages. 
```

dot/peerset/handler.go
```diff
@@ -113,40 +115,29 @@ func (h *Handler) PeerReputation(peerID peer.ID) (Reputation, error) {
...
 
-// Stop closes the actionQueue and result message chan.
-func (h *Handler) Stop() {
-	select {
-	case <-h.closeCh:
-	default:
-		h.cancelCtx()
-		close(h.closeCh)
-		close(h.actionQueue)
+func stringfyPeers(peers peer.IDSlice) string {
```
qdm12, line 136: nit should we have this as a `String()` method on `peer.IDSlice` instead?

dot/peerset/peerset.go
```diff
@@ -24,8 +24,6 @@ const (
 	// forgetAfterTime amount of time between the moment we disconnect
 	// from a node and the moment we remove it from the list.
 	forgetAfterTime = time.Second * 3600 // one hour
-	// default channel size for peerSet.
-	msgChanSize = 100
```
qdm12, line 28: :pray: one less thing to worry about :sweat_smile:

dot/peerset/peerset.go
```diff
@@ -294,15 +282,19 @@ func (ps *PeerSet) updateTime() error {
...
+
 				if lastDiscoveredTime.Add(forgetAfterTime).Second() >= currTime.Second() {
 					continue
 				}
 
 				// forget peer removes the peer from the list of members of the set.
 				err = ps.peerState.forgetPeer(set, peerID)
 				if err != nil {
-					return err
+					return fmt.Errorf("cannot forget peer %s from set %d: %w", peerID, set, err)
```
qdm12, line 297: supernit ideally the `forgetPeer` function should add the context about `set` and `peerID`. But this is fine too, just a tiny advice for future stuff.

dot/peerset/peerset.go
```diff
@@ -370,27 +367,29 @@ func (ps *PeerSet) allocSlots(setIdx int) error {
 			peerState.discover(setIdx, reservePeer)
 		}
 
-		var n *node
-		n, err = ps.peerState.getNode(reservePeer)
+		var node *node
+		node, err = ps.peerState.getNode(reservePeer)
```
qdm12, line 371: nit works the same:

```suggestion
		node, err := ps.peerState.getNode(reservePeer)
```

dot/peerset/peerset.go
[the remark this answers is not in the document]
qdm12: @EclesioMeloJunior I think this is still not done. Also you have pre-define the initial message with the set ID and pid which is common for all cases.

dot/peerset/peerset.go
```diff
@@ -357,7 +354,7 @@ func (ps *PeerSet) reportPeer(change ReputationChange, peers ...peer.ID) error {
 func (ps *PeerSet) allocSlots(setIdx int) error {
 	err := ps.updateTime()
```
jimjbrettj, line 355: nit: `if err := ps.updateTime(); err != nil {` for consistancy

dot/peerset/peerset.go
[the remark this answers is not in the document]
qdm12: Yes out of scope I agree :+1:

dot/peerset/peerset.go
```diff
@@ -665,20 +693,13 @@ func (ps *PeerSet) disconnect(setIdx int, reason DropReason, peers ...peer.ID) e
...
 
 // start handles all the action for the peerSet.
-func (ps *PeerSet) start(ctx context.Context, actionQueue chan action) {
-	ps.actionQueue = actionQueue
-
-	ps.resultMsgCh = make(chan Message, msgChanSize)
-	go ps.doWork(ctx)
+func (ps *PeerSet) start(ctx context.Context) {
+	go ps.periodicallyAllocateSlots(ctx)
 }
```
qdm12, line 698: Maybe remove this and call directly `go ps.periodicallyAllocateSlots(ctx)` from the caller calling `.start(ctx)`?

So it's more explicit it's launching a goroutine, and also saves one function that's not so needed imo.

dot/peerset/peerset_test.go
```diff
@@ -125,58 +120,60 @@ func TestReAllocAfterBanned(t *testing.T) {
...
 	time.Sleep(time.Millisecond * 100)
 
-	checkMessageStatus(t, <-ps.resultMsgCh, Drop)
-
 	// Check that an incoming connection from that node gets refused.
-
+	processor.EXPECT().Process(Message{Status: Connect, setID: 0, PeerID: peer1})
 	handler.Incoming(0, peer1)
-	checkMessageStatus(t, <-ps.resultMsgCh, Reject)
+	time.Sleep(time.Second * 2)
```
qdm12, line 134: Why do we need a sleep at the end of the test??

dot/peerset/peerstate.go
```diff
@@ -180,69 +181,99 @@ func (ps *PeersState) peers() []peer.ID {
...
-		Node   *node
+	type connectedPeerReputation struct {
+		peerID     peer.ID
+		reputation Reputation
 	}
 
-	var ss []kv
-	for k, v := range ps.nodes {
-		state := v.state[idx]
+	connectedPeersReps := make([]connectedPeerReputation, 0)
```
qdm12, line 197:
```suggestion
	connectedPeersReps := make([]connectedPeerReputation, 0, len(ps.nodes))
```

or just don't use a `make`.

dot/peerset/peerstate.go
```diff
@@ -180,69 +181,99 @@ func (ps *PeersState) peers() []peer.ID {
...
+func (ps *PeersState) updateReputationByTick(peerID peer.ID) (after Reputation, err error) {
+	ps.Lock()
+	defer ps.Unlock()
+
+	node, has := ps.nodes[peerID]
+	if !has {
+		return 0, ErrPeerDoesNotExist
+	}
+
+	after = reputationTick(node.reputation)
```
qdm12, line 231: nit maybe rename `after` to `newReputation`? I thought it was some kind of time at a quick glance.

dot/peerset/peerset.go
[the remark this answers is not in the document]
jimjbrettj: I like this idea

dot/peerset/peerset.go
```diff
@@ -645,18 +670,21 @@ func (ps *PeerSet) disconnect(setIdx int, reason DropReason, peers ...peer.ID) e
...
+
+		dropMessage := Message{
 			Status: Drop,
 			setID:  uint64(setIdx),
 			PeerID: pid,
 		}
 
+		ps.processor.Process(dropMessage)
+
 		// TODO: figure out the condition of connection refuse.
```
jimjbrettj, line 684: is this still a todo and/or is there an issue for it?

dot/peerset/peerstate.go
```diff
@@ -258,61 +289,73 @@ func (ps *PeersState) hasFreeIncomingSlot(set int) bool {
...
 	}
 
 	delete(ps.sets[idx].noSlotNodes, peerID)
-	n, err := ps.getNode(peerID)
-	if err != nil {
-		return fmt.Errorf("could not get node for peer id %s: %w", peerID, err)
+
+	node, has := ps.nodes[peerID]
+	if !has {
+		return fmt.Errorf("could not get node for peer id %s: %w", peerID, ErrPeerDoesNotExist)
```
qdm12, line 331: a bit of a SuPeRNiT...

`could not get node for peer id` should in the sentinel error `ErrPeerDoesNotExist`, which may be renamed to `ErrNodeNotFoundForPeer` to have i.e.:

```suggestion
		return fmt.Errorf("%w: for peer id %s", ErrNodeNotFoundForPeer, peerID)
```

wrap an error as the rightmost part, but wrap a first error as the leftmost part.

qdm12 (approved):
A bunch of last nits, but looking good :+1:

dot/peerset/handler.go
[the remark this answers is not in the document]
qdm12: Yes out of scope, never mind

jimjbrettj (changes requested):

dot/peerset/handler.go
EclesioMeloJunior: the `peer.IDSlice` is from `libp2p` external package

EclesioMeloJunior: Btw, I found useful to make a PR to add this -> https://github.com/libp2p/go-libp2p-core/pull/238

dot/peerset/peerset.go
EclesioMeloJunior: I didn't understand correctly what you mean by `the context about "set" and "peerID"`, is this related to the error message?

dot/peerset/peerset_test.go
```diff
@@ -25,96 +29,87 @@ func TestPeerSetBanned(t *testing.T) {
 
 	// we ban a node by setting its reputation under the threshold.
 	rpc := newReputationChange(BannedThresholdValue-1, "")
+
 	// we need one for the message to be processed.
+	processor.EXPECT().Process(Message{Status: Drop, setID: 0x0, PeerID: "testPeer1"})
```
kishansagathiya, line 34: There isn't any assertion after this point. Is this deliberate?

dot/peerset/peerset_test.go
```diff
@@ -25,96 +29,87 @@ func TestPeerSetBanned(t *testing.T) {
...
 	t.Parallel()
 
-	handler := newTestPeerSet(t, 0, 2, []peer.ID{bootNode}, []peer.ID{}, false)
+	ctrl := gomock.NewController(t)
+	processor := NewMockMessageProcessor(ctrl)
+	processor.EXPECT().Process(Message{Status: Connect, setID: 0, PeerID: bootNode})
+	processor.EXPECT().Process(Message{Status: Connect, setID: 0, PeerID: reservedPeer})
+	processor.EXPECT().Process(Message{Status: Connect, setID: 0, PeerID: reservedPeer2})
+
+	handler := newTestPeerSet(t, 0, 2, []peer.ID{bootNode}, []peer.ID{}, false, processor)
```
kishansagathiya, line 59: I think you would want to check for more things here. Check that reserved peers get added in `handler.peerSet.reservedNode` and noSlotNodes get added.

The previous decoupled code did seem a bit tough, but now after looking at some of the tests it does seem like it was more testable. For example, for this tests, we could have just checked for whether certain message is getting sent or not (previously).

dot/peerset/peerset_test.go
```diff
@@ -125,58 +120,60 @@ func TestReAllocAfterBanned(t *testing.T) {
...
+	processor.EXPECT().Process(Message{Status: Reject, setID: 0, PeerID: peer1})
 	handler.ReportPeer(rep, peer1)
 	time.Sleep(time.Millisecond * 100)
 
-	checkMessageStatus(t, <-ps.resultMsgCh, Drop)
-
 	// Check that an incoming connection from that node gets refused.
-
+	processor.EXPECT().Process(Message{Status: Connect, setID: 0, PeerID: peer1})
 	handler.Incoming(0, peer1)
```
kishansagathiya, line 133: put assert/require statements to show impact of handler.ReportPeer and handler.Incoming

dot/peerset/peerset_test.go
```diff
@@ -25,96 +29,87 @@ func TestPeerSetBanned(t *testing.T) {
...
-	handler.AddPeer(0, discovered2)
 
-	time.Sleep(200 * time.Millisecond)
-
-	require.Equal(t, 3, len(ps.resultMsgCh))
-	for len(ps.resultMsgCh) == 0 {
-		checkMessageStatus(t, <-ps.resultMsgCh, Connect)
-	}
+	processor.EXPECT().Process(Message{Status: Connect, setID: 0, PeerID: discovered2})
+	handler.AddPeer(0, discovered2)
```
kishansagathiya, line 104: put assert/require to show impact of addpeer (here and at other places in this test)

dot/peerset/peerset_test.go
```diff
@@ -25,96 +29,87 @@ func TestPeerSetBanned(t *testing.T) {
...
+	processor.EXPECT().Process(Message{Status: Accept, setID: 0, PeerID: incomingPeer})
 	handler.Incoming(0, incomingPeer)
-	checkMessageStatus(t, <-ps.resultMsgCh, Accept)
 
+	processor.EXPECT().Process(Message{Status: Accept, setID: 0, PeerID: incoming2})
 	handler.Incoming(0, incoming2)
-	checkMessageStatus(t, <-ps.resultMsgCh, Accept)
 
+	processor.EXPECT().Process(Message{Status: Reject, setID: 0, PeerID: incoming3})
 	handler.Incoming(0, incoming3)
```
kishansagathiya, line 87: put assert/require to show impact of handler.incoming (here and at other places in this test)

dot/network/host.go
EclesioMeloJunior: Right now Reject case is processed but not the right way I think. The `network.handleConn(conn libp2pnetwork.Conn)` which started the incoming calls the `peersetHandler.Incoming` method which should check if the peer should be accepted or rejected but this answer is sent to the processor instead of returned to the `network.handleConn` to stop the handshake with the peer.

I think we should change the following method:

````
func (s *Service) handleConn(conn libp2pnetwork.Conn) {
	// TODO: currently we only have one set so setID is 0, 
        // change this once we have more set in peerSet.
	s.host.cm.peerSetHandler.Incoming(0, conn.RemotePeer())
```` 

to 
````
func (s *Service) handleConn(conn libp2pnetwork.Conn) {
	// TODO: currently we only have one set so setID is 0, 
        // change this once we have more set in peerSet.
	result := s.host.cm.peerSetHandler.Incoming(0, conn.RemotePeer())
        switch result {
        case Reject: 
            logger.Infof("reject ...")
            return
        }

        // exchange BlockAnnounceHandshake with peer so we can start to
	// sync if necessary.
	prtl, has := s.notificationsProtocols[BlockAnnounceMsgType]
	if !has {
		return
	}

	hs, err := prtl.getHandshake()
        ...
```` 
@noot, @timwu20 what do you think?

File: dot/peerset/peerset_test.go (hunk 27 above)
Comment (EclesioMeloJunior) [reply dot/peerset/peerset_test.go]: Because we need to wait for the `periodicallyAllocSlots` being triggered and reject the connection

File: dot/peerset/peerset.go (hunk 25 above)
Comment (EclesioMeloJunior) [reply dot/peerset/peerset.go:355]: I prefer the split one:
```
err := ps.updateTime()
if err != nil
```
changing the occurrences of `if err := ps.updateTime(); err != nil {`

File: dot/peerset/peerset.go (hunk 30 above)
Comment (EclesioMeloJunior) [reply dot/peerset/peerset.go:682]: I'm not sure, @noot do you know if is this still a todo?

File: dot/peerset/peerset.go (hunk 23 above)
Comment (qdm12) [reply dot/peerset/peerset.go:297]: Yeah about the error wrapping.

For example if you pass argument `arg1` to function `F`, `F` should wrap a sentinel error together with the context it has, in this case `arg1`. The caller of `F` should not wrap the error returned by `F` with context passed to it, but only with context of itself.

File: dot/peerset/peerset_test.go (hunk 27 above)
Comment (qdm12) [reply dot/peerset/peerset_test.go]: And this one is called from where? Isn't there a way to wait for it with some event-driven approach?
A sleep is already bad, but at the end of a test this really looks super sketchy :smile: 
Or at least add a comment why we sleep. I guess we wait for all mock calls to be executed so the mock controller doesn't complain right?

Comment (EclesioMeloJunior) [reply dot/peerset/peerset_test.go]: The `periodicallyAllocSlots` is started in the `newTestPeerSet` helper function when we call `handler.Start()`

File: dot/peerset/peerset_test.go (hunk 32 above)
Comment (EclesioMeloJunior) [reply dot/peerset/peerset_test.go]: Added more assertions!

File: dot/network/host.go (hunk 14 above)
Comment (noot) [reply dot/network/host.go:468]: @EclesioMeloJunior that makes sense, also in the Reject case you would need to call `conn.Close()` to close the connection with the peer

**Comment by EclesioMeloJunior** [comment]
@kishansagathiya I think I've addressed all comments related to improve the peerset assertion tests

File: dot/peerset/peerset.go (hunk 23 above)
Comment (EclesioMeloJunior) [reply dot/peerset/peerset.go:297]: Got it! update to just `fmt.Errorf("cannot forget peer: %w", err)`

File: dot/peerset/handler.go (hunk 16 above)
Comment (EclesioMeloJunior) [reply dot/peerset/handler.go:40]: All these methods were intended to be non-blocking, so the caller executes a handler method that will dispatch a message through a channel, and if an error occurs during the message it just logs. Now I have removed the channel and as all those methods was not expected to return any error I keep the log

File: dot/peerset/test_helpers.go [hunk 37]
```diff
@@ -25,22 +25,28 @@ const (
 	peer2         = peer.ID("testPeer2")
 )
 
-func newTestPeerSet(t *testing.T, in, out uint32, bootNodes, reservedPeers []peer.ID, reservedOnly bool) *Handler {
+const allocTimeDuration = time.Second * 2
```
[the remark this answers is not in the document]
EclesioMeloJunior: No, this const is used by some unit tests but changed the format to `2 * time.Second`

timwu20:
What about the previous code was creating race conditions?

kishansagathiya (commented):
LGTM, I am happy to approve after checks pass

EclesioMeloJunior:
> What about the previous code was creating race conditions?

in the previous code, there were channels that were being created and closed by different goroutines, in addition to not protected maps.

timwu20:
> > What about the previous code was creating race conditions?
> 
> @timwu20 in the previous code, there were channels that were being created and closed by different goroutines, in addition to not protected maps.

I'm concerned that performance will decrease now that we made all the peer changes synchronous.  I kinda preferred the separate goroutine approach to handle these peer actions.  

I'll need to keep an eye on the performance of the node after this gets merged.

dot/peerset/handler.go
```diff
@@ -30,80 +30,79 @@ func NewPeerSetHandler(cfg *ConfigSet) (*Handler, error) {
 	}, nil
 }
 
+// SetReservedOnlyPeer not yet implemented
+func (h *Handler) SetReservedOnlyPeer(setID int, peers ...peer.ID) {
```
timwu20, line 34: I feel like these functions should return an error now, and the caller should log it.

dot/peerset/handler.go
```diff
@@ -30,80 +30,79 @@ func NewPeerSetHandler(cfg *ConfigSet) (*Handler, error) {
...
+	// TODO: not yet implemented (#1888)
+	logger.Errorf("failed to do action %s on peerSet: not implemented yet", setReservedOnly)
+}
+
 // AddReservedPeer adds reserved peer into peerSet.
 func (h *Handler) AddReservedPeer(setID int, peers ...peer.ID) {
```
timwu20, line 40: I think it should return an error and the caller should log the error.

timwu20 (changes requested):

kishansagathiya:
> I'm concerned that performance will decrease now that we made all the peer changes synchronous. I kinda preferred the separate goroutine approach to handle these peer actions.

Is there a way to fix the race condition without refactoring peerset package

dot/network/service.go
```diff
@@ -439,7 +439,31 @@ func (s *Service) sentBlockIntervalTelemetry() {
 
 func (s *Service) handleConn(conn libp2pnetwork.Conn) {
 	// TODO: currently we only have one set so setID is 0, change this once we have more set in peerSet.
-	s.host.cm.peerSetHandler.Incoming(0, conn.RemotePeer())
+	results, err := s.host.cm.peerSetHandler.Incoming(0, conn.RemotePeer())
+	if err != nil {
+		logger.Errorf("cannot accept incoming peer: %w", err)
```
qdm12, line 444:
```suggestion
		logger.Errorf("cannot accept incoming peer: %s", err)
```

dot/network/service.go
```diff
@@ -439,7 +439,31 @@ func (s *Service) sentBlockIntervalTelemetry() {
...
+		err := conn.Close()
+		if err != nil {
+			logger.Errorf("cannot close connection with peer: %w", err)
```
qdm12, line 447:
```suggestion
			logger.Errorf("cannot close connection with peer: %s", err)
```

dot/network/service.go
```diff
@@ -439,7 +439,31 @@ func (s *Service) sentBlockIntervalTelemetry() {
...
+		return
+	}
+
+	incomingPeerResult := results[0]
+	switch incomingPeerResult {
+	case peerset.Drop, peerset.Reject:
+		logger.Warnf("connection rejected with peer %s", conn.RemotePeer())
+		err := conn.Close()
+		if err != nil {
+			logger.Errorf("cannot close connection with peer: %w", err)
```
qdm12, line 463:
```suggestion
			logger.Errorf("cannot close connection with peer: %s", err)
```

dot/peerset/handler.go
```diff
@@ -30,80 +30,79 @@ func NewPeerSetHandler(cfg *ConfigSet) (*Handler, error) {
...
-		peers:      peers,
+	err := h.peerSet.addReservedPeers(setID, peers...)
+	if err != nil {
+		msg := fmt.Sprintf(logStringPattern, addReservedPeer, setID, "", stringfyPeers(peers))
+		logger.Errorf("failed to do action %s on peerSet: %s", msg, err)
 	}
 }
 
-// RemoveReservedPeer remove reserved peer from peerSet.
+// RemoveReservedPeer removes reserved peer from peerSet.
```
qdm12, line 48:
```suggestion
// RemoveReservedPeer removes the reserved peers from the peerSet of the handler.
```

dot/peerset/handler.go
```diff
@@ -30,80 +30,79 @@ func NewPeerSetHandler(cfg *ConfigSet) (*Handler, error) {
...
-		peers:      peers,
+	err := h.peerSet.removeReservedPeers(setID, peers...)
+	if err != nil {
+		msg := fmt.Sprintf(logStringPattern, removeReservedPeer, setID, "", stringfyPeers(peers))
+		logger.Errorf("failed to do action %s on peerSet: %s", msg, err)
 	}
 }
 
-// SetReservedPeer set the reserve peer into peerSet
+// SetReservedPeer sets the reserve peer into peerSet
```
qdm12, line 57:
```suggestion
// SetReservedPeer sets the reserved peer into the peerSet of the handler
```

dot/peerset/handler.go
```diff
@@ -30,80 +30,79 @@ func NewPeerSetHandler(cfg *ConfigSet) (*Handler, error) {
...
 func (h *Handler) AddReservedPeer(setID int, peers ...peer.ID) {
-	h.actionQueue <- action{
-		actionCall: addReservedPeer,
-		setID:      setID,
-		peers:      peers,
+	err := h.peerSet.addReservedPeers(setID, peers...)
+	if err != nil {
+		msg := fmt.Sprintf(logStringPattern, addReservedPeer, setID, "", stringfyPeers(peers))
+		logger.Errorf("failed to do action %s on peerSet: %s", msg, err)
 	}
```
qdm12, line 45: Maybe have a tiny helper function for these formatting lines? They show up quite a bit in this file

In example

```suggestion
	logOnError(err, addReservedPeer, setID, "", peers)
```

dot/peerset/handler.go
```diff
@@ -30,80 +30,79 @@ func NewPeerSetHandler(cfg *ConfigSet) (*Handler, error) {
...
-		setID:      setID,
-	}
-}
-
-// Messages return result message chan.
-func (h *Handler) Messages() chan Message {
-	return h.peerSet.resultMsgCh
+func (h *Handler) Incoming(setID int, peers ...peer.ID) (status []Status, err error) {
+	peersStatus, err := h.peerSet.incoming(setID, peers...)
+	return peersStatus, err
```
qdm12, line 97:
```suggestion
	return h.peerSet.incoming(setID, peers...)
```

dot/peerset/peerset.go
```diff
@@ -316,14 +308,15 @@ func (ps *PeerSet) updateTime() error {
...
+	err := ps.updateTime()
+	if err != nil {
+		return fmt.Errorf("cannot update time: %w", err)
 	}
 
 	for _, pid := range peers {
 		rep, err := ps.peerState.addReputation(pid, change)
 		if err != nil {
-			return err
+			return fmt.Errorf("cannot add reputation (%s) to peer %s: %w", pid, change.String(), err)
```
qdm12, line 319:
```suggestion
			return fmt.Errorf("cannot add reputation (%s) to peer %s: %w", change, pid, err)
```

dot/peerset/peerset.go
```diff
@@ -530,24 +542,26 @@ func (ps *PeerSet) removePeer(setID int, peers ...peer.ID) error {
...
 				PeerID: pid,
 			}
 
+			ps.processor.Process(dropMessage)
+
 			// disconnect and forget
 			err := ps.peerState.disconnect(setID, pid)
 			if err != nil {
-				return err
+				return fmt.Errorf("cannot disconnect peer %s at set %d: %w", pid, setID, err)
```
qdm12, line 556: nit maybe `from` instead of `at? Maybe not though

```suggestion
				return fmt.Errorf("cannot disconnect peer %s from set %d: %w", pid, setID, err)
```

dot/peerset/peerset.go
```diff
@@ -586,31 +606,40 @@ func (ps *PeerSet) incoming(setID int, peers ...peer.ID) error {
...
+			message.Status = Reject
 		default:
-			logger.Debugf("incoming connection accepted from peer %s", pid)
-			ps.resultMsgCh <- Message{
-				Status: Accept,
-				setID:  uint64(setID),
-				PeerID: pid,
+			err := state.tryAcceptIncoming(setID, pid)
+			if err != nil {
+				logger.Errorf("cannot accept incomming peer %pid: %w", pid, err)
```
qdm12, line 630:
```suggestion
				logger.Errorf("cannot accept incomming peer %pid: %s", pid, err)
```

dot/peerset/test_helpers.go
```diff
@@ -25,22 +25,28 @@ const (
 	peer2         = peer.ID("testPeer2")
 )
 
-func newTestPeerSet(t *testing.T, in, out uint32, bootNodes, reservedPeers []peer.ID, reservedOnly bool) *Handler {
+const allocTimeDuration = 2 * time.Second
+
+//go:generate mockgen -destination=mock_message_processor_test.go -package $GOPACKAGE . MessageProcessor
+func newTestPeerSet(t *testing.T, maxIn, maxOut uint32, bootNodes,
```
qdm12, line 31: nit add an empty line so it doesn't show as a comment of the function when hovering over it in an IDE

```suggestion

func newTestPeerSet(t *testing.T, maxIn, maxOut uint32, bootNodes,
```

dot/peerset/peerstate.go
```diff
@@ -258,61 +289,74 @@ func (ps *PeersState) hasFreeIncomingSlot(set int) bool {
...
 
 	// Insert peerStatus
 	ps.sets[idx].noSlotNodes[peerID] = struct{}{}
-	n, err := ps.getNode(peerID)
-	if err != nil {
-		return fmt.Errorf("could not get node for peer id %s: %w", peerID, err)
+
+	node, has := ps.nodes[peerID]
+	if !has {
+		fmt.Println("error here")
```
qdm12, line 305:
```suggestion
```

qdm12 (changes requested):

timwu20 (approved):
the diff of this PR could've been a lot smaller, but I liked how you wrapped a lot of the errors and made the code easier to read.

dot/peerset/peerset.go
```diff
@@ -24,7 +25,7 @@ const (
 	// forgetAfterTime amount of time between the moment we disconnect
 	// from a node and the moment we remove it from the list.
 	forgetAfterTime = time.Second * 3600 // one hour
-	// default channel size for peerSet.
+
 	msgChanSize = 100
```
qdm12, line 29: A bit out of scope, but do we actually need a buffer channel??

dot/peerset/peerset.go
```diff
@@ -729,3 +772,23 @@ func (ps *PeerSet) doWork(ctx context.Context) {
...
 }
+
+func (ps *PeerSet) periodicallyAllocateSlots(ctx context.Context) {
+	ticker := time.NewTicker(ps.nextPeriodicAllocSlots)
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-ctx.Done():
+			// TODO: log context error?
```
qdm12, line 783: nit left TODO, maybe log it as debug?

dot/peerset/peerset_test.go
```diff
@@ -113,73 +211,186 @@ func TestPeerSetDiscovered(t *testing.T) {
...
-	handler := newTestPeerSet(t, 0, 2, []peer.ID{discovered1, discovered2}, nil, false)
 	ps := handler.peerSet
-
 	require.Equal(t, 2, len(ps.resultMsgCh))
 	for len(ps.resultMsgCh) != 0 {
 		checkMessageStatus(t, <-ps.resultMsgCh, Connect)
 	}
 
-	handler.RemovePeer(0, discovered1, discovered2)
+	require.Equal(t, 2, len(ps.peerState.nodes))
```
qdm12, line 274: nit

```suggestion
	require.Len(t, ps.peerState.nodes, 2)
```

dot/peerset/peerset_test.go
```diff
@@ -113,73 +211,186 @@ func TestPeerSetDiscovered(t *testing.T) {
...
+	// when calling Incoming method the peer1 is with status notConnectedPeer
+	// so we update its lastConnected field to time.Now() again
+	n, exists = getNodePeer(t, ps.peerState, peer1)
+	require.True(t, exists)
+
+	currentLastTimeConnected := n.lastConnected[testSetID]
+	require.True(t, lastTimeConnected.Before(currentLastTimeConnected))
+
+	// wait a bit for the node's reputation to go above the threshold.
+	<-time.After(allocTimeDuration + time.Second)
```
qdm12, line 257: nit same thing really

```suggestion
	time.Sleep(allocTimeDuration + time.Second)
```

dot/peerset/peerset_test.go
```diff
@@ -113,73 +211,186 @@ func TestPeerSetDiscovered(t *testing.T) {
 
 func TestReAllocAfterBanned(t *testing.T) {
 	t.Parallel()
-
```
qdm12, line 116: nit I personally like the empty line after `t.Parallel()` to separate the 'run this test in parallel' with the test body. But up to you and please don't change, it doesn't matter :wink:

dot/peerset/peerset_test.go
```diff
@@ -113,73 +211,186 @@ func TestPeerSetDiscovered(t *testing.T) {
...
 
-	handler := newTestPeerSet(t, 0, 2, nil, []peer.ID{reservedPeer, reservedPeer2}, true)
 	ps := handler.peerSet
-
 	require.Equal(t, 2, len(ps.resultMsgCh))
 	for len(ps.resultMsgCh) != 0 {
 		checkMessageStatus(t, <-ps.resultMsgCh, Connect)
 	}
 
+	require.Equal(t, 2, len(ps.reservedNode))
```
qdm12, line 300: nit

```suggestion
	require.Len(t, ps.reservedNode, 2)
```

dot/peerset/peerset_test.go
```diff
@@ -113,73 +211,186 @@ func TestPeerSetDiscovered(t *testing.T) {
...
-	require.Equal(t, len(newRsrPeerSet), len(ps.reservedNode))
+	checkPeerSetReservedNodeCount(t, ps, 2)
 	for _, p := range newRsrPeerSet {
-		require.Contains(t, ps.reservedNode, p)
+		checkReservedNodePeerExists(t, ps, p)
+	}
+}
+
+func getNodePeer(t *testing.T, ps *PeersState, pid peer.ID) (node, bool) {
+	t.Helper()
```
qdm12, line 314: no need to pass the `t` we don't actually use it.

dot/peerset/peerset_test.go
```diff
@@ -113,73 +211,186 @@ func TestPeerSetDiscovered(t *testing.T) {
...
+
+	ps.Lock()
+	defer ps.Unlock()
+
+	_, exists := ps.reservedNode[pid]
+	require.True(t, exists)
+}
+
+func checkPeerIsInNoSlotsNode(t *testing.T, ps *PeersState, pid peer.ID, setID int) {
```
qdm12, line 360:
```suggestion
func checkPeerIsInNoSlotsNode(t *testing.T, ps *PeersState, pid peer.ID, setID int) {
	t.Helper()

```

dot/peerset/peerset_test.go
```diff
@@ -113,73 +211,186 @@ func TestPeerSetDiscovered(t *testing.T) {
...
+	ps.RLock()
+	defer ps.RUnlock()
+
+	_, exists := ps.sets[setID].noSlotNodes[pid]
+	require.True(t, exists)
+}
+
+func checkPeerStateSetNumOut(t *testing.T, ps *PeersState, setID int, expectedNumOut uint32) {
```
qdm12, line 368:
```suggestion
func checkPeerStateSetNumOut(t *testing.T, ps *PeersState, setID int, expectedNumOut uint32) {
	t.Helper()

```

dot/peerset/peerset_test.go
```diff
@@ -113,73 +211,186 @@ func TestPeerSetDiscovered(t *testing.T) {
...
+	ps.RLock()
+	defer ps.RUnlock()
+
+	gotNumOut := ps.sets[testSetID].numOut
+	require.Equal(t, expectedNumOut, gotNumOut)
+}
+
+func checkPeerStateSetNumIn(t *testing.T, ps *PeersState, setID int, expectedNumIn uint32) {
```
qdm12, line 376:
```suggestion
func checkPeerStateSetNumIn(t *testing.T, ps *PeersState, setID int, expectedNumIn uint32) {
	t.Helper()

```

dot/peerset/peerset_test.go
```diff
@@ -113,73 +211,186 @@ func TestPeerSetDiscovered(t *testing.T) {
...
+	ps.RLock()
+	defer ps.RUnlock()
+
+	gotNumIn := ps.sets[testSetID].numIn
+	require.Equal(t, expectedNumIn, gotNumIn)
+}
+
+func checkPeerStateNodeCount(t *testing.T, ps *PeersState, expectedCount int) {
```
qdm12, line 384:
```suggestion
func checkPeerStateNodeCount(t *testing.T, ps *PeersState, expectedCount int) {
	t.Helper()

```

dot/peerset/peerset_test.go
```diff
@@ -113,73 +211,186 @@ func TestPeerSetDiscovered(t *testing.T) {
...
+	ps.RLock()
+	defer ps.RUnlock()
+
+	require.Equal(t, expectedCount, len(ps.nodes))
+}
+
+func checkPeerSetReservedNodeCount(t *testing.T, ps *PeerSet, expectedCount int) {
```
qdm12, line 391:
```suggestion
func checkPeerSetReservedNodeCount(t *testing.T, ps *PeerSet, expectedCount int) {
	t.Helper()

```

qdm12 (changes requested):

kishansagathiya (changes requested):
Much of the code is variable renaming and restructuring of code, which sort of hides the actual problem that you have fixed.

I have commented when I am not on board with changes. I would say, we should not restyle unless it is obviously necessary.

Mostly I am requesting changes for some of the channel related comments I have made.

dot/peerset/peerset.go
```diff
@@ -586,28 +631,36 @@ func (ps *PeerSet) incoming(setID int, peers ...peer.ID) error {
...
+			message.Status = Reject
 		default:
-			logger.Debugf("incoming connection accepted from peer %s", pid)
-			ps.resultMsgCh <- Message{
-				Status: Accept,
-				setID:  uint64(setID),
-				PeerID: pid,
+			err := state.tryAcceptIncoming(setID, pid)
+			if err != nil {
+				logger.Errorf("cannot accept incomming peer %pid: %w", pid, err)
```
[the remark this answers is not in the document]
EclesioMeloJunior: I changed to be able to log the error message from `state.tryAcceptIncoming(setID, pid)`

dot/peerset/peerset.go
EclesioMeloJunior: We have both `actionQueue` and `resultMsgCh` as buffered channels. In fact, I believe, we don't need this `msgChanSize`, should I open an issue to turn them into unbuffered channels?

kishansagathiya (approved):

dot/network/service.go
```diff
@@ -487,22 +487,19 @@ func (s *Service) Stop() error {
...
-mainloop:
 	for {
 		select {
 		case _, hasMore := <-s.closeCh:
 			if !hasMore {
-				break mainloop
+				return nil
 			}
+		case <-s.closeCh:
+			return nil
```
qdm12, line 498: Wait a minuteee.... this looks incredibly wrong i think?

Why do we have two cases pulling from the same channel?

dot/peerset/handler.go
```diff
@@ -12,10 +12,10 @@ import (
 // Handler manages peerSet.
 type Handler struct {
 	actionQueue chan<- action
-	peerSet     *PeerSet
 	closeCh     chan struct{}
+	cancelCtx   context.CancelFunc
 
-	cancelCtx context.CancelFunc
+	peerSet *PeerSet
```
qdm12, line 18: nit keep order as it was, these diffs are quite unnecessary I think.

dot/peerset/peerset.go
qdm12: @EclesioMeloJunior reminder :wink:

dot/peerset/peerset.go
```diff
@@ -332,21 +349,24 @@ func (ps *PeerSet) reportPeer(change ReputationChange, peers ...peer.ID) error {
...
+			}
+
+			ps.resultMsgCh <- Message{
+				Status: Drop,
+				setID:  uint64(i),
+				PeerID: pid,
+			}
+
+			if err = ps.allocSlots(i); err != nil {
+				return fmt.Errorf("could not allocate slots: %w", err)
```
qdm12, line 369: nit add peer id context perhaps?

```suggestion
				return fmt.Errorf("could not allocate slots for peer %s: %w", pid, err)
```

dot/peerset/peerset.go
```diff
@@ -469,23 +494,27 @@ func (ps *PeerSet) removeReservedPeers(setID int, peers ...peer.ID) error {
...
-				return err
+				return fmt.Errorf("cannot disconnect peer %s at set %d: %w", peerID, setID, err)
 			}
 
 			ps.resultMsgCh <- Message{
 				Status: Drop,
 				setID:  uint64(setID),
 				PeerID: peerID,
 			}
+
```
qdm12, line 505: nit

```suggestion
```

dot/peerset/peerset.go
```diff
@@ -539,15 +574,15 @@ func (ps *PeerSet) removePeer(setID int, peers ...peer.ID) error {
 			// disconnect and forget
 			err := ps.peerState.disconnect(setID, pid)
 			if err != nil {
-				return err
+				return fmt.Errorf("cannot disconnect peer %s at set %d: %w", pid, setID, err)
```
qdm12, line 577: pid and setid error context should be added by the `disconnect` method and not here

dot/peerset/peerset.go
```diff
@@ -539,15 +574,15 @@ func (ps *PeerSet) removePeer(setID int, peers ...peer.ID) error {
...
 			}
 
 			if err = ps.peerState.forgetPeer(setID, pid); err != nil {
-				return err
+				return fmt.Errorf("cannot forget peer %s from set %d: %w", pid, setID, err)
```
qdm12, line 581: pid and setid error context should be added by the `forgetPeer` method and not here

dot/peerset/peerset.go
```diff
@@ -539,15 +574,15 @@ func (ps *PeerSet) removePeer(setID int, peers ...peer.ID) error {
...
 			}
 		} else if status == notConnectedPeer {
 			if err := ps.peerState.forgetPeer(setID, pid); err != nil {
-				return err
+				return fmt.Errorf("cannot forget peer %s from set %d: %w", pid, setID, err)
```
qdm12, line 585: pid and setid error context should be added by the `forgetPeer` method and not here

dot/peerset/peerset.go
qdm12: @EclesioMeloJunior reminder

dot/peerset/peerset.go
```diff
@@ -586,26 +622,42 @@ func (ps *PeerSet) incoming(setID int, peers ...peer.ID) error {
...
+					PeerID: pid,
+				}
+			} else {
+				logger.Debugf("incoming connection accepted from peer %s", pid)
+
+				ps.resultMsgCh <- Message{
+					Status: Accept,
+					setID:  uint64(setID),
+					PeerID: pid,
+				}
```
qdm12, line 660: nit can you instead set the message in each test case:

```go
message = Message{...}
```

and then send it in the channel at the end of the `switch` with `ps.resultMsgCh <- message`

You can even pre-set a common message with the peer id and set id to reduce the amount of lines/repetition.

dot/peerset/peerset.go
```diff
@@ -645,8 +697,9 @@ func (ps *PeerSet) disconnect(setIdx int, reason DropReason, peers ...peer.ID) e
 		state.nodes[pid] = n
 
 		if err = state.disconnect(setIdx, pid); err != nil {
-			return err
+			return fmt.Errorf("cannot disconnect peer %s at set %d: %w", pid, setIdx, err)
```
qdm12, line 700: setid and peer context should be wrapped in `disconnect`, not here

dot/peerset/peerset.go
```diff
@@ -656,7 +709,7 @@ func (ps *PeerSet) disconnect(setIdx int, reason DropReason, peers ...peer.ID) e
 		// TODO: figure out the condition of connection refuse.
 		if reason == RefusedDrop {
 			if err = ps.removePeer(setIdx, pid); err != nil {
-				return err
+				return fmt.Errorf("cannot remove peer %s at set %d: %w", pid, setIdx, err)
```
qdm12, line 712: setid and peer context should be wrapped in `removePeer`, not here

dot/peerset/peerset.go
```diff
@@ -667,32 +720,19 @@ func (ps *PeerSet) disconnect(setIdx int, reason DropReason, peers ...peer.ID) e
...
 			return
-		case <-ticker.C:
-			l := ps.peerState.getSetLength()
-			for i := 0; i < l; i++ {
-				if err := ps.allocSlots(i); err != nil {
-					logger.Warnf("failed to do action on peerSet: %s", err)
-				}
-			}
 		case act, ok := <-ps.actionQueue:
+
```
qdm12, line 735: nit

```suggestion
```

dot/peerset/peerset.go
```diff
@@ -729,3 +769,29 @@ func (ps *PeerSet) doWork(ctx context.Context) {
...
+		ticker.Stop()
+		close(ps.resultMsgCh)
+	}()
+
+	for {
+		select {
+		case <-ctx.Done():
+			err := ctx.Err()
+			if err != nil {
+				logger.Debugf("peerset context finished with error: %w", err)
```
qdm12, line 786: a context always finishes with an error :wink: 

Also don't use `%w` for logging errors, but `%s`

```suggestion
				logger.Debugf("peerset slot allocation exiting: %s", err)
```

dot/peerset/peerset.go
```diff
@@ -729,3 +769,29 @@ func (ps *PeerSet) doWork(ctx context.Context) {
...
+			}
+			return
+		case <-ticker.C:
+			for i := 0; i < ps.peerState.getSetLength(); i++ {
+				if err := ps.allocSlots(i); err != nil {
+					logger.Warnf("failed to do action on peerSet: %s", err)
```
qdm12, line 792: nit this seems more informative/updated I think:

```suggestion
					logger.Warnf("failed to allocate slots: %s", err)
```

dot/peerset/peerset.go
```diff
@@ -729,3 +769,29 @@ func (ps *PeerSet) doWork(ctx context.Context) {
...
+	for {
+		select {
+		case <-ctx.Done():
+			err := ctx.Err()
+			if err != nil {
+				logger.Debugf("peerset context finished with error: %w", err)
+			}
+			return
+		case <-ticker.C:
+			for i := 0; i < ps.peerState.getSetLength(); i++ {
```
qdm12, line 790: nit maybe you can rename `i` to `setID` or something more explicit? :pray:

dot/peerset/peerset_test.go
```diff
@@ -11,57 +11,103 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-func TestPeerSetBanned(t *testing.T) {
+const testSetID = 0
```
qdm12, line 14: nit inline it in each test where it's needed to reduce the Go package API pollution (even if it's just the test API one)

dot/peerset/peerset_test.go
```diff
@@ -11,57 +11,103 @@ import (
...
+
 	// we ban a node by setting its reputation under the threshold.
 	rpc := newReputationChange(BannedThresholdValue-1, "")
+
 	// we need one for the message to be processed.
+	// report peer will disconnect the peer and set the `lastConnected` to time.Now
 	handler.ReportPeer(rpc, peer1)
-	time.Sleep(time.Millisecond * 100)
 
+	time.Sleep(100 * time.Millisecond)
```
qdm12, line 43: nit can we have this event-driven? Can this sleep be removed?
Maybe by mocking `time.Now` wherever it's used?

dot/peerset/peerset_test.go
```diff
@@ -11,57 +11,103 @@ import (
...
+	reservedPeers := []struct {
+		peerID peer.ID
+	}{
+		{
+			peerID: reservedPeer,
+		},
+		{
+			peerID: reservedPeer2,
+		},
+	}
```
qdm12, line 87: nit make it a slice of peer.ID

dot/peerset/peerset_test.go
```diff
@@ -75,36 +121,91 @@ func TestAddReservedPeers(t *testing.T) {
 
 func TestPeerSetIncoming(t *testing.T) {
 	t.Parallel()
+	handler := newTestPeerSet(t, 2, 1, []peer.ID{bootNode},
+		[]peer.ID{}, false)
 
-	handler := newTestPeerSet(t, 2, 1, []peer.ID{bootNode}, []peer.ID{}, false)
```
qdm12, line 79: nit unneeded change, revert it please

dot/peerset/peerset_test.go
```diff
@@ -113,73 +214,196 @@ func TestPeerSetDiscovered(t *testing.T) {
...
+	// when calling Incoming method the peer1 is with status notConnectedPeer
+	// so we update its lastConnected field to time.Now() again
+	n, exists = getNodePeer(ps.peerState, peer1)
+	require.True(t, exists)
+
+	currentLastTimeConnected := n.lastConnected[testSetID]
+	require.True(t, lastTimeConnected.Before(currentLastTimeConnected))
+
+	// wait a bit for the node's reputation to go above the threshold.
+	time.Sleep(allocTimeDuration + time.Second)
```
qdm12, line 261: Can we have this event-driven? Or at least config the allocTimeDuration to be a `time.Nanosecond`?

dot/peerset/peerset_test.go
```diff
@@ -113,73 +214,196 @@ func TestPeerSetDiscovered(t *testing.T) {
...
+	checkPeerStateSetNumOut(t, ps.peerState, testSetID, 1)
 	checkMessageStatus(t, <-ps.resultMsgCh, Connect)
 }
 
 func TestRemovePeer(t *testing.T) {
 	t.Parallel()
+	handler := newTestPeerSet(t, 0, 2, []peer.ID{discovered1, discovered2},
+		nil, false)
 
-	handler := newTestPeerSet(t, 0, 2, []peer.ID{discovered1, discovered2}, nil, false)
```
qdm12, line 147: nit unneeded deltas

dot/peerset/peerset_test.go
```diff
@@ -113,73 +214,196 @@ func TestPeerSetDiscovered(t *testing.T) {
...
+	checkPeerStateNodeCount(t, ps.peerState, 0)
+	checkPeerStateSetNumOut(t, ps.peerState, testSetID, 0)
 }
 
 func TestSetReservePeer(t *testing.T) {
 	t.Parallel()
+	handler := newTestPeerSet(t, 0, 2, nil, []peer.ID{reservedPeer, reservedPeer2},
+		true)
 
-	handler := newTestPeerSet(t, 0, 2, nil, []peer.ID{reservedPeer, reservedPeer2}, true)
```
qdm12, line 169: nit unneeded deltas

dot/peerset/test_helpers.go
```diff
@@ -25,15 +25,20 @@ const (
 	peer2         = peer.ID("testPeer2")
 )
 
-func newTestPeerSet(t *testing.T, in, out uint32, bootNodes, reservedPeers []peer.ID, reservedOnly bool) *Handler {
+const allocTimeDuration = 2 * time.Second
```
qdm12, line 28: Is it possible to have this smaller? Like 1 nanosecond small?

qdm12 (changes requested):

dot/peerset/peerset.go
EclesioMeloJunior: Is not needed as the `ps.allocSlots`error already has the context of some peer

EclesioMeloJunior:
@qdm12 I don't have addressed the `allocTimeDuration` test value because looks like some changes in some tests and maybe in production code to remove the `time.Sleep()` and use an event-driven approach so I've created an issue https://github.com/ChainSafe/gossamer/issues/2520 in order to address your suggestion

dot/peerset/peerset.go
qdm12: @EclesioMeloJunior I don't think this is done for that particular line :wink:

dot/peerset/peerset.go
```diff
@@ -316,14 +332,15 @@ func (ps *PeerSet) updateTime() error {
...
+	err := ps.updateTime()
+	if err != nil {
+		return fmt.Errorf("cannot update time: %w", err)
 	}
 
 	for _, pid := range peers {
 		rep, err := ps.peerState.addReputation(pid, change)
 		if err != nil {
-			return err
+			return fmt.Errorf("cannot add reputation (%s) to peer %s: %w", change, pid, err)
```
qdm12, line 343: nit the change and pid context should be added in `addReputation` instead of here

dot/peerset/peerset.go
```diff
@@ -332,21 +349,24 @@ func (ps *PeerSet) reportPeer(change ReputationChange, peers ...peer.ID) error {
...
-					setID:  uint64(i),
-					PeerID: pid,
-				}
-				if err = ps.allocSlots(i); err != nil {
-					return err
-				}
+			// disconnect peer
+			err = ps.peerState.disconnect(i, pid)
+			if err != nil {
+				return fmt.Errorf("cannot disconnect peer %s at set %d: %w", pid, i, err)
```
qdm12, line 359: nit the `pid` and `i` context should be added in the `disconnect` method instead of here

dot/peerset/peerset.go
qdm12: I agree with the split one :+1: 
Although I am an old [`ifshort`](https://github.com/esimonov/ifshort) user, so I have tried both teams :smile:

dot/peerset/peerset.go
```diff
@@ -370,20 +390,19 @@ func (ps *PeerSet) allocSlots(setIdx int) error {
 			peerState.discover(setIdx, reservePeer)
 		}
 
-		var n *node
-		n, err = ps.peerState.getNode(reservePeer)
+		node, err := ps.peerState.getNode(reservePeer)
 		if err != nil {
-			return err
+			return fmt.Errorf("cannot get node using peer id %s: %w", reservePeer, err)
```
qdm12, line 395: nit the `reservePeer` context should be wrapped by `getNode` instead of here.

dot/peerset/peerset.go
```diff
@@ -469,7 +494,7 @@ func (ps *PeerSet) removeReservedPeers(setID int, peers ...peer.ID) error {
 		if ps.peerState.peerStatus(setID, peerID) == connectedPeer {
 			err := ps.peerState.disconnect(setID, peerID)
 			if err != nil {
-				return err
+				return fmt.Errorf("cannot disconnect peer %s at set %d: %w", peerID, setID, err)
```
qdm12, line 497: nit wrap the error context about the peer id and set id in `disconnect` instead of here

dot/peerset/peerset.go
```diff
@@ -586,28 +631,36 @@ func (ps *PeerSet) incoming(setID int, peers ...peer.ID) error {
...
+		node, has := state.nodes[pid]
+		if has {
+			nodeReputation = node.reputation
+		}
+		state.RUnlock()
+
+		message := Message{
+			setID:  uint64(setID),
+			PeerID: pid,
+		}
```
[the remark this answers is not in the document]
qdm12: I suggested that change, I think it's more readable with the message variable, although yes, a tiny bit less 'local'.

dot/peerset/peerset.go
```diff
@@ -586,28 +621,36 @@ func (ps *PeerSet) incoming(setID int, peers ...peer.ID) error {
...
-				setID:  uint64(setID),
-				PeerID: pid,
-			}
-		case state.tryAcceptIncoming(setID, pid) != nil:
-			ps.resultMsgCh <- Message{
-				Status: Reject,
-				setID:  uint64(setID),
-				PeerID: pid,
-			}
+		case nodeReputation < BannedThresholdValue:
```
qdm12, line 640: nit you could change that switch to an if else

dot/peerset/peerset.go
```diff
@@ -729,3 +759,29 @@ func (ps *PeerSet) doWork(ctx context.Context) {
...
+		close(ps.resultMsgCh)
+	}()
+
+	for {
+		select {
+		case <-ctx.Done():
+			err := ctx.Err()
+			if err != nil {
+				logger.Debugf("peerset slot allocation exiting: %s", err)
+			}
```
qdm12, line 777: this will always be not nil:

```suggestion
			logger.Debugf("peerset slot allocation exiting: %s", ctx.Err())
```

dot/peerset/peerstate.go
```diff
@@ -180,69 +181,99 @@ func (ps *PeersState) peers() []peer.ID {
...
 func (ps *PeersState) highestNotConnectedPeer(set int) peer.ID {
-	var maxRep = math.MinInt32
-	var peerID peer.ID
-	for id, n := range ps.nodes {
-		if n.state[set] != notConnected {
+	ps.RLock()
+	defer ps.RUnlock()
+
+	maxRep := math.MinInt32
+	var highestPeerID peer.ID
```
qdm12, line 262: nit maybe name your returned value to save that one line and make things a tiny bit clearer?

qdm12 (approved):
Just a few nits here and there :+1:

dot/peerset/peerset.go
qdm12: Maybe open an issue to investigate all buffered channels requirement? You can add that one as a first bullet point?

dot/peerset/peerset.go
```diff
@@ -318,7 +318,7 @@ func (ps *PeerSet) updateTime() error {
 				// forget peer removes the peer from the list of members of the set.
 				err = ps.peerState.forgetPeer(set, peerID)
 				if err != nil {
-					return fmt.Errorf("cannot forget peer %s from set %d: %w", peerID, set, err)
+					return fmt.Errorf("cannot forget peer: %w", err)
```
qdm12, line 321: nit add the set context in the `forgetPeer` error `fmt.Errorf("%w: for peer id %s", ErrPeerDoesNotExist, peerID)`

dot/peerset/peerset.go
```diff
@@ -340,7 +340,7 @@ func (ps *PeerSet) reportPeer(change ReputationChange, peers ...peer.ID) error {
 	for _, pid := range peers {
 		rep, err := ps.peerState.addReputation(pid, change)
 		if err != nil {
-			return fmt.Errorf("cannot add reputation (%s) to peer %s: %w", change, pid, err)
+			return fmt.Errorf("cannot add reputation: %w", err)
```
qdm12, line 343: nit add the change to the error context in `addReputation`: `fmt.Errorf("%w: for peer id %s", ErrPeerDoesNotExist, peerID)`

dot/peerset/peerset.go
```diff
@@ -356,7 +356,7 @@ func (ps *PeerSet) reportPeer(change ReputationChange, peers ...peer.ID) error {
 			// disconnect peer
 			err = ps.peerState.disconnect(i, pid)
 			if err != nil {
-				return fmt.Errorf("cannot disconnect peer %s at set %d: %w", pid, i, err)
+				return fmt.Errorf("cannot disconnect: %w", err)
```
qdm12, line 359: nit add the set context to the error wrapping in `fmt.Errorf("%w: for peer id %s", ErrPeerDoesNotExist, peerID)` in the disconnect method

dot/peerset/peerset.go
```diff
@@ -402,7 +402,7 @@ func (ps *PeerSet) allocSlots(setIdx int) error {
 		}
 
 		if err = peerState.tryOutgoing(setIdx, reservePeer); err != nil {
-			return fmt.Errorf("cannot set peer %s from set %d as outgoing: %w", reservePeer, setIdx, err)
+			return fmt.Errorf("cannot set as outgoing: %w", err)
```
qdm12, line 405: nit add the setidx context to the error wrapping in `tryOutgoing`

qdm12 (approved):
just a few additional nit changes to do :pray: 
But otherwise ready to merge :rocket:

jimjbrettj (approved):

PR #2282, opened 2022-02-09
feat(cross-client): create docker-compose.yml for local devnet 

### LINKED ISSUE #2149: create docker-compose.yml for local `gssmr` devnet
## Issue summary

Create docker-compose.yml file for a local cross client devnet which orchestrates 2 gossamer nodes, and 1 substrate node.  Would require a new docker context so adding a new folder to `\devnet` would suffice.

- Ensure metrics for all nodes are flowing to datadog with proper keys as tags.
- Ensure that block production and network messaging is occurring
- Create any issues that halt block production or networking that stop the consistency of the network

## Other Information
- https://github.com/ChainSafe/gossamer/pull/2008

EclesioMeloJunior:
## Changes

- Add `devnet/cross-client/docker-compose.yml` file to setup 2 gossamer nodes and 1 substrate node
- Update the genesis file to use a reasonable runtime version

## Tests

```
docker-compose -f ./devnet/cross-client/docker-compose.yml up --build
```

## Issues

- Closes #2149
 
## Primary Reviewer

- @timwu20


devnet/cross-client/docker-compose.yml
```diff
@@ -0,0 +1,47 @@
...
+      dockerfile: devnet/alice.Dockerfile
+      args:
+        DD_API_KEY: ${DD_API_KEY}
+    ports:
+      - 7001
+      - 8545
+      - 8546
+      - 8540
+      - 9876
+      - 6060
```
timwu20, line 15: do we expose pprof on bob?

EclesioMeloJunior: yes, bind pprof port (`6060`) to bob gossamer node.

devnet/cross-client/docker-compose.yml
```diff
@@ -0,0 +1,49 @@
+services: 
+  alice:
+    platform: linux/amd64
+    build:
+      context: ./../..
+      dockerfile: devnet/alice.Dockerfile
+      args:
+        DD_API_KEY: ${DD_API_KEY}
+    ports:
```
qdm12, line 9:
```suggestion
    expose:
```

devnet/cross-client/docker-compose.yml
```diff
@@ -0,0 +1,49 @@
...
+
+  bob:
+    platform: linux/amd64
+    build:
+      context: ./../..
+      dockerfile: devnet/bob.Dockerfile
+      args:
+        key: bob
+        DD_API_KEY: ${DD_API_KEY}
+    ports:
```
qdm12, line 25:
```suggestion
    expose:
```

devnet/cross-client/docker-compose.yml
```diff
@@ -0,0 +1,49 @@
...
+
+  charlie:
+    platform: linux/amd64
+    build:
+      context: ./../..
+      dockerfile: devnet/substrate_bob.Dockerfile
+      args:
+        key: charlie
+        DD_API_KEY: ${DD_API_KEY}
+    ports:
```
qdm12, line 43:
```suggestion
    expose:
```

qdm12 (approved):
Early approval, but use `expose:` instead of `ports:`, since ports will publish the ports you listed to some random ports on the host, and I don't think we care about this. If you need to publish i.e. the pprof port, then having it as `6060:6060` in a `ports:` block.

qdm12:
> since all these ports are being exposed by the docker image, do we need the `expose` or `ports` in the docker-compose file?

It's nice for documentation, I'd suggest to keep it.

> Just a note here: we may need to bind the rpc/ws port to host ports to use/connect with polkadot-js website.

If it's Gossamer initiating the connection/handshake, then there is no need to publish a port. I would think this is the case here?

timwu20 (approved):

devnet/cross-client/docker-compose.yml
```diff
@@ -0,0 +1,49 @@
...
+      args:
+        key: bob
+        DD_API_KEY: ${DD_API_KEY}
+    ports:
+      - 7001
+      - 8545
+      - 8546
+      - 8540
+      - 9876
+      - 6060
```
qdm12, line 31: If you do want to publish ports on the host, then change them to a mapping. This will just port map them to random available ports on the host, I don't really see the use here.

For example:

```yml
version: '3'

services:
    alpine:
        image: alpine
        init: true
        command: [ "sleep", "100" ]
        ports:
            - 8000
```

```sh
docker-compose up -d
docker ps | grep "8000"
```

Will show

```
5118bd0982d9   alpine                         "sleep 100"              2 seconds ago   Up 2 seconds          0.0.0.0:49154->8000/tcp, :::49154->8000/tcp     workspace-alpine-1
```

qdm12 (changes requested):

qdm12 (approved):

Issue #2281, opened 2022-02-09
Remove all usage of `sync.Map`


qdm12:
`sync.Map` proved troublesome since it would not be clearing deleted tries.
It was rather hard to debunk when doing #1973 even with various profiling.
This may not bring an advantage memory usage-wise, but could also be one / the main cause of our memory problems.
Anyway, it would be good to remove this doubt.
We should remove all usages of it (except when heavily read and no deletion) and use a mutex protected standard map instead.

PR #2286, opened 2022-02-09
chore(dot/state): replace `sync.Map` with map+mutex

qdm12:
## Tests

```
go test ./dot/state
```

## Issues

- #2281


dot/state/hashtoblock.go
```diff
@@ -0,0 +1,45 @@
...
+}
+
+func newHashToBlock() *hashToBlock {
+	return &hashToBlock{
+		mapping: make(map[common.Hash]*types.Block),
+	}
+}
+
+func (h *hashToBlock) get(hash common.Hash) (block *types.Block) {
+	h.mutex.RLock()
```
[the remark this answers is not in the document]
qdm12: No we need to lock the actual `map` so that it's not read and written at the same time (even at different keys) or it would panic.

Essentially this sums it up:

> Before Golang 1.6, concurrent read is OK, concurrent write is not OK, but write and concurrent read is OK. Since Golang 1.6, map cannot be read when it's being written. 

From https://stackoverflow.com/a/38140573

dot/state/hashtoblock.go
```diff
@@ -0,0 +1,45 @@
...
+package state
+
+import (
+	"sync"
+
+	"github.com/ChainSafe/gossamer/dot/types"
+	"github.com/ChainSafe/gossamer/lib/common"
+)
+
+type hashToBlock struct {
```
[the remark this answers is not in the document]
qdm12: Indeed, done in 4f04b115ab71d79d1ee9eabd3e4e79659c15f55e

noot (approved):

dot/state/hashtoblockmap_test.go
```diff
@@ -0,0 +1,327 @@
...
+	}
+}
+
+func Test_hashToBlockMap_threadSafety(t *testing.T) {
+	// This test consists in checking for concurrent access
+	// using the -race detector.
+	t.Parallel()
+
+	var startWg, endWg sync.WaitGroup
+	ctx, cancel := context.WithCancel(context.Background())
```
timwu20, line 267: why do we need this context?

qdm12: To cancel all `runInLoop` goroutines all at once.

```go
		for ctx.Err() == nil {
			f()
		}
```

and the cancel is called by the timer.

To be fair, all this feels like over-engineering but also like the right way to test for data races *quickly*, since it has to be 'tight' and as event-driven as possible

timwu20: I don't really see the need to use a context to handle cancellation.  Could you not just close some channel to signify cancellation?

timwu20 (approved):

dot/state/hashtoblockmap_test.go
qdm12: Yes we can use a struct{} channel that gets closed, it's just that it then becomes:

```go
stop := false
for !stop {
	select {
	case <-stopGoroutines:
		stop = true
	default:
	}
	f()
}
```

Which is longer/more confusing I think.

PR #2287, opened 2022-02-10
fix(dot/state): inject mutex protected tries to states

qdm12:
## Changes

Previously after each finalized block, a trie would not be found from memory and would be loaded from the database.
You can print it out at [this code line](https://github.com/ChainSafe/gossamer/blob/3ae34011ee4d00bc0c439d21e93c121ba7701294/dot/state/storage.go#L131). This caused some memory bloat in the long run.

I initially tried to refactor a bit by injecting the tries (mutex protected already) instead of relying on channels and a goroutine to send hashes to the state, and this has been fixed. I think it was a race condition introduced by the buffered channel + goroutine, where `TrieState` was called too late after the trie got deleted from memory.

## Issues

- #1973 

## Primary Reviewer

- @timwu20

### DIFF
--- a/dot/rpc/modules/dev_integration_test.go
+++ b/dot/rpc/modules/dev_integration_test.go
@@ -42,8 +42,8 @@ func newState(t *testing.T) (*state.BlockState, *state.EpochState) {
 
 	db := state.NewInMemoryDB(t)
 
-	_, _, genesisHeader := genesis.NewTestGenesisWithTrieAndHeader(t)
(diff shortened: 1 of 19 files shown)


timwu20:
Let's just go back to the initial approach of passing around the `*tries` pointer to both `StorageState` and `BlockState`.  I'll create another issue to handle refactoring this part of the code.

timwu20 (approved):

edwardmack (approved):

noot (approved):

Issue #2288, opened 2022-02-11
Use `*bytes.Buffer` instead of `*[]byte` in `dot/network` buffer pool to accommodate larger messages


qdm12:
## Task summary

This shows in the warnings logs when syncing:

```
received message with size 75041 greater than allocated message buffer size 64512
```

This comes from https://github.com/ChainSafe/gossamer/blob/3ae34011ee4d00bc0c439d21e93c121ba7701294/dot/network/utils.go#L198

and it looks like we drop the message. I'm not sure of the implications but we might want to use `*bytes.Buffer` in our sync.Pool which dynamically will resize, instead of using fixed sized `*[]byte`.

PR #2291, opened 2022-02-11
fix(dot/network): resize bytes slice buffer if needed

qdm12:
## Changes

- See why in #2288 

## Issues

- Related to #2288

## Primary Reviewer

- @noot

### DIFF
--- a/dot/network/helpers_test.go
+++ b/dot/network/helpers_test.go
@@ -86,7 +86,7 @@ func (s *testStreamHandler) readStream(stream libp2pnetwork.Stream,
 	}()
 
 	for {
-		tot, err := readStream(stream, msgBytes)
+		tot, err := readStream(stream, &msgBytes)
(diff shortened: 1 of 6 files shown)


kishansagathiya (approved):

noot (changes requested):
why 100kb exactly? we might still receive messages larger than this, we should just change it to dynamically get more buffers if the size is larger

qdm12:
> why 100kb exactly? we might still receive messages larger than this

We receive up to 78KB messages as far as I've seen in the logs. This was an ugly quick fix to not drop them, a better more complex fix would be to use `*bytes.Buffer` as suggested in the issue so we can have dynamic sizes.

I changed it to increase the buffer slice size when needed, although the bytes.Buffer's `Grow` method is really more appropriate here.

noot (approved):

PR #2139, opened 2022-01-04
chore(internal): move packages to internal directory

qdm12:
## Changes

:tipping_hand_man: This PR is designed to be nice with other PRs, and keep history of file movements for easy rebase/merge on development once it gets merged.

:tipping_hand_person: 95% of deltas are import strings updated to have the `internal/` part. File movements are not part of them. There are a few extra changes in logic to find the chains directory in tests in `internal/lib/utils/utils.go`

- [x] Rename `*.out` files to `*.json` to avoid gitignore conflicts
- [x] Do not .gitignore `test_data` paths
- [x] Move `lib`, `dot`, `tests` and `chain` to `internal` directory
- [x] Adapt import paths by adding `/internal` suffix to `dot`, `lib`, `chain` and `tests` import paths
- [x] Improve logic to find the `chain` directory in `internal/lib/utils/utils.go`
- [x] Rebase interactively to have a 3-commits only PR
    - [x] Squash gitignore changes in a single commit
    - [x] Squash file movements in a single commit
    - [x] Squash import paths renamings in a single commit
- [ ] **MERGE** (:warning:) the PR in development, **not** squash and merge so file movements are tracked and people can easily rebase their branches on `development` with `git rebase development`.

## Issues

- #1527 

## Primary Reviewer

- @timwu20

Issue #1527, opened 2021-04-15
Implement Go Standard Package folder structure


timwu20:
## Issue summary

- Many go projects with external packages utilize the `pkg` and `internal` root level folders.  Issue #1072 is looking to move SCALE into it's own package for external use.  Migrating `scale` to the `pkg` folder is a nice compromise and we can always move it out into it's own repo afterwards.
- Rename `lib` to `internal`.  The go compiler enforces that external applications that import packages will not be able to import packages that stored in an `internal` folder.  
- Create the `pkg` folder with the intention of putting the finished package from #1072 into the `pkg` folder.
- Audit the existing packages in `lib` and see which other packages we'd like to put in `pkg`.

## Other information and links

- https://github.com/golang-standards/project-layout

qdm12:
- `lib/trie/proof`'s [trie proof verification](https://github.com/ChainSafe/gossamer/blob/a1e127bc4f8504a5e5db6146944b3a51320a8fbe/lib/trie/proof/verify.go#L28)
- `lib/trie`'s `NewEmptyTrie()` function and trie `Put`, `Hash`, `Delete`, methods are used [here](https://github.com/w3f/polkadot-tests/blob/9f9b437b716e6b9bb0a564563c7400b2c040234d/adapters/gossamer/state_trie.go)
- `pkg/scale`  is used by https://github.com/w3f/polkadot-tests/blob/9f9b437b716e6b9bb0a564563c7400b2c040234d/adapters/gossamer/scale_codec.go#L25
- `lib/common`'s `Hash` is used by https://github.com/w3f/polkadot-tests/blob/9f9b437b716e6b9bb0a564563c7400b2c040234d/adapters/gossamer/state_trie.go#L90
- `lib/runtime` is used by https://github.com/w3f/polkadot-tests/tree/main/adapters/gossamer/host_api

PR #2292, opened 2022-02-11
feat(all): detect chain directory dynamically

qdm12:
💁 as the branch name suggests, that was me having some friday night fun and killing the 2022 new year PR #2139 to move everything to `internal/` for now.

## Changes

- Detect project root based on `chain` directory (and its subdirectories)
- Better detection of chain directory location for tests and production code
- Allow to move source files more easily
- Note this is just a takeaway from now defunct #2139 

## Issues

- Helping for #1527

### DIFF
--- a/lib/utils/utils.go
+++ b/lib/utils/utils.go
@@ -4,6 +4,7 @@
 package utils
 
 import (
+	"errors"
 	"fmt"
 	"os"
 	"os/user"
@@ -11,9 +12,11 @@ import (
 	"path/filepath"
 	"runtime"
 	"strings"
+	"testing"
 
 	"github.com/ChainSafe/chaindb"
 	"github.com/dgraph-io/badger/v2"
+	"github.com/stretchr/testify/require"
 )
 
 // DefaultDatabaseDir directory inside basepath where database contents are stored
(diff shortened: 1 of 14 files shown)


noot (approved):
nice fix!

lib/utils/utils.go
```diff
@@ -145,52 +148,133 @@ func KeystoreFilepaths(basepath string) ([]string, error) {
...
+// GetProjectRootPath finds the root of the project where `go.mod` is
+// and returns it as an absolute path.
+func GetProjectRootPath() (rootPath string, err error) {
+	_, fullpath, _, _ := runtime.Caller(0)
+	rootPath = path.Dir(fullpath)
+
+	const directoryToFind = "chain"
+	const subPathsToFind = "dev,gssmr,kusama,polkadot"
+
+	subPaths := strings.Split(subPathsToFind, ",")
```
EclesioMeloJunior, line 226:
```suggestion
	subPaths := [...]string{"dev", "gssmr", "kusama", "polkadot"}
```
Then `subPaths` will be a fixed size array

EclesioMeloJunior (approved):

lib/utils/utils.go
qdm12: 😢 I need constant immutable arrays so bad. I'm sure Rust has that.

I like your suggestion but since it doesn't prevent mutation of the array either, and especially that I got graced by the CI fully passing, I'll just merge it as is.

Issue #2171, opened 2022-01-11
Test trie store/load to cover its error cases


kishansagathiya:
## Issue summary

https://github.com/ChainSafe/gossamer/blob/790dfb5048fafbbabf7dfd450860eb6995be48e3/lib/trie/database.go#L27

I think this should be possible by mocking the database

## Other information and links

- extension of this comment https://github.com/ChainSafe/gossamer/pull/2122#discussion_r778856497

kishansagathiya:
closed in https://github.com/ChainSafe/gossamer/pull/2302

PR #2302, opened 2022-02-16
fix(lib/trie): Make sure writing and reading a trie to disk gives the same trie  and cover more store/load child trie related test cases

### LINKED ISSUE #1490: implement store/load of child trie from DB
## Describe the bug

- If `GetStorageChild` or `GetStorageFromChild` is called (in state.StorageState) It fails to load child trie from DB.
- the trie functions `Store` and `WriteDirty` should be updated to write the child tries to the database
- the trie function `Load` should be updated to load the child tries from the database and store them in memory

## To Reproduce

1. Clear the child trie from cache.
2. Try to load the child trie from DB.

## Log output

<details>
  <summary>Log Output</summary>
  
  ```Paste log output here
  paste log output...
  ```
</details> 
</br>

## Specification

- go version:
- gossamer version:
- gossamer commit tag:
- gossamer commit hash:
- operating system:
- additional links:

kishansagathiya:
While trying to write these tests I came across some problems, as following,  
- If you store a trie with child tries in db, when you load that trie from db, the loaded trie did not have child tries. 
- We were not able to load certain tries from db, despite storing them.

This was due to mis-match in the keys that gets used in loading  and storing a trie. key used during storing a trie was a `[]byte`, which could be of length equal to or less than 32, but while loading we were always using keys of length 32. So in case, say you are storing the trie with key `0x44102080304545b0b6579496e736964656368696c6414766f696c61`, we were loading with a key `0x00000000044102080304545b0b6579496e736964656368696c6414766f696c61`. And since these two keys are different, we were not able to load the trie.

## Changes
- Load now takes `[]byte` key instead of previous `common.Hash`
- Changed hashing related functions when appropriate
- Added a test to load and store child trie in lib/trie

## Tests

```

```

## Issues

- #2171
- closes #1490

## Primary Reviewer

- @qdm12


dot/state/storage_test.go
```diff
@@ -227,7 +228,12 @@ func TestGetStorageChildAndGetStorageFromChild(t *testing.T) {
 		"0",
 	))
 
-	testChildTrie := trie.NewEmptyTrie()
+	key := []byte{1, 2}
+	value := []byte{3, 4}
+	const dirty = true
+	const generation = 0
+	testChildTrie := trie.NewTrie(node.NewLeaf(key, value, dirty, generation))
```
qdm12, line 235: nit a bit simpler, since now most node fields are exported (for testing really) so `NewLeaf` is to be deprecated

```suggestion
	childTrieRoot := &node.Leaf{
		Key: []byte{1, 2},
		Value: []byte{3, 4}
		Dirty: true,
		Generation: 0,
	}
	testChildTrie := trie.NewTrie(childTrieRoot)
```

lib/trie/child_storage.go
```diff
@@ -16,17 +16,20 @@ var ChildStorageKeyPrefix = []byte(":child_storage:default:")
 var ErrChildTrieDoesNotExist = errors.New("child trie does not exist")
 
 // PutChild inserts a child trie into the main trie at key :child_storage:[keyToChild]
+// A child trie is added as a node (K, V) in main trie. K is the child storage key
+// associated to the child trie, and V is the Merkle value of its corresponding child trie.
+// To get merkle value of a node, we get value of the node. If the value is shorter than
+// 32 bits, we use it as merkle value. Otherwise, we use it's blake2b hash as merkle value.
```
qdm12, line 22: - Clarify difference between `node value` and `root hash`, using `value` for both is confusing
- It's 32 bytes, not 32 bits
- Add some details of what is blake2b-ed if the node value is not shorter than 32 bytes
- 

```suggestion
// A child trie is added as a node (K, V) in the main trie. K is the child storage key
// associated to the child trie, and V is the root hash of the child trie.
// If the child trie root node value is shorter than 32 bytes, this one is used as V.
// Otherwise, we use the blake2b hash digest of the child trie root node encoding as V.
```

lib/trie/database.go
```diff
@@ -145,24 +145,23 @@ func (t *Trie) loadProof(proofHashToNode map[string]Node, n Node) {
 
 // Load reconstructs the trie from the database from the given root hash.
 // It is used when restarting the node to load the current state trie.
-func (t *Trie) Load(db chaindb.Database, rootHash common.Hash) error {
-	if rootHash == EmptyHash {
+func (t *Trie) Load(db chaindb.Database, rootHashBytes []byte) error {
```
qdm12, line 148: How does this help?

I ran a tiny test

```go
func Test_ToBytes(t *testing.T) {
	h := Hash{1, 2, 3}
	t.Log(h)
	b := h.ToBytes()
	t.Log(b)
}
```

Printing

```
    /workspace/lib/common/hash_test.go:48: 0x0102030000000000000000000000000000000000000000000000000000000000
    /workspace/lib/common/hash_test.go:50: [1 2 3 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0]
```

So aren't these two the same?

noot: rootHash should always be hash I think (see above comment)?

dot/state/storage_test.go
kishansagathiya: trie.NewTrie is now inside a loop, so these variables are needed there.

lib/trie/child_storage.go
kishansagathiya: > I think since child tries are tries in themselves, the value (ie. their root hash) should always be a 32-byte hash, right?

True, will change the comment accordingly

lib/trie/database.go
kishansagathiya: > How does this help?

It helped me to keep both the keys (when storing and loading) same. TBH, I did not consider that root node should always be hashed here. I think the right solution should have been something like https://github.com/ChainSafe/gossamer/pull/2359 instead of changing this arg. I am going to change this back to common.Hash.

kishansagathiya:
blocking this for https://github.com/ChainSafe/gossamer/pull/2359 and https://github.com/ChainSafe/gossamer/pull/2360

lib/trie/database.go
kishansagathiya: addressed

lib/trie/child_storage.go
kishansagathiya: changed

lib/trie/database.go
```diff
@@ -377,8 +379,8 @@ func (t *Trie) writeDirty(db chaindb.Batch, n Node) error {
 			n.GetHash(), err)
 	}
 
-	if n == t.root {
-		// hash root node even if its encoding is under 32 bytes
+	// hash root node even if its encoding is under 32 bytes
+	if n == t.root && len(encoding) < 32 {
```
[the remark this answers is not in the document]
kishansagathiya: It checks if the the node is root and encoding less than 32, and in the if block it sets the hash. So, it matches.
This code has been moved now and is part of EncodeAndHash

dot/state/storage.go
```diff
@@ -167,7 +167,7 @@ func (s *StorageState) loadTrie(root *common.Hash) (*trie.Trie, error) {
 
 	tr, err := s.LoadFromDB(*root)
 	if err != nil {
-		return nil, errTrieDoesNotExist(*root)
+		return nil, fmt.Errorf("%s, %w", err, errTrieDoesNotExist(*root))
```
qdm12, line 170: A tiny bit out of scope, but let's just remove `errTrieDoesNotExist` please, it makes everything just more confusing. It's also only used in one other place, remove it there too :pray: 

```suggestion
		return nil, fmt.Errorf("trie does not exist at root %s: %w", *root, err)
```

dot/state/storage_test.go
```diff
@@ -195,6 +202,7 @@ func TestGetStorageChildAndGetStorageFromChild(t *testing.T) {
 
 	trieState, err := runtime.NewTrieState(genTrie)
 	require.NoError(t, err)
+	fmt.Printf("genTrie.RootNode().GetHash() %s\n", genTrie.RootNode().GetHash())
```
qdm12, line 205:
```suggestion
```

dot/state/storage_test.go
qdm12: They aren't used anywhere, I think you are mistaking this with another test? :thinking:

lib/trie/child_storage.go
```diff
@@ -16,17 +16,17 @@ var ChildStorageKeyPrefix = []byte(":child_storage:default:")
 var ErrChildTrieDoesNotExist = errors.New("child trie does not exist")
 
 // PutChild inserts a child trie into the main trie at key :child_storage:[keyToChild]
+// A child trie is added as a node (K, V) in main trie. K is the child storage key
```
qdm12, line 19: supernit

```suggestion
// A child trie is added as a node (K, V) in the main trie. K is the child storage key
```

lib/trie/child_storage.go
```diff
@@ -16,17 +16,17 @@ var ChildStorageKeyPrefix = []byte(":child_storage:default:")
...
+// associated to the child trie, and V is the root hash of the child trie.
 func (t *Trie) PutChild(keyToChild []byte, child *Trie) error {
-	childHash, err := child.Hash()
+	hash, err := child.Hash()
```
qdm12, line 22: nit keep it as `childHash` to keep deltas lower? Why the renaming :thinking:

lib/trie/child_storage.go
```diff
@@ -38,9 +38,7 @@ func (t *Trie) GetChild(keyToChild []byte) (*Trie, error) {
 		return nil, fmt.Errorf("%w at key 0x%x%x", ErrChildTrieDoesNotExist, ChildStorageKeyPrefix, keyToChild)
 	}
 
-	hash := [32]byte{}
-	copy(hash[:], childHash)
-	return t.childTries[common.Hash(hash)], nil
+	return t.childTries[common.BytesToHash(childHash)], nil
```
qdm12, line 41: This seems wrong? Where is `childHash` defined since it got renamed to `hash`?

lib/trie/database.go
```diff
@@ -209,16 +210,17 @@ func (t *Trie) load(db chaindb.Database, n Node) error {
...
-			return fmt.Errorf("failed to load child trie with root hash=0x%x: %w", value, err)
+			return fmt.Errorf("failed to load child trie with root hash=%s: %w", rootHash, err)
 		}
 
-		err = t.PutChild(value, childTrie)
+		hash, err := childTrie.Hash()
 		if err != nil {
-			return fmt.Errorf("failed to insert child trie with root hash=0x%x into main trie: %w",
-				childTrie.root.GetHash(), err)
+			return err
```
qdm12, line 221:
```suggestion
			return fmt.Errorf("cannot hash chilld trie at key %s: %w", key, err)
```

lib/trie/database.go
```diff
@@ -209,16 +210,17 @@ func (t *Trie) load(db chaindb.Database, n Node) error {
...
 		}
+		t.childTries[hash] = childTrie
```
qdm12, line 223: Dumb question, why not use `PutChild`?

lib/trie/database_test.go
```diff
@@ -66,6 +67,7 @@ func TestTrie_DatabaseStoreAndLoad(t *testing.T) {
 		err = res.Load(db, trie.MustHash())
 		require.NoError(t, err)
 		require.Equal(t, trie.MustHash(), res.MustHash())
+		require.Equal(t, trie.String(), res.String())
```
qdm12, line 70: Maybe then remove the hash assertion above? I guess diffs looks easier with this than comparing the root hashes so this addition makes sense.

lib/trie/database_test.go
```diff
@@ -387,3 +389,59 @@ func TestTrie_GetFromDB(t *testing.T) {
 		}
 	}
 }
+
+func TestStoreAndLoadWithChildTries(t *testing.T) {
+	testCase := []Test{
```
qdm12, line 394: nit Run these tests with `t.Parallel()`

qdm12, line 394:
```suggestion
func TestStoreAndLoadWithChildTries(t *testing.T) {
	testCase := []keyValue{
```

also please rename from `testCase` to `keyValues`

qdm12 (changes requested):

dot/state/storage_test.go
timwu20: I prefer using constructors whenever we can even if the fields are exported.

timwu20 (approved):
great work @kishansagathiya!

noot (approved):

lib/trie/child_storage.go
qdm12: Oh yeah dummy me, my bad!

qdm12 (approved):

Issue #2185, opened 2022-01-11
[Core] rewrite dot/core service unit tests part2


jimjbrettj:
## Issue summary

- Rewrite dot/core service tests to increase speed and code coverage
- This is a 2 part issue, so this one will cover the second have of the function

PR #2308, opened 2022-02-17
testing(dot/core): rewrite dot/core service tests (part 2)

jimjbrettj:
## Changes

- Finished writing unit tests for dot/core.Service.go
- Should wait to review till P1 is merged to reduce diff (at time of opening its awaiting 1 review)

## Tests

```
go test ./dot/core -v -coverprofile coverage.out
```

## Issues

- #2185 

## Primary Reviewer

- @timwu20


coverage.outc
```diff
@@ -0,0 +1 @@
+mode: set
```
noot, line 1: wat is this

dot/core/service_integration_test.go
```diff
@@ -37,6 +38,98 @@ import (
...
+func newTestDigest(t *testing.T) scale.VaryingDataTypeSlice {
+	vdts := types.NewDigest()
+	err := vdts.Add(
+		types.PreRuntimeDigest{
+			ConsensusEngineID: types.BabeEngineID,
+			Data:              common.MustHexToBytes("0x0201000000ef55a50f00000000"),
+		},
+		types.ConsensusDigest{
+			ConsensusEngineID: types.BabeEngineID,
+			Data:              common.MustHexToBytes("0x0118ca239392960473fe1bc65f94ee27d890a49c1b200c006ff5dcc525330ecc16770100000000000000b46f01874ce7abbb5220e8fd89bede0adad14c73039d91e28e881823433e723f0100000000000000d684d9176d6eb69887540c9a89fa6097adea82fc4b0ff26d1062b488f352e179010000000000000068195a71bdde49117a616424bdc60a1733e96acb1da5aeab5d268cf2a572e94101000000000000001a0575ef4ae24bdfd31f4cb5bd61239ae67c12d4e64ae51ac756044aa6ad8200010000000000000018168f2aad0081a25728961ee00627cfe35e39833c805016632bf7c14da5800901000000000000000000000000000000000000000000000000000000000000000000000000000000"), //nolint:lll
```
noot, line 60: where is this from? can you generate it instead of hardcoding it?

dot/core/service_integration_test.go
```diff
@@ -37,6 +38,98 @@ import (
...
+	require.NoError(t, err)
+	return append(append(append(h0, h1...), h2...), pub...)
+}
+
+func newTestDigest(t *testing.T) scale.VaryingDataTypeSlice {
+	vdts := types.NewDigest()
+	err := vdts.Add(
+		types.PreRuntimeDigest{
+			ConsensusEngineID: types.BabeEngineID,
+			Data:              common.MustHexToBytes("0x0201000000ef55a50f00000000"),
```
noot, line 56: same here

dot/core/service_integration_test.go
```diff
@@ -37,6 +38,98 @@ import (
...
+		},
+		types.ConsensusDigest{
+			ConsensusEngineID: types.BabeEngineID,
+			Data:              common.MustHexToBytes("0x0118ca239392960473fe1bc65f94ee27d890a49c1b200c006ff5dcc525330ecc16770100000000000000b46f01874ce7abbb5220e8fd89bede0adad14c73039d91e28e881823433e723f0100000000000000d684d9176d6eb69887540c9a89fa6097adea82fc4b0ff26d1062b488f352e179010000000000000068195a71bdde49117a616424bdc60a1733e96acb1da5aeab5d268cf2a572e94101000000000000001a0575ef4ae24bdfd31f4cb5bd61239ae67c12d4e64ae51ac756044aa6ad8200010000000000000018168f2aad0081a25728961ee00627cfe35e39833c805016632bf7c14da5800901000000000000000000000000000000000000000000000000000000000000000000000000000000"), //nolint:lll
+		},
+		types.SealDigest{
+			ConsensusEngineID: types.BabeEngineID,
+			Data:              common.MustHexToBytes("0x4625284883e564bc1e4063f5ea2b49846cdddaa3761d04f543b698c1c3ee935c40d25b869247c36c6b8a8cbbd7bb2768f560ab7c276df3c62df357a7e3b1ec8d"), //nolint:lll
```
noot, line 64: and here

dot/core/service_integration_test.go
```diff
@@ -37,6 +38,98 @@ import (
...
+	cfg.Storage = genState
+	cfg.LogLvl = 4
+	nodeStorage := runtime.NodeStorage{}
+	nodeStorage.BaseDB = runtime.NewInMemoryDB(t)
+	cfg.NodeStorage = nodeStorage
+
+	rt, err := wasmer.NewRuntimeFromGenesis(cfg)
+	require.NoError(t, err)
+
+	alicePub := common.MustHexToBytes("0xd43593c715fdd31c61141abd04a99fd6822c8558854ccde39a5684e7a56da27d")
```
noot, line 92: same here - is this from the `keystore.Sr25519Keyring`?

dot/core/service_integration_test.go
```diff
@@ -361,97 +454,70 @@ func TestHandleChainReorg_WithReorg_Transactions(t *testing.T) {
...
+		Validity:  &transaction.Validity{Priority: 1},
 	}
+	_ = ts.AddToPool(tx)
 
-	s := &Service{
-		transactionState: ts,
+	s := NewTestService(t, cfg)
+	s.transactionState = ts
+
+	provides := common.MustHexToBytes("0xd43593c715fdd31c61141abd04a99fd6822c8558854ccde39a5684e7a56da27d00000000")
```
noot, line 476: what is this? the previous tx hash?

dot/core/service_test.go
```diff
@@ -28,6 +36,68 @@ import (
...
+	rv := runtime.NewVersionData(
+		[]byte("polkadot"),
+		[]byte("parity-polkadot"),
+		0,
+		25,
+		0,
+		[]runtime.APIItem{testAPIItem},
+		5,
+	)
+	require.NoError(t, err)
```
noot, line 62: can you fetch this from the runtime instead of hardcoding it?

dot/core/service_test.go
```diff
@@ -28,6 +36,68 @@ import (
...
+
+	bob, err := ctypes.NewMultiAddressFromHexAccountID(
+		"0x90b5ab205c6974c9ea841be688864633dc9ca8a357843eeacf2314649965fe22")
+	require.NoError(t, err)
+
+	call, err := ctypes.NewCall(meta, "Balances.transfer", bob, ctypes.NewUCompactFromUInt(12345))
```
noot, line 68: make `Balances.transfer`  a constant

dot/core/service_test.go
```diff
@@ -28,6 +36,68 @@ import (
...
+		[]byte("parity-polkadot"),
+		0,
+		25,
+		0,
+		[]runtime.APIItem{testAPIItem},
+		5,
+	)
+	require.NoError(t, err)
+
+	bob, err := ctypes.NewMultiAddressFromHexAccountID(
```
noot, line 64: same here, is this from the keyring?

dot/core/service_test.go
```diff
@@ -28,6 +36,68 @@ import (
...
+		"0x90b5ab205c6974c9ea841be688864633dc9ca8a357843eeacf2314649965fe22")
+	require.NoError(t, err)
+
+	call, err := ctypes.NewCall(meta, "Balances.transfer", bob, ctypes.NewUCompactFromUInt(12345))
+	require.NoError(t, err)
+
+	// Create the extrinsic
+	extrinsic := ctypes.NewExtrinsic(call)
+	genHash, err := ctypes.NewHashFromHexString("0x35a28a7dbaf0ba07d1485b0f3da7757e3880509edc8c31d0850cb6dd6219361d")
```
noot, line 73: also where is this from

noot (commented):
can you update the hardcoded values to be dynamically created? it's a pain to update tests when stuff is hardcoded

dot/core/service.go
```diff
@@ -373,14 +375,14 @@ func (s *Service) handleChainReorg(prev, curr common.Hash) error {
 
 		for _, ext := range *body {
 			logger.Tracef("validating transaction on re-org chain for extrinsic %s", ext)
-			encExt, err := scale.Marshal(ext)
+			decExt := &ctypes.Extrinsic{}
+			decoder := cscale.NewDecoder(bytes.NewReader(ext))
+			_, err := decoder.DecodeUintCompact()
```
qdm12, line 380: Why the change? 🤔

dot/core/service.go
```diff
@@ -424,17 +425,16 @@ func (s *Service) maintainTransactionPool(block *types.Block) {
 			continue
 		}
 
-		txnValidity, err := rt.ValidateTransaction(tx.Extrinsic)
+		val, err := rt.ValidateTransaction(tx.Extrinsic)
```
qdm12, line 428: revert to `txnValidity` or make it at least `validity`

dot/core/service.go
```diff
@@ -424,17 +425,16 @@ func (s *Service) maintainTransactionPool(block *types.Block) {
...
 		if err != nil {
+			// failed to validate tx, remove it from the pool or queue
 			s.transactionState.RemoveExtrinsic(tx.Extrinsic)
 			continue
 		}
 
-		tx = transaction.NewValidTransaction(tx.Extrinsic, txnValidity)
-
+		tx = transaction.NewValidTransaction(tx.Extrinsic, val)
```
qdm12, line 435: nit revert, new lines are nice here and there

dot/core/service.go
```diff
@@ -424,17 +425,16 @@ func (s *Service) maintainTransactionPool(block *types.Block) {
...
 		// Err is only thrown if tx is already in pool, in which case it still gets removed
 		h, _ := s.transactionState.Push(tx)
-
```
qdm12, line 437: nit Why the new line removed 😢

dot/core/service_integration_test.go
```diff
@@ -37,6 +38,98 @@ import (
...
 //go:generate mockgen -destination=mock_telemetry_test.go -package $GOPACKAGE github.com/ChainSafe/gossamer/dot/telemetry Client
 
+func balanceKey(t *testing.T, pub []byte) []byte {
+	h0, err := common.Twox128Hash([]byte("System"))
+	require.NoError(t, err)
+	h1, err := common.Twox128Hash([]byte("Account"))
+	require.NoError(t, err)
+	h2, err := common.Blake2b128(pub)
+	require.NoError(t, err)
+	return append(append(append(h0, h1...), h2...), pub...)
```
qdm12, line 48: nit perhaps split this in multiple instruction lines?

dot/core/service_integration_test.go
```diff
@@ -37,6 +38,98 @@ import (
...
+}
+
+func newTestDigest(t *testing.T) scale.VaryingDataTypeSlice {
```
qdm12, line 51: Maybe pass in a data slice `data [][]byte`?

I'm a bit inclined to be against helping functions with actual data, it's easier to read/more explicit if the test passes the data to it.

dot/core/service_integration_test.go
```diff
@@ -37,6 +38,98 @@ import (
...
+}
+
+func generateTestValidTxns(t *testing.T) ([]byte, runtime.Instance) {
+	gen, err := genesis.NewGenesisFromJSONRaw("../../chain/gssmr/genesis.json")
+	require.NoError(t, err)
+
+	genTrie, err := genesis.NewTrieFromGenesis(gen)
+	require.NoError(t, err)
+
+	// set state to genesis state
```
qdm12, line 78: nit not needed

```suggestion
```

dot/core/service_integration_test.go
```diff
@@ -37,6 +38,98 @@ import (
...
+	genState, err := storage.NewTrieState(genTrie)
+	require.NoError(t, err)
+
+	cfg := &wasmer.Config{}
+	cfg.Storage = genState
+	cfg.LogLvl = 4
+	nodeStorage := runtime.NodeStorage{}
+	nodeStorage.BaseDB = runtime.NewInMemoryDB(t)
+	cfg.NodeStorage = nodeStorage
```
qdm12, line 87: nit also changed `4` to the corresponding log level, I guess it's `log.Error`?

```suggestion
	nodeStorage := runtime.NodeStorage{}
	nodeStorage.BaseDB = runtime.NewInMemoryDB(t)

	cfg := &wasmer.Config{
		Storage:		genState,
		LogLvl:		log.Error
		NodeStorage:	nodeStorage,
	}
```

dot/core/service_integration_test.go
```diff
@@ -37,6 +38,98 @@ import (
...
+	aliceBalanceKey := balanceKey(t, alicePub)
+
+	accInfo := types.AccountInfo{
+		Nonce: 0,
+		Data: struct {
+			Free       *scale.Uint128
+			Reserved   *scale.Uint128
+			MiscFrozen *scale.Uint128
+			FreeFrozen *scale.Uint128
+		}{
```
qdm12, line 102: nit Can we make a named struct for this?

dot/core/service_integration_test.go
```diff
@@ -37,6 +38,98 @@ import (
...
+
+	// Hash of encrypted centrifuge extrinsic
+	extHex := runtime.NewTestExtrinsic(t, rt, genesisHeader.Hash(), genesisHeader.Hash(),
+		0, "System.remark", []byte{0xab, 0xcd})
+
+	extBytes := common.MustHexToBytes(extHex)
+	extBytes = append([]byte{byte(types.TxnExternal)}, extBytes...)
+
+	runtime.InitializeRuntimeToTest(t, rt, genesisHeader.Hash())
+	return extBytes, rt
```
qdm12, line 130: Try to pass in more data to this helper function so the constant values are injected by the test and not in the helping test function. Although maybe that's just my opinion as well.

dot/core/service_integration_test.go
```diff
@@ -462,8 +528,9 @@ func TestMaintainTransactionPool_BlockWithExtrinsics(t *testing.T) {
 		}
 		res = append(res, tx)
 	}
-	require.Equal(t, 1, len(res))
-	require.Equal(t, res[0], txs[1])
+	// Extrinsic is removed. so empty res
+	require.Equal(t, 0, len(res))
+	require.Equal(t, res, []*transaction.ValidTransaction{})
```
qdm12, line 533: no need to compare length due to the 2nd assertion + expected argument is mean to go as 2nd argument, not 3rd.

```suggestion
	require.Equal(t, []*transaction.ValidTransaction{}, res)
```

dot/core/service_test.go
```diff
@@ -586,3 +656,699 @@ func Test_Service_handleBlocksAsync(t *testing.T) {
...
+		mockStorageState.EXPECT().GenerateTrieProof(common.Hash{3}, [][]byte{{1}}).
+			Return([][]byte{}, errDummyErr)
+		s := &Service{
+			blockState:   mockBlockState,
+			storageState: mockStorageState,
+		}
+		execTest(t, s, common.Hash{}, [][]byte{{1}}, common.Hash{}, nil, errDummyErr)
+	})
+
+	t.Run("happy path", func(t *testing.T) {
```
qdm12, line 1339: nit in the future perhaps call it `success`? I mean `happy path` is cute but not too serious either 😄

dot/core/service_test.go
```diff
@@ -586,3 +656,699 @@ func Test_Service_handleBlocksAsync(t *testing.T) {
...
+
+func TestService_GetReadProofAt(t *testing.T) {
+	t.Parallel()
+	execTest := func(t *testing.T, s *Service, block common.Hash, keys [][]byte,
+		expHash common.Hash, expProofForKeys [][]byte, expErr error) {
+		resHash, resProofForKeys, err := s.GetReadProofAt(block, keys)
+		assert.ErrorIs(t, err, expErr)
+		if expErr != nil {
+			assert.EqualError(t, err, expErr.Error())
+		}
```
qdm12, line 1306: nit in the future, the `expErr` might be wrapped, so maybe have an argument `expErr error` and another `expErrMessage string` for the `assert.EqualError` assertion?

dot/core/service_test.go
```diff
@@ -586,3 +656,699 @@ func Test_Service_handleBlocksAsync(t *testing.T) {
...
+	t.Run("get block state root error", func(t *testing.T) {
+		t.Parallel()
+		ctrl := gomock.NewController(t)
+		mockBlockState := NewMockBlockState(ctrl)
+		mockBlockState.EXPECT().BestBlockHash().Return(common.Hash{2})
+		mockBlockState.EXPECT().GetBlockStateRoot(common.Hash{2}).Return(common.Hash{}, errDummyErr)
+		s := &Service{
+			blockState: mockBlockState,
+		}
+		execTest(t, s, common.Hash{}, nil, common.Hash{}, nil, errDummyErr)
```
qdm12, line 1320: nit just a suggestion, I started doing this not too long ago and I think it's nice readability-wise; variable assign each argument so it's clear i.e. without an IDE what's what.

```suggestion

		blockHash := common.Hash{}
		keys := ([][]byte)(nil)
		expectedHash := common.Hash{}
		expectedProofForKeys := ([][]byte)(nil)
		expectedErr := errDummyErr
		execTest(t, s, blockHash, keys, expectedHash, expectedProofForKeys, expectedErr)
```

Also `s` for service is sus, call it `service`.

dot/core/service.go
jimjbrettj: I believe we just did not need to marshal here. Then we were using our own type that was a wrapper for cscale and calling DecodeVersion(). Since this function is the only place its used, i moved the logic here to make it more clear (I spent a lot of time trying to figure out what was going on) and deleted the wrapper type.

coverage.outc
jimjbrettj: oops its a go tool coverage report I forgot to delete

dot/core/service_integration_test.go
jimjbrettj: hmmm I got this digest from the types/digest_test file and it isn't marked there where they are from. Do you have a suggestion of how I could generate this?

dot/core/service_test.go
jimjbrettj: Since this is just for mocks do you think this is necessary (i'm not even sure this function is needed for the unit tests)?

dot/core/service_test.go
jimjbrettj: yeah it is, ill update it

dot/core/service_test.go
jimjbrettj: hmm I got it from here https://github.com/ChainSafe/gossamer/blob/4617175169f6e475764a553de653f34e6977ba9f/lib/babe/build_integration_test.go#L283 but not sure where that was from :/

dot/core/service_integration_test.go
noot: I think you can just generate any random consensus digest and put it in here, it doesn't even look like this is a real digest lol. if the data doesn't matter you could also just put some random bytes

jimjbrettj: ohhh I think they are hardcoded BabePreDigests. They do need to be valid, its updated to generate a babePreDigest now rather than just have a hex string

jimjbrettj: let me know if this should be fixed further!

dot/core/service_integration_test.go
jimjbrettj: I recently updated it so that the data isn't hardcoded, and I think doing this would mean repeating that code for every call to this function which I think would add to much duplicated code

EclesioMeloJunior (changes requested):

dot/core/service_integration_test.go
timwu20: so you are expecting this `Provides` tags to be added via the runtime? I would put a comment about it.  It would be nice to have a link to the runtime code that appends these tags, and what these tags are.

dot/core/service_test.go
```diff
@@ -28,6 +36,72 @@ import (
...
 
+func generateExtrinsic(t *testing.T) (ext types.Extrinsic, externExt types.Extrinsic, body *types.Body) {
+	rawMeta := common.MustHexToBytes(testdata.NewTestMetadata())
+	var decoded []byte
+	err := scale.Unmarshal(rawMeta, &decoded)
+	require.NoError(t, err)
+
+	meta := &ctypes.Metadata{}
+	err = ctypes.DecodeFromBytes(decoded, meta)
+	require.NoError(t, err)
```
timwu20, line 47:
```suggestion
	rawMeta := common.MustHexToBytes(testdata.NewTestMetadata())
	var decoded []byte
	err := scale.Unmarshal(rawMeta, &decoded)
	require.NoError(t, err)

	meta := &ctypes.Metadata{}
	err = ctypes.DecodeFromBytes(decoded, meta)
	require.NoError(t, err)
```
this could be put in a helper.

dot/core/service_test.go
```diff
@@ -577,12 +651,708 @@ func Test_Service_handleBlocksAsync(t *testing.T) {
...
+		}
+	}
+
+	testPrevHash := common.MustHexToHash("0x01")
+	testCurrentHash := common.MustHexToHash("0x02")
+	testAncestorHash := common.MustHexToHash("0x03")
+	testSubChain := []common.Hash{testPrevHash, testCurrentHash, testAncestorHash}
+	ext, externExt, body := generateExtrinsic(t)
+	testValidity := &transaction.Validity{Propagate: true}
+	vtx := transaction.NewValidTransaction(ext, testValidity)
```
timwu20, line 680: we do we need all of this if we're just mocking it out anyways? Looks like this is the only reference to `generateExtrinsic`

dot/core/service.go
```diff
@@ -372,14 +374,14 @@ func (s *Service) handleChainReorg(prev, curr common.Hash) error {
 
 		for _, ext := range *body {
 			logger.Tracef("validating transaction on re-org chain for extrinsic %s", ext)
-			encExt, err := scale.Marshal(ext)
+			decExt := &ctypes.Extrinsic{}
+			decoder := cscale.NewDecoder(bytes.NewReader(ext))
+			_, err := decoder.DecodeUintCompact()
```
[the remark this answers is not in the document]
jimjbrettj: yeah I believe this was the cause of the double encoding that was happening before. 

As to the aside: completely agree. It caused me some confusion when working on this having to deal with both our extrinsics and centrifuge's. I think having scale being able to take care of it is just one less thing the developer can get mixed up on

dot/core/service.go
```diff
@@ -372,14 +374,14 @@ func (s *Service) handleChainReorg(prev, curr common.Hash) error {
...
 			if err != nil {
 				return err
 			}
 
-			// decode extrinsic and make sure it's not an inherent.
-			decExt := &types.ExtrinsicData{}
-			if err = decExt.DecodeVersion(encExt); err != nil {
+			if err = decoder.Decode(&decExt.Version); err != nil {
```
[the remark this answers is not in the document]
jimjbrettj: So the thing is we don't need the version itself in our code. However we call the IsSigned method https://github.com/centrifuge/go-substrate-rpc-client/blob/db7f2af2992f9713bad11b88ce06d23759ee929a/types/extrinsic.go#L113 does need the version (and no other attributes) to check version. So the only reason we decode this at all, as far as I can tell, is just to get the version to see if the extrinsic is signed.

Just thought that was kind of interesting, but I agree with you decoding the whole thing is cleaner (and reduces code actually no more need to decode in two steps).

dot/core/service.go
```diff
@@ -388,14 +390,13 @@ func (s *Service) handleChainReorg(prev, curr common.Hash) error {
 				continue
 			}
 
-			externalExt := types.Extrinsic(append([]byte{byte(types.TxnExternal)}, encExt...))
+			externalExt := types.Extrinsic(append([]byte{byte(types.TxnExternal)}, ext...))
```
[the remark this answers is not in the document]
jimjbrettj: That is what it seems to me. Maybe @noot could give some more insight to this?

dot/core/service_integration_test.go
```diff
@@ -11,31 +11,158 @@ import (
...
+	bKey = append(bKey, h2...)
+	bKey = append(bKey, pub...)
+	return bKey
+}
+
+func newTestDigest(t *testing.T) scale.VaryingDataTypeSlice {
+	testBabeDigest := types.NewBabeDigest()
+	err := testBabeDigest.Set(types.BabePrimaryPreDigest{
+		VRFOutput: [sr25519.VRFOutputLength]byte{
+			0, 91, 50, 25, 214, 94, 119, 36, 71, 216, 33,
```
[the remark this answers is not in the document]
jimjbrettj: I got it from the babe digest test here https://github.com/ChainSafe/gossamer/blob/567c1ab1cf8cfd328a4eb4dfad4e48e50d91b2ce/dot/types/babe_digest_test.go#L15 but I am not sure where that was generated

dot/core/service_integration_test.go
jimjbrettj: I believe so? Haven't been able to find much on it. In substrate the provides is a vector of transaction tags, which is just a vector of bytes. Haven't been able to find that the possible values. My understanding is that the provides tags are added by the runtime to deal with transaction dependencies. So maybe it is the previous txn hash @noot? Or at least A previous transaction of which the current txn depends?

dot/core/service.go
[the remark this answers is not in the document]
noot: yeah it's basically just what the runtime wants... maybe we could update our types to handle this better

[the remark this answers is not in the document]
jimjbrettj: okay, I think I will open an issue for this rather than make type changes in this PR if that's alright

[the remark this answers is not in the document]
qdm12: nit and maybe a bit out of scope, but that would be clearer :wink: 

```suggestion
			externalExt := make(types.Extrinsic, 1, 1 + len(ext))
			externalExt[0] = byte(types.TxnExternal)
			externalExt = append(externalExt, ext...)
```

dot/core/service_integration_test.go
```diff
@@ -11,31 +11,147 @@ import (
...
 	"github.com/ChainSafe/gossamer/pkg/scale"
 	"github.com/golang/mock/gomock"
+	"github.com/libp2p/go-libp2p-core/peer"
 	"github.com/stretchr/testify/mock"
 	"github.com/stretchr/testify/require"
 )
 
 //go:generate mockgen -destination=mock_telemetry_test.go -package $GOPACKAGE github.com/ChainSafe/gossamer/dot/telemetry Client
 
+//nolint
```
qdm12, line 44: Precise what linter this should be against. In example `//nolint:lll` for the `lll` linter.

dot/core/service_integration_test.go
```diff
@@ -11,31 +11,147 @@ import (
...
+type testAccountData struct {
+	Free       *scale.Uint128
+	Reserved   *scale.Uint128
+	MiscFrozen *scale.Uint128
+	FreeFrozen *scale.Uint128
+}
+
+func balanceKey(t *testing.T, pub []byte) []byte {
+	var bKey []byte
```
qdm12, line 53: nit

```suggestion
func balanceKey(t *testing.T, pub []byte) (bKey []byte) {
```

dot/core/service_integration_test.go
```diff
@@ -11,31 +11,147 @@ import (
...
+	h0, err := common.Twox128Hash([]byte("System"))
```
qdm12, line 54: nit

```suggestion
	t.Helper()

	h0, err := common.Twox128Hash([]byte("System"))
```

dot/core/service_integration_test.go
```diff
@@ -11,31 +11,147 @@ import (
...
+			ConsensusEngineID: types.BabeEngineID,
+			Data:              data,
+		},
+	)
+	require.NoError(t, err)
+	return vdts
+}
+
+func generateTestValidRemarkTxns(t *testing.T, pubKey []byte, accInfo types.AccountInfo) ([]byte, runtime.Instance) {
+	projectRootPath := utils.GetProjectRootPathTest(t) + "/chain/gssmr/genesis.json"
```
qdm12, line 107: nit so it works on all weird OSes/Filesystems

```suggestion
	projectRootPath := filepath.Join(utils.GetProjectRootPathTest(t), "chain/gssmr/genesis.json")
```

dot/core/service_integration_test.go
```diff
@@ -11,31 +11,147 @@ import (
...
+	require.NoError(t, err)
+
+	genTrie, err := genesis.NewTrieFromGenesis(gen)
+	require.NoError(t, err)
+
+	genState, err := storage.NewTrieState(genTrie)
+	require.NoError(t, err)
+
+	nodeStorage := runtime.NodeStorage{}
+	nodeStorage.BaseDB = runtime.NewInMemoryDB(t)
```
qdm12, line 118: nit

```suggestion
	nodeStorage := runtime.NodeStorage{
		BaseDB:	runtime.NewInMemoryDB(t),
	}
```

dot/core/service_integration_test.go
```diff
@@ -11,31 +11,147 @@ import (
...
+	rt.(*wasmer.Instance).GetContext().Storage.Set(common.UpgradedToDualRefKey, []byte{1})
+
+	genesisHeader := &types.Header{
+		Number:    big.NewInt(0),
+		StateRoot: genTrie.MustHash(),
+	}
+
+	// Hash of encrypted centrifuge extrinsic
+	extHex := runtime.NewTestExtrinsic(t, rt, genesisHeader.Hash(), genesisHeader.Hash(),
+		0, "System.remark", []byte{0xab, 0xcd})
```
qdm12, line 146: nit please name that last byte slice argument to a variable with a meaningful name. As a reviewer, I have no clue what this is :smile:

dot/core/service_integration_test.go
```diff
@@ -11,31 +11,147 @@ import (
...
+
+	extBytes := common.MustHexToBytes(extHex)
+	extBytes = append([]byte{byte(types.TxnExternal)}, extBytes...)
```
qdm12, line 149: nit a tiny bit clearer imo

```suggestion
	const txnType = byte(types.TxnExternal)
	extBytes = append([]byte{txnType}, extBytes...)
```

dot/core/service_integration_test.go
```diff
@@ -197,6 +318,11 @@ func TestService_HasKey_UnknownType(t *testing.T) {
 	cfg := &Config{
 		Keystore: ks,
 	}
+
+	ctrl := gomock.NewController(t)
+	digestHandler := NewMockDigestHandler(ctrl)
+	cfg.DigestHandler = digestHandler
```
qdm12, line 324: If we don't configure it, I think leaving to nil will work too so we could remove those new lines

dot/core/service_integration_test.go
```diff
@@ -461,8 +585,8 @@ func TestMaintainTransactionPool_BlockWithExtrinsics(t *testing.T) {
 		}
 		res = append(res, tx)
 	}
-	require.Equal(t, 1, len(res))
-	require.Equal(t, res[0], txs[1])
+	// Extrinsic is removed. so empty res
+	require.Equal(t, []*transaction.ValidTransaction{}, res)
```
qdm12, line 589:
```suggestion
	require.Empty(t, res)
```

dot/core/service_integration_test.go
```diff
@@ -479,7 +603,21 @@ func TestService_GetRuntimeVersion(t *testing.T) {
...
+	digestHandler.EXPECT().HandleDigests(gomock.AssignableToTypeOf(new(types.Header)))
+	cfg.DigestHandler = digestHandler
+
+	net := NewMockNetwork(ctrl)
+	net.EXPECT().GossipMessage(gomock.AssignableToTypeOf(new(network.TransactionMessage))).AnyTimes()
+	net.EXPECT().IsSynced().Return(true).AnyTimes()
+	net.EXPECT().ReportPeer(
+		gomock.AssignableToTypeOf(peerset.ReputationChange{}),
+		gomock.AssignableToTypeOf(peer.ID("")),
+	).AnyTimes()
```
qdm12, line 618: have these more restrictive if it's not a helper test function

dot/core/service_integration_test.go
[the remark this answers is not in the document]
jimjbrettj: great call!

dot/core/helpers_test.go
```diff
@@ -35,9 +31,7 @@ func NewTestService(t *testing.T, cfg *Config) *Service {
 	}
 
 	if cfg.DigestHandler == nil {
-		digestHandler := NewMockDigestHandler(ctrl)
-		digestHandler.EXPECT().HandleDigests(gomock.AssignableToTypeOf(new(types.Header)))
-		cfg.DigestHandler = digestHandler
+		cfg.DigestHandler = NewMockDigestHandler(ctrl)
```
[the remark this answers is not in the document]
jimjbrettj: yeah we do, NewService() does a nil check on the digestHandler

dot/core/helpers_test.go
```diff
@@ -126,14 +120,7 @@ func NewTestService(t *testing.T, cfg *Config) *Service {
...
 	if cfg.Network == nil {
-		net := NewMockNetwork(ctrl)
-		net.EXPECT().GossipMessage(gomock.AssignableToTypeOf(new(network.TransactionMessage)))
-		net.EXPECT().IsSynced().Return(true)
-		net.EXPECT().ReportPeer(
-			gomock.AssignableToTypeOf(peerset.ReputationChange{}),
-			gomock.AssignableToTypeOf(peer.ID("")),
-		)
-		cfg.Network = net
+		cfg.Network = NewMockNetwork(ctrl)
```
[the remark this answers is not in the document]
jimjbrettj: same as above

dot/core/messages_integration_test.go
```diff
@@ -77,6 +80,10 @@ func TestService_HandleBlockProduced(t *testing.T) {
 		Keystore: keystore.NewGlobalKeystore(),
 	}
 
+	digestHandler := NewMockDigestHandler(ctrl)
+	digestHandler.EXPECT().HandleDigests(gomock.AssignableToTypeOf(new(types.Header)))
```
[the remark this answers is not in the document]
jimjbrettj: yeah I think this would be to much since it has to be passed as part of the config in generating the service but the parent hash of the headeris retrieved through the service.

dot/core/service_integration_test.go
jimjbrettj: How do I know what linter is used? I was getting linting error for testAccountData being unused, which is not the case

dot/core/service_integration_test.go
jimjbrettj: good call!

dot/core/service_test.go
timwu20: Thanks for the explanation!  Maybe add a comment saying the transaction is checked.

dot/core/service_integration_test.go
```diff
@@ -673,8 +793,9 @@ func TestTryQueryStore_WhenThereIsDataToRetrieve(t *testing.T) {
 	storageStateTrie.Set(testKey, testValue)
 	require.NoError(t, err)
 
+	digest := newTestDigest(t, 420)
```
qdm12, line 796: what's `420`?

Maybe have it as an inlined named constant above, like:

```suggestion
	const myNameThatMakesSense = 420
	digest := newTestDigest(t, myNameThatMakesSense)
```

dot/core/service_integration_test.go
```diff
@@ -873,78 +997,3 @@ func TestDecodeSessionKeys_WhenGetRuntimeReturnError(t *testing.T) {
 	require.Error(t, err, "problems")
 	require.Nil(t, b)
 }
-
-func TestGetReadProofAt(t *testing.T) {
-	keysToProof := [][]byte{[]byte("first_key"), []byte("another_key")}
-	mockedProofs := [][]byte{[]byte("proof01"), []byte("proof02")}
-
```
qdm12, line 880: Why are these tests gone? Are they moved somewhere else?

dot/core/service_test.go
```diff
@@ -28,6 +36,78 @@ import (
 
 var errTestDummyError = errors.New("test dummy error")
 
+func generateTestCentrifugeMetadata(t *testing.T) *ctypes.Metadata {
+	t.Helper()
+	rawMeta := common.MustHexToBytes(testdata.NewTestMetadata())
```
qdm12, line 41: nit let's avoid panicing in tests

```suggestion

	metadataHex := testdata.NewTestMetadata()
	rawMeta, err := common.HexToBytes(metadataHex)
	require.NoError(t, err)

```

dot/core/service_test.go
```diff
@@ -28,6 +36,78 @@ import (
...
+	err := scale.Unmarshal(rawMeta, &decoded)
+	require.NoError(t, err)
+
+	meta := &ctypes.Metadata{}
+	err = ctypes.DecodeFromBytes(decoded, meta)
+	require.NoError(t, err)
+	return meta
+}
+
+func generateExtrinsic(t *testing.T) (ext types.Extrinsic, externExt types.Extrinsic, body *types.Body) {
```
qdm12, line 52: nit

```suggestion
func generateExtrinsic(t *testing.T) (ext, externExt types.Extrinsic, body *types.Body) {
```

Also imo `extrinsic` > `ext` and `externalExtrinsic` > `externExt` :wink:

dot/core/service_test.go
```diff
@@ -28,6 +36,78 @@ import (
...
+		Ver:  99,
+	}
+	rv := runtime.NewVersionData(
+		[]byte("polkadot"),
+		[]byte("parity-polkadot"),
+		0,
+		25,
+		0,
+		[]runtime.APIItem{testAPIItem},
+		5,
```
qdm12, line 67: nit it would be nice to have these as named inlined constants above `rv := runtime.NewVersionData(` so a reader doesn't have to lookup the method signature and match arguments

dot/core/service_test.go
```diff
@@ -28,6 +36,78 @@ import (
...
+
+	keyring, err := keystore.NewSr25519Keyring()
+	require.NoError(t, err)
+	bobPub := keyring.Bob().Public().Hex()
+
+	bob, err := ctypes.NewMultiAddressFromHexAccountID(bobPub)
+	require.NoError(t, err)
+
+	const balanceTransfer = "Balances.transfer"
+	call, err := ctypes.NewCall(meta, balanceTransfer, bob, ctypes.NewUCompactFromUInt(12345))
```
qdm12, line 78: That's what I'm talking about :muscle:

dot/core/service_test.go
```diff
@@ -577,12 +657,708 @@ func Test_Service_handleBlocksAsync(t *testing.T) {
...
+			Return(testValidity, nil)
+		mockTxnStateOk := NewMockTransactionState(ctrl)
+		mockTxnStateOk.EXPECT().AddToPool(vtx).Return(common.Hash{})
+
+		service := &Service{
+			blockState:       mockBlockState,
+			transactionState: mockTxnStateOk,
+		}
+		execTest(t, service, testPrevHash, testCurrentHash, nil)
+	})
```
qdm12, line 810: nit this is fine, although it might a bit cleaner with mock configuration functions as test case fields ([example](https://gist.github.com/qdm12/99358e2529a5887ae13a936a7f1c8ae5#file-potato_test-go-L234))

dot/core/service_test.go
```diff
@@ -577,12 +657,708 @@ func Test_Service_handleBlocksAsync(t *testing.T) {
...
+		expErrMsg string
+	}{
+		{
+			name: "ok case",
+			service: &Service{
+				keys: &keyStore,
+			},
+			args: args{
+				kp:           aliceKeypair,
+				keystoreType: (string)(keystore.BabeName),
```
qdm12, line 839: nit You could also do the following which is more common for some reason

```suggestion
				keystoreType: string(keystore.BabeName),
```

dot/core/service_test.go
```diff
@@ -577,12 +657,708 @@ func Test_Service_handleBlocksAsync(t *testing.T) {
...
+			},
+		},
+		{
+			name: "err case",
+			service: &Service{
+				keys: &keyStore,
+			},
+			args: args{
+				pubKeyStr:    aliceKeypair.Public().Hex(),
+				keystoreType: "jimbo",
```
qdm12, line 905: Leaving the mark of the tester uh? :smile_cat:

qdm12 (approved):
A bunch of nits, but looks good otherwise :100:

dot/core/helpers_test.go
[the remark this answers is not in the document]
qdm12: Then just make it an empty concrete implementation:

```suggestion
		cfg.DigestHandler = &digest.Handler{}
```

dot/core/helpers_test.go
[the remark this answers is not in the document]
qdm12:
```suggestion
		cfg.Network = &network.Service{}
```

dot/core/service_integration_test.go
jimjbrettj: Will try this now, just getting to this round of feedback

jimjbrettj: It seems that this test does require the Provides param to pass. @noot correct me if im wrong but it seems to be added by the runtime. When i remove it a check fails because its getting set somewhere, but I haven't been able to find where its set.

Unless told otherwise, thinking I will leave as is and add a comment

dot/core/service_integration_test.go
```diff
@@ -360,97 +473,95 @@ func TestHandleChainReorg_WithReorg_Transactions(t *testing.T) {
...
+			Free:       scale.MustNewUint128(big.NewInt(1152921504606846976)),
+			Reserved:   scale.MustNewUint128(big.NewInt(0)),
+			MiscFrozen: scale.MustNewUint128(big.NewInt(0)),
+			FreeFrozen: scale.MustNewUint128(big.NewInt(0)),
 		},
 	}
+	keyring, err := keystore.NewSr25519Keyring()
+	require.NoError(t, err)
+	alicePub := common.MustHexToBytes(keyring.Alice().Public().Hex())
+	encExt, _ := generateTestValidRemarkTxns(t, alicePub, accInfo)
```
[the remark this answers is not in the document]
jimjbrettj: good call!

dot/core/service_integration_test.go
jimjbrettj: updated!

dot/core/service_integration_test.go
jimjbrettj: These were using mocks and struck me as unit tests. Yup the respective tests exists as part of the unit tests

dot/core/service_test.go
jimjbrettj: updated, might want to double check I did what you had in mind @qdm12

dot/core/service_test.go
```diff
@@ -901,7 +912,7 @@ func TestServiceHasKey(t *testing.T) {
 			},
 			args: args{
 				pubKeyStr:    aliceKeypair.Public().Hex(),
-				keystoreType: "jimbo",
+				keystoreType: "test",
```
qdm12, line 915: You could had left `jimbo` :smile_cat:

dot/types/account.go
```diff
@@ -21,3 +21,11 @@ type AccountInfo struct {
 		FreeFrozen *scale.Uint128
 	}
 }
+
+// AccountData represents account data of an account. Currently is used in core tests
+type AccountData struct {
```
qdm12, line 26: Can you use this type instead of the anonymous struct in the Account struct in the `types` package as well?
And then update the comment as well for this struct.

dot/core/service_test.go
qdm12: Yes that works! Although supernitting here, inlined constants just above their usage are better readability-wise since it's more 'local'.

qdm12 (approved):

dot/core/service_integration_test.go
qdm12: Ah ok this makes sense, thanks!

dot/core/service_test.go
jimjbrettj: i got scared haha

dot/types/account.go
jimjbrettj: yup good call!

dot/core/service_integration_test.go
noot: weird, ok feel free to leave it and just document it then

noot (approved):

lib/runtime/wasmer/instance.go
```diff
@@ -187,6 +187,11 @@ func (in *Instance) GetCodeHash() common.Hash {
 	return in.codeHash
 }
 
+// GetContext returns the context of the instance
+func (in *Instance) GetContext() *runtime.Context {
```
timwu20, line 191: Seems odd that we need to modify the context only in tests

timwu20 (approved):

lib/runtime/wasmer/instance.go
jimjbrettj: Agreed, I needed it to create valid extristics based on the current/correct runtime context. I think its modified elsewhere in the package, it just is not exported

Issue #2229, opened 2022-01-24
Instrument the trie with Prometheus metrics


qdm12:

qdm12:
There is a gauge metric to keep track of the number of tries in memory in https://github.com/ChainSafe/gossamer/pull/2272 which should be merged soon (restarting CI jobs to fix it), but 'inside-trie' are still work in progress in #2310 where implementation is done but tests are being worked on.

PR #2310, opened 2022-02-18
feat(lib/trie): add descendants count to branches

qdm12:
## Changes

- [x] Count nodes added/removed for branches in the trie
- [x] Adapt unit tests
- [x] Add test cases to have full coverage

💁 descendants are stored in **every node** since we have a delete operation that cuts of an entire branch, so it's more performant to have it get the count of descendants to remove from the branch being removed, rather than iterating over this branch and all its children.

Next PR will be to publish the count of descendants of the root node to the Prometheus server

## Tests

```
```

## Issues

- #2229 

## Primary Reviewer

- @timwu20

### DIFF
--- a/internal/trie/node/stats.go
+++ b/internal/trie/node/stats.go
@@ -1,0 +1,30 @@
+package node
+
+// Stats contains statistical information for a node.
+type Stats struct {
+	// Descendants is the number of descendant nodes for
+	// this particular node.
+	Descendants uint32
+}
+
+// NewStats creates a new Stats structure given the arguments.
+func NewStats(descendants uint32) Stats {
+	return Stats{
+		Descendants: descendants,
+	}
+}
+
+// GetDescendants returns the number of descendants in the branch.
+func (b *Branch) GetDescendants() (descendants uint32) {
+	return b.Stats.Descendants
+}
+
+// AddDescendants adds descendant nodes count to the node stats.
+func (b *Branch) AddDescendants(n uint32) {
+	b.Stats.Descendants += n
+}
+
+// SubDescendants subtracts descendant nodes count from the node stats.
+func (b *Branch) SubDescendants(n uint32) {
+	b.Stats.Descendants -= n
+}
--- a/internal/trie/node/stats.go
+++ b/internal/trie/node/stats.go
@@ -1,3 +1,6 @@
+// Copyright 2022 ChainSafe Systems (ON)
+// SPDX-License-Identifier: LGPL-3.0-only
+
 package node
 
 // GetDescendants returns the number of descendants in the branch.
--- a/internal/trie/node/stats.go
+++ b/internal/trie/node/stats.go
@@ -1,30 +1,16 @@
 package node
 
-// Stats contains statistical information for a node.
-type Stats struct {
-	// Descendants is the number of descendant nodes for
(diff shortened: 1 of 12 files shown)


internal/trie/node/stats.go
```diff
@@ -0,0 +1,30 @@
...
+
+// AddDescendants adds descendant nodes count to the node stats.
+func (b *Branch) AddDescendants(n uint32) {
+	b.Stats.Descendants += n
+}
+
+// SubDescendants subtracts descendant nodes count from the node stats.
+func (b *Branch) SubDescendants(n uint32) {
+	b.Stats.Descendants -= n
+}
```
qdm12, line 30: Note methods are used here since I would like to pass in more arguments in future PRs such as descendants bytes sizes.

internal/trie/node/stats.go
```diff
@@ -0,0 +1,30 @@
+package node
+
+// Stats contains statistical information for a node.
+type Stats struct {
```
timwu20, line 4: Do we really need this `Stats` type? You are adding receiver methods to `Branch` to make this work already, so why not just add a `Descendants uint32` attribute to `Branch`?

timwu20: nit: I would've introduced `Stats` when you have more than just `Descendants`.

timwu20 (approved):

internal/trie/node/stats.go
qdm12: Agreed, I removed it, only took me a minute 😉

lib/trie/trie.go
```diff
@@ -390,16 +395,19 @@ func (t *Trie) insertInLeaf(parentLeaf *node.Leaf, key,
 		Generation: t.generation,
 		Dirty:      true,
 	}
+	newBranchParent.AddDescendants(1)
+	nodesCreated++
 
-	return newBranchParent
+	return newBranchParent, nodesCreated
```
noot, line 401: is `nodesCreated` always 0 or 1 or am i missing something here?

noot (commented):
seems ok but I think I'm missing the reason why nodesCreated/nodesDeleted needs to be returned from a lot of the funcs? seems to be 0/1 only in some cases?

lib/trie/trie.go
qdm12: It can be 2 if the else block on line 382 is run

qdm12:
> seems ok but I think I'm missing the reason why nodesCreated/nodesDeleted needs to be returned from a lot of the funcs? seems to be 0/1 only in some cases?

Agreed, I'm doing a refactoring to deduce those numbers from the branch statistics (especially since there might be more fields in the future).

However, although it's working it really does run some weird/sketchy code which should be removed by #2352 can you please review this :smiley_cat: Thanks!!

qdm12:
Do not merge, this is not valid (found with a fuzz test), I'm working on fixing it.

qdm12:
@noot I reverted back to return recursively nodesRemoved/nodesCreated.

The trie code modifies the passed branches in-place, so to deduce descendants without returning values, it is required to either have:

- extremely confusing and ugly code; or
- deep copy branch nodes (not recursively) before passing them, but that hurts my feelings performance-wise

PR #2311, opened 2022-02-18
docs(codestyle): mocks configuration in subtests

qdm12:
## Changes

- Describe how mocks should be configured in subtests
- Describe exception where for loop / test cases should be broken in individual subtests

## Primary Reviewer

- @jimjbrettj @timwu20


noot (approved):

qdm12:
Let's wait before merging this, Tim has come up with a nice workaround solution for both cases mentioned.
I'll play around with it and then modify this document.

qdm12:
Actually merging for now since Tim's solution does not work when mocks depend/return other mocks. I'll investigate next time I toy with mocks, and probably do another PR to add a section to this document on that.

PR #2324, opened 2022-02-21
chore(lib/trie): add Trie `DeepCopy` unit test

qdm12:
## Changes

- Simple test addition for DeepCopy to reach full test coverage on trie.go

I often check for 'red' code paths in this file to make sure I cover everything since it's a rather complex source file. And that's the only red path showing up, and I wanted it gone since I would check it every time.

## Tests

```
go test ./lib/trie/...
```

### DIFF
--- a/lib/trie/trie_test.go
+++ b/lib/trie/trie_test.go
@@ -6,12 +6,14 @@ package trie
 import (
 	"bytes"
 	"encoding/hex"
+	"reflect"
 	"testing"
 
 	"github.com/ChainSafe/gossamer/internal/trie/node"
 	"github.com/ChainSafe/gossamer/lib/common"
 	gomock "github.com/golang/mock/gomock"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
 func Test_NewEmptyTrie(t *testing.T) {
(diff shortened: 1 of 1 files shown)


noot (approved):

kishansagathiya (changes requested):
The code looks good, but I am not sure why we need to check for mutations!

Why can't we just compare two structs (old struct and new one after copy) and its properties by reference?https://go.dev/play/p/zbmQpUbIEXp

timwu20 (approved):

qdm12:
> The code looks good, but I am not sure why we need to check for mutations!
> 
> Why can't we just compare two structs (old struct and new one after copy) and its properties by reference?https://go.dev/play/p/zbmQpUbIEXp

That works if fields are not pointers/slices but otherwise we should *ideally* check for mutation for pointer/slices fields. Even if the root element is copied, we want to check everything is deep copied including fields of the struct. So check for pointer fields, interfaces, maps and slices fields.

kishansagathiya (approved):
> That works if fields are not pointers/slices but otherwise we should ideally check for mutation for pointer/slices fields. Even if the root element is copied, we want to check everything is deep copied including fields of the struct. So check for pointer fields, interfaces, maps and slices fields.

I think all that should be possible. 
- You can compare generation with `&`
- You can compare root directly since it is going to be pointer mostly
- You can run a loop through childTries and compare tries using recursion of this whole trie-compare function.
- You can run a loop through deletedKeys and compare values using `&`

but, I will leave that upto you!

qdm12:
@kishansagathiya good suggestion, done 👍

I went a bit overboard and wrote [some helper functions to get the pointer of anything](https://github.com/ChainSafe/gossamer/blob/518a15ec030bcce2cb02308ae15f2520fd19e36b/lib/trie/trie_test.go#L177-L217) and code looks much cleaner now. All this `reflect` just made me realize I wrote https://github.com/qdm12/reprint which I could had used here 😄 but anyway too late

Issue #1785, opened 2021-09-10
update `*types.Header.Number` to be of type uint32


noot:
## Task summary

- currently the block number is represented by a `*big.Int`
- however substrate represents it with a uint32 as the blockchain will not grow higher than this
- gossamer should update this as it will likely make the math to do with block numbers cleaner

PR #2330, opened 2022-02-24
chore(all): replace block header type from `*big.Int` to `uint`

qdm12:
## Changes

- [x] Replace block header number field from `*big.Int` to `uint`
- [x] Minimal changes to production code
- [x] Remove now unused `MustHexToBigInt` and replace with `UintToHex`
- [x] Replace `big.NewInt(0).SetBytes` with  `common.BytesToUint`
- [x] Adapt tests
- [x] Change logs to use `%d` instead of `%s`
- [x] Test run the node for a while

## Notable/questionable changes

- Telemetry JSON has now `"height": 0` and `finalized_height": 0` not showing if zero because of the existing `omitempty` tag, whereas before a `&big.Int{}` would show as 0 (since it's not zero).

### Side changes

- Use `MustNewUint64OrHash` when using `NewUint64OrHash` without checking the error
- Remove `errWorkerMissingStartNumber` since we should really panic on nil pointer in `workerToRequests` I think.
- Use `const` when appropriate in tests
- Add names to interface arguments when modified
- Name returned values when modified
- Remove some comments when it's now obvious what is going on
- Use `assert.Regexp` in a test instead of `regexp.MustCompile`
- Refactor `equalNodeValue` and `equalLeaves` since I had trouble debugging a test

## Tests

All test suites passing

## Issues

- #1785 

## Primary Reviewer


noot (commented):
looks good! just one question, how is block number encoded now? it should be encoded as 4 bytes. I forget the reasoning for uint over uint32?

qdm12:
> it should be encoded as 4 bytes. I forget the reasoning for uint over uint32?

- For scale, `uint` is encoded like `*big.Int` so that's why it was picked over uint32
- I added some helping functions in `common` to encode uint 'compactly' like `*big.Int`'s `Bytes()` function does.

All in all no encoding/decoding was changed (at least for this PR and for now).

noot:
@qdm12 can you change this log to format nice haha

```
2022-03-04T16:01:17-05:00 INFO 🚣 currently syncing, 39 peers connected, target block number 9286741, 27.21 average blocks/second, 27.27 overall average, finalised block number %!s(uint=0) with hash 0x91b171bb158e2d3848fa23a9f1c25182fb8e20313b2c1eb49219da7a70ce90c3	chain_sync.go:L372	pkg=sync
```

noot (approved):

qdm12:
@noot indeed, my bad I thought I fixed them all, but a few slipped through (fixed in 1f96760ea9378f17aa9fbc229018645934f1e7a0)!

EclesioMeloJunior (approved):

kishansagathiya:
@qdm12 
> Telemetry JSON has now "height": 0 and finalized_height": 0 not showing if zero because of the existing omitempty tag, whereas before a &big.Int{} would show as 0 (since it's not zero).

Only `SystemInterval` has omitempty tag for height. It is okay if 0 does not show up. height if optional in system.interval. 

However, this PR uses uint, but substrate-telemetry always expects uint64 for height.

PR #2333, opened 2022-02-25
chore(lint): upgrade `golangci-lint` to `v1.44.x`

qdm12:
## Changes

- [x] Upgrade from `v1.43.0` to `v1.44`
- [x] Downloads and installs the latest feature release instead of pinning to a certain bug fix version
- [x] Fix newly detected errors

(Got sick of these 3 errors always showing up on my linter v1.44.x and that took under 5 minutes 😉)

## Tests

```
go install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.44
golangci-lint run
```

### DIFF
--- a/dot/telemetry/mailer.go
+++ b/dot/telemetry/mailer.go
@@ -57,7 +57,7 @@ func BootstrapMailer(ctx context.Context, conns []*genesis.TelemetryEndpoint, en
 		const maxRetries = 5
 
 		for connAttempts := 0; connAttempts < maxRetries; connAttempts++ {
-			conn, _, err := websocket.DefaultDialer.Dial(v.Endpoint, nil)
+			conn, response, err := websocket.DefaultDialer.Dial(v.Endpoint, nil)
 			if err != nil {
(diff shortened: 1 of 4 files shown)


noot (approved):

dot/telemetry/mailer.go
```diff
@@ -78,6 +78,11 @@ func BootstrapMailer(ctx context.Context, conns []*genesis.TelemetryEndpoint, en
 				}
 			}
 
+			err = response.Body.Close()
+			if err != nil {
+				mailer.logger.Errorf("cannot close body of response from %s: %s", v.Endpoint, err)
```
noot, line 83: does this need to be an error log? it doesn't really seem to matter if we were ignoring it before, change to debug perhaps?

qdm12: Changed it to warn because:

- I don't think this ever happens, I can't find any documentation/reason why the body close would fail on an http response
- The user can control the log level of the telemetry individually
- It's not really an error since execution continues
- It might happen i.e. if we run out of memory or some catastrophic situation perhaps, so using warn seems appropriate

edwardmack (approved):

PR #2339, opened 2022-03-02
fix(dot/digest): BABE NextEpochData and NextConfigData should be set on finalization

### LINKED ISSUE #2269: BABE NextEpochData and NextConfigData should be set on finalization, not import
## Task summary

- currently, BABE `NextEpochData` and `NextConfigData` are set on import of a block containing them (first block in an epoch)
- however, there is a chance there are two conflicting blocks that both contain these digests. these digests should be set in the database when the block is finalized, not when it's imported, to prevent us from using upcoming BABE data from a dropped fork.

### LINKED ISSUE #2367: Remove `DigestHandler` dependency from `dot/core` package
## Issue summary

- Looking at `dot/core` package the `DigestHandler` dependency is not used

- N/A

EclesioMeloJunior:
## Changes

- Only stores the `NextEpochData` and `NextConfigData` digests when the block is finalized.
- Stores the possible digests in a map and reset it once the chosen block is finalized.
- Just change `IncrementSetID` to return the new set id instead of doing a database call

## Tests

Running the unit tests
```
go test github.com/ChainSafe/gossamer/dot/digest
```
or running 3 gossamer nodes, with `--log-digest debug` flag enabled, you will be able to see, when the first block of a new epoch is produced all 3 nodes will store the possible `BABENextEpochData` digest and when the block is finalized you will see `BABENextEpochData` digest being defined for the finalized block.

## Issues

- Closes #2269
- Closes #2367

## Primary Reviewer

- @noot


dot/digest/digest.go
```diff
@@ -229,6 +256,37 @@ func (h *Handler) handleBlockFinalisation(ctx context.Context) {
 	}
 }
 
+// defineHeaderBabeDigest is called only when a block is finalised
+// and defines the correct next epoch data and next config data
+func (h *Handler) defineHeaderBabeDigest(header *types.Header) error {
```
noot, line 261: I'd name this function `setBABEDigestsOnFinalization` or something like that

dot/digest/digest.go
```diff
@@ -229,6 +256,37 @@ func (h *Handler) handleBlockFinalisation(ctx context.Context) {
...
+	h.babeConsensusDigestLock.Lock()
+	defer h.babeConsensusDigestLock.Unlock()
+
+	nextEpochData, has := h.nextEpochData[header.Hash()]
+	if has {
+		h.logger.Debugf("defining BABENextEpochData data: %v for hash: %s", nextEpochData, header.Hash())
```
noot, line 267:
```suggestion
		h.logger.Debugf("setting BABENextEpochData data: %v for hash: %s", nextEpochData, header.Hash())
```
```suggestion
		h.logger.Debugf("setting BABENextEpochData data: %v for hash: %s", nextEpochData, header.Hash())
```

dot/digest/digest.go
```diff
@@ -403,7 +451,7 @@ func (h *Handler) handleNextEpochData(act types.NextEpochData, header *types.Hea
 		return err
 	}
 
-	h.logger.Debugf("setting data for block number %s and epoch %d with data: %v",
+	h.logger.Debugf("defined data for block number %s and epoch %d with data: %v",
```
noot, line 454: maybe change this back? I don't think `defined` makes as much sense here

dot/digest/digest.go
```diff
@@ -229,6 +254,37 @@ func (h *Handler) handleBlockFinalisation(ctx context.Context) {
...
+
+	nextEpochData, has := h.nextEpochData[header.Hash()]
+	if has {
+		h.logger.Debugf("setting BABENextEpochData data: %v for hash: %s", nextEpochData, header.Hash())
+		err := h.handleNextEpochData(nextEpochData, header)
+		if err != nil {
+			return err
+		}
+
+		h.nextEpochData = make(map[common.Hash]types.NextEpochData)
```
noot, line 271: I don't think deleting the map entirely is the right move here, as there might be unfinalized digest info from the next epoch still in the map waiting to be finalized.

noot (commented):
I noticed this breaks syncing on polkadot :( I think this is because blocks are only finalized every ~512 blocks when first syncing, but epochs have around 200 blocks, so the data from the previous epoch isn't being set in time to process the next epoch.

the exact log I'm seeing is:
```
2022-03-04T15:55:44-05:00 EROR block data processing for block with hash 0xe940bd35e441a6061d2b20bd3f3c028bc55db56be5d01198e05afc7f46b9663b failed: could not verify block: failed to get verifier info for block 2520: failed to get epoch data for epoch 1: Key not found	chain_processor.go:L90	pkg=sync
```

so it's trying to process a block in epoch 1 but the epoch data hasn't been set yet. I think we need to add functionality to use the `NextEpochData`/`NextConfigData` that was set on the chain we're currently on, even if it's not finalized yet

EclesioMeloJunior:
> I noticed this breaks syncing on polkadot :( I think this is because blocks are only finalized every ~512 blocks when first syncing, but epochs have around 200 blocks, so the data from the previous epoch isn't being set in time to process the next epoch.
> 
> the exact log I'm seeing is:
> 
> ```
> 2022-03-04T15:55:44-05:00 EROR block data processing for block with hash 0xe940bd35e441a6061d2b20bd3f3c028bc55db56be5d01198e05afc7f46b9663b failed: could not verify block: failed to get verifier info for block 2520: failed to get epoch data for epoch 1: Key not found	chain_processor.go:L90	pkg=sync
> ```
> 
> so it's trying to process a block in epoch 1 but the epoch data hasn't been set yet. I think we need to add functionality to use the `NextEpochData`/`NextConfigData` that was set on the chain we're currently on, even if it's not finalized yet

@noot fixed the problem by: 
1. Handle the BABE Digest messages on every block import to keep the host updated with the next epoch config/data and store this configuration in a map.
2. On finalization the function `setBABEDigestsOnFinalization` is called to choose the right next epoch config/data using the block hash, store the chosen one and delete all the others for the epoch.

On my local test I reach epoch 18
```
DBUG setting data for block number 40791 and epoch 18 with data
```

dot/digest/digest.go
```diff
@@ -179,15 +189,41 @@ func (h *Handler) handleGrandpaConsensusDigest(digest scale.VaryingDataType, hea
...
-		return h.handleNextEpochData(val, header)
+		h.logger.Debugf("stored BABENextEpochData data: %v for hash: %s to epoch: %d\n", digest, headerHash, nextEpoch)
+
+		_, has := h.nextEpochData[nextEpoch]
+		if !has {
+			h.nextEpochData[nextEpoch] = make(map[common.Hash]types.NextEpochData)
+		}
+		h.nextEpochData[nextEpoch][headerHash] = val
+
+		return h.handleNextEpochData(val, nextEpoch, header)
```
noot, line 213: this works for bootstrap mode, since there's only really 1 chain, but it doesn't entirely fix the problem in the case that the previous epoch hasn't been finalized yet. for example if we have two forks A and B, and B arrives later than A, so the epoch data for B will be set in the database. however if we process blocks on fork A, they will use the info from fork B, which will result in them being marked invalid and thrown away even if they are valid. I think you probably need to modify the babe verifier to get the data from the right fork, if it isn't set in the database already (and only set it in the database on finalization)

dot/digest/digest.go
```diff
@@ -229,6 +271,44 @@ func (h *Handler) handleBlockFinalisation(ctx context.Context) {
...
 
+// setBABEDigestsOnFinalization is called only when a block is finalised
+// and defines the correct next epoch data and next config data
+func (h *Handler) setBABEDigestsOnFinalization(header *types.Header) error {
+	h.babeConsensusDigestLock.Lock()
+	defer h.babeConsensusDigestLock.Unlock()
+
+	currEpoch, err := h.epochState.GetEpochForBlock(header)
+	if err != nil {
+		return err
```
qdm12, line 282:
```suggestion
		return fmt.Errorf("cannot get epoch for block: %w", err)
```

dot/digest/digest_test.go
```diff
@@ -423,9 +422,35 @@ func TestHandler_HandleNextEpochData(t *testing.T) {
...
 
+	finalisedCh := make(chan *types.FinalisationInfo)
+
+	handler := newTestHandler(t)
+	handler.finalised = finalisedCh
+
+	ctx, cancel := context.WithCancel(context.Background())
+
+	var wg sync.WaitGroup
+	wg.Add(1)
```
qdm12, line 433: if it's just one in the wait group, just close a signal channel `done := make(chan struct{})` instead

Same for below changes

dot/state/grandpa.go
```diff
@@ -160,14 +160,19 @@ func (s *GrandpaState) SetNextChange(authorities []types.GrandpaVoter, number *b
...
 	if err != nil {
-		return err
+		return 0, err
 	}
 
 	nextSetID := currSetID + 1
-	return s.setCurrentSetID(nextSetID)
+	err = s.setCurrentSetID(nextSetID)
+	if err != nil {
+		return 0, err
```
qdm12, line 172:
```suggestion
		return 0, fmt.Errorf("cannot set current set ID: %w", err)
```

dot/state/epoch.go
```diff
@@ -45,6 +47,10 @@ type EpochState struct {
 	blockState  *BlockState
 	epochLength uint64 // measured in slots
 	skipToEpoch uint64
+
+	nextEpochLock  sync.RWMutex
+	nextEpochData  map[uint64]map[common.Hash]types.NextEpochData
+	nextConfigData map[uint64]map[common.Hash]types.NextConfigData
```
noot, line 53: can you add a comment saying what the `uint64` and `Hash` represent?

dot/state/epoch.go
```diff
@@ -365,3 +446,89 @@ func (s *EpochState) SkipVerify(header *types.Header) (bool, error) {
...
+		delete(s.nextEpochData, epoch)
+		return s.SetEpochData(epoch, epochData)
+	}
+
+	return nil
+}
+
+// FinalizeBABENextConfigDataToFinalize retrieves the types.NextConfigData by epoch and hash keys
+// and delete all the entries for the epoch
+func (s *EpochState) FinalizeBABENextConfigDataToFinalize(epoch uint64) error {
```
noot, line 509:
```suggestion
func (s *EpochState) FinalizeBABENextConfigData(epoch uint64) error {
```

dot/state/epoch.go
```diff
@@ -365,3 +446,89 @@ func (s *EpochState) SkipVerify(header *types.Header) (bool, error) {
...
+
+	for hash, nextEpochData := range epochData {
+		persisted, err := s.blockState.HasHeaderInDatabase(hash)
+		if err != nil {
+			return fmt.Errorf("failed to check header exists at database: %w", err)
+		}
+
+		if !persisted {
+			continue
+		}
```
noot, line 493: maybe add a comment here explaining what's happening since it's not obvious, if the header is in the database then it's been finalized and thus we can also set the corresponding EpochData in the database

dot/state/epoch.go
```diff
@@ -365,3 +446,89 @@ func (s *EpochState) SkipVerify(header *types.Header) (bool, error) {
...
+		epochData, err := nextEpochData.ToEpochData()
+		if err != nil {
+			return fmt.Errorf("cannot transform epoch data: %w", err)
+		}
+
+		delete(s.nextEpochData, epoch)
+		return s.SetEpochData(epoch, epochData)
+	}
+
+	return nil
```
noot, line 504: return an error here since no epoch data was set in the db

lib/babe/verify.go
```diff
@@ -219,16 +227,24 @@ func (v *VerificationManager) getVerifierInfo(epoch uint64) (*verifierInfo, erro
...
-			return v.epochState.GetConfigData(uint64(i))
+		configData, err := v.epochState.GetConfigData(uint64(i))
+		if errors.Is(chaindb.ErrKeyNotFound, err) {
+			configData, err = v.epochState.GetConfigDataForHeader(epoch, header)
+			if err != nil {
+				return nil, fmt.Errorf("config data was not found for epoch %d: %w", epoch, err)
+			}
 		}
+
+		return configData, err
```
noot, line 247: I don't think this is exactly right, in the case of unfinalized config data, then the loop will `continue` and not find it. I think you need to move the logic that checks `GetConfigData` and `GetConfigDataForHeader` to the top of the loop, and if both are nil, then continue. otherwise if one of them exists, then return it

EclesioMeloJunior: I have updated the `HasConfigData(uint64(i))` to check the in-memory next config data as well, so:

```
has, err := v.epochState.HasConfigData(uint64(i))
if !has { continue }
```
means that the config data is not in the database nor in the map otherwise, the data exists in one of them (database or map), I first check for the database if I get an `chaindb.ErrKeyNotFound` error then I search in the map.

dot/state/epoch.go
```diff
@@ -45,6 +52,13 @@ type EpochState struct {
 	blockState  *BlockState
 	epochLength uint64 // measured in slots
 	skipToEpoch uint64
+
+	nextEpochLock sync.RWMutex
+
+	// nextEpochData and nextConfigData is splited into
+	// epoch (uint64) and the block hash (common.Hash) that contains the digest
+	nextEpochData  map[uint64]map[common.Hash]types.NextEpochData
+	nextConfigData map[uint64]map[common.Hash]types.NextConfigData
```
qdm12, line 61: - Split the comments so it appears when hovering over it in IDEs
- They don't contain the `digest`, but a next-epoch-data or next-config-data

```suggestion
	// nextEpochData follows the format map[epoch]map[block hash]next epoch data
	nextEpochData  map[uint64]map[common.Hash]types.NextEpochData
	// nextConfigData follows the format map[epoch]map[block hash]next config data
	nextConfigData map[uint64]map[common.Hash]types.NextConfigData
```

dot/state/epoch.go
```diff
@@ -236,6 +254,32 @@ func (s *EpochState) GetEpochData(epoch uint64) (*types.EpochData, error) {
...
+		if err != nil {
+			return nil, fmt.Errorf("cannot verify the ancestry: %w", err)
+		}
+
+		if isDescendant {
+			return value.ToEpochData()
+		}
+	}
+
+	return nil, fmt.Errorf("cannot find epoch data for header %s", headerHash)
```
qdm12, line 280: add a sentinel error for this

dot/state/epoch.go
```diff
@@ -288,6 +341,32 @@ func (s *EpochState) GetConfigData(epoch uint64) (*types.ConfigData, error) {
...
 }
 
+// GetConfigDataForHeader retrieves the right epoch configuration data that belongs to the header parameter
+func (s *EpochState) GetConfigDataForHeader(epoch uint64, header *types.Header) (*types.ConfigData, error) {
+	s.nextEpochLock.RLock()
+	defer s.nextEpochLock.RUnlock()
+
+	atEpoch, has := s.nextConfigData[epoch]
+	if !has {
+		return nil, fmt.Errorf("epoch %d not found in memory stored config data", epoch)
```
qdm12, line 351: use a sentinel error

dot/state/epoch.go
```diff
@@ -301,7 +380,17 @@ func (s *EpochState) GetLatestConfigData() (*types.ConfigData, error) {
...
+
+	if !has || errors.Is(chaindb.ErrKeyNotFound, err) {
+		s.nextEpochLock.Lock()
+		defer s.nextEpochLock.Unlock()
+
+		_, has = s.nextConfigData[epoch]
+		return has, nil
+	}
+
+	return has, err
```
qdm12, line 393: - Less indentation
- Wrap database error if it's not key not found
- I think `has` can never be false from the database

```suggestion
	_, err := s.db.Has(configDataKey(epoch))
	if err == nil {
		return true, nil
	}

	if !errors.Is(chaindb.ErrKeyNotFound, err) {
		return false, fmt.Errorf("cannot check database for config data key of epoch %d: %w", epoch, err)
	}
	
	// Key not found in database
	s.nextEpochLock.Lock()
	defer s.nextEpochLock.Unlock()
	_, has := s.nextConfigData[epoch]
	return has, nil
```

dot/state/epoch.go
```diff
@@ -45,6 +52,13 @@ type EpochState struct {
 	blockState  *BlockState
 	epochLength uint64 // measured in slots
 	skipToEpoch uint64
+
+	nextEpochLock sync.RWMutex
```
qdm12, line 56: nit should we have two mutexes, one for each map?

dot/state/epoch.go
```diff
@@ -365,3 +454,143 @@ func (s *EpochState) SkipVerify(header *types.Header) (bool, error) {
...
+// StoreBABENextEpochData stores the types.NextEpochData under epoch and hash keys
+func (s *EpochState) StoreBABENextEpochData(epoch uint64, hash common.Hash, val types.NextEpochData) {
+	s.nextEpochLock.Lock()
+	defer s.nextEpochLock.Unlock()
+
+	_, has := s.nextEpochData[epoch]
+	if !has {
+		s.nextEpochData[epoch] = make(map[common.Hash]types.NextEpochData)
+	}
+	s.nextEpochData[epoch][hash] = val
```
qdm12, line 467: nit `nextEpochData` would be better than `val`

dot/state/epoch.go
```diff
@@ -365,3 +454,143 @@ func (s *EpochState) SkipVerify(header *types.Header) (bool, error) {
...
+// StoreBABENextConfigData stores the types.NextConfigData under epoch and hash keys
+func (s *EpochState) StoreBABENextConfigData(epoch uint64, hash common.Hash, val types.NextConfigData) {
+	s.nextEpochLock.Lock()
+	defer s.nextEpochLock.Unlock()
+
+	_, has := s.nextConfigData[epoch]
+	if !has {
+		s.nextConfigData[epoch] = make(map[common.Hash]types.NextConfigData)
+	}
+	s.nextConfigData[epoch][hash] = val
```
qdm12, line 479: nit `nextConfigData` would be better than `val`

dot/state/epoch.go
```diff
@@ -365,3 +454,143 @@ func (s *EpochState) SkipVerify(header *types.Header) (bool, error) {
...
+	if !has {
+		s.nextEpochData[epoch] = make(map[common.Hash]types.NextEpochData)
+	}
+	s.nextEpochData[epoch][hash] = val
+}
+
+// StoreBABENextConfigData stores the types.NextConfigData under epoch and hash keys
+func (s *EpochState) StoreBABENextConfigData(epoch uint64, hash common.Hash, val types.NextConfigData) {
+	s.nextEpochLock.Lock()
+	defer s.nextEpochLock.Unlock()
```
qdm12, line 473: Are these new methods called a lot? I doubt they are, but I'm just thinking we could have a mutex per submap as well to reduce lock contention. 90% probability we don't, but I'm wondering :wink:

dot/state/epoch.go
```diff
@@ -365,3 +454,143 @@ func (s *EpochState) SkipVerify(header *types.Header) (bool, error) {
...
+		return fmt.Errorf("cannot transform epoch data: %w", err)
+	}
+
+	err = s.SetEpochData(epoch, ed)
+	if err != nil {
+		return fmt.Errorf("cannot set epoch data: %w", err)
+	}
+
+	// remove previous epochs from the memory
+	epochsToDelete := make([]uint64, 0)
```
qdm12, line 506: nit Let's assign it some approximate capacity

```suggestion
	epochsToDelete := make([]uint64, 0, len(s.nextEpochData)/2)
```

dot/state/epoch.go
```diff
@@ -365,3 +454,143 @@ func (s *EpochState) SkipVerify(header *types.Header) (bool, error) {
...
+	}
+
+	cd := finalizedNextConfigData.ToConfigData()
+	err = s.SetConfigData(epoch, cd)
+	if err != nil {
+		return fmt.Errorf("cannot set config data: %w", err)
+	}
+
+	// remove previous epochs from the memory
+	epochsToDelete := make([]uint64, 0)
```
qdm12, line 540: nit no need to precise the `0` length, this works too:

```suggestion
	epochsToDelete := make([]uint64)
```

or just

```go
	var epochsToDelete []uint64
```

dot/state/epoch.go
```diff
@@ -365,3 +454,143 @@ func (s *EpochState) SkipVerify(header *types.Header) (bool, error) {
...
+	for e := range s.nextConfigData {
+		if e <= epoch {
+			epochsToDelete = append(epochsToDelete, e)
```
qdm12, line 543: nit by the way, you can safely delete keys inside the map as you iterate through it. It may look odd though :smile:

dot/state/epoch.go
```diff
@@ -365,3 +454,143 @@ func (s *EpochState) SkipVerify(header *types.Header) (bool, error) {
...
+func (s *EpochState) lookForPersistedHashForEpochData(epoch uint64) (types.NextEpochData, error) {
+	epochData, has := s.nextEpochData[epoch]
+	if !has {
+		return types.NextEpochData{}, errEpochNotFound
+	}
+
+	for hash, inMemoryEpochData := range epochData {
+		persisted, err := s.blockState.HasHeaderInDatabase(hash)
+		if err != nil {
+			return types.NextEpochData{}, fmt.Errorf("failed to check header exists at database: %w", err)
```
qdm12, line 563: nit

```suggestion
			return types.NextEpochData{}, fmt.Errorf("failed to check header exist in database: %w", err)
```

dot/state/epoch.go
```diff
@@ -365,3 +454,143 @@ func (s *EpochState) SkipVerify(header *types.Header) (bool, error) {
...
+func (s *EpochState) lookForPersistedHashForConfigData(epoch uint64) (types.NextConfigData, error) {
+	configData, has := s.nextConfigData[epoch]
+	if !has {
+		return types.NextConfigData{}, errEpochNotFound
+	}
+
+	for hash, inMemoryNextConfigData := range configData {
+		persisted, err := s.blockState.HasHeaderInDatabase(hash)
+		if err != nil {
+			return types.NextConfigData{}, fmt.Errorf("failed to check header exists at database: %w", err)
```
qdm12, line 585: nit

```suggestion
			return types.NextConfigData{}, fmt.Errorf("failed to check header exist in database: %w", err)
```

lib/babe/verify.go
```diff
@@ -127,6 +129,7 @@ func (v *VerificationManager) SetOnDisabled(index uint32, header *types.Header)
 }
 
 // VerifyBlock verifies that the block producer for the given block was authorized to produce it.
+// It checks the next epoch and config data stored in memory only if it cannot retrieves the data from database
```
qdm12, line 132: tinynit

```suggestion
// It checks the next epoch and config data stored in memory only if it cannot retrieve the data from database
```

dot/state/epoch.go
```diff
@@ -365,3 +454,143 @@ func (s *EpochState) SkipVerify(header *types.Header) (bool, error) {
...
+		persisted, err := s.blockState.HasHeaderInDatabase(hash)
+		if err != nil {
+			return types.NextEpochData{}, fmt.Errorf("failed to check header exists at database: %w", err)
+		}
+
+		if !persisted {
+			continue
+		}
+
+		return inMemoryEpochData, nil
```
noot, line 570: I'd add a comment here to explain what's happening, since it's not obvious. same for `lookForPersistedHashForConfigData`

dot/state/epoch.go
```diff
@@ -236,6 +254,32 @@ func (s *EpochState) GetEpochData(epoch uint64) (*types.EpochData, error) {
 	return raw.ToEpochData()
 }
 
+// GetEpochDataForHeader retrieves the right epoch data that belongs to the header parameter
+func (s *EpochState) GetEpochDataForHeader(epoch uint64, header *types.Header) (*types.EpochData, error) {
```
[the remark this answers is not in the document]
noot: nevermind, I see that in the babe verifier it first checks the database, then checks this for the in-memory epoch data. maybe it would be simpler to have the babe verifier call one function called `GetEpochData` which first checks the database, then the in-memory map?

lib/babe/verify_integration_test.go
```diff
@@ -428,3 +433,316 @@ func TestVerifyAuthorshipRight_Equivocation(t *testing.T) {
...
+		{
+			Key: keyring.KeyGeorge.Public().(*sr25519.PublicKey).AsBytes(),
+		},
+		{
+			Key: keyring.KeyHeather.Public().(*sr25519.PublicKey).AsBytes(),
+		},
+		{
+			Key: keyring.KeyIan.Public().(*sr25519.PublicKey).AsBytes(),
+		},
+	}
```
qdm12, line 470: nit simpler

```suggestion
	keyPairs := []*sr25519.Keypair{
		keyring.KeyAlice, keyring.KeyBob, keyring.KeyCharlie,
		 keyring.KeyDave,  keyring.KeyEve,keyring.KeyFerdie,
		 keyring.KeyGeorge, keyring.KeyHeather, keyring.KeyIan,
	}
	authorities := make([]types.AuthorityRaw, len(keyPairs))
	for i, keyPair := range keyPairs {
		authorities[i] = types.AuthorityRaw{
			Key: keyPair.Public().(*sr25519.PublicKey).AsBytes(),
		}
	}
```

lib/babe/verify_integration_test.go
```diff
@@ -428,3 +433,316 @@ func TestVerifyAuthorshipRight_Equivocation(t *testing.T) {
...
+	})
+	require.NoError(t, err)
+
+	return headerWhoOwnsNextEpochDigest
+}
+
+// issueNewBlockFrom will create and store a new block following a chain
+func issueNewBlockFrom(t *testing.T, parentHeader *types.Header,
+	kp *sr25519.Keypair, stateService *state.Service) *types.Header {
+	output, proof, err := kp.VrfSign(makeTranscript(Randomness{}, uint64(1), 1))
```
qdm12, line 718: nit

```suggestion
	kp *sr25519.Keypair, stateService *state.Service) *types.Header {
	t.Helper()

	output, proof, err := kp.VrfSign(makeTranscript(Randomness{}, uint64(1), 1))
```

lib/babe/verify_integration_test.go
```diff
@@ -428,3 +433,316 @@ func TestVerifyAuthorshipRight_Equivocation(t *testing.T) {
...
+	require.True(t, verifierInfo.secondarySlots)
+	require.Equal(t, expectedThreshold, verifierInfo.threshold)
+}
+
+// issueConsensusDigestsBlocksFromGenesis will create different
+// blocks that contains different consensus messages digests
+func issueConsensusDigestsBlockFromGenesis(t *testing.T, genesisHeader *types.Header,
+	kp *sr25519.Keypair, stateService *state.Service,
+	nextEpoch types.NextEpochData, nextConfig types.NextConfigData) *types.Header {
+	output, proof, err := kp.VrfSign(makeTranscript(Randomness{}, uint64(0), 0))
```
qdm12, line 663:
```suggestion
	nextEpoch types.NextEpochData, nextConfig types.NextConfigData) *types.Header {
	t.Helper()

	output, proof, err := kp.VrfSign(makeTranscript(Randomness{}, uint64(0), 0))
```

dot/state/epoch_test.go
```diff
@@ -9,13 +9,45 @@ import (
...
+	"github.com/ChainSafe/gossamer/lib/common"
 	"github.com/ChainSafe/gossamer/lib/crypto/sr25519"
 	"github.com/ChainSafe/gossamer/lib/keystore"
 	"github.com/ChainSafe/gossamer/pkg/scale"
 
 	"github.com/stretchr/testify/require"
 )
 
+var keyring, _ = keystore.NewSr25519Keyring()
+var authoritiesRaw = []types.AuthorityRaw{
```
qdm12, line 21: Move these in the test where they're used, let's avoid using global variables.

dot/digest/digest_test.go
```diff
@@ -420,9 +419,34 @@ func TestHandler_HandleNextEpochData(t *testing.T) {
...
+	handler.finalised = finalisedCh
+
+	ctx, cancel := context.WithCancel(context.Background())
+
+	doneCh := make(chan struct{})
+
+	go func() {
+		defer close(doneCh)
+		handler.handleBlockFinalisation(ctx)
+	}()
```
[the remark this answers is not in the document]
EclesioMeloJunior: This does not work, `handleBlockFinalisation` contains:
````
for {
		select {
		case info := <-h.finalised:
                 case <-ctx.Done():
}
```
And if I send the data through `finalisedCh` before synchronously calling `handler.handleBlockFinalisation(ctx)`, then it will block since the `h.finalised` keeps open and the `<-ctx.Done()` is not trigged yet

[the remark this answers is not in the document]
EclesioMeloJunior: An alternative is: creating a `context.WithTimeout` and passing it to the `handler.handleBlockFinalisation(ctx)` and then wait for the `<-ctx.Done()` be trigged

dot/state/epoch.go
```diff
@@ -248,7 +292,16 @@ func (s *EpochState) GetLatestEpochData() (*types.EpochData, error) {
...
+	has, err := s.db.Has(epochDataKey(epoch))
+	if errors.Is(chaindb.ErrKeyNotFound, err) {
+		s.nextEpochLock.Lock()
+		defer s.nextEpochLock.Unlock()
+
+		_, has = s.nextEpochData[epoch]
+		return has, nil
+	}
+
+	return has, err
```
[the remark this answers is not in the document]
EclesioMeloJunior: `has` can be false if `err == nil`

dot/digest/digest.go
```diff
@@ -197,7 +200,7 @@ func (h *Handler) handleBabeConsensusDigest(digest scale.VaryingDataType, header
 		return h.handleBABEOnDisabled(val, header)
 
 	case types.NextConfigData:
-		h.logger.Debugf("stored BABENextConfigData data: %v for hash: %s to epoch: %d", digest, headerHash, nextEpoch)
+		h.logger.Debugf("stored BABENextConfigData data: %v for hash: %s to epoch: %d\n", digest, headerHash, nextEpoch)
```
qdm12, line 203: no need for the new line at the end?

```suggestion
		h.logger.Debugf("stored BABENextConfigData data: %v for hash: %s to epoch: %d", digest, headerHash, nextEpoch)
```

dot/digest/digest.go
```diff
@@ -234,7 +237,7 @@ func (h *Handler) handleBlockFinalisation(ctx context.Context) {
 
 			err := h.setBABEDigestsOnFinalization(&info.Header)
 			if err != nil {
-				h.logger.Errorf("failed to store babe next epoch digest: %s", err)
+				h.logger.Errorf("failed to store babe next epoch digest: %s\n", err)
```
qdm12, line 240:
```suggestion
				h.logger.Errorf("failed to store babe next epoch digest: %s", err)
```

dot/digest/digest.go
```diff
@@ -259,20 +262,19 @@ func (h *Handler) setBABEDigestsOnFinalization(finalizedHeader *types.Header) er
 	nextEpoch := currEpoch + 1
 
 	err = h.epochState.FinalizeBABENextEpochData(nextEpoch)
-	if err != nil {
-		h.logger.Debugf("failed to store BABENextEpochData for block number %d (%s): %w",
+	if !errors.Is(err, state.ErrEpochNotInMemory) {
+		return fmt.Errorf("cannot finalize babe next epoch data for block number %d (%s): %w",
 			finalizedHeader.Number, finalizedHeader.Hash(), err)
-		return err
 	}
```
qdm12, line 268: Potato question, if we get `state.ErrEpochNotInMemory`, should we just continue execution?

dot/digest/digest.go
```diff
@@ -259,20 +262,19 @@ func (h *Handler) setBABEDigestsOnFinalization(finalizedHeader *types.Header) er
...
 
 	err = h.epochState.FinalizeBABENextConfigData(nextEpoch)
-	if err != nil {
-		h.logger.Debugf("failed to store BABENextConfigData for block number %d (%s): %w",
+	if !errors.Is(err, state.ErrEpochNotInMemory) {
```
qdm12, line 271: if `err == nil`, this block will trigger I think, you should change this right?

dot/digest/digest_test.go
```diff
@@ -418,36 +420,37 @@ func TestHandler_HandleNextEpochData(t *testing.T) {
...
 
-	go func() {
-		defer close(doneCh)
-		handler.handleBlockFinalisation(ctx)
-	}()
+	handler.finalised = make(chan *types.FinalisationInfo, 1)
 
-	err = handler.handleConsensusDigest(d, header)
-	require.NoError(t, err)
+	const timeout = 1 * time.Second
```
qdm12, line 437: nit

```suggestion
	const timeout = time.Second
```

dot/digest/digest_test.go
```diff
@@ -479,39 +484,42 @@ func TestHandler_HandleNextConfigData(t *testing.T) {
...
-	err = handler.handleConsensusDigest(d, header)
+	blockHeaderKey := append([]byte("hdr"), header.Hash().ToBytes()...)
+	blockHeaderKey = append([]byte("block"), blockHeaderKey...)
+	err = stateSrv.DB().Put(blockHeaderKey, []byte{})
 	require.NoError(t, err)
 
-	finalisedCh <- &types.FinalisationInfo{
+	handler.finalised = make(chan *types.FinalisationInfo, 1)
+
+	const timeout = 1 * time.Second
```
qdm12, line 501: nit

```suggestion
	const timeout = time.Second
```

dot/digest/digest_test.go
```diff
@@ -479,39 +484,42 @@ func TestHandler_HandleNextConfigData(t *testing.T) {
...
 		SetID:  1,
 	}
 
+	handler.handleBlockFinalisation(ctx)
+
 	// Before check the config data was stored
 	// we need to wait for both handle functions finish
-	cancel()
-	<-doneCh
+	<-ctx.Done()
```
qdm12, line 515: potato question: Do we need to wait for the context to be canceled here?

dot/digest/digest_test.go
```diff
@@ -418,36 +420,37 @@ func TestHandler_HandleNextEpochData(t *testing.T) {
...
 		SetID:  1,
 	}
 
+	handler.handleBlockFinalisation(ctx)
+
 	// Before check the epoch data was stored
 	// we need to wait for both handle functions to finish
-	cancel()
-	<-doneCh
+	<-ctx.Done()
```
qdm12, line 451: potato question: Do we need to wait for the context to be canceled here?

qdm12 (changes requested):

dot/digest/digest.go
EclesioMeloJunior: Yes, we should. First we try to finalize the `next-epoch-data` if the `state.ErrEpochNotInMemory` error occur then we should try to finalize `next-config-data` and if the `state.ErrEpochNotInMemory` error occur again then we return `ErrCannotDefineNextEpoch`. We should finalize at least one next epoch information, normally those info comes in the same block but sometimes only one of them appears

dot/digest/digest_test.go
qdm12: Need an answer please :wink:

dot/digest/digest.go
```diff
@@ -15,6 +17,8 @@ import (
 
 var (
 	_ services.Service = &Handler{}
+
+	ErrCannotDefineNextEpoch = errors.New("cannot define next epoch data and config")
```
qdm12, line 21: nit maybe rename to `ErrNextEpochDefinition`?

dot/digest/digest.go
```diff
@@ -225,6 +257,35 @@ func (h *Handler) handleBlockFinalisation(ctx context.Context) {
...
+	err = h.epochState.FinalizeBABENextConfigData(nextEpoch)
+	if err == nil {
+		return nil
+	} else if !errors.Is(err, state.ErrEpochNotInMemory) {
+		return fmt.Errorf("cannot finalize babe next config data for block number %d (%s): %w",
+			finalizedHeader.Number, finalizedHeader.Hash(), err)
+	}
+
+	// the epoch state does not contains any information about the next epoch
+	return ErrCannotDefineNextEpoch
```
qdm12, line 286: nit maybe flip these around

```suggestion
	} else if errors.Is(err, state.ErrEpochNotInMemory) {
		// the epoch state does not contains any information about the next epoch
		return ErrCannotDefineNextEpoch
	}

	return fmt.Errorf("cannot finalize babe next config data for block number %d (%s): %w",
		finalizedHeader.Number, finalizedHeader.Hash(), err)
```

dot/state/epoch.go
```diff
@@ -220,8 +240,30 @@ func (s *EpochState) SetEpochData(epoch uint64, info *types.EpochData) error {
...
+// if the header params is nil then it will search only in database
+func (s *EpochState) GetEpochData(epoch uint64, header *types.Header) (*types.EpochData, error) {
+	epochData, err := s.GetEpochDataInDatabase(epoch)
+	if err == nil && epochData != nil {
+		return epochData, nil
+	}
+
+	// if no header is given then skip the lookup in-memory
+	if !errors.Is(err, chaindb.ErrKeyNotFound) || header == nil {
+		return nil, fmt.Errorf("failed to get epoch data from database: %w", err)
```
qdm12, line 254: If the error is `nil` and the header is nil as well, the error will be `nil` and the error wrap will not wrap properly. Maybe have these two conditions as two separate if conditions?

dot/state/epoch.go
```diff
@@ -272,8 +353,29 @@ func (s *EpochState) setLatestConfigData(epoch uint64) error {
 	return s.db.Put(latestConfigDataKey, buf)
 }
 
-// GetConfigData returns the BABE config data for a given epoch
-func (s *EpochState) GetConfigData(epoch uint64) (*types.ConfigData, error) {
+// GetConfigData returns the config data for a given epoch persisted in database
+// otherwise will try to get the data from the in-memory map using the header
+// if the header params is nil then it will search only in database
```
qdm12, line 358: nit

```suggestion
// GetConfigData returns the config data for a given epoch persisted in the database
// and otherwise tries to get the data from the in-memory map using the header.
// If the header params is nil then it will search only in the database
```

dot/state/epoch.go
```diff
@@ -365,3 +506,125 @@ func (s *EpochState) SkipVerify(header *types.Header) (bool, error) {
...
+		return fmt.Errorf("cannot transform epoch data: %w", err)
+	}
+
+	err = s.SetEpochData(epoch, ed)
+	if err != nil {
+		return fmt.Errorf("cannot set epoch data: %w", err)
+	}
+
+	// remove previous epochs from the memory
+	epochsToDelete := make([]uint64, 0)
```
qdm12, line 558: nit what's the point to assign it to a length of `0`?

Just do this instead?

```suggestion
	var epochsToDelete []uint64
```

dot/state/epoch.go
qdm12: @EclesioMeloJunior is this done? My potato near midnight eyes still see `return nil`  :eyes:

dot/state/epoch.go
qdm12: There is no reason to keep the `0` length precision. Also if this is a path executed often or there are a big number of keys in the map, it might be worth deleting keys in the map whilst iterating over it.

dot/state/epoch.go
```diff
@@ -365,3 +506,125 @@ func (s *EpochState) SkipVerify(header *types.Header) (bool, error) {
...
+	}
+
+	for _, e := range epochsToDelete {
+		delete(s.nextConfigData, e)
+	}
+
+	return nil
+}
+
+// lookupForNextEpochPersistedHash given a specif epoch (the key) will go through the hashes looking
```
qdm12, line 606:
```suggestion
// lookupForNextEpochPersistedHash given a specific epoch (the key) will go through the hashes looking
```

dot/state/epoch.go
```diff
@@ -365,3 +506,125 @@ func (s *EpochState) SkipVerify(header *types.Header) (bool, error) {
...
+// for a database persisted hash (belonging to the finalized chain)
+// which contains the right configuration to be persisted and safely used
+func lookupForNextEpochPersistedHash[T types.NextConfigData | types.NextEpochData](
```
qdm12, line 609: :exploding_head: !!!

dot/state/epoch.go
```diff
@@ -365,3 +506,125 @@ func (s *EpochState) SkipVerify(header *types.Header) (bool, error) {
...
+	nextEpochMap map[uint64]map[common.Hash]T, es *EpochState, epoch uint64) (next *T, err error) {
+	hashes, has := nextEpochMap[epoch]
+	if !has {
+		return nil, ErrEpochNotInMemory
+	}
+
+	for hash, inMemoryNextEpochConfig := range hashes {
```
qdm12, line 616: nit change `inMemoryNextEpochConfig` to something more generic... since it uses generics

dot/state/epoch.go
kishansagathiya: exactly!!! mind == blown!! For a while I thought it is a bug!

kishansagathiya (changes requested):
Haven't gone through everything!! will continue!

dot/digest/digest.go
```diff
@@ -225,6 +257,35 @@ func (h *Handler) handleBlockFinalisation(ctx context.Context) {
 	}
 }
 
+// setBABEDigestsOnFinalization is called only when a block is finalised
+// and defines the correct next epoch data and next config data.
+func (h *Handler) setBABEDigestsOnFinalization(finalizedHeader *types.Header) error {
```
kishansagathiya, line 262: We are setting next epoch data and next config data here. Rename the function signature to reflect that. `setBABEDigestsOnFinalization` seems misleading to me.


tbh, I liked the separate functions as they were previously,

dot/digest/digest.go
qdm12: Maybe `ErrDefineNextEpoch`? The `Cannot` part shouldn't be part of the error variable name I think.

dot/state/epoch.go
EclesioMeloJunior: good catch!!

dot/digest/digest.go
```diff
@@ -118,12 +122,14 @@ func (h *Handler) NextGrandpaAuthorityChange() (next uint) {
...
-					header.Number, i, d.Value(), err)
-			}
+		if !ok {
+			continue
+		}
+
+		err := h.handleConsensusDigest(&val, header)
+		if err != nil {
+			h.logger.Errorf("cannot handle digests for block number %d, index %d, digest %s: %s",
+				header.Number, i, d.Value(), err)
```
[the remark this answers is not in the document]
EclesioMeloJunior: Using the `val` and changed the `func (d *ConsensusDigest) String()` to `func (d ConsensusDigest) String()` so now the `%s` can call the `String()` method in both cases `val` or `d.Value()`, before it was not possible because the `String()` method belonging to the type `*ConsensusDigest` and `val` and `d.Value()` is of type `ConsensusDigest`.

dot/digest/digest.go
EclesioMeloJunior: Can we change the name to `persistBABEDigestsForNextEpoch` and keep this function as it is?

dot/digest/digest.go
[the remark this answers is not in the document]
qdm12: Ahhhh good job

dot/digest/digest.go
kishansagathiya: > @kishansagathiya can we change the name to persistBABEDigestsForNextEpoch and keep this function as it is?

why?
that's basically same as what you have right now.

An appropriate name would be setNextEpochDataAndSetNextConfigData, And then the error that you get could be because of not being able to set epoch data or config data. But a better would be to just split this function into setNextEpochData and setNextConfigData. `setNextEpochDataAndSetNextConfigData` looks like two separate funcs are merged together.

If addressing this makes your life difficult (having to change a lot of code, not being able to write tests), I am cool with you not splitting the function.

dot/digest/digest.go
```diff
@@ -225,6 +257,35 @@ func (h *Handler) handleBlockFinalisation(ctx context.Context) {
...
+	if err != nil && !errors.Is(err, state.ErrEpochNotInMemory) {
+		return fmt.Errorf("cannot finalize babe next epoch data for block number %d (%s): %w",
+			finalizedHeader.Number, finalizedHeader.Hash(), err)
+	}
+
+	err = h.epochState.FinalizeBABENextConfigData(nextEpoch)
+	if err == nil {
+		return nil
+	} else if errors.Is(err, state.ErrEpochNotInMemory) {
+		return ErrDefineNextEpoch
```
qdm12, line 280: nit should we also add the detail error as a string?

```suggestion
		return fmt.Errorf("%w: %s", ErrDefineNextEpoch, err)
```

dot/state/epoch.go
```diff
@@ -220,8 +240,32 @@ func (s *EpochState) SetEpochData(epoch uint64, info *types.EpochData) error {
...
+	if err != nil && !errors.Is(err, chaindb.ErrKeyNotFound) {
+		return nil, fmt.Errorf("failed to get epoch data from database: %w", err)
+	} else if header == nil {
+		// if no header is given then skip the lookup in-memory
+		return epochData, nil
+	}
+
+	epochData, err = s.GetEpochDataForHeader(epoch, header)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get epoch data from memory: %w", err)
```
qdm12, line 261: nit should `GetEpochDataForHeader` be renamed to `GetEpochDataInMemory` so it uses the same naming format as `GetEpochDataInDatabase`?

dot/state/epoch.go
```diff
@@ -220,8 +240,32 @@ func (s *EpochState) SetEpochData(epoch uint64, info *types.EpochData) error {
...
+	}
+
+	return epochData, nil
+}
+
+// GetEpochDataInDatabase returns the epoch data for a given epoch persisted in database
+func (s *EpochState) GetEpochDataInDatabase(epoch uint64) (*types.EpochData, error) {
```
qdm12, line 268: unexport this

dot/state/epoch.go
```diff
@@ -236,19 +280,58 @@ func (s *EpochState) GetEpochData(epoch uint64) (*types.EpochData, error) {
 	return raw.ToEpochData()
 }
 
+// GetEpochDataForHeader retrieves the right epoch data that belongs to the header parameter
+func (s *EpochState) GetEpochDataForHeader(epoch uint64, header *types.Header) (*types.EpochData, error) {
```
qdm12, line 284: ideally unexport this, and use `GetEpochData` in the lib/babe integration test?

dot/state/epoch.go
```diff
@@ -288,6 +395,32 @@ func (s *EpochState) GetConfigData(epoch uint64) (*types.ConfigData, error) {
 	return info, nil
 }
 
+// GetConfigDataForHeader retrieves the BABE config data for a given epoch that belongs to the header parameter
+func (s *EpochState) GetConfigDataForHeader(epoch uint64, header *types.Header) (*types.ConfigData, error) {
```
qdm12, line 399: nit rename to `GetConfigDataFromMemoryForHeader`?

dot/state/epoch.go
```diff
@@ -365,3 +511,115 @@ func (s *EpochState) SkipVerify(header *types.Header) (bool, error) {
...
+	nextEpochMap map[uint64]map[common.Hash]T, es *EpochState, epoch uint64) (next *T, err error) {
+	hashes, has := nextEpochMap[epoch]
+	if !has {
+		return nil, ErrEpochNotInMemory
+	}
+
+	for hash, inMemory := range hashes {
+		persisted, err := es.blockState.HasHeaderInDatabase(hash)
+		if err != nil {
+			return nil, fmt.Errorf("failed to check header exists at database: %w", err)
```
qdm12, line 614:
```suggestion
			return nil, fmt.Errorf("failed to check header exists in database: %w", err)
```

qdm12 (approved):

lib/babe/verify_integration_test.go
```diff
@@ -427,3 +432,296 @@ func TestVerifyAuthorshipRight_Equivocation(t *testing.T) {
 	err = verifier.verifyAuthorshipRight(&block2.Header)
 	require.Equal(t, ErrProducerEquivocated, err)
 }
+
+func TestVerifyForkBlocksWithRespectiveEpochData(t *testing.T) {
```
kishansagathiya, line 436: this is really cool

kishansagathiya (approved):
:tada: 

Really nice!!

PR #2341, opened 2022-03-02
bug(dot/state): delete previous previous finalised trie from `BlockState.tries`

### LINKED ISSUE #2340: Delete previously finalised trie from `BlockState.tries`

timwu20:
## Changes

- Deletes the previous trie after a block is finalised from `BlockState.tries`.
- Significantly reduces memory usage over time
- Added a tries set/delete gauge
- Removed gauge mocks from `dot/state` tests.

## Tests

```
go test ./dot/state
```

## Issues

- closes #2340 

## Primary Reviewer

- @noot

### DIFF
--- a/dot/state/block_finalisation.go
+++ b/dot/state/block_finalisation.go
@@ -176,6 +176,17 @@ func (bs *BlockState) SetFinalisedHash(hash common.Hash, round, setID uint64) er
 		),
 	)
 
+	lastFinalisedHeader, err := bs.GetHeader(bs.lastFinalised)
+	if err != nil {
+		return fmt.Errorf("unable to retrieve header for last finalised block, hash: %s, err: %s", bs.lastFinalised, err)
+	}
+	stateRootTrie := bs.tries.get(lastFinalisedHeader.StateRoot)
+	if stateRootTrie != nil {
+		bs.tries.delete(lastFinalisedHeader.StateRoot)
+	} else {
+		return fmt.Errorf("unable to find trie with stateroot hash: %s", lastFinalisedHeader.StateRoot)
+	}
+
 	bs.lastFinalised = hash
 	return nil
 }
@@ -243,7 +254,6 @@ func (bs *BlockState) handleFinalisedBlock(curr common.Hash) error {
 
 		logger.Tracef("cleaned out finalised block from memory; block number %s with hash %s", block.Header.Number, hash)
 	}
-
 	return batch.Flush()
 }
 
--- a/dot/state/block_finalisation.go
+++ b/dot/state/block_finalisation.go
@@ -136,9 +136,9 @@ func (bs *BlockState) SetFinalisedHash(hash common.Hash, round, setID uint64) er
 		return fmt.Errorf("failed to set highest round and set ID: %w", err)
 	}
 
-	if err := bs.handleFinalisedBlock(hash); err != nil {
-		return fmt.Errorf("failed to set number->hash mapping on finalisation: %w", err)
-	}
+	// if err := bs.handleFinalisedBlock(hash); err != nil {
+	// 	return fmt.Errorf("failed to set number->hash mapping on finalisation: %w", err)
+	// }
 
 	if round > 0 {
 		bs.notifyFinalized(hash, round, setID)
@@ -176,15 +176,17 @@ func (bs *BlockState) SetFinalisedHash(hash common.Hash, round, setID uint64) er
 		),
 	)
 
-	lastFinalisedHeader, err := bs.GetHeader(bs.lastFinalised)
-	if err != nil {
-		return fmt.Errorf("unable to retrieve header for last finalised block, hash: %s, err: %s", bs.lastFinalised, err)
-	}
-	stateRootTrie := bs.tries.get(lastFinalisedHeader.StateRoot)
-	if stateRootTrie != nil {
-		bs.tries.delete(lastFinalisedHeader.StateRoot)
-	} else {
-		return fmt.Errorf("unable to find trie with stateroot hash: %s", lastFinalisedHeader.StateRoot)
+	if !bs.lastFinalised.Equal(hash) {
+		lastFinalisedHeader, err := bs.GetHeader(bs.lastFinalised)
+		if err != nil {
+			return fmt.Errorf("unable to retrieve header for last finalised block, hash: %s, err: %s", bs.lastFinalised, err)
+		}
+		stateRootTrie := bs.tries.get(lastFinalisedHeader.StateRoot)
+		if stateRootTrie != nil {
+			bs.tries.delete(lastFinalisedHeader.StateRoot)
+		} else {
+			return fmt.Errorf("unable to find trie with stateroot hash: %s", lastFinalisedHeader.StateRoot)
+		}
 	}
 
 	bs.lastFinalised = hash
--- a/dot/state/block_test.go
+++ b/dot/state/block_test.go
@@ -37,6 +37,12 @@ func newTestBlockState(t *testing.T, header *types.Header, tries *Tries) *BlockS
 
 	bs, err := NewBlockStateFromGenesis(db, tries, header, telemetryMock)
 	require.NoError(t, err)
+
+	tr := trie.NewEmptyTrie()
+	err = tr.Load(bs.db, header.StateRoot)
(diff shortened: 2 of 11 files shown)


dot/state/block_finalisation.go
```diff
@@ -176,6 +171,21 @@ func (bs *BlockState) SetFinalisedHash(hash common.Hash, round, setID uint64) er
 		),
 	)
 
+	if !bs.lastFinalised.Equal(hash) {
+		defer func(lastFinalised common.Hash) {
```
qdm12, line 175: Why defer? Can we just have

```go
previousLastFinalised := bs.lastFinalised
bs.lastFinalised = hash
if !previousLastFinalised.Equal(hash) {
    // defer function body, replacing lastFinalised with previousLastFinalised
}
return nil
```

dot/state/block_test.go
```diff
@@ -482,14 +479,7 @@ func TestAddBlockToBlockTree(t *testing.T) {
...
 
 func TestNumberIsFinalised(t *testing.T) {
-	ctrl := gomock.NewController(t)
-
-	triesGauge := NewMockGauge(ctrl)
-	triesGauge.EXPECT().Set(0.00).Times(2)
-	tries := &Tries{
-		rootToTrie: make(map[common.Hash]*trie.Trie),
-		triesGauge: triesGauge,
-	}
```
qdm12, line 492: Can you add your gauges to the previous test code instead of removing it? It should only take ~4 more lines for the 2 new gauges. We should test our metrics (at least new ones we add).

tests/stress/stress_test.go
```diff
@@ -118,6 +118,7 @@ func TestSync_SingleBlockProducer(t *testing.T) {
 			require.NoError(t, err, i)
 			continue
 		}
+		t.Log("hashes", hashes)
```
qdm12, line 121: remove?

```suggestion
```

qdm12 (changes requested):
Nice, although all the metrics testing removal hurts my soul, can you please adapt the test code to take care of the new metrics added?

Also please remove the tries gauge in favor of the new set and delete counter metrics

dot/state/block_finalisation.go
timwu20: I chose a defer, cause in normal operation the in memory tries should always contain the last finalised trie.  But if we return an error here then a lot of tests will fail outside the `dot/state` package.  I'd rather not just ignore the error since every other error in this function isn't ignored.

dot/state/block_test.go
timwu20: I don't see the need to test gauge mocks in every test that utilises the in-memory tries.  We should have a test to solely test metrics functionality.

dot/state/tries.go
```diff
@@ -4,40 +4,52 @@
...
 // to trie.
 type Tries struct {
-	rootToTrie map[common.Hash]*trie.Trie
-	mapMutex   sync.RWMutex
-	triesGauge prometheus.Gauge
+	rootToTrie    map[common.Hash]*trie.Trie
+	mapMutex      sync.RWMutex
+	triesGauge    prometheus.Gauge
+	setCounter    prometheus.Counter
+	deleteCounter prometheus.Counter
```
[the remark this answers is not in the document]
timwu20: they are different gauge collector types (gauge vs counter), so datadog and grafana will handle them differently.  I'd rather not do computation on the frontend for this metric.

dot/state/block_test.go
timwu20: I added the mock counters to `trie_test.go`.

dot/state/block_finalisation.go
qdm12: I didn't suggest to return an error, just have it without the defer function.

dot/state/tries_test.go
```diff
@@ -130,15 +133,20 @@ func Test_Tries_delete(t *testing.T) {
...
-				rootToTrie: testCase.rootToTrie,
-				triesGauge: triesGauge,
+				rootToTrie:    testCase.rootToTrie,
+				triesGauge:    triesGauge,
+				setCounter:    setCounter,
+				deleteCounter: deleteCounter,
 			}
 
+			tries.rootToTrie = testCase.rootToTrie
+
```
qdm12, line 149: it's already done above

```suggestion
```

dot/state/tries_test.go
```diff
@@ -130,15 +133,20 @@ func Test_Tries_delete(t *testing.T) {
 		t.Run(name, func(t *testing.T) {
 			t.Parallel()
 			ctrl := gomock.NewController(t)
-
-			triesGauge := NewMockGauge(ctrl)
-			triesGauge.EXPECT().Set(testCase.triesGaugeSet)
```
qdm12, line 135: please restore the triesGauge expectation here

dot/state/tries_test.go
```diff
@@ -27,18 +26,16 @@ func Test_NewTries(t *testing.T) {
...
+		setCounter:    setCounter,
+		deleteCounter: deleteCounter,
 	}
 
 	assert.Equal(t, expectedTries, rootToTrie)
 }
 
 //go:generate mockgen -destination=mock_gauge_test.go -package $GOPACKAGE github.com/prometheus/client_golang/prometheus Gauge
-
+//go:generate mockgen -destination=mock_counter_test.go -package $GOPACKAGE github.com/prometheus/client_golang/prometheus Counter
```
qdm12, line 38: nit adding empty line below generate command

```suggestion
//go:generate mockgen -destination=mock_counter_test.go -package $GOPACKAGE github.com/prometheus/client_golang/prometheus Counter

```

dot/state/block_test.go
qdm12: > I don't see the need to test gauge mocks in every test that utilises the in-memory tries. We should have a test to solely test metrics functionality.

I think we should mock everything that is called as long as it's in the same package, to have better testing depth / more passes on the same code. Although it doesn't test the metrics functionality directly, you can find other inconsistencies with this strategy. But up to you, juts sharing my 2 cents 😉

dot/state/block_test.go
```diff
@@ -37,6 +37,12 @@ func newTestBlockState(t *testing.T, header *types.Header, tries *Tries) *BlockS
 
 	bs, err := NewBlockStateFromGenesis(db, tries, header, telemetryMock)
 	require.NoError(t, err)
+
+	tr := trie.NewEmptyTrie()
+	err = tr.Load(bs.db, header.StateRoot)
+	require.NoError(t, err)
+	bs.tries.softSet(header.StateRoot, tr)
```
qdm12, line 44: Maybe add a comment for what this block does?

dot/state/block_finalisation.go
timwu20: yea.. I dislike ignoring errors, and I feel like cleaning up the in-memory tries is something that can be encompassed in a defer after the finalised block hash is updated.  If we end up refactoring some of this code, I think this should return an error since we should be ensuring the integrity of what's in-memory and what's on disk.

dot/state/block_test.go
timwu20: For each test that uses `Tries` for metric I would have to create a mock gauge or counter, and copy and paste the intended behaviour of the mock.  I think it's overkill.  IMO metric tests should be within the unit tests of the type that is publishing the metrics.

dot/state/tries_test.go
timwu20: whoops!

dot/state/block_test.go
qdm12: > overkill

Agreed! But a bit like those stress tests that helped us find the flags errors :eyes: Not their direct purpose, but it can indirectly fish out some unexpected behavior. Anyway resolving this!

dot/state/block_finalisation.go
qdm12: If you want to keep the defer statement, at least make it a separate function?
It's a bit confusing/ugly as it is now I think

dot/state/tries.go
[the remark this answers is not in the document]
qdm12: Alright :wink:

noot (approved):
nice!

qdm12 (approved):

PR #2342, opened 2022-03-02
feat(lib/trie): metrics interface with nodes gauge

qdm12:
## Issues

- #2229


qdm12:
Will re-open the PR later, this is based off 4 branches so it's rather far.

Issue #2357, opened 2022-03-08
Check for root in EncodeAndHash


kishansagathiya:
## Issue summary

- Looking at the comments and code, trie.Hash() and trie.root.EncodeAndHash() should produce the same hash, but they don't. condition of `len(key) < 32` is not checked in `trie.Hash` ~and it hashes everything~, because in case of root nodes we are supposed to return hash of encoding as hash regardless of length of encoding. This creates mismatching hashes from both functions.

- Right now the logic used in `node.EncodeAndHash() ` is repeating in `trie.Hash`. 
- ~`trie.Hash` should get hash from `trie.RootNode.EncodeAndHash() ` and return it~
- In `node.EncodeAndHash` we should check if this node is root or not. If the node is root we should return blake2b hash of its encoding regardless of length of its encoding.
- And then, trie.Hash should get hash from trie.RootNode.EncodeAndHash() and return it.

noot:
the root node should always be hashed, so I think the functionality of `trie.Hash` is correct

kishansagathiya:
> the root node should always be hashed when getting the trie hash, so I think the functionality of trie.Hash is correct. EncodeAndHash is for nodes that are not necessarily the root so it's not subject to the same rule

Ah, right. My bad!

I feel like we should check for root in `EncodeAndHash` and hash it. That should fix many problems. If we don't do this, we will have to do this separately every time we using EncodeAndHash.

kishansagathiya:
closed in https://github.com/ChainSafe/gossamer/pull/2359

PR #2359, opened 2022-03-09
fix(lib/trie): Check for root in EncodeAndHash

kishansagathiya:
## Changes

- In EncodeAndHash, If a node is root, we return blake2b hash of its encoding as hash
regardless of whether length of encoding is smaller or greater than 32

## Tests

```

```

## Issues

- #2357

## Primary Reviewer

- @qdm12


internal/trie/node/hash_test.go
```diff
@@ -119,7 +119,7 @@ func Test_Branch_EncodeAndHash(t *testing.T) {
 		t.Run(name, func(t *testing.T) {
 			t.Parallel()
 
-			encoding, hash, err := testCase.branch.EncodeAndHash()
+			encoding, hash, err := testCase.branch.EncodeAndHash(false)
```
qdm12, line 122: please add `isRoot` as a field to the test case. And also add a test case with `isRoot: true`.

internal/trie/node/hash_test.go
```diff
@@ -241,7 +241,7 @@ func Test_Leaf_EncodeAndHash(t *testing.T) {
 		t.Run(name, func(t *testing.T) {
 			t.Parallel()
 
-			encoding, hash, err := testCase.leaf.EncodeAndHash()
+			encoding, hash, err := testCase.leaf.EncodeAndHash(false)
```
qdm12, line 244: please add `isRoot` as a field to the test case. And also add a test case with `isRoot: true`.

lib/trie/lookup.go
```diff
@@ -18,24 +18,24 @@ type recorder interface {
 
 // findAndRecord search for a desired key recording all the nodes in the path including the desired node
 func findAndRecord(t *Trie, key []byte, recorder recorder) error {
-	return find(t.root, key, recorder)
+	return find(t.root, key, recorder, true)
```
qdm12, line 21: nit for extreme readability

```suggestion
	const isRoot = true
	return find(t.root, key, recorder, isRoot)
```

lib/trie/lookup.go
```diff
@@ -18,24 +18,24 @@ type recorder interface {
...
 }
 
-func find(parent Node, key []byte, recorder recorder) error {
-	enc, hash, err := parent.EncodeAndHash()
+func find(current Node, key []byte, recorder recorder, isCurrentRoot bool) error {
```
qdm12, line 24: why not leave it as `parent`? As in it's `parent` elsewhere mostly.
Although `current` works as well I guess.

lib/trie/lookup.go
```diff
@@ -49,5 +49,5 @@ func find(parent Node, key []byte, recorder recorder) error {
 		return nil
 	}
 
-	return find(b.Children[key[length]], key[length+1:], recorder)
+	return find(b.Children[key[length]], key[length+1:], recorder, false)
```
qdm12, line 52:
```suggestion
	const isChildRoot = false
	return find(b.Children[key[length]], key[length+1:], recorder, isChildRoot)
```

qdm12 (approved):
Nice, just a few nits.

Also, would it be possible to instead change the database store/fetch code to be able to understand smaller than 32 bytes values instead of hash for the root node? I guess not, just asking this out of curiosity.

lib/trie/lookup.go
kishansagathiya: parent seems rather misleading. parent suggest that there is some other node that is more interesting for this function

qdm12: We are finding a node which can either be the parent itself or one of its children. Why is it misleading?

In some function like `insert` we return a new node given an existing node and a node to insert; how should we call the returned node? I'm happy to do a batch rename if you have better naming. For now I was using `parent` as the argument and `newParent` as the returned node, but maybe `currentNode` and `newNode` would be better? :thinking:

kishansagathiya:
> Also, would it be possible to instead change the database store/fetch code to be able to understand smaller than 32 bytes values instead of hash for the root node?

This could be possible. I prefer the current approach however. I prefer function's behaviour to be obvious based on its signature. If I edit `key` inside say db.Put(key, value), that is not an obvious behaviour. A behaviour like that could lead to bugs!

qdm12:
>  If I edit key inside say db.Put(key, value), that is not an obvious behaviour. A behaviour like that could lead to bugs!

Definitely. But can't we write keys as byte slices (not fixed size arrays) to database? That way there might not be any custom code for the root, and we can know if it's a node value or the node encoding hash digest depending on the database retrieved byte slice length? :thinking: Maybe I'm wrong though..

lib/trie/lookup.go
EclesioMeloJunior: the `isChildRoot` variable does not make much sense, the root node already has been processed and the other nodes are not the root, so it is explicit that for other `find` calls the `false` argument means the nodes are not the root

EclesioMeloJunior (approved):

lib/trie/database.go
```diff
@@ -97,7 +97,7 @@ func (t *Trie) loadFromProof(rawProof [][]byte, rootHash []byte) error {
 		decodedNode.SetDirty(dirty)
 		decodedNode.SetEncodingAndHash(rawNode, nil)
 
-		_, hash, err := decodedNode.EncodeAndHash()
+		_, hash, err := decodedNode.EncodeAndHash(decodedNode == t.root)
```
noot, line 100: `decodedNode == t.root` is always false here (which is fine, since it's the same functionality as before)

noot: but perhaps just replace the case with `false`?

lib/trie/lookup.go
kishansagathiya: changed back to parent

PR #2366, opened 2022-03-10
chore(dot/network): replace `sync.Map` with map for stream manager

qdm12:
## Tests

All tests passing, no test added.

## Issues

- #2281

## Primary Reviewer

- @EclesioMeloJunior


dot/network/stream_manager.go
```diff
@@ -23,14 +23,15 @@ type streamData struct {
 // This prevents keeping stale streams open and continuously trying to
 // read from it, which takes up lots of CPU over time.
 type streamManager struct {
-	ctx           context.Context
-	streamDataMap *sync.Map //map[string]*streamData
+	ctx          context.Context
+	streamDataMu sync.RWMutex
```
noot, line 27: should it just be `sync.Mutex` since `RLock` isn't used?

noot (approved):

dot/network/stream_manager.go
qdm12: Totally, done! :+1:

dot/network/stream_manager.go
```diff
@@ -51,18 +52,18 @@ func (sm *streamManager) start() {
...
-		stream := sdata.stream
+	sm.streamDataMu.Lock()
+	defer sm.streamDataMu.Unlock()
+
+	for id, data := range sm.streamData {
+		lastReceived := data.lastReceivedMessage
+		stream := data.stream
 
 		if time.Since(lastReceived) > cleanupStreamInterval {
 			_ = stream.Close()
```
EclesioMeloJunior, line 63: as `stream.Close` returns an error, can we log this error
```suggestion
			err := stream.Close()
			if err != nil {
			    logger.Wanrf("failed to close stream: %w", err)
			}
```

EclesioMeloJunior (approved):
LGMT, just a comment

dot/network/stream_manager.go
qdm12: Out of scope in this PR (also, not going to lie, all the CI passed :smile:)

On the other hand, this close error might happen regularly too so we may not want to log it (at least no as info/warn/error). Anyway, sorry I'll ignore this and mergggge but I would had totally took a look if the CI would had failed or would be more stable :smile:

Issue #2345, opened 2022-03-03
Investigate Finality Issues on Cross-Client Devnets


danforbes:
When Gossamer is run with Substrate nodes in a cross-client devnet, GRANDPA finality stops after a few rounds. This ticket is being created to investigate the cause of this problem and possibly fix it (depending on complexity).

jimjbrettj:
stale issue

PR #2368, opened 2022-03-10
fix(lib/grandpa): various finality fixes, improves cross-client finality

noot:
## Changes

- add ticker to `messageTracker` to allow for regular processing of cached vote/commit messages, I noticed that some messages were being put into the tracker *after* the block it was for was already imported, so the message was essentially lost in memory forever
- remove check for if we have vote already in `validateVoteMessage`, if it's a duplicate it doesn't matter since it's stored in a map anyways. (this seems weird, but I noticed sometimes it would error with `errVoteExists` but not seem to have the vote :/)

## Tests

```
go test ./lib/grandpa
```

## Issues

- related to #2345 

## Primary Reviewer

- @EclesioMeloJunior


lib/grandpa/message_tracker.go
```diff
@@ -122,3 +127,32 @@ func (t *tracker) handleBlock(b *types.Block) {
...
+	for _, vms := range t.voteMessages {
+		for _, v := range vms {
+			// handleMessage would never error for vote message
+			_, err := t.handler.handleMessage(v.from, v.msg)
+			if err != nil {
+				logger.Debugf("failed to handle vote message %v: %s", v, err)
+			}
+
+			if v.msg.Round < t.handler.grandpa.state.round && v.msg.SetID == t.handler.grandpa.state.setID {
+				delete(t.voteMessages, v.msg.Message.Hash)
```
EclesioMeloJunior, line 144: I think it is not a good idea to modify the map that is being iterated, we can create a slice with hashes to delete and remove them right after the iteration finishes

EclesioMeloJunior (approved):

EclesioMeloJunior (approved):

lib/grandpa/message_tracker.go
EclesioMeloJunior, line 144: I think it is not a good idea to modify the map that is being iterated, we can create a slice with hashes to delete and remove them right after the iteration finishes

lib/grandpa/message_tracker.go
```diff
@@ -122,3 +127,32 @@ func (t *tracker) handleBlock(b *types.Block) {
...
+	}
+
+	for _, cm := range t.commitMessages {
+		_, err := t.handler.handleMessage("", cm)
+		if err != nil {
+			logger.Debugf("failed to handle commit message %v: %s", cm, err)
+			continue
+		}
+
+		delete(t.commitMessages, cm.Vote.Hash)
```
EclesioMeloJunior, line 156: Same here

EclesioMeloJunior (changes requested):

lib/grandpa/message_tracker_test.go
```diff
@@ -180,3 +181,50 @@ func TestMessageTracker_MapInsideMap(t *testing.T) {
...
+	msg := &VoteMessage{
+		Round: 100,
+	}
+	gs.tracker.addVote(&networkVoteMessage{
+		msg: msg,
+	})
+
+	gs.tracker.handleTick()
+
+	select {
```
qdm12, line 201: nit define it inlined

```suggestion
	const testTimeout = time.Second
	select {
```

lib/grandpa/message_tracker_test.go
```diff
@@ -180,3 +181,50 @@ func TestMessageTracker_MapInsideMap(t *testing.T) {
...
+	case v := <-in:
+		require.Equal(t, msg, v.msg)
+	case <-time.After(testTimeout):
+		t.Errorf("did not receive vote message")
```
qdm12, line 225: nit

```suggestion
		t.Errorf("did not receive vote message %v", msg)
```

lib/grandpa/message_tracker_test.go
```diff
@@ -180,3 +181,50 @@ func TestMessageTracker_MapInsideMap(t *testing.T) {
...
```
qdm12, line 205: nit

```suggestion
		t.Errorf("did not receive vote message %v", msg)
```

lib/grandpa/message_tracker_test.go
```diff
@@ -180,3 +181,50 @@ func TestMessageTracker_MapInsideMap(t *testing.T) {
...
+	}
+
+	// should be deleted as round in message < grandpa round
+	require.Equal(t, 0, len(gs.tracker.voteMessages[common.Hash{}]))
```
qdm12, line 229: nit

```suggestion
	require.Empty(t, gs.tracker.voteMessages[common.Hash{}])
```

lib/grandpa/message_tracker_test.go
```diff
@@ -180,3 +181,50 @@ func TestMessageTracker_MapInsideMap(t *testing.T) {
...
+
+	select {
+	case v := <-in:
+		require.Equal(t, msg, v.msg)
+	case <-time.After(testTimeout):
+		t.Errorf("did not receive vote message")
+	}
+
+	// shouldn't be deleted as round in message >= grandpa round
+	require.Equal(t, 1, len(gs.tracker.voteMessages[common.Hash{}]))
```
qdm12, line 209: nit can we use a non-empty `common.Hash` as the key to make sure it's passed/set correctly in the production code? Same for below.

lib/grandpa/vote_message.go
```diff
@@ -230,6 +217,7 @@ func (s *Service) validateVoteMessage(from peer.ID, m *VoteMessage) (*Vote, erro
 		s.prevotes.Store(pk.AsBytes(), just)
 	case precommit:
 		s.precommits.Store(pk.AsBytes(), just)
+		logger.Debugf("stored precommit: num=%d", s.lenVotes(precommit))
```
qdm12, line 220: nit why not more human string like

```suggestion
		logger.Debugf("stored %d precommit votes", s.lenVotes(precommit))
```

qdm12 (approved):
Nice, just a few *nits*

lib/grandpa/message_tracker.go
noot: @EclesioMeloJunior @qdm12 what's the vibe on this? should I update it to delete all the keys after the loop? wouldn't that require another for loop to delete them all though?

lib/grandpa/message_tracker.go
```diff
@@ -122,3 +127,32 @@ func (t *tracker) handleBlock(b *types.Block) {
...
+func (t *tracker) handleTick() {
+	t.mapLock.Lock()
+	defer t.mapLock.Unlock()
+
+	for _, vms := range t.voteMessages {
+		for _, v := range vms {
+			// handleMessage would never error for vote message
+			_, err := t.handler.handleMessage(v.from, v.msg)
+			if err != nil {
+				logger.Debugf("failed to handle vote message %v: %s", v, err)
```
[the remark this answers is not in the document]
noot: honestly, this will never return an error, since `t.handler.handleMessage` for `VoteMessage` never errors. should I just get rid of this check completely?

lib/grandpa/message_tracker.go
```diff
@@ -82,6 +83,8 @@ func (t *tracker) addCatchUpResponse(cr *CatchUpResponse) {
 }
 
 func (t *tracker) handleBlocks() {
+	ticker := time.NewTicker(time.Second)
```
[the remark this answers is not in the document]
noot: updated

lib/grandpa/message_tracker_test.go
noot: updated

lib/grandpa/message_tracker_test.go
noot: updated

lib/grandpa/message_tracker.go
EclesioMeloJunior: @qdm12 thanks for sharing this post, @noot it should be ok to keep it as it is.

lib/grandpa/vote_message.go
```diff
@@ -197,7 +184,7 @@ func (s *Service) validateVoteMessage(from peer.ID, m *VoteMessage) (*Vote, erro
 	// if the vote is from ourselves, ignore
```
EclesioMeloJunior, line 184:
```suggestion
	// if the vote is from ourselves, return an error
```

EclesioMeloJunior (approved):
lgtm!! just a super nit

lib/grandpa/message_tracker.go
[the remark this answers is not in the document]
qdm12: Wellll, for now it never errors, but one could change `handleMessage` and that could blow up in our face :smile: :bomb: 

Maybe just `panic(err)`?

lib/grandpa/message_tracker_test.go
```diff
@@ -189,28 +190,36 @@ func TestMessageTracker_handleTick(t *testing.T) {
 	gs, in, _, _ := setupGrandpa(t, kr.Bob().(*ed25519.Keypair))
 	gs.tracker = newTracker(gs.blockState, gs.messageHandler)
 
+	var testHash common.Hash = [32]byte{1, 2, 3}
```
qdm12, line 193: nit

```suggestion
	testHash := common.Hash{1, 2, 3}
```

qdm12 (approved):

lib/grandpa/message_tracker.go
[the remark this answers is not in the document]
timwu20: I would suggest a `logger.Critical` for now.  After looking at grandpa, it looks like the message is just popped onto the `in` channel and read later via `Service.receiveVoteMessages` where any errors would be logged at debug level.  It would be nice to be able track errors to specific messages that flow through these pipelines.  We could probably track them at the moment with critical log errors.

lib/grandpa/message_tracker.go
```diff
@@ -122,3 +129,32 @@ func (t *tracker) handleBlock(b *types.Block) {
...
+				logger.Debugf("failed to handle vote message %v: %s", v, err)
+			}
+
+			if v.msg.Round < t.handler.grandpa.state.round && v.msg.SetID == t.handler.grandpa.state.setID {
+				delete(t.voteMessages, v.msg.Message.Hash)
+			}
+		}
+	}
+
+	for _, cm := range t.commitMessages {
```
timwu20, line 151: could we split these loops to run in parallel?

qdm12: Why? Does each iteration take a significant amount of time? :thinking:

Issue #2329, opened 2022-02-24
Paritytech/trie proofs are incompatible with gossamer/lib/trie


seunlanlege:
## Describe the bug
Read proofs generated by the rust implementation of the substrate node don't seem to work with the `lib/trie` library.

<img width="1726" alt="Screenshot 2022-02-24 at 11 33 20 AM" src="https://user-images.githubusercontent.com/16494314/155508502-c93c23ad-66c1-4a73-86aa-749d683c6599.png">
<img width="1728" alt="Screenshot 2022-02-24 at 11 35 54 AM" src="https://user-images.githubusercontent.com/16494314/155508517-bb6faff6-1232-4fd5-9d82-d061a7a13fbb.png">
 
```golang

func TestTrieProof(t *testing.T) {
	key, err := hex.DecodeString("f0c365c3cf59d671eb72da0e7a4113c49f1f0515f462cdcf84e0f1d6045dfcbb")
	if err != nil {
		panic(err)
	}
	root, err := hex.DecodeString("dc4887669c2a6b3462e9557aa3105a66a02b6ec3b21784613de78c95dc3cbbe0")
	if err != nil {
		panic(err)
	}
[...]

PR #2369, opened 2022-03-11
fix(trie): decode inline child nodes

notbdu:
## Changes

- Adds handler for decoding inline child nodes based on expected hash length. 
- Achieves behavioral parity with the parity implementation here: https://github.com/paritytech/substrate/blob/542d587c3db7e46b6fff423e8efc4f327f7b79a1/primitives/trie/src/node_codec.rs#L141-L145 

## Tests

- Adds a test for decoding a branch with inlined child nodes.

## Issues
- Fixes proof issue: https://github.com/ChainSafe/gossamer/issues/2329

## Primary Reviewer

-


internal/trie/node/decode.go
```diff
@@ -105,6 +105,17 @@ func decodeBranch(reader io.Reader, header byte) (branch *Branch, err error) {
 				ErrDecodeChildHash, i, err)
 		}
 
+		// Handle inlined leaf nodes.
+		if Type(hash[0]>>6) == LeafType && len(hash) != HashLength {
```
qdm12, line 109: nit it would look better to have this hash length as an inlined constant here, since it's not really used elsewhere :wink: 

```suggestion
		const hashLength = 32
		if Type(hash[0]>>6) == LeafType && len(hash) < hashLength {
```

internal/trie/node/decode.go
```diff
@@ -105,6 +105,17 @@ func decodeBranch(reader io.Reader, header byte) (branch *Branch, err error) {
...
+			leaf, err := decodeLeaf(bytes.NewReader(hash[1:]), hash[0])
+			if err != nil {
+				return nil, fmt.Errorf("%w: at index %d: %s",
+					ErrDecodeValue, i, err)
+			}
+			branch.Children[i] = leaf
+			continue
+		}
+
```
qdm12, line 118: Quick question, I guess this inline format cannot be used for branches right? Only for leaves?

internal/trie/node/encode_decode_test.go
```diff
@@ -57,7 +57,9 @@ func Test_Branch_Encode_Decode(t *testing.T) {
 				Key: []byte{5},
 				Children: [16]Node{
 					&Leaf{
-						HashDigest: []byte{0x41, 0x9, 0x4, 0xa},
+						Key:   []byte{9},
+						Value: []byte{10},
+						Dirty: true,
```
qdm12, line 62: Can you instead increase that hash digest to be 32 bytes, and add a test case for the inlined leaf in this test?

internal/trie/node/types.go
```diff
@@ -17,3 +17,6 @@ const (
 	// InvalidType is used in tests only
 	InvalidType
 )
+
+// HashLength is 256 bits
+const HashLength = 32
```
qdm12, line 22: nit remove in favor of inlined constant.

internal/trie/node/decode_test.go
```diff
@@ -95,6 +95,42 @@ func Test_Decode(t *testing.T) {
...
+				Children: [16]Node{
+					nil, nil, nil, nil,
+					&Leaf{
+						Key:   []byte{14, 7, 11, 9, 0, 1, 2, 0, 9, 6, 11, 4, 1, 12, 4, 14, 11, 3, 10, 10, 15, 9, 4, 7, 15, 6, 14, 10, 4, 2, 9},
+						Value: []byte{0, 0},
+						Dirty: true,
+					},
+					nil, nil, nil, nil,
+					&Leaf{
+						Key:   []byte{15, 1, 15, 0, 5, 1, 5, 15, 4, 6, 2, 12, 13, 12, 15, 8, 4, 14, 0, 15, 1, 13, 6, 0, 4, 5, 13, 15, 12, 11, 11},
```
qdm12, line 125: You may have to split those lines to satisfy the line length linter (`lll`) we have in our CI

internal/trie/node/decode.go
notbdu: Ah so a few things here I think. Technically there _could_ be, my impression of the logic is that any child node < 32 bytes can be in-lined leaf or branch. 

The child parsing seems to be wrong in general? It looks like its currently parsing every child as a leaf node. Although I'm not sure it semantically makes a difference when its just a hash pointer. 

Also, there are a few header types that are missing from the impl:
```
impl Decode for NodeHeader {
    fn decode<I: Input>(input: &mut I) -> Result<Self, codec::Error> {
        let i = input.read_byte()?;
        if i == trie_constants::EMPTY_TRIE {
            return Ok(NodeHeader::Null);
        }
        println!("header byte {:b} {:b}", i, i & (0b11 << 6));
        match i & (0b11 << 6) {
            trie_constants::LEAF_PREFIX_MASK => Ok(NodeHeader::Leaf(decode_size(i, input, 2)?)),
            trie_constants::BRANCH_WITH_MASK => {
                Ok(NodeHeader::Branch(true, decode_size(i, input, 2)?))
            }
            trie_constants::BRANCH_WITHOUT_MASK => {
                Ok(NodeHeader::Branch(false, decode_size(i, input, 2)?))
            }
            trie_constants::EMPTY_TRIE => {
                if i & (0b111 << 5) == trie_constants::ALT_HASHING_LEAF_PREFIX_MASK {
                    Ok(NodeHeader::HashedValueLeaf(decode_size(i, input, 3)?))
                } else if i & (0b1111 << 4) == trie_constants::ALT_HASHING_BRANCH_WITH_MASK {
                    Ok(NodeHeader::HashedValueBranch(decode_size(i, input, 4)?))
                } else {
                    // do not allow any special encoding
                    Err("Unallowed encoding".into())
                }
            }
            _ => unreachable!(),
        }
    }
}
```

The mask size for hashed leaves and branches are different but after digging through the scale decoder implementation it looks like it handles that internally?

I think it's probably worth looking into some of the other issues but outside the scope of this PR? Maybe just have some open issues for now?

danforbes:
Hi, @notbdu - thank you so much for your contribution! I would love to learn a bit more about your interest in Gossamer 🤓 If you'd like to chat, can you please shoot me an email @ dan.forbes@chainsafe.io?

notbdu:
Sent an email!

Ah, I've also broken the integration tests :smiling_face_with_tear: .
```
   babe_integration_test.go:119:
               Error Trace:    babe_integration_test.go:119
                                                       babe_integration_test.go:231
               Error:          Received unexpected error:
                               failed to load storage trie from database: cannot load child at index 2 with hash 0x73c356dac7914c2574ed8cacd4e65147bf93ed93b56cce9792c2e2fb846c3941: cannot find child node key 0x in database: Key not found
               Test:           TestService_ProducesBlocks
-- FAIL: TestService_ProducesBlocks (1.38s)
```

I think the test is expecting all decoded children to be hash pointers but now it's loading the child inline.

internal/trie/node/decode.go
qdm12: > It looks like its currently parsing every child as a leaf node. Although I'm not sure it semantically makes a difference when its just a hash pointer.

As the `decodeBranch` comment mentions:

```go
// Note that since the encoded branch stores the hash of the children nodes, we are not
// reconstructing the child nodes from the encoding. This function instead stubs where the
// children are known to be with an empty leaf. The children nodes hashes are then used to
// find other values using the persistent database.
```

In another PR, it would be interesting to refactor this to load from database as it decodes, especially since you can have in-lined encoded nodes, otherwise it's rather confusing. Created #2375.

> there are a few header types that are missing from the impl

That might be the case, I opened #2374, thanks so much for your investigation!

> The mask size for hashed leaves and branches are different but after digging through the scale decoder implementation it looks like it handles that internally?

Are you sure? Can you point to some code you've seen? @timwu20 maybe do you have a clue on this?

> I think it's probably worth looking into some of the other issues but outside the scope of this PR? Maybe just have some open issues for now?

Totally :+1:

noot:
@notbdu fixed the integration tests, but can't push to your branch, feel free to add this to `*Trie.Load` in `trie/database.go` line 187:
```
		hash := child.GetHash()
		_, isLeaf := child.(*node.Leaf)
		if len(hash) == 0 && isLeaf {
			// node has already been loaded inline
			continue
		}
```

lib/trie/database.go
```diff
@@ -185,6 +185,11 @@ func (t *Trie) load(db chaindb.Database, n Node) error {
 		}
 
 		hash := child.GetHash()
+		_, isLeaf := child.(*node.Leaf)
+		if len(hash) == 0 && isLeaf {
+			// node has already been loaded inline
+			continue
+		}
```
noot, line 192: nit but please add space after this line

notbdu:
Thanks for the reviews @noot, @qdm12 and team!

notbdu:
Added same fix to `getFromDB()` as well to fix `trie/database` unit tests.

internal/trie/node/encode_decode_test.go
```diff
@@ -83,6 +127,7 @@ func Test_Branch_Encode_Decode(t *testing.T) {
 			resultBranch, err := decodeBranch(buffer, header)
 			require.NoError(t, err)
 
+			log.Println(resultBranch)
```
qdm12, line 130: remove

```suggestion
```

qdm12 (approved):
Looks good although there is some investigation to be done about missing types and inline encoding/decoding of branches.

qdm12:
@notbdu I think there are linting issues, can you please fix it so we can merge this? Thanks!

lib/trie/database.go
```diff
@@ -432,14 +444,14 @@ func (t *Trie) writeDirty(db chaindb.Batch, n Node) error {
 // We need to compute the hash values of each newly inserted node.
 func (t *Trie) GetInsertedNodeHashes() (hashesSet map[common.Hash]struct{}, err error) {
 	hashesSet = make(map[common.Hash]struct{})
-	err = t.getInsertedNodeHashes(t.root, hashesSet)
+	err = t._getInsertedNodeHashes(t.root, hashesSet)
```
qdm12, line 447: Why the name change? Please revert

qdm12 (changes requested):

notbdu:
Deepsource go was complaining of this issue (link from another repo): https://deepsource.io/gh/xiaods/k8e/issue/RVV-B0001/occurrences/

The private method is the lower cased version of the public method.

lib/trie/database.go
notbdu: Commented below on RVV-B0001 linter issue.

qdm12 (commented):
@notbdu Oh sorry for the confusion; I was talking about the required linter workflow (golangci-lint named as `checks / linting (pull_request)`), we don't really use Deepsource and it's not required either.

You have to `git rebase development` / `git merge development`, we made a small CI workflow requirement change so you need new changes :pray: Thanks!!

notbdu:
@qdm12 sick sick, just rebased!

qdm12:
Oh-oh there is some decoding/encoding issue it seems: https://github.com/ChainSafe/gossamer/runs/5681474712?check_suite_focus=true

notbdu:
Ah interesting, TAL now - remember it passing before :thinking: ... haha

notbdu:
Let me know if this fix maxes sense to you. The already inline loaded nodes were missing hash digest and encoding so I'm just setting them here: https://github.com/ChainSafe/gossamer/pull/2369/commits/c94e83cea2325b285caf47f07504d996add31195

kishansagathiya (approved):
thanks @notbdu and @seunlanlege

Great to have contributions like these!

kishansagathiya:
@danforbes Is this something that should be in the spec? I did a quick search for `inline` in the spec pdf and did not find anything!

internal/trie/node/decode.go
```diff
@@ -105,6 +105,18 @@ func decodeBranch(reader io.Reader, header byte) (branch *Branch, err error) {
 				ErrDecodeChildHash, i, err)
 		}
 
+		// Handle inlined leaf nodes.
+		const hashLength = 32
+		if Type(hash[0]>>6) == LeafType && len(hash) != hashLength {
+			leaf, err := decodeLeaf(bytes.NewReader(hash[1:]), hash[0])
+			if err != nil {
+				return nil, fmt.Errorf("%w: at index %d: %s",
```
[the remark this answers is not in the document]
kishansagathiya: > So we can keep it as is (and elsewhere), or I can create an issue to change each of these 'inversed wrapping'.

I will leave that up to you! resolving this!

internal/trie/node/decode.go
```diff
@@ -105,6 +105,18 @@ func decodeBranch(reader io.Reader, header byte) (branch *Branch, err error) {
 				ErrDecodeChildHash, i, err)
 		}
 
+		// Handle inlined leaf nodes.
+		const hashLength = 32
+		if Type(hash[0]>>6) == LeafType && len(hash) != hashLength {
```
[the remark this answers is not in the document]
notbdu: Just FYI, parity's impl is an equality check and it doesn't seem like they have any check if `count > H::LENGTH` :thinking: .
https://github.com/paritytech/substrate/blob/542d587c3db7e46b6fff423e8efc4f327f7b79a1/primitives/trie/src/node_codec.rs#L141-L145
```
						children[i] = Some(if count == H::LENGTH {
							NodeHandlePlan::Hash(range)
						} else {
							NodeHandlePlan::Inline(range)
						});
```

But yea I'm under the impression that inline nodes should be < 32 bytes. Modified the check :+1:.

danforbes:
@kishansagathiya this is the relevant part of the spec (thank you, @FlorianFranzen 🙏🏻) https://w3f.github.io/polkadot-spec/develop/sect-state-storage.html#defn-node-header

qdm12:
To add to @danforbes's comment, #2418 was created to address this.

Issue #2326, opened 2022-02-22
Turn Off Flaky C/I Integration Tests & Design Plan to Improve/Replace Them


danforbes:
We have some C/I tasks that are flaky (e.g. `docker-rpc` and `docker-stress`, I think). These tests oftentimes fail because the test infrastructure itself is not robust and deterministic - the tests are not failing because there are bugs in actual Gossamer production code, necessarily. This is bad because it wastes developer time, and also we can have very little confidence that these tests are actually checking anything meaningful if the test infrastructure itself is not correct.

This Issue can be closed when there is a plan in place to improve/replace the flaky tests (an Issue(s) should be created to track this) and the flaky tests have been turned off in C/I.

qdm12:
The following comment will be updated as I progress

A few simple refactoring that looks needed for now:

- Timeouts are a defined in each function locally, making it hard to define timeouts from the actual test. We should instead use contexts and injected timeout values (as arguments) for operations.
- All operations should have retry mechanisms retrying until the context is canceled or times out
- RPC operations should have a client function for each 'method', instead of passing method constant strings

More refactoring that can be done:

- Run gossamer using `go run` instead of pre-building it and then running it. That shouldn't impact performance since Go will not rebuild it on the second time, and should remove the need for pre-tests setups
- dependency inject loggers (#1946) so that logs can be inspected from the tests using a io.Writer interface such as as a buffer.

qdm12:
Issues were created and reference this issue.

Since it was rather hard to plan ahead how to fix the messy-ish test code, work for them is mostly done already.

The issues created address refactoring of the test code, and this should allow us to have more robust tests with easily adjustable timeout values.

Closing this since there is a plan in the forms of issues. More issues may be created later if it turns out the refactoring + timeout adjustments is not sufficient.

PR #2370, opened 2022-03-11
chore(tests/stress): refactor `compareBlocksByNumber`

qdm12:
## Changes

*First in a serie of `tests/` refactor PRs*
The end goal is to move timeouts responsibility to the test instead of hidden in helpers.

- Takes in context
- Retry as long as context is not canceled
- Assert errors and common values

## Tests

```
make it-stress
```

## Issues

- #2326 

## Primary Reviewer

- Anyone it's just a single function changed.

### DIFF
--- a/tests/stress/helpers.go
+++ b/tests/stress/helpers.go
@@ -4,9 +4,9 @@
 package stress
 
 import (
+	"context"
 	"errors"
 	"fmt"
-	"sync"
 	"testing"
 	"time"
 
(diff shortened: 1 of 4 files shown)


noot (approved):

edwardmack (approved):
Great work!

tests/stress/helpers.go
```diff
@@ -65,74 +65,60 @@ func compareChainHeadsWithRetry(t *testing.T, nodes []*utils.Node) error {
...
 // it returns a map of block hashes to node key names, and an error if the hashes don't all match
-func compareBlocksByNumber(t *testing.T, nodes []*utils.Node, num string) (map[common.Hash][]string, error) {
-	hashes := make(map[common.Hash][]string)
-	var errs []error
-	var mapMu sync.Mutex
-	var wg sync.WaitGroup
-	wg.Add(len(nodes))
+func compareBlocksByNumber(ctx context.Context, t *testing.T, nodes []*utils.Node,
+	num string) (hashToKeys map[common.Hash][]string) {
+	type resultContainer struct {
```
timwu20, line 70: nit:
```suggestion
	type result struct {
```

tests/stress/stress_test.go
```diff
@@ -111,14 +112,18 @@ func TestSync_SingleBlockProducer(t *testing.T) {
...
 		time.Sleep(3 * time.Second)
 		t.Log("comparing...", i)
-		hashes, err := compareBlocksByNumberWithRetry(t, nodes, strconv.Itoa(i))
-		if len(hashes) > 1 || len(hashes) == 0 {
-			require.NoError(t, err, i)
-			continue
-		}
+
+		const comparisonTimeout = 5 * time.Second
+		compareCtx, cancel := context.WithTimeout(ctx, comparisonTimeout)
```
timwu20, line 122:
```suggestion
		const comparisonTimeout = 5 * time.Second
		compareCtx, cancel := context.WithTimeout(ctx, 5 * time.Second)
```
why not inline?

tests/stress/helpers.go
qdm12: If I rename to `result` `govet` complains about shadowing with the `result`-named variable below.
And I know I could rename that variable below from `result` to `r`, but that defeats my crusade against single letter variable names :crossed_swords: (except `i` for indexes). If you have a better name, feel free to say it :smile: 

Also that's an inlined type so it won't pollute the API so it's not a big deal either.

tests/stress/stress_test.go
qdm12: Because I got tortured by [`gomnd`](https://github.com/tommy-muehle/go-mnd) for years... So now I constant most of my various 'constant' numbers.

PR #2382, opened 2022-03-16
feat(metrics): replace metrics port with address

qdm12:
I needed to listen on all interfaces, and was surprise how complicated it was to find out I had to adjust the RPC host for this. This simplifies is and allows the user to specify the complete listening address instead of just the port.

## Changes

- Replace toml `metrics-port` with `metrics-address`
- Replace flag `--metrics-port` with `--metrics-address`
- Defaults to `localhost:9876` for all configurations except `dev` defaulting to `:9876` (all interfaces)
- Update documentation
- Remove a log line that is duplicated in the HTTP server logs.

## Tests

Existing tests

## Issues

Needed for the Dockerized grafana+prometheus local setup I am working on, to show graphs for next week's trie presentation

## Primary Reviewer

- Anyone


chain/gssmr/defaults.go
```diff
@@ -23,8 +23,8 @@ var (
 	// DefaultBasePath Default node base directory path
 	DefaultBasePath = string("~/.gossamer/gssmr")
 
-	// DefaultMetricsPort is the metrics server port
-	DefaultMetricsPort = uint32(9876)
+	// DefaultMetricsAddress is the default metrics server listening address.
+	DefaultMetricsAddress = "localhost:9876"
```
noot, line 27: potato question but this doesn't need the `http://`?

noot (approved):

chain/gssmr/defaults.go
qdm12: :potato: no, this is just the `host:port` for the tcp net listener, [go source code relevant line](https://cs.opensource.google/go/go/+/refs/tags/go1.18:src/net/http/server.go;l=2964) :wink: 

Although you can also use `localhost:http` for example, and Go will understand it as `localhost:80`, but that's another thing.

jimjbrettj:
out of curiosity, why default to localhost:9876?

jimjbrettj (approved):

cmd/gossamer/README.md
```diff
@@ -188,7 +188,7 @@ defined in [lib/runtime/wasmer/exports.go](../../lib/runtime/wasmer/exports.go).
 Gossamer publishes telemetry data and also includes an embedded Prometheus server that reports metrics. The metrics
 capabilities are defined in the [dot/metrics](../../dot/metrics) package and build on
 [the metrics library that is included with Go Ethereum](https://github.com/ethereum/go-ethereum/blob/master/metrics/README.md).
-The default port for Prometheus metrics is 9090, and Gossamer allows the user to configure this parameter with the
-`--metrics-port` command-line parameter. The Gossamer telemetry server publishes telemetry data that is compatible with
+The default listening address for Prometheus metrics is `localhost:9090`, and Gossamer allows the user to configure this parameter with the
```
qdm12, line 191: Actually this is false right @danforbes ? It's `9876` not `9090` I think?

qdm12:
@jimjbrettj the default to `localhost:9876` is such that it can only be accessed on the same host, for security purposes. So if you run the Gossamer binary, it would only be accessible by the same machine, and not other machines on the same network. One can set it to `:9876` (=accept from all IPv4+IPv6) to listen on all network interfaces.

Also thanks for your question, you made me realize I forgot to change Dockerfiles' commands to use `--metrics-address=":9876"`, since `localhost` for a Docker container is the container itself which is kind of pointless to listen on (other containers won't be able to access it, and even port mapping won't work for the Prometheus server). This is now fixed :+1:

timwu20 (approved):
should we do the same for rpc? We could probably get rid of `--rpc-external` option.

qdm12:
> should we do the same for rpc? We could probably get rid of `--rpc-external` option.

Definitely, it's better to let the user pick the interface they want to listen on. Some exotic users have several network interfaces :smiley_cat: 

Maybe for WS as well?

danforbes:
It may be constructive to review the configuration parameters that Substrate exposes to help get some ideas.

PR #2383, opened 2022-03-16
chore(dot/network): gossip sync.Map to std map

qdm12:
## Changes

:warning: :question: Shouldn't we clear the hashes from the gossip map at some point? That would be a memory leak in the long run.

## Tests

```
go test dot/network
```

## Issues

- #2281 

## Primary Reviewer

Anyone :wink:


noot (approved):
yeah, we should definetely clear the messages out after some time period

qdm12:
> we should definetely clear the messages out after some time period

Should we clear them after having N messages? So it's more 'event-driven'? Or do you see a reason to have time based? :thinking:

qdm12:
#2398 tracks this memory leak, we can merge and then address 2398.

jimjbrettj (approved):

timwu20 (approved):

Issue #2351, opened 2022-03-05
lib/trie: copy node and update its generation only when needed


qdm12:
## Task summary

Right now, when the trie is a new snapshotted trie (forked tries or after a finalized trie), we deep copy nodes and update their generation whenever we are in a path where modification is possible. This works but implies potentially a lot more memory operations (deep copying and garbage collection) when no modification is actually done.

Instead we should deep copy nodes and update their generation only when modification is certain to be done.

PR #2384, opened 2022-03-16
feat(trie): finer deep copy of nodes

qdm12:
## Changes

- Do not copy cached fields when not needed
- Do not copy key field when not needed
- Do not copy value field when not needed

## Tests

```
go test -race ./internal/trie/... ./lib/trie/...
```

## Issues

- #2351 - already done but this is finer to save a few extra operations

## Primary Reviewer

- Anyone :tada:

### DIFF
--- a/internal/trie/node/copy.go
+++ b/internal/trie/node/copy.go
@@ -3,76 +3,107 @@
 
 package node
 
+// CopySettings contains settings to configure the deep copy
+// of a node. By default, it:
+// - does not deep copy children recrursively
+// - does not copy cached fields HashDigest and Encoding
+// - deep copies the key field
+// - deep copies the value field
+type CopySettings struct {
+	// CopyChildren can be set to true to recursively deep copy the eventual
+	// children of the node. This is false by default and should only be used
+	// in tests since it is quite inefficient.
+	CopyChildren bool
+	// CopyCached can be set to true to deep copy the cached digest
+	// and encoding fields on the current node copied.
+	// This is false by default because in production, a node is copied
+	// when it is about to be mutated, hence making its cached fields
+	// no longer valid.
+	CopyCached bool
+	// LeaveKeyEmpty can be set to true to not deep copy the key field of
+	// the node. This is useful if the key is about to be assigned after the
+	// Copy operation, to save a memory operation.
+	LeaveKeyEmpty bool
+	// LeaveValueEmpty can be set to true to not deep copy the value field of
+	// the node. This is useful if the value is about to be assigned after the
+	// Copy operation, to save a memory operation.
+	LeaveValueEmpty bool
+}
+
 // Copy deep copies the branch.
 // Setting copyChildren to true will deep copy
 // children as well.
-func (b *Branch) Copy(copyChildren bool) Node {
+func (b *Branch) Copy(settings CopySettings) Node {
 	cpy := &Branch{
 		Dirty:      b.Dirty,
 		Generation: b.Generation,
 	}
 
-	if copyChildren {
+	if settings.CopyChildren {
 		for i, child := range b.Children {
 			if child == nil {
 				continue
 			}
-			cpy.Children[i] = child.Copy(copyChildren)
+			cpy.Children[i] = child.Copy(settings)
 		}
 	} else {
 		cpy.Children = b.Children // copy interface pointers only
 	}
 
-	if b.Key != nil {
+	if !settings.LeaveKeyEmpty && b.Key != nil {
 		cpy.Key = make([]byte, len(b.Key))
 		copy(cpy.Key, b.Key)
 	}
 
 	// nil and []byte{} are encoded differently, watch out!
-	if b.Value != nil {
+	if !settings.LeaveValueEmpty && b.Value != nil {
 		cpy.Value = make([]byte, len(b.Value))
 		copy(cpy.Value, b.Value)
 	}
 
-	if b.HashDigest != nil {
-		cpy.HashDigest = make([]byte, len(b.HashDigest))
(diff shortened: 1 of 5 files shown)


EclesioMeloJunior (approved):

internal/trie/node/copy.go
```diff
@@ -3,76 +3,107 @@
 
 package node
 
+// CopySettings contains settings to configure the deep copy
+// of a node. By default, it:
+// - does not deep copy children recrursively
+// - does not copy cached fields HashDigest and Encoding
+// - deep copies the key field
+// - deep copies the value field
+type CopySettings struct {
```
noot, line 12: it would be nice to also have a `DefaultCopySettings` variable potentially?

internal/trie/node/copy.go
```diff
@@ -3,76 +3,107 @@
 
 package node
 
+// CopySettings contains settings to configure the deep copy
+// of a node. By default, it:
+// - does not deep copy children recrursively
```
noot, line 8:
```suggestion
// - does not deep copy children recursively
```

noot (approved):

internal/trie/node/node.go
```diff
@@ -20,7 +20,7 @@ type Node interface {
 	GetValue() (value []byte)
 	GetGeneration() (generation uint64)
 	SetGeneration(generation uint64)
-	Copy(copyChildren bool) Node
+	Copy(settings CopySettings) (cpy Node)
```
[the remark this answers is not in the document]
qdm12: Added `node.DeepCopySettings()` instead, which returns copy settings to deep copy everything (for tests).

internal/trie/node/copy.go
qdm12: Added `DefaultCopySettings() CopySettings` for this, and also changed field names

- `LeaveKeyEmpty` ==> `CopyKey` (inverted)
- `LeaveValueEmpty` ==> `CopyValue` (inverted)

so now the empty CopySettings struct does not copy anything, the dev has to use `node.DefaultCopySettings()` or fill what needs to be copied.

internal/trie/node/copy.go
qdm12: Nice spot! Fixed!

internal/trie/node/copy.go
```diff
@@ -3,76 +3,136 @@
...
+	}
+}
+
+// DeepCopySettings returns the following copy settings:
+// - children are deep copied recursively
+// - the HashDigest field is deep copied
+// - the Encoding field is deep copied
+// - the key field is deep copied
+// - the value field is deep copied
+func DeepCopySettings() CopySettings {
```
timwu20, line 25: same here

timwu20 (changes requested):

internal/trie/node/copy.go
```diff
@@ -3,76 +3,136 @@
...
+// - the HashDigest field is left empty on the copy
+// - the Encoding field is left empty on the copy
+// - the key field is deep copied
+// - the value field is deep copied
+func DefaultCopySettings() CopySettings {
+	return CopySettings{
+		CopyKey:   true,
+		CopyValue: true,
+	}
+}
```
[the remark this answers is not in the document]
qdm12: To have it more 'immutable' / less 'global variable'.
This allows to easily get the default copy settings and then modify it, without affecting other parts of the code.

internal/trie/node/copy.go
qdm12: See comment https://github.com/ChainSafe/gossamer/pull/2384#discussion_r833079804

internal/trie/node/copy.go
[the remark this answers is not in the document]
timwu20: I think it's understood that you shouldn't be modifying package level variables.  Even if someone decides to copy and modify a few attributes, it won't update the package level instance.  I don't see the need to call a function for this and allocating more memory.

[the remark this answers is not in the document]
qdm12: > I think it's understood that you shouldn't be modifying package level variables.

Disagree :smile_cat: So many times I have seen (I guess even myself) inadvertently modifying a global variable.

> Even if someone decides to copy and modify a few attributes, it won't update the package level instance.

Someone could do `node.DefaultCopySettings.SomeField = true` and modify the package level variable.
But yes in our case, we always assign it to a variable, which copies it on the stack and since there is no pointer, it won't modify the package level variable this way.

> I don't see the need to call a function for this and allocating more memory.

Yes I agree, I changed it to global variable, a function call isn't really necessary here.
**However**, on the memory allocation side, TIL and also now you know:

```go
var globalVar = CopySettings{}

func getGlobalVar() CopySettings {
	return CopySettings{}
}

func BenchmarkXxx(b *testing.B) {
	b.Run("global variable", func(b *testing.B) {
		for i := 0; i < b.N; i++ {
			x := globalVar
			x.CopyCached = true
		}
	})

	b.Run("function", func(b *testing.B) {
		for i := 0; i < b.N; i++ {
			x := getGlobalVar()
			x.CopyCached = true
		}
	})
}
```

Gives

```
BenchmarkXxx/global_variable
BenchmarkXxx/global_variable-8         	1000000000	         0.3357 ns/op	       0 B/op	       0 allocs/op
BenchmarkXxx/function
BenchmarkXxx/function-8                	1000000000	         0.3395 ns/op	       0 B/op	       0 allocs/op
```

So both don't use any memory (I guess at least on the heap) and take really the same time.

EclesioMeloJunior (approved):

timwu20 (approved):

PR #2385, opened 2022-03-17
chore(tests): refactor tests/utils/request_utils.go

qdm12:
## Changes

- Inject context to `PostRPC` and `PostRPCWithRetry`
- Swap method and url arguments for `PostRPC` and `PostRPCWithRetry`
- Remove some global variables
- Wrap all possible errors
- `PostRPCWithRetry` retries until context is canceled
- Keep 1s timeout for RPC calls

## Tests

Existing tests suites passing

## Issues

- #2386

### DIFF
--- a/tests/rpc/rpc_00_test.go
+++ b/tests/rpc/rpc_00_test.go
@@ -4,6 +4,7 @@
 package rpc
 
 import (
+	"context"
 	"fmt"
 	"os"
 	"reflect"
@@ -38,13 +39,14 @@ type testCase struct {
 	skip        bool
 }
 
-func getResponse(t *testing.T, test *testCase) interface{} {
+func getResponse(ctx context.Context, t *testing.T, test *testCase) interface{} {
 	if test.skip {
 		t.Skip("RPC endpoint not yet implemented")
 		return nil
 	}
 
-	respBody, err := utils.PostRPC(test.method, utils.NewEndpoint(currentPort), test.params)
+	endpoint := utils.NewEndpoint(currentPort)
+	respBody, err := utils.PostRPC(ctx, endpoint, test.method, test.params)
 	require.NoError(t, err)
 
 	target := reflect.New(reflect.TypeOf(test.expected)).Interface()
--- a/tests/utils/chain.go
+++ b/tests/utils/chain.go
@@ -4,9 +4,11 @@
 package utils
 
 import (
+	"context"
 	"fmt"
 	"strconv"
 	"testing"
+	"time"
 
 	"github.com/ChainSafe/gossamer/dot/rpc/modules"
 	"github.com/ChainSafe/gossamer/dot/types"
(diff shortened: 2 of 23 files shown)


noot (approved):

tests/rpc/rpc_00_test.go
```diff
@@ -38,13 +39,14 @@ type testCase struct {
 	skip        bool
 }
 
-func getResponse(t *testing.T, test *testCase) interface{} {
+func getResponse(ctx context.Context, t *testing.T, test *testCase) interface{} {
```
timwu20, line 42: could this function not just accept a `time.Duration` for timeout and handle context creation and cancellation within this function?  Doesn't look like the contexts are used for anything other than timeout functionality.

tests/utils/chain.go
```diff
@@ -40,12 +46,14 @@ func GetChainHeadWithError(t *testing.T, node *Node) (*types.Header, error) {
...
 
-// GetBlockHash calls the endpoint chain_getBlockHash to get the latest chain head
-func GetBlockHash(t *testing.T, node *Node, num string) (common.Hash, error) {
-	respBody, err := PostRPCWithRetry(ChainGetBlockHash, NewEndpoint(node.RPCPort), "["+num+"]", 5)
-	if err != nil {
-		return common.Hash{}, err
-	}
+// GetBlockHash calls the endpoint chain_getBlockHash to get the latest chain head.
+// It will block until a response is received or the context gets canceled.
+func GetBlockHash(ctx context.Context, t *testing.T, node *Node, num string) (common.Hash, error) {
```
timwu20, line 51: Not sure why we're adding `ctx` to all these methods.  I haven't seen a case where we have to cancel these requests bound to some other context.  The actual test code diff will be smaller, if we just create a context with cancel in this function, and handle it internally.  It should return an error if it's past a supplied timeout duration.

timwu20 (changes requested):

tests/rpc/rpc_00_test.go
qdm12: Yes for now, but the idea I have at the back of my head is to use it with `cancel()` in the future once it's more event-driven.

Context offer the flexibility of timeouts + event driven cancellation, and inherit from each other, so we should really be using them in e2e tests.

tests/utils/chain.go
qdm12: See above comment: https://github.com/ChainSafe/gossamer/pull/2385#discussion_r833071646

EclesioMeloJunior (changes requested):

tests/rpc/rpc_00_test.go
qdm12: Further down the line in my branches, this context is actually used to cancel operations, for example if a node crashes unexpectedly.

timwu20: I'd argue that we shouldn't be using context for cancellation.  Dave Cheney wrote about it in his [blog](https://dave.cheney.net/2017/08/20/context-isnt-for-cancellation). 

> The documentation for the context package strongly recommends that context.Context is only for request scoped values:

 With regards to cascading cancellation, you can just have a cancel channel that is closed and can be listened on by multiple goroutines.  

I know using contexts for cancellation is a common pattern.  It's something to keep an eye on for Go 2.  They may remove the cancellation from the context package.

timwu20 (approved):

tests/rpc/rpc_00_test.go
qdm12: Interesting blog post, and I 100% agree context values are the fruit of demons developers :imp: 
It's also sad they didn't figure out how to wait for a goroutine to finish on context cancellation, forcing to still have another signal/error channel to wait.

Now for the closing of a signal channel, there are two limitations I can think of:

- Handling timers to close a channel is painful (always need to check to empty the timer channel as well), and even more when nesting them.
- You cannot close a channel multiple times, so you would have to pretty much re-invent contexts for this (check if the channel is closed etc.)
- The standard library accepts contexts to cancel IO operations, but not close channels and often have no easy `Stop` method. One could use a signal channel to cancel a context down the callers, but it feels wrong to me.

PR #2408, opened 2022-03-21
fix(dot/network): re-add nil mutex check for disconnected peer

qdm12:
## Changes

- Re-add nil mutex check
- Fixes the SIGABRT ugly nil pointer issue

## Primary Reviewer

- @kishansagathiya

### DIFF
--- a/dot/network/notifications.go
+++ b/dot/network/notifications.go
@@ -284,6 +284,8 @@ func (s *Service) sendData(peer peer.ID, hs Handshake, info *notificationsProtoc
 	}, peer)
 }
 
+var errPeerDisconnected = errors.New("peer disconnected")
+
 func (s *Service) sendHandshake(peer peer.ID, hs Handshake, info *notificationsProtocol) (libp2pnetwork.Stream, error) {
 	// multiple processes could each call this upcoming section, opening multiple streams and
 	// sending multiple handshakes. thus, we need to have a per-peer and per-protocol lock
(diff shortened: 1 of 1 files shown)


dot/network/notifications.go
```diff
@@ -293,6 +295,12 @@ func (s *Service) sendHandshake(peer peer.ID, hs Handshake, info *notificationsP
 	// so we cannot have a method on peersData to lock and unlock the mutex
 	// from the map
 	peerMutex := info.peersData.getMutex(peer)
+	if peerMutex == nil {
+		// Note: the only place the mutex is deleted is when the peer disconnects.
+		// If it doesn't exist, the peer never connected either.
```
kishansagathiya, line 300: If it doesn't exist, the peer either never connected or got disconnected, right?

kishansagathiya (approved):
Tested it to be fixing the problem!

Great stuff :tada:

noot (approved):

timwu20 (approved):

qdm12:
@kishansagathiya 

> Do we already have something like this? Something basic that runs gossamer network for a while should do.

No, created https://github.com/ChainSafe/gossamer/issues/2409 for this.

> If it doesn't exist, the peer either never connected or got disconnected, right?

Yes that's right. The only panics I've seen where when we delete a mutex then try to get it, so only when a peer disconnects. I haven't seen the case where we try to get from a peer that did not connect before, but it could be be maybe.

PR #2425, opened 2022-03-22
fix(lib/runtime): avoid caching version in runtime instance

FlorianFranzen:
## Changes

- No longer provide workaround for runtimes with old babe implementation
- Stop caching runtime version inside life and wasmer instance

-> The runtime environment no longer expect the `Version()` Runtime API to be present in all runtimes.

## Tests

I have not tested this well yet, beyond checking if it fixes w3f/polkadot-tests#19.

## Issues

- w3f/polkadot-tests#19

## Primary Reviewer

Anyone

### DIFF
--- a/lib/runtime/life/exports.go
+++ b/lib/runtime/life/exports.go
@@ -4,7 +4,6 @@
 package life
 
 import (
-	"bytes"
 	"errors"
 	"fmt"
 	"strings"
@@ -141,12 +140,6 @@ func (in *Instance) ExecuteBlock(block *types.Block) ([]byte, error) {
 
 	// remove seal digest only
 	for _, d := range block.Header.Digest.Types {
-		// hack since substrate node_runtime can't seem to handle BABE pre-runtime digests
-		// with type prefix (ie Primary, Secondary...)
-		if bytes.Equal(in.version.SpecName(), []byte("node")) {
-			break
-		}
-
 		switch d.Value().(type) {
 		case types.SealDigest:
 			continue
--- a/lib/runtime/life/instance.go
+++ b/lib/runtime/life/instance.go
@@ -39,11 +39,10 @@ type Config struct {
 	Resolver exec.ImportResolver
 }
 
-// Instance represents a v0.8 runtime life instance
+// Instance is a runtime life instance
 type Instance struct {
-	vm      *exec.VirtualMachine
-	mu      sync.Mutex
-	version runtime.Version
+	vm *exec.VirtualMachine
+	mu sync.Mutex
 }
 
 // GetCodeHash returns code hash of the runtime
@@ -123,10 +122,6 @@ func NewInstance(code []byte, cfg *Config) (*Instance, error) {
 	}
 
 	ctx = runtimeCtx
-	inst.version, err = inst.Version()
-	if err != nil {
-		logger.Errorf("error checking instance version: %s", err)
-	}
 	return inst, nil
 }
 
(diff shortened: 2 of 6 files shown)


lib/runtime/life/instance.go
```diff
@@ -41,9 +41,8 @@ type Config struct {
 
 // Instance represents a v0.8 runtime life instance
```
qdm12, line 42: Should we amend this comment? :thinking: Since the version isn't contained in the instance and also it wasn't fixed :thinking:

lib/runtime/life/exports.go
```diff
@@ -141,12 +140,6 @@ func (in *Instance) ExecuteBlock(block *types.Block) ([]byte, error) {
 
 	// remove seal digest only
 	for _, d := range block.Header.Digest.Types {
-		// hack since substrate node_runtime can't seem to handle BABE pre-runtime digests
-		// with type prefix (ie Primary, Secondary...)
-		if bytes.Equal(in.version.SpecName(), []byte("node")) {
-			break
-		}
```
qdm12, line 148: Do we still need this?

If so, should we then keep it and change it to query the version from the runtime and do this if condition check?

qdm12 (approved):
Nice, thank you!

I wonder why we cached the version initially as well :thinking:

lib/runtime/life/exports.go
FlorianFranzen: The unit test currently seems to indicate so, probably because the `gssmr` runtime is not secondary slot aware.

lib/runtime/life/instance.go
FlorianFranzen: Indeed, this should probably be updated.

jimjbrettj (approved):
lgtm

lib/runtime/life/exports.go
noot: I agree, I'm currently working on updating the host API and dev/gssmr runtimes to be compatible with polkadot v0.9.17. if we merge this, it will temporarily break the unit test but we can just skip it until the runtime is updated

noot (approved):

lib/runtime/life/exports.go
qdm12: Which test should we skip? The `t.Skip("reason")` should be part of this PR.

lib/runtime/life/instance.go
qdm12, line 42:
```suggestion
// Instance is a runtime life instance
```

lib/runtime/life/exports.go
timwu20: `TestInstance_ExecuteBlock_GossamerRuntime`

FlorianFranzen:
Update the PR last night. Test is now disabled and everything passes again. (The CI is still angry I removed the "primary reviewer" section.)

qdm12 (approved):

FlorianFranzen:
> Can you do `git commit --amend --no-edit && git push --force` to re-trigger it please? 

TIL that the commit timestamp is part of the git hash...

FlorianFranzen:
> I've added the `Primary reviewer` section to the PR body.

Looks like the bot was still not happy with that, as it expects "Primary __R__eviewer". So I fixed that and ran the CI again.

qdm12:
Merged! :tada: 

> TIL that the commit timestamp is part of the git hash...

TIL You can actually amend the commit and re-use the older timestamp using

```
git commit --amend --no-edit --date="$(git log -n 1 --format=%aD)"
```

But you still get a different commit hash :shrug: 

> Looks like the bot was still not happy with that

Yeah we may remove that CI workflow, it's rather cumbersome, sorry about that.

PR #2428, opened 2022-03-23
Dependabot should only use `dependencies` label

danforbes:
It looks like some of our "weird" labels were coming from Dependabot as some people had suggested. At the moment, this PR configures Dependabot to only use the `dependabot` label, but I'm happy to hear suggestions for other approaches.

Ref: https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#labels

Primary reviewers: @qdm12 & @timwu20 🙏🏻

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,16 +5,16 @@ updates:
     schedule:
       interval: "weekly"
     labels:
-      - "dependabot"
+      - "dependencies"
   - package-ecosystem: docker
     directory: /
     schedule:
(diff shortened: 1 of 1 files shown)


timwu20 (approved):

qdm12:
@danforbes how about a `dependencies` label instead of `dependabot`? So that it can be used in other contexts as well? :thinking:

qdm12 (approved):

PR #2433, opened 2022-03-24
docs(cmd/gossamer): Metrics port `9090` -> `9876`

qdm12:
## Changes

- Fix Gossamer server metrics port from `9090` to `9876`

## Primary Reviewer

- @danforbes

### DIFF
--- a/cmd/gossamer/README.md
+++ b/cmd/gossamer/README.md
@@ -188,7 +188,7 @@ defined in [lib/runtime/wasmer/exports.go](../../lib/runtime/wasmer/exports.go).
 Gossamer publishes telemetry data and also includes an embedded Prometheus server that reports metrics. The metrics
 capabilities are defined in the [dot/metrics](../../dot/metrics) package and build on
 [the metrics library that is included with Go Ethereum](https://github.com/ethereum/go-ethereum/blob/master/metrics/README.md).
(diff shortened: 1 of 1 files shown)


danforbes (approved):

EclesioMeloJunior (approved):

timwu20:
why are we doing this? to match substrate?

qdm12:
> why are we doing this? to match substrate?

Gossamer uses port `9876`, so this fixes the documentation. Now why we chose `9876`, I guess probably for no reason. [Subtrate uses `9615`](https://docs.substrate.io/tutorials/v3/node-metrics/#start-a-substrate-template-node) and most applications use `9090` in my experience. I'd vote to change it to `9090` but I'm fine with whatever port between 1024 and 49151 as well :smiley_cat:

timwu20:
> > why are we doing this? to match substrate?
> 
> Gossamer uses port `9876`, so this fixes the documentation. Now why we chose `9876`, I guess probably for no reason. [Subtrate uses `9615`](https://docs.substrate.io/tutorials/v3/node-metrics/#start-a-substrate-template-node) and most applications use `9090` in my experience. I'd vote to change it to `9090` but I'm fine with whatever port between 1024 and 49151 as well 😺

Thanks for the explanation.  Let's keep this as is then, I'd rather not change the port now.

PR #2434, opened 2022-03-24
chore(golang-ci): upgrade `golangci-lint` to `v1.45`

EclesioMeloJunior:
## Changes

- In order to check golang 1.18 generic code we should upgrade golangci-lint to version `v1.45`
- Fixes PR https://github.com/ChainSafe/gossamer/pull/2339 CI

## Tests

```
N/A
```

## Issues

- N/A

## Primary Reviewer

- @qdm12

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -79,7 +79,7 @@ linters-settings:
 
 linters:
   enable:
-    - bodyclose
+    # - bodyclose
     - depguard
     - errcheck
     - exportloopref
@@ -96,13 +96,13 @@ linters:
     - megacheck
     - megacheck
     - misspell
-    - nilerr
+    # - nilerr
     - nilnil
     - nolintlint
     - revive
-    - staticcheck
+    # - staticcheck
     - unconvert
-    - unparam
+    # - unparam
     - varcheck
 
   fast: false
@@ -165,6 +165,10 @@ issues:
     - linters:
         - nolintlint
       source: "^//nolint:revive"
+    
+    - linters:
+        - nolintlint
+      source: "^?//nolint(:staticcheck)?"
 
     # Exclude lll issues for long lines with go:generate
     - linters:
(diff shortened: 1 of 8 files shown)


qdm12 (approved):

dot/rpc/modules/grandpa.go
```diff
@@ -65,7 +65,7 @@ func (gm *GrandpaModule) ProveFinality(r *http.Request, req *ProveFinalityReques
 	}
 
 	// Leaving check in for linter
-	if req.authorityID != uint64(0) { // nolint:staticcheck
+	if req.authorityID != uint64(0) { //// nolint:staticcheck
```
qdm12, line 68: I would like to suggest to leave these `//nolint` comments here as they were before, and comment out the linters not working with Go 1.18 in `.golangci-lint.yml`, and add custom rules to `nolintlint` such that it ignores `//nolint` comments for these disabled linters.

I'll monitor when these get fixed to work with generics and will re-add them over time.

qdm12 (changes requested):

.golangci.yml
```diff
@@ -165,6 +165,10 @@ issues:
     - linters:
         - nolintlint
       source: "^//nolint:revive"
+    
+    - linters:
+        - nolintlint
+      source: "^?//nolint(:staticcheck)?"
```
qdm12, line 171: Don't we need one for `bodyclose`, `nilerr` and `unparam` as well? :thinking: 

Also I think the regex shouldn't start with `^` if the comment if put at the end of a line right?

qdm12: Oh I get it, can you change that source to

```suggestion
      source: "//nolint:[bodyclose|nilerr|staticcheck|unparama]"
```

And remove that single `//nolint` you found (it was a false positive from staticcheck I think)

qdm12 (changes requested):

qdm12 (approved):

noot (approved):

PR #2435, opened 2022-03-25
Check for duplicate message earlier

### LINKED ISSUE #2098: gossamer should not ban other gossamer nodes

kishansagathiya:
## Changes

- Checks if we have seen the notification message before any processing
happens on it
- Adds more logs and handles more errors

## Tests
- Run a gossamer devnet. (I ran with 4 nodes, all of them authorities)
- You should never see the critical error `highest rated peer is below bannedThresholdValue` 
- Generally you start seeing the error at 37th to 43rd block. So, If I don't see it happening till 80th block, I would assume it never happens.
## Issues

- Fixes #2098

## Primary Reviewer

- @timwu20


dot/network/block_announce.go
```diff
@@ -69,11 +69,14 @@ func (bm *BlockAnnounceMessage) Decode(in []byte) error {
...
 // Hash returns the hash of the BlockAnnounceMessage
-func (bm *BlockAnnounceMessage) Hash() common.Hash {
+func (bm *BlockAnnounceMessage) Hash() (common.Hash, error) {
 	// scale encode each extrinsic
-	encMsg, _ := bm.Encode()
-	hash, _ := common.Blake2bHash(encMsg)
-	return hash
+	encMsg, err := bm.Encode()
+	if err != nil {
+		return common.Hash{}, err
```
qdm12, line 76: nit wrap the error

```suggestion
		return common.Hash{}, fmt.Errorf("cannot encode message: %w", err)
```

dot/network/gossip.go
```diff
@@ -24,13 +25,17 @@ func newGossip() *gossip {
 }
 
 // hasSeen broadcasts messages that have not been seen
-func (g *gossip) hasSeen(msg NotificationsMessage) bool {
+func (g *gossip) hasSeen(msg NotificationsMessage) (bool, error) {
```
qdm12, line 30: nit named returns would be nice

```suggestion
func (g *gossip) hasSeen(msg NotificationsMessage) (seen bool, err error) {
```

dot/network/message.go
```diff
@@ -389,11 +389,13 @@ func (cm *ConsensusMessage) Decode(in []byte) error {
...
 // Hash returns the Hash of ConsensusMessage
-func (cm *ConsensusMessage) Hash() common.Hash {
+func (cm *ConsensusMessage) Hash() (common.Hash, error) {
 	// scale encode each extrinsic
-	encMsg, _ := cm.Encode()
-	hash, _ := common.Blake2bHash(encMsg)
-	return hash
+	encMsg, err := cm.Encode()
+	if err != nil {
+		return common.Hash{}, err
```
qdm12, line 396: nit

```suggestion
		return common.Hash{}, fmt.Errorf("cannot encode message: %w", err)
```

dot/network/message_cache.go
```diff
@@ -55,6 +55,7 @@ func (m *messageCache) put(peer peer.ID, msg NotificationsMessage) (bool, error)
 func (m *messageCache) exists(peer peer.ID, msg NotificationsMessage) bool {
 	key, err := generateCacheKey(peer, msg)
 	if err != nil {
+		logger.Errorf("could not generate cache key: %w", err)
```
qdm12, line 58:
```suggestion
		logger.Errorf("could not generate cache key: %s", err)
```

dot/network/message_cache.go
```diff
@@ -67,7 +68,12 @@ func generateCacheKey(peer peer.ID, msg NotificationsMessage) ([]byte, error) {
 		return nil, errors.New("cache does not support handshake messages")
 	}
 
-	peerMsgHash, err := common.Blake2bHash(append([]byte(peer), msg.Hash().ToBytes()...))
+	msgHash, err := msg.Hash()
+	if err != nil {
+		return nil, err
```
qdm12, line 73:
```suggestion
		return nil, fmt.Errorf("cannot hash notification message: %w", err)
```

dot/network/notifications.go
```diff
@@ -238,7 +243,14 @@ func (s *Service) sendData(peer peer.ID, hs Handshake, info *notificationsProtoc
 		return
 	}
 
-	if support, err := s.host.supportsProtocol(peer, info.protocolID); err != nil || !support {
+	support, err := s.host.supportsProtocol(peer, info.protocolID)
+	if err != nil {
+		logger.Criticalf("could not check if protocol %s is supported by peer %s: %s", info.protocolID, peer, err)
+	} else if !support {
+		logger.Criticalf("protocol %s is not supported by peer %s", info.protocolID, peer)
```
qdm12, line 250: Should these be logged at the critical level? Why?

dot/network/transaction.go
```diff
@@ -54,10 +54,12 @@ func (tm *TransactionMessage) Decode(in []byte) error {
...
 
 // Hash returns the hash of the TransactionMessage
-func (tm *TransactionMessage) Hash() common.Hash {
-	encMsg, _ := tm.Encode()
-	hash, _ := common.Blake2bHash(encMsg)
-	return hash
+func (tm *TransactionMessage) Hash() (common.Hash, error) {
+	encMsg, err := tm.Encode()
+	if err != nil {
+		return common.Hash{}, nil
```
qdm12, line 60: Should we return an error here?

dot/network/transaction.go
```diff
@@ -129,6 +131,7 @@ func (s *Service) startTxnBatchProcessing(txnBatchCh chan *BatchMessage, slotDur
 				case txnMsg := <-txnBatchCh:
 					propagate, err := s.handleTransactionMessage(txnMsg.peer, txnMsg.msg)
 					if err != nil {
+						logger.Warnf("could not handle transaction message: %s ", err)
```
qdm12, line 134: supernit

```suggestion
						logger.Warnf("could not handle transaction message: %s", err)
```

dot/network/transaction.go
```diff
@@ -137,7 +140,16 @@ func (s *Service) startTxnBatchProcessing(txnBatchCh chan *BatchMessage, slotDur
 						continue
 					}
 
-					if !s.gossip.hasSeen(txnMsg.msg) {
+					// TODO: Check if s.gossip.hasSeen should be moved before handleTransactionMessage.
+					// That we could avoid handling the transactions again, which we would have already seen.
```
qdm12, line 144: TODO reminder

dot/peerset/handler.go
```diff
@@ -77,6 +77,10 @@ func (h *Handler) RemovePeer(setID int, peers ...peer.ID) {
 
 // ReportPeer reports ReputationChange according to the peer behaviour.
 func (h *Handler) ReportPeer(rep ReputationChange, peers ...peer.ID) {
+	for _, pid := range peers {
+		logger.Debugf("reporting reputation change of %d to peer %s, reason: %s", rep.Value, pid, rep.Reason)
+	}
+
```
qdm12, line 83: A tiny bit out of scope, but please change this method to accept a single peer ID. I checked its references, and it always get passed a single peer. `ReportPeer` doesn't match accepting multiple peers in its signature as well.
That way you can simplify that debug logging for loop you have as well.

dot/state/block.go
```diff
@@ -164,6 +164,7 @@ func headerKey(hash common.Hash) []byte {
 }
 
 // headerHashKey = headerHashPrefix + num (uint64 big endian)
+// TODO: Shouldn't we change this block number to uint32?
```
qdm12, line 167: Created https://github.com/ChainSafe/gossamer/issues/2436 for it. It would break database storage, that's why I didn't change it a few weeks ago when changing block numbers to use `uint`

```suggestion
```

dot/types/block_data.go
```diff
@@ -28,7 +28,7 @@ func NewEmptyBlockData() *BlockData {
 
 // Number returns the block header number.
 func (bd *BlockData) Number() uint {
-	return bd.Header.Number
+	return uint(bd.Header.Number)
```
qdm12, line 31: no need right? It's already a `uint`

```suggestion
	return bd.Header.Number
```

go.mod
```diff
@@ -58,7 +58,7 @@ require (
 	github.com/btcsuite/btcd v0.22.0-beta // indirect
 	github.com/cespare/xxhash v1.1.0 // indirect
 	github.com/cespare/xxhash/v2 v2.1.2 // indirect
-	github.com/cpuguy83/go-md2man/v2 v2.0.0 // indirect
+	github.com/cpuguy83/go-md2man/v2 v2.0.1 // indirect
```
qdm12, line 61: Should these go.mod dependency upgrades be discarded? Aren't they out of scope?

lib/grandpa/network.go
```diff
@@ -64,8 +64,8 @@ func (*GrandpaHandshake) Type() byte {
 }
 
 // Hash ...
-func (*GrandpaHandshake) Hash() common.Hash {
-	return common.Hash{}
+func (*GrandpaHandshake) Hash() (common.Hash, error) {
+	return common.Hash{}, nil
```
qdm12, line 68: for all these message types not hashing anything, should we `panic("unimplemented")` to make sure these Hash are never called?

dot/network/gossip_test.go
```diff
@@ -89,23 +89,26 @@ func TestGossip(t *testing.T) {
...
+	if hasSeenB, ok := nodeB.gossip.seen.Load(hash); !ok || hasSeenB.(bool) == false {
 		t.Error(
 			"node B did not receive block request message from node A",
 			"\nreceived:", hasSeenB,
 			"\nexpected:", true,
 		)
 	}
 
-	if hasSeenC, ok := nodeC.gossip.seen.Load(announceMessage.Hash()); !ok || hasSeenC.(bool) == false {
+	if hasSeenC, ok := nodeC.gossip.seen.Load(hash); !ok || hasSeenC.(bool) == false {
```
qdm12: rebase on development / merge development in, this is now a standard map with mutex.

qdm12 (changes requested):
Nice :+1: Just a few comments

dot/network/gossip.go
```diff
@@ -26,18 +27,22 @@ func newGossip() *gossip {
...
 	// check if message has not been seen
```
EclesioMeloJunior, line 31: move this comment to one line before the `_, ok := g.seenMap[msgHash]`

dot/network/message_cache.go
```diff
@@ -55,6 +56,7 @@ func (m *messageCache) put(peer peer.ID, msg NotificationsMessage) (bool, error)
 func (m *messageCache) exists(peer peer.ID, msg NotificationsMessage) bool {
 	key, err := generateCacheKey(peer, msg)
 	if err != nil {
+		logger.Errorf("could not generate cache key: %s", err)
```
EclesioMeloJunior, line 59: can we change the `exists` method to return an `(bool, error)`? and at `generateCacheKey` if the `msg` is a handshake return and error `ErrHandshakeNotSupported`

EclesioMeloJunior (changes requested):

dot/network/notifications.go
kishansagathiya: fixed

dot/network/transaction.go
kishansagathiya: fixed

dot/network/transaction.go
kishansagathiya: out of scope, created an issue https://github.com/ChainSafe/gossamer/issues/2445

dot/peerset/handler.go
kishansagathiya: It's a feature not a bug. Accepting multiple peers is a common theme amongst all peerset handler funcs.

lib/grandpa/network.go
kishansagathiya: linked this comment in the issue for refactoring opportunities https://github.com/ChainSafe/gossamer/issues/2437

dot/network/message_cache.go
kishansagathiya: we don't store handshake in `messageCache`. I dont think you would you want to check if message from messageCache is a handshake. Also, messageCache is for out going messages, so we have cntrol over it.

> can we change the exists method to return an (bool, error)?
No other piece of code needs this error. Only thing we are going to do with it, is logging it. We want to keep the code running in case of error here. return false in case of error is enough here,

go.mod
kishansagathiya: >  Aren't they out of scope?

May be.

But, all of those go.mod and go.sum changes are done by `go mod tidy`. As long as it is not hurting the code, I see no reason to go in it and manually change things!

dot/network/gossip.go
```diff
@@ -25,19 +26,23 @@ func newGossip() *gossip {
 	}
 }
 
-// hasSeen broadcasts messages that have not been seen
-func (g *gossip) hasSeen(msg NotificationsMessage) bool {
-	// check if message has not been seen
-	msgHash := msg.Hash()
+// hasSeen checks if we have seen given message before.
```
qdm12, line 29: nit

```suggestion
// hasSeen checks if we have seen the given message before.
```

dot/network/transaction.go
qdm12: Please then mention the issue #2445 in this comment

go.mod
qdm12: That's strange, `go mod tidy` doesn't do that for me. Please re-check?

kishansagathiya: reverted go.mod changes and ran `go mod tidy` again!

qdm12:
```sh
git checkout kishan/fix/dont-ban-us
git reset d2ec68d3e319d08f32dc0df4adf1fd7531d12217
git checkout go.mod go.sum
go mod tidy
```

does not produce any change, make sure you revert go.sum as well before re-running `go mod tidy`?

qdm12 (changes requested):

qdm12 (approved):
Nice :+1: Please just add a comment to [this code](https://github.com/ChainSafe/gossamer/pull/2435/files#r836394168) with the issue number you created

dot/peerset/handler.go
```diff
@@ -77,6 +77,10 @@ func (h *Handler) RemovePeer(setID int, peers ...peer.ID) {
 
 // ReportPeer reports ReputationChange according to the peer behaviour.
 func (h *Handler) ReportPeer(rep ReputationChange, peers ...peer.ID) {
+	for _, pid := range peers {
```
timwu20, line 80: is there a way to check logging level? so we don't range across peers if package logging level isn't trace or debug.

qdm12: We always ever have a single peer for this method ([slightly related comment](https://github.com/ChainSafe/gossamer/pull/2435#discussion_r835155766)), so it's fine really.

timwu20 (approved):

dot/network/notifications.go
```diff
@@ -238,7 +243,12 @@ func (s *Service) sendData(peer peer.ID, hs Handshake, info *notificationsProtoc
 		return
 	}
 
-	if support, err := s.host.supportsProtocol(peer, info.protocolID); err != nil || !support {
+	support, err := s.host.supportsProtocol(peer, info.protocolID)
+	if err != nil {
+		logger.Errorf("could not check if protocol %s is supported by peer %s: %s", info.protocolID, peer, err)
+	}
+
+	if err != nil || !support {
```
EclesioMeloJunior, line 251: should we report a peer if some error occurs while trying to check if the protocol is supported by the peer? @noot @timwu20. I think if an error occurs we should just log and then return.

EclesioMeloJunior (changes requested):

dot/network/notifications.go
noot: @EclesioMeloJunior I believe this logic is from substrate - we can double check if substrate downscores a peer based on them not supporting a protocol

noot (approved):

dot/network/notifications.go
kishansagathiya: that's a pretty good catch, thanks @EclesioMeloJunior

Issue #2388, opened 2022-03-17
tests/utils: handle `utils.Node` as a value instead of a pointer


qdm12:
## Task summary

There is no reason to use a pointer for `Node` and this makes the code more complicated.

- Change all functions to use `Node` instead of `*Node`
- Change RPC functions to take in RPC port string only instead of the entire `Node` struct

This is created from #2326 and because it blocks another refactor down the 'branches'.

qdm12:
Done in #2465

PR #2465, opened 2022-04-04
chore(end-to-end): handle `Node` structure without pointer

qdm12:
## Changes

- Change all functions to use `Node` instead of `*Node` since there is no reason to use a pointer really
- Change RPC functions to take in RPC port string only
- This unblocks a refactor in a next PR

## Tests

All end to end tests to pass

## Issues

- #2388 

## Primary Reviewer

- @edwardmack

### DIFF
--- a/tests/rpc/rpc_05-state_test.go
+++ b/tests/rpc/rpc_05-state_test.go
@@ -37,7 +37,7 @@ func TestStateRPCResponseValidation(t *testing.T) {
 	ctx := context.Background()
 
 	getBlockHashCtx, cancel := context.WithTimeout(ctx, time.Second)
-	blockHash, err := utils.GetBlockHash(getBlockHashCtx, t, nodes[0], "")
+	blockHash, err := utils.GetBlockHash(getBlockHashCtx, t, nodes[0].RPCPort, "")
 	cancel()
 	require.NoError(t, err)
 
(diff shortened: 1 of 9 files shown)


EclesioMeloJunior (approved):

kishansagathiya (commented):
> Change all functions to use Node instead of *Node since there is no reason to use a pointer really

Please provide some information on why this change is required.
Also, tag someone as primary reviewer. If you can't think of anyone who would be obviously aware of this piece of code, tag Tim.

qdm12:
> Please provide some information on why this change is required.

I've been refactoring pretty much the entire e2e helpers through a few commits. This one is one of them and doesn't have a useful value on its own, but I'm aiming at making small pull requests. Although I could do one single thick one if you prefer as well, let me know.

> Also, tag someone as primary reviewer. If you can't think of anyone who would be obviously aware of this piece of code, tag Tim.

On that, I would say it depends, sometimes the changes can be understood by anyone (like CI changes, or this simple PR) so I feel `anyone` is appropriate. I spotted Ed wrote a few lines in this file so I tagged him as primary reviewer, but I don't feel it's necessary :thinking: I'll bring it up in standup, I think it's a good question :wink:

timwu20 (approved):
Nice work! I'd recommend not using pointers to slices either.

qdm12:
> I'd recommend not using pointers to slices. Copying slices isn't performance intensive enough to justify the extra layer of complexity.

I assume you mean `not using pointers in slices` instead of `not using pointers to slices`; And yeah I agree, copying on the stack is quite efficient, and doesn't add load on the GC so not using pointers often result in the same or better performance. But more importantly, using pointers is risky due to unforeseen mutation, so using pointers for no reason is **meh**.

kishansagathiya (approved):

Issue #2397, opened 2022-03-18
Block Numbers are Not Verified Against Block Hashes in GRANDPA (GSR-04)


danforbes:
GRANDPA messages such as votes, commits and justifications often have both a `Hash` field and `Number` field. Where `Hash` is the block hash and `Number` is the block number associated with the hash.

Gossamer does not verify that the `Number` matches the `Hash`.

The following functions do not verify a `Number` is associated with a `Hash`:
1. `VerifyBlockJustification()` does not verify the `Justification.Commit.Number `against `Justification.Commit.Hash`
2. `VerifyBlockJustification()` does not verify the `GrandpaVote.Number` against `GrandpaVote.Hash` for each commit in `Justification.Commit.Percommits`
3. `handleCatchUpResponse()` does not verify `CatchUpResponse.Number` against `CatchUpResponse.Hash`
4. `handleCommitMessage()` does not verify `CommitMessage.Vote.Number` against `CommitMessage.Vote.Number`
5. `verifyPreCommitJustification()` does not verify `CatchUpResponse.PreCommitJustification.Vote.Number` against `CatchUpResponse.PreCommitJustification.Vote.Number`
6. `verifyPreVoteJustification()` does not verify `CatchUpResponse.PreCommitJustification.Vote.Number` against `CatchUpResponse.PreCommitJustification.Vote.Number`
7. `handleCommitMessage()` does not verify `CommitMessage.Vote.Number` against `CommitMessage.Vote.Number`

This issue may be resolved by adding validation checks for each of these messages to ensure that the `Number` matches the `Hash`. The checks may be implemented by using the `Hash` to fetch the block header from the block state. Then ensuring the `Number` equals the number in the block header.

Note that items 5-7 may be resolved by handing these checks in `validateJustification()`

PR #2468, opened 2022-04-06
task(lib/grandpa): Verify block hashes and number

kishansagathiya:
## Changes

- Verify block hashes against block number in justification
- Fix some of the tests
## Tests

there are tests for verifying justification in `message_handler_test.go`. I think if those pass, that should be enough.

## Issues

- Fixes https://github.com/ChainSafe/gossamer/issues/2397

## Primary Reviewer

- @timwu20


lib/grandpa/message_handler.go
```diff
@@ -573,6 +609,18 @@ func (s *Service) VerifyBlockJustification(hash common.Hash, justification []byt
 	return nil
 }
 
+func verifyBlockHashAgainstBlockNumber(bs BlockState, hash common.Hash, number uint) error {
+	header, err := bs.GetHeader(hash)
+	if err != nil {
+		return fmt.Errorf("could not get header from block hash: %w", err)
```
noot, line 615: is this error is returned, we likely haven't synced this block yet, and should add the message to the `tracker` for processing later. otherwise this will cause finality to get stuck if syncing is lagging

lib/grandpa/message_handler.go
```diff
@@ -386,6 +403,14 @@ func (h *MessageHandler) verifyPreVoteJustification(msg *CatchUpResponse) (commo
 }
 
 func (h *MessageHandler) verifyPreCommitJustification(msg *CatchUpResponse) error {
+
+	for _, pcj := range msg.PreCommitJustification {
+		err := verifyBlockHashAgainstBlockNumber(h.blockState, pcj.Vote.Hash, uint(pcj.Vote.Number))
+		if err != nil {
+			return err
```
qdm12, line 410:
```suggestion
			return fmt.Errorf("failed verifying block hash against block number: %w", err)
```

lib/grandpa/message_handler.go
```diff
@@ -562,6 +587,17 @@ func (s *Service) VerifyBlockJustification(hash common.Hash, justification []byt
 		return ErrMinVotesNotMet
 	}
 
+	if err := verifyBlockHashAgainstBlockNumber(s.blockState, fj.Commit.Hash, uint(fj.Commit.Number)); err != nil {
+		return err
```
qdm12, line 591:
```suggestion
		return fmt.Errorf("failed verifying block hash against block number: %w", err)
```

lib/grandpa/message_handler.go
```diff
@@ -562,6 +587,17 @@ func (s *Service) VerifyBlockJustification(hash common.Hash, justification []byt
...
+	}
+
+	for _, preCommit := range fj.Commit.Precommits {
+		err := verifyBlockHashAgainstBlockNumber(s.blockState, preCommit.Vote.Hash, uint(preCommit.Vote.Number))
+		if err != nil {
+			return err
```
qdm12, line 597:
```suggestion
			return fmt.Errorf("failed verifying block hash against block number: %w", err)
```

lib/grandpa/message_handler.go
```diff
@@ -573,6 +609,18 @@ func (s *Service) VerifyBlockJustification(hash common.Hash, justification []byt
...
 }
 
+func verifyBlockHashAgainstBlockNumber(bs BlockState, hash common.Hash, number uint) error {
+	header, err := bs.GetHeader(hash)
+	if err != nil {
+		return fmt.Errorf("could not get header from block hash: %w", err)
+	}
+
+	if header.Number != number {
+		return ErrBlockHashMismatch
```
qdm12, line 619:
```suggestion
		return fmt.Errorf("%w: expected number %d from header but got number %d", 
			ErrBlockHashMismatch, header.Number, number)
```

lib/grandpa/message_handler_test.go
```diff
@@ -553,7 +577,7 @@ func TestMessageHandler_HandleCatchUpResponse(t *testing.T) {
 
 	h := NewMessageHandler(gs, st.Block, telemetryMock)
 
-	round := uint64(77)
+	round := uint64(1)
```
qdm12, line 580: Why was it `77` before?

lib/grandpa/message_handler.go
```diff
@@ -104,6 +104,11 @@ func (h *MessageHandler) handleNeighbourMessage(msg *NeighbourMessage) error {
 func (h *MessageHandler) handleCommitMessage(msg *CommitMessage) error {
 	logger.Debugf("received commit message, msg: %+v", msg)
 
+	err := verifyBlockHashAgainstBlockNumber(h.blockState, msg.Vote.Hash, uint(msg.Vote.Number))
+	if err != nil {
+		return err
```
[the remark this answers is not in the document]
kishansagathiya: I have deliberately not wrapped the error `block hash does not correspond to given block number` is already enough. `failed verifying block hash against block number` does not provide more information here. `verifyBlockHashAgainstBlockNumber` is a tiny helping function rather than a task. If it was say a bigger task where context of the error was necessary, it would have made sense to wrap the error.

lib/grandpa/message_handler_test.go
kishansagathiya: So you catch up only when you are more than 2 block behind some other peer. There is not anything special about 77, it could be anything more than 3. I changed this because block number 77 is not in the block state, but I might have to change this piece code for address elizabeth's comment.

lib/grandpa/message_handler.go
```diff
@@ -215,10 +227,16 @@ func (h *MessageHandler) handleCatchUpResponse(msg *CatchUpResponse) error {
 
 	prevote, err := h.verifyPreVoteJustification(msg)
 	if err != nil {
+		// no need to check for chaindb.ErrKeyNotFound
+		// Since highest block was present (otherwise execution would not reach here),
+		// we expect previous blocks to exists as well. If they don't exist, we should error.
```
kishansagathiya, line 232: @noot This is what i inferred from the spec but I ain't super sure about this. Need confirmation!

So `msg.Hash` is highest ancestor of all blocks voted on in the arrays of justifications. I am assuming that we sync sequentially(true?). So, if we have the highest block we must have previous blocks, and thus ok to error. Same for verifyPreCommitJustification.

kishansagathiya: addressed

lib/grandpa/message_handler.go
kishansagathiya: addressed

lib/grandpa/message_handler.go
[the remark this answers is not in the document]
qdm12: Makes sense, thanks! :+1:

lib/grandpa/grandpa.go
```diff
@@ -852,6 +852,8 @@ func (s *Service) finalise() error {
 	return s.grandpaState.SetLatestRound(s.state.round)
 }
 
+// TODO: Given that this functions returns []SignedVote and not Justification, it should be renamed
+// accordingly.
```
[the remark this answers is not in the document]
kishansagathiya: removed the comment

lib/grandpa/message_handler.go
```diff
@@ -562,6 +608,17 @@ func (s *Service) VerifyBlockJustification(hash common.Hash, justification []byt
 		return ErrMinVotesNotMet
 	}
 
+	if err := verifyBlockHashAgainstBlockNumber(s.blockState, fj.Commit.Hash, uint(fj.Commit.Number)); err != nil {
```
EclesioMeloJunior, line 611: split the function call from the error check

EclesioMeloJunior (changes requested):

lib/grandpa/message_handler.go
```diff
@@ -573,6 +630,19 @@ func (s *Service) VerifyBlockJustification(hash common.Hash, justification []byt
...
 }
 
+func verifyBlockHashAgainstBlockNumber(bs BlockState, hash common.Hash, number uint) error {
+	header, err := bs.GetHeader(hash)
+	if err != nil {
+		return fmt.Errorf("could not get header from block hash: %w", err)
+	}
+
+	if header.Number != number {
+		return fmt.Errorf("%w: expected number %d from header but got number %d",
```
EclesioMeloJunior, line 640: what is the intended behavior when the numbers don't match?

lib/grandpa/message_handler.go
kishansagathiya: why?

EclesioMeloJunior: The function takes a lot of arguments and it ends up being large. In other parts of the code, you use the separated call and check tho.

lib/grandpa/message_handler.go
kishansagathiya: It means justification is invalid and we through an error

lib/grandpa/message_handler.go
kishansagathiya: done

lib/grandpa/errors.go
```diff
@@ -63,6 +63,8 @@ var (
 	// ErrNoJustification is returned when no justification can be found for a block, ie. it has not been finalised
 	ErrNoJustification = errors.New("no justification found for block")
 
+	ErrBlockHashMismatch = errors.New("block hash does not correspond to given block number")
```
qdm12, line 66: nit a tiny bit clearer

```suggestion
	ErrBlockHashMismatch = errors.New("block hash mismatches hash of block for the given block number")
```

lib/grandpa/message_handler.go
```diff
@@ -300,12 +322,13 @@ func (h *MessageHandler) verifyCommitMessageJustification(fm *CommitMessage) err
...
 		err := h.verifyJustification(just, fm.Round, h.grandpa.state.setID, precommit)
 		if err != nil {
+			logger.Errorf("could not verify justification: %s", err)
 			continue
 		}
 
 		isDescendant, err := h.blockState.IsDescendantOf(fm.Vote.Hash, just.Vote.Hash)
 		if err != nil {
-			logger.Warnf("verifyCommitMessageJustification: %s", err)
+			logger.Warnf("could not check for descendant: %s", err)
```
qdm12, line 331: nit maybe `ancestry` fits better?

```suggestion
			logger.Warnf("could not check for ancestry: %s", err)
```

qdm12 (approved):

lib/grandpa/message_handler.go
```diff
@@ -104,6 +105,17 @@ func (h *MessageHandler) handleNeighbourMessage(msg *NeighbourMessage) error {
...
+	err := verifyBlockHashAgainstBlockNumber(h.blockState, msg.Vote.Hash, uint(msg.Vote.Number))
+	if errors.Is(err, chaindb.ErrKeyNotFound) {
+		h.grandpa.tracker.addCommit(msg)
+		logger.Infof("we might not have synced to the given block %s yet: %s", msg.Vote.Hash, err)
+		return nil
+	}
+
+	if err != nil {
+		return err
+	}
```
timwu20, line 117:
```suggestion
	err := verifyBlockHashAgainstBlockNumber(h.blockState, msg.Vote.Hash, uint(msg.Vote.Number))
	if err != nil {
		if errors.Is(err, chaindb.ErrKeyNotFound) {
			h.grandpa.tracker.addCommit(msg)
			logger.Infof("we might not have synced to the given block %s yet: %s", msg.Vote.Hash, err)
			return nil
		}
		return err
	}

```

lib/grandpa/message_handler.go
```diff
@@ -184,6 +196,16 @@ func (h *MessageHandler) handleCatchUpResponse(msg *CatchUpResponse) error {
...
 
+	err := verifyBlockHashAgainstBlockNumber(h.blockState, msg.Hash, uint(msg.Number))
+	if errors.Is(err, chaindb.ErrKeyNotFound) {
+		h.grandpa.tracker.addCatchUpResponse(msg)
+		logger.Infof("we might not have synced to the given block %s yet: %s", msg.Hash, err)
+		return nil
+	}
+	if err != nil {
+		return err
+	}
```
timwu20, line 207: same here, can we put the `errors.Is` check inside the error check block.

lib/grandpa/message_handler.go
```diff
@@ -573,6 +631,19 @@ func (s *Service) VerifyBlockJustification(hash common.Hash, justification []byt
 	return nil
 }
 
+func verifyBlockHashAgainstBlockNumber(bs BlockState, hash common.Hash, number uint) error {
```
timwu20, line 634: nit: should this be a receiver function on `BlockState`

timwu20 (changes requested):

lib/grandpa/message_handler.go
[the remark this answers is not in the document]
kishansagathiya: Do you mean to add `verifyBlockHashAgainstBlockNumber` in the `BlockState` interface?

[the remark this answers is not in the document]
kishansagathiya: That would be possible, but it is an un-exported helper function. If I add it in the interface, I will have to un-necessarily export it, un-necessarily implement(or mock-implement) it.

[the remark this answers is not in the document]
timwu20: makes sense to me.

dot/state/grandpa.go
```diff
@@ -217,11 +218,11 @@ func (s *GrandpaState) GetSetIDByBlockNumber(num uint) (uint64, error) {
 
 		// if the given block number is greater or equal to the block number of the set ID change,
 		// return the current set ID
-		if num <= changeUpper && num > changeLower {
+		if blockNumber <= changeUpper && blockNumber > changeLower {
```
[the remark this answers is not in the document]
kishansagathiya: Set id change happens in the last block of a given set.

Say current set ID is `s` and at block `b` set is getting updated to `s+1`. Block `b` is part of the set `s`. `GetSetIDChange(setId)` returns the block at which set id was set to `setId`. so block b is smaller than any block in the set `setId`. Thus the condition `blockNumber > changeLower`

Now for `GetSetIDChange(setId+1)` return the last block in the `setId` (call it `b'`). So, now if you pick any block in the `setId` it could be equal to or smaller than `b'`. Thus, `blockNumber <= changeUpper`

Issue #2165, opened 2022-01-10
Fix Flaky Tests in `TestSync_Basic`


danforbes:
[Logs](https://github.com/ChainSafe/gossamer/runs/4164828765?check_suite_focus=true)

Issue #2391, opened 2022-03-17
tests/utils: refactor nodes starting and stopping and detect runtime errors


qdm12:
## Task summary

Right now we create nodes as subprocesses, write the log output to files and kill them using the OS kill process.
These functions should be changed such that:

- they are `context.Context` based (**so each node can be stopped gracefully by canceling the context**)
- they return a runtime error channel (**so we can detect runtime failures**)

The following functions have to be modified:

- `InitGossamer`
- `startGossamer`
- `RunGossamer`

These may be moved as methods on the `Node` type, which is more intuitive to use.

qdm12:
Done in 62204027a47ec40b2c9e9ecf6529018a06e244d1

Issue #2390, opened 2022-03-17
tests/utils: add `waitForNode` context-based function to wait for a node


qdm12:
## Task summary

Implement a context-based waiting function `waitForNode` to wait and detect when a node is ready (through a periodic RPC call), instead of relying on trying up to N (=24) times.

qdm12:
Done in 62204027a47ec40b2c9e9ecf6529018a06e244d1

Issue #2389, opened 2022-03-17
tests/utils: make helping functions not test aware


qdm12:
## Task summary

Several helping functions are test aware, and get passed a `t *testing.T`.
These should all be changed to only return an error and not fail the test.
This would allow for a graceful shutdown of each test, and to separate responsabilities.

Created from #2326

qdm12:
Done in 62204027a47ec40b2c9e9ecf6529018a06e244d1

PR #2470, opened 2022-04-06
end-to-end tests: refactor and fix flaky tests

### LINKED ISSUE #2161: Fix Flaky Tests in `TestSystemRPC`
[Logs](https://github.com/ChainSafe/gossamer/runs/4009896511?check_suite_focus=true)

qdm12:
:bulb: **Review commit by commit** by starting with the [first commit](https://github.com/ChainSafe/gossamer/pull/2470/commits/f6013846bfcd8be3de6d81c0d52829a8b041c6dd)

### TLDR

- Fix flakky tests and add testing depth
- More robust code
- Detect runtime errors
- Better concerns separation
- Can easily run a single test independently & locally (after `make gossamer`)

### Details

- Faster `TestAuthorSubmitExtrinsic`
- Fix `TestSync_SingleBlockProducer` (#2167)
- refactor `compareBlocksByNumber`
- refactor `TestSystemRPC` (#2161)
  - add empty skipped tests for missing cases
  - split each subtest individually
  - keep on retrying until main context is canceled
  - fix `networkState` test case
  - assert more fields
  - fix #2161 and #807
- `retry.UntilNoError` function
  - Use in `compareBlocksByNumber`
  - Use in `waitForNode`
  - change `GetBlockHash` to not retry RPC
  - remove `PostWithRetry`
- remove global variables `maxTries` and `testTimeout` in stress tests
- refactor `getResponse`
  - not test aware (#2389)
  - get passed target interface
  - push test assertion/skip to actual test
- rework start and stop of nodes (#2391)
  - Remove unneeded logs
  - Prefix writer with node string for multiple nodes
  - Test writer using t.Logf
  - Wait for nodes after starting all nodes
  - Log when node is ready only
  - Init nodes in parallel
- split `tests/utils` package
  - `tests/utils/rpc` package
  - `tests/utils/websocket` package
  - `tests/utils/config` package
- remove `DecodeRPC_NT`
- remove `HOSTNAME` variable
- inline RPC method strings
- remove global variable `GenesisTwoAuthsSecondaryVRF0_9_10`
- remove global configuration variables
  - write no auth config to temp dir per test
  - write no grandpa config to temp dir per test
  - write no babe config to temp dir per test
  - write grandpa config to temp dir per test
  - write default config to temp dir per test
- remove global framework variable
- remove global variables `GenesisDefault` and `GenesisDev`
- Add `GenerateGenesisAuths` function
- use `t.TempDir()` for base paths
- `waitForNode` context based function (#2390)
- make RPC helpers not test aware (#2389)
- make `DecodeRPC` not test aware (#2389)

## Tests

- RPC suite passing
- Stress tests passing

## Issues

See issues mentioned in changes above.

Indirectly fixed: #2165


tests/utils/config/config.go
```diff
@@ -0,0 +1,51 @@
...
+func NoGrandpa() (cfg toml.Config) {
+	cfg = Default()
+	cfg.Core.GrandpaAuthority = false
+	cfg.Core.BABELead = true
+	cfg.Core.GrandpaInterval = 1
+	return cfg
+}
+
+// NotAuthority generates an non-authority config.
+func NotAuthority() (cfg toml.Config) {
```
timwu20, line 45: nit:
```suggestion
func NotAuthorities() (cfg toml.Config) {
```

tests/utils/node/nodes.go
```diff
@@ -0,0 +1,205 @@
...
+
+		// Set defaults using index `i`
+		nodes[i].tomlConfig.Core.BABELead = i == 0
+
+		if nodes[i].index == nil {
+			nodes[i].index = intPtr(i)
+		}
+
+		// Set node defaults on the remaining unset fields
+		nodes[i].setDefaults(t)
```
timwu20, line 42: should this use `node.NewNode`

tests/utils/node/node.go
```diff
@@ -0,0 +1,275 @@
...
+}
+
+// GetRPCPort returns the rpc port of the node.
+func (n Node) GetRPCPort() (port string) { return fmt.Sprint(n.tomlConfig.RPC.Port) }
+
+// GetWSPort returns the websocket port of the node.
+func (n Node) GetWSPort() (port string) { return fmt.Sprint(n.tomlConfig.RPC.WSPort) }
+
+// GetKey returns the key of the node.
+func (n Node) GetKey() (key string) { return n.tomlConfig.Account.Key }
```
timwu20, line 58: nit: remove the `Get`

tests/utils/node/node.go
```diff
@@ -0,0 +1,275 @@
...
+	return nil
+}
+
+// StartAndWait starts a Gossamer node using the node configuration of
+// the receiving struct. It returns a start error if the node cannot
+// be started, and runs the node until the context gets canceled.
+// When the node crashes or is stopped, an error (nil or not) is sent
+// in the waitErrCh.
+// It waits for the node to respond to an RPC health call before returning.
+func (n *Node) StartAndWait(ctx context.Context, waitErrCh chan<- error) (startErr error) {
```
timwu20, line 167: Usually the function should return a `chan error`.  As a best practice, producers should return channels.

tests/utils/rpc/dev.go
```diff
@@ -0,0 +1,76 @@
...
+
+	"github.com/ChainSafe/gossamer/lib/common"
+)
+
+// PauseBABE calls the endpoint dev_control with the params ["babe", "stop"]
+func PauseBABE(ctx context.Context, rpcPort string) error {
+	endpoint := NewEndpoint(rpcPort)
+	const method = "dev_control"
+	const params = `["babe", "stop"]`
+	_, err := Post(ctx, endpoint, method, params)
```
timwu20, line 20: @EclesioMeloJunior babe can be controlled by RPC?

tests/utils/chain.go
```diff
@@ -7,132 +7,123 @@ import (
...
+	if err != nil {
+		return nil, fmt.Errorf("cannot decode RPC response body: %w", err)
+	}
 
-	stateRoot, err := common.HexToHash(header.StateRoot)
-	require.NoError(t, err)
+	rpcHeader := rpcBlock.Block.Header
+	header, err := headerResponseToHeader(rpcHeader)
+	if err != nil {
+		return nil, fmt.Errorf("malformed RPC header: %w", err)
```
EclesioMeloJunior: I think the variable `header` is not from the RPC but from the block returned by the RPC call, so the message error could be `malformed block header: %w`

tests/utils/chain.go
```diff
@@ -7,132 +7,123 @@ import (
...
+	}
 
-	extrinsicsRoot, err := common.HexToHash(header.ExtrinsicsRoot)
-	require.NoError(t, err)
+	body, err := types.NewBodyFromExtrinsicStrings(rpcBlock.Block.Body)
+	if err != nil {
+		return nil, fmt.Errorf("cannot create body from RPC block blody: %w", err)
```
EclesioMeloJunior: fmt.Errorf("cannot create body from RPC block <strike>blody</strike>: %w", err)
fmt.Errorf("cannot create body from RPC block **body**: %w", err)

EclesioMeloJunior (changes requested):

tests/utils/rpc/dev.go
EclesioMeloJunior: @timwu20 yes! In the `./dot/rpc/modules/dev.go` we expose the method `Control` which calls the `blockProducerAPI.Pause()/Resume()`

tests/utils/websocket/decode.go
```diff
@@ -0,0 +1,51 @@
+// Copyright 2022 ChainSafe Systems (ON)
+// SPDX-License-Identifier: LGPL-3.0-only
+
+package websocket
```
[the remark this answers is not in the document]
qdm12: Well it did, and the decode function was also a bit different. But it might just work, I merged them and removed the `websocket` package. The CI seems to pass so let's just assume this is resolved :wink:

tests/stress/grandpa_test.go
```diff
@@ -42,51 +43,47 @@ func TestStress_Grandpa_OneAuthority(t *testing.T) {
...
+		const retryWait = time.Second
+		fin, err := compareFinalizedHeadsWithRetry(ctx,
+			nodes, uint64(i), getFinalizedHeadByRoundTimeout, retryWait)
 		require.NoError(t, err)
 		t.Logf("finalised hash in round %d: %s", i, fin)
 	}
 }
 
 func TestStress_Grandpa_SixAuthorities(t *testing.T) {
 	t.Skip()
```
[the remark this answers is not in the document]
qdm12: I tried to unskip it, but it hangs forever. I'll create an issue for it.

tests/utils/chain.go
qdm12: Bloody body! Fixed :+1:

tests/utils/node/node.go
qdm12: I ended up making the changes, returning error channels instead of injecting them. It made things a bit more complicated, I had to add a 'errors fan in' object to handle stuff in `tests/utils/runtime/errors.go` but it did make responsibilities clearer, for example a consumer CANNOT close a channel (shame on my previous self!).

tests/utils/node/nodes.go
qdm12: Done 👍 
Now it always sets each node to have their 'index' as the index of the nodes slice, so the `SetIndex` option is ignored. We don't use anywhere so all good as well.

tests/rpc/rpc_01-system_test.go
```diff
@@ -9,137 +9,194 @@ import (
...
+
+	t.Run("system_nodeRoles", func(t *testing.T) {
+		t.Parallel()
+		t.Skip("test not implemented")
+	})
+
+	t.Run("system_accountNextIndex", func(t *testing.T) {
+		t.Parallel()
+		t.Skip("test not implemented")
+	})
```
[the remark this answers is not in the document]
qdm12: Done in #2546 (Dan created sub-issues as well)

tests/stress/grandpa_test.go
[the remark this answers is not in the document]
qdm12: Created #2545

tests/stress/stress_test.go
```diff
@@ -61,127 +49,140 @@ func TestMain(m *testing.M) {
...
-	ctx := context.Background()
 	const getChainHeadTimeout = time.Second
 
-	err = compareChainHeadsWithRetry(ctx, t, nodes, getChainHeadTimeout)
+	err := compareChainHeadsWithRetry(ctx, nodes, getChainHeadTimeout)
 	require.NoError(t, err)
 }
 
 func TestSync_MultipleEpoch(t *testing.T) {
 	t.Skip("skipping TestSync_MultipleEpoch")
```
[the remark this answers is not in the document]
qdm12: Created #2544

tests/stress/stress_test.go
```diff
@@ -334,98 +340,138 @@ func TestSync_Restart(t *testing.T) {
...
 		time.Sleep(time.Second * 5)
 	}
-	close(done)
+
+	mainCancel()
+	<-onOffRoutineDone
 }
 
 func TestSync_SubmitExtrinsic(t *testing.T) {
 	t.Skip()
```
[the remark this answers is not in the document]
qdm12: Created #2542

tests/utils/node/node.go
qdm12: I'll add it to the engineering wiki Go page

tests/stress/stress_test.go
```diff
@@ -61,127 +49,137 @@ func TestMain(m *testing.M) {
...
+	ctx, cancel := context.WithCancel(context.Background())
 
-	err = utils.StartNodes(t, nodes)
-	require.NoError(t, err)
+	runtimeErrors, startErr := nodes.Start(ctx)
+	if startErr != nil {
+		cancel()
+		for _, runtimeError := range runtimeErrors {
+			<-runtimeError
+		}
```
timwu20, line 72: could you put this in a helper if you don't actually care?

tests/utils/runtime/errors.go
```diff
@@ -0,0 +1,156 @@
...
+		nodeToRuntimeError: make(map[string]<-chan error),
+		nodeToFaninCancel:  make(map[string]context.CancelFunc),
+		nodeToFaninDone:    make(map[string]<-chan struct{}),
+		fifo:               make(chan nodeError),
+	}
+}
+
+// Add adds a runtime error receiving channel to the fan in mechanism
+// for the particular node string given. Note each node string must be
+// unique or the code willp panic.
```
timwu20, line 44:
```suggestion
// unique or the code will panic.
```

tests/utils/node/nodes.go
```diff
@@ -0,0 +1,177 @@
...
+		t.Helper()
+		// Disengage node watchdog goroutine
+		watchDogCancel()
+		<-watchDogDone
+		// Stop and wait for nodes to exit
+		stopNodes(t, nodesCancel, runtimeErrors)
+	})
+}
+
+func stopNodes(t *testing.T, nodesCancel context.CancelFunc,
```
timwu20, line 165: curious as to why this isn't just a receiver function on `Nodes`?

tests/utils/node/nodes.go
```diff
@@ -0,0 +1,177 @@
...
+		err := waitForNode(ctx, node.RPCPort())
+		if err == nil {
+			t.Logf("Node %s is ready", node)
+			continue
+		}
+
+		t.Errorf("Node %s failed to be ready: %s", node, err)
+		stopNodes(t, nodesCancel, runtimeErrors)
+		t.FailNow()
+	}
```
timwu20, line 134: why not just for the nodes in parallel and use a waitgroup?

tests/utils/node/nodes.go
timwu20, line 165: curious as to why this isn't a receiver function on `Nodes`?

tests/utils/retry/common.go
```diff
@@ -0,0 +1,28 @@
...
+
+package retry
+
+import (
+	"context"
+	"fmt"
+	"time"
+)
+
+func waitAfterFail(ctx context.Context, retryWait time.Duration,
```
timwu20, line 12: I find all this decorating of contexts very difficult to follow.

tests/utils/rpc/models.go
```diff
@@ -0,0 +1,30 @@
+// Copyright 2022 ChainSafe Systems (ON)
+// SPDX-License-Identifier: LGPL-3.0-only
+
```
timwu20, line 3: I would name this file `types.go`.

tests/utils/runtime/errors.go
```diff
@@ -0,0 +1,156 @@
...
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"sync"
+	"time"
+)
+
+// ErrorsFanIn takes care of fanning runtime errors from
```
timwu20, line 14: Can you explain why this needs to be exported? I would have assumed the `Nodes` type just handles this internally.

Dockerfile
```diff
@@ -17,11 +17,6 @@ RUN wget -qO- https://deb.nodesource.com/setup_14.x | bash - && \
 RUN wget -O /usr/local/bin/subkey https://chainbridge.ams3.digitaloceanspaces.com/subkey-v2.0.0 && \
     chmod +x /usr/local/bin/subkey
 
-# Polkadot JS dependencies
-WORKDIR /go/src/github.com/ChainSafe/gossamer/tests/polkadotjs_test
-COPY tests/polkadotjs_test/package.json tests/polkadotjs_test/package-lock.json ./
-RUN npm install
-
```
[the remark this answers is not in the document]
qdm12: No it's done in Go code at

https://github.com/ChainSafe/gossamer/blob/qdm12/tests-refactor-tip/tests/polkadotjs_test/start_polkadotjs_test.go#L28-L41

when the (single) test needing it is run

tests/utils/node/nodes.go
qdm12: See the comment

```go
	// this is run sequentially since all nodes start almost at the same time
	// so waiting for one node will also wait for all the others.
	// You can see this since the test logs out that all the nodes are ready
	// at the same time.
```

We could do it in parallel, but it adds quite a bit of complexity for no practical advantage.

tests/utils/node/nodes.go
qdm12: Because:

- all the arguments for this function are not part of the `Nodes`
- the calling tests do not need to stop multiple nodes so far

So I don't think there is a need to have it as a method. If one day a test needs to stop multiple nodes, then yeah we should move it as a receiver.

tests/utils/retry/common.go
qdm12: I understand you're not a fan of contexts (yet hehe!), I also like to keep it simple with signal channels.

But in this case we have different triggers that would cancel the context (aka close the channel), which can be the `cancel` function or the timer(s). Having these with a signal channel, multiple nested timers and select blocks would make it unreadable and needlessly complex. Even if context would not be part of std lib, I might consider using them in those cases 😉

tests/utils/runtime/errors.go
qdm12: I moved it to `utils/nodes` as unexported.

It was the Nodes initially handling this internally but that got messy rather quickly, so I split it away in that struct.
I then moved it to its own package to avoid polluting the namespace in utils/nodes, but that might had been overkill indeed.

tests/stress/stress_test.go
qdm12: Done!

tests/utils/retry/common.go
timwu20: This function is called by `retry.UntilOk(ctx context.Context, poll time.Duration, func() (ok bool, err error)`.  If you made the signature of this function `retry.UntilOk(timeout time.Duration, poll time.Duration, func() (ok bool, err error)`, you could just handle the timeout and deadline internally without having to worry about passing in a context.   

In your tests you declare a global `ctx` before using this, and in the function you pass to `retry.UntilOK`, you add a timeout to that context using `context.WithTimeout`, and you do it again in `waitAfterFail`.  Not sure how this approach of decorating contexts from a parent to handle long running processes is easier to read.  I just want to reiterate, that contexts should be request scoped, and not for long running processes.  It would make sense to use a context to facilitate cancellation and deadline for an RPC call.  But cancellation shouldn't be confused with explicit stopping of long running services/goroutines.

Taken from Dave Cheney's blog: 
> However, context.Context‘s most important facility, broadcasting a cancellation signal, is incomplete as there is no way to wait for the signal to be acknowledged.
This is the problem when conflating cancellation and explicit stopping.  There's no way for us to know if services have actually stopped if you use contexts for stopping long running processes.

timwu20 (approved):

tests/utils/retry/common.go
qdm12: > you could just handle the timeout and deadline internally without having to worry about passing in a context.

No because the top level context is used for nodes and is canceled when one of them crashes. So it's a mix of timeout and event driven, which would be very tedious to implement with channels and timers only, especially considering the different timeouts values here and there, there would be a lot more select cases all over the place.

> But cancellation shouldn't be confused with explicit stopping of long running services/goroutines.

At the end of the day, you do need a signal channel (internally or not) to signal to stop a long running goroutine. Contexts are just here to ease mixing event driven and time driven cancellation, without having to create a bunch of sub-signal-channels and sub-timers.

> is incomplete as there is no way to wait for the signal to be acknowledged.

Yes, but this is a bit out of scope; it is the same when closing a channel to stop a goroutine, you need another signal channel to wait for the goroutine to exit (and I agree this sucks too).
Such signal channels are implemented (or it's registered in `t.Cleanup`) in those tests.

PR #2472, opened 2022-04-07
fix(dot/digest): verify if next epoch already contains some definition

### LINKED ISSUE #2471: check if the BABE next epoch is already persisted
## Describe the bug

In the `dot/digest` package we listen for finalized headers and when a new finalized header arrives we execute the function `persistBABEDigestsForNextEpoch(finalizedHeader *types.Header)`. This function retrieves the epoch for the finalized header add `1` and uses the result as the `nextEpoch` and then call the `h.epochState.FinalizeBABENextEpochData(nextEpoch)` and `h.epochState.FinalizeBABENextConfigData(nextEpoch)` in order to persist the next epoch BABE data and config and remove the in memory data.

The problem is that when another header is finalized in the same epoch we call the same functions again but this time they will fail as there is no data to persist in the database.

A possible solution is to check if the `nextEpoch` already has defined some data in the database and if so we skip the respective persistency step

EclesioMeloJunior:
## Changes

- Add `epochState.AlreadyDefined(epoch uint64)` method that verifies if the epoch already contains the epoch or config data
- Use the previous method in the `persistBABEDigestsForNextEpoch` to avoid persistence many times in the same epoch

## Tests

Test the `dot/digest` and `dot/state` pkgs
```sh
go test github.com/ChainSafe/gossamer/dot/digest
go test github.com/ChainSafe/gossamer/dot/state
```

## Issues

closes #2471

## Primary Reviewer

@timwu20


dot/state/epoch_test.go
```diff
@@ -591,12 +597,24 @@ func TestStoreAndFinalizeBabeNextConfigData(t *testing.T) {
 			err := epochState.blockState.db.Put(headerKey(tt.finalizeHash), []byte{})
 			require.NoError(t, err)
 
+			// before finalize next config data this epoch should not be defined
```
qdm12, line 600: nit maybe a word missing I think?

```suggestion
			// before finalize next config data for this epoch should not be defined
```

dot/state/epoch_test.go
```diff
@@ -591,12 +597,24 @@ func TestStoreAndFinalizeBabeNextConfigData(t *testing.T) {
...
+			require.False(t, definedEpochData)
+			require.False(t, definedConfigData)
+
 			err = epochState.FinalizeBABENextConfigData(tt.finalizeEpoch)
 			if tt.expectErr != nil {
 				require.ErrorIs(t, err, tt.expectErr)
 			} else {
 				require.NoError(t, err)
 
+				// after finalize next config data this epoch should be defined
```
qdm12, line 612: nit maybe a word missing I think?

```suggestion
				// after finalize next config data for this epoch should be defined
```

dot/state/epoch_test.go
```diff
@@ -591,12 +597,24 @@ func TestStoreAndFinalizeBabeNextConfigData(t *testing.T) {
...
+				definedEpochData, definedConfigData, err = epochState.AlreadyDefined(tt.finalizeEpoch)
+				require.NoError(t, err)
+				require.False(t, definedEpochData)
```
qdm12, line 615: This is meant to be false right? (after finalize)

dot/state/epoch_test.go
```diff
@@ -401,12 +401,24 @@ func TestStoreAndFinalizeBabeNextEpochData(t *testing.T) {
...
+			require.False(t, definedEpochData)
+			require.False(t, definedConfigData)
+
 			err = epochState.FinalizeBABENextEpochData(tt.finalizeEpoch)
 			if tt.expectErr != nil {
 				require.ErrorIs(t, err, tt.expectErr)
 			} else {
 				require.NoError(t, err)
 
+				// after finalize next epoch data this epoch should not defined
```
qdm12, line 416: nit I think `for` is missing here?

```suggestion
				// after finalize next epoch data for this epoch should not defined
```

dot/state/epoch_test.go
```diff
@@ -401,12 +401,24 @@ func TestStoreAndFinalizeBabeNextEpochData(t *testing.T) {
 			err := epochState.blockState.db.Put(headerKey(tt.finalizeHash), []byte{})
 			require.NoError(t, err)
 
+			// before finalize next epoch data this epoch should not be defined
```
qdm12, line 404: nit I think `for` is missing here?

```suggestion
			// before finalize next epoch data for this epoch should not be defined
```

dot/state/epoch_test.go
```diff
@@ -226,10 +226,10 @@ func TestEpochState_GetEpochFromTime(t *testing.T) {
...
 }
 
-type inMemoryNextEpochData struct {
-	epoch          uint64
-	hashes         []common.Hash
-	nextEpochDatas []types.NextEpochData
+type inMemoryBABEData[T any] struct {
+	epoch  uint64
+	hashes []common.Hash
+	Data   []T
```
qdm12, line 232: nit Can we have `Data` unexported as `data`?

Also generics are going wild :tiger: !!

dot/state/epoch.go
```diff
@@ -512,6 +512,30 @@ func (s *EpochState) SkipVerify(header *types.Header) (bool, error) {
 	return false, nil
 }
 
+// AlreadyDefined will true for each returned value if it is already defined in the database for the given epoch
+func (s *EpochState) AlreadyDefined(epoch uint64) (epochData bool, configData bool, err error) {
+	keys := [...][]byte{epochDataKey(epoch), configDataKey(epoch)}
+	applied := [...]bool{false, false}
```
qdm12, line 518: Why use arrays? Can't we use some `epochDataKeyByte` and `configDataKeyByte` variables instead, and update `epochData` and `configData` without a for loop?

That would be a bit longer code, but quite easier to follow/read I think.

Also I would think splitting this function in two, one for epochData and one for configData would be nice (without using generics lol).

dot/digest/interface.go
```diff
@@ -18,6 +18,7 @@ type BlockState interface {
 	FreeFinalisedNotifierChannel(ch chan *types.FinalisationInfo)
 }
 
+//go:generate mockgen -destination=mock_epoch_state_test.go -package $GOPACKAGE . EpochState
```
qdm12, line 21: add an empty line so it's not part of `EpochState`'s comment

```suggestion
//go:generate mockgen -destination=mock_epoch_state_test.go -package $GOPACKAGE . EpochState

```

dot/digest/digest_test.go
```diff
@@ -515,3 +516,73 @@ func TestHandler_HandleNextConfigData(t *testing.T) {
 	require.NoError(t, err)
 	require.Equal(t, act.ToConfigData(), stored)
 }
+
+func Test_persistBABEDigestsForNextEpoch(t *testing.T) {
+	type testStruct struct {
+		defineNextEpochData          bool
+		defineNextConfigData         bool
+		expectedErr                  error
+		alreadyDefinedFuncThrowError bool
```
qdm12, line 525: nit you could define your mocks using functional fields if you want. [example](https://gist.github.com/qdm12/99358e2529a5887ae13a936a7f1c8ae5#file-potato_test-go-L230)

dot/digest/digest.go
```diff
@@ -267,23 +265,29 @@ func (h *Handler) persistBABEDigestsForNextEpoch(finalizedHeader *types.Header)
...
 	}
 
-	err = h.epochState.FinalizeBABENextConfigData(nextEpoch)
-	if err == nil {
-		return nil
-	} else if errors.Is(err, state.ErrEpochNotInMemory) {
-		return fmt.Errorf("%w: %s", ErrDefineNextEpoch, err)
+	if !appliedEpochData {
+		err = h.epochState.FinalizeBABENextEpochData(nextEpoch)
+		if err != nil && !errors.Is(err, state.ErrEpochNotInMemory) {
```
qdm12, line 276: What do we do in the case where `errors.Is(err, state.ErrEpochNotInMemory)`? Should we just continue execution?

dot/digest/digest.go
```diff
@@ -267,23 +265,29 @@ func (h *Handler) persistBABEDigestsForNextEpoch(finalizedHeader *types.Header)
...
+				finalizedHeader.Number, finalizedHeader.Hash(), err)
+		}
 	}
 
-	// the epoch state does not contains any information about the next epoch
-	return fmt.Errorf("cannot finalize babe next config data for block number %d (%s): %w",
-		finalizedHeader.Number, finalizedHeader.Hash(), err)
+	if !appliedConfigData {
+		err = h.epochState.FinalizeBABENextConfigData(nextEpoch)
+		if err != nil && !errors.Is(err, state.ErrEpochNotInMemory) {
```
qdm12, line 284: What do we do in the case where `errors.Is(err, state.ErrEpochNotInMemory)`? Should we just continue execution?

dot/state/epoch_test.go
EclesioMeloJunior: Yeah, we finalize only the `NextConfigData` so `definedEpochData` should be false

dot/digest/digest_test.go
EclesioMeloJunior: Nice suggestion, shrink the test conditions

dot/digest/digest.go
```diff
@@ -267,23 +265,29 @@ func (h *Handler) persistBABEDigestsForNextEpoch(finalizedHeader *types.Header)
...
+			return fmt.Errorf("cannot finalize babe next config data for block number %d (%s): %w",
+				finalizedHeader.Number, finalizedHeader.Hash(), err)
+		}
```
noot, line 287: not every epoch will have `ConfigData`, if it doesn't, it should just use the last set config data. does this handle that case? it seems to me  like it doesn't

dot/digest/digest.go
```diff
@@ -267,23 +265,29 @@ func (h *Handler) persistBABEDigestsForNextEpoch(finalizedHeader *types.Header)
...
 	nextEpoch := currEpoch + 1
-	err = h.epochState.FinalizeBABENextEpochData(nextEpoch)
-	if err != nil && !errors.Is(err, state.ErrEpochNotInMemory) {
-		return fmt.Errorf("cannot finalize babe next epoch data for block number %d (%s): %w",
-			finalizedHeader.Number, finalizedHeader.Hash(), err)
+
+	appliedEpochData, appliedConfigData, err := h.epochState.AlreadyDefined(nextEpoch)
+	if err != nil {
+		return fmt.Errorf("cannot check if next epoch is already defined: %w", err)
 	}
```
noot, line 272: I'm kind of confused still as to why this is needed - a block won't be finalized twice, so what is the case where the digest data would already be set in the database? also, it would be cleaner to handle the digest check + writing to disk all the in `EpochState` instead of here.

dot/digest/digest.go
noot: I think this function should only be called if `finalizedHeader` is the first block in its epoch, as that's the only block that would contain `NextEpoch` or `NextConfig` digests. that would also prevent the spamming

EclesioMeloJunior: But, what about the case of the epoch 1 started at block `N` but the finalized block is the block `N + 2`, we will not persist the next epoch data because the finalized block is not the first in its epoch.

dot/state/epoch.go
```diff
@@ -573,24 +591,45 @@ func (s *EpochState) FinalizeBABENextEpochData(epoch uint64) error {
...
+		return fmt.Errorf("cannot check if next epoch config is already defined for epoch %d: %w", nextEpoch, err)
+	}
+
+	// config already defined we don't need to lookup in the map
+	if configInDatabase != nil {
+		return nil
+	}
+
+	// not every epoch will have `ConfigData`
+	finalizedNextConfigData, err := lookupForNextEpochPersistedHash(s.nextConfigData, s, nextEpoch)
```
noot, line 617: can we rename this function `lookupForNextEpochPersistedHash` to something like `findNextInfoFromMemory`?

noot (approved):
looks good!

dot/state/epoch.go
EclesioMeloJunior: This function checks in the memory for the hash which is persisted in the database, what about `findFinalizedHeaderForEpoch`? meaning that we need to have in the epoch map one hash that is persisted in the database.

dot/state/epoch_test.go
```diff
@@ -539,38 +557,11 @@ func TestStoreAndFinalizeBabeNextConfigData(t *testing.T) {
...
-							C1:             3,
-							C2:             4,
-							SecondarySlots: 0,
-						},
-					},
-				},
-			},
+		"in_memory_config_not_found_shouldnt_return_error": {
+			shouldRemainInMemory: 0,
+			finalizeEpoch:        1, // try to finalize a epoch that does not exists
```
qdm12, line 562: supernit

```suggestion
			finalizeEpoch:        1, // try to finalize an epoch that does not exist
```

dot/state/epoch_test.go
```diff
@@ -422,33 +423,50 @@ func TestStoreAndFinalizeBabeNextEpochData(t *testing.T) {
...
+		types.PreRuntimeDigest{}))
+
+	require.NoError(t, digest.Add(*preRuntimeDigest))
+
+	// a random finalized header for testing purposes
+	finalizedHeader := &types.Header{
+		ParentHash: common.Hash{},
+		Number:     1,
+		Digest:     digest,
+	}
```
qdm12, line 446: nit it's not really random, is it? :thinking:

qdm12 (approved):
:+1:

dot/state/epoch.go
```diff
@@ -573,35 +591,56 @@ func (s *EpochState) FinalizeBABENextEpochData(epoch uint64) error {
...
+	}
+
+	// config already defined we don't need to lookup in the map
+	if configInDatabase != nil {
+		return nil
+	}
+
+	// not every epoch will have `ConfigData`
+	finalizedNextConfigData, err := findFinalizedHeaderForEpoch(s.nextConfigData, s, nextEpoch)
+	if errors.Is(err, ErrEpochNotInMemory) {
```
kishansagathiya, line 618: Are we likely to come across `ErrEpochNotInMemory` a lot?

dot/state/epoch_test.go
kishansagathiya: I am starting to see why go did not have generics. The code without them looks so much better!

dot/state/epoch.go
```diff
@@ -540,11 +540,29 @@ func (s *EpochState) StoreBABENextConfigData(epoch uint64, hash common.Hash, nex
...
-	finalizedNextEpochData, err := lookupForNextEpochPersistedHash(s.nextEpochData, s, epoch)
+	finalizedBlockEpoch, err := s.GetEpochForBlock(finalizedHeader)
+	if err != nil {
+		return fmt.Errorf("cannot get epoch for block %d (%s): %w",
+			finalizedHeader.Number, finalizedHeader.Hash(), err)
+	}
+
+	nextEpoch := finalizedBlockEpoch + 1
+
+	epochInDatabase, err := s.getEpochDataInDatabase(nextEpoch)
```
kishansagathiya, line 555: better to rename this as `getEpochDataFromDatabase`

dot/state/epoch.go
```diff
@@ -540,11 +540,29 @@ func (s *EpochState) StoreBABENextConfigData(epoch uint64, hash common.Hash, nex
...
+	if err != nil && !errors.Is(err, chaindb.ErrKeyNotFound) {
+		return fmt.Errorf("cannot check if next epoch data is already defined for epoch %d: %w", nextEpoch, err)
+	}
+
+	// config already defined we don't need to lookup in the map
```
kishansagathiya, line 560: do you mean epoch data already defined (and not config)?
because there is also a config data which also gets written to the db

dot/state/epoch.go
```diff
@@ -573,35 +591,56 @@ func (s *EpochState) FinalizeBABENextEpochData(epoch uint64) error {
...
-	finalizedNextConfigData, err := lookupForNextEpochPersistedHash(s.nextConfigData, s, epoch)
+	finalizedBlockEpoch, err := s.GetEpochForBlock(finalizedHeader)
 	if err != nil {
+		return fmt.Errorf("cannot get epoch for block %d (%s): %w",
+			finalizedHeader.Number, finalizedHeader.Hash(), err)
+	}
+
+	nextEpoch := finalizedBlockEpoch + 1
+
+	configInDatabase, err := s.getConfigDataInDatabase(nextEpoch)
```
kishansagathiya, line 606: getConfigDataInDatabase -> getConfigDataFromDatabase

kishansagathiya (changes requested):

dot/state/epoch.go
```diff
@@ -573,35 +591,56 @@ func (s *EpochState) FinalizeBABENextEpochData(epoch uint64) error {
...
+	if err != nil && !errors.Is(err, chaindb.ErrKeyNotFound) {
+		return fmt.Errorf("cannot check if next epoch config is already defined for epoch %d: %w", nextEpoch, err)
+	}
+
```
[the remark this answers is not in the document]
EclesioMeloJunior: This is deliberate, I call the `getConfigDataInDatabase` and if I got the `chaindb.ErrKeyNotFound` error I don't need to return an error since this error tells me a key is not present in the database

dot/state/epoch.go
EclesioMeloJunior: I'm not sure but, in this case, we will always use the previous epoch config data.

dot/digest/digest.go
noot: right, yeah this should be fine then

dot/state/epoch.go
noot: the function doesn't return a header though so that's kind of confusing I think

kishansagathiya (approved):

dot/state/epoch.go
EclesioMeloJunior: Added a `logger.Debugf` inside this if condition

kishansagathiya (approved):

dot/state/epoch.go
```diff
@@ -573,35 +594,60 @@ func (s *EpochState) FinalizeBABENextEpochData(epoch uint64) error {
...
 		}
 	}
 
 	return nil
 }
 
-// lookupForNextEpochPersistedHash given a specific epoch (the key) will go through the hashes looking
+// findFinalizedHeaderForEpoch given a specific epoch (the key) will go through the hashes looking
 // for a database persisted hash (belonging to the finalized chain)
 // which contains the right configuration to be persisted and safely used
```
qdm12, line 649: Update comment

```suggestion
// which contains the right configuration or data to be persisted and safely used
```

dot/state/epoch_test.go
```diff
@@ -580,24 +571,32 @@ func TestStoreAndFinalizeBabeNextConfigData(t *testing.T) {
...
 				for i, hash := range e.hashes {
-					epochState.StoreBABENextConfigData(e.epoch, hash, e.nextConfigDatas[i])
+					epochState.StoreBABENextConfigData(e.epoch, hash, e.nextData[i])
 				}
 			}
 
 			require.Len(t, epochState.nextConfigData, len(tt.inMemoryEpoch))
 
-			expectedConfigData := epochState.nextConfigData[tt.finalizeEpoch][tt.finalizeHash]
+			// if there is no data in memory we try to finalize the net config data
```
qdm12, line 580:
```suggestion
			// if there is no data in memory we try to finalize the next config data
```

qdm12 (approved):
Just two small suggestions

PR #2474, opened 2022-04-08
fix(ci): fix staging workflow

timwu20:
## Changes

- Use `go install` instead of `go build`.  Something must have changed for default target path of `go build` in 1.18.

## Tests

Ensure staging workflow is able to build the container and deploy to staging

## Primary Reviewer

@EclesioMeloJunior

### DIFF
--- a/Dockerfile.staging
+++ b/Dockerfile.staging
@@ -20,7 +20,7 @@ RUN ["sh", "-c", "mv .github/workflows/staging/openmetrics.d/${chain}-conf.yaml 
 RUN service datadog-agent start
 
 RUN go get ./...
-RUN go build github.com/ChainSafe/gossamer/cmd/gossamer
+RUN go install -trimpath github.com/ChainSafe/gossamer/cmd/gossamer
 
 RUN ["sh", "-c", "gossamer init --chain=${chain}"]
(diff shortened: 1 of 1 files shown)


jimjbrettj (approved):

EclesioMeloJunior (approved):

Dockerfile.staging
```diff
@@ -20,7 +20,7 @@ RUN ["sh", "-c", "mv .github/workflows/staging/openmetrics.d/${chain}-conf.yaml
 RUN service datadog-agent start
 
 RUN go get ./...
-RUN go build github.com/ChainSafe/gossamer/cmd/gossamer
+RUN go install -trimpath github.com/ChainSafe/gossamer/cmd/gossamer
```
qdm12, line 23: That will build from the master branch from github.com, I don't think that's what we want?
Just add `-o gossamer` to the `go build` command instead?

qdm12 (changes requested):

Dockerfile.staging
timwu20: I validated this works.  I think it's cause we are currently in the same directory as `go.mod` which references this module as `github.com/ChainSafe/gossamer`

qdm12: Interesting, TIL, sorry about the wrong assumption!

qdm12 (approved):

PR #2480, opened 2022-04-14
chore(lint): simplify .golangci.yml

qdm12:
## Changes

- Set timeout to 10min, remove timeout flag from Makefile
- Remove copy pasted comments from example .golangci.yml
- Add URL at the top to the .golangci.yml configuration documentation webpage
- Remove linter settings for linters not enabled `dupl` and `gocritic`
- Remove default `fast: false`
- Limit to default 50 errors per linter, instead of no limit
- Limit to default 3 same issues, instead of no limit
- Remove `megacheck` linter which is no longer available
- Remove copyright notice (not really anything to protect)

## Tests

```sh
golangci-lint run
```

## Issues

For context, this was done whilst adding linting documentation to Chainsafe's engineering Wiki on Go.

## Primary Reviewer

@jimjbrettj

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,7 +20,7 @@ help: Makefile
 .PHONY: lint
 lint:
 	go install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.45
-	golangci-lint run --build-tags integration --timeout 10m
+	golangci-lint run
 
 clean:
 	rm -fr ./bin
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,26 +1,8 @@
-# Copyright 2019 ChainSafe Systems (ON) Corp.
-# This file is part of gossamer.
-#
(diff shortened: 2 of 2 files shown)


timwu20 (approved):

Makefile
```diff
@@ -20,7 +20,7 @@ help: Makefile
 .PHONY: lint
 lint:
 	go install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.45
-	golangci-lint run --build-tags integration --timeout 10m
+	golangci-lint run
```
jimjbrettj, line 23: We don't want to have the integration build tag?

jimjbrettj (approved):

Makefile
qdm12: It's in the .golangci.yml file :wink: 
No point having it twice :wink:

PR #2483, opened 2022-04-14
fix(dot/rpc): include unsafe flags to be considered by RPC layer

### LINKED ISSUE #2482: rpc unsafe flags are not exposing unsafe RPC endpoints
## Describe the bug

- Currently we contains `--rpc-unsafe`, `--rpc-unsafe-external`, `--ws-unsafe` and `--ws-unsafe-external` but those flags are not exposing the unsafe rpc endpoints

## Expected Behavior

- We should be able to expose unsafe endpoints when setting those falgs 

## Current Behavior

- When I set those flags and we try to call the `author.rotateKeys()` for example, I got `-32000: unsafe rpc method author_rotateKeys cannot be reachable: null`

EclesioMeloJunior:
## Changes

- RPC unsafe flags are not being placed in the RPC services, modified to use those flags values

## Tests

- Spins up a gossamer node with `--rpc-unsafe` and then connect the [polkadot.js.org](https://polkadot.js.org/apps/#/rpc) with the gossamer node and in the [rpc](https://polkadot.js.org/apps/#/rpc) tab you should be able to call the following methods: `system_addReservedPeer`, `system_removeReservedPeer`, `author_submitExtrinsic`, `author_removeExtrinsic`, `author_insertKey`, `author_rotateKeys`, `state_getPairs`, `state_getKeysPaged`, `state_queryStorage`.

## Issues

closes #2482

## Primary Reviewer

@edwardmack


tests/utils/gossamer_utils.go
```diff
@@ -119,6 +119,7 @@ func startGossamer(t *testing.T, node Node, websocket bool) (
 		"--rpcport", node.RPCPort,
 		"--rpcmods", "system,author,chain,state,dev,rpc",
 		"--rpc",
+		"--rpc-unsafe",
```
qdm12, line 122: Do we need this flag for end to end tests?

tests/utils/gossamer_utils.go
```diff
@@ -135,7 +136,7 @@ func startGossamer(t *testing.T, node Node, websocket bool) (
 	}
 
 	if websocket {
-		params = append(params, "--ws",
+		params = append(params, "--ws", "--ws-unsafe",
```
qdm12, line 139: Do we need this flag?

tests/utils/gossamer_utils.go
EclesioMeloJunior: yes, we call the `author_submitExtrinsic` which is unsafe

EclesioMeloJunior:
@qdm12 the problem was: if in the `config.toml` file the `usafe = true` but the flag is not defined in the `gossamer --rpc-unsafe` then we overwrite the config.toml value and set unsafe to false. I've fixed this behaviour now and if the flag is not defined it will use the value of the config toml without overwrite it

qdm12 (approved):

jimjbrettj (approved):
looks good!

PR #2484, opened 2022-04-14
chore(lib/babe): epoch error wrapping fixed up

qdm12:
## Changes

I just saw this ugliest error logged out on Dan's test runtime run, so I had to fix it:

- Error is returned only if it's not nil and not `chaindb.ErrKeyNotFound` in `HasEpochData` (it was logged as `%w (nil)` previously)
- Only `HasEpochData` wraps the epoch number context, not all above callers, to avoid repeating information

## Primary Reviewer

@EclesioMeloJunior

### DIFF
--- a/lib/babe/babe.go
+++ b/lib/babe/babe.go
@@ -399,7 +399,7 @@ func (b *Service) initiate() {
 func (b *Service) initiateAndGetEpochHandler(epoch uint64) (*epochHandler, error) {
 	epochData, err := b.initiateEpoch(epoch)
 	if err != nil {
-		return nil, fmt.Errorf("failed to initiate epoch %d: %w", epoch, err)
+		return nil, fmt.Errorf("failed to initiate epoch: %w", err)
 	}
 
 	logger.Debugf("initiated epoch with threshold %s, randomness 0x%x and authorities %v",
@@ -430,7 +430,7 @@ func (b *Service) runEngine() error {
 		if errors.Is(err, errServicePaused) || errors.Is(err, context.Canceled) {
 			return nil
 		} else if err != nil {
-			return err
+			return fmt.Errorf("cannot handle epoch: %w", err)
 		}
 
 		epoch = next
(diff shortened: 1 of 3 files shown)


EclesioMeloJunior (approved):

lib/babe/babe.go
```diff
@@ -399,7 +399,7 @@ func (b *Service) initiate() {
 func (b *Service) initiateAndGetEpochHandler(epoch uint64) (*epochHandler, error) {
 	epochData, err := b.initiateEpoch(epoch)
 	if err != nil {
-		return nil, fmt.Errorf("failed to initiate epoch %d: %w", epoch, err)
```
kishansagathiya, line 402: why remove epoch number?

lib/babe/babe.go
```diff
@@ -442,7 +442,7 @@ func (b *Service) handleEpoch(epoch uint64) (next uint64, err error) {
 	defer cancel()
 	b.epochHandler, err = b.initiateAndGetEpochHandler(epoch)
 	if err != nil {
-		return 0, fmt.Errorf("cannot initiate and get epoch handler for epoch %d: %w", epoch, err)
```
kishansagathiya, line 445: same

lib/babe/epoch.go
```diff
@@ -92,7 +92,7 @@ func (b *Service) getEpochDataAndStartSlot(epoch uint64) (*epochData, uint64, er
 
 	has, err := b.epochState.HasEpochData(epoch)
 	if err != nil {
-		return nil, 0, fmt.Errorf("cannot check for epoch data for epoch %d: %w", epoch, err)
```
kishansagathiya, line 95: same

kishansagathiya (changes requested):

lib/babe/babe.go
qdm12: Because it's added in the `initiateEpoch` method, so to avoid repeating the epoch number a bunch of times in the error.

Anyway, the function handling the argument should be the only one adding context to the error with it, not its callers.

lib/babe/babe.go
qdm12: There is no point having the epoch number repeated several times in the same error

lib/babe/epoch.go
qdm12: There is no point having the epoch number repeated several times in the same error

timwu20 (approved):

Issue #2398, opened 2022-03-18
Insufficient Garbage Collection for GRANDPA Message Tracker (GSR-05)


danforbes:
During the GRANDPA protocol, if vote or commit messages are received which we are unable to process, we add this block to the `tracker` for future processing. This may occur due to a block not existing in our block tree or that the round in a vote message is greater than our current round. Future processing is triggered when a new block is imported, the `tracker` then compares the block hash of the new block against the hash included in the messages. When the hashes match the message is passed back to the handler for processing.

There is a potential denial of service (DoS) attack that can exploit the storage of these messages. When a vote is received with a round greater than our current round it will be added to the `tracker` . An attacker is able to exploit this by repeatedly sending vote messages with rounds much higher than the current round. These votes will be stored in the tacker indefinitely if the hash is not from a block that will not be imported (e.g. a hash of random bytes). Thus, memory will continue to grow indefinitely as the attacker continues to sending malicious votes with high rounds. Each round must be different else it is an equivocatory vote, however `round` is a `uint64` which gives 2^64 different rounds to use.

Furthermore, if there are valid votes received from a future round while our round is behind and we have imported the block associated with the `Hash` , then these votes will be added to the `tracker` and never processed. This is due to the fact we only iterate through the `tracker` when we import a new block, checking the vote hash against the new block hash. Hence, votes for blocks already in the tree will never be processed.

The votes or commits which are stored in the `tracker` will be stored indefinitely when they have a Hash related to blocks that will never be imported. Since the commit messages are validated and thus will related to finalised blocks, it can be assumed that all commit messages will be processed eventually as all finalised blocks will be imported. However, due to race conditions between storing the messages in the maps and blocks being imported it is possible to have commit messages included in the maps which relate to blocks already imported. Furthermore, a malfunctioning or malicious node may send a vote with a block hash not related to a real block. These block hashes will not be processed and will remain in the `tracker` indefinitely.

The issue may be partially resolved by ignoring votes with the `Round` higher than our current `Round` . Valid nodes will periodically resend votes for the duration of a round, thus the current round votes will eventually be received. This can be implemented in `lib/grandpa/vote_message.go::validateMessage()`.

Also consider setting an expiry time for vote and commit messages that are stored in the `tracker`. A garbage collector routine may then be used to periodically iterate over the `voteMessages` and `commitMessages` maps and delete those that have expired.

PR #2485, opened 2022-04-15
fix(lib/grandpa): capped number of tracked vote messages

qdm12:
## Changes

- [x] Vote messages tracker
  - Removes oldest vote message when tracker capacity is reached
  - Efficient removal of multiple messages at any place in the tracker queue (linked list) if they get processed
  - Efficient removal of oldest message
  - Uses a bit more space to store each block hash + authority ID, for each vote message
  - Order is not modified for the same vote message (same block hash and authority id)
- [x] Discard vote messages for more than 1 round in the future from the state round (thanks [andresilva](https://github.com/andresilva)
- [x] Discard vote messages for more than 1 round in the past from the state round (thanks [andresilva](https://github.com/andresilva)
- [x] (A bit out-of-scope) Disable `addCatchUpResponse` (not implemented yet) to avoid a possible memory leak/abuse.
- [x] Full unit test coverage of new code
- [x] Create issues and comment with issue number
  - [x] Change reputation of peer for bad vote messages
- [ ] Do #2506 

## Tests

```sh
go test -tags integration github.com/ChainSafe/gossamer/lib/grandpa
```

## Issues

- #2398 

Another PR will be done to address the commit messages to close the issue for good.

## Primary Reviewer

@jimjbrettj (since you were curious about the linked list :smile:)


lib/grandpa/message_tracker.go
```diff
@@ -108,18 +102,17 @@ func (t *tracker) handleBlock(b *types.Block) {
...
+		_, err := t.handler.handleMessage(v.from, v.msg)
+		if err != nil {
+			logger.Warnf("failed to handle vote message %v: %s", v, err)
 		}
-
-		delete(t.voteMessages, h)
 	}
 
+	t.votes.delete(h)
+
```
qdm12, line 115: Comment here since I got self-confused reading the diffs;
- before we would delete only if the block hash existed
- now we delete the block hash from the votes tracker whether it exists or not

lib/grandpa/message_tracker.go
```diff
@@ -134,17 +127,16 @@ func (t *tracker) handleTick() {
 	t.mapLock.Lock()
 	defer t.mapLock.Unlock()
 
-	for _, vms := range t.voteMessages {
-		for _, v := range vms {
+	for _, voteMessage := range t.votes.getMessages() {
```
qdm12, line 130: nit to nit or not to nit

qdm12: nit to nit or not to nit

qdm12: nit to nit or not to nit

qdm12: nit to nit or not to nit

qdm12: nit to nit or not to nit

qdm12: nit to nit or not to nit

lib/grandpa/message_tracker.go
jimjbrettj: Thanks for this!

lib/grandpa/tracker_votes.go
```diff
@@ -0,0 +1,179 @@
...
+// not be processed, and removes the oldest ones once
+// its maximum capacity is reached.
+// It is NOT THREAD SAFE to use.
+type votesTracker struct {
+	// map of vote block hash to authority ID (ed25519 public Key)
+	// to data (peer id + message + tracking linked list element pointer)
+	mapping map[common.Hash]authorityIDToData
+	// double linked list of block hash + authority ID
+	// to track the order vote messages were added in.
+	linkedList *list.List
```
jimjbrettj, line 24: nit: maybe a more specific name

lib/grandpa/tracker_votes.go
```diff
@@ -0,0 +1,179 @@
...
+		authorityID: authorityID,
+	}
+}
+
+// add adds a vote message to the vote message tracker.
+// If the vote message tracker capacity is reached,
+// the oldest vote message is removed.
+func (vt *votesTracker) add(peerID peer.ID, voteMessage *VoteMessage) {
+	signedMessage := voteMessage.Message
+	blockHash := signedMessage.Hash
```
jimjbrettj, line 66: The hashed signedMessage is the blockHash?

lib/grandpa/tracker_votes.go
```diff
@@ -0,0 +1,179 @@
...
+			msg:  data.message,
+		}
+		messages = append(messages, message)
+	}
+	return messages
+}
+
+// forEach runs the function `f` on each
+// peer id + message stored in the tracker.
+func (vt *votesTracker) forEach(
```
jimjbrettj, line 172: nice!

lib/grandpa/vote_message.go
```diff
@@ -126,51 +126,67 @@ func (s *Service) validateVoteMessage(from peer.ID, m *VoteMessage) (*Vote, erro
 	// check for message signature
 	pk, err := ed25519.NewPublicKey(m.Message.AuthorityID[:])
 	if err != nil {
+		// TODO: affect peer reputation
```
jimjbrettj, line 129: This we need to add logic for a reputation change? Should there be a corresponding issue for this?

lib/grandpa/vote_message.go
```diff
@@ -126,51 +126,67 @@ func (s *Service) validateVoteMessage(from peer.ID, m *VoteMessage) (*Vote, erro
...
+		// we overflowed below 0 so set the minimum to 0.
+		minRoundAccepted = 0
+	}
 
-			cm, err := s.newCommitMessage(header, m.Round)
-			if err != nil {
-				return nil, err
-			}
+	const maxRoundsAhead = 1
+	maxRoundAccepted := s.state.round + maxRoundsAhead
```
jimjbrettj, line 151: Do we need to consider overflow here, or will round not get high enough for that?

jimjbrettj (approved):
Nice work!

kishansagathiya (commented):
The code looks pretty good, but the structure looks a little complicated. I will spend some time and see if a simpler structure is possible. Especially some thing general purpose that can be used for commit messages, vote message and catch up message would be great.

Since they use different types of values and keys, it might be tough or not possible, but worth trying.

lib/grandpa/tracker_votes.go
qdm12: Any suggestion in mind?

Mine would be `linkedListOfBlockHashAndAuthorityID` hahaha but that feels like overkill-super-mega-long name.

lib/grandpa/tracker_votes.go
qdm12: It's not the hash of the signed message, but the block hash. Comment on signed message:

```go
// SignedMessage represents a block hash and number signed by an authority
```

I opened #2491 to rename the field to `BlockHash` to avoid any future confusion as well :wink:

lib/grandpa/tracker_votes.go
qdm12: For the sake of knowledge and discussion, here's a discussion on the similar `forEach` method for the commits tracker:

https://github.com/ChainSafe/gossamer/pull/2490#discussion_r856406631

lib/grandpa/vote_message.go
qdm12: @jimjbrettj yeah that's the two checkboxes I have in the PR description that are, well, still unchecked :smile: But good spot nonetheless!

@danforbes please let me know if you plan on creating the issues I mentioned to you on Slack, otherwise let me know and I can create them :+1: So I can link them in the comments in this PR.

lib/grandpa/vote_message.go
qdm12: Good question. I doubt the round would ever reach the max uint64. It also gets reset on authority set changes. The spec defines it as an `uint64` as well. I'll link this up on Slack see if anyone thinks anything on this.

qdm12:
@kishansagathiya Yeah it's being repeated in #2490 maybe using generics could solve it? :thinking: I'll take a stab at it sometime next week...

lib/grandpa/vote_message.go
qdm12: From Elizabeth:

> nah I don't think so. that's a LOT of rounds

lib/grandpa/message_tracker.go
```diff
@@ -134,18 +129,21 @@ func (t *tracker) handleTick() {
...
-			}
+			logger.Debugf("failed to handle vote message %v from peer id %s: %s", message, peerID, err)
+		}
 
-			if v.msg.Round < t.handler.grandpa.state.round && v.msg.SetID == t.handler.grandpa.state.setID {
-				delete(t.voteMessages, v.msg.Message.Hash)
-			}
+		if message.Round < t.handler.grandpa.state.round && message.SetID == t.handler.grandpa.state.setID {
+			blockHash := message.Message.Hash
+			blockHashesDone = append(blockHashesDone, blockHash)
```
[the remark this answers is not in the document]
qdm12: I prefer it as it was, especially given `.Hash` is not `.BlockHash` which is confusing.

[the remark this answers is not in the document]
qdm12: Actually now that `.Hash` is `.BlockHash` I have put it to how you suggested it :+1:

lib/grandpa/votes_tracker.go
```diff
@@ -0,0 +1,179 @@
...
+		element := vt.linkedList.PushFront(elementData)
+		data := voteMessageMapData{
+			peerID:  peerID,
+			message: voteMessage,
+			element: element,
+		}
+		vt.mapping[blockHash] = authorityIDToData{
+			authorityID: data,
+		}
+		return
```
timwu20, line 83: Isn't this code exactly the same as lines 100 to 108?

lib/grandpa/votes_tracker.go
```diff
@@ -0,0 +1,179 @@
...
+)
+
+// votesTracker tracks vote messages that could
+// not be processed, and removes the oldest ones once
+// its maximum capacity is reached.
+// It is NOT THREAD SAFE to use.
+type votesTracker struct {
+	// map of vote block hash to authority ID (ed25519 public Key)
+	// to data (peer id + message + tracking linked list element pointer)
+	mapping map[common.Hash]authorityIDToData
```
timwu20, line 21: could this be changed to `mapping map[common.Hash]map[*VoteMessage]PeerID`? That way we can remove `voteMessageMapData`, and just `list.PushFront` the actual `*VoteMessage`? You will have to update the maps accordingly, but you can lookup on the address of the `VoteMessage` to retrieve the `PeerID`.

lib/grandpa/message_tracker.go
```diff
@@ -76,10 +71,11 @@ func (t *tracker) addCommit(cm *CommitMessage) {
 	t.commitMessages[cm.Vote.Hash] = cm
 }
 
-func (t *tracker) addCatchUpResponse(cr *CatchUpResponse) {
+func (t *tracker) addCatchUpResponse(_ *CatchUpResponse) {
 	t.catchUpResponseMessageMutex.Lock()
 	defer t.catchUpResponseMessageMutex.Unlock()
-	t.catchUpResponseMessages[cr.Round] = cr
+	// uncomment when usage is setup properly, see #1531
+	// t.catchUpResponseMessages[cr.Round] = cr
```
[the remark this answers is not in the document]
qdm12: Because it's a memory leak as it is now, we never clean these up.
It's commented until it's properly implemented.

lib/grandpa/votes_tracker.go
qdm12: I changed it in the last commit c6ccd6aa11959e70b537157e324ca5eadfeae359 to be like the commits tracker where the mapping is block hash -> authority ID -> list element pointer, and the list element has the peer id + vote message. What do you think?

I think we want to dedup by block hash + authority ID, so using a vote message pointer as second key is not possible? :thinking:

lib/grandpa/votes_tracker.go
qdm12: Yes, I simplified it to avoid repeating it. Good catch :+1:

lib/grandpa/vote_message.go
qdm12: I pushed a commit with comments linking to the grandpa politeness issue now #2505

timwu20 (approved):

qdm12:
Merging this as is, I'm working on #2506 which I'll do as a separate PR.

Issue #2396, opened 2022-03-18
GRANDPA Does Not Verify Equivocatory Votes (GSR-03)


danforbes:
If a GRANDPA voter has been recorded as an equivocatory voter then their vote will not be verified. [The specifications](https://w3f.github.io/polkadot-spec/develop/sect-finality.html) state that if any of the conditions are not met then a vote is invalid.

• The block hash does not correspond to a valid block.
• The block is not an (eventual) descendant of a previously finalized block.
• The signature is invalid.
• The AuthorityID is not in the authority list for the set.

Refer to the sample function below that can be used to verify these votes and use it to ensure each equivocatory vote is valid for the functions `verifyCommitMessageJustification()`, `verifyPreCommitJustification()` and `VerifyBlockJustification()` otherwise disregard the vote.

```go
func (h *MessageHandler) verifyPreCommitJustification(msg *CatchUpResponse) error {
	auths := make([]AuthData, len(msg.PreCommitJustification))
	for i, pcj := range msg.PreCommitJustification {
		auths[i] = AuthData{AuthorityID: pcj.AuthorityID}
	}
	eqvVoters := getEquivocatoryVoters(auths)
	// verify pre-commit justification
	var count uint64
	for idx := range msg.PreCommitJustification {
		just := &msg.PreCommitJustification[idx]
		if _, ok := eqvVoters[just.AuthorityID]; ok {
			continue
		}
		err := h.verifyJustification(just, msg.Round, msg.SetID, precommit)
		if err != nil {
			continue
		}
		if just.Vote.Hash == msg.Hash && just.Vote.Number == msg.Number {
			count++
		}
	}
	if count+uint64(len(eqvVoters)) < h.grandpa.state.threshold() {
		return ErrMinVotesNotMet
	}
	return nil
}
```

kishansagathiya:
Any vote cast by an equivocatory voter is called equivocatory vote. And equivocatory votes are invalid (https://spec.polkadot.network/develop/sect-finality.html#defn-equivocation). 

So, wouldn't it be appropriate to not check any verifications on equivocatory votes?

noot:
@kishansagathiya no, you need to check the signatures. equivocatory votes are still used as part of the calculation for number of votes for a block (total # of equivocations is used)

PR #2486, opened 2022-04-15
fix(lib/grandpa): verify equivocatory votes in grandpa justifications

kishansagathiya:
## Tests

```sh
```

## Issues
Fixes https://github.com/ChainSafe/gossamer/issues/2396

## Primary Reviewer

@timwu20


danforbes:
Does this relate to #2396?

kishansagathiya:
depends on https://github.com/ChainSafe/gossamer/pull/2468

lib/grandpa/message_handler.go
```diff
@@ -299,20 +299,43 @@ func getEquivocatoryVoters(votes []AuthData) map[ed25519.PublicKeyBytes]struct{}
...
+		return fmt.Errorf("could not get highest finalised header: %w", err)
+	}
+
+	isDescendant, err := blockState.IsDescendantOf(highestHeader.Hash(), hash)
+	if err != nil {
+		return fmt.Errorf("could not check if %s is descendant of %s: %w", hash, highestHeader.Hash(), err)
+	}
+	if !isDescendant {
+		// rename this error
+		return errInvalidVoteBlock
```
qdm12, line 314: Reminder

lib/grandpa/message_handler.go
```diff
@@ -299,20 +299,43 @@ func getEquivocatoryVoters(votes []AuthData) map[ed25519.PublicKeyBytes]struct{}
...
+	return nil
+
+}
+
 func (h *MessageHandler) verifyCommitMessageJustification(fm *CommitMessage) error {
 	if len(fm.Precommits) != len(fm.AuthData) {
 		return ErrPrecommitSignatureMismatch
 	}
 
+	if fm.SetID != h.grandpa.state.setID {
```
qdm12, line 326: A bit out of scope, but why is it named `fm`??

lib/grandpa/message_handler.go
```diff
@@ -299,20 +299,43 @@ func getEquivocatoryVoters(votes []AuthData) map[ed25519.PublicKeyBytes]struct{}
...
+		return ErrSetIDMismatch
```
qdm12, line 327: nit add the expected and actual set id context

lib/grandpa/message_handler.go
```diff
@@ -299,20 +299,43 @@ func getEquivocatoryVoters(votes []AuthData) map[ed25519.PublicKeyBytes]struct{}
...
+	}
+
+	err := verifyIfDescendantOfHighestFinalisedBlock(h.blockState, fm.Vote.Hash)
+	if err != nil {
+		return err
```
qdm12, line 332: nit

```suggestion
		return fmt.Errorf("cannot verify ancestry of highest finalised block: %w", err)
```

lib/grandpa/message_handler.go
```diff
@@ -321,7 +344,7 @@ func (h *MessageHandler) verifyCommitMessageJustification(fm *CommitMessage) err
 
 		err := h.verifyJustification(just, fm.Round, h.grandpa.state.setID, precommit)
 		if err != nil {
-			logger.Errorf("could not verify justification: %s", err)
+			logger.Errorf("failed to verify justification for vote from authority id: %s, for block hash: %")
```
qdm12, line 347: This has no arguments being formatted

lib/grandpa/message_handler.go
```diff
@@ -531,6 +567,11 @@ func (s *Service) VerifyBlockJustification(hash common.Hash, justification []byt
 		return fmt.Errorf("already have finalised block with setID=%d and round=%d", setID, fj.Round)
 	}
 
+	err = verifyIfDescendantOfHighestFinalisedBlock(s.blockState, fj.Commit.Hash)
+	if err != nil {
+		return err
```
qdm12, line 572: nit maybe wrap the error

lib/grandpa/message_handler.go
```diff
@@ -299,20 +299,43 @@ func getEquivocatoryVoters(votes []AuthData) map[ed25519.PublicKeyBytes]struct{}
 	return eqvVoters
 }
 
+func verifyIfDescendantOfHighestFinalisedBlock(blockState BlockState, hash common.Hash) error {
```
EclesioMeloJunior, line 302: Why not use this as a method of `*MessageHandler` and use `h.blockState` instead of passing the reference of the `BlockState`?
```suggestion
func (h *MessageHandler) verifyIfDescendantOfHighestFinalisedBlock(hash common.Hash) error {
```

EclesioMeloJunior, line 302: I would like to suggest changing this method name to `IsDescendantOHighestFinalisedBlock` and instead of returning an `error` return `(bool, error)`, makes easier to understand what the function does IMO

lib/grandpa/message_handler.go
```diff
@@ -419,36 +451,40 @@ func (h *MessageHandler) verifyPreVoteJustification(msg *CatchUpResponse) (commo
...
 	for idx := range msg.PreCommitJustification {
 		just := &msg.PreCommitJustification[idx]
 
-		if _, ok := eqvVoters[just.AuthorityID]; ok {
-			continue
+		err = verifyBlockHashAgainstBlockNumber(h.blockState, just.Vote.Hash, uint(just.Vote.Number))
+		if err != nil {
+			if errors.Is(err, chaindb.ErrKeyNotFound) {
+				h.grandpa.tracker.addCatchUpResponse(msg)
+				logger.Infof("we might not have synced to the given block %s yet: %s", just.Vote.Hash, err)
```
EclesioMeloJunior, line 475: Less identantion:
```suggestion
		if errors.Is(err, chaindb.ErrKeyNotFound) {
				h.grandpa.tracker.addCatchUpResponse(msg)
				logger.Infof("we might not have synced to the given block %s yet: %s", just.Vote.Hash, err)
				continue
	        } else if err != nil {
				return err
	        }   
```

lib/grandpa/message_handler.go
qdm12: Yes but then isn't the boolean just going to be the same as `err == nil`?

lib/grandpa/message_handler.go
qdm12: :smile: I agree, although Tim previously suggested to have the `errors.Is` inside the `if err != nil` block... I would just keep it as it is I think :laughing:

lib/grandpa/message_handler.go
EclesioMeloJunior: Not really, there is a case where the hash is not a descendant of the highest finalized block, so in this case, there is no error we can return nil but for me does not sounds correct to return nil when the value is a bool

lib/grandpa/message_handler.go
timwu20: I do not find this easier to read.  If there's an error handle it within the block.  There shouldn't be multiple `if err != nil` lines.

EclesioMeloJunior: I do not find nested if's easy to read and usually what I suggest for long errors case is to use `switch err`. eg:
```go
switch err {
case ErrSome:
case ErrAnother:
case nil:
}
```

and for simple error checks using the:

```go
if errors.Is(err, ErrToCheck) {
     handle here...
} else err != nil {
     handle the other errors here
}
```

lib/grandpa/message_handler.go
kishansagathiya: no clue!

lib/grandpa/message_handler.go
kishansagathiya: renaming this to `errVoteBlockMismatch`. This is similar to one more error `ErrPrecommitBlockMismatch = errors.New("precommit block is not descendant of committed block")`

lib/grandpa/message_handler.go
kishansagathiya: It is used in service.VerifyBlockJustification as well! If I create a method I can't use `verifyIfDescendantOfHighestFinalisedBlock` there

kishansagathiya: Existing way is better.
In case verification fails we want to return error.

> Not really, there is a case where the hash is not a descendant of the highest finalized block, so in this case, there is no error and we can return nil but for me does not sounds correct to return nil when the right value is a bool

returning bool will just add extra code for no reason. In case of verification failure, I would have to convert the boolean value again to an error. So, returning nil like mentioned above is not an option.

lib/grandpa/message_handler.go
kishansagathiya: resolving, this does not seem to need any action

timwu20 (changes requested):

lib/grandpa/message_handler.go
```diff
@@ -299,20 +299,48 @@ func getEquivocatoryVoters(votes []AuthData) map[ed25519.PublicKeyBytes]struct{}
...
+
+	// isDescendant, err := blockState.IsDescendantOf(highestHeader.Hash(), hash)
+	// if err != nil {
+	// 	return fmt.Errorf("could not check if %s is descendant of %s: %w", hash, highestHeader.Hash(), err)
+	// }
+	// if !isDescendant {
+	// 	return errVoteBlockMismatch
+	// }
+
+	// return nil
```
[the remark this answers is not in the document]
kishansagathiya: horrible mistake that happened in last commit, removed it

EclesioMeloJunior (approved):

timwu20 (approved):

lib/grandpa/message_handler.go
kishansagathiya: done

PR #2490, opened 2022-04-20
fix(lib/grandpa): capped number of tracked commit messages

qdm12:
## Changes

- [x] Commit messages tracker
  - Removes oldest commit message when tracker capacity is reached
  - Efficient removal of multiple messages at any place in the tracker queue (linked list) if they get processed
  - Efficient removal of oldest message
  - Uses a bit more space to store each block hash (+2 pointers) for each commit message
  - Order is not modified for the same commit message (same block hash)
- [x] Full unit test coverage for the new commits tracker code

## Tests

```sh
go test -tags integration github.com/ChainSafe/gossamer/lib/grandpa
```

## Issues

#2398 

## Primary Reviewer

Will appoint someone soon


lib/grandpa/tracker_commits.go
```diff
@@ -0,0 +1,117 @@
...
+	blockHash := commitMessage.Vote.Hash
+
+	data, has := ct.mapping[blockHash]
+	if has {
+		// commit already exists so override the commit for the block hash;
+		// do not move the list element in the linked list to avoid
+		// someone re-sending the same commit message and going at the
+		// front of the list, hence erasing other possible valid commit messages
+		// in the tracker.
+		data.message = commitMessage
```
timwu20, line 55: when would `commitMessage` differ in this case?

lib/grandpa/tracker_commits.go
```diff
@@ -0,0 +1,117 @@
...
+
+	ct.linkedList.Remove(data.element)
+	delete(ct.mapping, blockHash)
+}
+
+// getMessageForBlockHash returns the commit messages
+// for a particular block hash from the tracker.
+// It returns nil if the block hash does not exist.
+func (ct *commitsTracker) getMessageForBlockHash(
+	blockHash common.Hash) (message *CommitMessage) {
```
timwu20, line 102:
```suggestion
func (ct *commitsTracker) CommitMessage(blockHash common.Hash) (message *CommitMessage) {
```
nit: I don't like getters syntax.

lib/grandpa/tracker_commits.go
```diff
@@ -0,0 +1,117 @@
...
+	}
+
+	// add new block hash in tracker
+	ct.cleanup()
+	element := ct.linkedList.PushFront(blockHash)
+	data = commitMessageMapData{
+		message: commitMessage,
+		element: element,
+	}
+	ct.mapping[blockHash] = data
```
timwu20, line 67: not sure why we need the `commitMessageMapData` type? Could we not just use `list.PushFront` to return us an element and store that in `ct.mapping`? That way `ct.mapping` would be of type `map[common.Hash]*list.Element`

lib/grandpa/tracker_commits.go
```diff
@@ -0,0 +1,117 @@
...
+	if !ok {
+		return nil
+	}
+
+	return data.message
+}
+
+// getMessages returns all the commit messages contained in the
+// tracker in the form of slice of commit messages.
+func (ct *commitsTracker) forEach(f func(message *CommitMessage)) {
```
timwu20, line 113: Could we not just return `[]*CommitMessage` in this case rather than requiring us to pass in a callback function?  You should be able to just iterate through the list like:
```
for e := l.Front(); e != nil; e = e.Next() {
	// do something with e.Value
}
```

timwu20 (changes requested):

lib/grandpa/tracker_commits.go
qdm12: It was originally `getMessages` returning a slice of messages assembled from the mapping.
Depending on the number of tracked messages, this would have some impact on performance.
It also suggested there was some order in messages although there wasn't so that's also an element that pushed me towards the current approach.

I also thought about just returning the mapping directly but that feels wrong (separation of concerns, impl details, weird API, unwanted mutations).

I also wrote a benchmark (in `tracker_commits_test.go`) giving out:

```
Benchmark_ForEachVsSlice/forEach
Benchmark_ForEachVsSlice/forEach-8         	     295	   3818911 ns/op	  101656 B/op	     691 allocs/op
Benchmark_ForEachVsSlice/get_messages_for_iterate
Benchmark_ForEachVsSlice/get_messages_for_iterate-8         	     260	   4023858 ns/op	  918320 B/op	     785 allocs/op
```

lib/grandpa/message_tracker.go
```diff
@@ -19,25 +19,26 @@ type tracker struct {
...
 	votes      votesTracker
-
-	// map of commit block hash to commit message
-	commitMessages map[common.Hash]*CommitMessage
-	mapLock        sync.Mutex
-	in             chan *types.Block // receive imported block from BlockState
-	stopped        chan struct{}
+	commits    commitsTracker
+	mapLock    sync.Mutex
+	in         chan *types.Block // receive imported block from BlockState
```
EclesioMeloJunior, line 24: `in` can mean many things, I believe that `importedBlocks` is more clear

lib/grandpa/message_tracker.go
```diff
@@ -19,25 +19,26 @@ type tracker struct {
...
+	stopped    chan struct{}
```
EclesioMeloJunior, line 25:
```suggestion
	stopCh    chan struct{}
```

lib/grandpa/message_tracker.go
```diff
@@ -19,25 +19,26 @@ type tracker struct {
...
 	handler    *MessageHandler
 	votes      votesTracker
-
-	// map of commit block hash to commit message
-	commitMessages map[common.Hash]*CommitMessage
-	mapLock        sync.Mutex
-	in             chan *types.Block // receive imported block from BlockState
-	stopped        chan struct{}
+	commits    commitsTracker
+	mapLock    sync.Mutex
```
EclesioMeloJunior, line 23: `mapLock` refers to `catchUpResponseMessages` map? if so I think we can change to `catchUpMessagesLock`

lib/grandpa/tracker_commits.go
```diff
@@ -0,0 +1,117 @@
...
+	data, has := ct.mapping[blockHash]
+	if !has {
+		return
+	}
+
+	ct.linkedList.Remove(data.element)
+	delete(ct.mapping, blockHash)
+}
+
+// getMessageForBlockHash returns the commit messages
```
EclesioMeloJunior, line 98:
```suggestion
// getMessageForBlockHash returns a pointer to the commit messages
```

EclesioMeloJunior (changes requested):

lib/grandpa/message_tracker.go
qdm12: That's code that was there before and changes would be out of scope.
On top of your comment, I would add we should not use a channel and use methods instead (maybe with a mutex if we need). This is especially true when communicating between different interfaces.

lib/grandpa/message_tracker.go
qdm12: Out of scope, changes show up since formatting (alignment) changed

lib/grandpa/message_tracker.go
qdm12: Out of scope, changes show up since formatting (alignment) changed.

But... it looks like `mapLock` is for all the 'sub-trackers'/sub-maps. The commit tracker is not thread safe so I would guess we need this `mapLock` for it as well.

lib/grandpa/tracker_commits.go
qdm12: Done in https://github.com/ChainSafe/gossamer/pull/2490/commits/42fe8a13128dccc59d8d8ac15fbbf0390488550c

lib/grandpa/tracker_commits.go
```diff
@@ -0,0 +1,117 @@
+// Copyright 2022 ChainSafe Systems (ON)
```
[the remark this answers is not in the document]
timwu20: I would assume the `commits_tracker.go` file contains the `commitsTracker` type when I try to jump to a file.

lib/grandpa/tracker_commits.go
timwu20: See https://go.dev/doc/effective_go#Getters
> There's nothing wrong with providing getters and setters yourself, and it's often appropriate to do so, but it's neither idiomatic nor necessary to put Get into the getter's name. If you have a field called owner (lower case, unexported), the getter method should be called Owner (upper case, exported), not GetOwner.

The `commitsTracker` type isn't exported anyways.

lib/grandpa/tracker_commits.go
timwu20: You would have a pointer to the `list.Element` and you could just use the `list.Element.Value` attribute and cast it back to `*CommitMessage`. 

I took a stab at it on another branch: https://github.com/ChainSafe/gossamer/pull/2517/files

lib/grandpa/tracker_commits.go
timwu20: I'd like to see the results of this benchmark if we revise the mapping.

timwu20 (changes requested):

timwu20:
It may make sense to create some sort of generic `FIFOMappedCappedList` that we can use for all these trackers.  Should be easier if we just accept `any` like the `list.PushToXXX` functions accept.

kishansagathiya (commented):
Same as vote tracking PR. would more sense to see ways to join both commit and vote message tracker, If there is a way to merge them without much hassle.

lib/grandpa/tracker_commits.go
qdm12: I did the revised mapping, and now the for each is worst. Although that's on the same machine, and the older forEach with the older mapping was twice faster and twice less allocations than the newer mapping + copy all messages.

```
Benchmark_ForEachVsSlice/forEach
Benchmark_ForEachVsSlice/forEach-8         	     152	   7897919 ns/op	  212625 B/op	    1999 allocs/op
Benchmark_ForEachVsSlice/get_messages_for_iterate
Benchmark_ForEachVsSlice/get_messages_for_iterate-8         	     223	   5273454 ns/op	  947635 B/op	    1363 allocs/op
```

Do you think we should revert back (reset the commit) and keep it as it was?

lib/grandpa/tracker_commits.go
qdm12: Changed to `message(blockHash common.Hash)` because:

- do not have exported method to avoid any confusion that the tracker matches a certain interface
- from the tracker perspective, we get a message, so just `message` feels appropriate. The commitMessage type might later become an `any` as well.

lib/grandpa/tracker_commits.go
timwu20: I think it's a minor optimisation.  I think avoiding more types is probably best.

lib/grandpa/tracker_commits.go
timwu20: I would assume the order needs to be preserved if this is supposed to be a FIFO queue.  I guess the dereferencing of the pointer and typecast is slowing this down.  I would say keep it as is for now since the code is simpler.

timwu20 (approved):

kishansagathiya (approved):

EclesioMeloJunior (approved):

qdm12:
Merging as is, will do #2507 as a separate PR

PR #2496, opened 2022-04-25
chore(ci): simplify Docker setup

qdm12:
## Changes

- Use built-in `DOCKER_BUILDKIT=1`
- Remove QEMU setup (no emulation)
- Remove Buildx setup (no emulation)
- Remove default values: `file: ./Dockerfile`, `platforms: linux/amd64`, `push: false`
- Remove unneeded `name` and `id` of Docker steps

## Tests

CI suite passing

## Primary Reviewer

@EclesioMeloJunior

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -112,23 +112,14 @@ jobs:
 
   docker-build-n-push:
     runs-on: ubuntu-latest
+    env:
+      DOCKER_BUILDKIT: "1"
     steps:
-      - name: Set up QEMU
-        uses: docker/setup-qemu-action@v1
-      - name: Set up Docker Buildx
-        uses: docker/setup-buildx-action@v1
-      - name: Login to DockerHub
-        uses: docker/login-action@v1
+      - uses: docker/login-action@v1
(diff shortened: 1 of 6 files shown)


EclesioMeloJunior (approved):
nit: could we keep the `file: ./Dockerfile` argument?

jimjbrettj (approved):
lgtm

qdm12:
@EclesioMeloJunior 

> nit: could we keep the `file: ./Dockerfile` argument?

Why though? It's the default argument, I don't see the point keeping it :thinking:

EclesioMeloJunior:
Yea, I know, just suggested to be more explicit. For example in the case of changing the docker file name so other ppl know where to change.

qdm12:
> Yea, I know, just suggested to be more explicit. For example in the case of changing the docker file name so other ppl know where to change.

I get your point, although then you might end up with a hundred option fields set to their defaults. It's better to keep it to the minimum to reduce cognitive load :brain: 

Also we will likely never rename that `Dockerfile` to something else :smile:

PR #2497, opened 2022-04-26
fix(lib/grandpa): check equivocatory votes count

### LINKED ISSUE #2401: GRANDPA Does Not Check Multiplicity of Equivocatory Votes (GSR-07)
An equivocatory vote occurs when a validator has two or more votes at the same stage during a single round.

The specifications state that a valid justification must not contain more than two equivocatory votes from each voter.

The Gossamer function `getEquivocatoryVoters()` does not put a restriction on the number of equivocatory votes from each voter.

The impact is that commits that may be rejected in Substrate but will be accepted in Gossamer. Furthermore, it is an inefficient use of space and bandwidth to include more than two equivocatory votes.

The issue can be mitigated by rejecting commits with more than two equivocations per user.

This can be implemented by having `getEquivocatoryVoters()` return error if any authority has more than two equivocations as seen below.

```go
func getEquivocatoryVoters(votes []AuthData) (map[ed25519.PublicKeyBytes]struct{}, error) {
	eqvVoters := make(map[ed25519.PublicKeyBytes]struct{})
	voters := make(map[ed25519.PublicKeyBytes]int, len(votes))
	for _, v := range votes {
		voters[v.AuthorityID]++
		if voters[v.AuthorityID] > 2 {
			return nil, errors.New("Invalid Multiplicity for Voter")
		}
		if voters[v.AuthorityID] == 2 {
			eqvVoters[v.AuthorityID] = struct{}{}
		}
	}
	return eqvVoters, nil
}
```

kishansagathiya:
## Changes
A valid justification can't have more than two equivocatory votes per
voter. Check the number of equivocatory votes per voter and throw an
error if more than two.

## Tests
`lib/grandpa` tests should be enough

## Issues

Fixes #2401
## Primary Reviewer

@timwu20

### DIFF
--- a/lib/grandpa/message_handler.go
+++ b/lib/grandpa/message_handler.go
@@ -9,6 +9,7 @@ import (
 	"fmt"
 	"reflect"
 
+	"github.com/ChainSafe/chaindb"
 	"github.com/ChainSafe/gossamer/dot/network"
 	"github.com/ChainSafe/gossamer/dot/telemetry"
 	"github.com/ChainSafe/gossamer/dot/types"
@@ -104,6 +105,16 @@ func (h *MessageHandler) handleNeighbourMessage(msg *NeighbourMessage) error {
 func (h *MessageHandler) handleCommitMessage(msg *CommitMessage) error {
 	logger.Debugf("received commit message, msg: %+v", msg)
 
+	err := verifyBlockHashAgainstBlockNumber(h.blockState, msg.Vote.Hash, uint(msg.Vote.Number))
+	if err != nil {
+		if errors.Is(err, chaindb.ErrKeyNotFound) {
+			h.grandpa.tracker.addCommit(msg)
+			logger.Infof("we might not have synced to the given block %s yet: %s", msg.Vote.Hash, err)
+			return nil
+		}
+		return err
+	}
+
 	containsPrecommitsSignedBy := make([]string, len(msg.AuthData))
 	for i, authData := range msg.AuthData {
 		containsPrecommitsSignedBy[i] = authData.AuthorityID.String()
@@ -183,6 +194,16 @@ func (h *MessageHandler) handleCatchUpResponse(msg *CatchUpResponse) error {
 	logger.Debugf(
 		"received catch up response with hash %s for round %d and set id %d",
 		msg.Hash, msg.Round, msg.SetID)
+
+	err := verifyBlockHashAgainstBlockNumber(h.blockState, msg.Hash, uint(msg.Number))
+	if err != nil {
+		if errors.Is(err, chaindb.ErrKeyNotFound) {
+			h.grandpa.tracker.addCatchUpResponse(msg)
+			logger.Infof("we might not have synced to the given block %s yet: %s", msg.Hash, err)
+			return nil
+		}
+		return err
+	}
 
 	// TODO: re-add catch-up logic (#1531)
 	if true {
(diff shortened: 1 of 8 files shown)


lib/grandpa/message_handler_test.go
```diff
@@ -649,7 +649,7 @@ func TestMessageHandler_VerifyBlockJustification_WithEquivocatoryVotes(t *testin
 
 	round := uint64(1)
 	number := uint32(1)
-	precommits := buildTestJustification(t, 20, round, setID, kr, precommit)
+	precommits := buildTestJustification(t, 18, round, setID, kr, precommit)
```
kishansagathiya, line 652: there are 9 authorities. If we vote count is 20, there would be few authorities with more than 2 votes. More than two equivocatory votes make the justification invalid.

lib/grandpa/message_handler.go
```diff
@@ -284,27 +284,32 @@ func (h *MessageHandler) verifyCatchUpResponseCompletability(prevote, precommit
 	return nil
 }
 
-func getEquivocatoryVoters(votes []AuthData) map[ed25519.PublicKeyBytes]struct{} {
+func getEquivocatoryVoters(votes []AuthData) (map[ed25519.PublicKeyBytes]struct{}, error) {
```
qdm12, line 287: nit named return values would be nice

```suggestion
func getEquivocatoryVoters(votes []AuthData) (eqvVoters map[ed25519.PublicKeyBytes]struct{}, err error) {
```

lib/grandpa/message_handler.go
```diff
@@ -284,27 +284,32 @@ func (h *MessageHandler) verifyCatchUpResponseCompletability(prevote, precommit
...
 	eqvVoters := make(map[ed25519.PublicKeyBytes]struct{})
 	voters := make(map[ed25519.PublicKeyBytes]int, len(votes))
 
 	for _, v := range votes {
 		voters[v.AuthorityID]++
+		if voters[v.AuthorityID] > 2 {
+			return nil, errInvalidMultiplicity
```
qdm12, line 294: nit perhaps wrap some context information?

```suggestion
			return nil, fmt.Errorf("%w: authority id %x has %d votes", 
				errInvalidMultiplicity, v.AuthorityID, voters[v.AuthorityID])
```

lib/grandpa/message_handler.go
```diff
@@ -284,27 +284,32 @@ func (h *MessageHandler) verifyCatchUpResponseCompletability(prevote, precommit
...
+		}
 
-		if voters[v.AuthorityID] > 1 {
+		if voters[v.AuthorityID] == 2 {
```
qdm12, line 297: nit a switch could be nice:

```go
switch voters[v.AuthorityID] {
	case 1:
	case 2:
		eqvVoters[v.AuthorityID] = struct{}{}
	default:
		return nil, fmt.Errorf("%w: authority id %x has %d votes", 
			errInvalidMultiplicity, v.AuthorityID, voters[v.AuthorityID])
}
```

qdm12 (approved):
Good work! :+1:

lib/grandpa/message_handler.go
timwu20: I prefer a switch here as well.  Or at the very least an `else if` condition.

timwu20 (changes requested):

EclesioMeloJunior (approved):
LGMT! just a small comment

timwu20 (approved):

PR #2500, opened 2022-04-28
chore(tests): use mockery constructors to ensure assertions

qdm12:
## Changes

- [x] Use generated mockery constructors for mockery mocks to ensure assertions are done in the test
- [x] Fix bugs caused by the now-fixed mock assertions
- [x] Remove existing `.Assert*` mocks calls

## Tests

All unit tests passing

## Primary Reviewer

@EclesioMeloJunior

### DIFF
--- a/lib/babe/babe_integration_test.go
+++ b/lib/babe/babe_integration_test.go
@@ -54,26 +54,13 @@ var (
 
 //go:generate mockgen -destination=mock_telemetry_test.go -package $GOPACKAGE github.com/ChainSafe/gossamer/dot/telemetry Client
 
-func createTestService(t *testing.T, cfg *ServiceConfig) *Service {
+func createTestService(t *testing.T, cfg ServiceConfig) *Service {
 	wasmer.DefaultTestLogLvl = 1
 
 	gen, genTrie, genHeader := genesis.NewDevGenesisWithTrieAndHeader(t)
 	genesisHeader = genHeader
 
 	var err error
-
-	if cfg == nil {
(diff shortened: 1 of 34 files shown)


lib/babe/babe.go
```diff
@@ -169,10 +169,6 @@ func NewService(cfg *ServiceConfig) (*Service, error) {
 		return nil, errNilEpochState
 	}
 
-	if cfg.BlockImportHandler == nil {
-		return nil, errNilBlockImportHandler
-	}
-
```
qdm12, line 175: Note this is removed to allow to have `nil` block import handler in test code.
Anyway imo this check shouldn't be there, and we should panic if we try to pass a nil handler.

lib/babe/babe_integration_test.go
```diff
@@ -54,26 +54,14 @@ var (
...
-			Authority: true,
-		}
-	}
-
-	cfg.BlockImportHandler = new(mocks.BlockImportHandler)
-	cfg.BlockImportHandler.(*mocks.BlockImportHandler).
-		On("HandleBlockProduced",
-			mock.AnythingOfType("*types.Block"), mock.AnythingOfType("*storage.TrieState")).
-		Return(nil)
-
```
qdm12, line 76: note this is pushed back up to the caller (tests), since it was differing from one test to the other.

jimjbrettj (approved):
Looks good. I will probably wait to merge my branch in so that I can use these changes and not have to duplicate (or create more conflicts)

edwardmack (approved):

PR #2501, opened 2022-04-28
chore(tests): remove tests testing tests

qdm12:
## Changes

- [x] Remove assertions testing test helping functions, only :red_circle: 

## Primary Reviewer

@edwardmack


kishansagathiya (changes requested):
why are we removing this?

kishansagathiya (approved):
> If your question is about why we remove this, because these assertions were testing test helping functions. We really should not test our test helping functions; on top of that, they always return the same thing. So useless confusing code, that also might get copied pastaed later on and add more useless testing code.

:+1:

PR #2503, opened 2022-04-28
check if epoch is less than s.skipToEpoch

### LINKED ISSUE #2403: Incorrect Bounds Checking Skips Verification of Epoch 0 (GSR-09)
The function `SkipVerify()`, will skip BABE verification if the epoch is less than or equal to `skipToEpoch`.

When running a live node the variable `skipToEpoch` should be unused, however it will default to 0. Thus, headers will never be verified for epoch 0. As epochs begin counting from 0 the first epoch worth of blocks will not be verified allowing nodes to send malicious blocks that will be accepted.

Consider reducing the condition that compares `epoch` to `s.skipToEpoch` to use a strict inequality.

kishansagathiya:
## Changes

compares epoch to s.skipToEpoch with strict inequality. So, far we were
checking less than or equal to, in which epoch 0 never gets verified.

## Tests

```sh
```

## Issues

Fixes #2403
## Primary Reviewer

@timwu20


timwu20 (approved):
well done 😅

qdm12 (approved):

Issue #2187, opened 2022-01-12
Trie: merge branch and leaf structure in a single type


qdm12:
## Task summary

- Initially proposed by @kishansagathiya 
- Since trie's leaf and branch structures are rather similar we should merge them to simplify code

Proposed implementation:

- The new node structure can contain a (1 byte) type field containing the node type (leaf, branch, branch with value)
- The children structure should be changed from an array to something more space efficient such as a map or linked list. We don't want to store 4x16 bytes (for 16 pointers) for each leaf.
- The existing node interface should be removed

kishansagathiya:
> The State Trie is a radix-16 tree

This make me think that you will have to use arrays.

qdm12:
> This make me think that you will have to use arrays.

I need to investigate if a 16 sized array actually takes 16x4 bytes of memory even if its members are nil. If it doesn't we keep the array, otherwise we can use in example a map of index (0 to 15) -> child pointer.

PR #2504, opened 2022-04-28
chore(trie): merge leaf and branch in single node struct

qdm12:
## Changes

:information_source: the aim of this PR was to simplify and streamline encoding and decoding of nodes before moving to implement additional encoding/decoding for the v1 trie.

:sos: this is a substantial amount of changes, but most of it is just regex replacements in unit tests. All tests pass with the existing values, so it is quite safe :wink: 

- [x] Removed type assertions and `Node` interface
- [x] Leaf and branch structs merged in one common `Node` struct
- [x] Removed trivial methods that were used to avoid type switches (i.e. `GetHash()`)
- [x] Update all tests
- [x] Change `Children` field from an array to a slice so it can be nil for leaves
- [x] Deduct node type from children slice (not nil == leaf)

## Tests

```sh
go test -race ./internal/trie/... ./lib/trie/...
```

## Issues

#2187 

## Primary Reviewer

- @kishansagathiya


qdm12:
> I wonder if this was possible by keeping node interface, moving common function into one struct and build leaf and branch using that common struct with the help of composition.

I thought about this, but I really wanted to get rid of all the type assertions and have a 'dumb' single node struct.
Maybe we can revisit later and try with composition though.

internal/trie/node/decode.go
```diff
@@ -36,42 +34,37 @@ func Decode(reader io.Reader) (n Node, err error) {
...
 // Note that since the encoded branch stores the hash of the children nodes, we are not
 // reconstructing the child nodes from the encoding. This function instead stubs where the
 // children are known to be with an empty leaf. The children nodes hashes are then used to
 // find other values using the persistent database.
-func decodeBranch(reader io.Reader, header byte) (branch *Branch, err error) {
-	nodeType := Type(header >> 6)
-	switch nodeType {
-	case BranchType, BranchWithValueType:
-	default:
-		return nil, fmt.Errorf("%w: %d", ErrNodeTypeIsNotABranch, nodeType)
```
kishansagathiya, line 68: I understand that `decodeBranch` gets called only on a branch, so we could do without this check. 
By removing this check we rely on the caller to not use `decodeBranch` other than for a branch.
I think having this check does not hurt.

internal/trie/node/decode.go
```diff
@@ -108,41 +102,37 @@ func decodeBranch(reader io.Reader, header byte) (branch *Branch, err error) {
 
 		// Handle inlined leaf nodes.
 		const hashLength = 32
-		if Type(hash[0]>>6) == LeafType && len(hash) < hashLength {
+		nodeTypeHeaderByte := hash[0] >> 6
+		if nodeTypeHeaderByte == leafHeader && len(hash) < hashLength {
```
kishansagathiya, line 106: Again, I don't get why `Type` is getting changed to header everywhere

internal/trie/node/decode.go
```diff
@@ -108,41 +102,37 @@ func decodeBranch(reader io.Reader, header byte) (branch *Branch, err error) {
...
 }
 
 // decodeLeaf reads and decodes from a reader with the encoding specified in lib/trie/node/encode_doc.go.
-func decodeLeaf(reader io.Reader, header byte) (leaf *Leaf, err error) {
-	nodeType := Type(header >> 6)
-	if nodeType != LeafType {
-		return nil, fmt.Errorf("%w: %d", ErrNodeTypeIsNotALeaf, nodeType)
-	}
-
-	leaf = &Leaf{
```
kishansagathiya, line 140: same comment as above in decodeBranch

internal/trie/node/node.go
```diff
@@ -3,24 +3,87 @@
...
+	Key   []byte
+	Value []byte
+	// Generation is incremented on every trie Snapshot() call.
+	// Each node also contain a certain Generation number,
+	// which is updated to match the trie Generation once they are
+	// inserted, moved or iterated over.
+	Generation uint64
+	// Children is a slice of length 16 for branches.
+	// It is left to nil for leaves to reduce memory usage.
+	Children []*Node
```
kishansagathiya, line 26: Looks like this was sized to `Children [16]Node` earlier. why this change to unsized?

internal/trie/node/branch_encode.go
```diff
@@ -223,7 +115,7 @@ func encodeChildrenOpportunisticParallel(children [16]Node, buffer io.Writer) (e
 	return err
 }
 
-func encodeChildrenSequentially(children [16]Node, buffer io.Writer) (err error) {
+func encodeChildrenSequentially(children []*Node, buffer io.Writer) (err error) {
```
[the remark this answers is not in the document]
qdm12: It's for leaves, so that it can be `nil` and not take space (instead of 16*8 bytes).

internal/trie/node/branch_encode.go
```diff
@@ -158,13 +50,13 @@ var parallelEncodingRateLimit = make(chan struct{}, parallelLimit)
 // goroutines IF they are less than the parallelLimit number of goroutines already
 // running. This is designed to limit the total number of goroutines in order to
 // avoid using too much memory on the stack.
-func encodeChildrenOpportunisticParallel(children [16]Node, buffer io.Writer) (err error) {
+func encodeChildrenOpportunisticParallel(children []*Node, buffer io.Writer) (err error) {
```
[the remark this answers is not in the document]
qdm12: See https://github.com/ChainSafe/gossamer/pull/2504#discussion_r880790130

internal/trie/node/copy.go
```diff
@@ -52,85 +52,56 @@ type CopySettings struct {
...
-	if settings.CopyChildren {
-		// Copy all fields of children if we deep copy children
-		childSettings := settings
-		childSettings.CopyKey = true
-		childSettings.CopyValue = true
-		childSettings.CopyCached = true
-		for i, child := range b.Children {
-			if child == nil {
-				continue
+	if n.Type() == Branch {
```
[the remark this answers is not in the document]
qdm12: That block of code is for children only, and leaves don't have any, so there isn't something to copy for leaves here. All the rest of the Copy method is common for branches and leaves

internal/trie/node/decode.go
```diff
@@ -36,42 +34,37 @@ func Decode(reader io.Reader) (n Node, err error) {
...
-	case BranchType, BranchWithValueType:
+	case branchHeader, branchWithValueHeader:
 		n, err = decodeBranch(reader, header)
 		if err != nil {
 			return nil, fmt.Errorf("cannot decode branch: %w", err)
 		}
 		return n, nil
 	default:
-		return nil, fmt.Errorf("%w: %d", ErrUnknownNodeType, nodeType)
+		return nil, fmt.Errorf("%w: %d", ErrUnknownNodeType, nodeTypeHeaderByte)
```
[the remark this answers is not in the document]
qdm12: These are necessary:

- the `header >> 6` is only the (shifted) variant bits of the header byte. These header bits should be unexported and are used only be the codec.
- The `Type` for a node has now different meaning, and can be either `Leaf` or `Branch`. It's not related to the codec but only used for code paths here and there.

Although these are further modified to better match the newer Polkadot spec and be more understandable in #2530

internal/trie/node/decode.go
qdm12: This function is not exported and is used in only one place, I would still like to remove that check.
I recall I added it previously to detect decoding errors I had whilst testing it, I probably forgot to remove it.

On top of that:

- Tests (unit, e2e, fuzzing) should detect such bad use of this function
- The header (variant + partial key length) decoding is pushed before `decodeBranch` / `decodeLeaf`. This is even more robust in the next PR #2530

internal/trie/node/node.go
qdm12: So it can be nil for leaves and not use 16x8 bytes per leaf of space. It's always made to a slice of 16 children for branches though.

internal/trie/node/copy.go
```diff
@@ -52,85 +52,56 @@ type CopySettings struct {
...
+			for i, child := range n.Children {
+				if child == nil {
+					continue
+				}
+				cpy.Children[i] = child.Copy(childSettings)
 			}
-			cpy.Children[i] = child.Copy(childSettings)
+		} else {
+			cpy.Children = make([]*Node, ChildrenCapacity)
+			copy(cpy.Children, n.Children) // copy interface pointers only)
```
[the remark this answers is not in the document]
qdm12: Updated! 👍

qdm12 (commented):
Thanks for the review @kishansagathiya !
All your comments should be addressed now 😉

kishansagathiya (approved):

internal/trie/node/header.go
```diff
@@ -7,62 +7,43 @@ import (
...
-	if len(l.Key) < 63 {
-		header |= byte(len(l.Key))
+	if len(node.Key) < keyLenOffset {
+		header |= byte(len(node.Key))
 		_, err = writer.Write([]byte{header})
 		return err
 	}
 
-	header |= keyLenOffset
+	header = header | keyLenOffset
```
EclesioMeloJunior, line 41: In many places we're using short self-assigned vars I think would be nice to keep here as well
```suggestion
	header |= keyLenOffset
```

EclesioMeloJunior (changes requested):

internal/trie/node/encode.go
```diff
@@ -0,0 +1,75 @@
...
+		_, err = buffer.Write(childrenBitmap)
+		if err != nil {
+			return fmt.Errorf("cannot write children bitmap to buffer: %w", err)
+		}
+	}
+
+	// check value is not nil for branch nodes, even though
+	// leaf nodes always have a non-nil value.
+	if n.Type() == Leaf || n.Value != nil {
+		// TODO remove `n.Type() == Leaf` and update tests
```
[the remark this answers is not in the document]
qdm12: Well, yes and no, since changing this will affect test values and I wanted to keep this PR with the same test values to be sure it doesn't change behavior. I'd rather do it right after it gets merged as another PR. Let's keep this unresolved though so I don't forget :wink:

internal/trie/node/header.go
qdm12: Indeed, although most of these get thrown out in #2530 ([that particular line gets removed and I do use the short `|` form too!](https://github.com/ChainSafe/gossamer/pull/2530/files#diff-63521964d6b19c9f06c40c3276755b2070ec87f12744cbea72d74ca32f152561L41)), so I'll not add more diffs to have an easier rebase for the next PR which contains those changes :wink:

EclesioMeloJunior (approved):

internal/trie/node/encode.go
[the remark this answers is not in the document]
qdm12: @EclesioMeloJunior done in #2588 resolving this!

PR #2519, opened 2022-05-04
fix(state/grandpa): track changes across forks

### LINKED ISSUE #2349: GRANDPA Should Only Process Forced Changes When They Exist (GSR-12)
Gossamer should ignore scheduled changes when forced changes exist. In [Substrate](https://github.com/paritytech/substrate/blob/fb08d15bcbea68415dd4ee90d11556b653096f10/client/finality-grandpa/src/import.rs#L239) they will only process the forced change and ignore the scheduled changes. However in Gossamer both the scheduled change and forced change will be processed, which is incorrect behavior.

### LINKED ISSUE #2395: `GrandpaState` Should be Set on Finalization, Not Import (GSR-01)
Similar to #2269, except this problem relates to GRANDPA digests and the `GrandpaState`. Presumable, the fix will look like #2339, but for the `GrandpaState` data.

### LINKED ISSUE #2400: Forks in Blocks with Different GRANDPA Digests Result in a Consensus Split (GSR-02)
There is only one `Handler` which processes all blocks, on any fork. The `Handler` has fields relating to GRANDPA changes which are not updated if we switch to another fork. As a result the `Handler` will have incorrect values from the GRANDPA changes when we switch forks.

GRANDPA changes should be maintained separately for each fork.

EclesioMeloJunior:
## Changes

- [x] Track GRANDPA scheduled/forced changes by fork until and apply the change accordingly with the imported block (in case of forced change) or the finalized block (in case of scheduled change)

## Tests

```sh
```

## Issues

- Closes #2400
- Closes #2395
- Closes #2519
- Closes #2349


dot/digest/digest.go
```diff
@@ -249,159 +188,13 @@ func (h *Handler) handleBlockFinalisation(ctx context.Context) {
 				h.logger.Errorf("failed to persist babe next epoch config: %s", err)
 			}
 
-			err = h.handleGrandpaChangesOnFinalization(info.Header.Number)
+			err = h.grandpaState.ApplyScheduledChanges(&info.Header)
 			if err != nil {
-				h.logger.Errorf("failed to handle grandpa changes on block finalisation: %s", err)
+				h.logger.Errorf("failed to apply standard scheduled changes on block finalisation: %w", err)
```
noot, line 193:
```suggestion
				h.logger.Errorf("failed to apply scheduled change on block finalisation: %w", err)
```

not sure what's meant by standard :p also, there should only be one scheduled change per block right? or at least, only one that can take effect

dot/state/grandpa.go
```diff
@@ -24,44 +26,570 @@ var (
 	currentSetIDKey   = []byte("setID")
 )
 
+var (
```
noot, line 29: all of this pendingChange related code should probably go in its own file

dot/state/grandpa.go
```diff
@@ -24,44 +26,570 @@ var (
...
+}
+
+func (c *pendingChangeNode) importScheduledChange(blockHash common.Hash, blockNumber uint, pendingChange *pendingChange,
+	isDescendantOf isDescendantOfFunc) (imported bool, err error) {
+	announcingHash := c.change.announcingHeader.Hash()
+
+	if blockHash.Equal(announcingHash) {
+		return false, errDuplicateHashes
+	}
+
```
noot, line 81: what happens if there's two scheduled changes per block?

dot/state/grandpa.go
```diff
@@ -24,44 +26,570 @@ var (
...
+	if blockNumber <= c.change.announcingHeader.Number {
+		return false, nil
+	}
```
noot, line 84: should this be an error? also, why do we need blockHash and blockNumber parameters if the pendingChange contains them?

dot/state/grandpa.go
```diff
@@ -24,44 +26,570 @@ var (
...
+	case types.GrandpaScheduledChange:
+		return s.addScheduledChange(header, val)
+	case types.GrandpaForcedChange:
+		return s.addForcedChange(header, val)
+	case types.GrandpaOnDisabled:
+		return nil
+	case types.GrandpaPause:
+		return nil
+	case types.GrandpaResume:
+		return nil
```
noot, line 181: should we error for these too (as we are probably supposed to support them?)

dot/state/grandpa.go
```diff
@@ -24,44 +26,570 @@ var (
...
+		}
+
+		isDescendant, err := s.blockState.IsDescendantOf(changeBlockHash, headerHash)
+		if err != nil {
+			return fmt.Errorf("cannot verify ancestry: %w", err)
+		}
+
+		if isDescendant {
+			return errAlreadyHasForcedChanges
+		}
```
noot, line 204: is this here because there can only be one forced-change per fork?

dot/state/grandpa.go
```diff
@@ -24,44 +26,570 @@ var (
...
+
+func (s *GrandpaState) importScheduledChange(pendingChange *pendingChange) error {
+	highestFinalizedHeader, err := s.blockState.GetHighestFinalisedHeader()
+	if err != nil {
+		return fmt.Errorf("cannot get highest finalized header: %w", err)
+	}
+
+	if pendingChange.announcingHeader.Number <= highestFinalizedHeader.Number {
+		return errLowerThanBestFinalized
+	}
```
noot, line 249: won't this reject any changes that are in unfinalized blocks? but I think this function is called for imported (unfinalized) blocks, right?

dot/state/grandpa.go
```diff
@@ -24,44 +26,570 @@ var (
...
+		if effectiveNumber == importedBlockHeader.Number {
+			forcedChange = forced
+			break
+		}
+	}
+
+	if forcedChange == nil {
+		return nil
+	}
+
```
noot, line 466: should the forced changes also be removed from the set once it's applied?

dot/state/grandpa.go
```diff
@@ -24,44 +26,570 @@ var (
...
+	errDuplicateHashes         = errors.New("duplicated hashes")
+	errAlreadyHasForcedChanges = errors.New("already has a forced change")
+	errUnfinalizedAncestor     = errors.New("ancestor with changes not applied")
+	errLowerThanBestFinalized  = errors.New("current finalized is lower than best finalized header")
+
+	ErrNoChanges = errors.New("cannot get the next authority change block number")
+)
+
+type pendingChange struct {
+	bestFinalizedNumber uint32
```
noot, line 40: what is this for?

lib/grandpa/grandpa.go
```diff
@@ -689,6 +683,11 @@ func (s *Service) deleteVote(key ed25519.PublicKeyBytes, stage Subround) {
 func (s *Service) determinePreVote() (*Vote, error) {
 	var vote *Vote
 
+	beastBlockHeader, err := s.blockState.BestBlockHeader()
```
noot, line 686:
```suggestion
	bestBlockHeader, err := s.blockState.BestBlockHeader()
```

noot (commented):
looks pretty good, just a few questions/comments. also would be nice to split up the scheduled and forced change logic into different files as `grandpa_state.go` is getting pretty big!

dot/digest/digest.go
```diff
@@ -194,7 +132,8 @@ func (h *Handler) handleBabeConsensusDigest(digest scale.VaryingDataType, header
 		return nil
 
 	case types.BABEOnDisabled:
-		return h.handleBABEOnDisabled(val, header)
+		h.logger.Debug("handling BABEOnDisabled")
```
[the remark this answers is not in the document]
EclesioMeloJunior: Not really, removed!

dot/digest/digest.go
EclesioMeloJunior: substrate uses `standard` and `forced` scheduled changes, removed the `standard` term, and yes just one scheduled change should be applied in the `addScheduledChange` method we have a check to avoid adding more than one scheduled change per block. However, there is an edge case where in the same block there are scheduled and forced changes, in that case, we should apply only the forced change.

dot/state/grandpa.go
EclesioMeloJunior: Agreed!

dot/state/grandpa.go
EclesioMeloJunior: We consider just the first one the other ones will be dropped. This check is similar to the one made by the `ForkTree` struct when importing the data for a block

dot/state/grandpa.go
EclesioMeloJunior: Not really, this is a recursive function so when the block number is less than or equal to the one which is persisted (`c.change`) then we should add the pending change as a sibling and not as a child of the `c.change`. I'm using separated params to hash an number values to become simple to read and avoid terms like: `pendingChange.announcingHeader.Number` or `pendingChange.announcingHeader.Hash()`

dot/state/grandpa.go
EclesioMeloJunior: Yes, I'm fixing the tests cases for the `scheduled` and `forced` changes before implement the `pause` and `resume` digests

dot/state/grandpa.go
EclesioMeloJunior: Yes, is not allowed more than one forced change per fork

dot/state/grandpa.go
EclesioMeloJunior: Yes! Every time a block is imported we check the digest for scheduled changes to import and If the node has a finalized block n so we should not import a scheduled change from blocks before n but we should import changes from blocks ahead of n, right?

dot/state/grandpa.go
```diff
@@ -24,44 +26,570 @@ var (
...
+		isDescendant, err := s.blockState.IsDescendantOf(scheduledBlockHash, forcedChangeHash)
+		if err != nil {
+			return fmt.Errorf("cannot check ancestry: %w", err)
+		}
+
+		if isDescendant {
+			return errPendingScheduledChanges
+		}
+	}
+
```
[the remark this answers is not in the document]
EclesioMeloJunior: We must ensure that we don't have any scheduled changes pending to be applied in this range from `n` (where the forced change was announced) and `n + y` (where it should take effect).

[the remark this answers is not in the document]
EclesioMeloJunior: Answering your comment https://github.com/ChainSafe/gossamer/pull/2519#discussion_r868229134: 
The `bestFinalizedNumber` is a special number set by the sudo key when they need to issue a forced change, this number tells the new validators set the last finalized block (AFAIR from the meeting with Andre from parity). @danforbes @timwu20 correct me if I saying something wrong plz

dot/state/grandpa.go
```diff
@@ -24,44 +26,570 @@ var (
...
+		change = changeNode.change
+		s.scheduledChangeRoots = make([]*pendingChangeNode, len(changeNode.nodes))
+		copy(s.scheduledChangeRoots, changeNode.nodes)
+	}
+
+	return change, nil
+}
+
+// keepDescendantForcedChanges should keep the forced changes for later blocks that
+// are descendant of the finalized block
```
qdm12, line 342:
```suggestion
// keepDescendantForcedChanges removes the forced changes from GRANDPA state for any block
// that is not a descendant of the current finalised block.
```

dot/state/grandpa.go
```diff
@@ -24,44 +26,570 @@ var (
...
+func (s *GrandpaState) keepDescendantForcedChanges(finalizedHash common.Hash, finalizedNumber uint) error {
```
qdm12, line 343: you should rename to `pruneForcedChanges`, since it doesn't really `keep`, it `prunes` I would say :thinking:

dot/state/grandpa.go
```diff
@@ -24,44 +26,570 @@ var (
...
+	onBranchForcedChanges := []*pendingChange{}
+
+	for _, forcedChange := range s.forcedChanges {
+		isDescendant, err := s.blockState.IsDescendantOf(finalizedHash, forcedChange.announcingHeader.Hash())
+		if err != nil {
+			return fmt.Errorf("cannot verify ancestry: %w", err)
+		}
+
+		if forcedChange.effectiveNumber() > finalizedNumber && isDescendant {
```
qdm12, line 352: Do we need to check the effective number in this scenario? Doesn't the descendant check already do that (forgive my ignorance here :smile:)

dot/state/grandpa.go
```diff
@@ -24,44 +26,570 @@ var (
...
+		return fmt.Errorf("cannot keep descendant forced changes: %w", err)
+	}
+
+	if len(s.scheduledChangeRoots) == 0 {
+		return nil
+	}
+
+	changeToApply, err := s.getApplicableScheduledChange(finalizedHash, finalizedHeader.Number)
+	if err != nil {
+		return fmt.Errorf("cannot finalize scheduled change: %w", err)
```
qdm12, line 401: nit maybe this is more appropriate?

```suggestion
		return fmt.Errorf("cannot get applicable scheduled change: %w", err)
```

dot/state/grandpa.go
```diff
@@ -24,44 +26,570 @@ var (
...
+
+	err = s.setChangeSetIDAtBlock(newSetID, changeToApply.effectiveNumber())
+	if err != nil {
+		return fmt.Errorf("cannot set change set id at block")
+	}
+
+	logger.Debugf("Applying authority set change scheduled at block #%d",
+		changeToApply.announcingHeader.Number)
+
+	// TODO: add afg.applying_scheduled_authority_set_change telemetry info here
```
qdm12, line 429: nit just a TODO reminder :wink:

dot/state/grandpa.go
```diff
@@ -24,44 +26,570 @@ var (
...
+		}
+	}
+
+	s.scheduledChangeRoots = make([]*pendingChangeNode, len(onBranchScheduledChanges))
+	copy(s.scheduledChangeRoots, onBranchScheduledChanges)
+	return nil
+}
+
+// ApplyScheduledChanges will check the schedules changes in order to find a root
+// that is equals or behind the finalized number and will apply its authority set changes
```
qdm12, line 386: nit

```suggestion
// equal or behind the finalized number and will apply its authority set changes.
```

qdm12 (commented):
I'll continue reviewing tomorrow :wink: 
But this is looking exciting :+1:

dot/state/grandpa.go
```diff
@@ -24,44 +26,570 @@ var (
...
+		}
+
+		if !isDescendant {
+			continue
+		}
+
+		if effectiveNumber == importedBlockHeader.Number {
+			forcedChange = forced
+			break
+		}
```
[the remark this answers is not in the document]
EclesioMeloJunior: I think no, the consensus digest is a SCALE encoded enum, so the same block cannot have more than one forced change but it can have a finalized and a scheduled change

dot/digest/digest.go
noot: okay, makes sense!

dot/state/grandpa.go
noot: oh hmm, I think this check can probably be removed, because we might be syncing somewhat slowly so by the time we sync a block with a grandpa change digest, it could be finalized already, but we should still take it into account

dot/state/grandpa.go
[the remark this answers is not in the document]
noot: ok, interesting, so it forces them to set that block number as finalized? would be good to add a comment about that!

dot/state/grandpa.go
noot: ahh i see, i didn't see that it was recursive. that makes sense!

dot/state/grandpa.go
```diff
@@ -24,44 +26,570 @@ var (
...
+		copy(s.scheduledChangeRoots, changeNode.nodes)
+	}
+
+	return change, nil
+}
+
+// keepDescendantForcedChanges should keep the forced changes for later blocks that
+// are descendant of the finalized block
+func (s *GrandpaState) keepDescendantForcedChanges(finalizedHash common.Hash, finalizedNumber uint) error {
+	onBranchForcedChanges := []*pendingChange{}
```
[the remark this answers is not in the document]
EclesioMeloJunior: I checked in the `go.dev` and looks it keeps the capacity 0 for both cases -> https://go.dev/play/p/47y0HD3OGU4

dot/state/grandpa.go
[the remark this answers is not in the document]
qdm12: So should we not break and collect all the forced changes?

dot/state/grandpa.go
```diff
@@ -26,42 +37,477 @@ var (
...
+	case types.GrandpaOnDisabled:
+		return nil
+	case types.GrandpaPause:
+		logger.Warn("GRANDPA Pause consensus message not imeplemented yet")
+		return nil
+	case types.GrandpaResume:
+		logger.Warn("GRANDPA Resume consensus message not imeplemented yet")
+		return nil
+	default:
+		return fmt.Errorf("not supported digest")
```
qdm12, line 101: should we panic instead? That looks like a programming error

dot/state/grandpa.go
```diff
@@ -26,42 +37,477 @@ var (
...
+		return s.addScheduledChange(header, val)
+	case types.GrandpaForcedChange:
+		return s.addForcedChange(header, val)
+	case types.GrandpaOnDisabled:
+		return nil
+	case types.GrandpaPause:
+		logger.Warn("GRANDPA Pause consensus message not imeplemented yet")
+		return nil
+	case types.GrandpaResume:
+		logger.Warn("GRANDPA Resume consensus message not imeplemented yet")
```
qdm12, line 98:
```suggestion
		logger.Warn("GRANDPA Resume consensus message not implemented yet")
```

dot/state/grandpa.go
```diff
@@ -26,42 +37,477 @@ var (
...
+func (s *GrandpaState) HandleGRANDPADigest(header *types.Header, digest scale.VaryingDataType) error {
+	switch val := digest.Value().(type) {
+	case types.GrandpaScheduledChange:
+		return s.addScheduledChange(header, val)
+	case types.GrandpaForcedChange:
+		return s.addForcedChange(header, val)
+	case types.GrandpaOnDisabled:
+		return nil
+	case types.GrandpaPause:
+		logger.Warn("GRANDPA Pause consensus message not imeplemented yet")
```
qdm12, line 95:
```suggestion
		logger.Warn("GRANDPA Pause consensus message not implemented yet")
```

dot/state/grandpa.go
```diff
@@ -26,42 +37,477 @@ var (
...
+}
+
+func (s *GrandpaState) addForcedChange(header *types.Header, fc types.GrandpaForcedChange) error {
+	headerHash := header.Hash()
+
+	for _, change := range s.forcedChanges {
+		changeBlockHash := change.announcingHeader.Hash()
+
+		if changeBlockHash.Equal(headerHash) {
+			return errDuplicateHashes
```
qdm12, line 112:
```suggestion
			return fmt.Errorf("%w: for block hash %s", errDuplicateHashes, headerHash)
```

dot/state/grandpa.go
```diff
@@ -26,42 +37,477 @@ var (
...
+		}
+
+		if isDescendant {
+			return errAlreadyHasForcedChanges
+		}
+	}
+
+	auths, err := types.GrandpaAuthoritiesRawToAuthorities(fc.Auths)
+	if err != nil {
+		return fmt.Errorf("cannot parser GRANDPA authorities to raw authorities: %w", err)
```
qdm12, line 127:
```suggestion
		return fmt.Errorf("cannot parse GRANDPA authorities to raw authorities: %w", err)
```

dot/state/grandpa.go
```diff
@@ -26,42 +37,477 @@ var (
...
+
+	pendingChange := &pendingChange{
+		bestFinalizedNumber: fc.BestFinalizedBlock,
+		nextAuthorities:     auths,
+		announcingHeader:    header,
+		delay:               fc.Delay,
+	}
+
+	s.forcedChanges = append(s.forcedChanges, pendingChange)
+	sort.Sort(s.forcedChanges)
```
qdm12, line 138: If the forcedChanges are already sorted before this append, you could do a more efficient sorted insertion, see https://stackoverflow.com/a/55460931

dot/state/grandpa.go
```diff
@@ -26,42 +37,477 @@ var (
...
+		// neither the finalized header is descendant of the change then skip to the next root
+		if !finalizedHash.Equal(changeNodeHash) {
+			isDescendant, err := s.blockState.IsDescendantOf(changeNodeHash, finalizedHash)
+			if err != nil {
+				return nil, fmt.Errorf("cannot verify ancestry: %w", err)
+			}
+
+			if !isDescendant {
+				continue
+			}
```
qdm12, line 212: The comment is unclear, especially with the `neither`, can you clarify this please 🤔

dot/state/grandpa.go
```diff
@@ -26,42 +37,477 @@ var (
...
+				return nil, errUnfinalizedAncestor
+			}
+		}
+
+		changeNode = root
+		break
+	}
+
+	// if there is no change to be applied then we should keep only
+	// the scheduled changes which belongs to the finalized header
```
qdm12, line 233:
```suggestion
	// the scheduled changes which belong to the finalized header
```

dot/state/grandpa.go
```diff
@@ -26,42 +37,477 @@ var (
...
+			return fmt.Errorf("cannot verify ancestry: %w", err)
+		}
+
+		if isDescendant {
+			onBranchForcedChanges = append(onBranchForcedChanges, forcedChange)
+		}
+	}
+
+	s.forcedChanges = make(orderedPendingChanges, len(onBranchForcedChanges))
+	copy(s.forcedChanges, onBranchForcedChanges)
```
qdm12, line 267: no need for the copy, since `onBranchForcedChanges` is not used anywhere else.
```suggestion
	s.forcedChanges = onBranchForcedChanges
```

dot/state/grandpa.go
```diff
@@ -26,42 +37,477 @@ var (
...
+			return fmt.Errorf("cannot verify ancestry: %w", err)
+		}
+
+		if isDescendant {
+			onBranchScheduledChanges = append(onBranchScheduledChanges, scheduledChange)
+		}
+	}
+
+	s.scheduledChangeRoots = make([]*pendingChangeNode, len(onBranchScheduledChanges))
+	copy(s.scheduledChangeRoots, onBranchScheduledChanges)
```
qdm12, line 291: no need for the copy

```suggestion
	s.scheduledChangeRoots = onBranchScheduledChanges
```

dot/state/grandpa.go
```diff
@@ -26,42 +37,477 @@ var (
...
+}
+
+// forcedChangeOnChainOf walk through the forced change slice looking for
+// a forced change that belong to the same branch as blockHash parameter
+func (s *GrandpaState) forcedChangeOnChainOf(blockHash common.Hash) (forcedChange *pendingChange, err error) {
+	for _, change := range s.forcedChanges {
+		changeHeader := change.announcingHeader
+
+		var isDescendant bool
+		isDescendant, err = s.blockState.IsDescendantOf(
```
qdm12, line 441: nit wouldn't this work as well?

```suggestion
		isDescendant, err := s.blockState.IsDescendantOf(
```

dot/state/grandpa.go
```diff
@@ -26,42 +37,477 @@ var (
...
+		}
+
+		if !isDescendant {
+			continue
+		}
+
+		return change, nil
+	}
+
+	return forcedChange, nil
```
qdm12, line 455:
```suggestion
	return nil, nil
```

dot/state/grandpa.go
```diff
@@ -26,42 +37,477 @@ var (
...
+
+		if err != nil {
+			return nil, fmt.Errorf("cannot verify ancestry: %w", err)
+		}
+
+		if !isDescendant {
+			continue
+		}
+
+		return change, nil
```
qdm12, line 452: nit flip logic around in the for loop

```suggestion
		if isDescendant {
			return change, nil
		}		
```

dot/state/grandpa.go
```diff
@@ -26,42 +37,477 @@ var (
...
+
+		if err != nil {
+			return nil, fmt.Errorf("cannot verify ancestry: %w", err)
+		}
+
+		if !isDescendant {
+			continue
+		}
+
+		return change.change, nil
```
qdm12, line 473: nit

```suggestion
		if isDescendant {
			return change.change, nil
		}
```

dot/state/grandpa.go
```diff
@@ -26,42 +37,477 @@ var (
...
+	}
+
+	return scheduledChange, nil
```
qdm12, line 476:
```suggestion
	return nil, nil
```

dot/state/grandpa.go
```diff
@@ -7,13 +7,24 @@ import (
...
 	"github.com/ChainSafe/gossamer/dot/types"
 	"github.com/ChainSafe/gossamer/lib/common"
 	"github.com/ChainSafe/gossamer/pkg/scale"
 )
 
+var (
+	errPendingScheduledChanges = errors.New("pending scheduled changes needs to be applied")
+	errDuplicateHashes         = errors.New("duplicated hashes")
+	errAlreadyHasForcedChanges = errors.New("already has a forced change")
+	errUnfinalizedAncestor     = errors.New("ancestor with changes not applied")
```
qdm12, line 23: nit maybe rename the error name to match its error string? 🤔

dot/state/grandpa.go
```diff
@@ -7,13 +7,24 @@ import (
...
+
+	ErrNoChanges = errors.New("cannot get the next authority change block number")
```
qdm12, line 25: nit maybe rename the error name to match its error string? 🤔

dot/state/grandpa_changes.go
```diff
@@ -0,0 +1,82 @@
...
+		}
+
+		if imported {
+			return true, nil
+		}
+	}
+
+	isDescendant, err := isDescendantOf(announcingHash, blockHash)
+	if err != nil {
+		return false, fmt.Errorf("cannot define ancestry: %w", err)
```
qdm12, line 72: nit we're not really defining the ancestry, we are trying to find it instead

```suggestion
		return false, fmt.Errorf("cannot determine ancestry: %w", err)
```

dot/state/grandpa_changes.go
```diff
@@ -0,0 +1,82 @@
...
+	}
+
+	if blockNumber <= c.change.announcingHeader.Number {
+		return false, nil
+	}
+
+	for _, childrenNodes := range c.nodes {
+		imported, err := childrenNodes.importScheduledChange(blockHash, blockNumber, pendingChange, isDescendantOf)
+		if err != nil {
+			return false, err
```
qdm12, line 62:
```suggestion
			return false, fmt.Errorf("cannot import scheduled change for pending change node: %w", err)
```

dot/state/grandpa_changes.go
```diff
@@ -0,0 +1,82 @@
...
+
+	if blockHash.Equal(announcingHash) {
+		return false, errDuplicateHashes
+	}
+
+	if blockNumber <= c.change.announcingHeader.Number {
+		return false, nil
+	}
+
+	for _, childrenNodes := range c.nodes {
```
qdm12, line 59: rename from `childrenNodes` to `pendingChangeNode`?

EclesioMeloJunior: I prefree

EclesioMeloJunior: I prefer to keep `childrenNodes` as a reference to inner/nested nodes

dot/digest/digest.go
```diff
@@ -139,42 +158,19 @@ func (h *Handler) handleConsensusDigest(d *types.ConsensusDigest, header *types.
...
-	case types.GrandpaScheduledChange:
-		return h.handleScheduledChange(val, header)
-	case types.GrandpaForcedChange:
-		return h.handleForcedChange(val, header)
-	case types.GrandpaOnDisabled:
-		return nil // do nothing, as this is not implemented in substrate
-	case types.GrandpaPause:
-		return h.handlePause(val)
-	case types.GrandpaResume:
-		return h.handleResume(val)
```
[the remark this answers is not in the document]
EclesioMeloJunior: because we don't support pause/resume digest messages yet, but there is an issue to implement them https://github.com/ChainSafe/gossamer/issues/2556 (I started, but this is not a right priority since substrate doesn't support them as well)

dot/digest/digest.go
```diff
@@ -220,10 +218,14 @@ func (h *Handler) handleBlockImport(ctx context.Context) {
 				continue
 			}
 
-			h.HandleDigests(&block.Header)
-			err := h.handleGrandpaChangesOnImport(block.Header.Number)
+			err := h.HandleDigests(&block.Header)
 			if err != nil {
-				h.logger.Errorf("failed to handle grandpa changes on block import: %s", err)
+				h.logger.Errorf("failed to handle digest: %s", err)
```
qdm12, line 223:
```suggestion
				h.logger.Errorf("failed to handle digests: %s", err)
```

dot/state/grandpa_changes.go
```diff
@@ -0,0 +1,310 @@
...
+type changeTree []*pendingChangeNode
+
+func (ct changeTree) Len() int { return len(ct) }
+func (ct *changeTree) importChange(pendingChange *pendingChange, isDescendantOf isDescendantOfFunc) error {
+	for _, root := range *ct {
+		imported, err := root.importNode(pendingChange.announcingHeader.Hash(),
+			pendingChange.announcingHeader.Number, pendingChange, isDescendantOf)
+
+		if err != nil {
+			return err
```
qdm12, line 187:
```suggestion
		if err != nil {
			return fmt.Errorf("cannot import node: %w", err)
```

dot/state/grandpa_changes.go
```diff
@@ -0,0 +1,310 @@
...
+	isDescendant, err := isDescendantOf(announcingHash, blockHash)
+	if err != nil {
+		return false, fmt.Errorf("cannot check ancestry: %w", err)
+	}
+
+	if !isDescendant {
+		return false, nil
+	}
+
+	pendingChangeNode := &pendingChangeNode{change: pendingChange, nodes: []*pendingChangeNode{}}
```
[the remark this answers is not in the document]
EclesioMeloJunior: Yes, it is possible since we just use `nodes` for looping

dot/state/grandpa_changes.go
```diff
@@ -0,0 +1,310 @@
...
+}
+
+// changeTree keeps track of the changes per fork allowing
+// n forks in the same structure, this structure is intended
+// to be an acyclic directed graph where the change nodes are
+// placed by descendency order and number, you can ensure an
+// node ancestry using the `isDescendantOfFunc`
+type changeTree []*pendingChangeNode
+
+func (ct changeTree) Len() int { return len(ct) }
```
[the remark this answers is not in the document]
EclesioMeloJunior: This is my vscode linter acting and adding this line, do you mind if we keep it as it is?

dot/state/grandpa_changes.go
EclesioMeloJunior: This error is already wrapped by the caller and by the calling function, so adding this wrapper the error will look like this: `cannot import scheduled change: cannot import node: cannot import node: ..."`

dot/digest/digest.go
```diff
@@ -91,44 +76,80 @@ func (h *Handler) Stop() error {
...
-// HandleDigests handles consensus digests for an imported block
-func (h *Handler) HandleDigests(header *types.Header) {
-	for i, d := range header.Digest.Types {
-		val, ok := d.Value().(types.ConsensusDigest)
-		if !ok {
+// checkForGRANDPAForcedChanges removes any GrandpaScheduledChange in the presence of a
+// GrandpaForcedChange in the same block digest, returning a new slice of types.ConsensusDigest
+func checkForGRANDPAForcedChanges(digests []types.ConsensusDigest) ([]types.ConsensusDigest, error) {
+	var hasForcedChange bool
+	digestsWithoutScheduled := make([]types.ConsensusDigest, 0, len(digests))
```
timwu20, line 125: why are defining the capacity of the slice, but using `append` on lines 142 and 144.  `append` will actually double the capacity if it overflows.

dot/state/grandpa.go
```diff
@@ -7,13 +7,23 @@ import (
...
 	"github.com/ChainSafe/gossamer/dot/types"
 	"github.com/ChainSafe/gossamer/lib/common"
 	"github.com/ChainSafe/gossamer/pkg/scale"
 )
 
+var (
+	errPendingScheduledChanges = errors.New("pending scheduled changes needs to be applied")
+	errDuplicateHashes         = errors.New("duplicated hashes")
+	errAlreadyHasForcedChange  = errors.New("already has a forced change")
+	errUnfinalizedAncestor     = errors.New("unfinalized ancestor")
```
timwu20, line 22: any reason why these are unexported?

dot/state/grandpa_changes.go
```diff
@@ -0,0 +1,313 @@
...
+func (p pendingChange) String() string {
+	return fmt.Sprintf("announcing header: %s (%d), delay: %d, next authorities: %d",
+		p.announcingHeader.Hash(), p.announcingHeader.Number, p.delay, len(p.nextAuthorities))
+}
+
+func (p *pendingChange) effectiveNumber() uint {
+	return p.announcingHeader.Number + uint(p.delay)
+}
+
+type orderedPendingChanges []*pendingChange
```
timwu20, line 33: why `[]*pendingChange` vs `[]pendingChange`.  `pendingChange` doesn't contain a lot of data, so I don't see the harm in just copying the type and reducing complexity when dealing with a slice of pendingChanges.

dot/state/grandpa.go
```diff
@@ -26,40 +36,276 @@ var (
...
+	}
+
+	pendingChange := &pendingChange{
+		bestFinalizedNumber: fc.BestFinalizedBlock,
+		nextAuthorities:     auths,
+		announcingHeader:    header,
+		delay:               fc.Delay,
+	}
+
+	err = s.forcedChanges.importChange(pendingChange, s.blockState.IsDescendantOf)
```
timwu20, line 121: seems really odd to always call `importChange, pruneChanges, etc..` with this `IsDescendantOf` function.  I think it makes more sense to make `orderedPendingChanges` a struct with a field of type `isDescendantOfFunc`.  That way you include it as state for the type `orderedPendingChanges`.

dot/state/grandpa_changes.go
```diff
@@ -0,0 +1,313 @@
...
+	*oc = onBranchForcedChanges
+	return nil
+}
+
+type pendingChangeNode struct {
+	change *pendingChange
+	nodes  []*pendingChangeNode
+}
+
+func (c *pendingChangeNode) importNode(blockHash common.Hash, blockNumber uint, pendingChange *pendingChange,
```
timwu20, line 140: this function is a little difficult to read knowing that it calls itself recursively, and with all the different return cases.  Is there a way to refactor this to clearly show all the return cases.  If a `pendingChangeNode` has children and the supplied `pendingChange` not successfully imported from one of the children, then it's supposed to append it as a child? I'd consider refactoring this code or putting some more comments.

dot/state/grandpa_changes.go
```diff
@@ -0,0 +1,313 @@
...
+	orderedChanges[idxToInsert] = pendingChange
+	*oc = orderedChanges
+
+	return nil
+}
+
+// pruneChanges will remove changes whose are not descendant of the hash argument
+// this function updates the current state of the change tree
+func (oc *orderedPendingChanges) pruneChanges(hash common.Hash, isDescendantOf isDescendantOfFunc) error {
+	onBranchForcedChanges := make([]*pendingChange, 0, oc.Len())
```
timwu20, line 118: Using `append` on slice with a defined capacity will double the capacity on overflow.

dot/state/grandpa_changes.go
```diff
@@ -0,0 +1,313 @@
...
+		change: pendingChange,
+	}
+
+	*ct = append(*ct, pendingChangeNode)
+	return nil
+}
+
+// lookupChangesWhere returns the first change which satisfy the
+// condition whithout modify the current state of the change tree
+func (ct changeTree) lookupChangeWhere(condition conditionFunc[*pendingChangeNode]) (
```
timwu20, line 211: why do we need to use generics in this case? Seems like all calls of `lookupChangeWhere` would only accept type `*pendingChangeNode`, so why not simplify and accept a function with signature `func (n *pendingChangeNode) (bool, err)`

dot/state/grandpa.go
```diff
@@ -26,40 +36,276 @@ var (
...
+	})
+	if err != nil {
+		return fmt.Errorf("cannot check pending changes while applying forced change: %w", err)
+	} else if dependant != nil {
+		return fmt.Errorf("%w: %s", errPendingScheduledChanges, dependant.change)
+	}
+
+	logger.Debugf("applying forced change: %s", forcedChange)
+
+	// send the telemetry s messages here
```
timwu20, line 230: typo?

timwu20 (changes requested):

dot/digest/digest.go
EclesioMeloJunior: Yes, but in this case, the maximum amount of data in the `digestsWithoutScheduled` slice is `len(digest)`, so there is no problem to use append as it will never overflow the capacity

dot/state/grandpa.go
EclesioMeloJunior: Not being used anywhere, it is just as an error sentinel/signalization

dot/state/grandpa_changes.go
EclesioMeloJunior: Yes, but in this case, the maximum amount of data in the `onBranchForcedChanges` slice is `oc.Len()`, so there is no problem to use append as it will never overflow the capacity

dot/state/grandpa_changes.go
EclesioMeloJunior: Not really, to lookup in the `orderedPendingChanges` the `lookupChangeWhere` function should iterate over `*pendingChange` and for the `changeTree` the `lookupChangeWhere` function should iterate over `*pendingChangeNode`.

I decided to use the generic function type `type conditionFunc[T any] func(T) (bool, error)` as it is the same as creating the parameter for each lookup function, like: `func(*pendingChangeNode) (bool, error)` and `func(*pendingChange) (bool, error)`

dot/state/grandpa.go
EclesioMeloJunior: I don't think keeping the function as a field in the `orderedPendingChanges` is good as we will keep this reference in memory until we garbage collect, but passing it as an argument then once the function returns the argument will be dropped

timwu20: I don't think there's much more overhead keeping a single reference to `s.blockState.IsDescendantOf` in a struct that includes an attribute for `[]pendingChange`.  The question you should ask yourself is, is the type `orderedPendingChanges`  intended to be stateful.  I would argue yes, since it requires a subset of functions of `blockState`.  You could just add it to a constructor for `orderedPendingChanges` when it's called in `NewGrandpaState`.

timwu20 (approved):

edwardmack (approved):

dot/digest/digest_test.go
```diff
@@ -13,17 +14,20 @@ import (
...
 	"github.com/ChainSafe/gossamer/lib/keystore"
 	"github.com/ChainSafe/gossamer/pkg/scale"
 	"github.com/golang/mock/gomock"
+	"github.com/gtank/merlin"
 
 	"github.com/stretchr/testify/require"
 )
 
 //go:generate mockgen -destination=mock_telemetry_test.go -package $GOPACKAGE github.com/ChainSafe/gossamer/dot/telemetry Client
+//go:generate mockgen -destination=mock_grandpa_test.go -package $GOPACKAGE . GrandpaState
```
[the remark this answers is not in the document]
EclesioMeloJunior: Agree with you but we can do this in another PR since it is out of scope for this PR

PR #2527, opened 2022-05-09
fix(dot/subscription): check websocket message from untrusted data

### LINKED ISSUE #2413: Assumption of JSON Fields from Untrusted Input (GSR-15)
Websocket RPC messages are unmarshalled from JSON into a map, which should have certain fields.

If the expected fields are not present then `nil` will be returned. This may lead to unexpected `nil` pointer exceptions being raised when these pointers are used.

The issue may be resolved by first ensuring each of the required fields are present in the map before accessing them.

```go
var reqid float64
if val, ok := msg["id"]; ok {
	if reqid, ok := val.(float64); !ok {
		// handle error
		return
	}
} else {
	// handle error
	return
}
var method string
if val, ok := msg["method"]; ok {
	if method, ok := val.(string); !ok {
		// handle error
		return
	}
} else {
	// handle error
	return
}
```

EclesioMeloJunior:
## Changes

- Created a struct with the required fields: `ID float64`, `Method string` and `Params any` to ensure the types and avoid conversions over `interface{}`
- Created a condition to don't allow an empty `Method` field

## Tests

```sh
go test github.com/ChainSafe/gossamer/dot/rpc/subscription
```

## Issues

- Closes #2413

## Primary Reviewer

@jimjbrettj

### DIFF
--- a/dot/rpc/subscription/websocket.go
+++ b/dot/rpc/subscription/websocket.go
@@ -69,15 +69,14 @@ func (c *WSConn) HandleConn() {
 		mbytes, err := c.readWebsocketMessage()
 		if errors.Is(err, errCannotReadFromWebsocket) {
 			return
-		} else if errors.Is(err, errCannotUnmarshalMessage) {
+		} else if err != nil {
 			c.safeSendError(0, big.NewInt(InvalidRequestCode), InvalidRequestMessage)
-			continue
 		}
 
 		msg := new(websocketMessage)
 		err = json.Unmarshal(mbytes, &msg)
 		if err != nil {
-			logger.Debugf("failed to unmarshal websocket request message: %s", err)
+			c.safeSendError(0, big.NewInt(InvalidRequestCode), InvalidRequestMessage)
 			continue
 		}
 
--- a/dot/rpc/subscription/websocket.go
+++ b/dot/rpc/subscription/websocket.go
@@ -56,7 +56,7 @@ func (c *WSConn) readWebsocketMessage() (bytes []byte, err error) {
 	_, bytes, err = c.Wsconn.ReadMessage()
 	if err != nil {
 		logger.Debugf("websocket failed to read message: %s", err)
-		return bytes, errCannotReadFromWebsocket
+		return nil, errCannotReadFromWebsocket
 	}
 
 	logger.Tracef("websocket message received: %s", string(bytes))
@@ -77,7 +77,7 @@ func (c *WSConn) HandleConn() {
 		msg := new(websocketMessage)
 		err = json.Unmarshal(mbytes, &msg)
 		if err != nil {
-			logger.Debugf("websocket failed to unmarshal request message: %s", err)
+			logger.Debugf("failed to unmarshal websocket request message: %s", err)
 			continue
 		}
 
--- a/dot/rpc/subscription/websocket.go
+++ b/dot/rpc/subscription/websocket.go
@@ -51,50 +51,55 @@ type WSConn struct {
 }
 
 // readWebsocketMessage will read and parse the message data to a string->interface{} data
-func (c *WSConn) readWebsocketMessage() (bytes []byte, err error) {
-	_, bytes, err = c.Wsconn.ReadMessage()
+func (c *WSConn) readWebsocketMessage() (rawBytes []byte, wsMessage *websocketMessage, err error) {
+	_, rawBytes, err = c.Wsconn.ReadMessage()
 	if err != nil {
-		logger.Debugf("websocket failed to read message: %s", err)
-		return nil, errCannotReadFromWebsocket
+		return nil, nil, fmt.Errorf("%w: %s", errCannotReadFromWebsocket, err.Error())
 	}
 
-	logger.Tracef("websocket message received: %s", string(bytes))
-	return bytes, nil
+	msg := new(websocketMessage)
+	err = json.Unmarshal(rawBytes, &msg)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	return rawBytes, msg, nil
 }
 
 // HandleConn handles messages received on websocket connections
 func (c *WSConn) HandleConn() {
+	defer c.Wsconn.Close()
+
 	for {
-		mbytes, err := c.readWebsocketMessage()
+		rawBytes, wsMessage, err := c.readWebsocketMessage()
 		if err != nil {
-			return
-		}
+			logger.Debugf("websocket failed to read message: %s", err)
+			if errors.Is(err, errCannotReadFromWebsocket) {
+				return
+			}
 
-		msg := new(websocketMessage)
-		err = json.Unmarshal(mbytes, &msg)
-		if err != nil {
 			c.safeSendError(0, big.NewInt(InvalidRequestCode), InvalidRequestMessage)
 			continue
 		}
 
-		if msg.Method == "" {
+		logger.Tracef("websocket message received: %s", string(rawBytes))
+		if wsMessage.Method == "" {
 			c.safeSendError(0, big.NewInt(InvalidRequestCode), InvalidRequestMessage)
 			continue
 		}
 
-		logger.Debugf("ws method %s called with params %v", msg.Method, msg.Params)
-
-		if !strings.Contains(msg.Method, "_unsubscribe") && !strings.Contains(msg.Method, "_unwatch") {
-			setupListener := c.getSetupListener(msg.Method)
+		logger.Debugf("ws method %s called with params %v", wsMessage.Method, wsMessage.Params)
+		if !strings.Contains(wsMessage.Method, "_unsubscribe") && !strings.Contains(wsMessage.Method, "_unwatch") {
+			setupListener := c.getSetupListener(wsMessage.Method)
 
 			if setupListener == nil {
-				c.executeRPCCall(mbytes)
+				c.executeRPCCall(rawBytes)
 				continue
 			}
 
-			listener, err := setupListener(msg.ID, msg.Params)
+			listener, err := setupListener(wsMessage.ID, wsMessage.Params)
 			if err != nil {
-				logger.Warnf("failed to create listener (method=%s): %s", msg.Method, err)
+				logger.Warnf("failed to create listener (method=%s): %s", wsMessage.Method, err)
 				continue
 			}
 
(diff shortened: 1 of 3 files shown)


qdm12 (approved):
Nice :+1:

dot/rpc/subscription/websocket_test.go
```diff
@@ -21,6 +21,53 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+func TestWSConn_CheckWebsocketInvalidData(t *testing.T) {
```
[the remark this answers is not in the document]
EclesioMeloJunior: Yes, created a issue to address this -> https://github.com/ChainSafe/gossamer/issues/2555

dot/rpc/subscription/websocket.go
```diff
@@ -52,28 +52,27 @@ type WSConn struct {
...
+		return bytes, msg, errCannotUnmarshalMessage
 	}
 
-	return mbytes, msg, nil
+	return bytes, msg, nil
 }
 
-//HandleComm handles messages received on websocket connections
-func (c *WSConn) HandleComm() {
+//HandleConn handles messages received on websocket connections
```
qdm12, line 74: supernit

```suggestion
// HandleConn handles messages received on websocket connections
```

dot/rpc/subscription/websocket.go
```diff
@@ -52,28 +52,27 @@ type WSConn struct {
 }
 
 // readWebsocketMessage will read and parse the message data to a string->interface{} data
-func (c *WSConn) readWebsocketMessage() ([]byte, *websocketMessage, error) {
-	_, mbytes, err := c.Wsconn.ReadMessage()
+func (c *WSConn) readWebsocketMessage() (bytes []byte, msg websocketMessage, err error) {
```
qdm12, line 55: should we keep the raw bytes inside the `websocketMessage` struct? 🤔

qdm12 (approved):

dot/rpc/subscription/websocket.go
EclesioMeloJunior: I don't think so, the raw bytes don't belong to the `websocketMessage` struct as it is used to create one IMO

dot/rpc/subscription/websocket.go
```diff
@@ -22,6 +22,12 @@ import (
 	"github.com/gorilla/websocket"
 )
 
+type websocketMessage struct {
+	ID     float64 `json:"id"`
+	Method string  `json:"method"`
+	Params any
```
qdm12, line 28: 🤯

dot/rpc/subscription/websocket.go
```diff
@@ -46,87 +52,82 @@ type WSConn struct {
...
-	logger.Tracef("websocket message received: %s", string(mbytes))
-
-	// determine if request is for subscribe method type
-	var msg map[string]interface{}
-	err = json.Unmarshal(mbytes, &msg)
-
-	if err != nil {
-		logger.Debugf("websocket failed to unmarshal request message: %s", err)
-		return nil, nil, errCannotUnmarshalMessage
+		return bytes, errCannotReadFromWebsocket
```
qdm12, line 59: nit just to be clear / explicit to the maxxx

```suggestion
		return nil, errCannotReadFromWebsocket
```

dot/rpc/subscription/websocket.go
```diff
@@ -46,87 +52,82 @@ type WSConn struct {
...
+			c.safeSendError(0, big.NewInt(InvalidRequestCode), InvalidRequestMessage)
+			continue
 		}
 
-		if errors.Is(err, errCannotUnmarshalMessage) {
-			c.safeSendError(0, big.NewInt(InvalidRequestCode), InvalidRequestMessage)
+		msg := new(websocketMessage)
+		err = json.Unmarshal(mbytes, &msg)
+		if err != nil {
+			logger.Debugf("websocket failed to unmarshal request message: %s", err)
```
qdm12, line 80: nit

```suggestion
			logger.Debugf("failed to unmarshal websocket request message: %s", err)
```

qdm12 (approved):

dot/rpc/subscription/websocket.go
```diff
@@ -46,87 +51,81 @@ type WSConn struct {
...
 				continue
 			}
 		}
 
 		err = listener.Stop()
 		if err != nil {
-			logger.Warnf("failed to stop listener goroutine (method=%s): %s", method, err)
-			c.safeSend(newBooleanResponseJSON(false, reqid))
+			logger.Warnf("failed to stop listener goroutine (method=%s): %s", msg.Method, err)
+			c.safeSend(newBooleanResponseJSON(false, msg.ID))
```
qdm12, line 125: This is not covered by tests, do you think you can add a test case for this one?

dot/rpc/subscription/websocket.go
```diff
@@ -46,87 +51,81 @@ type WSConn struct {
...
 			if setupListener == nil {
 				c.executeRPCCall(mbytes)
 				continue
 			}
 
-			listener, err := setupListener(reqid, params)
+			listener, err := setupListener(msg.ID, msg.Params)
 			if err != nil {
-				logger.Warnf("failed to create listener (method=%s): %s", method, err)
+				logger.Warnf("failed to create listener (method=%s): %s", msg.Method, err)
```
qdm12, line 99: This is not covered by tests, do you think you can add a test case for this one?

dot/rpc/subscription/websocket.go
```diff
@@ -46,87 +51,81 @@ type WSConn struct {
...
-func (c *WSConn) HandleComm() {
+// HandleConn handles messages received on websocket connections
+func (c *WSConn) HandleConn() {
 	for {
-		mbytes, msg, err := c.readWebsocketMessage()
+		mbytes, err := c.readWebsocketMessage()
 		if errors.Is(err, errCannotReadFromWebsocket) {
 			return
+		} else if err != nil {
+			c.safeSendError(0, big.NewInt(InvalidRequestCode), InvalidRequestMessage)
```
qdm12, line 72: This is not covered by tests, do you think you can add a test case for this one?

dot/rpc/subscription/websocket.go
EclesioMeloJunior: This is being covered by the `TestWSConn_HandleConn` test, we try to connect to the `StorageChangeListener` without the `Storage State` then it will not set up the listener and return an error through the WebSocket connection that we assert.

dot/rpc/subscription/websocket.go
EclesioMeloJunior: We have tests that cover `*_unsubcribe` actions, I'm not sure if it is needed to implement more tests in this PR.

dot/rpc/subscription/websocket.go
EclesioMeloJunior: Actually, this branch condition is not needed as the `readWebsocketMessage` function will never return other error than `errCannotReadFromWebsocket`

dot/rpc/subscription/websocket.go
qdm12: Yes please add them / refactor existing tests 😉
We should especially pay attention to test cover new code added.

dot/rpc/subscription/websocket.go
```diff
@@ -22,12 +22,17 @@ import (
 	"github.com/gorilla/websocket"
 )
 
+type websocketMessage struct {
+	ID     float64 `json:"id"`
+	Method string  `json:"method"`
+	Params any
```
kishansagathiya, line 28: why not `json:"params"` here?

dot/rpc/subscription/websocket.go
```diff
@@ -46,87 +51,79 @@ type WSConn struct {
...
 			continue
 		}
 
-		params := msg["params"]
-		reqid := msg["id"].(float64)
-		method := msg["method"].(string)
+		if msg.Method == "" {
+			c.safeSendError(0, big.NewInt(InvalidRequestCode), InvalidRequestMessage)
+			continue
+		}
```
kishansagathiya, line 83: This is still reading websocket message. The selected piece of code could be moved inside `readWebsocketMessage` and then `readWebsocketMessage` could return `websocketMessage, err`

dot/rpc/subscription/websocket_test.go
```diff
@@ -21,12 +21,59 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-func TestWSConn_HandleComm(t *testing.T) {
+func TestWSConn_CheckWebsocketInvalidData(t *testing.T) {
```
kishansagathiya, line 24: I am not super sure, but I think if you move this test after `TestWSConn_HandleConn`, git would understand the differ better.

kishansagathiya (changes requested):
Apart from a few small comments, looks pretty good!!

dot/rpc/subscription/websocket.go
```diff
@@ -46,87 +51,84 @@ type WSConn struct {
...
 
-		params := msg["params"]
-		reqid := msg["id"].(float64)
-		method := msg["method"].(string)
+		logger.Tracef("websocket message received: %s", string(rawBytes))
+
+		if wsMessage.Method == "" {
+			c.safeSendError(0, big.NewInt(InvalidRequestCode), InvalidRequestMessage)
+			continue
+		}
```
kishansagathiya, line 88: Could move this inside `readWebsocketMessage`? 

No need to return rawBytes as well
```
func (c *WSConn) readWebsocketMessage() ( wsMessage *websocketMessage, err error)
```

EclesioMeloJunior: @kishansagathiya actually, we need to return the `rawBytes` because if we don't find a listener we execute an RPC call with `rawBytes`, on line 96

EclesioMeloJunior: @kishansagathiya I've moved the condition to `readWebsocketMessage()`

kishansagathiya: > we need to return the rawBytes because if we don't find a listener we execute an RPC call with rawBytes, on line 96

Ah, did not notice that, my bad!

kishansagathiya (approved):
:tada:

dot/rpc/subscription/websocket.go
EclesioMeloJunior: @qdm12 the tests assert the full WebSocket response and the changes I introduced are being testes by `TestWSConn_CheckWebsocketInvalidData`, so I will keep the `TestWSConn_HandleConn` as it is since I've opened an issue to improve/refactor it -> https://github.com/ChainSafe/gossamer/issues/2555

PR #2529, opened 2022-05-09
fix(dot/subscription): unsafe type casting from untrusted input

### LINKED ISSUE #2412: Unsafe Type Casting from Untrusted Input (GSR-14)
Websocket RPC messages are unmarshalled from JSON input before being processed.

After the JSON has been unmarshalled as a `map[string]interface{}` the fields are converted to their required types.

It is unsafe to assume data decoded from the websockets has the correct type as incorrect types will panic without using safe casting.

For example a malicious user could send a JSON message with `id` field as a `string` then `reqid := msg["id"].(float64)` will panic.

Consider doing safe, checked casting for all untrusted input. An example can be seen below.

Note: see GSR-15 for further additions to this code.

The `params` value also requires safe casting, however this must be done in each of the RPC methods individually when the type of `params` is known, such as `initExtrinsicWatch()`. 

```go
reqid, ok := msg["id"].(float64)
if !ok {
    // Handle Error
}
method, ok := msg["method"].(string)
if !ok {
    // Handle Error
}
```

EclesioMeloJunior:
## Changes

- Add checks in methods `initExtrinsicWatch` and `initStorageChangeListener` to safely casting types from untrusted inputs

## Tests

```sh
go test github.com/ChainSafe/gossamer/dot/rpc/subscription
```

## Issues

- Closes #2412

## Primary Reviewer

@edwardmack

### DIFF
--- a/dot/rpc/subscription/websocket.go
+++ b/dot/rpc/subscription/websocket.go
@@ -163,6 +163,8 @@ func (c *WSConn) initStorageChangeListener(reqID float64, params interface{}) (L
 	}
 
 	switch filter := params.(type) {
+	case string:
+		stgobs.filter[filter] = []byte{}
 	case []string:
 		for _, key := range filter {
 			stgobs.filter[key] = []byte{}
@@ -176,8 +178,6 @@ func (c *WSConn) initStorageChangeListener(reqID float64, params interface{}) (L
 
 			stgobs.filter[key] = []byte{}
 		}
-	case string:
-		stgobs.filter[filter] = []byte{}
 	default:
 		return nil, fmt.Errorf("expected type equals string or []string. got %T", params)
 	}
--- a/dot/rpc/subscription/websocket.go
+++ b/dot/rpc/subscription/websocket.go
@@ -35,7 +35,6 @@ type httpclient interface {
 var (
 	errUnexpectedType          = errors.New("unexpected type")
 	errUnexpectedParamLen      = errors.New("unexpected params length")
-	errCannotUnmarshalMessage  = errors.New("cannot unmarshal websocket message data")
 	errCannotReadFromWebsocket = errors.New("cannot read message from websocket")
 	errEmptyMethod             = errors.New("empty method")
 )
--- a/dot/rpc/subscription/websocket.go
+++ b/dot/rpc/subscription/websocket.go
@@ -27,7 +27,6 @@ type httpclient interface {
 }
 
 var (
-	errNilParameters           = errors.New("expecting non nil params. got nil")
 	errCannotReadFromWebsocket = errors.New("cannot read message from websocket")
 	errCannotUnmarshalMessage  = errors.New("cannot unmarshal webasocket message data")
 )
@@ -163,25 +162,24 @@ func (c *WSConn) initStorageChangeListener(reqID float64, params interface{}) (L
 		wsconn: c,
 	}
 
-	pA, ok := params.([]interface{})
-	if !ok {
-		return nil, fmt.Errorf("unknown parameter type")
-	}
-	for _, param := range pA {
-		switch p := param.(type) {
-		case []interface{}:
-			for _, pp := range param.([]interface{}) {
-				data, ok := pp.(string)
-				if !ok {
-					return nil, fmt.Errorf("unknown parameter type")
-				}
-				stgobs.filter[data] = []byte{}
-			}
-		case string:
-			stgobs.filter[p] = []byte{}
-		default:
-			return nil, fmt.Errorf("unknown parameter type")
+	switch filter := params.(type) {
+	case []string:
+		for _, key := range filter {
+			stgobs.filter[key] = []byte{}
 		}
+	case []interface{}:
+		for _, interfaceKey := range params.([]interface{}) {
+			key, ok := interfaceKey.(string)
+			if !ok {
+				return nil, fmt.Errorf("expected type equals string. got %T", interfaceKey)
+			}
+
+			stgobs.filter[key] = []byte{}
+		}
+	case string:
+		stgobs.filter[filter] = []byte{}
+	default:
+		return nil, fmt.Errorf("expected type equals string or []string. got %T", params)
 	}
 
 	c.mu.Lock()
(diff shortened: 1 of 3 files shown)


dot/rpc/subscription/websocket.go
```diff
@@ -26,8 +26,11 @@ type httpclient interface {
 	Do(*http.Request) (*http.Response, error)
 }
 
-var errCannotReadFromWebsocket = errors.New("cannot read message from websocket")
-var errCannotUnmarshalMessage = errors.New("cannot unmarshal webasocket message data")
+var (
+	errCannotReadFromWebsocket = errors.New("cannot read message from websocket")
+	errCannotUnmarshalMessage  = errors.New("cannot unmarshal webasocket message data")
```
qdm12, line 31:
```suggestion
	errCannotUnmarshalMessage  = errors.New("cannot unmarshal websocket message data")
```

dot/rpc/subscription/websocket.go
```diff
@@ -159,25 +162,24 @@ func (c *WSConn) initStorageChangeListener(reqID float64, params interface{}) (L
...
-			stgobs.filter[p] = []byte{}
-		default:
-			return nil, fmt.Errorf("unknown parameter type")
+
+			stgobs.filter[key] = []byte{}
 		}
+	case string:
+		stgobs.filter[filter] = []byte{}
+	default:
+		return nil, fmt.Errorf("expected type equals string or []string. got %T", params)
```
qdm12, line 182: use a sentinel error

dot/rpc/subscription/websocket.go
```diff
@@ -159,25 +162,24 @@ func (c *WSConn) initStorageChangeListener(reqID float64, params interface{}) (L
...
-					return nil, fmt.Errorf("unknown parameter type")
-				}
-				stgobs.filter[data] = []byte{}
+	switch filter := params.(type) {
+	case []string:
+		for _, key := range filter {
+			stgobs.filter[key] = []byte{}
+		}
+	case []interface{}:
+		for _, interfaceKey := range params.([]interface{}) {
```
qdm12, line 171: I think that would work right?

```suggestion
		for _, interfaceKey := range filter {
```

dot/rpc/subscription/websocket.go
```diff
@@ -265,14 +267,30 @@ func (c *WSConn) initAllBlocksListerner(reqID float64, _ interface{}) (Listener,
...
+		encodedExtrinsic = encodedHex[0]
+	case []interface{}:
+		if len(encodedHex) != 1 {
+			return nil, fmt.Errorf("expected only 1 param item. got %d", len(encodedHex))
+		}
+
+		var ok bool
+		encodedExtrinsic, ok = encodedHex[0].(string)
+		if !ok {
+			return nil, fmt.Errorf("expected type equals string. got %T", encodedHex[0])
```
qdm12, line 286: use a sentinel error

dot/rpc/subscription/websocket.go
```diff
@@ -265,14 +267,30 @@ func (c *WSConn) initAllBlocksListerner(reqID float64, _ interface{}) (Listener,
...
+		}
+	default:
+		return nil, fmt.Errorf("expected type equals []string. got %T", params)
```
qdm12, line 289: use a sentinel error

also that should be `[]string or []interface{}` I guess? :thinking:

dot/rpc/subscription/websocket.go
```diff
@@ -159,25 +162,24 @@ func (c *WSConn) initStorageChangeListener(reqID float64, params interface{}) (L
...
-				if !ok {
-					return nil, fmt.Errorf("unknown parameter type")
-				}
-				stgobs.filter[data] = []byte{}
+	switch filter := params.(type) {
+	case []string:
+		for _, key := range filter {
+			stgobs.filter[key] = []byte{}
+		}
+	case []interface{}:
```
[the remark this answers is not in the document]
EclesioMeloJunior: > Couldn't a string be interpreted as []interface{} (slice of runes)?

Not really -> https://go.dev/play/p/cCeKJJsIBU8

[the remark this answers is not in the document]
EclesioMeloJunior: the method `initStorageChangeListener(reqID float64, params interface{})` receives the `params` which is a dynamic websocket data like:  multi-type slice`["hash", 10]`, a hex string or just a number, so adding this case I make sure that we cover the type`[]interface{"a", "c"}` as well as `[]string{"a", "c"}`

dot/rpc/subscription/websocket.go
EclesioMeloJunior: The method `initExtrinsicWatch(reqID float64, params interface{})` expects a slice of 1 item that must be a string so there is the `case []interface{}` just to cover the case of `[]interface{"a"}` but the data should be always []string of 1 item

dot/rpc/subscription/websocket.go
```diff
@@ -265,14 +267,30 @@ func (c *WSConn) initAllBlocksListerner(reqID float64, _ interface{}) (Listener,
...
 
-	if len(pA) != 1 {
-		return nil, errors.New("expecting only one parameter")
+	switch encodedHex := params.(type) {
+	case []string:
+		if len(encodedHex) != 1 {
+			return nil, fmt.Errorf("expected only 1 param item. got %d", len(encodedHex))
+		}
+		encodedExtrinsic = encodedHex[0]
+	case []interface{}:
```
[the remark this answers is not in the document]
EclesioMeloJunior: https://github.com/ChainSafe/gossamer/pull/2529#discussion_r872648722

dot/rpc/subscription/websocket.go
```diff
@@ -265,14 +267,30 @@ func (c *WSConn) initAllBlocksListerner(reqID float64, _ interface{}) (Listener,
 }
 
 func (c *WSConn) initExtrinsicWatch(reqID float64, params interface{}) (Listener, error) {
```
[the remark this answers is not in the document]
EclesioMeloJunior: No, this param should be an `interface{}` as it is what we receive from websocket, so each function is responsible to cast the data in the format it expects

dot/rpc/subscription/websocket.go
```diff
@@ -159,25 +164,24 @@ func (c *WSConn) initStorageChangeListener(reqID float64, params interface{}) (L
...
-		case []interface{}:
-			for _, pp := range param.([]interface{}) {
-				data, ok := pp.(string)
-				if !ok {
-					return nil, fmt.Errorf("unknown parameter type")
-				}
-				stgobs.filter[data] = []byte{}
+	switch filter := params.(type) {
+	case string:
+		stgobs.filter[filter] = []byte{}
```
qdm12, line 169: why do we set this to an empty slice and not `nil`? Setting it to `nil` would still set the key in the map, and we can then set it or append to it as well later.

dot/rpc/subscription/websocket.go
qdm12: Oh yes I was talking about the error string here

dot/rpc/subscription/websocket.go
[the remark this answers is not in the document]
qdm12: Awesome TIL!

dot/rpc/subscription/websocket.go
qdm12: I was talking about the error string. I made a code suggestion below 😉

dot/rpc/subscription/websocket_test.go
```diff
@@ -207,7 +206,7 @@ func TestWSConn_HandleComm(t *testing.T) {
 	wsconn.CoreAPI = modules.NewMockCoreAPI()
 	wsconn.BlockAPI = nil
 	wsconn.TxStateAPI = modules.NewMockTransactionStateAPI()
-	listner, err := wsconn.initExtrinsicWatch(0, []interface{}{"NotHex"})
+	listner, err := wsconn.initExtrinsicWatch(0, []string{"NotHex"})
```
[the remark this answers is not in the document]
EclesioMeloJunior: for `initExtrinsicWatch` the cases: `[]string` and `[]interface` being tested and for `initStorageChangeListener` the cases: `string`, `[]string` and `[]interface` are being tested as well, I've created [an issue](https://github.com/ChainSafe/gossamer/issues/2555) to refactor those tests to support test cases and then they will be more readable

dot/rpc/subscription/websocket.go
qdm12: Just a few bytes less. I guess not a big deal either, so up to you!

dot/rpc/subscription/websocket.go
```diff
@@ -265,14 +271,32 @@ func (c *WSConn) initAllBlocksListerner(reqID float64, _ interface{}) (Listener,
...
 func (c *WSConn) initExtrinsicWatch(reqID float64, params interface{}) (Listener, error) {
-	pA := params.([]interface{})
+	var encodedExtrinsic string
 
-	if len(pA) != 1 {
-		return nil, errors.New("expecting only one parameter")
+	switch encodedHex := params.(type) {
+	case []string:
+		if len(encodedHex) != 1 {
+			return nil, fmt.Errorf("%w: %d, expected 1 param", errUnexpectedParamLen, len(encodedHex))
```
jimjbrettj, line 279: Nit: Should we be more clear on what we got vs what was expected? Something like fmt.Errorf("%w: expected 1 param, got %d", errUnexpectedParamLen, len(encodedHex)).

Issue #2418, opened 2022-03-21
Implement storage trie version 1


noot:
## Task summary

Current storage trie is version 0, we need to upgrade to version 1 which has an [updated header format](https://spec.polkadot.network/#defn-node-header) which especially adds inlined hashes nodes.

Trie encoding/hashing functions might need to accept a v1 parameter or some other method for versioning the trie encoding

Discussions:

- https://github.com/paritytech/substrate/discussions/11607

Related Gossamer issues:

- #2374
- #2329 (good information in there!)

Related Gossamer PRs:

- #2369  (good information in there too!)
- #2477 (merge in feature branch later)

Related Rust code:

- [**Codec**](https://github.com/paritytech/substrate/blob/345e3b1fcbbb0c479a031c45859d7154e7c4b5f7/primitives/trie/src/trie_codec.rs)
- [**Node header**](https://github.com/paritytech/substrate/blob/345e3b1fcbbb0c479a031c45859d7154e7c4b5f7/primitives/trie/src/node_header.rs#L85)
- Commit for inner hash in trie crate: https://github.com/paritytech/trie/commit/cbc0283ceb014ef29c5a6c6dbe22a66e8a712c48
- Substrate PR for trie upgrade: https://github.com/paritytech/substrate/pull/9732
- [Smoldot PR](https://github.com/smol-dot/smoldot/pull/230/files)

Related documentation:

- [State storage trie structure spec](https://spec.polkadot.network/#sect-state-storage-trie-structure)
- [v1 trie diff](https://github.com/w3f/polkadot-spec/pull/478/files)

Other issues:

- https://github.com/ComposableFi/ibc-go/issues/6

WASM functions to implement:

- [`ext_storage_root_version_2`](https://spec.polkadot.network/#sect-ext-storage-root-version-2)
- [`ext_default_child_storage_root_version_2`](https://spec.polkadot.network/#_version_2_prototype_4)
- [`ext_trie_blake2_256_root_version_2`](https://spec.polkadot.network/#_version_2_prototype_9)
- [`ext_trie_blake2_256_ordered_root_version_2`](https://spec.polkadot.network/#_version_2_prototype_10)
- [`ext_trie_keccak_256_root_version_2`](https://spec.polkadot.network/#_version_2_prototype_11)
- [`ext_trie_keccak_256_ordered_root_version_2`](https://spec.polkadot.network/#_version_2_prototype_12)
[...]

PR #2530, opened 2022-05-11
chore(trie): refactor existing header encoding

qdm12:
## Changes

TLDR: Match more closely the [**polkadot spec documentation**](https://spec.polkadot.network/#defn-node-header) in the code, to allow further changes for new node variants. This doesn't add new node variants yet, this is coming in the next PR.

- Declare node variants as bits+mask instead of just bits (or the more obscure enum integer we had)
- `encodeHeader` encodes the first header byte + extra partial key length bytes instead of the convoluted/not-modular `encodeHeader` + `encodeKeyLength`
- `decodeHeader` function to get variant + partial key length, instead of convoluted in-line header code in `Decode` + further header decoding in `decodeLeaf` and `decodeBranch`
- Get node variant header partial key length mask dynamically from the node variant header bit mask (not a constant anymore)
- Clarify codec documentation in a README.md document
- Remove single byte sync pool (slower than stack byte slice of length 1)
- ⚠️ Fix `decodeKey` maximum check for last byte
- ⚠️ Fix `decodeKey` maximum check to accept maximum value `65535`
- ⚠️ allow to encode partial key of size `65535`
- Panic for programming errors cases
- Revise some of the error wrapping
- Use `uint16` for `partialKeyLength`

## Tests

```sh
go test lib/trie/... internal/trie/...
```

## Issues

- #2418 

## Primary Reviewer

@EclesioMeloJunior


internal/trie/node/encode_doc.go
```diff
@@ -3,26 +3,21 @@
```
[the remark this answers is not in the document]
qdm12: I pushed a commit to:

- Remove `encode_doc.go`
- Add `README.md` with Codec section
- Update comments for `Encode` and `Decode` methods
- Add `node` package comment

Let me know if you think that's good or if anything can be improved :+1:

EclesioMeloJunior (changes requested):

internal/trie/node/header.go
```diff
@@ -4,44 +4,148 @@
...
-	keyLenOffset    = 0x3f
-	nodeHeaderShift = 6
-)
-
 // encodeHeader writes the encoded header for the node.
 func encodeHeader(node *Node, writer io.Writer) (err error) {
-	var header byte
+	partialKeyLength := len(node.Key)
+	if partialKeyLength > int(maxPartialKeyLength) {
+		panic(fmt.Sprintf("partial key length is too big: %d", partialKeyLength))
```
EclesioMeloJunior, line 16: Same as described here -> https://github.com/ChainSafe/gossamer/pull/2530#discussion_r891633172

internal/trie/node/decode.go
```diff
@@ -9,63 +9,60 @@ import (
...
+		n, err = decodeBranch(reader, variant, partialKeyLength)
 		if err != nil {
 			return nil, fmt.Errorf("cannot decode branch: %w", err)
 		}
 		return n, nil
 	default:
-		return nil, fmt.Errorf("%w: %d", ErrUnknownNodeType, nodeTypeHeaderByte)
+		// this is a programming error, an unknown node variant
+		// should be caught by decodeHeader.
+		panic(fmt.Sprintf("not implemented for node variant %08b", variant))
```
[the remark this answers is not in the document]
qdm12: No `decodeHeader` would return an error if it's unknown variant, and this function would return before the switch statement.

The panic is just for future development (what I'm currently doing) when adding variants to avoid going crazy. I think it's appropriate to use a panic when it can only be a programming error and not something coming from IO (files, OS, network, etc.)

internal/trie/node/decode.go
```diff
@@ -101,37 +94,37 @@ func decodeBranch(reader io.Reader, header byte) (node *Node, err error) {
...
-				return nil, fmt.Errorf("%w: at index %d: %s",
-					ErrDecodeValue, i, err)
+		childNode := &Node{
+			HashDigest: hash,
+		}
+		if len(hash) < hashLength {
+			// Handle inlined nodes
+			reader = bytes.NewReader(hash)
+			variant, partialKeyLength, err := decodeHeader(reader)
+			if err == nil && variant == leafVariant.bits {
```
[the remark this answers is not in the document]
qdm12: Yes totally, thank god you have good 👀 !!

Code changed in https://github.com/ChainSafe/gossamer/pull/2530/commits/2fd8ea96c60d9e74a4803e95fa3eae9f450bca06

internal/trie/node/decode.go
```diff
@@ -101,37 +94,37 @@ func decodeBranch(reader io.Reader, header byte) (node *Node, err error) {
...
-		// Handle inlined leaf nodes.
 		const hashLength = 32
-		nodeTypeHeaderByte := hash[0] >> 6
-		if nodeTypeHeaderByte == leafHeader && len(hash) < hashLength {
-			leaf, err := decodeLeaf(bytes.NewReader(hash[1:]), hash[0])
-			if err != nil {
-				return nil, fmt.Errorf("%w: at index %d: %s",
-					ErrDecodeValue, i, err)
+		childNode := &Node{
+			HashDigest: hash,
```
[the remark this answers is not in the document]
qdm12: Indeed, added!

internal/trie/node/decode.go
```diff
@@ -78,15 +75,11 @@ func decodeBranch(reader io.Reader, header byte) (node *Node, err error) {
...
-		err := sd.Decode(&value)
+	if variant == branchWithValueVariant.bits {
+		err := sd.Decode(&node.Value)
 		if err != nil {
 			return nil, fmt.Errorf("%w: %s", ErrDecodeValue, err)
 		}
-		node.Value = value
 	}
 
 	for i := 0; i < 16; i++ {
```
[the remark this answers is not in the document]
qdm12: Out of scope, but I like it!

EclesioMeloJunior (approved):

internal/trie/node/decode.go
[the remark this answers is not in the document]
qdm12: This is done in https://github.com/ChainSafe/gossamer/pull/2611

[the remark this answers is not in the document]
qdm12: This is done in https://github.com/ChainSafe/gossamer/pull/2611

internal/trie/node/header.go
```diff
@@ -4,44 +4,148 @@
...
+	partialKeyLengthHeader, partialKeyLengthHeaderMask byte, err error) {
+	// variants is a slice of all variants sorted in ascending
+	// order by the number of bits each variant mask occupy
+	// in the header byte.
+	// See https://spec.polkadot.network/#defn-node-header
+	variants := []variant{
+		leafVariant,            // mask 1100_0000
+		branchVariant,          // mask 1100_0000
+		branchWithValueVariant, // mask 1100_0000
+	}
```
timwu20, line 136: could this be moved to a package level `var`?

internal/trie/node/variants.go
```diff
@@ -0,0 +1,26 @@
...
+	leafVariant = variant{ // leaf 01
+		bits: 0b0100_0000,
+		mask: 0b1100_0000,
+	}
+	branchVariant = variant{ // branch 10
+		bits: 0b1000_0000,
+		mask: 0b1100_0000,
+	}
+	branchWithValueVariant = variant{ // branch 11
+		bits: 0b1100_0000,
```
timwu20, line 23: is this correct? Aren't the bits supposed to match this?
Screenshot 2022-06-29 at 3 12 44 PM

internal/trie/node/decode.go
```diff
@@ -9,63 +9,60 @@ import (
...
+	switch variant {
+	case leafVariant.bits:
+		n, err = decodeLeaf(reader, partialKeyLength)
 		if err != nil {
 			return nil, fmt.Errorf("cannot decode leaf: %w", err)
 		}
 		return n, nil
-	case branchHeader, branchWithValueHeader:
-		n, err = decodeBranch(reader, header)
+	case branchVariant.bits, branchWithValueVariant.bits:
```
timwu20, line 40: is `branchWithValueVariant` supposed to be a branch containing a hashed subvalue?

internal/trie/node/decode.go
```diff
@@ -9,63 +9,60 @@ import (
...
+		return nil, fmt.Errorf("cannot decode header: %w", err)
 	}
-	header := oneByteBuf[0]
 
-	nodeTypeHeaderByte := header >> 6
-	switch nodeTypeHeaderByte {
-	case leafHeader:
-		n, err = decodeLeaf(reader, header)
+	switch variant {
+	case leafVariant.bits:
```
timwu20, line 34: do we support the "Leaf containing a hashed subvalue" variant?

internal/trie/node/variants.go
qdm12: Yes it's coming in the next PR

internal/trie/node/decode.go
qdm12: No it's just a branch with its value (aka subvalue) set.
`001` (not defined in this PR) is the variant for a hashed subvalue.

internal/trie/node/decode.go
qdm12: No this in the next PR I'll open once this merges, adding the 4 new node variants (although it's based on that branch).

internal/trie/node/header.go
timwu20: I don't mutation should be an issue, it should be pretty well understood we shouldn't be mutating `variants`.  I would assume `decodeHeader` is called a lot? We should be only allocating once if possible.

internal/trie/node/decode.go
```diff
@@ -78,18 +83,14 @@ func decodeBranch(reader io.Reader, header byte) (node *Node, err error) {
...
+	if variant == branchWithValueVariant.bits {
+		err := sd.Decode(&node.Value)
 		if err != nil {
 			return nil, fmt.Errorf("%w: %s", ErrDecodeValue, err)
 		}
-		node.Value = value
 	}
 
-	for i := 0; i < 16; i++ {
+	for i := 0; i < ChildrenCapacity; i++ {
```
timwu20, line 93: OOS: seems odd that `ChildrenCapacity` is exported

timwu20 (approved):

internal/trie/node/header.go
qdm12: > I would assume decodeHeader is called a lot? We should be only allocating once if possible.

Counter-intuitively, it really does not matter performance wise, it's the exact same speed/allocation. I pushed https://github.com/ChainSafe/gossamer/pull/2530/commits/374a29b6759981bcf91d56ae84bfda12cf158a65 with a comment + benchmark to clarify for future readers.

As a consequence, there is no need to make it global which would make it harder to read and pollute the Go package API a bit more.

internal/trie/node/decode.go
qdm12: It has been exported previously (from a suggestion of Eclesio as I recall);
It's because it's used by the `lib/trie` package to iterate over children.
We could define as a local constant in both packages, but yeah not a big deal imo.

Issue #2540, opened 2022-05-12
Fix CI on v0.9.20 branch


timwu20:
CI checks both required and optional are failing with latest runtime updates to the branch.

PR #2560, opened 2022-05-17
fix(ci): fix `cmd/gossamer` unit tests

timwu20:
- Fixes the `cmd/gossamer` unit tests for the `v0.9.20` branch.
- The build spec command tests were using the wrong file.  Should be pointed to the raw build spec file.

## Tests

```sh
go test ./cmd/gossamer
```

## Issues

contributes to #2540 

## Primary Reviewer

@jimjbrettj


chain/dev/config.toml
```diff
@@ -15,10 +15,10 @@ sync = ""
 digest = ""
 
 [init]
-genesis = "./chain/dev/dev-spec-raw.json"
+genesis = "./chain/dev/genesis-spec.json"
 
 [account]
-key = "alice"
+key = ""
```
qdm12, line 21: Why is the key removed?

chain/gssmr/config-bob.toml
```diff
@@ -1,49 +0,0 @@
-[global]
```
qdm12, line 1: Why is this file removed?

qdm12 (approved):

chain/gssmr/config-bob.toml
timwu20: the config is expected to be general purpose.  You can specify accounts via command line.  This is just mirroring what it's like on `development`.

chain/dev/config.toml
timwu20: see [comment](https://github.com/ChainSafe/gossamer/pull/2560#discussion_r875106720)

EclesioMeloJunior (approved):

Issue #2462, opened 2022-04-03
Problems Submitting Extrinsics


danforbes:
## Describe the bug

- It appears that there are problems submitting extrinsics through the `gsrpc` client library, which is probably due to a number of intersecting factors.

## Expected Behavior

- When an extrinsic is submitted to a Gossamer-based Polkadot Host through a supported client library like `gsrpc`, it should work like it works with the Substrate-based Polkadot Host.

## Current Behavior

- When an extrinsic is submitted to a Gossamer-based Polkadot Host, it doesn't execute properly (currently timing-out, it seems).

PR #2562, opened 2022-05-17
fix(core): handle latest extrinsic version and fix handleTxnPool logic

jimjbrettj:
## Changes

- Change how we construct extrinsics to support latest runtime version
- Add gssmrLatestTxnQueue runtime to run integration tests with the latest runtime for maintaining the transaction pool (to be removed at a later date)
- Upgrade the centrifuge library to v4
- Build decoding logic to decode and handle data returned from validate transaction runtime call

## Tests

```sh
go test ./dot/core -tags integration 
go test ./dot/core
go test ./dot/rpc/modules
```

## Issues

#2462

## Primary Reviewer

@timwu20


dot/core/service_integration_test.go
```diff
@@ -82,9 +82,10 @@ func newTestDigest(t *testing.T, slotNumber uint64) scale.VaryingDataTypeSlice {
 	return vdts
 }
 
-func generateTestValidRemarkTxns(t *testing.T, pubKey []byte, accInfo types.AccountInfo) ([]byte, runtime.Instance) {
+func generateTestValidRemarkTxns(t *testing.T, genesisFilePath string,
+	pubKey []byte, accInfo types.AccountInfo) ([]byte, runtime.Instance) {
 	t.Helper()
-	projectRootPath := filepath.Join(utils.GetProjectRootPathTest(t), "chain/gssmr/genesis.json")
+	projectRootPath := filepath.Join(utils.GetProjectRootPathTest(t), genesisFilePath)
```
qdm12, line 88: Maybe add a function in `lib/utils` with a switch statement returning one of `GetGssmrGenesisRawPathTest`, `GetGssmrGenesisPathTest`, etc. depending on a string. Also ideally define each constant string as a custom type for code completion

```go
type GenesisName string

const (
    GssmrGenesis GenesisName = "gssmr genesis"
    GssmrGenesisRaw GenesisName = "gssmr genesis raw"
)
```

so you would have something like

```go
func generateTestValidRemarkTxns(t *testing.T, genesisName utils.GenesisName,
	pubKey []byte, accInfo types.AccountInfo) ([]byte, runtime.Instance) {
	t.Helper()
	genesisPath := utils.GetGenesisPathTest(t, genesisName)
```

and use it like

```go
generateTestValidRemarkTxns(t, utils.GssmrGenesis, ...)
```

dot/core/service_integration_test.go
```diff
@@ -589,6 +590,7 @@ func TestService_HandleSubmittedExtrinsic(t *testing.T) {
 
 	extBytes := createExtrinsic(t, rt, genHeader.Hash(), 0)
 
+	// Tests use old extrinsic version currently
```
qdm12, line 593: Should we leave it as is? 🤔

dot/core/service_test.go
```diff
@@ -24,16 +24,29 @@ import (
...
+}
+var runtimeVersionData = runtime.NewVersionData(
+	[]byte("polkadot"),
+	[]byte("parity-polkadot"),
+	authoringVersion,
+	specVersion,
+	implVersion,
+	[]runtime.APIItem{testAPIItem},
+	transactionVersion,
+)
```
qdm12, line 49: having them inlined / locally defined as it was, was better imo.
Better to copy paste some code rather than having dependency and funny mutation surprises

dot/core/service_test.go
```diff
@@ -111,7 +112,9 @@ func generateExtrinsic(t *testing.T) (extrinsic, externalExtrinsic types.Extrins
 	require.NoError(t, err)
 
 	encExt := []types.Extrinsic{extEnc.Bytes()}
-	testExternalExt := types.Extrinsic(append([]byte{byte(types.TxnExternal)}, encExt[0]...))
+	testHeader := types.NewEmptyHeader()
+	testExternalExt := types.Extrinsic(append([]byte{byte(types.TxnExternal)},
+		append(encExt[0], testHeader.StateRoot.ToBytes()...)...))
```
qdm12, line 117: use a concatenateBytes helper function 😉 
Maybe you can tailor to produce a `types.Extrinsic` since that's all of its usage really.

Same for other uses of `types.Extrinsic(append(` below in this file.

dot/core/service_test.go
```diff
@@ -546,24 +559,32 @@ func Test_Service_maintainTransactionPool(t *testing.T) {
...
 		mockTxnState.EXPECT().RemoveExtrinsic(types.Extrinsic{21})
 		mockTxnState.EXPECT().PendingInPool().Return([]*transaction.ValidTransaction{vt})
 		mockTxnState.EXPECT().Push(tx).Return(common.Hash{}, nil)
 		mockTxnState.EXPECT().RemoveExtrinsicFromPool(types.Extrinsic{21})
 		mockBlockStateOk := NewMockBlockState(ctrl)
 		mockBlockStateOk.EXPECT().GetRuntime(nil).Return(runtimeMock, nil)
+		mockBlockStateOk.EXPECT().BestBlockHash().Return(common.Hash{})
+		mockStorageState := NewMockStorageState(ctrl)
+		mockStorageState.EXPECT().Lock()
+		mockStorageState.EXPECT().Unlock()
```
qdm12, line 583: nit a bit of a paranoid approach, but checking the order of calls can be nice especially for locking:

```suggestion
		lockCall := mockStorageState.EXPECT().Lock()
		mockStorageState.EXPECT().Unlock().After(lockCall)
```

Same for other mock expectations of lock/unlock in this test file

dot/core/messages.go
```diff
@@ -29,7 +29,10 @@ func (s *Service) validateTransaction(peerID peer.ID, head *types.Header, rt run
 	rt.SetContextStorage(ts)
 
 	// validate each transaction
-	externalExt := types.Extrinsic(append([]byte{byte(types.TxnExternal)}, tx...))
+	externalExt, err := s.buildTransaction(rt, tx)
+	if err != nil {
+		logger.Errorf("Unable to build transaction \n")
```
EclesioMeloJunior, line 24:
```suggestion
		logger.Errorf("unable to build transaction: %s", err)
```

dot/core/service.go
```diff
@@ -635,3 +666,20 @@ func (s *Service) GetReadProofAt(block common.Hash, keys [][]byte) (
...
+
+func (s *Service) buildTransaction(rt runtime.Instance, ext types.Extrinsic) (types.Extrinsic, error) {
+	runtimeVersion, err := rt.Version()
+	if err != nil {
+		return types.Extrinsic{}, err
+	}
+
+	txQueueVersion := runtimeVersion.TaggedTransactionQueueVersion(runtimeVersion)
+	var externalExt types.Extrinsic
+	if txQueueVersion == 3 {
```
EclesioMeloJunior, line 673: would be nice to have the `3` as a named const
```suggestion
        const supportedTxVersion = 3
	if txQueueVersion ==  supportedTxVersion {
```

lib/runtime/version.go
```diff
@@ -86,6 +87,19 @@ func (lvd *LegacyVersionData) TransactionVersion() uint32 {
 	return 0
 }
 
+// TaggedTransactionQueueVersion returns the TaggedTransactionQueueAPI version
+func (lvd *LegacyVersionData) TaggedTransactionQueueVersion(runtimeVersion Version) uint32 {
+	txQueueVersion := uint32(0)
+	for _, v := range runtimeVersion.APIItems() {
+		// Blake2-8("TaggedTransactionQueue")
+		if v.Name == [8]byte{0xd2, 0xbc, 0x98, 0x97, 0xee, 0xd0, 0x8f, 0x15} {
```
EclesioMeloJunior, line 95:
```suggestion
var encodedTaggedTransactionQueue = [8]byte{0xd2, 0xbc, 0x98, 0x97, 0xee, 0xd0, 0x8f, 0x15}
func (lvd *LegacyVersionData) TaggedTransactionQueueVersion(runtimeVersion Version) uint32 {
	txQueueVersion := uint32(0)
	for _, v := range runtimeVersion.APIItems() {
		// Blake2-8("TaggedTransactionQueue")
		if v.Name ==  encodedTaggedTransactionQueue {
```

lib/runtime/version.go
```diff
@@ -192,6 +206,19 @@ func (vd *VersionData) TransactionVersion() uint32 {
 	return vd.transactionVersion
 }
 
+// TaggedTransactionQueueVersion returns the TaggedTransactionQueueAPI version
+func (vd *VersionData) TaggedTransactionQueueVersion(runtimeVersion Version) uint32 {
+	txQueueVersion := uint32(0)
+	for _, v := range runtimeVersion.APIItems() {
+		// Blake2-8("TaggedTransactionQueue")
+		if v.Name == [8]byte{0xd2, 0xbc, 0x98, 0x97, 0xee, 0xd0, 0x8f, 0x15} {
```
EclesioMeloJunior, line 214:
```suggestion
    if v.Name ==  encodedTaggedTransactionQueue {
```

EclesioMeloJunior (changes requested):

chain/gssmrLatestTxnQueue/defaults.go
```diff
@@ -0,0 +1,119 @@
...
+	"github.com/ChainSafe/gossamer/lib/genesis"
+	"github.com/ChainSafe/gossamer/lib/runtime/wasmer"
+)
+
+var (
+	// GlobalConfig
+
+	// DefaultName Default node name
+	DefaultName = string("Gossamer")
+	// DefaultID Default chain ID
```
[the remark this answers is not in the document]
jimjbrettj: Since this is just a temporary runtime to be deleted, and was just copied from the 9.20 branch, I think I will leave it here and then open an issue to update on other branches

dot/core/messages.go
```diff
@@ -16,20 +16,13 @@ import (
...
+func (s *Service) validateTransaction(peerID peer.ID, rt runtime.Instance,
 	tx types.Extrinsic) (validity *transaction.Validity, valid bool, err error) {
-	s.storageState.Lock()
 
-	ts, err := s.storageState.TrieState(&head.StateRoot)
-	s.storageState.Unlock()
+	externalExt, err := s.BuildExternalTransaction(rt, tx)
 	if err != nil {
-		return nil, false, fmt.Errorf("cannot get trie state from storage for root %s: %w", head.StateRoot, err)
+		logger.Errorf("Unable to build transaction \n")
```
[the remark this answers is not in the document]
jimjbrettj: good catch!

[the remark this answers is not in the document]
jimjbrettj: Actually, im seeing that in this function other errors are just logged and not returned. Interesting, @timwu20 would you maybe be able to provide some context here?

dot/core/service.go
```diff
@@ -635,3 +659,25 @@ func (s *Service) GetReadProofAt(block common.Hash, keys [][]byte) (
 
 	return block, proofForKeys, nil
 }
+
+// BuildExternalTransaction builds an external transaction based on the current TransactionQueueAPIVersion
+// See https://paritytech.github.io/polkadot/src/sp_transaction_pool/runtime_api.rs.html#25-55
```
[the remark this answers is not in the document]
jimjbrettj: hmmm this is a link to the rust docs so I'm not sure how I would do that, do you know?

lib/runtime/wasmer/errors.go
```diff
@@ -0,0 +1,301 @@
...
+	invalidCustom InvalidCustom
+	unknownCustom UnknownCustom
+)
+
+func newUnknownError(data scale.VaryingDataTypeValue) error {
+	return fmt.Errorf("unknown error: %d", data)
+}
+
+// UnmarshalError occurs when unmarshalling fails
+type UnmarshalError struct {
```
timwu20, line 44: why do we need this?

devnet/docker-compose-substrate.yml
```diff
@@ -8,8 +8,8 @@ services:
         DD_API_KEY: ${DD_API_KEY}
     ports:
       - 7001
-      - 8545
-      - 8546
+      - 8545:8545
+      - 8546:8546
```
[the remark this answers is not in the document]
qdm12: Containers can reach other without port publishing if they're in the same Docker network, publishing ports is only useful if you want to access it i.e. from your host / terminal.

If these ports are there only for container comm, I'd suggest to change that `ports:` to `expose:` so it's just for documentation and doesn't publish ports to the host.

dot/core/service.go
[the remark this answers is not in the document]
qdm12: @danforbes do you know if you can pin the rust doc to a certain commit? 🤔

dot/core/service.go
qdm12: See my previous comment https://github.com/ChainSafe/gossamer/pull/2562#discussion_r919253153

We should define the constants at global scope so someone doesn't re-inline them in another function in the future, especially since these could be used in different places.

> I wonder if there is a difference here when the const is defined inline.

Performance wise, no difference, so it's nice for only-local-constant-not-shared-anywhere-else

lib/runtime/wasmer/errors.go
```diff
@@ -0,0 +1,301 @@
+// Copyright 2021 ChainSafe Systems (ON)
+// SPDX-License-Identifier: LGPL-3.0-only
+
+package wasmer
```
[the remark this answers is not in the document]
qdm12: nit or another subpackage in runtime.
Just my brain is crying when I see subpackages (wasmer) importing parent package (lib/runtime).

dot/core/service.go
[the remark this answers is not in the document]
danforbes: I do not believe that anyone is hosting that detail of historical information at this time. The team that I used to work on Parity would maintain the previous three versions of documentation, but I believe that process has been discontinued. There are docs at crates.parity.io and here https://paritytech.github.io/substrate/ The latter contains multiple versions, but not necessarily ones that are up-to-date.

devnet/docker-compose-substrate.yml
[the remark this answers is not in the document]
jimjbrettj: oh actually that is totally what i did it for. I was running a local script against the container. 

My thoughts: I think I leave it since I think for my txn testing it will be largely local, then can unexposed when my tests are migrated to a container.

Open to thoughts if anyone thinks otherwise

dot/core/service_integration_test.go
jimjbrettj: ya I think so. My current understanding is not that this PR is to migrate to newest runtime, but more to add logic to support without breaking development to then migrate further based on the 9.20 branch.

That being said, I do have a newer runtime that I am using to test maintainTxnPool, so if you think its appropriate i can create another test function that uses the latest runtime here too. This could be a good call

lib/runtime/version.go
jimjbrettj: good call

devnet/alice.Dockerfile
```diff
@@ -38,6 +38,6 @@ RUN go run cmd/update-dd-agent-confd/main.go -n=${METRICS_NAMESPACE} -t=key:alic
 
 WORKDIR /gossamer
 
-ENTRYPOINT service datadog-agent start && gossamer --key=alice --babe-lead --publish-metrics  --metrics-address=":9876" --rpc --rpc-external=true --pubdns=alice --port 7001
+ENTRYPOINT service datadog-agent start && gossamer --key=alice --babe-lead --publish-metrics --metrics-address=":9876" --rpc --rpc-external=true --pubdns=alice --port 7001 --ws --ws-external --rpc-unsafe --rpc-unsafe-external
```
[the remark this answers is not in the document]
jimjbrettj: good question, are all of these flags still used? @EclesioMeloJunior @danforbes

dot/core/service.go
```diff
@@ -346,6 +348,9 @@ func (s *Service) handleChainReorg(prev, curr common.Hash) error {
 		return nil
 	}
 
+	s.storageState.Lock()
```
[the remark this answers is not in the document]
jimjbrettj: For the storageState locks in this file, I had observed that with the locks the errors I was getting around the timestamp issue would go away. However, it doesn't permanently fix it and since @kishansagathiya has been investigating this issue maybe i remove these locks (as long as it still works ofc)?

dot/core/messages.go
```diff
@@ -72,15 +65,25 @@ func (s *Service) HandleTransactionMessage(peerID peer.ID, msg *network.Transact
...
+		return false, fmt.Errorf("cannot get trie state from storage for root %s: %w", head.StateRoot, err)
+	}
+
 	hash := head.Hash()
 	rt, err := s.blockState.GetRuntime(&hash)
 	if err != nil {
 		return false, err
 	}
 
+	rt.SetContextStorage(ts)
```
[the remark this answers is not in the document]
jimjbrettj: Unless I am mistaken this function had this logic before

dot/core/messages_test.go
```diff
@@ -63,6 +64,11 @@ type mockSetContextStorage struct {
 	trieState *storage.TrieState
 }
 
+type mockVersion struct {
+	version *mocksruntime.Version
+	err     error
+}
```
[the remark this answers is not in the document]
jimjbrettj: This is referenced by MockRuntime which is used a lot of places

dot/core/service_integration_test.go
jimjbrettj: Update: Talked with Tim about this and we are going to leave as is and test in latest Runtime branch

jimjbrettj:
Closed and replaced with

PR #2566, opened 2022-05-23
Implement parachain inherents

### LINKED ISSUE #2426: Implement Parachain Inherents
## Task summary

- implement stubbed/empty `newheads` and `parachn0` inherents to allow block building w/ polkadot runtimes >= v0.9.17
- for now we can use "empty" values, as we require parachain-specific logic to actually provide the data.

## Specification reference

`newheads` inherent type: I don't think this is actually set in polkadot, as searching the codebase for `NEW_HEADS_IDENTIFIER` doesn't show it being set.

`parachn0` inherent type: https://github.com/paritytech/polkadot/blob/v0.9.17/primitives/src/v1/mod.rs#L1345
```rust
/// Parachains inherent-data passed into the runtime by a block author
#[derive(Encode, Decode, Clone, PartialEq, RuntimeDebug, TypeInfo)]
pub struct InherentData<HDR: HeaderT = Header> {
	/// Signed bitfields by validators about availability.
	pub bitfields: UncheckedSignedAvailabilityBitfields,
	/// Backed candidates for inclusion in the block.
	pub backed_candidates: Vec<BackedCandidate<HDR::Hash>>,
	/// Sets of dispute votes for inclusion,
	pub disputes: MultiDisputeStatementSet,
	/// The parent block header. Used for checking state proofs.
	pub parent_header: HDR,
}
```

kishansagathiya:
## Changes

- Implements `parachn0` inherent

## Tests

```sh
```

## Issues

Fixes #2426
## Primary Reviewer

@timwu20


lib/babe/build.go
```diff
@@ -237,6 +237,147 @@ func (b *BlockBuilder) buildBlockExtrinsics(slot Slot, rt runtime.Instance) []*t
...
+type UpwardMessage []byte
+
+// OutboundHrmpMessage is an HRMP message seen from the perspective of a sender.
+type OutboundHrmpMessage struct {
+	Recipient uint32
+	Data      []byte
+}
+
+// All Vec<u8> in rust have become []byte here
+
```
qdm12, line 269: nit I don't think that's a really necessary comment 🤔 

```suggestion
```

lib/babe/build.go
```diff
@@ -237,6 +237,147 @@ func (b *BlockBuilder) buildBlockExtrinsics(slot Slot, rt runtime.Instance) []*t
...
+// ValidationCode is Parachain validation code.
+type ValidationCode []byte
+
+// HeadData is Parachain head data included in the chain.
+type HeadData []byte
```
qdm12, line 274: Do we want all these types if they don't have any methods defined on them?

Maybe just using `[]byte` directly would be clearer, with the right variable names. That would also reduce the Go API which is already too large imo.

lib/babe/build.go
```diff
@@ -237,6 +237,147 @@ func (b *BlockBuilder) buildBlockExtrinsics(slot Slot, rt runtime.Instance) []*t
...
+
+// CommittedCandidateReceipt is a candidate-receipt with commitments directly included.
+type CommittedCandidateReceipt struct {
+	Descriptor  *CandidateDescriptor
+	Commitments *CandidateCommitments
+}
+
+// ValidityAttestation is an implicit or explicit attestation to the validity of a parachain
+// candidate.
+type ValidityAttestation int
```
qdm12, line 300: nit to take less space

```suggestion
type ValidityAttestation uint8
```

lib/babe/build.go
```diff
@@ -237,6 +237,147 @@ func (b *BlockBuilder) buildBlockExtrinsics(slot Slot, rt runtime.Instance) []*t
...
+	ValidatorIndices []byte
+}
+
+// MultiDisputeStatementSet is a set of dispute statements.
+type MultiDisputeStatementSet []DisputeStatementSet
+
+// DisputeStatement is a statement about a candidate, to be used within the dispute
+// resolution process. Statements are either in favour of the candidate's validity
+// or against it.
+type DisputeStatement int
```
qdm12, line 337: nit to take less space

```suggestion
type DisputeStatement uint8
```

lib/runtime/wasmer/imports.go
```diff
@@ -875,8 +875,50 @@ func ext_trie_blake2_256_ordered_root_version_1(context unsafe.Pointer, dataSpan
...
+	instanceContext := wasm.IntoInstanceContext(context)
+	memory := instanceContext.Memory().Data()
+	runtimeCtx := instanceContext.Data().(*runtime.Context)
+	data := asMemorySlice(instanceContext, dataSpan)
+
+	t := trie.NewEmptyTrie()
+	var values [][]byte
+	err := scale.Unmarshal(data, &values)
+	if err != nil {
+		logger.Errorf("[ext_trie_blake2_256_ordered_root_version_2]: %s", err)
```
qdm12, line 889: The function name is already logged with the wasmer logger automagically, see line

```go
logger.Patch(log.SetLevel(cfg.LogLvl), log.SetCallerFunc(true))
```

```suggestion
		logger.Errorf("%s", err)
```

lib/runtime/wasmer/imports.go
```diff
@@ -875,8 +875,50 @@ func ext_trie_blake2_256_ordered_root_version_1(context unsafe.Pointer, dataSpan
...
+		return 0
+	}
+
+	for i, val := range values {
+		key, err := scale.Marshal(big.NewInt(int64(i)))
+		if err != nil {
+			logger.Errorf("[ext_trie_blake2_256_ordered_root_version_2]: %s", err)
```
qdm12, line 896:
```suggestion
			logger.Errorf("%s", err)
```

lib/runtime/wasmer/imports.go
```diff
@@ -875,8 +875,50 @@ func ext_trie_blake2_256_ordered_root_version_1(context unsafe.Pointer, dataSpan
...
+			"put key=0x%x and value=0x%x",
+			key, val)
+
+		t.Put(key, val)
+	}
+
+	// allocate memory for value and copy value to memory
+	ptr, err := runtimeCtx.Allocator.Allocate(32)
+	if err != nil {
+		logger.Errorf("[ext_trie_blake2_256_ordered_root_version_2]: %s", err)
```
qdm12, line 909:
```suggestion
		logger.Errorf("%s", err)
```

lib/runtime/wasmer/imports.go
```diff
@@ -875,8 +875,50 @@ func ext_trie_blake2_256_ordered_root_version_1(context unsafe.Pointer, dataSpan
...
+		return 0
+	}
+
+	hash, err := t.Hash()
+	if err != nil {
+		logger.Errorf("[ext_trie_blake2_256_ordered_root_version_2]: %s", err)
```
qdm12, line 915:
```suggestion
		logger.Errorf("%s", err)
```

lib/runtime/wasmer/imports.go
```diff
@@ -875,8 +875,50 @@ func ext_trie_blake2_256_ordered_root_version_1(context unsafe.Pointer, dataSpan
...
+
+	t := trie.NewEmptyTrie()
+	var values [][]byte
+	err := scale.Unmarshal(data, &values)
+	if err != nil {
+		logger.Errorf("[ext_trie_blake2_256_ordered_root_version_2]: %s", err)
+		return 0
+	}
+
+	for i, val := range values {
```
qdm12, line 893: nit `value` would be better than `val`

lib/babe/build.go
```diff
@@ -237,6 +237,147 @@ func (b *BlockBuilder) buildBlockExtrinsics(slot Slot, rt runtime.Instance) []*t
...
+	Valid DisputeStatement = iota
+	// Invalid is for an invalid statement
+	Invalid
+)
+
+// ValidatorIndex is the index of the validator.
+type ValidatorIndex uint32
+
+// ValidatorSignature is the signature with which parachain validators sign blocks.
+type ValidatorSignature []byte
```
qdm12, line 350: Do we really need those types?

dot/types/inherents.go
```diff
@@ -64,6 +68,14 @@ func (d *InherentsData) SetInt64Inherent(key []byte, data uint64) error {
 	return nil
 }
 
+// SetBytesInherent sets an inherent of type bytes.
```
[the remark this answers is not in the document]
kishansagathiya: An inherent could be a struct, int, etc. That's what i mean when i write type bytes.

lib/babe/build.go
```diff
@@ -237,6 +237,147 @@ func (b *BlockBuilder) buildBlockExtrinsics(slot Slot, rt runtime.Instance) []*t
 	return included
 }
 
+// CandidateDescriptor is a unique descriptor of the candidate receipt.
+type CandidateDescriptor struct {
+	// The ID of the para this is a candidate for.
```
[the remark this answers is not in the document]
kishansagathiya: linter is already green

lib/babe/build.go
```diff
@@ -237,6 +237,147 @@ func (b *BlockBuilder) buildBlockExtrinsics(slot Slot, rt runtime.Instance) []*t
...
+	// The hash of the relay-chain block this should be executed in
+	// the context of.
+	// NOTE: the fact that the hash includes this value means that code depends
+	// on this for deduplication. Removing this field is likely to break things.
+	RelayParent common.Hash
+	// The collator's relay-chain account ID
+	Collator []byte // CollatorId
+	// Signature on blake2-256 of components of this receipt:
+	// The para ID, the relay parent, and the `pov_hash`.
+	// TODO: I have made a guess that all signatures are []byte
```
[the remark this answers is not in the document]
kishansagathiya: In this case, this is more for the reviewer or for me, when i am uncertain about something. But yes, it is meant to be removed.

lib/babe/build.go
qdm12: in substrate you mean right?

kishansagathiya: Nope, polkadot. That's where all these structs are. https://github.com/paritytech/polkadot/blob/v0.9.17/primitives/src/v1/mod.rs#L1345

dot/types/inherents.go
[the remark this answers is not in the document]
qdm12: Would this this fit then?

```suggestion
// SetBytesInherent sets an inherent from its key and value bytes encoding.
```

lib/babe/build.go
qdm12: Interesting!

@timwu20 what do you think we should do in cases like such (I'd like to add it to the code style document of the repo): should we mirror what the Polkadot Rust code has or not care about it?

timwu20: Yes I think we should.  Even though they are just byte arrays right now, I'd like to know what they are.  We can add methods down the road if we need to.  It would be nice to be able to search through both codebases and see something similar.  Ideally I'd like to break this out into a separate repo the way parity does it.

lib/babe/parachain_inherents.go
```diff
@@ -0,0 +1,240 @@
...
+// Explicit is for explicit attestation.
+type Explicit ValidatorSignature
+
+// Index Returns VDT index
+func (ex Explicit) Index() uint {
+	return 2
+}
+
+// GetValidityAttestation returns a ValidityAttestation varying data type.
+func GetValidityAttestation() ValidityAttestation {
```
timwu20, line 30:
```suggestion
func NewValidityAttestation() ValidityAttestation {
```

lib/babe/parachain_inherents.go
```diff
@@ -0,0 +1,240 @@
...
+// resolution process. Statements are either in favour of the candidate's validity
+// or against it.
+type DisputeStatement scale.VaryingDataType
+
+// Valid is for a valid statement
+type Valid ValidDisputeStatementKind
+
+// Index Returns VDT index
+func (v Valid) Index() uint {
+	return 0
```
timwu20, line 49: Usually this should start at `1` I believe.  But I did notice that in the polkadot code it had this macro above the definition:
```
#[codec(index = 0)]
Valid(ValidDisputeStatementKind),
```
Can you verify that that it does use the index when serialising on the rust side? We should probably have a test for the whole nested `VaryingDataType` just to ensure this is working as expected.

lib/babe/parachain_inherents.go
```diff
@@ -0,0 +1,240 @@
...
+type ExplicitInvalidDisputeStatementKind struct{}
+
+// Index Returns VDT index
+func (ex ExplicitInvalidDisputeStatementKind) Index() uint {
+	return 0
+}
+
+// GetDisputeStatement returns a GetDisputeStatement varying data type.
+func GetDisputeStatement() DisputeStatement {
+	invalidDisputeStatementKind, err := scale.NewVaryingDataType(ExplicitInvalidDisputeStatementKind{})
```
timwu20, line 108: so odd that they were use an enum with only one option.

lib/babe/parachain_inherents.go
```diff
@@ -0,0 +1,240 @@
...
+		panic(err)
+	}
+
+	validDisputeStatementKind, err := scale.NewVaryingDataType(
+		ExplicitValidDisputeStatementKind{}, BackingSeconded{}, BackingValid{}, ApprovalChecking{})
+	if err != nil {
+		panic(err)
+	}
+
+	vdt, err := scale.NewVaryingDataType(validDisputeStatementKind.Value(), invalidDisputeStatementKind.Value())
```
timwu20, line 119: I don't think this is right.  I think you need to implement the `scale.VaryingDataTypeValue` interface  on the `ValidDisputeStatementKind` and `InvalidDisputeStatementKind` since this is a nested enum.

lib/babe/parachain_inherents.go
```diff
@@ -0,0 +1,240 @@
...
+	// NOTE: the fact that the hash includes this value means that code depends
+	// on this for deduplication. Removing this field is likely to break things.
+	RelayParent common.Hash
+	// The collator's relay-chain account ID
+	Collator []byte // CollatorId
+	// Signature on blake2-256 of components of this receipt:
+	// The para ID, the relay parent, and the `pov_hash`.
+	Signature []byte // CollatorSignature
+	// The hash of the `pov-block`.
+	PovHash common.Hash
```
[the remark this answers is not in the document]
kishansagathiya: addressed! Added scale tag to all the structs

lib/babe/parachain_inherents.go
kishansagathiya: `DisputeStatement` has two values `Valid` (0) and `Invalid` (1). So, I have structs Valid and Invalid which implement Index() method (thus scale.VaryingDataTypeValue interface). ValidDisputeStatementKind is type of Valid and InvalidDisputeStatementKind is type of Invalid. Isn't that appropriate?

Checkout 
https://github.com/ChainSafe/gossamer/pull/2566/files#diff-3abb401d0f0c9e3a6c9f813400f674af4b2a53d9f3af0f627213df77b053c06cR45 and
https://github.com/ChainSafe/gossamer/pull/2566/files#diff-3abb401d0f0c9e3a6c9f813400f674af4b2a53d9f3af0f627213df77b053c06cR53

lib/babe/build.go
```diff
@@ -252,6 +252,18 @@ func buildBlockInherents(slot Slot, rt runtime.Instance) ([][]byte, error) {
...
+	// for now we can use "empty" values, as we require parachain-specific
+	// logic to actually provide the data.
+
+	if err = idata.SetStructInherent(types.Parachn0, parachainInherent); err != nil {
+		return nil, err
+	}
+	if err = idata.SetStructInherent(types.Newheads, []byte{0}); err != nil {
+		return nil, err
+	}
+
```
kishansagathiya, line 266: @timwu20 does this scale encoding look good to you? The error that i am getting tells that the encoding is not right!
```
2022-07-04T10:14:05Z CRIT target=runtime message=panicked at 'Bad input data provided to inherent_extrinsics: Codec error', /home/elizabeth/substrate/bin/node/runtime/src/lib.rs:1099:1	ext_logging_log_version_1	pkg=runtime module=go-wasmer
badger 2022/07/04 10:14:05 INFO: [Compactor: 173] Running compaction: {level:0 score:1.73 dropPrefixes:[]} for level: 0
badger 2022/07/04 10:14:05 INFO: LOG Compact 0->1, del 1 tables, add 1 tables, took 1.592695ms
badger 2022/07/04 10:14:05 INFO: [Compactor: 173] Compaction for level: 0 DONE
badger 2022/07/04 10:14:05 INFO: Force compaction on level 0 done
--- FAIL: TestVerifyAuthorshipRight (8.20s)
    build_integration_test.go:97: 
        	Error Trace:	build_integration_test.go:97
        	            				verify_integration_test.go:376
        	Error:      	Received unexpected error:
        	            	cannot build inherents: Failed to call the `BlockBuilder_inherent_extrinsics` exported function.
        	Test:       	TestVerifyAuthorshipRight

```

lib/runtime/wasmer/imports.go
kishansagathiya: think i have changed this

kishansagathiya:
blocked on https://github.com/ChainSafe/gossamer/issues/2634

dot/types/inherents.go
```diff
@@ -64,6 +68,29 @@ func (d *InherentsData) SetInt64Inherent(key []byte, data uint64) error {
...
+
+	data, err := scale.Marshal(value)
+	if err != nil {
+		return err
+	}
+
+	venc, err := scale.Marshal(data)
+	if err != nil {
+		return err
+	}
```
noot, line 85: wait why is this marshaling twice?

noot: since @kishansagathiya mentioned this method works, i think it's actually correct: https://spec.polkadot.network/#defn-scale-dictionary

maybe double check with the spec team, but it says here the encoding of each (k,v) in the entire map is firstly encoded before concatentation, so if polkadot is interpreting the value as a byte slice instead of a struct this would be correct

lib/babe/build.go
```diff
@@ -252,6 +393,22 @@ func buildBlockInherents(slot Slot, rt runtime.Instance) ([][]byte, error) {
 		return nil, err
 	}
 
+	// add parachn0 and newheads
+	bz, err := scale.Marshal(&ParachainInherentData{})
+	if err != nil {
+		return nil, err
+	}
```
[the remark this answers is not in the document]
qdm12: > Also is the the computational efficiency gained by making it a global variable worth the confusion or worth the effort to discuss it? I don't think so.

Fair point, I agree as well. And yes a comment about changing the empty struct would be nice 👍

dot/types/inherents.go
qdm12: Ahh interesting, thanks for clarifying 💯  Kishan and I went a bit crazy figuring this out last week, without really knowing why it was suddenly working by encoding it twice. We should probably add a comment to why it's like that.

qdm12: Actually, re-reading the spec (and also PR-ing the slightly odd grammar):

1. We need to encode the tuple (`key`, `value`) as `encodedTuple` := Enc(`key`) || Enc(`value`)
2. Then we encode the map ((key: value)) as: Enc(size) || `encodedTuple` = Enc(1) || `encodedTuple`

Now our code is quite wrong I think since:

- the key gets lost (we only have `encodedTuple := Enc(value)`)
- the size is wrong, since we set the length of the encoded value instead of the map number of elements (`1` here)

IMO we should add support for maps in our scale package before fixing this.

qdm12 (changes requested):

dot/types/inherents.go
qdm12: It's still quite strange they want it as an encoded byte slice though, why not have it as a struct? 🤔

dot/types/inherents.go
```diff
@@ -64,6 +68,29 @@ func (d *InherentsData) SetInt64Inherent(key []byte, data uint64) error {
 	return nil
 }
 
+// SetStructInherent sets a struct inherent.
+func (d *InherentsData) SetStructInherent(key []byte, value interface{}) error {
+	if len(key) != 8 {
+		return errors.New("inherent key must be 8 bytes")
```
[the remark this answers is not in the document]
qdm12: Sure for out of scope elements, totally makes sense. But this is not out of scope, let's address it here.

lib/babe/parachain_inherents_test.go
```diff
@@ -0,0 +1,295 @@
...
+			encodingValue: []byte{0x1, 0x0},
+		},
+	}
+
+	for _, c := range testCases {
+		t.Run(c.name, func(t *testing.T) {
+			bytes, err := scale.Marshal(c.vdt)
+			require.NoError(t, err)
+
+			fmt.Printf("0x%x\n", c.encodingValue)
```
qdm12, line 187:
```suggestion
```

lib/babe/parachain_inherents_test.go
```diff
@@ -0,0 +1,295 @@
...
+			require.Equal(t, c.encodingValue, bytes)
+
+			newDst := NewDisputeStatement()
+			err = scale.Unmarshal(bytes, &newDst)
+			require.NoError(t, err)
+
+			if !reflect.DeepEqual(c.vdt, newDst) {
+				panic(fmt.Errorf("uh oh: \n%+v \n\n%+v", c.vdt, newDst))
+			}
```
qdm12, line 196:
```suggestion
			assert.Equal(c.vdt, newDst)
```

lib/babe/parachain_inherents_test.go
```diff
@@ -0,0 +1,295 @@
+// Copyright 2022 ChainSafe Systems (ON)
+// SPDX-License-Identifier: LGPL-3.0-only
+
+package babe
+
+import (
+	"bytes"
+	"fmt"
+	reflect "reflect"
```
qdm12, line 9:
```suggestion
	"reflect"
```

lib/babe/parachain_inherents_test.go
```diff
@@ -0,0 +1,295 @@
...
+			fmt.Printf("bytes 0x%x\n", bytes)
+			require.Equal(t, c.encodingValue, bytes)
+
+			newDst := NewValidityAttestation()
+			err = scale.Unmarshal(bytes, &newDst)
+			require.NoError(t, err)
+
+			if !reflect.DeepEqual(validityAttestation, newDst) {
+				panic(fmt.Errorf("uh oh: \n%+v \n\n%+v", validityAttestation, newDst))
+			}
```
qdm12, line 237:
```suggestion
			assert.Equal(validityAttestation, newDst)
```

lib/babe/parachain_inherents_test.go
```diff
@@ -0,0 +1,295 @@
...
+	for _, c := range testCases {
+		t.Run(c.name, func(t *testing.T) {
+			validityAttestation := NewValidityAttestation()
+			err := validityAttestation.Set(c.enumValue)
+			require.NoError(t, err)
+
+			bytes, err := scale.Marshal(validityAttestation)
+			require.NoError(t, err)
+
+			fmt.Printf("bytes 0x%x\n", bytes)
```
qdm12, line 228:
```suggestion
```

lib/babe/parachain_inherents_test.go
```diff
@@ -0,0 +1,295 @@
...
+		BackedCandidates: []BackedCandidate{},
+		Disputes:         MultiDisputeStatementSet{},
+		ParentHeader: types.Header{
+			ParentHash:     common.Hash{},
+			Number:         0,
+			StateRoot:      common.Hash{},
+			ExtrinsicsRoot: common.Hash{},
+			Digest:         scale.VaryingDataTypeSlice{},
+		},
+	}
```
qdm12, line 278: nit why not just (to avoid the cognitive load of checking every single field for a non-zero value):

```suggestion
	var parachainInherent ParachainInherentData
```

As a side note, shouldn't you set some fields to their non-zero value to test the encoding, just in case ™️

lib/babe/parachain_inherents_test.go
```diff
@@ -0,0 +1,295 @@
...
+
+	actualParaInherentBytes, err := scale.Marshal(parachainInherent)
+	require.NoError(t, err)
+
+	require.Equal(t, len(expectedParaInherentsbytes), len(actualParaInherentBytes))
+	require.True(t, bytes.Equal(actualParaInherentBytes, expectedParaInherentsbytes))
```
qdm12, line 284: nit same thing simpler

```suggestion
	assert.Equal(t, expectedParaInherentsbytes, actualParaInherentBytes)
```

lib/babe/parachain_inherents_test.go
```diff
@@ -0,0 +1,295 @@
...
+
+	idata := types.NewInherentsData()
+	err = idata.SetStructInherent(types.Parachn0, parachainInherent)
+	require.NoError(t, err)
+
+	actualInherentsBytes, err := idata.Encode()
+	require.NoError(t, err)
+	require.Equal(t, len(expectedInherentsBytes), len(actualInherentsBytes))
+	require.True(t, bytes.Equal(expectedInherentsBytes, actualInherentsBytes))
```
qdm12, line 293:
```suggestion
	require.Equal(t, expectedInherentsBytes, actualInherentsBytes)
```

lib/runtime/constants.go
```diff
@@ -26,8 +26,13 @@ const (
 
 	// v0.9.17 polkadot runtime
 	POLKADOT_RUNTIME_v0917     = "polkadot_runtime-v917"
-	POLKADOT_RUNTIME_FP_v0917  = "polkadot_runtime-v917.compact.wasm"
-	POLKADOT_RUNTIME_URL_v0917 = "https://github.com/paritytech/polkadot/blob/c583355a599200d64e9202d05f47860d824b4dcf/target/release/wbuild/polkadot-runtime/polkadot_runtime.compact.wasm?raw=true" //nolint:lll
+	POLKADOT_RUNTIME_FP_v0917  = "polkadot_runtime-v9170.compact.compressed.wasm"
+	POLKADOT_RUNTIME_URL_v0917 = "https://github.com/paritytech/polkadot/releases/download/v0.9.17-rc4/polkadot_runtime-v9170.compact.compressed.wasm?raw=true" //nolint:lll
```
qdm12, line 30: this is nice, but isn't it out of scope? Would you mind reverting and add this in a dumb dumb / easy PR?

lib/runtime/wasmer/imports.go
```diff
@@ -895,10 +895,52 @@ func ext_trie_blake2_256_ordered_root_version_1(context unsafe.Pointer, dataSpan
 }
 
 //export ext_trie_blake2_256_ordered_root_version_2
-func ext_trie_blake2_256_ordered_root_version_2(context unsafe.Pointer,
-	dataSpan C.int64_t, version C.int32_t) C.int32_t {
-	// TODO: update to use state trie version 1 (#2418)
-	return ext_trie_blake2_256_ordered_root_version_1(context, dataSpan)
+func ext_trie_blake2_256_ordered_root_version_2(context unsafe.Pointer, dataSpan C.int64_t, version C.int32_t,
+) C.int32_t {
+	logger.Debug("executing...")
```
qdm12, line 900: Does this work? It shouldn't since v1 state trie is still not there. Why is it needed here as well?

Also this was implemented already in [#host functions](https://github.com/ChainSafe/gossamer/pull/2664)

lib/runtime/wasmer/imports.go
```diff
@@ -2090,8 +2132,26 @@ func ext_storage_root_version_1(context unsafe.Pointer) C.int64_t {
 
 //export ext_storage_root_version_2
 func ext_storage_root_version_2(context unsafe.Pointer, version C.int32_t) C.int64_t {
-	// TODO: update to use state trie version 1 (#2418)
-	return ext_storage_root_version_1(context)
+	logger.Trace("executing...")
+
+	instanceContext := wasm.IntoInstanceContext(context)
```
qdm12, line 2137: Why is this needed in this PR?

qdm12 (changes requested):
Pretty good, thanks also for all the comments on every field and method 👍 

Why are we exporting so many new types in `lib/babe` if they are unused outside the package. We should unexport them and export them gradually as we need them in other packages.

lib/babe/parachain_inherents.go
```diff
@@ -0,0 +1,347 @@
...
+// ParachainInherentData is parachains inherent-data passed into the runtime by a block author.
+type ParachainInherentData struct {
+	// Signed bitfields by validators about availability.
+	Bitfields []UncheckedSignedAvailabilityBitfield `scale:"1"`
+	// Backed candidates for inclusion in the block.
+	BackedCandidates []BackedCandidate `scale:"2"`
+	// Sets of dispute votes for inclusion,
+	Disputes MultiDisputeStatementSet `scale:"3"` // []DisputeStatementSet
+	// The parent block header. Used for checking state proofs.
+	ParentHeader types.Header `scale:"4"`
```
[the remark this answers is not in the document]
kishansagathiya: It cannot, it does mess up encoding.

dot/types/inherents.go
```diff
@@ -64,6 +68,29 @@ func (d *InherentsData) SetInt64Inherent(key []byte, data uint64) error {
...
+
+	venc, err := scale.Marshal(data)
+	if err != nil {
+		return err
+	}
+
+	kb := [8]byte{}
+	copy(kb[:], key)
+
+	d.data[kb] = venc
```
timwu20, line 90: I don't think we should be storing the scale encoded version of this map in `InherentsData`.  I agree we should be supporting maps in the scale package. In the meantime move this code to `InherentsData.Encode()` which already loops through the map and is already called whenever we need to scale encode `InherentsData`.

timwu20 (changes requested):

lib/babe/parachain_inherents.go
kishansagathiya: done

lib/babe/parachain_inherents.go
kishansagathiya: done

lib/babe/parachain_inherents.go
```diff
@@ -0,0 +1,347 @@
...
+}
+
+// Value will return value from underying VaryingDataType
+func (v *ValidDisputeStatementKind) Value() (val scale.VaryingDataTypeValue) {
+	vdt := scale.VaryingDataType(*v)
+	return vdt.Value()
+}
+
+// ExplicitValidDisputeStatementKind is an explicit statement issued as part of a dispute.
+type ExplicitValidDisputeStatementKind struct{}
```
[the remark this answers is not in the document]
timwu20: If this is exported in the polkadot repo, I don't mind keeping them public.  I'd like to verify that the package name is the same as the crate name in substrate.

lib/runtime/wasmer/imports.go
kishansagathiya: Looks like things work even without changing this function, removed it.

lib/runtime/wasmer/imports.go
kishansagathiya: Looks like things work even without changing this function, removed it.

lib/babe/parachain_inherents.go
```diff
@@ -0,0 +1,347 @@
...
+	// The payload is part of the signed data. The rest is the signing context,
+	// which is known both at signing and at validation.
+	Payload []byte `scale:"1"`
+	// The index of the validator signing this statement.
+	ValidatorIndex uint32 `scale:"2"`
+	/// The signature by the validator of the signed payload.
+	Signature Signature `scale:"3"`
+	// go does not have phantom data
+	// /// This ensures the real payload is tracked at the typesystem level.
+	// real_payload: sp_std::marker::PhantomData<RealPayload>,
```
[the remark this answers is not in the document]
kishansagathiya: I removed this comment. In case of an unused argument, rust compiler gives an error. PhantomData gets used to suppress this compile error.

lib/babe/parachain_inherents.go
```diff
@@ -0,0 +1,345 @@
...
+import (
+	"fmt"
+
+	"github.com/ChainSafe/gossamer/dot/types"
+	"github.com/ChainSafe/gossamer/lib/common"
+	"github.com/ChainSafe/gossamer/pkg/scale"
+)
+
+// signature could be one of Ed25519 signature, Sr25519 signature or ECDSA/SECP256k1 signature.
+type signature [64]byte
```
qdm12, line 15: nit do we really need a `signature` type? Can't we just use `[64]byte` directly? I.e. `type collatorSignature [64]byte`, `type validatorSignature [64]byte`, `Signature [64]byte`
Asking since it conflicts with signature import, and `signature` is rather generic and would be nice to use as a variable name here and there.

lib/babe/parachain_inherents.go
```diff
@@ -0,0 +1,345 @@
...
+	if err != nil {
+		return fmt.Errorf("setting value to varying data type: %w", err)
+	}
+	// store original ParentVDT with VaryingDataType that has been set
+	*distputedStatement = disputeStatement(vdt)
+	return nil
+}
+
+// Value will return value from underying VaryingDataType
+func (distputedStatement *disputeStatement) Value() (val scale.VaryingDataTypeValue) {
```
qdm12, line 85: Typo `distputedStatement`, at least rename to `disputedStatement`. Ideally rename to `d` to avoid any confusion (past dispute or present dispute?)

lib/babe/parachain_inherents.go
```diff
@@ -0,0 +1,345 @@
...
+	return validityAttestation(vdt)
+}
+
+// disputeStatement is a statement about a candidate, to be used within the dispute
+// resolution process. Statements are either in favour of the candidate's validity
+// or against it.
+type disputeStatement scale.VaryingDataType
+
+// Set will set a VaryingDataTypeValue using the underlying VaryingDataType
+func (distputedStatement *disputeStatement) Set(val scale.VaryingDataTypeValue) (err error) {
```
qdm12, line 72: Typo `distputedStatement`, at least rename to `disputedStatement`. Ideally rename to `d` to avoid any confusion (past dispute or present dispute?)

lib/babe/parachain_inherents.go
```diff
@@ -0,0 +1,345 @@
...
+		panic(err)
+	}
+
+	return disputeStatement(vdt)
+}
+
+// collatorID is the collator's relay-chain account ID
+type collatorID []byte
+
+// CollatorSignature is the signature on a candidate's block data signed by a collator.
```
qdm12, line 209: nit

```suggestion
// collatorSignature is the signature on a candidate's block data signed by a collator.
```

lib/babe/parachain_inherents.go
```diff
@@ -0,0 +1,345 @@
...
+// outboundHrmpMessage is an HRMP message seen from the perspective of a sender.
+type outboundHrmpMessage struct {
+	Recipient uint32 `scale:"1"`
+	Data      []byte `scale:"2"`
+}
+
+// validationCode is Parachain validation code.
+type validationCode []byte
+
+// HeadData is Parachain head data included in the chain.
```
qdm12, line 261:
```suggestion
// headData is Parachain head data included in the chain.
```

lib/babe/parachain_inherents.go
```diff
@@ -0,0 +1,345 @@
...
+
+// uncheckedSignedAvailabilityBitfield is a set of unchecked signed availability bitfields.
+// Should be sorted by validator index, ascending.
+type uncheckedSignedAvailabilityBitfield struct {
+	// The payload is part of the signed data. The rest is the signing context,
+	// which is known both at signing and at validation.
+	Payload []byte `scale:"1"`
+	// The index of the validator signing this statement.
+	ValidatorIndex uint32 `scale:"2"`
+	/// The signature by the validator of the signed payload.
```
qdm12, line 294: supernit 
```suggestion
	// The signature by the validator of the signed payload.
```

qdm12 (approved):

timwu20:
[I created a PR](https://github.com/ChainSafe/gossamer/pull/2851) against this branch after doing a little more investigation.   I noticed that the encoding differed when using `InherentsData.SetInt64Inherent` vs the same `InherentData` struct from substrate.  I ended up revising the receiver functions to a more generic `SetInherent` function which accepts `interface{}`.  Maybe we can constrain the values using generics and the new interface constraints for types that we know that the scale package will support.

timwu20 (changes requested):
[I created a PR](https://github.com/ChainSafe/gossamer/pull/2851) against this branch after doing a little more investigation.   I noticed that the encoding differed when using `InherentsData.SetInt64Inherent` vs the same `InherentData` struct from substrate.  I ended up revising the receiver functions to a more generic `SetInherent` function which accepts `interface{}` and fixes the encoding.  Maybe we can constrain the values using generics and the new interface constraints for types that we know that the scale package will support.

kishansagathiya:
@timwu20 
>https://github.com/ChainSafe/gossamer/pull/2851 against this branch after doing a little more investigation. I noticed that the encoding differed when using InherentsData.SetInt64Inherent vs the same InherentData struct from substrate. I ended up revising the receiver functions to a more generic SetInherent function which accepts interface{} and fixes the encoding. I [added some tests](https://github.com/ChainSafe/gossamer/pull/2851/files#diff-53aaae327ea03dd5123be933da4cfe3ebcad72aafe919b2d9006ac727805050c) to ensure the encodings are the same. Maybe we can constrain the values using generics and the new interface constraints for types that we know that the scale package will support.

I already have a a PR for that (https://github.com/ChainSafe/gossamer/pull/2815/).

kishansagathiya:
@timwu20 When you try to scale encode a struct, its private (not exported) field can be read by our scale package. I observed this with InherentsData struct which has `data` unexported. With that in mind, does it make sense to keep things exported in parachain inherents struct?

@qdm12

pkg/scale/encode.go
```diff
@@ -223,6 +225,32 @@ func (es *encodeState) encodeArray(in interface{}) (err error) {
 	return
 }
 
+func (es *encodeState) encodeMap(in interface{}) (err error) {
```
qdm12, line 228: Add a unit test for this

pkg/scale/encode.go
```diff
@@ -223,6 +225,32 @@ func (es *encodeState) encodeArray(in interface{}) (err error) {
...
+	v := reflect.ValueOf(in)
+	err = es.encodeLength(v.Len())
+	if err != nil {
+		return
+	}
+
+	for _, key := range v.MapKeys() {
+		err = es.marshal(key.Interface())
+		if err != nil {
+			return
```
qdm12, line 238:
```suggestion
			return fmt.Errorf("marshaling key: %w", err)
```

pkg/scale/encode.go
```diff
@@ -223,6 +225,32 @@ func (es *encodeState) encodeArray(in interface{}) (err error) {
...
+		}
+
+		mapValue := v.MapIndex(key)
+		if !mapValue.CanInterface() {
+			continue
+		}
+
+		err = es.marshal(mapValue.Interface())
+		if err != nil {
+			return
```
qdm12, line 248:
```suggestion
			return fmt.Errorf("marshaling value: %w", err)
```

pkg/scale/encode.go
```diff
@@ -223,6 +225,32 @@ func (es *encodeState) encodeArray(in interface{}) (err error) {
...
+	for _, key := range v.MapKeys() {
+		err = es.marshal(key.Interface())
+		if err != nil {
+			return
+		}
+
+		mapValue := v.MapIndex(key)
+		if !mapValue.CanInterface() {
+			continue
+		}
```
qdm12, line 244: Can this ever happen? :thinking: A unit test would be a nice way to find out.

lib/babe/build.go
```diff
@@ -222,22 +222,38 @@ func (b *BlockBuilder) buildBlockExtrinsics(slot Slot, rt runtime.Instance) []*t
...
+	parachainInherent := ParachainInherentData{
+		ParentHeader: *parent,
+	}
+
+	// add parachn0 and newheads
+	// for now we can use "empty" values, as we require parachain-specific
+	// logic to actually provide the data.
+
+	if err = idata.SetInherent(types.Parachn0, parachainInherent); err != nil {
+		return nil, fmt.Errorf("setting struct inherent %q: %w", types.Parachn0, err)
```
qdm12, line 249: nit since the method name changed to `SetInherent`, maybe this is better:

```suggestion
		return nil, fmt.Errorf("setting inherent %q: %w", types.Parachn0, err)
```

lib/babe/build.go
```diff
@@ -222,22 +222,38 @@ func (b *BlockBuilder) buildBlockExtrinsics(slot Slot, rt runtime.Instance) []*t
...
+	}
+
+	if err = idata.SetInherent(types.Newheads, []byte{0}); err != nil {
+		return nil, fmt.Errorf("setting struct inherent %q: %w", types.Newheads, err)
```
qdm12, line 253: nit since the method name changed to `SetInherent`, maybe this is better:

```suggestion
		return nil, fmt.Errorf("setting inherent %q: %w", types.Newheads, err)
```

dot/types/inherents.go
```diff
@@ -4,90 +4,78 @@
...
 type InherentsData struct {
-	data map[[8]byte]([]byte)
+	Data map[[8]byte][]byte
 }
 
 // NewInherentsData returns InherentsData
 func NewInherentsData() *InherentsData {
 	return &InherentsData{
-		data: make(map[[8]byte]([]byte)),
+		Data: make(map[[8]byte]([]byte)),
```
qdm12, line 59: nit / a bit out of scope: doen't this work:

```suggestion
		Data: make(map[[8]byte][]byte),
```

dot/types/inherents.go
```diff
@@ -4,90 +4,78 @@
...
+	// Parachn0 is an inherent key for parachains inherent.
+	Parachn0
+	// Newheads is an inherent key for new minimally-attested parachain heads.
+	Newheads
 )
 
+// Bytes returns a byte array of given inherent identifier.
+func (ii InherentIdentifier) Bytes() [8]byte {
+
+	kb := [8]byte{}
```
qdm12, line 31: nit

```suggestion
func (ii InherentIdentifier) Bytes() (kb [8]byte) {
```

dot/types/inherents.go
```diff
@@ -4,90 +4,78 @@
...
+	case Timstap0:
+		copy(kb[:], []byte("timstap0"))
+	case Babeslot:
+		copy(kb[:], []byte("babeslot"))
+	case Uncles00:
+		copy(kb[:], []byte("uncles00"))
+	case Parachn0:
+		copy(kb[:], []byte("parachn0"))
+	case Newheads:
+		copy(kb[:], []byte("newheads"))
```
qdm12, line 42: nit why is it named `kb`? Shouldn't it be `inherentIndentifier` / `identifier` / `identifierBytes`

qdm12:
> @timwu20 When you try to scale encode a struct, its private (not exported) field can be read by our scale package. I observed this with InherentsData struct which has `data` unexported. With that in mind, does it make sense to keep things exported in parachain inherents struct?
> 
> @qdm12

That's interesting. Most Go codecs usually don't have access to unexported fields, should we do the same thing @timwu20 ? Or is there a use case where we need access to unexported fields?

lib/babe/parachain_inherents_test.go
```diff
@@ -0,0 +1,290 @@
...
+	//    },
+	// };
+	// ----------------------------------------
+	// way to get inherents encoding from rust
+	// ----------------------------------------
+	// let mut inherents: sp_inherents::InherentData = sp_inherents::InherentData::new();
+	// inherents.put_data(*b"parachn0", &para_int).unwrap();
+	// println!("{:?}", inherents.encode());
+
+	parachainInherent := ParachainInherentData{}
```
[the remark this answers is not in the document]
kishansagathiya: done

timwu20:
> > @timwu20 When you try to scale encode a struct, its private (not exported) field can be read by our scale package. I observed this with InherentsData struct which has `data` unexported. With that in mind, does it make sense to keep things exported in parachain inherents struct?
> > @qdm12
> 
> That's interesting. Most Go codecs usually don't have access to unexported fields, should we do the same thing @timwu20 ? Or is there a use case where we need access to unexported fields?

That's how they implemented `InherentData` in substrate.  They manually encode it, and the trait ensures that the type that you are using can be scale encoded.

dot/types/inherents.go
```diff
@@ -4,90 +4,78 @@
...
+	}
+
+	return kb
+}
+
 // InherentsData contains a mapping of inherent keys to values
 // keys must be 8 bytes, values are a scale-encoded byte array
 type InherentsData struct {
-	data map[[8]byte]([]byte)
+	Data map[[8]byte][]byte
```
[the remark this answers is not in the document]
kishansagathiya: https://github.com/ChainSafe/gossamer/pull/2566#issuecomment-1259049249

lib/babe/build.go
```diff
@@ -222,22 +222,38 @@ func (b *BlockBuilder) buildBlockExtrinsics(slot Slot, rt runtime.Instance) []*t
...
 
 	// add babeslot
-	err = idata.SetInt64Inherent(types.Babeslot, slot.number)
+	err = idata.SetInherent(types.Babeslot, slot.number)
 	if err != nil {
 		return nil, err
 	}
 
-	ienc, err := idata.Encode()
+	parachainInherent := ParachainInherentData{
```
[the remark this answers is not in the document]
kishansagathiya: > I wonder if runtimes that don't require these inherents will just ignore this.

Yes, but is that problematic?

> shouldn't we only add the parachain inherent data for the polkadot/kusama/westend runtimes?

 It seems like currently polkadot uses `9281`,  kusama uses `9291` and westend uses `9290`. So, they all will read parachain inherents.

Can you elaborate which other runtimes should we care about? Do you mean previous versions? Do you mean custom runtimes?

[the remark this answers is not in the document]
timwu20: these 3 networks/runtimes are relay chains for parachains.  I don't think these inherents are required for running a standalone chain or a parachain.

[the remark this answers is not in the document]
kishansagathiya: tested this to be working on following chains

- dev
- dev-staking
- dev-v3substrate
- gssmr
- gssmr-staking
- gssmr-v3substrate

So, it looks like in case parachain inherents are not needed, they just get ignored

PR #2568, opened 2022-05-26
chore(ci): split unit test workflow by package

qdm12:
## Changes

Advantages:

- Faster feedback since jobs are parallelised
- All unit tests 'fail fast', if one matrix job fails, others are canceled

Disadvantages:

- we need to add new packages to the matrix
- code coverage is no longer CI-tracked (like integration tests already)

We should one day revert these changes once unit tests are fast (< 5 minutes imo)

## Tests

```sh
```


qdm12:
Closed in favor of #2887

PR #2572, opened 2022-05-30
fix(ci): fix dot/network unit tests

timwu20:
- Fixes the `dot/network` unit tests for the `v0.9.20` branch.

## Tests

```sh
go test ./dot/network
```

## Issues

contributes to #2540 

## Primary Reviewer

@kishansagathiya


dot/network/gossip_integration_test.go
```diff
@@ -0,0 +1,111 @@
...
+
+	"github.com/stretchr/testify/require"
+)
+
+// test gossip messages to connected peers
+func TestGossip(t *testing.T) {
+	if testing.Short() {
+		t.Skip("skipping TestGossip; currently, nothing is gossiped")
+	}
+
```
qdm12, line 22: Do we really want this? It's already tagged as integration, I think we don't need to check for `-short`

dot/network/gossip_integration_test.go
```diff
@@ -0,0 +1,111 @@
...
+	require.NoError(t, err)
+
+	time.Sleep(10 * time.Second)
+
+	hash, err := announceMessage.Hash()
+	require.NoError(t, err)
+
+	t.Logf("B %+v\n", nodeB.gossip.seenMap)
+	t.Logf("C %+v\n", nodeC.gossip.seenMap)
+	t.Logf("A %+v\n", nodeA.gossip.seenMap)
```
qdm12, line 101:
```suggestion
```

dot/network/gossip_integration_test.go
```diff
@@ -0,0 +1,111 @@
...
+		Number: 128 * 7,
+		Digest: types.NewDigest(),
+	}
+
+	time.Sleep(10 * time.Second)
+
+	_, err = nodeA.host.send(addrInfoB.ID, "/gossamer/test/0/block-announces/1", announceMessage)
+	require.NoError(t, err)
+
+	time.Sleep(10 * time.Second)
```
qdm12, line 94: I guess there is no way around those 10s sleeps right?

dot/network/gossip_integration_test.go
```diff
@@ -0,0 +1,111 @@
...
+	handlerC := newTestStreamHandler(testBlockAnnounceMessageDecoder)
+	nodeC.host.registerStreamHandler(nodeC.host.protocolID, handlerC.handleStream)
+
+	err = nodeC.host.connect(addrInfoA)
+	// retry connect if "failed to dial" error
+	if failedToDial(err) {
+		time.Sleep(TestBackoffTimeout)
+		err = nodeC.host.connect(addrInfoA)
+	}
+	require.NoError(t, err)
```
qdm12, line 73: write a helper function for those lines? it's used 3 times

dot/network/gossip_integration_test.go
```diff
@@ -0,0 +1,111 @@
...
+	"testing"
+	"time"
+
+	"github.com/ChainSafe/gossamer/dot/types"
+
+	"github.com/stretchr/testify/require"
+)
+
+// test gossip messages to connected peers
+func TestGossip(t *testing.T) {
```
qdm12, line 18: from the linter:

> other declaration of TestGossip (typecheck)

dot/network/gossip_test.go
```diff
@@ -84,10 +84,12 @@ func TestGossip(t *testing.T) {
...
 	}
 
-	_, err = nodeA.host.send(addrInfoB.ID, "", announceMessage)
+	time.Sleep(10 * time.Second)
+
+	_, err = nodeA.host.send(addrInfoB.ID, "/gossamer/test/0/block-announces/1", announceMessage)
 	require.NoError(t, err)
 
-	time.Sleep(TestMessageTimeout)
+	time.Sleep(10 * time.Second)
```
kishansagathiya, line 92: locally the test passes even with one second delay. I am guessing the delay requirement might be slightly different. 10s seems little high though. 
But seems like only potential way to find an optimum delay would be to run ci bunch of times.

.github/workflows/integration-tests.yml
```diff
@@ -26,7 +26,7 @@ jobs:
             github.com/ChainSafe/gossamer/dot,
             github.com/ChainSafe/gossamer/dot/core,
             github.com/ChainSafe/gossamer/dot/rpc/modules,
-            github.com/ChainSafe/gossamer/lib/babe,
+            github.com/ChainSafe/gossamer/lib/babe
```
qdm12, line 29: Is this necessary? :thinking: A bit OOS too. But sure, why not as well :smile:

qdm12 (approved):
Approved although CI seems to be failing still :thinking:

.github/workflows/integration-tests.yml
timwu20: :joy: didn't mean to do this.  I'll remove it

timwu20:
> Approved although CI seems to be failing still 🤔

It isn't gonna be green. I'm trying to get the unit tests suite green, but that will require more changes.

Issue #2567, opened 2022-05-25
Sending on a closed channel in networks package


kishansagathiya:
## Describe the bug

- found this on one of the ci unit tests logs `panic: send on closed channel`
- Relevant logs might be this 
```
2022-05-25T13:43:24.3252053Z 2022-05-25T13:43:24Z INFO started network service with supported protocols /p2p/id/delta/1.0.0, /ipfs/id/1.0.0, /ipfs/id/push/1.0.0, /ipfs/ping/1.0.0, /gossamer/test/0/sync/2, /gossamer/test/0/light/2, /gossamer/test/0/block-announces/1, /gossamer/test/0/transactions/1	pkg=network
2022-05-25T13:43:24.3252553Z panic: send on closed channel
2022-05-25T13:43:24.3252708Z 
2022-05-25T13:43:24.3252815Z goroutine 10060 [running]:
2022-05-25T13:43:24.3253197Z github.com/ChainSafe/gossamer/dot/peerset.(*PeerSet).disconnect(0xc04b8d8200, 0x0, 0x0, {0xc04d65fe00, 0x1, 0x1?})
2022-05-25T13:43:24.3253660Z 	/home/runner/work/gossamer/gossamer/dot/peerset/peerset.go:693 +0x26f
2022-05-25T13:43:24.3254166Z github.com/ChainSafe/gossamer/dot/peerset.(*PeerSet).listenAction(0xc04b8d8200, {0x13b6200, 0xc04d62c380})
2022-05-25T13:43:24.3254611Z 	/home/runner/work/gossamer/gossamer/dot/peerset/peerset.go:752 +0x1b4
2022-05-25T13:43:24.3254996Z created by github.com/ChainSafe/gossamer/dot/peerset.(*PeerSet).start
2022-05-25T13:43:24.3255402Z 	/home/runner/work/gossamer/gossamer/dot/peerset/peerset.go:715 +0xea
2022-05-25T13:43:24.3255751Z FAIL	github.com/ChainSafe/gossamer/dot/network	77.287s
```

## Expected Behavior

- We should not be sending things on closed channels

## Current Behavior

-

## Possible Solution

-

## To Reproduce
- Running`./dot/network` unit tests should do it.

## Log output
- I have uploaded full CI logs here https://gist.github.com/kishansagathiya/b56af71482671c6c64aed8a1e9a60afa

## Specification

- go version:
- gossamer version:
- gossamer commit tag:
- gossamer commit hash:
- operating system:
[...]

PR #2573, opened 2022-05-31
fix(dot/peerset): fix sending on closed channel race condition when dropping peer

timwu20:
## Changes

- sets channel to nil on close
- nil check before producing to channel

## Tests

```sh
go test ./dot/network
```

## Issues

#2567

## Primary Reviewer

@kishansagathiya

### DIFF
--- a/dot/peerset/peerset.go
+++ b/dot/peerset/peerset.go
@@ -690,10 +690,12 @@ func (ps *PeerSet) disconnect(setIdx int, reason DropReason, peers ...peer.ID) e
 			return fmt.Errorf("cannot disconnect: %w", err)
 		}
 
-		ps.resultMsgCh <- Message{
-			Status: Drop,
-			setID:  uint64(setIdx),
-			PeerID: pid,
+		if ps.resultMsgCh != nil {
+			ps.resultMsgCh <- Message{
+				Status: Drop,
+				setID:  uint64(setIdx),
+				PeerID: pid,
+			}
 		}
 
 		// TODO: figure out the condition of connection refuse.
@@ -765,6 +767,7 @@ func (ps *PeerSet) periodicallyAllocateSlots(ctx context.Context) {
 	defer func() {
 		ticker.Stop()
 		close(ps.resultMsgCh)
+		ps.resultMsgCh = nil
 	}()
 
 	for {
--- a/dot/peerset/peerset.go
+++ b/dot/peerset/peerset.go
@@ -690,12 +690,10 @@ func (ps *PeerSet) disconnect(setIdx int, reason DropReason, peers ...peer.ID) e
 			return fmt.Errorf("cannot disconnect: %w", err)
 		}
 
-		if ps.resultMsgCh != nil {
-			ps.resultMsgCh <- Message{
-				Status: Drop,
-				setID:  uint64(setIdx),
-				PeerID: pid,
-			}
+		ps.resultMsgCh <- Message{
+			Status: Drop,
+			setID:  uint64(setIdx),
+			PeerID: pid,
 		}
 
 		// TODO: figure out the condition of connection refuse.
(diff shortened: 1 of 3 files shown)


dot/peerset/peerset.go
```diff
@@ -690,10 +690,12 @@ func (ps *PeerSet) disconnect(setIdx int, reason DropReason, peers ...peer.ID) e
...
-		ps.resultMsgCh <- Message{
-			Status: Drop,
-			setID:  uint64(setIdx),
-			PeerID: pid,
+		if ps.resultMsgCh != nil {
+			ps.resultMsgCh <- Message{
+				Status: Drop,
+				setID:  uint64(setIdx),
+				PeerID: pid,
+			}
```
qdm12, line 698: Shouldn't the producer close the channel? So somewhere here :thinking:

kishansagathiya (commented):
So, channel `ps.resultMsgCh` is used in two goroutines `ps.listenAction(ctx)` and `ps.periodicallyAllocateSlots(ctx)`.

It looks like problems is happening because when a context is done `ps.periodicallyAllocateSlots(ctx)` processes it and closes the channel, but `ps.listenAction(ctx)` does not get to know about it until next for loop iteration.

@timwu20 It looks like you have fixed the problem for `disconnect` action, but the give problem could still occur in other actions such as `removeReservedPeers`, `removePeer` and `incoming`.

We can make a similar fix in the remaining actions or check if context was done using `ctx.Err()` before using the channel.

timwu20:
> So, channel `ps.resultMsgCh` is used in two goroutines `ps.listenAction(ctx)` and `ps.periodicallyAllocateSlots(ctx)`.
> 
> It looks like problems is happening because when a context is done, `ps.periodicallyAllocateSlots(ctx)` processes it and closes the channel, but `ps.listenAction(ctx)` does not get to know about it until next for loop iteration.
> 
> @timwu20 It looks like you have fixed the problem for `disconnect` action, but the given problem could still occur in other actions such as `removeReservedPeers`, `removePeer` and `incoming`.

Thanks for pointing this out.  I'll add the check for all the writes to the channel.  I was just debugging the panic produced by running the tests.

> We can make a similar fix in the remaining actions or check if context was done using `ctx.Err()` before using the channel. I like the later approach better (I don't have a reason why :)).

I'd prefer not to check `ctx.Err()` since in the future I'd rather remove context from the service architecture entirely.

dot/peerset/peerset.go
timwu20: I think there are problems with the arch of `peerset.Handler` that we should address at some point.  
- Right now this channel is returned by `Handler.Messages()`, but it's not a true pipeline.  If `handler.Messages()` was called again and was consumed, then messages would be split among all readers.  
- IMO the producer is `peerset.Peerset` so the producer is actually closing the channel.  Just on another goroutine.

qdm12 (approved):

kishansagathiya (approved):
Looks good! Just two more comments with easy fix!!

dot/peerset/peerset.go
```diff
@@ -712,15 +712,27 @@ func (ps *PeerSet) start(ctx context.Context, actionQueue chan action) {
...
+	go ps.listenActionAllocSlots(ctx)
 }
 
-func (ps *PeerSet) listenAction(ctx context.Context) {
+func (ps *PeerSet) listenActionAllocSlots(ctx context.Context) {
+	ticker := time.NewTicker(ps.nextPeriodicAllocSlots)
+
+	defer func() {
+		ticker.Stop()
+		close(ps.resultMsgCh)
```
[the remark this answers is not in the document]
timwu20: `actionQueue` is closed in `Handler.Stop()`.  Seems like it's passed into `PeerSet.start`.  It's kind of an odd design to be honest, but I'd rather not refactor the `Handler` to `PeerSet` relationship.

Issue #2591, opened 2022-06-07
Gossamer cannot produce blocks in the next epoch if finalization stales


EclesioMeloJunior:
## Issue summary

- While running a gossamer node in a cross-client network where the finalization is not occurring the gossamer node is able to produce blocks until the epoch changes, once the epoch changes it cannot produce or sync more blocks since we rely on database persisted epoch data and the node only persist the next epoch data once the finalization occurs and in a case where finalization is not working the next epoch data is not persisted. 

- The gossamer node should keep producing blocks if the finalization stops as both gadgets (production and finalization) should be independant.

## Other information and links

- N/A

kishansagathiya:
Even when we wait for finalisation, we are failing to write next epoch data and config data to database. So sometime `info.Header` does not contain pre runtime digests, resulting in this
```
header digest len  0
2022-06-08T17:52:16+05:30 EROR failed to persist babe next epoch data: cannot get epoch for block 0 (0xee87cc7ac5e87460094ec4ba29bbf8f28fb7741b69e7edb142233c3dd990b365): header does not contain pre-runtime digest	digest.go:L244	pkg=digest
header digest len  0
2022-06-08T17:52:16+05:30 EROR failed to persist babe next epoch config: cannot get epoch for block 0 (0xee87cc7ac5e87460094ec4ba29bbf8f28fb7741b69e7edb142233c3dd990b365): header does not contain pre-runtime digest	digest.go:L249	pkg=digest
2
```

kishansagathiya:
There are more problem with epoch handling. Seeing a lot of these logs
```
2022-06-08T18:30:12+05:30 EROR block data processing for block with hash 0xd671d24ff29cf990010bfc98a54981e06079ad71f900f86881e30b186ea5d158 failed: could not verify block: failed to get verifier info for block 23: failed to get epoch data for epoch 1: failed to get epoch data from memory: cannot verify the ancestry: end node does not exist	chain_processor.go:L84	pkg=sync
```

PR #2592, opened 2022-06-08
fix(state/epoch): assign epoch 1 when block number is 0

EclesioMeloJunior:
## Changes

- The genesis block does not contain any BABE digest since this information is already persisted using the information in the `spec.json`

## Tests

When you start the gossamer node you should not see 
```
2022-06-08T17:52:16+05:30 EROR failed to persist babe next epoch data: cannot get epoch for block 0 (0xee87cc7ac5e87460094ec4ba29bbf8f28fb7741b69e7edb142233c3dd990b365): header does not contain pre-runtime digest	digest.go:L244	pkg=digest
2022-06-08T17:52:16+05:30 EROR failed to persist babe next epoch config: cannot get epoch for block 0 (0xee87cc7ac5e87460094ec4ba29bbf8f28fb7741b69e7edb142233c3dd990b365): header does not contain pre-runtime digest	digest.go:L249	pkg=digest
```
## Issues

- Related #2591

## Primary Reviewer

@kishansagathiya

### DIFF
--- a/dot/state/epoch.go
+++ b/dot/state/epoch.go
@@ -543,20 +543,19 @@ func (s *EpochState) StoreBABENextConfigData(epoch uint64, hash common.Hash, nex
 // check if the header is in the database then it's been finalized and
 // thus we can also set the corresponding EpochData in the database
 func (s *EpochState) FinalizeBABENextEpochData(finalizedHeader *types.Header) error {
-	if finalizedHeader.Number == 0 {
-		return nil
-	}
-
 	s.nextEpochDataLock.Lock()
 	defer s.nextEpochDataLock.Unlock()
 
-	finalizedBlockEpoch, err := s.GetEpochForBlock(finalizedHeader)
-	if err != nil {
-		return fmt.Errorf("cannot get epoch for block %d (%s): %w",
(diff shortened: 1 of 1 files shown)


kishansagathiya (approved):

EclesioMeloJunior:
> Nice 👍
> 
> Probably a dumb question, but should we force-set the BABE digest for the first block finalized header somewhere above in the call stack?

This is a good question and I think it is possible but requires a larger refactor

dot/state/epoch.go
```diff
@@ -543,6 +543,10 @@ func (s *EpochState) StoreBABENextConfigData(epoch uint64, hash common.Hash, nex
 // check if the header is in the database then it's been finalized and
 // thus we can also set the corresponding EpochData in the database
 func (s *EpochState) FinalizeBABENextEpochData(finalizedHeader *types.Header) error {
+	if finalizedHeader.Number == 0 {
```
timwu20, line 546: Can we add this check once in `Handler.HandleBlockFinalisation` instead of in both of these functions?

timwu20 (changes requested):

qdm12 (approved):

timwu20 (approved):

PR #2593, opened 2022-06-09
Eclesio/fix/block production for epochs

EclesioMeloJunior:
## Tests

```sh
```

## Primary Reviewer

@timwu20

### DIFF
--- a/dot/state/epoch.go
+++ b/dot/state/epoch.go
@@ -578,13 +578,15 @@ func (s *EpochState) FinalizeBABENextEpochData(finalizedHeader *types.Header) er
 	s.nextEpochDataLock.Lock()
 	defer s.nextEpochDataLock.Unlock()
 
-	finalizedBlockEpoch, err := s.GetEpochForBlock(finalizedHeader)
-	if err != nil {
-		return fmt.Errorf("cannot get epoch for block %d (%s): %w",
-			finalizedHeader.Number, finalizedHeader.Hash(), err)
+	var nextEpoch uint64 = 1
+	if finalizedHeader.Number != 0 {
+		finalizedBlockEpoch, err := s.GetEpochForBlock(finalizedHeader)
+		if err != nil {
+			return fmt.Errorf("cannot get epoch for block %d (%s): %w",
+				finalizedHeader.Number, finalizedHeader.Hash(), err)
+		}
+		nextEpoch = finalizedBlockEpoch + 1
 	}
-
-	nextEpoch := finalizedBlockEpoch + 1
 
 	epochInDatabase, err := s.getEpochDataFromDatabase(nextEpoch)
 
@@ -632,13 +634,16 @@ func (s *EpochState) FinalizeBABENextConfigData(finalizedHeader *types.Header) e
 	s.nextConfigDataLock.Lock()
 	defer s.nextConfigDataLock.Unlock()
 
-	finalizedBlockEpoch, err := s.GetEpochForBlock(finalizedHeader)
-	if err != nil {
-		return fmt.Errorf("cannot get epoch for block %d (%s): %w",
-			finalizedHeader.Number, finalizedHeader.Hash(), err)
-	}
+	var nextEpoch uint64 = 1
+	if finalizedHeader.Number != 0 {
+		finalizedBlockEpoch, err := s.GetEpochForBlock(finalizedHeader)
+		if err != nil {
+			return fmt.Errorf("cannot get epoch for block %d (%s): %w",
+				finalizedHeader.Number, finalizedHeader.Hash(), err)
+		}
 
-	nextEpoch := finalizedBlockEpoch + 1
+		nextEpoch = finalizedBlockEpoch + 1
+	}
 
 	configInDatabase, err := s.getConfigDataFromDatabase(nextEpoch)
 
--- a/dot/state/epoch.go
+++ b/dot/state/epoch.go
@@ -18,10 +18,11 @@ import (
 )
 
 var (
-	ErrEpochNotInMemory  = errors.New("epoch not found in memory map")
-	errHashNotInMemory   = errors.New("hash not found in memory map")
-	ErrEpochDataNotFound = errors.New("epoch data not found in the database")
-	errHashNotPersisted  = errors.New("hash with next epoch not found in database")
+	ErrEpochNotInMemory   = errors.New("epoch not found in memory map")
+	errHashNotInMemory    = errors.New("hash not found in memory map")
+	errEpochNotInDatabase = errors.New("epoch data not found in the database")
+	errConfigNotFound     = errors.New("config data not found")
+	errHashNotPersisted   = errors.New("hash with next epoch not found in database")
 )
 
 var (
@@ -257,14 +258,19 @@ func (s *EpochState) GetEpochData(epoch uint64, header *types.Header) (*types.Ep
 
 	//  lookup in-memory only if header is given
 	if header != nil && errors.Is(err, chaindb.ErrKeyNotFound) {
-		epochData, err = s.getEpochDataFromMemory(epoch, header)
+		inMemoryEpochData, err := retrieveFromMemory(s.nextEpochData, s, epoch, header)
 		if err != nil {
 			return nil, fmt.Errorf("failed to get epoch data from memory: %w", err)
+		}
+
+		epochData, err = inMemoryEpochData.ToEpochData()
+		if err != nil {
+			return nil, fmt.Errorf("cannot transform into epoch data: %w", err)
 		}
 	}
 
 	if epochData == nil {
-		return nil, ErrEpochDataNotFound
+		return nil, errEpochNotInDatabase
 	}
 
 	return epochData, nil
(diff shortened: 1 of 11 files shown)


dot/state/epoch.go
```diff
@@ -397,26 +379,36 @@ func (s *EpochState) getConfigDataFromDatabase(epoch uint64) (*types.ConfigData,
...
+		// is not fully imported by the blocktree, in this case we will use
+		// its parent header which migth be already imported.
+		if errors.Is(err, blocktree.ErrEndNodeNotFound) {
+			parentHeader, err := es.blockState.GetHeader(header.ParentHash)
+			if err != nil {
+				return nil, fmt.Errorf("cannot get parent header: %w", err)
+			}
+
+			return retrieveFromMemory(nextEpochMap, es, epoch, parentHeader)
+		}
```
noot, line 404: what do you mean the header is not fully imported? a block can only be imported or not imported afaik

lib/babe/epoch.go
```diff
@@ -24,15 +24,25 @@ func (b *Service) initiateEpoch(epoch uint64) (*epochData, error) {
 		}
 	}
 
-	epochData, startSlot, err := b.getEpochDataAndStartSlot(epoch)
+	bestBlockHeader, err := b.blockState.BestBlockHeader()
+	if err != nil {
+		return nil, fmt.Errorf("cannot get the best block header")
```
noot, line 29:
```suggestion
		return nil, fmt.Errorf("cannot get the best block header: %w", err)
```

noot (commented):
looks good so far, I'm surprised it didn't work previously actually as we were already storing the digests in memory if not finalized. but if this fixes it that's good!

dot/state/epoch.go
EclesioMeloJunior: The problem I notice is that at the block where we should change to the next epoch, I got the error `blocktree.ErrEndNodeNotFound` on function `chainProcessor.processBlockData`.

This method (`processBlockData`) calls the method `handleHeader` which calls the `babeVerifier.VerifyBlock(header)`, and the variable `header` contains the header value of the first block in the next epoch. The `VerifyBlock` method (`lib/babe/verify.go`) calls the method `getVerifierInfo` which calls the `epochState.GetEpochData` using the variables `header` and `epoch`. 

The `epochState.GetEpochData` did not find anything in the database for epoch 1 so we look up in the memory map using the `header` and that is when the problem happens, the block state is not aware of the `header` (first block in the next epoch) and return: `end node does not exist`. 

The fully error is:
```
2022-06-08T18:30:12+05:30 EROR block data processing for block with hash 0xd671d24ff29cf990010bfc98a54981e06079ad71f900f86881e30b186ea5d158 failed: could not verify block: failed to get verifier info for block 23: failed to get epoch data for epoch 1: failed to get epoch data from memory: cannot verify the ancestry: end node does not exist	chain_processor.go:L84	pkg=sync
```

noot: that makes sense, this seems like a decent fix, but maybe we should just pass the parent hash into this function instead, since the parent would always be imported already (i think). what do you think?

dot/state/epoch.go
```diff
@@ -357,28 +326,41 @@ func (s *EpochState) setLatestConfigData(epoch uint64) error {
...
-	if err != nil && !errors.Is(err, chaindb.ErrKeyNotFound) {
-		return nil, fmt.Errorf("failed to get config data from database: %w", err)
-	} else if header == nil {
-		// if no header is given then skip the lookup in-memory
-		return configData, nil
-	}
+	for tryEpoch := epoch; tryEpoch >= 0; tryEpoch-- {
+		configData, err = s.getConfigDataFromDatabase(tryEpoch)
+		if err == nil && configData != nil {
+			return configData, nil
```
kishansagathiya, line 343: Are you sure about this part? I seems like we are returning config data for a different epoch that the one asked in the argument. 
I did not find anything explicit in the spec

dot/state/epoch.go
```diff
@@ -357,28 +326,41 @@ func (s *EpochState) setLatestConfigData(epoch uint64) error {
...
-	configData, err := s.getConfigDataFromDatabase(epoch)
-	if err == nil && configData != nil {
-		return configData, nil
+// GetConfigData returns the newest config data for a given epoch persisted in database
+// otherwise tries to get the data from the in-memory map using the header. If we don't
+// find any config data for the current epoch we lookup in the previous epochs, as the spec says:
+// - The supplied configuration data are intended to be used from the next epoch onwards.
+// If the header params is nil then it will search only in the database.
+func (s *EpochState) GetConfigData(epoch uint64, header *types.Header) (configData *types.ConfigData, err error) {
+	if header != nil {
```
kishansagathiya, line 335: this does not seem right

lib/babe/epoch.go
```diff
@@ -14,7 +14,7 @@ import (
 // initiateEpoch sets the epochData for the given epoch, runs the lottery for the slots in the epoch,
 // and stores updated EpochInfo in the database
 func (b *Service) initiateEpoch(epoch uint64) (*epochData, error) {
-	logger.Debugf("initiating epoch %d", epoch)
+	logger.Infof("initiating epoch %d", epoch)
```
kishansagathiya, line 17: debugging leftover

dot/state/epoch.go
```diff
@@ -251,14 +254,26 @@ func (s *EpochState) GetEpochData(epoch uint64, header *types.Header) (*types.Ep
...
+		}
+
+		epochData, err = inMemoryEpochData.ToEpochData()
+		if err != nil {
+			return nil, fmt.Errorf("cannot transform into epoch data: %w", err)
+		}
+	}
+
+	if epochData == nil {
+		return nil, errEpochNotInDatabase
```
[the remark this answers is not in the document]
EclesioMeloJunior: Got it! Just a small problem with your suggestion is the `if header != nil` that should be `if header == nil`

dot/state/epoch.go
EclesioMeloJunior: In the `polkadot-host-spec_v0.1.1.pdf` in the section `6.1.2 Runtime-to-Consensus Engine Message` there is the description of the `Next Config Data` which says:

> - Next Config Data: These messages are only issued on configuration change and in the first block of an epoch. The supplied configuration data are intended to be used from the next epoch onwards.

The next config data is not issued every epoch as next epoch data. So in the spec when they say:  _The supplied configuration data are intended to be used from the next epoch onwards_ means we should use the closest config data for epoch `X`, if epoch `X` does not contain any config data we should lookup in the epoch `X-1` trying to retrieve the config data.

dot/state/epoch.go
```diff
@@ -404,26 +380,36 @@ func (s *EpochState) getConfigDataFromDatabase(epoch uint64) (*types.ConfigData,
 	return info, nil
 }
 
-// getConfigDataFromMemory retrieves the BABE config data for a given epoch that belongs to the header parameter
-func (s *EpochState) getConfigDataFromMemory(epoch uint64, header *types.Header) (*types.ConfigData, error) {
-	s.nextConfigDataLock.RLock()
-	defer s.nextConfigDataLock.RUnlock()
+func retrieveFromMemory[T types.NextEpochData | types.NextConfigData](nextEpochMap map[uint64]map[common.Hash]T,
```
timwu20, line 383: could this not still be a receiver function on `EpochState`?

timwu20 (commented):
why are we changing `chain/dev/genesis-spec.json`?

dot/state/epoch.go
EclesioMeloJunior: Yes, I can pass the pointer to `BlockState` instead of `EpochState`

dot/state/epoch.go
kishansagathiya: thanks!

kishansagathiya (approved):
Good Stuff :tada:

dot/state/epoch.go
timwu20, line 383: Did you consider using a type for `nextEpochMap`? 

```
type nextEpochMap[T types.NextEpochData | types.NextConfigData] map[uint64]map[common.Hash]T

func (nem nextEpochMap[T]) Retrieve(blockState *BlockState, epoch uint64, header *types.Header) (*T, error) {
	atEpoch, has := nem[epoch]
	if !has {
		return nil, fmt.Errorf("%w: %d", ErrEpochNotInMemory, epoch)
	}

	headerHash := header.Hash()
	for hash, value := range atEpoch {
		isDescendant, err := blockState.IsDescendantOf(hash, headerHash)

		// sometimes while moving to the next epoch is possible the header
		// is not fully imported by the blocktree, in this case we will use
		// its parent header which migth be already imported.
		if errors.Is(err, blocktree.ErrEndNodeNotFound) {
			parentHeader, err := blockState.GetHeader(header.ParentHash)
			if err != nil {
				return nil, fmt.Errorf("cannot get parent header: %w", err)
			}

			return retrieveFromMemory(nem, blockState, epoch, parentHeader)
		}

		if err != nil {
			return nil, fmt.Errorf("cannot verify the ancestry: %w", err)
		}

		if isDescendant {
			return &value, nil
		}
	}

	return nil, fmt.Errorf("%w: %s", errHashNotInMemory, headerHash)
}
```

I took a stab and pushed to another branch: https://github.com/ChainSafe/gossamer/pull/2633/files

EclesioMeloJunior: Oh, I didn't, this is really interesting!

timwu20 (approved):
looks like CI is still failing?

Issue #2541, opened 2022-05-12
Gossamer Can't Build Blocks on a Cross-Client v0.9.20 Devnet 


danforbes:
Ensure that Gossamer can author blocks on-par with Substrate nodes in a cross-client devnet.

kishansagathiya:
I think there are a bunch of issues here. One of them is https://github.com/ChainSafe/gossamer/issues/2584. This is rather frequent. Like you will rarely not see this issue.

Other issue is that substrate drops gossamer nodes after a while. I don't see a clear pattern over here, like I have gossamer nodes getting dropped after building few blocks and after building 42 blocks as well. I will know more about this problem after I fix https://github.com/ChainSafe/gossamer/issues/2584. We have seen gossamer dropping issues before and fixed them. This time blocks might be dropping for different reason.

There are few different errors that I see as well regarding epoch handling. But It is not clear how to reproduce them. Some of those error happen on the second run of gossamer node. So, I am guessing based on what ends up getting written to the db (this would mostly depend on where the program is when you stop the gossamer node), and how we get info from db on restarting the node, the errors could change. But again fixing the first issue will give me better idea about other issues.

kishansagathiya:
Looks like sometimes blocks built by gossamer are improper with `Import failed: Unexpected epoch change  `
```
022-06-06 20:44:52 Error with block built on 0x41d30e3127c88687f1ce81a3ffd8adec0a3f59d224987aa6fffd4ba44ecb5d5e: Import failed: Unexpected epoch change    
2022-06-06 20:44:55 💤 Idle (1 peers), best: #3 (0x41d3…5d5e), finalized #0 (0xee87…b365), ⬇ 0.3kiB/s ⬆ 0.2kiB/s    
2022-06-06 20:44:56 🙌 Starting consensus session on top of parent 0x41d30e3127c88687f1ce81a3ffd8adec0a3f59d224987aa6fffd4ba44ecb5d5e    
2022-06-06 20:44:56 🎁 Prepared block for proposing at 4 (0 ms) [hash: 0xd18a26f25d6d3783f6d1b035285ef03261b0fd8f8a5a1908428c67f2f452fbbe; parent_hash: 0x41d3…5d5e; extrinsics (1): [0xd558…c4eb]]    
[...]

PR #2596, opened 2022-06-10
Remove checks for equivocatory blocks

### LINKED ISSUE #2597: When checking for equivocatory blocks error only if both are created by primary author
## Describe the bug

- We have logic to find blocks that were authored by same node and have same block number. In such case we error. This could happen when one block is authored as secondary author and other as primary author. We reduce this check only to primary author. That is error only if both blocks we authored as primary author. 

https://github.com/ChainSafe/gossamer/blob/8582cb20a37d574e6397f980f919c6dd4dfb24a0/lib/babe/verify.go#L367-L369

kishansagathiya:
## Changes
We have logic to find different blocks with same block number and same author. If we find such blocks we throw and error.
Some of the discussion is here https://github.com/ChainSafe/gossamer/issues/2541#issuecomment-1152010241
https://github.com/ChainSafe/gossamer/issues/2541#issuecomment-1152395681

As discussed with @noot , it's probably authoring one as primary and one as secondary. If we don't error the fork should get resolved eventually.

## Tests

```sh
```

## Issues

Fixes #2597
## Primary Reviewer

@timwu20


lib/babe/verify.go
```diff
@@ -354,17 +354,40 @@ func (b *verifier) verifyAuthorshipRight(header *types.Header) error {
 			continue
 		}
 
+		currentPreDigestItem := currentHeader.Digest.Types[0]
+
+		currentPreDigest, ok := currentPreDigestItem.Value().(types.PreRuntimeDigest)
+		if !ok {
+			return fmt.Errorf("first digest item is not pre-digest")
```
EclesioMeloJunior, line 361: I would suggest to create the error `ErrNoFirstPreDigest` and set in the error message the type of `currentPreDigestItem.Value()`
```suggestion
			return fmt.Errorf("%w: got %T", ErrNoFirstPreDigest, currentPreDigestItem.Value())
```

lib/babe/verify.go
```diff
@@ -354,17 +354,40 @@ func (b *verifier) verifyAuthorshipRight(header *types.Header) error {
 			continue
 		}
 
+		currentPreDigestItem := currentHeader.Digest.Types[0]
```
EclesioMeloJunior, line 357: should we check the `len()` of `Types` before getting the first index?

EclesioMeloJunior (changes requested):

lib/babe/verify_test.go
```diff
@@ -697,22 +665,134 @@ func Test_verifier_verifyAuthorshipRight(t *testing.T) {
...
+		name     string
+		verifier verifier
+		header   *types.Header
+		expErr   error
+	}{
 		{
 			name:     "equivocate - primary",
-			verifier: *babeVerifier7,
-			header:   header7,
+			verifier: *verifierEquivocatoryPrimary,
```
EclesioMeloJunior, line 782: This verifier is using a mock that defines `mockBlockStateEquiv1.EXPECT().GetAllBlocksAtDepth(gomock.Any())` but it is possible to replace the `gomock.Any()` with the `headerEquivocatoryPrimary.ParentHash`, same for the others mocks using the `gomock.Any()`

EclesioMeloJunior (changes requested):

edwardmack (approved):

lib/babe/verify_test.go
kishansagathiya: :+1:  addressed

lib/babe/verify.go
```diff
@@ -347,26 +347,55 @@ func (b *verifier) verifyAuthorshipRight(header *types.Header) error {
...
 			existingBlockProducerIndex = d.AuthorityIndex
 		case types.BabeSecondaryPlainPreDigest:
 			existingBlockProducerIndex = d.AuthorityIndex
 		}
 
-		// same authority won't produce two different blocks at the same block number
-		if currentBlockProducerIndex == existingBlockProducerIndex && hash != header.Hash() {
+		// same authority won't produce two different blocks at the same block number as primary block producer
+		if currentBlockProducerIndex == existingBlockProducerIndex &&
+			hash != header.Hash() &&
```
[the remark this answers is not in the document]
kishansagathiya: :+1: addressed

lib/babe/verify.go
```diff
@@ -347,26 +347,55 @@ func (b *verifier) verifyAuthorshipRight(header *types.Header) error {
...
 		}
 
 		currentBlockProducerIndex, err := getAuthorityIndex(currentHeader)
 		if err != nil {
+			logger.Errorf("failed to get authority index %s", err)
 			continue
 		}
 
+		if len(currentHeader.Digest.Types) == 0 {
+			return fmt.Errorf("current header missing digest")
```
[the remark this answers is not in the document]
kishansagathiya: :+1: addressed

dot/types/babe.go
```diff
@@ -107,7 +110,7 @@ func GetSlotFromHeader(header *Header) (uint64, error) {
 
 	preDigest, ok := header.Digest.Types[0].Value().(PreRuntimeDigest)
 	if !ok {
-		return 0, fmt.Errorf("first digest item is not pre-digest")
+		return 0, ErrNoFirstPreDigest
```
EclesioMeloJunior, line 113: Would be nice to put in the message the type we got
```suggestion
		return 0, fmt.Errorf("%w: got %T", ErrNoFirstPreDigest, header.Digest.Types[0].Value())
```

lib/babe/verify.go
```diff
@@ -277,12 +277,12 @@ func (b *verifier) verifyAuthorshipRight(header *types.Header) error {
...
 	preDigest, ok := preDigestItem.Value().(types.PreRuntimeDigest)
 	if !ok {
-		return fmt.Errorf("first digest item is not pre-digest")
+		return fmt.Errorf("%w: got %T", types.ErrNoFirstPreDigest, preDigestItem.Value())
 	}
 
 	seal, ok := sealItem.Value().(types.SealDigest)
 	if !ok {
-		return fmt.Errorf("last digest item is not seal")
+		return errLastDigestItemNotSeal
```
EclesioMeloJunior, line 285:
```suggestion
		return fmt.Errorf("%w: got %T", errLastDigestItemNotSeal, sealItem.Value())
```

lib/babe/verify.go
```diff
@@ -344,29 +344,58 @@ func (b *verifier) verifyAuthorshipRight(header *types.Header) error {
...
 		}
 
 		currentBlockProducerIndex, err := getAuthorityIndex(currentHeader)
 		if err != nil {
+			logger.Errorf("failed to get authority index %s", err)
 			continue
 		}
 
+		if len(currentHeader.Digest.Types) == 0 {
+			logger.Errorf("current header missing digest")
```
EclesioMeloJunior, line 361: should we return or continue to the next iteration?

lib/babe/verify.go
```diff
@@ -344,29 +344,58 @@ func (b *verifier) verifyAuthorshipRight(header *types.Header) error {
...
 			existingBlockProducerIndex = d.AuthorityIndex
 		case types.BabeSecondaryPlainPreDigest:
 			existingBlockProducerIndex = d.AuthorityIndex
 		}
 
-		// same authority won't produce two different blocks at the same block number
-		if currentBlockProducerIndex == existingBlockProducerIndex && hash != header.Hash() {
+		// same authority won't produce two different blocks at the same block number as primary block producer
+		if currentBlockProducerIndex == existingBlockProducerIndex &&
+			currentHash.Equal(header.Hash()) &&
```
EclesioMeloJunior, line 397: should be not equal, right?
```suggestion
			!currentHash.Equal(header.Hash()) &&
```

EclesioMeloJunior (changes requested):

dot/types/babe.go
EclesioMeloJunior, line 113:
```suggestion
		return 0, fmt.Errorf("%w: type=%T", ErrNoFirstPreDigest, header.Digest.Types[0].Value())
```

lib/babe/verify.go
```diff
@@ -277,12 +277,12 @@ func (b *verifier) verifyAuthorshipRight(header *types.Header) error {
 
 	preDigest, ok := preDigestItem.Value().(types.PreRuntimeDigest)
 	if !ok {
-		return fmt.Errorf("first digest item is not pre-digest")
+		return fmt.Errorf("%w: got %T", types.ErrNoFirstPreDigest, preDigestItem.Value())
```
EclesioMeloJunior, line 280: in some places, you've used `fmt.Errorf("%w: type=%T")` and in others you've used `fmt.Errorf("%w: got %T")` so would be nice to choose a single one and keep it as the default. IMO I would choose `fmt.Errorf("%w: got %T")` as the error explain that we did not receive the predigest but we got another type -> `ErrNoFirstPreDigest: got %T`

EclesioMeloJunior (changes requested):

lib/babe/verify.go
```diff
@@ -344,29 +344,58 @@ func (b *verifier) verifyAuthorshipRight(header *types.Header) error {
...
+		if err != nil {
+			return fmt.Errorf("failed to verify pre-runtime digest: %w", err)
+		}
+
+		var isCurrentBlockProducerPrimary bool
+		switch currentBabePreDigest.(type) {
+		case types.BabePrimaryPreDigest:
+			isCurrentBlockProducerPrimary = true
+		default:
+		}
```
timwu20, line 381:
```suggestion
		_, isCurrentBlockProducerPrimary := currentBabePreDigest.(types.BabePrimaryPreDigest)
```

lib/babe/verify.go
```diff
@@ -344,29 +344,53 @@ func (b *verifier) verifyAuthorshipRight(header *types.Header) error {
...
 		}
 
 		currentBlockProducerIndex, err := getAuthorityIndex(currentHeader)
 		if err != nil {
+			logger.Errorf("failed to get authority index %s", err)
 			continue
 		}
 
+		if len(currentHeader.Digest.Types) == 0 {
+			logger.Errorf("current header missing digest")
```
EclesioMeloJunior, line 361: I think we should jump to the next item otherwise this condition will let empty `currentHeader.Digest.Types` go further and panic the node when tried to `currentPreDigestItem := currentHeader.Digest.Types[0]`

EclesioMeloJunior (changes requested):

lib/babe/verify.go
kishansagathiya: I have started returning things, it makes more sense since it's a verification function.

EclesioMeloJunior (approved):

PR #2601, opened 2022-06-13
fix(lib/babe): check if authority index is in the `authorities` range

### LINKED ISSUE #2415: Casting of `uint32` to `int` Will Overflow In 32-bit OS (GSR-18)
The function `verifyPreRuntimeDigest()` which verifies a digest message received from the network. It ensure the authority ID received in the message is in the current authority list by checking the ID is less than the length.

An issue arises on 32-bit operating systems where `int` is a 32 bit signed integer. Casting a `uint32` which is larger than 2^31 will result in a negative `int` . A negative number will not trigger the conditional statement in `verifyPreRuntimeDigest()`. The function will continue to execute and later use `authIdx` as an index to the `b.authorities` list which will panic due to an Index Out Of Bounds error.

Consider modifying the conditional statement in `verifyPreRuntimeDigest()` to include negative numbers.

EclesioMeloJunior:
## Changes

- Check for negative numbers after casting a value to `int`

## Tests

N/A

## Issues

- Closes #2415

## Primary Reviewer

@jimjbrettj

### DIFF
--- a/lib/babe/verify.go
+++ b/lib/babe/verify.go
@@ -388,7 +388,7 @@ func (b *verifier) verifyPreRuntimeDigest(digest *types.PreRuntimeDigest) (scale
 		authIdx = d.AuthorityIndex
 	}
 
-	if uint64(len(b.authorities)) < uint64(authIndex) {
+	if uint64(len(b.authorities)) < uint64(authIdx) {
 		logger.Tracef("verifyPreRuntimeDigest invalid auth index %d, we have %d auths",
 			authIdx, len(b.authorities))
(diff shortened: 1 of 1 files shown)


timwu20 (changes requested):
`authIndex` is undefined

qdm12 (approved):

jimjbrettj (approved):

qdm12:
@timwu20 we need you to approve 😉 

Also regarding your comment, it might had been your IDE/language server going crazy? 🤔 

> `authIndex` is undefined

PR #2602, opened 2022-06-14
chore(trie): add `MerkleValue` method and function

qdm12:
## Changes

- Add (`internal/trie/node`) `MerkleValue` function taking in encoding + isRoot boolean and returning the "Merkle value", to match the polkadot spec closer. This is also needed in an upcoming proof generation/verification refactoring.
- Add `MerkleValue` method to `node.Node` to return the merkle value byte slice only, instead of using `_, hash, err := node.EncodeAndHash(isRoot)` in a few places, since it was quite confusing

## Tests

```sh
```

## Issues

Related to #2418 

## Primary Reviewer

@timwu20


EclesioMeloJunior (changes requested):

internal/trie/node/hash.go
```diff
@@ -5,51 +5,78 @@ package node
...
 
-	if !isRoot && buffer.Len() < 32 {
-		n.HashDigest = make([]byte, len(bufferBytes))
-		copy(n.HashDigest, bufferBytes)
-		hash = n.HashDigest // no need to copy
-		return encoding, hash, nil
+		bufferBytes := buffer.Bytes()
+
+		// TODO remove this copying since it defeats the purpose of `buffer`
+		// and the sync.Pool.
```
EclesioMeloJunior, line 69: I'm curious about how copying defeats the purpose of `buffer` and the sync.Pool?

qdm12: That's just copy paste from older code.

Don't forget `bytes.Buffer` are really just byte slices.
If we copy from the buffer to a byte slice, the GC will keep track of the copied byte slice.
So we might just use a single byte slice (or a single local buffer) without the sync.pool buffer, to save the copying operation.

We should ideally address this later for the sake of performance, but it's not a priority.

internal/trie/node/hash.go
```diff
@@ -5,51 +5,78 @@ package node
...
 	"bytes"
+	"fmt"
 
 	"github.com/ChainSafe/gossamer/internal/trie/pools"
 	"github.com/ChainSafe/gossamer/lib/common"
 )
 
-// EncodeAndHash returns the encoding of the node and
+// MerkleValue produces the Merkle value from the encoding of a node.
+// For root nodes, the Merkle value is always the Blak2b hash of the encoding.
```
qdm12, line 15:
```suggestion
// For root nodes, the Merkle value is always the Blake2b hash of the encoding.
```

EclesioMeloJunior (approved):

internal/trie/node/hash.go
```diff
@@ -5,51 +5,93 @@ package node
...
-// the blake2b hash digest of the encoding of the node.
-// If the encoding is less than 32 bytes, the hash returned
-// is the encoding and not the hash of the encoding.
-func (n *Node) EncodeAndHash(isRoot bool) (encoding, hash []byte, err error) {
-	if !n.Dirty && n.Encoding != nil && n.HashDigest != nil {
-		return n.Encoding, n.HashDigest, nil
+// MerkleValue produces the Merkle value from the encoding of a non-root node.
+// If the encoding is less or equal to 32 bytes, the Merkle value is the encoding.
+// Otherwise, the Merkle value is the Blake2b hash digest of the encoding.
+func MerkleValue(encoding []byte) (merkleValue []byte, err error) {
```
timwu20, line 17: any specific reason why we are exporting this?

qdm12: sorry I forgot to mention, it's needed in the `lib/trie/proof` code on #2604

qdm12: This now used in `lib/trie/proof` within this PR since #2604 got merged.

timwu20:
@qdm12 can you resolve the conflicts on this.

qdm12:
Conflicts resolved 😉

noot (approved):
basically looks fine, just one comment about removing duplicate code

internal/trie/node/hash.go
```diff
@@ -5,84 +5,176 @@ package node
...
+		}
+
+		bufferBytes := buffer.Bytes()
+
+		// TODO remove this copying since it defeats the purpose of `buffer`
+		// and the sync.Pool.
+		n.Encoding = make([]byte, len(bufferBytes))
+		copy(n.Encoding, bufferBytes)
+	}
 	encoding = n.Encoding // no need to copy
```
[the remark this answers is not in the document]
qdm12: Indeed! 👍 

Added `encodeIfNeeded() (encoding []byte, err error)` method for each.

By the way, there is a bit of code repetition between root and non root but deepsource complains about boolean control argument, so I thought I would satisfy it.

PR #2604, opened 2022-06-15
chore(trie): proof code refactoring and fixes

qdm12:
## Changes

- New package `lib/trie/proof` containing:
  - `lib/trie/proof.go` (refactored)
  - `internal/trie/record` files (then removed)
  - `lib/trie/database.go` (moved since it had nothing to do with database)
- Proof `Verify` function:
  - ⚠️ returns an error when root is not found in proof encoded nodes
  - ⚠️ fix `buildTrie` to detect root node for roots with encoding smaller than 32 bytes
  - ⚠️ Removes child with no matching hash from proof trie
  - ⚠️ Only decode root node, and then lazy decode nodes. **This is to allow for 'values' (not node encodings) to be together node encodings with v1 generated proofs**.
  - Only verify for a single key and value instead of pairs (no usage at all)
  - Produce ordered (from root to leaf) proof encoded nodes
  - Returns only an error and no boolean (was unneeded before)
- Proof `Generate` function:
  - ⚠️ returns an error when one key is not found in trie when generating proof
  - Remove `internal/trie/recorder` code (now unneeded)
  - Generate encoded nodes ordered from root to leaf (to reduce verify hash computations)
  - Do not add inlined nodes
- Add full coverage unit tests and remove older tests
- Minor changes
  - Better error wrapping
  - Database arguments uses a smaller locally defined `Database` interface in `proof` and `trie` packages
  - rename `proofEncodedNodes` to `encodedProofNodes`
  - rename `decProofs` to `encodedProofNodes`

## Tests

```sh
```

## Issues

One more step towards #2418 

## Primary Reviewer

@EclesioMeloJunior


lib/trie/proof/verify.go
```diff
@@ -0,0 +1,160 @@
...
+		// We compute the Merkle value of nodes treating them all
+		// as non-root nodes, meaning nodes with encoding smaller
+		// than 33 bytes will have their Merkle value set as their
+		// encoding. The Blake2b hash of the encoding is computed
+		// later if needed to compare with the root hash given to find
+		// which node is the root node.
+		const isRoot = false
+		decodedNode.HashDigest, err = node.MerkleValue(encodedProofNode, isRoot)
+		if err != nil {
+			return nil, fmt.Errorf("cannot calculate Merkle value of node at index %d: %w", i, err)
```
qdm12, line 82:
```suggestion
			return nil, fmt.Errorf("merkle value of node at index %d: %w", i, err)
```

lib/trie/proof/verify_test.go
```diff
@@ -0,0 +1,435 @@
...
+			rootHash:   []byte{1, 2, 3},
+			errWrapped: ErrEmptyProof,
+			errMessage: "cannot build trie from proof encoded nodes: " +
+				"proof slice empty: for Merkle root hash 0x010203",
+		},
+		"value not found": {
+			encodedProofNodes: [][]byte{
+				encodeNode(t, branch),
+				encodeNode(t, leafB),
+				// TODO Note leaf A is small enough to be inlined in branch
```
qdm12, line 66:
```suggestion
				// Note leaf A is small enough to be inlined in branch
```

lib/trie/proof/verify_test.go
```diff
@@ -0,0 +1,435 @@
...
+				}),
+			}),
+			expectedTrie: trie.NewTrie(&node.Node{
+				Key:         []byte{1},
+				Descendants: 4,
+				Children: padRightChildren([]*node.Node{
+					{
+						Key:   leafAShort.Key,
+						Value: leafAShort.Value,
+						Dirty: true, // TODO: this should be false
```
qdm12, line 243: TODO this should be set to false in the production code.

lib/trie/proof/generate_test.go
```diff
@@ -0,0 +1,380 @@
...
+			encodedProofNodes: [][]byte{
+				{
+					0b1100_0000 | 2,          // Node variant and partial key length
+					0x12,                     // partial key
+					0b0000_0001, 0b0000_0000, // children bitmap
+					0x4, 0x3, // scale encoded value
+					0x10,
+					0b0100_0000 | 1, // Node variant and partial key length
+					0x4, 0x4,        // partial key
+					0x5, // leaf value
```
qdm12, line 217: use `encodeNode(t, node)` for these test cases

lib/trie/proof/verify_test.go
qdm12: Changed to set it to true in all cases, since the trie isn't used outside of the proof code (nor the database), and also because to be consistent we would have to recursively check child from a decoded node and set their dirty flag to false. that's more, it's more confusing and hits performance a bit.

seunlanlege:
> Only verify for a single key and value instead of pairs (no usage at all)

Hi @qdm12 I'm currently using the trie lib in gossamer to implement a light client in golang here: [composableFi/ibc-go](https://github.com/composableFi/ibc-go)

We actually depend on this functionality and wondering if it'll be possible to keep it?

seunlanlege:
Also with these changes, would you say that the lib can be used to verify proofs from [paritytech/trie](https://github.com/paritytech/trie)

qdm12:
Hi @seunlanlege sorry for the delay, I was off the last few days.

> Only verify for a single key and value instead of pairs (no usage at all)
> We actually depend on this functionality and wondering if it'll be possible to keep it?

I have checked your codebase, you only check a single key-value pair in every of your calls so far, so it won't affect your codebase. The functions you use are still exported as well, although their signature and import path do change a bit, but Gossamer is still in its v0.x.x versions so this is expected.

> Also with these changes, would you say that the lib can be used to verify proofs from [paritytech/trie](https://github.com/paritytech/trie)

This is a work in progress to be able to verify proofs generated by the v1 trie state.
Together with #2530 and an additional PR adding headers (got a local branch with it already), it should allow to verify proofs at the very least. I'll notify you once it's ready 😉

lib/trie/database.go
```diff
@@ -76,74 +76,9 @@ func (t *Trie) store(db chaindb.Batch, n *Node) error {
...
-
-		branch.Children[i] = node
-		t.loadProof(proofHashToNode, node)
-	}
-}
-
 // Load reconstructs the trie from the database from the given root hash.
 // It is used when restarting the node to load the current state trie.
-func (t *Trie) Load(db chaindb.Database, rootHash common.Hash) error {
+func (t *Trie) Load(db Database, rootHash common.Hash) error {
```
timwu20, line 81: I would like to remove the `chaindb` dependency completely at some point.

timwu20 (approved):

lib/trie/database.go
qdm12: Agreed, although not a priority, but a not-too-complex task either.

Also we should use locally defined interfaces exported so that they get injected (accept interfaces), it should be up to the caller to define a (narrow as possible) interface for a concrete implementation. I guess one step closer now 😉

EclesioMeloJunior (approved):

PR #2606, opened 2022-06-15
fix(trie): descendants count for clear prefix

qdm12:
## Changes

Fix removing descendants for the ClearPrefix operation.
That was found with a sometimes flaky trie test in the CI.

## Tests

```sh
```

## Primary Reviewer

@kishansagathiya

### DIFF
--- a/lib/trie/trie.go
+++ b/lib/trie/trie.go
@@ -877,10 +877,11 @@ func (t *Trie) clearPrefix(parent *Node, prefix []byte) (
 			return parent, nodesRemoved
 		}
 
-		nodesRemoved = 1
+		nodesRemoved = 1 + child.Descendants
 		copySettings := node.DefaultCopySettings
 		branch = t.prepBranchForMutation(branch, copySettings)
 		branch.Children[childIndex] = nil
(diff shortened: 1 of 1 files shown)


EclesioMeloJunior (approved):

lib/trie/trie.go
```diff
@@ -892,7 +893,7 @@ func (t *Trie) clearPrefix(parent *Node, prefix []byte) (
 	noPrefixForNode := len(prefix) <= len(branch.Key) ||
 		lenCommonPrefix(branch.Key, prefix) < len(branch.Key)
 	if noPrefixForNode {
-		const nodesRemoved = 0
+		const nodesRemoved = 0 // XXX
```
timwu20, line 896:
```suggestion
		const nodesRemoved = 0 // XXX
```
`XXX`?

qdm12: Oopsi

For context, I used that to see test coverage and see where that bug was coming from, since some of the tests are fuzzing-like and generate random values.

kishansagathiya (approved):

PR #2608, opened 2022-06-15
feat(e2e): build Gossamer on any test run

### LINKED ISSUE #2458: end-to-end tests: build the binary in each test
## Task summary

For now the Go binary has to be built to `./bin/gossamer` (for which a `go build` command is in `make xxx` commands).
This works for CI, but it's rather unpractical for development.

If some production code is changed, and the developer is testing a single end to end test, he/she has to rebuild the binary manually pretty-much which amounts to quite some time wasted (and headaches when forgetting to re-build).

A solution is to run, for each test, the `go build -trimpath -o ./bin/gossamer ./cmd/gossamer` command from an `exec.CommandContext` call, before calling the built binary. The `go build` command will re-use the build cache so re-builds will be quite fast, making this change a good solution.

qdm12:
## Changes

This happened one too many times, us developers forgetting to re-build Gossamer before running an e2e test locally.

This changes e2e to build the binary every time a test is run. The first build takes 10s, but it only takes about 1.5s on rebuilds, which is well worth the avoided trouble of dev time wasted.

## Tests

End-to-end suites passing

## Issues

Fixes #2458 

## Primary Reviewer

@edwardmack

### DIFF
--- a/tests/polkadotjs_test/start_polkadotjs_test.go
+++ b/tests/polkadotjs_test/start_polkadotjs_test.go
@@ -25,9 +25,12 @@ func TestStartGossamerAndPolkadotAPI(t *testing.T) {
 		return
 	}
 
+	err := utils.BuildGossamer()
+	require.NoError(t, err)
+
 	const nodePackageManager = "npm"
 	t.Logf("Checking %s is available...", nodePackageManager)
-	_, err := exec.LookPath(nodePackageManager)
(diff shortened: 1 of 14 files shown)


edwardmack (approved):
This looks great, just address lint and copyright issues.  Nice job moving the MODE != "rpc" into TestMain, make more sense there instead of for every test.

EclesioMeloJunior (approved):

Issue #2404, opened 2022-03-18
Panic When Deleting Some Nonexistent Keys From Trie (GSR-10)


danforbes:
Under some circumstances, deleting a key from a trie triggers a slice bounds out of range runtime panic, causing the Gossamer node to crash.

In particular, deleting a nonexistent key kx will raise a panic when the trie contains valid key `k_y` for which `k_x` is a prefix i.e. where `bytes.HasPrefix(k_y, k_x)` holds.

While this cannot be exploited by network messages from unauthenticated nodes, it may be readily possible for the runtime to trigger this crash; in particular via the `ext_storage_clear and ext_default_child_storage_clear` Host API functions, which may be possible for unprivileged users to execute in some runtimes. Malicious block proposers may also be able to directly exploit this.

Once a block contains such an operation, it becomes impossible for Gossamer nodes to execute the state transition
without crashing — resulting in repeated crashes and possibly a chain split until the issue is patched.

Consider interfacing with the maintainers of the Host Specification and reference implementation — to clarify and explicitly specify whether a compliant runtime should ever call `ext_storage_clear` and `ext_default_child_storage_clear` with nonexistent keys and, if so, what should occur.

If, under those circumstances,` (*Trie).Delete()` should be effectively a no-op, properly ensure that it can handle nonexistent key arguments without panicking.

Solutions could involve:
• Checking that the value `length + 1` is within the bounds of the `key` slice i.e. that `length < len(key)`.
• Handling the specific case where the parent key `p.key` is longer than the `key` argument.

PR #2609, opened 2022-06-16
fix(trie): Panic when deleting nonexistent keys from trie (GSR-10)

edwardmack:
## Changes

- Implement check to handle condition when deleting a nonexistent key K_x when the trie contains valid key K_y for which K_x is a prefix.  i.e. where `bytes.HasPrefix(k_y, k_x)` holds will no longer cause panic

## Tests

```sh
 go test -run ^Test_Trie_delete$ github.com/ChainSafe/gossamer/lib/trie -v
```

## Issues
closes: #2404 

## Primary Reviewer

@timwu20

### DIFF
--- a/lib/trie/trie.go
+++ b/lib/trie/trie.go
@@ -969,7 +969,8 @@ func (t *Trie) deleteBranch(branch *Node, key []byte) (
 	}
 
 	commonPrefixLength := lenCommonPrefix(branch.Key, key)
-	if (commonPrefixLength + 1) > len(key) {
+	keyDoesNotExist := commonPrefixLength == len(key)
+	if keyDoesNotExist {
 		return branch, false, 0
 	}
 	childIndex := key[commonPrefixLength]
--- a/lib/trie/trie.go
+++ b/lib/trie/trie.go
@@ -969,6 +969,9 @@ func (t *Trie) deleteBranch(branch *Node, key []byte) (
 	}
 
 	commonPrefixLength := lenCommonPrefix(branch.Key, key)
+	if (commonPrefixLength + 1) > len(key) {
+		return branch, false, 0
+	}
 	childIndex := key[commonPrefixLength]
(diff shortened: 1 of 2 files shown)


lib/trie/trie.go
```diff
@@ -969,6 +969,9 @@ func (t *Trie) deleteBranch(branch *Node, key []byte) (
 	}
 
 	commonPrefixLength := lenCommonPrefix(branch.Key, key)
+	if (commonPrefixLength + 1) > len(key) {
+		return branch, false, 0
+	}
```
qdm12, line 974: `commonPrefixLength` is the common prefix length so it cannot be longer than the length of `key`. We can thus change the if condition to the simpler `commonPrefixLength == len(key)`.

I would suggest also assigning it to a `keyDoesNotExist` variable so that is servers as code documentation (or add a comment if you prefer), since it's not super obvious what that block does.

```suggestion
	keyDoesNotExist := commonPrefixLength == len(key)
	if keyDoesNotExist {
		return branch, false, 0
	}

```

lib/trie/trie_test.go
```diff
@@ -3573,6 +3573,42 @@ func Test_Trie_delete(t *testing.T) {
...
+						Value: []byte{1},
+					},
+					{ // full key 1, 1, 2
+						Key:   []byte{2},
+						Value: []byte{2},
+					},
+				}),
+			},
+			updated:      false,
+			nodesRemoved: 0,
```
qdm12, line 3610: nit we can remove those default values

```suggestion
```

qdm12 (approved):
The trie and its million pitfalls 😄 
One less! 👍

lib/trie/trie.go
edwardmack: Yes, 2 == 2 makes more sense than (2 + 1) > 2.  Updated.

EclesioMeloJunior (approved):

PR #2611, opened 2022-06-16
feat(trie): decode all inlined node variants

qdm12:
## Changes

Extend from previously limited-to-leaf inlined decoding to all node variants.

## Tests

```sh
```

## Issues

One step towards #2418 

## Primary Reviewer

@EclesioMeloJunior

### DIFF
--- a/internal/trie/node/decode_test.go
+++ b/internal/trie/node/decode_test.go
@@ -14,6 +14,10 @@ import (
 )
 
 func scaleEncodeBytes(t *testing.T, b ...byte) (encoded []byte) {
+	return scaleEncodeByteSlice(t, b)
+}
+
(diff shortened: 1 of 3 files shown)


internal/trie/node/decode_test.go
```diff
@@ -98,66 +102,6 @@ func Test_Decode(t *testing.T) {
 				Dirty:    true,
 			},
 		},
-		"branch with two inlined children": {
```
EclesioMeloJunior, line 101: it is not possible to have to inline children in the branch node?

qdm12: Oh yes, it's just that test case is no longer needed since it's covered by a new test case in another test.

timwu20 (approved):

EclesioMeloJunior (approved):

Issue #1819, opened 2021-09-30
feature: Add SCALE custom type functionality for Result, VaryingDataType, and VaryingDataTypeSlice 


timwu20:
## Task summary

- SCALE requires the destination data structure to be known before decoding.  This isn't a problem when we deal with initialised `Result`, `VaryingDataType` and `VaryingDataTypeSlice` instances.  
- The issue is when `Result`, `VaryingDataType` and `VaryingDataTypeSlice` are used as attributes in structs.  The zero case for the `Result`, `VaryingDataType` and `VaryingDataTypeSlice` have not been intialized with their associated value types, which causes a decode error.
- The proposed solution is to allow for shadowing (custom types) of `Result`, `VaryingDataType` and `VaryingDataTypeSlice`, and to utilise the constructor pattern (ex. `func NewMyResult() *NewResult`) which will be called by the scale decoding process to initialise these special types.  In the event that there is no constructor defined in the package we could throw a friendly error to notify the developer.

## Specification reference

- 

## Other information and links

- https://github.com/ChainSafe/gossamer/pull/1548

danforbes:
Is this a place where we could be able to make use of generics?

PR #2612, opened 2022-06-17
feat(pkg/scale): support for custom `VaryingDataType` types

timwu20:
## Changes

- Allow for shadowing (custom types) of `VaryingDataType`.  
- Example in the test code and readme
- Unit tests for custom `VaryingDataType`, and for nested custom `VaryingDataType`.

## Tests

```sh
go test ./pkg/scale
```

## Issues

- contributes to #1819 

## Primary Reviewer

@kishansagathiya


pkg/scale/varying_data_type_test.go
```diff
@@ -324,6 +326,48 @@ func Test_decodeState_decodeVaryingDataType(t *testing.T) {
...
+	}
+}
+func Test_decodeState_decodeCustomVaryingDataType(t *testing.T) {
+	for _, tt := range varyingDataTypeTests {
+		t.Run(tt.name, func(t *testing.T) {
+			vdt, err := NewVaryingDataType(VDTValue{}, VDTValue1{}, VDTValue2{}, VDTValue3(0))
+			if err != nil {
+				t.Errorf("%v", err)
+				return
+			}
```
qdm12, line 351: Maybe use testify test libs? 🤔 
Or are you trying to keep without dependency so it's movable outside of this repo with zero-dependency?

```suggestion
			require.NoError(t, err)
```

qdm12 (approved):
👍

EclesioMeloJunior (approved):

pkg/scale/varying_data_type_nested_example_test.go
```diff
@@ -0,0 +1,186 @@
...
+import (
+	"fmt"
+	"reflect"
+
+	"github.com/ChainSafe/gossamer/pkg/scale"
+)
+
+// ParentVDT is a VaryingDataType that consists of multiple nested VaryingDataType
+// instances (aka. a rust enum containing multiple enum options)
+type ParentVDT scale.VaryingDataType
```
jimjbrettj, line 15: Is index receiver not needed for the ParentVDT because you attempt to case to a VDT while decoding, and if you can then call custom method?

pkg/scale/varying_data_type_nested_example_test.go
```diff
@@ -0,0 +1,186 @@
...
+// Value will return value from underying VaryingDataType
+func (pvdt *ParentVDT) Value() (val scale.VaryingDataTypeValue) {
+	vdt := scale.VaryingDataType(*pvdt)
+	return vdt.Value()
+}
+
+// NewParentVDT is constructor for ParentVDT
+func NewParentVDT() ParentVDT {
+	// use standard VaryingDataType constructor to construct a VaryingDataType
+	vdt, err := scale.NewVaryingDataType(NewChildVDT(), NewOtherChildVDT())
```
jimjbrettj, line 39: Is it needed/is there a way to make the number of children here a param?

jimjbrettj (approved):
really cool nice stuff

kishansagathiya (commented):
It seems like for a custom `VaryingDataType`, we are supposed to implement methods `Set` and `Value`. Actually, I am some what confused on when to implement those methods, because I see that at some place only Set is implemented.
If we are supposed to implement those methods, may be worth adding in the read me.

pkg/scale/varying_data_type_nested_example_test.go
timwu20: index receiver isn't required for `ParentVDT` since this type isn't used as a `VaryingDataTypeValue` anywhere.

pkg/scale/varying_data_type_nested_example_test.go
timwu20: The types of children and number of child options is supplied via the `NewVaryingDataType` constructor.  This is supposed to resemble the declaration of the enum in Rust.

pkg/scale/decode.go
```diff
@@ -344,6 +349,19 @@ func (ds *decodeState) decodeVaryingDataTypeSlice(dstv reflect.Value) (err error
...
+	converted := dstv.Convert(reflect.TypeOf(VaryingDataType{}))
+	tempVal := reflect.New(converted.Type())
+	tempVal.Elem().Set(converted)
+	err = ds.decodeVaryingDataType(tempVal.Elem())
+	if err != nil {
+		return
+	}
+	dstv.Set(tempVal.Elem().Convert(initialType))
+	return
+}
```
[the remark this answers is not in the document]
timwu20: dstv is not modified.  `dstv.Convert` gives you a new instance that is converted but not addressable.  So a `tempVal` needs to be created which is then decoded to.  Then I set the original `dstv` with the decoded data converted back to the initial custom type.

timwu20:
> It seems like for a custom `VaryingDataType`, we are supposed to implement methods `Set` and `Value`. Actually, I am some what confused on when to implement those methods, because I see that at some place only Set is implemented. If we are supposed to implement those methods, may be worth adding in the read me.
> 
> I wonder if there is any value in having a vdt interface that has Set and Value method!

I agree it's quite cumbersome right now when using custom types in go.  It doesn't expose the `Set` and `Value` methods unless you embed the `VaryingDataType` into your own type.  Embedding would require more changes to the scale package to make it work.  If I were to make `VaryingDataType` an interface, you would still need to embed some sort of `DefaultVaryingDataType` implementation within your own struct, which differs from the primitive custom types . Let me know if that's what you prefer.  So instead of a custom `VaryingDataType` type, the scale package would be ensuring that the type implements a `VaryingDataType` interface where an implementation is supplied to you as `DefaultVaryingDataType` so the user doesn't have to implement their own logic to fulfil the interface.

Issue #2411, opened 2022-03-21
Storing Justification Allows Excess Bytes (GSR-13)


danforbes:
Justifications are sent as raw bytes over the network which are then decoded when they are verified in `VerifyBlockJustification()`.

They are decoded using SCALE decoding, which allows excess bytes to be appended to the end of the data being decoded.

After verification, the original bytes are stored in the block state in `handleJustification()`.

This poses a threat, as a malicious actor may append garbage bytes to the end of the justification. These bytes will then be stored in the block state. This is an attack vector for a user who may unnecessarily fill up disk space on a node with garbage bytes.

The Substrate implementation will decode then encode the justification and store the encoded bytes to ensure excess bytes are not included. Consider taking this approach or alternatively rejecting justifications with excess bytes appended as valid nodes will not include these.

PR #2618, opened 2022-06-22
fix(lib/grandpa): Storing Justification Allows Extra Bytes (GSR-13)

edwardmack:
## Changes
- Implement changes to fix condition where extra justification bytes were stored in block state.
- Create tests to validate fix cures this issue.

## Tests

```sh
go test -run ^TestService_VerifyBlockJustification$ github.com/ChainSafe/gossamer/lib/grandpa -v
```

## Issues
closes: #2411 

## Primary Reviewer

@EclesioMeloJunior


EclesioMeloJunior (changes requested):

dot/sync/syncer_integration_test.go
```diff
@@ -120,7 +120,10 @@ func newTestSyncer(t *testing.T) *Service {
 	cfg.LogLvl = log.Trace
 	mockFinalityGadget := NewMockFinalityGadget(ctrl)
 	mockFinalityGadget.EXPECT().VerifyBlockJustification(gomock.AssignableToTypeOf(common.Hash{}),
-		gomock.AssignableToTypeOf([]byte{})).AnyTimes()
+		gomock.AssignableToTypeOf([]byte{})).DoAndReturn(func(hash common.Hash, justification []byte) ([]byte, error) {
+		return justification, nil
+	}).AnyTimes()
```
jimjbrettj, line 125: Is there a way to know exactly how many times this runs?

lib/grandpa/grandpa.go
```diff
@@ -826,7 +826,6 @@ func (s *Service) finalise() error {
 	if err != nil {
 		return err
 	}
-
```
jimjbrettj, line 829: nit: remove this diff

lib/grandpa/message_handler.go
```diff
@@ -661,30 +662,30 @@ func (s *Service) VerifyBlockJustification(hash common.Hash, justification []byt
...
 	if err != nil {
-		return err
+		return nil, err
 	}
 
 	logger.Debugf(
 		"set finalised block with hash %s, round %d and set id %d",
 		hash, fj.Round, setID)
-	return nil
+	return scale.Marshal(fj)
```
jimjbrettj, line 688: A little confused, just for my own understanding where are the extra bytes trimmed here? Is it just by unmarshalling then marshalling?

lib/grandpa/message_handler_test.go
```diff
@@ -1055,3 +1063,152 @@ func signFakeFullVote(
 
 	return sig
 }
+
+func TestService_VerifyBlockJustification(t *testing.T) {
+	validJustification := "0x05000000000000008306f3c4dda0211e863e52c8bc15a3251e5ba38c7dd2f241f896653665e5e7e0080000" +
```
jimjbrettj, line 1068: Maybe better to generate this if possible? If not feasible within unit test, maybe comment where you got it?

dot/sync/syncer_integration_test.go
edwardmack: No, since this is in `newTestSyncer` which is used in multiple tests that have different conditions we don't know if this will be called or not.  Ideally the tests that use this would be refactored to put the mocks in-line for each individual test so that we would know exactly how many times it's called.

lib/grandpa/message_handler.go
edwardmack: Yes, the extra bytes are removed when the un-marshaled Justification is marshaled.

lib/grandpa/message_handler_test.go
edwardmack: Good point, I've refactored this to use `buildTestJustification` to generate this, it's clearer now what the test is testing.

timwu20 (approved):
great work Ed!

EclesioMeloJunior (approved):
LGTM!

PR #2619, opened 2022-06-23
(lib/grandpa): verify justification hash relates to block being justified 

### LINKED ISSUE #2414: `VerifyBlockJustification()` Does Not Verify Justification Hash Relates to Block Being Justified (GSR-17)
The function `VerifyBlockJustification(hash common.Hash, justification []byte)` verifies that a `justification` justifies a block with `hash`.

There is no check to ensure that `Justification.Hash` equals `hash` . As a result it is possible that this justification is for a future block.

Substrate performs this check to ensure the justification of for this specific block and thus will reject justifications that are accepted by Gossamer.

The impact is rated low as the verification will only succeed if there are sufficient votes for a descendant of `hash` and thus a future block is already finalised.

Consider including a check that ensures the justification is for this specific block. That is, ensure `Justification.Hash` equals `hash`. Alternatively, it's possible that the specification could be updated to allow justifications for descendants to be used.

EclesioMeloJunior:
## Changes
- Check if the given `justification commit hash` is equal the given header hash in the method `VerifyBlockJustification(hash common.Hash, justification []byte)`

## Tests

```sh
go test github.com/ChainSafe/gossamer/lib/grandpa
```

## Issues

- Closes #2414

## Primary Reviewer

@timwu20


lib/common/hash.go
```diff
@@ -64,6 +64,12 @@ func (h Hash) String() string {
 	return fmt.Sprintf("0x%x", h[:])
 }
 
+// Short returns the first 3 bytes and the last 3 bytes of the hex string for the hash
+func (h Hash) Short() string {
+	const nBytes = 3
```
qdm12, line 69: 👍 Might be useful!
Although just in case I'd just have 4+4 characters, since 16^6 = 16,777,216‬ and 16^8 = 4,294,967,296 which is quite more collision resistant.

```suggestion
// Short returns the first 4 bytes and the last 4 bytes of the hex string for the hash
func (h Hash) Short() string {
	const nBytes = 4
```

lib/grandpa/errors.go
```diff
@@ -61,7 +61,8 @@ var (
 	ErrNotCommitMessage = errors.New("cannot get finalised hash from VoteMessage")
 
 	// ErrNoJustification is returned when no justification can be found for a block, ie. it has not been finalised
-	ErrNoJustification = errors.New("no justification found for block")
+	ErrNoJustification       = errors.New("no justification found for block")
+	ErrJustificationMismatch = errors.New("justication has does not correspond to given block hash")
```
qdm12, line 65:
```suggestion
	ErrJustificationMismatch = errors.New("justification does not correspond to given block hash")
```

lib/grandpa/message_handler.go
```diff
@@ -561,6 +561,10 @@ func (s *Service) VerifyBlockJustification(hash common.Hash, justification []byt
 		return err
 	}
 
+	if !hash.Equal(fj.Commit.Hash) {
+		return fmt.Errorf("%w: %s not equal %s", ErrJustificationMismatch, fj.Commit.Hash.Short(), hash.Short())
```
qdm12, line 565: mention which hash you expected and which was the received hash, since it's hard to know here which hash is which

qdm12 (approved):

edwardmack (approved):

lib/grandpa/message_handler.go
```diff
@@ -561,6 +561,10 @@ func (s *Service) VerifyBlockJustification(hash common.Hash, justification []byt
 		return err
 	}
 
+	if !hash.Equal(fj.Commit.Hash) {
+		return fmt.Errorf("%w: expected %s. got %s", ErrJustificationMismatch, fj.Commit.Hash.Short(), hash.Short())
```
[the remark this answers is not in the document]
EclesioMeloJunior: This will generate too much noise in the error message:
`justification does not correspond to given block hash: justification commit hash is %s and does not match hash %s`

Another way could be:
`justification does not correspond to given block hash: justification %s. block hash %s`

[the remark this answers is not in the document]
qdm12: Yeah sounds good 👍 just that full stop `.` is weird to me, maybe just replace it with `and` 😉

lib/grandpa/message_handler_test.go
```diff
@@ -789,6 +790,19 @@ func TestMessageHandler_VerifyBlockJustification_invalid(t *testing.T) {
...
+	just = newJustification(round+1, testHash, number, precommits)
+	data, err = scale.Marshal(*just)
+	require.NoError(t, err)
+	otherHeader := types.NewEmptyHeader()
+	_, err = gs.VerifyBlockJustification(otherHeader.Hash(), data)
+	require.ErrorIs(t, err, ErrJustificationMismatch)
+
+	expectedErr := fmt.Sprintf("%s: justification %s and block hash %s", ErrJustificationMismatch,
+		testHash.Short(), otherHeader.Hash().Short())
+	require.EqualError(t, err, expectedErr)
```
qdm12, line 805: nit you could assert the wrapped error:

```suggestion
	expectedErr := fmt.Sprintf("%s: justification %s and block hash %s", ErrJustificationMismatch,
		testHash.Short(), otherHeader.Hash().Short())
	assert.ErrorIs(t, err, ErrJustificationMismatch)
	require.EqualError(t, err, expectedErr)
```

Issue #2410, opened 2022-03-21
Duplicate Votes in GRANDPA are Counted as Equivocatory Votes (GSR-11)


danforbes:
The function `getEquivocatoryVoters()` is used to determine if a voter has equivocated by voting twice in a single round of GRANDPA.

It does not account for the case where a vote appears twice in the AuthData list. As a result, an authority may be incorrectly judged as an equivocator when an attacker includes the vote twice.

Duplicates are currently ignored in [the Substrate implementation](https://github.com/paritytech/finality-grandpa/blob/v0.15.0/src/lib.rs#L438) and should be treated the same in Gossamer to avoid a consensus split.

However, it seems beneficial to both protocols to reject commits which have malicious or wasteful data to reduce database size and network bandwidth. Consider updating the specifications to reject commits which include invalid or duplicate votes.

PR #2624, opened 2022-06-28
fix(lib/grandpa): Duplicate votes is GRANDPA are counted as equivocatory votes (GSR-11)

edwardmack:
## Changes
- Implement fix to `getEquivocatoryVoters` that does not include duplicate votes as equivocatory votes, addressing Sigma Prime - Gossamer Audit issue 11 (GSR-11)
- Add unit tests for testing duplicate votes condition.

## Tests

```sh
go test -run ^Test_getEquivocatoryVoters$ github.com/ChainSafe/gossamer/lib/grandpa -v
```

## Issues
closes: #2410 

## Primary Reviewer

@kishansagathiya

### DIFF
--- a/lib/grandpa/message_handler.go
+++ b/lib/grandpa/message_handler.go
@@ -284,22 +284,21 @@ func (h *MessageHandler) verifyCatchUpResponseCompletability(prevote, precommit 
 	return nil
 }
 
-func getEquivocatoryVoters(votes []AuthData) (map[ed25519.PublicKeyBytes]struct{}, error) {
+func getEquivocatoryVoters(votes []AuthData) map[ed25519.PublicKeyBytes]struct{} {
 	eqvVoters := make(map[ed25519.PublicKeyBytes]struct{})
-	voters := make(map[ed25519.PublicKeyBytes]int, len(votes))
+	voters := make(map[ed25519.PublicKeyBytes][64]byte, len(votes))
 
 	for _, v := range votes {
-		voters[v.AuthorityID]++
-		switch voters[v.AuthorityID] {
-		case 1:
-		case 2:
-			eqvVoters[v.AuthorityID] = struct{}{}
-		default:
-			return nil, fmt.Errorf("%w: authority id %x has %d votes",
-				errInvalidMultiplicity, v.AuthorityID, voters[v.AuthorityID])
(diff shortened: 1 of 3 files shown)


lib/grandpa/message_handler.go
```diff
@@ -284,22 +284,21 @@ func (h *MessageHandler) verifyCatchUpResponseCompletability(prevote, precommit
...
+func getEquivocatoryVoters(votes []AuthData) map[ed25519.PublicKeyBytes]struct{} {
 	eqvVoters := make(map[ed25519.PublicKeyBytes]struct{})
-	voters := make(map[ed25519.PublicKeyBytes]int, len(votes))
+	voters := make(map[ed25519.PublicKeyBytes][64]byte, len(votes))
 
 	for _, v := range votes {
-		voters[v.AuthorityID]++
-		switch voters[v.AuthorityID] {
-		case 1:
-		case 2:
```
kishansagathiya, line 295: Does not this already take cared of the case when a vote appears twice in the AuthData list?

lib/grandpa/message_handler.go
```diff
@@ -284,22 +284,21 @@ func (h *MessageHandler) verifyCatchUpResponseCompletability(prevote, precommit
...
-			eqvVoters[v.AuthorityID] = struct{}{}
-		default:
-			return nil, fmt.Errorf("%w: authority id %x has %d votes",
-				errInvalidMultiplicity, v.AuthorityID, voters[v.AuthorityID])
+		signature, present := voters[v.AuthorityID]
+		if present {
+			if !bytes.Equal(signature[:], v.Signature[:]) {
```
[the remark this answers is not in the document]
edwardmack: If there is a vote present for the authority, and the signature of that vote doesn't equal the votes signature then we add that voter to the equivocatory voters map.  The voter is equivocatory since there are votes for that voter that do not match.

lib/grandpa/message_handler.go
edwardmack: The audit issue and spec say that duplicate votes are not equivocatory, just when votes don't match they are then equivocatory.

EclesioMeloJunior (changes requested):

PR #2628, opened 2022-06-29
Make sure message that we receive are below their threshold sizes 

### LINKED ISSUE #2399: DOS via Decoding of Malicious Messages (GSR-06)
A malicious node can readily cause all connected Gossamer peers to crash by broadcasting maliciously crafted messages, without requiring significant expenditure of resources.

SCALE messages from untrusted sources are decoded without appropriately validating length values, restricting them to appropriate maximums. Large length values result in an attempt to allocate excessive memory, triggering an unrecoverable “out of memory” (OOM) crash via an unrecoverable [`runtime.throw`](https://cs.opensource.google/go/go/+/refs/tags/go1.17.7:src/runtime/panic.go;l=1188;drc=refs%2Ftags%2Fgo1.17.7;bpv=0;bpt=1).

As malformed or malicious messages from unauthenticated sources can cause all receiving Gossamer nodes to crash, this has been deemed of critical severity. Such an attack requires little expertise (sending random data is likely to trigger this) and would have a significant impact on the health of any network running a suitable proportion of Gossamer nodes.

Validate length values within all SCALE-encoded objects received from external sources.

The testing team recommend that the Polkadot Host specification be extended to define appropriate maximum length limits for all fields of variable length transmitted via the network. Compliant implementations should enforce these limits.

Additionally, SCALE doesn’t involve compression so decoded message content (without length info) should be shorter than the encoded message (containing length info). As such, it should be impossible for any valid SCALE-encoded data to contain a length value larger than total length of the data. This could be helpful for validating length values.

If validation is expensive, it could be reasonable to avoid length validation when decoding trusted, internal data (e.g. from the database). If so, it is crucial that all data received from untrusted sources are clearly separated from trusted data. One solution could involve enforcing this via the type system e.g. a `BlockResponseMessage` contains a sequence of `types.BlockData`, but the BlockDB stores data of type `TrustedBlockData` , which is decoded using `scale.UnsafeUnmarshal()`, with unchecked lengths.

kishansagathiya:
## Tests

```sh
```

## Issues
Fixes #2399

## Primary Reviewer

@timwu20


pkg/scale/decode.go
```diff
@@ -541,6 +541,7 @@ func (ds *decodeState) decodeBytes(dstv reflect.Value) (err error) {
 		return
 	}
 
+	// check the length
```
kishansagathiya, line 544:
```suggestion
```

dot/network/utils.go
```diff
@@ -20,6 +20,16 @@ import (
...
+var (
+	// maxBlockRequestSize              uint64 = 1024 * 1024      // 1mb
+	maxBlockResponseSize uint64 = 1024 * 1024 * 16 // 16mb
+	// MaxGrandpaNotificationSize is maximum size for a grandpa notification message.
+	MaxGrandpaNotificationSize       uint64 = 1024 * 1024      // 1mb
+	maxTransactionsNotificationSize  uint64 = 1024 * 1024 * 16 // 16mb
+	maxBlockAnnounceNotificationSize uint64 = 1024 * 1024      // 1mb
+
+)
+
```
EclesioMeloJunior, line 32: could you change them to `const`?

EclesioMeloJunior (changes requested):

EclesioMeloJunior (approved):

timwu20 (approved):

Issue #2416, opened 2022-03-21
Unrestricted Loop When Building Blocks (GSR-19)


danforbes:
There is a loop when building BABE blocks that is unrestricted. While the slot has not finished, the builder will continue to poll the `transactionState` for new transactions. This is done in a loop without sleeps in between. The result is an excessive use of resources and constant grabbing and releasing of the transaction queue lock.

One solution is to add a small sleep when `txn == nil` to avoid continued, excessive polling of the `transactionState`.

PR #2632, opened 2022-06-29
fix(lib/babe): Unrestricted Loop When Building Blocks (GSR-19)

edwardmack:
## Changes
- Add function `PopChannel` which returns a channel that notifies when data is push to the stack, this channel is used to determine when the transactionState stack is not empty.
- Add to unit tests

## Tests

```sh
go test -run ^TestBlockBuilder_buildBlockExtrinsics$ github.com/ChainSafe/gossamer/lib/babe -v
```

## Issues
closes: #2416 

## Primary Reviewer

@timwu20


lib/transaction/priority_queue.go
```diff
@@ -134,11 +135,43 @@ func (spq *PriorityQueue) Push(txn *ValidTransaction) (common.Hash, error) {
...
 
 	transactionQueueGauge.Set(float64(spq.pq.Len()))
 	return hash, nil
 }
 
+// NewPushWatcher returns a read only channel to be signalled
+// when the next Push() is called on the queue.
+// Note the returned channel is closed when the Push() function
+// is called.
+func (spq *PriorityQueue) NewPushWatcher() (nextPushWatcher <-chan struct{}) {
```
qdm12, line 156: nit `NextPushWatcher` would fit better, since we are not creating a watcher every time

lib/transaction/priority_queue.go
```diff
@@ -134,11 +135,43 @@ func (spq *PriorityQueue) Push(txn *ValidTransaction) (common.Hash, error) {
...
+	spq.Lock()
+	defer spq.Unlock()
+
+	var ok bool
+
+	select {
+	case _, ok = <-spq.nextPushWatcher:
+
```
qdm12, line 164:
```suggestion
```

edwardmack:
NOTE: there seems to be an issue after the determineErr check in `lib/babe/build.go` lines 250-254 are not reachable because this checks for errors that are type `TransactionValidityError` however the check at 239 continues the loop for any error that is not type `DispatchOutcomeError` thus making it impossible for `TransactionValidityError` types to make it past 239.  

I didn't fix this (yet) in the PR because it's outside the scope if this PR, and becasue I don't fully understand the requirement for the checks with the results from `deterrminErr`.

Please advise.

lib/transaction/priority_queue.go
```diff
@@ -134,11 +135,36 @@ func (spq *PriorityQueue) Push(txn *ValidTransaction) (common.Hash, error) {
 	spq.currOrder++
 	heap.Push(&spq.pq, item)
 	spq.txs[hash] = item
+	if spq.nextPushWatcher != nil {
+		close(spq.nextPushWatcher)
+	}
```
EclesioMeloJunior, line 140: This code can cause a panic if `close(spq.nextPushWatcher)` is called more than once, I notice that you only assign a value to `spq.nextPushWatcher` when the method  `NextPushWatcher` is called. My question is: Let's say we call `NextPushWatcher` and then the `Push` method is called closing the channel how do we ensure that when the `Push` method is called a second time the `close(spq.nextPushWatcher)` not happen?

EclesioMeloJunior: Sorry, miss clicking on the approval 😞

EclesioMeloJunior (changes requested):

lib/transaction/priority_queue.go
qdm12: Indeed, I guess we could check if the channel is closed

```go
var closed bool
select {
case _, closed = <- spq.nextPushWatcher:
default:
}

if !closed {
  close(spq.nextPushWatcher)
}
```

Or instead, perhaps cleaner, change it to create a new signal channel for each call to `NextPushWatcher`, add each new channel to a slice in the queue struct, and close each in a for loop in the `Push` method (and clear the slice of channels). That would work I think right? 🤔

lib/babe/mocks/block_import_handler.go
```diff
@@ -1,4 +1,4 @@
-// Code generated by mockery v2.10.6. DO NOT EDIT.
+// Code generated by mockery v2.14.0. DO NOT EDIT.
```
qdm12, line 1: You should probably `git rebase development` since those mocks were bumped to 2.14.0 last week on development

lib/babe/build.go
```diff
@@ -189,11 +190,35 @@ func (b *BlockBuilder) buildBlockSeal(header *types.Header) (*types.SealDigest,
...
+	timeout := time.Until(slotEnd)
+	slotTimer := time.NewTimer(timeout)
+
+	for {
+		select {
+		case <-b.testForceReturn:
+			if !slotTimer.Stop() {
+				<-slotTimer.C
+			}
+			return included
```
qdm12, line 203: You can remove that test channel + select case, but we need an integration test 😉

edwardmack: Removed test channel and added integration test TestBuildBlock_addToTransactionState

EclesioMeloJunior (approved):
LGTM

lib/babe/build.go
```diff
@@ -189,11 +189,30 @@ func (b *BlockBuilder) buildBlockSeal(header *types.Header) (*types.SealDigest,
 func (b *BlockBuilder) buildBlockExtrinsics(slot Slot, rt runtime.Instance) []*transaction.ValidTransaction {
 	var included []*transaction.ValidTransaction
 
-	for !hasSlotEnded(slot) {
+	slotEnd := slot.start.Add(slot.duration * 2 / 3) // reserve last 1/3 of slot for block finalisation
+	timeout := time.Until(slotEnd)
+	slotTimer := time.NewTimer(timeout)
```
qdm12, line 194: @EclesioMeloJunior PTSD right there 🤣

lib/babe/build_integration_test.go
```diff
@@ -139,6 +139,49 @@ func TestBuildBlock_ok(t *testing.T) {
...
+	parentHash := babeService.blockState.GenesisHash()
+	rt, err := babeService.blockState.GetRuntime(nil)
+	require.NoError(t, err)
+
+	epochData, err := babeService.initiateEpoch(testEpochIndex)
+	require.NoError(t, err)
+
+	ext := runtime.NewTestExtrinsic(t, rt, parentHash, parentHash, 0, "System.remark", []byte{0xab, 0xcd})
+	block := createTestBlock(t, babeService, emptyHeader, [][]byte{common.MustHexToBytes(ext)},
+		1, testEpochIndex, epochData)
```
qdm12, line 163: `block` is never used

Also I'm not a genius, my linter is, you might want to bump your linter up 😉 

```
go install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.47
```

lib/babe/build_integration_test.go
```diff
@@ -139,6 +139,49 @@ func TestBuildBlock_ok(t *testing.T) {
 	require.Equal(t, 1, len(extsBytes))
 }
 
+func TestBuildBlock_addToTransactionState(t *testing.T) {
```
qdm12, line 142: Can you add to this test a push before the slot timer expires?

Right now the block

```go
			case <-pushWatcher:
				txn = b.transactionState.Pop()
				// check in case another goroutine popped the
				// transaction before the Pop() call above.
				retry = txn == nil
```

is not covered.

Ideally I'd say the following pattern would be nice to integrate test this:

1. Get item from the push watcher
1. Get item from queue directly
1. Get item from the push watcher
1. Slot timer expires so we exit (either at top of loop or within the push watcher select block)

qdm12 (changes requested):

lib/babe/build_integration_test.go
edwardmack: Created test `TestBuildBlock_buildBlockExtrinsics` which now covers above block.

lib/babe/build.go
```diff
@@ -189,11 +189,30 @@ func (b *BlockBuilder) buildBlockSeal(header *types.Header) (*types.SealDigest,
...
+		case <-slotTimer.C:
+			return included
+		default:
+		}
+
 		txn := b.transactionState.Pop()
-		// Transaction queue is empty.
-		if txn == nil {
-			continue
+		retry := txn == nil // Transaction queue is empty.
```
timwu20, line 204: I think this code would be more straight forward if we had some sort of `TransactionState.PopChannel() (tx *transaction.ValidTransaction, cancel func() (err error)` method.  That way you can move the retry logic within the `TransactionState`. The calling code would look a lot simpler:
```
	for {
                popChan, cancel := b.transactionState.PopChannel()
		select {
		case <-slotTimer.C
		        // ensure that popChan is closed
		        var wg sync.WaitGroup
		        wg.Add(1)
		        go func(){ 
		            <-popChan
                             wg.Done()
		         }()
		        cancel()
		        wg.Wait()
			return included
		case txn := <-popChan:
			// handle txn
			extrinsic := txn.Extrinsic
			...
		}
        }
```

timwu20 (changes requested):

lib/babe/build.go
qdm12: Good point, we could just return a channel from the Pop method.
Although we should probably just have one Pop method (and not one returning nil, one returning a channel) to avoid any data race / reduce complexity.

a few nits on that:

- the method is `TransactionState.PopChannel() (tx <-chan *transaction.ValidTransaction, cancel func())` I think
- Use `done := make(chan struct{})`, not empty interface which does occupy a few bytes vs struct{}

```go
		done := make(chan interface{})
		go func(){ 
			<-popChan
			close(done)
		}()
		cancel()
		<-done
```

can just be

```go
		cancel()
		<-popChan
```

lib/babe/build.go
```diff
@@ -189,52 +189,78 @@ func (b *BlockBuilder) buildBlockSeal(header *types.Header) (*types.SealDigest,
...
+			done := make(chan interface{})
+			go func() {
+				<-popChan
+				close(done)
+			}()
+			err := cancel()
+			if err != nil {
+				logger.Errorf("cannot cancel pop channel: %s", err)
+			}
+			<-done
```
qdm12, line 217: same thing afaik

```suggestion
			err := cancel()
			if err != nil {
				logger.Errorf("cannot cancel pop channel: %s", err)
			}
			<-popChan
```

lib/babe/build.go
```diff
@@ -189,52 +189,78 @@ func (b *BlockBuilder) buildBlockSeal(header *types.Header) (*types.SealDigest,
...
+		select {
+		case <-slotTimer.C:
+			return included
+		default:
 		}
 
-		extrinsic := txn.Extrinsic
-		logger.Tracef("build block, applying extrinsic %s", extrinsic)
+		popChan, cancel := b.transactionState.PopChannel()
+		defer func() {
```
qdm12, line 204: we probably should not rely on `defer` since this would stack N defers for N iterations of that for loop.

edwardmack: Addressed by refactoring how popChan is created and handled.

dot/state/transaction.go
```diff
@@ -47,6 +48,12 @@ func (s *TransactionState) Pop() *transaction.ValidTransaction {
 	return s.queue.Pop()
 }
 
+// PopChannel returns a *transaction.ValidTransaction channel to be signalled
+// when the next Push() is called on the queue.
+func (s *TransactionState) PopChannel(timer *time.Timer) (tx chan *transaction.ValidTransaction) {
```
qdm12, line 53: Restrict it to be read-only?

```suggestion
func (s *TransactionState) PopChannel(timer *time.Timer) (tx <-chan *transaction.ValidTransaction) {
```

lib/babe/build.go
```diff
@@ -189,13 +189,13 @@ func (b *BlockBuilder) buildBlockSeal(header *types.Header) (*types.SealDigest,
...
-		if txn == nil {
-			continue
-		}
+	slotEnd := slot.start.Add(slot.duration * 2 / 3) // reserve last 1/3 of slot for block finalisation
+	timeout := time.Until(slotEnd)
+	slotTimer := time.NewTimer(timeout)
+	popChan := b.transactionState.PopChannel(slotTimer)
 
+	for txn := range popChan {
+		// handle txn
```
qdm12, line 198: nit unneeded comment I would say 😉 

```suggestion
```

lib/transaction/priority_queue.go
```diff
@@ -80,14 +81,16 @@ type PriorityQueue struct {
...
 }
 
 // NewPriorityQueue creates new instance of PriorityQueue
 func NewPriorityQueue() *PriorityQueue {
 	spq := &PriorityQueue{
-		pq:  make(priorityQueue, 0),
-		txs: make(map[common.Hash]*Item),
+		pq:        make(priorityQueue, 0),
+		txs:       make(map[common.Hash]*Item),
+		sleepTime: time.Millisecond * 10,
```
qdm12, line 93: supernit - unit after factor

```suggestion
		sleepTime: 10 * time.Millisecond,
```

qdm12 (changes requested):

lib/babe/build.go
edwardmack: I've refactored this to create function `TransactionState.PopChannel(*time.Timer) (tx chan *transaction.ValidTransaction)` which take a time that will cancel the channel when the timer expires.  This seems like a good solution that eliminates the need for a cancel function.

lib/transaction/priority_queue.go
```diff
@@ -139,6 +142,30 @@ func (spq *PriorityQueue) Push(txn *ValidTransaction) (common.Hash, error) {
...
+			select {
+			case <-timer.C:
+				close(popChannel)
+			default:
+				txn := spq.Pop()
+
+				if txn != nil {
+					popChannel <- txn
+				} else {
+					time.Sleep(spq.sleepTime)
```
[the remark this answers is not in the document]
edwardmack: I think the point of the PR is to solve the unrestricted loop, the audit suggested adding a sleep into the loop to solve this.  I created PopChannel function on PriorityQueue so that notification and time out logic is contained within that function.  The sleep is used when the queue is empty to let that thread rest.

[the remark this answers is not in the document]
qdm12: The solution proposed in the audit has the problem that what value should be the poll period? What if it's smaller/larger than the slot time? @timwu20 any thoughts on that?

Also if we still want to do time based, you can simplify all this to be something like https://github.com/ChainSafe/gossamer/pull/2829 and not have this extra PopChannel method and not expose channels at all.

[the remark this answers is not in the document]
timwu20: > The solution proposed in the audit has the problem that we will eventually miss some extrinsics whilst we sleep the last time before the slot ends.

@qdm12 if you think this is a problem then state this explicitly.  Don't conflate this with what the auditors have recommended. 

I paired with @edwardmack on this, and designed the PopChannel interface to abstract away the idea of reading all `ValidTransaction` instances from the priority queue over a given time.  

> The solution proposed in the audit has the problem that we will eventually miss some extrinsics whilst we sleep the last time before the slot ends. This is true even if the sleep is very short. Time based is just not the right approach imo.

The existing code reserved 1/3 of the time for finalisation.  "missed" extrinsics during this duration of time will be added to the priority queue and included in the next block.  This is the already existing behaviour on the base branch.  So what approach would be the right approach in your opinion?

timwu20 (approved):

lib/transaction/priority_queue.go
```diff
@@ -80,14 +81,16 @@ type PriorityQueue struct {
 	pq        priorityQueue
 	currOrder uint64
 	txs       map[common.Hash]*Item
+	sleepTime time.Duration
```
timwu20, line 84:
```suggestion
	pollInterval time.Duration
```

timwu20 (changes requested):
After thinking about it some more, depending on the timeout provided to the sleep, the `PopChannel` may not close at an expected time.  I ended up changing the code a little in [this commit](https://github.com/ChainSafe/gossamer/commit/a5f3b865be09b00bee8c11b39725dde81e3cc86e) and added a test to ensure the channel closes within an expected time range.

edwardmack:
> After thinking about it some more, depending on the timeout provided to the sleep, the `PopChannel` may not close at an expected time. I ended up changing the code a little in [this commit](https://github.com/ChainSafe/gossamer/commit/a5f3b865be09b00bee8c11b39725dde81e3cc86e) and added a test to ensure the channel closes within an expected time range.

Good point, I've updated code to implement suggested changes.

timwu20 (approved):

lib/transaction/priority_queue.go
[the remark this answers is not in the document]
qdm12: We had an implementation fully event driven (no timer to retry polling) from commit https://github.com/ChainSafe/gossamer/pull/2632/commits/7ae3a2185c9a12bd30ef7ae97e3512f3fe96b3d6

It got dropped later (not sure which commit) in favor of the current periodic polling. Anyway if we want to keep the periodic polling, why not just do what the first commit was doing: https://github.com/ChainSafe/gossamer/pull/2632/commits/b6ecc25bc8cb2932d9a268064eb4a2391ffe8518 🤔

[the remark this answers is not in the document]
timwu20: > We had an implementation fully event driven (no timer to retry polling) from commit https://github.com/ChainSafe/gossamer/commit/7ae3a2185c9a12bd30ef7ae97e3512f3fe96b3d6

Again I had paired with @edwardmack and designed the `PopChannel` idea together.  We wanted to simplify the calling code to not have to worry about timers, or how/when to pop form the queue or create some sort of `pushWatcher`, which can all be contained within the bounded context of the priority queue.

I'm not sure how beneficial it is to rehash old commits, but my feedback around that commit specifically was that the handling of `pushWatchers` is quite cumbersome.  Additionally, there's a chance the priority queue could panic as a result of trying to close a `nil` channel or an already closed channel.

We could've taken the `pushWatcher` component and implemented it into the `popChannel`, but it seemed like overkill, since the calling code is now interfacing with a separate producing goroutine, which would sleep on its own.  And it can still be added as an enhancement.

> It got dropped later (not sure which commit) in favor of the current periodic polling. Anyway if we want to keep the periodic polling, why not just do what the first commit was doing: https://github.com/ChainSafe/gossamer/commit/b6ecc25bc8cb2932d9a268064eb4a2391ffe8518 🤔

Again, we wanted to simplify the calling code to be able to just get all the messages for a supplied amount of time.  If we just added the sleep, there's a chance that that we would exit late if the priority queue was empty which initiated a sleep, at the same time the timer had elapsed.

lib/babe/build.go
```diff
@@ -174,13 +174,18 @@ func (b *BlockBuilder) buildBlockSeal(header *types.Header) (*types.SealDigest,
...
+	slotTimer := time.NewTimer(timeout)
+
+	for {
+		txn := b.transactionState.PopWithTimer(slotTimer)
+		slotTimerExpired := txn == nil
+		if slotTimerExpired {
+			break
 		}
 
+		// handle txn
```
qdm12, line 188: nit

```suggestion
```

lib/babe/build.go
```diff
@@ -218,7 +223,6 @@ func (b *BlockBuilder) buildBlockExtrinsics(slot Slot, rt runtime.Instance) []*t
 		logger.Debugf("build block applied extrinsic %s", extrinsic)
 		included = append(included, txn)
 	}
-
```
qdm12, line 221: nit unneeded delta removal of newline

lib/transaction/priority_queue.go
```diff
@@ -77,17 +78,19 @@ func (pq *priorityQueue) Pop() interface{} {
...
+	pollInterval time.Duration
 	sync.Mutex
 }
 
 // NewPriorityQueue creates new instance of PriorityQueue
 func NewPriorityQueue() *PriorityQueue {
 	spq := &PriorityQueue{
-		pq:  make(priorityQueue, 0),
-		txs: make(map[common.Hash]*Item),
+		pq:           make(priorityQueue, 0),
```
qdm12, line 91: nit / out of scope, but since it's a delta, that `make` is really useless

```suggestion
```

lib/transaction/priority_queue.go
```diff
@@ -139,6 +142,35 @@ func (spq *PriorityQueue) Push(txn *ValidTransaction) (common.Hash, error) {
 	return hash, nil
 }
 
+// PopWithTimer returns the next valid transaction from the queue.
+// When the timer expires, it returns `nil`.
+func (spq *PriorityQueue) PopWithTimer(timer *time.Timer) (transaction *ValidTransaction) {
```
qdm12, line 147: :warning: Check if the queue has a transaction first to avoid spinning an extra goroutine AND waiting spq.pollInterval even a transaction is already present in the queue.

```suggestion
func (spq *PriorityQueue) PopWithTimer(timer *time.Timer) (transaction *ValidTransaction) {
	transaction = spq.Pop()
	if transaction != nil {
		return transaction
	}

```

lib/transaction/priority_queue_test.go
```diff
@@ -270,3 +272,100 @@ func TestRemoveExtrinsic(t *testing.T) {
...
+		assert.Equal(t, tests[expected[counter]], txn)
+		counter++
+	}
+}
+
+func Test_PopWithTimer_Ends(t *testing.T) {
+	pq := NewPriorityQueue()
+	// increase sleep time greater than timer
+	pq.pollInterval = 2 * time.Second
+	slotTimer := time.NewTimer(time.Second)
```
qdm12, line 324: nit it would be nice to have this test take a shorter time and cover more cases, you can do that using the :warning: change mentioned above, and test for:

- empty queue polled once (at first if block, no ticker)
- empty queue polled multiple times (twice or more, you can set the period to 1ns)
- queue with one element polled once (at first if block, no ticker)
- empty queue polled multiple times until new element is added.. period can also be 1ns, but you need to add an element after i.e. 1ms.

I wrote such test for contexts, and adapted it for PopWithTimer, and it should cover all code paths, cases above and take up to 1ms (I didn't try it though!):

```go
func Test_PriorityQueue_PopWithTimer(t *testing.T) {
	t.Parallel()

	testCases := map[string]struct {
		queueBuilder  func() *PriorityQueue
		queueModifier func(queue *PriorityQueue, done chan<- struct{})
		timer         *time.Timer
		transaction   *ValidTransaction
	}{
		"empty queue polled once": {
			// test should last 1ns
			queueBuilder: NewPriorityQueue,
			timer:        time.NewTimer(time.Nanosecond),
		},
		"empty queue polled multiple times": {
			// test should last 1ms
			queueBuilder: func() *PriorityQueue {
				queue := NewPriorityQueue()
				queue.pollInterval = time.Nanosecond
				return queue
			},
			timer: time.NewTimer(time.Millisecond),
		},
		"queue with one element polled once": {
			// test should be instantaneous
			queueBuilder: func() *PriorityQueue {
				queue := NewPriorityQueue()
				queue.Push(&ValidTransaction{Validity: &Validity{Priority: 1}})
				return queue
			},
			timer:       time.NewTimer(time.Nanosecond),
			transaction: &ValidTransaction{Validity: &Validity{Priority: 1}},
		},
		"queue polled multiple times until new element": {
			// test should last 1ms
			queueBuilder: func() *PriorityQueue {
				queue := NewPriorityQueue()
				queue.pollInterval = time.Nanosecond
				return queue
			},
			queueModifier: func(queue *PriorityQueue, done chan<- struct{}) {
				close(done)
				time.Sleep(time.Millisecond)
				queue.Push(&ValidTransaction{Validity: &Validity{Priority: 1}})
			},
			timer:       time.NewTimer(time.Second),
			transaction: &ValidTransaction{Validity: &Validity{Priority: 1}},
		},
	}

	for name, testCase := range testCases {
		testCase := testCase
			t.Parallel()

			queue := testCase.queueBuilder()

			modifyDone := make(chan struct{})
			if testCase.queueModifier != nil {
				// modify queue asynchronously while popping
				go testCase.queueModifier(queue, modifyDone)
			} else {
				close(modifyDone)
			}

			transaction := queue.PopWithTimer(testCase.timer)
			<-modifyDone
			testCase.timer.Stop()
			assert.Equal(t, testCase.transaction, transaction)
		})
	}
}

```

qdm12 (changes requested):
Requesting changes really for the initial queue check in PopWithTimer that is quite important.

PR #2641, opened 2022-07-04
trie: add new variant headers

qdm12:
## Changes

*Note*: This is based on #2665 

- [x] Add node headers
- [ ] Reject decoding new headers when using `v0` state trie
- [ ] Proof verification to use compact encoding escape header
- [ ] Implement leaf with hashed subvalue
- [ ] Implement branch with hashed subvalue
- [ ] Support child tries upgrade
- [ ] Support proof generation/verification

Questions:

- [ ] Should `Encode` return the too-large values or take in a Database interface to store them directly? How about for proof generation?
- [ ] Should `Decode` take in a database interface to load values directly, or be passed a map of hash<->value? How about for proof verification?
- [x] How does Substrate retrieves subvalues from hashes? Using database lookup
- [ ] How to know which version to use to encode/decode nodes in the state trie?
- [x] How does Substrate delete subvalues? For 'dumb' KV, they prepend the subvalue hash with the node key to go around the storing another key value in another table from hash subvalue to reference counter.
- [x] Regarding the `ESCAPE_HEADER` (aka 'reserved for compact encoding' aka `00010000`), is the 'value node' just the scale encoding of the node subvalue bytes? Also why is this header needed if leaf and branch with hashed subvalue already have their own header `001` and `0001` (or the opposite, why did we need those 2 new headers if we have this compact encoding escape header)?
    ```rust
	/// Escape header byte sequence to indicate next node is a
	/// branch or leaf with hash of value, followed by the value node.
	const ESCAPE_HEADER: Option<u8> = None;
    ```

    See https://github.com/paritytech/substrate/discussions/11607#discussioncomment-3102924

Post checks:

- [ ] Review tests using `version.V0` and make them test cases with V1 as well

## Tests

```sh
go test ./lib/trie/... ./internal/trie/...
```

## Issues

- #2418

### DIFF
--- a/internal/trie/node/decode.go
+++ b/internal/trie/node/decode.go
@@ -39,13 +39,13 @@ func Decode(reader io.Reader) (n *Node, err error) {
 	}
 
 	switch variant {
-	case leafVariant.bits:
+	case leafVariant:
 		n, err = decodeLeaf(reader, partialKeyLength)
 		if err != nil {
 			return nil, fmt.Errorf("cannot decode leaf: %w", err)
 		}
 		return n, nil
-	case branchVariant.bits, branchWithValueVariant.bits:
+	case branchVariant, branchWithValueVariant:
 		n, err = decodeBranch(reader, variant, partialKeyLength)
 		if err != nil {
 			return nil, fmt.Errorf("cannot decode branch: %w", err)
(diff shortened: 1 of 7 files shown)


dimartiro:
Closed in favour of #3295

PR #2648, opened 2022-07-06
fix(lib/babe): ensure the slot time is correct before build a block

### LINKED ISSUE #2634: Runtime error `Timestamp slot must match CurrentSlot` while producing block
## Describe the bug

- While running gossamer as Bob and two other substrate nodes as Alice and Charlie, respectively I notice that the gossamer node throw several of the following error:

```
2022-06-30T14:39:16-04:00 CRIT target=runtime message=panicked at 'Timestamp slot must match `CurrentSlot`', /Users/eclesiojunior/.cargo/git/checkouts/substrate-7e08433d4c370a21/563f482/frame/babe/src/lib.rs:845:9	imports.go:L139:ext_logging_log_version_1	pkg=runtime module=go-wasmer
2022-06-30T14:39:16-04:00 WARN failed to handle slot 414153514: cannot build inherents: Failed to call the `BlockBuilder_apply_extrinsic` exported function.	epoch_handler.go:L125	pkg=babe
```

I noticed that running gossamer as Alice (babe-lead: true) and two other substrate nodes as Bob and Charlie this error rarely appears.

Obs: In fact gossamer waits for the first block to be produced to initiate the epoch and set up the slots.

EclesioMeloJunior:
## Changes

- Replace `time.Until` by `startTime.Sub(time.Now())`
- Use `time.Sleep` to correct the time if there is a negative difference between the current time (when the channel is triggered) and the time defined in the slot

## Tests

It is possible to see the time being adjusted after the modifications

Screen Shot 2022-07-06 at 15 35 38

## Issues

- Closes #2634

## Primary Reviewer

@qdm12

### DIFF
--- a/lib/babe/epoch_handler.go
+++ b/lib/babe/epoch_handler.go
@@ -91,17 +91,16 @@ func (h *epochHandler) run(ctx context.Context, errCh chan<- error) {
 		}
 
 		startTime := getSlotStartTime(authoringSlot, h.constants.slotDuration)
-
 		waitTime := startTime.Sub(time.Now())
 		timer := time.NewTimer(waitTime)
 
 		slotTimeTimers = append(slotTimeTimers, &slotWithTimer{
 			timer:     timer,
-			startTime: startTime,
 			slotNum:   authoringSlot,
+			startTime: startTime,
 		})
 
-		logger.Debugf("start time of slot %d: %v", authoringSlot, startTime.UnixMilli())
+		logger.Debugf("start time of slot %d: %v", authoringSlot, startTime)
 	}
 
 	defer func() {
@@ -125,9 +124,10 @@ func (h *epochHandler) run(ctx context.Context, errCh chan<- error) {
 			// we must do a time correction as the slot timer sometimes is triggered
 			// before the time defined in the constructor due to an inconsistency
 			// of the language -> https://github.com/golang/go/issues/17696
-			diff := time.Now().Sub(swt.startTime).Nanoseconds()
+
+			diff := time.Since(swt.startTime)
 			if diff < 0 {
-				time.Sleep(-time.Duration(diff))
+				time.Sleep(-diff)
 			}
 
 			if _, has := h.slotToPreRuntimeDigest[swt.slotNum]; !has {
--- a/lib/babe/epoch_handler.go
+++ b/lib/babe/epoch_handler.go
@@ -78,8 +78,9 @@ func (h *epochHandler) run(ctx context.Context, errCh chan<- error) {
 	authoringSlots := getAuthoringSlots(h.slotToPreRuntimeDigest)
 
 	type slotWithTimer struct {
-		timer   *time.Timer
-		slotNum uint64
+		startTime time.Time
+		timer     *time.Timer
+		slotNum   uint64
 	}
 
 	slotTimeTimers := make([]*slotWithTimer, 0, len(authoringSlots))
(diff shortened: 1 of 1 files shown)


jimjbrettj (approved):
lgtm

lib/babe/epoch_handler.go
```diff
@@ -115,6 +122,14 @@ func (h *epochHandler) run(ctx context.Context, errCh chan<- error) {
 		case <-ctx.Done():
 			return
 		case <-swt.timer.C:
+			// we must do a time correction as the slot timer sometimes is triggered
+			// before the time defined in the constructor due to an inconsistency
+			// of the language -> https://github.com/golang/go/issues/17696
```
qdm12, line 127: I'm still skeptical on this; I ran the following test (adding load and running a bunch of timers) and none occur before the expected end time... Can you try running it on your end?

```go

func Test_Timers(t *testing.T) {
	// Add CPU load
	ctx, cancel := context.WithCancel(context.Background())
	wg := new(sync.WaitGroup)
	numGoroutines := runtime.NumCPU()
	wg.Add(numGoroutines)
	for i := 0; i < numGoroutines; i++ {
		go func() {
			defer wg.Done()
			i := 0
			b := make([]byte, 1024)
			for {
				for j := range b {
					b[j] = 2*b[j] ^ 0b1111_1111 + 1
				}

				i++
				if i%1000 == 0 && ctx.Err() != nil {
					return
				}
			}
		}()
	}

	type timerData struct {
		startTime time.Time
		timer     *time.Timer
	}
	const timeout = 2 * time.Second
	const sleepBetween = 2 * time.Millisecond
	const numberOfTimers = 100
	timers := make([]timerData, numberOfTimers)

	for i := 0; i < numberOfTimers; i++ {
		timers[i].startTime = time.Now()
		timers[i].timer = time.NewTimer(timeout)
		time.Sleep(sleepBetween)
	}

	for i := 0; i < numberOfTimers; i++ {
		expectedEndTime := timers[i].startTime.Add(timeout)
		received := <-timers[i].timer.C
		endTime := time.Now()
		if endTime.Before(expectedEndTime) {
			t.Errorf("expected end time %s to be after %s (received %s from channel)",
				endTime, expectedEndTime, received)
		}
	}

	cancel()
	wg.Wait()
}

```

qdm12 (changes requested):

timwu20:
This actually fixes the timestamp issue?

kishansagathiya (commented):
Shouldn't you test this against latest runtime?

EclesioMeloJunior:
> This fixes the timestamp issue when building blocks?

@timwu20 we, in the cases where the timer is triggered before the intended time it adjusts correctly and keep the block production in the right slot's timestamp

lib/babe/epoch_handler.go
EclesioMeloJunior: I tested this code many times using `go clean -testcache && go test -run ^Test_Timers$ github.com/ChainSafe/gossamer/lib/babe -v` and it did not fail

lib/babe/epoch_handler.go
```diff
@@ -115,6 +122,14 @@ func (h *epochHandler) run(ctx context.Context, errCh chan<- error) {
 		case <-ctx.Done():
 			return
 		case <-swt.timer.C:
```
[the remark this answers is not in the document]
EclesioMeloJunior: I added the following code:
```
case t := <-swt.timer.C:
  logger.Debugf(
    "timer triggered: time: %v, diff of %v", 
    t.UnixMilli(), swt.startTime.Sub(t).Milliseconds())
```
and the logged time is inconsistent (sometimes it log the timestamp before the defined)
```
DBUG timer triggered: time: 1657195472000, diff of 0
DBUG timer triggered: time: 1657195476000, diff of 0
DBUG timer triggered: time: 1657195515998, diff of 1
DBUG timer triggered: time: 1657195551997, diff of 2
DBUG timer triggered: time: 1657195579996, diff of 3
DBUG timer triggered: time: 1657195623995, diff of 4
DBUG timer triggered: time: 1657195648047, diff of -47 
DBUG timer triggered: time: 1657196003994, diff of 5
```

lib/babe/epoch_handler.go
```diff
@@ -115,6 +122,14 @@ func (h *epochHandler) run(ctx context.Context, errCh chan<- error) {
...
+			// we must do a time correction as the slot timer sometimes is triggered
+			// before the time defined in the constructor due to an inconsistency
+			// of the language -> https://github.com/golang/go/issues/17696
+			diff := time.Now().Sub(swt.startTime).Nanoseconds()
+			if diff < 0 {
+				time.Sleep(-time.Duration(diff))
+			}
```
qdm12, line 131:
```suggestion
			diff := time.Since(swt.startTime)
			if diff < 0 {
				time.Sleep(-diff)
			}
```

qdm12 (approved):

qdm12 (approved):

qdm12 (approved):

qdm12 (approved):

PR #2650, opened 2022-07-07
chore(lib/runtime/wasmer): import functions in a for loop

qdm12:
## Changes

- `ImportsNodeRuntime`: append functions in a for loop
- Remove no longer needed `//nolint:gocyclo`
- Update now deprecated `.Append` to be `.AppendFunction`

## Issues

Done whilst adding functions

## Primary Reviewer

@kishansagathiya

### DIFF
--- a/lib/runtime/wasmer/imports.go
+++ b/lib/runtime/wasmer/imports.go
@@ -2231,350 +2231,102 @@ func toWasmMemoryFixedSizeOptional(context wasm.InstanceContext, data []byte) (i
 }
 
 // ImportsNodeRuntime returns the imports for the v0.8 runtime
-func ImportsNodeRuntime() (*wasm.Imports, error) { //nolint:gocyclo
-	var err error
+func ImportsNodeRuntime() (imports *wasm.Imports, err error) {
+	imports = wasm.NewImports()
 
-	imports := wasm.NewImports()
-
-	_, err = imports.Append("ext_allocator_free_version_1", ext_allocator_free_version_1, C.ext_allocator_free_version_1)
-	if err != nil {
-		return nil, err
-	}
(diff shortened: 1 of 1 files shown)


lib/runtime/wasmer/imports.go
```diff
@@ -2230,351 +2230,103 @@ func toWasmMemoryFixedSizeOptional(context wasm.InstanceContext, data []byte) (i
...
-		return nil, err
-	}
-	_, err = imports.Append("ext_transaction_index_renew_version_1", ext_transaction_index_renew_version_1, C.ext_transaction_index_renew_version_1)
-	if err != nil {
-		return nil, err
+// ImportsNodeRuntime returns the WASM imports for the node runtime.
+func ImportsNodeRuntime() (imports *wasm.Imports, err error) {
+	imports = wasm.NewImports()
+
+	for _, toRegister := range []struct {
```
EclesioMeloJunior, line 2237: Would be more clear to have this as a fixed-sized array outside of the function scope IMO

```go
var importedFuncs = [...]struct{
  import_name string
  implementation interface{}
  cgoPointer     unsafe.Pointer
}{
    ...
}

for _, toRegister := importedFuncs {
    _, err = imports.AppendFunction(toRegister.importName, toRegister.implementation, toRegister.cgoPointer)
    if err != nil {
        return nil, fmt.Errorf("importing function: %w", err)
    }
}
```

EclesioMeloJunior (approved):

lib/runtime/wasmer/imports.go
qdm12: Definitely taking the fixed size array (god damn second time I don't use it, thanks for the reminder!).
Although I don't really see the point to have it at global scope 🤔 I don't think it would change performance wise, and it's only used in that function. It would also pollute more the (unexported) package Go API.

kishansagathiya (approved):

Issue #2489, opened 2022-04-18
Fix libp2p Dependabot PR


danforbes:
Dependabot opened #2488 to upgrade libp2p from v0.9.0 to v0.15.0, but it's broken our builds. The build should be fixed so that we upgrade our libp2p version. If the upgrade doesn't directly relate to patching some vulnerability, or allowing us to participate in a Polkadot network, then this probably isn't very high priority.

qdm12:
Also github.com/libp2p/go-libp2p-discovery needs to be upgraded

PR #2668, opened 2022-07-13
chore(network): update libp2p libraries

edwardmack:
## Changes
Upgrade the following libp2p libraries:
- go-libp2p from v0.15.1 to v0.20.3
- go-libp2p-core from v0.9.0 to v0.16.1
- go-libp2p-kad-dht from v0.11.1 to v0.16.0
- go-libp2p-peerstore from v0.3.0 to v0.6.0

Addressed breaking changes:
- In go-libp2p v0.16.0 the `context.Context` in the host constructor (`libp2p.New`) was removed, see: https://github.com/libp2p/go-libp2p/releases/tag/v0.16.0
- In go-libp2p-core v0.16.0 The `OpenedStream` and `ClosedStream` methods were removed from the `network.Notifiee` interface.  See: https://github.com/libp2p/go-libp2p-core/releases/tag/v0.16.0
- In go-libp2p-core v0.11.0 the `ConnHandler` interface, both its definition and use was removed from `Network` interface.  See:  https://github.com/libp2p/go-libp2p-core/releases/tag/v0.11.0

## Tests

```sh
go test github.com/ChainSafe/gossamer/dot/network -short
```
Also tested on 3 node gossamer devnet and reviewed logs to confirm that nodes were connecting and communicating with each other. 

## Issues
closes: #2489 

## Primary Reviewer

@timwu20


dot/network/service_test.go
```diff
@@ -295,7 +295,7 @@ func TestBroadcastDuplicateMessage(t *testing.T) {
 	}
 
 	time.Sleep(time.Millisecond * 500)
-	require.Equal(t, 2, len(handler.messages[nodeA.host.id()]))
+	require.Equal(t, 1, len(handler.messages[nodeA.host.id()]))
```
edwardmack, line 298: Note, this was changed to 1 since we are expecting only one message to have been sent here, see comment on line 291.  I'm not sure why it was working with 2 in development, may be related to changes when `ConnHandler` interface was removed.

go.mod
```diff
@@ -181,4 +160,36 @@ require (
 	gotest.tools/v3 v3.0.3 // indirect
 )
 
+require (
+	github.com/benbjohnson/clock v1.3.0 // indirect
+	github.com/btcsuite/btcd/chaincfg/chainhash v1.0.1 // indirect
+	github.com/cheekybits/genny v1.0.0 // indirect
+	github.com/containerd/cgroups v0.0.0-20201119153540-4cbc285b3327 // indirect
```
qdm12, line 167: Please move these in the existing `require` block above

Also running `go mod tidy` may help removing deps

dot/network/service.go
```diff
@@ -277,16 +277,13 @@ func (s *Service) Start() error {
...
-	// NOTE: this only handles *incoming* connections
-	s.host.p2pHost.Network().SetConnHandler(s.handleConn)
-
 	// this handles all new connections (incoming and outgoing)
 	// it creates a per-protocol mutex for sending outbound handshakes to the peer
 	s.host.cm.connectHandler = func(peerID peer.ID) {
 		for _, prtl := range s.notificationsProtocols {
 			prtl.peersData.setMutex(peerID)
 		}
+		s.host.cm.peerSetHandler.Incoming(0, peerID)
```
qdm12, line 286: 1. Adding the `setID` constant since the interface doesn't have named arguments (sus) and it's hard to know what the first arg is.
2. Re-adding the older TODO comment
3. Don't we need the extra logic from the now removed `handleConn`?

```suggestion
		// TODO: currently we only have one set so setID is 0, change this once we have more set in peerSet.
		const setID = 0
		s.host.cm.peerSetHandler.Incoming(setID, peerID)
```

dot/network/stream_manager_test.go
```diff
@@ -41,15 +37,22 @@ func setupStreamManagerTest(t *testing.T) (context.Context, []libp2phost.Host, [
...
-		ctx, libp2p.ListenAddrs(addrB),
+		libp2p.ListenAddrs(addrB),
 	)
 	require.NoError(t, err)
 
+	t.Cleanup(func() {
+		cleanupStreamInterval = time.Minute
+		cancel()
+		ha.Close()
+		hb.Close()
```
qdm12, line 53:
```suggestion
		assert.NoError(t, ha.Close())
		assert.NoError(t, hb.Close())
```

qdm12 (changes requested):

dot/network/service.go
edwardmack: It seems that the extra logic in handleConn is handled somewhere else, when ConnHandler interface was removed.  Testing with 3 node devnet confirms that nodes are connecting and communicating.

EclesioMeloJunior (approved):

dot/network/stream_manager_test.go
```diff
@@ -41,15 +37,22 @@ func setupStreamManagerTest(t *testing.T) (context.Context, []libp2phost.Host, [
...
 	require.NoError(t, err)
 
 	hb, err := libp2p.New(
-		ctx, libp2p.ListenAddrs(addrB),
+		libp2p.ListenAddrs(addrB),
 	)
 	require.NoError(t, err)
 
+	t.Cleanup(func() {
+		cleanupStreamInterval = time.Minute
```
qdm12, line 50: nit A bit out of scope, but having this horrendous global variable `cleanupStreamInterval` within the `streamManager` struct would be much nicer than a global var we fiddle with in tests

qdm12 (approved):

dot/network/host.go
```diff
@@ -187,7 +187,7 @@ func newHost(ctx context.Context, cfg *Config) (*host, error) {
 	}
 
 	// create libp2p host instance
-	h, err := libp2p.New(ctx, opts...)
+	h, err := libp2p.New(opts...)
```
timwu20, line 190: Don't we need to handle the context cancellation explicitly now and call `h.Close()`?

edwardmack: We are handling it in our close function at line 239 in dot/host.go

PR #2673, opened 2022-07-15
chore(lib/runtime): refactor version struct and codec

qdm12:
## Changes

- Version code
  - Single `Version` struct
  - Encoding the `Version` always encode all the last fields
  - Decoding the version data to `Version` sets possible absent fields as zero if not in the data (`transaction_version`)
  - Remove `LegacyVersionData` struct, `Version` interface, getter methods, cloned structs for codec
  - Move decoding logic from `lib/runtime/wasmer/exports.go` to `lib/runtime/version.go` in `DecodeVersion` function
  - Use struct value for `Version` instead of its pointer (no reason to use a pointer)
- Test code
  - Refactor multiple `TestInstance_Version*` in one test with test cases
  - Add runtime API items as expected fields in tests
  - Remove version runtime mockery mock
  - Simplify testing in other packages since we no longer need to use a test-only exported constructor `NewVersionData`

## Issues

I had to change a whole bunch of version related code to persist the state version, and I thought I would simplify it since it was rather confusing as a reader.

## Primary Reviewer

@timwu20


noot (approved):
i think this is fine assuming the encoding is the same (should be cause tests pass :p)

EclesioMeloJunior (approved):

dot/core/service.go
```diff
@@ -454,26 +454,27 @@ func (s *Service) DecodeSessionKeys(enc []byte) ([]byte, error) {
 }
 
 // GetRuntimeVersion gets the current RuntimeVersion
-func (s *Service) GetRuntimeVersion(bhash *common.Hash) (runtime.Version, error) {
+func (s *Service) GetRuntimeVersion(bhash *common.Hash) (
+	version runtime.Version, err error) {
```
timwu20, line 458: did you consider returning a `*runtime.Version`.  I feel like the zero case of nil is more applicable here.

dot/rpc/subscription/listeners.go
```diff
@@ -410,13 +410,13 @@ func (l *RuntimeVersionListener) Listen() {
...
-	ver.ImplVersion = rtVersion.ImplVersion()
-	ver.TransactionVersion = rtVersion.TransactionVersion()
-	ver.Apis = modules.ConvertAPIs(rtVersion.APIItems())
+	ver.SpecName = string(rtVersion.SpecName)
+	ver.ImplName = string(rtVersion.ImplName)
+	ver.AuthoringVersion = rtVersion.AuthoringVersion
+	ver.SpecVersion = rtVersion.SpecVersion
+	ver.ImplVersion = rtVersion.ImplVersion
+	ver.TransactionVersion = rtVersion.TransactionVersion
+	ver.Apis = modules.ConvertAPIs(rtVersion.APIItems)
```
timwu20, line 419: can we move this to a helper function `newStateRuntimeVersionResponse(ver Version)`.

lib/runtime/version.go
```diff
@@ -4,89 +4,32 @@
...
+	AuthoringVersion   uint32
+	SpecVersion        uint32
+	ImplVersion        uint32
+	APIItems           []APIItem
+	TransactionVersion uint32
+	legacy             bool
 }
 
-type legacyVersionData struct {
+type legacyData struct {
```
timwu20, line 32: why not embed this struct in `Version`?

timwu20 (changes requested):

lib/runtime/version.go
qdm12: I tried it, it encodes wrong with scale, so we would need another clone struct. But good suggestion nonetheless!!

dot/core/service.go
qdm12: I think we should keep it dumb without pointer whenever possible (and for performance as well).
If an error happens, the caller should discard the version returned anyway and handle the error.

dot/rpc/subscription/listeners.go
qdm12: Since it's a bit OOS, it's done in #2740 pointing to this branch.
Feel free to review now or later, we can merge it in this branch or merge it later in development if this branch gets merged first 😉

qdm12:
⚠️ do not merge, this needs more changes (to support N versions of the version since the new state_version field is yet another version).

lib/runtime/version.go
qdm12: Now irrelevant, this is changed to be:

- encoding: always encode to last format with all fields (some to their zero value `0`)
- decoding: decodes all required fields and tries to decode optional fields one by one without erroring when reaching EOF

timwu20 (approved):

PR #2674, opened 2022-07-15
fix(tests): fix block body regex in `TestChainRPC`

qdm12:
## Changes

`TestChainRPC` broke on our development branch.

```
Expect "0x280403000bbbb895028201" to match "^0x280403000b[0-9a-z]{8}8101$"
```

I wrote that regex, and it was consistently ending with 8101, now it's 8201 out of the blue.
This PR updates the regex to use `8201` instead.
But ideally, someone can point why this changed? 🤔 

## Tests

```sh
MODE=rpc go test -run ^TestChainRPC$ github.com/ChainSafe/gossamer/tests/rpc
```

## Primary Reviewer

@timwu20

### DIFF
--- a/tests/rpc/rpc_03-chain_test.go
+++ b/tests/rpc/rpc_03-chain_test.go
@@ -102,7 +102,7 @@ func TestChainRPC(t *testing.T) {
 	}
 	block.Block.Header.Digest.Logs = nil
 	assert.Len(t, block.Block.Body, 1)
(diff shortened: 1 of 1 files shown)


jimjbrettj (approved):
lgtm

EclesioMeloJunior (approved):
Looks like the extrinsic that is put in the block body changed slightly but I'm not sure... 🤔

PR #2675, opened 2022-07-15
chore(lib/runtime): remove imports function field

qdm12:
## Changes

- Remove wasmer `Config` struct
- Use `runtime.InstanceConfig` as wasmer config
- Always use `ImportsNodeRuntime` (as it was)
- Remove `imports` field from `Instance`
- Change tests runtime log levels to critical

## Issues

Longshot to be able to dependency inject the state version in the Wasmer instance constructor.

## Primary Reviewer

@edwardmack


dot/core/service.go
```diff
@@ -34,7 +34,7 @@ var (
 // QueryKeyValueChanges represents the key-value data inside a block storage
 type QueryKeyValueChanges map[string]string
 
-type wasmerInstanceFunc func(code []byte, cfg *wasmer.Config) (instance *wasmer.Instance, err error)
+type wasmerInstanceFunc func(code []byte, cfg runtime.InstanceConfig) (instance *wasmer.Instance, err error)
```
jimjbrettj, line 37: No longer a pointer?

jimjbrettj: Saw the other changes, answered my own question

jimjbrettj (approved):
lgtm, just one comment asking about a TODO comment

dot/core/service.go
qdm12: Nah just a dumb struct. No need for a pointer, the struct isn't that big and contains slices/maps which are themselves pointers. FYI and counter-intuitively, most of the time in Go, it's faster to copy structs on the stack than passing a pointer (which also involves load on the GC).

Issue #2631, opened 2022-06-29
pkg/scale: define sentinel errors and add error wrapping


qdm12:
## Task summary

The `pkg/scale` package is cumbersome to use together with error wrapping aware code, since no sentinel error is defined and exported.

All defined `errors.New(` and `fmt.Errorf(` constant strings errors should be moved to global scope and exported so callers can use them together with `errors.Is(err, scale.ErrXXX)`. This can be useful for testing packages importing the scale package as well, where the (programmatic) error has to be asserted.

Standard library error wrapping with `fmt.Errorf("doing something: %w", err)` should also be added at the same time to add more context to errors returned.

timwu20:
@qdm12 where in the gossamer codebase do we actually check error types?

PR #2676, opened 2022-07-15
feat(scale): implement max length constraint for decoding bytes

### LINKED ISSUE #2629: Bug: payload makes scale decoder go out of memory
## Describe the bug

Scale decoder can go out of memory with specific input bytes, this will panic:

```go
func Test_Decoder_Decode_Fail(t *testing.T) {
	reader := bytes.NewBuffer([]byte{0x7, 0x8, 0x9, 0x10, 0x0, 0x40})
	decoder := scale.NewDecoder(reader)
	var result []byte
	_ = decoder.Decode(&result)
}
```

The message was found a bit randomly whilst writing tests for the trie.

Maybe we need something like https://docs.rs/parity-scale-codec/3.1.5/parity_scale_codec/trait.DecodeLimit.html to prevent crazy recursions 🤔 

As a side note, we should definitely do #2630 

## Expected Behavior

Do not go out of memory / do not panic.

## Current Behavior

```
fatal error: runtime: out of memory
```

edwardmack:
## Changes
- Add maxLength constraint to `decodeState` where error is raised if `decodeState.decodeLength` is greater than `decodeState.maxLength` to avoid trying to decode very long bytes values.
- Add constant `defaultMaxLength` of 16 GB to use as default `maxLength` value, so that decoder will not attempt to decode bytes encoded with a length of greater than 16 GB.
- Updated function `NewDecoder` and `Unmarshal` to accept variadic option, following pattern explained in https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis 
- Added function `OptionMaxLength` for overriding default `maxLength`
- Added unit tests for testing use of `OptionMaxLength` with `NewDecoder` and `Unmarshal`
- Added unit tests for testing decoding bytes with length greater than `defaultMaxLength`

## Tests

```sh
go test github.com/ChainSafe/gossamer/pkg/scale/
```

## Issues

fixes #2629, which was causing out of memory errors when trying to decode very long bytes encodings.
## Primary Reviewer

@timwu20

### DIFF
--- a/pkg/scale/decode.go
+++ b/pkg/scale/decode.go
@@ -56,8 +56,10 @@ func indirect(dstv reflect.Value) (elem reflect.Value) {
 	return
 }
 
+const defaultMaxLength = 16 * 1024 * 1024 * 1024 // 16GB
+
 // Unmarshal takes data and a destination pointer to unmarshal the data to.
-func Unmarshal(data []byte, dst interface{}) (err error) {
+func Unmarshal(data []byte, dst interface{}, options ...func(*Decoder)) (err error) {
 	dstv := reflect.ValueOf(dst)
 	if dstv.Kind() != reflect.Ptr || dstv.IsNil() {
 		err = fmt.Errorf("unsupported dst: %T, must be a pointer to a destination", dst)
@@ -69,15 +71,21 @@ func Unmarshal(data []byte, dst interface{}) (err error) {
 		return
 	}
 
+	d := &Decoder{}
+	d.maxLength = defaultMaxLength
+
+	// Option parameters values:
+	for _, option := range options {
+		option(d)
+	}
 	buf := &bytes.Buffer{}
-	ds := decodeState{}
 	_, err = buf.Write(data)
 	if err != nil {
 		return
 	}
-	ds.Reader = buf
+	d.Reader = buf
 
-	err = ds.unmarshal(elem)
+	err = d.decodeState.unmarshal(elem)
 	if err != nil {
 		return
 	}
@@ -105,15 +113,29 @@ func (d *Decoder) Decode(dst interface{}) (err error) {
 }
 
 // NewDecoder is constructor for Decoder
-func NewDecoder(r io.Reader) (d *Decoder) {
-	d = &Decoder{
-		decodeState{r},
+func NewDecoder(r io.Reader, options ...func(*Decoder)) (d *Decoder) {
+	d = &Decoder{}
+	// Default values...
+	d.decodeState.Reader = r
+	d.decodeState.maxLength = defaultMaxLength
+
+	// Option parameters values:
+	for _, option := range options {
+		option(d)
 	}
 	return
 }
 
+// OptionMaxLength sets the maximum length of the data to be decoded
+func OptionMaxLength(maxLength int) func(*Decoder) {
+	return func(options *Decoder) {
+		options.maxLength = maxLength
+	}
+}
+
 type decodeState struct {
 	io.Reader
+	maxLength int
 }
 
 func (ds *decodeState) unmarshal(dstv reflect.Value) (err error) {
@@ -539,6 +561,9 @@ func (ds *decodeState) decodeBytes(dstv reflect.Value) (err error) {
 	length, err := ds.decodeLength()
 	if err != nil {
 		return
+	}
+	if length > ds.maxLength {
+		return fmt.Errorf("could not decode length %v greater than max length %v", length, ds.maxLength)
 	}
 
 	b := make([]byte, length)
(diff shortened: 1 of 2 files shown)


pkg/scale/decode.go
```diff
@@ -105,15 +113,29 @@ func (d *Decoder) Decode(dst interface{}) (err error) {
...
+// OptionMaxLength sets the maximum length of the data to be decoded
+func OptionMaxLength(maxLength int) func(*Decoder) {
+	return func(options *Decoder) {
+		options.maxLength = maxLength
+	}
+}
+
 type decodeState struct {
 	io.Reader
+	maxLength int
```
qdm12, line 138: That's gonna be a niche nit...

getting the size through file.Stat() in Go returns a `int64`;
maybe we should have this `maxLength` as `int64` if i.e. the writer is a file, and the OS is 32 bit (then `int` = `int32`)

pkg/scale/decode.go
```diff
@@ -540,6 +562,9 @@ func (ds *decodeState) decodeBytes(dstv reflect.Value) (err error) {
 	if err != nil {
 		return
 	}
+	if length > ds.maxLength {
+		return fmt.Errorf("could not decode length %v greater than max length %v", length, ds.maxLength)
```
qdm12, line 566: make this a sentinel error. I understand this package doesn't use sentinel errors, but that would help one more step for #2631

pkg/scale/decode.go
```diff
@@ -105,15 +113,29 @@ func (d *Decoder) Decode(dst interface{}) (err error) {
...
+		option(d)
 	}
 	return
 }
 
+// OptionMaxLength sets the maximum length of the data to be decoded
+func OptionMaxLength(maxLength int) func(*Decoder) {
+	return func(options *Decoder) {
+		options.maxLength = maxLength
+	}
```
qdm12, line 133: an error would be useful here to prevent the user from settings a limit of `0` for example here.

pkg/scale/decode_test.go
```diff
@@ -70,6 +71,103 @@ func Test_decodeState_decodeBytes(t *testing.T) {
...
+	for _, tt := range maxLengthDecodeTests {
+		t.Run(tt.name, func(t *testing.T) {
+			dst := reflect.New(reflect.TypeOf(tt.in)).Elem().Interface()
+			if err := Unmarshal(tt.want, &dst, OptionMaxLength(64)); (err != nil) != tt.wantErr {
+				t.Errorf("decodeState.unmarshal() error = %v, wantErr %v", err, tt.wantErr)
+				return
+			}
+			if !reflect.DeepEqual(dst, tt.in) {
+				t.Errorf("decodeState.unmarshal() = %#v, want %#v", dst, tt.in)
+			}
```
qdm12, line 166: Use `assert.Equal`

(also the `t.Errorf` message is wrong here)

pkg/scale/decode_test.go
```diff
@@ -70,6 +71,103 @@ func Test_decodeState_decodeBytes(t *testing.T) {
 	}
 }
 
+var maxLengthDecodeTests = tests{
```
qdm12, line 74: make this a function returning a `tests`, so we can be (more) sure there is no funny mutation between each test using it.

pkg/scale/decode_test.go
```diff
@@ -70,6 +71,103 @@ func Test_decodeState_decodeBytes(t *testing.T) {
...
+	{
+		name: "myCustomBytes(testStrings[0])",
+		in:   myCustomBytes(testStrings[0]),
+
+		want: append([]byte{0x0D, 0x01}, testStrings[0]...),
+	},
+}
+
+func Test_decodeState_decodeBytes_maxLength(t *testing.T) {
+
```
qdm12, line 156:
```suggestion
	t.Parallel()
```

pkg/scale/decode.go
```diff
@@ -105,15 +113,29 @@ func (d *Decoder) Decode(dst interface{}) (err error) {
 }
 
 // NewDecoder is constructor for Decoder
-func NewDecoder(r io.Reader) (d *Decoder) {
-	d = &Decoder{
-		decodeState{r},
+func NewDecoder(r io.Reader, options ...func(*Decoder)) (d *Decoder) {
```
kishansagathiya, line 116: What is the `option` here for exactly?
Are they like helper functions for decoding? In which case, may be they could be passed only here and be saved as property of `decoder` and use them in the `Decode` function. If that makes sense, decode function would not need to be passed options.

kishansagathiya (commented):
- Do we have places where `Unmarshal` with options is getting used?
I like the idea of `option` function. It is helpful if we have a bunch of options.

pkg/scale/decode.go
qdm12: That makes sense. Although you would still have to pass options to `scale.Unmarshal` which creates its decoder within AFAIK.

edwardmack:
Closed, replaced by PR #2683

Issue #2636, opened 2022-07-01
Import failed: Slot number must increase


EclesioMeloJunior:
## Describe the bug

- While running a gossamer node with two other substrate nodes I notice that the substrate throws the following error:
```
2022-06-30 14:32:51 💔 Error importing block 0xf919e19068653baf95680af89d1751ce43cf4cc9422c0fa8e98300f470a430fc: consensus error: Import failed: Slot number must increase: parent slot: 414153490, this slot: 414153490
```

This block was created by the gossamer node:
```
2022-06-30T14:32:42-04:00 INFO built block 155 with hash 0xf919e19068653baf95680af89d1751ce43cf4cc9422c0fa8e98300f470a430fc, state root 0x8305859db24d16d1a34cee766dbf7d5b3132f3d4d0c57d31f21ec9e765752d68, epoch 6 and slot 414153490	babe.go:L541	pkg=babe
```

The slot number specified in the error messages was claimed by the gossamer node as a secondary vrf slot in the epoch 6:
```
2022-06-30T14:32:40-04:00 DBUG claimed secondary slot, for slot number: 414153490	crypto.go:L110	pkg=babe
2022-06-30T14:32:40-04:00 DBUG epoch 6: claimed secondary vrf slot 414153490	epoch.go:L239	pkg=babe
```

kishansagathiya:
I don't think this is related to secondary slot claiming. I have seen this happen for primary block as well. This has more to do with some time related inaccuracies. This error occurs only for the first slot in the epoch.
If I just change https://github.com/ChainSafe/gossamer/blob/3a471d91f61c1ba10022ff11403b9cebb71c8666/lib/babe/epoch_handler.go#L88 to 
```
		if authoringSlot <= currSlot {
```

we would not see the problem anymore. But that would mean that first slot would never get used, so that's bad.

kishansagathiya:
Different error `block has an unknown parent`, fun!!
```
2022-07-18 17:19:07 💔 Error importing block 0xaf9a9621dd46c50b668967ccc78f983aad789127658143b780296121e34c0f97: block has an unknown parent    
```

kishansagathiya:
> @kishansagathiya I think this is not the reason, the error `genesis hash mismatch` does not enable us to sync with the peers, could you compare the block 0 hash in the gossamer and in the peer they should be the same otherwise the block will not be imported. Maybe the genesis spec you're using is different from the one is used in the peer

[...]

PR #2681, opened 2022-07-19
Added checks to avoid decoding handshakes as messages

kishansagathiya:
## Changes

These changes  would show the appropriate error messages instead of genesis mismatch and also avoid banning the other peer.

## Tests

```sh
```

## Issues

#2636
## Primary Reviewer

@timwu20

### DIFF
--- a/dot/network/block_announce.go
+++ b/dot/network/block_announce.go
@@ -18,6 +18,8 @@ import (
 var (
 	_ NotificationsMessage = &BlockAnnounceMessage{}
 	_ NotificationsMessage = &BlockAnnounceHandshake{}
+
+	errExpectedBlockAnnounceMsg = errors.New("found block announce handshake instead of message")
 )
 
 // BlockAnnounceMessage is a state block header
@@ -223,6 +225,10 @@ func (s *Service) validateBlockAnnounceHandshake(from peer.ID, hs Handshake) err
 // if some more blocks are required to sync the announced block, the node will open a sync stream
 // with its peer and send a BlockRequest message
 func (s *Service) handleBlockAnnounceMessage(from peer.ID, msg NotificationsMessage) (propagate bool, err error) {
+	if msg.IsHandshake() {
+		return false, errExpectedBlockAnnounceMsg
+	}
+
 	bam, ok := msg.(*BlockAnnounceMessage)
 	if !ok {
 		return false, errors.New("invalid message")
--- a/dot/network/transaction.go
+++ b/dot/network/transaction.go
@@ -18,6 +18,8 @@ import (
 var (
 	_ NotificationsMessage = &TransactionMessage{}
 	_ NotificationsMessage = &transactionHandshake{}
+
+	errExpectedTransactionMsg = errors.New("found transaction handshake instead of message")
 )
 
 // txnBatchChTimeout is the timeout for adding a transaction to the batch processing channel
@@ -191,6 +193,10 @@ func decodeTransactionMessage(in []byte) (NotificationsMessage, error) {
 }
 
 func (s *Service) handleTransactionMessage(peerID peer.ID, msg NotificationsMessage) (bool, error) {
+	if msg.IsHandshake() {
+		return false, errExpectedTransactionMsg
+	}
+
 	txMsg, ok := msg.(*TransactionMessage)
 	if !ok {
 		return false, errors.New("invalid transaction type")


dot/network/notifications.go
```diff
@@ -158,6 +158,8 @@ func (s *Service) createNotificationsMessageHandler(
 
 			hs, ok := msg.(Handshake)
 			if !ok {
+				// this error will never happen.
+				// Handshake interface and NotificationMessage interfaces are same.
```
EclesioMeloJunior, line 162: since both interfaces are the same, what if:
```
var hs Handshake = msg
```
is that possible? if so, we don't need to cast and don't need to check the `ok` value

EclesioMeloJunior (changes requested):

dot/network/block_announce.go
```diff
@@ -183,6 +188,10 @@ func (s *Service) validateBlockAnnounceHandshake(from peer.ID, hs Handshake) err
 		return errors.New("invalid handshake type")
 	}
 
+	if int(bhs.Roles) > 4 {
+		return errors.New("invalid handshake role")
```
qdm12, line 192: Use a sentinel error, and wrap the actual role with it

```suggestion
		return fmt.Errorf("%w: %d", errHandshakeRoleNotValid, bhs.Roles)
```

dot/network/notifications.go
qdm12: I'd just use `handshake, ok := msg.(Handshake)` instead of `msg.IsHandshake()` + if condition, right?

dot/network/transaction.go
```diff
@@ -18,6 +18,8 @@ import (
 var (
 	_ NotificationsMessage = &TransactionMessage{}
 	_ NotificationsMessage = &transactionHandshake{}
+
+	errExpectedTransactionMsg = errors.New("found transaction handshake instead of message")
```
qdm12, line 22: nit

```suggestion
	errExpectedTransactionMsg = errors.New("received a transaction handshake but expected a transaction message")
```

dot/network/transaction.go
```diff
@@ -191,6 +193,10 @@ func decodeTransactionMessage(in []byte) (NotificationsMessage, error) {
 }
 
 func (s *Service) handleTransactionMessage(peerID peer.ID, msg NotificationsMessage) (bool, error) {
+	if msg.IsHandshake() {
+		return false, errExpectedTransactionMsg
+	}
```
qdm12, line 198: A `*TransactionMessage` always returns false to `IsHandshake`, so this check is unneeded given the below type assertion.

Also why not having `msg *TransactionMessage` instead of using an interface `msg NotificationsMessage` for the argument? Especially given the function name, type check/switching should be done at the caller 🤔

dot/network/block_announce.go
```diff
@@ -223,6 +232,10 @@ func (s *Service) validateBlockAnnounceHandshake(from peer.ID, hs Handshake) err
 // if some more blocks are required to sync the announced block, the node will open a sync stream
 // with its peer and send a BlockRequest message
 func (s *Service) handleBlockAnnounceMessage(from peer.ID, msg NotificationsMessage) (propagate bool, err error) {
+	if msg.IsHandshake() {
+		return false, errExpectedBlockAnnounceMsg
+	}
+
```
qdm12, line 238: This will always be false since `*BlockAnnounceMessage` always return `false`, given the type assertion right below.
Also why is this function not taking a `*BlockAnnounceMessage` directly instead of `NotificationsMessage`? The type assertion should be done at the calling layer imo, given the function name.

dot/network/notifications.go
EclesioMeloJunior: it is not necessary to use the cast + `ok` check if both `Handshake` and `NotificationMessage` are the same

dot/network/block_announce.go
```diff
@@ -183,6 +188,10 @@ func (s *Service) validateBlockAnnounceHandshake(from peer.ID, hs Handshake) err
 		return errors.New("invalid handshake type")
 	}
 
+	if int(bhs.Roles) > 4 {
```
EclesioMeloJunior, line 191: `3` is an invalid role, right? should you include an error for it?

EclesioMeloJunior (changes requested):

dot/network/block_announce.go
qdm12: We should use a typed byte and constants to be sure about role values etc. (and have meaningful constant names too to know which is which). Dealing with numbers like this is just sus

dot/network/block_announce.go
```diff
@@ -108,6 +110,9 @@ func decodeBlockAnnounceMessage(in []byte) (NotificationsMessage, error) {
 
 // BlockAnnounceHandshake is exchanged by nodes that are beginning the BlockAnnounce protocol
 type BlockAnnounceHandshake struct {
+	// 1 for node is a full node
+	// 2 for node is a light client
+	// 4 for node is a validator
```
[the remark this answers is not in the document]
kishansagathiya: > Maybe add a dot/network/roles package (or somewhere else) to have it shared with other packages as well.

I don't think other packages need it though. I like the idea of typed bytes though. Maybe I could make it enum.

dot/network/block_announce.go
kishansagathiya: that's cause it's a notification message handler. It has to follow certain signature of a message handler to be accepted by `RegisterNotificationsProtocol`.

Take a look at dot/network/service.go line 245

dot/network/transaction.go
kishansagathiya: same reply as https://github.com/ChainSafe/gossamer/pull/2681#discussion_r925750884
handleTransactionMessage is similar to handleBlockAnnounceMessage

dot/network/block_announce.go
kishansagathiya: addressed

dot/network/transaction.go
kishansagathiya: addressed

dot/network/block_announce.go
kishansagathiya: changed to enum now

dot/network/notifications.go
kishansagathiya: > is that possible? if so, we don't need to cast and don't need to check the ok value

Maybe, but I don't see the point of that change. The reason I put that comment is to point out a problem. The solution to that is larger than one fix or so.

kishansagathiya: > I'd just use handshake, ok := msg.(Handshake) instead of msg.IsHandshake() + if condition, right?

` handshake, ok := msg.(Handshake) ` is not enough to check, because it will always be ok, because they follow the same interface.

EclesioMeloJunior: I understand you want to point out a problem that need to be addressed in a future larger refactor, but my point is: the cast + ok check is a useless code and we should avoid it, you can point out the problem using a `//TODO` comment that points to an issue and then pass the `msg` variable to the `info.handshakeValidator(peer, msg)` without casting to `Handshake` interface:

```go
err := info.handshakeValidator(peer, msg)
```
instead of

```go
hs, ok := msg.(Handshake)
if !ok {
  // this error will never happen.
  // Handshake interface and NotificationMessage interfaces are same.
  return errMessageIsNotHandshake
}

...
err := info.handshakeValidator(peer, hs)
```

EclesioMeloJunior (changes requested):

dot/network/block_announce.go
qdm12: > that's cause it's a notification message handler. It has to follow certain signature of a message handler to be accepted by

Why not change the `NotificationsMessageHandler` function argument? I don't see the point to have it as an interface at all. The caller `createNotificationsMessageHandler` should do the type assertion of the NotificationsMessage and pass it down as `*BlockAnnounceMessage`

> The type assertion below could fail.

Yes that's my point. If the type assertion succeeds, `IsHandshake()` is always false. So no point doing both checks, the existing type assertion is enough.

dot/network/service.go
```diff
@@ -629,7 +629,7 @@ func (s *Service) Peers() []common.PeerInfo {
 		peerHandshakeMessage := data.handshake
 		peers = append(peers, common.PeerInfo{
 			PeerID:     p.String(),
-			Roles:      peerHandshakeMessage.(*BlockAnnounceHandshake).Roles,
+			Roles:      byte(peerHandshakeMessage.(*BlockAnnounceHandshake).Roles),
```
qdm12, line 632: Could we change the `Roles` in `PeerInfo` to be a `Roles` type you defined?

dot/network/transaction.go
qdm12: I still don't understand.

`TransactionMessage` always returns false for `IsHandshake()`, so the type assertion below is fine.
Unless you are doing something with the error returned, such as `errors.Is()`, but I don't think that's the case right?

qdm12 (changes requested):

dot/network/block_announce.go
```diff
@@ -108,12 +110,26 @@ func decodeBlockAnnounceMessage(in []byte) (NotificationsMessage, error) {
...
+type Roles byte
+
+const (
+	// FullNode allow you to read the current state of the chain and to submit and validate
+	// extrinsics directly on the network without relying on a centralised infrastructure provider.
+	FullNode Roles = 1
+	// LightClient node has only the runtime and the current state, but does not store past
+	// blocks and so cannot read historical data without requesting it from a node that has it.
+	LightClient Roles = 2
+	// Validator node helps seal new blocks.
```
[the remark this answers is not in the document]
kishansagathiya: Not that I know of. Not that it 3 is getting used anywhere.

dot/network/block_announce.go
```diff
@@ -108,12 +110,26 @@ func decodeBlockAnnounceMessage(in []byte) (NotificationsMessage, error) {
...
 type BlockAnnounceHandshake struct {
-	Roles           byte
+	Roles           Roles
 	BestBlockNumber uint32
 	BestBlockHash   common.Hash
 	GenesisHash     common.Hash
 }
 
+// Roles could be full node, light client or a validator.
+type Roles byte
```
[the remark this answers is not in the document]
kishansagathiya: Unfortunately, I don't know. The word `roles` is used in plural form at all places, including config.toml and even in polkadot code.

kishansagathiya:
Making roles enum does not fix the problem, i still have to check if it is 1, 2 or 4.

qdm12:
> Making roles enum does not fix the problem, i still have to check if it is 1, 2 or 4.

For sure, you can have them as typed byte set to those values (1,2,4), enums won't work (especially since 3 is missing lol)

dot/network/notifications.go
kishansagathiya: What you are saying makes sense. I can not cast it and I can directly pass a msg. But, when we don that we make the assumption that Handshake and NotificationMessage have the same interface. Which they have. But what happen when in future they are not the same interface?

If I were to directly pass the msg, I might as well remove the Handshake interface. But, I don't think we should remove handshake interface. Theoretically, we can remove Handshake interface, but Handshake interface makes the code a lot more readable. It tells us, when handshakes are getting used.

dot/network/block_announce.go
```diff
@@ -183,6 +200,10 @@ func (s *Service) validateBlockAnnounceHandshake(from peer.ID, hs Handshake) err
 		return errors.New("invalid handshake type")
 	}
 
+	if bhs.Roles != FullNode && bhs.Roles != LightClient && bhs.Roles != Validator {
+		return errInvalidRole
+	}
+
```
EclesioMeloJunior, line 206: @qdm12, @kishansagathiya I think there is a pattern in the role numbers `1, 2, 4`, which looks like when we raise 2 to the power of 0, 1, and 2 respectively. So instead of checking for `1, 2 or 4` separately you could do:

```go
type Role byte

const (
	FullNode Roles = 1 << iota // will be evaluated to 1
	LightClient Roles // will be evaluated to 2
	Validator Roles // will be evaluated to 4
)

...
// then in the getBlockAnnounceHandshake method
func (s *Service) getBlockAnnounceHandshake() (Handshake, error) {
  ...
  switch bhs.Roles {
    case 1, 2, 4:
    default:
      return errors.New("inexistent role")
    }
}
```

dot/network/block_announce.go
kishansagathiya: addressed

dot/network/transaction.go
kishansagathiya: addressed

dot/network/host_test.go
```diff
@@ -330,12 +330,11 @@ func TestStreamCloseMetadataCleanup(t *testing.T) {
 	require.NoError(t, err)
 
 	const (
-		roles           byte   = 4
 		bestBlockNumber uint32 = 77
 	)
 
 	testHandshake := &BlockAnnounceHandshake{
-		Roles:           roles,
+		Roles:           common.Validator,
```
qdm12, line 337: nit it would be nice to have a `roles` package somewhere (maybe in the common directory?), so you could call it using `roles.Validator` instead

dot/network/block_announce.go
```diff
@@ -183,6 +184,12 @@ func (s *Service) validateBlockAnnounceHandshake(from peer.ID, hs Handshake) err
 		return errors.New("invalid handshake type")
 	}
 
+	switch bhs.Roles {
+	case common.FullNode, common.LightClient, common.Validator:
+	default:
+		return errInvalidRole
```
qdm12, line 190: nit

```suggestion
		return fmt.Errorf("%w: %d", errInvalidRole, bhs.Roles)
```

dot/network/block_announce.go
```diff
@@ -15,6 +15,7 @@ import (
 	"github.com/libp2p/go-libp2p-core/peer"
 )
 
+var errInvalidRole = errors.New("invalid handshake role")
```
qdm12, line 18: nit should this be renamed to `errHandshakeRoleInvalid` since it's for handshakes? 🤔

dot/services.go
```diff
@@ -280,7 +280,7 @@ func (nodeBuilder) createNetworkService(cfg *Config, stateSrvc *state.Service,
 		LogLvl:            cfg.Log.NetworkLvl,
 		BlockState:        stateSrvc.Block,
 		BasePath:          cfg.Global.BasePath,
-		Roles:             cfg.Core.Roles,
+		Roles:             common.Roles(cfg.Core.Roles),
```
qdm12, line 283: nit do you think we could have the `cfg.Core.Roles` directly as `common.Roles` type?

dot/network/notifications.go
qdm12, line 162: Should we then remove the `, ok` in `hs, ok := msg.(Handshake)` so it panics in case it ever happens?

qdm12 (approved):

dot/network/notifications.go
kishansagathiya: Arhhmm, I am not really sure about that. I would prefer not to remove this.

I feel like NotificationMessage and Handshake happens to have the same interface at the moment, but they don't have to.  Also, not a big fan of panicking.

dot/network/host_test.go
kishansagathiya: I don't think it is a good idea to have a package just for one struct, which is barely 10 lines of code. I could however rename this to `common.ValidatorRole` to improve readability.

kishansagathiya: done

dot/network/block_announce.go
kishansagathiya: changed to `invalid handshake role` instead. `invalid handshake role` didn't really make sense, since role is a property of a node and not a handshake.

dot/services.go
kishansagathiya: with latest commit, roles are getting used everywhere instead of byte. I have propagated these changes til toml config. I would not like to change toml config, since that's what deals with marshalling and marshalling config. I don't want to disturb any behaviour there.

dot/network/notifications.go
qdm12: Then let's remove that comment.

Panicing is useful for programming errors, so it doesn't go unnoticed in some error log line.
I also was team anti-panic, but it really has its use for programming errors (not from user/IO of course)

dot/services.go
qdm12: Having a byte alias probably won't change config parsing. Don't we have unit tests to verify this?

dot/network/host_test.go
qdm12: this is more of a discussion than something to do here -

I agree I don't like 10 lines packages; but `common` / `utils` / `shared` packages are bad package names and they usually bundle a mix bag of things. Not something we may want to do anytime soon, but having meaningful package names would be nice.

EclesioMeloJunior (approved):
LGTM!

qdm12:
> Making roles enum does not fix the problem, i still have to check if it is 1, 2 or 4.

Just so you know, you could use `_` to skip `3` and have enums using iota.

```go
const (
    enum1 uint8 = iota // that's 0
    _
    enum2 // that's 2
)
```

But I prefer to just hardcode the constant as it is now, so really just a shower thought

PR #2683, opened 2022-07-20
feat(scale): add range checks to decodeUint function

### LINKED ISSUE #2629: Bug: payload makes scale decoder go out of memory

edwardmack:
## Changes
- Refactor `decodeUint` to add range checks that check to confirm decoded value is within range for given prefix mode. 
 This replicates behavior of parity scale codec decode functions. 
- Add tests to confirm updated functionality is working as expected.

## Tests

```sh
go test github.com/ChainSafe/gossamer/pkg/scale/
```

## Issues

fixes #2629, which was causing out of memory errors when trying to decode very long bytes encodings.
## Primary Reviewer

@timwu20


qdm12:
Thinking about this, 1GB per malicious message is still risky.
I'm wondering how quickly a peer would be banned if it sends a few 1GB messages at the same time (before program crash).

pkg/scale/decode.go
```diff
@@ -523,10 +524,68 @@ func (ds *decodeState) decodeUint(dstv reflect.Value) (err error) {
...
+	in := dstv.Interface()
+	temp := reflect.New(reflect.TypeOf(in))
+	// check mode of encoding, stored at 2 least significant bits
+	mode := b & 3
+	switch {
+	case mode <= 2:
+		var val int64
+		val, err = ds.decodeSmallInt(b, mode)
+		if err != nil {
+			return
```
qdm12, line 545:
```suggestion
			return fmt.Errorf("decoding small int: %w", err)
```

pkg/scale/decode.go
```diff
@@ -523,10 +524,68 @@ func (ds *decodeState) decodeUint(dstv reflect.Value) (err error) {
...
+		}
+		if mode == 1 {
+			if !(val > 0b0011_1111 && val <= 0b0011_1111_1111_1111) {
+				return errorU32OutOfRange
```
qdm12, line 549:
```suggestion
				return fmt.Errorf("%w: for mode %d and value %d", errorU32OutOfRange, mode, val)
```

pkg/scale/decode.go
```diff
@@ -523,10 +524,68 @@ func (ds *decodeState) decodeUint(dstv reflect.Value) (err error) {
...
+			}
+		}
+		if mode == 2 {
+			if !(val > 0b0011_1111_1111_1111 && val <= math.MaxUint32>>2) {
+				return errorU32OutOfRange
```
qdm12, line 554:
```suggestion
				return fmt.Errorf("%w: for mode %d and value %d", errorU32OutOfRange, mode, val)
```

pkg/scale/decode.go
```diff
@@ -523,10 +524,68 @@ func (ds *decodeState) decodeUint(dstv reflect.Value) (err error) {
...
+		return
+	}
+
+	in := dstv.Interface()
+	temp := reflect.New(reflect.TypeOf(in))
+	// check mode of encoding, stored at 2 least significant bits
+	mode := b & 3
+	switch {
+	case mode <= 2:
+		var val int64
```
qdm12, line 542: nit `value` would be cool 😎

pkg/scale/decode.go
```diff
@@ -523,10 +524,68 @@ func (ds *decodeState) decodeUint(dstv reflect.Value) (err error) {
...
+		if err != nil {
+			return
+		}
+		if mode == 1 {
+			if !(val > 0b0011_1111 && val <= 0b0011_1111_1111_1111) {
+				return errorU32OutOfRange
+			}
+		}
+		if mode == 2 {
+			if !(val > 0b0011_1111_1111_1111 && val <= math.MaxUint32>>2) {
```
qdm12, line 553: nit to remove the `math` import, you can have the max uint32 as a constant with:

```go
const maxUint32 = ^uint32(0)
```

Maybe inlined in the function 😉

pkg/scale/decode.go
```diff
@@ -523,10 +524,68 @@ func (ds *decodeState) decodeUint(dstv reflect.Value) (err error) {
...
+				return errorU32OutOfRange
+			}
+		}
+		temp.Elem().Set(reflect.ValueOf(val).Convert(reflect.TypeOf(in)))
+		dstv.Set(temp.Elem())
+	default:
+		// >4 byte mode
+		topSixBits := b >> 2
+		if mode == 3 && topSixBits != 0 {
+			return errorU32OutOfRange
```
qdm12, line 563:
```suggestion
			return fmt.Errorf("%w: for mode %d and top six bits %08b", 
				errorU32OutOfRange, mode, topSixBits)
```

pkg/scale/decode.go
```diff
@@ -523,10 +524,68 @@ func (ds *decodeState) decodeUint(dstv reflect.Value) (err error) {
...
+		}
+		byteLen := uint(topSixBits) + 4
+
+		buf := make([]byte, byteLen)
+		_, err = ds.Read(buf)
+		if err != nil {
+			return
```
qdm12, line 570:
```suggestion
			return fmt.Errorf("reading from buffer: %w", err)
```

pkg/scale/decode.go
```diff
@@ -523,10 +524,68 @@ func (ds *decodeState) decodeUint(dstv reflect.Value) (err error) {
...
+	default:
+		// >4 byte mode
+		topSixBits := b >> 2
+		if mode == 3 && topSixBits != 0 {
+			return errorU32OutOfRange
+		}
+		byteLen := uint(topSixBits) + 4
+
+		buf := make([]byte, byteLen)
+		_, err = ds.Read(buf)
```
qdm12, line 568: Aren't we just reading 0 bytes from that buffer? 🤔 Or the ds.Read is badly named and should be `Write` instead??

pkg/scale/decode.go
```diff
@@ -523,10 +524,68 @@ func (ds *decodeState) decodeUint(dstv reflect.Value) (err error) {
...
+		if err != nil {
+			return
+		}
+
+		var o uint32
```
qdm12, line 573: Can you name `o` to something more explicit? I'm not sure myself what it's meant to be 🤔

pkg/scale/decode.go
```diff
@@ -523,10 +524,68 @@ func (ds *decodeState) decodeUint(dstv reflect.Value) (err error) {
...
+		if byteLen == 4 {
+			o = binary.LittleEndian.Uint32(buf)
+		}
+		if !(o > math.MaxUint32>>2) {
```
qdm12, line 577:
```suggestion
		if o <= math.MaxUint32>>2 {
```

pkg/scale/decode_test.go
```diff
@@ -302,3 +302,70 @@ func Test_Decoder_Decode_MultipleCalls(t *testing.T) {
 		})
 	}
 }
+
+var (
+	decodeUint32Tests = tests{
```
qdm12, line 307: Please inline this in the relevant (only) test using it.

pkg/scale/decode.go
```diff
@@ -523,10 +524,68 @@ func (ds *decodeState) decodeUint(dstv reflect.Value) (err error) {
 	return
 }
 
+var errorU32OutOfRange = errors.New("uint32 out of range")
+
+// decodeCompactUint32 will decode compact unsigned 32-bit integer
+func (ds *decodeState) decodeCompactUint32(dstv reflect.Value) (err error) {
```
timwu20, line 530: not sure why we need another function for this? Nevertheless `decodeUint` requires the overflow checking as well.  I'll sync up with you to clarify things.

timwu20 (changes requested):

pkg/scale/decode.go
EclesioMeloJunior, line 553:
```suggestion
		if mode == 2 &&  !(val > 0b0011_1111_1111_1111 && val <= math.MaxUint32>>2) {
```

pkg/scale/decode.go
```diff
@@ -523,10 +524,68 @@ func (ds *decodeState) decodeUint(dstv reflect.Value) (err error) {
...
+	mode := b & 3
+	switch {
+	case mode <= 2:
+		var val int64
+		val, err = ds.decodeSmallInt(b, mode)
+		if err != nil {
+			return
+		}
+		if mode == 1 {
+			if !(val > 0b0011_1111 && val <= 0b0011_1111_1111_1111) {
```
EclesioMeloJunior, line 548:
```suggestion
		if mode == 1 && if !(val > 0b0011_1111 && val <= 0b0011_1111_1111_1111) {
```

EclesioMeloJunior (changes requested):

pkg/scale/decode.go
```diff
@@ -523,10 +524,68 @@ func (ds *decodeState) decodeUint(dstv reflect.Value) (err error) {
...
+		dstv.Set(reflect.ValueOf(o).Convert(reflect.TypeOf(in)))
+	}
+	return
+}
+
 // decodeLength is helper method which calls decodeUint and casts to int
 func (ds *decodeState) decodeLength() (l int, err error) {
 	dstv := reflect.New(reflect.TypeOf(l))
-	err = ds.decodeUint(dstv.Elem())
+	err = ds.decodeCompactUint32(dstv.Elem())
```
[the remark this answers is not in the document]
edwardmack: Refactor decodeUint to handle this.

pkg/scale/decode.go
edwardmack: Removed `decodeCompactUint32`, and refactored `decodeUint` to add overflow checking.

pkg/scale/encode_test.go
```diff
@@ -758,9 +763,11 @@ var (
...
 		},
 		{
-			name: "[]int{1 << 32, 2, 3, 1 << 32}",
-			in:   []int{1 << 32, 2, 3, 1 << 32},
-			want: []byte{0x10, 0x07, 0x00, 0x00, 0x00, 0x00, 0x01, 0x08, 0x0c, 0x07, 0x00, 0x00, 0x00, 0x00, 0x01},
+			name: "[]int64{1 << 32, 2, 3, 1 << 32}",
+			in:   []int64{1 << 32, 2, 3, 1 << 32},
+			want: []byte{0x10, 0x00, 0x00, 0x00, 0x00, 0x01, 0x00, 0x00, 0x00, 0x02, 0x00, 0x00, 0x00, 0x00, 0x00,
+				0x00, 0x00, 0x03, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x01, 0x00, 0x00,
+				0x00},
```
edwardmack, line 770: Note: this was changed because after refactoring `decodeUint` it was getting errors, which is consistent with parity scale codec behavior, confirm when testing with:
```
#[test]
	fn decode_vec() {
		let encoded = vec![0x10, 0x07, 0x00, 0x00, 0x00, 0x00, 0x01, 0x08, 0x0c, 0x07, 0x00, 0x00, 0x00, 0x00, 0x01];
		let result = <Vec<u8>>::decode_all(&mut encoded.as_slice());
		println!("result: {:?}", result);
	}
```
Correct encoding determined by parity scale codec:
```
#[test]
	fn vec_and_slice() {
		let slice: &[i64] = &[1 << 32, 2, 3, 1 << 32];
		let data: Vec<i64> = slice.iter().copied().collect();

		let data_encoded = data.encode();
		println!("result: {:?}", data_encoded);
	}
```

pkg/scale/encode_test.go
```diff
@@ -801,9 +808,11 @@ var (
...
 		},
 		{
-			name: "[4]int{1 << 32, 2, 3, 1 << 32}",
-			in:   [4]int{1 << 32, 2, 3, 1 << 32},
-			want: []byte{0x07, 0x00, 0x00, 0x00, 0x00, 0x01, 0x08, 0x0c, 0x07, 0x00, 0x00, 0x00, 0x00, 0x01},
+			name: "[4]int64{1 << 32, 2, 3, 1 << 32}",
+			in:   [4]int64{1 << 32, 2, 3, 1 << 32},
+			want: []byte{0x00, 0x00, 0x00, 0x00, 0x01, 0x00, 0x00, 0x00, 0x02, 0x00, 0x00, 0x00, 0x00, 0x00,
+				0x00, 0x00, 0x03, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x01, 0x00, 0x00,
+				0x00},
```
edwardmack, line 815: Note: this was changed because after refactoring decodeUint it was getting errors, which is consistent with parity scale codec behavior, confirm when testing with:
```
#[test]
	fn decode_vec() {
		let encoded = vec![0x07, 0x00, 0x00, 0x00, 0x00, 0x01, 0x08, 0x0c, 0x07, 0x00, 0x00, 0x00, 0x00, 0x01];
		let result = <Vec<u8>>::decode_all(&mut encoded.as_slice());
		println!("result: {:?}", result);
	}
```

pkg/scale/decode.go
```diff
@@ -478,51 +478,77 @@ func (ds *decodeState) decodeBool(dstv reflect.Value) (err error) {
...
-		var val int64
-		val, err = ds.decodeSmallInt(b, mode)
+	mode := prefix % 4
+	var value uint
+	switch mode {
+	case 0:
+		value = uint(prefix >> 2)
+	case 1:
+		var buf byte
+		buf, err = ds.ReadByte()
```
qdm12, line 498: nit this should do now we don't use a naked return

```suggestion
		buf, err = ds.ReadByte()
```

pkg/scale/decode.go
```diff
@@ -478,51 +478,77 @@ func (ds *decodeState) decodeBool(dstv reflect.Value) (err error) {
...
-		temp.Elem().Set(reflect.ValueOf(val).Convert(reflect.TypeOf(in)))
-		dstv.Set(temp.Elem())
-	default:
-		// >4 byte mode
-		topSixBits := b >> 2
-		byteLen := uint(topSixBits) + 4
-
+		value = uint(binary.LittleEndian.Uint16([]byte{prefix, buf}) >> 2)
+		if value <= 0b0011_1111 || value > 0b0111_1111_1111_1111 {
+			return ErrU64OutOfRange
```
qdm12, line 504: nit add context with it

```suggestion
			return fmt.Errorf("%w: %d (%16b)", ErrU64OutOfRange, value, value)
```

Note I'm not 100% sure `%16b` works but it might just work

pkg/scale/decode.go
```diff
@@ -478,51 +478,77 @@ func (ds *decodeState) decodeBool(dstv reflect.Value) (err error) {
...
+		}
+	case 2:
+		buf := make([]byte, 3)
+		_, err = ds.Read(buf)
+		if err != nil {
+			break
```
qdm12, line 510: Why do we break here? Shouldn't we return an error instead?

pkg/scale/decode.go
```diff
@@ -478,51 +478,77 @@ func (ds *decodeState) decodeBool(dstv reflect.Value) (err error) {
...
+		}
+		value = uint(binary.LittleEndian.Uint32(append([]byte{prefix}, buf...)) >> 2)
+		if value <= 0b0011_1111_1111_1111 || value > uint(maxUint32>>2) {
+			return ErrU64OutOfRange
```
qdm12, line 514: nit add context to the error

```suggestion
			return fmt.Errorf("%w: %d (%32b)", ErrU32OutOfRange, value, value)
```

Note I'm not 100% sure `%32b` works but it might just work

qdm12, line 514: Shouldn't this be `ErrU32OutOfRange`?

pkg/scale/decode.go
qdm12, line 504: Shouldn't this be `ErrU16OutOfRange`?

pkg/scale/decode.go
```diff
@@ -478,51 +478,77 @@ func (ds *decodeState) decodeBool(dstv reflect.Value) (err error) {
...
-
-		var o uint64
-		if byteLen == 4 {
-			o = uint64(binary.LittleEndian.Uint32(buf))
-		} else if byteLen > 4 && byteLen <= 8 {
+		switch byteLen {
+		case 4:
+			value = uint(binary.LittleEndian.Uint32(buf))
+			if value <= uint(maxUint32>>2) {
+				return ErrU64OutOfRange
```
qdm12, line 527: Shouldn't this be `ErrU32OutOfRange`?

pkg/scale/decode.go
```diff
@@ -478,51 +478,77 @@ func (ds *decodeState) decodeBool(dstv reflect.Value) (err error) {
...
+			}
+		case 8:
+			uintSize := 32 << (^uint(0) >> 32 & 1)
+			if uintSize == 32 {
+				return fmt.Errorf("uint64 is not supported")
```
qdm12, line 532: use a sentinel error

pkg/scale/decode.go
```diff
@@ -478,51 +478,77 @@ func (ds *decodeState) decodeBool(dstv reflect.Value) (err error) {
...
+			}
 			tmp := make([]byte, 8)
 			copy(tmp, buf)
-			o = binary.LittleEndian.Uint64(tmp)
-		} else {
-			err = errors.New("could not decode invalid integer")
-			return
+			value = uint(binary.LittleEndian.Uint64(tmp))
+			if value <= uint(maxUint64>>8) {
```
qdm12, line 537: That will fail on 32 bit OS, where `uint` = `uint32`. Can we have value as `uint64` perhaps?

qdm12 (changes requested):

pkg/scale/decode.go
```diff
@@ -478,51 +478,77 @@ func (ds *decodeState) decodeBool(dstv reflect.Value) (err error) {
...
+			if uintSize == 32 {
+				return fmt.Errorf("uint64 is not supported")
+			}
 			tmp := make([]byte, 8)
 			copy(tmp, buf)
-			o = binary.LittleEndian.Uint64(tmp)
-		} else {
-			err = errors.New("could not decode invalid integer")
-			return
+			value = uint(binary.LittleEndian.Uint64(tmp))
```
qdm12, line 536: If `tmp` bytes result in max uint32 + 1 (still valid uint64), `value` will be wrapping around on 32 bit OS since it converts it to `uint32` (and `uint` = `uint32` on 32 bit OS). Can we have value as `uint64` perhaps?

pkg/scale/decode.go
qdm12, line 537: `maxUint64 >> 8` is still larger than max uint32 I think right?
If so, this will fail on 32 bit OS since `uint` = `uint32` and we will convert this uint64 to a uint32, wrapping around (again from zero)

edwardmack: I updated value to be `uint64` and the check for uintSize will return errors for 32 bit OS.

pkg/scale/decode.go
```diff
@@ -478,51 +478,79 @@ func (ds *decodeState) decodeBool(dstv reflect.Value) (err error) {
...
 	return
 }
 
+var ErrU16OutOfRange = errors.New("uint16 out of range")
+var ErrU32OutOfRange = errors.New("uint32 out of range")
+var ErrU64OutOfRange = errors.New("uint64 out of range")
+var ErrU64NotSupported = errors.New("uint64 is not supported")
+
 // decodeLength is helper method which calls decodeUint and casts to int
 func (ds *decodeState) decodeLength() (l int, err error) {
```
timwu20, line 555:
```suggestion
func (ds *decodeState) decodeLength() (l uint, err error) {
```
can you change this to `uint`

timwu20 (changes requested):

pkg/scale/encode_test.go
timwu20: Since we use the same function for a target `uint` in a `struct`, and to a `uint` directly, it will be difficult to error out correctly without refactoring things significantly.  I was trying to figure out how substrate deals with this, but I kinda got lost navigating around.  I would assume this error only comes as a result of using `decode_all`, which implies that all the bytes should be read.  

Looking at the test code in [decode_all.rs](https://github.com/paritytech/parity-scale-codec/blob/9835ab3274821aeae520e43fd441166fd39fc76d/src/decode_all.rs#L81), for struct types you need to handle each attribute specifically when implementing the `Decode` trait.  This differs from our scale package, since it decodes optimistically across the struct attributes. 

I tried adding a test with the `Vec<u8>` as an attribute, and I passed in the bytes that produced that error using `decode_all` and It didn't produce that error when it's a struct attribute.  See `TestStruct2`

```
mod tests {
	use super::*;
	use crate::{Compact, Encode, EncodeLike, Input};

	macro_rules! test_decode_all {
		(
			$( $type:ty => $value:expr; )*
		) => {
			$(
				{
					let mut encoded = <$type as Encode>::encode(&$value);
					<$type>::decode_all(&mut encoded.as_slice()).expect(
						&format!("`{} => {}` decodes all!", stringify!($type), stringify!($value)),
					);

					encoded.extend(&[1, 2, 3, 4, 5, 6]);
					assert_eq!(
						<$type>::decode_all(&mut encoded.as_slice()).unwrap_err().to_string(),
						"Input buffer has still data left after decoding!",
					);
				}
			)*
		};
	}

	#[derive(Debug)]
	struct TestStruct {
		data: Vec<u32>,
		other: u8,
		compact: Compact<u128>,
	}

	impl EncodeLike for TestStruct {}

	impl Encode for TestStruct {
		fn encode(&self) -> Vec<u8> {
			let mut res = Vec::new();
			self.data.encode_to(&mut res);
			self.other.encode_to(&mut res);
			self.compact.encode_to(&mut res);
			res
		}
	}

	impl Decode for TestStruct {
		fn decode<I: Input>(input: &mut I) -> Result<Self, Error> {
			Ok(Self {
				data: Vec::<u32>::decode(input)?,
				other: u8::decode(input)?,
				compact: Compact::<u128>::decode(input)?,
			})
		}
	}

	#[derive(Debug)]
	struct TestStruct2 {
		data: Vec<u8>,
		other: u8,
		compact: Compact<u128>,
	}

	impl EncodeLike for TestStruct2 {}

	impl Encode for TestStruct2 {
		fn encode(&self) -> Vec<u8> {
			let mut res = Vec::new();
			self.data.encode_to(&mut res);
			self.other.encode_to(&mut res);
			self.compact.encode_to(&mut res);
			res
		}
	}

	impl Decode for TestStruct2 {
		fn decode<I: Input>(input: &mut I) -> Result<Self, Error> {
			Ok(Self {
				data: Vec::<u8>::decode(input)?,
				other: u8::decode(input)?,
				compact: Compact::<u128>::decode(input)?,
			})
		}
	}

	#[test]
	fn decode_all_works() {
		test_decode_all! {
			u8 => 120;
			u16 => 30;
			u32 => 1;
			u64 => 2343545;
			u128 => 34358394245459854;
			Vec<u8> => vec![1, 2, 3, 4, 5, 6, 7, 8, 9, 10];
			Vec<u32> => vec![1, 2, 3, 4, 5, 6, 7, 8, 9, 10];
			Compact<u32> => Compact(32445);
			Compact<u128> => Compact(34353454453545);
			TestStruct => TestStruct { data: vec![1, 2, 4, 5, 6], other: 45, compact: Compact(123234545) };
			TestStruct2 => TestStruct2 {
				data: vec![
					0x10, 0x07, 0x00, 0x00, 0x00, 0x00, 0x01, 0x08, 0x0c, 0x07, 0x00, 0x00, 0x00, 0x00,
					0x01,
				],
				other: 45,
				compact: Compact(123234545),
			};
		}
	}
}
```

timwu20 (approved):
Great work @edwardmack!

pkg/scale/encode_test.go
edwardmack: This makes sense, thanks.

pkg/scale/decode.go
```diff
@@ -478,59 +478,87 @@ func (ds *decodeState) decodeBool(dstv reflect.Value) (err error) {
...
-		dstv.Set(temp.Elem())
-	default:
-		// >4 byte mode
-		topSixBits := b >> 2
-		byteLen := uint(topSixBits) + 4
-
+		value = uint64(binary.LittleEndian.Uint16([]byte{prefix, buf}) >> 2)
+		if value <= 0b0011_1111 || value > 0b0111_1111_1111_1111 {
+			return fmt.Errorf("%w: %d (%b)", ErrU16OutOfRange, value, value)
+		}
```
timwu20, line 504: we should probably test this right? codecov/patch isn't a required workflow, but I agree with it here. We should test all the out of range cases.

pkg/scale/decode.go
```diff
@@ -478,59 +478,87 @@ func (ds *decodeState) decodeBool(dstv reflect.Value) (err error) {
 
 // decodeUint will decode unsigned integer
 func (ds *decodeState) decodeUint(dstv reflect.Value) (err error) {
-	b, err := ds.ReadByte()
+	const maxUint32 = ^uint32(0)
+	const maxUint64 = ^uint64(0)
+	prefix, err := ds.ReadByte()
 	if err != nil {
-		return
+		return fmt.Errorf("reading from buffer: %w", err)
```
qdm12, line 485: nit for `Decoder`s, it's just an `io.Reader` so not necessarily a buffer 😉

```suggestion
		return fmt.Errorf("reading byte: %w", err)
```

pkg/scale/decode.go
```diff
@@ -478,59 +478,87 @@ func (ds *decodeState) decodeBool(dstv reflect.Value) (err error) {
...
+	mode := prefix % 4
+	var value uint64
+	switch mode {
+	case 0:
+		value = uint64(prefix >> 2)
+	case 1:
+		buf, err := ds.ReadByte()
 		if err != nil {
-			return
+			return fmt.Errorf("reading from buffer: %w", err)
```
qdm12, line 499: nit

```suggestion
			return fmt.Errorf("reading byte: %w", err)
```

pkg/scale/decode.go
```diff
@@ -478,59 +478,87 @@ func (ds *decodeState) decodeBool(dstv reflect.Value) (err error) {
...
-
+		value = uint64(binary.LittleEndian.Uint16([]byte{prefix, buf}) >> 2)
+		if value <= 0b0011_1111 || value > 0b0111_1111_1111_1111 {
+			return fmt.Errorf("%w: %d (%b)", ErrU16OutOfRange, value, value)
+		}
+	case 2:
+		buf := make([]byte, 3)
+		_, err = ds.Read(buf)
+		if err != nil {
+			return fmt.Errorf("reading from buffer: %w", err)
```
qdm12, line 509: nit for `Decoder`s, it's just an `io.Reader` so not necessarily a buffer. Also we're reading from the `io.Reader` to the 3-bytes-buffer

```suggestion
			return fmt.Errorf("reading bytes: %w", err)
```

pkg/scale/decode.go
```diff
@@ -478,59 +478,87 @@ func (ds *decodeState) decodeBool(dstv reflect.Value) (err error) {
...
+		if value <= 0b0011_1111_1111_1111 || value > uint64(maxUint32>>2) {
+			return fmt.Errorf("%w: %d (%b)", ErrU32OutOfRange, value, value)
+		}
+	case 3:
+		byteLen := (prefix >> 2) + 4
 		buf := make([]byte, byteLen)
 		_, err = ds.Read(buf)
 		if err != nil {
-			return
+			return fmt.Errorf("reading from buffer: %w", err)
```
qdm12, line 520: nit

```suggestion
			return fmt.Errorf("reading bytes: %w", err)
```

pkg/scale/decode.go
```diff
@@ -478,59 +478,87 @@ func (ds *decodeState) decodeBool(dstv reflect.Value) (err error) {
...
-
-		var o uint64
-		if byteLen == 4 {
-			o = uint64(binary.LittleEndian.Uint32(buf))
-		} else if byteLen > 4 && byteLen <= 8 {
+		switch byteLen {
+		case 4:
+			value = uint64(binary.LittleEndian.Uint32(buf))
+			if value <= uint64(maxUint32>>2) {
+				return ErrU32OutOfRange
```
qdm12, line 526: nit

```suggestion
				return fmt.Errorf("%w: %d (%b)", ErrU32OutOfRange, value, value)
```

pkg/scale/decode.go
```diff
@@ -478,59 +478,87 @@ func (ds *decodeState) decodeBool(dstv reflect.Value) (err error) {
...
 	}
+	temp.Elem().Set(reflect.ValueOf(value).Convert(reflect.TypeOf(in)))
+	dstv.Set(temp.Elem())
 	return
 }
 
+var ErrU16OutOfRange = errors.New("uint16 out of range")
+var ErrU32OutOfRange = errors.New("uint32 out of range")
+var ErrU64OutOfRange = errors.New("uint64 out of range")
+var ErrU64NotSupported = errors.New("uint64 is not supported")
```
qdm12, line 552: supernit

```suggestion
var (
	ErrU16OutOfRange = errors.New("uint16 out of range")
	ErrU32OutOfRange = errors.New("uint32 out of range")
	ErrU64OutOfRange = errors.New("uint64 out of range")
	ErrU64NotSupported = errors.New("uint64 is not supported")
)
```

pkg/scale/decode.go
```diff
@@ -478,59 +478,87 @@ func (ds *decodeState) decodeBool(dstv reflect.Value) (err error) {
...
-			o = binary.LittleEndian.Uint64(tmp)
-		} else {
-			err = errors.New("could not decode invalid integer")
-			return
+			value = binary.LittleEndian.Uint64(tmp)
+			if value <= maxUint64>>8 {
+				return ErrU64OutOfRange
+			}
+		default:
+			return fmt.Errorf("unexpected prefix decoding compact uint: %d", prefix)
```
qdm12, line 540: nit use a sentinel error to have `fmt.Errorf("%w: %d", ErrCompactUintPrefixUnknown, prefix)`

together with something like

```go
var ErrCompactUintPrefixUnknown = errors.New("unknown prefix for compact uint")
```

pkg/scale/decode.go
```diff
@@ -478,59 +478,87 @@ func (ds *decodeState) decodeBool(dstv reflect.Value) (err error) {
...
+			}
 			tmp := make([]byte, 8)
 			copy(tmp, buf)
-			o = binary.LittleEndian.Uint64(tmp)
-		} else {
-			err = errors.New("could not decode invalid integer")
-			return
+			value = binary.LittleEndian.Uint64(tmp)
+			if value <= maxUint64>>8 {
+				return ErrU64OutOfRange
```
qdm12, line 537:
```suggestion
				return fmt.Errorf("%w: %d (%b)", ErrU64OutOfRange, value, value)
```

pkg/scale/decode.go
```diff
@@ -478,59 +478,87 @@ func (ds *decodeState) decodeBool(dstv reflect.Value) (err error) {
...
+		case 4:
+			value = uint64(binary.LittleEndian.Uint32(buf))
+			if value <= uint64(maxUint32>>2) {
+				return ErrU32OutOfRange
+			}
+		case 8:
+			uintSize := 32 << (^uint(0) >> 32 & 1)
+			if uintSize == 32 {
+				return ErrU64NotSupported
+			}
```
qdm12, line 532: That's a super interesting way to detect support for uint64! 👍

Tiny nit to set it as a constant

```suggestion
			const uintSize = 32 << (^uint(0) >> 32 & 1)
			if uintSize == 32 {
				return ErrU64NotSupported
			}
```

pkg/scale/decode_test.go
```diff
@@ -11,6 +11,7 @@ import (
 
 	"github.com/google/go-cmp/cmp"
 	"github.com/google/go-cmp/cmp/cmpopts"
+	"github.com/stretchr/testify/assert"
```
qdm12, line 14: Same question as in #2741 @timwu20 can we add those testing deps? Or should we not? Or should we remove them later? 🤔

pkg/scale/decode_test.go
```diff
@@ -302,3 +303,78 @@ func Test_Decoder_Decode_MultipleCalls(t *testing.T) {
...
+			want: []byte{0x10, 0x07, 0x00, 0x00, 0x00, 0x00, 0x01, 0x08, 0x0c, 0x07, 0x00, 0x00, 0x00, 0x00, 0x01},
+		},
+		{
+			name:    "[4]int{1 << 32, 2, 3, 1 << 32}",
+			in:      [4]int{0, 0, 0, 0},
+			want:    []byte{0x07, 0x00, 0x00, 0x00, 0x00, 0x01, 0x08, 0x0c, 0x07, 0x00, 0x00, 0x00, 0x00, 0x01},
+			wantErr: true,
+		},
+	}
+	t.Parallel()
```
qdm12, line 357: nit let's move this at the start of the test

qdm12 (commented):
Just a few minor nit commit suggestions 😉

pkg/scale/decode_test.go
timwu20: Since you wrote the book on go testing here at ChainSafe, I don't mind using testify packages in test code for ChainSafe repos.

qdm12: Actually the testing section is still drafty/wip 😄 So nothing official yet!
To be honest, I like having libraries without any dependency, it adds security/ease of mind for library users.
But then yeah writing tests becomes more of pain. I would suggest we can pay attention to it when migrating scale outside Gossamer later on, it shouldn't be hard to remove the dep on testify packages. (also Go needs dev/test-only dependencies imo)

qdm12:
This blocks a unit test case in #2673 and a unit test case in #2747 (since the error code path cannot be triggered without this PR!)

pkg/scale/decode_test.go
```diff
@@ -302,3 +303,103 @@ func Test_Decoder_Decode_MultipleCalls(t *testing.T) {
...
+			t.Parallel()
+			dst := reflect.New(reflect.TypeOf(tt.in)).Elem().Interface()
+			dstv := reflect.ValueOf(&dst)
+			elem := indirect(dstv)
+
+			buf := &bytes.Buffer{}
+			ds := decodeState{}
+			_, err := buf.Write(tt.want)
+			assert.NoError(t, err)
+			ds.Reader = buf
```
qdm12, line 395: nit

```suggestion
			ds := decodeState{
				Reader: bytes.NewBuffer(tt.want),
			}
```

pkg/scale/decode_test.go
```diff
@@ -302,3 +303,103 @@ func Test_Decoder_Decode_MultipleCalls(t *testing.T) {
...
+			err = ds.decodeUint(elem)
+			if (err != nil) != tt.wantErr {
+				t.Errorf("decodeState.decodeUint error = %v, wantErr %v", err, tt.wantErr)
+			}
+			if !reflect.DeepEqual(dst, tt.in) {
+				t.Errorf("decodeState.decodeUint = %v, want %v", dst, tt.in)
+			}
```
qdm12, line 402:
```suggestion
			if tt.wantErr {
				assert.Error(t, err)
			} else {
				assert.NoError(t, err)
			}
			assert.Equal(t, tt.in, dst)
```

qdm12 (approved):

EclesioMeloJunior (approved):

PR #2685, opened 2022-07-20
chore(lib/runtime/wasmer): common `setupVM` function

qdm12:
## Changes

The main goal (not yet achieved) is to get the runtime version from the wasm code byte slice **only**.
This moves one step closer, by having a shared `setupVM` function taking in only the wasm code.

- Fix: decompress WASM when checking runtime version
- Fix: clear allocator when closing instance
- Fix: add mutex protection for the entire `UpdateRuntimeCode` body
  - Add and use new `close` method on `Instance` (without mutex)
  - Use `close` method in `Stop` method
- `setupVM` only takes in the WASM code bytes and returns an instance+allocator
- WASM instance context data is set outside setupVM
- Minor changes
  - Sentinel error for 'code is empty'
  - Remove useless debug log (just logs pointer addresses)
  - `Instance` named field `mutex` to find references easily
  - Explicit runtime `Stop` call in `dot/state/initialize.go`

## Issues

#2418 

## Primary Reviewer

@timwu20

### DIFF
--- a/lib/runtime/wasmer/instance.go
+++ b/lib/runtime/wasmer/instance.go
@@ -138,12 +138,12 @@ func (in *Instance) GetContext() *runtime.Context {
 
 // UpdateRuntimeCode updates the runtime instance to run the given code
 func (in *Instance) UpdateRuntimeCode(code []byte) (err error) {
-	in.Stop() // TODO move after setupVM in case setup fails
-
 	wasmInstance, allocator, err := setupVM(code)
 	if err != nil {
 		return fmt.Errorf("setting up VM: %w", err)
 	}
+
+	in.Stop()
 
 	in.ctx.Allocator = allocator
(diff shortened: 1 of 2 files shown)


lib/runtime/wasmer/instance.go
```diff
@@ -43,7 +43,7 @@ type Instance struct {
 	ctx      *runtime.Context
 	isClosed bool
 	codeHash common.Hash
-	sync.Mutex
+	mutex    sync.Mutex
```
qdm12, line 46: Note this was changed since I was having trouble finding references to the Lock/Unlock methods.

As a side note, I think we should avoid embedding things due to that, and also because it exports methods we might not want exported actually (like here, the mutex should be private).

EclesioMeloJunior (approved):

timwu20 (approved):

PR #2686, opened 2022-07-21
chore(lib/runtime/wasmer): refactor `Exec` and `exec` methods

qdm12:
## Changes

Simple refactor to adapt `exec` to be more independent, in order to have an instance-independent check runtime function.

- Remove nil storage check for temporary instances (some calls don't need it, and anyway we should panic if this happens imo)
- Fix for bigger than max int32 pointer and data length (before > int32 pointers/data length would crash it)
- Minor simplifications
  - Only use `Exec` instead of both directly wrapping `exec`
  - Inline single/2-lines functions content in `exec`: `malloc`, `clear`, `store`, `load`
  - Add sentinel errors and error wrappings
  - Add named returns

## Issues

#2418 

## Primary Reviewer

@EclesioMeloJunior

### DIFF
--- a/lib/runtime/errors.go
+++ b/lib/runtime/errors.go
@@ -19,6 +19,3 @@ var ErrInvalidTransaction = &json2.Error{Code: 1010, Message: "Invalid Transacti
 // ErrUnknownTransaction is returned if the call to runtime function
 // TaggedTransactionQueueValidateTransaction fails with value of [1, 1, x]
 var ErrUnknownTransaction = &json2.Error{Code: 1011, Message: "Unknown Transaction Validity"}
-
-// ErrNilStorage is returned when the runtime context storage isn't set
-var ErrNilStorage = errors.New("runtime context storage is nil")
--- a/lib/runtime/wasmer/instance.go
+++ b/lib/runtime/wasmer/instance.go
@@ -264,12 +264,6 @@ func (in *Instance) Stop() {
 	}
 }
 
-// Store func
-func (in *Instance) store(data []byte, location int32) {
-	mem := in.vm.Memory.Data()
-	copy(mem[location:location+int32(len(data))], data)
-}
-
 // Load load
 func (in *Instance) load(location, length int32) []byte {
 	mem := in.vm.Memory.Data()
@@ -285,7 +279,8 @@ func (in *Instance) Exec(function string, data []byte) ([]byte, error) {
 		return nil, errors.New("instance is stopped")
 	}
 
-	ptr, err := in.ctx.Allocator.Allocate(uint32(len(data)))
+	dataLength := uint32(len(data))
+	ptr, err := in.ctx.Allocator.Allocate(dataLength)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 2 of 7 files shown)


jimjbrettj:
Would error -> graceful shutdown be better than panicking?

lib/runtime/errors.go
```diff
@@ -19,6 +19,3 @@ var ErrInvalidTransaction = &json2.Error{Code: 1010, Message: "Invalid Transacti
 // ErrUnknownTransaction is returned if the call to runtime function
 // TaggedTransactionQueueValidateTransaction fails with value of [1, 1, x]
 var ErrUnknownTransaction = &json2.Error{Code: 1011, Message: "Unknown Transaction Validity"}
-
-// ErrNilStorage is returned when the runtime context storage isn't set
-var ErrNilStorage = errors.New("runtime context storage is nil")
```
jimjbrettj, line 24: Nice! Im deleting this file in my PR (I'm getting rid of other errors)

lib/runtime/wasmer/instance.go
```diff
@@ -264,67 +264,45 @@ func (in *Instance) Stop() {
 	}
 }
 
-// Store func
-func (in *Instance) store(data []byte, location int32) {
```
jimjbrettj, line 268: why did we want to remove these functions?

lib/runtime/wasmer/instance.go
```diff
@@ -264,67 +264,45 @@ func (in *Instance) Stop() {
...
+	copy(memory[inputPtr:inputPtr+dataLength], data)
 
 	runtimeFunc, ok := in.vm.Exports[function]
 	if !ok {
-		return nil, fmt.Errorf("could not find exported function %s", function)
+		return nil, fmt.Errorf("%w: %s", ErrExportFunctionNotFound, function)
 	}
 
-	res, err := runtimeFunc(int32(ptr), datalen)
+	wasmValue, err := runtimeFunc(int32(inputPtr), int32(dataLength))
```
jimjbrettj, line 298: good work

jimjbrettj (commented):
looks good just few questions

lib/runtime/wasmer/instance.go
qdm12: That was me trying to understand everything that was needed in that `Exec` method, and those tiny methods made it just hard to understand tbh.

And these 1 or 2-lines functions were only used once (in prod code, and once in a test), so I just inlined them.

So I took the EXECUTIVE DECISION to inline them pretty much 😄

timwu20 (approved):

qdm12:
@jimjbrettj

> Would error -> graceful shutdown be better than panicking?

Well in standard Go, panic should really be used only for programming errors.
But in this case, we have no way to return errors given the CGO function signature, so we agreed to panic in case of errors. Anyway these errors are very unlikely except if for example we run out of memory or our disk is physically bad for example, so it's fine to panic. Maybe one day once our service architecture is reworked, we could signal through a channel on the instance we encountered a fatal error and recover from it... but not today 😄

PR #2687, opened 2022-07-21
chore(lib/runtime): `CheckRuntimeVersion` independent from existing runtime instance

qdm12:
## Changes

Blocked by #2686 and #2685

- `CheckRuntimeVersion` as function instead of instance method
  - No dependency on existing instance
  - Fix: do not modify parent instance allocator
  - Fix: close temporary instance on exit
  - Remove `CheckRuntimeVersion` from Instance interface
- `ext_misc_runtime_version_version_1` uses `CheckRuntimeVersion`

## Issues

#2418 

## Primary Reviewer

@timwu20

### DIFF
--- a/lib/runtime/interface.go
+++ b/lib/runtime/interface.go
@@ -16,7 +16,6 @@ import (
 // Instance is the interface a v0.8 runtime instance must implement
 type Instance interface {
 	UpdateRuntimeCode([]byte) error
-	CheckRuntimeVersion([]byte) (Version, error)
 	Stop()
 	NodeStorage() NodeStorage
 	NetworkService() BasicNetwork
--- a/lib/runtime/wasmer/instance.go
+++ b/lib/runtime/wasmer/instance.go
@@ -156,9 +156,9 @@ func (in *Instance) UpdateRuntimeCode(code []byte) (err error) {
 	return nil
 }
 
-// CheckRuntimeVersion finds the runtime version by initiating a temporary
+// GetRuntimeVersion finds the runtime version by initiating a temporary
 // runtime instance using the WASM code provided, and querying it.
-func CheckRuntimeVersion(code []byte) (version runtime.Version, err error) {
+func GetRuntimeVersion(code []byte) (version runtime.Version, err error) {
 	config := runtime.InstanceConfig{
 		LogLvl: log.DoNotChange,
 	}
--- a/lib/runtime/wasmer/instance.go
+++ b/lib/runtime/wasmer/instance.go
@@ -156,25 +156,31 @@ func (in *Instance) UpdateRuntimeCode(code []byte) (err error) {
 	return nil
 }
 
-// CheckRuntimeVersion calculates runtime Version for runtime blob passed in
-func (in *Instance) CheckRuntimeVersion(code []byte) (runtime.Version, error) {
-	in.mutex.Lock()
-	defer in.mutex.Unlock()
-
+// CheckRuntimeVersion finds the runtime version by initiating a temporary
+// runtime instance using the WASM code provided, and querying it.
+func CheckRuntimeVersion(code []byte) (version runtime.Version, err error) {
 	wasmInstance, allocator, err := setupVM(code)
 	if err != nil {
 		return nil, fmt.Errorf("setting up VM: %w", err)
 	}
 
-	in.ctx.Allocator = allocator // TODO we should no change the allocator of the parent instance
-	wasmInstance.SetContextData(in.ctx)
+	ctx := &runtime.Context{
+		Allocator: allocator,
+	}
+	wasmInstance.SetContextData(ctx)
 
 	instance := Instance{
 		vm:  wasmInstance,
-		ctx: in.ctx,
+		ctx: ctx,
+	}
+	defer instance.close()
+
+	version, err = instance.Version()
+	if err != nil {
+		return nil, fmt.Errorf("running runtime: %w", err)
 	}
 
-	return instance.Version()
+	return version, nil
 }
 
 var (
--- a/lib/runtime/wasmer/instance.go
+++ b/lib/runtime/wasmer/instance.go
@@ -159,21 +159,14 @@ func (in *Instance) UpdateRuntimeCode(code []byte) (err error) {
 // CheckRuntimeVersion finds the runtime version by initiating a temporary
 // runtime instance using the WASM code provided, and querying it.
 func CheckRuntimeVersion(code []byte) (version runtime.Version, err error) {
-	wasmInstance, allocator, err := setupVM(code)
+	config := runtime.InstanceConfig{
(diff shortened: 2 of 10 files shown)


lib/runtime/interface.go
```diff
@@ -16,7 +16,6 @@ import (
 // Instance is the interface a v0.8 runtime instance must implement
 type Instance interface {
 	UpdateRuntimeCode([]byte) error
-	CheckRuntimeVersion([]byte) (Version, error)
```
timwu20, line 19: I understand why you removed this function from the interface since it creates a new instance to do the version check, but how will we mock this out now? We'd have to change the function signatures of all callers of this function to include a `func([]byte) (Version, error)`

timwu20: I don't think it's fast at all.  It's instantiating a wasmer instance, loading the code in, binding the imports, etc.  Now all of the unit tests are runtime specific if we remove it from the `Instance` interface.  

Why aren't we just implementing a `RuntimeVersion() (Version, error)` (no blob) on `Instance`.  Then anyone who needs to version check needs to instantiate an instance via `NewInstance`.

qdm12: It takes 500ms to load the version, so not that fast indeed.
**But** if you have a look, there was zero place we were configuring the `CheckRuntimeVersion` mock expectation, so it won't change a thing with our code as it is. And as I mentioned above, if in the future we want to mock the runtime check (for test coverage or testing speed), we can just set it as a field on the receiving struct (like we would with `timeNow func() time.Time` to mock time).

timwu20: I looked at the interface again.  The `Instance` interface includes a `Version() (Version, error)` method.  I think we should remove `CheckRuntimeVersion` from the runtime package and make this a private method in `dot/state` and a helper function for the tests. What do you think?

qdm12: Good point, although that would require to export a bunch of things such as `setupVM`, the wasmer `Instance` struct fields `vm` and `ctx`, and I think ultimately we should keep this in the wasmer package since it's quite tightly coupled with wasmer thingies? I would rather export `CheckRuntimeVersion` than these other function/fields, what do you think?

timwu20: why do you need to export those private functions? To check a version of a given wasm blob, you instantiate an instance by calling `NewInstance(code []byte, cfg runtime.InstanceConfig)` and call `Version()` on that instance.

qdm12: Good point, sorry I misunderstood.
I changed `CheckRuntimeVersion` to use `NewInstance`, simplifying it a bit / reducing code duplication (although it sets up extra stuff it doesn't need, but that's a dep injection problem for another day).

On the other hand, `ext_misc_runtime_version_version_1` in `wasmer/imports.go` uses `CheckRuntimeVersion` so I think we should still keep it in the wasmer package? 🤔

jimjbrettj (approved):

lib/runtime/interface.go
timwu20: I have dissonance with having a function on the public package API that achieves the same functionality that can be achieved by instantiating an instance and calling a receiver method on that instance.  But that would lead to code duplication cause we use it in various places.  I gave it a shot in this [commit](https://github.com/ChainSafe/gossamer/commit/29e178ea16ce1caee7fbc775b9549172f315d90e), and I'm not really happy with it.  So I'm ok with leaving with leaving it in the `wasmer` package.

lib/runtime/wasmer/instance.go
```diff
@@ -156,25 +156,24 @@ func (in *Instance) UpdateRuntimeCode(code []byte) (err error) {
...
 
-// CheckRuntimeVersion calculates runtime Version for runtime blob passed in
-func (in *Instance) CheckRuntimeVersion(code []byte) (runtime.Version, error) {
-	in.mutex.Lock()
-	defer in.mutex.Unlock()
-
-	wasmInstance, allocator, err := setupVM(code)
+// CheckRuntimeVersion finds the runtime version by initiating a temporary
+// runtime instance using the WASM code provided, and querying it.
+func CheckRuntimeVersion(code []byte) (version runtime.Version, err error) {
```
timwu20, line 161:
```suggestion
func GetRuntimeVersion(code []byte) (version runtime.Version, err error) {
```
even though I don't like `GetXXX` syntax in go, I think it's better than `Check..` since it's not checking anything.

timwu20 (changes requested):

lib/runtime/wasmer/instance.go
qdm12: Or `QueryRuntimeVersion`?

noot, line 161: like @timwu20 mentioned there isn't really a need for this function to remain in the runtime/wasmer package anymore, might make more sense for it to be in dot/state

qdm12:
> I have dissonance with having a function on the public package API that achieves the same functionality that can be achieved by instantiating an instance and calling a receiver method on that instance.

If it's stateless, I would say it's fine to have as a 'pure' function (without interfacing etc.). The caller can always have it as a function struct field (or other).

> But that would lead to code duplication cause we use it in various places. I gave it a shot in this [commit](https://github.com/ChainSafe/gossamer/commit/29e178ea16ce1caee7fbc775b9549172f315d90e), and I'm not really happy with it. So I'm ok with leaving with leaving it in the wasmer package.
> there isn't really a need for this function to remain in the runtime/wasmer package anymore, might make more sense for it to be in dot/state

The thing is it calls `NewInstance` which is wasmer specific, so it kind of belongs in the wasmer package I would say. And we also don't want code duplication (especially when it adds more package dependency).

Also a thought which is quite a bit out of scope, since it looks so far we only want to support 1 wasm runtime, I feel it would be nice some day some time to merge `lib/runtime` and `lib/runtime/wasmer` together (i.e. `QueryVersion` would be nice from `lib/runtime`); again just a shower thought really.

lib/runtime/wasmer/instance.go
qdm12: Actually I'm setting it as `GetRuntimeVersion` since a bunch of other methods are also called `GetRuntimeVersion`, for consistency. Maybe we can rename all of them to `QueryRuntimeVersion` later.

timwu20 (approved):

lib/runtime/wasmer/instance.go
qdm12: Kept it there to avoid replicating 3 times (lib/runtime/wasmer, dot/state, and somewhere else too) as per what Tim described.

Also it might get a bit more complex due to `Core_version` being deprecated, and we won't need to create an instance anymore for newer runtimes, so we should definitely have it in one place. (see https://github.com/paritytech/substrate/pull/8688)

PR #2696, opened 2022-07-25
Separate Handshake and Notification message interfaces and other tiny changes

### LINKED ISSUE #2437: dot/network: refactor `createNotificationsMessageHandle` into smaller functions
## Issue summary

- It would be nice to break up `createNotificationsMessageHandler` into two smaller functions for handshake and notification message
- right now, it is easier to confuse between notification message and handshake. They both use the same interface. It might be worth it to separate them, that could potentially make it easier for errors to get spot. Hash() method is never implemented in case of Handshake.
https://github.com/ChainSafe/gossamer/blob/7ab31f0c84b16df6565f98553dd9fc6f9ae39e88/dot/network/block_announce.go#L148-L150

And then we have `isHandshake()` method differentiate between them. Same interface looks forced than natural here. We should split it into two different interfaces.

- https://github.com/ChainSafe/gossamer/blob/7ab31f0c84b16df6565f98553dd9fc6f9ae39e88/dot/network/notifications.go#L258
Don't think we should be allowing for message cache to be nil. We should throw and error if it is nil. 
- Should change this `...` comments to something more meaningful. They are at a bunch of places.
https://github.com/ChainSafe/gossamer/blob/7ab31f0c84b16df6565f98553dd9fc6f9ae39e88/dot/network/block_announce.go#L147
- I think more ideas will start coming on the task
- https://github.com/ChainSafe/gossamer/pull/2435#discussion_r835161686

kishansagathiya:
## Changes
- Separated Handshake and Notification message interfaces. Removed Hash method from Handshake interface. It was not being used anyway. This would create fewer opportunities where code would get confused between those two. We would not need to cast one interface into other anymore.
- Broke `createNotificationsMessageHandler` into two functions
- Some extra comments

## Tests

```sh
```

## Issues
fixes #2437 

## Primary Reviewer

@timwu20


dot/network/block_announce.go
```diff
@@ -18,7 +18,7 @@ import (
 var errInvalidRole = errors.New("invalid role")
 var (
 	_ NotificationsMessage = &BlockAnnounceMessage{}
-	_ NotificationsMessage = &BlockAnnounceHandshake{}
+	_ Handshake            = &BlockAnnounceHandshake{}
```
qdm12, line 21: out of scope nit, but you can change to have zero allocation with

```suggestion
	_ Handshake            = (*BlockAnnounceHandshake)(nil)
```

dot/network/block_announce.go
```diff
@@ -138,20 +138,14 @@ func (*BlockAnnounceHandshake) Type() byte {
...
 }
 
-// Hash returns blake2b hash of block announce handshake.
-func (hs *BlockAnnounceHandshake) Hash() (common.Hash, error) {
-	// scale encode each extrinsic
-	encMsg, err := hs.Encode()
-	if err != nil {
-		return common.Hash{}, fmt.Errorf("cannot encode handshake: %w", err)
+// IsHandshake returns true if handshakes's role is valid.
+func (hs *BlockAnnounceHandshake) IsHandshake() bool {
```
qdm12, line 142: Since `BlockAnnounceHandshake` is named with `Handshake`, shouldn't it always be a handshake?

dot/network/notifications.go
```diff
@@ -128,8 +133,13 @@ func (s *Service) createNotificationsMessageHandler(
 			ok   bool
 			msg  NotificationsMessage
 			peer = stream.Conn().RemotePeer()
+			hs   Handshake
 		)
 
+		if hs, ok = m.(Handshake); ok && hs.IsHandshake() {
```
qdm12, line 139: That feels weird to do an interface assertion and then check if it's a handshake again.

We should do one or the other, not both

dot/network/notifications.go
```diff
@@ -226,6 +180,56 @@ func (s *Service) createNotificationsMessageHandler(
...
+	hs Handshake, peer peer.ID) error {
+	logger.Tracef("received handshake on notifications sub-protocol %s from peer %s, message is: %s",
+		info.protocolID, stream.Conn().RemotePeer(), hs)
+
+	// if we are the receiver and haven't received the handshake already, validate it
+	// note: if this function is being called, it's being called via SetStreamHandler,
+	// ie it is an inbound stream and we only send the handshake over it.
+	// we do not send any other data over this stream, we would need to open a new outbound stream.
+	hsData := info.peersData.getInboundHandshakeData(peer)
+	if hsData == nil {
```
qdm12, line 193: nit (a bit out of scope but since these are deltas now) flip the if condition around to return early and have less indentation

dot/network/transaction.go
```diff
@@ -17,7 +17,7 @@ import (
 
 var (
 	_ NotificationsMessage = &TransactionMessage{}
-	_ NotificationsMessage = &transactionHandshake{}
+	_ Handshake            = &transactionHandshake{}
```
qdm12, line 20: nit

```suggestion
	_ Handshake            = (*transactionHandshake)(nil)
```

qdm12 (approved):

EclesioMeloJunior (changes requested):

dot/network/notifications.go
timwu20: Yea do we still need `IsHandshake()` if we can do the type assertions?

dot/network/block_announce.go
kishansagathiya: Renaming this function to `IsValidHandshake` to make things more clear

dot/network/notifications.go
kishansagathiya: > That feels weird to do an interface assertion and then check if it's a handshake again.

`IsHandshake` not only checks the type but checks if it is a valid handshake. I am going to rename `IsHandshake` to `IsValidHandshake`, that would make things more clear.

dot/network/notifications.go
```diff
@@ -226,6 +179,58 @@ func (s *Service) createNotificationsMessageHandler(
...
+		return errCannotValidateHandshake
+	}
+
+	hsData.validated = true
+	info.peersData.setInboundHandshakeData(peer, hsData)
+
+	// once validated, send back a handshake
+	resp, err := info.getHandshake()
+	if err != nil {
+		logger.Warnf("failed to get handshake using protocol %s: %s", info.protocolID, err)
```
qdm12, line 215: nit I know it's copy pasta, but since it shows as deltas, one more suggestion...

Why is this logged at the warn level whereas others are logged at the trace level? Would it feel appropriate to remove this log and rely on the caller Trace log instead?



```suggestion
```

qdm12 (commented):
Nice changes 👍 👍 

Just a few nits (most should just be easy-to-batch-suggestion-commit) about the copied pasted code since it shows as deltas.

dot/network/notifications.go
kishansagathiya: > Would it feel appropriate to remove this log and rely on the caller Trace log instead?

I prefer not to ignore errors. If we are not returning the err, at least we should log it as warn.

lib/grandpa/network.go
```diff
@@ -63,8 +63,8 @@ func (*GrandpaHandshake) Hash() (common.Hash, error) {
 	return common.Hash{}, nil
 }
 
-// IsHandshake returns true
-func (*GrandpaHandshake) IsHandshake() bool {
+// IsValidHandshake returns true
+func (*GrandpaHandshake) IsValidHandshake() bool {
```
EclesioMeloJunior, line 67: would be nice to have a similar check here since the `GrandpaHandshake` should come from other authority peers
```go
switch hs.Roles {
	case common.AuthorityRole:
		return true
	default:
		return false
	}
```

EclesioMeloJunior (changes requested):
LGTM, just a small adjustment

dot/network/notifications.go
```diff
@@ -20,7 +20,9 @@ const handshakeTimeout = time.Second * 10
 
 // Handshake is the interface all handshakes for notifications protocols must implement
 type Handshake interface {
-	NotificationsMessage
+	Message
+	Type() byte
+	IsValidHandshake() bool
```
timwu20, line 25: do we even need an interface for this anymore? Could we not do type checks on the types that are handshake messages? Or even use the constraint feature in go generics.

dot/network/message.go
```diff
@@ -36,7 +36,7 @@ type NotificationsMessage interface {
 	Message
 	Type() byte
 	Hash() (common.Hash, error)
-	IsHandshake() bool
+	IsValidHandshake() bool
```
timwu20, line 39: Do we still need an interface for this? Could this be refactored further using generics? As it stands, all handshake messages implement every method of this interface except `Hash()`.

dot/network/block_announce.go
```diff
@@ -18,7 +18,7 @@ import (
 var errInvalidRole = errors.New("invalid role")
 var (
 	_ NotificationsMessage = &BlockAnnounceMessage{}
-	_ NotificationsMessage = &BlockAnnounceHandshake{}
+	_ Handshake            = (*BlockAnnounceHandshake)(nil)
```
timwu20, line 21: why the nil case?

dot/network/transaction.go
```diff
@@ -17,7 +17,7 @@ import (
 
 var (
 	_ NotificationsMessage = &TransactionMessage{}
-	_ NotificationsMessage = &transactionHandshake{}
+	_ Handshake            = (*transactionHandshake)(nil)
```
timwu20, line 20: why the nil case?

dot/network/block_announce.go
kishansagathiya: This is a check to make sure that BlockAnnounceHandshake implements Handshake interface. I think, Unlike the previous times, with nil, we end up not allocating any memory.

dot/network/transaction.go
kishansagathiya: This is a check to make sure that transactionHandshake implements Handshake interface. I think, Unlike the previous times, with nil, we end up not allocating any memory.

dot/network/notifications.go
kishansagathiya: `IsValidHandshake` validates the handshake as well. It does more that type checking.

kishansagathiya:
@timwu20 There was a opportunity to do more refactor. Removed IsValidHandshake() from NotificationsMessage Interface and removed Type() from Handshake interface. Bother interfaces are different now (can't be casted into each other)

lib/grandpa/network.go
kishansagathiya: addressed

EclesioMeloJunior (approved):

timwu20 (approved):

PR #2697, opened 2022-07-25
feat(build): binary built-in timezone data

qdm12:
## Changes

- Avoid relying on OS time zone data to provide correct times

## Primary Reviewer

@jimjbrettj

### DIFF
--- a/cmd/gossamer/main.go
+++ b/cmd/gossamer/main.go
@@ -7,6 +7,7 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	_ "time/tzdata"
 
 	"github.com/ChainSafe/gossamer/dot"
 	"github.com/ChainSafe/gossamer/dot/state"


jimjbrettj (approved):
looks good but to clarify, this is only imported for its side effects?

qdm12:
> looks good but to clarify, this is only imported for its side effects?

Yep. Essentially to have timezoned time, you need timezone data, which usually comes on the OS, but not always the case on Docker images (such as `debian` or `alpine` or `scratch`, to keep their base size small).

For example without that timezone data (aka `tzdata`), logs would show at GMT always even though you set `TZ=America/New York`, but will log in the NY timezone if that data is installed.

More on it: https://pkg.go.dev/time/tzdata

Why? No idea, but adding #2695 reminded me of that one, which might turn useful in some corner cases / docker stuff.

timwu20 (approved):

PR #2709, opened 2022-07-28
don't throw ErrProducerEquivocated if authored blocks have different parents

### LINKED ISSUE #2682: could not verify slot claim VRF proof
## Describe the bug

- Run two substrate nodes (using substrate node template) and one gossamer node
- Wait for a few blocks somewhere between 40 to 100. You sometimes see

```
2022-07-19T16:36:37+05:30 EROR block data processing for block with hash 0xe5956255bd91183531280c4b69209787697ff3b09b6ec1098887b1326dda2c2d failed: failed to verify pre-runtime digest: could not verify slot claim VRF proof	chain_processor.go:L86	pkg=sync
```

## Log output

https://gist.github.com/kishansagathiya/b951113719a5bc62748e7e622b5342e5

kishansagathiya:
## Changes
Waiting on some spec related clarification

## Tests

```sh
```

## Issues
Fixes #2682

## Primary Reviewer

@timwu20


lib/babe/verify.go
```diff
@@ -372,6 +372,7 @@ func (b *verifier) verifyAuthorshipRight(header *types.Header) error {
 		// same authority won't produce two different blocks at the same block number as primary block producer
 		if currentBlockProducerIndex == existingBlockProducerIndex &&
 			!currentHash.Equal(header.Hash()) &&
+			currentHeader.ParentHash.Equal(header.ParentHash) &&
```
qdm12, line 375: nit maybe assign this boolean to a named var for documentation purposes

i.e.

```go
authoredBlocksHaveSameParent := currentHeader.ParentHash.Equal(header.ParentHash)
if ... &&
   authoredBlocksHaveSameParent &&
   ... {

}
```

qdm12 (approved):

qdm12:
Also would you be able to add a test case to cover such condition?
Or are unit tests messy such that's it's not easily feasible?

noot (commented):
i think the whole equivocation check is wrong, afaik equivocation for block producers is defined as authoring two blocks in the same slot. we're allowed to author multiple blocks w the same number as long as they're in different slots, but the check is currently checking for multiple blocks w the same number, instead of multiple blocks in the same slot

kishansagathiya:
> afaik equivocation for block producers is defined as authoring two blocks in the same slot.

Seems right. Will edit equivocation check accordingly.

https://github.com/paritytech/substrate/blob/8e600ee0515122667cd091d632104708d343ca34/client/consensus/slots/src/aux_schema.rs#L53

kishansagathiya:
@timwu20 made it draft, have to re-do equivocation logic after noot's comments

lib/babe/verify.go
```diff
@@ -327,62 +327,58 @@ func (b *verifier) verifyAuthorshipRight(header *types.Header) error {
 		return ErrBadSignature
 	}
 
-	// check if the producer has equivocated, ie. have they produced a conflicting block?
-	// hashes is hashes of all blocks with same block number as header.Number
-	hashes := b.blockState.GetAllBlocksAtDepth(header.ParentHash)
+	equivocated, err := b.verifyBlockEquivocation(header)
+	if err != nil {
+		return fmt.Errorf("could not verify block equivocation for header %s: %w", header.Hash(), err)
```
qdm12, line 332: header should be part of the error context set by verifyBlockEquivocation, not the caller (to avoid repeating it):

```suggestion
		return fmt.Errorf("could not verify block equivocation: %w", err)
```

lib/babe/verify.go
```diff
@@ -327,62 +327,58 @@ func (b *verifier) verifyAuthorshipRight(header *types.Header) error {
...
+	}
+	if equivocated {
+		return ErrProducerEquivocated
```
qdm12, line 335: nit to force using errors.Is(err, ErrGenesisHeader) instead of err == ErrGenesisHeader (as recommended on the Go blog):

```suggestion
		return fmt.Errorf("%w", ErrProducerEquivocated)
```

Also it would be nice to add some context, i.e. the block header hash I guess?

lib/babe/verify.go
```diff
@@ -327,62 +327,58 @@ func (b *verifier) verifyAuthorshipRight(header *types.Header) error {
...
-		currentBlockProducerIndex, err := getAuthorityIndex(currentHeader)
-		if err != nil {
-			return fmt.Errorf("failed to get authority index %s", err)
-		}
+// verifyBlockEquivocation checks if given block's author has occupied corresponding slot more than once.
+// It returns true if block was equivocated.
+func (b *verifier) verifyBlockEquivocation(header *types.Header) (bool, error) {
+	author, err := getAuthorityIndex(header)
+	if err != nil {
+		return false, fmt.Errorf("failed to get authority index for %s: %w", header.Hash(), err)
```
qdm12, line 346: header should be part of the error context set by getAuthorityIndex, not the caller (to avoid repeating it):

```suggestion
		return false, fmt.Errorf("failed to get authority index: %w", err)
```

lib/babe/verify.go
```diff
@@ -327,62 +327,58 @@ func (b *verifier) verifyAuthorshipRight(header *types.Header) error {
...
-			return fmt.Errorf("failed get header %s", err)
-		}
+	return nil
+}
 
-		currentBlockProducerIndex, err := getAuthorityIndex(currentHeader)
-		if err != nil {
-			return fmt.Errorf("failed to get authority index %s", err)
-		}
+// verifyBlockEquivocation checks if given block's author has occupied corresponding slot more than once.
```
qdm12, line 341:
```suggestion
// verifyBlockEquivocation checks if the given block's author has occupied the corresponding slot more than once.
```

lib/babe/verify.go
```diff
@@ -327,62 +327,58 @@ func (b *verifier) verifyAuthorshipRight(header *types.Header) error {
...
+// It returns true if block was equivocated.
+func (b *verifier) verifyBlockEquivocation(header *types.Header) (bool, error) {
```
qdm12, line 343: nit I'd suggest naming your return so there is no need to mention it returns true if the the block is equivocated:

```suggestion
func (b *verifier) verifyBlockEquivocation(header *types.Header) (equivocated bool, err error) {
```

lib/babe/verify.go
```diff
@@ -327,62 +327,58 @@ func (b *verifier) verifyAuthorshipRight(header *types.Header) error {
...
+	if err != nil {
+		return false, fmt.Errorf("failed to get authority index for %s: %w", header.Hash(), err)
+	}
 
-		if len(currentHeader.Digest.Types) == 0 {
-			return fmt.Errorf("current header missing digest")
+	currentHash := header.Hash()
+	slot, err := types.GetSlotFromHeader(header)
+	if err != nil {
+		return false, fmt.Errorf("failed to get slot for block %s: %w", currentHash, err)
```
qdm12, line 352: `currentHash` should be part of the error context set by `types.GetSlotFromHeader`, not the caller (to avoid repeating it):

```suggestion
		return false, fmt.Errorf("failed to get slot from header: %w", err)
```

lib/babe/verify.go
```diff
@@ -327,62 +327,58 @@ func (b *verifier) verifyAuthorshipRight(header *types.Header) error {
...
+	}
+
+	blocksInSlot, err := b.blockState.GetBlocksBySlot(slot)
+	if err != nil {
+		return false, fmt.Errorf("failed to get blocks produced in slot %d: %w", slot, err)
```
qdm12, line 357: `slot` should be part of the error context set by `GetBlocksBySlot`, not the caller (to avoid repeating it):

```suggestion
		return false, fmt.Errorf("failed to get blocks produced by slot: %w", err)
```

lib/babe/verify.go
```diff
@@ -327,62 +327,58 @@ func (b *verifier) verifyAuthorshipRight(header *types.Header) error {
...
+	}
+
+	for _, blockInSlot := range blocksInSlot {
```
qdm12, line 360: nit maybe rename:

- `blockInSlot` -> `blockHashInSlot`
- `blocksInSlot` -> `blockHashesInSlot`

lib/babe/verify_test.go
```diff
@@ -499,18 +500,15 @@ func Test_verifier_verifyAuthorshipRight(t *testing.T) {
 	assert.NoError(t, err)
 	testHeaderPrimary := types.NewEmptyHeader()
 	testHeaderPrimary.Digest = testDigestPrimary
-
```
qdm12, line 502: nit I'd keep those empty lines / these are unneeded deltas

lib/babe/verify_test.go
```diff
@@ -682,6 +677,21 @@ func Test_verifier_verifyAuthorshipRightEquivocatory(t *testing.T) {
 	output, proof, err := kp.VrfSign(makeTranscript(Randomness{}, uint64(1), 1))
 	assert.NoError(t, err)
 
+	output2, proof2, err := kp.VrfSign(makeTranscript(Randomness{}, uint64(1), 2))
+	assert.NoError(t, err)
+	secDigestExisting := types.BabePrimaryPreDigest{
```
qdm12, line 682: Rename to `secondDigestExisting`, `sec` can mean a lot of things / is confusing

lib/babe/verify_test.go
```diff
@@ -714,9 +725,9 @@ func Test_verifier_verifyAuthorshipRightEquivocatory(t *testing.T) {
 	hashEquivocatoryPrimary := encodeAndHashHeader(t, headerEquivocatoryPrimary)
 	signAndAddSeal(t, kp, headerEquivocatoryPrimary, hashEquivocatoryPrimary[:])
 
-	mockBlockStateEquiv1.EXPECT().GetAllBlocksAtDepth(headerEquivocatoryPrimary.ParentHash).Return(
-		[]common.Hash{hashEquivocatoryPrimary})
 	mockBlockStateEquiv1.EXPECT().GetHeader(hashEquivocatoryPrimary).Return(headerEquivocatoryPrimary, nil)
+	mockBlockStateEquiv1.EXPECT().GetBlocksBySlot(gomock.Any()).Return(
```
qdm12, line 729: Is it not possible to expect an actual slot integer instead of `gomock.Any()`?

lib/babe/verify_test.go
```diff
@@ -737,9 +748,9 @@ func Test_verifier_verifyAuthorshipRightEquivocatory(t *testing.T) {
 	signAndAddSeal(t, kp, headerEquivocatorySecondaryPlain, hashEquivocatorySecondaryPlain[:])
 	babeVerifier8 := newTestVerifier(kp, mockBlockStateEquiv2, scale.MaxUint128, true)
 
-	mockBlockStateEquiv2.EXPECT().GetAllBlocksAtDepth(headerEquivocatorySecondaryPlain.ParentHash).Return(
-		[]common.Hash{hashEquivocatorySecondaryPlain})
 	mockBlockStateEquiv2.EXPECT().GetHeader(hashEquivocatorySecondaryPlain).Return(headerEquivocatorySecondaryPlain, nil)
+	mockBlockStateEquiv2.EXPECT().GetBlocksBySlot(gomock.Any()).Return(
```
qdm12, line 752: Is it not possible to expect an actual slot integer instead of `gomock.Any()`?

lib/babe/verify_test.go
```diff
@@ -749,17 +760,16 @@ func Test_verifier_verifyAuthorshipRightEquivocatory(t *testing.T) {
...
 	headerEquivocatorySecondaryVRF := newTestHeader(t, *types.NewBABEPreRuntimeDigest(encVrfDigest))
-
 	hashEquivocatorySecondaryVRF := encodeAndHashHeader(t, headerEquivocatorySecondaryVRF)
 	signAndAddSeal(t, kp, headerEquivocatorySecondaryVRF, hashEquivocatorySecondaryVRF[:])
 	babeVerifierEquivocatorySecondaryVRF := newTestVerifier(kp, mockBlockStateEquiv3, scale.MaxUint128, true)
-	mockBlockStateEquiv3.EXPECT().GetAllBlocksAtDepth(headerEquivocatorySecondaryVRF.ParentHash).Return(
-		[]common.Hash{hashEquivocatorySecondaryVRF})
+
 	mockBlockStateEquiv3.EXPECT().GetHeader(hashEquivocatorySecondaryVRF).Return(headerEquivocatorySecondaryVRF, nil)
+	mockBlockStateEquiv3.EXPECT().GetBlocksBySlot(gomock.Any()).Return(
```
qdm12, line 771: Is it not possible to expect an actual slot integer instead of `gomock.Any()`?

lib/babe/verify_test.go
```diff
@@ -645,21 +640,21 @@ func Test_verifier_verifyAuthorshipRight(t *testing.T) {
...
+			expErr:   errors.New("failed to get header for block"),
 		},
 	}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
 			b := &tt.verifier
 			err := b.verifyAuthorshipRight(tt.header)
 			if tt.expErr != nil {
-				assert.EqualError(t, err, tt.expErr.Error())
+				assert.ErrorContains(t, err, tt.expErr.Error())
```
qdm12, line 657: Revert this and change the expected errors to `checking for justification: test mock error` and `getting justification: test mock error` for their respective test case.

lib/blocktree/blocktree.go
```diff
@@ -330,6 +330,19 @@ func (bt *BlockTree) GetAllBlocks() []Hash {
 	return bt.root.getAllDescendants(nil)
 }
 
+// GetAllDescendants returns all blocks that are descendants of the given block.
+func (bt *BlockTree) GetAllDescendants(a common.Hash) ([]Hash, error) {
+	bt.RLock()
+	defer bt.RUnlock()
+
+	an := bt.getNode(a)
```
qdm12, line 338: Please rename `a` -> `blockHash` and `an` to `node` (or something else, I'm not sure what `an` should be here.

lib/blocktree/blocktree.go
```diff
@@ -330,6 +330,19 @@ func (bt *BlockTree) GetAllBlocks() []Hash {
...
+	if an == nil {
+		return []common.Hash{}, ErrNodeNotFound
```
qdm12, line 340:
```suggestion
		return []common.Hash{}, fmt.Errorf("%w: for block hash %s", ErrNodeNotFound, a)
```

lib/blocktree/blocktree.go
```diff
@@ -330,6 +330,19 @@ func (bt *BlockTree) GetAllBlocks() []Hash {
 	return bt.root.getAllDescendants(nil)
 }
 
+// GetAllDescendants returns all blocks that are descendants of the given block.
+func (bt *BlockTree) GetAllDescendants(a common.Hash) ([]Hash, error) {
```
qdm12, line 334: Please add a unit test for this method

lib/babe/errors.go
```diff
@@ -72,6 +72,7 @@ var (
 	errNoBABEAuthorityKeyProvided = errors.New("cannot create BABE service as authority; no keypair provided")
 	errLastDigestItemNotSeal      = errors.New("last digest item is not seal")
 	errLaggingSlot                = errors.New("current slot is smaller than slot of best block")
+	errNoDigest                   = errors.New("no digest provided")
```
qdm12, line 75: nit It ultimately gets returned by `VerifyBlock` which is exported, so we *should* export this error

qdm12 (changes requested):

lib/babe/verify.go
qdm12, line 343: Can you add a unit test for this new method?

dot/state/block.go
```diff
@@ -250,6 +250,37 @@ func (bs *BlockState) GetHashByNumber(num uint) (common.Hash, error) {
 	return common.NewHash(bh), nil
 }
 
+// GetBlocksBySlot gets all blocks that were produced in given slot.
+func (bs *BlockState) GetBlocksBySlot(slotNum uint64) ([]common.Hash, error) {
```
qdm12, line 254: Please add a unit test for this new method

dot/state/block.go
```diff
@@ -250,6 +250,37 @@ func (bs *BlockState) GetHashByNumber(num uint) (common.Hash, error) {
...
+	highestFinalisedHash, err := bs.GetHighestFinalisedHash()
+	if err != nil {
+		return nil, fmt.Errorf("failed to get highest finalised hash: %w", err)
+	}
+
+	descendants, err := bs.bt.GetAllDescendants(highestFinalisedHash)
```
EclesioMeloJunior, line 260: Does this function work for finalized blocks? Because all finalized blocks are pruned from the memory once finalized and the `blocktree` only contains a reference for blocks in memory (not finalized yet)

dot/state/block.go
```diff
@@ -250,6 +250,37 @@ func (bs *BlockState) GetHashByNumber(num uint) (common.Hash, error) {
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to get descendants of %s: %w", highestFinalisedHash, err)
+	}
+
+	blocksWithGivenSlot := []common.Hash{}
```
EclesioMeloJunior, line 265:
```suggestion
	blocksWithGivenSlot := make([]common.Hash, 0, len(descendants))
```

dot/types/babe.go
```diff
@@ -108,6 +109,10 @@ type ConfigData struct {
 
 // GetSlotFromHeader returns the BABE slot from the given header
 func GetSlotFromHeader(header *Header) (uint64, error) {
+	if header.Number == 0 {
+		return 0, ErrGenesisHeader
```
EclesioMeloJunior, line 113: 👍 nice!

lib/babe/verify.go
```diff
@@ -327,62 +327,58 @@ func (b *verifier) verifyAuthorshipRight(header *types.Header) error {
...
 
-		if len(currentHeader.Digest.Types) == 0 {
-			return fmt.Errorf("current header missing digest")
+	currentHash := header.Hash()
+	slot, err := types.GetSlotFromHeader(header)
+	if err != nil {
+		return false, fmt.Errorf("failed to get slot for block %s: %w", currentHash, err)
+	}
+
+	blocksInSlot, err := b.blockState.GetBlocksBySlot(slot)
```
EclesioMeloJunior, line 355: I have a question about this method: Since `GetBlocksBySlot` retrieves finalized blocks (given that `GetBlocksBySlot` get all descendants of `GetHighestFinalisedHash`) produced on slot `X` why do we need to verify the block equivocation?

I mean, if `GetBlocksBySlot` already returns finalized blocks on slot X any other block being created on the same epoch should be invalid, right?

I believe you should change the method `GetBlocksBySlot` to rely on `BestBlock` (best block not finalized yet) to get blocks produced on a given slot and then verify equivocation

EclesioMeloJunior (changes requested):

dot/state/block.go
qdm12: I don't know if this is that necessary, if we only set one element in the slice, iterating from 100s of elements, it might not be worth setting a 100-sized capacity in memory. Unless most of the time most elements are appended in the slice then it's worth it. Not a big deal performance either I'd guess.

lib/babe/verify.go
kishansagathiya: descendants of `GetHighestFinalisedHash` blocks will not be finalised.

kishansagathiya: lets' say `GetHighestFinalisedHash` gives you `x`. If `x` has a descendant block `d` that is finalized, that would make `d` the highest finalised block.

dot/state/block.go
kishansagathiya: all descendants will be not finalised

lib/babe/errors.go
kishansagathiya: > so we should export this error

When it is needed outside this package, not now

lib/babe/verify.go
kishansagathiya: `Test_verifier_verifyAuthorshipRightEquivocatory` and `TestVerifyAuthorshipRight_Equivocation`

dot/state/block.go
```diff
@@ -250,6 +250,37 @@ func (bs *BlockState) GetHashByNumber(num uint) (common.Hash, error) {
...
+
+	for _, desc := range descendants {
+		descSlot, err := bs.GetSlotForBlock(desc)
+		if errors.Is(err, types.ErrGenesisHeader) {
+			continue
+		}
+		if err != nil {
+			return blocksWithGivenSlot, fmt.Errorf("could not get slot for block %s: %w", desc, err)
```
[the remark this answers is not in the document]
kishansagathiya: > to avoid repeating it

desc gets printed only once in this case. I would rather add that context here once than adding it in all errors in `GetSlotForBlock`

lib/babe/verify.go
kishansagathiya: > to avoid repeating it

desc gets printed only once in this case. I would rather add that context here once than adding it in all errors in `verifyBlockEquivocation`

dot/types/babe.go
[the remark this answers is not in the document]
kishansagathiya: I don't really care much for this, tbh. But if you want to start wrapping errors, instead of doing this here, we can just create our errors with `fmt.Errorf` instead of `errors.New`. 

@timwu20 what do you think about this change?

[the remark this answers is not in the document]
EclesioMeloJunior: > But if you want to start wrapping errors, instead of doing this here, we can just create our errors with fmt.Errorf instead of errors.New

Only changing `errors.New` for `fmt.Errorf` we don't wrap errors, to wrap an error we should explicity use `fmt.Errorf("%w", err)`. 

The suggestion is to enforce the usage of `errors.Is` and changing`errors.New` for `fmt.Errorf` we keep the problem (allowing error assertion with `==` instead of `errors.Is`)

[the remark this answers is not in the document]
timwu20: What are we discussing here? Whether or not we define errors via `errors.New` or `fmt.Errorf`?

lib/babe/state.go
```diff
@@ -28,6 +28,7 @@ type BlockState interface {
 	GetHeader(common.Hash) (*types.Header, error)
 	GetBlockByNumber(blockNumber uint) (*types.Block, error)
 	GetBlockByHash(common.Hash) (*types.Block, error)
+	GetBlocksBySlot(uint64) ([]common.Hash, error)
```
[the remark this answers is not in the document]
timwu20: maybe rename to `GetBlockHashesBySlot`?

dot/types/babe.go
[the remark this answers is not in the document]
kishansagathiya: @EclesioMeloJunior my bad, why do you want to force `errors.Is`?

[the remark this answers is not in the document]
EclesioMeloJunior: @kishansagathiya I am not advocating for enforcing `errors.Is`, I commented that using `fmt.Errorf` to create errors instead of `errors.New` (as you seggested in https://github.com/ChainSafe/gossamer/pull/2897) is not the same as wrapping an error and is really odd to me.

And regards to the @qdm12 suggestion, I tested in the [go.dev](https://go.dev/play/p/oXGHNROAgm6) if using `fmt.Errorf` to create errors enforces the usage of `errors.Is` over the `==` and it does not enforce.

EclesioMeloJunior (approved):

timwu20 (commented):
bumping my previous feedback

dot/types/babe.go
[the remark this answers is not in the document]
timwu20: pretty sure `errors.Is` works in this case on the calling side if testing for `ErrGenesisHeader`.  I'm not sure why we need to wrap this error when returning as based on the suggestion.

lib/babe/verify.go
qdm12: Problem is we should standardize this; you could have a local function within `verifyBlockEquivocation`:

```go
func verifyBlockEquivocation(...) {
   wrapError := func(err error) error {
     return fmt.Errorf("for block header %s: %w", header, err)
  }
  // ...
  if err != nil {
    return false, wrapError(err)
  }

}
```

lib/babe/verify_test.go
kishansagathiya: There are hashes included in the some error context, for example 
```
could not verify block equivocation for header 0x2eb61ed30b3663f18c53f29d0d6ad3dd7f40aafdbc73b04727e5e3e9891d9a8c: failed to get authority index for block 0x0100000000000000000000000000000000000000000000000000000000000000: for block hash 0xdcdd89927d8a348e00257e1ecc8617f45edb5118efff3ea2f9961b2ad9b7690a: no digest provided
```

Can't find a test case where `checking for justification: test mock error` and `getting justification: test mock error` are expected test cases

kishansagathiya: addressed

lib/babe/verify_test.go
kishansagathiya: reverted

dot/types/babe.go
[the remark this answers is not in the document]
kishansagathiya: @qdm12 does not look like I need to do anything here. Please un-resolve if you think otherwise.

lib/babe/verify.go
kishansagathiya: I prefer to explicitly declare variables rather than declaring with named returns.

dot/state/block.go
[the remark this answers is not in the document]
kishansagathiya: I prefer to explicitly declare variables instead of declaring them using named returns.

lib/babe/verify.go
kishansagathiya: https://github.com/ChainSafe/gossamer/pull/2709#discussion_r1007990573

lib/babe/verify.go
qdm12: No since it opens the door to adding multiple times the same context at different levels of the call stack.
Adding the context once at the bottom layer actually handling the argument is the way to avoid this and not have to think about it.

dot/state/block_test.go
```diff
@@ -195,6 +195,62 @@ func TestGetSlotForBlock(t *testing.T) {
 	require.Equal(t, expectedSlot, res)
 }
 
+func TestGetBlockHashesBySlot(t *testing.T) {
```
qdm12, line 198: nit run it in parallel

```suggestion
func TestGetBlockHashesBySlot(t *testing.T) {
	t.Parallel()

```

dot/state/block_test.go
```diff
@@ -195,6 +195,62 @@ func TestGetSlotForBlock(t *testing.T) {
...
+		Body: types.Body{},
+	}
+	err = bs.AddBlock(block2)
+	require.NoError(t, err)
+
+	blocks, err := bs.GetBlockHashesBySlot(slot)
+	require.NoError(t, err)
+	require.Equal(t, len(blocks), 2)
+	require.Contains(t, blocks, block.Header.Hash())
+	require.Contains(t, blocks, block2.Header.Hash())
```
qdm12, line 251: simplifying with

```suggestion
	expectedBlockHashes := []common.Hash{block.Header.Hash(), block2.Header.Hash()}
	require.ElementsMatch(t, blocks, expectedBlockHashes)
```

also `require.Equal` has the expected as first arg, actual as 2nd arg, otherwise it's hard-to-debug test messages when it fails

lib/babe/errors.go
qdm12: I get your point, although then it defeats the purpose of having a sentinel error, we might as well just have a string with `fmt.Errorf` without a sentinel error. I guess then it's up for debate which is better, but having an unexported sentinel error is just wasteful in this case.

lib/babe/state.go
```diff
@@ -28,6 +28,7 @@ type BlockState interface {
 	GetHeader(common.Hash) (*types.Header, error)
 	GetBlockByNumber(blockNumber uint) (*types.Block, error)
 	GetBlockByHash(common.Hash) (*types.Block, error)
+	GetBlockHashesBySlot(uint64) ([]common.Hash, error)
```
qdm12, line 31: nit adding named arguments and return values is nice for documentation

```suggestion
	GetBlockHashesBySlot(slot uint64) (blockHashes []common.Hash, err error)
```

lib/babe/verify.go
qdm12: I'd still argue that documentation with code is better than documentation with comments

> Good code is its own best documentation. As you’re about to add a comment, ask yourself, 'How can I improve the code so that this comment isn’t needed?' Improve the code and then document it to make it even clearer

qdm12: - `Test_verifier_verifyAuthorshipRightEquivocatory` tests its caller `verifyAuthorshipRightEquivocatory` not `verifyBlockEquivocation`. It does not cover all code paths in that function either.
- `TestVerifyAuthorshipRight_Equivocation` takes 4 seconds to execute and is an integration test, not a unit test. And also does not cover all code paths in `verifyBlockEquivocation`

lib/babe/verify_test.go
```diff
@@ -525,11 +527,11 @@ func Test_verifier_verifyAuthorshipRight(t *testing.T) {
 	h := common.MustHexToHash("0x01")
 	h1 := []common.Hash{h}
 
-	mockBlockState.EXPECT().GetAllBlocksAtDepth(gomock.Any()).Return(h1)
 	mockBlockState.EXPECT().GetHeader(h).Return(types.NewEmptyHeader(), nil)
+	mockBlockState.EXPECT().GetBlockHashesBySlot(gomock.AssignableToTypeOf(uint64(0))).Return(h1, nil)
```
qdm12, line 531: please don't use `gomock.AssignableToTypeOf` or `gomock.Any` when it's very easy to expect an actual value.

```suggestion
	mockBlockState.EXPECT().GetBlockHashesBySlot(uint64(1)).Return(h1, nil)
```

lib/babe/verify_test.go
```diff
@@ -525,11 +527,11 @@ func Test_verifier_verifyAuthorshipRight(t *testing.T) {
...
 
-	mockBlockStateErr.EXPECT().GetAllBlocksAtDepth(gomock.Any()).Return(h1)
 	mockBlockStateErr.EXPECT().GetHeader(h).Return(nil, errors.New("get header error"))
+	mockBlockStateErr.EXPECT().GetBlockHashesBySlot(gomock.AssignableToTypeOf(uint64(0))).Return(h1, nil)
```
qdm12, line 534: please don't use `gomock.AssignableToTypeOf` or `gomock.Any` when it's very easy to expect an actual value.

```suggestion
	mockBlockStateErr.EXPECT().GetBlockHashesBySlot(uint64(1)).Return(h1, nil)
```

lib/babe/verify_test.go
```diff
@@ -550,23 +552,20 @@ func Test_verifier_verifyAuthorshipRight(t *testing.T) {
...
 	header5 := newTestHeader(t, *babeSecPlainPrd)
-
 	signAndAddSeal(t, kp, header5, []byte{1})
 	babeVerifier3 := newTestVerifier(kp, mockBlockState, scale.MaxUint128, true)
 
 	// Case 6: Invalid signature - BabeSecondaryVrfPreDigest
 	encSecVrfDigest := newEncodedBabeDigest(t, testBabeSecondaryVRFPreDigest)
 	assert.NoError(t, err)
 	header6 := newTestHeader(t, *types.NewBABEPreRuntimeDigest(encSecVrfDigest))
-
```
qdm12, line 569: nit a lot of empty lines removed, not sure why. Try being careful next time to reduce deltas :wink:

lib/babe/verify_test.go
```diff
@@ -648,13 +647,14 @@ func Test_verifier_verifyAuthorshipRight(t *testing.T) {
 			name:     "valid digest items, getAuthorityIndex error",
 			verifier: *babeVerifier5,
 			header:   header7,
-			expErr:   errors.New("failed to get authority index no digest provided"),
+			expErr: fmt.Errorf("could not verify block equivocation: failed to get authority index for block %s: for block hash %s: %w", //nolint:lll
+				h, types.NewEmptyHeader().Hash(), errNoDigest),
```
qdm12, line 651: removing the `//nolint` + expecting constant string since there is no need to have a dynamic expectation (remove complexity/debugging hassle)

```suggestion
			expErr: fmt.Errorf("could not verify block equivocation: " +
				"failed to get authority index for block " +
				"0x0100000000000000000000000000000000000000000000000000000000000000: " +
				"for block hash 0xdcdd89927d8a348e00257e1ecc8617f45edb5118efff3ea2f9961b2ad9b7690a: " +
				"no digest provided"),
```

lib/babe/verify_test.go
```diff
@@ -648,13 +647,14 @@ func Test_verifier_verifyAuthorshipRight(t *testing.T) {
...
 		},
 		{
 			name:     "get header err",
 			verifier: *babeVerifier6,
 			header:   header7,
-			expErr:   errors.New("failed get header get header error"),
+			expErr:   fmt.Errorf("could not verify block equivocation: failed to get header for block %s: get header error", h),
```
qdm12, line 657: expecting constant string since there is no need to have a dynamic expectation.

```suggestion
			expErr: fmt.Errorf("could not verify block equivocation: " +
				"failed to get header for block 0x0100000000000000000000000000000000000000000000000000000000000000: " +
				"get header error"),
```

lib/babe/verify_test.go
```diff
@@ -717,9 +733,9 @@ func Test_verifier_verifyAuthorshipRightEquivocatory(t *testing.T) {
 	hashEquivocatoryPrimary := encodeAndHashHeader(t, headerEquivocatoryPrimary)
 	signAndAddSeal(t, kp, headerEquivocatoryPrimary, hashEquivocatoryPrimary[:])
 
-	mockBlockStateEquiv1.EXPECT().GetAllBlocksAtDepth(headerEquivocatoryPrimary.ParentHash).Return(
-		[]common.Hash{hashEquivocatoryPrimary})
 	mockBlockStateEquiv1.EXPECT().GetHeader(hashEquivocatoryPrimary).Return(headerEquivocatoryPrimary, nil)
+	mockBlockStateEquiv1.EXPECT().GetBlockHashesBySlot(gomock.AssignableToTypeOf(uint64(0))).Return(
```
qdm12, line 737:
```suggestion
	mockBlockStateEquiv1.EXPECT().GetBlockHashesBySlot(uint64(1)).Return(
```

lib/babe/verify_test.go
```diff
@@ -740,9 +756,9 @@ func Test_verifier_verifyAuthorshipRightEquivocatory(t *testing.T) {
 	signAndAddSeal(t, kp, headerEquivocatorySecondaryPlain, hashEquivocatorySecondaryPlain[:])
 	babeVerifier8 := newTestVerifier(kp, mockBlockStateEquiv2, scale.MaxUint128, true)
 
-	mockBlockStateEquiv2.EXPECT().GetAllBlocksAtDepth(headerEquivocatorySecondaryPlain.ParentHash).Return(
-		[]common.Hash{hashEquivocatorySecondaryPlain})
 	mockBlockStateEquiv2.EXPECT().GetHeader(hashEquivocatorySecondaryPlain).Return(headerEquivocatorySecondaryPlain, nil)
+	mockBlockStateEquiv2.EXPECT().GetBlockHashesBySlot(gomock.AssignableToTypeOf(uint64(0))).Return(
```
qdm12, line 760:
```suggestion
	mockBlockStateEquiv2.EXPECT().GetBlockHashesBySlot(uint64(1)).Return(
```

lib/babe/verify_test.go
```diff
@@ -752,17 +768,16 @@ func Test_verifier_verifyAuthorshipRightEquivocatory(t *testing.T) {
...
 	headerEquivocatorySecondaryVRF := newTestHeader(t, *types.NewBABEPreRuntimeDigest(encVrfDigest))
-
 	hashEquivocatorySecondaryVRF := encodeAndHashHeader(t, headerEquivocatorySecondaryVRF)
 	signAndAddSeal(t, kp, headerEquivocatorySecondaryVRF, hashEquivocatorySecondaryVRF[:])
 	babeVerifierEquivocatorySecondaryVRF := newTestVerifier(kp, mockBlockStateEquiv3, scale.MaxUint128, true)
-	mockBlockStateEquiv3.EXPECT().GetAllBlocksAtDepth(headerEquivocatorySecondaryVRF.ParentHash).Return(
-		[]common.Hash{hashEquivocatorySecondaryVRF})
+
 	mockBlockStateEquiv3.EXPECT().GetHeader(hashEquivocatorySecondaryVRF).Return(headerEquivocatorySecondaryVRF, nil)
+	mockBlockStateEquiv3.EXPECT().GetBlockHashesBySlot(gomock.AssignableToTypeOf(uint64(0))).Return(
```
qdm12, line 779:
```suggestion
	mockBlockStateEquiv3.EXPECT().GetBlockHashesBySlot(uint64(1)).Return(
```

lib/blocktree/blocktree.go
```diff
@@ -330,6 +330,19 @@ func (bt *BlockTree) GetAllBlocks() []Hash {
 	return bt.root.getAllDescendants(nil)
 }
 
+// GetAllDescendants returns all blocks that are descendants of the given block.
```
qdm12, line 333:
```suggestion
// GetAllDescendants returns all block hashes that are descendants of the given block hash.
```

lib/blocktree/blocktree.go
```diff
@@ -330,6 +330,19 @@ func (bt *BlockTree) GetAllBlocks() []Hash {
...
+func (bt *BlockTree) GetAllDescendants(hash common.Hash) ([]Hash, error) {
+	bt.RLock()
+	defer bt.RUnlock()
+
+	node := bt.getNode(hash)
+	if node == nil {
+		return []common.Hash{}, fmt.Errorf("%w: for block hash %s", ErrNodeNotFound, hash)
```
qdm12, line 340: nit `nil` seems more appropriate/saves some allocation

```suggestion
		return nil, fmt.Errorf("%w: for block hash %s", ErrNodeNotFound, hash)
```

lib/blocktree/blocktree_test.go
```diff
@@ -347,6 +347,15 @@ func TestBlockTree_GetAllBlocksAtNumber(t *testing.T) {
 	require.Equal(t, expected, hashes)
 }
 
+func TestBlockTreeGetAllDescendants(t *testing.T) {
```
qdm12, line 350: nit run in parallel

```suggestion
func TestBlockTreeGetAllDescendants(t *testing.T) {
	t.Parallel()

```

dot/types/babe.go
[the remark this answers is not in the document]
qdm12: Oh well this really was a simple suggestion and the conversation really went overboard :smile: 

The reason for `errors.Is` and 'disallowing' `==` is to allow later to add context without breaking compatibility with callers.

From https://go.dev/blog/go1.13-errors

> If a function is defined as returning an error wrapping some sentinel or type, do not return the underlying error directly.

```go
...
return fmt.Errorf("%w", ErrPermission)
...
```

lib/babe/verify_test.go
kishansagathiya: @timwu20 

> when it's very easy to expect an actual value.

@qdm12 Just write your suggestion. There is no need for this condescending behaviour.

lib/babe/verify_test.go
kishansagathiya: @timwu20 I modify the code based on what seems appropriate. Is that what I am supposed to do or write code based on how I create the least amount of deltas?

lib/babe/errors.go
kishansagathiya: `errNoDigest` gets used at multiple places, that's purpose here.

lib/babe/verify.go
kishansagathiya: Naming return variables in the signature leads to two things
- you can get away with not explicitly declaring variables
- you can get away with having naked returns instead of explicit returns.

I can't really keep on arguing about every small small things. Even if it is a nit comment, I have to resolve it. I am very humbly requesting you to stop giving out reviews based on preferences and opinions.

kishansagathiya: added the test

dot/types/babe.go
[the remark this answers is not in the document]
timwu20: I don't really agree with the go blog in this case.  Like I see how this would enforce callers of this function to use `errors.Is`, but we are copying an error and not adding context and passing that around.  We can just enforce the use of `errors.Is` through PRs and mutual understanding.

timwu20 (approved):

dot/state/block.go
```diff
@@ -250,6 +250,37 @@ func (bs *BlockState) GetHashByNumber(num uint) (common.Hash, error) {
 	return common.NewHash(bh), nil
 }
 
+// GetBlockHashesBySlot gets all blocks that were produced in given slot.
```
qdm12, line 253: nit

```suggestion
// GetBlockHashesBySlot gets all block hashes that were produced in the given slot.
```

dot/state/block.go
```diff
@@ -250,6 +250,37 @@ func (bs *BlockState) GetHashByNumber(num uint) (common.Hash, error) {
...
+
+	blocksWithGivenSlot := []common.Hash{}
+
+	for _, desc := range descendants {
+		descSlot, err := bs.GetSlotForBlock(desc)
+		if errors.Is(err, types.ErrGenesisHeader) {
+			continue
+		}
+		if err != nil {
+			return nil, fmt.Errorf("could not get slot for block %s: %w", desc, err)
```
qdm12, line 273: `desc` is to be wrapped by `GetSlotForBlock`, not here

```suggestion
			return nil, fmt.Errorf("could not get slot for block: %w", err)
```

dot/state/block_test.go
```diff
@@ -195,6 +195,62 @@ func TestGetSlotForBlock(t *testing.T) {
...
+			Digest:     digest2,
+		},
+		Body: types.Body{},
+	}
+	err = bs.AddBlock(block2)
+	require.NoError(t, err)
+
+	blocks, err := bs.GetBlockHashesBySlot(slot)
+	require.NoError(t, err)
+	require.ElementsMatch(t, blocks, []common.Hash{block.Header.Hash(), block2.Header.Hash()})
```
qdm12, line 251: nit `blocks` -> `blockHashes`

```suggestion
	blockHashes, err := bs.GetBlockHashesBySlot(slot)
	require.NoError(t, err)
	require.ElementsMatch(t, blockHashes, []common.Hash{block.Header.Hash(), block2.Header.Hash()})
```

lib/babe/verify.go
```diff
@@ -327,62 +327,58 @@ func (b *verifier) verifyAuthorshipRight(header *types.Header) error {
...
-		}
+	return nil
+}
 
-		currentBlockProducerIndex, err := getAuthorityIndex(currentHeader)
-		if err != nil {
-			return fmt.Errorf("failed to get authority index %s", err)
-		}
+// verifyBlockEquivocation checks if the given block's author has occupied the corresponding slot more than once.
+// It returns true if block was equivocated.
```
qdm12, line 342: nit

```suggestion
// It returns true if the block was equivocated.
```

lib/babe/verify.go
```diff
@@ -327,62 +327,58 @@ func (b *verifier) verifyAuthorshipRight(header *types.Header) error {
...
 
-		currentPreDigestItemValue, err := currentHeader.Digest.Types[0].Value()
+		existingHeader, err := b.blockState.GetHeader(blockHashInSlot)
 		if err != nil {
-			return fmt.Errorf("getting current header first digest type value: %w", err)
-		}
-		currentPreDigest, ok := currentPreDigestItemValue.(types.PreRuntimeDigest)
-		if !ok {
-			return fmt.Errorf("%w: got %T", types.ErrNoFirstPreDigest, currentPreDigestItemValue)
+			return false, fmt.Errorf("failed to get header for block %s: %w", blockHashInSlot, err)
```
qdm12, line 367: See https://chainsafeio.slack.com/archives/CEATMPULS/p1666965925368239

```suggestion
			return false, fmt.Errorf("failed to get header: %w", err)
```

lib/babe/verify_integration_test.go
```diff
@@ -466,7 +466,7 @@ func TestVerifyAuthorshipRight_Equivocation(t *testing.T) {
 	require.NoError(t, err)
 
 	err = verifier.verifyAuthorshipRight(&block2.Header)
-	require.Equal(t, ErrProducerEquivocated, err)
+	require.Equal(t, fmt.Errorf("%w for block header %s", ErrProducerEquivocated, block2.Header.Hash()), err)
```
qdm12, line 469: Test the `errors.Is` behavior + string separately if you want

```suggestion
	require.ErrorIs(t, ErrProducerEquivocated, err)
	require.EqualError(t, err, fmt.Sprintf("%s for block header %s", ErrProducerEquivocated, block2.Header.Hash())
```

lib/babe/verify_test.go
```diff
@@ -671,6 +673,144 @@ func Test_verifier_verifyAuthorshipRight(t *testing.T) {
...
+			verifier:    *verifier5,
+			header:      testHeader5,
+			equivocated: true,
+			expErr:      nil,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			tt := tt
```
qdm12, line 802: unneeded unless `t.Parallel()` is called after it to pin the `tt` variable in the loop.

```suggestion
```

lib/babe/verify_test.go
```diff
@@ -671,6 +673,144 @@ func Test_verifier_verifyAuthorshipRight(t *testing.T) {
 	}
 }
 
+func Test_verifier_verifyBlockEquivocation(t *testing.T) {
+	ctrl := gomock.NewController(t)
```
qdm12, line 677: It would be nice to have a controller + mocks configured per subtest as described in https://engineering-handbook.onrender.com/development/tech-stack/go#mocks-with-subtests instead of in the parent test for the subtests

qdm12 (approved):

dot/types/babe.go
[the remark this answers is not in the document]
kishansagathiya: think, it is safe to resolve this.

PR #2717, opened 2022-08-02
chore(deps): remove dependency on centrifuge v3

qdm12:
## Changes

Use centrifuge v4 calls only.

I thought I would do it when resolving #2646 

## Tests

```sh
```


jimjbrettj (approved):
I had actually done this in my PR too lol, but looks good!

EclesioMeloJunior (approved):

PR #2722, opened 2022-08-03
feat: include nested varying data type on neighbor messages

### LINKED ISSUE #2716: Update Neighbor Message
Currently, substrate only share vote messages with peers who already sent `Neighbor Messages` and gossamer implements the message in the wrong way. In substrate the `GossipMessage` enum contains the `Neighbor` which is another enum called `VersionedNeighborPacket`:

```rs
#[derive(Debug, Encode, Decode)]
pub(super) enum GossipMessage<Block: BlockT> {
        ...
	Neighbor(VersionedNeighborPacket<NumberFor<Block>>),
}
```

And the `VersionedNeighborPacket` is another enum that contains the `NeighborPacket`:

```rs
#[derive(Debug, Encode, Decode)]
pub(super) enum VersionedNeighborPacket<N> {
	#[codec(index = 1)]
	V1(NeighborPacket<N>),
}

#[derive(Debug, Encode, Decode, Clone)]
pub(super) struct NeighborPacket<N> {
	pub(super) round: Round,
	pub(super) set_id: SetId,
	pub(super) commit_finalized_height: N,
}
```

We should use the gossamer scale varying nested data type to support this type

EclesioMeloJunior:
## Changes

- Update gossamer `NeighborMessages` to match the definition:
```rs
#[derive(Debug, Encode, Decode)]
pub(super) enum GossipMessage<Block: BlockT> {
        ...
	Neighbor(VersionedNeighborPacket<NumberFor<Block>>),
}

#[derive(Debug, Encode, Decode)]
pub(super) enum VersionedNeighborPacket<N> {
	#[codec(index = 1)]
	V1(NeighborPacket<N>),
}

#[derive(Debug, Encode, Decode, Clone)]
pub(super) struct NeighborPacket<N> {
	pub(super) round: Round,
	pub(super) set_id: SetId,
	pub(super) commit_finalized_height: N,
}
```

## Tests

```sh
go test github.com/ChainSafe/gossamer/lib/grandpa
```

## Issues

- Closes #2716

## Primary Reviewer

@timwu20


lib/grandpa/message.go
```diff
@@ -77,21 +77,30 @@ func (v *VoteMessage) ToConsensusMessage() (*ConsensusMessage, error) {
 	}, nil
 }
 
-// NeighbourMessage represents a network-level neighbour message
-type NeighbourMessage struct {
-	Version byte
-	Round   uint64
-	SetID   uint64
-	Number  uint32
+type VersionedNeighborMessage scale.VaryingDataType
```
qdm12, line 80: Should we just keep it as `NeighbourMessage` though? I don't think there is a need to rename it to `VersionedNeighborMessage` right?

lib/grandpa/message_handler.go
```diff
@@ -21,6 +21,10 @@ import (
 	"github.com/libp2p/go-libp2p-core/peer"
 )
 
+var (
+	ErrNeighborVersionNotSupported = errors.New("neighbor version not supported")
```
qdm12, line 25: nit Rename to `ErrNeighbourVersionNotSupported`

lib/grandpa/network.go
```diff
@@ -177,29 +175,35 @@ func (s *Service) sendMessage(msg GrandpaMessage) error {
 func (s *Service) sendNeighbourMessage(interval time.Duration) {
 	t := time.NewTicker(interval)
 	defer t.Stop()
+
+	var neighborMessage *V1NeighbourMessage
```
qdm12, line 179: nit rename to `neighbourMessage`

lib/grandpa/network.go
```diff
@@ -222,8 +226,15 @@ func decodeMessage(cm *network.ConsensusMessage) (m GrandpaMessage, err error) {
 		m = &val
 	case CommitMessage:
 		m = &val
-	case NeighbourMessage:
-		m = &val
+
+	case VersionedNeighborMessage:
+		switch neighborMessage := msg.Value().(type) {
```
qdm12, line 231: nit rename to `neighbourMessage`

lib/grandpa/message.go
EclesioMeloJunior: from what I understand we could support more than one neighbour messages but for now, only the version 1 is in usage. I believe we can keep it as it is

lib/grandpa/message.go
```diff
@@ -77,21 +77,55 @@ func (v *VoteMessage) ToConsensusMessage() (*ConsensusMessage, error) {
...
 }
 
-// NeighbourMessage represents a network-level neighbour message
-type NeighbourMessage struct {
-	Version byte
-	Round   uint64
-	SetID   uint64
-	Number  uint32
+// VersionedNeighbourMessage represents the enum of nighbor messages
+type VersionedNeighbourMessage scale.VaryingDataType
```
timwu20, line 81:
```suggestion
type VersionedNeighbourPacket scale.VaryingDataType
```
Can we start just matching the substrate names.  It'll make it much easier to read both codebases.

lib/grandpa/message.go
```diff
@@ -77,21 +77,55 @@ func (v *VoteMessage) ToConsensusMessage() (*ConsensusMessage, error) {
...
+}
+
+// Value returns the current VDT value
+func (vnm *VersionedNeighbourMessage) Value() (val scale.VaryingDataTypeValue) {
+	vdt := scale.VaryingDataType(*vnm)
+	return vdt.Value()
+}
+
+// V1NeighbourMessage represents a network-level neighbour message
+type V1NeighbourMessage struct {
```
timwu20, line 110:
```suggestion
type V1NeighbourPacket struct {
```
Or even just `NeighbourPacket`

timwu20 (changes requested):

timwu20:
Nice change @EclesioMeloJunior. I can see how the previous encoding used to work, but I feel like the types you've introduced better reflect the structure of the `NeighbourPacket`.

lib/grandpa/message.go
```diff
@@ -77,21 +77,55 @@ func (v *VoteMessage) ToConsensusMessage() (*ConsensusMessage, error) {
...
+	Round  uint64
+	SetID  uint64
```
timwu20, line 112: this is actually a struct containing a u64 in the substrate codebase [(line)](https://github.com/paritytech/substrate/blob/86a0fc25e3aaeee273b6b3706f2aef99df72a21e/core/finality-grandpa/src/communication/mod.rs#L633).  Maybe provide a comment here referencing this.  If they add an attribute to this struct it will break encoding/decoding.

lib/grandpa/message.go
qdm12: supernit maybe `NeighbourPacketV1` would be nicer than `V1NeighbourPacket` for autocompletion purposes

lib/grandpa/message.go
qdm12: Any idea why they are using a struct? 🤔

EclesioMeloJunior: I would say that is a common practice in rust wrapper primitive types within another meaningful type, I believe its called [zero-cost abstraction](https://blog.polybdenum.com/2021/08/09/when-zero-cost-abstractions-aren-t-zero-cost.html)

lib/grandpa/message.go
```diff
@@ -21,7 +21,7 @@ type GrandpaMessage interface { //nolint:revive
 // NewGrandpaMessage returns a new VaryingDataType to represent a GrandpaMessage
 func newGrandpaMessage() scale.VaryingDataType {
 	return scale.MustNewVaryingDataType(
-		VoteMessage{}, CommitMessage{}, NeighbourMessage{},
+		VoteMessage{}, CommitMessage{}, newVersionedNighborMessage(),
```
qdm12, line 24: nit rename `newVersionedNighborMessage` to `newVersionedNighbourMessage`

lib/grandpa/message.go
```diff
@@ -77,21 +77,57 @@ func (v *VoteMessage) ToConsensusMessage() (*ConsensusMessage, error) {
 	}, nil
 }
 
-// NeighbourMessage represents a network-level neighbour message
-type NeighbourMessage struct {
-	Version byte
-	Round   uint64
-	SetID   uint64
-	Number  uint32
+// VersionedNeighbourPacket represents the enum of nighbor messages
```
qdm12, line 80:
```suggestion
// VersionedNeighbourPacket represents the enum of neighbour messages
```

also is this an enum?

lib/grandpa/message.go
```diff
@@ -77,21 +77,57 @@ func (v *VoteMessage) ToConsensusMessage() (*ConsensusMessage, error) {
...
+func (VersionedNeighbourPacket) Index() uint { return 2 }
+
+func newVersionedNighborMessage() VersionedNeighbourPacket {
+	vdt := scale.MustNewVaryingDataType(NeighbourPacketV1{})
+
+	return VersionedNeighbourPacket(vdt)
+}
+
+// Set updates the current VDT value to be `val`
+func (vnm *VersionedNeighbourPacket) Set(val scale.VaryingDataTypeValue) (err error) {
```
qdm12, line 93: Rename `vnm` to `vnp`, same for other methods.

lib/grandpa/message.go
```diff
@@ -77,21 +77,57 @@ func (v *VoteMessage) ToConsensusMessage() (*ConsensusMessage, error) {
...
+	vdt := scale.VaryingDataType(*vnm)
+	err = vdt.Set(val)
+	if err != nil {
+		return
```
qdm12, line 97:
```suggestion
		return Errorf("setting varying data type value: %w", err)
```

lib/grandpa/message.go
```diff
@@ -77,21 +77,57 @@ func (v *VoteMessage) ToConsensusMessage() (*ConsensusMessage, error) {
...
+	}
+	*vnm = VersionedNeighbourPacket(vdt)
+	return
```
qdm12, line 100: nit naked returns just makes things confusing imo

```suggestion
	return nil
```

lib/grandpa/message.go
```diff
@@ -77,21 +77,57 @@ func (v *VoteMessage) ToConsensusMessage() (*ConsensusMessage, error) {
...
-func (NeighbourMessage) Index() uint { return 2 }
+func (NeighbourPacketV1) Index() uint { return 1 }
 
 // ToConsensusMessage converts the NeighbourMessage into a network-level consensus message
-func (m *NeighbourMessage) ToConsensusMessage() (*network.ConsensusMessage, error) {
+func (m *NeighbourPacketV1) ToConsensusMessage() (*network.ConsensusMessage, error) {
+	versionedNeighbourPacket := newVersionedNighborMessage()
+	err := versionedNeighbourPacket.Set(*m)
+	if err != nil {
+		return nil, err
```
qdm12, line 126:
```suggestion
		return nil, fmt.Errorf("setting neighbour packet v1: %w", err)
		
```

lib/grandpa/network.go
```diff
@@ -222,8 +226,14 @@ func decodeMessage(cm *network.ConsensusMessage) (m GrandpaMessage, err error) {
...
 		m = &val
-	case NeighbourMessage:
-		m = &val
+
+	case VersionedNeighbourPacket:
+		switch NeighbourMessage := val.Value().(type) {
+		case NeighbourPacketV1:
+			m = &NeighbourMessage
+		default:
+			return nil, ErrInvalidMessageType
```
qdm12, line 235: A bit of a new thing, we should wrap sentinel errors even without context, so it forces callers to use `errors.Is` instead of `==` which would fail later if some context is added.

```suggestion
			return nil, fmt.Errorf("%w", ErrInvalidMessageType)
```

qdm12, line 235: Shouldn't we panic in this case? That looks like a programming error right?

lib/grandpa/network_test.go
```diff
@@ -102,32 +102,40 @@ func TestSendNeighbourMessage(t *testing.T) {
 	hash := block.Header.Hash()
 	round := uint64(7)
 	setID := uint64(33)
+
+	// waits 1.5 seconds and then finalize the block
+	// we will first send a neighbour message with the initial values
+	// ant then send another nighbor message with the finalized block values
```
qdm12, line 108:
```suggestion
	// ant then send another neighbour message with the finalized block values
```

lib/grandpa/network_test.go
```diff
@@ -102,32 +102,40 @@ func TestSendNeighbourMessage(t *testing.T) {
...
+		nm, ok := msg.(*NeighbourPacketV1)
 		require.True(t, ok)
 		require.Equal(t, expected, nm)
 	}
 
-	require.Equal(t, expected, gs.neighbourMessage)
-
 	select {
-	case <-time.After(time.Second * 2):
+	case <-time.After(time.Second * 1):
```
qdm12, line 129: nit

```suggestion
	case <-time.After(time.Second):
```

lib/grandpa/network.go
EclesioMeloJunior: Not really, sometimes is just an unsupported neighbor packet, and since this code deals with outside data inserting a panic there could be an attack vector by allowing peers to send us wrong neighbor messages that we don't support and then panicking us

lib/grandpa/message.go
```diff
@@ -77,21 +77,57 @@ func (v *VoteMessage) ToConsensusMessage() (*ConsensusMessage, error) {
...
-	Round   uint64
-	SetID   uint64
-	Number  uint32
+// VersionedNeighbourPacket represents the enum of neighbour messages
+type VersionedNeighbourPacket scale.VaryingDataType
+
+// Index Returns VDT index
+func (VersionedNeighbourPacket) Index() uint { return 2 }
+
+func newVersionedNeighbourMessage() VersionedNeighbourPacket {
```
timwu20, line 86:
```suggestion
func newVersionedNeighbourPacket() VersionedNeighbourPacket {
```

timwu20 (changes requested):

qdm12 (approved):

lib/grandpa/message.go
```diff
@@ -77,21 +77,57 @@ func (v *VoteMessage) ToConsensusMessage() (*ConsensusMessage, error) {
...
+// Value returns the current VDT value
+func (vnp *VersionedNeighbourPacket) Value() (val scale.VaryingDataTypeValue) {
+	vdt := scale.VaryingDataType(*vnp)
+	return vdt.Value()
+}
+
+// NeighbourPacketV1 represents a network-level neighbour message
+// currently, round and setID represents a struct containing an u64
+// https://github.com/paritytech/substrate/blob/master/client/finality-grandpa/src/communication/mod.rs#L660
+type NeighbourPacketV1 struct {
```
jimjbrettj, line 112: Is there a V2?

jimjbrettj (approved):
Nice work!

lib/grandpa/message.go
EclesioMeloJunior: Currently not, but as in substrate they use the 'V1' suffix I believe will exists another version in the future

timwu20 (approved):

PR #2723, opened 2022-08-03
chore(lint): remove unneeded `nolint` comments

qdm12:
## Primary Reviewer

@timwu20


qdm12:
Not needed, this was due to user error running golangci-lint v1.47.2 locally instead of v1.47.3

PR #2724, opened 2022-08-04
chore(trie): Node `HashDigest` -> `MerkleValue`

qdm12:
## Changes

Polkadot spec refers to it as *merkle value*, and even in the implementation it's not always a hash digest, it can also be the encoding itself if the node is not the trie root and its encoding is <= 32 bytes.

## Primary Reviewer

@jimjbrettj


EclesioMeloJunior (approved):

lib/trie/database.go
```diff
@@ -406,7 +406,7 @@ func (t *Trie) getInsertedNodeHashesAtNode(n *Node, hashes map[common.Hash]struc
 	if err != nil {
 		return fmt.Errorf(
 			"cannot encode and hash node with hash 0x%x: %w",
```
qdm12, line 408:
```suggestion
			"encoding and hashing node with Merkle value 0x%x: %w",
```

lib/trie/proof/verify.go
```diff
@@ -124,7 +124,7 @@ func loadProof(digestToEncoding map[string][]byte, n *node.Node) (err error) {
 		// for inlined child nodes, the hash digest field is the
 		// encoding itself instead of the encoding hash digest, so we
 		// use the `merkleValue` variable name below to avoid confusion.
```
qdm12, line 126:
```suggestion
```

lib/trie/trie.go
```diff
@@ -101,7 +101,7 @@ func updateGeneration(currentNode *Node, trieGeneration uint64,
 
 	// The hash of the node from a previous snapshotted trie
 	// is usually already computed.
-	deletedHashBytes := currentNode.HashDigest
+	deletedHashBytes := currentNode.MerkleValue
```
qdm12, line 104: note this is changed in #2725

noot (approved):

PR #2725, opened 2022-08-04
fix(trie): use direct Merkle value for database keys

qdm12:
## Changes

The problem is as follows in this example.

For a node `A` encoding to `{1, 2}`, its database key is `{0, 0, 0, ..., 1, 2}`. Another node `B` with > 32B encoding could (unlikely but still) hash to the same value `{0, 0, 0, ..., 1, 2}` as for `A` and that would conflict in the database.

The solution is to NOT force the database key to be 32B always, and to keep it plainly as the Merkle value of the node (32B or less). This should also occupy less space in the database eventually.

- Storage KV database node key has now the length of the Merkle value and is no longer forced to 32B
- Renaming `XXXHashesSet` to `XXXMerkleValues` since it's not just hashes
- Renaming `deletedKeys` to `deletedMerkleValueToBlockNumber` in `dot/state`
- Renaming methods `XXXNodeHashes` to `XXXMerkleValues`

## Tests

```sh
```

## Primary Reviewer

@timwu20

### DIFF
--- a/lib/trie/trie.go
+++ b/lib/trie/trie.go
@@ -101,9 +101,9 @@ func updateGeneration(currentNode *Node, trieGeneration uint64,
 
 	// The hash of the node from a previous snapshotted trie
 	// is usually already computed.
-	deletedMerkleValueBytes := currentNode.MerkleValue
-	if len(deletedMerkleValueBytes) > 0 {
-		deletedMerkleValueString := string(deletedMerkleValueBytes)
+	deletedMerkleValue := currentNode.MerkleValue
+	if len(deletedMerkleValue) > 0 {
+		deletedMerkleValueString := string(deletedMerkleValue)
 		deletedMerkleValues[deletedMerkleValueString] = struct{}{}
 	}
 
--- a/lib/trie/trie.go
+++ b/lib/trie/trie.go
@@ -18,10 +18,10 @@ var EmptyHash, _ = NewEmptyTrie().Hash()
 
 // Trie is a base 16 modified Merkle Patricia trie.
 type Trie struct {
-	generation  uint64
-	root        *Node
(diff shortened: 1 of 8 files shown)


noot (approved):
lgtm!

lib/trie/trie.go
```diff
@@ -94,17 +94,17 @@ func (t *Trie) prepBranchForMutation(currentBranch *Node,
...
 
 	// The hash of the node from a previous snapshotted trie
 	// is usually already computed.
-	deletedHashBytes := currentNode.MerkleValue
-	if len(deletedHashBytes) > 0 {
-		deletedHash := common.BytesToHash(deletedHashBytes)
-		deletedHashes[deletedHash] = struct{}{}
+	deletedMerkleValueBytes := currentNode.MerkleValue
+	if len(deletedMerkleValueBytes) > 0 {
+		deletedMerkleValueString := string(deletedMerkleValueBytes)
```
jimjbrettj, line 106: nit deletedMerkleValue instead, string feels redundant

jimjbrettj:
It's enforced that the node value <= 32B yes? Seems like but just wanted to double check

jimjbrettj (approved):
lgtm

qdm12:
> It's enforced that the node value <= 32B yes? Seems like but just wanted to double check

The Merkle value is <= 32B (node scale encoding if it is smaller than 32B, otherwise the blake2b hash digest of the node scale encoding)

lib/trie/trie.go
qdm12: Good point, renamed to `deletedMerkleValue` as suggested

timwu20:
Just looked at this. Great work!

PR #2726, opened 2022-08-04
Author a sibling block in case best block's slot is same as current slot

kishansagathiya:
## Changes

- While authoring a block in a given slot S, it could happen that best block was also authored in the slot S. In this case, we can't create a new block as a child of best block, because then we would have a chain where slot numbers don't increase.
- In such scenario, we should author a sibling block to best block. That is we should author a block as a child of best block's parent.
- After authoring such a block, we would have two blocks with same parent, same block number and same slot. This would create a fork, but that would eventually get resolved.

## Tests

```sh
```

## Issues
Fixes https://github.com/ChainSafe/gossamer/issues/2636
## Primary Reviewer

@timwu20


lib/babe/babe.go
```diff
@@ -497,6 +497,21 @@ func (b *Service) handleSlot(epoch, slotNum uint64,
 		return err
 	}
 
+	slotNumOfTheBestBlock, err := b.blockState.GetSlotForBlock(parentHeader.Hash())
+	if err != nil {
+		return fmt.Errorf("could not get slot for block %s: %w", parentHeader.Hash(), err)
```
qdm12, line 502: nit

- do not prefix error wrappings with `failed`/`cannot`/`could not` since it builds up repetition for deeply wrapped errors
- the parent header hash context should be wrapped by `GetSlotForBlock`, not here

```suggestion
		return fmt.Errorf("getting slot for block: %w", err)
```

lib/babe/babe.go
```diff
@@ -497,6 +497,21 @@ func (b *Service) handleSlot(epoch, slotNum uint64,
 		return err
 	}
 
+	slotNumOfTheBestBlock, err := b.blockState.GetSlotForBlock(parentHeader.Hash())
```
qdm12, line 500: nit rename to `bestBlockSlotNum`? `slotNumOfTheBestBlock` feels a bit poetically verbose ✍️

lib/babe/babe.go
```diff
@@ -497,6 +497,21 @@ func (b *Service) handleSlot(epoch, slotNum uint64,
...
+	if err != nil {
+		return fmt.Errorf("could not get slot for block %s: %w", parentHeader.Hash(), err)
+	}
+
+	if slotNumOfTheBestBlock == slotNum {
+		// pick parent of best block instead to handle slot
+		newParentHeader, err := b.blockState.GetHeader(parentHeader.ParentHash)
+		if err != nil {
+			return fmt.Errorf("could not get header for hash %s: %w", parentHeader.ParentHash, err)
```
qdm12, line 509: nit:

- the hash should be added as context by `GetHeader`, not here
- removing the `could not` prefix
- adding the `grandparent` context? I guess parent hash of parent header is grandparent right?

```suggestion
			return fmt.Errorf("getting grandparent header: %w", err)
```

lib/babe/babe.go
```diff
@@ -497,6 +497,21 @@ func (b *Service) handleSlot(epoch, slotNum uint64,
...
 	}
 
+	slotNumOfTheBestBlock, err := b.blockState.GetSlotForBlock(parentHeader.Hash())
+	if err != nil {
+		return fmt.Errorf("could not get slot for block %s: %w", parentHeader.Hash(), err)
+	}
+
+	if slotNumOfTheBestBlock == slotNum {
+		// pick parent of best block instead to handle slot
+		newParentHeader, err := b.blockState.GetHeader(parentHeader.ParentHash)
```
qdm12, line 507: nit rename to `grandParentHeader`? 🤔

qdm12 (approved):

lib/babe/babe.go
kishansagathiya: can't use the word grandparent. Basically we are looking for a block on which we can put our new block. So it is always parent.

lib/babe/babe.go
EclesioMeloJunior: > since it builds up a repetition for deeply wrapped errors

I don't think so, would be much easier to read eg. `cannot build block: cannot get slot number: failed to get block from the tree` than `cannot build block: getting slot number: retrieving block from the tree`, I think writing better log messages is needed but in this case I think we should keep as it is

lib/babe/babe.go
```diff
@@ -497,6 +497,21 @@ func (b *Service) handleSlot(epoch, slotNum uint64,
 		return err
 	}
 
+	slotNumOfTheBestBlock, err := b.blockState.GetSlotForBlock(parentHeader.Hash())
+	if err != nil {
+		return fmt.Errorf("could not get slot for block %s: %w", parentHeader.Hash(), err)
+	}
+
+	if slotNumOfTheBestBlock == slotNum {
```
EclesioMeloJunior, line 505: should we check if `slotNumOfTheBestBlock >= slotNum` as well since it is possible to happen to build a block on slot `S` and the best block number was built in slot `S + 2`, right?

EclesioMeloJunior (changes requested):

lib/babe/babe.go
EclesioMeloJunior: > since it builds up a repetition for deeply wrapped errors

I don't think so, would be much easier to read eg. `cannot build block: cannot get slot number: failed to get block from the tree` than `cannot build block: getting slot number: retrieving block from the tree`, I think writing better log messages is needed but in this case I think we should keep as it is

qdm12: I don't see how `cannot build block: cannot get slot number: (more cannots): retrieving block from the tree` is not worst than `building block: getting slot number: (more verbings): retrieving block from the tree failed`. Only the wrapped (bottom) error should state is failed I would say.

I was a die hard fan or prefixing all wrappings with `cannot` for months if not years to be honest, but seeing horribly long repetitive cannot wrappings in error logs, I decided it was time for change😿 

Let's see what others feel like, I'll send the link to this on Slack, such that we use the convention for errors and error wrappings.

lib/babe/babe.go
kishansagathiya: > it is possible to happen to build a block on slot S and the best block number was built in slot S + 2

That does not seem possible. For that best block would have to be built in future, which does not make sense.

kishansagathiya: Slot number are calculated using absolute clock and not logical clock. Each belongs to an actual time interval in the real world. So, if we are at slot S and out best block was built in slot S+2, that could mean one of three things
1) Best block is invalid. This would mean that we are not validating blocks properly. This would mean a bug in our code.
2) Our local time is not accurate.
3) We are not calculating slot number properly. This would mean a bug in our code.

If `slotNumOfTheBestBlock > slotNum` happens, I don't see if it makes sense to produce further blocks. It seems like the right place to panic.

lib/babe/babe.go
kishansagathiya: > adding the grandparent context? I guess parent hash of parent header is grandparent right?

can't use the word grandparent. Basically we are looking for a block on which we can put our new block (parent of the block to be minted). So it is always parent.

kishansagathiya: > the hash should be added as context by GetHeader, not here

blockState.GetHeader doesn't do all that. I have passed hash in the argument, it is not unreasonable to add that here in the error.

> Let's see what others feel like, I'll send the link to this on Slack, such that we use the convention for errors and error wrappings.

I am happy with the current style of error wording. don't have a problem with `cannot` or `could not`.

lib/babe/babe.go
kishansagathiya: added a check for `slotNumOfTheBestBlock > slotNum`

lib/babe/babe.go
qdm12: > I am happy with the current style of error wording. don't have a problem with cannot or could not.

Sure up to you. On the other hand, really the hash should not be added here, or it will be twice in the error message. Only the last function handling the hash should wrap it as context to the error. Especially important since hashes are not that short.

kishansagathiya: > really the hash should not be added here, or it will be twice in the error message.

But none of the errors in GetHeader are wrapping hash in them 

https://github.com/ChainSafe/gossamer/blob/b56d7f91fd4f9603473f834b94d5e26b34ebd306/dot/state/block.go#L202-L233

lib/babe/babe.go
kishansagathiya: addressed

lib/babe/babe.go
```diff
@@ -497,6 +497,25 @@ func (b *Service) handleSlot(epoch, slotNum uint64,
...
 	}
 
+	bestBlockSlotNum, err := b.blockState.GetSlotForBlock(parentHeader.Hash())
+	if err != nil {
+		return fmt.Errorf("could not get slot for block %s: %w", parentHeader.Hash(), err)
+	}
+
+	if bestBlockSlotNum > slotNum {
+		return errLaggingSlot
+	}
```
timwu20, line 507: can we write a unit test to get coverage for these lines?

lib/babe/babe.go
```diff
@@ -497,6 +497,25 @@ func (b *Service) handleSlot(epoch, slotNum uint64,
...
+
+	if bestBlockSlotNum == slotNum {
+		// pick parent of best block instead to handle slot
+		newParentHeader, err := b.blockState.GetHeader(parentHeader.ParentHash)
+		if err != nil {
+			return fmt.Errorf("could not get header for hash %s: %w", parentHeader.ParentHash, err)
+		}
+
+		parentHeader = newParentHeader
+	}
```
timwu20, line 517: can we write a unit test to test this case?

timwu20 (changes requested):

kishansagathiya:
>  if gossamer imports block X with contains slot 10 but our next authoring slot is 8 so X should wait until we reach that or a great slot

@EclesioMeloJunior,  what do you mean when you say X should wait? Do you mean we should not import block X?
Your described scenario could happen because of two reason:
- We are lagging behind or
- X is an invalid block

I think in this scenario, we should just author blocks regularly. We author for slot 8 even though we received a block with slot 10.  It is problematic only if in a chain (without forks, so not a tree), slots of all block should be increasing. You could author a block for slot 8 and create a fork.

kishansagathiya:
I was facing problems with tests. Seeing a lot of 
```
2022-08-23T13:58:29+05:30 WARN failed to apply extrinsic 0xad018400d43593c715fdd31c61141abd04a99fd6822c8558854ccde39a5684e7a56da27d01e08635fff8fd7b307827575dbd8f17a96ad9de6c15e690806d8c22ef410ae2668ac8f88e697f73414a20fd905556860688a9e8b661c819f2ee00774b5023d082000000000108abcd: transaction validity error: bad proof	pkg=babe
2022-08-23T13:58:31+05:30 CRIT target=runtime message=panicked at 'Timestamp must be updated once in the block', /home/elizabeth/substrate/frame/timestamp/src/lib.rs:170:13	ext_logging_log_version_1	pkg=runtime module=go-wasmer
2022-08-23T13:58:31+05:30 WARN failed to handle slot 553747769: cannot finalise block: running runtime function: Failed to call the `BlockBuilder_finalize_block` exported function.	pkg=babe
```

noot mentioned that this could be because of a race condition. Below logs might help

https://gist.github.com/kishansagathiya/d64d70dc48bbde60c7ceb3813a4a266c

dot/types/authority.go
```diff
@@ -16,7 +16,9 @@ import (
 
 // Authority struct to hold authority data
 type Authority struct {
-	Key    crypto.PublicKey
+	Key crypto.PublicKey
+	// Weight exists for potential improvements in the protocol and could
+	// have a use-case in the future. In polkadot all authorities have the weight = 1.
```
timwu20, line 21: 👍

timwu20 (approved):

PR #2736, opened 2022-08-09
lib/trie: add `Version` type

qdm12:
## Changes

- `String()` method
- `ParseVersion` function

## Tests

```sh
```

## Issues

#2418 

## Primary Reviewer

@jimjbrettj


lib/trie/version.go
```diff
@@ -0,0 +1,62 @@
...
+var ErrParseVersion = errors.New("parsing version failed")
+
+// ParseVersion parses a state trie version string.
+func ParseVersion(s string) (version Version, err error) {
+	switch {
+	case strings.EqualFold(s, V0.String()):
+		return V0, nil
+	default:
+		return version, fmt.Errorf("%w: %q must be %s",
+			ErrParseVersion, s, orStrings([]string{V0.String()}))
```
qdm12, line 39: Note this code is made adaptable to support multiple versions (v1 in the foreseeable future)

lib/trie/version.go
```diff
@@ -0,0 +1,66 @@
...
+
+// ParseVersion parses a state trie version string.
+func ParseVersion(s string) (version Version, err error) {
+	switch {
+	case strings.EqualFold(s, V0.String()):
+		return V0, nil
+	default:
+		choices := orStrings([]string{V0.String()})
+		return version, fmt.Errorf("%w: %q must be %s",
+			ErrParseVersion, s, choices)
```
EclesioMeloJunior, line 43: I believe the usage of `strings.Join` could be enough to show the amount of options you have to `ParseVersion` and more simple to understand:
```suggestion
		options := strings.Join([]string{V0.String()}, ", ")
		return version, fmt.Errorf("%w: %q should be one of %s",
			ErrParseVersion, s, options)
```

EclesioMeloJunior (changes requested):

lib/trie/version.go
qdm12: I midagree (disagree and agree!)

Re-reading this, this is just brainiac thoughts spreaded in code, and simple code is important.
BUT it's also nice to see `"V2" must be V0 or V1` instead of `"V2" should be one of V0, V1`.

Now since it makes you happy, let's go with your suggestion 😉 Change to:

```go
		options := strings.Join([]string{V0.String()}, ", ")
		return version, fmt.Errorf("%w: %q must be one of %s",
			ErrParseVersion, s, options)
```

EclesioMeloJunior (approved):

lib/trie/version.go
qdm12: I actually just changed a bit to be `fmt.Errorf("%w: %q must be %s", ErrParseVersion, s, V0)` since there is only v0 for now. I'll update the error to be `fmt.Errorf("%w: %q must be %s or %s", ErrParseVersion, s, V0, V1)` when adding V1, otherwise error messages are just plain weird.

timwu20 (approved):

PR #2740, opened 2022-08-09
chore(dot/rpc): `NewStateRuntimeVersionResponse`

qdm12:
## Changes

- Add new conversion function `NewStateRuntimeVersionResponse`
- Inline `modules.ConvertAPIs` content in `NewStateRuntimeVersionResponse`
- Replace all conversions with `NewStateRuntimeVersionResponse`
- Update tests to have explicit expected API items to remove `modules.ConvertAPIs`

## Tests

```sh
```

## Issues

Created from a comment in #2673 

## Primary Reviewer

@timwu20

### DIFF
--- a/dot/rpc/subscription/listeners_test.go
+++ b/dot/rpc/subscription/listeners_test.go
@@ -344,7 +344,7 @@ func TestRuntimeChannelListener_Listen(t *testing.T) {
 
 	expectedInitialVersion := modules.StateRuntimeVersionResponse{
 		SpecName: "mock-spec",
-		Apis:     modules.ConvertAPIs(nil),
+		Apis:     []interface{}{},
 	}
 
 	expectedInitialResponse := newSubcriptionBaseResponseJSON()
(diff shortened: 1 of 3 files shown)


timwu20 (approved):

dot/rpc/subscription/listeners_test.go
```diff
@@ -344,7 +344,7 @@ func TestRuntimeChannelListener_Listen(t *testing.T) {
 
 	expectedInitialVersion := modules.StateRuntimeVersionResponse{
 		SpecName: "mock-spec",
-		Apis:     modules.ConvertAPIs(nil),
+		Apis:     []interface{}{},
```
jimjbrettj, line 347: Does this still need to be an interface or can we be more explicit with the type?

jimjbrettj (approved):
Looks good, just one comment

dot/rpc/subscription/listeners_test.go
qdm12: A bit out of scope here, I'm just replacing the return from `modules.ConvertAPIs(nil)`.

Anyway, yes, we need this, since we use it with `[]interface{}{[]interface{}{"0x0102030405060708", uint32(99)}}}` so we kinda need interfaces to mix types in the same slice.

PR #2741, opened 2022-08-10
feat(pkg/scale): add `Encoder` with `Encode` method

### LINKED ISSUE #1897: pkg/scale: Add `Encoder` and `NewEncoder(io.Writer)`

qdm12:
## Changes

- Change `encodeState` to use `io.Writer` instead of `bytes.Buffer`
- Define `Encoder` with `Encode(value interface{}) error` method
- Define constructor `NewEncoder(writer io.Writer) *Encoder`
- Add unit tests for encoder

## Tests

```sh
go test github.com/ChainSafe/gossamer/pkg/scale
```

## Issues

Fixes #1897

~Blocks PR #2673 which (long shot) blocks #2418~

## Primary Reviewer

@timwu20

### DIFF
--- a/pkg/scale/encode_test.go
+++ b/pkg/scale/encode_test.go
@@ -17,12 +17,16 @@ import (
 func Test_NewEncoder(t *testing.T) {
 	t.Parallel()
 
+	cache.Lock()
+	defer cache.Unlock()
+
 	writer := bytes.NewBuffer(nil)
 	encoder := NewEncoder(writer)
 
 	expectedEncoder := &Encoder{
 		encodeState: encodeState{
-			Writer: writer,
+			Writer:                 writer,
+			fieldScaleIndicesCache: cache,
 		},
 	}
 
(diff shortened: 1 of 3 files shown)


pkg/scale/encode_test.go
```diff
@@ -4,12 +4,48 @@
...
 
 import (
+	"bytes"
 	"math/big"
 	"reflect"
 	"strings"
 	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
```
qdm12, line 14: Let me know if you feel this is inappropriate to use to keep zero-dependency on scale.
Although right now since it's part of Gossamer it doesn't change anything really.

jimjbrettj: Tagging @timwu20 here, if we are planning to migrate scale out of gossamer might be better to keep dependencies out?

timwu20: https://github.com/ChainSafe/gossamer/pull/2683#discussion_r942741296

EclesioMeloJunior (approved):

PR #2743, opened 2022-08-10
chore(log): fix spelling and add padding for logger

jimjbrettj:
## Changes

There were spelling errors in the log levels to maintain a constant string length for all the levels. I have fixed spelling and added padding to make it cleaner and maintain consistency.

## Tests

```sh
go test ./internal/log
```

## Primary Reviewer

@qdm12

### DIFF
--- a/dot/config_test.go
+++ b/dot/config_test.go
@@ -446,8 +446,8 @@ func TestLogConfig_String(t *testing.T) {
 		{
 			name:      "default case",
 			logConfig: LogConfig{},
-			want: "core: CRIT , digest: CRIT , sync: CRIT , network: CRIT , rpc: CRIT , state: CRIT , " +
-				"runtime: CRIT , block producer: CRIT , finality gadget: CRIT ",
+			want: "core: CRITICAL, digest: CRITICAL, sync: CRITICAL, network: CRITICAL, rpc: CRITICAL, " +
+				"state: CRITICAL, runtime: CRITICAL, block producer: CRITICAL, finality gadget: CRITICAL",
 		},
 		{
 			name: "change fields case",
@@ -462,8 +462,8 @@ func TestLogConfig_String(t *testing.T) {
 				BlockProducerLvl:  log.Warn,
 				FinalityGadgetLvl: log.Error,
 			},
-			want: "core: DEBUG, digest: INFO , sync: WARN , network: ERROR, rpc: CRIT , state: DEBUG," +
-				" runtime: INFO , block producer: WARN , finality gadget: ERROR",
+			want: "core: DEBUG   , digest: INFO    , sync: WARN    , network: ERROR   , rpc: CRITICAL, " +
+				"state: DEBUG   , runtime: INFO    , block producer: WARN    , finality gadget: ERROR   ",
 		},
 	}
 	for _, tt := range tests {
--- a/dot/config_test.go
+++ b/dot/config_test.go
@@ -446,8 +446,8 @@ func TestLogConfig_String(t *testing.T) {
 		{
 			name:      "default case",
 			logConfig: LogConfig{},
-			want: "core: CRIT, digest: CRIT, sync: CRIT, network: CRIT, rpc: CRIT, state: CRIT, runtime: CRIT, " +
-				"block producer: CRIT, finality gadget: CRIT",
+			want: "core: CRIT , digest: CRIT , sync: CRIT , network: CRIT , rpc: CRIT , state: CRIT , " +
+				"runtime: CRIT , block producer: CRIT , finality gadget: CRIT ",
 		},
 		{
 			name: "change fields case",
@@ -462,8 +462,8 @@ func TestLogConfig_String(t *testing.T) {
 				BlockProducerLvl:  log.Warn,
 				FinalityGadgetLvl: log.Error,
 			},
-			want: "core: DBUG, digest: INFO, sync: WARN, network: EROR, rpc: CRIT, state: DBUG, runtime: INFO, " +
-				"block producer: WARN, finality gadget: EROR",
+			want: "core: DEBUG, digest: INFO , sync: WARN , network: ERROR, rpc: CRIT , state: DEBUG," +
+				" runtime: INFO , block producer: WARN , finality gadget: ERROR",
 		},
 	}
 	for _, tt := range tests {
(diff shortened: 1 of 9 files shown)


internal/log/level.go
```diff
@@ -98,13 +98,13 @@ func ParseLevel(s string) (level Level, err error) {
 		return Trace, nil
 	case Debug.String(), "DEBUG":
```
qdm12, line 99:
```suggestion
	case Debug.String():
```

internal/log/level.go
```diff
@@ -98,13 +98,13 @@ func ParseLevel(s string) (level Level, err error) {
...
 		return Debug, nil
-	case Info.String():
+	case Info.String(), "INFO":
 		return Info, nil
-	case Warn.String():
+	case Warn.String(), "WARN":
 		return Warn, nil
 	case Error.String(), "ERROR":
```
qdm12, line 105:
```suggestion
	case Error.String():
```

internal/log/level.go
```diff
@@ -98,13 +98,13 @@ func ParseLevel(s string) (level Level, err error) {
 		return Trace, nil
 	case Debug.String(), "DEBUG":
 		return Debug, nil
-	case Info.String():
+	case Info.String(), "INFO":
 		return Info, nil
-	case Warn.String():
+	case Warn.String(), "WARN":
```
qdm12, line 103: ahh not that trivial, good catch!

internal/log/level.go
```diff
@@ -36,17 +36,17 @@ const (
...
+		return "INFO "
 	case Warn:
-		return "WARN"
+		return "WARN "
 	case Error:
-		return "EROR"
+		return "ERROR"
 	case Critical:
-		return "CRIT"
+		return "CRIT "
```
qdm12, line 49: Shouldn't we change this to `CRITICAL`?
Opinions maybe @kishansagathiya @danforbes 🤔 
But then we'll need 4 spaces suffix for all other levels (which is fine but still). Maybe @jimjbrettj you can check how the code concat the level with the next string, it might be `\t` and in that case just change it to a single space? That would thus look like `4`+`1` spaces in most log lines, and to `1` space only for `CRITICAL` level.

qdm12 (commented):
👍

internal/log/level.go
danforbes: I would probably prefer `CRITICAL` but it's not a hill I'd die on.

jimjbrettj: @qdm12 updated!

internal/log/level.go
jimjbrettj: added these back with changing padding for CRITICAL

dot/config_test.go
```diff
@@ -462,8 +462,8 @@ func TestLogConfig_String(t *testing.T) {
 				BlockProducerLvl:  log.Warn,
 				FinalityGadgetLvl: log.Error,
 			},
-			want: "core: DBUG, digest: INFO, sync: WARN, network: EROR, rpc: CRIT, state: DBUG, runtime: INFO, " +
-				"block producer: WARN, finality gadget: EROR",
+			want: "core: DEBUG   , digest: INFO    , sync: WARN    , network: ERROR   , rpc: CRITICAL, " +
+				"state: DEBUG   , runtime: INFO    , block producer: WARN    , finality gadget: ERROR   ",
```
qdm12, line 466: Oh-oh... maybe add a method such as `IndentedString() string` for the one adding spaces? And keep `String() string` without the spaces? 🤔 
Since that config with spaces is just megaweird

qdm12:
Another thought, maybe let's just have an unexported method `func (l Level) getSpaceSuffixes() int` with a switch which return the number of spaces to add after when logging? Otherwise the whole logging code/config strings look rather weird.

kishansagathiya (approved):

dot/config_test.go
jimjbrettj: > Another thought, maybe let's just have an unexported method `func (l Level) getSpaceSuffixes() int` with a switch which return the number of spaces to add after when logging? Otherwise the whole logging code/config strings look rather weird.

@qdm12 Im a bit confused, how does this fix the strings looking weird? I can see how it cleans up the expected string, but if we want padding, for tests that test logs wouldn't the spaces still be present in the string (like above test)?

qdm12: So you would return the level as before (sorta) with `String()` (`INFO`, ..., `CRITICAL`)

And then have:

```go
func (l Level) formatWithTrailingSpaces() (formatted string) {
  switch l {
  case Info:
    return "info    "
  // ...
  case Critical:
    return "critical "
  }
}
```

and use `level.formatWithTrailingSpaces()` in the `log` function in the logger package, right before writing to the io.Writer.

jimjbrettj: ah okay yes that sounds good thank you

internal/log/level.go
```diff
@@ -73,7 +71,7 @@ func (level Level) ColouredString() (s string) {
 	}
 
 	c := color.New(attribute)
-	return c.Sprint(level.String())
+	return c.Sprintf("%-8s", level.String())
```
qdm12, line 74: Awesome 👍 Didn't know it would handle it properly that well!

internal/log/level.go
```diff
@@ -94,17 +92,17 @@ func ParseLevel(s string) (level Level, err error) {
...
 		return Debug, nil
 	case Info.String():
 		return Info, nil
 	case Warn.String():
 		return Warn, nil
-	case Error.String(), "ERROR":
+	case Error.String():
 		return Error, nil
-	case Critical.String(), "CRITICAL":
+	case Critical.String(), "CRIT":
```
qdm12, line 105: Let's remove support for `CRIT`? I don't think we use it anywhere, and we can just update flag we use to use `critical` instead of `crit`

```suggestion
	case Critical.String():
```

internal/log/level_test.go
```diff
@@ -11,7 +11,7 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-func Test_Level_ColouredString(t *testing.T) {
+func Test_Level_Format(t *testing.T) {
```
qdm12, line 14: nit nit

```suggestion
func Test_Level_format(t *testing.T) {
```

qdm12 (approved):
Awesome! 👍

internal/log/level.go
jimjbrettj: Me neither! Was pleasantly suprised at this

internal/log/level.go
jimjbrettj: yeah removing it broke tests and I wasn't sure why, imagine its the flag then will go look

PR #2688, opened 2022-07-22
fix(lib/grandpa): avoid spamming round messages

EclesioMeloJunior:
## Changes

1. Introduces the service pattern `lib/grandpa/engine.go`:
- GRANDPA counts with 2 mains services (`finalizationEngine`, `handleVotingRound`) and 1 orchestrator service (`finalizationHandler`)
- `finalizationEngine`: is responsible for deciding when the node does the voting, it owns the `determinePrevoteTimer`, `determinePrecommitTimer` and once the node does a pre-commit it starts attempting to finalize checking the vote thresholds, this engine ends once a valid commit message arrives.

- `handleVotingRound`: holds a channel from the finalization engine and is responsible for executes actions (`determinePrevote`, `determinePrecommit`, `finalize`, `alreadyFinalized`) based on messages sent through the finalization engine channel

- `finalizationHandler`: holds both previous services and is responsible for initiating the current round, bootstrapping both services, and waiting for them to finish to proceed to the next round. This service exposes an `observableErrs`, which is a fanout channel that aggregates errors from `finalizationEngine` and `handleVotingRound`, the caller can listen and handle those errors.

## Tests

```
go test github.com/ChainSafe/gossamer/lib/grandpa
```

Currently, the GRANDPA test suite at the development branch takes `234.980s` while in this branch the GRANDPA test suite takes `154.471s` a reduction of `80,5s` and in a branch called `eclesio/add/grandpa-paralell-tests` where I included `t.Parallel()` to all GRANDPA tests, it takes `115.824s` a reduction of `119,15s` from the development branch.

## Issues

- Closes #2700

## Primary Reviewer

@timwu20

Issue #2505, opened 2022-04-28
Polite GRANDPA


danforbes:
## Task summary

- Implement "polite GRANDPA" in Gossamer. Polite GRANDPA is a mechanism that allows participants in the GRANDPA protocol to score one another in order allow participants to prefer certain peers over others. Per @qdm12, per a conversation with Andre at Parity:
  - It's impolite to get a message for a round > state round + 1
  - It's impolite to get a message for a round < state round - 1
  - Message signature verification fails
  - Any other validation failure due to the message being malformed.
  - The relevant Gossamer function is `validateVoteMessage` in `lib/grandpa/vote_message.go`

## Specification reference

- Refer to the following GRANDPA implementations in Substrate: [`client/finality-grandpa/src/communication/gossip.rs`](https://github.com/paritytech/substrate/blob/88731ed619f6be1280f01cc8f02f10e2fb4cf199/client/finality-grandpa/src/communication/gossip.rs#L19) and [`client/finality-grandpa/src/communication/mod.rs`](https://github.com/paritytech/substrate/blob/88731ed619f6be1280f01cc8f02f10e2fb4cf199/client/finality-grandpa/src/communication/mod.rs#L19)

jimjbrettj:
Upon investigating we found these relevant parts of the grandpa package:
- https://github.com/paritytech/substrate/blob/88731ed619f6be1280f01cc8f02f10e2fb4cf199/client/finality-grandpa/src/communication/mod.rs#L93
- https://github.com/paritytech/substrate/blob/88731ed619f6be1280f01cc8f02f10e2fb4cf199/client/finality-grandpa/src/communication/gossip.rs#L434

We think it will be useful to have a similar misbehavior function where we can calculate the costs based on the type of misbehavior within grandpa. 

We will also create a README (or some other doc) to more clearly document reporting for polite grandpa.

jimjbrettj:
The list of tasks involved in this issue will be split between @EclesioMeloJunior and myself. The plan currently is to work together to develop the framework for handling reporting and then to divide and conquer implementing each specific behavior. This will involve finding all cases where this behavior is to be reported and handling it as well as adding to the documentation details about the behavior and where it is used.

- [ ] Create README to document how polite grandpa reporting works (possibly to be published somewhere since current documentation is very poor)
- [ ] Create a framework for handling reporting for polite grandpa within the grandpa package. This includes defining the costs as well as the misbehavior function.
- [ ] PAST_REJECTION
[...]

PR #2745, opened 2022-08-10
feat(lib/grandpa): improve `determinePrevote` method

EclesioMeloJunior:
## Changes

- Ensure the block we vote on belongs to the same chain as `latestFinalizedHeader`
- Don't include the `BeforeBestBlockBy` voting rule as I did not find too much information about it (probably include in other PR) refs
  - https://github.com/paritytech/polkadot/blob/270fb1e576ae2199c8b1803ec91a9a749ded6779/node/service/src/lib.rs#L1250
  - https://github.com/paritytech/substrate/blob/20b5aac7e09e8b25abaa48c6781e653b36ba3ca0/client/finality-grandpa/src/voting_rule.rs#L92
## Tests

```
N/A
```
## Issues

- Related to PR #2688
- Related to #2505

## Primary Reviewer

@timwu20


lib/grandpa/grandpa.go
```diff
@@ -679,6 +679,44 @@ func (s *Service) deleteVote(key ed25519.PublicKeyBytes, stage Subround) {
 	}
 }
 
+// implements `BeforeBestBlockBy` a custom voting rule that guarantees that our vote is always
+// behind the best block by at least N blocks, unless the base number is < N blocks behind the
+// best, in which case it votes for the base.
+// (https://github.com/paritytech/substrate/blob/1fd71c7845d6c28c532795ec79106d959dd1fe30/client/finality-grandpa/src/voting_rule.rs#L92)
+func (s *Service) determineBestHeaderToPrevote(finalizedHeader, bestBlockHeader *types.Header) (
+	headerToPrevote *types.Header, err error) {
+	gensisHash := s.blockState.GenesisHash()
```
qdm12, line 688: typo rename to `genesisHash`

lib/grandpa/grandpa.go
```diff
@@ -679,6 +679,44 @@ func (s *Service) deleteVote(key ed25519.PublicKeyBytes, stage Subround) {
...
+	isGenesisHash := gensisHash.Equal(bestBlockHeader.Hash())
```
qdm12, line 689: nit you could cache `bestBlockHeader.Hash()` in a variable `bestBlockHash := bestBlockHeader.Hash()` to avoid calling `.Hash()` a bunch of times.

lib/grandpa/grandpa.go
```diff
@@ -679,6 +679,44 @@ func (s *Service) deleteVote(key ed25519.PublicKeyBytes, stage Subround) {
...
+	if isGenesisHash || finalizedHeader.Hash().Equal(bestBlockHeader.Hash()) {
+		return bestBlockHeader, nil
+	}
+
+	isDescendant, err := s.blockState.IsDescendantOf(finalizedHeader.Hash(), bestBlockHeader.Hash())
+	if err != nil {
+		return headerToPrevote, fmt.Errorf("determine ancestry: %w", err)
```
qdm12, line 696: nit let's return explicitely `nil` to avoid any confusion 😉 

```suggestion
		return nil, fmt.Errorf("determining ancestry: %w", err)
```

lib/grandpa/grandpa.go
```diff
@@ -679,6 +679,44 @@ func (s *Service) deleteVote(key ed25519.PublicKeyBytes, stage Subround) {
...
+	}
+
+	if !isDescendant {
+		return headerToPrevote, fmt.Errorf("%w: %s is not ancestor of %s",
```
qdm12, line 700: nit

```suggestion
		return nil, fmt.Errorf("%w: %s is not ancestor of %s",
```

lib/grandpa/grandpa.go
```diff
@@ -679,6 +679,44 @@ func (s *Service) deleteVote(key ed25519.PublicKeyBytes, stage Subround) {
...
+			blocktree.ErrNoCommonAncestor, bestBlockHeader.Hash().Short(), bestBlockHeader.Hash().Short())
+	}
+
+	headerToPrevote = bestBlockHeader
+	for i := 0; i < 2; i++ {
```
qdm12, line 705: Please add a comment above that for loop explaining why we do this twice

lib/grandpa/grandpa.go
```diff
@@ -688,19 +726,15 @@ func (s *Service) determinePreVote() (*Vote, error) {
...
-	// otherwise, we simply choose the head of our chain.
-	primary := s.derivePrimary()
-	prm, has := s.loadVote(primary.PublicKeyBytes(), prevote)
-	if has && prm.Vote.Number >= uint32(s.head.Number) {
-		vote = &prm.Vote
-	} else {
-		vote = NewVoteFromHeader(bestBlockHeader)
+	headerToPrevote, err := s.determineBestHeaderToPrevote(s.head, bestBlockHeader)
+	if err != nil {
+		return nil, fmt.Errorf("determine best hash to prevote: %w", err)
```
qdm12, line 729:
```suggestion
		return nil, fmt.Errorf("determining best header to prevote: %w", err)
```

lib/grandpa/grandpa.go
```diff
@@ -678,15 +678,61 @@ func (s *Service) deleteVote(key ed25519.PublicKeyBytes, stage Subround) {
...
+	}
 
+	isDescendant, err := s.blockState.IsDescendantOf(finalizedHeader.Hash(), bestBlockHash)
+	if err != nil {
+		return nil, fmt.Errorf("determining ancestry: %w", err)
+	}
+
+	if !isDescendant {
+		return nil, fmt.Errorf("%w: %s is not ancestor of %s",
+			blocktree.ErrNoCommonAncestor, bestBlockHash.Short(), bestBlockHash.Short())
```
qdm12, line 698: the error name `ErrNoCommonAncestor` seems inappropriate here, shouldn't it be `ErrNotADescendant` instead?

lib/grandpa/grandpa.go
```diff
@@ -678,15 +678,61 @@ func (s *Service) deleteVote(key ed25519.PublicKeyBytes, stage Subround) {
...
+	}
+
+	// implements `BeforeBestBlockBy` a custom voting rule that guarantees that our vote is always
+	// behind the best block by at least N blocks, unless the base number is < N blocks behind the
+	// best, in which case it votes for the base.
+	// (https://github.com/paritytech/substrate/blob/master/client/finality-grandpa/src/voting_rule.rs#L92)
```
qdm12, line 704: link to a permalink please, this will eventually change over time.

qdm12, line 704: Should the code below be its own function with this comment above the function signature?

lib/grandpa/grandpa.go
```diff
@@ -678,15 +678,61 @@ func (s *Service) deleteVote(key ed25519.PublicKeyBytes, stage Subround) {
...
+	headerToPrevote = bestBlockHeader
+	const beforeBestBlockBy = 2
```
qdm12, line 706: In the comment above you mention `N` blocks, so should this always be hardcoded to `2`? 🤔

lib/grandpa/grandpa.go
```diff
@@ -678,15 +678,61 @@ func (s *Service) deleteVote(key ed25519.PublicKeyBytes, stage Subround) {
...
+	for i := 0; i < beforeBestBlockBy; i++ {
+		headerToPrevote, err = s.blockState.GetHeader(headerToPrevote.ParentHash)
+		if err != nil {
+			return headerToPrevote, fmt.Errorf("get parent header: %w", err)
```
qdm12, line 710: nit

```suggestion
			return headerToPrevote, fmt.Errorf("getting parent header: %w", err)
```

lib/grandpa/grandpa.go
```diff
@@ -678,15 +678,61 @@ func (s *Service) deleteVote(key ed25519.PublicKeyBytes, stage Subround) {
...
+		}
+
+		isGenesisHash := genesisHash.Equal(headerToPrevote.Hash())
+		if finalizedHeader.Hash().Equal(headerToPrevote.Hash()) || isGenesisHash {
```
qdm12, line 714: nit it feels weird to have a var for one of the two conditions but not the other one

```suggestion
		isFinalizedHash := finalizedHeader.Hash().Equal(headerToPrevote.Hash())
		if isFinalizedHash || isGenesisHash {
```

lib/grandpa/grandpa.go
```diff
@@ -678,15 +678,61 @@ func (s *Service) deleteVote(key ed25519.PublicKeyBytes, stage Subround) {
...
+// determinePreVote determines what block is our pre-voted block for the current round
+func (s *Service) determinePreVote() (vote *Vote, err error) {
 	bestBlockHeader, err := s.blockState.BestBlockHeader()
 	if err != nil {
 		return nil, fmt.Errorf("cannot get best block header: %w", err)
 	}
 
+	headerToPrevote, err := s.determineBestHeaderToPrevote(s.head, bestBlockHeader)
+	if err != nil {
+		return nil, fmt.Errorf("determining best hash to prevote: %w", err)
```
qdm12, line 731: nit

```suggestion
		return nil, fmt.Errorf("determining best block header to prevote: %w", err)
```

lib/grandpa/grandpa_test.go
```diff
@@ -966,6 +966,14 @@ func TestDeterminePreVote_NoPrimaryPreVote(t *testing.T) {
 
 	header, err := st.Block.BestBlockHeader()
 	require.NoError(t, err)
+
+	// the prevoted block should be 2 blocks behind the best block header
+	header, err = st.Block.GetHeader(header.ParentHash)
+	require.NoError(t, err)
+
+	header, err = st.Block.GetHeader(header.ParentHash)
+	require.NoError(t, err)
```
qdm12, line 975: Why is this twice here?

Also do we cover the case where the block is 3 blocks behind?

lib/grandpa/grandpa.go
timwu20: Can you make this configurable?

PR #2664, opened 2022-07-11
feat(wasmer): implement functions using v1 state trie

qdm12:
## Changes

Progress:

- [ ] Version state trie operations `Put` and `Hash`
- [x] Implement runtime functions
  - [x] `ext_trie_blake2_256_root_version_2`
  - [x] `ext_trie_blake2_256_ordered_root_version_2`
  - [x] `ext_trie_blake2_256_verify_proof_version_2`
  - [x] `ext_default_child_storage_root_version_2`
  - [x] `ext_storage_root_version_2`
- [x] Change version from 4 bytes pointed by pointer to `uint8` read from uint32 4 bytes.
- [x] Panic instead of `logger.Error`
- [ ] Add unit tests

Related PR: https://github.com/w3f/polkadot-spec/pull/518

## Tests

```sh
go test -tags integration github.com/ChainSafe/gossamer
```

## Issues

#2418 

## Primary Reviewer

PR #2747, opened 2022-08-11
feat(wasmer): get and cache state version in instance context

qdm12:
## Changes

Note this caches the state version for usages in several import functions (see #2664)

- [x] Codec support for Core_version and state_version
- [x] Cache core version in runtime instance context
  - Set when creating instance
  - Update when updating runtime code
- [x] Figure out how to handle `Core_version` not there in test runtimes

## Tests

```sh
go test -tags integration github.com/ChainSafe/gossamer/lib/runtime/...
```

## Issues

One more step towards #2418

## Primary Reviewer

@timwu20

### DIFF
--- a/lib/runtime/wasmer/test_helpers.go
+++ b/lib/runtime/wasmer/test_helpers.go
@@ -36,8 +36,7 @@ func NewTestInstanceWithTrie(t *testing.T, targetRuntime string, tt *trie.Trie) 
 	require.NoError(t, err)
 
 	r, err := NewInstanceFromFile(runtimeFilepath, cfg)
-	require.NoError(t, err, "Got error when trying to create new VM", "targetRuntime", targetRuntime)
-	require.NotNil(t, r, "Could not create new VM instance", "targetRuntime", targetRuntime)
+	require.NoError(t, err)
 	return r
 }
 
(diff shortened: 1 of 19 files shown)


lib/runtime/wasmer/test_helpers.go
```diff
@@ -52,14 +52,25 @@ func setupConfig(t *testing.T, tt *trie.Trie, lvl log.Level, role common.Roles)
...
 	}
 
+	var versions *testVersions
+	if targetRuntime == runtime.HOST_API_TEST_RUNTIME {
+		// Force state version to 0 since the host api test runtime
+		// does not implement the Core_version call so we cannot get the
+		// state version from it.
+		versions = &testVersions{
+			stateVersion: 0,
+		}
```
qdm12, line 62: When shall we bump this to 1 for the host api runtime? When we support the v1 state trie (finally) or would you see another milestone to bump this up for the host api test runtime? 🤔

qdm12: > i guess, do you want to test the imported trie functions w the different versions? if so you might need the two diff versions (or just override it)

EclesioMeloJunior (approved):
LGTM!

edwardmack (approved):

PR #2769, opened 2022-08-22
chore(dot/core): remove trie state calls from `GetRuntimeVersion`

qdm12:
## Tests

```sh
```


qdm12:
I really don't see why we need to update the trie state to just get the runtime version. But some integration tests are failing, I'll leave it as draft for now (and work my way around it).

qdm12:
It looks like we need to call TrieState() for whatever odd reason, so closing this for now.

qdm12:
Removed in #2912

Issue #2770, opened 2022-08-22
Running `./bin/gossamer init` fails


axaysagathiya:
## Describe the bug

While trying to initialize Gossamer errors.

```
axay@axay-Inspiron-3542:~/go/src/github.com/ChainSafe/gossamer$ ./bin/gossamer init --force
2022-08-22T22:17:08+05:30 INFO     🕸️ initialising node with name asset-wine-3519, id gssmr_test, base path /home/axay/.gossamer/gssmr and genesis ./chain/gssmr/genesis-spec.json...	node.go:L128	pkg=dot
2022-08-22T22:17:08+05:30 ERROR    failed to initialise node: failed to convert genesis-spec to raw genesis: key array less than 2	main.go:L342	pkg=cmd
failed to convert genesis-spec to raw genesis: key array less than 2
```

danforbes:
This is related to #2417

EclesioMeloJunior:
hey @axaysagathiya! Thanks for pointing this out, currently our default configuration values point to a human-readable genesis file which gossamer tries to convert into a raw genesis file but it cannot because there are missing parts we should cover however, there is an alternative that the team agreed that is using the raw genesis file directly to bootstrap the node, right now I'm working on to change those default values to always use a raw genesis file.

EclesioMeloJunior:
fixed at dd2fbc9d9c7df102f049c87134b92c3851850a2f

PR #2772, opened 2022-08-22
[fix] Use Raw Specs

danforbes:
Since Gossamer cannot compile human-readable specs, it should use raw specs.

## Issues

#2770

## Primary Reviewer

@qdm12


EclesioMeloJunior:
Closing in favor of #2775

PR #2778, opened 2022-08-23
chore(config): improve `StateConfig`

qdm12:
## Changes

Really just a tiny small thing I worked on as part of the trie versioning somewhere sometime.

- Change rewind flag to be `uint` instead of `int`
- Add `setStateConfig` with toml precedence over flags
- Add String() method to dot `StateConfig`
- Add `StateConfig` to `dot/config/toml`

## Primary Reviewer

@EclesioMeloJunior

### DIFF
--- a/cmd/gossamer/config.go
+++ b/cmd/gossamer/config.go
@@ -929,9 +929,9 @@ func setDotPprofConfig(ctx *cli.Context, tomlCfg ctoml.PprofConfig, cfg *dot.Ppr
 }
 
 func setStateConfig(ctx *cli.Context, tomlCfg ctoml.StateConfig, cfg *dot.StateConfig) {
-	if tomlCfg.Rewind > 0 {
-		cfg.Rewind = tomlCfg.Rewind
-	} else if ctx.GlobalIsSet(RewindFlag.Name) {
+	if ctx.GlobalIsSet(RewindFlag.Name) {
 		cfg.Rewind = ctx.GlobalUint(RewindFlag.Name)
+	} else if tomlCfg.Rewind > 0 {
+		cfg.Rewind = tomlCfg.Rewind
 	}
 }
--- a/cmd/gossamer/config.go
+++ b/cmd/gossamer/config.go
@@ -134,10 +134,7 @@ func createDotConfig(ctx *cli.Context) (*dot.Config, error) {
 	setDotNetworkConfig(ctx, tomlCfg.Network, &cfg.Network)
 	setDotRPCConfig(ctx, tomlCfg.RPC, &cfg.RPC)
 	setDotPprofConfig(ctx, tomlCfg.Pprof, &cfg.Pprof)
-
-	if rewind := ctx.GlobalUint(RewindFlag.Name); rewind != 0 {
-		cfg.State.Rewind = rewind
-	}
+	setStateConfig(ctx, tomlCfg.State, &cfg.State)
 
 	// set system info
 	setSystemInfoConfig(ctx, cfg)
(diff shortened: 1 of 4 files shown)


jimjbrettj:
Do we want toml having precedence over the flags?

dot/config.go
```diff
@@ -162,6 +162,11 @@ type StateConfig struct {
 	Rewind uint
 }
 
+func (s *StateConfig) String() string {
+	return "" +
+		"rewind " + fmt.Sprint(s.Rewind)
```
EclesioMeloJunior, line 167: is there any reason for this empty string concatenating with the rest of the string? also change to: `fmt.Sprintf("rewind %d", s.Rewind)`

cmd/gossamer/config.go
```diff
@@ -930,3 +927,11 @@ func setDotPprofConfig(ctx *cli.Context, tomlCfg ctoml.PprofConfig, cfg *dot.Ppr
 
 	logger.Debug("pprof configuration: " + cfg.String())
 }
+
+func setStateConfig(ctx *cli.Context, tomlCfg ctoml.StateConfig, cfg *dot.StateConfig) {
+	if tomlCfg.Rewind > 0 {
+		cfg.Rewind = tomlCfg.Rewind
+	} else if ctx.GlobalIsSet(RewindFlag.Name) {
+		cfg.Rewind = ctx.GlobalUint(RewindFlag.Name)
+	}
```
EclesioMeloJunior, line 936: is there any reason to change the precedence here?

EclesioMeloJunior (changes requested):

dot/config.go
qdm12: It's just so adding fields won't create new deltas. I think originally this contained a state version flag, which I later removed, hence the form like this.

cmd/gossamer/config.go
qdm12: The toml value was ignored previously so there is no change of precedence.
This is actually a fix I guess.

qdm12:
@jimjbrettj 

> Do we want toml having precedence over the flags?

Well honestly, I'm having a hard time figuring which precedence we should use, but happy to reverse it if the flag should take  precedence (which makes more sense too). I think I set it like that because I saw toml having precedence somewhere, but maybe that's the wrong approach.

cmd/gossamer/config.go
EclesioMeloJunior: Hm, but we should change to use the flag values with higher precedence, we might open an issue to update those places

dot/config.go
EclesioMeloJunior: But seems odd this way: `"" + "rewind " + fmt.Sprint(s.Rewind)`, is it possible change to `fmt.Sprintf("rewind %d", s.Rewind)` or at least `"rewind " + fmt.Sprint(s.Rewind)`?

cmd/gossamer/config.go
qdm12: Done! Also regarding settings

> I would like to propose to (aka create an issue) refactor our settings to have meaningless zero values (i.e. nil for boolean, "" for strings that cannot be empty, nil for strings that can be empty, etc.). Any opinions on that?
> Also we set default values before settings flag / toml values which feels very wrong, it should be set on remaining field not specified by flag & toml

dot/config.go
qdm12: Done!

EclesioMeloJunior (approved):

jimjbrettj (approved):

PR #2781, opened 2022-08-24
enhancement(dot/state) Clean up old runtimes on canonical chain

### LINKED ISSUE #2767: Clean Up Old Runtimes on Canonical Chain
## Task summary

- Currently, Gossamer maintains running instances of all runtimes that were included in the canonical chain - this behavior should be changed so that only the most up-to-date runtime on the canonical chain is maintained.

edwardmack:
## Changes
- Create function `ClearRuntimes` that deletes all runtimes stored in BlockTree runtimes mapping.
- Add call to `ClearRuntimes` from within `HandleRuntimeChanges` function to clear runtimes mapping before adding updated runtime to mapping.

## Tests

```sh
 go test -tags integration github.com/ChainSafe/gossamer/dot/core -v -run TestService_HandleRuntimeChanges
```

## Issues
closes #2767 

## Primary Reviewer

@qdm12

### DIFF
--- a/dot/core/interface.go
+++ b/dot/core/interface.go
@@ -71,6 +71,7 @@ type BlockState interface {
 	GetBlockBody(hash common.Hash) (*types.Body, error)
 	HandleRuntimeChanges(newState *rtstorage.TrieState, in runtime.Instance, bHash common.Hash) error
 	GetRuntime(*common.Hash) (runtime.Instance, error)
+	GetRuntimeFromDB(hash *common.Hash) (runtime.Instance, error)
 	StoreRuntime(common.Hash, runtime.Instance)
 }
 
--- a/dot/core/interface.go
+++ b/dot/core/interface.go
@@ -84,6 +84,7 @@ type StorageState interface {
 	GetStateRootFromBlock(bhash *common.Hash) (*common.Hash, error)
 	GetStorage(root *common.Hash, key []byte) ([]byte, error)
 	GenerateTrieProof(stateRoot common.Hash, keys [][]byte) ([][]byte, error)
+	GetRuntimeFromDB(blockHash common.Hash) (instance runtime.Instance, err error)
 	sync.Locker
 }
 
--- a/dot/core/interface.go
+++ b/dot/core/interface.go
@@ -71,7 +71,7 @@ type BlockState interface {
 	GetBlockBody(hash common.Hash) (*types.Body, error)
 	HandleRuntimeChanges(newState *rtstorage.TrieState, in runtime.Instance, bHash common.Hash) error
 	GetRuntime(*common.Hash) (runtime.Instance, error)
-	GetRuntimeFromDB(hash *common.Hash) (runtime.Instance, error)
+	GetRuntimeFromDB(blockHash *common.Hash) (instance runtime.Instance, err error)
 	StoreRuntime(common.Hash, runtime.Instance)
 }
 
--- a/dot/core/interface.go
+++ b/dot/core/interface.go
@@ -70,7 +70,7 @@ type BlockState interface {
 	SubChain(start, end common.Hash) ([]common.Hash, error)
 	GetBlockBody(hash common.Hash) (*types.Body, error)
 	HandleRuntimeChanges(newState *rtstorage.TrieState, in runtime.Instance, bHash common.Hash) error
-	GetBestBlockRuntime() runtime.Instance
+	//GetBestBlockRuntime() runtime.Instance
 	GetRuntime(hash *common.Hash) (instance runtime.Instance, err error)
 	StoreRuntime(common.Hash, runtime.Instance)
 }
@@ -84,7 +84,7 @@ type StorageState interface {
 	GetStateRootFromBlock(bhash *common.Hash) (*common.Hash, error)
 	GetStorage(root *common.Hash, key []byte) ([]byte, error)
 	GenerateTrieProof(stateRoot common.Hash, keys [][]byte) ([][]byte, error)
-	GetRuntime(blockHash common.Hash) (instance runtime.Instance, err error)
+	//GetRuntime(blockHash common.Hash) (instance runtime.Instance, err error)
 	sync.Locker
 }
 
--- a/dot/core/interface.go
+++ b/dot/core/interface.go
@@ -84,7 +84,7 @@ type StorageState interface {
 	GetStateRootFromBlock(bhash *common.Hash) (*common.Hash, error)
 	GetStorage(root *common.Hash, key []byte) ([]byte, error)
 	GenerateTrieProof(stateRoot common.Hash, keys [][]byte) ([][]byte, error)
-	GetRuntimeFromDB(blockHash common.Hash) (instance runtime.Instance, err error)
+	GetRuntime(blockHash common.Hash) (instance runtime.Instance, err error)
 	sync.Locker
 }
 
--- a/dot/core/interface.go
+++ b/dot/core/interface.go
@@ -70,7 +70,7 @@ type BlockState interface {
 	SubChain(start, end common.Hash) ([]common.Hash, error)
 	GetBlockBody(hash common.Hash) (*types.Body, error)
 	HandleRuntimeChanges(newState *rtstorage.TrieState, in runtime.Instance, bHash common.Hash) error
-	GetRuntime(*common.Hash) (runtime.Instance, error)
+	GetBestBlockRuntime() runtime.Instance
 	StoreRuntime(common.Hash, runtime.Instance)
 }
 
--- a/dot/core/interface.go
+++ b/dot/core/interface.go
@@ -71,7 +71,6 @@ type BlockState interface {
 	GetBlockBody(hash common.Hash) (*types.Body, error)
 	HandleRuntimeChanges(newState *rtstorage.TrieState, in runtime.Instance, bHash common.Hash) error
 	GetRuntime(*common.Hash) (runtime.Instance, error)
-	GetRuntimeFromDB(blockHash *common.Hash) (instance runtime.Instance, err error)
 	StoreRuntime(common.Hash, runtime.Instance)
 }
 
--- a/dot/core/interface.go
+++ b/dot/core/interface.go
@@ -71,6 +71,7 @@ type BlockState interface {
 	GetBlockBody(hash common.Hash) (*types.Body, error)
 	HandleRuntimeChanges(newState *rtstorage.TrieState, in runtime.Instance, bHash common.Hash) error
 	GetBestBlockRuntime() runtime.Instance
+	GetRuntime(hash *common.Hash) (instance runtime.Instance, err error)
 	StoreRuntime(common.Hash, runtime.Instance)
 }
 
--- a/dot/core/messages_test.go
+++ b/dot/core/messages_test.go
@@ -339,11 +339,12 @@ func TestServiceHandleTransactionMessage(t *testing.T) {
 					tt.mockBlockState.bestHeader.header,
 					tt.mockBlockState.bestHeader.err)
 
-				if tt.mockBlockState.getRuntime != nil {
-					blockState.EXPECT().GetRuntime(gomock.Any()).Return(
-						tt.mockBlockState.getRuntime.runtime,
-						tt.mockBlockState.getRuntime.err)
-				}
+				//if tt.mockBlockState.getRuntime != nil {
+				// todo(ed) re-implement
+				//blockState.EXPECT().GetRuntime(gomock.Any()).Return(
+				//	tt.mockBlockState.getRuntime.runtime,
+				//	tt.mockBlockState.getRuntime.err)
+				//}
 				s.blockState = blockState
 			}
 			if tt.mockStorageState != nil {
--- a/dot/core/messages_test.go
+++ b/dot/core/messages_test.go
@@ -162,26 +162,6 @@ func TestServiceHandleTransactionMessage(t *testing.T) {
 			expErr:    errDummyErr,
 			expErrMsg: errDummyErr.Error(),
 		},
-		//{
-		// todo(ed) confirm this is no longer necessary since GetBestBlockRuntime doesn't return error
-		//	name: "get runtime error",
-		//	mockNetwork: &mockNetwork{
-		//		IsSynced: true,
-		//	},
-		//	mockBlockState: &mockBlockState{
-		//		bestHeader: &mockBestHeader{
-		//			header: testEmptyHeader,
-		//		},
-		//		getRuntime: &mockGetRuntime{
-		//			err: errDummyErr,
-		//		},
-		//	},
-		//	args: args{
-		//		msg: &network.TransactionMessage{Extrinsics: []types.Extrinsic{}},
-		//	},
-		//	expErr:    errDummyErr,
-		//	expErrMsg: errDummyErr.Error(),
-		//},
 		{
 			name: "happy path no loop",
 			mockNetwork: &mockNetwork{
--- a/dot/core/messages_test.go
+++ b/dot/core/messages_test.go
@@ -38,13 +38,9 @@ type mockBestHeader struct {
 	err    error
 }
 
-type mockGetRuntime struct {
-	runtime RuntimeInstance
-}
-
 type mockBlockState struct {
 	bestHeader *mockBestHeader
-	getRuntime *mockGetRuntime
+	runtime    RuntimeInstance
 }
 
 type mockStorageState struct {
@@ -177,9 +173,7 @@ func TestServiceHandleTransactionMessage(t *testing.T) {
 				bestHeader: &mockBestHeader{
 					header: testEmptyHeader,
 				},
-				getRuntime: &mockGetRuntime{
-					runtime: runtimeMock,
-				},
+				runtime: runtimeMock,
 			},
 			args: args{
 				peerID: peer.ID("jimbo"),
@@ -195,9 +189,7 @@ func TestServiceHandleTransactionMessage(t *testing.T) {
 				bestHeader: &mockBestHeader{
 					header: testEmptyHeader,
 				},
-				getRuntime: &mockGetRuntime{
-					runtime: runtimeMock,
-				},
+				runtime: runtimeMock,
 			},
 			mockStorageState: &mockStorageState{
 				input: &common.Hash{},
@@ -229,9 +221,7 @@ func TestServiceHandleTransactionMessage(t *testing.T) {
 				bestHeader: &mockBestHeader{
 					header: testEmptyHeader,
 				},
-				getRuntime: &mockGetRuntime{
-					runtime: runtimeMock2,
-				},
+				runtime: runtimeMock2,
 			},
 			mockStorageState: &mockStorageState{
 				input:     &common.Hash{},
@@ -268,9 +258,7 @@ func TestServiceHandleTransactionMessage(t *testing.T) {
 				bestHeader: &mockBestHeader{
 					header: testEmptyHeader,
 				},
-				getRuntime: &mockGetRuntime{
-					runtime: runtimeMock3,
-				},
+				runtime: runtimeMock3,
 			},
 			mockStorageState: &mockStorageState{
 				input:     &common.Hash{},
@@ -315,13 +303,12 @@ func TestServiceHandleTransactionMessage(t *testing.T) {
 			}
 			if tt.mockBlockState != nil {
 				blockState := NewMockBlockState(ctrl)
-				blockState.EXPECT().BestBlockHeader().Return(
-					tt.mockBlockState.bestHeader.header,
-					tt.mockBlockState.bestHeader.err)
+				blockState.EXPECT().BestBlockHeader().
+					Return(tt.mockBlockState.bestHeader.header, tt.mockBlockState.bestHeader.err)
 
-				if tt.mockBlockState.getRuntime != nil {
-					blockState.EXPECT().GetBestBlockRuntime().Return(
-						tt.mockBlockState.getRuntime.runtime)
+				if tt.mockBlockState.runtime != nil {
+					blockState.EXPECT().GetBestBlockRuntime().
+						Return(tt.mockBlockState.runtime)
 				}
 				s.blockState = blockState
 			}
@@ -329,9 +316,8 @@ func TestServiceHandleTransactionMessage(t *testing.T) {
 				storageState := NewMockStorageState(ctrl)
 				storageState.EXPECT().Lock()
 				storageState.EXPECT().Unlock()
-				storageState.EXPECT().TrieState(tt.mockStorageState.input).Return(
-					tt.mockStorageState.trieState,
-					tt.mockStorageState.err)
+				storageState.EXPECT().TrieState(tt.mockStorageState.input).
+					Return(tt.mockStorageState.trieState, tt.mockStorageState.err)
 				s.storageState = storageState
 			}
 			if tt.mockTxnState != nil {
--- a/dot/core/messages_test.go
+++ b/dot/core/messages_test.go
@@ -40,7 +40,6 @@ type mockBestHeader struct {
 
 type mockGetRuntime struct {
 	runtime RuntimeInstance
-	err     error
 }
 
 type mockBlockState struct {
@@ -163,25 +162,26 @@ func TestServiceHandleTransactionMessage(t *testing.T) {
 			expErr:    errDummyErr,
 			expErrMsg: errDummyErr.Error(),
 		},
-		{
-			name: "get runtime error",
-			mockNetwork: &mockNetwork{
-				IsSynced: true,
-			},
-			mockBlockState: &mockBlockState{
-				bestHeader: &mockBestHeader{
-					header: testEmptyHeader,
-				},
-				getRuntime: &mockGetRuntime{
-					err: errDummyErr,
-				},
-			},
-			args: args{
-				msg: &network.TransactionMessage{Extrinsics: []types.Extrinsic{}},
-			},
-			expErr:    errDummyErr,
-			expErrMsg: errDummyErr.Error(),
-		},
+		//{
+		// todo(ed) confirm this is no longer necessary since GetBestBlockRuntime doesn't return error
+		//	name: "get runtime error",
+		//	mockNetwork: &mockNetwork{
+		//		IsSynced: true,
+		//	},
+		//	mockBlockState: &mockBlockState{
+		//		bestHeader: &mockBestHeader{
+		//			header: testEmptyHeader,
+		//		},
+		//		getRuntime: &mockGetRuntime{
+		//			err: errDummyErr,
+		//		},
+		//	},
+		//	args: args{
+		//		msg: &network.TransactionMessage{Extrinsics: []types.Extrinsic{}},
+		//	},
+		//	expErr:    errDummyErr,
+		//	expErrMsg: errDummyErr.Error(),
+		//},
 		{
 			name: "happy path no loop",
 			mockNetwork: &mockNetwork{
@@ -339,12 +339,10 @@ func TestServiceHandleTransactionMessage(t *testing.T) {
 					tt.mockBlockState.bestHeader.header,
 					tt.mockBlockState.bestHeader.err)
 
-				//if tt.mockBlockState.getRuntime != nil {
-				// todo(ed) re-implement
-				//blockState.EXPECT().GetRuntime(gomock.Any()).Return(
-				//	tt.mockBlockState.getRuntime.runtime,
-				//	tt.mockBlockState.getRuntime.err)
-				//}
+				if tt.mockBlockState.getRuntime != nil {
+					blockState.EXPECT().GetBestBlockRuntime().Return(
+						tt.mockBlockState.getRuntime.runtime)
+				}
 				s.blockState = blockState
 			}
 			if tt.mockStorageState != nil {
--- a/dot/core/service.go
+++ b/dot/core/service.go
@@ -453,10 +453,10 @@ func (s *Service) GetRuntimeVersion(bhash *common.Hash) (
 	}
 
 	rt, err := s.blockState.GetRuntime(bhash)
-	if err != nil && err == blocktree.ErrFailedToGetRuntime {
+	if errors.Is(err, blocktree.ErrFailedToGetRuntime) {
 		rt, err = s.blockState.GetRuntimeFromDB(bhash)
 		if err != nil {
-			return version, err
+			return version, fmt.Errorf("getting runtime from database: %w", err)
 		}
 	} else if err != nil {
 		return version, err
@@ -532,10 +532,10 @@ func (s *Service) GetMetadata(bhash *common.Hash) ([]byte, error) {
 	}
 
 	rt, err := s.blockState.GetRuntime(bhash)
-	if err != nil && err == blocktree.ErrFailedToGetRuntime {
+	if errors.Is(err, blocktree.ErrFailedToGetRuntime) {
 		rt, err = s.blockState.GetRuntimeFromDB(bhash)
 		if err != nil {
-			return nil, err
+			return nil, fmt.Errorf("getting runtime from database: %w", err)
 		}
 	} else if err != nil {
 		return nil, err
--- a/dot/core/service.go
+++ b/dot/core/service.go
@@ -453,7 +453,12 @@ func (s *Service) GetRuntimeVersion(bhash *common.Hash) (
 	}
 
 	rt, err := s.blockState.GetRuntime(bhash)
-	if err != nil {
+	if err != nil && err == blocktree.ErrFailedToGetRuntime {
+		rt, err = s.blockState.GetRuntimeFromDB(bhash)
+		if err != nil {
+			return version, err
+		}
+	} else if err != nil {
 		return version, err
 	}
 
@@ -527,7 +532,12 @@ func (s *Service) GetMetadata(bhash *common.Hash) ([]byte, error) {
 	}
 
 	rt, err := s.blockState.GetRuntime(bhash)
-	if err != nil {
+	if err != nil && err == blocktree.ErrFailedToGetRuntime {
+		rt, err = s.blockState.GetRuntimeFromDB(bhash)
+		if err != nil {
+			return nil, err
+		}
+	} else if err != nil {
 		return nil, err
 	}
 
--- a/dot/core/service.go
+++ b/dot/core/service.go
@@ -19,7 +19,6 @@ import (
 	"github.com/ChainSafe/gossamer/lib/keystore"
 	"github.com/ChainSafe/gossamer/lib/runtime"
 	rtstorage "github.com/ChainSafe/gossamer/lib/runtime/storage"
-	"github.com/ChainSafe/gossamer/lib/runtime/wasmer"
 	"github.com/ChainSafe/gossamer/lib/services"
 	"github.com/ChainSafe/gossamer/lib/transaction"
 	cscale "github.com/centrifuge/go-substrate-rpc-client/v4/scale"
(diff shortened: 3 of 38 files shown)


lib/blocktree/blocktree.go
```diff
@@ -418,3 +418,10 @@ func (bt *BlockTree) GetBlockRuntime(hash common.Hash) (runtime.Instance, error)
 	}
 	return ins, nil
 }
+
+// ClearRuntimes deletes all runtime instances from BlockTree runtime map
+func (bt *BlockTree) ClearRuntimes() {
+	for key := range bt.runtimes.mapping {
+		delete(bt.runtimes.mapping, key)
+	}
```
qdm12, line 426: This is not thread safe, you should instead add a method to `bt.runtimes` locking a mutex and re-making the map (since maps don't shrink when we delete elements)

qdm12 (changes requested):
- You need to store older runtime wasm blob on disk
- BlockState's `GetRuntime` should, if the instance is not found in the memory map, try to load an instance from the disk wasm blob and return that instance. The caller should I guess know it was an instance created only for a temporary usage, and call `.close()` on the instance once done with it.

dot/core/interface.go
```diff
@@ -71,6 +71,7 @@ type BlockState interface {
 	GetBlockBody(hash common.Hash) (*types.Body, error)
 	HandleRuntimeChanges(newState *rtstorage.TrieState, in runtime.Instance, bHash common.Hash) error
 	GetRuntime(*common.Hash) (runtime.Instance, error)
+	GetRuntimeFromDB(hash *common.Hash) (runtime.Instance, error)
```
qdm12, line 74: Didn't we want to not have this method, and instead call a method on storagestate in callers?

lib/blocktree/blocktree.go
```diff
@@ -418,3 +418,10 @@ func (bt *BlockTree) GetBlockRuntime(hash common.Hash) (runtime.Instance, error)
 	}
 	return ins, nil
 }
+
+// ClearRuntimes deletes all runtime instances from BlockTree runtime map
+func (bt *BlockTree) ClearRuntimes() {
+	bt.runtimes.mutex.Lock()
+	bt.runtimes.mapping = make(map[Hash]runtime.Instance)
+	bt.runtimes.mutex.Unlock()
```
qdm12, line 426: This should be a method on `bt.runtimes` ideally since there are already a few methods on it, so that it would be easier to move it to a subpackage later.

dot/core/interface.go
edwardmack: I'm not sure I understand what you mean here?  Should GetRuntimeFromDB be called from GetRuntime if that returns nil?

qdm12: No we should get the runtime from `StorageState`'s `LoadCode` instead of from `blockState` I think.

dot/rpc/modules/payment.go
```diff
@@ -49,7 +49,6 @@ func (p *PaymentModule) QueryInfo(_ *http.Request, req *PaymentQueryInfoRequest,
 	if err != nil {
 		return err
 	}
```
qdm12, line 51: nit out of scope / unneeded delta

```suggestion
	}

```

lib/blocktree/blocktree.go
```diff
@@ -418,3 +418,8 @@ func (bt *BlockTree) GetBlockRuntime(hash common.Hash) (runtime.Instance, error)
 	}
 	return ins, nil
 }
+
+// ClearRuntimes deletes all runtime instances from BlockTree runtime map
```
qdm12, line 422: nit

```suggestion
// ClearRuntimes deletes all runtime instances from the BlockTree runtime map
```

dot/state/block.go
```diff
@@ -626,6 +629,7 @@ func (bs *BlockState) HandleRuntimeChanges(newState *rtstorage.TrieState,
 		return err
 	}
 
+	bs.ClearRuntimes()
```
qdm12, line 632: this can be unexported I think?

qdm12 (changes requested):
I would suggest the following if it's possible:

1. Create a method on BlockState `GetBestBlockRuntime() (instance RuntimeInstance)` which always finds the runtime in memory. And then replace (many) uses of `GetRuntime` to use this method if `nil` or `&bestBlockHash` is passed to them
2. Change method on BlockState `GetRuntime(blockHash common.Hash) (instance RuntimeInstance, err error)` to return a sentinel error `ErrRuntimeNotFound`, but not check in storage state (and not have it in BlockState).
3. The caller of `GetRuntime` (only 2 occurrences I think) should check the error returned from `GetRuntime`, and, if it's not found, get the runtime from `StorageState`, which they probably already have access to (and if not they should have access, not BlockState).

dot/core/messages_test.go
```diff
@@ -40,7 +40,6 @@ type mockBestHeader struct {
 
 type mockGetRuntime struct {
 	runtime RuntimeInstance
-	err     error
 }
```
qdm12, line 43: nit should we remove this struct since it's just one field now?

dot/core/messages_test.go
```diff
@@ -335,9 +315,8 @@ func TestServiceHandleTransactionMessage(t *testing.T) {
 					tt.mockBlockState.bestHeader.err)
 
 				if tt.mockBlockState.getRuntime != nil {
-					blockState.EXPECT().GetRuntime(gomock.Any()).Return(
-						tt.mockBlockState.getRuntime.runtime,
-						tt.mockBlockState.getRuntime.err)
+					blockState.EXPECT().GetBestBlockRuntime().Return(
+						tt.mockBlockState.getRuntime.runtime)
```
qdm12, line 319: supernit better cut it after the `.`

```suggestion
					blockState.EXPECT().GetBestBlockRuntime().
						Return(tt.mockBlockState.getRuntime.runtime)
```

dot/core/service.go
```diff
@@ -187,7 +187,14 @@ func (s *Service) handleBlock(block *types.Block, state *rtstorage.TrieState) er
...
 	rt, err := s.blockState.GetRuntime(&block.Header.ParentHash)
-	if err != nil {
+	if errors.Is(err, blocktree.ErrFailedToGetRuntime) {
+		rt, err = s.getRuntimeFromDB(&block.Header.ParentHash)
+		if err != nil {
+			return err
+		}
+		defer rt.Stop()
+	}
+	if err != nil && !errors.Is(err, blocktree.ErrFailedToGetRuntime) {
```
qdm12, line 197:
```suggestion
	} else if err != nil {
```

dot/core/service.go
```diff
@@ -187,7 +187,14 @@ func (s *Service) handleBlock(block *types.Block, state *rtstorage.TrieState) er
 		block.Header.Hash(), state.MustRoot())
 
 	rt, err := s.blockState.GetRuntime(&block.Header.ParentHash)
-	if err != nil {
+	if errors.Is(err, blocktree.ErrFailedToGetRuntime) {
+		rt, err = s.getRuntimeFromDB(&block.Header.ParentHash)
+		if err != nil {
+			return err
+		}
+		defer rt.Stop()
```
qdm12, line 195: nit maybe let's add a comment why this is needed

```suggestion
		// ensure the runtime stops and releases resources since it was 
		// instantiated from disk just for this function call.
		defer rt.Stop()
```

dot/core/service.go
```diff
@@ -187,7 +187,14 @@ func (s *Service) handleBlock(block *types.Block, state *rtstorage.TrieState) er
 		block.Header.Hash(), state.MustRoot())
 
 	rt, err := s.blockState.GetRuntime(&block.Header.ParentHash)
-	if err != nil {
+	if errors.Is(err, blocktree.ErrFailedToGetRuntime) {
+		rt, err = s.getRuntimeFromDB(&block.Header.ParentHash)
+		if err != nil {
+			return err
```
qdm12, line 193:
```suggestion
			return fmt.Errorf("getting runtime from database: %w", err)
```

dot/core/service.go
```diff
@@ -449,7 +446,17 @@ func (s *Service) GetRuntimeVersion(bhash *common.Hash) (
 	}
 
 	rt, err := s.blockState.GetRuntime(bhash)
-	if err != nil {
+	if errors.Is(err, blocktree.ErrFailedToGetRuntime) {
+		rt, err = s.getRuntimeFromDB(bhash)
+		if err != nil {
+			return version, err
```
qdm12, line 452:
```suggestion
			return version, fmt.Errorf("getting runtime from database: %w", err)
```

dot/core/service.go
```diff
@@ -449,7 +446,17 @@ func (s *Service) GetRuntimeVersion(bhash *common.Hash) (
...
+		}
+		version = rt.Version()
+
+		rt.Stop()
+		return version, nil
+	}
+	if err != nil && !errors.Is(err, blocktree.ErrFailedToGetRuntime) {
```
qdm12, line 459:
```suggestion
	} else if err != nil {
```

dot/core/service.go
```diff
@@ -547,7 +552,12 @@ func (s *Service) GetMetadata(bhash *common.Hash) ([]byte, error) {
 	}
 
 	rt, err := s.blockState.GetRuntime(bhash)
-	if err != nil {
+	if err != nil && err == blocktree.ErrFailedToGetRuntime {
+		rt, err = s.blockState.GetRuntimeFromDB(bhash)
+		if err != nil {
+			return nil, err
```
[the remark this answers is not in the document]
qdm12: I think this is still not done

dot/core/service.go
```diff
@@ -523,7 +526,19 @@ func (s *Service) GetMetadata(bhash *common.Hash) ([]byte, error) {
...
 	rt, err := s.blockState.GetRuntime(bhash)
-	if err != nil {
+	if errors.Is(err, blocktree.ErrFailedToGetRuntime) {
+		rt, err = s.getRuntimeFromDB(bhash)
+		if err != nil {
+			return nil, err
+		}
+		metadata, err := rt.Metadata()
+		if err != nil {
+			return nil, err
```
qdm12, line 536:
```suggestion
			return nil, fmt.Errorf("getting runtime metadata: %w", err)
```

dot/core/service.go
```diff
@@ -523,7 +526,19 @@ func (s *Service) GetMetadata(bhash *common.Hash) ([]byte, error) {
...
+		}
+		rt.Stop()
```
qdm12, line 538: move this one above in a defer statement. If `rt.Metadata()` fails, we still want to stop the runtime

dot/core/service.go
```diff
@@ -523,7 +526,19 @@ func (s *Service) GetMetadata(bhash *common.Hash) ([]byte, error) {
...
+		return metadata, nil
+	}
+	if err != nil && !errors.Is(err, blocktree.ErrFailedToGetRuntime) {
```
qdm12, line 541:
```suggestion
	} else if err != nil {
```

dot/core/service.go
```diff
@@ -551,3 +566,30 @@ func (s *Service) GetReadProofAt(block common.Hash, keys [][]byte) (
...
+
+// getRuntimeFromDB gets the runtime for the corresponding block hash from storageState
+func (s *Service) getRuntimeFromDB(blockHash *common.Hash) (instance runtime.Instance, err error) {
+	var stateRootHash *common.Hash
+	if blockHash != nil {
+		stateRootHash, err = s.storageState.GetStateRootFromBlock(blockHash)
+		if err != nil {
+			return nil, fmt.Errorf("getting state root from block hash: %w", err)
+		}
+	}
```
qdm12, line 578: I would like to suggest we accept a `blockHash common.Hash` argument and not a `nil` hash.
We do this all over the codebase and it's really dark magic at the end of the day, which we should avoid in new code (until we refactor older code one day). Anyway I don't think we pass in `nil` anywhere so it shouldn't be too troublesome to change.

dot/core/service.go
```diff
@@ -551,3 +566,30 @@ func (s *Service) GetReadProofAt(block common.Hash, keys [][]byte) (
...
+
+	trieState, err := s.storageState.TrieState(stateRootHash)
+	if err != nil {
+		return nil, fmt.Errorf("getting trie state: %w", err)
+	}
+
+	code := trieState.LoadCode()
+	config := wasmer.Config{
+		LogLvl: log.DoNotChange,
```
qdm12, line 587: 👍 Nice attention to detail!

dot/core/service.go
```diff
@@ -551,3 +566,30 @@ func (s *Service) GetReadProofAt(block common.Hash, keys [][]byte) (
 
 	return block, proofForKeys, nil
 }
+
+// getRuntimeFromDB gets the runtime for the corresponding block hash from storageState
+func (s *Service) getRuntimeFromDB(blockHash *common.Hash) (instance runtime.Instance, err error) {
```
qdm12, line 571: Since this only uses `s.storageState`, should we define it as a method on `StorageState` instead?

Maybe it's not possible due to cycle dependency that I don't know yet.

dot/core/service_test.go
```diff
@@ -1126,36 +1100,13 @@ func TestServiceHandleSubmittedExtrinsic(t *testing.T) {
...
-	})
-
 	t.Run("validate txn err", func(t *testing.T) {
 		t.Parallel()
 		ctrl := gomock.NewController(t)
 		mockBlockState := NewMockBlockState(ctrl)
 		mockBlockState.EXPECT().BestBlockHash().Return(common.Hash{})
-		runtimeMockErr := mocksruntime.NewInstance(t)
-		mockBlockState.EXPECT().GetRuntime(&common.Hash{}).Return(runtimeMockErr, nil).MaxTimes(2)
+		runtimeMockErr := new(mocksruntime.Instance)
```
qdm12, line 1108:
```suggestion
		runtimeMockErr := mocksruntime.NewInstance(t)
```

dot/rpc/modules/api.go
```diff
@@ -12,6 +12,7 @@ import (
 	"github.com/ChainSafe/gossamer/lib/genesis"
 	"github.com/ChainSafe/gossamer/lib/grandpa"
 	"github.com/ChainSafe/gossamer/lib/runtime"
+	rtstorage "github.com/ChainSafe/gossamer/lib/runtime/storage"
```
qdm12, line 15: nit Why not without import alias `rtstorage`? I don't see any conflicting `storage` import 🤔

dot/rpc/modules/payment.go
```diff
@@ -46,10 +54,16 @@ func (p *PaymentModule) QueryInfo(_ *http.Request, req *PaymentQueryInfoRequest,
...
 	r, err := p.blockAPI.GetRuntime(&hash)
-	if err != nil {
+	if errors.Is(err, blocktree.ErrFailedToGetRuntime) {
+		r, err = p.getRuntimeFromDB(&hash)
+		if err != nil {
+			return err
+		}
+		defer r.Stop()
+	}
+	if err != nil && !errors.Is(err, blocktree.ErrFailedToGetRuntime) {
```
qdm12, line 64:
```suggestion
	} else if err != nil {
```

dot/rpc/modules/payment.go
```diff
@@ -70,3 +84,30 @@ func (p *PaymentModule) QueryInfo(_ *http.Request, req *PaymentQueryInfoRequest,
 
 	return nil
 }
+
+// getRuntimeFromDB gets the runtime for the corresponding block hash from storageState
+func (p *PaymentModule) getRuntimeFromDB(blockHash *common.Hash) (instance runtime.Instance, err error) {
+	var stateRootHash *common.Hash
+	if blockHash != nil {
+		stateRootHash, err = p.storageAPI.GetStateRootFromBlock(blockHash)
```
qdm12, line 92: It would be nice to use a  method on storage state if that's possible, so we can dedup this code which is here twice.

dot/state/offline_pruner.go
```diff
@@ -46,7 +46,8 @@ func NewOfflinePruner(inputDBPath, prunedDBPath string, bloomSize uint64,
 
 	// create blockState state
 	// NewBlockState on pruner execution does not use telemetry
-	blockState, err := NewBlockState(db, tries, nil)
+	// todo(ed) determine if setting storage to nil is correct
+	blockState, err := NewBlockState(db, tries, nil, nil)
```
qdm12, line 50: Remind me next week, I'm gonna have a look at the offline pruner.

dot/sync/chain_processor.go
```diff
@@ -218,7 +222,14 @@ func (s *chainProcessor) handleBlock(block *types.Block) error {
 
 	hash := parent.Hash()
 	rt, err := s.blockState.GetRuntime(&hash)
-	if err != nil {
+	if errors.Is(err, blocktree.ErrFailedToGetRuntime) {
+		rt, err = s.getRuntimeFromDB(&hash)
+		if err != nil {
+			return err
```
qdm12, line 228:
```suggestion
			return fmt.Errorf("getting runtime from database: %w", err)
```

dot/sync/chain_processor.go
```diff
@@ -218,7 +222,14 @@ func (s *chainProcessor) handleBlock(block *types.Block) error {
...
+		}
+		defer rt.Stop()
+	}
+	if err != nil && !errors.Is(err, blocktree.ErrFailedToGetRuntime) {
```
qdm12, line 232:
```suggestion
	} else if err != nil {
```

dot/sync/chain_processor.go
```diff
@@ -263,3 +274,30 @@ func (s *chainProcessor) handleJustification(header *types.Header, justification
 
 	logger.Infof("🔨 finalised block number %d with hash %s", header.Number, header.Hash())
 }
+
+// getRuntimeFromDB gets the runtime for the corresponding block hash from storageState
+func (s *chainProcessor) getRuntimeFromDB(blockHash *common.Hash) (instance runtime.Instance, err error) {
+	var stateRootHash *common.Hash
+	if blockHash != nil {
```
qdm12, line 281: Oh boy we really need a method on storage state, i hope it's possible 😆

dot/state/block.go
```diff
@@ -651,11 +652,25 @@ func (bs *BlockState) GetRuntime(hash *common.Hash) (runtime.Instance, error) {
 	return bs.bt.GetBlockRuntime(*hash)
 }
 
+//  GetBestBlockRuntime returns the current runtime instance for the best block
+func (bs *BlockState) GetBestBlockRuntime() (instance runtime.Instance) {
+	instance, err := bs.bt.GetBlockRuntime(bs.BestBlockHash())
+	if err != nil {
+		logger.Errorf("error retrieving block runtime: %w", err)
+	}
+	return
```
qdm12, line 661: So.... on that one....

I guess this is taking it from memory and it should ALWAYS be there right? If so replace the error log with a panic.
We definitely don't want this to go unnoticed in an error log.

```suggestion
	instance, err := bs.bt.GetBlockRuntime(bs.BestBlockHash())
	if err != nil {
		panic(fmt.Sprintf("we should always succeed getting the best block runtime but an error occurred: %s", err))
	}
	return instance
```

lib/babe/babe_integration_test.go
```diff
@@ -216,10 +216,6 @@ func TestService_ProducesBlocks(t *testing.T) {
 	}()
 
 	time.Sleep(babeService.constants.slotDuration * 2)
-	babeService.blockImportHandler.(*mocks.BlockImportHandler).
-		AssertCalled(t, "HandleBlockProduced",
-			mock.AnythingOfType("*types.Block"),
-			mock.AnythingOfType("*storage.TrieState"))
```
qdm12, line 222: Thank you 🙏

qdm12 (changes requested):
Nice work! 👍

I just hope (or wish if not possible) we can have that `getRuntimeFromDB` as a method on the storage state, so can deduplicate it at 3 places without affecting package dependency 🙏

dot/core/service.go
```diff
@@ -523,7 +526,20 @@ func (s *Service) GetMetadata(bhash *common.Hash) ([]byte, error) {
...
+			return nil, fmt.Errorf("getting runtime from database: %w", err)
+		}
+		metadata, err := rt.Metadata()
+		if err != nil {
+			return nil, fmt.Errorf("getting runtime metadata: %w", err)
+		}
+		// ensure the runtime stops and releases resources since it was
+		// instantiated from disk just for this function call.
+		defer rt.Stop()
+		return metadata, nil
```
EclesioMeloJunior, line 541: should the `defer rt.Stop()` statement be placed before `metadata, err := rt.Metadata()` to be called in a case where it is not possible to retrieve the `rt.Metadata`

dot/core/service.go
```diff
@@ -523,7 +526,20 @@ func (s *Service) GetMetadata(bhash *common.Hash) ([]byte, error) {
 	}
 
 	rt, err := s.blockState.GetRuntime(bhash)
-	if err != nil {
+	if errors.Is(err, blocktree.ErrFailedToGetRuntime) {
+		rt, err = s.getRuntimeFromDB(*bhash)
+		if err != nil {
+			return nil, fmt.Errorf("getting runtime from database: %w", err)
+		}
+		metadata, err := rt.Metadata()
```
EclesioMeloJunior, line 534: in the outer scope, after we `GetRuntime` successfully we call `rt.SetContextStorage(ts)`, and here, in the inner scope we just call `rt.Metadata` directly, is there any reason?

EclesioMeloJunior (changes requested):
Agreed with Quentin, we need a method on storage state to retrieve the runtime from the database when needed

dot/core/service.go
edwardmack: We don't set runtime context storage here because the runtime used from database is temporary and is closed as soon as metadata is found.

dot/core/interface.go
```diff
@@ -83,6 +84,7 @@ type StorageState interface {
 	GetStateRootFromBlock(bhash *common.Hash) (*common.Hash, error)
 	GetStorage(root *common.Hash, key []byte) ([]byte, error)
 	GenerateTrieProof(stateRoot common.Hash, keys [][]byte) ([][]byte, error)
+	GetRuntimeFromDB(blockHash common.Hash) (instance runtime.Instance, err error)
```
qdm12, line 87: nit I don't think the `FromDB` is necessary since this is on the storageState interface? 🤔

dot/state/block.go
```diff
@@ -651,11 +652,25 @@ func (bs *BlockState) GetRuntime(hash *common.Hash) (runtime.Instance, error) {
 	return bs.bt.GetBlockRuntime(*hash)
 }
 
+//  GetBestBlockRuntime returns the current runtime instance for the best block
+func (bs *BlockState) GetBestBlockRuntime() (instance runtime.Instance) {
+	instance, err := bs.bt.GetBlockRuntime(bs.BestBlockHash())
+	if err != nil {
+		panic(fmt.Sprintf("we should always succeed getting the best block runtime but an error orrurred: %s", err))
+	}
+	return
```
qdm12, line 661: nit

```suggestion
	return instance
```

dot/state/storage.go
```diff
@@ -305,3 +308,28 @@ func (s *StorageState) GenerateTrieProof(stateRoot common.Hash, keys [][]byte) (
...
 	return proof.Generate(stateRoot[:], keys, s.db)
 }
+
+// GetRuntimeFromDB gets the runtime for the corresponding block hash from storageState
+func (s *StorageState) GetRuntimeFromDB(blockHash common.Hash) (instance runtime.Instance, err error) {
+	var stateRootHash *common.Hash
+	stateRootHash, err = s.GetStateRootFromBlock(&blockHash)
+	if err != nil {
+		return nil, fmt.Errorf("getting state root from block hash: %w", err)
+	}
```
qdm12, line 318: This should be deduped as a method on StorageState I think? 🤔  We have similar bodies in multiple places in this PR

dot/sync/chain_processor.go
```diff
@@ -263,3 +271,30 @@ func (s *chainProcessor) handleJustification(header *types.Header, justification
 
 	logger.Infof("🔨 finalised block number %d with hash %s", header.Number, header.Hash())
 }
+
+//// getRuntimeFromDB gets the runtime for the corresponding block hash from storageState
+//func (s *chainProcessor) getRuntimeFromDB(blockHash *common.Hash) (instance runtime.Instance, err error) {
+//	var stateRootHash *common.Hash
+//	if blockHash != nil {
+//		stateRootHash, err = s.storageState.GetStateRootFromBlock(blockHash)
```
qdm12, line 279: remove

qdm12 (commented):
Almost done 💯 
I'm still wondering, it looks like we have multiple occurrences of the same function GetRuntimeFromDB, maybe we can move it as a method on storage state to deduplicate?

dot/state/block.go
```diff
@@ -72,7 +71,8 @@ type BlockState struct {
 }
 
 // NewBlockState will create a new BlockState backed by the database located at basePath
-func NewBlockState(db chaindb.Database, trs *Tries, telemetry telemetry.Client) (*BlockState, error) {
+func NewBlockState(db chaindb.Database, trs *Tries, telemetry telemetry.Client,
+	storageState *StorageState) (*BlockState, error) {
```
qdm12, line 75: Revert (other code changes needed)

```suggestion
func NewBlockState(db chaindb.Database, trs *Tries, telemetry telemetry.Client) (*BlockState, error) {
```

qdm12 (changes requested):

dot/core/service_test.go
```diff
@@ -593,16 +593,16 @@ func Test_Service_maintainTransactionPool(t *testing.T) {
 		tx := transaction.NewValidTransaction(types.Extrinsic{21}, &transaction.Validity{Propagate: true})
 
 		ctrl := gomock.NewController(t)
-		runtimeMock := NewMockRuntimeInstance(ctrl)
-		runtimeMock.EXPECT().ValidateTransaction(types.Extrinsic{21}).
+		runtimeMock := mocksruntime.NewInstance(t)
+		runtimeMock.On("ValidateTransaction", types.Extrinsic{21}).
```
qdm12, line 597: Let's keep the gomock mock instead of mockery (maybe a bad git operation?)

dot/core/service_test.go
```diff
@@ -558,13 +558,13 @@ func Test_Service_maintainTransactionPool(t *testing.T) {
 		vt := transaction.NewValidTransaction(extrinsic, validity)
 
 		ctrl := gomock.NewController(t)
-		runtimeMock := NewMockRuntimeInstance(ctrl)
-		runtimeMock.EXPECT().ValidateTransaction(types.Extrinsic{21}).Return(nil, errTestDummyError)
+		runtimeMock := mocksruntime.NewInstance(t)
+		runtimeMock.On("ValidateTransaction", types.Extrinsic{21}).Return(nil, errTestDummyError)
```
qdm12, line 562: Let's keep the gomock mock instead of mockery (maybe a bad git operation?)

dot/core/service_test.go
```diff
@@ -682,13 +682,13 @@ func Test_Service_handleBlocksAsync(t *testing.T) {
 		block.Header.Number = 21
 
 		ctrl := gomock.NewController(t)
-		runtimeMock := NewMockRuntimeInstance(ctrl)
-		runtimeMock.EXPECT().ValidateTransaction(types.Extrinsic{21}).Return(nil, errTestDummyError)
+		runtimeMock := mocksruntime.NewInstance(t)
+		runtimeMock.On("ValidateTransaction", types.Extrinsic{21}).Return(nil, errTestDummyError)
```
qdm12, line 686: Let's keep the gomock mock instead of mockery (maybe a bad git operation?)

dot/core/service_test.go
```diff
@@ -794,33 +794,18 @@ func TestService_handleChainReorg(t *testing.T) {
...
-			blockState: mockBlockState,
-		}
-		execTest(t, service, testPrevHash, testCurrentHash, errDummyErr)
-	})
-
 	t.Run("invalid transaction", func(t *testing.T) {
 		t.Parallel()
 		ctrl := gomock.NewController(t)
-		runtimeMockErr := NewMockRuntimeInstance(ctrl)
+		runtimeMockErr := mocksruntime.NewInstance(t)
```
qdm12, line 800: Let's keep the gomock mock instead of mockery (maybe a bad git operation?)

dot/core/service_test.go
```diff
@@ -980,26 +966,15 @@ func TestService_DecodeSessionKeys(t *testing.T) {
 	t.Run("ok case", func(t *testing.T) {
 		t.Parallel()
 		ctrl := gomock.NewController(t)
-		runtimeMock := NewMockRuntimeInstance(ctrl)
-		runtimeMock.EXPECT().DecodeSessionKeys(testEncKeys).Return(testEncKeys, nil)
+		runtimeMock := mocksruntime.NewInstance(t)
+		runtimeMock.On("DecodeSessionKeys", testEncKeys).Return(testEncKeys, nil)
```
qdm12, line 970: Let's keep the gomock mock instead of mockery (maybe a bad git operation?)

dot/core/service_test.go
```diff
@@ -1179,12 +1130,13 @@ func TestServiceHandleSubmittedExtrinsic(t *testing.T) {
 		t.Parallel()
 		ctrl := gomock.NewController(t)
 
-		runtimeMock := NewMockRuntimeInstance(ctrl)
+		runtimeMock := mocksruntime.NewInstance(t)
```
qdm12, line 1133: Let's keep the gomock mock instead of mockery (maybe a bad git operation?)

qdm12 (approved):
Nice, good job :100: !

Just a few minor/quick details to address :wink:

dot/state/block_test.go
```diff
@@ -543,3 +547,53 @@ func TestNumberIsFinalised(t *testing.T) {
...
 }
+
+func TestBlockState_GetBestBlockRuntime(t *testing.T) {
+	blockState := newTestBlockState(t, newTriesEmpty())
+	cfg := wasmer.Config{}
+	polkadotRuntimeFilepath, err := runtime.GetRuntime(
+		context.Background(), runtime.POLKADOT_RUNTIME)
+	require.NoError(t, err)
+	code, err := os.ReadFile(polkadotRuntimeFilepath)
+	require.NoError(t, err)
```
qdm12, line 558: nit can you use a mock instance to store in the runtimes instead of a full instance? To reduce the resource usage / time it takes to init.

dot/state/block_test.go
```diff
@@ -543,3 +547,53 @@ func TestNumberIsFinalised(t *testing.T) {
...
+		r := recover()
+		require.Equal(t, "we should always succeed getting the best block runtime but an error occurred: "+
+			"failed to get runtime instance", r)
+	}()
+
+	resultInstance := blockState.GetBestBlockRuntime()
+	require.Equal(t, nil, resultInstance)
+}
+
+func TestBlockState_clearRuntimes(t *testing.T) {
```
qdm12, line 580: nit you can run in parallel

```suggestion
func TestBlockState_clearRuntimes(t *testing.T) {
	t.Parallel()

```

dot/state/block_test.go
```diff
@@ -543,3 +547,53 @@ func TestNumberIsFinalised(t *testing.T) {
...
+	blockState := newTestBlockState(t, newTriesEmpty())
+	cfg := wasmer.Config{}
+	polkadotRuntimeFilepath, err := runtime.GetRuntime(
+		context.Background(), runtime.POLKADOT_RUNTIME)
+	require.NoError(t, err)
+	code, err := os.ReadFile(polkadotRuntimeFilepath)
+	require.NoError(t, err)
+	runtimeInstance, err := wasmer.NewInstance(code, cfg)
+	require.NoError(t, err)
```
qdm12, line 589: nit can you use a mock instance to store in the runtimes instead of a full instance? To reduce the resource usage / time it takes to init.

dot/core/interface.go
```diff
@@ -70,7 +70,8 @@ type BlockState interface {
 	SubChain(start, end common.Hash) ([]common.Hash, error)
 	GetBlockBody(hash common.Hash) (*types.Body, error)
 	HandleRuntimeChanges(newState *rtstorage.TrieState, in runtime.Instance, bHash common.Hash) error
-	GetRuntime(*common.Hash) (runtime.Instance, error)
+	GetBestBlockRuntime() runtime.Instance
+	GetRuntime(hash *common.Hash) (instance runtime.Instance, err error)
```
timwu20, line 74: Why doesn't `GetBlockBlockRuntime()` return an error?

dot/state/storage.go
```diff
@@ -305,3 +308,27 @@ func (s *StorageState) GenerateTrieProof(stateRoot common.Hash, keys [][]byte) (
 	encodedProofNodes [][]byte, err error) {
 	return proof.Generate(stateRoot[:], keys, s.db)
 }
+
+// GetRuntime gets the runtime for the corresponding block hash from storageState
+func (s *StorageState) GetRuntime(blockHash common.Hash) (instance runtime.Instance, err error) {
```
timwu20, line 313: why are we adding a `GetRuntime` method to the storage state? Shouldn't the `blockTree` contain all the runtimes?

dot/core/service.go
```diff
@@ -191,7 +191,15 @@ func (s *Service) handleBlock(block *types.Block, state *rtstorage.TrieState) er
...
 
 	rt, err := s.blockState.GetRuntime(&block.Header.ParentHash)
-	if err != nil {
+	if errors.Is(err, blocktree.ErrFailedToGetRuntime) {
+		rt, err = s.storageState.GetRuntime(block.Header.ParentHash)
+		if err != nil {
+			return fmt.Errorf("getting runtime from database: %w", err)
+		}
+		// ensure the runtime stops and releases resources since it was
+		// instantiated from disk just for this function call.
```
timwu20, line 200: Why is it the responsibility of the caller of `BlockState.GetRuntime` to get the runtime from the `StorageState` if it doesn't exist? Why doesn't it exist?

dot/core/interface.go
edwardmack: I had this returning an error, but updated based on this comment: https://github.com/ChainSafe/gossamer/pull/2781#discussion_r973095859

dot/state/storage.go
edwardmack: This PR introduces clearRuntimes that removes all but the current runtime from blockTree, so we need a way to get runtimes from storage state.

dot/core/service.go
edwardmack: It may not exist in BlockState because it's not the current runtime.  This was updated based on request from this comment: https://github.com/ChainSafe/gossamer/pull/2781#pullrequestreview-1099168287

dot/state/block_test.go
edwardmack: I didn't want to use a mock for blockState since then the test would then be just testing the mock and not the way that runtime is actually stored.

dot/state/block_test.go
edwardmack: I didn't want to use a mock for blockState since then the test would then be just testing the mock and not the way that runtime is actually stored.

qdm12 (approved):

dot/state/block.go
```diff
@@ -651,11 +651,25 @@ func (bs *BlockState) GetRuntime(hash *common.Hash) (runtime.Instance, error) {
 	return bs.bt.GetBlockRuntime(*hash)
 }
 
+//  GetBestBlockRuntime returns the current runtime instance for the best block
+func (bs *BlockState) GetBestBlockRuntime() (instance runtime.Instance) {
+	instance, err := bs.bt.GetBlockRuntime(bs.BestBlockHash())
+	if err != nil {
+		panic(fmt.Sprintf("we should always succeed getting the best block runtime but an error occurred: %s", err))
```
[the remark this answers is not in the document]
timwu20: At some point we need to stop relying on panics to stop the node.

dot/core/service.go
timwu20: Why is the responsibility of the caller to manage this? I'm still not sure why the runtime wouldn't exist.  Where should the runtimes exist? The block tree or the storage state? It shouldn't be on the onus of the caller to have to manage this.  Why can't we just retrieve the runtime from whatever is supposed to contain it, and leave the caching in memory to that package.

edwardmack:
Closing for now, will consider best design for handling runtimes in the future.

PR #2782, opened 2022-08-24
chore(dot/state): refactor `NewTries`

qdm12:
## Changes

- `NewTries` does not return an error (always nil)
- `NewTries` does not take an initial as argument (which won't get versioned)
- Add `(*Tries) SetEmptyTrie()` method (which won't get versioned)
- Add `(*Tries) SetTrie(*trie.Trie)` method (which **will** get versioned)

This PR is because the empty trie is the same for v0 and v1 state versions.
So we want a method `SetEmptyTrie` on Tries which doesn't require a state version, and another method `SetTrie` which does.

This allows to have way less code changes in the next chunky PR versioning trie calls.

## Tests

```sh
```

## Issues

#2418 

## Primary Reviewer

@kishansagathiya

### DIFF
--- a/dot/rpc/modules/dev_integration_test.go
+++ b/dot/rpc/modules/dev_integration_test.go
@@ -42,8 +42,7 @@ func newState(t *testing.T) (*state.BlockState, *state.EpochState) {
 	db := state.NewInMemoryDB(t)
 
 	_, genesisTrie, genesisHeader := genesis.NewTestGenesisWithTrieAndHeader(t)
-	tries, err := state.NewTries(genesisTrie)
-	require.NoError(t, err)
+	tries := state.NewTries(genesisTrie)
 	bs, err := state.NewBlockStateFromGenesis(db, tries, genesisHeader, telemetryMock)
 	require.NoError(t, err)
 	es, err := state.NewEpochStateFromGenesis(db, bs, genesisBABEConfig)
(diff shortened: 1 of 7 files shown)


dot/core/interface.go
```diff
@@ -29,7 +29,7 @@ type RuntimeInstance interface {
 	Exec(function string, data []byte) ([]byte, error)
 	SetContextStorage(s runtime.Storage)
 	GetCodeHash() common.Hash
-	Version() (runtime.Version, error)
+	Version() (version runtime.Version)
```
EclesioMeloJunior, line 32: was not these changes already merged in another branch? Should you rebase this branch with development?

dot/core/mocks_test.go
```diff
@@ -1164,12 +1164,11 @@ func (mr *MockRuntimeInstanceMockRecorder) Validator() *gomock.Call {
 }
 
 // Version mocks base method.
-func (m *MockRuntimeInstance) Version() (runtime.Version, error) {
+func (m *MockRuntimeInstance) Version() runtime.Version {
```
EclesioMeloJunior, line 1167: was not these changes already merged in another branch? Should you rebase this branch with development?

EclesioMeloJunior (changes requested):
should you rebase development into this branch

dot/core/interface.go
qdm12: Indeed, this is fixed up, sorry about that!

dot/core/mocks_test.go
qdm12: Indeed, this is fixed up, sorry about that!

EclesioMeloJunior (approved):

jimjbrettj (approved):

PR #2774, opened 2022-08-23
chore(all): version trie calls

qdm12:
## Changes

- [x] Version relevant trie calls (Put, Hash)
- [x] Convert wasmer Core_version `Version` to typed `trie.Version`
- [x] Adapt tests/test helpers to read the state version from the genesis runtime code
- [ ] Fix unit tests
- [ ] Fix integration tests

## Tests

Entire codebase test suite

## Issues

#2418 

## Primary Reviewer

@edwardmack

PR #2789, opened 2022-08-29
chore(dot/core): refactor `NewTestService`

qdm12:
## Changes

💁  this it to ease test changes in #2774  as well as doing less dark magic in tests.

## Tests

```sh
go test -tags integration github.com/ChainSafe/gossamer/dot/core/...
```

## Primary Reviewer

@EclesioMeloJunior

### DIFF
--- a/dot/core/helpers_test.go
+++ b/dot/core/helpers_test.go
@@ -130,7 +130,7 @@ func createTestService(t *testing.T, genesisFilePath string,
 
 	cfgCodeSubstitutedState = stateSrvc.Base
 
-	cfg := &Config{
+	cfg := Config{
 		Keystore:             cfgKeystore,
 		LogLvl:               log.Critical,
 		BlockState:           cfgBlockState,
@@ -256,7 +256,7 @@ func NewTestService(t *testing.T, cfg *Config) *Service {
 		cfg.CodeSubstitutedState = stateSrvc.Base
 	}
 
-	s, err := NewService(cfg)
+	s, err := NewService(*cfg)
 	require.NoError(t, err)
 
 	return s
--- a/dot/core/helpers_test.go
+++ b/dot/core/helpers_test.go
@@ -149,13 +149,9 @@ func createTestService(t *testing.T, genesisFilePath string,
 }
 
 // NewTestService creates a new test core service
-func NewTestService(t *testing.T, cfg *Config) *Service {
+func NewTestService(t *testing.T, cfg Config) *Service {
 	t.Helper()
 	ctrl := gomock.NewController(t)
-
-	if cfg == nil {
-		cfg = &Config{}
-	}
 
 	if cfg.Keystore == nil {
 		cfg.Keystore = keystore.NewGlobalKeystore()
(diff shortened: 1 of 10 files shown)


dot/core/helpers_test.go
```diff
@@ -148,27 +148,11 @@ func createTestService(t *testing.T, genesisFilePath string,
...
-		cfg.Keystore = keystore.NewGlobalKeystore()
-		kp, err := sr25519.GenerateKeypair()
-		if err != nil {
-			t.Fatal(err)
-		}
-		err = cfg.Keystore.Acco.Insert(kp)
-		require.NoError(t, err)
-	}
-
-	cfg.LogLvl = 3
```
kishansagathiya, line 170: I would rather have this piece of code here once than having this code being repeated everywhere newTestService is used.

That is way too much repetitive. I like the current approach better. In the current approach, `NewTestService` takes values from the config passed to it and for whatever that is not set in config, it would use default config values.

dot/core/helpers_test.go
```diff
@@ -148,27 +148,11 @@ func createTestService(t *testing.T, genesisFilePath string,
 	return service, encodedExtrinsic
 }
 
-// NewTestService creates a new test core service
-func NewTestService(t *testing.T, cfg *Config) *Service {
+// newTestService creates a new test core service
+func newTestService(t *testing.T, cfg Config) *Service {
```
kishansagathiya, line 152: There are arguments for and against using pointers.
- using a reference is obviously cheaper resource-wise, compared to passing a value, while
- values are safer. no nil pointer error, saves you from race conditions, etc.
- 
But, more often than not, people use pointers or values based on what seems more convenient at the time and based on their personal preferences. 

Unless there is a strong reason for changing such code, I would prefer to have it stay as it is.

kishansagathiya (changes requested):
Don't really agree with the contents of this PR.

I want to know the opinions of others devs on this one to change my mind.

dot/core/helpers_test.go
qdm12: I get your point, but when debugging a test this is quite horrible to dig into helping code (and helping code calling other helping code) to find what is configured how. On top of that, often tests don't need all the defaults set and might need only one or two fields set, which makes it a lot easier to understand and modify.

I really pulled my hair changing tests in #2774 which has ramifications in the entire codebase, that's why I made this separate PR.

dot/core/helpers_test.go
qdm12: > using a reference is obviously cheaper resource-wise, compared to passing a value

Not at all. Passing by value is most of the time faster and puts less pressure on the GC. A good rule of thumb is to pass by pointers >= 80 bytes structs. Now in this particular case, the config struct has 8 pointers (8x8 bytes) + log level (1 byte) + map (I think it's ~24B) = 89 bytes, so technically yes passing by pointer would yield a small performance advantage. BUT practically speaking, passing a config (even in production code) to a service is not something executed often so the optimisation is very useless. Worth dumbing down the code to use values and avoid nil pointer checks/panics.

> saves you from race conditions

I think you meant accidental mutation, which is also debatable, since often there are pointers in struct such a maps or slices, so it doesn't shield it completely either.

Sent a message on Slack about this to get alignment across the team :wink:

timwu20:
closing due to inactivity.

PR #2788, opened 2022-08-29
chore(deps): bump most libp2p dependencies

qdm12:
## Changes

Unblocks #2768 

- `github.com/libp2p/go-libp2p-core/*` -> `github.com/libp2p/go-libp2p/core/*`
- `github.com/libp2p/go-libp2p-peerstore/pstoreds` -> `github.com/libp2p/go-libp2p/p2p/host/peerstore/pstoreds`
- `github.com/libp2p/go-libp2p-discovery` -> `github.com/libp2p/go-libp2p/p2p/discovery/routing`
- Fix btcsuite/btcd module issues
- Update mocks

## Tests

```sh
go test -tags integration github.com/ChainSafe/gossamer
```

## Issues

Unblocks #2768 

## Primary Reviewer

@jimjbrettj

Issue #2768, opened 2022-08-22
Go 1.19 support


q9f:
Hello 😊 

## Issue summary

- compiling gossamer should work with the latest go version
- however, the dependency https://github.com/lucas-clemente/quic-go does not support 1.19 yet

```
~/.opt/ChainSafe/gossamer development 12s
❯ go version
go version go1.19 linux/amd64
```

## Other information and links

```
# github.com/lucas-clemente/quic-go/internal/qtls
../../../.go/pkg/mod/github.com/lucas-clemente/quic-go@v0.27.1/internal/qtls/go119.go:6:13: cannot use "The version of quic-go you're using can't be built on Go 1.19 yet. For more details, please see https://github.com/lucas-clemente/quic-go/wiki/quic-go-and-Go-versions." (untyped string constant "The version of quic-go you're using can't be built on Go 1.19 yet. F...) as int value in variable declaration
make: *** [Makefile:73: build] Error 2
```

How are you managing go versions usually?

q9f:
Just noticed #2719

Wondering if bumping quic to 0.28 would address this?

qdm12:
This is blocked by a bunch of dependency upgrade regarding libp2p libraries, and substantial manual work.
Definitely to do for the fun of it, but definitely not a priority either. I have a local branch with a few changes (regarding deps upgrade, not Go 1.19 yet), I'll probably finish it in my 'friday evening refactoring fun' 😉

qdm12:
Blocked by https://github.com/libp2p/go-libp2p-kad-dht/issues/790

PR #2793, opened 2022-08-30
chore(deps): btcsuite/btcd from v0.22.1 to v0.23.1

qdm12:
## Changes

Bump that dependency which is required to bump a bunch of libp2p dependencies.

## Issues

Unblocks #2788 to solve #2768 

## Primary Reviewer

@jimjbrettj


qdm12:
Not needed actually

PR #2794, opened 2022-09-01
chore(dot/network): use `mdns` instead of `mdns_legacy`

qdm12:
## Tests

```sh
MODE=rpc go test -run ^TestStableNetworkRPC$ github.com/ChainSafe/gossamer/tests/rpc -v
```

## Issues

Related to #2768 and #2788


qdm12:
@Stebalien This PR only focuses on moving to the newer `mdns` package. I also tried with the latest `github.com/libp2p/go-libp2p` without success (on #2788).

One of the test failing is an integration test `TestStableNetworkRPC` running 3 nodes, and no node is found (should be `2`).

To help you reproduce:

```sh
git clone --single-branch --branch qdm12/dot/network/mdns-upgrade --depth 1 https://github.com/ChainSafe/gossamer.git
cd gossamer
MODE=rpc go test -run ^TestStableNetworkRPC$ github.com/ChainSafe/gossamer/tests/rpc -v
```

Thank you in advance for any tips! 👍

Stebalien:
Hm. It looks like it could be related to https://github.com/libp2p/go-libp2p/issues/1333.

Stebalien:
cc @marten-seemann

marten-seemann:
> Hm. It looks like it could be related to https://github.com/libp2p/go-libp2p/issues/1333.

Not sure. Could be, but we have a similar test case in go-libp2p (some nodes discovering each other) that works fine.

qdm12:
Thanks @marten-seemann and @Stebalien for helping out! 👍 
I tried to separate each of the 3 Gossamer node launches by 1 minute, and then wait an extra 10 minutes and it still doesn't find other peers 😢 So maybe another problem?

For extra debugging information, the notifee `HandlePeerFound(p peer.AddrInfo)` is never called. The mdns is created though (`NewMDNSService` doesn't hang).

Do we have to perhaps change the libp2p host creation or change the service name string? (we use the same `string(m.host.protocolID)` as before for now)? Thanks!

qdm12:
> Could be, but we have a similar test case in go-libp2p (some nodes discovering each other) that works fine.

Can you point me to it please? Maybe I can spot what's missing/wrong in those code changes or our existing code.

qdm12:
I changed this to use our own `internal/mdns` implementation based off libp2p's `mdns_legacy`. It's still using `github.com/whyrusleeping/mdns` under the hood, maybe upgrading to `github.com/hashicorp/mdns` would be nice once their race condition is fixed in https://github.com/hashicorp/mdns/pull/75 but really not a necessity.

With this merged, #2788 should finally be unblocked.

internal/mdns/interfaces.go
```diff
@@ -0,0 +1,26 @@
...
+
+package mdns
+
+import (
+	"github.com/libp2p/go-libp2p-core/network"
+	"github.com/libp2p/go-libp2p-core/peer"
+)
+
+// Logger is a logger interface for the mDNS service.
+type Logger interface {
```
jimjbrettj, line 12: Why do we need this and not just use internal LeveledLogger instead?

jimjbrettj (approved):
lgtm

internal/mdns/dialable.go
```diff
@@ -0,0 +1,60 @@
...
+	"net"
+
+	manet "github.com/multiformats/go-multiaddr/net"
+)
+
+var (
+	ErrTCPListenAddressNotFound = errors.New("TCP listen address not found")
+)
+
+func getMDNSIPsAndPort(p2pHost Networker) (ips []net.IP, port uint16) {
```
timwu20, line 18: nit:
```suggestion
func getMDNSIPsAndPort(networker Networker) (ips []net.IP, port uint16) {
```

timwu20, line 18: no tests for this package?

internal/mdns/mdns.go
```diff
@@ -0,0 +1,225 @@
...
+
+	server, err := mdns.NewServer(&mdns.Config{Zone: mdnsService})
+	if err != nil {
+		return fmt.Errorf("creating mDNS server: %w", err)
+	}
+	s.server = server
+
+	s.started = true
+	s.stop = make(chan struct{})
+	s.done = make(chan struct{})
```
timwu20, line 72: Should we put these lines in the constructor?

internal/mdns/mdns.go
```diff
@@ -0,0 +1,225 @@
...
+	// so we wait for the run goroutine to be ready.
+	<-ready
+
+	return nil
+}
+
+// Stop stops the mDNS service and server.
+func (s *Service) Stop() (err error) {
+	if !s.started {
+		return nil
```
timwu20, line 86: we should probably return an error in this case?

internal/mdns/mdns.go
```diff
@@ -0,0 +1,225 @@
...
+	}()
+	close(s.stop)
+	<-s.done
+	return s.server.Shutdown()
+}
+
+func (s *Service) run(ready chan<- struct{}) {
+	defer close(s.done)
+
+	const pollPeriod = time.Minute
```
timwu20, line 100: this should be configurable on the service itself.  If there were tests we wouldn't want to wait a minute.

internal/mdns/mdns.go
```diff
@@ -0,0 +1,225 @@
...
+				err := s.handleEntry(entry)
+				if err != nil {
+					s.logger.Warnf("handling mDNS entry: %s", err)
+				}
+			case <-stopListening:
+				continueListening = false
+			case <-stop:
+				return
+			}
+		}
```
timwu20, line 178:
```suggestion
	listening:
		for {
			// Listen for entries until we receive a stop listening signal.
			select {
			case entry := <-entries:
				err := s.handleEntry(entry)
				if err != nil {
					s.logger.Warnf("handling mDNS entry: %s", err)
				}
			case <-stopListening:
				break listening
			case <-stop:
				return
			}
		}
```

internal/mdns/mdns.go
```diff
@@ -0,0 +1,225 @@
...
+		case <-ticker.C:
+		case <-s.stop:
+			close(entriesListeningLoopStop)
+			<-entriesListeningLoopDone
+			close(entriesCh)
+			close(entriesStartListening)
+			close(entriesStopListening)
+			return
+		}
+	}
```
timwu20, line 148: This whole thing seems overly complicated to just periodically query the mdns and handle the entries.  Couldn't we just run a goroutine in the loop itself and just process the events for each query? I created [another PR](https://github.com/ChainSafe/gossamer/pull/2986) against this one to demonstrate.

timwu20 (changes requested):

internal/mdns/mdns.go
qdm12: Actually that's a good question, what should a service implementer do if

- the service is stopped but was already stopped
- the service is started but was already started

Possible solutions:

1. just return nil, since the state transition is already done
2. return an error
3. panic, since it's a programming error on the caller side
4. not care about the state of the service

I think 4. is risky and would lead to possibly working but unexpected behavior, although it's easier for a service implementer. I would rather do 3. than 2., since really it's a programming error on the caller, not the service failing to stop due to an error (otherwise it might got unnoticed). Then 1. works fine (we don't need to crash in the end really), but allows for bad calling code. So in my mind it's 1. or 3., maybe leaning towards 3. Any thoughts?


-

internal/mdns/mdns.go
qdm12: These are here so the same service can be started -> stopped multiple times. I think we should design our services as such, what do you think?

internal/mdns/dialable.go
qdm12: I can add some, the reason I didn't was:

- the original mdns_legacy package we were using did not have any test (well it has one that was skipped and not working :cry:)
- this is a bit of a temporary fix, ideally we can upgrade to the newer mdns package, but this breaks all our stuff for now (see discussion above)
- rpc tests with multiple Gossamer nodes fail when that new mdns package is badly implemented

internal/mdns/mdns.go
qdm12: This increases indentation :cry: And also labels are not really common in Go code, so I tend to avoid for readers/newcomers. Keeping this as is :wink:

internal/mdns/mdns.go
qdm12: It was simpler (similar to your PR) before commit https://github.com/ChainSafe/gossamer/pull/2794/commits/9a0ae60f79ee59aa907faca4634741b31f5549d9

Why I changed it to this was to:
- avoid closing the channel on the reading side (our side)
- re-use the same buffered channel without creating it on every tick... but yeah the tick is 1 minute, I didn't realise :sweat: 

Reverted it to how it was (a bit better with ready and done channels)

internal/mdns/mdns.go
qdm12: Indeed :+1: 
Moved as a struct field. I didn't expose it as argument to the service creation, since it wasn't exposed in `mdns_legacy` and we would lose access to it once upgrading to the newer `mdns` (when they fix their stuff)

internal/mdns/dialable.go
qdm12: Changed to `network interfaceListenAddressesGetter` where `interfaceListenAddressesGetter` is an interface the `p2phost.Network()` returned value implements

internal/mdns/mdns.go
timwu20: I wouldn't say this is a requirement in this case.  Would we ever start and stop this service?

internal/mdns/dialable.go
timwu20: Do we have an issue to track using the new mdns package? If we're not gonna add tests to this, we should ensure everything is working on staging before merging this.

internal/mdns/mdns.go
timwu20: Well we can't really ensure that we don't call `Stop` more than once.  So I'd probably just return an error in this case.

qdm12: I don't see the point of an error, it's not something coming from the service failing to start or stop really. It's more of a programming error on the caller. If we want to pay attention to this, we should panic instead so the caller can fix up their code (instead of returning an error which might be just logged and never seen).

I'm also thinking more and more 1. is fine. If we send a network request to change state of something to X, and it's already at state X, I'd say it's fine to return an OK status (=nil error here).

internal/mdns/mdns.go
qdm12: Just thinking in terms of service design in general, it should be restart-able. There is no reason to restart this service, but why make code worst if it can be better for the same amount of lines :wink:

internal/mdns/dialable.go
qdm12: Created https://github.com/ChainSafe/gossamer/issues/2993
It's deploying now on staging.

internal/mdns/mdns.go
timwu20: I'd advise against panicing this case.  From previous experience on implementing this interface for data pipelines, there could be cases where `Stop()` could be called multiple times.  Take the `SIGINT` to `SIGTERM` example.  If you had something listening on both signals, you could trigger `Service.Stop()` in both cases.

timwu20 (approved):

internal/mdns/mdns.go
qdm12: Yeah ok, this could make sense. Should we then just return `nil`? We're trying to transition to the same state, I don't think we should really return an error right? :thinking: (haha I know jumping from panic to return nil feels odd)

timwu20: I think returning `nil` is fine.  In what case would you return an error for this service?

qdm12: - When starting fails due to the setup before launching the internal goroutine (setup files, download stuff, check settings, etc.)
- When stopping fails due to the internal goroutine failing to stop or cleanup (remove files, close connections etc.)

I'm very shared between panicing and returning nil, but on the other hand, sending a request to change a state from X to X often succeed, so I guess it's fine to return nil and not panic (leading to possibly worst calling code, but the caller might not care about the service telling them this :smile:)

qdm12: Asked on reddit: https://www.reddit.com/r/golang/comments/101ah0j

Looks like `return nil` is the preferred way so it stays idempotent.

PR #2795, opened 2022-09-01
chore(lib/genesis): less dependencies on other packages

qdm12:
## Changes

TLDR: clear up `lib/genesis` cycle dependency hell 🔥 on `lib/runtime/wasmer` and `lib/trie`

That's gonna look like it's useless changes, but I actually hated my day working on this and it's really needed 😄 

The reason is #2774 needs to run wasmer code to extract the state version from the runtime code in the genesis mapping.
And as it is, it makes lib/genesis import lib/runtime/wasmer which opens a whole can of worms of circular dependencies.

Anyway this fixes it and moves functions around such that `lib/genesis` does not import `lib/runtime/wasmer` nor `lib/trie` anymore.

- Remove all utils/test helping global variables from `lib/genesis` and inline them where needed
- **duplicate** test-only code helping functions where they are needed in `helpers_test.go` and unexport them. That way the Go production API is reduced and the lib/genesis imports are less. I know it's tempting to dedup the code, but when it comes to test helping functions we should 100% duplicate to avoid cycle dependency hell 👿
- Change test helping functions to return genesis, trie and block header as values instead of pointers. This is part of my movement to *not use pointers unless required* for performance reasons (stack is better) and safety (no nil check needed).
- Move lib/genesis's `NewGenesisBlockFromTrie` function as `lib/trie`  `Trie` method `GenesisBlock()` (same thing, less genesis dependencies)
- Move lib/genesis's `NewTrieFromGenesis` function to `lib/runtime/wasmer` since it will need in #2774 to use wasmer functions.
- Move key constants from `lib/runtime/constants.go` to `lib/genesis/constants.go`

## Tests

```sh
```

## Issues

Slow boat towards #2418 

## Primary Reviewer

@EclesioMeloJunior


dot/core/helpers_test.go
```diff
@@ -144,7 +167,7 @@ func getGssmrRuntimeCode(t *testing.T) (code []byte) {
 	gssmrGenesis, err := genesis.NewGenesisFromJSONRaw(path)
 	require.NoError(t, err)
 
-	trie, err := genesis.NewTrieFromGenesis(gssmrGenesis)
+	trie, err := wasmer.NewTrieFromGenesis(*gssmrGenesis)
```
qdm12, line 170: This is moved to wasmer since it will need to run wasmer code to get the state version from the runtime code contained in the genesis.

dot/node.go
```diff
@@ -144,13 +145,13 @@ func (*nodeBuilder) initNode(cfg *Config) error {
...
 	// create trie from genesis
-	t, err := genesis.NewTrieFromGenesis(gen)
+	t, err := wasmer.NewTrieFromGenesis(*gen)
 	if err != nil {
 		return fmt.Errorf("failed to create trie from genesis: %w", err)
 	}
 
 	// create genesis block from trie
-	header, err := genesis.NewGenesisBlockFromTrie(t)
+	header, err := t.GenesisBlock()
```
qdm12, line 154: this is moved as a method on the trie since it doesn't add dependencies in lib/trie, but allows to remove the lib/trie dependency in lib/genesis

dot/rpc/helpers_test.go
```diff
@@ -0,0 +1,37 @@
...
+	stateRoot := genesisTrie.MustHash()
+	extrinsicRoot := trie.EmptyHash
+	const number = 0
+	digest := types.NewDigest()
+	genesisHeader, err = types.NewHeader(parentHash,
+		stateRoot, extrinsicRoot, number, digest)
+	require.NoError(t, err)
+
+	return gen, genesisTrie, genesisHeader
+}
```
qdm12, line 37: That's repetition indeed, but this is TEST CODE only and should very much NOT be exported to reduce cycle dependency hell.

lib/genesis/helpers.go
```diff
@@ -16,8 +16,7 @@ import (
 	"github.com/ChainSafe/gossamer/dot/types"
 	"github.com/ChainSafe/gossamer/lib/common"
 	"github.com/ChainSafe/gossamer/lib/crypto"
-	"github.com/ChainSafe/gossamer/lib/runtime"
-	"github.com/ChainSafe/gossamer/lib/trie"
+	"github.com/ChainSafe/gossamer/lib/runtime/constants"
```
qdm12, line 19: New subpackage containing just a few hex trie keys, to avoid cycle dependency hell. Also an example why we need moare subpackages.

lib/runtime/wasmer/genesis_test.go
```diff
@@ -0,0 +1,30 @@
...
+	}
+
+	expTrie := trie.NewEmptyTrie()
+	expTrie.Put([]byte(`:code`), []byte{1, 2})
+
+	trie, err := NewTrieFromGenesis(rawGenesis)
+	require.NoError(t, err)
+
+	require.Equal(t, expTrie, trie)
+}
```
qdm12, line 30: note to self: make better

lib/trie/genesis.go
```diff
@@ -0,0 +1,30 @@
...
+	extrinsicRoot := EmptyHash
+	const blockNumber = 0
+	digest := types.NewDigest()
+	genesisHeaderPtr, err := types.NewHeader(parentHash, rootHash, extrinsicRoot, blockNumber, digest)
+	if err != nil {
+		return genesisHeader, fmt.Errorf("creating genesis block header: %w", err)
+	}
+
+	return *genesisHeaderPtr, nil
+}
```
qdm12, line 30: note to self: add test

lib/runtime/wasmer/genesis_test.go
qdm12: done

lib/genesis/genesis_test.go
```diff
@@ -6,47 +6,59 @@ package genesis
 import (
 	"testing"
 
-	"github.com/stretchr/testify/require"
+	"github.com/stretchr/testify/assert"
```
jimjbrettj, line 9: Why assert over require here?

jimjbrettj (approved):
nice stuff!

lib/genesis/genesis_test.go
qdm12: I didn't even realize that got changed 😅
In my defense (but i'll revert it), `assert` is better when you just want to report the test error but there is no need to fail the test right now. So I use `require` really when test execution should stop (i.e. failed to create a test service).

qdm12: reverted

dot/rpc/helpers_test.go
```diff
@@ -0,0 +1,39 @@
...
+	"github.com/ChainSafe/gossamer/lib/genesis"
+	"github.com/ChainSafe/gossamer/lib/runtime/wasmer"
+	"github.com/ChainSafe/gossamer/lib/trie"
+	"github.com/ChainSafe/gossamer/lib/utils"
+	"github.com/stretchr/testify/require"
+)
+
+func newTestGenesisWithTrieAndHeader(t *testing.T) (
+	gen genesis.Genesis, genesisTrie trie.Trie, genesisHeader types.Header) {
+	genesisPath := utils.GetGssmrV3SubstrateGenesisRawPathTest(t)
```
EclesioMeloJunior, line 20: would suggest to add `t.Helper()` at the beginning of each `newTestGenesisWithTrieAndHeader` test function
```suggestion
func newTestGenesisWithTrieAndHeader(t *testing.T) (
	gen genesis.Genesis, genesisTrie trie.Trie, genesisHeader types.Header) {
	t.Helper()
	genesisPath := utils.GetGssmrV3SubstrateGenesisRawPathTest(t)
```

EclesioMeloJunior (changes requested):

lib/genesis/helpers.go
EclesioMeloJunior: @qdm12 the `constants` package you introduce here just contains some constants that is only used by production code while initializing the node and building the Genesis struct `dot/state/initialize.go`, I would suggest to move those constants to `lib/genesis` since they are specifically keys at the genesis file

qdm12: These are are used by both wasmer and genesis packages (and wasmer imports genesis). We can thus either:

- keep it as is in a separate dependency-free package
- duplicate the constants in both wasmer and genesis

But it feels a bit better to have them in that constants package, what do you think?

lib/runtime/wasmer/genesis.go
```diff
@@ -0,0 +1,34 @@
...
+	tr = trie.NewEmptyTrie()
+	genesisFields := gen.GenesisFields()
+	keyValues, ok := genesisFields.Raw["top"]
+	if !ok {
+		return nil, fmt.Errorf("%w: in genesis %s",
+			ErrGenesisTopNotFound, gen.Name)
+	}
+
+	err = tr.LoadFromMap(keyValues)
+	if err != nil {
```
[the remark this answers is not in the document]
qdm12: Not possible since the wasmer package imports the trie package, so we cannot have this in the trie package. I wish it would be possible and totally agree apart from the cycle dependency problem!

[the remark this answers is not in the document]
qdm12: As I mentioned above:

> Soon this will instantiate a wasmer instance from the genesis runtime code, and load the state version from it. That's why this is in the wasmer package.

So it won't work. That PR was especially made for the upcoming branch based on it (#2774) 😉

lib/genesis/helpers.go
qdm12: Move to `lib/genesis/keys.go`

qdm12: Also these constants are used by some tests in the wasmer package, not just production code.

EclesioMeloJunior (approved):
LGTM!

PR #2804, opened 2022-09-06
chore(dot/state): `newTestStorageState` does not take `tries` argument

qdm12:
Tiny simplification for test helper always taking `newEmptyTries()` argument

### DIFF
--- a/dot/state/block_test.go
+++ b/dot/state/block_test.go
@@ -30,12 +30,9 @@ func newTestBlockState(t *testing.T, tries *Tries) *BlockState {
 	telemetryMock.EXPECT().SendMessage(gomock.Any()).AnyTimes()
 
 	db := NewInMemoryDB(t)
-	header := types.Header{
-		StateRoot: trie.EmptyHash,
-		Digest:    types.NewDigest(),
-	}
+	header := testGenesisHeader
 
-	bs, err := NewBlockStateFromGenesis(db, tries, &header, telemetryMock)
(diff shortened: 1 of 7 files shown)


EclesioMeloJunior (approved):

dot/state/block_test.go
```diff
@@ -24,17 +24,18 @@ var testGenesisHeader = &types.Header{
...
-func newTestBlockState(t *testing.T, header *types.Header, tries *Tries) *BlockState {
+func newTestBlockState(t *testing.T, tries *Tries) *BlockState {
 	ctrl := gomock.NewController(t)
 	telemetryMock := NewMockClient(ctrl)
 	telemetryMock.EXPECT().SendMessage(gomock.Any()).AnyTimes()
 
 	db := NewInMemoryDB(t)
-	if header == nil {
-		header = testGenesisHeader
+	header := types.Header{
```
timwu20, line 33: why are we not returning `testGenesisHeader` anymore?

timwu20 (changes requested):

dot/state/block_test.go
qdm12: Trying to steer away from global variables slowly

timwu20: But it's not intuitive that it still uses `testGenesisHeader` which is asserted on in a bunch of those tests that used to provide `testGenesisHeader` to `newTestBlockState`.

qdm12: Agreed, reverted 👍

timwu20 (approved):

PR #2809, opened 2022-09-07
fix(core): fix txn pool for latest runtime

jimjbrettj:
## Changes

Improved txnPool logic and added txn formatting based on txnQueueApiVersion to support the latest runtime.

## Tests

```sh
go test ./dot/core
```

## Issues

https://github.com/ChainSafe/gossamer/issues/2462

## Primary Reviewer

@noot


dot/core/errors.go
```diff
@@ -8,17 +8,12 @@ import (
 )
 
 var (
-
```
jimjbrettj, line 11: These were not being used

dot/core/service.go
```diff
@@ -317,6 +323,7 @@ func (s *Service) handleChainReorg(prev, curr common.Hash) error {
 	}
 
 	// Check transaction validation on the best block.
+	// TODO does this need to be updated like txnPool did
 	rt, err := s.blockState.GetRuntime(nil)
```
jimjbrettj, line 327: Should this be updated to take in a blockHash similar to the change to maintain transaction pool?

dot/core/service.go
```diff
@@ -266,6 +269,9 @@ func (s *Service) handleCodeSubstitution(hash common.Hash,
 // does not need to be completed before the next block can be imported.
 func (s *Service) handleBlocksAsync() {
 	for {
+		// TODO Look into if this is best way to get prev
+		// Maybe pass in block and get parent rather than assume that
+		// best block hash is the parent
 		prev := s.blockState.BestBlockHash()
```
jimjbrettj, line 275: Leaving this comment here to get reviewers thoughts on it

dot/core/helpers_test.go
```diff
@@ -17,10 +19,174 @@ import (
...
+	require.NoError(t, err)
+	bKey = append(bKey, h0...)
+	h1, err := common.Twox128Hash([]byte("Account"))
+	require.NoError(t, err)
+	bKey = append(bKey, h1...)
+	h2, err := common.Blake2b128(pub)
+	require.NoError(t, err)
+	bKey = append(bKey, h2...)
+	bKey = append(bKey, pub...)
+	return
```
qdm12, line 40: nit use `concatenateByteSlices`:

```suggestion
	h1, err := common.Twox128Hash([]byte("Account"))
	require.NoError(t, err)
	h2, err := common.Blake2b128(pub)
	require.NoError(t, err)
	return concatenateByteSlices(h0, h1, h2, pub)
```

dot/core/helpers_test.go
```diff
@@ -17,10 +19,174 @@ import (
...
+}
+
+// Note might need to pass in Config but will see
+// CreateTestService is a new helper function to clean up orchestration for testing the txnPool
+func CreateTestService(t *testing.T, genesisFilePath string,
+	pubKey []byte, accInfo types.AccountInfo, ctrl *gomock.Controller) (*Service, []byte) {
```
qdm12, line 46: nit name your returns, especially for the byte slice

dot/core/helpers_test.go
```diff
@@ -17,10 +19,174 @@ import (
...
+	gen, err := genesis.NewGenesisFromJSONRaw(genesisFilePath)
+	require.NoError(t, err)
+
+	// Trie Created
+	genTrie, err := genesis.NewTrieFromGenesis(gen)
+	require.NoError(t, err)
+
+	// Extrinsic and context related stuff
+	aliceBalanceKey := balanceKey(t, pubKey)
+	encBal, err := scale.Marshal(accInfo)
```
qdm12, line 59: nit rename `encBal` to `encodedAccountInfo`?

dot/core/helpers_test.go
```diff
@@ -17,10 +19,174 @@ import (
...
+	require.NoError(t, err)
+
+	// Genesis header made
```
qdm12, line 62:
```suggestion
```

dot/core/helpers_test.go
```diff
@@ -17,10 +19,174 @@ import (
...
+	genesisHeader := &types.Header{
+		StateRoot: genTrie.MustHash(),
+		Number:    0,
+	}
+
+	cfg := &Config{} // Use this for now but might need to pass in
```
qdm12, line 68: Unclear comment 🤔

dot/core/helpers_test.go
```diff
@@ -17,10 +19,174 @@ import (
...
+	if cfg.Keystore == nil {
+		cfg.Keystore = keystore.NewGlobalKeystore()
+		kp, err := sr25519.GenerateKeypair()
+		if err != nil {
+			t.Fatal(err)
+		}
```
qdm12, line 74: nit

```suggestion
		require.NoError(t, err)
```

dot/core/helpers_test.go
```diff
@@ -17,10 +19,174 @@ import (
...
+		err = cfg.Keystore.Acco.Insert(kp)
+		require.NoError(t, err)
+	}
+
+	cfg.LogLvl = 3
```
qdm12, line 79:
```suggestion
	cfg.LogLvl = log.Critical
```

dot/core/helpers_test.go
```diff
@@ -17,10 +19,174 @@ import (
...
+	require.NoError(t, err)
+	bKey = append(bKey, h1...)
+	h2, err := common.Blake2b128(pub)
+	require.NoError(t, err)
+	bKey = append(bKey, h2...)
+	bKey = append(bKey, pub...)
+	return
+}
+
+// Note might need to pass in Config but will see
```
qdm12, line 43: Please definitely do this. The dark magic checking nil fields in the config and setting in them is just horrible to debug later down the road.

dot/core/messages.go
```diff
@@ -29,18 +28,26 @@ func (s *Service) validateTransaction(peerID peer.ID, head *types.Header, rt Run
 	rt.SetContextStorage(ts)
 
 	// validate each transaction
-	externalExt := types.Extrinsic(append([]byte{byte(types.TxnExternal)}, tx...))
+	externalExt, err := s.buildExternalTransaction(rt, tx)
+	if err != nil {
+		return nil, false, fmt.Errorf("unable to build transaction: %s", err)
```
qdm12, line 33: nit to unclutter `unable to` repetition with errors coming from a nested call

```suggestion
		return nil, false, fmt.Errorf("building transaction: %s", err)
```

dot/core/helpers_test.go
jimjbrettj: Let me know what you think, but instead of passing in I can also just remove all these conditionals I believe. We define an empty config and then do a bunch of nil checks for no reason. So maybe I just can remove all these nil checks for this case? It seems the options are simplify this code and always build up the config (which I feel is prob more straightforward for testing) or be more flexible and take in config as a param to a separate buildConfig method and do all the nil checking there.

I feel like the second might feel nicer but I'm not sure its necessary. 1) it adds complexity since Im not sure in what use case we are passing in a portion config in the core tests and 2) it doesn't remove the nil checks then lol. I would lean towards leaving as is where we construct in this test helper from an empty config and then removing the nil checks to chean it up (and like you said make it easier to debug. I think this makes more sense for this use case, but let me know what you think

qdm12: Either don't pass a config or fully pass it really. In the second case, avoid nil checks so that it force the dev to repeat and be explicit of what's needed.

If we always pass an empty config, then just create it within the helper and remove the nil checks 👍  that's fine

dot/core/helpers_test.go
```diff
@@ -17,10 +19,168 @@ import (
...
+	require.NoError(t, err)
+	h2, err := common.Blake2b128(pub)
+	require.NoError(t, err)
+	bytes := [][]byte{h0, h1, h2, pub}
+	return concatenateByteSlices(bytes)
+}
+
+// Note might need to pass in Config but will see
+// CreateTestService is a new helper function to clean up orchestration for testing the txnPool
+func CreateTestService(t *testing.T, genesisFilePath string,
```
qdm12, line 42: nit Although it's unexported since it's in a test file, change this to a lowercase to avoid any confusion about exports

dot/core/helpers_test.go
```diff
@@ -17,10 +19,168 @@ import (
...
+	pubKey []byte, accInfo types.AccountInfo, ctrl *gomock.Controller) (s *Service, encExtrinsic []byte) {
+	t.Helper()
+
+	// Create genesis
+	gen, err := genesis.NewGenesisFromJSONRaw(genesisFilePath)
+	require.NoError(t, err)
+
+	// Trie Created
```
qdm12, line 50: nit unneeded comments

```suggestion
	gen, err := genesis.NewGenesisFromJSONRaw(genesisFilePath)
	require.NoError(t, err)

```

dot/core/helpers_test.go
```diff
@@ -17,10 +19,168 @@ import (
...
+	h2, err := common.Blake2b128(pub)
+	require.NoError(t, err)
+	bytes := [][]byte{h0, h1, h2, pub}
+	return concatenateByteSlices(bytes)
+}
+
+// Note might need to pass in Config but will see
+// CreateTestService is a new helper function to clean up orchestration for testing the txnPool
+func CreateTestService(t *testing.T, genesisFilePath string,
+	pubKey []byte, accInfo types.AccountInfo, ctrl *gomock.Controller) (s *Service, encExtrinsic []byte) {
```
qdm12, line 43: nit `accInfo` ->  `accountInfo`, `s` -> `service` and `encExtrinsic` -> `encodedExtrinsic` would be nice

dot/core/helpers_test.go
```diff
@@ -17,10 +19,168 @@ import (
...
+	encodedAccountInfo, err := scale.Marshal(accInfo)
+	require.NoError(t, err)
+
+	genesisHeader := &types.Header{
+		StateRoot: genTrie.MustHash(),
+		Number:    0,
+	}
+
+	cfg := &Config{}
+	if cfg.Keystore == nil {
```
qdm12, line 65: remove this and the other if conditions below. You can also directly set a bunch of fields in `&Config{...}`.

dot/core/messages.go
```diff
@@ -28,7 +28,11 @@ func (s *Service) validateTransaction(head *types.Header, rt RuntimeInstance,
 	rt.SetContextStorage(ts)
 
 	// validate each transaction
-	externalExt := types.Extrinsic(append([]byte{byte(types.TxnExternal)}, tx...))
+	externalExt, err := s.buildExternalTransaction(rt, tx)
+	if err != nil {
+		return nil, fmt.Errorf("building transaction: %s", err)
```
qdm12, line 33: nit

```suggestion
		return nil, fmt.Errorf("building external transaction: %s", err)
```

dot/core/service.go
```diff
@@ -31,6 +31,9 @@ var (
 	logger                  = log.NewFromGlobal(log.AddContext("pkg", "core"))
 )
 
+const supportedTxVersion = 3
+const previousSupportedTxVersion = 2
```
qdm12, line 35: nit Define these two as constants inlined in `buildExternalTransaction`

dot/core/service.go
qdm12: Is previous meant to be the parent of the block? If so, we should definitely get the parent of the block by using `block.ParentHash` to look up its parent block right?

dot/core/service.go
```diff
@@ -346,15 +352,17 @@ func (s *Service) handleChainReorg(prev, curr common.Hash) error {
...
-			txv, err := rt.ValidateTransaction(externalExt)
+			externalExt, err := s.buildExternalTransaction(rt, ext)
+			if err != nil {
+				return fmt.Errorf("building external transaction: %w", err)
+			}
+
+			transactionValidity, err := rt.ValidateTransaction(externalExt)
 			if err != nil {
 				logger.Debugf("failed to validate transaction for extrinsic %s: %s", ext, err)
 				continue
```
qdm12, line 363: nit a bit out of scope, but shouldn't we return the error?

dot/core/service.go
```diff
@@ -372,17 +380,33 @@ func (s *Service) maintainTransactionPool(block *types.Block) {
 		s.transactionState.RemoveExtrinsic(ext)
 	}
 
+	bestBlockHash := s.blockState.BestBlockHash()
+	stateRoot, err := s.storageState.GetStateRootFromBlock(&bestBlockHash)
+	if err != nil {
+		logger.Errorf("could not get state root from block %s: %w", bestBlockHash, err)
```
qdm12, line 386: Shouldn't we return an error instead here?

dot/core/service.go
```diff
@@ -372,17 +380,33 @@ func (s *Service) maintainTransactionPool(block *types.Block) {
...
+	}
+
+	ts, err := s.storageState.TrieState(stateRoot)
+	if err != nil {
+		logger.Errorf(err.Error())
```
qdm12, line 391: Shouldn't we return an error instead here?

dot/core/service.go
```diff
@@ -372,17 +380,33 @@ func (s *Service) maintainTransactionPool(block *types.Block) {
...
 		if err != nil {
 			logger.Warnf("failed to get runtime to re-validate transactions in pool: %s", err)
 			continue
 		}
 
-		txnValidity, err := rt.ValidateTransaction(tx.Extrinsic)
+		rt.SetContextStorage(ts)
+		externalExt, err := s.buildExternalTransaction(rt, tx.Extrinsic)
+		if err != nil {
+			logger.Errorf("Unable to build external transaction: %s", err)
```
qdm12, line 406: Shouldn't we return an error instead here?
Do  we really want to continue if one extrinsic fails to build/validate?

lib/runtime/version.go
```diff
@@ -30,9 +30,24 @@ type Version struct {
...
 var (
+	// Blake2-8("TaggedTransactionQueue")
+	encodedTaggedTransactionQueue = [8]byte{0xd2, 0xbc, 0x98, 0x97, 0xee, 0xd0, 0x8f, 0x15}
+
 	ErrDecodingVersionField = errors.New("decoding version field")
 )
 
+// TaggedTransactionQueueVersion returns the TaggedTransactionQueueAPI version
+func (*Version) TaggedTransactionQueueVersion(runtimeVersion Version) uint32 {
+	txQueueVersion := uint32(0)
```
qdm12, line 41: nit

```suggestion
func (*Version) TaggedTransactionQueueVersion(runtimeVersion Version) (txQueueVersion uint32) {
```

lib/runtime/version.go
```diff
@@ -30,9 +30,24 @@ type Version struct {
...
+	for _, v := range runtimeVersion.APIItems {
```
qdm12, line 42: nit `v` -> `apiItem`

lib/runtime/version.go
```diff
@@ -30,9 +30,24 @@ type Version struct {
 }
 
 var (
+	// Blake2-8("TaggedTransactionQueue")
+	encodedTaggedTransactionQueue = [8]byte{0xd2, 0xbc, 0x98, 0x97, 0xee, 0xd0, 0x8f, 0x15}
```
qdm12, line 34: nit it would be fancy to have a new function in `lib/common/hasher.go`:

```go
// MustBlake2b8 returns the 8-bit blake2b hash of the input data
// and panics on error.
func MustBlake2b8(data []byte) (digest [8]byte) {
	const bytes = 1
	hasher, err := blake2b.New(bytes, nil)
	if err != nil {
		panic(err)
	}

	_, err = hasher.Write(data)
	if err != nil {
		panic(err)
	}

	digestBytes := hasher.Sum(nil)
	copy(digest[:], digestBytes)
	return digest
}
```

and then use it here using

```suggestion
	encodedTaggedTransactionQueue = common.MustBlake2b8([]byte("TaggedTransactionQueue"))
```

lib/runtime/version.go
```diff
@@ -30,9 +30,24 @@ type Version struct {
...
 	ErrDecodingVersionField = errors.New("decoding version field")
 )
 
+// TaggedTransactionQueueVersion returns the TaggedTransactionQueueAPI version
+func (*Version) TaggedTransactionQueueVersion(runtimeVersion Version) uint32 {
+	txQueueVersion := uint32(0)
+	for _, v := range runtimeVersion.APIItems {
+		if v.Name == encodedTaggedTransactionQueue {
+			txQueueVersion = v.Ver
+			break
```
qdm12, line 45:
```suggestion
			return v.Ver
```

lib/runtime/version.go
```diff
@@ -30,9 +30,24 @@ type Version struct {
...
+		}
+	}
+	return txQueueVersion
```
qdm12, line 48: nit just to be explicit the default is `0` if not found.

```suggestion
	return 0
```

lib/blocktree/mocks_generate_test.go
```diff
@@ -0,0 +1,6 @@
+// Copyright 2022 ChainSafe Systems (ON)
+// SPDX-License-Identifier: LGPL-3.0-only
+
+package blocktree
+
+//go:generate mockgen -destination=mock_instance_test.go -package $GOPACKAGE github.com/ChainSafe/gossamer/lib/runtime Instance
```
qdm12, line 6: generate to mocks_test.go for consistency with other packages (also don't forget to re-`go generate`)

```suggestion
//go:generate mockgen -destination=mocks_test.go -package $GOPACKAGE github.com/ChainSafe/gossamer/lib/runtime Instance
```

lib/runtime/version.go
```diff
@@ -30,9 +30,24 @@ type Version struct {
...
 
 var (
+	// Blake2-8("TaggedTransactionQueue")
+	encodedTaggedTransactionQueue = [8]byte{0xd2, 0xbc, 0x98, 0x97, 0xee, 0xd0, 0x8f, 0x15}
+
 	ErrDecodingVersionField = errors.New("decoding version field")
 )
 
+// TaggedTransactionQueueVersion returns the TaggedTransactionQueueAPI version
+func (*Version) TaggedTransactionQueueVersion(runtimeVersion Version) uint32 {
```
qdm12, line 40: Why is this a method on Version and taking a Version argument? Remove the version argument and leave it as a method I'd say?

qdm12, line 40: What happens if the version returned is not supported by Gossamer?
You might need to handle it when the wasmer instance sets the Version field (`NewInstance` and `HandleRuntimeChanges` I think), and return an error.

dot/rpc/modules/author_integration_test.go
```diff
@@ -511,19 +556,11 @@ func TestAuthorModule_SubmitExtrinsic_WithVersion_V0910(t *testing.T) {
...
-	extHex := runtime.NewTestExtrinsic(t,
-		integrationTestController.runtime, genesisHash, genesisHash, 1, "System.remark", []byte{0xab, 0xcd})
-
-	// to extrinsic works with a runtime version 0910 we need to
-	// append the block hash bytes at the end of the extrinsics
-	hashBytes := genesisHash.ToBytes()
-	extBytes := append(common.MustHexToBytes(extHex), hashBytes...)
-
-	extHex = common.BytesToHex(extBytes)
+	ext := createExtrinsic(t, integrationTestController.runtime, genesisHash, 0)
```
qdm12, line 559: nit `ext` -> `extrinsic` 👏 🎉

dot/rpc/modules/author_integration_test.go
```diff
@@ -79,6 +83,47 @@ func useInstanceFromRuntimeV0910(t *testing.T, rtStorage *storage.TrieState) (in
...
+	var decoded []byte
+	err = scale.Unmarshal(rawMeta, &decoded)
+	require.NoError(t, err)
+
+	meta := &ctypes.Metadata{}
+
+	err = ctypes.Decode(decoded, meta)
+	require.NoError(t, err)
+
+	rv := rt.Version()
```
qdm12, line 100: nit `rv` -> `runtimeVersion`

dot/rpc/modules/author_integration_test.go
```diff
@@ -79,6 +83,47 @@ func useInstanceFromRuntimeV0910(t *testing.T, rtStorage *storage.TrieState) (in
...
+
+	c, err := ctypes.NewCall(meta, "System.remark", []byte{0xab, 0xcd})
```
qdm12, line 102: nit `c` -> `metaCall`

dot/rpc/modules/author_integration_test.go
```diff
@@ -79,6 +83,47 @@ func useInstanceFromRuntimeV0910(t *testing.T, rtStorage *storage.TrieState) (in
...
+	require.NoError(t, err)
+
+	ext := ctypes.NewExtrinsic(c)
```
qdm12, line 105: nit `ext` -> `extrinsic`

dot/rpc/modules/author_integration_test.go
```diff
@@ -79,6 +83,47 @@ func useInstanceFromRuntimeV0910(t *testing.T, rtStorage *storage.TrieState) (in
...
+	}
+
+	// Sign the transaction using Alice's key
+	err = ext.Sign(signature.TestKeyringPairAlice, options)
+	require.NoError(t, err)
+
+	extEnc, err := ctypes.EncodeToHex(ext)
+	require.NoError(t, err)
+
+	extBytes := types.Extrinsic(common.MustHexToBytes(extEnc))
```
qdm12, line 123: nit can't we do this in a single step instead of ext -> hex -> bytes

dot/rpc/modules/author_integration_test.go
```diff
@@ -79,6 +83,47 @@ func useInstanceFromRuntimeV0910(t *testing.T, rtStorage *storage.TrieState) (in
...
+	return extBytes
```
qdm12, line 124: nit  rename to `extrinsicBytes`

dot/core/helpers.go
```diff
@@ -0,0 +1,16 @@
...
+	length := 0
+	for i := range slices {
+		length += len(slices[i])
+	}
+	concatenated = make([]byte, 0, length)
+	for _, slice := range slices {
+		concatenated = append(concatenated, slice...)
+	}
+	return concatenated
+}
```
EclesioMeloJunior, line 16: would suggest to use [bytes.Join](https://pkg.go.dev/bytes#Join)
```go
s := [][]byte{[]byte("foo"), []byte("bar"), []byte("baz")}
bytes.Join(s, nil)
```

```suggestion
```

dot/core/helpers_test.go
```diff
@@ -17,10 +19,168 @@ import (
...
+	t.Helper()
+
+	h0, err := common.Twox128Hash([]byte("System"))
+	require.NoError(t, err)
+	h1, err := common.Twox128Hash([]byte("Account"))
+	require.NoError(t, err)
+	h2, err := common.Blake2b128(pub)
+	require.NoError(t, err)
+	bytes := [][]byte{h0, h1, h2, pub}
+	return concatenateByteSlices(bytes)
```
EclesioMeloJunior, line 37:
```suggestion
	return bytes.Join(bytes, nil)
```

dot/core/service.go
```diff
@@ -551,3 +579,27 @@ func (s *Service) GetReadProofAt(block common.Hash, keys [][]byte) (
...
+	runtimeVersion := rt.Version()
+	txQueueVersion := runtimeVersion.TaggedTransactionQueueVersion(runtimeVersion)
+	var externalExt types.Extrinsic
+	switch txQueueVersion {
+	case supportedTxVersion:
+		externalExt = types.Extrinsic(concatenateByteSlices([][]byte{
+			{byte(types.TxnExternal)},
+			ext,
+			s.blockState.BestBlockHash().ToBytes(),
+		}))
```
EclesioMeloJunior, line 595:
```suggestion
extrinsics := [][]byte{{byte(types.TxnExternal)}, ext, s.blockState.BestBlockHash().ToBytes()}
externalExt = types.Extrinsic(bytes.Join(extrinsics, nil))
```

dot/core/service.go
```diff
@@ -551,3 +579,27 @@ func (s *Service) GetReadProofAt(block common.Hash, keys [][]byte) (
...
+	case previousSupportedTxVersion:
+		externalExt = types.Extrinsic(concatenateByteSlices([][]byte{
+			{byte(types.TxnExternal)},
+			ext,
+		}))
```
EclesioMeloJunior, line 600:
```suggestion 
extrinsics := [][]byte{{byte(types.TxnExternal)}, ext}
externalExt = types.Extrinsic(bytes.Join(extrinsics, nil))
```

EclesioMeloJunior (changes requested):

dot/core/helpers.go
qdm12: Nice I didn't know about that one!! 👍

dot/core/service.go
qdm12: nit it's not really `extrinsics`, it's more of a `extrinsicParts` I guess

dot/core/service.go
qdm12: nit it's not really `extrinsics`, it's more of a `extrinsicParts` I guess

dot/core/service.go
jimjbrettj: I think I would agree. I will hold up on changing till I get more eyes on it

dot/core/service.go
jimjbrettj: seems to me no. Since we have to check every extrinsic, doesn't make sense to me that one invalid one should halt that process. Feels like attack vector. Maybe there is some other appropriate way to handle though.

dot/core/service.go
jimjbrettj: hmm this one maybe we should. Will ask for @timwu20 or @noot thoughts

dot/core/service.go
jimjbrettj: Also Im thinking so here. Will still wait for more feedback to double check

dot/core/service.go
jimjbrettj: I mean I think we do, see my explanation for why in an above comment. Let me know if my logic seems flawed though

lib/runtime/version.go
jimjbrettj: good call

lib/runtime/version.go
jimjbrettj: Really like the idea but didn't seem to be working for me. I feel like I remember its actually an encoding of the whole function signature, but I might be wrong. Perhaps @danforbes would know

lib/runtime/version.go
jimjbrettj: Very good point. I initially built this when we supported the legacy version as well, so had receivers for both. Just forgot to remove

jimjbrettj: We currently handle it in transaction construction, where an error is returned if this happens (for both new and legacy unsupported versions)

jimjbrettj: I'm not sure why we would globally need to track supported versions. Unless I'm missing something I feel like we can just check version where we actually construct transactions

dot/rpc/modules/author_integration_test.go
jimjbrettj: why metaCall and not call?

dot/rpc/modules/author_integration_test.go
jimjbrettj: updated to just do hex to bytes

dot/core/helpers.go
jimjbrettj: Nice!

lib/runtime/version.go
danforbes: The API ID is the first 8 bytes of the Blake2b hash of the API name. Ref https://github.com/paritytech/substrate/blob/polkadot-v0.9.29/primitives/api/proc-macro/src/decl_runtime_apis.rs#L553-L562

dot/core/service.go
timwu20: I think this is right.  We are trying to compare the best block to the incoming block we receive over the channel.  The channel is produced by `handleBlock()` via `HandleBlockImport` and `HandleBlockProduced`.  So we may receive blocks from other authors that may have forked, and that's what we're trying to detect when we call `handleChainReorg` when we receive a block over the channel.

I think what's wrong is that we do `prev := s.blockState.BestBlockHash()` but `prev` isn't used until line 288.  This method will block on like 278 until either the context is cancelled or a block comes in through `s.blockAddCh`.  So there's a probability we're using an outdated best block.

dot/core/service.go
timwu20: I agree, we return an error for `handleChainReorg` so I'm not sure why this function doesn't

timwu20: I don't think we should be continuing with the rest of this function if there is an error trying to get the `stateRoot`.

dot/core/service.go
timwu20: Same here, how can we continue if we don't have the trie state?

dot/core/service.go
timwu20: Shouldn't we `continue` to the next transaction? Why are we passing it to the runtime to validate if we can't build it?

timwu20 (changes requested):

lib/runtime/version.go
qdm12: As in, the transaction queue version given by the runtime to Gossamer has to be supported by Gossamer (like the v1 state trie) right? If so we should enforce to error out if the runtime gives a version not implemented yet.

jimjbrettj:
@timwu20 https://github.com/ChainSafe/gossamer/pull/2809#discussion_r983858307 so you think good to move this call to just above handleChainReorg? Just wanted to confirm before changing since will need to update tests

jimjbrettj:
@timwu20 https://github.com/ChainSafe/gossamer/pull/2809#discussion_r983864001 so should I change the function signature to return an error?

dot/core/messages.go
```diff
@@ -28,7 +28,11 @@ func (s *Service) validateTransaction(head *types.Header, rt RuntimeInstance,
 	rt.SetContextStorage(ts)
 
 	// validate each transaction
-	externalExt := types.Extrinsic(append([]byte{byte(types.TxnExternal)}, tx...))
+	externalExt, err := s.buildExternalTransaction(rt, tx)
+	if err != nil {
+		return nil, fmt.Errorf("building external transaction: %s", err)
```
EclesioMeloJunior, line 33:
```suggestion
		return nil, fmt.Errorf("building external transaction: %w", err)
```

dot/core/messages.go
```diff
@@ -70,32 +74,34 @@ func (s *Service) HandleTransactionMessage(peerID peer.ID, msg *network.Transact
...
+	allTxsAreValid := true
 	for _, tx := range txs {
-		txnIsValid := true
+		isValidTxn := true
 		validity, err := s.validateTransaction(head, rt, tx)
 		if err != nil {
-			txnIsValid = false
-			allTxnsAreValid = false
-			switch err := err.(type) {
+			switch err.(type) {
```
EclesioMeloJunior, line 82: Whenever `err != nil` you should update the `allTxsAreValid` to false
```suggestion
			allTxsAreValid = false
			switch err.(type) {
```

dot/core/messages.go
```diff
@@ -70,32 +74,34 @@ func (s *Service) HandleTransactionMessage(peerID peer.ID, msg *network.Transact
...
 			case runtime.InvalidTransaction:
+				isValidTxn = false
+				allTxsAreValid = false
```
EclesioMeloJunior, line 85: I would suggest using `continue` right after reporting the peer 
```suggestion
				s.net.ReportPeer(peerset.ReputationChange{
					Value:  peerset.BadTransactionValue,
					Reason: peerset.BadTransactionReason,
				}, peerID)
				continue
```

dot/core/messages.go
```diff
@@ -70,32 +74,34 @@ func (s *Service) HandleTransactionMessage(peerID peer.ID, msg *network.Transact
...
 				s.net.ReportPeer(peerset.ReputationChange{
 					Value:  peerset.BadTransactionValue,
 					Reason: peerset.BadTransactionReason,
 				}, peerID)
 			case runtime.UnknownTransaction:
+				isValidTxn = false
+				allTxsAreValid = false
```
EclesioMeloJunior, line 92:
```suggestion
				continue
```

dot/core/messages.go
```diff
@@ -70,32 +74,34 @@ func (s *Service) HandleTransactionMessage(peerID peer.ID, msg *network.Transact
...
 			default:
-				return false, fmt.Errorf("validating transaction from peerID %s: %w", peerID, err)
+				return false, fmt.Errorf("failed validating transaction for peerID %s: %w", peerID, err)
 			}
 		}
 
-		if txnIsValid && validity.Propagate {
+		if isValidTxn && validity.Propagate {
```
EclesioMeloJunior, line 98: using the `continue` in the following cases: `runtime.InvalidTransaction`, `runtime.UnknownTransaction` you don't need to have `isValidTxn` because you will jump to the next iteration item without passing through this if statement

```suggestion
		if validity.Propagate {
```

dot/core/service.go
```diff
@@ -282,11 +280,14 @@ func (s *Service) handleBlocksAsync() {
 				continue
 			}
 
+			prev := s.blockState.BestBlockHash()
```
EclesioMeloJunior, line 283: nit: would suggest renaming from `prev` to `bestBlockHash` since `prev` is not a clear name

dot/core/service.go
```diff
@@ -370,23 +373,42 @@ func (s *Service) handleChainReorg(prev, curr common.Hash) error {
...
 // them to the queue if valid.
 // See https://github.com/paritytech/substrate/blob/74804b5649eccfb83c90aec87bdca58e5d5c8789/client/transaction-pool/src/lib.rs#L545
-func (s *Service) maintainTransactionPool(block *types.Block) {
+func (s *Service) maintainTransactionPool(block *types.Block) error {
 	// remove extrinsics included in a block
 	for _, ext := range block.Body {
 		s.transactionState.RemoveExtrinsic(ext)
 	}
 
+	bestBlockHash := s.blockState.BestBlockHash()
```
EclesioMeloJunior, line 382: instead of retrieving the best block hash inside this function, would be nice to pass it as an argument. I saw that in the caller context the best block hash is already known (`dot/core/service.go` line `283`)

EclesioMeloJunior (changes requested):

dot/core/service.go
```diff
@@ -282,11 +280,14 @@ func (s *Service) handleBlocksAsync() {
 				continue
 			}
 
+			prev := s.blockState.BestBlockHash()
 			if err := s.handleChainReorg(prev, block.Header.Hash()); err != nil {
 				logger.Warnf("failed to re-add transactions to chain upon re-org: %s", err)
```
timwu20, line 285: I wonder if we ever get to this case.  Maybe it would be best to panic until we figure out reporting errors from `handleBlocksAsync`.

dot/core/service.go
```diff
@@ -282,11 +280,14 @@ func (s *Service) handleBlocksAsync() {
...
 			}
 
-			s.maintainTransactionPool(block)
+			if err := s.maintainTransactionPool(block); err != nil {
+				logger.Warnf("failed to maintain txn pool after re-org: %s", err)
```
timwu20, line 289: I wonder if we ever get to this case.  Maybe it would be best to panic until we figure out reporting errors from `handleBlocksAsync`.

dot/core/service.go
```diff
@@ -350,15 +351,17 @@ func (s *Service) handleChainReorg(prev, curr common.Hash) error {
 				continue
 			}
 
-			externalExt := make(types.Extrinsic, 0, 1+len(ext))
-			externalExt = append(externalExt, byte(types.TxnExternal))
-			externalExt = append(externalExt, ext...)
-			txv, err := rt.ValidateTransaction(externalExt)
+			externalExt, err := s.buildExternalTransaction(rt, ext)
+			if err != nil {
+				return fmt.Errorf("building external transaction: %w", err)
```
timwu20, line 356: why not `continue`? This is essentially stopping the chain reorg.

dot/core/messages.go
jimjbrettj: This does already happen, but I think this way if cleaner

dot/core/messages.go
jimjbrettj: Nice!

dot/core/service.go
jimjbrettj: really good catch

dot/core/interface.go
```diff
@@ -88,6 +88,7 @@ type StorageState interface {
 
 // TransactionState is the interface for transaction state methods
 type TransactionState interface {
+	Pop() *transaction.ValidTransaction
```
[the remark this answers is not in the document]
jimjbrettj: I'm a bit confused, wouldn't the type assertion still be on the transaction state tho? Are you saying have a receiver on that the transaction state in the test?

lib/runtime/version.go
```diff
@@ -30,9 +30,22 @@ type Version struct {
...
 
 var (
+	// Blake2-8("TaggedTransactionQueue")
+	encodedTaggedTransactionQueue = [8]byte{0xd2, 0xbc, 0x98, 0x97, 0xee, 0xd0, 0x8f, 0x15}
+
 	ErrDecodingVersionField = errors.New("decoding version field")
 )
 
+// TaggedTransactionQueueVersion returns the TaggedTransactionQueueAPI version
+func TaggedTransactionQueueVersion(runtimeVersion Version) (txQueueVersion uint32) {
```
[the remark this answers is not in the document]
jimjbrettj: This was the case before, but was changed/addressed in an above comment

go.mod
```diff
@@ -12,9 +12,9 @@ require (
 	github.com/dgraph-io/ristretto v0.0.3-0.20200630154024-f66de99634de
 	github.com/disiqueira/gotree v1.0.0
 	github.com/docker/docker v20.10.18+incompatible
-	github.com/ethereum/go-ethereum v1.10.25
+	github.com/ethereum/go-ethereum v1.10.23
 	github.com/fatih/color v1.13.0
-	github.com/go-playground/validator/v10 v10.11.1
+	github.com/go-playground/validator/v10 v10.11.0
```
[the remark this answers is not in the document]
jimjbrettj: hm yeah not sure what changed this, I didn't do it manually. Will fix

lib/runtime/version.go
jimjbrettj: Nice!

jimjbrettj:
Updated!

jimjbrettj:
It seems adding panics to handleBlockAsync has caused things to break. WIll further investigate

jimjbrettj:
@timwu20 I think the core tests should be fixed now. Elizabeth cautioned against having these added panics, but maybe good for testing still? Your call

dot/core/interface.go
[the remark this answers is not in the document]
qdm12: In the test just do `service.transactionState.(*state.TransactionState).Pop()` (from the top of my head), and you can remove that `Pop` method from the tx state interface (which should be focused on prod code only)

lib/runtime/version.go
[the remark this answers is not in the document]
qdm12: You mean https://github.com/ChainSafe/gossamer/pull/2809#discussion_r982718248 right? If so, we could change it to a method which seems cleaner/pollutes less the global package Go API, as:

```suggestion
// TaggedTransactionQueueVersion returns the TaggedTransactionQueueAPI version
func (v *Version) TaggedTransactionQueueVersion() (txQueueVersion uint32) {
```

dot/core/service.go
```diff
@@ -350,15 +352,18 @@ func (s *Service) handleChainReorg(prev, curr common.Hash) error {
 				continue
 			}
 
-			externalExt := make(types.Extrinsic, 0, 1+len(ext))
-			externalExt = append(externalExt, byte(types.TxnExternal))
-			externalExt = append(externalExt, ext...)
-			txv, err := rt.ValidateTransaction(externalExt)
+			externalExt, err := s.buildExternalTransaction(rt, ext)
+			if err != nil {
+				logger.Debugf("building external transaction: %w", err)
```
qdm12, line 357: Shouldn't this be at least at the warning level? :thinking: 

Or does it happen a lot?

dot/core/service.go
```diff
@@ -370,23 +375,42 @@ func (s *Service) handleChainReorg(prev, curr common.Hash) error {
...
 // them to the queue if valid.
 // See https://github.com/paritytech/substrate/blob/74804b5649eccfb83c90aec87bdca58e5d5c8789/client/transaction-pool/src/lib.rs#L545
-func (s *Service) maintainTransactionPool(block *types.Block) {
+func (s *Service) maintainTransactionPool(block *types.Block, bestBlockHash common.Hash) error {
 	// remove extrinsics included in a block
 	for _, ext := range block.Body {
 		s.transactionState.RemoveExtrinsic(ext)
 	}
 
+	fmt.Println(bestBlockHash)
```
qdm12, line 384:
```suggestion
```

dot/core/service.go
```diff
@@ -555,3 +584,22 @@ func (s *Service) GetReadProofAt(block common.Hash, keys [][]byte) (
 
 	return block, proofForKeys, nil
 }
+
+// buildExternalTransaction builds an external transaction based on the current TransactionQueueAPIVersion
+// See https://github.com/paritytech/substrate/blob/polkadot-v0.9.25/primitives/transaction-pool/src/runtime_api.rs#L25-L55
+func (s *Service) buildExternalTransaction(rt runtime.Instance, ext types.Extrinsic) (types.Extrinsic, error) {
+	runtimeVersion := rt.Version()
+	txQueueVersion := runtime.TaggedTransactionQueueVersion(runtimeVersion)
+	var externalExt types.Extrinsic
```
qdm12, line 593: nit

```suggestion
func (s *Service) buildExternalTransaction(rt runtime.Instance, ext types.Extrinsic) (
	externalExt types.Extrinsic, err error) {
	runtimeVersion := rt.Version()
	txQueueVersion := runtime.TaggedTransactionQueueVersion(runtimeVersion)
```

dot/core/service.go
```diff
@@ -555,3 +584,22 @@ func (s *Service) GetReadProofAt(block common.Hash, keys [][]byte) (
...
+	switch txQueueVersion {
+	case 3:
+		extrinsicParts := [][]byte{{byte(types.TxnExternal)}, ext, s.blockState.BestBlockHash().ToBytes()}
+		externalExt = types.Extrinsic(bytes.Join(extrinsicParts, nil))
+	case 2:
+		extrinsicParts := [][]byte{{byte(types.TxnExternal)}, ext}
+		externalExt = types.Extrinsic(bytes.Join(extrinsicParts, nil))
```
qdm12, line 600: nit you could take `externalExt` declaration + setting after the switch since it's common to all code branches

dot/core/service_integration_test.go
```diff
@@ -128,11 +60,16 @@ func TestAnnounceBlock(t *testing.T) {
 	err = digest.Add(*prd)
 	require.NoError(t, err)
 
+	// Used to define the state root of new block for testing
+	genesisBlock, err := s.blockState.GetBlockByHash(s.blockState.GenesisHash())
```
qdm12, line 64: nit Can we call something more specific to set the state root of our test block? Or not, just a quick question

dot/core/service_integration_test.go
```diff
@@ -481,39 +409,32 @@ func TestMaintainTransactionPool_BlockWithExtrinsics(t *testing.T) {
...
-	extrinsicBytes, _ := generateTestValidRemarkTxns(t, alicePub, accountInfo)
+	genesisFilePath, err := utils.GetGssmrGenesisRawPath()
+	require.NoError(t, err)
 
 	ctrl := gomock.NewController(t)
-	telemetryMock := NewMockClient(ctrl)
-	telemetryMock.EXPECT().SendMessage(gomock.Any()).AnyTimes()
-
-	ts := state.NewTransactionState(telemetryMock)
+	service, encExt := createTestService(t, genesisFilePath, alicePub, accountInfo, ctrl)
```
qdm12, line 416: nit `encExt` -> `encodedExtrinsic`

dot/core/service_test.go
```diff
@@ -555,22 +560,49 @@ func Test_Service_maintainTransactionPool(t *testing.T) {
 			Propagate: true,
 		}
 
-		extrinsic := types.Extrinsic{21}
-		vt := transaction.NewValidTransaction(extrinsic, validity)
+		ext := types.Extrinsic{21}
```
qdm12, line 563: nit what's wrong with previous `extrinsic`? :cry:

lib/common/hasher_test.go
```diff
@@ -33,6 +33,12 @@ func TestBlake128(t *testing.T) {
 	require.Equal(t, expected, h)
 }
 
+func TestBlake2b8(t *testing.T) {
+	exp := [8]byte{0xd2, 0xbc, 0x98, 0x97, 0xee, 0xd0, 0x8f, 0x15}
+	res := common.MustBlake2b8([]byte("TaggedTransactionQueue"))
```
qdm12, line 38: nit side note, having a `common_test` package in the common directory is just weird AF :smile:

dot/core/service_test.go
```diff
@@ -555,22 +560,49 @@ func Test_Service_maintainTransactionPool(t *testing.T) {
...
+		mockStorageState := NewMockStorageState(ctrl)
+		mockStorageState.EXPECT().TrieState(&common.Hash{}).Return(&rtstorage.TrieState{}, nil)
+		mockStorageState.EXPECT().GetStateRootFromBlock(&common.Hash{}).Return(&common.Hash{}, nil)
 		service := &Service{
 			transactionState: mockTxnState,
 			blockState:       mockBlockState,
+			storageState:     mockStorageState,
 		}
-		service.maintainTransactionPool(&block)
+		err := service.maintainTransactionPool(&block, common.Hash{})
```
qdm12, line 604: nit you should pass in a non-zero value i.e. `common.Hash{1}` to make sure it gets passed around correctly (without reset/empty value passed in)

dot/core/service_test.go
```diff
@@ -589,41 +621,63 @@ func Test_Service_maintainTransactionPool(t *testing.T) {
...
+		mockStorageState := NewMockStorageState(ctrl)
+		mockStorageState.EXPECT().TrieState(&common.Hash{}).Return(&rtstorage.TrieState{}, nil)
+		mockStorageState.EXPECT().GetStateRootFromBlock(&common.Hash{}).Return(&common.Hash{}, nil)
 		service := &Service{
 			transactionState: mockTxnState,
 			blockState:       mockBlockStateOk,
+			storageState:     mockStorageState,
 		}
-		service.maintainTransactionPool(&block)
+		err := service.maintainTransactionPool(&block, common.Hash{})
```
qdm12, line 668: nit you should pass in a non-zero value i.e. `common.Hash{1}` to make sure it gets passed around correctly (without reset/empty value passed in)

dot/core/service_test.go
```diff
@@ -665,46 +711,30 @@ func Test_Service_handleBlocksAsync(t *testing.T) {
...
-			blockAddCh:       blockAddChan,
-			ctx:              context.Background(),
+			blockState: mockBlockState,
+			blockAddCh: blockAddChan,
+			ctx:        context.Background(),
 		}
-		service.handleBlocksAsync()
+
+		assert.PanicsWithError(t, "failed to re-add transactions to chain upon re-org: test dummy error",
+			service.handleBlocksAsync, "The code panics as expected")
```
qdm12, line 737: nit unneeded message

```suggestion
			service.handleBlocksAsync)
```

dot/core/service_test.go
```diff
@@ -555,22 +560,49 @@ func Test_Service_maintainTransactionPool(t *testing.T) {
...
-		runtimeMock.EXPECT().ValidateTransaction(types.Extrinsic{21}).Return(nil, errTestDummyError)
+		runtimeMock.EXPECT().ValidateTransaction(externalExt).Return(nil, errTestDummyError)
+		runtimeMock.EXPECT().Version().Return(runtime.Version{
+			SpecName:         []byte("polkadot"),
+			ImplName:         []byte("parity-polkadot"),
+			AuthoringVersion: authoringVersion,
+			SpecVersion:      specVersion,
+			ImplVersion:      implVersion,
+			APIItems: []runtime.APIItem{{
+				Name: [8]byte{0xd2, 0xbc, 0x98, 0x97, 0xee, 0xd0, 0x8f, 0x15},
```
qdm12, line 581: nit

```suggestion
				Name: common.MustBlake2b8([]byte("TaggedTransactionQueue")),
```

dot/core/service_test.go
```diff
@@ -589,41 +621,63 @@ func Test_Service_maintainTransactionPool(t *testing.T) {
...
-			Return(&transaction.Validity{Propagate: true}, nil)
+		runtimeMock.EXPECT().ValidateTransaction(externalExt).Return(&transaction.Validity{Propagate: true}, nil)
+		runtimeMock.EXPECT().Version().Return(runtime.Version{
+			SpecName:         []byte("polkadot"),
+			ImplName:         []byte("parity-polkadot"),
+			AuthoringVersion: authoringVersion,
+			SpecVersion:      specVersion,
+			ImplVersion:      implVersion,
+			APIItems: []runtime.APIItem{{
+				Name: [8]byte{0xd2, 0xbc, 0x98, 0x97, 0xee, 0xd0, 0x8f, 0x15},
```
qdm12, line 643:
```suggestion
				Name: common.MustBlake2b8([]byte("TaggedTransactionQueue")),
```

dot/core/service_test.go
```diff
@@ -813,15 +843,31 @@ func TestService_handleChainReorg(t *testing.T) {
...
 		runtimeMockErr := NewMockRuntimeInstance(ctrl)
+		runtimeMockErr.EXPECT().ValidateTransaction(externExt).Return(nil, errTestDummyError)
+		runtimeMockErr.EXPECT().Version().Return(runtime.Version{
+			SpecName:         []byte("polkadot"),
+			ImplName:         []byte("parity-polkadot"),
+			AuthoringVersion: authoringVersion,
+			SpecVersion:      specVersion,
+			ImplVersion:      implVersion,
+			APIItems: []runtime.APIItem{{
+				Name: [8]byte{0xd2, 0xbc, 0x98, 0x97, 0xee, 0xd0, 0x8f, 0x15},
```
qdm12, line 856:
```suggestion
				Name: common.MustBlake2b8([]byte("TaggedTransactionQueue")),
```

dot/core/service_test.go
```diff
@@ -833,14 +879,29 @@ func TestService_handleChainReorg(t *testing.T) {
...
 		runtimeMockOk := NewMockRuntimeInstance(ctrl)
+		runtimeMockOk.EXPECT().ValidateTransaction(externExt).Return(testValidity, nil)
+		runtimeMockOk.EXPECT().Version().Return(runtime.Version{
+			SpecName:         []byte("polkadot"),
+			ImplName:         []byte("parity-polkadot"),
+			AuthoringVersion: authoringVersion,
+			SpecVersion:      specVersion,
+			ImplVersion:      implVersion,
+			APIItems: []runtime.APIItem{{
+				Name: [8]byte{0xd2, 0xbc, 0x98, 0x97, 0xee, 0xd0, 0x8f, 0x15},
```
qdm12, line 890:
```suggestion
				Name: common.MustBlake2b8([]byte("TaggedTransactionQueue")),
```

dot/core/service_test.go
```diff
@@ -1165,8 +1231,21 @@ func TestServiceHandleSubmittedExtrinsic(t *testing.T) {
...
-		runtimeMockErr.EXPECT().SetContextStorage(&rtstorage.TrieState{})
 		runtimeMockErr.EXPECT().ValidateTransaction(externalExt).Return(nil, errDummyErr)
+		runtimeMockErr.EXPECT().Version().Return(runtime.Version{
+			SpecName:         []byte("polkadot"),
+			ImplName:         []byte("parity-polkadot"),
+			AuthoringVersion: authoringVersion,
+			SpecVersion:      specVersion,
+			ImplVersion:      implVersion,
+			APIItems: []runtime.APIItem{{
+				Name: [8]byte{0xd2, 0xbc, 0x98, 0x97, 0xee, 0xd0, 0x8f, 0x15},
```
qdm12, line 1242:
```suggestion
				Name: common.MustBlake2b8([]byte("TaggedTransactionQueue")),
```

dot/core/service_test.go
```diff
@@ -1184,8 +1263,23 @@ func TestServiceHandleSubmittedExtrinsic(t *testing.T) {
...
+
 		runtimeMock.EXPECT().ValidateTransaction(externalExt).Return(&transaction.Validity{Propagate: true}, nil)
+		runtimeMock.EXPECT().Version().Return(runtime.Version{
+			SpecName:         []byte("polkadot"),
+			ImplName:         []byte("parity-polkadot"),
+			AuthoringVersion: authoringVersion,
+			SpecVersion:      specVersion,
+			ImplVersion:      implVersion,
+			APIItems: []runtime.APIItem{{
+				Name: [8]byte{0xd2, 0xbc, 0x98, 0x97, 0xee, 0xd0, 0x8f, 0x15},
```
qdm12, line 1276:
```suggestion
				Name: common.MustBlake2b8([]byte("TaggedTransactionQueue")),
```

dot/core/service.go
jimjbrettj: @timwu20 I talked to Elizabeth and she brought up a good point, why do we not use the block returned in the channel? I feel like we should use that instead of getting best block. Thoughts?

dot/core/service.go
jimjbrettj: yes i do! sweet sweet

lib/runtime/version.go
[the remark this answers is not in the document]
jimjbrettj: oh yes I did the reverse, will update

dot/core/service.go
jimjbrettj: Will leave panics and am updating maintaintxnPool to return more errors when appropriate

dot/core/service.go
jimjbrettj: Will leave panics and am updating maintaintxnPool to return more errors when appropriate

dot/core/service.go
```diff
@@ -282,11 +281,14 @@ func (s *Service) handleBlocksAsync() {
 				continue
 			}
 
-			if err := s.handleChainReorg(prev, block.Header.Hash()); err != nil {
-				logger.Warnf("failed to re-add transactions to chain upon re-org: %s", err)
+			bestBlockHash := s.blockState.BestBlockHash()
+			if err := s.handleChainReorg(bestBlockHash, block.Header.Hash()); err != nil {
+				panic(fmt.Errorf("failed to re-add transactions to chain upon re-org: %s", err))
```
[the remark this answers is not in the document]
jimjbrettj: This broke test so leaving as is

dot/core/service.go
jimjbrettj: You are right, but functions were now modified to return error here

dot/core/service_integration_test.go
jimjbrettj: I dont think so. Looked around and didn't see anything that jumped out at me

lib/common/hasher.go
```diff
@@ -27,6 +27,24 @@ func Blake2b128(in []byte) ([]byte, error) {
 	return h.Sum(nil), nil
 }
 
+// MustBlake2b8 returns the first 8 bytes of the Blake2b hash of the input data
+func MustBlake2b8(data []byte) (digest [8]byte) {
```
kishansagathiya, line 31: Can we make this `Blake2b8` and return errors instead of panicking?

dot/core/messages.go
```diff
@@ -70,32 +74,32 @@ func (s *Service) HandleTransactionMessage(peerID peer.ID, msg *network.Transact
...
+			allTxsAreValid = false
+			switch err.(type) {
 			case runtime.InvalidTransaction:
 				s.net.ReportPeer(peerset.ReputationChange{
 					Value:  peerset.BadTransactionValue,
 					Reason: peerset.BadTransactionReason,
 				}, peerID)
+				continue
 			case runtime.UnknownTransaction:
+				continue
```
kishansagathiya, line 90: Instead of this 
- you can put a single continue statement after switch statement finishes
or
- use `else if validity.Propagate {` and you won't need to continue.

dot/core/service.go
```diff
@@ -555,3 +580,21 @@ func (s *Service) GetReadProofAt(block common.Hash, keys [][]byte) (
 
 	return block, proofForKeys, nil
 }
+
+// buildExternalTransaction builds an external transaction based on the current TransactionQueueAPIVersion
```
kishansagathiya, line 584: Generally comments and logs should use explanatory names so, `TransactionQueueAPIVersion` would become say `transaction queue API version`.

dot/core/service.go
```diff
@@ -555,3 +580,21 @@ func (s *Service) GetReadProofAt(block common.Hash, keys [][]byte) (
...
+	switch txQueueVersion {
+	case 3:
+		extrinsicParts = [][]byte{{byte(types.TxnExternal)}, ext, s.blockState.BestBlockHash().ToBytes()}
+	case 2:
+		extrinsicParts = [][]byte{{byte(types.TxnExternal)}, ext}
+	default:
+		return types.Extrinsic{}, fmt.Errorf("%w: %d", errInvalidTransactionQueueVersion, txQueueVersion)
+	}
+	externalTransaction := types.Extrinsic(bytes.Join(extrinsicParts, nil))
+	return externalTransaction, nil
```
kishansagathiya, line 599:
```suggestion
	return types.Extrinsic(bytes.Join(extrinsicParts, nil)), nil
```

dot/core/service.go
```diff
@@ -350,15 +352,16 @@ func (s *Service) handleChainReorg(prev, curr common.Hash) error {
...
 
-			externalExt := make(types.Extrinsic, 0, 1+len(ext))
-			externalExt = append(externalExt, byte(types.TxnExternal))
-			externalExt = append(externalExt, ext...)
-			txv, err := rt.ValidateTransaction(externalExt)
+			externalExt, err := s.buildExternalTransaction(rt, ext)
 			if err != nil {
-				logger.Debugf("failed to validate transaction for extrinsic %s: %s", ext, err)
-				continue
+				return fmt.Errorf("building external transaction: %s", err)
```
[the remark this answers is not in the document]
kishansagathiya: while you are at this please log the error for 
```
if err = decoder.Decode(&decExt); err != nil {
```
line 346 in this file

dot/core/messages.go
```diff
@@ -70,32 +74,32 @@ func (s *Service) HandleTransactionMessage(peerID peer.ID, msg *network.Transact
 		return false, err
 	}
 
-	allTxnsAreValid := true
+	allTxsAreValid := true
```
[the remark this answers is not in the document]
jimjbrettj: wasn't intentional, good catch. Will change.

kishansagathiya (commented):
- `lib/blocktree/mock_instance_test.go` seems like an appropriate name based on its contents, why has it been changed to `lib/blocktree/mocks_test.go`
- We want to shut down the node only in case of really critical errors. If the node can continue despite some error, it should.

dot/core/service.go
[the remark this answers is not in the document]
jimjbrettj: I totally agree. However, Tim and I talked and decided to put these there to surface errors that may come up in an obvious way. This is just for now though and will be replaced with more robust error handling in the future. Will tag @timwu20 just to make sure I am right here.

dot/core/service.go
```diff
@@ -282,11 +281,14 @@ func (s *Service) handleBlocksAsync() {
...
-			if err := s.handleChainReorg(prev, block.Header.Hash()); err != nil {
-				logger.Warnf("failed to re-add transactions to chain upon re-org: %s", err)
+			bestBlockHash := s.blockState.BestBlockHash()
+			if err := s.handleChainReorg(bestBlockHash, block.Header.Hash()); err != nil {
+				panic(fmt.Errorf("failed to re-add transactions to chain upon re-org: %s", err))
 			}
 
-			s.maintainTransactionPool(block)
+			if err := s.maintainTransactionPool(block, bestBlockHash); err != nil {
+				panic(fmt.Errorf("failed to maintain txn pool after re-org: %s", err))
```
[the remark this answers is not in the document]
jimjbrettj: See above response

dot/core/helpers_test.go
```diff
@@ -19,10 +21,135 @@ import (
...
+	t.Helper()
+
+	h0, err := common.Twox128Hash([]byte("System"))
+	require.NoError(t, err)
+	h1, err := common.Twox128Hash([]byte("Account"))
+	require.NoError(t, err)
+	h2, err := common.Blake2b128(pub)
+	require.NoError(t, err)
+	bytesToConcat := [][]byte{h0, h1, h2, pub}
+	return bytes.Join(bytesToConcat, nil)
```
kishansagathiya, line 39:
```suggestion
	return bytes.Join([][]byte{h0, h1, h2, pub}, nil)
```

dot/core/helpers_test.go
```diff
@@ -19,10 +21,135 @@ import (
...
+	err = cfgKeystore.Acco.Insert(kp)
+	require.NoError(t, err)
+
+	// Create state service
+	var stateSrvc *state.Service
+	testDatadirPath := t.TempDir()
+
+	// Set up block and storage state
+	telemetryMock := NewMockClient(ctrl)
+	telemetryMock.EXPECT().SendMessage(gomock.Any()).AnyTimes()
```
kishansagathiya, line 75: In SendMessage try to make the type `telemetry.Message`

dot/core/helpers_test.go
```diff
@@ -19,10 +21,135 @@ import (
...
+
+	// Start state service
+	err = stateSrvc.Start()
+	require.NoError(t, err)
+
+	cfgBlockState := stateSrvc.Block
+	cfgStorageState := stateSrvc.Storage
+	cfgCodeSubstitutedState := stateSrvc.Base
+
+	// Runtime stuff
```
kishansagathiya, line 97:
```suggestion
```

dot/core/service_integration_test.go
```diff
@@ -423,52 +358,45 @@ func TestMaintainTransactionPool_EmptyBlock(t *testing.T) {
...
-	service := NewTestService(t, cfg)
-	service.transactionState = transactionState
+	_ = service.transactionState.AddToPool(tx)
 
 	// provides is a list of transaction hashes that depend on this tx, see:
 	// https://github.com/paritytech/substrate/blob/5420de3face1349a97eb954ae71c5b0b940c31de/core/sr-primitives/src/transaction_validity.rs#L195
 	provides := common.MustHexToBytes("0xd43593c715fdd31c61141abd04a99fd6822c8558854ccde39a5684e7a56da27d00000000")
 	txnValidity := &transaction.Validity{
-		Priority:  39325240425794630,
+		Priority:  36074,
```
kishansagathiya, line 377: Why did we have to change this?

dot/core/service_integration_test.go
```diff
@@ -661,8 +579,7 @@ func TestService_HandleCodeSubstitutes(t *testing.T) {
 
 	s.blockState.StoreRuntime(blockHash, rt)
 
-	ts := rtstorage.NewTrieState(nil)
-
```
kishansagathiya, line 665: I am going to guess that removed lines have happened because of leftovers after debugging and not very deliberate. revert them since they provide readability.

kishansagathiya (changes requested):

jimjbrettj:
> * `lib/blocktree/mock_instance_test.go` seems like an appropriate name based on its contents, why has it been changed to `lib/blocktree/mocks_test.go`
> 
>     * We want to shut down the node only in case of really critical errors. If the node can continue despite some error, it should.

@kishansagathiya This does indeed make sense since we are dealing with the runtime. However, we are changing the mocks in the code in general to have a mocks_generate_tests file that creates a mocks_test mock file. So even tho that makes sense for this specific case I changed it to keep with the convention used elsewhere in the code

dot/core/messages_test.go
```diff
@@ -344,6 +356,9 @@ func TestServiceHandleTransactionMessage(t *testing.T) {
 						tt.mockBlockState.getRuntime.runtime,
 						tt.mockBlockState.getRuntime.err)
 				}
+				if tt.mockBlockState.callsBestBlockHash {
+					blockState.EXPECT().BestBlockHash().Return(common.Hash{})
```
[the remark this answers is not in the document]
jimjbrettj: I think we are tryin to avoid using `.AnyTimes() `

dot/core/service.go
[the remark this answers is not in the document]
jimjbrettj: Good call. Other places around here I think good to return error, bu for validating transactions I think this is appropriate. I also changed the code to remove ext from the pool if it is invalid. Let me know if you think this is a good idea here

dot/core/helpers_test.go
jimjbrettj: Having trouble with this. Can I pass in an interface like that? Or do I need to instantiate it and am just being dumb somehow when I am tring

dot/core/service_integration_test.go
jimjbrettj: You know I am not entirely sure why it changed (since the runtime sets it), but since our runtimes as well as the transaction being tested has changed I am not surprised that this needed to be changed. However, if you find this suspicious let me know

dot/core/helpers_test.go
kishansagathiya: Try something like 
```
	var k telemetry.Message
	telemetryMock.EXPECT().SendMessage(gomock.AssignableToTypeOf(k)).AnyTimes()
```

dot/core/service_integration_test.go
kishansagathiya: I didn't ask to change anything here. I just wanted you to add the empty line back.

dot/core/service.go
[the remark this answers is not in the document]
timwu20: Yes. can we add a TODO in here to address this in a future PR.

dot/core/service.go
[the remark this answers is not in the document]
timwu20: Yes. can we add a TODO in here to address this in a future PR.

dot/core/service.go
```diff
@@ -342,6 +344,7 @@ func (s *Service) handleChainReorg(prev, curr common.Hash) error {
 			decExt := &ctypes.Extrinsic{}
 			decoder := cscale.NewDecoder(bytes.NewReader(ext))
 			if err = decoder.Decode(&decExt); err != nil {
+				logger.Error(err.Error())
```
timwu20, line 347: this seems to be a returnable error for me.  How do we have a transaction in here that doesn't decode?

dot/core/service.go
```diff
@@ -350,15 +353,18 @@ func (s *Service) handleChainReorg(prev, curr common.Hash) error {
...
-			externalExt = append(externalExt, ext...)
-			txv, err := rt.ValidateTransaction(externalExt)
+			externalExt, err := s.buildExternalTransaction(rt, ext)
+			if err != nil {
+				return fmt.Errorf("building external transaction: %s", err)
+			}
+
+			transactionValidity, err := rt.ValidateTransaction(externalExt)
 			if err != nil {
 				logger.Debugf("failed to validate transaction for extrinsic %s: %s", ext, err)
```
timwu20, line 363:
```suggestion
				logger.Debugf("failed to validate transaction for extrinsic %s: %s skipping in chain reorg", ext, err)
```

timwu20 (approved):
Just the last feedback about the TODOs.  Great work!

timwu20 (approved):

EclesioMeloJunior (commented):
I believe you should rebase this PR with development

PR #2810, opened 2022-09-08
chore(dot/sync): `handleJustification` returns errors

qdm12:
## Changes

- Remove nil check on header and justification (unused, see caller checks - and caller should be the ones checking what they pass down)
- Return wrapped errors to callers
- Change callers to handle non-nil errors and return them up the call stack
- Adapt tests

## Tests

```sh
go test -tags integration github.com/ChainSafe/gossamer/dot/sync/...
```

## Primary Reviewer

@noot

### DIFF
--- a/dot/sync/chain_processor.go
+++ b/dot/sync/chain_processor.go
@@ -134,7 +134,10 @@ func (s *chainProcessor) processBlockData(bd *types.BlockData) error {
 
 		if bd.Justification != nil {
 			logger.Debugf("handling Justification for block number %d with hash %s...", block.Header.Number, bd.Hash)
-			s.handleJustification(&block.Header, *bd.Justification)
+			err = s.handleJustification(&block.Header, *bd.Justification)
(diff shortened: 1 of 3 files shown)


dot/sync/chain_processor.go
```diff
@@ -244,22 +250,22 @@ func (s *chainProcessor) handleBlock(block *types.Block) error {
 	return nil
 }
 
-func (s *chainProcessor) handleJustification(header *types.Header, justification []byte) {
-	if len(justification) == 0 || header == nil {
-		return
+func (s *chainProcessor) handleJustification(header *types.Header, justification []byte) (err error) {
+	if len(justification) == 0 {
+		return nil
 	}
```
qdm12, line 256: I don't think that can ever happen really, but keeping it to be safe ™️

EclesioMeloJunior (approved):

noot (approved):

PR #2811, opened 2022-09-08
chore(dot/sync): simplify `processBlockData` and associated tests

qdm12:
## Changes

- `processBlockData`:
  - Split subblocks into functions:
    - `processBlockDataWithHeaderAndBody` function with test
    - `processBlockDataWithStateHeaderAndBody` function with test
  - takes block data `blockData` as value argument instead of pointer
  - Review error wrappings
  - Remove block data nil check
  - Rename receiver from `s` to `c` for `chainProcessor`
- Remove unneeded unit test cases for `processBlockData` (now covered in subfunctions associated tests)
- Push (possibly unneeded) justification length check up in caller of `handleJustification`.
- Return error on block import error
- Fix position of debug logs

Done whilst fixing tests in #2774 

:new: is kinda blocking me on #2912 (trying to fix the memmm)

## Tests

```sh
go test -tags integration github.com/ChainSafe/gossamer/dot/sync
```

## Primary Reviewer

@edwardmack

### DIFF
--- a/dot/sync/chain_processor.go
+++ b/dot/sync/chain_processor.go
@@ -123,45 +123,11 @@ func (s *chainProcessor) processBlockData(bd *types.BlockData) error {
 	// while in bootstrap mode we don't need to broadcast block announcements
 	announceImportedBlock := s.chainSync.syncState() == tip
 	if hasHeader && hasBody {
-		// TODO: fix this; sometimes when the node shuts down the "best block" isn't stored properly,
-		// so when the node restarts it has blocks higher than what it thinks is the best, causing it not to sync
-		// if we update the node to only store finalised blocks in the database, this should be fixed and the entire
-		// code block can be removed (#1784)
-		block, err := s.blockState.GetBlockByHash(bd.Hash)
+		err = s.processBlockDataWithStateHeaderAndBody(*bd, announceImportedBlock)
 		if err != nil {
-			return fmt.Errorf("getting block by hash: %w", err)
+			return fmt.Errorf("processing block data with header and "+
+				"body in block state: %w", err)
 		}
-
-		err = s.blockState.AddBlockToBlockTree(block)
-		if errors.Is(err, blocktree.ErrBlockExists) {
-			logger.Debugf(
-				"block number %d with hash %s already exists in block tree, skipping it.",
-				block.Header.Number, bd.Hash)
-			return nil
-		} else if err != nil {
-			return fmt.Errorf("adding block to blocktree: %w", err)
-		}
-
-		if bd.Justification != nil {
-			err = s.handleJustification(&block.Header, *bd.Justification)
-			if err != nil {
-				return fmt.Errorf("handling justification: %w", err)
-			}
-		}
-
-		// TODO: this is probably unnecessary, since the state is already in the database
-		// however, this case shouldn't be hit often, since it's only hit if the node state
-		// is rewinded or if the node shuts down unexpectedly (#1784)
-		state, err := s.storageState.TrieState(&block.Header.StateRoot)
-		if err != nil {
-			return fmt.Errorf("loading trie state: %w", err)
-		}
-
-		err = s.blockImportHandler.HandleBlockImport(block, state, announceImportedBlock)
-		if err != nil {
-			return fmt.Errorf("handling block import: %w", err)
-		}
-
 		return nil
 	}
 
@@ -193,6 +159,50 @@ func (s *chainProcessor) processBlockData(bd *types.BlockData) error {
 
 	if err := s.blockState.CompareAndSetBlockData(bd); err != nil {
 		return fmt.Errorf("comparing and setting block data: %w", err)
+	}
+
+	return nil
+}
+
+func (c *chainProcessor) processBlockDataWithStateHeaderAndBody(blockData types.BlockData,
+	announceImportedBlock bool) (err error) { //nolint:revive
+	// TODO: fix this; sometimes when the node shuts down the "best block" isn't stored properly,
+	// so when the node restarts it has blocks higher than what it thinks is the best, causing it not to sync
+	// if we update the node to only store finalised blocks in the database, this should be fixed and the entire
+	// code block can be removed (#1784)
+	block, err := c.blockState.GetBlockByHash(blockData.Hash)
+	if err != nil {
+		return fmt.Errorf("getting block by hash: %w", err)
+	}
+
+	err = c.blockState.AddBlockToBlockTree(block)
+	if errors.Is(err, blocktree.ErrBlockExists) {
+		logger.Debugf(
+			"block number %d with hash %s already exists in block tree, skipping it.",
+			block.Header.Number, blockData.Hash)
+		return nil
+	} else if err != nil {
+		return fmt.Errorf("adding block to blocktree: %w", err)
+	}
+
+	if blockData.Justification != nil {
+		err = c.handleJustification(&block.Header, *blockData.Justification)
+		if err != nil {
+			return fmt.Errorf("handling justification: %w", err)
+		}
+	}
+
+	// TODO: this is probably unnecessary, since the state is already in the database
+	// however, this case shouldn't be hit often, since it's only hit if the node state
+	// is rewinded or if the node shuts down unexpectedly (#1784)
+	state, err := c.storageState.TrieState(&block.Header.StateRoot)
+	if err != nil {
+		return fmt.Errorf("loading trie state: %w", err)
+	}
+
+	err = c.blockImportHandler.HandleBlockImport(block, state, announceImportedBlock)
+	if err != nil {
+		return fmt.Errorf("handling block import: %w", err)
 	}
 
 	return nil
--- a/dot/sync/chain_processor.go
+++ b/dot/sync/chain_processor.go
@@ -82,9 +82,12 @@ func (s *chainProcessor) stop() {
 
 func (s *chainProcessor) processReadyBlocks() {
 	for {
-		bd := s.readyBlocks.pop(s.ctx)
-		if bd == nil { // context was canceled
-			return
+		bd, err := s.readyBlocks.pop(s.ctx)
+		if err != nil {
+			if errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) {
+				return
+			}
+			panic(fmt.Sprintf("unhandled error: %s", err))
 		}
 
 		if err := s.processBlockData(*bd); err != nil {
--- a/dot/sync/chain_processor.go
+++ b/dot/sync/chain_processor.go
@@ -156,8 +156,9 @@ func (s *chainProcessor) processBlockData(bd *types.BlockData) error {
 			return fmt.Errorf("loading trie state: %w", err)
 		}
 
-		if err := s.blockImportHandler.HandleBlockImport(block, state, announceImportedBlock); err != nil {
-			logger.Warnf("failed to handle block import: %s", err)
+		err = s.blockImportHandler.HandleBlockImport(block, state, announceImportedBlock)
+		if err != nil {
+			return fmt.Errorf("handling block import: %w", err)
 		}
 
 		return nil
--- a/dot/sync/chain_processor.go
+++ b/dot/sync/chain_processor.go
@@ -109,6 +109,8 @@ func (s *chainProcessor) processReadyBlocks() {
 // returns the index of the last BlockData it handled on success,
 // or the index of the block data that errored on failure.
 func (s *chainProcessor) processBlockData(bd *types.BlockData) error {
+	logger.Debugf("processing block data with hash %s", bd.Hash)
+
 	hasHeader, err := s.blockState.HasHeader(bd.Hash)
 	if err != nil {
 		return fmt.Errorf("checking if block state has header: %w", err)
@@ -162,8 +164,6 @@ func (s *chainProcessor) processBlockData(bd *types.BlockData) error {
 
 		return nil
 	}
-
-	logger.Debugf("processing block data with hash %s", bd.Hash)
 
 	if bd.Header != nil && bd.Body != nil {
 		if err := s.babeVerifier.VerifyBlock(bd.Header); err != nil {
--- a/dot/sync/chain_processor.go
+++ b/dot/sync/chain_processor.go
@@ -140,7 +140,7 @@ func (c *chainProcessor) processBlockData(blockData types.BlockData) error { //n
 		logger.Debugf("block with hash %s processed", blockData.Hash)
 	}
 
-	if blockData.Justification != nil && blockData.Header != nil {
+	if blockData.Justification != nil && len(*blockData.Justification) > 0 && blockData.Header != nil {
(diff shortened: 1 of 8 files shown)


dot/sync/chain_processor.go
```diff
@@ -249,9 +259,7 @@ func (s *chainProcessor) handleBlock(block *types.Block) error {
 }
 
 func (s *chainProcessor) handleJustification(header *types.Header, justification []byte) (err error) {
-	if len(justification) == 0 {
-		return nil
-	}
```
qdm12, line 254: Here if the justification isn't `nil`, I think we should process it even if it's empty right?? That looks like a potential attack :thinking:

dot/sync/chain_processor_test.go
```diff
@@ -375,67 +363,7 @@ func Test_chainProcessor_processBlockData(t *testing.T) {
...
 			},
-			blockData:     &types.BlockData{},
-			expectedError: mockError,
-		},
-		"handle AddBlockToBlockTree error": {
-			chainProcessorBuilder: func(ctrl *gomock.Controller) chainProcessor {
-				mockBlock := &types.Block{
-					Header: types.Header{
-						Number: uint(1),
-					},
```
qdm12, line 386: Note these few test cases are now unneeded for this function and moved to the new tests testing sub-functions.

dot/sync/chain_processor.go
qdm12: length justification check is pushed up the call stack, although it's likely unnecessary since the blockdata would have it set to `nil` instead of the empty byte slice.

dot/sync/chain_processor.go
```diff
@@ -74,7 +74,7 @@ func (s *chainProcessor) processReadyBlocks() {
 			return
 		}
 
-		if err := s.processBlockData(bd); err != nil {
+		if err := s.processBlockData(*bd); err != nil {
```
EclesioMeloJunior, line 77: is there a guarantee that `bd` is not nil? I think you should put a check to ensure `bd` is not nil before dereference it

dot/sync/chain_processor.go
```diff
@@ -95,99 +95,109 @@ func (s *chainProcessor) processReadyBlocks() {
...
+		return fmt.Errorf("checking if block state has body: %w", err)
 	}
 
 	if hasHeader && hasBody {
-		// TODO: fix this; sometimes when the node shuts down the "best block" isn't stored properly,
-		// so when the node restarts it has blocks higher than what it thinks is the best, causing it not to sync
-		// if we update the node to only store finalised blocks in the database, this should be fixed and the entire
-		// code block can be removed (#1784)
-		block, err := s.blockState.GetBlockByHash(bd.Hash)
+		err = c.processBlockDataWithStateHeaderAndBody(blockData)
```
EclesioMeloJunior, line 112: I would suggest to rename `hasHeader, hasBody` to `headerInState, bodyInState` to avoid a sigthly confusion. We first check `if hasHeader && hasBody` and then we check `if blockData.Header != nil && blockData.Body != nil` and looks like we are doing a double check against header and body,  changing the first condition to  `if headerInState && bodyInState` becomes clear to understand the next if statement

```go
- if hasHeader && hasBody {...}
+ if headerInState && bodyInState {...}
if blockData.Header != nil && blockData.Body != nil {...}
```

dot/sync/chain_processor.go
```diff
@@ -95,99 +95,109 @@ func (s *chainProcessor) processReadyBlocks() {
...
+			return fmt.Errorf("processing block data with header and body: %w", err)
 		}
+		logger.Debugf("block with hash %s processed", blockData.Hash)
+	}
 
-		// TODO: this is probably unnecessary, since the state is already in the database
-		// however, this case shouldn't be hit often, since it's only hit if the node state
-		// is rewinded or if the node shuts down unexpectedly (#1784)
-		state, err := s.storageState.TrieState(&block.Header.StateRoot)
+	if blockData.Justification != nil && len(*blockData.Justification) > 0 && blockData.Header != nil {
```
EclesioMeloJunior, line 128: supermeganit: in case header is nil will be not needed to derefer justification and check its len
```suggestion
	if blockData.Header != nil && blockData.Justification != nil && len(*blockData.Justification) > 0 {
```

EclesioMeloJunior (changes requested):

dot/sync/chain_processor.go
qdm12: It's kinda obscure, but yes; `bd := s.readyBlocks.pop(s.ctx)` only returns `nil` if the context gets canceled, and there is a

```go
		if s.ctx.Err() != nil {
			return
		}
```

check below. I changed that if check to `bd == nil` to make it more explicit :wink:

dot/sync/chain_processor.go
EclesioMeloJunior: I did some benchmarks and this does not make it faster 😢

EclesioMeloJunior (approved):
image

dot/sync/chain_processor.go
qdm12: I can confirm, I did this benchmark mixing half with header and half with no header and justification set.

```go
func BenchmarkXxx(b *testing.B) {
	withHeader := types.BlockData{
		Header:        &types.Header{},
		Justification: &[]byte{1, 2, 3, 4, 5},
	}
	withNoHeader := types.BlockData{
		Justification: &[]byte{1, 2, 3, 4, 5},
	}

	mixedBlockData := make([]types.BlockData, 500)
	for i := range mixedBlockData {
		if i%2 == 0 {
			mixedBlockData[i] = withHeader
		} else {
			mixedBlockData[i] = withNoHeader
		}
	}

	b.ResetTimer()

	b.Run("original", func(b *testing.B) {
		for i := 0; i < b.N; i++ {
			for _, blockData := range mixedBlockData {
				_ = blockData.Justification != nil && len(*blockData.Justification) > 0 && blockData.Header != nil
			}
		}
	})

	b.Run("suggested", func(b *testing.B) {
		for i := 0; i < b.N; i++ {
			for _, blockData := range mixedBlockData {
				_ = blockData.Header != nil && blockData.Justification != nil && len(*blockData.Justification) > 0
			}
		}
	})
}
```

I then even went the assembly route to check if there was any difference, and there is:

image

So at least that's that :wink: I guess our CPUs are just too quick.

dot/sync/chain_processor.go
```diff
@@ -70,11 +70,11 @@ func (s *chainProcessor) stop() {
 func (s *chainProcessor) processReadyBlocks() {
 	for {
 		bd := s.readyBlocks.pop(s.ctx)
-		if s.ctx.Err() != nil {
+		if bd == nil { // context was canceled
```
q9f, line 73: not necessarily a helpful comment for your requested review but the code would be more readable if we did not use variable names like `s` or `bd`. 

```go
chainProcessorContext := chainProcessor.readyBlocks.pop(chainProcessor.ctx)
```

q9f: also, inline comments should be a separate line above, what do you think?

```
// check if the chain-processor context was canceled
if chainProcessorContext == nil {
```

dot/sync/chain_processor.go
```diff
@@ -95,99 +95,109 @@ func (s *chainProcessor) processReadyBlocks() {
 // processBlockData processes the BlockData from a BlockResponse and
 // returns the index of the last BlockData it handled on success,
 // or the index of the block data that errored on failure.
-func (s *chainProcessor) processBlockData(bd *types.BlockData) error {
-	if bd == nil {
-		return ErrNilBlockData
-	}
+func (c *chainProcessor) processBlockData(blockData types.BlockData) error { //nolint:revive
```
q9f, line 98: curious why you renamed `s` to `c`?

👍🏼 for renaming `bd` to `blockData`

dot/sync/chain_processor.go
```diff
@@ -95,99 +95,109 @@ func (s *chainProcessor) processReadyBlocks() {
...
 	if err != nil {
-		return fmt.Errorf("failed to check if block state has header for hash %s: %w", bd.Hash, err)
+		return fmt.Errorf("checking if block state has header: %w", err)
 	}
-	hasBody, err := s.blockState.HasBlockBody(bd.Hash)
+
+	bodyInState, err := c.blockState.HasBlockBody(blockData.Hash)
 	if err != nil {
-		return fmt.Errorf("failed to check block state has body for hash %s: %w", bd.Hash, err)
+		return fmt.Errorf("checking if block state has body: %w", err)
```
q9f, line 108:
```suggestion
		return fmt.Errorf("Block state has no body: %w", err)
```

dot/sync/chain_processor.go
```diff
@@ -95,99 +95,109 @@ func (s *chainProcessor) processReadyBlocks() {
...
-		return ErrNilBlockData
-	}
+func (c *chainProcessor) processBlockData(blockData types.BlockData) error { //nolint:revive
+	logger.Debugf("processing block data with hash %s", blockData.Hash)
 
-	hasHeader, err := s.blockState.HasHeader(bd.Hash)
+	headerInState, err := c.blockState.HasHeader(blockData.Hash)
 	if err != nil {
-		return fmt.Errorf("failed to check if block state has header for hash %s: %w", bd.Hash, err)
+		return fmt.Errorf("checking if block state has header: %w", err)
```
q9f, line 103: Technically you already checked.

```suggestion
		return fmt.Errorf("Block state has no header: %w", err)
```

dot/sync/chain_processor.go
```diff
@@ -95,99 +95,109 @@ func (s *chainProcessor) processReadyBlocks() {
...
-			block.Header.Number, bd.Hash) // TODO is this valid?
-
-		err = s.blockState.AddBlockToBlockTree(block)
-		if errors.Is(err, blocktree.ErrBlockExists) {
-			return nil
-		} else if err != nil {
-			logger.Warnf("failed to add block with hash %s to blocktree: %s", bd.Hash, err)
-			return err
+			return fmt.Errorf("processing block data with header and "+
+				"body in block state: %w", err)
```
q9f, line 115: Does this return an error or a string now?

dot/sync/chain_processor.go
```diff
@@ -95,99 +95,109 @@ func (s *chainProcessor) processReadyBlocks() {
...
-		if bd.Justification != nil {
-			logger.Debugf("handling Justification for block number %d with hash %s...", block.Header.Number, bd.Hash)
-			err = s.handleJustification(&block.Header, *bd.Justification)
-			if err != nil {
-				return fmt.Errorf("handling justification: %w", err)
-			}
+	if blockData.Header != nil && blockData.Body != nil {
+		err = c.processBlockDataWithHeaderAndBody(blockData)
+		if err != nil {
+			return fmt.Errorf("processing block data with header and body: %w", err)
```
q9f, line 123:
```suggestion
			return fmt.Errorf("Cannot process block data with header and body: %w", err)
```

q9f (commented):
I cannot review the logic but left some questions and comments regarding style :)

dot/sync/chain_processor.go
qdm12: - for method receivers, I made peace with having short or single letter variable names. Although here it should be `c`/`cp`, not `s` (historical reasons I guess).
- for all other variables, short non-explicit names are a big no-no for me. Completely agree. At least for new code we add/modify I think our team now enforces explicit variable names.

For both cases, as you mentioned, it's a bit out of scope to keep deltas somehow low in this PR.

> also, inline comments should be a separate line above, what do you think?

For me, it really depends, if the comment is short and relates to a single line (here `if bd == nil {`), it's fine to inline it. If the comment is long or refers to a block of code below, then yeah have it on a separate line.
Anyway, I think it's better to leave this up to the developer, it's not a big deal and we shouldn't enforce too much code style or contributors (us internally or external) would just get mad :exploding_head: :wink:

dot/sync/chain_processor.go
qdm12: > curious why you renamed `s` to `c`?

Copy-pastaing my answer above:
for method receivers, I made peace with having short or single letter variable names.
Also see https://github.com/golang/go/wiki/CodeReviewComments#receiver-names :wink: (also just learnt that tbh)

dot/sync/chain_processor.go
qdm12: - the error comes from `c.blockState.HasHeader` and, if non-nil, does not mean the block state has no header for that block hash. It just failed checking if the block state had the header (i.e. disk io error etc.).
- the error occurred whilst `checking if block state has header`. The bottom wrapped error usually describes what actually went wrong. The `checking if block state has header: ` is just context (or human readable stack trace if you prefer) for that wrapped error.
- Don't capitalize errors :wink: see https://github.com/golang/go/wiki/CodeReviewComments#error-strings

dot/sync/chain_processor.go
qdm12: See https://github.com/ChainSafe/gossamer/pull/2811/files/b00715a7d879d9288c78bdcf98b306ab114032e5#r1012884704

dot/sync/chain_processor.go
qdm12: We try to avoid prefixing error strings with `cannot` / `failed to` / etc. otherwise it builds up quite some repetition when logged for errors returned from deep calls, i.e. `failed to do A: failed to do B: failed to do C: what went wrong actually`. That's why the perhaps slightly odd format convention `fmt.Errorf("doing this: %w", err)`

dot/sync/chain_processor.go
timwu20: why did we change from `*types.BlockData` to `types.BlockData`?

dot/sync/chain_processor.go
qdm12: Yeah I agree with you too Tim, reverting the code to how it was is also confusing I think.

I did tentative commit 015c45d6425e44592c2b1db014fdaa6bd7e3f4b9 changing `pop` to return an error when the context gets canceled? That way it would be more explicit, what do you think? Or something else

dot/sync/chain_processor.go
```diff
@@ -108,101 +111,113 @@ func (s *chainProcessor) processReadyBlocks() {
...
-		// is rewinded or if the node shuts down unexpectedly (#1784)
-		state, err := s.storageState.TrieState(&block.Header.StateRoot)
+	if blockData.Header != nil && blockData.Justification != nil && len(*blockData.Justification) > 0 {
+		err = c.handleJustification(blockData.Header, *blockData.Justification)
 		if err != nil {
-			logger.Warnf("failed to load state for block with hash %s: %s", block.Header.Hash(), err)
-			return err
+			return fmt.Errorf("handling justification: %w", err)
 		}
+	}
```
timwu20, line 151: Can we move to a single `if blockData.Header != nil {` block

qdm12: Indeed! Done :wink:

timwu20 (approved):

edwardmack (approved):

PR #2815, opened 2022-09-09
inherent improvements

kishansagathiya:
## Tests

```sh
```

## Primary Reviewer

@timwu20

### DIFF
--- a/dot/types/inherents.go
+++ b/dot/types/inherents.go
@@ -5,22 +5,41 @@ package types
 
 import (
 	"bytes"
-	"encoding/binary"
-	"errors"
 	"fmt"
 	"math/big"
 
 	"github.com/ChainSafe/gossamer/pkg/scale"
 )
 
-var (
-	// Timstap0 is an inherent key.
-	Timstap0 = []byte("timstap0")
-	// Babeslot is an inherent key.
-	Babeslot = []byte("babeslot")
-	// Uncles00 is an inherent key.
-	Uncles00 = []byte("uncles00")
+// InherentIdentifier is an identifier for an inherent.
+type InherentIdentifier uint
+
+const (
+	// Timstap0 is the identifier for the `timestamp` inherent.
+	Timstap0 InherentIdentifier = iota
+	// Babeslot is the BABE inherent identifier.
+	Babeslot
+	// Uncles00 is the identifier for the `uncles` inherent.
+	Uncles00
 )
+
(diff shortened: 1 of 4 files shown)


dot/types/inherents.go
```diff
@@ -5,23 +5,42 @@ package types
...
+// InherentIdentifier is an identifier for an inherent.
+type InherentIdentifier uint
+
+const (
+	// Timstap0 is the identifier for the `timestamp` inherent.
+	Timstap0 InherentIdentifier = iota
+	// Babeslot is the BABE inherent identifier.
+	Babeslot
+	// Uncles00 is the identifier for the `uncles` inherent.
+	Uncles00
```
qdm12, line 23: why not instead have these as `string` and then just use `[]byte(Timstap0)` where needed? Seems better than iota + method.

Also probably too out of scope, but these are not types, maybe they can be moved to lib/babe or lib/babe/constants (in case of cycle dependency) 🤔

dot/types/inherents.go
```diff
@@ -43,24 +62,20 @@ func (d *InherentsData) String() string {
 	return str
 }
 
-// SetInt64Inherent set an inherent of type uint64
-func (d *InherentsData) SetInt64Inherent(key []byte, data uint64) error {
-	if len(key) != 8 {
-		return errors.New("inherent key must be 8 bytes")
+// SetInherent sets a inherent.
+func (d *InherentsData) SetInherent(inherentIdentifier InherentIdentifier, value interface{}) error {
```
qdm12, line 66: supernit we could use `any` now that we use go 1.18

```suggestion
func (d *InherentsData) SetInherent(inherentIdentifier InherentIdentifier, value any) error {
```

dot/types/inherents.go
kishansagathiya: its not `[]byte(Timstap0)`, its [8]byte. It save me from doing things like `kb := [8]byte{}; copy(kb[:], key)`

> Also probably too out of scope, but these are not types, maybe they can be moved to lib/babe or lib/babe/constants (in case of cycle dependency) thinking

`InherentIdentifier` should stay where InherentsData stays. We could however move inherents.go file to `lib/babe`. It makes more sense there anyway.

kishansagathiya: Inherents get used in lib/runtime, and lib/sync as well. Let's just keep them here in dot/types. This saves possible cyclical dependencies

PR #2827, opened 2022-09-13
Author a sibling block in case best block's slot is same as current slot 

kishansagathiya:
## Changes
Copy of https://github.com/ChainSafe/gossamer/pull/2726 with the merge fix

- There were files ending with `_test.go` which were not actually tests, but just things that were getting used in test files. I had to remove `_test` from them in order to use that helper code to be used in other packages.
- `createTestService` in babe tests had some un-necessary conditions. Removed them. It also helps me with the test that I have written. 
- Remove some  extra redundant code that created TransactionState
- Used actual block import handler instead of a fake one.
- 

## Tests

```sh
```

## Primary Reviewer

@timwu20


dot/core/mocks_generate.go
```diff
@@ -0,0 +1,6 @@
+// Copyright 2022 ChainSafe Systems (ON)
```
qdm12, line 1: Please keep the *_test.go mock files like this. You can generate duplicate mocks where needed from locally defined interfaces. Otherwise this creates more package dependencies which makes cycle dependency hell a closer reality (and hard to break up packages as well).

qdm12 (changes requested):

lib/babe/errors.go
```diff
@@ -71,6 +71,7 @@ var (
 	errInvalidSlotTechnique       = errors.New("invalid slot claiming technique")
 	errNoBABEAuthorityKeyProvided = errors.New("cannot create BABE service as authority; no keypair provided")
 	errLastDigestItemNotSeal      = errors.New("last digest item is not seal")
+	errLaggingSlot                = errors.New("cannot claim slot, current slot is smaller than slot of best block")
```
qdm12, line 74: nit the `claim slot` part seems out of scope given the error var name?

Perhaps just have it as 

```suggestion
	errLaggingSlot                = errors.New("current slot is smaller than slot of best block")
```

or rename the error variable name

lib/genesis/test_utils.go
```diff
@@ -100,8 +101,22 @@ func newGenesisTrieAndHeader(t *testing.T, gen *Genesis) (*trie.Trie, *types.Hea
...
+
+	bdEnc, err := scale.Marshal(babeDigest)
+	require.NoError(t, err)
+
+	digest := types.NewDigest()
+	err = digest.Add(types.PreRuntimeDigest{
+		ConsensusEngineID: types.BabeEngineID,
+		Data:              bdEnc,
+	})
+	require.NoError(t, err)
```
qdm12, line 116: self reminder: @qdm12 in #2795 this should be added where necessary

lib/babe/babe_integration_test.go
```diff
@@ -293,3 +300,160 @@ func TestService_PauseAndResume(t *testing.T) {
...
+	cfg := ServiceConfig{
+		Authority: true,
+		Lead:      true,
+	}
+	babeService := createTestService(t, cfg)
+
+	err := babeService.Start()
+	require.NoError(t, err)
+	defer func() {
+		_ = babeService.Stop()
```
qdm12, line 314: nit

```suggestion
		err = babeService.Stop()
		require.NoError(t, err)
```

lib/babe/babe_integration_test.go
```diff
@@ -293,3 +300,160 @@ func TestService_PauseAndResume(t *testing.T) {
...
+
+	epochData, err := babeService.initiateEpoch(testEpochIndex)
+	require.NoError(t, err)
+
+	ext := runtime.NewTestExtrinsic(t, rt, parentHash, parentHash, 0, "System.remark", []byte{0xab, 0xcd})
+	block := createTestBlock(t, babeService, emptyHeader, [][]byte{common.MustHexToBytes(ext)},
+		1, testEpochIndex, epochData)
+
+	babeService.blockState.AddBlock(block)
+	time.Sleep(babeService.constants.slotDuration * 1)
```
qdm12, line 330: nit

```suggestion
	time.Sleep(babeService.constants.slotDuration)
```

lib/babe/babe_integration_test.go
```diff
@@ -293,3 +300,160 @@ func TestService_PauseAndResume(t *testing.T) {
...
+	header, err := babeService.blockState.BestBlockHeader()
+	require.NoError(t, err)
+
+	bestBlockSlotNum, err := babeService.blockState.GetSlotForBlock(header.Hash())
+	require.NoError(t, err)
+
+	slotnum := uint64(1)
+	slot := Slot{
+		start:    time.Now(),
+		duration: 1 * time.Second,
```
qdm12, line 341: nit

```suggestion
		duration: time.Second,
```

lib/babe/babe_integration_test.go
```diff
@@ -293,3 +300,160 @@ func TestService_PauseAndResume(t *testing.T) {
...
+	cfgAlice.AuthData = []types.Authority{
+		{
+			Key:    alice.Public().(*sr25519.PublicKey),
+			Weight: 1,
+		},
+		{
+			Key:    bob.Public().(*sr25519.PublicKey),
+			Weight: 1,
+		},
+	}
```
qdm12, line 381: nit all in one

```suggestion
		Keypair:   alice,
		AuthData: []types.Authority{
			{
				Key:    alice.Public().(*sr25519.PublicKey),
				Weight: 1,
			},
			{
				Key:    bob.Public().(*sr25519.PublicKey),
				Weight: 1,
			},
		},
	}
```

lib/babe/babe_integration_test.go
```diff
@@ -293,3 +300,160 @@ func TestService_PauseAndResume(t *testing.T) {
...
+	cfgBob.AuthData = []types.Authority{
+		{
+			Key:    alice.Public().(*sr25519.PublicKey),
+			Weight: 1,
+		},
+		{
+			Key:    bob.Public().(*sr25519.PublicKey),
+			Weight: 1,
+		},
+	}
```
qdm12, line 398: nit all in one statement

```suggestion
		AuthData: []types.Authority{
			{
				Key:    alice.Public().(*sr25519.PublicKey),
				Weight: 1,
			},
			{
				Key:    bob.Public().(*sr25519.PublicKey),
				Weight: 1,
			},
		},
	}
```

lib/babe/babe_integration_test.go
```diff
@@ -293,3 +300,160 @@ func TestService_PauseAndResume(t *testing.T) {
...
+	babeServiceBob := createTestService(t, cfgBob)
+
+	err := babeServiceBob.Start()
+	require.NoError(t, err)
+	defer func() {
+		_ = babeServiceBob.Stop()
+	}()
+
+	// wait till bob creates a block
+	time.Sleep(babeServiceBob.constants.slotDuration * 1)
```
qdm12, line 409: nit

```suggestion
	time.Sleep(babeServiceBob.constants.slotDuration)
```

lib/babe/babe_integration_test.go
```diff
@@ -293,3 +300,160 @@ func TestService_PauseAndResume(t *testing.T) {
...
+	require.NoError(t, err)
+
+	block, err := babeServiceBob.blockState.GetBlockByNumber(1)
+	require.NoError(t, err)
+
+	err = babeServiceBob.Stop()
+	require.NoError(t, err)
+
+	time.Sleep(babeServiceBob.constants.slotDuration * 1)
```
qdm12, line 418: nit

```suggestion
	time.Sleep(babeServiceBob.constants.slotDuration)
```

lib/babe/babe_integration_test.go
```diff
@@ -293,3 +300,160 @@ func TestService_PauseAndResume(t *testing.T) {
...
+
+	babeServiceAlice := createTestService(t, cfgAlice)
+
+	// Add block created by Bob to Alice
+	err = babeServiceAlice.blockState.AddBlock(block)
+	require.NoError(t, err)
+
+	time.Sleep(babeServiceBob.constants.slotDuration * 1)
```
qdm12, line 426: nit

```suggestion
	time.Sleep(babeServiceBob.constants.slotDuration)
```

lib/babe/babe_integration_test.go
```diff
@@ -293,3 +300,160 @@ func TestService_PauseAndResume(t *testing.T) {
...
+
+	bestBlockHeader, err := babeServiceAlice.blockState.BestBlockHeader()
+	require.NoError(t, err)
+	require.Equal(t, block.Header.Hash().String(), bestBlockHeader.Hash().String())
```
qdm12, line 430: nit same thing I think?

```suggestion
	require.Equal(t, block.Header.Hash(), bestBlockHeader.Hash())
```

lib/babe/babe_integration_test.go
```diff
@@ -293,3 +300,160 @@ func TestService_PauseAndResume(t *testing.T) {
...
+
+	// If the slot we are claiming is same as slot in best header, test that we don't
+	// through any error and can claim slot.
```
qdm12, line 433:
```suggestion
	// If the slot we are claiming is the same as the slot of the best block header, test that we can
	// still claim the slot without error.
```

lib/babe/babe_integration_test.go
```diff
@@ -293,3 +300,160 @@ func TestService_PauseAndResume(t *testing.T) {
...
+	bestBlockSlotNum, err := babeServiceAlice.blockState.GetSlotForBlock(bestBlockHeader.Hash())
+	require.NoError(t, err)
+
+	slot := Slot{
+		start:    time.Now(),
+		duration: 1 * time.Second,
```
qdm12, line 439: nit

```suggestion
		duration: time.Second,
```

lib/babe/babe_integration_test.go
```diff
@@ -293,3 +300,160 @@ func TestService_PauseAndResume(t *testing.T) {
...
+		number:   bestBlockSlotNum,
+	}
+	testVRFOutputAndProof := &VrfOutputAndProof{}
+	preRuntimeDigest, err := types.NewBabePrimaryPreDigest(
+		0, slot.number,
+		testVRFOutputAndProof.output,
+		testVRFOutputAndProof.proof,
```
qdm12, line 446: nit no need for the extra variable

```suggestion
	preRuntimeDigest, err := types.NewBabePrimaryPreDigest(
		0, slot.number,
		[sr25519.VRFOutputLength]byte{},
		[sr25519.VRFProofLength]byte{},
```

lib/babe/babe_integration_test.go
```diff
@@ -293,3 +300,160 @@ func TestService_PauseAndResume(t *testing.T) {
...
+	slot := Slot{
+		start:    time.Now(),
+		duration: 1 * time.Second,
+		number:   slotnum,
+	}
+	testVRFOutputAndProof := &VrfOutputAndProof{}
+	preRuntimeDigest, err := types.NewBabePrimaryPreDigest(
+		0, slot.number,
+		testVRFOutputAndProof.output,
+		testVRFOutputAndProof.proof,
```
qdm12, line 348: nit no need for extra var

```suggestion
	preRuntimeDigest, err := types.NewBabePrimaryPreDigest(
		0, slot.number,
		[sr25519.VRFOutputLength]byte,
		[sr25519.VRFProofLength]byte,
```

EclesioMeloJunior:
@kishansagathiya looking at substrate code I notice they have a method to report equivocations, is it related to this PR? should we implement it? 				https://github.com/paritytech/substrate/blob/c24431eb6a95197550b30715a4c124be1aea79de/client/consensus/babe/src/lib.rs#L1224

dot/core/mocks_generate.go
kishansagathiya: > Otherwise this creates more package dependencies which makes cycle dependency hell a closer reality (and hard to break up packages as well).

This particular instance does not create an cycle dependency. 

I was trying to check if there is any convention to have `_test.go` in mock files. Add some reference for such mocking guidelines in go.

I checked in other popular go projects to see if they are adding `_test.go`, 
Check the kubernetes project https://github.com/kubernetes/kubernetes/search?q=Code+generated+by+MockGen
In most cases files containing mocks don't have `_test,go` in them.
Same in some aws code https://github.com/search?l=Go&q=org%3Aaws+Code+generated+by+MockGen.&type=Code
Similar things in grafana too https://github.com/grafana/grafana/blob/a14621fff679add99c494628f1fb9a001a734f47/pkg/services/live/runstream/mock.go

kishansagathiya: From what I understand anything that ends with `_test.go` has to be a test.

lib/babe/babe.go
```diff
@@ -467,6 +467,30 @@ func (b *Service) handleSlot(epoch, slotNum uint64,
...
+			return errLaggingSlot
+		}
+
+		if bestBlockSlotNum == slotNum {
+			// pick parent of best block instead to handle slot
+			newParentHeader, err := b.blockState.GetHeader(parentHeader.ParentHash)
+			if err != nil {
+				return fmt.Errorf("could not get header for hash %s: %w", parentHeader.ParentHash, err)
+			}
+			if newParentHeader == nil {
```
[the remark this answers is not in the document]
kishansagathiya: It probably isn't supposed to happen, but I have seen it happen.

lib/babe/babe.go
```diff
@@ -467,6 +467,30 @@ func (b *Service) handleSlot(epoch, slotNum uint64,
...
+				return errNilParentHeader
+			}
+			parentHeader = newParentHeader
```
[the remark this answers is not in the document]
kishansagathiya: in the previous line when I am returning the error, I am using hash from previous parentHeader, 
If I use it directly, in case of error, parentHeader will become nil and I won't be able to get its parentHeader.parentHash. Plus it also looks cleaner.

lib/babe/babe_integration_test.go
```diff
@@ -257,17 +266,15 @@ func TestService_GetAuthorityIndex(t *testing.T) {
 }
 
 func TestStartAndStop(t *testing.T) {
-	serviceConfig := ServiceConfig{}
-	bs := createTestService(t, serviceConfig)
+	bs := createTestService(t, ServiceConfig{})
```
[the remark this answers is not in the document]
kishansagathiya: None of this comment makes sense to me.

> createTestService(t, ServiceConfig{}) is uglier.

Ugly and beautiful is very subjective. What you are finding ugly here, is prettier for me.

there is an empty `serviceConfig` created only to pass in the very next line, which does not make any sense to me. Here you are advocating creation of variable which is used only once. 

While on the other side, in the same PR you have a suggestion (https://github.com/ChainSafe/gossamer/pull/2827#discussion_r973318034) to remove a variable creation which is getting used twice and makes the code readable. The code that you have suggested there, I find it uglier.

lib/babe/babe_integration_test.go
```diff
@@ -80,37 +79,32 @@ func createTestService(t *testing.T, cfg ServiceConfig) *Service {
...
+	})
 
-		cfg.BlockState = dbSrv.Block
-		cfg.StorageState = dbSrv.Storage
-		cfg.EpochState = dbSrv.Epoch
-	}
+	cfg.BlockState = dbSrv.Block
+	cfg.StorageState = dbSrv.Storage
+	cfg.EpochState = dbSrv.Epoch
+	cfg.TransactionState = dbSrv.Transaction
```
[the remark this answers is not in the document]
kishansagathiya: Probably not, 

But I need `dbSrv` to have an actual blockState. Now, `dbSrv` also gives a transaction state also, so there is no need to create a separate standalone state.NewTransactionState

lib/babe/babe_integration_test.go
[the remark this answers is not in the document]
qdm12: Still unneeded delta. If you had to change the code for another reason and you want to change the style, sure go ahead. But this is isolated and unneeded.

Also ideally in the future, we *should* define the config in the test and inject it to `createTestService` (and not have these obscure nil checks inside all over the place), so we'll have to define a separate variable again.

Regarding https://github.com/ChainSafe/gossamer/pull/2827#discussion_r973318034 it was not to remove a variable, but really since that was confusing to use `testVRFOutputAndProof.output` whereas you could use a more explicit `[sr25519.VRFOutputLength]byte{}`

kishansagathiya:
@EclesioMeloJunior 
> @kishansagathiya looking at substrate code I notice they have a method to report equivocations, is it related to this PR? should we implement it? https://github.com/paritytech/substrate/blob/c24431eb6a95197550b30715a4c124be1aea79de/client/consensus/babe/src/lib.rs#L1224

This is a really good find. I went through that piece of code and tried to figure out what it does. It appears to me that we should report equivocations. I have created an issue for that https://github.com/ChainSafe/gossamer/issues/2853. Take a look at it, let me know your thoughts.

dot/core/mocks_generate.go
kishansagathiya: I really don't enjoy these debates. I have added _test back, and have added a separate mock for network in babe.

lib/babe/babe_integration_test.go
```diff
@@ -79,37 +78,34 @@ func createTestService(t *testing.T, cfg ServiceConfig) *Service {
...
+		LogLevel:  log.Info,
+		Telemetry: telemetryMock,
+	}
+	dbSrv = state.NewService(config)
+	dbSrv.UseMemDB()
 
-		err = dbSrv.Initialise(&gen, &genHeader, &genTrie)
-		require.NoError(t, err)
+	// err = dbSrv.Initialise(gen, genHeader, genTrie)
+	// require.NoError(t, err)
```
EclesioMeloJunior, line 94:
```suggestion
```

EclesioMeloJunior (approved):
LGTM, just a small unneeded comment

lib/babe/babe.go
```diff
@@ -454,32 +454,65 @@ func (b *Service) handleEpoch(epoch uint64) (next uint64, err error) {
 	return next, nil
 }
 
-func (b *Service) handleSlot(epoch, slotNum uint64,
-	authorityIndex uint32,
-	preRuntimeDigest *types.PreRuntimeDigest,
-) error {
+func (b *Service) getParentForBlockAuthoring(slotNum uint64) (*types.Header, error) {
```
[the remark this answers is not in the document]
kishansagathiya: It gets tested `TestService_HandleSlotWithSameSlot` and `TestService_HandleSlotWithLaggingSlot`.

timwu20 (approved):

lib/babe/babe.go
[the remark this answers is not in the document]
timwu20: @qdm12 please remember this was a [PR](https://github.com/ChainSafe/gossamer/pull/2726/files) that was already approved before.  The lack of existing unit tests in the `babe` package is known.  I don't think it makes sense to add scope to this PR now by just unit testing this function, and not `handleSlot` as well.

lib/babe/mocks_generate_test.go
```diff
@@ -0,0 +1,6 @@
+// Copyright 2022 ChainSafe Systems (ON)
+// SPDX-License-Identifier: LGPL-3.0-only
+
+package babe
+
+//go:generate mockgen -destination=mocks/network.go -source=../../dot/core/interface.go -package mocks dot/core Network
```
kishansagathiya, line 6: @qdm12 Would you know whats wrong here? it mocks all interfaces instead of just the network interface

qdm12: Don't use the `-source` mode (also don't generate to a mocks directory, generate to a local _test file):

```go
//go:generate mockgen -destination=mocks_test.go -package=$GOPACKAGE github.com/ChainSafe/gossamer/dot/core 
 Network
```

lib/babe/babe.go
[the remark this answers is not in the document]
qdm12: I feel like new code added should be unit tested, but up to you if you feel it's not necessary. I wasn't referring to unit testing  `handleSlot`, that's indeed out of scope.

lib/babe/mocks_generate_test.go
kishansagathiya: done.

> also don't generate to a mocks directory, generate to a local _test file
There is already a mocks directory here. I would prefer to just let it stay the way it is.

qdm12 (approved):

PR #2829, opened 2022-09-13
Time based simpler unrestricted loop

qdm12:


dot/state/transaction.go
```diff
@@ -28,8 +30,9 @@ type TransactionState struct {
 
 // NewTransactionState returns a new TransactionState
 func NewTransactionState(telemetry telemetry.Client) *TransactionState {
+	const popTryWait = 10 * time.Millisecond
```
qdm12, line 33: Now comes the question... what should this value be? I guess we don't want to wait longer than the slot time... Also even if it's half of the slot time, that means we could miss some extrinsics whilst we sleep. IMO time based is not a solution here.

qdm12:
I'll close this to avoid polluting draft PR list view

PR #2831, opened 2022-09-13
chore(dot/state/pruner): refactoring and fixes

### LINKED ISSUE #2835: dot/state/pruner: refactor code to be more understandable and modular
The pruner code is (unit, integration and even usage-wise) untested, and hard to understand.

Since the trie lazy loading of nodes will interact a lot more with the database, and that the pruner already does some work with pruning database keys (node hashes), it feels required to understand the online pruner code and make it more modular.

qdm12:
## Changes

- [x] Correctly prune storage database
- [x] Prune journal database
- [x] Everything stored on disk so that it handles a Gossamer restart
- [x] Event driven pruning instead of periodic pruning
  - Does not allow to have more than the number of blocks to retain on disk (in storage and journal databases)
  - No performance impact since IO is already being done (storing journal data)
  - Fix thread safety locking for the entire store journal record operation
- [x] Do not store inserted node hashes in journal database (unneeded)
- [x] Safer/transactional database batches
- [x] Support node hash re-insertion when deleted by one or more ancestor/uncle blocks
- [x] Check descendant of a block using the block state (injected) to ensure we can remove a node hash to be pruned
- [x] Minor refactoring
  - [x] Move to `internal/pruner/full` and `internal/pruner/archive`
  - [x] Use `node hash` + `common.Hash` instead of `merkle value` + `string`, since it's only 32B node hashes
  - [x] Inject database and tables to the pruner constructor
  - [x] Remove no longer needed `NewIterator` database method from production code
  - [x] Rename `StoreJournalRecord` to `RecordAndPrune`
- [ ] Optimisations (create issue):
  - [ ] Read cache to reduce database lookups
  - [ ] Use custom binary encoding for quicker appends than scale decode + re-encode

## Issues

Fixes #2835 

- Throw out node subvalues for the v1 tries (for #2418) 
- Simplify/Dedup work in #2994 for the database interface change
- Fix the really wrong full pruner implementation code (not resistant to program restarts, pruning wrong things etc.)

## Primary Reviewer

@EclesioMeloJunior


chain/dev/defaults.go
```diff
@@ -28,8 +28,8 @@ var (
 	// DefaultLvl is the default log level
 	DefaultLvl = log.Info
 
-	// DefaultPruningMode is the default pruning mode
-	DefaultPruningMode = "archive"
+	// DefaultPruningEnabled describes if online pruning should be enabled by default.
+	DefaultPruningEnabled = false
```
qdm12, line 32: Let me know if you prefer to keep it as a string (might be needed for more pruning options later), but I changed it to a boolean to keep it simpler for now.

dot/state/base.go
```diff
@@ -155,26 +154,3 @@ func (s *BaseState) loadSlotDuration() (uint64, error) {
 
 	return binary.LittleEndian.Uint64(data), nil
 }
-
-// storePruningData stores the pruner configuration.
-func (s *BaseState) storePruningData(mode pruner.Config) error {
```
qdm12, line 160: That was persisting the pruning mode, whereas really it should be given at start from the toml config or flags, and be able to change from one mode to the other.

dot/state/storage.go
```diff
@@ -86,7 +76,19 @@ func (s *StorageState) StoreTrie(ts *rtstorage.TrieState, header *types.Header)
 			return fmt.Errorf("failed to get state trie inserted keys: block %s %w", header.Hash(), err)
 		}
 
-		err = s.pruner.StoreJournalRecord(deletedMerkleValues, insertedMerkleValues, header.Hash(), int64(header.Number))
+		// Temporary work-around until we drop merkle values for node hashes in another PR (already opened).
+		insertedNodeHashes := make(map[common.Hash]struct{}, len(insertedMerkleValues))
+		for k := range insertedMerkleValues {
```
qdm12, line 81: Conversion will be removed once #2915 gets merged.

pkg/scale/decode.go
```diff
@@ -424,6 +424,14 @@ func (ds *decodeState) decodeMap(dstv reflect.Value) (err error) {
 	if err != nil {
 		return fmt.Errorf("decoding length: %w", err)
 	}
+
+	if dstv.IsNil() {
+		keyType := dstv.Type().Key()
+		valueType := dstv.Type().Elem()
+		aMapType := reflect.MapOf(keyType, valueType)
+		dstv.Set(reflect.MakeMapWithSize(aMapType, int(numberOfTuples)))
+	}
```
qdm12, line 433: this is so it can be decoded to a nil typed destination map

internal/pruner/full/pruner_test.go
```diff
@@ -0,0 +1,175 @@
...
+			[]journalKey{{1, common.Hash{101}}}),
+	}
+	assertDatabaseContent(t, database, keyValuePairs)
+
+	// Block 1 hash 102
+	setNodeHashesInStorageDB(t, storageDB, []common.Hash{{5}, {6}})
+	logger.EXPECT().Debugf("journal data stored for block number %d and block hash %s",
+		uint32(1), "0x66000000...00000000")
+	err = pruner.RecordAndPrune(
+		map[common.Hash]struct{}{{3}: {}},          // deleted node hashes
```
qdm12, line 81: Should we prune 3?

dot/state/service_integration_test.go
```diff
@@ -176,9 +175,9 @@ func TestService_StorageTriePruning(t *testing.T) {
 	config := Config{
```
qdm12, line 175: The test `TestService_StorageTriePruning` which was modified in #3063 should now be unskipped.

PR #2834, opened 2022-09-15
fix(lib/trie): prepare trie nodes for mutation only when needed

qdm12:
## Changes

Done whilst working on the online pruning / lazy loading stuff.
EDIT: actually rather needed for the lazy loading

- Only prepare nodes for mutation if mutation is guaranteed
- **Fixes 'reported deleted hashes' when no mutation is done**, indirectly fixing the online pruning that would previously prune database keys that may still be in use.
- Add comment on `deletedKeys` trie struct field
- Update tests to maintain full test coverage on trie code

## Tests

```sh
```

## Issues

#2836 which blocks #2838 

## Primary Reviewer

@timwu20

### DIFF
--- a/lib/trie/trie.go
+++ b/lib/trie/trie.go
@@ -353,7 +353,7 @@ func (t *Trie) insertInLeaf(parentLeaf *Node, key, value []byte) (
 	newParent *Node, mutated bool, nodesCreated uint32) {
 	if bytes.Equal(parentLeaf.Key, key) {
 		nodesCreated = 0
-		if bytes.Equal(value, parentLeaf.SubValue) {
+		if parentLeaf.SubValueEqual(value) {
 			const mutated = false
 			return parentLeaf, mutated, nodesCreated
 		}
@@ -433,7 +433,7 @@ func (t *Trie) insertInBranch(parentBranch *Node, key, value []byte) (
 	copySettings := node.DefaultCopySettings
 
 	if bytes.Equal(key, parentBranch.Key) {
-		if bytes.Equal(parentBranch.SubValue, value) {
+		if parentBranch.SubValueEqual(value) {
 			const mutated = false
 			return parentBranch, mutated, 0
 		}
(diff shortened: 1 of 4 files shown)


lib/trie/trie.go
```diff
@@ -319,20 +323,22 @@ func findNextKeyChild(children []*Node, startIndex byte,
...
 }
 
 // insert inserts a value in the trie at the key specified.
 // It may create one or more new nodes or update an existing node.
-func (t *Trie) insert(parent *Node, key, value []byte) (newParent *Node, nodesCreated uint32) {
+func (t *Trie) insert(parent *Node, key, value []byte) (
+	newParent *Node, mutated bool, nodesCreated uint32) {
 	if parent == nil {
 		const nodesCreated = 1
+		const mutated = true
```
timwu20, line 335: is there a reason why you declare const here instead of just putting them inline? You can also just use the allocated named return params.

qdm12: The reason is kind of excessive thinking I guess, but here we go:

- `return x, y, true, z`  for a someone reading without an IDE (i.e. github PR files changed page) is sometimes confusing about what returned value is which, so I assign to a named var/constant so it's more readable locally without having to cross check the function signature
- I use `const mutated = true` instead of `mutated = true` to highlight it doesn't change anymore. It does help reading when something doesn't change anymore for a larger block of code within a condition, although sometimes (like here) it's not really needed either.
- One last tiny note is that compiler wise, these 3 approaches result in the same assembly

I can change it to `mutated = true` if you feel it's clearer though

timwu20: I think having them inline is nicer to read and reduces the number of lines.  I would consider using the named parameters if you were trying to reduce allocations.

timwu20 (approved):

EclesioMeloJunior (approved):

lib/trie/trie.go
qdm12: Changing them to use the named arguments instead of using `const`, especially for short blocks, since it doesn't really help and is to some extent confusing.
On the other hand, readability wise, my over-opinionated feeling is to name it near the `return` so one doesn't have to go at the function signature try to decipher what return position corresponds to what return argument. I won't ever nit anyone on that though :smiley_cat: 
Also note (as mentioned above) allocation-wise it's the exact same if you use `const` or a the named return or `return 1` directly, so it's really just code style and no performance difference.

PR #2851, opened 2022-09-21
revise SetStructInherent and SetInt64Inherent to generic SetInherent

timwu20:
… add tests against encoded data from rust examples

## Tests

```sh
```

## Primary Reviewer

@kishansagathiya

### DIFF
--- a/dot/types/inherents.go
+++ b/dot/types/inherents.go
@@ -5,7 +5,6 @@ package types
 
 import (
 	"bytes"
-	"encoding/binary"
 	"fmt"
 	"math/big"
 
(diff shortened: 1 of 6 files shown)


dot/types/inherents.go
```diff
@@ -46,29 +45,8 @@ func (d *InherentsData) String() string {
...
-		panic(fmt.Sprintf("inherent key must be 8 bytes but is: %v", key))
-	}
-
-	val := make([]byte, 8)
-	binary.LittleEndian.PutUint64(val, data)
-
-	venc, err := scale.Marshal(val)
-	if err != nil {
-		return err
-	}
```
timwu20, line 61: when comparing with the substrate code I noticed that when we try to encode `InherentData` in substrate with inherents added with key `timstap0` where the value is of type `i64` that the encoding differs.  I ended up added a test in `inherents_test.go` to ensure the same encoding.

Issue #2848, opened 2022-09-20
Return error on Value() call when VDT is not set


jimjbrettj:
## Issue summary

- Currently when you call Value() on a VDT and it is not set scale will return nil. Will change to return an error to allow for more clean and clear handling.

danforbes:
Should this Issue be closed with #2852?

PR #2708, opened 2022-07-27
feat(pkg/scale): add use of pkg/error Wrap for error handling

edwardmack:
## Changes
- When preparing to address issue #2631 I found this article regarding error handling https://dave.cheney.net/2016/04/27/dont-just-check-errors-handle-them-gracefully, which suggests that the use of Sentinel errors should be avoided for a number of reasons, so I decided to follow his advice and address this issue using `github.com/pkg/errors` package.  

## Tests

```sh
go test github.com/ChainSafe/gossamer/pkg/scale
```

## Issues
will close #2631 

## Primary Reviewer

@timwu20
@qdm12

PR #2852, opened 2022-09-21
enhancement(pkg/scale): return error on vdt.Value() call

jimjbrettj:
## Changes

- Return error on the Value() function for varyingDataType

## Tests

```sh
go test ./pkg/scale
```

## Issues

#2848

## Primary Reviewer

@timwu20

### DIFF
--- a/dot/core/service.go
+++ b/dot/core/service.go
@@ -140,7 +140,11 @@ func (s *Service) HandleBlockProduced(block *types.Block, state *rtstorage.TrieS
 
 	digest := types.NewDigest()
 	for i := range block.Header.Digest.Types {
-		err := digest.Add(block.Header.Digest.Types[i].Value())
+		digestValue, err := block.Header.Digest.Types[i].Value()
+		if err != nil {
+			return err
+		}
+		err = digest.Add(digestValue)
 		if err != nil {
 			return err
 		}
--- a/dot/core/service.go
+++ b/dot/core/service.go
@@ -142,7 +142,7 @@ func (s *Service) HandleBlockProduced(block *types.Block, state *rtstorage.TrieS
 	for i := range block.Header.Digest.Types {
 		digestValue, err := block.Header.Digest.Types[i].Value()
 		if err != nil {
-			return err
+			return fmt.Errorf("getting value of digest type at index %d: %w", i, err)
 		}
 		err = digest.Add(digestValue)
 		if err != nil {
--- a/dot/digest/digest.go
+++ b/dot/digest/digest.go
@@ -105,12 +105,12 @@ func (h *Handler) toConsensusDigests(scaleVaryingTypes []scale.VaryingDataType) 
 	for _, d := range scaleVaryingTypes {
 		digestValue, err := d.Value()
 		if err != nil {
-			h.logger.Debug(err.Error())
+			h.logger.Error(err.Error())
 			continue
 		}
 		digest, ok := digestValue.(types.ConsensusDigest)
 		if !ok {
-			h.logger.Debugf("digest type not supported: %T", digestValue)
+			h.logger.Errorf("digest type not supported: %T", digestValue)
 			continue
 		}
 
@@ -128,7 +128,7 @@ func (h *Handler) toConsensusDigests(scaleVaryingTypes []scale.VaryingDataType) 
 func checkForGRANDPAForcedChanges(digests []types.ConsensusDigest) ([]types.ConsensusDigest, error) {
 	var hasForcedChange bool
 	digestsWithoutScheduled := make([]types.ConsensusDigest, 0, len(digests))
-	for _, digest := range digests {
+	for i, digest := range digests {
 		if digest.ConsensusEngineID != types.GrandpaEngineID {
 			digestsWithoutScheduled = append(digestsWithoutScheduled, digest)
 			continue
@@ -142,7 +142,7 @@ func checkForGRANDPAForcedChanges(digests []types.ConsensusDigest) ([]types.Cons
 
 		dataValue, err := data.Value()
 		if err != nil {
-			return nil, err
+			return nil, fmt.Errorf("getting value of digest type at index %d: %w", i, err)
 		}
 		switch dataValue.(type) {
 		case types.GrandpaScheduledChange:
@@ -189,7 +189,7 @@ func (h *Handler) handleBabeConsensusDigest(digest scale.VaryingDataType, header
 
 	digestValue, err := digest.Value()
 	if err != nil {
-		return err
+		return fmt.Errorf("getting digest value: %w", err)
 	}
 	switch val := digestValue.(type) {
 	case types.NextEpochData:
--- a/dot/digest/digest.go
+++ b/dot/digest/digest.go
@@ -103,9 +103,14 @@ func (h *Handler) toConsensusDigests(scaleVaryingTypes []scale.VaryingDataType) 
 	consensusDigests := make([]types.ConsensusDigest, 0, len(scaleVaryingTypes))
 
 	for _, d := range scaleVaryingTypes {
-		digest, ok := d.Value().(types.ConsensusDigest)
+		digestValue, err := d.Value()
+		if err != nil {
+			h.logger.Debug(err.Error())
+			continue
+		}
+		digest, ok := digestValue.(types.ConsensusDigest)
 		if !ok {
-			h.logger.Debugf("digest type not supported: %T", d.Value())
+			h.logger.Debugf("digest type not supported: %T", digestValue)
 			continue
 		}
 
@@ -135,7 +140,11 @@ func checkForGRANDPAForcedChanges(digests []types.ConsensusDigest) ([]types.Cons
 			return nil, fmt.Errorf("cannot unmarshal GRANDPA consensus digest: %w", err)
 		}
 
-		switch data.Value().(type) {
+		dataValue, err := data.Value()
+		if err != nil {
+			return nil, err
+		}
+		switch dataValue.(type) {
 		case types.GrandpaScheduledChange:
 		case types.GrandpaForcedChange:
 			hasForcedChange = true
@@ -178,7 +187,11 @@ func (h *Handler) handleConsensusDigest(d *types.ConsensusDigest, header *types.
 func (h *Handler) handleBabeConsensusDigest(digest scale.VaryingDataType, header *types.Header) error {
 	headerHash := header.Hash()
 
-	switch val := digest.Value().(type) {
+	digestValue, err := digest.Value()
+	if err != nil {
+		return err
+	}
+	switch val := digestValue.(type) {
 	case types.NextEpochData:
 		currEpoch, err := h.epochState.GetEpochForBlock(header)
 		if err != nil {
--- a/dot/import.go
+++ b/dot/import.go
@@ -127,7 +127,11 @@ func newHeaderFromFile(filename string) (*types.Header, error) {
 			return nil, err
 		}
 
-		err = digest.Add(digestItem.Value())
+		digestItemVal, err := digestItem.Value()
+		if err != nil {
+			return nil, err
+		}
+		err = digest.Add(digestItemVal)
 		if err != nil {
 			return nil, err
 		}
--- a/dot/import.go
+++ b/dot/import.go
@@ -129,7 +129,7 @@ func newHeaderFromFile(filename string) (*types.Header, error) {
 
 		digestItemVal, err := digestItem.Value()
 		if err != nil {
-			return nil, err
+			return nil, fmt.Errorf("getting digest item value: %w", err)
 		}
 		err = digest.Add(digestItemVal)
 		if err != nil {
--- a/dot/state/epoch.go
+++ b/dot/state/epoch.go
@@ -202,7 +202,11 @@ func (s *EpochState) GetEpochForBlock(header *types.Header) (uint64, error) {
 	}
 
 	for _, d := range header.Digest.Types {
-		predigest, ok := d.Value().(types.PreRuntimeDigest)
+		digestValue, err := d.Value()
+		if err != nil {
+			continue
+		}
+		predigest, ok := digestValue.(types.PreRuntimeDigest)
 		if !ok {
 			continue
 		}
--- a/dot/state/grandpa.go
+++ b/dot/state/grandpa.go
@@ -84,7 +84,11 @@ func NewGrandpaState(db chaindb.Database, bs *BlockState) *GrandpaState {
 
 // HandleGRANDPADigest receives a decoded GRANDPA digest and calls the right function to handles the digest
 func (s *GrandpaState) HandleGRANDPADigest(header *types.Header, digest scale.VaryingDataType) error {
-	switch val := digest.Value().(type) {
+	digestValue, err := digest.Value()
+	if err != nil {
+		return err
+	}
+	switch val := digestValue.(type) {
 	case types.GrandpaScheduledChange:
 		return s.addScheduledChange(header, val)
 	case types.GrandpaForcedChange:
--- a/dot/state/grandpa.go
+++ b/dot/state/grandpa.go
@@ -86,7 +86,7 @@ func NewGrandpaState(db chaindb.Database, bs *BlockState) *GrandpaState {
 func (s *GrandpaState) HandleGRANDPADigest(header *types.Header, digest scale.VaryingDataType) error {
 	digestValue, err := digest.Value()
 	if err != nil {
-		return err
+		return fmt.Errorf("getting digest value: %w", err)
 	}
 	switch val := digestValue.(type) {
 	case types.GrandpaScheduledChange:
--- a/dot/types/babe.go
+++ b/dot/types/babe.go
@@ -112,9 +112,13 @@ func GetSlotFromHeader(header *Header) (uint64, error) {
 		return 0, ErrChainHeadMissingDigest
 	}
 
-	preDigest, ok := header.Digest.Types[0].Value().(PreRuntimeDigest)
+	digestValue, err := header.Digest.Types[0].Value()
+	if err != nil {
+		return 0, err
+	}
+	preDigest, ok := digestValue.(PreRuntimeDigest)
 	if !ok {
-		return 0, fmt.Errorf("%w: got %T", ErrNoFirstPreDigest, header.Digest.Types[0].Value())
+		return 0, fmt.Errorf("%w: got %T", ErrNoFirstPreDigest, digestValue)
 	}
 
 	digest, err := DecodeBabePreDigest(preDigest.Data)
@@ -145,9 +149,13 @@ func IsPrimary(header *Header) (bool, error) {
 		return false, ErrChainHeadMissingDigest
 	}
 
-	preDigest, ok := header.Digest.Types[0].Value().(PreRuntimeDigest)
+	digestValue, err := header.Digest.Types[0].Value()
+	if err != nil {
+		return false, err
+	}
+	preDigest, ok := digestValue.(PreRuntimeDigest)
 	if !ok {
-		return false, fmt.Errorf("%w: got %T", ErrNoFirstPreDigest, header.Digest.Types[0].Value())
+		return false, fmt.Errorf("%w: got %T", ErrNoFirstPreDigest, digestValue)
 	}
 
 	digest, err := DecodeBabePreDigest(preDigest.Data)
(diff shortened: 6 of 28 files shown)


pkg/scale/varying_data_type.go
```diff
@@ -66,8 +66,11 @@ func (vdt *VaryingDataType) Set(value VaryingDataTypeValue) (err error) {
 }
 
 // Value returns value stored in vdt
-func (vdt *VaryingDataType) Value() VaryingDataTypeValue {
-	return vdt.value
+func (vdt *VaryingDataType) Value() (VaryingDataTypeValue, error) {
+	if vdt.value == nil {
+		return nil, fmt.Errorf("vdt value is nil")
+	}
```
qdm12, line 72: Noob questions on that:

1. Shouldn't we have it as a panic if the value is left unset? (Is it a programming only error, or can it be caused by user input/io etc.)
2. Is having a varying data type without a value set not allowed?

pkg/scale/varying_data_type.go
```diff
@@ -66,8 +66,11 @@ func (vdt *VaryingDataType) Set(value VaryingDataTypeValue) (err error) {
 }
 
 // Value returns value stored in vdt
-func (vdt *VaryingDataType) Value() VaryingDataTypeValue {
-	return vdt.value
+func (vdt *VaryingDataType) Value() (VaryingDataTypeValue, error) {
+	if vdt.value == nil {
+		return nil, fmt.Errorf("vdt value is nil")
```
timwu20, line 71:
```suggestion
		return nil, fmt.Errorf("VaryingDataTypeValue has not been set")
```
See Ed's PR #2708, he defines sentinel errors for the scale package.  This should be a sentinel error as well.  Possibly a public one IMO.

pkg/scale/varying_data_type.go
timwu20: Unset VDTs are used when decoding.  You supply an unset VDT as a destination.

timwu20 (commented):
Can you update the README.md as well.  There's just some copy pasta'd examples that you already updated.

timwu20 (changes requested):
I meant to request changes

pkg/scale/varying_data_type.go
qdm12: Thanks! So then shouldn't the caller handle the `nil` value and return an error if it shouldn't be nil? 🤔 
Or should it never be nil anywhere? 🤔

jimjbrettj: IMO I dont think the caller should have the burden having to remember to do the nil check. As is what happened in my last PR, tim had to request changes to add the nil check. It feels more intuitive to me to return error from scale, that the caller then check because you always error check, and then handle as it sees fit

jimjbrettj:
https://github.com/ChainSafe/gossamer/pull/2852#pullrequestreview-1117520790 Updated, let me know if I missed anything

pkg/scale/errors.go
```diff
@@ -0,0 +1,10 @@
+// Copyright 2022 ChainSafe Systems (ON)
+// SPDX-License-Identifier: LGPL-3.0-only
+
+package scale
+
+import "errors"
+
+var (
+	ErrVaryingDataTypeValueNotSet = errors.New("VaryingDataTypeValue has not been set")
```
timwu20, line 9:
```suggestion
	ErrVaryingDataTypeNotSet = errors.New("VaryingDataTypeValue has not been set")
```

timwu20 (changes requested):
just the one thing, I'll pre-approve. good work!

pkg/scale/errors.go
jimjbrettj: good call!

pkg/scale/varying_data_type.go
qdm12: That also makes sense, I guess if it's a sentinel error, one could always programmatically `errors.Is` the error and allow it to be nil if needed. I'm just wondering really, I'm not really for or against one way or another, so all good. Resolving this!

dot/core/service.go
```diff
@@ -140,7 +140,11 @@ func (s *Service) HandleBlockProduced(block *types.Block, state *rtstorage.TrieS
 
 	digest := types.NewDigest()
 	for i := range block.Header.Digest.Types {
-		err := digest.Add(block.Header.Digest.Types[i].Value())
+		digestValue, err := block.Header.Digest.Types[i].Value()
+		if err != nil {
+			return err
```
qdm12, line 145: nit bunble some context and the index in the error context (only accessible at this level):

```suggestion
			return fmt.Errorf("getting value of digest type at index %d: %w", i, err)
```

dot/digest/digest.go
```diff
@@ -103,9 +103,14 @@ func (h *Handler) toConsensusDigests(scaleVaryingTypes []scale.VaryingDataType)
 	consensusDigests := make([]types.ConsensusDigest, 0, len(scaleVaryingTypes))
 
 	for _, d := range scaleVaryingTypes {
-		digest, ok := d.Value().(types.ConsensusDigest)
+		digestValue, err := d.Value()
+		if err != nil {
+			h.logger.Debug(err.Error())
```
qdm12, line 108: Shouldn't we log this at the error level? Or even panic? :thinking:

dot/digest/digest.go
```diff
@@ -103,9 +103,14 @@ func (h *Handler) toConsensusDigests(scaleVaryingTypes []scale.VaryingDataType)
...
+			continue
+		}
+		digest, ok := digestValue.(types.ConsensusDigest)
 		if !ok {
-			h.logger.Debugf("digest type not supported: %T", d.Value())
+			h.logger.Debugf("digest type not supported: %T", digestValue)
```
qdm12, line 113: nit a tiny bit out of scope: shouldn't we log this at the error level? Or even panic? :thinking:

dot/digest/digest.go
```diff
@@ -135,7 +140,11 @@ func checkForGRANDPAForcedChanges(digests []types.ConsensusDigest) ([]types.Cons
 			return nil, fmt.Errorf("cannot unmarshal GRANDPA consensus digest: %w", err)
 		}
 
-		switch data.Value().(type) {
+		dataValue, err := data.Value()
+		if err != nil {
+			return nil, err
```
qdm12, line 145: nit bundling the digest index in the error context would be nice

dot/digest/digest.go
```diff
@@ -178,7 +187,11 @@ func (h *Handler) handleConsensusDigest(d *types.ConsensusDigest, header *types.
 func (h *Handler) handleBabeConsensusDigest(digest scale.VaryingDataType, header *types.Header) error {
 	headerHash := header.Hash()
 
-	switch val := digest.Value().(type) {
+	digestValue, err := digest.Value()
+	if err != nil {
+		return err
```
qdm12, line 192: nit

```suggestion
		return fmt.Errorf("getting digest value: %w", err)
```

dot/import.go
```diff
@@ -127,7 +127,11 @@ func newHeaderFromFile(filename string) (*types.Header, error) {
 			return nil, err
 		}
 
-		err = digest.Add(digestItem.Value())
+		digestItemVal, err := digestItem.Value()
+		if err != nil {
+			return nil, err
```
qdm12, line 132: nit (also adding the log index would be nice)

```suggestion
			return nil, fmt.Errorf("getting digest item value: %w", err)
```

dot/state/epoch.go
```diff
@@ -202,7 +202,11 @@ func (s *EpochState) GetEpochForBlock(header *types.Header) (uint64, error) {
 	}
 
 	for _, d := range header.Digest.Types {
-		predigest, ok := d.Value().(types.PreRuntimeDigest)
+		digestValue, err := d.Value()
+		if err != nil {
+			continue
```
qdm12, line 207: Should we panic here? :thinking:

dot/state/grandpa.go
```diff
@@ -84,7 +84,11 @@ func NewGrandpaState(db chaindb.Database, bs *BlockState) *GrandpaState {
 
 // HandleGRANDPADigest receives a decoded GRANDPA digest and calls the right function to handles the digest
 func (s *GrandpaState) HandleGRANDPADigest(header *types.Header, digest scale.VaryingDataType) error {
-	switch val := digest.Value().(type) {
+	digestValue, err := digest.Value()
+	if err != nil {
+		return err
```
qdm12, line 89:
```suggestion
		return fmt.Errorf("getting digest value: %w", err)
```

dot/types/babe.go
```diff
@@ -112,9 +112,13 @@ func GetSlotFromHeader(header *Header) (uint64, error) {
 		return 0, ErrChainHeadMissingDigest
 	}
 
-	preDigest, ok := header.Digest.Types[0].Value().(PreRuntimeDigest)
+	digestValue, err := header.Digest.Types[0].Value()
+	if err != nil {
+		return 0, err
```
qdm12, line 117: nit

```suggestion
		return 0, fmt.Errorf("getting first digest type value: %w", err)
```

dot/types/babe.go
```diff
@@ -145,9 +149,13 @@ func IsPrimary(header *Header) (bool, error) {
 		return false, ErrChainHeadMissingDigest
 	}
 
-	preDigest, ok := header.Digest.Types[0].Value().(PreRuntimeDigest)
+	digestValue, err := header.Digest.Types[0].Value()
+	if err != nil {
+		return false, err
```
qdm12, line 154:
```suggestion
		return false, fmt.Errorf("getting first digest type value: %w", err)
```

dot/types/babe_digest.go
```diff
@@ -24,7 +24,11 @@ func DecodeBabePreDigest(in []byte) (scale.VaryingDataTypeValue, error) {
 		return nil, err
 	}
 
-	switch msg := babeDigest.Value().(type) {
+	babeDigestValue, err := babeDigest.Value()
+	if err != nil {
+		return nil, err
```
qdm12, line 29: nit

```suggestion
		return nil, fmt.Errorf("getting babe digest value: %w", err)
```

dot/types/header.go
```diff
@@ -68,7 +68,11 @@ func (bh *Header) DeepCopy() (*Header, error) {
 	if len(bh.Digest.Types) > 0 {
 		cp.Digest = NewDigest()
 		for _, d := range bh.Digest.Types {
-			err := cp.Digest.Add(d.Value())
+			digestValue, err := d.Value()
+			if err != nil {
+				return nil, err
```
qdm12, line 73: nit (also adding the digest type index would be nice)

```suggestion
				return nil, fmt.Errorf("getting digest type value: %w", err)
```

lib/babe/errors.go
```diff
@@ -233,7 +233,11 @@ type MandatoryDispatch struct{}
 func (MandatoryDispatch) Index() uint { return 9 }
 
 func determineErrType(vdt scale.VaryingDataType) error {
-	switch val := vdt.Value().(type) {
+	vdtVal, err := vdt.Value()
+	if err != nil {
+		return err
```
qdm12, line 238: nit

```suggestion
		return fmt.Errorf("getting vdt value: %w", err)
```

lib/babe/verify.go
```diff
@@ -245,14 +246,22 @@ func (b *verifier) verifyAuthorshipRight(header *types.Header) error {
 	preDigestItem := header.Digest.Types[0]
 	sealItem := header.Digest.Types[len(header.Digest.Types)-1]
 
-	preDigest, ok := preDigestItem.Value().(types.PreRuntimeDigest)
+	preDigestItemValue, err := preDigestItem.Value()
+	if err != nil {
+		return err
```
qdm12, line 251: nit

```suggestion
		return fmt.Errorf("getting pre digest item value: %w", err)
```

lib/babe/verify.go
```diff
@@ -245,14 +246,22 @@ func (b *verifier) verifyAuthorshipRight(header *types.Header) error {
...
+	preDigest, ok := preDigestItemValue.(types.PreRuntimeDigest)
 	if !ok {
-		return fmt.Errorf("%w: got %T", types.ErrNoFirstPreDigest, preDigestItem.Value())
+		return fmt.Errorf("%w: got %T", types.ErrNoFirstPreDigest, preDigestItemValue)
 	}
 
-	seal, ok := sealItem.Value().(types.SealDigest)
+	sealItemValue, err := sealItem.Value()
+	if err != nil {
+		return err
```
qdm12, line 260: nit

```suggestion
		return fmt.Errorf("getting seal item value: %w", err)
```

lib/babe/verify.go
```diff
@@ -277,15 +286,23 @@ func (b *verifier) verifyAuthorshipRight(header *types.Header) error {
 	// remove seal before verifying signature
 	h := types.NewDigest()
 	for _, val := range header.Digest.Types[:len(header.Digest.Types)-1] {
-		err = h.Add(val.Value())
+		digestValue, err := val.Value()
+		if err != nil {
+			return err
```
qdm12, line 291: nit (also adding index of digest type to error context would be nice)

```suggestion
			return fmt.Errorf("getting digest type value: %w", err)
```

lib/babe/verify.go
```diff
@@ -329,10 +346,13 @@ func (b *verifier) verifyAuthorshipRight(header *types.Header) error {
 			return fmt.Errorf("current header missing digest")
 		}
 
-		currentPreDigestItem := currentHeader.Digest.Types[0]
-		currentPreDigest, ok := currentPreDigestItem.Value().(types.PreRuntimeDigest)
+		currentPreDigestItemValue, err := currentHeader.Digest.Types[0].Value()
+		if err != nil {
+			return err
```
qdm12, line 351:
```suggestion
			return fmt.Errorf("getting current header first digest type value: %w", err)
```

lib/babe/verify.go
```diff
@@ -476,7 +496,11 @@ func getAuthorityIndex(header *types.Header) (uint32, error) {
 		return 0, fmt.Errorf("no digest provided")
 	}
 
-	preDigest, ok := header.Digest.Types[0].Value().(types.PreRuntimeDigest)
+	digestValue, err := header.Digest.Types[0].Value()
+	if err != nil {
+		return 0, err
```
qdm12, line 501:
```suggestion
		return 0, fmt.Errorf("getting first digest type value: %w", err)
```

lib/grandpa/network.go
```diff
@@ -217,13 +217,21 @@ func decodeMessage(cm *network.ConsensusMessage) (m GrandpaMessage, err error) {
 		return nil, err
 	}
 
-	switch val := msg.Value().(type) {
+	msgValue, err := msg.Value()
+	if err != nil {
+		return nil, err
```
qdm12, line 222: nit

```suggestion
		return nil, fmt.Errorf("getting message value: %w", err)
```

lib/grandpa/network.go
```diff
@@ -217,13 +217,21 @@ func decodeMessage(cm *network.ConsensusMessage) (m GrandpaMessage, err error) {
...
+	switch val := msgValue.(type) {
 	case VoteMessage:
 		m = &val
 	case CommitMessage:
 		m = &val
 	case VersionedNeighbourPacket:
-		switch NeighbourMessage := val.Value().(type) {
+		neighbourMessageVal, err := val.Value()
+		if err != nil {
+			return nil, err
```
qdm12, line 232:
```suggestion
			return nil, fmt.Errorf("getting versioned neighbour packet value: %w", err)
```

lib/runtime/invalid_transaction.go
```diff
@@ -29,16 +29,16 @@ func (i *InvalidTransaction) Set(val scale.VaryingDataTypeValue) (err error) {
...
 }
 
 // Error returns the error message associated with the InvalidTransaction
 func (i InvalidTransaction) Error() string {
-	value := i.Value()
-	if value == nil {
-		return "invalidTransaction hasn't been set"
+	value, err := i.Value()
+	if err != nil {
+		return err.Error()
```
qdm12, line 41: nit

```suggestion
		return fmt.Sprintf("getting invalid transaction value: %s", err)
```

lib/runtime/transaction_validity.go
```diff
@@ -30,17 +30,18 @@ func (tve *TransactionValidityError) Set(val scale.VaryingDataTypeValue) (err er
...
 	return vdt.Value()
 }
 
 // Error will return the error underlying TransactionValidityError
 func (tve TransactionValidityError) Error() string {
-	if tve.Value() == nil {
-		return "unset TransactionValidityError"
+	value, err := tve.Value()
+	if err != nil {
+		return err.Error()
```
qdm12, line 42: nit

```suggestion
		return fmt.Sprintf("getting transaction validity error value: %s", err)
```

lib/runtime/transaction_validity.go
```diff
@@ -78,14 +79,18 @@ func UnmarshalTransactionValidity(res []byte) (*transaction.Validity, error) {
 			panic(fmt.Sprintf("%s: %T", errInvalidTypeCast, scaleWrappedErr.Err))
 		}
 
-		switch val := txnValidityErr.Value().(type) {
+		txnValidityErrValue, err := txnValidityErr.Value()
+		if err != nil {
+			return nil, err
```
qdm12, line 84: nit

```suggestion
			return nil, fmt.Errorf("getting transaction validity error value: %w", err)
```

lib/runtime/unknown_transaction.go
```diff
@@ -29,15 +29,15 @@ func (u *UnknownTransaction) Set(val scale.VaryingDataTypeValue) (err error) {
...
 	return vdt.Value()
 }
 
 func (u UnknownTransaction) Error() string {
-	value := u.Value()
-	if value == nil {
-		return "unknownTransaction hasn't been set"
+	value, err := u.Value()
+	if err != nil {
+		return err.Error()
```
qdm12, line 40: nit

```suggestion
		return fmt.Sprintf("getting unknown transaction value: %s", err)
```

lib/runtime/wasmer/exports.go
```diff
@@ -135,11 +135,15 @@ func (in *Instance) ExecuteBlock(block *types.Block) ([]byte, error) {
 
 	// remove seal digest only
 	for _, d := range block.Header.Digest.Types {
-		switch d.Value().(type) {
+		digestValue, err := d.Value()
+		if err != nil {
+			return nil, err
```
qdm12, line 140: nit (also adding the digest type index to the error context would be nice)

```suggestion
			return nil, fmt.Errorf("getting digest type value: %w", err)
```

pkg/scale/varying_data_type.go
```diff
@@ -66,8 +66,11 @@ func (vdt *VaryingDataType) Set(value VaryingDataTypeValue) (err error) {
 }
 
 // Value returns value stored in vdt
-func (vdt *VaryingDataType) Value() VaryingDataTypeValue {
-	return vdt.value
+func (vdt *VaryingDataType) Value() (VaryingDataTypeValue, error) {
+	if vdt.value == nil {
+		return nil, ErrVaryingDataTypeNotSet
```
qdm12, line 71: nit to force using errors.Is instead of `==`

```suggestion
		return nil, fmt.Errorf("%w", ErrVaryingDataTypeNotSet)
```

tests/utils/rpc/request.go
```diff
@@ -113,7 +113,11 @@ func rpcLogsToDigest(logs []string) (digest scale.VaryingDataTypeSlice, err erro
 			return digest, fmt.Errorf("malformed digest item bytes: %w", err)
 		}
 
-		err = digest.Add(di.Value())
+		digestValue, err := di.Value()
+		if err != nil {
+			return digest, err
```
qdm12, line 118: nit

```suggestion
			return digest, fmt.Errorf("getting digest item value: %w", err)
```

qdm12 (approved):
Great work :medal_military: 

*Just a few nits about error wrappings that you can batch commit*

pkg/scale/errors.go
qdm12: I got quite confused with the the chicken or egg  debate.... is the varying data type not set or the varying data type value not set in the varying data type :rofl: Anyway that was a fun observation

dot/digest/digest.go
jimjbrettj: updated!

dot/state/epoch.go
jimjbrettj: I don't think so. This function iterates through all digests to check for the preruntime digest. I think technically this should be the first digest anyways, but possibly there is a case where there are other empty digests that we check before preruntime digest.

Even if not, the current logic is that if we don't find the preruntime digest we return error and I'm not sure why we would change that in this case.

timwu20 (approved):

PR #2864, opened 2022-09-30
scale support for map

### LINKED ISSUE #2796: Support encoding and decoding of maps in scale
## Issue summary

- Currently scale package cannot encode or decode maps, it a type that is not supported.
- inherents data is kept as a map. So we need to support maps in scale package.
- Currently, we have `InherentsData.Encode()` which sort of implements map encoding for scale. But I think that code actually belongs in scale package.
- Once we have support for maps in scale, we wouldn't need `InherentsData.Encode()` as we would be able to directly do `scale.Marshal(InherentsData)` ( which is also what substrate does https://github.com/paritytech/substrate/blob/e986f05062a5c7a8608967febcd5c1d2bb02eb40/primitives/inherents/src/lib.rs#L424)

- So, for this we have to implement `reflect.Map` case in the scale.marshal method https://github.com/ChainSafe/gossamer/blob/b98c45aa070d57a59799db4ef4ebeb571a6dadc4/pkg/scale/encode.go#L55

- InherentsData.Encode() https://github.com/ChainSafe/gossamer/blob/b98c45aa070d57a59799db4ef4ebeb571a6dadc4/dot/types/inherents.go#L68

kishansagathiya:
## Tests

```sh
```

## Issues
Fixes #2796

### DIFF
--- a/pkg/scale/encode.go
+++ b/pkg/scale/encode.go
@@ -106,6 +106,8 @@ func (es *encodeState) marshal(in interface{}) (err error) {
 			err = es.encodeArray(in)
 		case reflect.Slice:
 			err = es.encodeSlice(in)
+		case reflect.Map:
+			err = es.encodeMap(in)
 		default:
 			err = fmt.Errorf("unsupported type: %T", in)
 		}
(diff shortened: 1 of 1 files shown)


kishansagathiya:
scale has map support now.

PR #2865, opened 2022-09-30
remove `digest type not supported` error

kishansagathiya:
This is not an error. We are iterating over all digests and picking consensus digests. It is completely fine to see other types of digests.

## Tests

```sh
```

## Primary Reviewer

@timwu20

### DIFF
--- a/dot/digest/digest.go
+++ b/dot/digest/digest.go
@@ -110,7 +110,6 @@ func (h *Handler) toConsensusDigests(scaleVaryingTypes []scale.VaryingDataType) 
 		}
 		digest, ok := digestValue.(types.ConsensusDigest)
 		if !ok {
-			h.logger.Errorf("digest type not supported: %T", digestValue)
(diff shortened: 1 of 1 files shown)


qdm12 (approved):
:+1:

dot/digest/digest.go
```diff
@@ -110,7 +110,6 @@ func (h *Handler) toConsensusDigests(scaleVaryingTypes []scale.VaryingDataType)
 		}
 		digest, ok := digestValue.(types.ConsensusDigest)
 		if !ok {
-			h.logger.Errorf("digest type not supported: %T", digestValue)
```
qdm12, line 113: Actually... ideally... what you could do is a type `switch` with all digest type supported, and panic on the default case. That way we cannot forget to add a new digest type in the future. But really nitpicking I guess.

timwu20 (approved):

Issue #2854, opened 2022-09-22
Faster online pruning with cached trie inserted/modified merkle values


qdm12:
The online pruning finds inserted/modified nodes by going through the entire trie and checking each node for the dirty flag.

We can change that to have each trie cache merkle values (database keys) that were inserted/modified,  so it should be much faster.

This should be done after #2835

qdm12:
When inserting/updating a node in the trie, all the path of nodes from the root to that new or updated node is modified. In other words, for a node at depth N in the trie, N nodes are modified. That means for every `Put` operation, there are N node encoding and up to N blake2b hashing. As a consequence, performance is severely degraded, so tracking updates in the trie is not feasible. The current walk of the trie will be kept instead, especially since it would be needed to implement the 'write cache' for lazy loading (walk the trie and only write updated nodes since the last snapshot).

PR #2871, opened 2022-10-04
fix(trie): `PopulateMerkleValues` behavior

qdm12:
## Changes

- Include Merkle value of the parent node passed as argument
- Update comment all nodes are expected to have their Merkle value computed
- Enforce all nodes to have their Merkle value computed with a `panic` (this will be thrown out with lazy loading)
- Add unit test

For now this only affects the offline pruning, but this function is to be used in another PR to fix the tracking of deleted Merkle values and add cached tracking of inserted Merkle values

## Tests

```sh
go test github.com/ChainSafe/gossamer/lib/trie/...
```

## Issues

Related to #2854 

## Primary Reviewer

@kishansagathiya

### DIFF
--- a/lib/trie/database.go
+++ b/lib/trie/database.go
@@ -186,7 +186,7 @@ func (t *Trie) loadNode(db Database, n *Node) error {
 }
 
 // PopulateMerkleValues writes the Merkle values of the node given and of
-// all its descendant nodes as keys to the map merkleValues.
+// all its descendant nodes as keys to the merkleValues map.
 // It is assumed the node and its descendant nodes have their Merkle value already
 // computed.
 func PopulateMerkleValues(n *Node, merkleValues map[string]struct{}) {
--- a/lib/trie/database.go
+++ b/lib/trie/database.go
@@ -187,21 +187,44 @@ func (t *Trie) loadNode(db Database, n *Node) error {
 
 // PopulateMerkleValues writes the Merkle value of each children of the node given
 // as keys to the map merkleValues.
-func (t *Trie) PopulateMerkleValues(n *Node, merkleValues map[string]struct{}) {
-	if n.Kind() != node.Branch {
-		return
+func (t *Trie) PopulateMerkleValues(n *Node,
+	merkleValues map[string]struct{}) (err error) {
+	if n == nil {
+		return nil
+	}
+
+	merkleValue := n.MerkleValue
+	if len(merkleValue) == 0 {
+		// Compute and cache node Merkle value if it is absent.
+		if n == t.root {
+			merkleValue, err = n.CalculateRootMerkleValue()
+			if err != nil {
+				return fmt.Errorf("calculating Merkle value for root node: %w", err)
+			}
+		} else {
+			merkleValue, err = n.CalculateMerkleValue()
+			if err != nil {
+				return fmt.Errorf("calculating Merkle value for node: %w", err)
+			}
+		}
+	}
+
+	merkleValues[string(merkleValue)] = struct{}{}
+
+	if n.Kind() == node.Leaf {
+		return nil
 	}
 
 	branch := n
 	for _, child := range branch.Children {
-		if child == nil {
-			continue
+		err = t.PopulateMerkleValues(child, merkleValues)
+		if err != nil {
+			// Note: do not wrap error since this is recursive.
+			return err
 		}
-
-		merkleValues[string(child.MerkleValue)] = struct{}{}
-
-		t.PopulateMerkleValues(child, merkleValues)
 	}
+
+	return nil
 }
 
 // GetFromDB retrieves a value at the given key from the trie using the database.
--- a/lib/trie/database.go
+++ b/lib/trie/database.go
@@ -185,46 +185,29 @@ func (t *Trie) loadNode(db Database, n *Node) error {
 	return nil
 }
 
-// PopulateMerkleValues writes the Merkle value of each children of the node given
-// as keys to the map merkleValues.
-func (t *Trie) PopulateMerkleValues(n *Node,
-	merkleValues map[string]struct{}) (err error) {
+// PopulateMerkleValues writes the Merkle values of the node given and of
+// all its descendant nodes as keys to the map merkleValues.
+// It is assumed the node and its descendant nodes have their Merkle value already
+// computed.
+func PopulateMerkleValues(n *Node, merkleValues map[string]struct{}) {
 	if n == nil {
-		return nil
+		return
 	}
 
-	merkleValue := n.MerkleValue
-	if len(merkleValue) == 0 {
-		// Compute and cache node Merkle value if it is absent.
-		if n == t.root {
-			merkleValue, err = n.CalculateRootMerkleValue()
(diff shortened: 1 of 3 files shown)


EclesioMeloJunior (approved):

lib/trie/database.go
```diff
@@ -185,22 +185,28 @@ func (t *Trie) loadNode(db Database, n *Node) error {
 	return nil
 }
 
-// PopulateMerkleValues writes the Merkle value of each children of the node given
-// as keys to the map merkleValues.
-func (t *Trie) PopulateMerkleValues(n *Node, merkleValues map[string]struct{}) {
-	if n.Kind() != node.Branch {
+// PopulateMerkleValues writes the Merkle values of the node given and of
+// all its descendant nodes as keys to the map merkleValues.
+// It is assumed the node and its descendant nodes have their Merkle value already
```
jimjbrettj, line 190: Is this assumption reenforced elsewhere in the code?

lib/trie/database.go
```diff
@@ -185,22 +185,28 @@ func (t *Trie) loadNode(db Database, n *Node) error {
...
+	if n == nil {
 		return
 	}
 
-	branch := n
-	for _, child := range branch.Children {
-		if child == nil {
-			continue
-		}
+	if len(n.MerkleValue) == 0 {
```
jimjbrettj, line 197: is n.MerkleValue always created/checked for validity before this call? Basically want to make sure that user input cannot cause this function to panic

jimjbrettj (commented):
few comments

lib/trie/database.go
qdm12: Good point;

If this function was previously working and ignoring non-pre-calculated Merkle values, it was just plain wrong and dangerous, so we should panic.

Now practically speaking:

1. This is only used in the offline pruner in this particular git branch, so it's really not-a-big-deal, and it would work fine (due to LoadFromDB caching Merkle values for each trie node loaded)
2. #2873 (based on this branch) uses `PopulateMerkleValues` within the trie code to track deleted merkle values, so it becomes more critical. That's why I added https://github.com/ChainSafe/gossamer/pull/2873/files#diff-33f1852514318cc87b4ec66a5c67372b8161934cc6aa5c89cae7fbb09d2dfadfR1348 to enforce all Merkle values are computed before calling `PopulateMerkleValues`. (read 3. as well) It may or may not happen that all Merkle values are computed.
3. Lazy loading will load each trie node from their Merkle value... so really it will always be there, and that 'caching of Merkle value enforcement' won't be needed anymore. That is also partly why the enforcement is outside this `PopulateMerkleValues` function.

lib/trie/database.go
qdm12: See https://github.com/ChainSafe/gossamer/pull/2871/files#r992189720

User input for now is only the offline pruner which guarantees all nodes have their cached Merkle value set.

lib/trie/database.go
```diff
@@ -185,22 +185,28 @@ func (t *Trie) loadNode(db Database, n *Node) error {
...
 
-// PopulateMerkleValues writes the Merkle value of each children of the node given
-// as keys to the map merkleValues.
-func (t *Trie) PopulateMerkleValues(n *Node, merkleValues map[string]struct{}) {
-	if n.Kind() != node.Branch {
+// PopulateMerkleValues writes the Merkle values of the node given and of
+// all its descendant nodes as keys to the map merkleValues.
+// It is assumed the node and its descendant nodes have their Merkle value already
+// computed.
+func PopulateMerkleValues(n *Node, merkleValues map[string]struct{}) {
```
timwu20, line 192: would there be any value in storing the pointer to the `Node` in the map? `map[string]*Node`

qdm12: Nope :wink: 

This is (and will be) 100% only used for pruning which only cares about deleted and inserted trie node Merkle values (well, actually, I recently realized it's **trie node hashes**, since < 32B merkle values are inlined in parent node), which are the chaindb database keys (already now, and in the future for lazy loading).

qdm12:
@timwu20: @kishansagathiya raised a good question I'm not 100% confident in answering; when pruning, do we want to keep the top trie node (root hash <-> root node encoding) for blocks past our retained blocks setting? I would think we want to prune the block fully, but maybe there is a reason to keep the root node hash <-> root node encoding (no longer useful really)?

EclesioMeloJunior:
> @timwu20: @kishansagathiya raised a good question I'm not 100% confident in answering; when pruning, do we want to keep the root trie node (root hash <-> root node encoding) for blocks past our retained blocks setting? I would think we want to prune the block fully, but maybe there is a reason to keep the root node hash <-> root node encoding (no longer useful really)?

@qdm12 I don't know if this answers your question but: the root hash is a field in the block header `StateRootHash` so if we prune the entire block storing its contents in the database it is possible to get this information if needed so I don't see why we should keep the root trie node

kishansagathiya (commented):
It looks like this PR should be made to https://github.com/ChainSafe/gossamer/pull/2873 than in development, since that's where it is needed.

lib/trie/database.go
```diff
@@ -185,22 +185,28 @@ func (t *Trie) loadNode(db Database, n *Node) error {
 	return nil
 }
 
-// PopulateMerkleValues writes the Merkle value of each children of the node given
-// as keys to the map merkleValues.
-func (t *Trie) PopulateMerkleValues(n *Node, merkleValues map[string]struct{}) {
-	if n.Kind() != node.Branch {
+// PopulateMerkleValues writes the Merkle values of the node given and of
```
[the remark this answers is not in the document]
qdm12: You can say both, `node given` and `given node`... I'm not aware of a difference :thinking:

lib/trie/database.go
```diff
@@ -185,22 +185,28 @@ func (t *Trie) loadNode(db Database, n *Node) error {
...
 		return
 	}
 
-	branch := n
-	for _, child := range branch.Children {
-		if child == nil {
-			continue
-		}
+	if len(n.MerkleValue) == 0 {
+		panic(fmt.Sprintf("node with key 0x%x has no Merkle value computed", n.Key))
```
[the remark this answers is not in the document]
qdm12: We do want to panic, it would be a programming error (and catastrophic consequences) if this ever happens. This is essentially a guard against my changes on other branches based on this one (since it's a lot of mental overhead to think about everything, the trie code is quite hard). This will be removed soon enough with lazy loading since all Merkle values will be 100% computed anyway, so that panic will be removed.
I'll add a `// TODO` comment linking the lazy loading issue.

qdm12:
> It looks like this PR should be made to #2873 than in development, since that's where it is needed.

I initially spotted it wouldn't include the root node by using it in #2873 and thought it was a bug as I assumed the root node should also be pruned in the offline pruner.

From today's standup & Eclesio's response, it seems it is indeed a bug and the root node should be pruned as well as its descendants, so I'd say it's still fine to have it separate. Also #2873 is rather thick so breaking parts away from it whenever possible is nice.

kishansagathiya (approved):

Issue #2838, opened 2022-09-15
trie: implement lazy loading of trie nodes from disk


qdm12:
Our memory usage is high since we keep all trie nodes in memory. 

Instead we should (like Substrate) store all `node hash -> node encoding` on disk, and lazy load them as we need to access the trie (both read and write). Only the root hash should be kept in memory, and every node should be retrieved from disk as we go deeper in the trie (to insert or retrieve).

We should also implement pruning for nodes inserted/modified in uncle blocks (and make sure the finalized block doesn't contain them).

We can also implement a cache to keep the most frequently used nodes in memory and prevent decoding them again

Related with #2835 and #2836

dimartiro:
I found that Ethereum is also using a lazy loading strategy with the `node keys -> encoded nodes` 

https://github.com/ethereum/go-ethereum/blob/d76efbb9be403689bc8d677a16af453728a7a513/trie/triedb/hashdb/database.go#L186-L224

dimartiro:
In substrate, they are using this [trie crate ](https://github.com/paritytech/trie) for all trie operations, and they created a [TrieDB structure backed by a DB](https://github.com/paritytech/trie/blob/6eb4b88e0781da2db7dda0021249a8fa4244bc5d/trie-db/src/triedb.rs#L120-L128) with a cache, and they are implementing the [lazy loading there](https://github.com/paritytech/trie/blob/6eb4b88e0781da2db7dda0021249a8fa4244bc5d/trie-db/src/lookup.rs#L489-L570)

dimartiro:
This change will help us to improve our memory usage and our node startup time when we are using an snapshot, also improving our debugging time when we are trying to sync to the tip of the chain and we need to use snapshots for that.

dimartiro:
This issue will be Done when #3805 is completed

P1sar:
Design Design considered complete

Issue #2669, opened 2022-07-13
Gossamer Should Propogate Panics from the Runtime


danforbes:
Currently, Gossamer just logs errors that occur in the runtime - instead, when a Gossamer node experiences a runtime panic, the node itself should also panic.

EclesioMeloJunior:
@timwu20 I think we can close this one since wazero propagates panics as errors to the host layer so this is issue is actually done

PR #2873, opened 2022-10-05
fix(lib/trie): record deleted Merkle values fixed

qdm12:
## Changes

- Record Merkle values deleted only if they are NOT 'dirty' (the same as in the last trie snapshot)
- Record Merkle values deleted when one or more nodes are deleted from the trie
- Do NOT record Merkle values of inlined nodes in the deleted set of Merkle values
- Lazy calculate Merkle values if they are missing (especially for delete operations)
- Update tests to have full trie.go coverage again (except hashing error due to the buffer sync pools at global scope)
- Fix preparing branch for mutation only once before for loop (instead of N times)
- Check errors from merkle value computation (useful for lazy loading to handle database errors)

Eventually add more test cases to test for correct registration of deleted Merkle values, but meh might not be that useful with incoming lazy loading changes

## Tests

```sh
```

## Issues

Blocking ~#2854 and~ #2838 

## Primary Reviewer

@timwu20

### DIFF
--- a/lib/runtime/storage/trie.go
+++ b/lib/runtime/storage/trie.go
@@ -108,7 +108,7 @@ func (s *TrieState) Delete(key []byte) (err error) {
 	defer s.lock.Unlock()
 	err = s.t.Delete(key)
 	if err != nil {
-		return err
+		return fmt.Errorf("deleting from trie: %w", err)
 	}
 
 	return nil
--- a/lib/runtime/storage/trie.go
+++ b/lib/runtime/storage/trie.go
@@ -210,6 +210,7 @@ func (s *TrieState) DeleteChildLimit(key []byte, limit *[]byte) (
 		// If one deletion fails, the child trie and its parent trie are then in
 		// a bad intermediary state. Take also care of the caching of deleted Merkle
 		// values within the tries, which is used for online pruning.
+		// See https://github.com/ChainSafe/gossamer/issues/3032
 		err = tr.Delete([]byte(k))
 		if err != nil {
 			return deleted, allDeleted, fmt.Errorf("deleting from child trie located at key 0x%x: %w", key, err)
--- a/lib/runtime/storage/trie.go
+++ b/lib/runtime/storage/trie.go
@@ -5,6 +5,7 @@ package storage
 
 import (
 	"encoding/binary"
+	"fmt"
 	"sort"
 	"sync"
 
@@ -67,11 +68,11 @@ func (s *TrieState) RollbackStorageTransaction() {
 	s.oldTrie = nil
 }
 
-// Put puts thread safely a value at the specified key in the trie.
-func (s *TrieState) Put(key, value []byte) {
+// Put puts a key-value pair in the trie
+func (s *TrieState) Put(key, value []byte) (err error) {
 	s.lock.Lock()
 	defer s.lock.Unlock()
-	s.t.Put(key, value)
+	return s.t.Put(key, value)
 }
 
 // Get gets a value from the trie
@@ -97,15 +98,20 @@ func (s *TrieState) Has(key []byte) bool {
 }
 
 // Delete deletes a key from the trie
-func (s *TrieState) Delete(key []byte) {
+func (s *TrieState) Delete(key []byte) (err error) {
 	val := s.t.Get(key)
 	if val == nil {
-		return
+		return nil
 	}
 
 	s.lock.Lock()
 	defer s.lock.Unlock()
-	s.t.Delete(key)
+	err = s.t.Delete(key)
+	if err != nil {
+		return err
+	}
+
+	return nil
 }
 
 // NextKey returns the next key in the trie in lexicographical order. If it does not exist, it returns nil.
@@ -116,19 +122,19 @@ func (s *TrieState) NextKey(key []byte) []byte {
 }
 
 // ClearPrefix deletes all key-value pairs from the trie where the key starts with the given prefix
-func (s *TrieState) ClearPrefix(prefix []byte) {
+func (s *TrieState) ClearPrefix(prefix []byte) (err error) {
 	s.lock.Lock()
 	defer s.lock.Unlock()
-	s.t.ClearPrefix(prefix)
+	return s.t.ClearPrefix(prefix)
 }
 
 // ClearPrefixLimit deletes key-value pairs from the trie where the key starts with the given prefix till limit reached
-func (s *TrieState) ClearPrefixLimit(prefix []byte, limit uint32) (uint32, bool) {
+func (s *TrieState) ClearPrefixLimit(prefix []byte, limit uint32) (
+	deleted uint32, allDeleted bool, err error) {
 	s.lock.Lock()
 	defer s.lock.Unlock()
 
-	num, del := s.t.ClearPrefixLimit(prefix, limit)
-	return num, del
+	return s.t.ClearPrefixLimit(prefix, limit)
 }
 
 // TrieEntries returns every key-value pair in the trie
@@ -167,24 +173,29 @@ func (s *TrieState) GetChildStorage(keyToChild, key []byte) ([]byte, error) {
 }
 
 // DeleteChild deletes a child trie from the main trie
-func (s *TrieState) DeleteChild(key []byte) {
+func (s *TrieState) DeleteChild(key []byte) (err error) {
 	s.lock.Lock()
 	defer s.lock.Unlock()
-	s.t.DeleteChild(key)
+	return s.t.DeleteChild(key)
 }
 
-// DeleteChildLimit deletes up to limit of database entries by lexicographic order, return number
-//  deleted, true if all delete otherwise false
-func (s *TrieState) DeleteChildLimit(key []byte, limit *[]byte) (uint32, bool, error) {
+// DeleteChildLimit deletes up to limit of database entries by lexicographic order.
+func (s *TrieState) DeleteChildLimit(key []byte, limit *[]byte) (
+	deleted uint32, allDeleted bool, err error) {
 	s.lock.Lock()
 	defer s.lock.Unlock()
 	tr, err := s.t.GetChild(key)
 	if err != nil {
 		return 0, false, err
 	}
+
 	qtyEntries := uint32(len(tr.Entries()))
 	if limit == nil {
-		s.t.DeleteChild(key)
+		err = s.t.DeleteChild(key)
+		if err != nil {
+			return 0, false, err
+		}
+
 		return qtyEntries, true, nil
 	}
 	limitUint := binary.LittleEndian.Uint32(*limit)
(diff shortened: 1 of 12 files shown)


lib/trie/trie.go
```diff
@@ -80,52 +82,50 @@ func (t *Trie) handleTrackedDeltas(success bool, pendingDeletedMerkleValues map[
...
-// updateGeneration is called when the currentNode is from
-// an older trie generation (snapshot) so we deep copy the
-// node and update the generation on the newer copy.
-func updateGeneration(currentNode *Node, trieGeneration uint64,
-	deletedMerkleValues map[string]struct{}, copySettings node.CopySettings) (
-	newNode *Node) {
-	newNode = currentNode.Copy(copySettings)
-	newNode.Generation = trieGeneration
+func registerDeletedMerkleValue(node *Node, isRoot bool,
+	pendingDeletedMerkleValues map[string]struct{}) (err error) {
```
timwu20, line 109: consider defining a type for `map[string]struct{}`? maybe something like `type mappedMerklesValues map[string]struct{}`

lib/trie/trie.go
```diff
@@ -80,52 +82,50 @@ func (t *Trie) handleTrackedDeltas(success bool, pendingDeletedMerkleValues map[
...
-	// The hash of the node from a previous snapshotted trie
-	// is usually already computed.
-	deletedMerkleValue := currentNode.MerkleValue
-	if len(deletedMerkleValue) > 0 {
-		deletedMerkleValueString := string(deletedMerkleValue)
-		deletedMerkleValues[deletedMerkleValueString] = struct{}{}
+	if !node.Dirty {
+		// Only register deleted nodes that were not previously modified
+		// since the last trie snapshot.
+		pendingDeletedMerkleValues[string(node.MerkleValue)] = struct{}{}
```
timwu20, line 125: why don't we just add this key directly to `t.deletedMerkleValues`? I don't see the reason we are passing around this map of `deletedMerkleValues` which will eventually be used to update `t.deletedMerkleValues` in `t.handleTrackedDeltas`.

qdm12: This was implemented in #2876 in order to not have tracked trie deltas in a bad state in case of error during a single trie operation (method call). On top of that, this is especially important for error handling with the trie lazy loading since we IO from disk and each call returns an error (which may fail at any 'depth' in the trie).

lib/trie/trie.go
qdm12: I would rather have it without a custom type (for clarity/explicitness/portability) if that's ok with you. Otherwise, I can do this in a future PR (since deletedMerkleValues `map[string]struct{}` are already in development).

lib/trie/trie.go
```diff
@@ -1157,8 +1383,32 @@ func handleDeletion(branch *Node, key []byte) (newNode *Node, branchChildMerged
...
-		return newBranch, branchChildMerged
+		return newBranch, branchChildMerged, nil
+	}
+}
+
+// ensureMerkleValueIsCalculated is used before calling PopulateMerkleValues
+// to ensure the parent node and all its descendant nodes have their Merkle
+// value computed and ready to be used. This has a close to zero performance
+// impact if the parent node Merkle value is already computed.
+func ensureMerkleValueIsCalculated(parent *Node, isRoot bool) (err error) {
```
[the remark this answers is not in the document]
qdm12: I changed `registerDeletedMerkleValue` to be a receiver on Trie so we don't have to spam with `isRoot := t.root == parent` in the trie code.

However, I left `ensureMerkleValueIsCalculated` as is, since the next PR (branch based on this one) tracks inserted node hashes, and we need to inject `isRoot` from the caller to that function since the node to be created is never equal to the current trie root, but may replace it (so `isRoot` should be true).

Anyway, this is about to be thrown out with lazy loading since every node will have its Merkle value <-> encoding stored on disk.

lib/trie/trie.go
```diff
@@ -80,52 +82,50 @@ func (t *Trie) handleTrackedDeltas(success bool, pendingDeletedMerkleValues map[
...
 
-// updateGeneration is called when the currentNode is from
-// an older trie generation (snapshot) so we deep copy the
-// node and update the generation on the newer copy.
-func updateGeneration(currentNode *Node, trieGeneration uint64,
-	deletedMerkleValues map[string]struct{}, copySettings node.CopySettings) (
-	newNode *Node) {
-	newNode = currentNode.Copy(copySettings)
-	newNode.Generation = trieGeneration
+func registerDeletedMerkleValue(node *Node, isRoot bool,
```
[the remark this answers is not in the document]
qdm12: Done :+1:

lib/runtime/wasmer/imports.go
```diff
@@ -1187,7 +1197,8 @@ func ext_default_child_storage_storage_kill_version_1(context unsafe.Pointer, ch
 	storage := ctx.Storage
 
 	childStorageKey := asMemorySlice(instanceContext, childStorageKeySpan)
-	storage.DeleteChild(childStorageKey)
+	err := storage.DeleteChild(childStorageKey)
+	panicOnError(err)
```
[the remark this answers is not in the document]
qdm12: There are a lot of places in our runtime code where we should panic (or somehow handle the error), but I'd say it's out of scope since code was not modified there. Let's address this in #2669 instead

lib/runtime/storage/trie.go
```diff
@@ -97,15 +98,20 @@ func (s *TrieState) Has(key []byte) bool {
...
 
 	s.lock.Lock()
 	defer s.lock.Unlock()
-	s.t.Delete(key)
+	err = s.t.Delete(key)
+	if err != nil {
+		return err
+	}
+
+	return nil
```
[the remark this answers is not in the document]
qdm12: I wrapped the error with `fmt.Errorf("deleting from trie: %w", err)` :wink: Thanks for pointing it out still.

lib/trie/trie.go
qdm12: Never mind, stuff is getting out of hands on #2881 and I need a custom struct with methods in the end... So good suggestion :+1:

qdm12: Done in #2896 (next branch) to avoid adding even more deltas in this branch which is already thick.
I will definitely use that in the next-next branch to track inserted node hashes in the trie.

timwu20 (approved):

lib/trie/trie.go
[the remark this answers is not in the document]
qdm12: Changed `ensureMerkleValueIsCalculated` to check for the root node within its body since the PR on tracking node hashes on every insertion (#2881) got abandoned.

lib/trie/trie.go
```diff
@@ -706,59 +759,78 @@ func retrieveFromBranch(branch *Node, key []byte) (value []byte) {
...
 	prefix := codec.KeyLEToNibbles(prefixLE)
 	prefix = bytes.TrimSuffix(prefix, []byte{0})
 
-	t.root, deleted, _, allDeleted = t.clearPrefixLimitAtNode(
+	root, deleted, _, allDeleted, err := t.clearPrefixLimitAtNode(
 		t.root, prefix, limit, pendingDeletedMerkleValues)
-	return deleted, allDeleted
+	if err != nil {
+		// Note: no need to wrap the error really since the private function has
+		// the same name as the exported function `ClearPrefixLimit`.
```
qdm12, line 781: nit Not really needed

```suggestion
```

lib/trie/trie.go
```diff
@@ -948,135 +1080,208 @@ func (t *Trie) clearPrefixAtNode(parent *Node, prefix []byte,
...
+
+	newParent = nil
+
+	err = t.registerDeletedMerkleValue(parent, deletedMerkleValues)
+	if err != nil {
+		return nil, fmt.Errorf("registering deleted merkle value: %w", err)
 	}
-	return parent
+
+	return newParent, nil
```
qdm12, line 1215: note: logic was flipped around for earlier returns / less indentation

lib/trie/trie.go
```diff
@@ -1147,8 +1364,32 @@ func handleDeletion(branch *Node, key []byte) (newNode *Node, branchChildMerged
...
+			return fmt.Errorf("calculating Merkle value of root node: %w", err)
+		}
+	} else {
+		_, err = parent.CalculateMerkleValue()
+		if err != nil {
+			return fmt.Errorf("calculating Merkle value of node: %w", err)
+		}
+	}
+
+	return nil
```
qdm12, line 1392: nit change to early return

lib/trie/trie.go
```diff
@@ -79,52 +81,48 @@ func (t *Trie) handleTrackedDeltas(success bool, pendingDeletedMerkleValues map[
...
-		newLeaf = currentLeaf
+	pendingDeletedMerkleValues map[string]struct{}) (
+	newNode *Node, err error) {
+	if currentNode.Generation == t.generation {
+		// no need to track deleted node, deep copy the node and
+		// update the node generation.
+		newNode = currentNode
 	} else {
-		newLeaf = updateGeneration(currentLeaf, t.generation, pendingDeletedMerkleValues, copySettings)
+		err = t.registerDeletedMerkleValue(currentNode,
```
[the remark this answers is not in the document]
qdm12: Length check should be kept in there to avoid duplicating the check a bunch of times;

It's renamed in #2915 which is a branch based on this one.

lib/trie/trie.go
qdm12: Meh I'll keep as it is, this only saves 1 line and it's still short blocks of code.

lib/runtime/storage/trie.go
```diff
@@ -68,10 +69,10 @@ func (s *TrieState) RollbackStorageTransaction() {
 }
 
 // Set sets a key-value pair in the trie
-func (s *TrieState) Set(key, value []byte) {
+func (s *TrieState) Set(key, value []byte) (err error) {
```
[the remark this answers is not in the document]
qdm12: Done in https://github.com/ChainSafe/gossamer/pull/2954

lib/runtime/storage/trie.go
```diff
@@ -167,24 +173,29 @@ func (s *TrieState) GetChildStorage(keyToChild, key []byte) ([]byte, error) {
...
 	s.lock.Lock()
 	defer s.lock.Unlock()
-	s.t.DeleteChild(key)
+	return s.t.DeleteChild(key)
 }
 
-// DeleteChildLimit deletes up to limit of database entries by lexicographic order, return number
-//  deleted, true if all delete otherwise false
-func (s *TrieState) DeleteChildLimit(key []byte, limit *[]byte) (uint32, bool, error) {
+// DeleteChildLimit deletes up to limit of database entries by lexicographic order.
```
jimjbrettj, line 182:
```suggestion
// DeleteChildLimit deletes up to limit of database entries by lexicographic order
```

lib/runtime/storage/trie.go
```diff
@@ -194,20 +205,24 @@ func (s *TrieState) DeleteChildLimit(key []byte, limit *[]byte) (uint32, bool, e
 		keys = append(keys, k)
 	}
 	sort.Strings(keys)
-	deleted := uint32(0)
 	for _, k := range keys {
-		tr.Delete([]byte(k))
+		// TODO have a transactional/atomic way to delete multiple keys in trie.
```
jimjbrettj, line 209: This seems like a good idea, do we have an issue or discussion opened for this?

lib/trie/trie.go
```diff
@@ -354,33 +358,51 @@ func (t *Trie) insert(parent *Node, key, value []byte,
 			StorageValue: value,
 			Generation:   t.generation,
 			Dirty:        true,
-		}, mutated, nodesCreated
+		}, mutated, nodesCreated, nil
 	}
 
 	// TODO ensure all values have dirty set to true
```
jimjbrettj, line 364: is this still a valid TODO?

lib/trie/trie.go
```diff
@@ -809,51 +889,79 @@ func (t *Trie) clearPrefixLimitChild(branch *Node, prefix []byte, limit uint32,
 		const valuesDeleted, nodesRemoved = 0, 0
 		// TODO ensure this is the same behaviour as in substrate
```
jimjbrettj, line 890: Is this still a vaid TODO?

jimjbrettj (approved):
great work, just a few comments!

lib/trie/trie.go
qdm12: Yeah, I'll get to it eventually 💪 🥲

lib/runtime/storage/trie.go
qdm12: What's wrong with an ending full stop 😄 
I got terrorised by https://github.com/tetafro/godot that's why

lib/runtime/storage/trie.go
qdm12: Good point, created https://github.com/ChainSafe/gossamer/issues/3032 and linked in that comment

lib/trie/trie.go
qdm12: Meh not really, removed it.

PR #2875, opened 2022-10-06
chore(lib/trie): `LoadFromMap` as function for transactional trie mutation

qdm12:
## Changes

- Trie mutation should be transactional/atomic as a method, in case an error occurs
- Changed from method to function to avoid leaving a trie in a bad state
- Function returns a new empty trie with all the key and value pairs given

This will become important for lazy loading where a disk write operation could fail for real.

## Tests

```sh
```

## Issues

Related to #2838 

## Primary Reviewer

@jimjbrettj


EclesioMeloJunior (approved):

jimjbrettj (approved):
great work!

PR #2876, opened 2022-10-06
feat(lib/trie): atomic tracked merkle values

qdm12:
## Changes

- New trie field `newDeletedMerkleValues` map to track in-flight deleted Merkle values
- Move values from `newDeletedMerkleValues` to `deletedMerkleValues` **if** the exported operation succeeds
- Reset `newDeletedMerkleValues` at the end of every exported trie operation

For now this really only affects `LoadFromMap` if it fails after the first trie insertion.

However, this is **QUITE IMPORTANT** for lazy loading where most functions will now have to handle an error returned, and could fail at any step.

## Tests

```sh
```

## Issues

Related to #2838 

## Primary Reviewer

@edwardmack


lib/trie/trie.go
```diff
@@ -322,6 +342,9 @@ func findNextKeyChild(children []*Node, startIndex byte,
 // Put inserts a value into the trie at the
 // key specified in little Endian format.
 func (t *Trie) Put(keyLE, value []byte) {
+	defer func() {
+		t.handleTrackedDeltas(nil)
+	}()
```
EclesioMeloJunior, line 347: is there any reason to put `t.handleTrackedDeltas(nil)` inside a defer call?
```suggestion
	nibblesKey := codec.KeyLEToNibbles(keyLE)
	t.root, _, _ = t.insert(t.root, nibblesKey, value)
        t.handleTrackedDeltas(nil)
```

lib/trie/trie.go
```diff
@@ -515,6 +538,10 @@ func (t *Trie) insertInBranch(parentBranch *Node, key, value []byte) (
 // The keys are in hexadecimal little Endian encoding and the values
 // are hexadecimal encoded.
 func (t *Trie) LoadFromMap(data map[string]string) (err error) {
+	defer func() {
+		t.handleTrackedDeltas(err)
+	}()
+
```
EclesioMeloJunior, line 544:
```suggestion
	defer t.handleTrackedDeltas(err)
```

lib/trie/trie.go
```diff
@@ -681,6 +708,10 @@ func retrieveFromBranch(branch *Node, key []byte) (value []byte) {
 // keys and a boolean indicating if all keys with the prefix were deleted
 // within the limit.
 func (t *Trie) ClearPrefixLimit(prefixLE []byte, limit uint32) (deleted uint32, allDeleted bool) {
+	defer func() {
+		t.handleTrackedDeltas(nil)
+	}()
+
```
EclesioMeloJunior, line 714:
```suggestion
	defer t.handleTrackedDeltas(nil)
```

lib/trie/trie.go
```diff
@@ -865,6 +896,10 @@ func (t *Trie) deleteNodesLimit(parent *Node, limit uint32) (
 // ClearPrefix deletes all nodes in the trie for which the key contains the
 // prefix given in little Endian format.
 func (t *Trie) ClearPrefix(prefixLE []byte) {
+	defer func() {
+		t.handleTrackedDeltas(nil)
+	}()
+
```
EclesioMeloJunior, line 902:
```suggestion
	defer t.handleTrackedDeltas(nil)

```

lib/trie/trie.go
```diff
@@ -950,6 +985,10 @@ func (t *Trie) clearPrefixAtNode(parent *Node, prefix []byte) (
 // matching the key given in little Endian format.
 // If no node is found at this key, nothing is deleted.
 func (t *Trie) Delete(keyLE []byte) {
+	defer func() {
+		t.handleTrackedDeltas(nil)
+	}()
```
EclesioMeloJunior, line 990:
```suggestion
	defer t.handleTrackedDeltas(nil)
```

EclesioMeloJunior (changes requested):

lib/trie/trie.go
qdm12: Actually this is different I think; the original case will capture the returned named `err` error, but the second will capture the original (nil) error...

See https://go.dev/play/p/q1E3YATJ7ir

lib/trie/trie.go
qdm12: Keeping it as is (so I don't forget like in a few hours on another branch) due to https://github.com/ChainSafe/gossamer/pull/2876#discussion_r989956510

lib/trie/trie.go
qdm12: Keeping it as is (so I don't forget like in a few hours on another branch) due to https://github.com/ChainSafe/gossamer/pull/2876#discussion_r989956510

lib/trie/trie.go
qdm12: Keeping it as is (so I don't forget like in a few hours on another branch) due to https://github.com/ChainSafe/gossamer/pull/2876#discussion_r989956510

lib/trie/trie.go
qdm12: Well not for now, but there is on the next branch where `nil` gets replaced with `err` because `insert` returns an error.
If you don't mind can we keep it as is (especially so I don't forget to put it back in a defer call).

lib/trie/trie.go
```diff
@@ -515,6 +538,10 @@ func (t *Trie) insertInBranch(parentBranch *Node, key, value []byte) (
 // The keys are in hexadecimal little Endian encoding and the values
 // are hexadecimal encoded.
 func (t *Trie) LoadFromMap(data map[string]string) (err error) {
+	defer func() {
+		t.handleTrackedDeltas(err)
```
timwu20, line 542: I don't see how `err` is every set in this.  Won't it always be `nil`?

lib/trie/trie.go
```diff
@@ -54,28 +60,42 @@ func (t *Trie) Snapshot() (newTrie *Trie) {
...
 
+// handleTrackedDeltas modifies the trie deleted merkle values set
+// only if the error is nil. In all cases, it resets the
+// `newDeletedMerkleValues` map.
+func (t *Trie) handleTrackedDeltas(err error) {
+	if err == nil {
+		for merkleValue := range t.newDeletedMerkleValues {
+			t.deletedMerkleValues[merkleValue] = struct{}{}
+		}
+	}
```
timwu20, line 87: this seems really awkward.  Could you not just call this whenever `err == nil`, and explicitly reset the `t.newDeletedMerkleValues`.

lib/trie/trie.go
```diff
@@ -26,6 +26,11 @@ type Trie struct {
 	// pruner to detect with database keys (trie node Merkle values) can
 	// be deleted.
 	deletedMerkleValues map[string]struct{}
+	// newDeletedMerkleValues contains deleted Merkle values for operations
+	// in progress. Its goal is to prevent having the trie `deletedMerkleValues`
+	// in a bad state if one of the trie operation modified some (deep copied) trie
+	// node and then failed.
+	newDeletedMerkleValues map[string]struct{}
```
timwu20, line 33: is there a need for this? Could you not just lock and unlock the `trie` so it will never be in a bad state.

lib/trie/trie.go
qdm12: No it will capture the returned named error, so if any `return xxx` happens it will get caught in that deferred function

lib/trie/trie.go
qdm12: Well I get your point, although this function is about to be stuffed with more logic to track inserted merkle values and remove re-inserted merkle values too. Maybe even more stuff with lazy loading.

lib/trie/trie.go
qdm12: This is about tracking deleted merkle values in the trie since the last trie snapshot, not the actual trie data.
The trie data is unaffected if an error occurs (since we don't update the root node of the trie), but those tracked merkle values would be wrongly tracked in case of an error occuring in the middle of an operation.

jimjbrettj, line 33: nit: maybe change this to something like pendingDeletedMerkleValues or something like this?

jimjbrettj (approved):
small nit, but looks good to me

lib/trie/trie.go
qdm12: Done

qdm12:
@jimjbrettj @EclesioMeloJunior I changed this to inject a `pendingDeletedMerkleValues` to the (unexported) functions. This feels a bit better / more explicit, although it does change more function signatures than before. What do you think? :thinking:

lib/trie/trie.go
qdm12: I think every trie method should be transactional; the transaction mechanism in the state (based on the trie snapshot) allows to run multiple method calls and rollback (the trie data for now, not the tracked stuff, although this could be arranged).

However, there is a lot of trie calls without snapshotting the trie, and in that case, if the operation fails in the middle (for now it's only `LoadFromMap`, but soon all operations may fail at any time during their recursion), the trie will be left in a bad state.

Finally - that came up today - I actually need to track the deleted (and inserted) Merkle values during a single operation to track the inserted Merkle values correctly (i.e. a temporarily inserted node which then gets thrown out in the same method call should not be tracked as inserted).

qdm12:
@timwu20 merging this since I kind of need this badly now in a few other branches (and rebasing/changing PR target branch for the other 4 branches is rather time consuming). But feel free to comment back on the conversations which I hope would somehow be resolved, and I can address this in another PR.

qdm12:
Oops forgot to add a unit test for `handleTrackedDeltas`, added in #2873

Issue #2839, opened 2022-09-16
Change retain blocks type to be `uint32`


qdm12:
Since block numbers are `uint32`, retain blocks should also be `uint32`. For now we have a mix of `uint` and `int64` etc., it would be nice to change the types of retain blocks struct fields to be `uint32`

qdm12:
Change it to `Uint64Flag` (so it rejects negative values), and then cast it to uint32 using `cfg.RetainBlocks = uint32(ctx.Uint64(RetainBlockNumberFlag.Name))`.

Thinking about it some more, since we need to use uint64 for the flag, you should also add a check that the value given is less than the maximum uint32 value (`^uint32(0)` which equals 4,294,967,295) otherwise it would wrap around and produce unexpected surprises. For example:

- uint64 `4294967295` casted to uint32 results in `4294967295`
- uint64 `4294967296` casted to uint32 results in `0` (wraps around)
- uint64 `4294967297` casted to uint32 results in `1` (wraps around)

PR #2877, opened 2022-10-06
chore(chain, cmd, dot): Change retain blocks type to be `uint32`

axaysagathiya:
## Changes

Changed retain blocks type to `uint32` and 
modified the code to use `cli.Uint64Flag` struct instead of `cli.Int64Flag` struct.

## Tests

```sh
```

## Issues

#2839

## Primary Reviewer

@timwu20


cmd/gossamer/config.go
```diff
@@ -485,7 +485,7 @@ func setDotGlobalConfigFromFlags(ctx *cli.Context, cfg *dot.GlobalConfig) error
 		cfg.MetricsAddress = metricsAddress
 	}
 
-	cfg.RetainBlocks = ctx.Int64(RetainBlockNumberFlag.Name)
+	cfg.RetainBlocks = uint32(ctx.Uint64(RetainBlockNumberFlag.Name))
```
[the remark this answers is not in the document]
qdm12: Probably error out I'd say in this case, that warn log might get unnoticed with unwanted consequences.

cmd/gossamer/main.go
```diff
@@ -425,7 +425,7 @@ func pruneState(ctx *cli.Context) error {
 	}
 
 	bloomSize := ctx.GlobalUint64(BloomFilterSizeFlag.Name)
-	retainBlocks := ctx.GlobalInt64(RetainBlockNumberFlag.Name)
+	retainBlocks := uint32(ctx.GlobalUint64(RetainBlockNumberFlag.Name))
```
qdm12, line 428: We should add a check that the value `ctx.GlobalUint64(RetainBlockNumberFlag.Name)` is less than the max uint32 (`^uint32(0)`)

qdm12 (approved):
Nice, just add these 2 checks in case the value is above max uint32 and let's get this merged!

Pre-approving :wink:

cmd/gossamer/config.go
```diff
@@ -485,7 +485,14 @@ func setDotGlobalConfigFromFlags(ctx *cli.Context, cfg *dot.GlobalConfig) error
 		cfg.MetricsAddress = metricsAddress
 	}
 
-	cfg.RetainBlocks = ctx.Int64(RetainBlockNumberFlag.Name)
+	const uint32Max = ^uint32(0)
+	flagValue := ctx.Uint64(RetainBlockNumberFlag.Name)
+
+	if uint64(uint32Max) < flagValue {
+		return fmt.Errorf("retain blocks value overflows uint32 boundaries, must be less than, %d", uint32Max)
```
EclesioMeloJunior, line 492: supernit
```suggestion
		return fmt.Errorf("retain blocks value overflows uint32 boundaries, must be less or equal than: %d", uint32Max)
```

EclesioMeloJunior (changes requested):

EclesioMeloJunior (approved):

PR #2881, opened 2022-10-07
feat(lib/trie): cache inserted trie node Merkle values

### LINKED ISSUE #2854: Faster online pruning with cached trie inserted/modified merkle values

qdm12:
## Changes

- [x] Production code changes
- [ ] Fixes all tests
- [ ] Change delta tracking to use node pointers to improve performance
- [ ] Remove dirty bool field
- [ ] Create issue for child tries delta tracking

## Tests

```sh
```

## Issues

Fixes #2854 and unblocks lazy loading impl

### DIFF
--- a/dot/state/storage.go
+++ b/dot/state/storage.go
@@ -85,13 +85,10 @@ func (s *StorageState) StoreTrie(ts *rtstorage.TrieState, header *types.Header) 
 	}
 
 	if header != nil {
-		insertedMerkleValues, err := ts.GetInsertedMerkleValues()
-		if err != nil {
-			return fmt.Errorf("failed to get state trie inserted keys: block %s %w", header.Hash(), err)
-		}
+		insertedMerkleValues := ts.GetInsertedMerkleValues()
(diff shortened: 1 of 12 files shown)


qdm12:
This is simply not possible in terms of cpu performance.
Re-calculating the merkle value (scale encoding + blake2b) of N nodes when inserting a node at depth N in the trie is too expensive to be a working solution. This is also highlights the need for:

- lazy-hashing: only compute the state root of the trie when needed, and not on every insertion
- lazy loading needs a write cache, meaning we keep deltas in a in-memory part of the trie without encoding/hashing/writing to disk, and only do the root hash computation + write of deltas when needed (i.e. at the end of handling a block).

Closing this for now as a failed but taughtful experiment.

Issue #2801, opened 2022-09-05
Gossamer Miss Behavior `Reason: Request refused` and `Reason: Request timeout`


EclesioMeloJunior:
## Issue summary

Substrate nodes are punishing Gossamer since gossamer refuses a substrate block request

```
2022-09-05 10:25:12.026 DEBUG tokio-runtime-worker sync: Request to peer PeerId("12D3KooWGc4ohzsjYwBQnZdV1CpUNv46JpgCUfL5iGLqW2KfWxm2") failed: Refused.    
2022-09-05 10:25:12.026 TRACE tokio-runtime-worker sub-libp2p: External API => Disconnect(12D3KooWGc4ohzsjYwBQnZdV1CpUNv46JpgCUfL5iGLqW2KfWxm2, SetId(0))    
2022-09-05 10:25:12.026 TRACE tokio-runtime-worker sub-libp2p: PSM <= Dropped(12D3KooWGc4ohzsjYwBQnZdV1CpUNv46JpgCUfL5iGLqW2KfWxm2, SetId(0))    
2022-09-05 10:25:12.026 TRACE tokio-runtime-worker peerset: Dropping 12D3KooWGc4ohzsjYwBQnZdV1CpUNv46JpgCUfL5iGLqW2KfWxm2: -256 to 68    
2022-09-05 10:25:12.026 TRACE tokio-runtime-worker sub-libp2p: External API <= Closed(12D3KooWGc4ohzsjYwBQnZdV1CpUNv46JpgCUfL5iGLqW2KfWxm2, SetId(0))    
2022-09-05 10:25:12.026 TRACE tokio-runtime-worker sub-libp2p: Handler(PeerId("12D3KooWGc4ohzsjYwBQnZdV1CpUNv46JpgCUfL5iGLqW2KfWxm2"), ConnectionId(2)) <= Close(SetId(0))    
2022-09-05 10:25:12.026 DEBUG tokio-runtime-worker sync: 12D3KooWGc4ohzsjYwBQnZdV1CpUNv46JpgCUfL5iGLqW2KfWxm2 disconnected    
2022-09-05 10:25:12.026 TRACE tokio-runtime-worker sync: 0 blocks ready for import    
2022-09-05 10:25:12.027 TRACE tokio-runtime-worker sync: Block imported successfully Some(6) (0x3d60…24ee)    
[...]

PR #2885, opened 2022-10-10
fix(dot/sync): Gossip `BlockAnnounceMessage` only after  successfully imported

EclesioMeloJunior:
## Changes

- Gossamer is announcing a block right after receiving a block announcement from another peer which leads us to announce we contain a block that, in fact, we don't have.
- This PR fixes this problem by 
  - removing the gossip after receiving the block announcement
  - announcing the block only after the block was imported successfully

## Tests

```sh
go test -tags integration github.com/ChainSafe/gossamer/dot/sync
go test -tags integration github.com/ChainSafe/gossamer/dot/core
```

## Issues

- Closes https://github.com/ChainSafe/gossamer/issues/2801

## Primary Reviewer

@qdm12


dot/core/service.go
```diff
@@ -127,40 +127,68 @@ func (s *Service) StorageRoot() (common.Hash, error) {
...
+	if err != nil {
+		return fmt.Errorf("handling block: %s", err)
+	}
+
+	isBestBlock := false
+	bestBlockHash := s.blockState.BestBlockHash()
+
+	if bestBlockHash.Equal(block.Header.Hash()) {
+		isBestBlock = true
+	}
```
qdm12, line 140: nit

```suggestion
	bestBlockHash := s.blockState.BestBlockHash()
	isBestBlock := bestBlockHash.Equal(block.Header.Hash())
```

dot/core/service.go
```diff
@@ -127,40 +127,68 @@ func (s *Service) StorageRoot() (common.Hash, error) {
...
+
+	blockAnnounce, err := createBlockAnnounce(block, isBestBlock)
+	if err != nil {
+		logger.Errorf("creating block announce: %s", err)
```
qdm12, line 144: why not return the error?

dot/core/service.go
```diff
@@ -127,40 +127,68 @@ func (s *Service) StorageRoot() (common.Hash, error) {
...
-		return err
+	err := s.handleBlock(block, state)
+	if err != nil {
+		return fmt.Errorf("handling block: %s", err)
 	}
 
+	const isBestBlock = true
+	blockAnnounce, err := createBlockAnnounce(block, isBestBlock)
+	if err != nil {
+		logger.Errorf("creating block announce: %s", err)
```
qdm12, line 163: why not return the error?

dot/network/block_announce.go
```diff
@@ -203,14 +203,13 @@ func (s *Service) validateBlockAnnounceHandshake(from peer.ID, hs Handshake) err
 // if some more blocks are required to sync the announced block, the node will open a sync stream
 // with its peer and send a BlockRequest message
 func (s *Service) handleBlockAnnounceMessage(from peer.ID, msg NotificationsMessage) (propagate bool, err error) {
+	propagate = false
+
```
qdm12, line 207: nit not needed, the named return is already false

```suggestion
```

dot/network/block_announce.go
```diff
@@ -203,14 +203,13 @@ func (s *Service) validateBlockAnnounceHandshake(from peer.ID, hs Handshake) err
...
 	bam, ok := msg.(*BlockAnnounceMessage)
 	if !ok {
-		return false, errors.New("invalid message")
-	}
-
-	if err = s.syncer.HandleBlockAnnounce(from, bam); err != nil {
-		return false, err
+		return propagate, errors.New("invalid message")
```
qdm12, line 210: Can this error ever happen? Shouldn't we panic instead here? :thinking: 
Especially since the function is named `handleBlockAnnounceMessage`

dot/network/block_announce.go
```diff
@@ -203,14 +203,13 @@ func (s *Service) validateBlockAnnounceHandshake(from peer.ID, hs Handshake) err
...
 	}
 
-	return true, nil
+	err = s.syncer.HandleBlockAnnounce(from, bam)
+	return propagate, err
```
qdm12, line 214: is it normal `propagate` is always `false` here? Is there a point returning it anyway? (Maybe to match a signature I'm not sure).

dot/sync/chain_processor.go
```diff
@@ -238,6 +238,7 @@ func (s *chainProcessor) handleBlock(block *types.Block) error {
 	}
 
 	logger.Debugf("🔗 imported block number %d with hash %s", block.Header.Number, block.Header.Hash())
+	// should we announce the block here?
```
qdm12, line 241: did you figure it out? Or asking reviewers? :thinking:

dot/sync/chain_sync.go
```diff
@@ -632,7 +632,7 @@ func (cs *chainSync) dispatchWorker(w *worker) {
 		"start hash %s, target hash %s, "+
 		"request data %d, direction %s",
 		w.id,
-		w.startNumber, w.targetNumber,
+		*w.startNumber, *w.targetNumber,
```
qdm12, line 635: nit can this ever be nil? We should be careful that log doesn't panic if startNumber/targetNumber is nil :thinking:

dot/sync/chain_processor.go
EclesioMeloJunior: Yeah, I figured out we should do the block announcement right after it is imported

dot/network/block_announce.go
EclesioMeloJunior: > Can this error ever happen?

I would say no. This function (`handleBlockAnnounceMessage`) is called after we call the `decodeBlockAnnounceMessage` which returns a `*BlockAnnounceMessage`, so in a case where we receive a wrong block announce message the function `decodeBlockAnnounceMessage` will fail and return an error and then the function `handleBlockAnnounceMessage` will never be called. I can remove this guard check

dot/sync/chain_sync.go
EclesioMeloJunior: Nice catch, moving bellow an if condition that checks this case `if w.targetNumber == nil || w.startNumber == nil`

dot/core/service.go
```diff
@@ -127,40 +127,64 @@ func (s *Service) StorageRoot() (common.Hash, error) {
...
+
+	bestBlockHash := s.blockState.BestBlockHash()
+	isBestBlock := bestBlockHash.Equal(block.Header.Hash())
+
+	blockAnnounce, err := createBlockAnnounce(block, isBestBlock)
+	if err != nil {
+		return fmt.Errorf("creating block announce: %w", err)
+	}
+
+	s.net.GossipMessage(blockAnnounce)
```
jimjbrettj, line 143: Is there a risk/is it allowed to gossip a block announce back to the peer that sent it to us?

dot/core/service.go
```diff
@@ -127,40 +127,64 @@ func (s *Service) StorageRoot() (common.Hash, error) {
...
+		return fmt.Errorf("handling block: %w", err)
 	}
 
+	const isBestBlock = true
+	blockAnnounce, err := createBlockAnnounce(block, isBestBlock)
+	if err != nil {
+		return fmt.Errorf("creating block announce: %w", err)
+	}
+
+	s.net.GossipMessage(blockAnnounce)
```
jimjbrettj, line 162: Is there a risk/is it allowed to gossip a block announce back to the peer that sent it to us?

dot/network/block_announce.go
```diff
@@ -208,9 +209,11 @@ func (s *Service) handleBlockAnnounceMessage(from peer.ID, msg NotificationsMess
 		return false, errors.New("invalid message")
 	}
 
-	if err = s.syncer.HandleBlockAnnounce(from, bam); err != nil {
-		return false, err
+	// TODO announce if we already have the block
```
jimjbrettj, line 212: Is this addressed here? If not, is there an issue created for it?

jimjbrettj (commented):
Looks good, just one question

dot/core/service.go
EclesioMeloJunior: I am not sure about the risks/penalties (I did not see any while looking into substrate logs), I would say that the intended behavior is to don't send the message back but right now this is out of the scope of this PR, should we create an issue to address the creation of a method `GossipMessageExcluding(message, peers...)`? @timwu20 @danforbes

dot/core/service.go
EclesioMeloJunior: In this case, there is no risk at all because we are gossipping a block announce of a block we've created

dot/core/service.go
danforbes: I would need a little more information to be able to definitively answer this question or to be able to comment on the priority of such a task.

dot/core/service.go
jimjbrettj: great point!

dot/network/block_announce.go
EclesioMeloJunior: > Is this addressed here?

yes! removed the comment

jimjbrettj (approved):
great work!

dot/core/service.go
```diff
@@ -127,40 +127,64 @@ func (s *Service) StorageRoot() (common.Hash, error) {
 
 // HandleBlockImport handles a block that was imported via the network
 func (s *Service) HandleBlockImport(block *types.Block, state *rtstorage.TrieState) error {
-	return s.handleBlock(block, state)
+	err := s.handleBlock(block, state)
+	if err != nil {
+		return fmt.Errorf("handling block: %s", err)
```
qdm12, line 132:
```suggestion
		return fmt.Errorf("handling block: %w", err)
```

dot/core/service.go
```diff
@@ -127,40 +127,64 @@ func (s *Service) StorageRoot() (common.Hash, error) {
...
 	for i := range block.Header.Digest.Types {
 		digestValue, err := block.Header.Digest.Types[i].Value()
 		if err != nil {
-			return fmt.Errorf("getting value of digest type at index %d: %w", i, err)
+			return nil, fmt.Errorf("getting value of digest type at index %d: %w", i, err)
 		}
 		err = digest.Add(digestValue)
 		if err != nil {
-			return err
+			return nil, err
```
qdm12, line 176: nit whilst we're at it

```suggestion
			return nil, fmt.Errorf("adding digest value for type at index %d: %w", i, err)
```

dot/core/service_test.go
```diff
@@ -480,9 +480,10 @@ func Test_Service_HandleBlockProduced(t *testing.T) {
 	t.Parallel()
 	execTest := func(t *testing.T, s *Service, block *types.Block, trieState *rtstorage.TrieState, expErr error) {
 		err := s.HandleBlockProduced(block, trieState)
-		assert.ErrorIs(t, err, expErr)
+		require.ErrorIs(t, err, expErr)
```
qdm12, line 483:
```suggestion
		require.ErrorIs(t, expErr, err)
```

dot/core/service_test.go
```diff
@@ -480,9 +480,10 @@ func Test_Service_HandleBlockProduced(t *testing.T) {
...
 		if expErr != nil {
-			assert.EqualError(t, err, expErr.Error())
+			wrapped := fmt.Errorf("handling block: %w", expErr)
+			assert.EqualError(t, err, wrapped.Error())
```
qdm12, line 486: nit

```suggestion
			assert.EqualError(t, err, "handling block: " + expErr.Error())
```

also it would be nice to inject the expected error string, since it might not always be prefixed with `handling block: `

dot/network/block_announce_test.go
```diff
@@ -126,24 +128,58 @@ func TestDecodeBlockAnnounceHandshake(t *testing.T) {
...
+		mockSyncer func(*testing.T, peer.ID, *BlockAnnounceMessage) Syncer
+	}{
+		"block already exists": {
+			mockSyncer: func(t *testing.T, peer peer.ID, blockAnnounceMessage *BlockAnnounceMessage) Syncer {
+				ctrl := gomock.NewController(t)
+				syncer := NewMockSyncer(ctrl)
+				syncer.EXPECT().
+					HandleBlockAnnounce(peer, blockAnnounceMessage).
+					Return(blocktree.ErrBlockExists).
+					Times(1)
```
qdm12, line 142: nit

```suggestion
					Return(blocktree.ErrBlockExists)
```

dot/network/block_announce_test.go
```diff
@@ -126,24 +128,58 @@ func TestDecodeBlockAnnounceHandshake(t *testing.T) {
...
+			msg := &BlockAnnounceMessage{
+				Number: 10,
+				Digest: types.NewDigest(),
+			}
+
+			if tt.mockSyncer != nil {
+				config.Syncer = tt.mockSyncer(t, peerID, msg)
+			}
+
+			s := createTestService(t, config)
```
qdm12, line 176: nit `s` -> `service` would be nice

dot/sync/chain_sync_test.go
```diff
@@ -1246,7 +1247,11 @@ func Test_chainSync_start(t *testing.T) {
 	}
 }
 
+var blockAnnounceHeader = &types.Header{Number: 2}
+
```
qdm12, line 1251: nit Move this locally in the test(s) where it's needed, remove from global scope :pray:

dot/sync/chain_sync_test.go
```diff
@@ -1271,13 +1277,82 @@ func Test_chainSync_setBlockAnnounce(t *testing.T) {
...
+					// will hang until we read on this channel and the goal is to
+					// put the work on the channel and don't block
+					workQueue: make(chan *peerState, 1),
+				}
+			},
+		},
 	}
 	for name, tt := range tests {
 		t.Run(name, func(t *testing.T) {
+			t.Parallel()
```
qdm12, line 1339:
```suggestion
	for name, tt := range tests {
		tt := tt
		t.Run(name, func(t *testing.T) {
			t.Parallel()
```

dot/sync/chain_sync_test.go
```diff
@@ -1271,13 +1277,82 @@ func Test_chainSync_setBlockAnnounce(t *testing.T) {
...
+				assert.EqualError(t, err, tt.wantErr.Error())
+			} else {
+				assert.NoError(t, err)
+			}
+
+			if sync.workQueue != nil {
+				assert.LessOrEqual(t, len(sync.workQueue), 1)
+				if len(sync.workQueue) > 0 {
+					<-sync.workQueue
+					close(sync.workQueue)
```
qdm12, line 1353: nit call me an async nitter :smile: But the producer should close the channel, not the reader (so not here). You can probably leave it there and it'll get GCed

dot/sync/chain_sync_test.go
```diff
@@ -1271,13 +1277,82 @@ func Test_chainSync_setBlockAnnounce(t *testing.T) {
...
-			assert.ErrorIs(t, err, tt.wantErr)
+			if tt.wantErr != nil {
+				assert.EqualError(t, err, tt.wantErr.Error())
+			} else {
+				assert.NoError(t, err)
+			}
+
+			if sync.workQueue != nil {
+				assert.LessOrEqual(t, len(sync.workQueue), 1)
+				if len(sync.workQueue) > 0 {
```
qdm12, line 1351: meganit you could have this as a for loop instead (change `if` to `for` and remove `close` of channel), in case there is more than 1 element in the channel

EclesioMeloJunior: This channel cannot be closed easily by the writer unless we do some refactoring in this code, this channel is created and read based on a timer, so if I put a `for work := range sync.workQueue` then it will block until we close the channel which never happens, I will keep it as it is but removing the `close`, something like:

```go
// if workQueue is not nil then I expect one element in the channel
if sync.workQueue != nil {
	assert.Equal(t, len(sync.workQueue), 1)
}
```

dot/core/service.go
```diff
@@ -127,40 +127,64 @@ func (s *Service) StorageRoot() (common.Hash, error) {
...
 func (s *Service) HandleBlockProduced(block *types.Block, state *rtstorage.TrieState) error {
-	if err := s.handleBlock(block, state); err != nil {
-		return err
+	err := s.handleBlock(block, state)
+	if err != nil {
+		return fmt.Errorf("handling block: %w", err)
 	}
 
+	const isBestBlock = true
+	blockAnnounce, err := createBlockAnnounce(block, isBestBlock)
```
timwu20, line 157: why not just inline?
```suggestion
	blockAnnounce, err := createBlockAnnounce(block, true)
```

timwu20 (approved):

Issue #2481, opened 2022-04-14
CI: split unit testing in different jobs


qdm12:
## Task summary

1. Split unit testing in different jobs in the `unit-tests` workflow, by running one job per top-level package (i.e. one job for `lib/grandpa`)
2. Merge code coverage with https://github.com/wadey/gocovmerge before uploading it to codecov

**Why**:

- Unit testing takes 25 minutes overall, so parallelising it will make it quicker, even with the new job overhead
- Unit testing sometimes hangs forever, and we don't know why. It would be useful to be able to isolate where the fault is coming from and from which package

If one day we manage to have stable and fast (<2 minutes) unit testing, we can revert this and have it back to a single `go test ./...`.

PR #2887, opened 2022-10-11
chore(all): move long tests to integration tests

qdm12:
## Changes

:tipping_hand_person: This should possibly speed up dramatically our entire CI. Long overdued and got sick of it today.

:tipping_hand_man: there is barely any change, it's just moving tests around really. I think I removed a single `time.Sleep(time.Second)` to keep a test as a unit test somewhere, but that's it.

- [x] Move test files with >75% long tests to integration-tagged-and-named test files. Some of the tests are fast in those files, but this is to minimize deltas and bad git blame.
- [x] Add package names to the integration tests workflow matrix
- [x] Remove usage of `t.Short()` + Skip `Test_Gossip` since it doesn't work
- [x] Move `//go:generate` commands to non-integration-tagged test files (to avoid passing `-tags integration` to go generate)
- [x] Remove older `// +build integration`

Packages addressed:

- [x] dot
- [x] dot/digest
- [x] dot/network
- [x] dot/rpc
- [x] dot/rpc/subscription
- [x] dot/state
- [x] lib/grandpa

## Tests

```sh
go test -tags integration github.com/ChainSafe/gossamer/...
```

## Issues

#2481 

## Primary Reviewer

@timwu20


lib/grandpa/vote_message_test.go
```diff
@@ -1,3 +1,5 @@
+//go:build integration
```
timwu20, line 1: shouldn't this be renamed to `vote_message_integration_test.go`?

qdm12:
@timwu20 

I'm thinking to just move files, even if some of tests are fast, provided at least one test is slow.
That way it would still have correct git blame information, and it would be easier to rebase this branch/other branches with git.
Or should we keep it as is (where rebase may get tough and some of git blame information got lost)?

lib/grandpa/vote_message_test.go
qdm12: Indeed! Done.

qdm12:
Workflows taking > 5 minutes, ordered by time spent:

1. 10 minutes: lib/grandpa
2. 8 minutes: stress, dot/rpc/modules
3. 7 minutes: RPC, dot
4. 6 minutes: unit tests, lib/babe

Since stress and RPC might not be possible to accelerate, I think aiming for 8 minutes total CI time is a good goal practically speaking.

timwu20 (approved):

codecov.yml
```diff
@@ -4,64 +4,9 @@ coverage:
     patch: off
     project:
       default:
-        target: 60%
```
jimjbrettj, line 7: Why did we remove all of these targets?

dot/node_integration_test.go
```diff
@@ -9,24 +9,197 @@ import (
...
+	dotConfig.Core.WasmInterpreter = wasmer.Name
+	dotConfig.Global.Name = "TestNode"
+
+	ks, err := initKeystore(t, dotConfig)
+	assert.NoError(t, err)
+
+	mockServiceRegistry := NewMockServiceRegisterer(ctrl)
+	mockServiceRegistry.EXPECT().RegisterService(gomock.Any()).Times(8)
+
+	m := NewMocknodeBuilderIface(ctrl)
```
jimjbrettj, line 88: To make sure i understand, this is a temp PR to speed up CI, but in future will be more intentional about where individual tests are?

dot/rpc/websocket_integration_test.go
```diff
@@ -1,3 +1,5 @@
+//go:build integration
```
jimjbrettj, line 1: subscription/websocket_integration_test.go also needs this header I believe

jimjbrettj (changes requested):
Looks good, just a few small things

codecov.yml
qdm12: TLDR: these targets are useless and are impossible to update properly (I set them initially, my bad!)

Since a bunch of tests were moved from 'unit' to 'integration, coverage got lowered a bunch in various packages.
I initially tried to adjust these % and packages position in each % target, pulled my hair off, and realized that each % target listed in codecov.yml are the average % coverage for all the packages listed in them (I thought it was a % target for each package, not their average). And I am still not too sure how they calculate this 'average' (go test doesn't offer such thing either).

One solution would be to have a coverage target per package, but, visually speaking, that would make our workflows list in our CI become mega long and I'd personally avoid.

dot/node_integration_test.go
qdm12: This PR renames entire files so git blame information is kept as much as possible. After this PR (or never), we could gradually fish out fast tests from these integration test files and move them to unit test files, although that would also lose git blame information :cry:

dot/rpc/websocket_integration_test.go
qdm12: Indeed :rocket: Fixed!

jimjbrettj (approved):
nice work!

PR #2894, opened 2022-10-17
feat(pkg/scale): encoding and decoding of maps in scale

axaysagathiya:
## Changes

written the methods for encoding and decoding maps in scale.

## Tests

```sh
```

## Issues

#2796 

## Primary Reviewer

@timwu20


pkg/scale/decode.go
```diff
@@ -426,6 +428,34 @@ func (ds *decodeState) decodeArray(dstv reflect.Value) (err error) {
 	return
 }
 
+func (ds *decodeState) decodeMap(dstv reflect.Value) (err error) {
+	l, err := ds.decodeLength()
```
qdm12, line 432: nit rename `l` to `numberOfTuples`

pkg/scale/decode.go
```diff
@@ -426,6 +428,34 @@ func (ds *decodeState) decodeArray(dstv reflect.Value) (err error) {
...
+	if err != nil {
+		return
```
qdm12, line 434: nit add error context so we can trace back what failed and which code path

```suggestion
		return fmt.Errorf("decoding length: %w", err)
```

kishansagathiya (commented):
Fix this test `Test_unmarshal_optionality/testMap1 `

pkg/scale/decode.go
```diff
@@ -426,6 +428,34 @@ func (ds *decodeState) decodeArray(dstv reflect.Value) (err error) {
...
+	}
+	in := dstv.Interface()
+
+	for i := uint(0); i < l; i++ {
+		tempKeyType := reflect.TypeOf(in).Key()
+		tempKey := reflect.New(tempKeyType).Elem()
+		err = ds.unmarshal(tempKey)
+		if err != nil {
+			return
```
qdm12, line 443:
```suggestion
			return fmt.Errorf("decoding key %d of %d: %w", i+1, l, err)
```

pkg/scale/decode.go
```diff
@@ -426,6 +428,34 @@ func (ds *decodeState) decodeArray(dstv reflect.Value) (err error) {
...
+		}
+
+		tempElemType := reflect.TypeOf(in).Elem()
+		tempElem := reflect.New(tempElemType).Elem()
+		err = ds.unmarshal(tempElem)
+		if err != nil {
+			return
```
qdm12, line 450:
```suggestion
			return fmt.Errorf("decoding value %d of %d: %w", i+1, l, err)
```

pkg/scale/encode.go
```diff
@@ -223,6 +225,35 @@ func (es *encodeState) encodeArray(in interface{}) (err error) {
...
 
+func (es *encodeState) encodeMap(in interface{}) (err error) {
+	v := reflect.ValueOf(in)
+	err = es.encodeLength(v.Len())
+	if err != nil {
+		return
+	}
+
+	for i := v.MapRange(); i.Next(); {
+		fmt.Println(i.Key(), "\t:", i.Value())
```
EclesioMeloJunior, line 236:
```suggestion
```

pkg/scale/decode.go
```diff
@@ -426,6 +428,34 @@ func (ds *decodeState) decodeArray(dstv reflect.Value) (err error) {
 	return
 }
 
+func (ds *decodeState) decodeMap(dstv reflect.Value) (err error) {
```
qdm12, line 431: nit rename `dstv` to `destination`

pkg/scale/decode.go
```diff
@@ -426,6 +428,34 @@ func (ds *decodeState) decodeArray(dstv reflect.Value) (err error) {
...
+		tempElem := reflect.New(tempElemType).Elem()
+		err = ds.unmarshal(tempElem)
+		if err != nil {
+			return
+		}
+
+		dstv.SetMapIndex(tempKey, tempElem)
+	}
+
+	return
```
qdm12, line 456: nit to be explicit / the function is not that short

```suggestion
	return nil
```

pkg/scale/encode_test.go
```diff
@@ -1096,6 +1114,24 @@ func Test_encodeState_encodeArray(t *testing.T) {
 	}
 }
 
+func Test_encodeState_encodeMap(t *testing.T) {
+	for _, tt := range mapTests {
+		t.Run(tt.name, func(t *testing.T) {
```
EclesioMeloJunior, line 1119: you should pin the `tt` variable to the loop context otherwise it can cause inconsistency
```suggestion
	for _, tt := range mapTests {
	        tt := tt
		t.Run(tt.name, func(t *testing.T) {
```

pkg/scale/encode.go
```diff
@@ -223,6 +225,35 @@ func (es *encodeState) encodeArray(in interface{}) (err error) {
 	return
 }
 
+func (es *encodeState) encodeMap(in interface{}) (err error) {
+	v := reflect.ValueOf(in)
+	err = es.encodeLength(v.Len())
+	if err != nil {
+		return
```
qdm12, line 232:
```suggestion
		return fmt.Errorf("encoding length: %w", err)
```

pkg/scale/encode.go
```diff
@@ -223,6 +225,35 @@ func (es *encodeState) encodeArray(in interface{}) (err error) {
...
+	}
+
+	for i := v.MapRange(); i.Next(); {
+		fmt.Println(i.Key(), "\t:", i.Value())
+
```
qdm12, line 237:
```suggestion
```

pkg/scale/decode_test.go
```diff
@@ -132,6 +132,40 @@ func Test_decodeState_decodeSlice(t *testing.T) {
...
+			},
+			want: []byte{
+				8, 2, 44, 115, 111, 109, 101, 32, 115, 116, 114, 105, 110, 103, 16, 44, 108, 111, 114, 101, 109, 32,
+				105, 112, 115, 117, 109,
+			},
+		},
+	}
+
+	for _, tt := range mapTests {
+		t.Run(tt.name, func(t *testing.T) {
```
EclesioMeloJunior, line 156: you should pin the tt variable to the loop context otherwise it can cause inconsistency
```suggestion
	for _, tt := range mapTests {
	        tt := tt
		t.Run(tt.name, func(t *testing.T) {
```

pkg/scale/encode.go
```diff
@@ -223,6 +225,35 @@ func (es *encodeState) encodeArray(in interface{}) (err error) {
...
 }
 
+func (es *encodeState) encodeMap(in interface{}) (err error) {
+	v := reflect.ValueOf(in)
+	err = es.encodeLength(v.Len())
+	if err != nil {
+		return
+	}
+
+	for i := v.MapRange(); i.Next(); {
```
qdm12, line 235: nit rename `i` to `keyValue`, since `i` is usually for an index which isn't the case here :thinking:

pkg/scale/encode.go
```diff
@@ -223,6 +225,35 @@ func (es *encodeState) encodeArray(in interface{}) (err error) {
...
+		fmt.Println(i.Key(), "\t:", i.Value())
+
+		key := i.Key()
+		err = es.marshal(key.Interface())
+		if err != nil {
+			return
```
qdm12, line 241:
```suggestion
			return fmt.Errorf("encoding map key: %w", err)
```

pkg/scale/encode.go
```diff
@@ -223,6 +225,35 @@ func (es *encodeState) encodeArray(in interface{}) (err error) {
...
+		}
+
+		mapValue := i.Value()
+		if !mapValue.CanInterface() {
+			continue
+		}
```
qdm12, line 247: Shouldn't we return an error in that case, to avoid silently discarding a map value when encoding?

@timwu20

pkg/scale/encode.go
```diff
@@ -223,6 +225,35 @@ func (es *encodeState) encodeArray(in interface{}) (err error) {
...
+
+		err = es.marshal(mapValue.Interface())
+		if err != nil {
+			return
```
qdm12, line 251:
```suggestion
			return fmt.Errorf("encoding map value: %w", err)
```

pkg/scale/encode.go
```diff
@@ -223,6 +225,35 @@ func (es *encodeState) encodeArray(in interface{}) (err error) {
...
+		}
+	}
+	return
```
qdm12, line 254: nit

```suggestion
	return nil
```

qdm12 (commented):
Nice, thanks :+1: 
Note one test is failing in the CI.

EclesioMeloJunior (changes requested):
Nice PR! I would like to suggest providing more test cases, for example a map from string to a struct, like:

```go
type exampleStruct struct {
  Acc string
  Value int
}

toEncode := map[string]exampleStruct{}
```

pkg/scale/decode.go
kishansagathiya: It is dstv everywhere in that file, there is no need to change this

pkg/scale/decode_test.go
qdm12: Only if `t.Parallel()` is used on the subtest :wink:

jimjbrettj:
Seems rpc_03-chain_test is failing

pkg/scale/decode_test.go
```diff
@@ -167,7 +282,14 @@ func Test_unmarshal_optionality(t *testing.T) {
 					t.Errorf("decodeState.unmarshal() = %s", diff)
 				}
 			default:
-				dst := reflect.New(reflect.TypeOf(tt.in)).Interface()
```
jimjbrettj, line 170: Why are can we not treat maps like the other types here? Could we not define dst as the type of the input and unmarshal?

pkg/scale/encode.go
jimjbrettj: Def wait for Tims answer, but this happens at other places in this file as well (i.e. structs)

pkg/scale/encode_test.go
```diff
@@ -909,9 +909,28 @@ var (
 		},
 	}
 
+	mapTests = tests{
+		{
+			name: "testMap1",
+			in:   map[int8][]byte{2: []byte("some string")},
+			want: []byte{4, 2, 44, 115, 111, 109, 101, 32, 115, 116, 114, 105, 110, 103},
```
jimjbrettj, line 916: For both encoding and decoding, were any of the expected values derived from substrate? Or were the all derived from our own scale implementation?

jimjbrettj (commented):
Great work, just have a few questions

pkg/scale/decode_test.go
axaysagathiya: if we use `dst := reflect.New(reflect.TypeOf(tt.in)).Interface()` for map,

- it panics in Unmarshal    
>  **panic: assignment to entry in nil map**

-   for `in` of type `map[int8][]byte`, it returns variable of type `*map[int8][]uint8` instead of `*map[int8][]byte`

so need to initialize the map and specify the type of `dst` same as type of `in`.

pkg/scale/decode_test.go
```diff
@@ -132,6 +132,121 @@ func Test_decodeState_decodeSlice(t *testing.T) {
 	}
 }
 
+// // Rust code to encode a map of string to struct.
+// let mut btree_map: BTreeMap<String, User> = BTreeMap::new();
```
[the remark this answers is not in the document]
axaysagathiya: This is Rust code used to get expected encoded data.

pkg/scale/encode_test.go
axaysagathiya: yes all derived from substrate.

axaysagathiya:
@qdm12  

Can you please rerun the test.
 Intermittent test `rpc_03-chain_test.go ` is  failing.

pkg/scale/decode_test.go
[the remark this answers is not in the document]
EclesioMeloJunior: @axaysagathiya I think it is not needed to put this comment block here, could you provide a link to the actual Rust code that does the encode?

EclesioMeloJunior (changes requested):

pkg/scale/decode_test.go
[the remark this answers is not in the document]
axaysagathiya: @EclesioMeloJunior this Rust code is written by me to get scale encoded data. Do you want me to provide a Gist link ?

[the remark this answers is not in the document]
EclesioMeloJunior: Not really, I mean your comment starts with: 

```
// // Rust code to encode a map of string to struct.
```

But actually, there is no code that encodes a map in your comment block, you're just adding two entries into a `BTreeMap`, that is why I would suggest removing this comment

pkg/scale/decode_test.go
jimjbrettj: cool cool that makes sense, thank you!

pkg/scale/decode_test.go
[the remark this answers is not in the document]
jimjbrettj: @kishansagathiya cool, I thought it was taken from somewhere else so I am fine to leave it as is. Want to avoid situations where we don't know where bytes came from.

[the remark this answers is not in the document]
EclesioMeloJunior: Oh, okay I did understand that you were trying to explain how rust encodes but you're explaining how you generated the encoded bytes that you're using in the test (my bad 😅)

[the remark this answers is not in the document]
kishansagathiya: > Oh, okay I did understand that you were trying to explain how rust encodes but you're explaining how you generated the encoded bytes that you're using in the test (my bad sweat_smile)

:laughing:

EclesioMeloJunior (approved):

pkg/scale/encode.go
timwu20: 'CanInterface` is essentially determining if this is a public attribute of a struct or public method.  I think it should be fine in this case.  This will almost always return true in the case we're trying to decode into a map value.  I wonder what cases this would return false though, maybe we can provide a test case.

@kishansagathiya please don't merge the PR with unresolved conversations.  My bad for not getting to this earlier.

pkg/scale/encode.go
```diff
@@ -223,6 +226,40 @@ func (es *encodeState) encodeArray(in interface{}) (err error) {
...
+		return fmt.Errorf("encoding length: %w", err)
+	}
+
+	mapKeys := v.MapKeys()
+
+	sort.Slice(mapKeys, func(i, j int) bool {
+		keyByteOfI, _ := Marshal(mapKeys[i].Interface())
+		keyByteOfJ, _ := Marshal(mapKeys[j].Interface())
+		return bytes.Compare(keyByteOfI, keyByteOfJ) < 0
+	})
```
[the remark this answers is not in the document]
axaysagathiya: yeah

pkg/scale/encode.go
kishansagathiya: I realised about this comment few moments after merging it. Wasn't deliberate.

pkg/scale/encode.go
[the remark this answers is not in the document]
timwu20: @axaysagathiya can you create another PR to only do this in the tests? This is not required when encoding maps in production, and it's not very performant.  I'd suggest adding a `sortKeys` parameter to the `encodeMap` function signature.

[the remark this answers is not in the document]
axaysagathiya: Oh, my bad! 
> This is not required when encoding maps

This is not just needed in test, this is needed in encoding as well. keys in go maps are unordered. so running encode function on the same go map could result into different encodings. Rust equivalent of go map is BTree map, where keys are sorted. Scale encode function in the Rust implementation does not produces different values on running it multiple times.

So, in order to make sure that our scale package and Rust's scale package encodes a map to the same value sorting is necessary.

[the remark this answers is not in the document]
timwu20: Just because the `BTreeMap` has deterministic sorting of keys, doesn't mean that the `parity-scale-codec` isn't able to decode maps that were encoded with keys in non-deterministic order.  In go it's understood that the order of the keys is non-deterministic.  In this case, we should be testing that we can encode a map and decode into a map, and assert that the maps are equal.  We could also ensure that decoding into a rust `BTreeMap` via the `parity-scale-codec` works as expected.  To make the tests more deterministic, you could provide a map with a single key or a map with two keys, and just assert that it's equal to one of the two encodings and remove the sorting logic altogether.

PR #2897, opened 2022-10-20
changing errors.New to fmt.Errorf

kishansagathiya:
## Tests

```sh
```

## Primary Reviewer

@timwu20


EclesioMeloJunior (commented):
There is no need to change `errors.New` (used to creating sentinel errors) for `fmt.Errorf` (used to wrap errors in an error chain)

PR #2902, opened 2022-10-27
fix(dot/state): fix tries not being GCed

qdm12:


qdm12:
Does not work, memory usage unchanged, the leak is coming from somewhere else maybe #2912

PR #2906, opened 2022-10-29
fix(dot/sync): remove block announcement in `bootstrap` sync mode

### LINKED ISSUE #2905: Remove block announcement in `bootstrap` sync mode
## Issue summary

- Remove block announcement in `bootstrap` sync mode

- N/A

EclesioMeloJunior:
## Changes

- Do not announce blocks while in the `bootstrap` sync mode

## Tests

```sh
go test -tags integration ./dot/sync/...
```

## Issues

- Closes #2905

## Primary Reviewer

@timwu20


dot/network/notifications_test.go
```diff
@@ -196,6 +197,11 @@ func TestCreateNotificationsMessageHandler_BlockAnnounceHandshake(t *testing.T)
 
 	err = handler(stream, testHandshake)
 	require.ErrorIs(t, err, errCannotValidateHandshake)
+
+	expectedError := fmt.Errorf("handling handshake: %w from peer %s using protocol %s: %s",
+		errCannotValidateHandshake, testPeerID, info.protocolID, errors.New("genesis hash mismatch"))
+	require.EqualError(t, err, expectedError.Error())
```
qdm12, line 203: nit

```suggestion
	expectedErrorMessage := fmt.Sprintf("handling handshake: %s from peer %s using protocol %s: genesis hash mismatch",
		errCannotValidateHandshake, testPeerID, info.protocolID)
	require.EqualError(t, err, expectedErrorMessage)
```

dot/sync/chain_processor.go
```diff
@@ -42,24 +44,35 @@ type chainProcessor struct {
...
+	blockState         BlockState
+	storageState       StorageState
+	transactionState   TransactionState
+	babeVerifier       BabeVerifier
+	finalityGadget     FinalityGadget
+	blockImportHandler BlockImportHandler
+	telemetry          telemetry.Client
+}
+
+func newChainProcessor(cfg *chainProcessorConfig) *chainProcessor {
```
qdm12, line 60: nit I'd revert this, I don't think this delta is needed nor related to this PR. Especially for unexported constructors, I'd say it's fine to list many arguments.

dot/core/service.go
```diff
@@ -127,12 +127,16 @@ func (s *Service) StorageRoot() (common.Hash, error) {
 }
 
 // HandleBlockImport handles a block that was imported via the network
-func (s *Service) HandleBlockImport(block *types.Block, state *rtstorage.TrieState) error {
+func (s *Service) HandleBlockImport(block *types.Block, state *rtstorage.TrieState, announce bool) error {
```
qdm12, line 130: nit maybe rename `announce` to `announceImportedBlock`?

qdm12 (approved):

qdm12:
Why not have the base branch as development though?

dot/sync/chain_processor.go
EclesioMeloJunior: I will keep it as it is because we already use a configuration struct for the `chain_sync` so adding this configuration struct for `chain_processor` keep them in the same pattern. I have changed to accept a value instead of a pointer.

dot/network/notifications_test.go
```diff
@@ -196,6 +197,11 @@ func TestCreateNotificationsMessageHandler_BlockAnnounceHandshake(t *testing.T)
 
 	err = handler(stream, testHandshake)
 	require.ErrorIs(t, err, errCannotValidateHandshake)
+
+	expectedErrorMessage := fmt.Sprintf("handling handshake: %s from peer %s using protocol %s: genesis hash mismatch",
+		errCannotValidateHandshake, testPeerID, info.protocolID)
+	require.EqualError(t, err, expectedErrorMessage)
```
timwu20, line 203: we're expecting this error now?

EclesioMeloJunior: No, this is just an improvement in the test, not related to the block announcement directly

dot/sync/chain_processor.go
qdm12: Ok, although it's still kinda out of scope. Would you mind opening another (easy) PR for those config changes?

qdm12 (approved):
Just ideally move out the config struct changes to its own PR :wink:

dot/sync/chain_processor.go
```diff
@@ -42,24 +44,35 @@ type chainProcessor struct {
 	telemetry          telemetry.Client
 }
 
-func newChainProcessor(readyBlocks *blockQueue, pendingBlocks DisjointBlockSet,
```
kishansagathiya, line 45: `chainProcessorConfig` and `chainProcessor` are basically the same struct apart from ctx and cancel. `chainProcessorConfig` isn't useful apart from providing better indentation.
I would suggest to remove `chainProcessorConfig` and modify `newChainProcessor` everywhere to provide better indentation.

Like given below, (here and everywhere `newChainProcessor` is used)
```
func newChainProcessor(
	readyBlocks *blockQueue, 
	pendingBlocks DisjointBlockSet,
	blockState BlockState, 
	storageState StorageState,
	transactionState TransactionState, 
	babeVerifier BabeVerifier,
	finalityGadget FinalityGadget, 
	blockImportHandler BlockImportHandler, 
	telemetry telemetry.Client) 
*chainProcessor {
```

dot/sync/chain_sync.go
```diff
@@ -176,7 +176,7 @@ type chainSyncConfig struct {
 	slotDuration       time.Duration
 }
 
-func newChainSync(cfg *chainSyncConfig) *chainSync {
```
kishansagathiya, line 179: What's the motivation behind making `chainSyncConfig` from pointer to value?

kishansagathiya (commented):
LGTM apart from two comments

dot/sync/chain_processor.go
qdm12: For PR size you're right, it's more for git history/git blame in IDE. It happens from time to time someone checks some old commit for X or Y reason, and it helps when it's focused. It's also fine to do side code changes if you need to change that code too, for example refactor an affected test

dot/sync/chain_sync.go
qdm12: I had [this suggestion](https://github.com/ChainSafe/gossamer/pull/2906#discussion_r1009571677) for `func newChainProcessor(cfg *chainProcessorConfig) *chainProcessor`

> Also side-nit, if you do pass a settings struct, pass it as a value, not a pointer, there is really need no need to pass it as a pointer (no performance impact + fields are already interfaces + opens the door to pass nil)

Although I guess it applies in several places like here. Anyway imo those changes do not belong to this PR

dot/sync/chain_processor.go
qdm12: :warning: repeating myself, but these are out of scope changes, we should have them in a separate ez PR.

Eclesio's [argument](https://github.com/ChainSafe/gossamer/pull/2906#discussion_r1010818200) was:

> I will keep it as it is because we already use a configuration struct for the chain_sync so adding this configuration struct for chain_processor keep them in the same pattern. I have changed to accept a value instead of a pointer.

Although I agree we should standardize our codebase as we progress through it, I agree with Kishan's suggestion especially for unexported/internal constructors.

I would also add in my very opinionated opinion that config structs should be reserved for the public Go API (so exported public constructors) and every field can be set to a default (so the user can pass an empty struct). It also allows to add fields without breaking the constructor signature (so it stays retro-compat). 

For internal constructors, it's far easier to have plain and simple compulsory named arguments which would detect compilation problems when adding an argument.

timwu20 (approved):

kishansagathiya:
@timwu20 This has many unresolved conversations

PR #2912, opened 2022-11-01
chore(state): review usages of `TrieState`

qdm12:
## Changes

- Split out snapshotting feature out of the `TrieState` method
- Add `TrieState` method `Snapshot` returning a snapshotted trie state
- `GetRuntimeMetadata` and `GetRuntimeVersion` do not need the trie
- Temporary modifiable trie snapshot for each pending transaction (instead of sharing the same trie)
- Move snapshotting operations closer to where they are needed
- Rework some related error wrappings and tests
- Remove impossible nil trie error check

## Tests

```sh
```

## Primary Reviewer

@timwu20

### DIFF
--- a/dot/core/messages.go
+++ b/dot/core/messages.go
@@ -20,12 +20,18 @@ func (s *Service) validateTransaction(head *types.Header, rt RuntimeInstance,
 	s.storageState.Lock()
 
 	ts, err := s.storageState.TrieState(&head.StateRoot)
-	s.storageState.Unlock()
 	if err != nil {
+		s.storageState.Unlock()
 		return nil, fmt.Errorf("getting trie state from storage: %w", err)
 	}
 
-	rt.SetContextStorage(ts)
+	// ValidateTransaction modifies the trie state so we want to snapshot it
+	// so that the original trie state remains unaffected.
+	temporaryState := ts.Snapshot()
+
+	s.storageState.Unlock()
+
+	rt.SetContextStorage(temporaryState)
 
 	// validate each transaction
 	externalExt, err := s.buildExternalTransaction(rt, tx)
@@ -50,8 +56,8 @@ func (s *Service) validateTransaction(head *types.Header, rt RuntimeInstance,
 
 // HandleTransactionMessage validates each transaction in the message and
 // adds valid transactions to the transaction queue of the BABE session
-// returns boolean for transaction propagation, true - transactions should be propagated
-func (s *Service) HandleTransactionMessage(peerID peer.ID, msg *network.TransactionMessage) (bool, error) {
+func (s *Service) HandleTransactionMessage(peerID peer.ID, msg *network.TransactionMessage) (
+	propagateTransactions bool, err error) {
 	logger.Debug("received TransactionMessage")
 
 	if !s.net.IsSynced() {
@@ -65,13 +71,13 @@ func (s *Service) HandleTransactionMessage(peerID peer.ID, msg *network.Transact
 
 	head, err := s.blockState.BestBlockHeader()
 	if err != nil {
-		return false, err
+		return false, fmt.Errorf("getting best block header: %w", err)
 	}
 
 	bestBlockHash := head.Hash()
 	rt, err := s.blockState.GetRuntime(bestBlockHash)
 	if err != nil {
-		return false, err
+		return false, fmt.Errorf("getting runtime from block state: %w", err)
 	}
 
 	allTxnsAreValid := true
--- a/dot/core/messages.go
+++ b/dot/core/messages.go
@@ -31,14 +31,13 @@ func (s *Service) validateTransaction(head *types.Header, rt RuntimeInstance,
 
 	s.storageState.Unlock()
 
-	rt.SetContextStorage(temporaryState)
-
 	// validate each transaction
 	externalExt, err := s.buildExternalTransaction(rt, tx)
 	if err != nil {
 		return nil, fmt.Errorf("building external transaction: %w", err)
 	}
 
+	rt.SetContextStorage(temporaryState)
 	validity, err = rt.ValidateTransaction(externalExt)
 	if err != nil {
 		logger.Debugf("failed to validate transaction: %s", err)
--- a/dot/core/messages.go
+++ b/dot/core/messages.go
@@ -43,6 +43,7 @@ func (s *Service) validateTransaction(head *types.Header, rt RuntimeInstance,
 		logger.Debugf("failed to validate transaction: %s", err)
 		return nil, err
 	}
+	rt.SetContextStorage(nil)
 
 	vtx := transaction.NewValidTransaction(tx, validity)
 
--- a/dot/core/messages.go
+++ b/dot/core/messages.go
@@ -22,7 +22,7 @@ func (s *Service) validateTransaction(head *types.Header, rt RuntimeInstance,
 	ts, err := s.storageState.TrieState(&head.StateRoot)
 	s.storageState.Unlock()
 	if err != nil {
-		return nil, fmt.Errorf("cannot get trie state from storage for root %s: %w", head.StateRoot, err)
+		return nil, fmt.Errorf("getting trie state from storage: %w", err)
 	}
 
 	rt.SetContextStorage(ts)
--- a/dot/core/messages_test.go
+++ b/dot/core/messages_test.go
@@ -179,6 +179,7 @@ func Test_Service_HandleTransactionMessage(t *testing.T) {
 				}
 				runtimeInstance.EXPECT().Version().Return(version)
 				runtimeInstance.EXPECT().SetContextStorage(trieState.Snapshot())
+				runtimeInstance.EXPECT().SetContextStorage(nil)
 				validity := &transaction.Validity{Propagate: true}
 				externalExtrinsic := []byte{2, 1, 2, 3}
 				runtimeInstance.EXPECT().ValidateTransaction(externalExtrinsic).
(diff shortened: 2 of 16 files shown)


dot/core/service.go
```diff
@@ -526,17 +536,21 @@ func (s *Service) HandleSubmittedExtrinsic(ext types.Extrinsic) error {
...
+	// ValidateTransaction modifies the trie state so we want to snapshot it
+	// so that the original trie state remains unaffected.
+	temporaryState := ts.Snapshot()
+	rt.SetContextStorage(temporaryState)
 	transactionValidity, err := rt.ValidateTransaction(externalExt)
 	if err != nil {
 		return err
 	}
+	// TODO adding this fails polkadotjs tests
+	// rt.SetContextStorage(nil)
```
qdm12, line 553: Not sure why, the whole program crashes badly, anyone would know why?

dot/sync/chain_processor.go
```diff
@@ -196,6 +196,11 @@ func (c *chainProcessor) processBlockDataWithStateHeaderAndBody(blockData types.
 		return fmt.Errorf("loading trie state: %w", err)
 	}
 
+	// Note we don't want to snapshot the state since we want to modify the underlying trie
+	// for the block header state root.
+	// TODO remove this:
+	state = state.Snapshot()
```
qdm12, line 202: TODO remove this

dot/sync/chain_processor.go
```diff
@@ -258,15 +263,19 @@ func (s *chainProcessor) handleBlock(block *types.Block, announceImportedBlock b
...
 	if err != nil {
+		rt.SetContextStorage(nil)
 		return fmt.Errorf("failed to execute block %d: %w", block.Header.Number, err)
 	}
 
-	if err = s.blockImportHandler.HandleBlockImport(block, ts, announceImportedBlock); err != nil {
-		return err
+	rt.SetContextStorage(nil)
+
+	if err = s.blockImportHandler.HandleBlockImport(block, nextStorageState, announceImportedBlock); err != nil {
```
qdm12, line 277: nit split

dot/sync/chain_processor.go
```diff
@@ -196,6 +196,9 @@ func (c *chainProcessor) processBlockDataWithStateHeaderAndBody(blockData types.
 		return fmt.Errorf("loading trie state: %w", err)
 	}
 
+	// Note we don't want to snapshot the state since we want to modify the underlying trie
+	// for the block header state root.
+
```
qdm12, line 201: If someone could confirm this just in case :tm:

dot/core/service.go
jimjbrettj: maybe tests are wrong? I feel like we should be setting the context yes?

qdm12: Oh yeah definitely, but clearing it (setting it to `nil`) makes it panic ugly (cgo horrible trace)

dot/core/messages_test.go
```diff
@@ -112,295 +58,362 @@ func TestService_TransactionsCount(t *testing.T) {
...
+		propagateTransactions bool
+		errSentinel           error
+		errMessage            string
+		expectedMessage       *network.TransactionMessage
 	}{
-		{
-			name: "not synced",
-			mockNetwork: &mockNetwork{
-				IsSynced: false,
+		"not synced": {
```
EclesioMeloJunior, line 81: would be nice to change the tests names to `snake_case` pattern as they all appears as diffs

dot/core/service.go
```diff
@@ -520,17 +530,21 @@ func (s *Service) HandleSubmittedExtrinsic(ext types.Extrinsic) error {
...
 
+	// ValidateTransaction modifies the trie state so we want to snapshot it
+	// so that the original trie state remains unaffected.
+	temporaryState := ts.Snapshot()
+	rt.SetContextStorage(temporaryState)
 	transactionValidity, err := rt.ValidateTransaction(externalExt)
 	if err != nil {
 		return err
 	}
+	// TODO adding this fails polkadotjs tests
```
EclesioMeloJunior, line 546: could an issue be created for this TODO?

EclesioMeloJunior (approved):

dot/core/messages_test.go
qdm12: Indeed, done 👍 

Created https://github.com/ChainSafe/gossamer/issues/3039 to do this for the entire codebase (with a regex and batch all)

dot/core/service.go
qdm12: Just fishing for a reviewer that may have an answer on this. Otherwise yes I'll create an issue for this.

dot/core/messages.go
```diff
@@ -20,24 +20,30 @@ func (s *Service) validateTransaction(head *types.Header, rt RuntimeInstance,
...
 		return nil, fmt.Errorf("building external transaction: %w", err)
 	}
 
+	rt.SetContextStorage(temporaryState)
 	validity, err = rt.ValidateTransaction(externalExt)
 	if err != nil {
 		logger.Debugf("failed to validate transaction: %s", err)
 		return nil, err
 	}
+	rt.SetContextStorage(nil)
```
timwu20, line 46: why do we have to do this now?

dot/core/service.go
```diff
@@ -426,13 +425,16 @@ func (s *Service) maintainTransactionPool(block *types.Block, bestBlockHash comm
...
 		externalExt, err := s.buildExternalTransaction(rt, tx.Extrinsic)
 		if err != nil {
 			return fmt.Errorf("building external transaction: %s", err)
 		}
 
+		// ValidateTransaction modifies the trie state so we use the temporary state.
+		temporaryState := ts.Snapshot()
+		rt.SetContextStorage(temporaryState)
 		txnValidity, err := rt.ValidateTransaction(externalExt)
+		rt.SetContextStorage(nil)
```
timwu20, line 437: why do we set it as `nil`? Shouldn't we restore back to original `ts`?

dot/core/service.go
```diff
@@ -482,14 +484,22 @@ func (s *Service) DecodeSessionKeys(enc []byte) ([]byte, error) {
...
+	}
+
+	runtime, err := s.blockState.GetRuntime(blockHashValue)
 	if err != nil {
-		return version, fmt.Errorf("setting up runtime: %w", err)
+		return version, fmt.Errorf("getting runtime: %w", err)
 	}
-	return rt.Version(), nil
+
+	return runtime.Version(), nil
```
timwu20, line 502: it would be nice to just cache this, so we don't need to execute the `Version` function every time.  Can also help when we don't store every running runtime instance.

dot/core/service.go
```diff
@@ -543,12 +557,19 @@ func (s *Service) HandleSubmittedExtrinsic(ext types.Extrinsic) error {
...
+		blockHashValue = s.blockState.BestBlockHash()
+	}
+	runtime, err := s.blockState.GetRuntime(blockHashValue)
 	if err != nil {
-		return nil, fmt.Errorf("setting up runtime: %w", err)
+		return nil, fmt.Errorf("getting runtime: %w", err)
 	}
-	return rt.Metadata()
+
+	return runtime.Metadata()
```
timwu20, line 572: same with this.. should we consider caching this?

dot/state/storage_notify_test.go
```diff
@@ -60,6 +61,7 @@ func TestStorageState_RegisterStorageObserver_Multi(t *testing.T) {
 	ss := newTestStorageState(t)
 	ts, err := ss.TrieState(nil)
 	require.NoError(t, err)
+	ts = ts.Snapshot()
```
timwu20, line 64: why do we need to snapshot if we are overwriting?

dot/state/test_helpers.go
```diff
@@ -240,6 +240,8 @@ func generateBlockWithRandomTrie(t *testing.T, serv *Service,
 	trieState, err := serv.Storage.TrieState(nil)
 	require.NoError(t, err)
 
+	trieState = trieState.Snapshot()
```
timwu20, line 243: again why do we need to snapshot here?

dot/sync/chain_processor.go
```diff
@@ -258,15 +261,20 @@ func (s *chainProcessor) handleBlock(block *types.Block, announceImportedBlock b
...
 
 	_, err = rt.ExecuteBlock(block)
 	if err != nil {
+		rt.SetContextStorage(nil)
 		return fmt.Errorf("failed to execute block %d: %w", block.Header.Number, err)
 	}
 
-	if err = s.blockImportHandler.HandleBlockImport(block, ts, announceImportedBlock); err != nil {
-		return err
+	rt.SetContextStorage(nil)
```
timwu20, line 273: could we not just defer this line before line 267

dot/core/messages.go
qdm12: Clearing the storage trie state for that runtime, so that it can be garbage collected. Otherwise the runtime may hold it forever.
The point is to reduce/avoid memory leaks, so this aims at setting the state and then clearing it from the runtime structure; all calls get a new trie state anyway

dot/core/service.go
qdm12: It's already cached, see https://github.com/ChainSafe/gossamer/blob/4ea0a703339ca06e617b06eeb7d56308e7b48093/lib/runtime/types.go#L59

dot/core/service.go
qdm12: Yes, created https://github.com/ChainSafe/gossamer/issues/3065

dot/state/storage_notify_test.go
qdm12: Because

1. `Snapshot()` was part of `TrieState()` previously
2. We don't want to modify the underlying data of the original `ts` from the storage state. We are overriding the `ts` variable, but that's just a pointer variable, and Snapshot gives it another pointer.

dot/state/test_helpers.go
qdm12: See https://github.com/ChainSafe/gossamer/pull/2912#discussion_r1085494292

dot/sync/chain_processor.go
qdm12: We could, I did that more-explicit-less-deferred to highlight the storage is no longer needed right after `rt.ExecuteBlock(block)`

dot/core/service.go
qdm12: Somehow this got fixed and the storage can now be cleared after this runtime call.

PR #2915, opened 2022-11-02
fix(state): node hashes vs merkle values

qdm12:
## Changes

- Pruners only care about node hashes
- Rename variables, functions and methods only dealing with node hashes
- Do not write or read inlined nodes with a non-hash Merkle value
- Clarify error wrappings and comments

:information_source: this may feel familiar since I made previous PRs to rename node hashes -> merkle value, but it turned out it was not applicable for database related things (but only learnt/realised that recently).

## Tests

```sh
```

## Primary Reviewer

@timwu20

### DIFF
--- a/dot/state/offline_pruner.go
+++ b/dot/state/offline_pruner.go
@@ -149,7 +149,7 @@ func (p *OfflinePruner) SetBloomFilter() (err error) {
 		}
 	}
 
-	logger.Infof("Total keys added in bloom filter: %d", len(nodeHashes))
+	logger.Infof("Total keys added in filter database: %d", len(nodeHashes))
 	return nil
 }
 
--- a/dot/state/offline_pruner.go
+++ b/dot/state/offline_pruner.go
@@ -116,7 +116,7 @@ func (p *OfflinePruner) SetBloomFilter() (err error) {
 	}
 
 	latestBlockNum := header.Number
-	merkleValues := make(map[string]struct{})
+	nodeHashes := make(map[common.Hash]struct{})
 
 	logger.Infof("Latest block number is %d", latestBlockNum)
 
@@ -132,7 +132,7 @@ func (p *OfflinePruner) SetBloomFilter() (err error) {
 			return err
 		}
 
-		trie.PopulateNodeHashes(tr.RootNode(), merkleValues)
+		trie.PopulateNodeHashes(tr.RootNode(), nodeHashes)
 
 		// get parent header of current block
 		header, err = p.blockState.GetHeader(header.ParentHash)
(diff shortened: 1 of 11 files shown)


dot/state/offline_pruner.go
```diff
@@ -131,14 +131,14 @@ func (p *OfflinePruner) SetBloomFilter() (err error) {
...
-		err = p.bloom.put([]byte(key))
+	for key := range nodeHashes {
+		err = p.bloom.put(key.ToBytes())
 		if err != nil {
 			return err
 		}
 	}
 
-	logger.Infof("Total keys added in bloom filter: %d", len(merkleValues))
+	logger.Infof("Total keys added in bloom filter: %d", len(nodeHashes))
```
EclesioMeloJunior, line 141: are we replacing the bloom filter? Should we change this log message to be more general?

```suggestion
	logger.Infof("Total keys added in prune filter: %d", len(nodeHashes))
```

dot/state/storage.go
```diff
@@ -64,12 +64,12 @@ func (s *StorageState) StoreTrie(ts *rtstorage.TrieState, header *types.Header)
...
 	if header != nil {
-		insertedMerkleValues, deletedMerkleValues, err := ts.GetChangedNodeHashes()
+		insertedNodeHashes, deletedNodeHashes, err := ts.GetChangedNodeHashes()
 		if err != nil {
-			return fmt.Errorf("failed to get state trie inserted keys: block %s %w", header.Hash(), err)
+			return fmt.Errorf("getting trie changed node hashes for block hash %s: %w", header.Hash(), err)
 		}
 
-		err = s.pruner.StoreJournalRecord(deletedMerkleValues, insertedMerkleValues, header.Hash(), int64(header.Number))
+		err = s.pruner.StoreJournalRecord(deletedNodeHashes, insertedNodeHashes, header.Hash(), int64(header.Number))
```
EclesioMeloJunior, line 72: as we are wrapping error, can we wrap this error as well?

lib/trie/database.go
```diff
@@ -305,11 +307,19 @@ func (t *Trie) writeDirtyNode(db Putter, n *Node) (err error) {
 			n.MerkleValue, err)
 	}
 
-	err = db.Put(merkleValue, encoding)
+	if len(merkleValue) < 32 {
+		// Inlined node, there is no need to write it to database.
```
EclesioMeloJunior, line 311: could you elaborate a bit more on this statement? I didn't understand it properly 😅. We don't need to write it on DB because it is already part of another encoding that is or will be stored in DB (is that right?)

EclesioMeloJunior (changes requested):

lib/trie/database.go
timwu20: I think this part of the polkadot spec describes it https://spec.polkadot.network/#defn-hashed-subvalue

timwu20 (approved):

lib/trie/database.go
qdm12: Changed to:

```go
		// Merkle value is the node encoding which is less than 32 bytes.
		// That means this node encoding is inlined in its parent node encoding,
		// and so it is not needed to write it in the database.
```

dot/state/storage.go
qdm12: Done 😉

dot/state/offline_pruner.go
qdm12: Changed the log line completely (should had done it in #3084 my bad)

EclesioMeloJunior (approved):

PR #2918, opened 2022-11-02
feat(lib/trie): offload > 32B subvalues to database

qdm12:
## Issues

Partially address #2838 and #1936 

## Primary Reviewer

@timwu20

### DIFF
--- a/dot/state/block.go
+++ b/dot/state/block.go
@@ -614,8 +614,10 @@ func (bs *BlockState) HandleRuntimeChanges(newState *rtstorage.TrieState,
 
 	logger.Infof("🔄 detected runtime code change, upgrading with block %s from previous code hash %s to new code hash %s...", //nolint:lll
 		bHash, codeHash, currCodeHash)
-	code := newState.LoadCode()
-	if len(code) == 0 {
+	code, err := newState.LoadCode()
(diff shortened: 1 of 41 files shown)


qdm12:
This is potentially quite wrong, so closing this PR.

When inserting a (key, value) in the trie, the subvalue of the node to be created (even if it already exists in the trie) should be computed, and, if it's larger than 32B, the node inserted should be one of the new node variants (such as branch with hash of subvalue), where its key is the original key from the (key, value) inserted, but its storage value is the hash of the subvalue computed previously. Another node must be inserted (as v0 trie node variants such as branch with value) with its full key being the hash of the subvalue computed previously, and the storage value being the actual subvalue bytes from previously. Note this second node is part of the trie and the merkle computation.

Another catch is that writing a (key, value) to the trie, where the key already exists and the value is the same as the one already there, should replace that node with the mechanism written in the paragraph above. This allows to later fully upgrade a v0 state trie to a v1 state trie.

Finally when getting a storage value from the trie given a certain key, if the node at the key has a node variant of leaf/branch containing a hashed subvalue, the storage value to return should be extracted from the storage value (the subvalue initially computed when inserting) of the other node looked up using the subvalue hash.

PR #2922, opened 2022-11-04
chore(lib/trie): remove `Store` method

qdm12:
## Changes

- `WriteDirty` is just the clever version of `Store`, where we write only nodes that got changed in-memory from the version we have in the database.
- This also ensures all the nodes are set as dirty correctly

## Tests

```sh
go test -tags integration github.com/ChainSafe/gossamer/...
```

## Issues

Trying to fix up strange root hash mismatch panic found in #2919 

## Primary Reviewer

@timwu20

### DIFF
--- a/dot/state/service_test.go
+++ b/dot/state/service_test.go
@@ -72,13 +72,24 @@ func TestService_Initialise(t *testing.T) {
 	state := newTestService(t)
 
 	genData, genTrie, genesisHeader := newTestGenesisWithTrieAndHeader(t)
+
+	// Deep copy the trie because of the following:
+	// Initialise clears the database, writes only dirty nodes to disk
+	// and marks them as clean.
+	// Initialise is called twice.
+	// The first call writes all nodes to disk and marks them as clean in the
+	// in-memory trie representation.
+	// If the same trie is re-used for the second call, the database is cleared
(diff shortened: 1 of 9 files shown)


dot/state/service_test.go
```diff
@@ -72,13 +72,24 @@ func TestService_Initialise(t *testing.T) {
...
+	genTrieCopy := genTrie.DeepCopy()
+
 	err := state.Initialise(&genData, &genesisHeader, &genTrie)
 	require.NoError(t, err)
 
 	genesisHeaderPtr := types.NewHeader(common.NewHash([]byte{77}),
 		genTrie.MustHash(), trie.EmptyHash, 0, types.NewDigest())
 
-	err = state.Initialise(&genData, genesisHeaderPtr, &genTrie)
+	err = state.Initialise(&genData, genesisHeaderPtr, genTrieCopy)
```
timwu20, line 92: Why is this test calling `Initialise` twice?

timwu20 (approved):

dot/state/service_test.go
qdm12: I think to test it overrides the previous `Initialise` call. Honestly that drove me mad for some minutes, and it's kinda weird, but a bit out of scope to change.

PR #2923, opened 2022-11-04
remove Deadlock from lib/grandpa/message_tracker.go 

### LINKED ISSUE #2841: Deadlock at `lib/grandpa/message_tracker.go`
## Describe the bug

The `message_tracker` contains the responsability to store consensus messages to be processed in the future, and the tracker checks every second if some message should be handled by the grandpa engine. But there is a possible bug inside the `handleTick()` method, the `handleTick()` method locks the message map and only unlocks after the method ends.

After locking it starts to iterate over `vote messages` which is not a problem and then starts to iterate over the `commit messages` and here we could face the deadlock.

For each `commit message` stored by the tracker, we call the `handleMessage` method from the grandpa engine which will process the commit message via `handleCommitMessage` and inside this function, in some cases (e.g. the block hash in the commit is not in the block tree yet), it is possible that it calls the tracker `addCommit()` method and this method will try to lock the tracker map which is already locked by the `handleTick()` and then the problem happens since `addCommit()` will wait until the release but the locker will only be released once the `handleTick()` finishes which will never happen.

## Expected Behavior

- Every second the tracker should retrieve and process the commits

## Current Behavior

- In some cases the tracker could hang forever while processing a specific commit

## Possible Solution

- We should lock the tracker map, retrieve all the commit messages from there, and then release the lock and then process each commit message.

kishansagathiya:
remove common mapLock from map tracker and have two separate locks for commit tracker and vote tracker

## Tests

```sh
```

## Issues

Fixes #2841

## Primary Reviewer

@EclesioMeloJunior


lib/grandpa/commits_tracker.go
```diff
@@ -14,6 +15,7 @@ import (
 // its maximum capacity is reached.
 // It is NOT THREAD SAFE to use.
 type commitsTracker struct {
+	sync.Mutex
```
qdm12, line 18: please don't embed the mutex to avoid exposing `.Lock()` and `.Unlock()` methods (even though it's not exported for now, it might be in the future). It also makes asserting the entire object cumbersome in tests as you've seen.

lib/grandpa/commits_tracker.go
```diff
@@ -90,6 +98,9 @@ func (ct *commitsTracker) delete(blockHash common.Hash) {
 // does not exist in the tracker
 func (ct *commitsTracker) message(blockHash common.Hash) (
 	message *CommitMessage) {
+	ct.Lock()
+	defer ct.Unlock()
```
qdm12, line 102: I guess you should lock the `forEach` method as well?

lib/grandpa/commits_tracker_test.go
```diff
@@ -51,7 +51,9 @@ func Test_newCommitsTracker(t *testing.T) {
 	}
 	vt := newCommitsTracker(capacity)
 
-	assert.Equal(t, expected, vt)
+	assert.Equal(t, expected.mapping, vt.mapping)
+	assert.Equal(t, expected.linkedList, vt.linkedList)
+	assert.Equal(t, expected.capacity, vt.capacity)
```
qdm12, line 56: Revert this once the mutex is a field and not embedded in the tracker.
The point of this assertion is to assert everything returned, asserting field by field opens room for errors for future modifications.

lib/grandpa/votes_tracker_test.go
```diff
@@ -70,7 +70,9 @@ func Test_newVotesTracker(t *testing.T) {
 	}
 	vt := newVotesTracker(capacity)
 
-	assert.Equal(t, expected, vt)
+	assert.Equal(t, expected.mapping, vt.mapping)
+	assert.Equal(t, expected.linkedList, vt.linkedList)
+	assert.Equal(t, expected.capacity, vt.capacity)
```
qdm12, line 75: Revert this once the mutex is a field and not embedded in the tracker.
The point of this assertion is to assert everything returned, asserting field by field opens room for errors for future modifications.

lib/grandpa/votes_tracker.go
```diff
@@ -16,6 +17,7 @@ import (
 // its maximum capacity is reached.
 // It is NOT THREAD SAFE to use.
 type votesTracker struct {
+	sync.Mutex
```
qdm12, line 20: please don't embed the mutex to avoid exposing `.Lock()` and `.Unlock()` methods (even though it's not exported for now, it might be in the future). It also makes asserting the entire object cumbersome in tests as you've seen.

lib/grandpa/votes_tracker.go
```diff
@@ -119,6 +127,9 @@ func (vt *votesTracker) delete(blockHash common.Hash) {
 // It returns nil if the block hash does not exist.
 func (vt *votesTracker) messages(blockHash common.Hash) (
 	messages []networkVoteMessage) {
+	vt.Lock()
+	defer vt.Unlock()
```
qdm12, line 131: nit this could be a rw mutex with read lock here

lib/grandpa/votes_tracker.go
```diff
@@ -138,6 +149,9 @@ func (vt *votesTracker) messages(blockHash common.Hash) (
 // as a slice of networkVoteMessages.
 func (vt *votesTracker) networkVoteMessages() (
 	messages []networkVoteMessage) {
+	vt.Lock()
+	defer vt.Unlock()
```
qdm12, line 153: nit this could be a rw mutex with read lock here

lib/grandpa/message_tracker_test.go
```diff
@@ -18,10 +19,10 @@ import (
 
 // getMessageFromVotesTracker returns the vote message
 // from the votes tracker for the given block hash and authority ID.
-func getMessageFromVotesTracker(votes votesTracker,
+func getMessageFromVotesTracker(votesMapping map[common.Hash]map[ed25519.PublicKeyBytes]*list.Element,
```
qdm12, line 22: nit rename the function to `getMessageFromVotesMapping`

lib/grandpa/commits_tracker.go
```diff
@@ -90,6 +98,9 @@ func (ct *commitsTracker) delete(blockHash common.Hash) {
 // does not exist in the tracker
 func (ct *commitsTracker) message(blockHash common.Hash) (
 	message *CommitMessage) {
+	ct.Lock()
```
qdm12, line 101: nit this could be a rw mutex with read lock here

qdm12 (changes requested):
~When do you get such deadlock? A bit more context would be nice :wink:~

EDIT: never mind I'm an idiot, I just read the issue linked :+1:

lib/grandpa/votes_tracker_test.go
```diff
@@ -70,7 +70,9 @@ func Test_newVotesTracker(t *testing.T) {
...
 }
```
qdm12, line 77: It would be nice to add a test ensuring thread safety. I wrote one up for the votes tracker, it can be adapted as well for the commits tracker:

```go
func Test_votesTracker_threadSafety(t *testing.T) {
	// This test is meant to be run with the `-race` flag
	// to detect any data race.
	t.Parallel()

	const capacity = 2
	votesTracker := newVotesTracker(capacity)

	const parallelism = 4

	var startWg, endWg sync.WaitGroup
	startWg.Add(parallelism)
	endWg.Add(parallelism)

	timedOut := make(chan struct{})
	go func() {
		startWg.Wait()
		const duration = 50 * time.Millisecond
		time.Sleep(duration)
		close(timedOut)
	}()

	for i := 0; i < parallelism; i++ {
		go func(i int) {
			defer endWg.Done()
			startWg.Done()

			const peerID = peer.ID("abc")
			blockHash := common.Hash{byte(i)}
			voteMessage := &VoteMessage{
				Message: SignedMessage{
					BlockHash: blockHash,
				},
			}

			for {
				select {
				case <-timedOut:
					return
				default:
				}

				votesTracker.add(peerID, voteMessage)
				votesTracker.delete(blockHash)
				_ = votesTracker.messages(blockHash)
				_ = votesTracker.networkVoteMessages()
			}
		}(i)
	}

	endWg.Wait()
}

```

lib/grandpa/commits_tracker.go
timwu20: I don't mind embedding the mutex in private types.  In the event that an instance of `commitsTracker` needs to be public in the future, we can revise the embedding then.

qdm12: Sure, but there is really no need to embed it

qdm12 (approved):

lib/grandpa/commits_tracker.go
EclesioMeloJunior: It is not possible to lock the `forEach`, the reason is: we call the `forEach` method and we pass to it, as an argument, a function that calls `t.commits.delete` which locks to remove the commit

```go
-> cannot lock for reading
t.commits.forEach(func(...) {
   ...
  -> already locks for delete
   t.commits.delete
})
```

EclesioMeloJunior (changes requested):
There are no tests related to the commits in the tracker so I would like to suggest adding tests to it

lib/grandpa/commits_tracker.go
EclesioMeloJunior: But I agree that we should lock while we are getting the commits from the tracker, but with the `forEach` method is not possible so I would like to suggest to replace the `forEach(func(cm *CommitMessage))` for a method that only retrieves the commits in the tracker instead of a method that receives a function and apply the function to each commit in the tracker.

EclesioMeloJunior (changes requested):

lib/grandpa/commits_tracker.go
qdm12: Good point, my bad for not checking.

Although, thinking more on this, shouldn't the entire 'forEach' operation lock then? Even if we don't need that much the mutex, one might later assume the commits tracker is thread safe due to most operations locking, so that's a risk for future changes.

> for a method that only retrieves the commits in the tracker

I wrote it as a forEach method for performance if we have a lot of tracked messages. See https://github.com/ChainSafe/gossamer/pull/2490#discussion_r856153149
Maybe there is no need for it though, I haven't had the time to add Prometheus metrics to check this out.

A (hopefully better) alternative would be to add a `deleteWithoutLock` method not locking the mutex, which you can then call in that `forEach` argument function, as well as in the `delete` method.

qdm12:
> There are no tests related to the commits in the tracker so I would like to suggest adding tests to it

What do you mean? There is `lib/grandpa/commits_tracker_test.go` right?

There are no tests for the callers of the commits tracker, although I have a local WIP branch with these (waiting for a grandpa refactor and/or something like #2575)

EclesioMeloJunior:
> What do you mean? There is lib/grandpa/commits_tracker_test.go right?

> There are no tests for the callers of the commits tracker, although I have a local WIP branch with these (waiting for a grandpa refactor and/or something like https://github.com/ChainSafe/gossamer/pull/2575)

The message tracker handles vote messages and commit messages, but in the `message_tracker_test.go` there is no tests related to adding a commit to the tracker and calling the `(t *tracker) handleTick` method, the test `TestMessageTracker_handleTick` only tests vote messages, so I would suggest to add one more case to tests the commit messages as well

qdm12:
> The message tracker handles vote messages and commit messages, but in the `message_tracker_test.go` there is no tests related to adding a commit to the tracker and calling the `(t *tracker) handleTick` method, the test `TestMessageTracker_handleTick` only tests vote messages, so I would suggest to add one more case to tests the commit messages as well

Yeah.... That's what I was trying to do a few months ago, but it's kinda of a pain in the ass to write good enough unit tests due to the impl details of the subtrackers (and the unexported fields of the linked lists)... so we would need interfaces... ideally internal subpackages for the 2 trackers and dep inject them to the parent tracker... but then we need #2575 ... anyway probably overthinking it lol...

We can for sure make an integration test without too much testing depth.

lib/grandpa/votes_tracker_test.go
kishansagathiya: @qdm12 Thanks for this. Adding this. 
It doesn't look like we run tests with `-race` automatically in CI. I only see `test-state-race`, which tests only for `dot/state`. I found similar data race tests in `dot/sync` and `lib/blocktree` as well. It doesn't look like they are running with `-race` at the moment. 

If that makes sense to you, I can create a separate small PR, adding  a make target say `test-race` and run those in CI as well.

qdm12: Yes that's true, the problem is `-race` makes tests a lot slower and use more memory, so we should really reserve it for unit tests. Now a bunch of unit tests are really integration tests, but it should be fixed more or less in #2887. What you can do is create an issue (or draft PR up to you) to add some `go test -race` workflows, once #2887 is merged?

lib/grandpa/votes_tracker.go
kishansagathiya: https://github.com/ChainSafe/gossamer/pull/2923#discussion_r1021518744

lib/grandpa/commits_tracker.go
kishansagathiya: https://github.com/ChainSafe/gossamer/pull/2923#discussion_r1021518744

lib/grandpa/votes_tracker_test.go
kishansagathiya: Created an issue https://github.com/ChainSafe/gossamer/issues/2946

EclesioMeloJunior (approved):
LGTM!

lib/grandpa/commits_tracker_test.go
kishansagathiya: resolving embedding related comments

lib/grandpa/votes_tracker_test.go
kishansagathiya: resolving embedding related comments

lib/grandpa/votes_tracker.go
kishansagathiya: not embedding does not fix this. I still can't use `assert.Equal(t, expected, vt)` as long as there is mutex in the struct

PR #2927, opened 2022-11-07
fix(trie): disallow empty byte slice node values

qdm12:
## Changes

- Always encode empty values for leaf nodes only
- Force empty byte slice values to `nil` when inserting
- Remove now unneeded `SubValueEqual` method
- Add extra test cases

Notes:

- Scale encoding:
  - `nil` -> `{0}`
  - `{}` -> `{0}`
- Scale decoding:
  - `nil` -> `nil` (because of `io.EOF` error)
  - `{}` -> `nil` (because of `io.EOF` error)
  -  `{0}` -> `{}` **NOT NIL!**

Unblocks #2919 and might potentially resolve other bugs

## Tests

```sh
```

## Issues

Related to #2838 

## Primary Reviewer

@timwu20


internal/trie/node/encode.go
```diff
@@ -57,14 +59,14 @@ func (n *Node) Encode(buffer Buffer) (err error) {
...
-	if n.Kind() == Branch {
+	if nodeIsBranch {
 		err = encodeChildrenOpportunisticParallel(n.Children, buffer)
 		if err != nil {
 			return fmt.Errorf("cannot encode children of branch: %w", err)
 		}
 	}
 
-	if n.Kind() == Leaf {
+	if kind == Leaf {
```
timwu20, line 69: seems like this should be an `else` statement

EclesioMeloJunior (approved):

timwu20 (approved):

Issue #2920, opened 2022-11-03
Panic during syncing of Polkadot


timwu20:
## Describe the bug

- The node panics intermittently (not at the same block height) on staging syncing setups. 
- From the AWS logs of the polkadot gossamer node:
```
panic: failed to re-add transactions to chain upon re-org: failed to get runtime instance
```

## Expected Behavior

- Does not panic

## Current Behavior

- We introduced these panics in a previous PR to notify us if these things happen on long running nodes.

## Log output

<details>
  <summary>Log Output</summary>
  
  ```Paste log output here
  panic: failed to re-add transactions to chain upon re-org: failed to get runtime instance
  ```
</details> 
</br>

## Specification

- go version: `1.18`
- gossamer version: `development`
- gossamer commit tag: NA
- gossamer commit hash: NA
- operating system: Ubuntu 19.10
- additional links: NA

jimjbrettj:
resolved with #2934

PR #2934, opened 2022-11-09
bug(lib/blocktree): fix blocktree highest common ancestor

jimjbrettj:
## Changes

- rewrite logic for finding the highest ancestor between 2 nodes
- rename to lowest to mimic parity
- remove old function and replaces usages with new one
- write tests for new func

## Tests

```sh
go test -tags integration ./lib/grandpa
go test ./core/service
go test ./lib/blocktree
```

## Issues

#2920 

## Primary Reviewer

@timwu20


dot/core/interface.go
```diff
@@ -66,12 +66,12 @@ type BlockState interface {
 	FreeImportedBlockNotifierChannel(ch chan *types.Block)
 	GetFinalisedNotifierChannel() chan *types.FinalisationInfo
 	FreeFinalisedNotifierChannel(ch chan *types.FinalisationInfo)
-	HighestCommonAncestor(a, b common.Hash) (common.Hash, error)
 	SubChain(start, end common.Hash) ([]common.Hash, error)
 	GetBlockBody(hash common.Hash) (*types.Body, error)
 	HandleRuntimeChanges(newState *rtstorage.TrieState, in runtime.Instance, bHash common.Hash) error
 	GetRuntime(blockHash common.Hash) (instance runtime.Instance, err error)
 	StoreRuntime(common.Hash, runtime.Instance)
+	LowestCommonAncestor(a, b common.Hash) (common.Hash, error)
```
qdm12, line 74: nit whilst we're at it, it would be nice to name the return (I guess `commonAncestorBlockHash`?)

dot/state/block.go
```diff
@@ -561,9 +561,9 @@ func (bs *BlockState) IsDescendantOf(parent, child common.Hash) (bool, error) {
 	return bs.bt.IsDescendantOf(parent, child)
 }
 
-// HighestCommonAncestor returns the block with the highest number that is an ancestor of both a and b
-func (bs *BlockState) HighestCommonAncestor(a, b common.Hash) (common.Hash, error) {
-	return bs.bt.HighestCommonAncestor(a, b)
+// LowestCommonAncestor returns the lowest common ancestor between two blocks in the tree.
+func (bs *BlockState) LowestCommonAncestor(a, b common.Hash) (common.Hash, error) {
```
qdm12, line 565: nit a bit of a long name, but still useful for documentation

```suggestion
func (bs *BlockState) LowestCommonAncestor(a, b common.Hash) (
	lowestCommonAncestorBlockHash common.Hash, err error) {
```

lib/blocktree/blocktree.go
```diff
@@ -297,29 +298,51 @@ func (bt *BlockTree) Leaves() []Hash {
...
 
-// HighestCommonAncestor returns the highest block that is a Ancestor to both a and b
-func (bt *BlockTree) HighestCommonAncestor(a, b Hash) (Hash, error) {
-	bt.RLock()
-	defer bt.RUnlock()
-
-	an := bt.getNode(a)
-	if an == nil {
+// LowestCommonAncestor returns the lowest common ancestor between two blocks in the tree.
+func (bt *BlockTree) LowestCommonAncestor(a, b Hash) (Hash, error) {
```
qdm12, line 302: nit

```suggestion
func (bt *BlockTree) LowestCommonAncestor(a, b Hash) (
	lowestCommonAncestorBlockHash Hash, err error) {
```

lib/blocktree/blocktree.go
```diff
@@ -297,29 +298,51 @@ func (bt *BlockTree) Leaves() []Hash {
 	return la
 }
 
-// HighestCommonAncestor returns the highest block that is a Ancestor to both a and b
-func (bt *BlockTree) HighestCommonAncestor(a, b Hash) (Hash, error) {
-	bt.RLock()
-	defer bt.RUnlock()
```
qdm12, line 303: is it ok we don't lock the block tree anymore? :thinking:

lib/blocktree/blocktree.go
```diff
@@ -297,29 +298,51 @@ func (bt *BlockTree) Leaves() []Hash {
...
-
-	an := bt.getNode(a)
-	if an == nil {
+// LowestCommonAncestor returns the lowest common ancestor between two blocks in the tree.
+func (bt *BlockTree) LowestCommonAncestor(a, b Hash) (Hash, error) {
+	// Get nodes to iterate through
```
qdm12, line 303: unneeded comment and also kinda wrong, we just get a single node corresponding to block hash A.. Reading more, maybe you meant getting nodes for A and B, but this is a bit confusing to read, I'd just remove :wink: 

```suggestion
```

lib/blocktree/blocktree.go
```diff
@@ -297,29 +298,51 @@ func (bt *BlockTree) Leaves() []Hash {
...
+	return lowestCommonAncestor(aNode, bNode)
+}
+func lowestCommonAncestor(aNode, bNode *node) (Hash, error) {
+	// balance out block height to compare
+	aNum := aNode.number
+	bNum := bNode.number
+	if aNum > bNum {
+		diff := aNum - bNum
+		if diff > aNum {
+			return common.Hash{}, errors.New("out of bounds ancestor check")
```
qdm12, line 322: nit define a sentinel error (same for below), and also add context such as `aNum` and `bNum`

lib/blocktree/blocktree.go
```diff
@@ -297,29 +298,51 @@ func (bt *BlockTree) Leaves() []Hash {
...
+		}
+		for diff > 0 {
+			aNode = aNode.parent
+			diff--
+		}
+	} else if bNum > aNum {
```
qdm12, line 328: these two if blocks are rather similar, maybe you can find a way to dedup them? I'd guess setting a `higherNode` and `lowerNode` variables depending on `a` and `b` block numbers.

lib/blocktree/blocktree.go
```diff
@@ -297,29 +298,51 @@ func (bt *BlockTree) Leaves() []Hash {
...
+		}
+		for diff > 0 {
+			bNode = bNode.parent
+			diff--
+		}
 	}
 
-	return ancestor.hash, nil
+	for {
+		if aNode.hash == bNode.hash {
```
qdm12, line 340: nit does this work fine? If yes, we should remove the `Equal` method it's rather pointless (not in this PR)...

lib/blocktree/blocktree.go
```diff
@@ -297,29 +298,51 @@ func (bt *BlockTree) Leaves() []Hash {
...
+			return aNode.hash, nil
+		}
+		aNode = aNode.parent
+		bNode = bNode.parent
+	}
```
qdm12, line 345: Add a check when the parent is nil so we don't go above the root, that would panic

lib/blocktree/blocktree_test.go
```diff
@@ -371,7 +371,82 @@ func TestBlockTree_IsDecendantOf(t *testing.T) {
...
+	require.NotNil(t, ancestor)
+	require.Equal(t, root.hash, ancestor)
+
+	ancestor, err = lowestCommonAncestor(finalChild[0], childrenChildren[0])
+	require.NoError(t, err)
+	require.NotNil(t, ancestor)
+	require.Equal(t, root.hash, ancestor)
+}
+
+func TestBlockTree_LowestCommonAncestor(t *testing.T) {
```
qdm12, line 449: supernit

```suggestion
func Test_BlockTree_LowestCommonAncestor(t *testing.T) {
```

lib/blocktree/blocktree_test.go
```diff
@@ -389,31 +464,31 @@ func TestBlockTree_HighestCommonAncestor(t *testing.T) {
...
 	b := leaves[1]
 
-	p, err := bt.HighestCommonAncestor(a, b)
+	p, err := bt.LowestCommonAncestor(a, b)
 	require.NoError(t, err)
 	require.Equal(t, expected, p)
 }
 
-func TestBlockTree_HighestCommonAncestor_SameNode(t *testing.T) {
+func TestBlockTree_LowestCommonAncestor_SameNode(t *testing.T) {
```
qdm12, line 472: supernit

```suggestion
func Test_BlockTree_LowestCommonAncestor_SameNode(t *testing.T) {
```

lib/blocktree/blocktree_test.go
```diff
@@ -389,31 +464,31 @@ func TestBlockTree_HighestCommonAncestor(t *testing.T) {
...
 	a := leaves[0]
 
-	p, err := bt.HighestCommonAncestor(a, a)
+	p, err := bt.LowestCommonAncestor(a, a)
 	require.NoError(t, err)
 	require.Equal(t, a, p)
 }
 
-func TestBlockTree_HighestCommonAncestor_SameChain(t *testing.T) {
+func TestBlockTree_LowestCommonAncestor_SameChain(t *testing.T) {
```
qdm12, line 483: supernit

```suggestion
func Test_BlockTree_LowestCommonAncestor_SameChain(t *testing.T) {
```

lib/blocktree/blocktree.go
qdm12, line 322: Should we panic in this case? :thinking:

lib/blocktree/blocktree.go
```diff
@@ -297,29 +298,51 @@ func (bt *BlockTree) Leaves() []Hash {
...
+			return common.Hash{}, errors.New("out of bounds ancestor check")
+		}
+		for diff > 0 {
+			aNode = aNode.parent
+			diff--
+		}
+	} else if bNum > aNum {
+		diff := bNum - aNum
+		if diff > bNum {
+			return common.Hash{}, errors.New("out of bounds ancestor check")
```
qdm12, line 331: Should we panic in this case? :thinking:

lib/blocktree/blocktree_test.go
```diff
@@ -371,7 +371,82 @@ func TestBlockTree_IsDecendantOf(t *testing.T) {
 	require.False(t, isDescendant)
 }
 
-func TestBlockTree_HighestCommonAncestor(t *testing.T) {
+func TestNodeLowestCommonAncestor(t *testing.T) {
```
qdm12, line 374: this doesn't cover the error cases, please add test cases for these..

also do you think it could be re-arranged as test-cases + subtests?

lib/grandpa/mocks/network.go
```diff
@@ -1,4 +1,4 @@
-// Code generated by mockery v2.14.1. DO NOT EDIT.
+// Code generated by mockery v2.14.0. DO NOT EDIT.
```
qdm12, line 1: upgrade your mockery :wink: 

```suggestion
// Code generated by mockery v2.14.1. DO NOT EDIT.
```

lib/grandpa/state.go
```diff
@@ -19,7 +19,7 @@ type BlockState interface {
 	GetHeader(hash common.Hash) (*types.Header, error)
 	GetHeaderByNumber(num uint) (*types.Header, error)
 	IsDescendantOf(parent, child common.Hash) (bool, error)
-	HighestCommonAncestor(a, b common.Hash) (common.Hash, error)
+	LowestCommonAncestor(a, b common.Hash) (common.Hash, error)
```
qdm12, line 22: nit

```suggestion
	LowestCommonAncestor(a, b common.Hash) (blockHash common.Hash, err error)
```

dot/core/interface.go
EclesioMeloJunior, line 74: what I think is not so clear is what means `lower`? eg:

Given the following chain:
```
                               / -> 4 -> 5 (0xA) -> 6
1 (0xU) -> 2 (0xY) -> 3 (0xZ) ->
                               \ -> 4 (0xB) -> 5 -> 6
```
the common ancestors between hashes `0xA` and `0xB` are hashes `0xZ`, `0xY` and `0xU` but the **lowest** common ancestor is `0xZ` (contains the higher block number and which is more close) or is `0xU` (contains the lowest block number and which is far)?

EclesioMeloJunior (changes requested):

dot/core/interface.go
qdm12: Good point (and thanks for the diagram)!

From how `highestCommonAncestor` was and how the new function is, it would be `0XZ`. I guess maybe `DeepestCommonAncestor` would remove any doubt (since depth is from the root, unlike highest/lowest which can be from both perspective).

timwu20: the function in substarte is `lowest_common_ancestor`.  I'm not sure why we named it `HighestCommonAncestor`.

timwu20: I think it's understood that it's the block hash.

lib/blocktree/blocktree_test.go
```diff
@@ -371,7 +371,82 @@ func TestBlockTree_IsDecendantOf(t *testing.T) {
...
+
+	ancestor, err = lowestCommonAncestor(root, childrenChildren[0])
+	require.NoError(t, err)
+	require.NotNil(t, ancestor)
+	require.Equal(t, root.hash, ancestor)
+
+	ancestor, err = lowestCommonAncestor(finalChild[0], childrenChildren[0])
+	require.NoError(t, err)
+	require.NotNil(t, ancestor)
+	require.Equal(t, root.hash, ancestor)
```
timwu20, line 446: I think these should be different tests.  I would use anonymous struct method to define tests.

lib/blocktree/blocktree.go
jimjbrettj: Yeah I think so too, good catch

lib/blocktree/blocktree.go
jimjbrettj: I dont really understand this, could you elaborate @qdm12 ?

jimjbrettj: oh I think I know what you mean actually

lib/blocktree/blocktree.go
jimjbrettj: I mean it seems to be yeah

lib/blocktree/blocktree.go
jimjbrettj: Im not sure this is a possibility. Can add it in anyways, but in what case would this be possible? It would have to be us processing a block that isn't a part of the chain I think ...

lib/blocktree/blocktree.go
jimjbrettj: I am not entirely sure, what do you think @timwu20 ?

lib/blocktree/blocktree_test.go
jimjbrettj: Actually with code changes I dont believe we can even hit the out of bounds error, so removing it

jimjbrettj: actually take it back, will add error test case

EclesioMeloJunior (approved):

dot/core/interface.go
qdm12: Meh for us sure, not for a newcomer

lib/blocktree/blocktree.go
qdm12: Cool, created #2939 for it (good first time issue :wink:)

lib/blocktree/blocktree.go
```diff
@@ -16,11 +17,14 @@ import (
...
-	Name:      "leaves_total",
-	Help:      "total number of blocktree leaves",
-})
+var (
+	leavesGauge = promauto.NewGauge(prometheus.GaugeOpts{
+		Namespace: "gossamer_block",
+		Name:      "leaves_total",
+		Help:      "total number of blocktree leaves",
+	})
+	errorAncestorOutOfBoundsCheck = errors.New("out of bounds ancestor check")
```
qdm12, line 26: supernit not sure why but convention seems to be `errXXX` not `errorXXX`... but really I feel like this convention I got used to for years is retarded... any thoughts anyone? Should we prefix our errors with.... `error` instead of `err`/`Err`? :thinking: Not a big deal either of course

lib/blocktree/blocktree.go
```diff
@@ -297,29 +301,53 @@ func (bt *BlockTree) Leaves() []Hash {
...
+func lowestCommonAncestor(aNode, bNode *node) (Hash, error) {
+	var higherNode *node
+	var lowerNode *node
+	if aNode.number > bNode.number {
+		higherNode = aNode
+		lowerNode = bNode
+	} else {
+		higherNode = bNode
+		lowerNode = aNode
+	}
```
qdm12, line 329: supernit shorter form:

```suggestion
	higherNode := bNode
	lowerNode := aNode
	if aNode.number > bNode.number {
		higherNode = aNode
		lowerNode = bNode
	}
```

lib/blocktree/blocktree.go
```diff
@@ -297,29 +301,53 @@ func (bt *BlockTree) Leaves() []Hash {
...
+
+	higherNum := higherNode.number
+	lowerNum := lowerNode.number
+	diff := higherNum - lowerNum
+	for diff > 0 {
+		if higherNode.parent == nil {
+			return common.Hash{}, fmt.Errorf("%w: for block number %v", errorAncestorOutOfBoundsCheck, higherNum)
```
qdm12, line 336: Should we panic here instead of returning an error? Same for below check?
That way we can simplify the signature to not return an error

Unless this can possibly happen

lib/blocktree/blocktree_test.go
```diff
@@ -371,7 +371,141 @@ func TestBlockTree_IsDecendantOf(t *testing.T) {
 	require.False(t, isDescendant)
 }
 
-func TestBlockTree_HighestCommonAncestor(t *testing.T) {
+func Test_Node_LowestCommonAncestorNew(t *testing.T) {
```
qdm12, line 374: nit we test `lowestCommonAncestor` so it should be:

```suggestion
func Test_lowestCommonAncestor(t *testing.T) {
```

lib/blocktree/blocktree.go
qdm12: Duplicate comment sorry

dot/core/interface.go
timwu20: Even for a newcomer, it's part of the `BlockState` interface.  I think it's pretty clear it's a block hash and not a trie node hash.

lib/blocktree/blocktree.go
timwu20, line 26:
```suggestion
	errAncestorOutOfBoundsCheck = errors.New("out of bounds ancestor check")
```

lib/blocktree/blocktree_test.go
```diff
@@ -371,7 +371,141 @@ func TestBlockTree_IsDecendantOf(t *testing.T) {
...
+		},
+	}
+	for _, tt := range tests {
+		tt := tt
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+			ancestor, err := lowestCommonAncestor(tt.args.nodeA, tt.args.nodeB)
+			assert.ErrorIs(t, err, tt.expErr)
+			if tt.expErr != nil {
+				assert.EqualError(t, err, tt.expErrMsg)
```
timwu20, line 501: This seems like overkill, since we already do the `assert.ErrorIs` check.  I don't think we need to string comparisons on error messages.

timwu20 (approved):

lib/blocktree/blocktree.go
jimjbrettj: the old code used a for loop and had the check just as the break conditions. I added this check, but do not think it's needed or will every be hit. This is also why there is no test case.

In typing this, deciding I will add a panic, that we can remove later if we want, just in case I am wrong and we somehow we hit this case. However, I do not think we can or will

lib/blocktree/blocktree.go
jimjbrettj: I just kept with error

lib/blocktree/blocktree.go
jimjbrettj: I have added panics instead of errors

PR #2943, opened 2022-11-14
fix(trie): use cached Merkle values for root hash

qdm12:
## Changes

- Use in-memory cached merkle value for encoding children, instead of re-encoding all children (mega expensive)
- Use in-memory cached root hash for returning the trie root hash
- Set the `EmptyHash` global variable using `hash([]byte{0})` directly (to avoid function cycle)

## Tests

```sh
```

## Primary Reviewer

@timwu20


lib/trie/trie_test.go
```diff
@@ -11,11 +11,22 @@ import (
...
 
+func Test_EmptyHash(t *testing.T) {
+	t.Parallel()
+
+	expected := common.Hash{
+		0x3, 0x17, 0xa, 0x2e, 0x75, 0x97, 0xb7, 0xb7,
+		0xe3, 0xd8, 0x4c, 0x5, 0x39, 0x1d, 0x13, 0x9a,
+		0x62, 0xb1, 0x57, 0xe7, 0x87, 0x86, 0xd8, 0xc0,
+		0x82, 0xf2, 0x9d, 0xcf, 0x4c, 0x11, 0x13, 0x14,
+	}
```
qdm12, line 26: note: that was taken from the `EmptyHash` variable value before these PR changes

lib/trie/trie_test.go
```diff
@@ -435,6 +353,12 @@ func Test_Trie_Hash(t *testing.T) {
 				root: &Node{
 					Key:      []byte{1, 2, 3},
 					SubValue: []byte{1},
+					MerkleValue: []byte{
+						0xa8, 0x13, 0x7c, 0xee, 0xb4, 0xad, 0xea, 0xac,
+						0x9e, 0x5b, 0x37, 0xe2, 0x8e, 0x7d, 0x64, 0x78,
+						0xac, 0xba, 0xb0, 0x6e, 0x90, 0x76, 0xe4, 0x67,
+						0xa1, 0xd8, 0xa2, 0x29, 0x4e, 0x4a, 0xd9, 0xa3,
+					},
```
qdm12, line 361: note: merkle values are only calculated and cached if they are absent. In this case, they were absent in the original trie, so they are cached in the final trie.

internal/trie/node/branch_encode.go
```diff
@@ -133,7 +133,7 @@ func encodeChild(child *Node, buffer io.Writer) (err error) {
 // and then SCALE encodes it. This is used to encode children
 // nodes of branches.
 func scaleEncodeHash(node *Node) (encoding []byte, err error) {
-	_, merkleValue, err := node.EncodeAndHash()
+	merkleValue, err := node.CalculateMerkleValue()
```
qdm12, line 136: `CalculateMerkleValue` uses the cached node Merkle value if it's already computed, and doesn't require to re-encode all the descendant nodes.

Since before #2919 we were caching encoding fields, we did not see the impact of this since hashing the cached encoding was cheap. This is now performance critical to use the cached merkle value to avoid re-encoding+hashing all the children.

lib/trie/trie.go
```diff
@@ -195,13 +183,17 @@ func (t *Trie) MustHash() common.Hash {
...
-	if err != nil {
-		return [32]byte{}, err
+	if t.root == nil {
+		return EmptyHash, nil
 	}
 
-	return common.Blake2bHash(buffer.Bytes()) // TODO optimisation: use hashers sync pools
+	merkleValue, err := t.root.CalculateRootMerkleValue()
+	if err != nil {
+		// no need to wrap the error really
```
EclesioMeloJunior, line 192:
```suggestion
```

EclesioMeloJunior (approved):
LGTM, just a small comment

lib/trie/trie.go
qdm12: Indeed, done :+1:

timwu20 (approved):

PR #2947, opened 2022-11-14
feat(lib/babe): Submit BABE equivocation report 

kishansagathiya:
## Tests

```sh
```

## Primary Reviewer

@timwu20


lib/babe/verify.go
```diff
@@ -331,13 +332,49 @@ func (b *verifier) verifyAuthorshipRight(header *types.Header) error {
...
 
 	return nil
 }
 
+func (b *verifier) submitAndReportEquivocation(
+	slot uint64, authorityIndex uint32, firstHeader, secondHeader types.Header) error {
+
+	// TODO: Check if it is initial sync
+	// don't report any equivocations during initial sync
+	// as they are most likely stale.
```
kishansagathiya, line 348: I will create a separate issue for this. substrate has something called `BlockOrigin`, which tells us how did we get this block. It could be things like we got it in the initial sync, was broadcasted in the network, genesis block, got it from a file, etc.

I think adding this initial sync check could be its own PR. It mostly isn't unsafe to not have this check for now.

dot/types/equivocation_test.go
```diff
@@ -0,0 +1,27 @@
...
+	"github.com/ChainSafe/gossamer/pkg/scale"
+	"github.com/stretchr/testify/require"
+)
+
+func TestEquivocationProof(t *testing.T) {
+	exp := []byte{
+		0, 91, 50, 25, 214, 94, 119, 36, 71, 216, 33, 152,
+		85, 184, 34, 120, 61, 161, 164, 223, 76, 53, 40,
+		246, 76, 38, 235, 204, 43, 31, 179, 28, 1, 0, 0,
+		0, 0, 0, 0, 0}
```
kishansagathiya, line 18: I have lost the rust code that I was using to produce these bytes. Will either wait to find that code or re-write that rust code. 
After that I will be able to do tests with some extra value in BabeEquivocationProof struct.
Currently this test fails

lib/runtime/wasmer/exports.go
```diff
@@ -86,6 +86,60 @@ func (in *Instance) GrandpaAuthorities() ([]types.Authority, error) {
...
+	if err != nil {
+		return nil, fmt.Errorf("encoding set id: %w", err)
+	}
+	combinedArg = append(combinedArg, encodedSetID...)
+
+	encodedAuthorityID, err := scale.Marshal(authorityID)
+	if err != nil {
+		return nil, fmt.Errorf("encoding authority id: %w", err)
+	}
+	combinedArg = append(combinedArg, encodedAuthorityID...)
```
qdm12, line 104: You might want to use the `scale.Encoder` to directly write to the final buffer, and also pre-allocate the max size for that buffer which can be known in advance:

```go
const maxBufferLength = 8 + 33 // scale encoded slot uint64 + scale encoded array of 32 bytes
buffer := bytes.NewBuffer(make([]byte, 0, maxBufferLength))
encoder := scale.NewEncoder(buffer)
err = encoder.Encode(slot)
// ...
err = encoder.Encode(authorityID)
// ...
ret, err := in.Exec(runtime.BabeAPIGenerateKeyOwnershipProof, buffer.Bytes())
```

lib/runtime/wasmer/exports.go
```diff
@@ -86,6 +86,60 @@ func (in *Instance) GrandpaAuthorities() ([]types.Authority, error) {
...
+
+	ret, err := in.Exec(runtime.BabeAPIGenerateKeyOwnershipProof, combinedArg)
+	if err != nil {
+		return nil, err
```
qdm12, line 108: context wrapping here could help us debug in the future

```suggestion
		return nil, fmt.Errorf("execing %s: %w", runtime.BabeAPIGenerateKeyOwnershipProof, err)
```

lib/runtime/wasmer/exports.go
```diff
@@ -86,6 +86,60 @@ func (in *Instance) GrandpaAuthorities() ([]types.Authority, error) {
...
+	}
+	combinedArg = append(combinedArg, encodedSetID...)
+
+	encodedAuthorityID, err := scale.Marshal(authorityID)
+	if err != nil {
+		return nil, fmt.Errorf("encoding authority id: %w", err)
+	}
+	combinedArg = append(combinedArg, encodedAuthorityID...)
+
+	ret, err := in.Exec(runtime.BabeAPIGenerateKeyOwnershipProof, combinedArg)
```
qdm12, line 106: nit `ret` -> `encodedKeyOwnershipProof`  would be better

lib/runtime/wasmer/exports.go
```diff
@@ -86,6 +86,60 @@ func (in *Instance) GrandpaAuthorities() ([]types.Authority, error) {
...
+	if err != nil {
+		return nil, err
+	}
+
+	keyOwnershipProof := types.OpaqueKeyOwnershipProof{}
+	err = scale.Unmarshal(ret, &keyOwnershipProof)
+	if err != nil {
+		return nil, fmt.Errorf("unmarshalling: %w", err)
```
qdm12, line 114: nit add what is being unmarshaled:

```suggestion
		return nil, fmt.Errorf("unmarshalling key ownership proof: %w", err)
```

also in my experience, most end users are not familiar with `unmarshalling` so I personally tend to use `scale decoding` / `scale encoding` in error wrappings/logs, but up to you, not really a needed change either!

lib/runtime/wasmer/exports.go
```diff
@@ -86,6 +86,60 @@ func (in *Instance) GrandpaAuthorities() ([]types.Authority, error) {
...
+	if err != nil {
+		return fmt.Errorf("encoding equivocation proof: %w", err)
+	}
+	combinedArg = append(combinedArg, encodedEquivocationProof...)
+
+	encodedKeyOwnershipProof, err := scale.Marshal(keyOwnershipProof)
+	if err != nil {
+		return fmt.Errorf("encoding key ownership proof: %w", err)
+	}
+	combinedArg = append(combinedArg, encodedKeyOwnershipProof...)
```
qdm12, line 137: same here you can use the more efficient pre-allocated buffer + scale encoder

lib/runtime/wasmer/exports.go
```diff
@@ -86,6 +86,60 @@ func (in *Instance) GrandpaAuthorities() ([]types.Authority, error) {
...
+
+	_, err = in.Exec(runtime.BabeAPISubmitReportEquivocationUnsignedExtrinsic, combinedArg)
+	return err
```
qdm12, line 140:
```suggestion
	if err != nil {
		return fmt.Errorf("execing %s: %w", runtime.BabeAPISubmitReportEquivocationUnsignedExtrinsic, err)
	}
	return nil
```

lib/runtime/wasmer/exports_test.go
```diff
@@ -408,6 +408,32 @@ func TestInstance_GrandpaAuthorities_PolkadotRuntime(t *testing.T) {
 	require.Equal(t, expected, auths)
 }
 
+func TestInstance_BabeGenerateKeyOwnershipProof_NodeRuntime(t *testing.T) {
+	tt := trie.NewEmptyTrie()
+	rt := NewTestInstanceWithTrie(t, runtime.NODE_RUNTIME, tt)
+	authorityID := types.AuthorityID{}
+	slot := uint64(1)
+	_, err := rt.BabeGenerateKeyOwnershipProof(slot, authorityID)
```
qdm12, line 416: The returned value is always an empty byte slice, can't we somehow use another runtime to test with actual values?

lib/runtime/wasmer/exports_test.go
```diff
@@ -408,6 +408,32 @@ func TestInstance_GrandpaAuthorities_PolkadotRuntime(t *testing.T) {
 	require.Equal(t, expected, auths)
 }
 
+func TestInstance_BabeGenerateKeyOwnershipProof_NodeRuntime(t *testing.T) {
+	tt := trie.NewEmptyTrie()
+	rt := NewTestInstanceWithTrie(t, runtime.NODE_RUNTIME, tt)
+	authorityID := types.AuthorityID{}
+	slot := uint64(1)
```
qdm12, line 415: nit use a constant

```suggestion
	const slot = 1
```

lib/runtime/wasmer/exports_test.go
```diff
@@ -408,6 +408,32 @@ func TestInstance_GrandpaAuthorities_PolkadotRuntime(t *testing.T) {
...
+	_, err := rt.BabeGenerateKeyOwnershipProof(slot, authorityID)
+	require.NoError(t, err)
+}
+
+func TestInstance_BabeSubmitReportEquivocationUnsignedExtrinsic_NodeRuntime_(t *testing.T) {
```
qdm12, line 420:
```suggestion
func TestInstance_BabeSubmitReportEquivocationUnsignedExtrinsic_NodeRuntime(t *testing.T) {
```

lib/runtime/wasmer/exports_test.go
```diff
@@ -408,6 +408,32 @@ func TestInstance_GrandpaAuthorities_PolkadotRuntime(t *testing.T) {
...
+	tt := trie.NewEmptyTrie()
+	rt := NewTestInstanceWithTrie(t, runtime.NODE_RUNTIME, tt)
+	authorityID := types.AuthorityID{}
```
qdm12, line 423: you should test using non-empty values to ensure they are passed around correctly in the call stack

```suggestion
	authorityID := types.AuthorityID{1}
```

lib/runtime/wasmer/exports_test.go
```diff
@@ -408,6 +408,32 @@ func TestInstance_GrandpaAuthorities_PolkadotRuntime(t *testing.T) {
...
+	slot := uint64(1)
```
qdm12, line 424: nit you can use a constant

```suggestion
	const slot = 1
```

lib/runtime/wasmer/exports_test.go
```diff
@@ -408,6 +408,32 @@ func TestInstance_GrandpaAuthorities_PolkadotRuntime(t *testing.T) {
...
+	keyOwnershipProof, err := rt.BabeGenerateKeyOwnershipProof(slot, authorityID)
```
qdm12, line 425: The returned value is always an empty byte slice, can't we somehow use another runtime to test with actual values?

lib/runtime/wasmer/exports.go
```diff
@@ -86,6 +86,60 @@ func (in *Instance) GrandpaAuthorities() ([]types.Authority, error) {
...
 }
 
+// BabeGenerateKeyOwnershipProof returns babe key ownership proof from runtime.
+func (in *Instance) BabeGenerateKeyOwnershipProof(slot uint64, authorityID [32]byte) (
+	types.OpaqueKeyOwnershipProof, error) {
+
+	combinedArg := []byte{}
+	encodedSetID, err := scale.Marshal(slot)
+	if err != nil {
+		return nil, fmt.Errorf("encoding set id: %w", err)
```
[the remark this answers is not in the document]
jimjbrettj: Is set id not more precise here?

lib/runtime/wasmer/exports.go
jimjbrettj: Maybe "executing" instead of "execing"? "execing sounds weird/isn't a word I don't think lol

lib/runtime/wasmer/exports.go
jimjbrettj: Instead of wrapping the errors here, why not have the caller just wrap it? Feels a little redundant to me, but also a genuine question. My intuition is that the returned error will be enough for the caller (or us reading logs) to know what is wrong. For example, a formatting error is an error from the runtime which is logged, so we don't need that info in the log because its already obvious. 

But like I said, genuine question! Just wanna make sure we are adding things for good reason

dot/core/interface.go
```diff
@@ -42,6 +42,12 @@ type RuntimeInstance interface {
 	DecodeSessionKeys(enc []byte) ([]byte, error)
 	PaymentQueryInfo(ext []byte) (*types.RuntimeDispatchInfo, error)
 	CheckInherents()
+	BabeGenerateKeyOwnershipProof(slot uint64, authorityID [32]byte) (
```
jimjbrettj, line 45: Really really like shifting from passing in bytes to actual params and handling from their. This is how substrate does it and is much more intuitive in my opinion

dot/types/equivocation.go
```diff
@@ -0,0 +1,30 @@
...
+}
+
+// AuthorityID represents a babe authority identifier.
+type AuthorityID [32]byte
+
+// OpaqueKeyOwnershipProof is an opaque type used to represent the key ownership proof at the
+// runtime API boundary. The inner value is an encoded representation of the actual key
+// ownership proof which will be parameterized when defining the runtime. At
+// the runtime API boundary this type is unknown and as such we keep this
+// opaque representation, implementors of the runtime API will have to make
```
jimjbrettj, line 28: Is this part of the comment not rust specific? Doesnt golang do type checking on the underlying type? Example, you could pass this as a para to a function that takes []byte as an argument and this would compile right?

Other thing is, do we need this type? Since its just a byte wrapper and doesn't provide (I think) the type checking benefits that rust does, is it necessary? If its just for readability and to mimick substrate I am okay with that, but if its not serving a good purpose I am okay with replacing the type with relevant variable names

dot/types/equivocation_test.go
jimjbrettj: I like it, but do we really need this? Is this not just testing the ability to encode/decode this data? Since its not an assertion on the actual results of the runtime call, do we need it?

lib/runtime/wasmer/exports.go
jimjbrettj: Curious to see what method gets chosen here, will follow suit with what @kishansagathiya does in my grandpa version of this PR

lib/runtime/wasmer/exports.go
jimjbrettj: If not, maybe have a helper appendBytes([][]byte) func to combine bytes?

jimjbrettj (commented):
Looks good, jus a few comments. Great stuff!

lib/runtime/wasmer/exports.go
```diff
@@ -86,6 +86,60 @@ func (in *Instance) GrandpaAuthorities() ([]types.Authority, error) {
 	return types.GrandpaAuthoritiesRawToAuthorities(gar)
 }
 
+// BabeGenerateKeyOwnershipProof returns babe key ownership proof from runtime.
+func (in *Instance) BabeGenerateKeyOwnershipProof(slot uint64, authorityID [32]byte) (
```
jimjbrettj, line 90: Maybe make authorityID be of type ed25519.PublicKeyBytes for clarity?

dot/core/interface.go
timwu20:
```
p_api::decl_runtime_apis! {
	/// API necessary for block authorship with BABE.
	#[api_version(2)]
	pub trait BabeApi {
		...
		/// Generates a proof of key ownership for the given authority in the
		/// current epoch. An example usage of this module is coupled with the
		/// session historical module to prove that a given authority key is
		/// tied to a given staking identity during a specific session. Proofs
		/// of key ownership are necessary for submitting equivocation reports.
		/// NOTE: even though the API takes a `slot` as parameter the current
		/// implementations ignores this parameter and instead relies on this
		/// method being called at the correct block height, i.e. any point at
		/// which the epoch for the given slot is live on-chain. Future
		/// implementations will instead use indexed data through an offchain
		/// worker, not requiring older states to be available.
		fn generate_key_ownership_proof(
			slot: Slot,
			authority_id: AuthorityId,
		) -> Option<OpaqueKeyOwnershipProof>;
		...
	}
}
```
Looks right to me? `AuthorityID` resolves to a sr25519 public key, which is a `[u8; 32]`.

dot/types/equivocation.go
timwu20: This comment is copied from the substrate code.  I don't mind defining types that don't have receiver methods.  It's the same on the substrate side.  Maybe we need a constructor for this?

lib/runtime/wasmer/exports.go
timwu20: You can also put it into a helper struct and just encode that instead of appending the encodings.

lib/runtime/wasmer/exports.go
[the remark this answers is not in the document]
timwu20: It's slot in the babe context, and set id in the grandpa context.

dot/types/equivocation_test.go
kishansagathiya: replace these bytes.

lib/babe/verify_test.go
```diff
@@ -814,6 +815,75 @@ func Test_verifier_verifyBlockEquivocation(t *testing.T) {
...
+		authorities: []types.Authority{*auth, *auth},
+		threshold:   scale.MaxUint128,
+	}
+
+	ctrl := gomock.NewController(t)
+	mockBlockState := NewMockBlockState(ctrl)
+	verifier := newVerifier(mockBlockState, 1, vi)
+
+	slot := uint64(1)
+	authorityIndex := uint32(1)
```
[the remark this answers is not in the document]
kishansagathiya: I can use const, but I need precise types. Otherwise, it would change encodings.

lib/babe/verify_test.go
```diff
@@ -814,6 +815,75 @@ func Test_verifier_verifyBlockEquivocation(t *testing.T) {
...
+		VRFProof:       proof,
+	}
+	prd, err := testDigest.ToPreRuntimeDigest()
+	assert.NoError(t, err)
+
+	firstHeader := newTestHeader(t, *prd)
+	firstHash := encodeAndHashHeader(t, firstHeader)
+	signAndAddSeal(t, kp, firstHeader, firstHash[:])
+
+	output2, proof2, err := kp.VrfSign(makeTranscript(Randomness{}, slot, 2))
```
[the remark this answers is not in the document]
kishansagathiya: >  I tried with randomly set values and the test passes which itches me

It shouldn't itch.

The goal of this test is to submit and report equivocation. For that it creates blocks that are equivocated. Now, what is used here (secondHeader) is just one example of equivocated block. You could change randomness (and/or other values as well) and that would create an another equivocated block. We are not doing any equality check of blocks here, in fact we need a different block which is equivocated.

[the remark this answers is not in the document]
kishansagathiya: > do you know what that randomness field does?

Randomness is used to mine blocks.

If randomness ( that you generated as a validator) is less than a threshold you get to produce that block.
https://wiki.polkadot.network/docs/learn-randomness#heres-how-it-works-in-detail

lib/runtime/wasmer/exports.go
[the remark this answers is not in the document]
kishansagathiya: changed. Looks like this was left over code from previous iteration. First time I implemented this I ended up implementing this function for grandpa, without realizing that babe and grandpa both have this runtime function.

lib/runtime/wasmer/exports_test.go
kishansagathiya: > The returned value is always an empty byte slice, can't we somehow use another runtime to test with actual values?

So, authorityID has to be part of authorities in order to see a non empty proof. Our current runtime instance is created without authorities. I have tried changing some basic things like `role`, but that hasn't work. Will look more into it.

Once I can create a runtime instance with babe authorities, I can use ` rt.BabeConfiguration()` to get authorities and can use one of them to create a proof. Currently ` rt.BabeConfiguration()` gives you no authorities.

kishansagathiya: I have added another copy of these tests though, with Polkadot runtime.

kishansagathiya: Hmm, Ok.
This is definitely a problem. I ran a network and modified the code to see values of key Ownership and I am always seeing empty bytes. 
But I haven't figured out yet why this is happening.

lib/babe/verify.go
kishansagathiya: created issue here https://github.com/ChainSafe/gossamer/issues/3004

lib/runtime/wasmer/exports.go
```diff
@@ -86,6 +86,61 @@ func (in *Instance) GrandpaAuthorities() ([]types.Authority, error) {
...
+
+	combinedArg := []byte{}
+
+	encodedEquivocationProof, err := scale.Marshal(equivocationProof)
+	if err != nil {
+		return fmt.Errorf("encoding equivocation proof: %w", err)
+	}
+	combinedArg = append(combinedArg, encodedEquivocationProof...)
+
+	encodedKeyOwnershipProof, err := scale.Marshal(keyOwnershipProof)
```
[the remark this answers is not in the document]
kishansagathiya: As discussed separately, it is expected to scale marshal keyOwnershipProof here

lib/runtime/wasmer/exports_test.go
```diff
@@ -408,6 +408,32 @@ func TestInstance_GrandpaAuthorities_PolkadotRuntime(t *testing.T) {
 	require.Equal(t, expected, auths)
 }
 
+func TestInstance_BabeGenerateKeyOwnershipProof_NodeRuntime(t *testing.T) {
```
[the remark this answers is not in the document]
kishansagathiya: makes sense. 

Will remove use of BabeGenerateKeyOwnershipProof inside TestInstance_BabeSubmitReportEquivocationUnsignedExtrinsic_NodeRuntime and will use a hardcoded key ownership proof.

lib/runtime/wasmer/exports_test.go
kishansagathiya: resolving for https://github.com/ChainSafe/gossamer/issues/3003

lib/babe/mocks_generate_test.go
```diff
@@ -5,3 +5,4 @@ package babe
 
 //go:generate mockgen -destination=mocks/network.go -package=mocks github.com/ChainSafe/gossamer/dot/core Network
 //go:generate mockgen -destination=mock_telemetry_test.go -package $GOPACKAGE github.com/ChainSafe/gossamer/dot/telemetry Client
+//go:generate mockgen -destination=mocks/runtime.go -package=mocks github.com/ChainSafe/gossamer/dot/core RuntimeInstance
```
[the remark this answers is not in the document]
kishansagathiya: there is a dedicated mocks folder already, before this PR. Also, I prefer it that way rather than having all mocks in the same file.

dot/types/equivocation.go
kishansagathiya: @jimjbrettj This is mostly for readability and to add some extra info about what key ownership proof is. I could just replace this with []byte and things would work just fine.

> Maybe we need a constructor for this?

Seems unnecessary. I could wrap bytes with `OpaqueKeyOwnershipProof([]byte{})` or use `NewOpaqueKeyOwnershipProof([]byte{})`. Both seems like pretty much the same thing.

lib/runtime/wasmer/exports.go
kishansagathiya: > Instead of wrapping the errors here, why not have the caller just wrap it?

Yep, that goes with the idea of adding context to errors while returning errors. So, function names should already be there in the returned error.

There isn't any need for changes here, because of that.

lib/runtime/wasmer/exports.go
kishansagathiya: > You can also put it into a helper struct and just encode that instead of appending the encodings.

Have gone with this one, seem more appropriate

lib/runtime/wasmer/exports.go
kishansagathiya: addressed

lib/babe/verify.go
```diff
@@ -331,13 +330,49 @@ func (b *verifier) verifyAuthorshipRight(header *types.Header) error {
...
 		return fmt.Errorf("%w for block header %s", ErrProducerEquivocated, header.Hash())
 	}
 
 	return nil
 }
 
+func (b *verifier) submitAndReportEquivocation(
+	slot uint64, authorityIndex uint32, firstHeader, secondHeader types.Header) error {
+
+	// TODO: Check if it is initial sync
```
jimjbrettj, line 344: I saw you created issue for this, maybe include the number in this comment

lib/babe/verify_test.go
```diff
@@ -814,6 +834,75 @@ func Test_verifier_verifyBlockEquivocation(t *testing.T) {
...
+	require.NoError(t, err)
+
+	firstHeader := newTestHeader(t, *preRuntimeDigest)
+	firstHash := encodeAndHashHeader(t, firstHeader)
+	signAndAddSeal(t, keyPair, firstHeader, firstHash[:])
+
+	output2, proof2, err := keyPair.VrfSign(makeTranscript(Randomness{}, slot, epochNumber))
+	require.NoError(t, err)
+
+	testDigest2 := types.BabePrimaryPreDigest{
```
jimjbrettj, line 875: nit: can just be called digest2

lib/runtime/wasmer/exports_test.go
```diff
@@ -408,6 +408,96 @@ func TestInstance_GrandpaAuthorities_PolkadotRuntime(t *testing.T) {
...
+	tt := trie.NewEmptyTrie()
+
+	rvalue, err := common.HexToHash("0x01")
```
jimjbrettj, line 414: Is there a better name than avalue to use?

lib/runtime/wasmer/exports_test.go
```diff
@@ -408,6 +408,96 @@ func TestInstance_GrandpaAuthorities_PolkadotRuntime(t *testing.T) {
...
+	require.NoError(t, err)
+	key := common.MustHexToBytes(genesis.BABERandomnessKeyHex)
+	tt.Put(key, rvalue[:])
+
+	avalue, err := common.HexToBytes("0x08eea1eabcac7d2c8a6459b7322cf997874482bfc3d2ec7a80888a3a7d714103640100000000000000b64994460e59b30364cad3c92e3df6052f9b0ebbb8f88460c194dc5794d6d7170100000000000000") //nolint:lll
```
jimjbrettj, line 419: Is there a better name than avalue to use?

lib/runtime/wasmer/exports_test.go
```diff
@@ -408,6 +408,96 @@ func TestInstance_GrandpaAuthorities_PolkadotRuntime(t *testing.T) {
...
+
+	const slot = uint64(10)
+	_, err = rt.BabeGenerateKeyOwnershipProof(slot, authorityID)
+	require.NoError(t, err)
+}
+
+func TestInstance_BabeGenerateKeyOwnershipProof_PolkadotRuntime(t *testing.T) {
+	tt := trie.NewEmptyTrie()
+
+	rvalue, err := common.HexToHash("0x01")
```
jimjbrettj, line 442: Is there a better name than rvalue to use?

lib/runtime/wasmer/exports_test.go
```diff
@@ -408,6 +408,96 @@ func TestInstance_GrandpaAuthorities_PolkadotRuntime(t *testing.T) {
...
+	require.NoError(t, err)
+	key := common.MustHexToBytes(genesis.BABERandomnessKeyHex)
+	tt.Put(key, rvalue[:])
+
+	avalue, err := common.HexToBytes("0x08eea1eabcac7d2c8a6459b7322cf997874482bfc3d2ec7a80888a3a7d714103640100000000000000b64994460e59b30364cad3c92e3df6052f9b0ebbb8f88460c194dc5794d6d7170100000000000000") //nolint:lll
```
jimjbrettj, line 447: Is there a better name than avalue to use?

jimjbrettj (approved):
Few comments, but overall looks great!

lib/runtime/wasmer/exports_test.go
```diff
@@ -408,6 +408,96 @@ func TestInstance_GrandpaAuthorities_PolkadotRuntime(t *testing.T) {
 	require.Equal(t, expected, auths)
 }
 
+func TestInstance_BabeGenerateKeyOwnershipProof_NodeRuntime(t *testing.T) {
```
EclesioMeloJunior, line 411: Is it possible to test against Westend runtime instead?

lib/runtime/wasmer/exports_test.go
```diff
@@ -408,6 +408,96 @@ func TestInstance_GrandpaAuthorities_PolkadotRuntime(t *testing.T) {
...
+	key := common.MustHexToBytes(genesis.BABERandomnessKeyHex)
+	tt.Put(key, rvalue[:])
+
+	avalue, err := common.HexToBytes("0x08eea1eabcac7d2c8a6459b7322cf997874482bfc3d2ec7a80888a3a7d714103640100000000000000b64994460e59b30364cad3c92e3df6052f9b0ebbb8f88460c194dc5794d6d7170100000000000000") //nolint:lll
+	require.NoError(t, err)
+
+	key = common.MustHexToBytes(genesis.BABEAuthoritiesKeyHex)
+	tt.Put(key, avalue)
+
+	rt := NewTestInstanceWithTrie(t, runtime.POLKADOT_RUNTIME, tt)
```
EclesioMeloJunior, line 453: The `POLKADOT_RUNTIME` targets to polkadot 0.8.25 (old one) could you change to `POLKADOT_RUNTIME_v0929`?

EclesioMeloJunior (changes requested):

lib/runtime/wasmer/exports.go
```diff
@@ -86,6 +87,55 @@ func (in *Instance) GrandpaAuthorities() ([]types.Authority, error) {
...
+	err := encoder.Encode(equivocationProof)
+	if err != nil {
+		return fmt.Errorf("encoding equivocation proof: %w", err)
+	}
+	err = encoder.Encode(keyOwnershipProof)
+	if err != nil {
+		return fmt.Errorf("encoding key ownership proof: %w", err)
+	}
+	_, err = in.Exec(runtime.BabeAPISubmitReportEquivocationUnsignedExtrinsic, buffer.Bytes())
+	return err
```
qdm12, line 136: it would be nice to cover the error != nil case in a unit test

lib/runtime/wasmer/exports_test.go
```diff
@@ -408,6 +408,96 @@ func TestInstance_GrandpaAuthorities_PolkadotRuntime(t *testing.T) {
...
+	equivocationProof := types.BabeEquivocationProof{
+		Offender: authorityID,
+		Slot:     slot,
+	}
+
+	err := rt.BabeSubmitReportEquivocationUnsignedExtrinsic(equivocationProof, keyOwnershipProof)
+	require.NoError(t, err)
+}
+
+func TestInstance_BabeSubmitReportEquivocationUnsignedExtrinsic_PolkadotRuntime(t *testing.T) {
```
qdm12, line 484: Please merge the two tests `TestInstance_BabeSubmitReportEquivocationUnsignedExtrinsic_XXX` in a single test with 2 test cases

lib/runtime/wasmer/exports.go
```diff
@@ -86,6 +87,55 @@ func (in *Instance) GrandpaAuthorities() ([]types.Authority, error) {
...
+		return nil, fmt.Errorf("encoding slot: %w", err)
+	}
+	err = encoder.Encode(authorityID)
+	if err != nil {
+		return nil, fmt.Errorf("encoding authority id: %w", err)
+	}
+
+	encodedKeyOwnershipProof, err := in.Exec(runtime.BabeAPIGenerateKeyOwnershipProof, buffer.Bytes())
+	if err != nil {
+		return nil, fmt.Errorf("executing %s: %w", runtime.BabeAPIGenerateKeyOwnershipProof, err)
```
qdm12, line 109: nit potentially bad suggestion from me, let's revert it (sorry!)

```suggestion
		return nil, err
```

lib/runtime/wasmer/exports_test.go
```diff
@@ -408,6 +408,96 @@ func TestInstance_GrandpaAuthorities_PolkadotRuntime(t *testing.T) {
...
+	require.NotEmpty(t, babeConfig.GenesisAuthorities)
+
+	authorityID := babeConfig.GenesisAuthorities[0].Key
+
+	const slot = uint64(10)
+	_, err = rt.BabeGenerateKeyOwnershipProof(slot, authorityID)
+	require.NoError(t, err)
+}
+
+func TestInstance_BabeGenerateKeyOwnershipProof_PolkadotRuntime(t *testing.T) {
```
qdm12, line 439: Merge this test with the other test for the node runtime in a single test with test cases since they are very similar.

lib/runtime/wasmer/exports.go
qdm12, line 109: It would be nice to have this error case covered in a unit test (that is, if invalid slot+authority id is possible)

lib/babe/verify.go
timwu20: can you link the created issue?

timwu20:
I changed the PR name on this to conform to [semantic commit messages](https://gist.github.com/joshbuchea/6f47e86d2510bce28f8e7f42ae84c716).  It's used by our release workflow to generate the release notes.

lib/babe/verify.go
kishansagathiya: https://github.com/ChainSafe/gossamer/issues/3004

dot/interfaces.go
```diff
@@ -64,6 +64,8 @@ type runtimeInterface interface {
 	DecodeSessionKeys(enc []byte) ([]byte, error)
 	PaymentQueryInfo(ext []byte) (*types.RuntimeDispatchInfo, error)
 	CheckInherents()
+	BabeGenerateKeyOwnershipProof(slot uint64, offenderPublicKey [32]byte) (types.OpaqueKeyOwnershipProof, error)
+	BabeSubmitReportEquivocationUnsignedExtrinsic(types.BabeEquivocationProof, types.OpaqueKeyOwnershipProof) error
```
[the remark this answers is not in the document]
kishansagathiya: rofl

lib/runtime/wasmer/exports.go
```diff
@@ -86,6 +87,55 @@ func (in *Instance) GrandpaAuthorities() ([]types.Authority, error) {
 	return types.GrandpaAuthoritiesRawToAuthorities(gar)
 }
 
+// BabeGenerateKeyOwnershipProof returns the babe key ownership proof from the runtime.
+func (in *Instance) BabeGenerateKeyOwnershipProof(slot uint64, authorityID [32]byte) (
+	types.OpaqueKeyOwnershipProof, error) {
+
+	// scale encoded slot uint64 + scale encoded array of 32 bytes
+	const maxBufferLength = 8 + 33
```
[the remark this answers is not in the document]
kishansagathiya: I think there is one extra byte for length or something

lib/runtime/wasmer/exports_test.go
```diff
@@ -408,6 +408,96 @@ func TestInstance_GrandpaAuthorities_PolkadotRuntime(t *testing.T) {
...
+
+	babeConfig, err := rt.BabeConfiguration()
+	require.NoError(t, err)
+
+	require.NotEmpty(t, babeConfig.GenesisAuthorities)
+
+	authorityID := babeConfig.GenesisAuthorities[0].Key
+
+	const slot = uint64(10)
+	_, err = rt.BabeGenerateKeyOwnershipProof(slot, authorityID)
```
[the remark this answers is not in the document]
kishansagathiya: we had a discussion about the empty struct issue in one of the standups and had decided to delay it and do it seperately
https://github.com/ChainSafe/gossamer/issues/3003

lib/runtime/wasmer/exports_test.go
```diff
@@ -408,6 +408,96 @@ func TestInstance_GrandpaAuthorities_PolkadotRuntime(t *testing.T) {
...
```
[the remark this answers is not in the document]
kishansagathiya: we had a discussion about the empty struct issue in one of the standups and had decided to delay it and do it seperately
https://github.com/ChainSafe/gossamer/issues/3003

lib/runtime/wasmer/exports.go
kishansagathiya: doesn't seem like it, I have tried this with empty authority id and the call doesn't throw error

kishansagathiya: I would just let it be, it's not bad

lib/runtime/wasmer/exports.go
kishansagathiya: way to do that would be to feed some invalid values to `BabeSubmitReportEquivocationUnsignedExtrinsic`. I have tried feeding empty values to `BabeSubmitReportEquivocationUnsignedExtrinsic`, but `runtime.BabeAPISubmitReportEquivocationUnsignedExtrinsic` does not throw error.

kishansagathiya: I can feed bad data directly to `runtime.BabeAPISubmitReportEquivocationUnsignedExtrinsic` which would throw error, but anything I send to `BabeSubmitReportEquivocationUnsignedExtrinsic` gets scale encoded and does not throw error.

but then, that would be more like testing `runtime.BabeAPISubmitReportEquivocationUnsignedExtrinsic` runtime func.
So, for now, I have n't seen any data on which  `BabeSubmitReportEquivocationUnsignedExtrinsic` could throw error.

lib/runtime/test_helpers.go
```diff
@@ -282,6 +282,8 @@ type Instance interface {
 	ExecuteBlock(block *types.Block) ([]byte, error)
 	DecodeSessionKeys(enc []byte) ([]byte, error)
 	PaymentQueryInfo(ext []byte) (*types.RuntimeDispatchInfo, error)
+	BabeGenerateKeyOwnershipProof(slot uint64, offenderPublicKey [32]byte) (types.OpaqueKeyOwnershipProof, error)
+	BabeSubmitReportEquivocationUnsignedExtrinsic(types.BabeEquivocationProof, types.OpaqueKeyOwnershipProof) error
```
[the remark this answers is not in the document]
kishansagathiya: It is needed in the test TestPaymentModule_QueryInfo

lib/runtime/wasmer/exports_test.go
```diff
@@ -408,6 +408,96 @@ func TestInstance_GrandpaAuthorities_PolkadotRuntime(t *testing.T) {
...
+	require.NoError(t, err)
+}
+
+func TestInstance_BabeSubmitReportEquivocationUnsignedExtrinsic_NodeRuntime(t *testing.T) {
+	tt := trie.NewEmptyTrie()
+	rt := NewTestInstanceWithTrie(t, runtime.NODE_RUNTIME, tt)
```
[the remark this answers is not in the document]
kishansagathiya: there is a runtime package already, also rt is not bad

lib/babe/verify.go
```diff
@@ -375,6 +410,11 @@ func (b *verifier) verifyBlockEquivocation(header *types.Header) (bool, error) {
 			continue
 		}
 
+		err = b.submitAndReportEquivocation(slot, authorOfExistingHeader, *existingHeader, *header)
+		if err != nil {
+			return true, fmt.Errorf("submitting and reporting equivocation: %w", err)
```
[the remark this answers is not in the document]
kishansagathiya: added

lib/babe/verify.go
```diff
@@ -331,13 +330,49 @@ func (b *verifier) verifyAuthorshipRight(header *types.Header) error {
...
 	}
+
 	if equivocated {
 		return fmt.Errorf("%w for block header %s", ErrProducerEquivocated, header.Hash())
 	}
 
 	return nil
 }
 
+func (b *verifier) submitAndReportEquivocation(
```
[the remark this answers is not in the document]
kishansagathiya: I have added one extra case (error case). Some of cases, I haven't seen erroring, like `runtimeInstance.BabeSubmitReportEquivocationUnsignedExtrinsic` and `runtimeInstance.BabeGenerateKeyOwnershipProof`

lib/babe/verify.go
```diff
@@ -330,13 +330,50 @@ func (b *verifier) verifyAuthorshipRight(header *types.Header) error {
...
+	slot uint64, authorityIndex uint32, firstHeader, secondHeader types.Header) error {
+
+	// TODO: Check if it is initial sync
+	// don't report any equivocations during initial sync
+	// as they are most likely stale.
+	// Issue #3004
```
timwu20, line 347:
```suggestion
	// https://github.com/ChainSafe/gossamer/issues/3004
```

lib/babe/verify.go
```diff
@@ -330,13 +330,50 @@ func (b *verifier) verifyAuthorshipRight(header *types.Header) error {
...
+
+	bestBlockHash := b.blockState.BestBlockHash()
+	runtimeInstance, err := b.blockState.GetRuntime(bestBlockHash)
+	if err != nil {
+		return fmt.Errorf("getting runtime: %w", err)
+	}
+
+	offenderPublicKey := b.authorities[authorityIndex].ToRaw().Key
```
timwu20, line 355: what if `authorityIndex` is out of bounds? Maybe we should return an error in this case? Better than a panic.

timwu20 (changes requested):

lib/babe/mocks/runtime.go
```diff
@@ -0,0 +1,417 @@
+// Code generated by MockGen. DO NOT EDIT.
+// Source: github.com/ChainSafe/gossamer/dot/core (interfaces: RuntimeInstance)
+
+// Package mocks is a generated GoMock package.
+package mocks
```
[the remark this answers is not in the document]
kishansagathiya: added to mocks_generate_test.go, I don't really mind the mocks package. Looks cleaner to me that creating one file with all mocks.

timwu20 (approved):

PR #2952, opened 2022-11-15
fix(peerset): check for incoming slot error

### LINKED ISSUE #2914: [peerset]: cannot accept incomming peer: not enough incoming slots
## Describe the bug

- Investigation purpose issue
- ERROR cannot accept incomming peer 12D3KooWL3WgKWRhCMaxSCbyGVzYNa7DxF1SuBk9cRvffCBidR5y: not enough incoming slots peerset.go:L644 pkg=peerset

edwardmack:
## Changes

- Added check for error `ErrIncomingSlotsUnavailable` and logging condition to Info logs, since this happens when the peer set has filled it's incoming slot allowance.
- Fix spelling of `incomming` to `incoming` in log message.

## Tests

```sh
go test github.com/ChainSafe/gossamer/dot/peerset
```

## Issues
closes #2914 

## Primary Reviewer

@timwu20

### DIFF
--- a/dot/peerset/peerset.go
+++ b/dot/peerset/peerset.go
@@ -641,8 +641,13 @@ func (ps *PeerSet) incoming(setID int, peers ...peer.ID) error {
 		} else {
 			err := state.tryAcceptIncoming(setID, pid)
 			if err != nil {
-				logger.Errorf("cannot accept incomming peer %s: %s", pid, err)
-				message.Status = Reject
+				if err == ErrIncomingSlotsUnavailable {
+					logger.Infof("cannot accept incoming peer %s: %s", pid, err)
+					message.Status = Reject
+				} else {
+					logger.Errorf("cannot accept incoming peer %s: %s", pid, err)
+					message.Status = Reject
+				}
 			} else {
 				logger.Debugf("incoming connection accepted from peer %s", pid)
 				message.Status = Accept
--- a/dot/peerset/peerset.go
+++ b/dot/peerset/peerset.go
@@ -5,6 +5,7 @@ package peerset
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"math"
 	"strings"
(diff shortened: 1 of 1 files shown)


dot/peerset/peerset.go
```diff
@@ -641,8 +641,13 @@ func (ps *PeerSet) incoming(setID int, peers ...peer.ID) error {
 		} else {
 			err := state.tryAcceptIncoming(setID, pid)
 			if err != nil {
-				logger.Errorf("cannot accept incomming peer %s: %s", pid, err)
-				message.Status = Reject
+				if err == ErrIncomingSlotsUnavailable {
```
qdm12, line 644:
```suggestion
				if errors.Is(err, ErrIncomingSlotsUnavailable) {
```

dot/peerset/peerset.go
```diff
@@ -641,8 +641,13 @@ func (ps *PeerSet) incoming(setID int, peers ...peer.ID) error {
...
+					logger.Infof("cannot accept incoming peer %s: %s", pid, err)
+					message.Status = Reject
+				} else {
+					logger.Errorf("cannot accept incoming peer %s: %s", pid, err)
+					message.Status = Reject
```
qdm12, line 649: take the `message.Status = Reject` outside the `if`/`else` (before or after it)

jimjbrettj:
Even though it is no longer an error-level log, these logs still happen often and pollute the logs. Maybe this is okay, but maybe its worth discussing if there is another way to handle this that doesn't spam as many logs?

qdm12:
> Even though it is no longer an error-level log, these logs still happen often and pollute the logs. Maybe this is okay, but maybe its worth discussing if there is another way to handle this that doesn't spam as many logs?

Yeah good point, should these be at the debug level? :thinking:

edwardmack:
> > Even though it is no longer an error-level log, these logs still happen often and pollute the logs. Maybe this is okay, but maybe its worth discussing if there is another way to handle this that doesn't spam as many logs?
> 
> Yeah good point, should these be at the debug level? thinking

Yes, I agree, I've updated to report this message as log level Debug.

qdm12 (approved):

timwu20 (approved):

PR #2953, opened 2022-11-16
chore(wasmer): `nil` helper functions

qdm12:
## Changes

Just something I thought would be nice whilst reviewing #2873 

It also panics on error when we would previously discard the error (and really we should not).


jimjbrettj (approved):
feels like i nice improvement to me

timwu20 (approved):

Issue #2937, opened 2022-11-11
CRITICAL: Bitfields and heads must be included every block


EclesioMeloJunior:
## Describe the bug

- Running gossamer using the chain configuration `westend-local` I face the following **critical** log message
```
CRITICAL target=runtime message=panicked at 'Bitfields and heads must be included every block', /builds/runtime/parachains/src/paras_inherent/mod.rs:200:17	imports.go:L140:ext_logging_log_version_1

WARN     failed to handle slot 278029812: cannot finalise block: running runtime function: Failed to call the `BlockBuilder_finalize_block` exported function.
```
- This error happens in the right after I start the nodes, here is the full log: https://gist.github.com/EclesioMeloJunior/4eff5eb0a2c88ad6313fdd19c100b370

## Additional Informations

- I am using the latest changes in development branch
- the config file I am using is:
```toml
[global]
basepath = "./tmp/westend"
log = "info"
metrics-address = ":9876"

[log]
core = ""
network = "trace"
rpc = ""
state = ""
runtime = ""
babe = ""
grandpa = ""
sync = "trace"
digest = ""

[init]
genesis = "./chain/westend-local/westend-local-spec-raw.json"

[account]
key = "bob"
unlock = ""

[core]
roles = 4
babe-authority = true
grandpa-authority = true
babe-lead = true

[network]
port = 7001
nobootstrap = false
nomdns = false

[rpc]
enabled = true
ws = true
port = 8545
host = "localhost"
modules = [
    "system",
    "author",
    "chain",
    "state",
    "rpc",
    "grandpa",
    "offchain",
    "childstate",
    "syncstate",
    "payment",
]
ws-port = 8546

[pprof]
listening-address = "localhost:6060"
block-rate = 0
mutex-rate = 0

```

kishansagathiya:
Can reproduce this. I will check what's causing this.

PR #2955, opened 2022-11-16
fixing `bitfields and heads must be included every block`

kishansagathiya:
## Tests

```sh
```

## Issues
Fixes https://github.com/ChainSafe/gossamer/issues/2937

## Primary Reviewer

@qdm12

### DIFF
--- a/lib/trie/trie.go
+++ b/lib/trie/trie.go
@@ -333,11 +333,6 @@ func (t *Trie) Put(keyLE, value []byte) {
 
 func (t *Trie) insertKeyLE(keyLE, value []byte, deletedMerkleValues map[string]struct{}) {
 	nibblesKey := codec.KeyLEToNibbles(keyLE)
-	if len(value) == 0 {
-		// Force value to be inserted to nil since we don't
-		// differentiate between nil and empty values.
-		value = nil
-	}
 	t.root, _, _ = t.insert(t.root, nibblesKey, value, deletedMerkleValues)
 }
 


lib/trie/trie.go
```diff
@@ -333,11 +333,11 @@ func (t *Trie) Put(keyLE, value []byte) {
...
-	if len(value) == 0 {
-		// Force value to be inserted to nil since we don't
-		// differentiate between nil and empty values.
-		value = nil
-	}
+	// if len(value) == 0 {
+	// 	// Force value to be inserted to nil since we don't
+	// 	// differentiate between nil and empty values.
+	// 	value = nil
+	// }
```
qdm12, line 340: Just peeking out of curiosity; we need this and the spec does not differentiate empty byte slices and nil byte slices (it's just 'empty'). The `bitfields and heads must be included every block` error happened before this change as well AFAIK. Finally, one more note is that before #2927 we would have funny trie root hashes (because of empty byte slices and kinda wrong encoding), maybe that could had been the cause?

kishansagathiya: Also, please note that the reason why I have kept this draft is because I am still exploring other solutions.

qdm12: Yes, just peeking out of curiosity as I said.
Try rebasing development and dropping that commit to check if it works fine?

As you said, Rust doesn't have `nil`. For now our trie code only have `nil` node values (and not `{}`), but that can be changed as well to `{}` instead of `nil`, provided we have consistency. Consistency might not even be that needed, but it helps mentally when debugging stuff. What is important is the `n.SubValue != nil` check in

https://github.com/ChainSafe/gossamer/blob/8a17f3779140a736d2e348749b7708963c458558/internal/trie/node/encode.go#L40-L48

You could change that to check for length > 0 instead of a nil check.

kishansagathiya: dropping that commit has a bunch of conflict. So, instead I removed the part which was likely to create such a change, which is this branch. And this branch works fine.

lib/trie/trie.go
```diff
@@ -333,11 +333,6 @@ func (t *Trie) Put(keyLE, value []byte) {
 
 func (t *Trie) insertKeyLE(keyLE, value []byte, deletedMerkleValues map[string]struct{}) {
 	nibblesKey := codec.KeyLEToNibbles(keyLE)
-	if len(value) == 0 {
-		// Force value to be inserted to nil since we don't
-		// differentiate between nil and empty values.
-		value = nil
-	}
```
qdm12, line 340: Sorry to block this; but all our code checks for empty values with `!= nil` and although that might the issue here, I don't want to risk other ramifications.

Especially:

- the trie node encoding would result in encoding empty node values as `{0}` for branch-without-value nodes, which is wrong according to spec and [substate code](https://github.com/paritytech/substrate/blob/a7ba55d3c54b9957c242f659e02f5b5a0f47b3ff/primitives/trie/src/node_codec.rs#L123). Our decoding would still work since we interpret `EOF` and `{0}` the same, but it's not necessarily the same with substrate. And it's an unneeded `{0}` as well.
- pretty much all trie methods rely on `SubValue != nil` checks

After re-reading Substrate code and the spec, #2927 is still correct and our code before was basically wrong. It was perhaps making it (still wrongly) thanks to the cached encoding of nodes, which got removed in #2919 

As much as I hate to throw the ball elsewhere, I think the problem you are facing is due to something else (and didn't show up because of our wrong node encoding caching before). I'll ask on the elements chat about node encoding/decoding just to double check though.

qdm12 (changes requested):

EclesioMeloJunior:
Since https://github.com/ChainSafe/gossamer/pull/2964 was merged and it points out:

```
and closes https://github.com/ChainSafe/gossamer/pull/2955
```

I think we can close this PR, right? @kishansagathiya @qdm12

qdm12:
Closing this in favor of #2964

PR #2961, opened 2022-11-18
chore(trie): add trie heap usage workflow step

qdm12:
## Changes

- Fix `mutateTrieLeavesAtPrefix` (shame on previous me!)
- Fix `mutateTrieLeavesAtPrefix` to work with empty values
- Tigthen ratio brackets to newer values
- Add trie memory workflow step

## Primary Reviewer

@timwu20

### DIFF
--- a/lib/trie/mem_test.go
+++ b/lib/trie/mem_test.go
@@ -43,8 +43,8 @@ func Test_Trie_MemoryUsage(t *testing.T) {
 	// Check heap memory usage - it should be 2X
 	filledTrieHeap := getHeapUsage()
 	ratio := getApproximateRatio(halfFilledTrieHeap, filledTrieHeap)
-	assert.Greater(t, ratio, 1.5)
-	assert.Less(t, ratio, 2.1)
+	assert.Greater(t, ratio, 1.6)
+	assert.Less(t, ratio, 1.7)
 
 	// Snapshot the trie
 	triesMap["second"] = triesMap["first"].Snapshot()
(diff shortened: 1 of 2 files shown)


lib/trie/mem_test.go
```diff
@@ -16,7 +16,7 @@ func Test_Trie_MemoryUsage(t *testing.T) {
 	// Set skip to false to run the test.
 	// This test should be run on its own since it interacts
 	// with the Go garbage collector.
-	const skip = true
+	const skip = false
```
jimjbrettj, line 19: Will this effect effect tests if this test isn't run in isolation? Looking at above comment

jimjbrettj (changes requested):

lib/trie/mem_test.go
qdm12: Good point, changed it back to `true` since it should be run isolated. The CI takes care of changing that constant (ugly `sed` hack) to `true` when needed.

lib/trie/mem_test.go
```diff
@@ -102,13 +102,17 @@ func mutateTrieLeavesAtPrefix(trie *Trie,
...
-			newValue[i], newValue[j] = newValue[j], newValue[i]
+		var newValue []byte
+		if len(value) == 0 {
+			newValue = []byte{1}
+		} else {
+			newValue = make([]byte, len(value))
+			copy(newValue, value)
+			for i := range newValue {
+				newValue[i]++
+			}
```
EclesioMeloJunior, line 113:
```suggestion
		newValue := []byte{1}
		if len(value) != 0 {
		        newValue = make([]byte, len(value))
			for i := range newValue {
				newValue[i] = value[i] + 1
			}
		}
			
```

EclesioMeloJunior (approved):

timwu20 (approved):

lib/trie/mem_test.go
qdm12: Nice nit

jimjbrettj (approved):

Issue #1945, opened 2021-10-30
chore(logs): add `String() string` methods on complex types logged


qdm12:
## Issue summary

When reviewing most of the log usage in #1928 I noticed several structures did not have an associated `.String() string` method, which makes the log ugly and bit pointless (e.g. who cares about a pointer address).

The following structures were identified as logged out, complex enough and without a String method:

- `GrandpaMessage`
- `CommitMessage`
- `VoteMessage`
- `runtime.Context`
- `PublicKey` (mirror `Hex` with `String`)
- `babe.Slot`
- `dot.LogConfig`
- `types.GrandpaScheduledChange`
- `scale.VaryingDataType`
- `types.EpochData`
- `types.GrandpaForcedChange`
- `types.ConfigData`
- `types.Header`

PR #2965, opened 2022-11-20
chore(types): add String unit tests for Digest and fix NewDigest comment

qdm12:
## Changes

Sunday evening fun after getting sick of seeing horrible digest logged as some weird convoluted fields that don't make sense.

- `NewDigest` comment fixed
- `Test_BlockAnnounceMessage_String` unit test added
- `Test_Digest_String` unit test added

## Tests

```sh
go test -run ^Test_DigestToString$ github.com/ChainSafe/gossamer/dot/types
```

## Issues

One step towards #1945 

## Primary Reviewer

@EclesioMeloJunior


dot/types/digest.go
```diff
@@ -15,11 +16,40 @@ func NewDigestItem() scale.VaryingDataType {
...
 }
 
-// NewDigest returns a new Digest from the given DigestItems
+// NewDigest returns a new Digest as a varying data type slice.
 func NewDigest() scale.VaryingDataTypeSlice {
 	return scale.NewVaryingDataTypeSlice(NewDigestItem())
 }
 
+// DigestToString renders a digest varying data type slice as a string.
+func DigestToString(digest scale.VaryingDataTypeSlice) (s string) {
```
EclesioMeloJunior, line 25: is it possible to create a `String` method  at `VaryingDataTypeSlice` struct?

dot/types/digest.go
```diff
@@ -15,11 +16,40 @@ func NewDigestItem() scale.VaryingDataType {
...
+			element = impl.String()
+		case PreRuntimeDigest:
+			element = impl.String()
+		case ConsensusDigest:
+			element = impl.String()
+		case SealDigest:
+			element = impl.String()
+		default:
+			panic(fmt.Sprintf("digest type not implemented for String: %T", impl))
+		}
```
EclesioMeloJunior, line 45: I think almost all `VaryingDataType` implementations implements `String` method we could change the `VaryingDataTypeValue` interface to:

```go
type VaryingDataTypeValue interface {
	fmt.Stringer
	Index() uint
}
```

Then you can just use `elements = append(elements, vdtValue.String())`

EclesioMeloJunior (changes requested):

dot/types/digest.go
qdm12: Great suggestion. It required a bit more work than expected, a bunch of types were missing the String method, but it's done now in #2970 and `DigestToString` is no longer needed (this PR just ends up being a unit test :wink:)

dot/types/digest.go
qdm12: Yes, done in https://github.com/ChainSafe/gossamer/pull/2970 :+1:

jimjbrettj:
Can this be closed now due to https://github.com/ChainSafe/gossamer/pull/2970 ?

timwu20 (approved):

PR #2966, opened 2022-11-20
chore(all): review embedded mutex fields

qdm12:
## Changes

Why:
- Embedded fields do not show as unused when unused
- Embedded mutexes export Lock and Unlock methods which should often not be exported
- Better to name the field for readability/explicitness
- Embedded fields are annoying to navigate / ctrl+click in IDEs
- Embedding makes asserting entire struct more complicated
- There is no real need to embed fields most of the time, so why do it anyway

:warning: before merging, this should be added to eng wiki

Changes done:

- `dot/network`: remove unused embedded mutex in `ConnManager`
- Use named mutex field for:
  - lib/grandpa: `commitsTracker`, `votesTracker`
  - dot/network: `ConnManager`
  - dot/peerset: `PeerSet`, `PeerState`
  - dot/state: `BlockState`, `StorageState`
  - dot/state/pruner: `FullNode`
  - dot/sync: `chainSync`, `disjoingBlockSet`, `workerState`
  - dot/telemetry: `telemetryConnection`
  - lib/babe: `Service`
  - lib/blocktree: `BlockTree`, `leafMap`
  - lib/crypto: `SignatureVerifier`
  - lib/transaction: `PriorityQueue`
  - pkg/scale: `fieldScaleIndicesCache`
- dot/core: remove embedded mutex in `Service`, which was protecting... a channel?
  - Change service to use `stopCh` instead of context
  - Do not close block channel in stop method
  - Do not add to block queue in goroutine (bad backpressure)
  - Only exit from handleBlockAsync when stop channel gets closed

### DIFF
--- a/dot/core/service.go
+++ b/dot/core/service.go
@@ -35,6 +35,7 @@ type QueryKeyValueChanges map[string]string
 // and blocks by calling their respective validation functions in the runtime.
 type Service struct {
 	stopCh     chan struct{}
+	doneCh     chan struct{}
 	blockAddCh chan *types.Block // for asynchronous block handling
 
 	// Service interfaces
@@ -71,16 +72,15 @@ type Config struct {
 func NewService(cfg *Config) (*Service, error) {
 	logger.Patch(log.SetLevel(cfg.LogLvl))
 
-	blockAddCh := make(chan *types.Block, 256)
-
 	srv := &Service{
 		stopCh:               make(chan struct{}),
+		doneCh:               make(chan struct{}),
 		keys:                 cfg.Keystore,
 		blockState:           cfg.BlockState,
 		storageState:         cfg.StorageState,
 		transactionState:     cfg.TransactionState,
 		net:                  cfg.Network,
-		blockAddCh:           blockAddCh,
+		blockAddCh:           make(chan *types.Block, 256),
 		codeSubstitute:       cfg.CodeSubstitutes,
 		codeSubstitutedState: cfg.CodeSubstitutedState,
 	}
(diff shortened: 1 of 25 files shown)


dot/core/service.go
```diff
@@ -299,11 +284,7 @@ func (s *Service) handleCodeSubstitution(hash common.Hash,
 func (s *Service) handleBlocksAsync() {
 	for {
 		select {
-		case block, ok := <-s.blockAddCh:
-			if !ok {
-				return
-			}
```
qdm12, line 305: only return when signaled from `stopCh`

dot/core/service.go
```diff
@@ -236,15 +226,10 @@ func (s *Service) handleBlock(block *types.Block, state *rtstorage.TrieState) er
...
 
-	go func() {
-		s.Lock()
-		defer s.Unlock()
-		if s.ctx.Err() != nil {
-			return
-		}
-
-		s.blockAddCh <- block
-	}()
```
qdm12, line 247: We have a buffered block add channel (256), so we should not 'add to the queue' with goroutines, otherwise goroutines act as the 2nd queue when the channel is full. We should instead block to have proper back pressure.

EclesioMeloJunior (changes requested):
LGTM! Just a small comment over the `Stop` method

dot/core/service.go
```diff
@@ -102,11 +96,7 @@ func (s *Service) Start() error {
 
 // Stop stops the core service
 func (s *Service) Stop() error {
-	s.Lock()
-	defer s.Unlock()
-
-	s.cancel()
-	close(s.blockAddCh)
+	close(s.stopCh)
```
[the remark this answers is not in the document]
qdm12: Kinda out of scope, but I couldn't hold myself, done in 592d4cbd95e9f0a5b1cb4bb8a455a1bf0b89ba73

Note I changed `handleBlocksAsync` to be `handleBlocksAsync(stop <-chan struct{}, done chan<- struct{})` to force only consumption/production on each channel, since I have seen horrible things in the past I really want to avoid!

PR #2970, opened 2022-11-23
feat(pkg/scale): `VaryingDataTypeValue` String method

qdm12:
## Changes

As proposed by Eclesio in https://github.com/ChainSafe/gossamer/pull/2965#discussion_r1028345455

1. ~`VaryingDataTypeValue` interface is now composed of `fmt.Stringer`~
2. Add missing `String() string` methods to types needing to implement `VaryingDataTypeValue`
3. Add `String() string` method on `VaryingDataType` struct
4. Add `String() string` method on `VaryingDataTypeSlice` struct

## Issues

A few more steps towards #1945 

## Primary Reviewer

@EclesioMeloJunior

### DIFF
--- a/pkg/scale/varying_data_type.go
+++ b/pkg/scale/varying_data_type.go
@@ -10,6 +10,7 @@ import (
 // VaryingDataTypeValue is used to represent scale encodable types of an associated VaryingDataType
 type VaryingDataTypeValue interface {
 	Index() uint
+	String() string
 }
 
 // VaryingDataTypeSlice is used to represent []VaryingDataType. SCALE requires knowledge
--- a/pkg/scale/varying_data_type.go
+++ b/pkg/scale/varying_data_type.go
@@ -75,6 +75,13 @@ func (vdt *VaryingDataType) Value() (VaryingDataTypeValue, error) {
 	return vdt.value, nil
 }
 
+func (vdt VaryingDataType) String() string {
+	if vdt.value == nil {
+		return "VaryingDataType(nil)"
+	}
+	return vdt.value.String()
+}
+
 // NewVaryingDataType is constructor for VaryingDataType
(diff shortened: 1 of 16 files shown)


EclesioMeloJunior (approved):
LGTM! Happy with log improvements as well as relying on type systems instead of runtime type checks!! 🚀

pkg/scale/varying_data_type.go
```diff
@@ -5,11 +5,13 @@ package scale
 
 import (
 	"fmt"
+	"strings"
 )
 
 // VaryingDataTypeValue is used to represent scale encodable types of an associated VaryingDataType
 type VaryingDataTypeValue interface {
 	Index() uint
+	String() string
```
timwu20, line 14: I don't think we need to enforce this for all `VaryingDataTypeValue` implementations.

timwu20 (changes requested):

pkg/scale/varying_data_type.go
jimjbrettj: I would say I agree with this

jimjbrettj (approved):
Aside from Tims comment this looks good to me

pkg/scale/varying_data_type.go
qdm12: Good point, removed from interface and changed `VaryingDataType`'s String method to be:

```go
func (vdt VaryingDataType) String() string {
	if vdt.value == nil {
		return "VaryingDataType(nil)"
	}
	stringer, ok := vdt.value.(fmt.Stringer)
	if !ok {
		return fmt.Sprintf("VaryingDataType(%v)", vdt.value)
	}
	return stringer.String()
}
```

Does that work with you? 🤔

timwu20 (approved):

PR #2973, opened 2022-11-24
fix(wasmer): `ext_storage_exists_version_1` for empty values

qdm12:
## Changes

- If the value is `[]byte{}`, the key exists (leaf node, branch with value)
- If the value is `[]byte(nil)`, the key does not exist (node not found or branch without value)
- Change unit test to be cases-based and add case

## Tests

```sh
go test -run ^Test_ext_storage_exists_version_1$ github.com/ChainSafe/gossamer/lib/runtime/wasmer
```

## Primary Reviewer

@timwu20

### DIFF
--- a/lib/runtime/wasmer/imports.go
+++ b/lib/runtime/wasmer/imports.go
@@ -1910,8 +1910,8 @@ func ext_storage_exists_version_1(context unsafe.Pointer, keySpan C.int64_t) C.i
 	key := asMemorySlice(instanceContext, keySpan)
 	logger.Debugf("key: 0x%x", key)
 
-	val := storage.Get(key)
-	if len(val) > 0 {
+	value := storage.Get(key)
+	if value != nil {
 		return 1
 	}
 
--- a/lib/runtime/wasmer/imports_test.go
+++ b/lib/runtime/wasmer/imports_test.go
@@ -509,26 +509,51 @@ func Test_ext_storage_get_version_1(t *testing.T) {
 
 func Test_ext_storage_exists_version_1(t *testing.T) {
 	t.Parallel()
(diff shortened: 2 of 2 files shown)


lib/runtime/wasmer/imports.go
```diff
@@ -1920,8 +1920,8 @@ func ext_storage_exists_version_1(context unsafe.Pointer, keySpan C.int64_t) C.i
 	key := asMemorySlice(instanceContext, keySpan)
 	logger.Debugf("key: 0x%x", key)
 
-	val := storage.Get(key)
-	if len(val) > 0 {
+	value := storage.Get(key)
```
EclesioMeloJunior, line 1923: Wondering if we could return an error from `Get(key)` method like we do for database `Get` method where it returns `ErrKeyNotFound`

qdm12: Substrate returns a result with byte array or `None`, I think the analogy `[]byte(nil)` <-> `None` fits fine, especially since it also matches the implementation (`StorageValue = nil` in the trie code).

EclesioMeloJunior: Yeah, I thought would be more golang-ish this way:

```go
value, err := storage.Get(key)
if errors.Is(err, ErrKeyNotFound) {}
```
but that was just a thought and I belive it will require a more effort/deltas to accomplish so I am good with the current changes

EclesioMeloJunior (approved):

lib/runtime/wasmer/imports.go
qdm12: Yeah I'm kinda shared too :smile: 
But since we now (finally duh) have a clear understanding `nil` means empty in the trie implementation/interface/spec (unlike `[]byte{}`), I guess we can keep it as is too.

PR #2984, opened 2022-11-29
chore(lib/grandpa): use map set for authority keys for faster lookups

### LINKED ISSUE #2960: lib/grandpa: use authority keys set for `verifyJustification`
`authorities` is only used to check against a justification key in `verifyJustification`, it would be better to inject only authority keys as a set `map[string]struct{}` down to `verifyJustification` for faster lookups and to only inject what's necessary.

The grandpa service `authorities` method can be changed to:

```go
func (s *Service) authorityKeysSet() (authorityKeys map[string]struct{}) {
	authorityKeys = make(map[string]struct{}, len(s.state.voters))
	for _, voter := range s.state.voters {
		authority := &types.Authority{
			Key:    &voter.Key,
			Weight: voter.ID,
		}
		encodedKey := authority.Key.Encode()
		authorityKeys[string(encodedKey)] = struct{}{}
	}

	return authorityKeys
}
```

That way the block in `verifyJustification`:

```go
	for _, auth := range authorities {
		if bytes.Equal(auth.Key.Encode(), justificationKey) {
			return nil
		}
	}
```

can be replaced with faster lookups using

```go
_, has := authorityKeys[justificationKey]
if has {
  return nil
}
```

lakshya-sky:
## Issues

Fixes #2960 

## Primary Reviewer

@timwu20

### DIFF
--- a/lib/grandpa/grandpa.go
+++ b/lib/grandpa/grandpa.go
@@ -200,16 +200,18 @@ func (s *Service) Stop() error {
 }
 
 // authorities returns the current grandpa authorities
-func (s *Service) authorities() []*types.Authority {
-	ad := make([]*types.Authority, len(s.state.voters))
-	for i := 0; i < len(s.state.voters); i++ {
-		ad[i] = &types.Authority{
-			Key:    &s.state.voters[i].Key,
-			Weight: s.state.voters[i].ID,
+func (s *Service) authorityKeySet() (authorityKeys map[string]*types.Authority) {
+	authorityKeys = make(map[string]*types.Authority, len(s.state.voters))
+	for _, voter := range s.state.voters {
+		authority := &types.Authority{
+			Key:    &voter.Key,
+			Weight: voter.ID,
 		}
+		encodedKey := authority.Key.Encode()
+		authorityKeys[string(encodedKey)] = authority
 	}
 
-	return ad
+	return authorityKeys
 }
 
 // updateAuthorities updates the grandpa voter set, increments the setID, and resets the round numbers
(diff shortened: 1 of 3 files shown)


lib/grandpa/grandpa.go
```diff
@@ -1354,10 +1356,9 @@ func verifyJustification(justification *SignedVote, round, setID uint64,
 		return fmt.Errorf("encoding authority key: %w", err)
 	}
 
-	for _, auth := range authorities {
-		if bytes.Equal(auth.Key.Encode(), justificationKey) {
-			return nil
-		}
+	_, not := authorityKeys[string(justificationKey)]
+	if not {
```
qdm12, line 1360: rename `not` to `has`/`exists`/`ok`/`keyInAuthorities`

lib/grandpa/grandpa.go
```diff
@@ -199,16 +199,18 @@ func (s *Service) Stop() error {
 }
 
 // authorities returns the current grandpa authorities
-func (s *Service) authorities() []*types.Authority {
-	ad := make([]*types.Authority, len(s.state.voters))
-	for i := 0; i < len(s.state.voters); i++ {
-		ad[i] = &types.Authority{
-			Key:    &s.state.voters[i].Key,
-			Weight: s.state.voters[i].ID,
+func (s *Service) authorityKeySet() (authorityKeys map[string]*types.Authority) {
```
qdm12, line 202: As mentioned in https://github.com/ChainSafe/gossamer/issues/2960 we want a **set** so no value in the map. Change `map[string]*types.Authority` to `map[string]struct{}`. This is because we never use the map values, we only care about the key being present or not. Also `struct{}{}` uses 0 byte unlike any other type.

FYI to set a key in such map 'set', use `authorityKeys[string(encodedKey)] = struct{}{}`

qdm12 (changes requested):
Thanks!!

Please change the map to a set (well a map, without value)

EclesioMeloJunior (approved):
LGTM! thanks for the modifications! 🚀

qdm12 (approved):
Performance whoop whoop :rocket: 
Thanks! :+1:

PR #2985, opened 2022-11-29
chore(lib/common): remove Equal method for Hash and replace with ==

### LINKED ISSUE #2939: `lib/common`: remove `Equal` method for `Hash` and replace with `==`
We can equal hashes (32 bytes arrays) using just `==`, there is no point to have an `Equal` method, and even less to check equality using the method function body which converts both arrays to bytes slices and then to strings to check for equality

lakshya-sky:
## Issues

Fixes #2939


dot/state/block.go
```diff
@@ -621,7 +621,7 @@ func (bs *BlockState) HandleRuntimeChanges(newState *rtstorage.TrieState,
 
 	codeSubBlockHash := bs.baseState.LoadCodeSubstitutedBlockHash()
 
-	if !codeSubBlockHash.Equal(common.Hash{}) {
+	if codeSubBlockHash != common.EmptyHash() {
```
qdm12, line 624: nit keep it to

```suggestion
	if codeSubBlockHash != common.Hash{} {
```

not sure why we have `EmptyHash` in the first place, might be worth removing it as well.

lib/common/hash.go
```diff
@@ -22,6 +22,13 @@ const (
 // Hash used to store a blake2b hash
 type Hash [32]byte
 
+// EmptyHash returns Hash which is an array of 32bytes with
+// every byte set to zero.
+func EmptyHash() Hash {
+	res := Hash{}
+	return res
+}
```
qdm12, line 30: Yeah we don't want that, there is no point for it and performance-wise it adds a function call.

qdm12 (changes requested):
Thanks! Just need to remove that `EmptyHash` function :wink:

dot/state/block.go
lakshya-sky:
```
dot/state/block.go:624:39: expected ';', found '{' (typecheck)
        if codeSubBlockHash != common.Hash{} {
```

lakshya-sky: if I use 
```
codeSubBlockHash != (common.Hash{})
``` 
then error goes away, golang is so strange.

lib/common/hash.go
```diff
@@ -41,7 +41,7 @@ func HashValidator(field reflect.Value) interface{} {
 	// Try to convert to hash type.
 	if valuer, ok := field.Interface().(Hash); ok {
 		// Check if the hash is empty.
-		if valuer.Equal(Hash{}) {
+		if valuer == (Hash{}) {
 			return ""
```
qdm12, line 45: Let's remove `func (h Hash) Equal(g Hash) bool {` below

qdm12 (changes requested):

qdm12 (approved):

EclesioMeloJunior (approved):

qdm12:
Awesome, merged & thanks Darshan :+1:

Issue #2951, opened 2022-11-15
Implement `ext_crypto_ecdsa_generate` Host API


danforbes:
## Task summary

- Implement the `ext_crypto_ecdsa_generate` [Host API](https://spec.polkadot.network/#chap-host-api).

## Specification reference

- https://spec.polkadot.network/#_ext_crypto_ecdsa_generate

axaysagathiya:
merged @lakshya-sky's [PR](https://github.com/ChainSafe/polkadot-spec/pull/14) on chainsafe/polkadot-spec repo

PR #2987, opened 2022-12-01
feat(lib/runtime/wasmer) Implement `ext_crypto_ecdsa_generate_version_1`

lakshya-sky:
Added two new functions for secp256k1 to generate priv key from mnemonic, and get public key from encoded bytes.

## Issues

#2951


lakshya-sky:
Unit test for this one fails, I don't know what is wrong with it, I am doing everything same as other PRs for Wamser.

EclesioMeloJunior:
hey @lakshya-sky! This error happens because the runtime we use for this tests is not an production runtime but a mocked runtime, it is placed in another repository called https://github.com/ChainSafe/polkadot-spec and you should include this exported method in it. Take a look at this pull request for example: https://github.com/ChainSafe/polkadot-spec/pull/11

lib/runtime/wasmer/imports_test.go
```diff
@@ -881,6 +881,39 @@ func Test_ext_crypto_ed25519_verify_version_1(t *testing.T) {
...
+	var out []byte
+	err = scale.Unmarshal(ret, &out)
+	require.NoError(t, err)
+
+	pubKey, err := secp256k1.NewPublicKey(out)
+	require.NoError(t, err)
+	require.Equal(t, 1, ks.Size())
+
+	kp := ks.GetKeypair(pubKey)
+	require.NotNil(t, kp)
```
qdm12, line 914: Can we assert against an actual value instead of not nil only?

lib/runtime/wasmer/imports_test.go
```diff
@@ -881,6 +881,39 @@ func Test_ext_crypto_ed25519_verify_version_1(t *testing.T) {
 	require.NotNil(t, read)
 }
 
+func Test_ext_crypto_ecdsa_generate_version_1(t *testing.T) {
+	t.Parallel()
+	inst := NewTestInstance(t, runtime.HOST_API_TEST_RUNTIME)
```
qdm12, line 886: nit `inst` -> `instance`

lib/crypto/secp256k1/secp256k1.go
```diff
@@ -102,6 +111,27 @@ func NewKeypairFromPrivate(priv *PrivateKey) (*Keypair, error) {
...
+	}
+	edpriv, err := secp256k1.ToECDSA(seed)
+	if err != nil {
+		return nil, err
+	}
+	return NewKeypair(*edpriv), nil
+}
+
+// NewKeypairFromMnenomic returns a new Keypair using the given mnemonic and password.
+func NewKeypairFromMnenomic(mnemonic, password string) (*Keypair, error) {
```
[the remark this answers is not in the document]
lakshya-sky: we could, but I just used API from other keypairs and they provide passsword arg.

lib/runtime/wasmer/imports_test.go
```diff
@@ -881,6 +881,38 @@ func Test_ext_crypto_ed25519_verify_version_1(t *testing.T) {
...
+	seedData, err := scale.Marshal(&mnemonicBytes)
+	require.NoError(t, err)
+
+	params := append(idData, seedData...)
+
+	encodedKeyPair, err := instance.Exec("rtm_ext_crypto_ecdsa_generate_version_1", params)
+	require.NoError(t, err)
+
+	var out []byte
+	err = scale.Unmarshal(encodedKeyPair, &out)
```
lakshya-sky, line 905: @EclesioMeloJunior @qdm12 I would need a small favour, this function here gives 32 bytes slice while the input is 33 bytes slice. I tried looking at how scale codec works but couldn't find out why it wouldn't work.

lakshya-sky:
can anyone review https://github.com/ChainSafe/polkadot-spec/pull/14 so this PR can pass tests.

timwu20:
> can anyone review [ChainSafe/polkadot-spec#14](https://github.com/ChainSafe/polkadot-spec/pull/14) so this PR can pass tests. Thank you!

@lakshya-sky we are updating our runtime integration tests to use the `westend-local` chain.  You should be able to use that runtime in your integration test.  We will be removing the `gssmr` and `dev` chains soon.

PR #2977, opened 2022-11-25
feat(dot/sync): Remove the `EndBlockHash` from `BlockRequestMessage`

EclesioMeloJunior:
## Changes

- Remove the `EndBlockHash` from `BlockRequestMessage`
- Remove the `to_block` from the `api.v1.proto` file
## Tests

```sh
go test -tags integration github.com/ChainSafe/gossamer/dot/sync
go test -tags integration github.com/ChainSafe/gossamer/dot/network
```

## Issues

- Closes #2883

## Primary Reviewer

@edwardmack

PR #2992, opened 2022-12-06
chore(proto): fixes and CI check

qdm12:
## Changes

- `make proto` was not working, found whilst reviewing #2977 
- Use `go generate` to generate Go code from proto file
- Add CI check to ensure Go code is generated using expected proto-gen-go and protoc versions
- Re-generate using protoc v3.21.10 and proto-gen-go v1.28.1
- Print out `protoc` requirement when running `make proto`

## Primary Reviewer

@EclesioMeloJunior


jimjbrettj (approved):
lgtm!

PR #2994, opened 2022-12-06
chore(all): dependency inject databases and tables

### LINKED ISSUE #2981: Dependency inject databases and tables
In many places we create the database or the table in constructors (or even worst in Start methods), whereas it should ideally be injected as dependency instead.

This blocks some modifications such as using a database in the trie code, or creates duplicate table objects which may or may not be thread safe. It also makes changing our database implementation harder.

qdm12:
## Changes

Improves codebase health + simplify migration to newer database interface

- Dependency inject to `storeInitialValues`
- Dependency inject to `createRuntimeStorage`
- Dependency inject to `NewGrandpaState`
- Dependency inject to `NewEpochStateFromGenesis`
- Dependency inject to `NewEpochState`
- Dependency inject to `NewGrandpaStateFromGenesis`
- Dependency inject to `NewBlockStateFromGenesis`
- Dependency inject to `NewBlockState`
- Remove `LoadChainDB`, `LoadBadgerDB`, `runtime.NewInMemoryDB` and `utils.SetupDatabase`, `newInMemoryGrandpaDatabase`
- Minor test database setup simplifications
  - dot/state: make `newInMemoryDB` test only
  - Remove unneeded datadir setting for in memory databases
  - Add test `.Close()` no error assertions
  - Add missing test database `.Close()` calls
  - Use in memory database when amount is not big (`TestTrieDiff`)
  - Do not create unneeded `trie` table in `lib/trie`'s `newTestDB`
  - `dot/state`: use `t.TempDir()` as data dir directly
  - lib/trie: do not create table for TestTrieDiff

## Issues

Fixes #2981

### DIFF
--- a/cmd/gossamer/config.go
+++ b/cmd/gossamer/config.go
@@ -7,10 +7,12 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"path/filepath"
 	"strconv"
 	"strings"
 	"time"
 
+	"github.com/ChainSafe/chaindb"
 	"github.com/ChainSafe/gossamer/dot"
 	ctoml "github.com/ChainSafe/gossamer/dot/config/toml"
 	"github.com/ChainSafe/gossamer/dot/state"
@@ -897,8 +899,9 @@ func updateDotConfigFromGenesisJSONRaw(tomlCfg ctoml.Config, cfg *dot.Config) {
 
 // updateDotConfigFromGenesisData updates the configuration from genesis data of an initialised node
 func updateDotConfigFromGenesisData(ctx *cli.Context, cfg *dot.Config) error {
-	// initialise database using data directory
-	db, err := utils.SetupDatabase(cfg.Global.BasePath, false)
+	db, err := chaindb.NewBadgerDB(&chaindb.Config{
+		DataDir: filepath.Join(cfg.Global.BasePath, "db"),
+	})
 	if err != nil {
 		return fmt.Errorf("failed to create database: %s", err)
 	}
--- a/dot/node_integration_test.go
+++ b/dot/node_integration_test.go
@@ -14,6 +14,7 @@ import (
 	"path/filepath"
 	"testing"
 
+	"github.com/ChainSafe/chaindb"
 	"github.com/ChainSafe/gossamer/dot/core"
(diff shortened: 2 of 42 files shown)


cmd/gossamer/config.go
```diff
@@ -6,10 +6,12 @@ package main
 import (
 	"errors"
 	"fmt"
+	"path/filepath"
 	"strconv"
 	"strings"
 	"time"
 
+	"github.com/ChainSafe/chaindb"
```
timwu20, line 14: I thought we were getting rid of this?

timwu20 (commented):
Not sure why we need to merge this PR if you're going to be swapping out the DB entirely.

qdm12:
> Not sure why we need to merge this PR if you're going to be swapping out the DB entirely.

Since we dependency inject the database/tables as locally-defined interfaces, it means there will be less changes to be made (no need to repeat update database/table constructors)

cmd/gossamer/config.go
qdm12: This is just to remove the `utils.SetupDatabase` in favor of using the database implementation directly (since really that helper function isn't that necessary and adds complexity).

dot/node_integration_test.go
```diff
@@ -117,7 +117,7 @@ func TestNewNode(t *testing.T) {
 		return stateSrvc, nil
 	})
 
-	m.EXPECT().createRuntimeStorage(gomock.AssignableToTypeOf(&state.Service{})).Return(&runtime.
+	m.EXPECT().createRuntimeStorage(gomock.AssignableToTypeOf(&state.BaseState{}), gomock.Any()).Return(&runtime.
```
jimjbrettj, line 120: Can we not mock the specific type here?

jimjbrettj (approved):

dot/node_integration_test.go
qdm12: Lazy me, done 👍 
Required a bit of test refactoring, but worth it 💯

qdm12:
I can't push to this branch, it would be nice if someone can pick it up and

```sh
git checkout development
git pull
git checkout qdm12/dep-inject-db
git rebase development
git push -f
```

Thanksss

PR #1234, opened 2020-12-11
buildBlockExtrinsics to continue when apply fails

RyRy79261:
## Changes

- Removed error returns from buildBlockExtrinsics loop

## Tests

```
go test ./lib/babe/... -short
go test ./... -short
```

## Checklist

- [x] I have read [CODE_OF_CONDUCT](https://github.com/ChainSafe/gossamer/blob/development/.github/CODE_OF_CONDUCT.md) and [CONTRIBUTING](https://github.com/ChainSafe/gossamer/blob/development/.github/CONTRIBUTING.md) 
- [x] I have provided as much information as possible and necessary
- [ ] I have reviewed my own pull request before requesting a review
- [ ] All integration tests and required coverage checks are passing

## Issues

- closes #1225

Issue #2933, opened 2022-11-09
Submit GRANDPA equivocations to the runtime


EclesioMeloJunior:
## Task summary

- We should submit GRANDPA equivocations to the runtime when we identifies equivocators

## Specification reference

- https://spec.polkadot.network/#sect-grandpaapi_submit_report_equivocation_unsigned_extrinsic

## Other information and links

- https://github.com/paritytech/substrate/blob/a6da808575fe403ab2bd7f6cd009896cdc6fd71a/client/finality-grandpa/src/environment.rs#L562

EclesioMeloJunior:
Closed due https://github.com/ChainSafe/gossamer/pull/3007

PR #2997, opened 2022-12-07
dot/types: add equivocation vdt

jimjbrettj:
## Changes

- Add an equivocation type to create a wrapper around signed grandpa votes for proper encoding/decoding in runtime calls. This mimicks substrate as seen here https://crates.parity.io/sp_finality_grandpa/enum.Equivocation.html

## Tests

```sh
go test ./dot/types 
```

## Issues

#2933

## Primary Reviewer

@timwu20

### DIFF
--- a/dot/types/grandpa.go
+++ b/dot/types/grandpa.go
@@ -207,7 +207,7 @@ type GrandpaEquivocation struct {
 // GrandpaEquivocationVote is a custom vdt type for a grandpa equivocation
 type GrandpaEquivocationVote scale.VaryingDataType
 
-// Set will set a VaryingDataTypeValue using the underlying VaryingDataType
+// Set sets a VaryingDataTypeValue using the underlying VaryingDataType
 func (ge *GrandpaEquivocationVote) Set(value scale.VaryingDataTypeValue) (err error) {
 	vdt := scale.VaryingDataType(*ge)
 	err = vdt.Set(value)
--- a/dot/types/grandpa.go
+++ b/dot/types/grandpa.go
@@ -193,3 +193,20 @@ type GrandpaVote struct {
 func (v *GrandpaVote) String() string {
 	return fmt.Sprintf("hash=%s number=%d", v.Hash, v.Number)
 }
+
+// NewGrandpaEquivocation returns a new VaryingDataType to represent a grandpa Equivocation
+func NewGrandpaEquivocation() scale.VaryingDataType {
+	return scale.MustNewVaryingDataType(PreVoteEquivocation{}, PreCommitEquivocation{})
+}
+
+// PreVoteEquivocation equivocation type for a prevote
+type PreVoteEquivocation GrandpaSignedVote
+
+// Index returns VDT index
+func (PreVoteEquivocation) Index() uint { return 0 }
+
+// PreCommitEquivocation equivocation type for a precommit
+type PreCommitEquivocation GrandpaSignedVote
+
+// Index returns VDT index
+func (PreCommitEquivocation) Index() uint { return 1 }
--- a/dot/types/grandpa.go
+++ b/dot/types/grandpa.go
@@ -194,9 +194,34 @@ func (v *GrandpaVote) String() string {
 	return fmt.Sprintf("hash=%s number=%d", v.Hash, v.Number)
 }
 
+// GrandpaEquivocation is a custom vdt type for a grandpa equivocation
+type GrandpaEquivocation scale.VaryingDataType
+
+// Set will set a VaryingDataTypeValue using the underlying VaryingDataType
+func (ge *GrandpaEquivocation) Set(val scale.VaryingDataTypeValue) (err error) {
+	vdt := scale.VaryingDataType(*ge)
+	err = vdt.Set(val)
+	if err != nil {
+		return err
+	}
+	*ge = GrandpaEquivocation(vdt)
+	return nil
+}
+
+// Value will return the value from the underlying VaryingDataType
+func (ge *GrandpaEquivocation) Value() (val scale.VaryingDataTypeValue, err error) {
+	vdt := scale.VaryingDataType(*ge)
+	return vdt.Value()
+}
+
 // NewGrandpaEquivocation returns a new VaryingDataType to represent a grandpa Equivocation
-func NewGrandpaEquivocation() scale.VaryingDataType {
-	return scale.MustNewVaryingDataType(PreVoteEquivocation{}, PreCommitEquivocation{})
+func NewGrandpaEquivocation() *GrandpaEquivocation {
+	vdt, err := scale.NewVaryingDataType(PreVoteEquivocation{}, PreCommitEquivocation{})
+	if err != nil {
+		panic(err)
+	}
+	ge := GrandpaEquivocation(vdt)
+	return &ge
 }
 
 // PreVoteEquivocation equivocation type for a prevote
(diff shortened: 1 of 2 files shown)


dot/types/grandpa.go
```diff
@@ -193,3 +193,45 @@ type GrandpaVote struct {
...
+func (ge *GrandpaEquivocation) Value() (val scale.VaryingDataTypeValue, err error) {
+	vdt := scale.VaryingDataType(*ge)
+	return vdt.Value()
+}
+
+// NewGrandpaEquivocation returns a new VaryingDataType to represent a grandpa Equivocation
+func NewGrandpaEquivocation() *GrandpaEquivocation {
+	vdt, err := scale.NewVaryingDataType(PreVoteEquivocation{}, PreCommitEquivocation{})
+	if err != nil {
+		panic(err)
```
kishansagathiya, line 221: opinion: return is better than panic

last line in go proverbs https://go-proverbs.github.io/

dot/types/grandpa_test.go
```diff
@@ -12,6 +12,35 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+func TestEncodeAndDecodeEquivocationPreVote(t *testing.T) {
+	exp := common.MustHexToBytes("0x000a0b0c0d00000000000000000000000000000000000000000000000000000000e7030000010203040000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000506070800000000000000000000000000000000000000000000000000000000") //nolint:lll
```
kishansagathiya, line 16: I think we should test that encoding of equivocation produced by substrate can be decoded by us.
Did we get the above value from encoding [sp_finality_grandpa](https://crates.parity.io/sp_finality_grandpa/index.html)::[Equivocation](https://crates.parity.io/sp_finality_grandpa/enum.Equivocation.html#) ?

kishansagathiya (commented):
Looks good other than a few comments

dot/types/grandpa_test.go
qdm12, line 16: nit `exp` -> `expectedEncoding`

dot/types/grandpa_test.go
```diff
@@ -12,6 +12,35 @@ import (
...
+	var testVote = GrandpaVote{
+		Hash:   common.Hash{0xa, 0xb, 0xc, 0xd},
+		Number: 999,
+	}
+	var testSignature = [64]byte{1, 2, 3, 4}
+	var testAuthorityID = [32]byte{5, 6, 7, 8}
+
+	sv := GrandpaSignedVote{
```
qdm12, line 24: nit `sv` -> `signedVote`

dot/types/grandpa_test.go
```diff
@@ -12,6 +12,35 @@ import (
...
+		Vote:        testVote,
+		Signature:   testSignature,
+		AuthorityID: testAuthorityID,
+	}
+
+	equivPreVote := PreVoteEquivocation(sv)
+	equivVote := NewGrandpaEquivocation()
+	err := equivVote.Set(equivPreVote)
+	require.NoError(t, err)
+	enc, err := scale.Marshal(*equivVote)
```
qdm12, line 34: nit `enc` -> `encoding` or `equivPreVoteEncoding`

dot/types/grandpa_test.go
```diff
@@ -12,6 +12,35 @@ import (
...
+	require.NoError(t, err)
+	require.Equal(t, exp, enc)
+
+	res := NewGrandpaEquivocation()
```
qdm12, line 38: nit `res` -> `grandpaEquivocation`

qdm12 (approved):
:+1: 

Also for the PR issues section, we should use `Fixes #1234` so merging the PR auto-closes the issue.

dot/types/grandpa_test.go
jimjbrettj: I believe so? I actually wasnt the one to get the bytes, I used it from another test in this file, and since its encoded as a VDT it just pretends a 0

jimjbrettj:
@qdm12 actually we dont want it to close the issue in this case as its just a component of the overall issue. But will do that for my next PR!

dot/types/grandpa.go
jimjbrettj: In other use cases on VDT constructors in the code I see that we use Must... so I changed it to that

dot/types/grandpa_test.go
```diff
@@ -12,6 +12,35 @@ import (
...
 )
 
+func TestEncodeAndDecodeEquivocationPreVote(t *testing.T) {
+	exp := common.MustHexToBytes("0x000a0b0c0d00000000000000000000000000000000000000000000000000000000e7030000010203040000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000506070800000000000000000000000000000000000000000000000000000000") //nolint:lll
+	var testVote = GrandpaVote{
+		Hash:   common.Hash{0xa, 0xb, 0xc, 0xd},
+		Number: 999,
+	}
+	var testSignature = [64]byte{1, 2, 3, 4}
+	var testAuthorityID = [32]byte{5, 6, 7, 8}
```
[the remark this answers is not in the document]
jimjbrettj: fixed this in below tests as well!

dot/types/grandpa_test.go
jimjbrettj: Applied naming changes to rest of file as well!

jimjbrettj:
This is actually not correct, the equivocation should not be of type GrandpaSignedVote, but should be a different equivocation struct as seen here https://github.com/paritytech/finality-grandpa/blob/eb26b0c86dbb6949435b627f85eceb026f96ebe9/src/lib.rs#L210

Will update and rerequest review

jimjbrettj:
updated

dot/types/grandpa.go
```diff
@@ -193,3 +193,52 @@ type GrandpaVote struct {
...
+	FirstVote       GrandpaVote
+	FirstSignature  [64]byte
+	SecondVote      GrandpaVote
+	SecondSignature [64]byte
+}
+
+// GrandpaEquivocationVote is a custom vdt type for a grandpa equivocation
+type GrandpaEquivocationVote scale.VaryingDataType
+
+// Set will set a VaryingDataTypeValue using the underlying VaryingDataType
```
qdm12, line 210: supernit `will` feels like it's doing async stuff, just use present?

```suggestion
// Set sets a VaryingDataTypeValue using the underlying VaryingDataType
```

dot/types/grandpa_test.go
```diff
@@ -7,55 +7,94 @@ import (
 	"testing"
 
 	"github.com/ChainSafe/gossamer/lib/common"
+	"github.com/ChainSafe/gossamer/lib/crypto/ed25519"
 	"github.com/ChainSafe/gossamer/pkg/scale"
 
 	"github.com/stretchr/testify/require"
 )
 
+func TestEncodeAndDecodeEquivocationPreVote(t *testing.T) {
```
qdm12, line 16:
```suggestion
func TestEncodeAndDecodeEquivocationPreVote(t *testing.T) {
	t.Parallel()

```

qdm12 (approved):
Nice :medal_military:

dot/types/grandpa_test.go
kishansagathiya: @jimjbrettj 
Would nice if you can take that rust struct and encode and and use those bytes here. That way we know that the struct created by us is correct. It would also prevent people from un-necessarily modifying that struct.

EclesioMeloJunior (approved):

dot/types/grandpa_test.go
jimjbrettj: Yes I will do this, good call

PR #3002, opened 2022-12-12
chore(all): remove unneeded timer channel drains

qdm12:
## Changes

:sob: Sorry I've been wrong and annoying everyone with draining timer channels when stopping them...
Basically (duh) a channel is just like any other pointer, so if it goes out of scope it gets GC'd. There is no need to drain it, even if it's buffered, as long as it goes out of scope, it'll get GC'd. So that means timers don't need to have their channels drained. The only reason to drain them (the timer documentation isn't so clear on that) is to reset them, which we don't use at all in our codebase (except once where the timer already expired and was drained)

### DIFF
--- a/dot/network/discovery.go
+++ b/dot/network/discovery.go
@@ -155,9 +155,7 @@ func (d *discovery) advertise() {
 
 		select {
 		case <-d.ctx.Done():
-			if !timer.Stop() {
-				<-timer.C
-			}
+			timer.Stop()
 			return
 		case <-timer.C:
 			logger.Debug("advertising ourselves in the DHT...")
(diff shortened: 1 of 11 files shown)


dot/network/discovery.go
```diff
@@ -217,10 +215,6 @@ func (d *discovery) findPeers() {
 			logger.Tracef("found new peer %s via DHT", peer.ID)
 			d.h.Peerstore().AddAddrs(peer.ID, peer.Addrs, peerstore.PermanentAddrTTL)
 			d.handler.AddPeer(0, peer.ID)
-
-			if !timer.Stop() {
-				<-timer.C
-			}
```
qdm12, line 223: removed like in #3001

EclesioMeloJunior (approved):

timwu20 (approved):

PR #3005, opened 2022-12-13
feat(internal/database): badger v3 and memory implementations

qdm12:
## Changes

:information_source: This is only the standalone database implementation. I need the in-memory implementation for #2831 which in turn unblocks #2981 which will unblock the full replacement of chaindb. This doesn't affect the existing codebase for now.

- [x] `internal/database` shared (return) interfaces and sentinel errors
- [x] `internal/database/memory`
  - [x] implementation
  - [x] integration test
  - [x] full unit test coverage
  - [x] thread safety test
- [x] `internal/database/badger` - based on badger/v3
  - [x] implementation
  - [x] integration test
  - [x] full unit test coverage
  - [x] thread safety test
- [x] Add race tests to CI

## Tests

```sh
go test github.com/ChainSafe/gossamer/internal/database/...
```

## Issues

Related #1973 / will create an issue for replacing the database implementation

## Primary Reviewer

@timwu20

### DIFF
--- a/internal/database/badger/settings_test.go
+++ b/internal/database/badger/settings_test.go
@@ -16,13 +16,13 @@ func Test_Settings_SetDefaults(t *testing.T) {
 		originalSettings Settings
 		expectedSettings Settings
 	}{
-		"empty settings": {
+		"empty_settings": {
 			expectedSettings: Settings{
 				Path:     ptrTo(""),
 				InMemory: ptrTo(false),
 			},
 		},
-		"non-empty settings": {
+		"non-empty_settings": {
 			originalSettings: Settings{
 				Path:     ptrTo("x"),
 				InMemory: ptrTo(true),
@@ -54,7 +54,7 @@ func Test_Settings_Validate(t *testing.T) {
 		errWrapped error
 		errMessage string
 	}{
-		"path set in-memory": {
+		"path_set_in-memory": {
 			settings: Settings{
 				Path:     ptrTo("."),
 				InMemory: ptrTo(true),
@@ -64,7 +64,7 @@ func Test_Settings_Validate(t *testing.T) {
 		},
 		// Note we cannot test for a bad path since we would
 		// need os.Getcwd() to fail.
-		"valid settings": {
+		"valid_settings": {
 			settings: Settings{
 				Path:     ptrTo("."),
 				InMemory: ptrTo(false),
--- a/internal/database/memory/database_test.go
+++ b/internal/database/memory/database_test.go
@@ -28,14 +28,14 @@ func Test_Database_Get(t *testing.T) {
 		errWrapped error
 		errMessage string
 	}{
-		"database closed": {
+		"database_closed": {
 			db: &Database{
 				closed: atomicTrue(),
 			},
 			errWrapped: database.ErrClosed,
 			errMessage: "database closed",
 		},
-		"key not found": {
+		"key_not_found": {
 			db: &Database{
 				closed:    new(atomic.Bool),
 				keyValues: map[string][]byte{},
@@ -44,7 +44,7 @@ func Test_Database_Get(t *testing.T) {
 			errWrapped: database.ErrKeyNotFound,
 			errMessage: "key not found: 0x01",
 		},
-		"key found": {
+		"key_found": {
 			db: &Database{
 				closed: new(atomic.Bool),
 				keyValues: map[string][]byte{
@@ -83,7 +83,7 @@ func Test_Database_Set(t *testing.T) {
 		errMessage string
 		expectedDB *Database
 	}{
-		"database is closed": {
+		"database_is_closed": {
 			db: &Database{
 				closed: atomicTrue(),
 			},
@@ -93,7 +93,7 @@ func Test_Database_Set(t *testing.T) {
 				closed: atomicTrue(),
 			},
 		},
-		"set at new key": {
+		"set_at_new_key": {
 			db: &Database{
 				closed:    new(atomic.Bool),
 				keyValues: map[string][]byte{},
@@ -107,7 +107,7 @@ func Test_Database_Set(t *testing.T) {
 				},
 			},
 		},
-		"override value at key": {
+		"override_value_at_key": {
 			db: &Database{
 				closed: new(atomic.Bool),
 				keyValues: map[string][]byte{
@@ -168,7 +168,7 @@ func Test_Database_Delete(t *testing.T) {
 		errWrapped error
 		expectedDB *Database
 	}{
-		"database closed": {
+		"database_closed": {
 			db: &Database{
 				closed: atomicTrue(),
 			},
@@ -177,7 +177,7 @@ func Test_Database_Delete(t *testing.T) {
 				closed: atomicTrue(),
 			},
 		},
-		"key not found": {
+		"key_not_found": {
 			db: &Database{
 				closed:    new(atomic.Bool),
 				keyValues: map[string][]byte{},
@@ -188,7 +188,7 @@ func Test_Database_Delete(t *testing.T) {
 				keyValues: map[string][]byte{},
 			},
 		},
-		"key deleted": {
+		"key_deleted": {
 			db: &Database{
 				closed: new(atomic.Bool),
 				keyValues: map[string][]byte{


internal/database/badger/writebatch.go
```diff
@@ -0,0 +1,40 @@
...
+	return &writeBatch{
+		prefix:           prefix,
+		badgerWriteBatch: badgerWriteBatch,
+	}
+}
+
+// Set sets a value at the given key prefixed with the given prefix.
+func (wb *writeBatch) Set(key, value []byte) (err error) {
+	key = append(wb.prefix, key...)
+	return wb.badgerWriteBatch.Set(key, value)
```
qdm12, line 22: here we use the badger db writer batch instead of writing our own (chaindb one is faulty on top of that)

internal/database/memory/database.go
```diff
@@ -0,0 +1,96 @@
...
+
+import (
+	"fmt"
+	"sync"
+
+	"github.com/ChainSafe/gossamer/internal/database"
+)
+
+// Database is an in-memory database implementation.
+type Database struct {
```
qdm12, line 12: The point of the in-memory database implementation package is to force the internal/database code to be flexible for additional implementations.

On top of that, I would suggest us to use this one instead of the in-memory badger database, so the badger database is only on disk.

internal/database/memory/writebatch.go
```diff
@@ -0,0 +1,76 @@
...
+}
+
+// writeBatch implements an in-memory write batch and prefixes
+// all keys with a certain given prefix.
+// It is NOT thread safe, but its Flush operation is thread safe for
+// the database injected.
+type writeBatch struct {
+	prefix     string
+	database   *Database
+	operations []operation
```
qdm12, line 23: Note this fixes the current issue in the chaindb batch, where the order of operation is kept.

internal/database/badger/database.go
```diff
@@ -0,0 +1,109 @@
...
+		}
+
+		return nil
+	})
+
+	if errors.Is(err, badger.ErrKeyNotFound) {
+		return nil, fmt.Errorf("%w: 0x%x", database.ErrKeyNotFound, key)
+	}
+
+	return value, err
```
EclesioMeloJunior, line 65: Could be simplified to wrap any error that happens
```suggestion
        if err != nil {
          fmt.Errorf("getting key 0x%x: %w", key, err)
        }
	return value, err
```

internal/database/badger/database.go
```diff
@@ -0,0 +1,109 @@
...
+}
+
+// Set sets a value at the given key in the database.
+func (db *Database) Set(key, value []byte) (err error) {
+	return db.badgerDatabase.Update(func(txn *badger.Txn) error {
+		err = txn.Set(key, value)
+		return err
```
EclesioMeloJunior, line 72:
```suggestion
		return txn.Set(key, value)
```

internal/database/badger/settings.go
```diff
@@ -0,0 +1,33 @@
...
+import (
+	"fmt"
+	"path/filepath"
+)
+
+// Settings is the database settings.
+type Settings struct {
+	// Path is the database directory path to use.
+	// It defaults to the current directory if left unset.
+	Path *string
```
EclesioMeloJunior, line 15: Are there any problems with using `Path: string` with the default value being `""`?

EclesioMeloJunior (changes requested):

internal/database/badger/settings.go
qdm12: Well `""` is a valid path (current directory) and is also the default.
I use a string pointer so we can know when the field is left unset.
I need to write a blog post sometime to explain the whole reasoning behind this, but this should explain it more or less: https://github.com/qdm12/gosettings#settings-struct

> Each component has a settings struct, where the zero value of a field should be INVALID. For example, if the value 0 is allowed for a field, then it must be an *int field. On the contrary, you could have an int field if the zero value 0 is not valid. The reasoning behind this is that you want the zero Go value to be considered as 'unset field' so that the field value can be defaulted, merged with or overridden by another settings struct.

qdm12: ... never mind, `""` is not valid and crashes the badger open function, so changing this to a non-pointer string!

internal/database/badger/helpers.go
```diff
@@ -0,0 +1,16 @@
+// Copyright 2022 ChainSafe Systems (ON)
+// SPDX-License-Identifier: LGPL-3.0-only
+
+package badger
+
+func makePrefixedKey(prefix, key []byte) (prefixedKey []byte) {
```
qdm12, line 6: This function fixes a kinda critical problem that exists for tables in chaindb

kishansagathiya (commented):
Looks good overall. 

Yet to look at test files.

internal/database/memory/database.go
qdm12: It could if we add a config field for the database implementation later.
Not that useful I have to admit, but I would say it serves as a good example for another database implementation (as in it's a 'dumb' base just using Go std types).

internal/database/memory/writebatch.go
qdm12: > Are you saying that currently chaindb does not keep order of operations?

Yes 😄 

https://github.com/ChainSafe/chaindb/blob/f900218c88f853b64138c628c2761383f4180dd3/database.go#L89-L90

> Wouldn't that mess up things while executing them?

Yes 😄

internal/database/badger/settings.go
```diff
@@ -0,0 +1,39 @@
...
+	"path/filepath"
+)
+
+// Settings is the database settings.
+type Settings struct {
+	// Path is the database directory path to use.
+	// It defaults to the current directory if left unset.
+	Path string
+	// InMemory is whether to use an in-memory database.
+	InMemory *bool
```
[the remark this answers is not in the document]
qdm12: Actually the point of this is to detect when it's left unset, since `false` is not 'unset'. That way we can set defaults as we want with a method.

More details on https://github.com/qdm12/gosettings#settings-struct I have been using this for about a year, and it has been quite scalable and nice to use.

internal/database/memory/database.go
```diff
@@ -0,0 +1,99 @@
...
+
+// NewTable returns a new table using the database.
+// All keys on the table will be prefixed with the given prefix.
+func (db *Database) NewTable(prefix string) (writeBatch database.Table) {
+	db.panicOnClosed()
+	return newTable(prefix, db)
+}
+
+// Close closes the database.
+func (db *Database) Close() (err error) {
```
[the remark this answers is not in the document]
qdm12: > Is this getting used anywhere other than tests?

Both implementations are not used anywhere in this PR. The badger implementation will be used in production code and in tests in the next PR. The in memory implementation could be used in tests, but really we should test using the in-memory badger implementation. So at the end of the day, that memory implementation is really just a base to serve for another implementation (zero dependency, just Go code with what should be expected in terms of implementation)

internal/database/memory/table.go
```diff
@@ -0,0 +1,49 @@
+// Copyright 2022 ChainSafe Systems (ON)
+// SPDX-License-Identifier: LGPL-3.0-only
+
+package memory
+
+import (
```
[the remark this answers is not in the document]
qdm12: They are similar for all methods but `NewWriteBatch()`. The problem is the writeBatch implementation for both packages is quite different, so I don't think this can be abstracted away really. Unless we would use our own in-memory write batch for both database implementations, but I would rather let it flexible + use the badger write batch. Feel free to give it a stab in a draft PR if you want, but I couldn't make it work

internal/database/badger/race_test.go
```diff
@@ -0,0 +1,78 @@
...
+
+	"github.com/stretchr/testify/require"
+)
+
+func Test_Database_threadSafety(t *testing.T) {
+	// This test consists in checking for concurrent access
+	// using the -race detector.
+	t.Parallel()
+
+	var startWg, endWg sync.WaitGroup
```
kishansagathiya, line 20: I think we should be able to use just one wait group. Let me know what I might be missing.

internal/database/badger/settings.go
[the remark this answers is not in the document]
kishansagathiya: why do  we need to check if it is set or unset?

kishansagathiya (changes requested):
Considering we have two database implementations, we should have the same test testing for both the implementations.

internal/database/badger/race_test.go
qdm12: We want to wait for all goroutines (aka workers) to be ready to make sure they all operate at the same time (since starting a goroutine takes a few milliseconds) using `startWg`, and then we want to wait for all of them to exit (`endWg`)

qdm12:
> It not nice that none of these functions are not getting used in this PR. Ideally, it will be nice if we have a db interface that is getting used. Then, In the same PR we can implement a db and plug it will rest of the codebase to use it to test if things are working fine. Let me know if that can't be done at the moment.

Yeah I get your point; using this newer interface/packages is blocked by #2831 which I'm working on now. Once this is merged, I'll rebase this branch on it and make a separate PR based on this one, which we can merge into this PR. Trying to keep this PR and the future interface-replacement PR separate to the amount of deltas in each (for reviewers), but we should merge it as a single commit in development indeed 👍

internal/database/badger/settings.go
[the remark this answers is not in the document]
qdm12: To be able to set defaults if it's left unset (and it extends beyond eventually, such as merging/overriding settings structs)

internal/database/badger/database_test.go
```diff
@@ -0,0 +1,356 @@
...
+import (
+	"testing"
+
+	"github.com/ChainSafe/gossamer/internal/database"
+	"github.com/dgraph-io/badger/v3"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func Test_Database(t *testing.T) {
```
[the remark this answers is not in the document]
qdm12: I really don't see the need for comments in this test.
Method names are quite explicit and it's tiny/meaningless values that any programmer can understand.
Comments would bloat more the test code without bringing more than what one can understand from just reading the code.

For example

```go
		err := db.Set([]byte{1}, []byte{2})
		require.NoError(t, err)
```

versus

```go
		// Set key []byte{1} with value []byte{2}
		err := db.Set([]byte{1}, []byte{2})
		require.NoError(t, err)
```

the comment is rather redundant/unneeded.

internal/database/badger/database.go
```diff
@@ -0,0 +1,112 @@
...
+	"github.com/dgraph-io/badger/v3"
+)
+
+// Database is database implementation using a badger/v3 database.
+type Database struct {
+	badgerDatabase *badger.DB
+}
+
+// New returns a new database based on a badger v3 database.
+func New(settings Settings) (database *Database, err error) {
```
timwu20, line 21: nit:
```suggestion
func NewDatabase(settings Settings) (database *Database, err error) {
```

internal/database/badger/database.go
```diff
@@ -0,0 +1,112 @@
...
+
+// Package badger provides a database implementation using badger v3.
+package badger
+
+import (
+	"errors"
+	"fmt"
+
+	"github.com/ChainSafe/gossamer/internal/database"
+	"github.com/dgraph-io/badger/v3"
```
timwu20, line 12: nit: it's not obvious the package name in this folder is still `badger`
```suggestion
	badger "github.com/dgraph-io/badger/v3"
```

internal/database/badger/database.go
```diff
@@ -0,0 +1,112 @@
...
+)
+
+// Database is database implementation using a badger/v3 database.
+type Database struct {
+	badgerDatabase *badger.DB
```
timwu20, line 17: nit: embed?
```suggestion
	*badger.DB
```

internal/database/badger/helpers.go
```diff
@@ -0,0 +1,33 @@
...
+
+import (
+	"errors"
+	"fmt"
+
+	"github.com/ChainSafe/gossamer/internal/database"
+	"github.com/dgraph-io/badger/v3"
+)
+
+func makePrefixedKey(prefix, key []byte) (prefixedKey []byte) {
```
timwu20, line 14:
```suggestion
func newPrefixedKey(prefix, key []byte) (prefixedKey []byte) {
```

internal/database/badger/settings.go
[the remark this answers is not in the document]
timwu20: I don't see why this is necessary.  IMO `InMemory` should default to false.

internal/database/badger/writebatch.go
```diff
@@ -0,0 +1,46 @@
+// Copyright 2022 ChainSafe Systems (ON)
+// SPDX-License-Identifier: LGPL-3.0-only
+
+package badger
+
+import "github.com/dgraph-io/badger/v3"
+
+// writeBatch uses the badger write batch and prefixes
+// all keys with a certain given prefix.
+type writeBatch struct {
```
timwu20, line 10: what uses this?

internal/database/memory/database.go
timwu20: if tests can use the badger in memory version, do we really need this in memory implementation?

internal/database/badger/settings.go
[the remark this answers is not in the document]
kishansagathiya: > To be able to set defaults if it's left unset (and it extends beyond eventually, such as merging/overriding settings structs).

I don't think I understand this completely.

>  it extends beyond eventually, such as merging/overriding settings structs

May be elaborate this part. Do we need to merge or override this setting at some point?

What I am really not able to understand is that why can't we treat false as unset. 

Say someone assigned a false value to inMemory, why do we care if it was set or it is there because it is default value?

internal/database/badger/database.go
qdm12: I think it's fine as `New` since the package is only dedicated to the badger database (and also the signature can be checked). A bit like `log.New` instead of `log.NewLogger`

internal/database/badger/database.go
qdm12: I prefer to avoid embedding if it's not needed because:

- it reduces method and field name conflicts
- it exports `DB` as a field accessible outside (big no no in that case)
- it's more cumbersome to assert against

internal/database/badger/settings.go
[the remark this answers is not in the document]
qdm12: > What I am really not able to understand is that why can't we treat false as unset.

Because `false` or `true` are set values. The user can specify `false`.
Say we want to read from two configuration sources such as flags and toml, and we want flags to take precedence.
The flag specifies `--inmemory=false` and the toml config specifies `inmemory = true`. If we consider `false` as unset, then we would end up with a final value of `true` whereas it should be `false` due to precedence.
Same applies with merging/overriding settings.

It's perhaps out of scope and not used yet, but it's good code writing imo and there is no point neglecting this.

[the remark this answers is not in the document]
kishansagathiya: > If we consider false as unset, then we would end up with a final value of true whereas it should be false due to precedence

If I want flag to take precedence, I would just take flag value and assign it to inMemory. It doesn't matter if it is set or not. So, if the flag says false, the value of inMemory will become false.

The way we deal with such setting is, 
- initialise with default settings with, say `Default()`.
- Override with say toml config (we don't need to know existing values in settings here)
- Override with say flag (we don't need existing values here as well)

I don't think we need inMemory to be pointer here.

internal/database/badger/helpers.go
qdm12: Done

internal/database/badger/database.go
qdm12: Done

internal/database/badger/writebatch.go
qdm12: The database `NewWriteBatch()` method and the database's table method `NewWriteBatch()`

internal/database/memory/database.go
qdm12: I would like to keep it as a base for another future implementation. For example `badger` uses the badger/v3 batch implementation, but some other implementations might need to implement their own batching, and it's already there in the (zero-dependency) memory implementation. I can create an issue to remove it later on once we have at least another implementation?

timwu20:
I think we should wait to merge this into `development`.  We should probably do some testing to ensure things are working as expected before merging in.  Can you create a release branch that will contain this PR and the PR that utilises this code?

internal/database/badger/settings.go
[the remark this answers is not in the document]
timwu20: I don't see why we would want to support detecting an unset value for this attribute.  I see no reason why the default value for `InMemory` should not be `false`.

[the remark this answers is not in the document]
qdm12: Read https://github.com/ChainSafe/gossamer/pull/3005#discussion_r1071986475 again. Empty value should be invalid (and for a boolean, false and true are both 'set values').

> If I want flag to take precedence, I would just take flag value and assign it to inMemory. It doesn't matter if it is set or not. So, if the flag says false, the value of inMemory will become false.

Say the flag is `--inmemory=false`, then your field is set to false. Then your next source (i.e. toml) sets `inmemory = true`. How does it know if it should override this `false` with its `true`?? You can have ugly logic for every field to do this, but it's much more modular to just use a boolean pointer and if it's nil, then just set it to the value read from the settings source.

internal/database/memory/database.go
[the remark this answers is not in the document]
qdm12: It's fully replaced in #3088 now 😉 (based on this PR)

Issue #2974, opened 2022-11-24
Gossamer does not answer descending block requests


EclesioMeloJunior:
## Describe the bug

- I am using `westend-local` with `bob` as the unique block producer and I start gossamer as `bob` after the gossamer node produced around 30 blocks I started substrate node as `alice` and as usual substrate executes a block request like the bellow example:

```
DEBUG tokio-runtime-worker sync: Connected 12D3KooWHw4UXtfTnxDpLRYqP1svSxTikrMrYSG1UwUg3sFc4LCx    
TRACE tokio-runtime-worker sync: New block request for 12D3KooWHw4UXtfTnxDpLRYqP1svSxTikrMrYSG1UwUg3sFc4LCx, (best:41, common:0) BlockRequest { id: 0, fields: HEADER | BODY | JUSTIFICATION, from: Hash(0xdf56ae17a070025c7c63b5d81dbbd34deb2f9fded7dfe0465482317b51df9902), direction: Descending, max: Some(41) }    
```

- Gossamer receives this message as printed in the bellow logs:

```
TRACE    host 12D3KooWHw4UXtfTnxDpLRYqP1svSxTikrMrYSG1UwUg3sFc4LCx received message from peer 12D3KooWFRAiaCpbSXPiSJ8isS9xyHkn4dXWTETSPZwmJ99U6RKW: BlockRequestMessage RequestedData=19 StartingBlock={[223 86 174 23 160 112 2 92 124 99 181 216 29 187 211 77 235 47 159 222 215 223 224 70 84 130 49 123 81 223 153 2]} EndBlockHash=0x0000000000000000000000000000000000000000000000000000000000000000 Direction=1 Max=41	inbound.go:L43	pkg=network
DEBUG    handling BlockRequestMessage with direction descending from start block with hash 0xdf56ae17a070025c7c63b5d81dbbd34deb2f9fded7dfe0465482317b51df9902 to end block with hash 0x3e54455287ee4bfbfcb0576d7747505bc8981ec602a01b3ea0a067134fb6f72b	message.go:L212	pkg=sync
DEBUG    cannot create response for request: start node does not exist	sync.go:L117	pkg=network
```

- The point is that the hashes/blocks requested by substrate exists in the Gossamer node, but for some reason the Gossamer node cannot retrieve them

[EDIT]

- Currently, `blockState.SubChain` only retrieve blocks from memory that is a limitation when we try to search for a subchain that is stored on disk.
[...]

PR #3006, opened 2022-12-15
fix(dot/sync): use `Range` instead of `SubChain` at `handleDescedingRequest`

EclesioMeloJunior:
## Changes

- Replace `SubChain` (only look-up in the in-memory blocktree) for `Range` (look-up in database and in disk if needed) to get a range

## Tests

```sh
```

## Issues

- Ref https://github.com/ChainSafe/gossamer/issues/2974

## Primary Reviewer

@timwu20


kishansagathiya (commented):
Looks good.

There are some tests that needs to be fixed though. I think It is mostly failing for not re-generating mocks everywhere.

Also, do we still need `SubChain()` func? It is used at two places. Does it makes sense to replace it with `Range` everywhere?

EclesioMeloJunior:
> Looks good.
> 
> There are some tests that needs to be fixed though. I think It is mostly failing for not re-generating mocks everywhere.
> 
> Also, do we still need `SubChain()` func? It is used at two places. Does it makes sense to replace it with `Range` everywhere?

I decided to keep `SubChain` since it retrieves blocks only in the blocktree (in-memory) and I've created `Range` to retrieve blocks from the database and in-memory blocktree. So the `SubChain` is used in other places that only require lookups in the in-memory blocktree that is why it is important yet

qdm12 (approved):

qdm12:
> So the SubChain is used in other places that only require lookups in the in-memory blocktree that is why it is important yet

Should we perhaps then rename `SubChain` to `RangeInMemory`?

EclesioMeloJunior:
> > So the SubChain is used in other places that only require lookups in the in-memory blocktree that is why it is important yet
> 
> Should we perhaps then rename `SubChain` to `RangeInMemory`?

Done here: https://github.com/ChainSafe/gossamer/pull/3011

dot/sync/message_test.go
```diff
@@ -27,7 +27,7 @@ func TestService_CreateBlockResponse(t *testing.T) {
 		want              *network.BlockResponseMessage
 		err               error
 	}{
-		"invalid block request": {
+		"invalid_block_request": {
```
kishansagathiya, line 30: why change test case names like this?

kishansagathiya (changes requested):
LGTM, 

I am not sure why the test case names have changed though. Is that some new convention that seems better than what we have already?

dot/sync/message_test.go
qdm12: > Is that some new convention that seems better than what we have already?

Yeah it is, since logs written by `go test` replaces spaces with underscores. So it's easier to copy paste and search in the codebase. We should use underscores instead of spaces from now on (feel free to harass me on that, I'm having trouble remembering this 😄)

EclesioMeloJunior: Yeah, if you have the following test case name: `invalid block request` then the go test output will be `invalid_block_request`, so will be more convenient to introduce the underscore by default while naming a test case.

kishansagathiya (approved):

PR #3007, opened 2022-12-16
feat(lib/runtime/wasmer): report grandpa equivocations

jimjbrettj:
## Changes

- Create new runtime function calls for reporting equivocations in grandpa and getting a key ownership proof
- Write Unit tests for both these runtime functions
- Create function reportEquivocation in lib/grandpa that, when a user equivocates, uses grandpa and block state to create the appropriate data types to get the keyOwnershipProof from the runtime and submit the equivocationProof to the runtime
- Write unit test for reportEquivocation function

Notes:
- Used existing equivocation integration test in grandpa to verify there are no errors in my reportEquivocation function

## Tests

```sh
go test ./lib/runtime/wasmer
go test ./dot/types
go test ./lib/grandpa
```

## Issues

 #2933

## Primary Reviewer

@kishansagathiya


lib/grandpa/vote_message_test.go
```diff
@@ -0,0 +1,186 @@
...
+		stage        Subround
+		existingVote *SignedVote
+		currentVote  *SignedVote
+	}
+	tests := []struct {
+		name      string
+		service   *Service
+		args      args
+		expErr    error
+		expErrMsg string
```
qdm12, line 113: define these mocks for each test case using a functional field in the test struct:

```go
serviceBuilder func(ctrl *gomock.Controller) *Service
```

And configure the mocks for each test case.

The test loop body can then do

```go
service := tt.serviceBuilder(ctrl)
```

That way a reader doesn't have to jump up and down, everything is in the test case definition block (and also controller isn't shared across parallel subtests)

lib/runtime/interfaces.go
```diff
@@ -4,11 +4,55 @@
...
+	Keystore() *keystore.GlobalKeystore
+	Validator() bool
+	Exec(function string, data []byte) ([]byte, error)
+	SetContextStorage(s Storage) // used to set the TrieState before a runtime call
+
+	GetCodeHash() common.Hash
+	// Version returns the version from the runtime.
+	// This should return the cached version and be cheap to execute.
+	Version() (version Version)
+	Metadata() ([]byte, error)
```
qdm12, line 32:
```suggestion
	Versioner
	Metadataer
```

lib/runtime/interfaces.go
```diff
@@ -4,11 +4,55 @@
...
+	DecodeSessionKeys(enc []byte) ([]byte, error)
+	PaymentQueryInfo(ext []byte) (*types.RuntimeDispatchInfo, error)
+	CheckInherents() // TODO: use this in block verification process (#1873)
+	GrandpaGenerateKeyOwnershipProof(authSetID uint64, authorityID ed25519.PublicKeyBytes) (
+		types.OpaqueKeyOwnershipProof, error)
+	GrandpaSubmitReportEquivocationUnsignedExtrinsic(
+		equivocationProof types.GrandpaEquivocationProof, keyOwnershipProof types.OpaqueKeyOwnershipProof,
+	) error
+
+	// parameters and return values for these are undefined in the spec
```
qdm12, line 50: let's remove comments in interfaces, better have comments on implementation methods, otherwise we should duplicate comments on every interface definitions (narrower and wider interfaces of the same impl)

```suggestion
```

lib/runtime/interfaces.go
```diff
@@ -4,11 +4,55 @@
...
+	GrandpaAuthorities() ([]types.Authority, error)
+	ValidateTransaction(e types.Extrinsic) (*transaction.Validity, error)
+	InitializeBlock(header *types.Header) error
+	InherentExtrinsics(data []byte) ([]byte, error)
+	ApplyExtrinsic(data types.Extrinsic) ([]byte, error)
+	FinalizeBlock() (*types.Header, error)
+	ExecuteBlock(block *types.Block) ([]byte, error)
+	DecodeSessionKeys(enc []byte) ([]byte, error)
+	PaymentQueryInfo(ext []byte) (*types.RuntimeDispatchInfo, error)
+	CheckInherents() // TODO: use this in block verification process (#1873)
```
qdm12, line 43: is this a new `TODO`? If not let's remove it, it should be on the `CheckInherents` implementation method

```suggestion
	CheckInherents()
```

lib/runtime/interfaces.go
```diff
@@ -4,11 +4,55 @@
...
+	"github.com/ChainSafe/gossamer/lib/crypto/ed25519"
+	"github.com/ChainSafe/gossamer/lib/keystore"
 	"github.com/ChainSafe/gossamer/lib/transaction"
 	"github.com/ChainSafe/gossamer/lib/trie"
 )
 
+//go:generate mockery --name Instance --structname Instance --case underscore --keeptree
+
+// Instance is the interface a v0.8 runtime instance must implement
+type Instance interface {
```
qdm12, line 18: Perhaps keep this `test_helpers.go` for the sake of lower deltas / git blame?
I'm planning on doing a PR sometime to group all interfaces in `interfaces.go` later

lib/runtime/wasmer/exports.go
```diff
@@ -227,6 +228,58 @@ func (in *Instance) QueryCallFeeDetails(ext []byte) (*types.FeeDetails, error) {
...
+	}
+	combinedArg = append(combinedArg, encodedSetID...)
+
+	encodedAuthorityID, err := scale.Marshal(authorityID)
+	if err != nil {
+		return nil, fmt.Errorf("encoding authority id: %w", err)
+	}
+	combinedArg = append(combinedArg, encodedAuthorityID...)
+
+	ret, err := in.Exec(runtime.GrandpaGenerateKeyOwnershipProof, combinedArg)
```
qdm12, line 248: nit use scale.NewEncoder + pre-determined buffer capacity ⚡  So we write to the same underlying memory array, without having to resize/copy

```suggestion
	const bufferSize = 8 + 32 // authSetID uint64 + ed25519.PublicKeyBytes
	buffer := bytes.NewBuffer(make([]byte, 0, bufferSize))
	encoder := scale.NewEncoder(buffer)
	err := encoder.Encode(authSetID)
	if err != nil {
		return nil, fmt.Errorf("encoding auth set id: %w", err)
	}
	err = encoder.Encode(authorityID)
	if err != nil {
		return nil, fmt.Errorf("encoding authority id: %w", err)
	}

	ret, err := in.Exec(runtime.GrandpaGenerateKeyOwnershipProof, buffer.Bytes())
```

lib/runtime/wasmer/exports.go
```diff
@@ -227,6 +228,58 @@ func (in *Instance) QueryCallFeeDetails(ext []byte) (*types.FeeDetails, error) {
...
+	}
+	combinedArg = append(combinedArg, encodedEquivocationProof...)
+
+	encodedKeyOwnershipProof, err := scale.Marshal(keyOwnershipProof)
+	if err != nil {
+		return fmt.Errorf("encoding key ownership proof: %w", err)
+	}
+	combinedArg = append(combinedArg, encodedKeyOwnershipProof...)
+
+	_, err = in.Exec(runtime.GrandpaSubmitReportEquivocation, combinedArg)
```
qdm12, line 279: nit

```suggestion
	buffer := bytes.NewBuffer(nil)
	encoder := scale.NewEncoder(buffer)
	err := encoder.Encode(equivocationProof)
	if err != nil {
		return fmt.Errorf("encoding equivocation proof: %w", err)
	}
	err = encoder.Encode(keyOwnershipProof)
	if err != nil {
		return fmt.Errorf("encoding key ownership proof: %w", err)
	}
	_, err = in.Exec(runtime.GrandpaSubmitReportEquivocation, buffer.Bytes())
```

lib/runtime/wasmer/exports_test.go
```diff
@@ -1210,3 +1210,126 @@ func TestInstance_TransactionPaymentCallApi_QueryCallFeeDetails(t *testing.T) {
...
+		Hash:   common.MustHexToHash("0xc314327941fdd924bc67fd72651c40aececd485ca3e878c21e02abb40feae5bd"),
+		Number: uint32(10),
+	}
+
+	firstSignature, _ := common.HexToBytes("0xd7292caacc62504365f179892a7399f233944bf261f8a3f66260f70e0016f2d" +
+		"b63922726b015c82dc7131f4730fbec61f71672a571453e51029bfb469070900f")
+	var firstSignatureArray [64]byte
+	copy(firstSignatureArray[:], firstSignature)
+
+	secondSignature, _ := common.HexToBytes("0xb3c408b74905dfedfffa66f99f16fe8b938fd8df76a92225228a1ca07523" +
```
qdm12, line 1248: nit

```suggestion
	secondSignature := common.MustHexToBytes("0xb3c408b74905dfedfffa66f99f16fe8b938fd8df76a92225228a1ca07523" +
```

lib/runtime/wasmer/exports.go
```diff
@@ -227,6 +228,58 @@ func (in *Instance) QueryCallFeeDetails(ext []byte) (*types.FeeDetails, error) {
 // TODO: use this in block verification process (#1873)
 func (in *Instance) CheckInherents() {}
 
+// GrandpaGenerateKeyOwnershipProof returns grandpa key ownership proof from runtime.
+func (in *Instance) GrandpaGenerateKeyOwnershipProof(authSetID uint64, authorityID ed25519.PublicKeyBytes) (
```
qdm12, line 232: This function is not test covered, can you add a test for it?

lib/runtime/interfaces.go
jimjbrettj: Oh okay I was thinking good to incrementally move to interfaces.go, but if you're gonna batch do it I can revert and leave as it was

lib/grandpa/vote_message.go
```diff
@@ -248,12 +250,81 @@ func (s *Service) checkForEquivocation(voter *Voter, vote *SignedVote, stage Sub
...
+		if err != nil {
+			return fmt.Errorf("setting grandpa equivocation VDT: %w", err)
+		}
+	case precommit:
+		err = equivocationVote.Set(types.PreCommitEquivocation(grandpaEquivocation))
+		if err != nil {
+			return fmt.Errorf("setting grandpa equivocation VDT: %w", err)
+		}
+	default:
+		return fmt.Errorf("invalid stage for equivocating: %w", err)
```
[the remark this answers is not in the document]
jimjbrettj: Good call for using sentinel error, will update. No, there is a third value of Subround where stage = 2 that represents primaryProposal. However, I do not believe you can equivocate by voting in this subround (and the enum in substrate to create this proof does not include it) so I added the default case to account for it.

As Im typing this I'm thinking perhaps I can add a specific case for this and return an error there, and then the default can be panic? @qdm12

lib/grandpa/vote_message_test.go
jimjbrettj: Really like this!

lib/runtime/interfaces.go
jimjbrettj: No it was there and in both places, removing from here

lib/runtime/wasmer/exports.go
jimjbrettj: Since the types are defined as the function arguments I don't know what I can pass in to make it fail, so I wrote a simple test for the happy path. However if there is a case I didn't think of please let me know and I can add it.

lib/grandpa/vote_message.go
```diff
@@ -216,7 +217,7 @@ func (s *Service) validateVoteMessage(from peer.ID, m *VoteMessage) (*Vote, erro
 // checkForEquivocation checks if the vote is an equivocatory vote.
 // it returns true if so, false otherwise.
 // additionally, if the vote is equivocatory, it updates the service's votes and equivocations.
```
timwu20, line 219: update this comment if you're going to change the signature.

lib/grandpa/vote_message.go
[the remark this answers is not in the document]
timwu20: > As Im typing this I'm thinking perhaps I can add a specific case for this and return an error there, and then the default can be panic? @qdm12

this sounds good.

timwu20 (commented):
Great work Jimmy! Just a few small things remanining.

lib/runtime/wasmer/exports.go
qdm12: You cannot fail the encode, and probably not the decode (unless the instance is mocked etc. which is, well for another day), but would it be possible to make Exec fail with bad authSetID+authorityID values? If not possible, the happy path only is fine.

lib/grandpa/vote_message_test.go
qdm12: Credits to Tim a few months back 🎖️

lib/runtime/interfaces.go
qdm12: Yeah that's the thing. I agree with incremental changes... but on the other hand, it messes up the git blame (like wtf did he move this from a file to another, it's OOS)... And also focusing on doing 1 cleanup in the entire codebase is much more efficient (example change mockery to mockgen), especially given the current tight-ish coupling between packages.

dot/types/grandpa.go
```diff
@@ -193,3 +193,62 @@ type GrandpaVote struct {
 func (v *GrandpaVote) String() string {
 	return fmt.Sprintf("hash=%s number=%d", v.Hash, v.Number)
 }
+
+// GrandpaEquivocation is used to create a proof of equivocation
+type GrandpaEquivocation struct {
```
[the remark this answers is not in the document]
qdm12: Do you think we can already have in `lib/grandpa`? Or could we kick off a `lib/gandpa/types` package with that struct?

lib/runtime/wasmer/exports_test.go
```diff
@@ -1210,3 +1210,138 @@ func TestInstance_TransactionPaymentCallApi_QueryCallFeeDetails(t *testing.T) {
...
+	encoded := scale.MustMarshal(keyOwnershipProof)
+	var proof types.OpaqueKeyOwnershipProof
+	err := scale.Unmarshal(encoded, &proof)
+	require.NoError(t, err)
+	require.Equal(t, keyOwnershipProof, proof)
+}
+
+func TestInstance_GrandpaGenerateKeyOwnershipProof(t *testing.T) {
+	t.Parallel()
+	rt := NewTestInstance(t, runtime.WESTEND_RUNTIME_v0929)
```
qdm12, line 1227: nit `rt` -> `runtime`

lib/runtime/wasmer/exports_test.go
```diff
@@ -1210,3 +1210,138 @@ func TestInstance_TransactionPaymentCallApi_QueryCallFeeDetails(t *testing.T) {
...
+	identity := common.MustHexToBytes("0x88dc3417d5058ec4b4503e0c12ea1a0a89be200fe98922423d4334014fa6b0ee")
+	identityPubKey, _ := ed25519.NewPublicKey(identity)
+	var authorityID [32]byte
+	copy(authorityID[:], identityPubKey.Encode())
+
+	_, err := rt.GrandpaGenerateKeyOwnershipProof(uint64(0), authorityID)
```
qdm12, line 1233: Can we check the returned value as well?

lib/runtime/wasmer/exports.go
```diff
@@ -227,6 +229,53 @@ func (in *Instance) QueryCallFeeDetails(ext []byte) (*types.FeeDetails, error) {
...
+	encoder := scale.NewEncoder(buffer)
+	err := encoder.Encode(authSetID)
+	if err != nil {
+		return nil, fmt.Errorf("encoding auth set id: %w", err)
+	}
+	err = encoder.Encode(authorityID)
+	if err != nil {
+		return nil, fmt.Errorf("encoding authority id: %w", err)
+	}
+	ret, err := in.Exec(runtime.GrandpaGenerateKeyOwnershipProof, buffer.Bytes())
```
qdm12, line 247: nit `ret` -> `encodedOpaqueKeyOwnershipProof`

lib/runtime/wasmer/exports.go
```diff
@@ -227,6 +229,53 @@ func (in *Instance) QueryCallFeeDetails(ext []byte) (*types.FeeDetails, error) {
...
+	err := encoder.Encode(equivocationProof)
+	if err != nil {
+		return fmt.Errorf("encoding equivocation proof: %w", err)
+	}
+	err = encoder.Encode(keyOwnershipProof)
+	if err != nil {
+		return fmt.Errorf("encoding key ownership proof: %w", err)
+	}
+	_, err = in.Exec(runtime.GrandpaSubmitReportEquivocation, buffer.Bytes())
+	return err
```
qdm12, line 276: nit have it as

```suggestion
	if err != nil {
		return err
	}
	
	return nil
```

so that you can more easily spot code not covered. I would guess the exec error case is not test covered 🤔 Would it be possible to plug in bad equivocationProof + keyOwnershipProof to make the runtime return an error?

lib/runtime/wasmer/exports_test.go
```diff
@@ -1210,3 +1210,138 @@ func TestInstance_TransactionPaymentCallApi_QueryCallFeeDetails(t *testing.T) {
...
+
+func TestInstance_GrandpaGenerateKeyOwnershipProofEncoding(t *testing.T) {
+	t.Parallel()
+	keyOwnershipProof := types.OpaqueKeyOwnershipProof([]byte{64, 138, 252, 29, 127, 102, 189, 129, 207, 47, 157,
+		60, 17, 138, 194, 121, 139, 92, 176, 175, 224, 16, 185, 93, 175, 251, 224, 81, 209, 61, 0, 71})
+	encoded := scale.MustMarshal(keyOwnershipProof)
+	var proof types.OpaqueKeyOwnershipProof
+	err := scale.Unmarshal(encoded, &proof)
+	require.NoError(t, err)
+	require.Equal(t, keyOwnershipProof, proof)
```
qdm12, line 1222: Shouldn't this test be in the `types` package? Since we are really testing `types.OpaqueKeyOwnershipProof` + `pkg/scale`

lib/runtime/wasmer/exports_test.go
```diff
@@ -1210,3 +1210,138 @@ func TestInstance_TransactionPaymentCallApi_QueryCallFeeDetails(t *testing.T) {
...
+	}
+	secondVote := types.GrandpaVote{
+		Hash:   common.MustHexToHash("0xc314327941fdd924bc67fd72651c40aececd485ca3e878c21e02abb40feae5bd"),
+		Number: uint32(10),
+	}
+
+	firstSignature := common.MustHexToBytes("0xd7292caacc62504365f179892a7399f233944bf261f8a3f66260f70e0016f2d" +
+		"b63922726b015c82dc7131f4730fbec61f71672a571453e51029bfb469070900f")
+	var firstSignatureArray [64]byte
+	copy(firstSignatureArray[:], firstSignature)
```
qdm12, line 1259: maybe add a little helping function `mustHexTo64BArray` to do these 3 lines in `helpers_test.go`?

lib/runtime/wasmer/exports_test.go
```diff
@@ -1210,3 +1210,138 @@ func TestInstance_TransactionPaymentCallApi_QueryCallFeeDetails(t *testing.T) {
...
+	require.NoError(t, err)
+
+	firstVote := types.GrandpaVote{
+		Hash:   common.MustHexToHash("0x4801b8e62d31167d30c893cc1970f6a0e289420282a4b245b75f2c46fb308af1"),
+		Number: 10,
+	}
+	secondVote := types.GrandpaVote{
+		Hash:   common.MustHexToHash("0xc314327941fdd924bc67fd72651c40aececd485ca3e878c21e02abb40feae5bd"),
+		Number: 10,
+	}
```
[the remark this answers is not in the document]
qdm12: nit I think it's nicer to have local values for readability (no need to lookup around, CTRL+click etc.). Except maybe huge values, but a single line hex string is fine imo.

lib/grandpa/vote_message_test.go
```diff
@@ -0,0 +1,197 @@
...
+	"github.com/ChainSafe/gossamer/dot/types"
+	"github.com/ChainSafe/gossamer/lib/common"
+	"github.com/golang/mock/gomock"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+var (
+	errTestError  = errors.New("test dummy error")
+	testDummyHash = common.NewHash([]byte{1})
```
qdm12, line 19: supernit `testDummyHash` -> `dummyHash` / `testHash`

lib/grandpa/vote_message_integration_test.go
```diff
@@ -120,20 +121,20 @@ func TestCheckForEquivocation_WithExistingEquivocation(t *testing.T) {
...
 	require.Equal(t, 1, len(gs.pvEquivocations))
 
 	vote3 := vote1
 
-	equivocated = gs.checkForEquivocation(&voter, &SignedVote{
+	err = gs.checkForEquivocation(&voter, &SignedVote{
 		Vote: *vote3,
 	}, prevote)
-	require.True(t, equivocated)
+	require.True(t, errors.Is(err, ErrEquivocation))
```
qdm12, line 137:
```suggestion
	require.ErrorIs(t, err, ErrEquivocation)
```

lib/grandpa/vote_message_integration_test.go
```diff
@@ -120,20 +121,20 @@ func TestCheckForEquivocation_WithExistingEquivocation(t *testing.T) {
 	vote2, err := NewVoteFromHash(leaves[0], gs.blockState)
 	require.NoError(t, err)
 
-	equivocated := gs.checkForEquivocation(&voter, &SignedVote{
+	err = gs.checkForEquivocation(&voter, &SignedVote{
 		Vote: *vote2,
 	}, prevote)
-	require.True(t, equivocated)
+	require.True(t, errors.Is(err, ErrEquivocation))
```
qdm12, line 127:
```suggestion
	require.ErrorIs(t, err, ErrEquivocation)
```

lib/grandpa/vote_message_integration_test.go
```diff
@@ -78,10 +79,10 @@ func TestCheckForEquivocation_WithEquivocation(t *testing.T) {
 	vote2, err := NewVoteFromHash(leaves[1], st.Block)
 	require.NoError(t, err)
 
-	equivocated := gs.checkForEquivocation(&voter, &SignedVote{
+	err = gs.checkForEquivocation(&voter, &SignedVote{
 		Vote: *vote2,
 	}, prevote)
-	require.True(t, equivocated)
+	require.True(t, errors.Is(err, ErrEquivocation))
```
qdm12, line 85:
```suggestion
	require.ErrorIs(t, err, ErrEquivocation)
```

lib/grandpa/vote_message.go
```diff
@@ -234,24 +235,94 @@ func (s *Service) checkForEquivocation(voter *Voter, vote *SignedVote, stage Sub
...
 		s.deleteVote(v, stage)
-		return true
+
+		err := s.reportEquivocation(stage, existingVote, vote)
+		if err != nil {
+			// TODO get feedback on if this is a appropriate way to handle error
+			logger.Errorf("%s: %s", errReportingEquivocation, err)
+			return fmt.Errorf("%w: %s", ErrEquivocation, err)
+		}
+		return fmt.Errorf("%w", ErrEquivocation)
```
qdm12, line 260: I think it's good to log as an error the equivocation report error, but it doesn't feel right to put it as context with the returned error. On the other hand, could we add some context about the equivocation (on line `258` and line `260`, which can return the same wrapped error and context)?

Maybe something like

```suggestion
		if err != nil {
			// TODO get feedback on if this is a appropriate way to handle error
			logger.Errorf("failed reporting equivocation: %s", err)
		}
		return fmt.Errorf("%w: voter %s has existing vote %s and new vote %s", 
			ErrEquivocation, v, existingVote.Vote.Hash, vote.Vote.Hash)
```

you can also remove `errReportingEquivocation`

lib/grandpa/vote_message.go
```diff
@@ -234,24 +235,94 @@ func (s *Service) checkForEquivocation(voter *Voter, vote *SignedVote, stage Sub
...
-	return false
+	round, err := s.grandpaState.GetLatestRound()
+	if err != nil {
+		return fmt.Errorf("getting latest round: %w", err)
+	}
+
+	pubKey := existingVote.AuthorityID
+
+	bestBlockHash := s.blockState.BestBlockHash()
+	rt, err := s.blockState.GetRuntime(bestBlockHash)
```
qdm12, line 280: nit `rt` -> `runtime`

lib/grandpa/vote_message.go
```diff
@@ -198,9 +199,9 @@ func (s *Service) validateVoteMessage(from peer.ID, m *VoteMessage) (*Vote, erro
 		AuthorityID: pk.AsBytes(),
 	}
 
-	equivocated := s.checkForEquivocation(voter, just, m.Message.Stage)
-	if equivocated {
-		return nil, fmt.Errorf("%w", ErrEquivocation)
+	err = s.checkForEquivocation(voter, just, m.Message.Stage)
+	if err != nil {
+		return nil, err
```
qdm12, line 204: nit maybe?

```suggestion
		return nil, fmt.Errorf("checking for equivocation: %w", err)
```

lib/grandpa/interfaces.go
```diff
@@ -3,7 +3,16 @@
...
+import (
+	"encoding/json"
+
+	"github.com/ChainSafe/gossamer/dot/types"
+)
+
+// BabeConfigurer returns the babe configuration of the runtime.
+type BabeConfigurer interface {
+	BabeConfiguration() (*types.BabeConfiguration, error)
+}
```
qdm12, line 15: Maybe this can be removed now?

dot/types/grandpa.go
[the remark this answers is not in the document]
jimjbrettj: @timwu20 is there any action item for me on this, or just a note for the future?

lib/grandpa/vote_message.go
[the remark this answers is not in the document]
jimjbrettj: updated and added a unit test to test that case

lib/runtime/wasmer/exports.go
jimjbrettj: I do not believe I can, with invalid data (with respect to the runtime, not formatting) it seems to return an empty proof

lib/runtime/wasmer/exports_test.go
jimjbrettj: yeah I can add an empty check for this case (the more integration-y tests are out of scope of this PR)

lib/runtime/wasmer/exports.go
jimjbrettj: Commented above as well, but it appears to just return an empty proof when provided with incorrect (but correctly encoded) data

jimjbrettj: Similar to the above case, it doesn't seem to error unless due to formatting

lib/runtime/wasmer/exports_test.go
jimjbrettj: good call!

lib/runtime/wasmer/exports_test.go
jimjbrettj: good call!

dot/types/grandpa.go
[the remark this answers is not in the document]
timwu20: No action item for this. Was just adding context.

lib/grandpa/vote_message.go
```diff
@@ -236,22 +238,93 @@ func (s *Service) checkForEquivocation(voter *Voter, vote *SignedVote, stage Sub
...
+		}
+	case precommit:
+		err = equivocationVote.Set(types.PreCommitEquivocation(grandpaEquivocation))
+		if err != nil {
+			return fmt.Errorf("setting grandpa equivocation VDT as precommit equivocation: %w", err)
+		}
+	case primaryProposal:
+		return fmt.Errorf("%w", errInvalidEquivocationStage)
+	default:
+		panic(fmt.Errorf("building equivocation proof: %s", errInvalidEquivocationStage))
```
qdm12, line 314: nit no need to error wrap + stack trace provide function name & line

```suggestion
		panic(fmt.Sprintf("equivocation stage not implemented: %s (%d)", stage, stage))
```

lib/runtime/wasmer/exports.go
```diff
@@ -227,6 +229,56 @@ func (in *Instance) QueryCallFeeDetails(ext []byte) (*types.FeeDetails, error) {
 // TODO: use this in block verification process (#1873)
 func (in *Instance) CheckInherents() {}
 
+// GrandpaGenerateKeyOwnershipProof returns grandpa key ownership proof from runtime.
```
qdm12, line 232: supernit

```suggestion
// GrandpaGenerateKeyOwnershipProof returns grandpa key ownership proof from the runtime.
```

lib/runtime/wasmer/exports.go
```diff
@@ -227,6 +229,56 @@ func (in *Instance) QueryCallFeeDetails(ext []byte) (*types.FeeDetails, error) {
...
+func (in *Instance) GrandpaGenerateKeyOwnershipProof(authSetID uint64, authorityID ed25519.PublicKeyBytes) (
+	types.OpaqueKeyOwnershipProof, error) {
+
```
qdm12, line 235: supernit unneeded start new line (some linters would shout on you 😄)

```suggestion
```

lib/runtime/wasmer/exports.go
```diff
@@ -227,6 +229,56 @@ func (in *Instance) QueryCallFeeDetails(ext []byte) (*types.FeeDetails, error) {
...
+	keyOwnershipProof := types.OpaqueKeyOwnershipProof{}
+	err = scale.Unmarshal(encodedOpaqueKeyOwnershipProof, &keyOwnershipProof)
+	if err != nil {
+		return nil, fmt.Errorf("scale decoding: %w", err)
+	}
+
+	return keyOwnershipProof, nil
+}
+
+// GrandpaSubmitReportEquivocationUnsignedExtrinsic reports equivocation report to the runtime.
```
qdm12, line 261: supernit to avoid confusion with `reports` + `report` 😄 

```suggestion
// GrandpaSubmitReportEquivocationUnsignedExtrinsic reports an equivocation report to the runtime.
```

lib/runtime/wasmer/exports_test.go
```diff
@@ -26,6 +26,13 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+func mustHexTo64BArray(t *testing.T, inputHex string) [64]byte {
+	t.Helper()
+	var outputArray [64]byte
+	copy(outputArray[:], common.MustHexToBytes(inputHex))
+	return outputArray
+}
+
```
qdm12, line 35: Move this to helpers_test.go instead?

lib/runtime/wasmer/exports_test.go
```diff
@@ -1210,3 +1217,120 @@ func TestInstance_TransactionPaymentCallApi_QueryCallFeeDetails(t *testing.T) {
...
+	t.Parallel()
+	runtime := NewTestInstance(t, runtime.WESTEND_RUNTIME_v0929)
+	identity := common.MustHexToBytes("0x88dc3417d5058ec4b4503e0c12ea1a0a89be200fe98922423d4334014fa6b0ee")
+	identityPubKey, _ := ed25519.NewPublicKey(identity)
+	var authorityID [32]byte
+	copy(authorityID[:], identityPubKey.Encode())
+
+	encodedOpaqueKeyOwnershipProof, err := runtime.GrandpaGenerateKeyOwnershipProof(uint64(0), authorityID)
+	require.NoError(t, err)
+	// Since the input is not valid with respect to the runtime and empty proof is returned
```
qdm12, line 1231:
```suggestion
	// Since the input is not valid with respect to the runtime, an empty proof is returned
```

lib/runtime/wasmer/exports_test.go
```diff
@@ -1210,3 +1217,120 @@ func TestInstance_TransactionPaymentCallApi_QueryCallFeeDetails(t *testing.T) {
...
+	require.Empty(t, encodedOpaqueKeyOwnershipProof)
```
qdm12, line 1232: Should we return an error from `runtime.GrandpaGenerateKeyOwnershipProof` is the proof is empty? 🤔

jimjbrettj: I was duplicating @kishansagathiya's PR so perhaps good to talk about with him as well on if this should be an error or not

dot/types/grandpa.go
```diff
@@ -196,3 +196,61 @@ type GrandpaVote struct {
...
+	RoundNumber     uint64
+	ID              [32]byte
+	FirstVote       GrandpaVote
+	FirstSignature  [64]byte
+	SecondVote      GrandpaVote
+	SecondSignature [64]byte
+}
+
+// GrandpaEquivocationVote is a custom vdt type for a grandpa equivocation
```
kishansagathiya, line 210: redo the comment to explain better what it does.

```
/// GrandpaEquivocationVote is a wrapper object for GRANDPA equivocation proofs, useful for unifying prevote
/// and precommit equivocations under a common type.
```

dot/types/grandpa.go
```diff
@@ -196,3 +196,61 @@ type GrandpaVote struct {
...
+type GrandpaEquivocationVote scale.VaryingDataType
```
kishansagathiya, line 211: this makes it sound like it represents vote, which it doesn't.
rename to something like GrandpaEquivocationEnum or GrandpaEquivocationWrapper

dot/types/grandpa.go
```diff
@@ -196,3 +196,61 @@ type GrandpaVote struct {
...
+
+// GrandpaEquivocation is used to create a proof of equivocation
+type GrandpaEquivocation struct {
+	RoundNumber     uint64
+	ID              [32]byte
+	FirstVote       GrandpaVote
+	FirstSignature  [64]byte
+	SecondVote      GrandpaVote
+	SecondSignature [64]byte
+}
```
kishansagathiya, line 208: does this produce same scale encoding as the rust struct?

```
pub struct Equivocation<Id, V, S> {
	/// The round number equivocated in.
	pub round_number: u64,
	/// The identity of the equivocator.
	pub identity: Id,
	/// The first vote in the equivocation.
	pub first: (V, S),
	/// The second vote in the equivocation.
	pub second: (V, S),
}
```

dot/types/grandpa_test.go
```diff
@@ -7,55 +7,161 @@ import (
...
+	encoded := scale.MustMarshal(keyOwnershipProof)
+	var proof GrandpaOpaqueKeyOwnershipProof
+	err := scale.Unmarshal(encoded, &proof)
+	require.NoError(t, err)
+	require.Equal(t, keyOwnershipProof, proof)
+}
+
+func TestInstance_GrandpaSubmitReportEquivocationUnsignedExtrinsicEncoding(t *testing.T) {
+	t.Parallel()
+	expectedEncoding := common.MustHexToBytes("0x010000000000000000010000000000000088dc3417d5058ec4b4503e0c12ea" +
```
kishansagathiya, line 35: put some info about how you got this encoding. Either some rust code or if it is taken from substrate or polkadot, permalink to the code

dot/types/grandpa_test.go
```diff
@@ -7,55 +7,161 @@ import (
...
+	encoding := scale.MustMarshal(*equivVote)
 
-	dec := GrandpaVote{}
-	err = scale.Unmarshal(enc, &dec)
+	grandpaEquivocation := NewGrandpaEquivocation()
+	err = scale.Unmarshal(encoding, grandpaEquivocation)
 	require.NoError(t, err)
-	require.Equal(t, testVote, dec)
+	require.Equal(t, equivVote, grandpaEquivocation)
+}
```
kishansagathiya, line 119: This test that after marshaling we unmarshal to the same struct. I could replace `equivVote` with any struct and this test would work.

Instead, we could test it against some other scale encoding. So we could get rust equivalent of  `equivVote`  and marshal it using rust scale library and test that
- our struct also marshals to the same value.
- we can unmarshal it to our struct

dot/types/grandpa_test.go
```diff
@@ -7,55 +7,161 @@ import (
...
+		Number: 999,
+	}
+
+	encoding := scale.MustMarshal(vote)
+	require.Equal(t, expectedEncoding, encoding)
+
+	grandpaVote := GrandpaVote{}
+	err := scale.Unmarshal(encoding, &grandpaVote)
+	require.NoError(t, err)
+	require.Equal(t, vote, grandpaVote)
```
kishansagathiya, line 135: there isn't much value in such test that use our scale package for both marshal and unmarshal. 

Was about expectedEncoding created using some rust equivalent? Add info to get expectedEncoding in a comment

dot/types/grandpa_test.go
```diff
@@ -7,55 +7,161 @@ import (
...
+	signedVote := GrandpaSignedVote{
+		Vote:        vote,
+		Signature:   signature,
+		AuthorityID: authorityID,
 	}
-	enc, err := scale.Marshal(sv)
-	require.NoError(t, err)
-	require.Equal(t, exp, enc)
+	encoding := scale.MustMarshal(signedVote)
+	require.Equal(t, expectedEncoding, encoding)
```
kishansagathiya, line 154: All this is unnecessary renaming. Please change it back.

dot/types/grandpa_test.go
```diff
@@ -7,55 +7,161 @@ import (
...
+		ID:              authorityID,
+		FirstVote:       firstVote,
+		FirstSignature:  signature,
+		SecondVote:      secondVote,
+		SecondSignature: signature,
+	}
 
-	enc, err := scale.Marshal(testVote)
+	equivPreVote := PreVoteEquivocation(equivocation)
+	equivVote := NewGrandpaEquivocation()
```
kishansagathiya, line 110: rename `equivVote` to something else, as it is not a vote.

dot/types/grandpa_test.go
```diff
@@ -64,21 +170,20 @@ func TestGrandpaAuthoritiesRawToAuthorities(t *testing.T) {
...
+	require.Equal(t, auths[0].ID, authorities[0].Weight)
 
-	a := Authority{}
-	err = a.FromRawEd25519(dec[1])
+	authority := Authority{}
+	err = authority.FromRawEd25519(grandpaAuthoritiesRaw[1])
 	require.NoError(t, err)
-	require.Equal(t, a, authoritys[1])
+	require.Equal(t, authority, authorities[1])
 }
```
kishansagathiya, line 189: Unnecessary renaming, change it back please

lib/grandpa/vote_message.go
```diff
@@ -214,9 +215,11 @@ func (s *Service) validateVoteMessage(from peer.ID, m *VoteMessage) (*Vote, erro
 }
 
 // checkForEquivocation checks if the vote is an equivocatory vote.
-// it returns true if so, false otherwise.
-// additionally, if the vote is equivocatory, it updates the service's votes and equivocations.
-func (s *Service) checkForEquivocation(voter *Voter, vote *SignedVote, stage Subround) bool {
+// If it is an equivocatory vote, the wrapped error `ErrEquivocation` is returned,
+// which can be checked with `error.Is(err, ErrEquivocation)`.
+// Additionally, if the vote is equivocatory, it updates the service's votes and equivocations
+// and reports the equivocation to the runtime.
```
kishansagathiya, line 221:
```suggestion
// If it is an equivocatory vote, the error `ErrEquivocation` is returned, the service's votes and equivocations are updated
// and reports the equivocation to the runtime.
```

lib/grandpa/vote_message.go
```diff
@@ -214,9 +215,11 @@ func (s *Service) validateVoteMessage(from peer.ID, m *VoteMessage) (*Vote, erro
...
+func (s *Service) checkForEquivocation(voter *Voter, vote *SignedVote, stage Subround) error {
```
kishansagathiya, line 222: rename it to checkAndReportEquivocation

kishansagathiya (changes requested):

dot/types/grandpa_test.go
EclesioMeloJunior: IMO it is a necessary renaming since it is fixing the typo `authoritys -> authorities` as well as making the variable name `a` more clear with `authority`

dot/types/grandpa.go
jimjbrettj: Yup! We handle tuples using structs

dot/types/grandpa_test.go
jimjbrettj: Do you think since all this is already tested in an above test, I should just remove this one and test against anencoding above?

jimjbrettj: I opted to delete this test since its already covered in above test

dot/types/grandpa_test.go
jimjbrettj: I think there is still value if the expectedEncoding is derived from rust scale encodings. It allows us to assert that both the encoded and raw data are correct. I will like how the expected encoding was generated

dot/types/grandpa_test.go
jimjbrettj: Why is it unnecessary? I think we opt to use := over var when we can in our code (correct time if I am wrong) so I changed that, and then sv and enc are vague so I made them more clear

lib/grandpa/vote_message.go
jimjbrettj: Is this preferred? Since it is not always reported (most cases won't be equivocations) and we have a separate function for reporting. Can change still if you but I wasn't sure if this was actually more clear

kishansagathiya: `checkForEquivocation` tells you that it is only checking, thus it should return boolean. The function name should tell that it does more than just check.

jimjbrettj: this makes sense to me! Will update

dot/types/grandpa.go
```diff
@@ -196,3 +196,62 @@ type GrandpaVote struct {
...
+
+// GrandpaEquivocationEnum is a wrapper object for GRANDPA equivocation proofs, useful for unifying prevote
+// / and precommit equivocations under a common type.
+type GrandpaEquivocationEnum scale.VaryingDataType
```
timwu20, line 212:
```suggestion
type GrandpaEquivocationEnum scale.VaryingDataType
```
I think this should be named `GrandpaEquivocation` instead of on line 201.  It reflects what is in the `primitives/finality-grandpa` package in substrate.  You can make the struct on line 201 a private base struct and redefine the `PreVoteEquivocation` and `PrecommitEquivocation` definitions.

dot/types/grandpa.go
```diff
@@ -196,3 +196,62 @@ type GrandpaVote struct {
...
+
+// NewGrandpaEquivocation returns a new VaryingDataType to represent a grandpa Equivocation
+func NewGrandpaEquivocation() *GrandpaEquivocationEnum {
+	vdt := scale.MustNewVaryingDataType(PreVoteEquivocation{}, PreCommitEquivocation{})
+	ge := GrandpaEquivocationEnum(vdt)
+	return &ge
+}
+
+// PreVoteEquivocation equivocation type for a prevote
+type PreVoteEquivocation GrandpaEquivocation
```
timwu20, line 239:
```suggestion
type Prevote GrandpaEquivocation
```
the type infers that it is an equivocation

dot/types/grandpa.go
```diff
@@ -196,3 +196,62 @@ type GrandpaVote struct {
...
+
+// Index returns VDT index
+func (PreVoteEquivocation) Index() uint { return 0 }
+
+// PreCommitEquivocation equivocation type for a precommit
+type PreCommitEquivocation GrandpaEquivocation
```
timwu20, line 245:
```suggestion
type PreCommit GrandpaEquivocation
```

dot/types/grandpa.go
```diff
@@ -196,3 +196,62 @@ type GrandpaVote struct {
...
+
+// Index returns VDT index
+func (PreCommitEquivocation) Index() uint { return 1 }
+
+// GrandpaOpaqueKeyOwnershipProof contains a key ownership proof for reporting equivocations
+type GrandpaOpaqueKeyOwnershipProof []byte
+
+// GrandpaEquivocationProof is used to report grandpa equivocations
+type GrandpaEquivocationProof struct {
```
timwu20, line 254: I think it may make sense to create a separate sub package to house these primitives that map 1 to 1 to the `primitives/finality-grandpa` crate.  Would need to do the same for BABE.

timwu20 (changes requested):

dot/types/grandpa.go
jimjbrettj: Unless I am misunderstanding, this causes issues because the current GrandpaEquivocation struct on line 201 is used in other packages. What I think we can do is when we have the standalone grandpa package, they can both be GrandpaEquivocation. I do believe this is what substrate does

dot/types/grandpa.go
jimjbrettj: I think this is related to my comment here, https://github.com/ChainSafe/gossamer/pull/3007#discussion_r1093771317

Let me know what you think

dot/types/grandpa.go
timwu20: I see.  There are two types named `Equivocation` in substrate.  One is in `sp_finality_grandpa::Equivocation` and one is `finality_grandpa::Equivocation`.  The `sp_finality_grandpa::Equivocation` type references `finality_grandpa::Equivocation`.  I think the types within the `types` package should hold `sp_finality_grandpa::Equivocation` and we should move the type on line 201 to the `grandpa` package which will be replaced by the standalone grandpa package in the future.

lib/runtime/wasmer/test_helpers.go
```diff
@@ -45,6 +46,17 @@ func NewTestInstanceWithTrie(t *testing.T, targetRuntime string, tt *trie.Trie)
 	return r
 }
 
+func mustHexTo64BArray(t *testing.T, inputHex string) (outputArray [64]byte) {
+	t.Helper()
+	copy(outputArray[:], common.MustHexToBytes(inputHex))
+	return outputArray
+}
+
+func ed25519PubKeyTo32BArray(pubKey ed25519.PublicKey) (array [32]byte) {
```
kishansagathiya, line 55: `PublicKeyLength` is 32 bytes anyway, so you dont need this function. just do `pubKey.AsBytes()`

dot/types/grandpa_test.go
```diff
@@ -7,55 +7,129 @@ import (
...
+	"github.com/ChainSafe/gossamer/lib/crypto/ed25519"
 	"github.com/ChainSafe/gossamer/pkg/scale"
 
 	"github.com/stretchr/testify/require"
 )
 
-func TestEncodeGrandpaVote(t *testing.T) {
-	exp := common.MustHexToBytes("0x0a0b0c0d00000000000000000000000000000000000000000000000000000000e7030000")
-	var testVote = GrandpaVote{
+func mustHexTo64BArray(t *testing.T, inputHex string) (outputArray [64]byte) {
```
kishansagathiya, line 16: Since this function is defined twice and seems rather general purpose, I think we can add this in `lib/common`, you can move this function there and introduce common.Signature

`type Signature [64]byte`

lib/runtime/wasmer/test_helpers.go
```diff
@@ -45,6 +46,17 @@ func NewTestInstanceWithTrie(t *testing.T, targetRuntime string, tt *trie.Trie)
 	return r
 }
 
+func mustHexTo64BArray(t *testing.T, inputHex string) (outputArray [64]byte) {
```
kishansagathiya, line 49: Since this function is defined twice and seems rather general purpose, I think we can add this in lib/common, you can move this function there and introduce common.Signature

`type Signature [64]byte`

kishansagathiya (approved):
I have some minor comments, but this is green for me!

lib/runtime/wasmer/test_helpers.go
jimjbrettj: Nice catch!

lib/common/common.go
```diff
@@ -86,6 +90,14 @@ func MustHexToBytes(in string) []byte {
...
 }
 
+// MustHexTo64BArray turns a 0x prefixed hex string into a Signature
+// it panic if it cannot decode the string
+func MustHexTo64BArray(t *testing.T, inputHex string) (outputArray Signature) {
+	t.Helper()
+	copy(outputArray[:], MustHexToBytes(inputHex))
+	return outputArray
+}
+
```
timwu20, line 100: This is a helper function, should be in a test file.

timwu20 (changes requested):
just the small change.

timwu20 (approved):

Issue #3010, opened 2022-12-23
Adding support for using locally provided runtime


radkomih:
## Issue summary

Hi, I am using a fork of Gossamer for running tests by providing local runtime instead of fetching a remote one. I am using the fork only because of this snippet:
```go
// NewLocalTestInstanceWithTrie ...
func NewLocalTestInstanceWithTrie(t *testing.T, runtimeFilepath string, tt *trie.Trie) *Instance {
	t.Helper()

	cfg := setupConfig(t, tt, DefaultTestLogLvl, common.NoNetworkRole, runtimeFilepath)

	r, err := NewInstanceFromFile(runtimeFilepath, cfg)
	require.NoError(t, err)
	return r
}
```
Is it an option for you to integrate it into the official Gossamer?

Thanks!

q9f:
Hi, which fork is that?

radkomih:
It is this one https://github.com/LimeChain/gossamer, but the only difference is the above snippet (and currently a bit outdated)

EclesioMeloJunior:
Hey @radkomih, instead of creating a new function I propose an update in the `NewTestInstanceWithTrie` to check if the `targetRuntime` argument is a valid file path if so we call `NewInstanceFromFile` otherwise we fetch it remotely, take a look in the following PR https://github.com/ChainSafe/gossamer/pull/3012.

This can be really helpful for testing as you're using  👍

radkomih:
thanks!

EclesioMeloJunior:
@radkomih fyi PR #3012 was merged!

PR #3012, opened 2023-01-03
chore(lib/wasmer): `NewTestInstanceWithTrie` checks if `targetRuntime` is a valid filepath

EclesioMeloJunior:
## Changes

- `NewTestInstanceWithTrie` is a function that given a target runtime it builds a `wasmer` instance for testing, now this function will check if the target runtime is a valid file path, if so we will use it to load the `wasmer` instance otherwise we will try to fetch it remotely.

## Tests

N/A

## Issues

- Closes https://github.com/ChainSafe/gossamer/issues/3010

## Primary Reviewer

@timwu20


kishansagathiya (approved):

qdm12 (commented):
Doesn't `runtime.GetRuntime` already do this check?

```go
	if utils.PathExists(runtimePath) {
		return runtimePath, nil
	}
```

lib/runtime/wasmer/test_helpers.go
```diff
@@ -31,11 +33,16 @@ func NewTestInstanceWithTrie(t *testing.T, targetRuntime string, tt *trie.Trie)
 	t.Helper()
 
 	cfg := setupConfig(t, tt, DefaultTestLogLvl, common.NoNetworkRole, targetRuntime)
-	runtimeFilepath, err := runtime.GetRuntime(context.Background(), targetRuntime)
-	require.NoError(t, err)
 
-	r, err := NewInstanceFromFile(runtimeFilepath, cfg)
+	_, err := os.Stat(targetRuntime)
+	if errors.Is(err, os.ErrNotExist) {
+		targetRuntime, err = runtime.GetRuntime(context.Background(), targetRuntime)
```
qdm12, line 39: 1. Should we move this check within `runtime.GetRuntime`? 🤔 
2. We should document (aka comment the function) that `targetRuntime` can be a constant string defined or a file path, in this case in the `NewTestInstanceWithTrie` comment

lib/runtime/wasmer/test_helpers.go
```diff
@@ -26,16 +26,18 @@ func NewTestInstance(t *testing.T, targetRuntime string) *Instance {
 	return NewTestInstanceWithTrie(t, targetRuntime, nil)
 }
 
-// NewTestInstanceWithTrie will create a new runtime (polkadot/test) with the supplied trie as the storage
+// NewTestInstanceWithTrie will return a wasmer.Instance given the targetRuntime, using the provided trie as the storage
+// the targetRuntime can be a file path to a local .wasm file or a already defined constant in runtime/constants.go
```
qdm12, line 30: Maybe also update the comment of `runtime.GetRuntime` 🤔

qdm12 (approved):

lib/runtime/test_helpers.go
```diff
@@ -99,6 +101,9 @@ func GetRuntime(ctx context.Context, runtime string) (
 		runtimeFilename = WESTEND_RUNTIME_V0929_FP
 		url = WESTEND_RUNTIME_V0929_URL
 	default:
+		if utils.PathExists(runtime) {
+			return runtime, nil
+		}
```
qdm12, line 106: supernit maybe move this up at the top of the function, so that:

- the directory `/tmp/gossamer/runtimes/` isn't created if we use a runtime file we specify
- if we have a file with the same name as the runtime string constant (i.e. `WESTEND_RUNTIME_v0929`), the file specified takes precedence over the hardcoded one

The comment above `If the runtime argument is not defined in the constants.go and is a valid` may have to be changed a bit.

lib/runtime/test_helpers.go
```diff
@@ -57,6 +57,8 @@ var (
 // GetRuntime returns the runtime file path located in the
 // /tmp/gossamer/runtimes directory (depending on OS and environment).
 // If the file did not exist, the runtime WASM blob is downloaded to that file.
+// If the runtime argument is not defined in the constants.go and is a valid
+// file path this function will return the runtime argument
```
qdm12, line 61: supernit 
```suggestion
// file path, the runtime argument is returned.
```

lib/runtime/wasmer/test_helpers.go
qdm12, line 30: supernit no need to specify `wasmer.Instance` (check the function signature) + hopefully simplify/clarify a tiny bit


```suggestion
// NewTestInstanceWithTrie returns an instance based on the target runtime string specified, 
// which can be a file path or a constant from the constants defined in `lib/runtime/constants.go`.
// The instance uses the trie given as argument for its storage.
```

qdm12 (approved):
Just supernits, looks good to me already 👍

Issue #3019, opened 2023-01-06
`TaggedTransactionQueue_validate_transaction` fails in `TestBuildAndApplyExtrinsic` while using westend-dev spec file


EclesioMeloJunior:
In order to address the epic https://github.com/ChainSafe/gossamer/issues/2895 we should replace the usage of `GetGssmrV3SubstrateGenesisRawPathTest` to `GetWestendDevGenesisPath` but in the test function `TestBuildAndApplyExtrinsic` (`lib/babe/build_integration_test.go`) we got the following error:

```
CRITICAL target=runtime message=panicked at 'Bad input data provided to validate_transaction: Codec error', /builds/runtime/westend/src/lib.rs:1276:1	ext_logging_log_version_1	pkg=runtime module=go-wasmer
```
and the test fails with the following assertion problem:

```
--- FAIL: TestBuildAndApplyExtrinsic (17.02s)
    build_integration_test.go:269: 
        	Error Trace:	/home/runner/work/gossamer/gossamer/lib/babe/build_integration_test.go:269
        	Error:      	Received unexpected error:
        	            	running runtime function: Failed to call the `TaggedTransactionQueue_validate_transaction` exported function.
        	Test:       	TestBuildAndApplyExtrinsic
```

We should be able to run this test against the westend-dev spec file without errors

EclesioMeloJunior:
Resolved by #3037

Issue #2704, opened 2022-07-26
tests(lib/babe): Add tests against latest `dev` runtime


timwu20:
## Issue summary

- The following tests have been updated in PR #2671 to point to the old `dev` runtime:
  - `TestBuildBlock_ok`
  - `TestApplyExtrinsic`
  - `TestBuildAndApplyExtrinsic`
  - `TestBuildBlockTimeMonitor`
  - `TestVerificationManager_OnDisabled_InvalidIndex`
  - `TestVerificationManager_OnDisabled_NewDigest`
  - `TestVerificationManager_OnDisabled_DuplicateDigest`
  - `TestVerificationManager_VerifyBlock_Ok`
  - `TestVerificationManager_VerifyBlock_Secondary`
  - `TestVerificationManager_VerifyBlock_InvalidBlockOverThreshold`
  - `TestVerificationManager_VerifyBlock_InvalidBlockAuthority`
  - `TestVerifyAuthorshipRight`
  - `TestVerifyAuthorshipRight_Equivocation`
- Investigate and add new tests against latest `dev` runtime  

## Other information and links

- #2671

EclesioMeloJunior:
@jimjbrettj the PR #3037 resolves this issue?

Issue #2336, opened 2022-02-28
Refactor existing test helping functions


qdm12:
## Task summary

Most of our unit and integration tests use test helping functions to setup things. This is fine for some integration tests, but it really does feel like it's over-used.

It also allows to justify using `gomock.Any()` and `.AnyTimes()` for mocks, which both reduce testing depth.

This task should aim at reducing their usage or make these helping functions more configurable.

PR #3037, opened 2023-01-12
chore(lib/babe): fix TestBuildAndApplyExtrinsic for westend 

jimjbrettj:
## Changes

- Fix all babe integration tests to use Westend
- Pass in genesis, genesisTrie, and GenesisHeader to service builder in tests to allow for more flexible testing
- Create a helper function to get Westend genesis
- Create issue to address skipped test in this PR that require a more in-depth refactor or rewrite

## Tests

```sh
go test --tags=integration ./lib/babe
```

## Issues

#3019 #2704

## Primary Reviewer

@EclesioMeloJunior

### DIFF
--- a/lib/babe/babe_integration_test.go
+++ b/lib/babe/babe_integration_test.go
@@ -229,8 +229,8 @@ func TestService_HandleSlotWithSameSlot(t *testing.T) {
 		},
 	}
 
-	gen, genTrie, genHeader := newWestendDevGenesisWithTrieAndHeader(t)
-	babeServiceBob := createTestService(t, cfgBob, gen, genTrie, genHeader)
+	genBob, genTrieBob, genHeaderBob := newWestendDevGenesisWithTrieAndHeader(t)
+	babeServiceBob := createTestService(t, cfgBob, genBob, genTrieBob, genHeaderBob)
 
 	err := babeServiceBob.Start()
 	require.NoError(t, err)
@@ -250,8 +250,8 @@ func TestService_HandleSlotWithSameSlot(t *testing.T) {
 
 	time.Sleep(babeServiceBob.constants.slotDuration)
 
-	gen, genTrie, genHeader = newWestendDevGenesisWithTrieAndHeader(t)
-	babeServiceAlice := createTestService(t, cfgAlice, gen, genTrie, genHeader)
+	genAlice, genTrieAlice, genHeaderAlice := newWestendDevGenesisWithTrieAndHeader(t)
+	babeServiceAlice := createTestService(t, cfgAlice, genAlice, genTrieAlice, genHeaderAlice)
 
 	// Add block created by Bob to Alice
 	err = babeServiceAlice.blockState.AddBlock(block)
--- a/lib/babe/babe_integration_test.go
+++ b/lib/babe/babe_integration_test.go
@@ -16,6 +16,7 @@ import (
 	"github.com/ChainSafe/gossamer/lib/babe/mocks"
 	"github.com/ChainSafe/gossamer/lib/common"
 	"github.com/ChainSafe/gossamer/lib/crypto/sr25519"
+	"github.com/ChainSafe/gossamer/lib/genesis"
 	"github.com/ChainSafe/gossamer/lib/runtime"
 	rtstorage "github.com/ChainSafe/gossamer/lib/runtime/storage"
 	"github.com/ChainSafe/gossamer/lib/runtime/wasmer"
@@ -50,10 +51,17 @@ var (
 	}
 )
 
-func createTestService(t *testing.T, cfg ServiceConfig) *Service {
+func createTestService(t *testing.T, cfg ServiceConfig, useWestend bool) *Service {
 	wasmer.DefaultTestLogLvl = 1
 
-	gen, genTrie, genHeader := newDevGenesisWithTrieAndHeader(t)
+	var gen genesis.Genesis
+	var genTrie trie.Trie
+	var genHeader types.Header
+	if useWestend {
+		gen, genTrie, genHeader = newWestendGenesisWithTrieAndHeader(t)
+	} else {
+		gen, genTrie, genHeader = newDevGenesisWithTrieAndHeader(t)
+	}
 	genesisHeader = &genHeader
 
 	var err error
@@ -212,7 +220,7 @@ func TestService_ProducesBlocks(t *testing.T) {
 		Lead:               true,
 		BlockImportHandler: blockImportHandler,
 	}
-	babeService := createTestService(t, cfg)
+	babeService := createTestService(t, cfg, true)
 
 	err := babeService.Start()
 	require.NoError(t, err)
@@ -256,7 +264,7 @@ func TestService_GetAuthorityIndex(t *testing.T) {
 }
 
 func TestStartAndStop(t *testing.T) {
-	bs := createTestService(t, ServiceConfig{})
+	bs := createTestService(t, ServiceConfig{}, true)
 	err := bs.Start()
 	require.NoError(t, err)
 	err = bs.Stop()
@@ -264,7 +272,7 @@ func TestStartAndStop(t *testing.T) {
 }
 
 func TestService_PauseAndResume(t *testing.T) {
-	bs := createTestService(t, ServiceConfig{})
+	bs := createTestService(t, ServiceConfig{}, true)
 	err := bs.Start()
 	require.NoError(t, err)
 	time.Sleep(time.Second)
@@ -296,7 +304,7 @@ func TestService_HandleSlotWithLaggingSlot(t *testing.T) {
 		Authority: true,
 		Lead:      true,
 	}
-	babeService := createTestService(t, cfg)
+	babeService := createTestService(t, cfg, false)
 
 	err := babeService.Start()
 	require.NoError(t, err)
@@ -389,7 +397,7 @@ func TestService_HandleSlotWithSameSlot(t *testing.T) {
 		},
 	}
 
-	babeServiceBob := createTestService(t, cfgBob)
+	babeServiceBob := createTestService(t, cfgBob, false)
 
 	err := babeServiceBob.Start()
 	require.NoError(t, err)
@@ -409,7 +417,7 @@ func TestService_HandleSlotWithSameSlot(t *testing.T) {
 
 	time.Sleep(babeServiceBob.constants.slotDuration)
 
-	babeServiceAlice := createTestService(t, cfgAlice)
+	babeServiceAlice := createTestService(t, cfgAlice, false)
 
 	// Add block created by Bob to Alice
 	err = babeServiceAlice.blockState.AddBlock(block)
@@ -446,5 +454,4 @@ func TestService_HandleSlotWithSameSlot(t *testing.T) {
 		0,
 		preRuntimeDigest)
 	require.NoError(t, err)
-
 }
--- a/lib/babe/babe_integration_test.go
+++ b/lib/babe/babe_integration_test.go
@@ -127,7 +127,9 @@ func TestService_PauseAndResume(t *testing.T) {
 	require.NoError(t, err)
 }
 
+// Since this test and TestService_HandleSlotWithSameSlot are very similar, improve this one along with it in #3060
 func TestService_HandleSlotWithLaggingSlot(t *testing.T) {
+	t.Skip()
 	cfg := ServiceConfig{
 		Authority: true,
 		Lead:      true,
--- a/lib/babe/babe_integration_test.go
+++ b/lib/babe/babe_integration_test.go
@@ -6,6 +6,7 @@
 package babe
 
 import (
+	"fmt"
 	"testing"
 	"time"
 
@@ -51,17 +52,17 @@ var (
 	}
 )
 
-func createTestService(t *testing.T, cfg ServiceConfig, useWestend bool) *Service {
+func createTestService(t *testing.T, cfg ServiceConfig) *Service {
 	wasmer.DefaultTestLogLvl = 1
 
 	var gen genesis.Genesis
 	var genTrie trie.Trie
 	var genHeader types.Header
-	if useWestend {
-		gen, genTrie, genHeader = newWestendGenesisWithTrieAndHeader(t)
-	} else {
-		gen, genTrie, genHeader = newDevGenesisWithTrieAndHeader(t)
-	}
+
+	gen, genTrie, genHeader = newWestendGenesisWithTrieAndHeader(t)
+
+	//gen, genTrie, genHeader = newDevGenesisWithTrieAndHeader(t)
+
 	genesisHeader = &genHeader
 
 	var err error
@@ -220,7 +221,7 @@ func TestService_ProducesBlocks(t *testing.T) {
 		Lead:               true,
 		BlockImportHandler: blockImportHandler,
 	}
-	babeService := createTestService(t, cfg, true)
+	babeService := createTestService(t, cfg)
 
 	err := babeService.Start()
 	require.NoError(t, err)
@@ -264,7 +265,7 @@ func TestService_GetAuthorityIndex(t *testing.T) {
 }
 
 func TestStartAndStop(t *testing.T) {
-	bs := createTestService(t, ServiceConfig{}, true)
+	bs := createTestService(t, ServiceConfig{})
 	err := bs.Start()
 	require.NoError(t, err)
 	err = bs.Stop()
@@ -272,7 +273,7 @@ func TestStartAndStop(t *testing.T) {
 }
 
 func TestService_PauseAndResume(t *testing.T) {
-	bs := createTestService(t, ServiceConfig{}, true)
+	bs := createTestService(t, ServiceConfig{})
 	err := bs.Start()
 	require.NoError(t, err)
 	time.Sleep(time.Second)
@@ -299,12 +300,12 @@ func TestService_PauseAndResume(t *testing.T) {
 	require.NoError(t, err)
 }
 
-func TestService_HandleSlotWithLaggingSlot(t *testing.T) {
+func TestService_HandleSlotWithLaggingSlotNew(t *testing.T) {
 	cfg := ServiceConfig{
 		Authority: true,
 		Lead:      true,
 	}
-	babeService := createTestService(t, cfg, false)
+	babeService := createTestService(t, cfg)
 
 	err := babeService.Start()
 	require.NoError(t, err)
@@ -322,8 +323,82 @@ func TestService_HandleSlotWithLaggingSlot(t *testing.T) {
 	epochData, err := babeService.initiateEpoch(testEpochIndex)
 	require.NoError(t, err)
 
-	ext := runtime.NewTestExtrinsic(t, rt, parentHash, parentHash, 0,
-		signature.TestKeyringPairAlice, "System.remark", []byte{0xab, 0xcd})
+	slot := getSlot(t, rt, time.Now())
+	fmt.Println(slot.duration)
+
+	ext := runtime.NewTestExtrinsic(t, rt, parentHash, parentHash, 0, signature.TestKeyringPairAlice, "System.remark", []byte{0xab, 0xcd})
+	block := createTestBlockWithSlot(t, babeService, emptyHeader, [][]byte{common.MustHexToBytes(ext)}, testEpochIndex, epochData, slot)
+
+	babeService.blockState.AddBlock(block)
+	time.Sleep(babeService.constants.slotDuration)
+
+	header, err := babeService.blockState.BestBlockHeader()
+	require.NoError(t, err)
+
+	bestBlockSlotNum, err := babeService.blockState.GetSlotForBlock(header.Hash())
+	require.NoError(t, err)
+
+	slotnum := uint64(1)
+	slot = Slot{
+		start:    time.Now(),
+		duration: time.Second,
+		number:   slotnum,
+	}
+	preRuntimeDigest, err := types.NewBabePrimaryPreDigest(
+		0, slot.number,
+		[sr25519.VRFOutputLength]byte{},
+		[sr25519.VRFProofLength]byte{},
+	).ToPreRuntimeDigest()
+
+	require.NoError(t, err)
+
+	err = babeService.handleSlot(
+		babeService.epochHandler.epochNumber,
+		bestBlockSlotNum-1,
+		babeService.epochHandler.epochData.authorityIndex,
+		preRuntimeDigest)
+
+	require.ErrorIs(t, err, errLaggingSlot)
+}
+
+func TestService_HandleSlotWithLaggingSlot(t *testing.T) {
+	cfg := ServiceConfig{
+		Authority: true,
+		Lead:      true,
+	}
+	babeService := createTestService(t, cfg)
+
+	err := babeService.Start()
+	require.NoError(t, err)
+	defer func() {
+		err = babeService.Stop()
+		require.NoError(t, err)
+	}()
+
+	// add a block
+	parentHash := babeService.blockState.GenesisHash()
+	bestBlockHash := babeService.blockState.BestBlockHash()
+	rt, err := babeService.blockState.GetRuntime(bestBlockHash)
+	require.NoError(t, err)
+
+	epochData, err := babeService.initiateEpoch(testEpochIndex)
+	require.NoError(t, err)
+
+	//babeConfig, err := rt.BabeConfiguration()
+	//require.NoError(t, err)
+
+	//babeConfig.SlotDuration
+	//
+	//duration, err := time.ParseDuration("1s")
+	//require.NoError(t, err)
+	//
+	//slot := Slot{
+	//	start:    time.Now(),
+	//	duration: time.Duration(babeConfig.SlotDuration),
+	//	number:   slotNumber,
+	//}
+
+	ext := runtime.NewTestExtrinsic(t, rt, parentHash, parentHash, 0, "System.remark", []byte{0xab, 0xcd})
 	block := createTestBlock(t, babeService, emptyHeader, [][]byte{common.MustHexToBytes(ext)},
 		1, testEpochIndex, epochData)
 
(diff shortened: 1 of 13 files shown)


lib/babe/babe_integration_test.go
```diff
@@ -50,10 +51,17 @@ var (
 	}
 )
 
-func createTestService(t *testing.T, cfg ServiceConfig) *Service {
+func createTestService(t *testing.T, cfg ServiceConfig, useWestend bool) *Service {
```
qdm12, line 54: could we instead inject the gen+genTrie+genHeader to this helper function? So we move one inch closer to #2336 and also it would be easier to change when adding more runtime genesis

qdm12 (approved):

lib/babe/babe_integration_test.go
EclesioMeloJunior, line 54: we should use `westend-dev` or `westend-local` as our spec file and remove the other ones, so the `useWestend` argument is not needed here.

EclesioMeloJunior (changes requested):

lib/babe/babe_integration_test.go
jimjbrettj: Yeah i like this. Let me finish fixing these babe tests, then depending on size of PR will either do here or in different PR

qdm12: Check out Eclesio's requested changes first, we might not need those changes at all

jimjbrettj: Even with eclesio's suggestion I decided to incase we want to test with other runtimes in the future

jimjbrettj: I had done this initially to allow for fixing tests to use westend incrementally across multiple PRs, but I have resolved all of them in this one now so the boolean is gone.

lib/babe/build.go
```diff
@@ -261,6 +260,8 @@ func buildBlockInherents(slot Slot, rt ExtrinsicHandler, parent *types.Header) (
 		return nil, err
 	}
 
+	// ienc is same length
+
```
EclesioMeloJunior, line 264:
```suggestion
```

lib/babe/build_integration_test.go
```diff
@@ -54,41 +54,10 @@ func TestSeal(t *testing.T) {
...
-	block, err := babeService.buildBlock(parent, slot, rt, epochData.authorityIndex, preRuntimeDigest)
-	require.NoError(t, err)
-
-	babeService.blockState.(*state.BlockState).StoreRuntime(block.Header.Hash(), rt)
-	return block
-}
-
-// TODO: add test against latest dev runtime
-// See https://github.com/ChainSafe/gossamer/issues/2704
+// TODO see if there can be better assertions on block body
```
EclesioMeloJunior, line 57: I don't think this TODO is needed the test function `TestBuildBlock_ok` includes the `System.remark` extrinsic and AFAIK the timestamp is in the block body as well, so makes sense `require.Equal(t, 2, len(extsBytes))`

EclesioMeloJunior (changes requested):

lib/babe/build_integration_test.go
jimjbrettj: @timwu20 what you think?

lib/babe/build.go
jimjbrettj: good catch!

lib/babe/babe_integration_test.go
```diff
@@ -349,7 +187,9 @@ func TestService_HandleSlotWithLaggingSlot(t *testing.T) {
 	require.ErrorIs(t, err, errLaggingSlot)
 }
 
+// TODO Rewrite this test to utilise westend. Since its built for 2 nodes, doesnt work with existing setup
```
kishansagathiya, line 190: add issue number here

lib/babe/babe_integration_test.go
```diff
@@ -349,7 +187,9 @@ func TestService_HandleSlotWithLaggingSlot(t *testing.T) {
...
 func TestService_HandleSlotWithSameSlot(t *testing.T) {
+	t.Skip()
```
kishansagathiya, line 192: may be I don't have all the context. Why are we skipping this?

lib/babe/babe_integration_test.go
```diff
@@ -387,7 +227,8 @@ func TestService_HandleSlotWithSameSlot(t *testing.T) {
 		},
 	}
 
-	babeServiceBob := createTestService(t, cfgBob)
+	gen, genTrie, genHeader := newWestendDevGenesisWithTrieAndHeader(t)
```
kishansagathiya, line 230:
```suggestion
	genBob, genTrieBob, genHeaderBob := newWestendDevGenesisWithTrieAndHeader(t)
```

kishansagathiya: better to have different names for alice and bob

lib/babe/babe_integration_test.go
```diff
@@ -407,7 +248,8 @@ func TestService_HandleSlotWithSameSlot(t *testing.T) {
 
 	time.Sleep(babeServiceBob.constants.slotDuration)
 
-	babeServiceAlice := createTestService(t, cfgAlice)
+	gen, genTrie, genHeader = newWestendDevGenesisWithTrieAndHeader(t)
```
kishansagathiya, line 251:
```suggestion
	genAlice, genTrieAlice, genHeaderAlice = newWestendDevGenesisWithTrieAndHeader(t)
```

lib/babe/build.go
```diff
@@ -196,7 +196,7 @@ func (b *BlockBuilder) buildBlockExtrinsics(slot Slot, rt ExtrinsicHandler) []*t
 
 		err = determineErr(ret)
 		if err != nil {
-			logger.Warnf("failed to apply extrinsic %s: %s", extrinsic, err)
+			logger.Warnf("error when applying extrinsic %s: %s", extrinsic, err)
```
kishansagathiya, line 199: why? It's the same thing as before

lib/babe/build.go
```diff
@@ -229,8 +229,7 @@ func (b *BlockBuilder) buildBlockExtrinsics(slot Slot, rt ExtrinsicHandler) []*t
 func buildBlockInherents(slot Slot, rt ExtrinsicHandler, parent *types.Header) ([][]byte, error) {
 	// Setup inherents: add timstap0
 	idata := types.NewInherentData()
-	timestamp := uint64(time.Now().UnixMilli())
-	err := idata.SetInherent(types.Timstap0, timestamp)
+	err := idata.SetInherent(types.Timstap0, uint64(slot.start.UnixMilli()))
```
kishansagathiya, line 232: Shouldn't this be current time stamp? I have asked in the elements chat for confirmation.

lib/babe/build_integration_test.go
kishansagathiya, line 57: Is this TODO still pending?

lib/babe/build.go
kishansagathiya: Looks like @EclesioMeloJunior also thinks that it should be current timestamp

lib/babe/build_integration_test.go
```diff
@@ -278,10 +236,12 @@ func TestBuildAndApplyExtrinsic(t *testing.T) {
 	require.Equal(t, []byte{0, 0}, res)
 }
 
+// TODO investigate if this is a needed test
```
kishansagathiya, line 239: reminder to remove or address this

timwu20: do we have any other tests that assert an error returned from `buildBlock`?  If not, it makes sense to cover some error case.

lib/babe/build_integration_test.go
```diff
@@ -364,8 +324,6 @@ func TestDecodeExtrinsicBody(t *testing.T) {
 	require.True(t, contains)
 }
 
-// TODO: add test against latest dev runtime
-// See https://github.com/ChainSafe/gossamer/issues/2704
 func TestBuildBlockTimeMonitor(t *testing.T) {
```
timwu20, line 327: This is just testing some metrics?

lib/babe/build_integration_test.go
```diff
@@ -374,18 +332,23 @@ func TestBuildBlockTimeMonitor(t *testing.T) {
 		Authority: true,
 	}
```
timwu20, line 333: do we still need this?

lib/babe/build_integration_test.go
```diff
@@ -374,18 +332,23 @@ func TestBuildBlockTimeMonitor(t *testing.T) {
...
 
-	babeService := createTestService(t, cfg)
+	genesis, genesisTrie, genesisHeader := newWestendDevGenesisWithTrieAndHeader(t)
+	babeService := createTestService(t, cfg, genesis, genesisTrie, genesisHeader)
```
timwu20, line 336: is `cfg` even needed anymore?

lib/babe/build.go
jimjbrettj: oh really? Okay will update this

jimjbrettj: So to be clear, by "current timestamp" they meant current as in when the inherent are created as opposed to current being when the slot is claimed? @kishansagathiya @EclesioMeloJunior

jimjbrettj: I think we do this in other parts of the code as well https://github.com/ChainSafe/gossamer/blob/2d5ead18ed3aa8fd1e412b26755a165de2248c02/dot/sync/chain_processor_integration_test.go#L46

lib/babe/babe_integration_test.go
jimjbrettj: No worries! So there is kinda two answers. 
1) I have created an issue to address fixes in babe integration tests using Westend that require larger changes or rewrites. So I am skipping in this PR to allow a more straightforward review and will pick up that issue next.
2) This test specifically needs to be rewritten. It was created for a two-node gossamer test net, but with the Westend runtimes we have we only have 1 and 3-node network options. Since we need 2 nodes, we cannot use the one-node network. As for the three-node network, we cannot know currently that our nodes have authorship rights for a slot, so this results in this test being flaky.

lib/babe/build.go
jimjbrettj: I changed this because there is a case above that has the exact same log, but it continues rather than handle it. I changed the logs to differentiate between the two cases. But can leave if this is what we want.

lib/babe/build.go
jimjbrettj: Talked with @EclesioMeloJunior and I understand now, will update!

lib/babe/build_integration_test.go
jimjbrettj: Added some of these questions to the comment to provide some research guidance

lib/babe/build_integration_test.go
jimjbrettj: Don't believe so. I also think this test can be cleaned up. Its on my list of ones to address in my follow up PR

lib/babe/babe_integration_test.go
```diff
@@ -255,38 +89,40 @@ func TestService_GetAuthorityIndex(t *testing.T) {
...
 	}()
 
 	go func() {
-		err := bs.Resume()
+		err := babeService.Resume()
 		require.NoError(t, err)
 	}()
 
-	err = bs.Stop()
+	err = babeService.Stop()
```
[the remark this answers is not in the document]
jimjbrettj: I believe it is to make sure nothing breaks if we call one of these function when we are already in the respective state. Example, calling pause when we are already paused

lib/babe/build_integration_test.go
```diff
@@ -115,101 +85,87 @@ func TestBuildBlock_ok(t *testing.T) {
 
 	// confirm block body is correct
 	extsBytes := types.ExtrinsicsArrayToBytesArray(block.Body)
-	require.Equal(t, 1, len(extsBytes))
+	require.Equal(t, 2, len(extsBytes))
 }
 
-// TODO: add test against latest dev runtime
-// See https://github.com/ChainSafe/gossamer/issues/2704
 func TestApplyExtrinsic(t *testing.T) {
```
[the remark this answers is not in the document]
jimjbrettj: Yup for sure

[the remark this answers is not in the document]
jimjbrettj: I went with TestApplyExtrinsicAfterFirstBlockFinalized, but let me know if you have any suggestions

lib/babe/build_integration_test.go
jimjbrettj: Yeah think so. Perhaps this is another one I can investigate in follow up PR if you think its worth it

lib/babe/build_integration_test.go
jimjbrettj: nope, good catch

lib/babe/build_integration_test.go
jimjbrettj: I'm not entirely sure, there are a couple of things to consider.
1) Config and the VerificationManager were both used and weirdly coupled in the verification manager integrations test. My plan was in my follow-up PR to investigate this further and find out what we need and what we don't for the tests to work and be intuitive.
2) Even if we don't use it now, perhaps we would want to in the future. I think this still relates to my first point though

lib/babe/build.go
jimjbrettj: @kishansagathiya Upon further investigation, the slot.start value is set when the slot is activated, so the time will always be within the appropriate slot. So I think leaving it as is is okay since it allows easier testing and is still producing a correct timestamp.

https://github.com/ChainSafe/gossamer/blob/4ea0a703339ca06e617b06eeb7d56308e7b48093/lib/babe/babe.go#L485

kishansagathiya: > it allows easier testing 

Is there a test or a usecase that fails, if you use `time.Now()`

> is still producing a correct timestamp.

Sort of. slot duration is generally in seconds (right?). timestamp's accuracy is a lot more than that.

> the slot.start value is set when the slot is activated, so the time will always be within the appropriate slot

That's kinda what I have  a problem with. I don't think it's a good idea to forcefully match slot and timestamp. They should match naturally. Maybe, using `slot.start` is good enough and doesn't create any problems. But, we know that that is not what is being expected from the spec. And I don't know what kind of problems it might lead.

This change looks like a hack to (may be) make things easier for now. But unless there is a strong need for it, we shouldn't have such hacks in our codebase.
@timwu20

kishansagathiya (dismissed):
https://github.com/ChainSafe/gossamer/pull/3037#discussion_r1086217256

lib/babe/verify_integration_test.go
```diff
@@ -436,37 +425,47 @@ func TestVerifyAuthorshipRight_Equivocation(t *testing.T) {
...
 	}
 
-	babeService := createTestService(t, cfg)
+	genesis, genesisTrie, genesisHeader := newWestendDevGenesisWithTrieAndHeader(t)
+	babeService := createTestService(t, cfg, genesis, genesisTrie, genesisHeader)
 	epochData, err := babeService.initiateEpoch(testEpochIndex)
 	require.NoError(t, err)
 
+	bestBlockHash := babeService.blockState.BestBlockHash()
+	rt, err := babeService.blockState.GetRuntime(bestBlockHash)
```
qdm12, line 434: nit `rt` -> `runtime`

lib/babe/build.go
```diff
@@ -229,8 +229,9 @@ func (b *BlockBuilder) buildBlockExtrinsics(slot Slot, rt ExtrinsicHandler) []*t
 func buildBlockInherents(slot Slot, rt ExtrinsicHandler, parent *types.Header) ([][]byte, error) {
 	// Setup inherents: add timstap0
 	idata := types.NewInherentData()
-	timestamp := uint64(time.Now().UnixMilli())
-	err := idata.SetInherent(types.Timstap0, timestamp)
+	fmt.Println(slot.start.UnixMilli())
+	fmt.Println(slot.number)
```
qdm12, line 233: nit remove prints

```suggestion
```

lib/babe/build.go
qdm12: maybe

```suggestion
			logger.Warnf("determining apply extrinsic call error: %s", err)
```

lib/babe/build_integration_test.go
```diff
@@ -366,28 +331,27 @@ func TestDecodeExtrinsicBody(t *testing.T) {
...
-	}
-
-	babeService := createTestService(t, cfg)
+	genesis, genesisTrie, genesisHeader := newWestendDevGenesisWithTrieAndHeader(t)
+	babeService := createTestService(t, ServiceConfig{}, genesis, genesisTrie, genesisHeader)
 
 	parent, err := babeService.blockState.BestBlockHeader()
 	require.NoError(t, err)
 
+	rt, err := babeService.blockState.GetRuntime(parent.Hash())
```
qdm12, line 344: nit `rt` -> `runtime`

lib/babe/helpers_test.go
```diff
@@ -4,28 +4,56 @@
...
+	"github.com/ChainSafe/gossamer/pkg/scale"
 	"github.com/golang/mock/gomock"
 	"github.com/stretchr/testify/require"
 )
 
+var (
+	defaultTestLogLvl = log.Info
+	emptyHash         = trie.EmptyHash
+	testEpochIndex    = uint64(0)
+	maxThreshold      = scale.MaxUint128
```
qdm12, line 36: since these are 'new' deltas, maybe have these as constants?

lib/babe/helpers_test.go
```diff
@@ -133,12 +159,209 @@ func NewTestService(t *testing.T, cfg *core.Config) *core.Service {
 	return s
 }
 
-// newDevGenesisWithTrieAndHeader generates test dev genesis, genesis trie and genesis header
-func newDevGenesisWithTrieAndHeader(t *testing.T) (
+func createTestService(t *testing.T, cfg ServiceConfig, genesis genesis.Genesis,
+	genesisTrie trie.Trie, genesisHeader types.Header) *Service {
+	wasmer.DefaultTestLogLvl = 1
```
qdm12, line 164: nit that's a copy paste but since it shows as new deltas, maybe set to the constant `log.Error` instead of `1`

lib/babe/helpers_test.go
```diff
@@ -133,12 +159,209 @@ func NewTestService(t *testing.T, cfg *core.Config) *core.Service {
...
+	}
+
+	ctrl := gomock.NewController(t)
+	telemetryMock := NewMockTelemetry(ctrl)
+	telemetryMock.EXPECT().SendMessage(gomock.Any()).AnyTimes()
+
+	cfg.Telemetry = telemetryMock
+
+	testDatadirPath := t.TempDir()
+	require.NoError(t, err)
```
qdm12, line 187:
```suggestion
```

lib/babe/helpers_test.go
```diff
@@ -133,12 +159,209 @@ func NewTestService(t *testing.T, cfg *core.Config) *core.Service {
...
 }
 
-// newDevGenesisWithTrieAndHeader generates test dev genesis, genesis trie and genesis header
-func newDevGenesisWithTrieAndHeader(t *testing.T) (
+func createTestService(t *testing.T, cfg ServiceConfig, genesis genesis.Genesis,
+	genesisTrie trie.Trie, genesisHeader types.Header) *Service {
+	wasmer.DefaultTestLogLvl = 1
+
+	var err error
+
```
qdm12, line 167: Remove this and change below `err = dbSrv.Initialise(&genesis, &genesisHeader, &genesisTrie)` to `err := dbSrv.Initialise(&genesis, &genesisHeader, &genesisTrie)`

lib/babe/helpers_test.go
```diff
@@ -133,12 +159,209 @@ func NewTestService(t *testing.T, cfg *core.Config) *core.Service {
...
+	testDatadirPath := t.TempDir()
+	require.NoError(t, err)
+
+	var dbSrv *state.Service
+	config := state.Config{
+		Path:      testDatadirPath,
+		LogLevel:  log.Info,
+		Telemetry: telemetryMock,
+	}
+	dbSrv = state.NewService(config)
```
qdm12, line 195: nit

```suggestion
	config := state.Config{
		Path:      testDatadirPath,
		LogLevel:  log.Info,
		Telemetry: telemetryMock,
	}
	dbSrv := state.NewService(config)
```

lib/babe/helpers_test.go
```diff
@@ -133,12 +159,209 @@ func NewTestService(t *testing.T, cfg *core.Config) *core.Service {
...
+
+	t.Cleanup(func() {
+		_ = dbSrv.Stop()
+	})
+
+	rtCfg := wasmer.Config{
+		Storage: rtstorage.NewTrieState(&genesisTrie),
+	}
+
+	rt, err := wasmer.NewRuntimeFromGenesis(rtCfg)
```
qdm12, line 285: nit `rt` -> `runtime`

lib/babe/helpers_test.go
```diff
@@ -133,12 +159,209 @@ func NewTestService(t *testing.T, cfg *core.Config) *core.Service {
...
+func createTestBlockWithSlot(t *testing.T, babeService *Service, parent *types.Header,
+	exts [][]byte, epoch uint64, epochData *epochData, slot Slot) *types.Block {
+	for _, ext := range exts {
+		vtx := transaction.NewValidTransaction(ext, &transaction.Validity{})
+		_, err := babeService.transactionState.Push(vtx)
+		require.NoError(t, err)
+	}
+
+	bestBlockHash := babeService.blockState.BestBlockHash()
+	rt, err := babeService.blockState.GetRuntime(bestBlockHash)
```
qdm12, line 346: nit `rt` -> `runtime`

lib/babe/helpers_test.go
```diff
@@ -133,12 +159,209 @@ func NewTestService(t *testing.T, cfg *core.Config) *core.Service {
...
+		duration: time.Duration(babeConfig.SlotDuration) * time.Millisecond,
+		number:   currentSlot,
+	}
+}
+
+func createTestBlockWithSlot(t *testing.T, babeService *Service, parent *types.Header,
+	exts [][]byte, epoch uint64, epochData *epochData, slot Slot) *types.Block {
+	for _, ext := range exts {
+		vtx := transaction.NewValidTransaction(ext, &transaction.Validity{})
+		_, err := babeService.transactionState.Push(vtx)
```
qdm12, line 341: nit `vtx` -> `validTransaction`

```suggestion
		validTransaction := transaction.NewValidTransaction(ext, &transaction.Validity{})
		_, err := babeService.transactionState.Push(validTransaction)
```

lib/babe/helpers_test.go
```diff
@@ -152,7 +375,8 @@ func newDevGenesisWithTrieAndHeader(t *testing.T) (
 	return gen, genesisTrie, genesisHeader
 }
 
-func newTestGenesisWithTrieAndHeader(t *testing.T) (
+// newWestendDevGenesisWithTrieAndHeader generates westend genesis, genesis trie and genesis header
```
qdm12, line 378: supernit

```suggestion
// newWestendDevGenesisWithTrieAndHeader returns the westend genesis, genesis trie and genesis header
```

lib/babe/verify_integration_test.go
```diff
@@ -246,32 +221,29 @@ func TestVerificationManager_VerifyBlock_Secondary(t *testing.T) {
...
+	vm := NewVerificationManager(babeService.blockState, babeService.epochState)
 
 	bestBlockHash := babeService.blockState.BestBlockHash()
 	rt, err := babeService.blockState.GetRuntime(bestBlockHash)
 	require.NoError(t, err)
 
-	cfg, err := rt.BabeConfiguration()
-	require.NoError(t, err)
+	//babeCfg, err := rt.BabeConfiguration()
+	//require.NoError(t, err)
```
qdm12, line 242: Remove?

lib/babe/verify_integration_test.go
```diff
@@ -284,28 +256,35 @@ func TestVerificationManager_VerifyBlock_MultipleEpochs(t *testing.T) {
...
-		cfg.EpochLength*futureEpoch+1, futureEpoch, futureEpochData)
-	block2 := createTestBlock(t, babeService, &block1.Header, [][]byte{},
-		cfg.EpochLength*futureEpoch+2, futureEpoch, futureEpochData)
+	slot1 := getSlot(t, rt, time.Now())
+	//slot1.number = babeCfg.EpochLength*futureEpoch + 1
+	block1 := createTestBlockWithSlot(t, babeService, &genesisHeader, [][]byte{}, futureEpoch, futureEpochData, slot1)
+
+	slot2 := getSlot(t, rt, time.Now())
+	//slot2.number = babeCfg.EpochLength*futureEpoch + 2
+	block2 := createTestBlockWithSlot(t, babeService, &block1.Header, [][]byte{}, futureEpoch, futureEpochData, slot2)
```
qdm12, line 265: Remove commented code?

lib/babe/verify_integration_test.go
```diff
@@ -284,28 +256,35 @@ func TestVerificationManager_VerifyBlock_MultipleEpochs(t *testing.T) {
...
 
 	err = vm.VerifyBlock(&block2.Header)
 	require.NoError(t, err)
 
 	// create block in epoch 1
-	block := createTestBlock(t, babeService, genesisHeader, [][]byte{}, cfg.EpochLength-10, testEpochIndex, epochData)
+	slot := getSlot(t, rt, time.Now())
+	//slot1.number = babeCfg.EpochLength-10
```
qdm12, line 272: Remove commented code?

lib/babe/verify_integration_test.go
```diff
@@ -397,24 +384,28 @@ func TestVerifyPimarySlotWinner(t *testing.T) {
...
 func TestVerifyAuthorshipRight(t *testing.T) {
 	serviceConfig := ServiceConfig{
 		Authority: true,
 	}
-	babeService := createTestService(t, serviceConfig)
+	genesis, genesisTrie, genesisHeader := newWestendLocalGenesisWithTrieAndHeader(t)
+	babeService := createTestService(t, serviceConfig, genesis, genesisTrie, genesisHeader)
+
+	bestBlockHash := babeService.blockState.BestBlockHash()
+	rt, err := babeService.blockState.GetRuntime(bestBlockHash)
```
qdm12, line 400: nit `rt` -> `runtime`

lib/babe/verify_integration_test.go
jimjbrettj: I actually kept this one intentionally to use as reference in my follow up PR

PR #3042, opened 2023-01-12
fix(tests/rpc): ensure new blocks are created before assert

### LINKED ISSUE #3041: Fix Flaky Test `TestChainRPC`
## Description

The test `TestChainRPC` waits for a block with a number greater than `0` or `1` to be finalized. When the block is finalized this test calls the `chain_getBlock` RPC endpoint to retrieve the full block header and calls `chain_getBlock` again but this time to retrieve the parent block full header. 

The last step in the test is to call `chain_getBlockHash` and assert that the returned hash is not equal to the finalized hash retrieved in the very first step, this last part is flaky since it is not guaranteed that when we call `chain_getBlockHash` a new block ahead of the finalized one was produced

EclesioMeloJunior:
## Changes

- Wait for Gossamer to produce a new block ahead of the finalized one

## Tests

```sh
MODE=rpc go test -timeout 10m -run ^TestChainRPC$ github.com/ChainSafe/gossamer/tests/rpc --tags=integration -v
```

## Issues

- Closes #3041

## Primary Reviewer

@jimjbrettj

### DIFF
--- a/tests/rpc/rpc_03-chain_test.go
+++ b/tests/rpc/rpc_03-chain_test.go
@@ -124,11 +124,7 @@ func TestChainRPC(t *testing.T) {
 	err = retry.UntilOK(ctx, retryWaitDuration, func() (ok bool, err error) {
 		var blockHash string
 		fetchWithTimeout(ctx, t, "chain_getBlockHash", "[]", &blockHash)
-
-		if finalizedHash == blockHash {
-			return false, nil
-		}
-		return true, nil
+		return finalizedHash != blockHash, nil
 	})
 	require.NoError(t, err)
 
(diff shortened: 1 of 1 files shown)


tests/rpc/rpc_03-chain_test.go
```diff
@@ -112,17 +113,29 @@ func TestChainRPC(t *testing.T) {
...
 
+	// Wait for Gossamer to produce a new block ahead of the finalized one
+	err = retry.UntilOK(ctx, retryWaitDuration, func() (ok bool, err error) {
+		var blockHash string
+		fetchWithTimeout(ctx, t, "chain_getBlockHash", "[]", &blockHash)
+
+		if finalizedHash == blockHash {
+			return false, nil
+		}
+		return true, nil
```
qdm12, line 131: nit

```suggestion
		return finalizedHash != blockHash, nil
```

qdm12 (approved):
Awesome 🎉 That drove me mad and didn't get the time to fix it, thanks!! 👍 🎖️

jimjbrettj (approved):
Good work!

PR #3044, opened 2023-01-13
chore(tests): test cases spaces to underscores

### LINKED ISSUE #3039: Batch replace all test case strings to have underscores
`go test` logs out subtest names converting spaces to underscores.
It's convenient to copy paste a failed test in your search to find the matching subtest that failed.
However, since for now we use spaces and not underscores, this is troublesome and most likely cost us quite a bit of development time.

1. We should replace all spaces with underscores matching the regex `\t".+ .+": \{` in `*_test.go` files. That might be done with standard Linux utilities, or we can write a tiny Go program to walk files.
2. Using 1., a CI step should ensure we don't add test cases using spaces, and use underscores instead

qdm12:
## Changes

(Sorry got mad at replacing myself test case spaces in my own PRs, now we have a program to do it)

- Add program in `cmd/testcases/main.go`
- Add CI step running the testcases program
- Replace all spaces in test cases with underscores

## Issues

Fixes #3039 

## Primary Reviewer

@EclesioMeloJunior


EclesioMeloJunior (approved):

timwu20:
why do we need to change test names to underscores? Doesn't the go testing framework already do that?

qdm12:
> Doesn't the go testing framework already do that when displaying results?

See https://github.com/ChainSafe/gossamer/issues/3039 it explains why (credits to Eclesio).

jimjbrettj (approved):

PR #3046, opened 2023-01-16
fix(ci): fix all Deepsource issues

qdm12:
## Tests

```sh
```


EclesioMeloJunior (approved):

jimjbrettj (approved):

internal/pprof/settings.go
```diff
@@ -18,13 +18,13 @@ type Settings struct {
 	MutexProfileRate int
 }
 
-func (s *Settings) setDefaults() {
+func (s *Settings) setDefaults() { //skipcq: GO-W1029
```
timwu20, line 21: why add the skips if we can just conform to pointer receivers if we need it?

qdm12: Deepsource complains if we mix pointer and value receivers for the same type.
I like to use value receivers to highlight we don't modify the receiving variable / we don't use the receiving variable (example VDT Index method). Any thoughts on that? I don't mind changing everything to pointer receivers though (since a value receiver can still be modified if it contains a pointer).

timwu20: I think if deepsource is gonna complain about mixing both, we should just conform to one.  I used to mix them previously, to show what is mutating the state, but kinda unnecessary thinking about it more.  I think using pointer receivers when we never mutate state is unnecessary, but if there are methods that do, it makes sense to me to make them all pointer receivers.

qdm12: After a bit of investigation, value receivers got changed to pointer receivers for:

- `VaryingDataTypeSlice`'s `String` method
- `VaryingDataType`'s `String` method
- Trie `Node`'s `Copy` method (no impact performance wise)
- pprof `Settings`'s `String` method

BUT some value receivers are kept (mixed with pointer receivers) in the following:

- VDTs `Index()` (and `String()` since if we have 1 mixed receiver it's the same as 2 for Deepsource) methods, since otherwise we need to pass pointers instead of values of types as VDTs and that's a lot of code change and perhaps unnecessary complexity
- Some methods for `common.Hash` since these are used all around the place, and dereferencing them may impact performance somehow
- `lib/transaction`'s `priorityQueue` since it may have a performance impact to dereference (not 100% sure, but it's just 3 deepsource skip comments)

- Keep value receivers for VDTs, `common.Hash`, `transaction.priorityQueue`

timwu20 (approved):

PR #3047, opened 2023-01-16
fix(lib/runtime): use `westend-dev` spec file in `TestNodeRuntime_ValidateTransaction`

EclesioMeloJunior:
## Changes

- Replace `GetGssmrV3SubstrateGenesisRawPathTest` for `GetWestendDevRawGenesisPath`

## Tests

N/A

## Issues

- Ref: https://github.com/chainsafe/gossamer/issues/3015

## Primary Reviewer

@jimjbrettj

PR #3048, opened 2023-01-16
feat(dot/types): moves `ParachainInherentData` type to `types` package

EclesioMeloJunior:
## Changes

- Extract `ParachainInherentData` type from `lib/babe` package and moves it to `dot/types` package, this is needed since there is a bunch of tests that need it but cannot import due to cyclic dependency.

## Tests

N/A

## Issues

- Unblocks: https://github.com/ChainSafe/gossamer/pull/3047

## Primary Reviewer

@timwu20
@qdm12

### DIFF
--- a/dot/types/parachain_inherents.go
+++ b/dot/types/parachain_inherents.go
@@ -1,12 +1,11 @@
 // Copyright 2022 ChainSafe Systems (ON)
 // SPDX-License-Identifier: LGPL-3.0-only
 
-package babe
+package types
 
 import (
 	"fmt"
 
-	"github.com/ChainSafe/gossamer/dot/types"
 	"github.com/ChainSafe/gossamer/lib/common"
 	"github.com/ChainSafe/gossamer/pkg/scale"
 )
@@ -379,5 +378,5 @@ type ParachainInherentData struct {
 	// Sets of dispute votes for inclusion,
 	Disputes multiDisputeStatementSet `scale:"3"`
 	// The parent block header. Used for checking state proofs.
-	ParentHeader types.Header `scale:"4"`
+	ParentHeader Header `scale:"4"`
 }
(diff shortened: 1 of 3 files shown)


qdm12 (approved):
👍 Although I would instead have it in the `interrnal/` directory to limit the external usage of these types from outside repositories.

Crossing my mind are:

- `internal/babe/parachain`
- `internal/babe/inherents`
- `internal/primitives` (=internal replacement for `dot/types`)
- `internal/primitives/parachain`
- `internal/primitives/inherents`

If not, maybe it would be the time to separate things in the `dot/types` package and add a `dot/types/parachain` package, to reduce the tight coupling/dependency contention.

kishansagathiya:
I like `lib/babe/types` or `lib/babe/inherents` if that solves your cyclical dependencies.

kishansagathiya (approved):
you could have just added this change as part of https://github.com/ChainSafe/gossamer/pull/3047

That way reviewer would have more context of circular dependency

qdm12:
> you could have just added this change as part of #3047
> 
> That way reviewer would have more context of circular dependency

Good point, @EclesioMeloJunior does #3047 modify these files? If not we can move those changes to #3047 and close this PR? File movements will be tracked correctly even if commits are squashed, as long as we don't modify these babe files too much.


> I like `lib/babe/types` or `lib/babe/inherents` if that solves your cyclical dependencies.

- In `./internal/` would be better in order to hide from public before our Go API is stable / needs to be used outside
- I like `lib/babe/inherents`
- We should stop throwing moareee in the the kitchen sink aka `dot/types`

EclesioMeloJunior:
I will include these changes in https://github.com/ChainSafe/gossamer/pull/3047, and I will move these files to `lib/babe/inherents`

PR #3050, opened 2023-01-17
feat(lib/blocktree): `Prune` returns ordered pruned forks

qdm12:
## Changes

The blocktree `Prune` method returns a mapping from fork origin block hash (part of the canonical chain) to an ordered slice of block hashes forming the now-dropped forked chain of blocks.

This allows the pruner to be 'fork-aware' and to find which deltas can be deleted safely from each fork, on finalisation. For example if a dropped fork chain inserted database key XYZ, we should remove it BUT we need to ensure this same key was not inserted by a block on the canonical chain as well. To do this, we need all the dropped forks ordered to work with their deltas and compare them with the canonical chain.

## Tests

```sh
```

## Issues

Needed for PR #2831

## Primary Reviewer

@timwu20


dot/state/block_finalisation.go
```diff
@@ -142,16 +142,20 @@ func (bs *BlockState) SetFinalisedHash(hash common.Hash, round, setID uint64) er
 		bs.notifyFinalized(hash, round, setID)
 	}
 
-	pruned := bs.bt.Prune(hash)
-	for _, hash := range pruned {
-		blockHeader := bs.unfinalisedBlocks.delete(hash)
-		if blockHeader == nil {
-			continue
-		}
+	forkOriginToChain := bs.bt.Prune(hash)
```
EclesioMeloJunior, line 145: what means this `forkOriginToChain`?

qdm12: It's a mapping from the fork origin (block hash of the block on the canonical chain where the fork originates from) to the forked chain (slice of block hashes corresponding to the blocks of the forked chain, from lower block number to higher). Would you foresee a clearer name (maybe a bit longer is fine)?

timwu20:
@qdm12 is this still required given we're removing the online pruner?

qdm12:
> @qdm12 is this still required given we're removing the online pruner?

Yes it's required for #2831 - I can bundle it in #2831 if you prefer though.

EclesioMeloJunior (approved):

PR #3070, opened 2023-01-25
fix(lib/babe): use slot start time in handleSlot

jimjbrettj:
## Changes

- Pass in slot to slot handler rather than declare it in the function
- Instead of using time.Now() use the slot start time as the start field of Slot

## Tests

```sh
go test -tags integration lib/babe
```

## Primary Reviewer

@EclesioMeloJunior

### DIFF
--- a/lib/babe/babe.go
+++ b/lib/babe/babe.go
@@ -478,19 +478,13 @@ func (b *Service) getParentForBlockAuthoring(slotNum uint64) (*types.Header, err
 	return parent, nil
 }
 
-func (b *Service) handleSlot(epoch, slotNum uint64,
+func (b *Service) handleSlot(epoch uint64, slot Slot,
 	authorityIndex uint32,
 	preRuntimeDigest *types.PreRuntimeDigest,
 ) error {
-	currentSlot := Slot{
-		start:    time.Now(),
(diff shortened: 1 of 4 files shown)


EclesioMeloJunior (approved):

lib/babe/babe.go
```diff
@@ -478,19 +478,13 @@ func (b *Service) getParentForBlockAuthoring(slotNum uint64) (*types.Header, err
...
-		start:    time.Now(),
-		duration: b.constants.slotDuration,
-		number:   slotNum,
-	}
-
-	parent, err := b.getParentForBlockAuthoring(slotNum)
+	parent, err := b.getParentForBlockAuthoring(slot.number)
 	if err != nil {
-		return fmt.Errorf("could not get parent for claiming slot %d: %w", slotNum, err)
+		return fmt.Errorf("could not get parent for claiming slot %d: %w", slot.number, err)
```
qdm12, line 487: nit whilst we're at it:

```suggestion
		return fmt.Errorf("getting parent for block authoring for claiming slot %d: %w", slot.number, err)
```

qdm12 (approved):
👍

Issue #3066, opened 2023-01-24
dot/state: load older canonical chain runtimes from storage where needed


qdm12:
Blocked by #2767 which now cleans up old runtimes from memory (on finalisation, removes all runtimes different from the finalised block runtime).

For certain RPC calls such as

https://github.com/ChainSafe/gossamer/blob/4ea0a703339ca06e617b06eeb7d56308e7b48093/dot/rpc/modules/payment.go#L40

And possibly babe:

https://github.com/ChainSafe/gossamer/blob/4ea0a703339ca06e617b06eeb7d56308e7b48093/lib/babe/babe.go#L506

There is a need to retrieve the runtime for an older block, that may have a runtime different from the finalised runtime we currently have in memory.
Runtimes are already stored in the `StorageState`, and can be retrieved already.

We should, for those uses above, check from memory first, and if it's not found, retrieve it from the storage state.
I would suggest we do NOT inject storage state to block state or block state to storage state, and just let the caller handle the not found error, to reduce dependency contention.

See #2781 which has some of the work already done (maybe not the last commit which inject storage state into block state, see the previous ones).

timwu20:
This is an optimization with no related milestone, but is ready to work on.

timwu20:
@jimjbrettj can you put this in an appropriate epic.

PR #3071, opened 2023-01-26
fix(blocktree): stop and prune runtimes from memory

### LINKED ISSUE #2767: Clean Up Old Runtimes on Canonical Chain

qdm12:
## Changes

- [x] Stop runtimes from forks before pruning them
- [x] Stop and prune runtimes from the canonical chain, on finalisation for runtimes different than the finalised block runtime
- [x] Add unit test for the blocktree `Prune` method

Note this PR breaks RPC calls `payment_Queryinfo`, `runtime_getVersion` and `runtime_getMetadata` which no longer have access to historical runtimes.

## Tests

```sh
```

## Issues

Fixes #2767 

Sibling issue to fix next is #3066 to re-establish the 3 RPC calls working for blocks older than the last finalised block.

## Primary Reviewer

@timwu20

### DIFF
--- a/lib/blocktree/blocktree.go
+++ b/lib/blocktree/blocktree.go
@@ -262,19 +262,15 @@ func (bt *BlockTree) Prune(finalised Hash) (pruned []Hash) {
 
 	// Cleanup in-memory runtimes from the canonical chain.
 	// The runtime used in the newly finalised block is kept
-	// instantiated in memory, and other runtimes from the
-	// previous finalised block to the newly finalised block
-	// included are stopped and removed from memory. Note
-	// these are still accessible through the storage as WASM blob.
-	finalisedBlockRuntime := bt.runtimes.get(finalised)
-	canonicalChainBlock := n
+	// instantiated in memory, and all other runtimes are
+	// stopped and removed from memory. Note these are still
+	// accessible through the storage as WASM blob.
 	previousFinalisedBlock := bt.root
-	for canonicalChainBlock != previousFinalisedBlock.parent {
-		runtime := bt.runtimes.get(canonicalChainBlock.hash)
-		if runtime != finalisedBlockRuntime {
-			runtime.Stop()
-			bt.runtimes.delete(canonicalChainBlock.hash)
-		}
+	newCanonicalChainBlocksCount := n.number - previousFinalisedBlock.number
+	canonicalChainBlock := n
+	newCanonicalChainBlockHashes := make([]Hash, newCanonicalChainBlocksCount)
+	for i := int(newCanonicalChainBlocksCount) - 1; i >= 0; i-- {
+		newCanonicalChainBlockHashes[i] = canonicalChainBlock.hash
 		canonicalChainBlock = canonicalChainBlock.parent
 	}
 
@@ -288,13 +284,7 @@ func (bt *BlockTree) Prune(finalised Hash) (pruned []Hash) {
 		bt.leaves.store(leaf.hash, leaf)
 	}
 
-	for _, hash := range pruned {
-		runtimeFromFork := bt.runtimes.get(hash)
-		if runtimeFromFork != finalisedBlockRuntime {
-			runtimeFromFork.Stop()
-		}
-		bt.runtimes.delete(hash)
-	}
+	bt.runtimes.onFinalisation(newCanonicalChainBlockHashes, pruned)
 
 	leavesGauge.Set(float64(len(bt.leaves.nodes())))
 	return pruned
--- a/lib/blocktree/blocktree.go
+++ b/lib/blocktree/blocktree.go
@@ -289,8 +289,10 @@ func (bt *BlockTree) Prune(finalised Hash) (pruned []Hash) {
 	}
 
 	for _, hash := range pruned {
-		instance := bt.runtimes.get(hash)
-		instance.Stop()
+		runtimeFromFork := bt.runtimes.get(hash)
+		if runtimeFromFork != finalisedBlockRuntime {
+			runtimeFromFork.Stop()
+		}
 		bt.runtimes.delete(hash)
 	}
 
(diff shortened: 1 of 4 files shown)


kishansagathiya (changes requested):
Ran this with on westend chain. I see size of `bt.runtime` getting increased as blocks get imported. imported a good number of blocks, but it doesn't looks like runtimes are getting pruned

```
number of runtimes stored: 26238
2023-02-10T19:18:53+05:30 DEBUG    imported block 0xc179598e8992f82d9bf88afb690d8e7f8e65d002a3e336c87f980cc7c1c4cd50 and stored state trie with root 0x5a2ac0194ba3f39789bd212ebf3cee4bd6398bdde1d7223f5e54988a1f7b6d36	service.go:L212	pkg=core
number of runtimes stored: 26238
2023-02-10T19:18:53+05:30 DEBUG    🔗 imported block number 111742 with hash 0xc179598e8992f82d9bf88afb690d8e7f8e65d002a3e336c87f980cc7c1c4cd50	chain_processor.go:L272	pkg=sync
2023-02-10T19:18:53+05:30 DEBUG    block with hash 0xc179598e8992f82d9bf88afb690d8e7f8e65d002a3e336c87f980cc7c1c4cd50 processed	chain_processor.go:L144	pkg=sync
2023-02-10T19:18:53+05:30 DEBUG    processing block data with hash 0xe24646cd5c86b20a5c1c62d0e48f077019409f8feba9458fdfab947c79b0ef4b	chain_processor.go:L115	pkg=sync
number of runtimes stored: 26239
2023-02-10T19:18:53+05:30 DEBUG    imported block 0xe24646cd5c86b20a5c1c62d0e48f077019409f8feba9458fdfab947c79b0ef4b and stored state trie with root 0x2fb5be18eadcbdbeddaf57c4103b0f48ff3c4346d1381344428e11d3609a7e98	service.go:L212	pkg=core
number of runtimes stored: 26239
2023-02-10T19:18:53+05:30 DEBUG    🔗 imported block number 111743 with hash 0xe24646cd5c86b20a5c1c62d0e48f077019409f8feba9458fdfab947c79b0ef4b	chain_processor.go:L272	pkg=sync
2023-02-10T19:18:53+05:30 DEBUG    block with hash 0xe24646cd5c86b20a5c1c62d0e48f077019409f8feba9458fdfab947c79b0ef4b processed	chain_processor.go:L144	pkg=sync
2023-02-10T19:18:53+05:30 DEBUG    processing block data with hash 0x1d5f85989e3ca6afc12f30434877a9646879496780527a9ce3e5cdd5046c9da2	chain_processor.go:L115	pkg=sync
number of runtimes stored: 26240
```

I am printing following in `(bt *BlockTree) GetBlockRuntime`
```
	fmt.Printf("number of runtimes stored: %d\n", len(bt.runtimes.mapping))
```

qdm12:
This should now work, and it has tests with it.

@kishansagathiya can you try running it until there is a runtime upgrade? Pruning of the runtimes mapping and stopping of runtime(s) only happen after a runtime upgrade, so make sure such upgrade happens before checking the mapping gets trimmed down. I gotta run and didn't have time to sync for more than a few thousands blocks without runtime upgrade.

There is `TestHandler_GrandpaScheduledChange` which fails due to `runtime not found for finalised block hash 0x68bf703270f01d73e6bc47aa911d9f6e5c62148e99c8f63c53c29f68b4db2519`, but this looks like an error in that existing test, maybe someone can have a look to fix it? I don't think it should be possible to get a finalised block hash without previously having set the mapping block hash <-> runtime pointer right?

EclesioMeloJunior:
> There is TestHandler_GrandpaScheduledChange which fails due to runtime not found for finalised block hash 0x68bf703270f01d73e6bc47aa911d9f6e5c62148e99c8f63c53c29f68b4db2519, but this looks like an error in that existing test, maybe someone can have a look to fix it?

@qdm12 I will take a look on it

kishansagathiya:
> Pruning of the runtimes mapping and stopping of runtime(s) only happen after a runtime upgrade, so make sure such upgrade happens before checking the mapping gets trimmed down.

I am not sure at what block do I see the first runtime. So far I have ran it till only 25k blocks and  number of saves runtimes are also 25k. This still stores a lot of runtimes in memory that we don't need. Can we not remove them before runtime upgrade or store them cleverly such that it takes less memory?

EclesioMeloJunior:
The test `TestHandler_GrandpaScheduledChange` does not load a runtime what means that the `blocktree.runtimes` map is empty, so when the function `onFinalisation` is called it panics since there is no entry for the given block hash, I raised the PR to address it -> https://github.com/ChainSafe/gossamer/pull/3142

qdm12:
Sorry for the delayed response;

> So far I have ran it till only 91k blocks and number of saves runtimes are also 91k. This still stores a lot of runtimes in memory that we don't need. Can we not remove them before runtime upgrade or store them cleverly such that it takes less memory?

It only stores the mapping block hash <-> runtime pointer, so not a big deal. And we need this mapping, ideally from memory since we lookup a lot from it. It doesn't store a separate runtime instance for each entry. So really per entry it's 32B (block hash) + 8B (runtime pointer) + map overhead, so even a few million entries is fine. That was discussed previously (I think at standup?) if it was okay to keep this map growing, especially if we have no runtime upgrade for millions of blocks. I think we can address having a more performant system (maybe on disk + lru cache) once we encounter a problem later down the line (we might never do at least on westend/kusama/polkadot).

> For blocks where runtime is not changing, can we use the same runtime instance for all those blocks?

This is the case. We only have one instantiated runtime from the last finalised block, and may have more instantiated runtimes until the next finalisation (i.e. 2 more runtimes if we do a runtime upgrade on 2 forks).

> so it is possible that for a new finalized block a mapping entry could be not there.

Are you sure? Don't we add the mapping for every handled block??

EclesioMeloJunior:
@timwu20 I think we can close this PR

PR #3059, opened 2023-01-23
Fixing descendant search

kishansagathiya:
If blocktree fails to get highest finalised hash (which is stored in memory), load it from disk.

## Tests
- Run gossamer on westend network and see that block importing never stops, (westend's genesis spec `polkadot/node/service/chain-specs/westend.json`) 

```sh
go test -tags integration github.com/ChainSafe/gossamer
```

## Issues
Fixes #3053

## Primary Reviewer

@timwu20

PR #3081, opened 2023-01-30
fix(lib/grandpa):  on verifying block justification, compare given block hash with finalised hash

### LINKED ISSUE #3061: failed to get runtime instance
## Describe the bug

- on restarting gossamer node on westend this happens
```
2023-01-24T18:32:49+05:30 ERROR    block data processing for block with hash 0xe0c8567481cb548652db81ff7715996ce97656d9027ce4e45e047d3a85241472 failed: processing block data with header and body: handling block: failed to get runtime instance	chain_processor.go:L96	pkg=sync
```
This prevents us from importing blocks

- On restarting gossamer importing did not work for any block
- I saw this on restarting after importing some 200k block.
- I have tried restarting after 13k blocks and did not see this error. So, it looks like this error doesn't happen in all cases.
- looks like this error starts to occur at 198656 block number
## Expected Behavior

-

## Current Behavior

-

## Possible Solution

-

## To Reproduce
Steps to reproduce the behaviour:

1. 
2. 
3. 

## Log output

https://gist.github.com/kishansagathiya/ebe4448aca406776866c58826e1b9567
## Specification

- go version:
- gossamer version:
- gossamer commit tag:
- gossamer commit hash:
- operating system:
- additional links:

kishansagathiya:
## Changes
So far, while verifying block justification in grandpa, we were just checking if we already know a finalised block that has the same set id and round as the block in question.

we should check if the finalised block that we know already and the block that we are verifying are not the same and not throw error in this case.

We should through error if we see two finalised blocks with the same set id and round. In this case, both blocks that we were comparing were the same. We were not comparing their hashes and were throwing an error thinking that we have two finalised blocks in the same set id and round.

After this change, I was able to sync further than 198656.

## Tests
Merge https://github.com/ChainSafe/gossamer/pull/3059 into this branch on testing, otherwise you will end up seeing that error.

this is zip of my basepath https://1drv.ms/u/s!AgmhJ6kdAKF3hHqBnSfpKialBZYC?e=VdPb6q
Use this and run gossamer on westend network. 
- On development, syncing would fail with following error,
```
2023-01-24T18:32:49+05:30 ERROR    block data processing for block with hash 0xe0c8567481cb548652db81ff7715996ce97656d9027ce4e45e047d3a85241472 failed: processing block data with header and body: handling block: failed to get runtime instance	chain_processor.go:L96	pkg=sync
```
- On this branch syncing would continue after 198656

## Issues

Fixes #3061
## Primary Reviewer

@timwu20


lib/grandpa/message_handler.go
```diff
@@ -427,7 +427,13 @@ func (s *Service) VerifyBlockJustification(hash common.Hash, justification []byt
 	}
 
 	if has {
-		return nil, fmt.Errorf("already have finalised block with setID=%d and round=%d", setID, fj.Round)
+		finalisedHeader, err := s.blockState.GetFinalisedHeader(fj.Round, setID)
```
EclesioMeloJunior, line 430: The function `GetFinalisedHeader` performs more operations (retrieve the hash, decode it, build the header struct) than the method `GetFinalisedHash` which basically does a database call to get the stored hash.
```suggestion
		storedFinalisedHash, err := s.blockState.GetFinalisedHash(fj.Round, setID)
		if err != nil {...}
		if storedFinalisedHash != hash {...}
```

lib/grandpa/message_handler.go
```diff
@@ -427,7 +427,13 @@ func (s *Service) VerifyBlockJustification(hash common.Hash, justification []byt
...
+		if err != nil {
+			return nil, fmt.Errorf("getting finalised header: %w", err)
+		}
+		if finalisedHeader.Hash() != hash {
+			return nil, fmt.Errorf("already have finalised block with setID=%d and round=%d", setID, fj.Round)
```
EclesioMeloJunior, line 435: Add a sentinel error here

lib/grandpa/message_handler.go
kishansagathiya: there is not function `s.blockState.GetFinalisedHash`

EclesioMeloJunior: That is because `s.blockState` is an interface that does not contains the method, but the implementation is here -> https://github.com/ChainSafe/gossamer/blob/9f19d317258c5f7d5128b3819ff2534dd4dbfbb9/dot/state/block_finalisation.go#L56

So, basically what you should do is introduce the method in the interface:

```go
type BlockState interface {
    ...
    GetFinalisedHash(round, setID uint64) (common.Hash, error)
}
```

and now you are able to use the method `s.blockState.GetFinalisedHash`

dot/state/block.go
```diff
@@ -250,14 +250,62 @@ func (bs *BlockState) GetHashByNumber(num uint) (common.Hash, error) {
 	return common.NewHash(bh), nil
 }
 
+// GetAllDescendants gets all the descendants even if hash is not stored in memory.
+func (bs *BlockState) GetAllDescendants(hash common.Hash) ([]common.Hash, error) {
```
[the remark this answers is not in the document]
kishansagathiya: removed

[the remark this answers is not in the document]
kishansagathiya: removed

EclesioMeloJunior (approved):

lib/grandpa/message_handler.go
```diff
@@ -427,7 +427,13 @@ func (s *Service) VerifyBlockJustification(hash common.Hash, justification []byt
...
 
 	if has {
-		return nil, fmt.Errorf("already have finalised block with setID=%d and round=%d", setID, fj.Round)
+		storedFinalisedHash, err := s.blockState.GetFinalisedHash(fj.Round, setID)
+		if err != nil {
+			return nil, fmt.Errorf("getting finalised hash: %w", err)
+		}
+		if storedFinalisedHash != hash {
+			return nil, fmt.Errorf("%w, setID=%d and round=%d", errFinalisedBlocksMismatch, setID, fj.Round)
+		}
```
timwu20, line 436: do we really want to continue scale encoding the justification and then calling `s.finalityGadget.VerifyBlockJustification` in `chainProcessor.handleJustification` if no error is returned? I'm assuming this has already been done in a previous call to `Service.VerifyBlockJustification`

timwu20:
Really good work @kishansagathiya ! Must've been a tough one to investigate.

lib/grandpa/message_handler.go
kishansagathiya: > do we really want to continue scale encoding the justification and then calling s.finalityGadget.VerifyBlockJustification

justification is already scale encoded at this point. I don't see a scale.Marshal before this point, so we already got it marshaled. 

I see that VerifyBlockJustification returns justification after scale encoding it again, which is not necessary. I have removed returning justification there. Is that what you meant?

timwu20: Let me put it this way.  Say we enter this `has == true` case so we have a finalised block for a given round and set id. and the `storedFinalisedHash == hash`.  Based on the current code we will exit this `if` block and continue with lines 439 to line 537 where we `s.blockState.SetFinalisedHash(hash, fj.Round, setID)`.  Is this really required given that we found the block already?

timwu20 (approved):

PR #3082, opened 2023-01-30
chore(types/digest): remove ChangesTrieRootDigest

kishansagathiya:
## Changes
This PR removes ChangesTrieRootDigest and introduces a HeaderDigest struct 

In polkadot spec, there is no mention of ChangesTrieRootDigest. I checked git blame, it was added 2 years ago. Chances are that specs changed since then.

https://spec.polkadot.network/#defn-digest

## Tests

```sh
```

## Primary Reviewer

@EclesioMeloJunior


qdm12 (approved):
Red is always (almost) good! 💯

EclesioMeloJunior (approved):

PR #3084, opened 2023-01-30
chore(dot/state): offline pruner changes

qdm12:
## Changes

- Directly prune input database, instead of streaming/copying it to a new directory.
- Do NOT use in-memory bloom filter, use a file-backed filter database instead
  - Bloom filter is probabilistic, some keys may not be pruned
  - Bloom filter is in memory, memory usage may be too high
- Fix input database path to have `/db` path suffix (like in other places of the codebase)

## Tests

```sh
```

## Issues

Greatly simplifies my life changing the database #3005 

## Primary Reviewer

@EclesioMeloJunior


dot/state/offline_pruner.go
```diff
@@ -160,54 +171,35 @@ func (p *OfflinePruner) Prune() error {
...
-	if err = writer.Prepare(); err != nil {
-		return fmt.Errorf("cannot create stream writer in out DB at %s error %w", p.prunedDBPath, err)
-	}
-
-	// Stream contents of DB to the output DB.
 	stream := inputDB.NewStream()
-	stream.LogPrefix = fmt.Sprintf("Streaming DB to new DB at %s ", p.prunedDBPath)
-
 	stream.ChooseKey = func(item *badger.Item) bool {
 		key := string(item.Key())
```
EclesioMeloJunior, line 176: wondering if it is a good use case for using the `[]byte <> string` conversion w/o allocation: https://go.dev/play/p/P__twwbKG0O

qdm12: NO 🤣

EclesioMeloJunior (approved):

jimjbrettj (approved):

PR #3085, opened 2023-01-30
ci: add more juicy github action runners

q9f:
## Changes

uses build jet external runners. should not make any difference (except for speed).

## Tests

n/a

## Issues

n/a

## Primary Reviewer

n/a

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -27,7 +27,7 @@ jobs:
     timeout-minutes: 60
     strategy:
       matrix:
-        platform: [macos-latest, ubuntu-latest]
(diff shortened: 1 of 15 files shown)


EclesioMeloJunior (approved):
🚀

jimjbrettj (approved):

qdm12 (approved):
@q9f is this something ChainSafe has started paying for? 🤔

q9f:
> @q9f is this something ChainSafe has started paying for? thinking

Yes, mainly for the Rust pipelines, but since the 4vCPU machines on Buildjet cost the same as the 2vCPU Github Actions runners, there is no need to not use the slow ones here.

Let me know if you think you need even more power somewhere.

PR #3086, opened 2023-01-30
feat(chain): remove unneeded spec files

EclesioMeloJunior:
## Changes

- Remove: `dev-staking/dev-v3substrate/dev`, `gssmr-staking/gssmr-v3substrate/gssmr`, `rococo-dev`

## Tests

```sh
N/A
```

## Primary Reviewer

@danforbes + @jimjbrettj


chain/westend_dev/defaults.go
```diff
@@ -1,11 +1,9 @@
 // Copyright 2021 ChainSafe Systems (ON)
 // SPDX-License-Identifier: LGPL-3.0-only
 
-package gssmr
+package westend_dev
```
jimjbrettj, line 4: nit: No underscores in package name

jimjbrettj (commented):
Looks great! Might be a few conflicts with my docker PR since changed folder name, but should be easy to fix. All looks good to me, will approve when tests are passing

chain/westend_dev/defaults.go
danforbes: Why are we using the underscores at all as opposed to hyphens?

danforbes (approved):
Amazing! You can also remove the rococo-dev one, I think.

README.md
```diff
@@ -151,29 +151,29 @@ and Charlie test accounts. In three separate terminals, initialize the data
 directories for the three Gossamer instances:
 
 ```sh
-./bin/gossamer init --force --config ./chain/westend-local/config-alice.toml
+./bin/gossamer init --force --config ./chain/westend_local/config-alice.toml
```
timwu20, line 154: why did we change to underscores for the folders? I like hyphens over underscores when it comes to folder names.

cmd/gossamer/config.go
```diff
@@ -718,10 +712,12 @@ func setDotRPCConfig(ctx *cli.Context, tomlCfg ctoml.RPCConfig, cfg *dot.RPCConf
 	cfg.WSUnsafeExternal = tomlCfg.WSUnsafeExternal
 
 	// check --rpc flag and update node configuration
-	if enabled := ctx.GlobalBool(RPCEnabledFlag.Name); enabled || cfg.Enabled {
-		cfg.Enabled = true
-	} else if ctx.IsSet(RPCEnabledFlag.Name) && !enabled {
-		cfg.Enabled = false
+	rpcFlagIsSet := ctx.IsSet(RPCEnabledFlag.Name)
```
jimjbrettj, line 715: Leaving here for reviewer, please double check this change and the one below are okay

dot/services_integration_test.go
```diff
@@ -726,6 +726,8 @@ func TestNewWebSocketServer(t *testing.T) {
 	cfg.Init.Genesis = genFile
 	cfg.RPC.External = false
 	cfg.RPC.WS = true
+	cfg.RPC.Port = 9545
```
jimjbrettj, line 729: Note for reviewer, ports were already in use in the CI causing this test to fail. Changed to bind to different port, but let me know if there is a better solution

jimjbrettj (approved):
Awesome work!!

chain/westend-dev/config.toml
```diff
@@ -25,12 +25,13 @@ unlock = ""
...
 babe-authority = true
 grandpa-authority = true
-babe-lead = true
+grandpa-interval = 1
 
 [network]
 port = 7001
 nobootstrap = false
 nomdns = false
+discovery-interval = 10
```
qdm12, line 34: Forgive my ignorance, but why these changes 🥲

cmd/gossamer/config.go
```diff
@@ -26,24 +24,22 @@ import (
...
-	defaultDevConfigPath      = "./chain/dev/config.toml"
-
-	gossamerName = "gssmr"
-	kusamaName   = "kusama"
-	polkadotName = "polkadot"
-	devName      = "dev"
+	DefaultCfg                  = dot.WestendDevConfig
+	defaultKusamaConfigPath     = "./chain/kusama/config.toml"
+	defaultPolkadotConfigPath   = "./chain/polkadot/config.toml"
+	defaultWestendDevConfigPath = "./chain/westen-dev/config.toml"
```
qdm12, line 30: typo this is meant to be

```suggestion
	defaultWestendDevConfigPath = "./chain/westend-dev/config.toml"
```

cmd/gossamer/config.go
```diff
@@ -26,24 +24,22 @@ import (
...
+
+	kusamaName     = "kusama"
+	polkadotName   = "polkadot"
+	westendDevName = "westend-dev"
```
qdm12, line 34: nit maybe change these few string constants to... you guessed it, constants, since it's all changed up now

cmd/gossamer/config.go
```diff
@@ -161,8 +153,10 @@ func createInitConfig(ctx *cli.Context) (*dot.Config, error) {
 		return nil, fmt.Errorf("--%s must be %s", PruningFlag.Name, pruner.Archive)
 	}
 
-	if cfg.Global.RetainBlocks < dev.DefaultRetainBlocks {
-		return nil, fmt.Errorf("--%s cannot be less than %d", RetainBlockNumberFlag.Name, dev.DefaultRetainBlocks)
+	const defaultRetainBlocks = uint32(512)
+
+	if cfg.Global.RetainBlocks < defaultRetainBlocks {
```
qdm12, line 158: nit rename to `minRetainBlocks`?

cmd/gossamer/config.go
```diff
@@ -625,10 +619,10 @@ func setDotCoreConfig(ctx *cli.Context, tomlCfg ctoml.CoreConfig, cfg *dot.CoreC
 	case wasmer.Name:
 		cfg.WasmInterpreter = wasmer.Name
 	case "":
-		cfg.WasmInterpreter = gssmr.DefaultWasmInterpreter
+		cfg.WasmInterpreter = wasmer.Name
 	default:
-		cfg.WasmInterpreter = gssmr.DefaultWasmInterpreter
-		logger.Warn("invalid wasm interpreter set in config, defaulting to " + gssmr.DefaultWasmInterpreter)
+		cfg.WasmInterpreter = wasmer.Name
+		logger.Warn("invalid wasm interpreter set in config, defaulting to " + wasmer.Name)
```
qdm12, line 625: nit/a bit OOS: we should error or panic (maybe if this isn't user input error) here

cmd/gossamer/config.go
qdm12: LGTM 👍

cmd/gossamer/config_test.go
```diff
@@ -41,29 +42,24 @@ func TestConfigFromChainFlag(t *testing.T) {
...
-			"Test gossamer --chain gssmr",
-			[]string{"chain", "name", "pruning", "retain-blocks"},
-			[]interface{}{"gssmr", dot.GssmrConfig().Global.Name, gssmr.DefaultPruningMode, gssmr.DefaultRetainBlocks},
-			dot.GssmrConfig(),
-		},
 		{
 			"Test gossamer --chain kusama",
 			[]string{"chain", "name", "pruning", "retain-blocks"},
-			[]interface{}{"kusama", dot.KusamaConfig().Global.Name, gssmr.DefaultPruningMode, gssmr.DefaultRetainBlocks},
+			[]interface{}{"kusama", dot.KusamaConfig().Global.Name, kusama.DefaultPruningMode, kusama.DefaultRetainBlocks},
```
qdm12, line 48: thanks for the fixes 💯

cmd/gossamer/utils.go
```diff
@@ -72,15 +72,15 @@ func newTestConfig(t *testing.T) *dot.Config {
 
 	cfg := &dot.Config{
 		Global: dot.GlobalConfig{
-			Name:           dot.GssmrConfig().Global.Name,
-			ID:             dot.GssmrConfig().Global.ID,
+			Name:           dot.WestendDevConfig().Global.Name,
```
qdm12, line 75: supernit should we assign this to a variable, `westendDevConfig := dot.WestendDevConfig()` to avoid calling the function multiple times?

devnet/README.md
```diff
@@ -52,26 +52,25 @@ There are four Docker files used in the devnet.
...
 ### docker-compose.yml
 
-The Docker Compose file.  Specifies the IP addresses of all the nodes. 
-
+The Docker Compose file. Specifies the IP addresses of all the nodes.
 
 ## ECS Fargate Deployment
 
-The [`docker-compose.yml`](gssmr-ecs/docker-compose.yml) file within `devnet/gssmr-ecs` folder uses Docker Compose ECS plugin to deploy and update an existing AWS ECS Cluster using the Fargate launch type running a Gossamer devnet with 3 services corresponding to the 3 keys used `alice`, `bob`, and `charlie`.  
+The [`docker-compose.yml`](gssmr-ecs/docker-compose.yml) file within `devnet/gssmr-ecs` folder uses Docker Compose ECS plugin to deploy and update an existing AWS ECS Cluster using the Fargate launch type running a Gossamer devnet with 3 services corresponding to the 3 keys used `alice`, `bob`, and `charlie`.
```
qdm12, line 67: nit a lot of auto-formatting fixes, would it be possible to save without formatting and open another PR fixing all markdown documents? Otherwise it'll eventually be strange in the git history/git blame why those lines got modified here.

dot/config.go
```diff
@@ -185,64 +184,66 @@ func (p PprofConfig) String() string {
 	return p.Settings.String()
 }
 
-// GssmrConfig returns a new test configuration using the provided basepath
-func GssmrConfig() *Config {
+// WestendDevConfig returns a kusama node configuration
```
qdm12, line 187: Isn't it westend?

```suggestion
// WestendDevConfig returns a westend node configuration
```

dot/config.go
```diff
@@ -185,64 +184,66 @@ func (p PprofConfig) String() string {
...
-			LogLvl:         gssmr.DefaultLvl,
-			MetricsAddress: gssmr.DefaultMetricsAddress,
-			RetainBlocks:   gssmr.DefaultRetainBlocks,
-			Pruning:        pruner.Mode(gssmr.DefaultPruningMode),
-			TelemetryURLs:  gssmr.DefaultTelemetryURLs,
+			Name:           "Westend",
+			ID:             "westend_dev",
+			BasePath:       "~/.gossamer/westend-dev",
+			LogLvl:         log.Info,
+			MetricsAddress: ":9876",
```
qdm12, line 195: Should we only listen on localhost with

```suggestion
			MetricsAddress: "localhost:9876",
```

dot/config.go
```diff
@@ -185,64 +184,66 @@ func (p PprofConfig) String() string {
...
+			RetainBlocks:   512,
+			Pruning:        pruner.Archive,
+			TelemetryURLs:  []genesis.TelemetryEndpoint(nil),
```
qdm12, line 198: nit remove?

```suggestion
```

dot/utils_test.go
```diff
@@ -145,7 +144,7 @@ func TestNewTestConfig(t *testing.T) {
 					Modules: []string{"system", "author", "chain", "state", "rpc", "grandpa", "offchain",
 						"childstate", "syncstate", "payment"},
 					WSPort:           8546,
-					WS:               false,
+					WS:               true,
```
qdm12, line 147: nit why this change for `TestNewTestConfig`

cmd/gossamer/config.go
```diff
@@ -26,24 +24,22 @@ import (
...
+	westendDevName = "westend-dev"
 )
 
 // loadConfigFile loads a default config file if --chain is specified, a specific
 // config if --config is specified, or the default gossamer config otherwise.
 func loadConfigFile(ctx *cli.Context, cfg *ctoml.Config) (err error) {
 	cfgPath := ctx.GlobalString(ConfigFlag.Name)
 	if cfgPath == "" {
-		return loadConfig(cfg, defaultGssmrConfigPath)
+		return loadConfig(cfg, defaultPolkadotConfigPath)
```
kishansagathiya, line 42: `DefaultCfg = dot.WestendDevConfig`

So, default config is westend dev config, but if while loading config file path is not provided you load it from `defaultPolkadotConfigPath`, why?

`return loadConfig(cfg, defaultPolkadotConfigPath)`

Personally, I feel like our default config should be whatever immediate chain that we are trying to support. So, `westend` instead of westend dev

cmd/gossamer/config.go
```diff
@@ -625,10 +619,10 @@ func setDotCoreConfig(ctx *cli.Context, tomlCfg ctoml.CoreConfig, cfg *dot.CoreC
 	case wasmer.Name:
 		cfg.WasmInterpreter = wasmer.Name
 	case "":
-		cfg.WasmInterpreter = gssmr.DefaultWasmInterpreter
+		cfg.WasmInterpreter = wasmer.Name
```
kishansagathiya, line 622: We can do without `case ""`. That can be handled by default case.

cmd/gossamer/config.go
kishansagathiya: Looks good to me.  It improves things.

In exising code, say `cfg.Enabled` is true, but in the flag we pass false, `cfg.Enabled` will be set to true. Which is wrong. `cfg.Enabled` should be set to false since flags takes precedence. 

Proposed code fixes that. If we don't pass the flag, `cfg.Enabled` would stay as it is, but if we do pass it, it will be overridden.

kishansagathiya (approved):
Mostly looks good to me apart from comments made by @qdm12

It would nice if westend chain spec was here. then we could have made that as default config.

kishansagathiya (dismissed):
As per the discussion in standup, our default chain config should be polkadot config.

tagging @danforbes

chain/westend-dev/config.toml
jimjbrettj: Was for compatibility with existing tests I do believe

cmd/gossamer/config.go
jimjbrettj: I mean do we really need to? All cases just end up using wasmer anyways. IMO even if its bad input since we still end up using the intended interpreter a log is fine, but let me know if you think otherwise

cmd/gossamer/utils.go
jimjbrettj: absolutely

dot/config.go
jimjbrettj: tagging Tim for this one @timwu20 what you think?

dot/utils_test.go
jimjbrettj: to match with the westend config

devnet/README.md
jimjbrettj: updated

chain/polkadot/defaults.go
```diff
@@ -21,6 +21,8 @@ var (
 	DefaultConfig = string("./chain/polkadot/config.toml")
 	// DefaultBasePath Default node base directory path
 	DefaultBasePath = string("~/.gossamer/polkadot")
+	// DefaultMetricsAddress is the default metrics server listening address.
+	DefaultMetricsAddress = "localhost:9876"
```
timwu20, line 25: why is this the default now?

cmd/gossamer/config.go
```diff
@@ -26,20 +24,24 @@ import (
...
+	DefaultCfg                  = dot.PolkadotConfig
+	defaultKusamaConfigPath     = "./chain/kusama/config.toml"
+	defaultPolkadotConfigPath   = "./chain/polkadot/config.toml"
+	defaultWestendDevConfigPath = "./chain/westend-dev/config.toml"
+)
+
+const (
+	kusamaName     = "kusama"
+	polkadotName   = "polkadot"
+	westendDevName = "westend-dev"
```
timwu20, line 36: I put these as inline case statements in my westend PR since we don't reference them more than once.

cmd/gossamer/config.go
```diff
@@ -162,8 +155,10 @@ func createInitConfig(ctx *cli.Context) (*dot.Config, error) {
 		return nil, fmt.Errorf("--%s must be %s", PruningFlag.Name, pruner.Archive)
 	}
 
-	if cfg.Global.RetainBlocks < dev.DefaultRetainBlocks {
-		return nil, fmt.Errorf("--%s cannot be less than %d", RetainBlockNumberFlag.Name, dev.DefaultRetainBlocks)
+	const minRetainBlocks = uint32(512)
```
timwu20, line 158: why are we using this `const` now?

cmd/gossamer/config.go
timwu20: We should remove this if we don't plan to use something else.  If we are going to investigate `wazero` we should leave this in.

cmd/gossamer/utils.go
```diff
@@ -67,20 +67,21 @@ func confirmMessage(msg string) bool {
 }
 
 // newTestConfig returns a new test configuration using the provided basepath
-func newTestConfig(t *testing.T) *dot.Config {
+func newTestConfig(t *testing.T, chainConfig *dot.Config) *dot.Config {
```
timwu20, line 70: nice change.

dot/config.go
timwu20: I think it should listen on all addresses.  It's not understood that metrics are only accessible locally.

timwu20 (changes requested):

kishansagathiya:
@jimjbrettj I am assuming that you have taken over this PR. Please ask re-review. Don't dismiss reviews unless you have a very good reason for it.

cmd/gossamer/config.go
jimjbrettj: okay sounds good ill update to do that

chain/polkadot/defaults.go
jimjbrettj: It seems that it is what we were using before https://github.com/ChainSafe/gossamer/blob/64425445e6be472ae1e2dffba85f8ce4221d2eb8/chain/gssmr/defaults.go#L28

cmd/gossamer/config.go
jimjbrettj: Polkadot, kusama, and westend all have a defaultRetainBlocks value of uint32(512). However, westend dev and westend local do not have this value set. This was a change made before I took over the PR but I imagine that this seemed like the most straightforward was to set the minRetainBlocks value. However, the name of the chain being used is stored in the context, so I could do a switch and set it based on what chain we were using. 

If this is preferred let me know, and then we also need to decide the minimum retained blocks for chains where this default value isn't present.

timwu20: This works for me.

timwu20 (approved):

Issue #3020, opened 2023-01-06
Unable to start node as Authority Role without toml config


edwardmack:
## Describe the bug

- I does not seem possible to start gossamer node as Authority Role (building and finalizing blocks) without using a toml config file.

## Expected Behavior

- Should be able to start node with Role Authority using command line arguments (without config file).
- The CLI attribute `--roles` should work, it doesn't seem to be working when passed in as CLI attribute.
- There should be a CLI attribute for `babe-authority` to define this from CLI (there is a babe-authority config attribute).

## Current Behavior

- In order to start node as Authority Role it is necessary to:
  - Use `--config` CLI flag that points to a .toml file that contains attributes:
  - core attribute `role = 4`
  - core attribute `babe-authority = true`

## Possible Solution

- These attributes should be working as CLI attributes.

## To Reproduce
Steps to reproduce the behaviour:

1. Try starting gossamer node as authority role without using config file: 
```
gossamer --genesis /path/to/genesis.json --name alice --rpc  --ws  --ws-external  --rpc-external  --publish-metrics --key alice --babe-lead --rpcmods system,author,chain,state,rpc,grandpa,offchain,childstate,syncstate,payment --no-telemetry --roles 4 --rpcport 40775 --wsport 45543 --port 33403 --metrics-address :40433 --basepath /tmp/gossamer/basepath/alice/data
```
2. The node will start, but will not build or finalize any blocks. 

## Specification

- go version: 1.18
- gossamer version: development
- gossamer commit tag:
- gossamer commit hash:
- operating system:
- additional links:

edwardmack:
Update: further investigations shows that gossamer can not be started if it doesn't have access to toml config file.  Steps to reproduce issue:
```
emack@system76-pc:~/projects/ChainSafe/gossamer$ make gossamer
rm -fr ./bin
  >  Building binary... 
go build -trimpath -o ./bin/gossamer -ldflags="-s -w" ./cmd/gossamer
emack@system76-pc:~/projects/ChainSafe/gossamer$ ./bin/gossamer 
2023-02-01T15:04:12-05:00 INFO     loaded package log configuration: core: INFO, digest: INFO, sync: INFO, network: INFO, rpc: INFO, state: INFO, runtime: INFO, block producer: INFO, finality gadget: INFO   config.go:L122  pkg=cmd
[...]

PR #3091, opened 2023-02-01
fix(cmd/gossamer): embed default toml config files

edwardmack:
## Changes
Embed default toml config files as resources so that they can be loaded from binary instead of local file system.  This includes files that match pattern `chain/*/*.toml`

## Tests

- Build gossamer (from project directory) `make gossamer`
- `cd` to a different directory
- Run gossamer `path/bin/gossamer`

Notice that gossamer runs without errors:

```
2023-02-01T15:04:49-05:00 ERROR    failed to open toml configuration file: open /home/emack/chain/gssmr/config.toml: no such file or directory  toml_config.go:L29      pkg=cmd
2023-02-01T15:04:49-05:00 ERROR    failed to load toml configuration: open /home/emack/chain/gssmr/config.toml: no such file or directory       config.go:L66   pkg=cmd
2023-02-01T15:04:49-05:00 ERROR    failed to set chain configuration: open /home/emack/chain/gssmr/config.toml: no such file or directory       config.go:L109  pkg=cmd
2023-02-01T15:04:49-05:00 ERROR    failed to create node configuration: open /home/emack/chain/gssmr/config.toml: no such file or directory     main.go:L224    pkg=cmd
open /home/emack/chain/gssmr/config.toml: no such file or directory
```

## Issues
fixes: #3020 

## Primary Reviewer

@timwu20


cmd/gossamer/config.go
```diff
@@ -27,10 +27,10 @@ import (
...
 	// DefaultCfg is the default configuration for the node.
 	DefaultCfg                = dot.GssmrConfig
-	defaultGssmrConfigPath    = "./chain/gssmr/config.toml"
-	defaultKusamaConfigPath   = "./chain/kusama/config.toml"
-	defaultPolkadotConfigPath = "./chain/polkadot/config.toml"
-	defaultDevConfigPath      = "./chain/dev/config.toml"
+	defaultGssmrConfigPath    = "chain/gssmr/config.toml"
+	defaultKusamaConfigPath   = "chain/kusama/config.toml"
+	defaultPolkadotConfigPath = "chain/polkadot/config.toml"
+	defaultDevConfigPath      = "chain/dev/config.toml"
```
qdm12, line 33: is this change needed? 🤔

cmd/gossamer/toml_config.go
```diff
@@ -11,25 +11,36 @@ import (
...
+	"github.com/naoina/toml"
 )
 
-// loadConfig loads the values from the toml configuration file into the provided configuration
-func loadConfig(cfg *ctoml.Config, fp string) error {
+func loadConfigFromResource(cfg *ctoml.Config, resourcePath string) error {
+	file, err := gossamer.DefaultConfigTomlFiles.Open(resourcePath)
+	if err != nil {
+		logger.Errorf("failed to open toml configuration file: %s", err)
+		return err
```
qdm12, line 23: nit maybe return a wrapped error and let the caller log this out instead (to avoid duplicate log):

```suggestion
		return fmt.Errorf("opening toml configuration file: %w", err)
```

resources.go
```diff
@@ -0,0 +1,9 @@
+// Copyright 2023 ChainSafe Systems (ON)
+// SPDX-License-Identifier: LGPL-3.0-only
+package gossamer
+
+import "embed"
+
+//go:embed chain/*/*.toml
+// DefaultConfigTomlFiles resource files for default toml configurations
```
qdm12, line 8: nit maybe:

```suggestion
// DefaultConfigTomlFiles is the embedded file system containing the default toml 
// configuration files.
```

resources.go
```diff
@@ -0,0 +1,9 @@
+// Copyright 2023 ChainSafe Systems (ON)
+// SPDX-License-Identifier: LGPL-3.0-only
+package gossamer
```
qdm12, line 3: Having this file at the root of the project isn't that great-great, but it's necessary to have it as a parent of the chains subdirectories. We could have it in the ./chain directory but that would mean changing all default paths. Since we need to change those paths (if we have that resource file in a subdir), maybe we can move those chain files in `./internal/chain/` and update the default paths? That way:

- Embedding is modular (if we later embed more stuff, we can pick what to embed depending on subpackage imports)
- `DefaultConfigTomlFiles` is not importable by other Go projects

qdm12 (approved):

internal/resources.go
```diff
@@ -0,0 +1,9 @@
+// Copyright 2023 ChainSafe Systems (ON)
+// SPDX-License-Identifier: LGPL-3.0-only
+package internal
```
qdm12, line 3: Can we move this to internal/chain/resources.go and have it as `package chain`?

lib/utils/utils.go
```diff
@@ -191,14 +191,15 @@ var (
...
+// and subdirectories `gossamer, testcases` are
 // and returns it as an absolute path.
 func GetProjectRootPath() (rootPath string, err error) {
 	_, fullpath, _, _ := runtime.Caller(0)
 	rootPath = path.Dir(fullpath)
 
-	const directoryToFind = "chain"
-	const subPathsToFind = "dev,gssmr,kusama,polkadot"
+	const directoryToFind = "cmd"
+	const subPathsToFind = "gossamer,testcases"
```
qdm12, line 202: supernit let's just have `gossamer`, since `testcases` might be renamed/disappear/be-moved in the future

```suggestion
	const subPathsToFind = "gossamer"
```

lib/utils/utils.go
```diff
@@ -191,14 +191,15 @@ var (
 	ErrFindProjectRoot = errors.New("cannot find project root")
 )
 
-// GetProjectRootPath finds the root of the project where `go.mod` is
+// GetProjectRootPath finds the root of the project where directory `cmd`
+// and subdirectories `gossamer, testcases` are
```
qdm12, line 195: to match suggestion below

```suggestion
// and subdirectory `gossamer` are
```

resources.go
qdm12: I personally like the changes with `internal/chain`, but let's keep it unresolved in case someone doesn't / see something wrong with it 🤔

cmd/gossamer/toml_config.go
```diff
@@ -11,25 +11,36 @@ import (
...
+	"github.com/naoina/toml"
 )
 
-// loadConfig loads the values from the toml configuration file into the provided configuration
-func loadConfig(cfg *ctoml.Config, fp string) error {
+func loadConfigFromResource(cfg *ctoml.Config, resourcePath string) error {
+	file, err := internal.DefaultConfigTomlFiles.Open(resourcePath)
+	if err != nil {
+		logger.Errorf("opening toml configuration file: %w", err)
+		return err
```
qdm12, line 23: nit

```suggestion
		return fmt.Errorf("opening toml configuration file: %w", err)
```

cmd/gossamer/toml_config.go
```diff
@@ -46,7 +57,7 @@ func loadConfig(cfg *ctoml.Config, fp string) error {
 		},
 	}
 
-	if err = tomlSettings.NewDecoder(file).Decode(cfg); err != nil {
+	if err := tomlSettings.NewDecoder(file).Decode(cfg); err != nil {
```
qdm12, line 60: supermeganit, you could change the function signature (already changed for other reasons) to 

```go
func loadConfig(cfg *ctoml.Config, file fs.File) (err error) {
```

that way you can keep this line untouched. Verysupermeganit please forgive me 😄

timwu20 (commented):
Why do we have to move `chain` to `internal/chain`? Could you not have just created a go file in `chain` which embeds the resources?

edwardmack:
> Why do we have to move `chain` to `internal/chain`? Could you not have just created a go file in `chain` which embeds the resources?

The embed tag needs to be relative to the files it's referencing (and can't start with `.`), so I originally had the embed resource tag in the project root directory to keep the chain resources in the `chain` directory.  But decided to move it based on these comments: https://github.com/ChainSafe/gossamer/pull/3091#discussion_r1094779333

edwardmack:
I don't know how to handle this linter issue:
```
internal/resources.go:7: File is not `gofmt`-ed with `-s` (gofmt)
//go:embed chain/*/*.toml
```
I've done a gofmt of that file and it still complains about this issue.  Any ideas?

kishansagathiya (approved):

chain/westend/config.toml
```diff
@@ -56,4 +56,4 @@ ws-external = false
 [pprof]
 listening-address = "localhost:6060"
 block-rate = 0
-mutex-rate = 0
+mutex-rate = 0
```
qdm12, line 59: nit unneeded diff

```suggestion
mutex-rate = 0

```

qdm12 (approved):
💯

cmd/gossamer/main_test.go
```diff
@@ -202,10 +204,7 @@ func TestMain(m *testing.M) {
 		panic(err)
 	}
 
-	defaultKusamaConfigPath = filepath.Join(rootPath, "./chain/kusama/config.toml")
-	defaultPolkadotConfigPath = filepath.Join(rootPath, "./chain/polkadot/config.toml")
-	defaultWestendDevConfigPath = filepath.Join(rootPath, "./chain/westend-dev/config.toml")
-
+	testWestendDevConfigPath = filepath.Join(rootPath, "./chain/westend-dev/config.toml")
```
timwu20, line 207: You can't be sure that `TestMain` will run first and `testWestendDevConfigPath` set with the correct file path.  If you ran `TestInitCommand_RenameNodeWhenCalled` it'll come be empty string.

timwu20 (changes requested):

cmd/gossamer/README.md
```diff
@@ -20,7 +20,7 @@ This is the default Gossamer execution method, which invokes the `gossamerAction
 blockchain client are [described below in the Client Components section](#client-components).
 
 - `--basepath` - the path to the directory where Gossamer will store its data
-- `--chain` - specifies the [chain configuration](../../chain) that the
+- `--chain` - specifies the [chain configuration](../../internal/chain) that the
```
timwu20, line 23:
```suggestion
- `--chain` - specifies the [chain configuration](../../chain) that the
```

cmd/gossamer/README.md
```diff
@@ -86,7 +86,7 @@ The `importStateAction` function is defined in [`main.go`](main.go).
 The `export` subcommand transforms a genesis configuration and Gossamer state into a TOML configuration file. This
 subcommand invokes the `exportAction` function defined in [`export.go`](export.go).
 
-- `--config` - path to a TOML configuration file (e.g. those defined in [the `chain` directory](../../chain))
+- `--config` - path to a TOML configuration file (e.g. those defined in [the `chain` directory](../../internal/chain))
```
timwu20, line 89:
```suggestion
- `--config` - path to a TOML configuration file (e.g. those defined in [the `chain` directory](../../chain))
```

cmd/gossamer/main_test.go
qdm12: I'm not sure I understand correctly, but `TestMain` always runs before any other tests of the package it's in. Printing `testWestendDevConfigPath` at the start of `TestInitCommand_RenameNodeWhenCalled` is not empty either 🤔

qdm12 (approved):

cmd/gossamer/main_test.go
timwu20: Oh I didn't see it was `*testing.M`.  I thought it was a test.

timwu20 (approved):

PR #3097, opened 2023-02-06
 Change GetBlockByNumber to GetBlocksByNumber and return multiple blocks

### LINKED ISSUE #2748: Change GetBlockByNumber to GetBlocksByNumber and return multiple blocks
## Issue summary
- There could be multiple blocks with the same block number, so `GetBlockByNumber` is not very useful. Currently, it returns best block. I believe, it would make more sense if we start returning all blocks with same block number. 
- We could also have `GetBestBlockByNumber` is that seems important or useful.

- code links 
https://github.com/ChainSafe/gossamer/blob/a0a1804b8572e1eeba8faceccc3254409786ab46/lib/babe/state.go#L29
https://github.com/ChainSafe/gossamer/blob/491af9516dad3cef582d04162558d83acc225e77/dot/state/block.go#L264

kishansagathiya:
## Tests

```sh
```

## Issues
Fixes #2748

## Primary Reviewer

@timwu20


dot/state/block.go
```diff
@@ -250,6 +251,37 @@ func (bs *BlockState) GetHashByNumber(num uint) (common.Hash, error) {
 	return common.NewHash(bh), nil
 }
 
+// GetHashesByNumber returns the block hash on our best chain with the given number
+func (bs *BlockState) GetHashesByNumber(num uint) ([]common.Hash, error) {
```
timwu20, line 255:
```suggestion
func (bs *BlockState) GetBestHashesByNumber(num uint) ([]common.Hash, error) {
```
Let's make it more explicit that it's traversing the best chain that we have.

timwu20 (changes requested):

dot/state/block.go
kishansagathiya: my bad! The comment was not accurate. I fixed that.

Best chain means best fork. So, best chain hashes for given number would not make sense, since there would be only one.

here, GetBlockByNumber gets the block from best chain, but later we can use that block to get all the blocks on that level.

dot/state/block.go
```diff
@@ -250,6 +251,37 @@ func (bs *BlockState) GetHashByNumber(num uint) (common.Hash, error) {
...
+// GetHashesByNumber returns the block hashes with the given number
+func (bs *BlockState) GetHashesByNumber(num uint) ([]common.Hash, error) {
+	blockHashes := []common.Hash{}
+	block, err := bs.GetBlockByNumber(num)
+	if err == nil {
+		blockHashes = bs.bt.GetAllBlocksAtNumber(block.Header.ParentHash)
+	}
+
+	hash, err := bs.bt.GetHashByNumber(num)
+	if err == nil {
```
jimjbrettj, line 263: Nit: Could be nice to reframe this as if err != nil to stick with normal pattern

jimjbrettj (commented):
Small comment, but looks good otherwise!

dot/state/block.go
```diff
@@ -250,6 +251,37 @@ func (bs *BlockState) GetHashByNumber(num uint) (common.Hash, error) {
 	return common.NewHash(bh), nil
 }
 
+// GetHashesByNumber returns the block hashes with the given number
+func (bs *BlockState) GetHashesByNumber(num uint) ([]common.Hash, error) {
+	blockHashes := []common.Hash{}
+	block, err := bs.GetBlockByNumber(num)
+	if err == nil {
+		blockHashes = bs.bt.GetAllBlocksAtNumber(block.Header.ParentHash)
+	}
```
timwu20, line 260: Seems odd that we don't care if this return an error.  Eventually this will call `BlockTree.GetHashByNumber` which returns a number of different sentinel errors.  Should we proceed in every error case?

dot/state/block.go
timwu20: I agree. Can we switch this around? You can do the `errors.Is` check in a `if err != nil ` block, and have all of this code on the same base indentation.

dot/state/block.go
kishansagathiya: fixed it

dot/state/block.go
```diff
@@ -250,6 +251,39 @@ func (bs *BlockState) GetHashByNumber(num uint) (common.Hash, error) {
...
+		if err != nil {
+			return blockHashes, fmt.Errorf("cannot get block %d: %w", num, err)
+		}
+
+		if !slices.Contains(blockHashes, hash) {
+			blockHashes = append(blockHashes, common.NewHash(bh))
+		}
+
+		return blockHashes, nil
+	}
```
EclesioMeloJunior, line 279:
```suggestion
	hash := block.Header.Hash()
```

dot/state/block.go
```diff
@@ -250,6 +251,39 @@ func (bs *BlockState) GetHashByNumber(num uint) (common.Hash, error) {
...
+
+	if !slices.Contains(blockHashes, hash) {
```
EclesioMeloJunior, line 281: Since the function `GetAllBlocksAtNumber` returns all block hashes with the number of the given hash plus one why we need to check if the `blockHashes` contains `hash`?

EclesioMeloJunior (changes requested):

dot/state/block.go
kishansagathiya: `GetAllBlocksAtNumber` does not return all block hashes, only the ones that are in memory.

`bs.GetBlockByNumber(num)` looks into both disk and memory to give you a block.
`bs.bt.GetAllBlocksAtNumber` looks only inside memory.

If hash was read from memory, blockHashes would have it.
If it was read from disk, that means it was pruned from memory so blockHashes won't have it.

dot/state/block.go
kishansagathiya: don't remember why i did that. must be some confusion, fixing it.

EclesioMeloJunior (approved):

timwu20 (approved):

dot/state/block.go
```diff
@@ -250,6 +251,23 @@ func (bs *BlockState) GetHashByNumber(num uint) (common.Hash, error) {
 	return common.NewHash(bh), nil
 }
 
+// GetHashesByNumber returns the block hashes with the given number
+func (bs *BlockState) GetHashesByNumber(num uint) ([]common.Hash, error) {
+	block, err := bs.GetBlockByNumber(num)
+	if err != nil {
+		return []common.Hash{}, fmt.Errorf("getting block by number %w", err)
```
qdm12, line 258: no point allocating an empty slice

```suggestion
		return nil, fmt.Errorf("getting block by number: %w", err)
```

(also missing `:` at the end)

dot/state/block.go
```diff
@@ -250,6 +251,23 @@ func (bs *BlockState) GetHashByNumber(num uint) (common.Hash, error) {
 	return common.NewHash(bh), nil
 }
 
+// GetHashesByNumber returns the block hashes with the given number
+func (bs *BlockState) GetHashesByNumber(num uint) ([]common.Hash, error) {
```
qdm12, line 255: `num` -> `blockNumber` would be nice

dot/state/block_test.go
```diff
@@ -198,6 +198,62 @@ func TestGetSlotForBlock(t *testing.T) {
 	require.Equal(t, expectedSlot, res)
 }
 
+func TestGetHashesByNumber(t *testing.T) {
+	t.Parallel()
+
+	// create two blocks with the same block number and test if GetHashesByNumber gets us
+	// both the blocks
+	bs := newTestBlockState(t, newTriesEmpty())
```
qdm12, line 206: nit `bs` -> `blockState` we're not being charged for using more characters 😉

dot/state/block_test.go
```diff
@@ -198,6 +198,62 @@ func TestGetSlotForBlock(t *testing.T) {
...
+	slot := uint64(77)
```
qdm12, line 207: nit 
```suggestion
	const slot = uint64(77)
```

dot/state/block_test.go
```diff
@@ -198,6 +198,62 @@ func TestGetSlotForBlock(t *testing.T) {
...
+
+	babeHeader := types.NewBabeDigest()
+	err := babeHeader.Set(*types.NewBabePrimaryPreDigest(0, slot, [32]byte{}, [64]byte{}))
+	require.NoError(t, err)
+	data, err := scale.Marshal(babeHeader)
+	require.NoError(t, err)
```
qdm12, line 213: nit

```suggestion
	data := scale.MustMarshal(babeHeader)
```

dot/state/block_test.go
```diff
@@ -198,6 +198,62 @@ func TestGetSlotForBlock(t *testing.T) {
...
+		Body: types.Body{},
+	}
+
+	err = bs.AddBlock(block)
+	require.NoError(t, err)
+
+	babeHeader2 := types.NewBabeDigest()
+	err = babeHeader2.Set(*types.NewBabePrimaryPreDigest(1, slot+1, [32]byte{}, [64]byte{}))
+	require.NoError(t, err)
+	data2, err := scale.Marshal(babeHeader2)
```
qdm12, line 234: nit `data2` -> `encodedBabeHeader2` would be nice

qdm12 (approved):

kishansagathiya:
ignoring some of the nit comments

Issue #2843, opened 2022-09-17
Investigate ZombieNet Integration Requirements


danforbes:
The outcome of this Issue should be a well defined list of the APIs and integrations that Gossamer must implement in order to allow Gossamer nodes to be included in ZombieNet networks and testing pipelines.

edwardmack:
Notes: 
- ZombieNet has two primary functions: spawn and test.
  - Spawn starts ephemeral networks based on configurations defined by the network definition file (toml or json).
  - Test provides a way to conduct test/assertions on the nodes in the networks.  Tests are constructed using the ZombieNet DSL which handles interacting with the using polkadot.js api, metrics and logs.

I've started investigating requirements necessary to spawn gossamer nodes.   The network definition spec defines attributes for defining a nodes configuration.  A network is defines a provider that will host the network: kubernetes, Podman or native.  I've started testing using native provider, which is starting the network on local hardware.

When executing the `spawn` command ZombieNet, the default behavior is to call `build-spec` to build a raw (machine readable) chain spec (based on `chain` attribute in network definition) to use when starting the nodes.  The attributes sent to the build spec command for polkadot nodes are different than those used with gossamer's build-spec command, so using the default behavior with gossamer nodes does not work.  Using the `chain_spec_path` attribute overrides this behavior by using the chain spec defined by the file set in `chain_spec_path` and does not call the build spec command.  Using this does work with gossamer nodes.  

The `default_command` attribute defines the commands that will be executed to start nodes, the default_command also starts the node with a set of pre-defined default arguments that are passed into the default command.  Some of these defaults arguments are arguments that the gossamer node does not recognize, so the gossamer node does not start when using the 'default_command' or 'command' attributes due to errors of unrecognized CLI arguments.  

Using the attribute `command_with_args` solves this issue since using this command there are no default arguments that are included.  Only the args that are defined with the attribute are passed to the command when starting.  

[...]

PR #3098, opened 2023-02-06
feature(cmd/gossamer): add aliases for cli flags

edwardmack:
## Changes
This PR contains changes to make gossamer more congruent with polkadot CLI commands in order to enable gossamer to work with paritytech zombienet testing framework.  
- Add CLI flag `no-mdns` to alias `nomdns`.
- Add CLI flag `base-path` to alias `basepath`
- Add CLI flag `rpc-port` to alias `rpcport`
- Add CLI flag `ws-port` to alias `wsport`
- Implement new CLI flag `node-key` which overrides the secret Ed25519 key to use for libp2p
- Add unit tests for new flags

## Tests

```sh
go test github.com/ChainSafe/gossamer/cmd/gossamer -run ^TestGlobalConfigFromFlags$ -v
go test github.com/ChainSafe/gossamer/cmd/gossamer -run ^TestNetworkConfigFromFlags$ -v
 go test github.com/ChainSafe/gossamer/cmd/gossamer -run ^TestRPCConfigFromFlags$ -v

```

## Issues
Work done researching issue #2843 

## Primary Reviewer

@timwu20

### DIFF
--- a/cmd/gossamer/config.go
+++ b/cmd/gossamer/config.go
@@ -10,6 +10,8 @@ import (
 	"strings"
 	"time"
 
+	"github.com/ChainSafe/gossamer/chain/dev"
+	"github.com/ChainSafe/gossamer/chain/gssmr"
 	"github.com/ChainSafe/gossamer/dot"
 	ctoml "github.com/ChainSafe/gossamer/dot/config/toml"
 	"github.com/ChainSafe/gossamer/dot/state"
@@ -19,7 +21,7 @@ import (
 	"github.com/ChainSafe/gossamer/lib/genesis"
 	"github.com/ChainSafe/gossamer/lib/runtime/wasmer"
 	"github.com/ChainSafe/gossamer/lib/utils"
-	"github.com/urfave/cli"
+	"github.com/urfave/cli/v2"
 )
 
 var (
@@ -34,7 +36,7 @@ var (
 // loadConfigFile loads a default config file if --chain is specified, a specific
 // config if --config is specified, or the default gossamer config otherwise.
 func loadConfigFile(ctx *cli.Context, cfg *ctoml.Config) (err error) {
-	cfgPath := ctx.GlobalString(ConfigFlag.Name)
+	cfgPath := ctx.String(ConfigFlag.Name)
 	if cfgPath == "" {
 		return loadConfig(cfg, defaultPolkadotConfigPath)
 	}
@@ -61,7 +63,7 @@ func setupConfigFromChain(ctx *cli.Context) (*ctoml.Config, *dot.Config, error) 
 	}
 
 	// check --chain flag and load configuration from defaults.go
-	if id := ctx.GlobalString(ChainFlag.Name); id != "" {
+	if id := ctx.String(ChainFlag.Name); id != "" {
 		switch id {
 		case "kusama":
 			logger.Info("loading toml configuration from " + defaultKusamaConfigPath + "...")
@@ -456,7 +458,7 @@ func setDotGlobalConfigFromToml(tomlCfg *ctoml.Config, cfg *dot.GlobalConfig) {
 // setDotGlobalConfigFromFlags sets dot.GlobalConfig using flag values from the cli context
 func setDotGlobalConfigFromFlags(ctx *cli.Context, cfg *dot.GlobalConfig) error {
 	// check --basepath flag and update node configuration
-	if basepath := ctx.GlobalString(strings.Split(BasePathFlag.Name, ",")[0]); basepath != "" {
+	if basepath := ctx.String(BasePathFlag.Name); basepath != "" {
 		cfg.BasePath = basepath
 	}
 
@@ -474,7 +476,7 @@ func setDotGlobalConfigFromFlags(ctx *cli.Context, cfg *dot.GlobalConfig) error 
 	cfg.PublishMetrics = ctx.Bool("publish-metrics")
 
 	// check --metrics-address flag and update node configuration
-	if metricsAddress := ctx.GlobalString(MetricsAddressFlag.Name); metricsAddress != "" {
+	if metricsAddress := ctx.String(MetricsAddressFlag.Name); metricsAddress != "" {
 		cfg.MetricsAddress = metricsAddress
 	}
 
@@ -490,7 +492,7 @@ func setDotGlobalConfigFromFlags(ctx *cli.Context, cfg *dot.GlobalConfig) error 
 	cfg.NoTelemetry = ctx.Bool("no-telemetry")
 
 	var telemetryEndpoints []genesis.TelemetryEndpoint
-	for _, telemetryURL := range ctx.GlobalStringSlice(TelemetryURLFlag.Name) {
+	for _, telemetryURL := range ctx.StringSlice(TelemetryURLFlag.Name) {
 		splits := strings.Split(telemetryURL, " ")
 		if len(splits) != 2 {
 			return fmt.Errorf("%s must be in the format 'URL VERBOSITY'", TelemetryURLFlag.Name)
@@ -517,8 +519,8 @@ func setDotGlobalConfigName(ctx *cli.Context, tomlCfg *ctoml.Config, cfg *dot.Gl
 	initialised := dot.IsNodeInitialised(globalBasePath)
 
 	// consider the --name flag as higher priority
-	if ctx.GlobalString(NameFlag.Name) != "" {
-		cfg.Name = ctx.GlobalString(NameFlag.Name)
+	if ctx.String(NameFlag.Name) != "" {
+		cfg.Name = ctx.String(NameFlag.Name)
 		return nil
 	}
 
@@ -556,12 +558,12 @@ func setDotAccountConfig(ctx *cli.Context, tomlCfg ctoml.AccountConfig, cfg *dot
 	}
 
 	// check --key flag and update node configuration
-	if key := ctx.GlobalString(KeyFlag.Name); key != "" {
+	if key := ctx.String(KeyFlag.Name); key != "" {
 		cfg.Key = key
 	}
 
 	// check --unlock flag and update node configuration
-	if unlock := ctx.GlobalString(UnlockFlag.Name); unlock != "" {
+	if unlock := ctx.String(UnlockFlag.Name); unlock != "" {
 		cfg.Unlock = unlock
 	}
 
@@ -578,11 +580,11 @@ func setDotCoreConfig(ctx *cli.Context, tomlCfg ctoml.CoreConfig, cfg *dot.CoreC
 
 	cfg.BABELead = tomlCfg.BABELead
 	if ctx.IsSet(BABELeadFlag.Name) {
-		cfg.BABELead = ctx.GlobalBool(BABELeadFlag.Name)
+		cfg.BABELead = ctx.Bool(BABELeadFlag.Name)
 	}
 
 	// check --roles flag and update node configuration
-	if roles := ctx.GlobalString(RolesFlag.Name); roles != "" {
+	if roles := ctx.String(RolesFlag.Name); roles != "" {
 		// convert string to byte
 		n, err := strconv.Atoi(roles)
 		b := common.Roles(n)
@@ -643,13 +645,13 @@ func setDotNetworkConfig(ctx *cli.Context, tomlCfg ctoml.NetworkConfig, cfg *dot
 	cfg.DiscoveryInterval = time.Second * time.Duration(tomlCfg.DiscoveryInterval)
 
 	// check --port flag and update node configuration
-	if port := ctx.GlobalUint(PortFlag.Name); port != 0 {
+	if port := ctx.Uint(PortFlag.Name); port != 0 {
 		cfg.Port = uint16(port)
 	}
 
 	// check --bootnodes flag and update node configuration
-	if bootnodes := ctx.GlobalString(BootnodesFlag.Name); bootnodes != "" {
-		cfg.Bootnodes = strings.Split(ctx.GlobalString(BootnodesFlag.Name), ",")
+	if bootnodes := ctx.String(BootnodesFlag.Name); bootnodes != "" {
+		cfg.Bootnodes = strings.Split(ctx.String(BootnodesFlag.Name), ",")
 	}
 
 	// format bootnodes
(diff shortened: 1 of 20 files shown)


cmd/gossamer/config_test.go
```diff
@@ -417,341 +399,290 @@ func TestCoreConfigFromFlags(t *testing.T) {
...
-	for _, c := range testcases {
+	for key, c := range testcases {
 		c := c // bypass scopelint false positive
-		t.Run(c.description, func(t *testing.T) {
-			ctx, err := newTestContext(c.description, c.flags, c.values)
-			require.NoError(t, err)
-			cfg, err := createDotConfig(ctx)
-			require.NoError(t, err)
-			require.Equal(t, c.expected, cfg.Network)
+		t.Run(key, func(t *testing.T) {
```
qdm12, line 509: nit run in parallel

```suggestion
		t.Run(key, func(t *testing.T) {
			t.Parallel()

```

you can then remove the comment `// bypass scopelint false positive`

cmd/gossamer/flags.go
```diff
@@ -100,7 +100,7 @@ var (
 	}
 	// BasePathFlag data directory for node
 	BasePathFlag = cli.StringFlag{
-		Name:  "basepath",
+		Name:  "base-path,basepath", // base-path is argument used by polkadot node
```
qdm12, line 103: Should we just have breaking changes and change these flags to match polkadot? @danforbes ?

dot/network/config.go
```diff
@@ -94,6 +96,9 @@ type Config struct {
 	// privateKey the private key for the network p2p identity
 	privateKey crypto.PrivKey
 
+	//NodeKey is the private hex encoded Ed25519 key to build the p2p identity
```
qdm12, line 99:
```suggestion
	// NodeKey is the private hex encoded Ed25519 key to build the p2p identity
```

dot/network/config.go
```diff
@@ -161,6 +166,19 @@ func (c *Config) checkState() (err error) {
...
+			return fmt.Errorf("parsing hex encoding of ed25519 private key: %w", err)
+		}
+		key := ed25519.NewKeyFromSeed(privateKeySeed)
+		privateKey, err := crypto.UnmarshalEd25519PrivateKey(key)
+		if err != nil {
+			return fmt.Errorf("decoding ed25519 bytes: %w", err)
+		}
+		c.privateKey = privateKey
+		return nil
+	}
```
qdm12, line 181: supernit an empty line below would be nice

```suggestion
	}

```

cmd/gossamer/config.go
```diff
@@ -770,7 +775,7 @@ func setDotRPCConfig(ctx *cli.Context, tomlCfg ctoml.RPCConfig, cfg *dot.RPCConf
 		cfg.Modules = strings.Split(ctx.GlobalString(RPCModulesFlag.Name), ",")
 	}
 
-	if wsport := ctx.GlobalUint(WSPortFlag.Name); wsport != 0 {
+	if wsport := ctx.GlobalUint(strings.Split(WSPortFlag.Name, ",")[0]); wsport != 0 {
```
qdm12, line 778: Can we listen on multiple ports?

cmd/gossamer/config.go
```diff
@@ -674,7 +674,7 @@ func setDotNetworkConfig(ctx *cli.Context, tomlCfg ctoml.NetworkConfig, cfg *dot
 	}
 
 	// check --nomdns flag and update node configuration
-	if nomdns := ctx.GlobalBool(NoMDNSFlag.Name); nomdns {
+	if nomdns := ctx.GlobalBool(strings.Split(NoMDNSFlag.Name, ",")[0]); nomdns {
```
qdm12, line 677: Why splitting??

cmd/gossamer/flags.go
danforbes: I'm fine with either approach.

cmd/gossamer/config.go
```diff
@@ -460,7 +460,7 @@ func setDotGlobalConfigFromToml(tomlCfg *ctoml.Config, cfg *dot.GlobalConfig) {
 // setDotGlobalConfigFromFlags sets dot.GlobalConfig using flag values from the cli context
 func setDotGlobalConfigFromFlags(ctx *cli.Context, cfg *dot.GlobalConfig) error {
 	// check --basepath flag and update node configuration
-	if basepath := ctx.GlobalString(BasePathFlag.Name); basepath != "" {
+	if basepath := ctx.GlobalString(strings.Split(BasePathFlag.Name, ",")[0]); basepath != "" {
```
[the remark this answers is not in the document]
edwardmack: Fixed by upgrading urfave/cli to v2

cmd/gossamer/config_test.go
```diff
@@ -257,15 +238,16 @@ func TestGlobalConfigFromFlags(t *testing.T) {
...
+	for key, c := range testcases {
 		c := c // bypass scopelint false positive
-		t.Run(c.description, func(t *testing.T) {
-			ctx, err := newTestContext(c.description, c.flags, c.values)
-			require.NoError(t, err)
-			cfg, err := createDotConfig(ctx)
-			require.NoError(t, err)
-
-			require.Equal(t, c.expected, cfg.Global)
+		t.Run(key, func(t *testing.T) {
```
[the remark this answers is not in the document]
edwardmack: Updated.

cmd/gossamer/config.go
edwardmack: because of the way urfave/cli V1 handles aliases, this is updated when upgraded to V2.

dot/network/host.go
```diff
@@ -82,7 +82,11 @@ type host struct {
 
 func newHost(ctx context.Context, cfg *Config) (*host, error) {
 	// create multiaddress (without p2p identity)
-	addr, err := ma.NewMultiaddr(fmt.Sprintf("/ip4/0.0.0.0/tcp/%d", cfg.Port))
+	addrString := fmt.Sprintf("/ip4/0.0.0.0/tcp/%d", cfg.Port)
+	if cfg.ListenAddress != "" {
+		addrString = cfg.ListenAddress
+	}
+	addr, err := ma.NewMultiaddr(addrString)
```
qdm12, line 89: nit

```suggestion
	listenAddress := fmt.Sprintf("/ip4/0.0.0.0/tcp/%d", cfg.Port)
	if cfg.ListenAddress != "" {
		listenAddress = cfg.ListenAddress
	}
	addr, err := ma.NewMultiaddr(addrString)
```

dot/network/config.go
```diff
@@ -82,6 +84,8 @@ type Config struct {
 	NoBootstrap bool
 	// NoMDNS disables MDNS discovery
 	NoMDNS bool
+	// ListenAddress multiaddress to listen on
```
qdm12, line 87: nit perhaps you meant this:

```suggestion
	// ListenAddress is the multiaddress to listen on
```

dot/network/config.go
```diff
@@ -94,6 +98,9 @@ type Config struct {
 	// privateKey the private key for the network p2p identity
 	privateKey crypto.PrivKey
 
+	// NodeKey is the private hex encoded Ed25519 key to build the p2p identity
+	NodeKey string
+
```
qdm12, line 103: supernit move this up above unexported fields (here `privateKey`) so it's grouped with exported fields?

cmd/gossamer/main.go
```diff
@@ -207,8 +207,8 @@ func importRuntimeAction(ctx *cli.Context) error {
 // then creates and starts the node and node services
 func gossamerAction(ctx *cli.Context) error {
 	// check for unknown command arguments
-	if arguments := ctx.Args(); len(arguments) > 0 {
-		return fmt.Errorf("failed to read command argument: %q", arguments[0])
+	if arguments := ctx.Args(); arguments.Len() > 0 {
+		return fmt.Errorf("failed to read command argument: %q", arguments.Get(0))
```
qdm12, line 211: nit whilst we're at it changing the line:

```suggestion
		return fmt.Errorf("%d extra unknown command line arguments", arguments.Len())
```

Ideally also:

- List all unknown arguments (not just the number of them, nor the first one only)
- Have a sentinel error; but since we're in main.go I guess it's okay to have just a formatted string.

qdm12 (approved):

kishansagathiya (commented):
Why are we changing flags from global to local? It changes where we can place these flag in a command.

dot/network/host.go
```diff
@@ -82,7 +82,11 @@ type host struct {
 
 func newHost(ctx context.Context, cfg *Config) (*host, error) {
 	// create multiaddress (without p2p identity)
-	addr, err := ma.NewMultiaddr(fmt.Sprintf("/ip4/0.0.0.0/tcp/%d", cfg.Port))
+	addrString := fmt.Sprintf("/ip4/0.0.0.0/tcp/%d", cfg.Port)
+	if cfg.ListenAddress != "" {
+		addrString = cfg.ListenAddress
+	}
```
timwu20, line 88: In this case we need to parse the port from the address supplied and store it in `cfg.Port`. `cfg.Port` is used to populate `externalAddr` further down.

timwu20 (changes requested):

dot/network/host.go
edwardmack: Why wouldn't we continue to use the value in `cfg.Port` for building the `externalAddr`?  Since listen address and external address may be different, should we assume that external address uses the same port as listen address when listen address cfg value has been set?

edwardmack:
> Why are we changing flags from global to local? It changes where we can place these flag in a command.

This was done because we upgraded to v2 of urfave/cli and GlobalString, etc have been deprecated. https://cli.urfave.org/migrate-v1-to-v2/#globalstring-globalbool-and-its-likes-are-deprecated

I don't believe this changes where we can place these flags in a command.

kishansagathiya (approved):

EclesioMeloJunior (changes requested):

cmd/gossamer/config.go
```diff
@@ -625,7 +633,14 @@ func setDotCoreConfig(ctx *cli.Context, tomlCfg ctoml.CoreConfig, cfg *dot.CoreC
 }
 
 // setDotNetworkConfig sets dot.NetworkConfig using flag values from the cli context
-func setDotNetworkConfig(ctx *cli.Context, tomlCfg ctoml.NetworkConfig, cfg *dot.NetworkConfig) {
+func setDotNetworkConfig(ctx *cli.Context, tomlCfg ctoml.NetworkConfig, cfg *dot.NetworkConfig) error {
+	if listenAddress := ctx.String(ListenAddressFlag.Name); listenAddress != "" {
+		if (tomlCfg.Port != cfg.Port) || (ctx.Uint(PortFlag.Name) != 0) {
+			return fmt.Errorf("can not set both port and listen address")
+		}
+		cfg.ListenAddress = listenAddress
```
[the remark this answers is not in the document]
edwardmack: good point. removed.

cmd/gossamer/config.go
```diff
@@ -625,7 +633,14 @@ func setDotCoreConfig(ctx *cli.Context, tomlCfg ctoml.CoreConfig, cfg *dot.CoreC
 }
 
 // setDotNetworkConfig sets dot.NetworkConfig using flag values from the cli context
-func setDotNetworkConfig(ctx *cli.Context, tomlCfg ctoml.NetworkConfig, cfg *dot.NetworkConfig) {
+func setDotNetworkConfig(ctx *cli.Context, tomlCfg ctoml.NetworkConfig, cfg *dot.NetworkConfig) error {
+	if listenAddress := ctx.String(ListenAddressFlag.Name); listenAddress != "" {
+		if (tomlCfg.Port != cfg.Port) || (ctx.Uint(PortFlag.Name) != 0) {
```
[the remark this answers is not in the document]
edwardmack: At this point `cfg.Port` has been initialized with default values in `createDotConfig` which calls this.  This check if tomlCfg.Port is different from cfg.Port (which has been set to default).  If tomlCfg.Port is different it means that the default value has changed.

cmd/gossamer/config.go
```diff
@@ -474,8 +493,13 @@ func setDotGlobalConfigFromFlags(ctx *cli.Context, cfg *dot.GlobalConfig) error
...
 	// check --metrics-address flag and update node configuration
-	if metricsAddress := ctx.GlobalString(MetricsAddressFlag.Name); metricsAddress != "" {
-		cfg.MetricsAddress = metricsAddress
+	if metricsAddress := ctx.String(MetricsAddressFlag.Name); metricsAddress != "" {
+		port, err := strconv.Atoi(metricsAddress)
+		if err != nil {
+			cfg.MetricsAddress = metricsAddress
+		} else {
+			cfg.MetricsAddress = ":" + fmt.Sprint(port)
+		}
```
timwu20, line 502: This is supposed to parse as an integer now?

timwu20 (commented):
LGTM, just the one question.

cmd/gossamer/config.go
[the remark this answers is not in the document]
EclesioMeloJunior: Ok, got it. So if we have a listen address and `tomlCfg.Port` is different we return an error but lets say that we have a listen address and both `tomlCfg.Port` and `cfg.Port` is 5000 and there is not port flag value, that means we don't trigger the error right but we still have a port and a listen address, right? please correct me if I am missing something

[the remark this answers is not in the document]
edwardmack: The condition we're checking for is if the user has changed the port from default (by cli or toml) and is setting the `listen-addr` cli flag.  In this case we don't know which change has priority, so we throw the error.  It's ok if the user changes the port or the listen address, but not both.  The way that we build the cfg from toml and cli is a little wonky, but where we have this check `cfg.Port` has the default port value, if it doesn't equal the `tomlCfg.Port` we know the user has changed the port value in their config.toml to something different from default.  If `tomlCfg.Port` == `cfg.Port` we know to use the default port value (cfg.Port).

cmd/gossamer/config.go
edwardmack: The parsing integer is to handle the new alias for MetriceAddressFlag `--prometheus-port` which is used by zombienet.  The original `--metrics-address` flag used a string in the form `:<port>`.  So this new code checks if the value received is an int, than prefix that with a `:`  If the value received is not an int than use the value directly, as the original flag is handled.

EclesioMeloJunior (approved):
LGTM!!

timwu20 (approved):
Great work Ed!

Issue #2454, opened 2022-03-31
Add Westend testnet to staging setup


timwu20:
## Issue summary

- Add the Westend testnet to our staging setup to sync continuously and collect metrics.  
- Metrics should be namespaced with `westend`.
- Ensure that Westend runtime is correct in gossamer codebase

## Other information and links

-

PR #3103, opened 2023-02-10
feat(chain): Add Westend network as command line `chain` option

timwu20:
## Changes

- Adds `chain=westend` option

## Tests

- Test that gossamer now syncs to Westend test net using the `chain=westend` option

```sh
```

## Issues

- contributes to #2454 

## Primary Reviewer

@kishansagathiya


chain/westend/config.toml
```diff
@@ -0,0 +1,55 @@
...
+    "chain",
+    "state",
+    "rpc",
+    "grandpa",
+    "offchain",
+    "childstate",
+    "syncstate",
+    "payment",
+]
+ws-port = 8546
```
kishansagathiya, line 50: We have kept ws as false in other test chains like kusama. May be do so here?
```
ws = false
ws-external = false
```

kishansagathiya (approved):

jimjbrettj (approved):

PR #3117, opened 2023-02-14
chore(lint): use crypto/rand instead of math/rand

kishansagathiya:
## Changes
`make lint` and deepsource both keep on complaining about how we should use crypto/rand package instead of math/rand.

Considering that `make lint` and deepsource both are compulsary checks in our CI, I require this fix to merge my existing PRs.

## Tests

```sh
make lint
```

## Primary Reviewer

@jimjbrettj

### DIFF
--- a/dot/state/helpers_test.go
+++ b/dot/state/helpers_test.go
@@ -4,9 +4,9 @@
 package state
 
 import (
-	"math/rand"
+	crand "crypto/rand"
+	"math/big"
 	"testing"
-	"time"
 
 	"github.com/ChainSafe/gossamer/dot/types"
 	"github.com/ChainSafe/gossamer/lib/common"
@@ -26,34 +26,25 @@ func newTriesEmpty() *Tries {
 	}
 }
 
-// newGenerator creates a new PRNG seeded with the
-// unix nanoseconds value of the current time.
-func newGenerator() (prng *rand.Rand) {
-	seed := time.Now().UnixNano()
-	source := rand.NewSource(seed)
-	return rand.New(source) //skipcq: GSC-G404
-}
-
-func generateKeyValues(tb testing.TB, generator *rand.Rand, size int) (kv map[string][]byte) {
+func generateKeyValues(tb testing.TB, size int) (kv map[string][]byte) {
 	tb.Helper()
 
 	kv = make(map[string][]byte, size)
 
 	const maxKeySize, maxValueSize = 510, 128
 	for i := 0; i < size; i++ {
-		populateKeyValueMap(tb, kv, generator, maxKeySize, maxValueSize)
+		populateKeyValueMap(tb, kv, maxKeySize, maxValueSize)
 	}
 
 	return kv
 }
 
-func populateKeyValueMap(tb testing.TB, kv map[string][]byte,
-	generator *rand.Rand, maxKeySize, maxValueSize int) {
+func populateKeyValueMap(tb testing.TB, kv map[string][]byte, maxKeySize, maxValueSize int) {
 	tb.Helper()
 
 	for {
 		const minKeySize = 2
-		key := generateRandBytesMinMax(tb, minKeySize, maxKeySize, generator)
+		key := generateRandBytesMinMax(tb, minKeySize, maxKeySize)
 
 		keyString := string(key)
 
@@ -64,7 +55,7 @@ func populateKeyValueMap(tb testing.TB, kv map[string][]byte,
 		}
 
 		const minValueSize = 0
-		value := generateRandBytesMinMax(tb, minValueSize, maxValueSize, generator)
+		value := generateRandBytesMinMax(tb, minValueSize, maxValueSize)
 
 		kv[keyString] = value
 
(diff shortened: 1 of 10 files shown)


qdm12 (changes requested):
I'm opposed to this  because:

- `crypto/rand` is a lot slower than `math/rand`'s PRNG, so that could make tests with many values slower
- Code is more complex (i.e. Intn is very convoluted now)
- it's now hard to have a deterministic test (we could just change `time.Now().UnixNano()` to `0` when debugging)

For golangci-lint, use `//nolint:gosec`. Version v1.48 we currently use does not complain, maybe match your local linter with the CI one?

For Deepsource, you can use `//skipcq: XX-9999` where `XX-9999` is the issue number reported by Deepsource.

kishansagathiya:
> crypto/rand is a lot slower than math/rand's PRNG, so that could make tests with many values slower

How slower exactly? Would we notice it?

> Code is more complex (i.e. Intn is very convoluted now)

Somewhat true. But look at the code changes, Amount of code added and removed is more or less the same. Plus crypto/rand removes the need to use a seed and generator argument that getting passed through so many nested function calls.

> it's now hard to have a deterministic test (we could just change time.Now().UnixNano() to 0 when debugging)

I have modified the tests, they are all passing. It is in fact, everytime different data would be tested.

dot/state/helpers_test.go
```diff
@@ -26,34 +26,25 @@ func newTriesEmpty() *Tries {
 	}
 }
 
-// newGenerator creates a new PRNG seeded with the
-// unix nanoseconds value of the current time.
-func newGenerator() (prng *rand.Rand) {
```
jimjbrettj, line 31: So the new lib takes care of generating the seed and generator for us?

lib/trie/proof/verify_test.go
```diff
@@ -64,7 +66,7 @@ func Test_Verify(t *testing.T) {
 			errWrapped: ErrKeyNotFoundInProofTrie,
 			errMessage: "key not found in proof trie: " +
 				"0x0101 in proof trie for root hash " +
-				"0xec4bb0acfcf778ae8746d3ac3325fc73c3d9b376eb5f8d638dbf5eb462f5e703",
```
jimjbrettj, line 67: Is it better in these cases to no longer test against hard-coded values?

jimjbrettj (commented):
Looks good, just a few questions

lib/trie/proof/verify_test.go
kishansagathiya: math/rand is deterministic. It will always give us the same value, that's why we can use hardcoded values. crypto/rand always gives us different values. So, this hash value changes, and thus we can't use hardcoded values.

In general, it is a good idea to use hardcoded values when when possible.

dot/state/helpers_test.go
kishansagathiya: yes

lib/trie/proof/verify_test.go
jimjbrettj: Okay this makes sense to me thank you. I was initially concerned that we could run into issues with it no longer being deterministic, but it seems this only affects test files so we should be okay I think.

jimjbrettj (commented):
All looks good to me. Want to see where the discussion with @qdm12 goes before approving. Once that is resolved will approve

qdm12:
TLDR: it's just a linting false positive, so use `//nolint:gosec` on your `helpers_test.go` file in your Git branch, and close this PR.

Another important argument in my view is `crypto/rand` exists for when you **NEED cryptographically secure random data**, which we don't need here. The linter `gosec` often triggers false positives on `math/rand`, since it's not context aware, but this also makes sense to warn you in case you would need crypto secure random data.

Plus now we also have Deepsource as non-required, so only using `//nolint:gosec` on the new line added is the appropriate fix.

> How slower exactly? Would we notice it?

You're right, we don't, so never mind on that one

> Somewhat true. But look at the code changes, Amount of code added and removed is more or less the same.

Even outside lines plus and minus, the complexity of each line is way worst for example

```go
	size := minSize + generator.Intn(maxSize-minSize)
	return generateRandBytes(tb, size, generator)
```

versus

```go
	randN, err := crand.Int(crand.Reader, big.NewInt(int64(maxSize-minSize)))
	require.NoError(tb, err)
	size := minSize + int(randN.Int64())
	return generateRandBytes(tb, size)
```

> Plus crypto/rand removes the need to use a seed and generator argument that getting passed through so many nested function calls.

We could just do the same using the global math/rand generator if we would want to, but it's injected on purpose to be able to manipulate seeds easily & for clarity.

> It is in fact, everytime different data would be tested.

That was the case as well for math/rand generator seeded with `time.UnixNano()` (except it's not cryptographically secure), the difference being that some generators could be seeded with a constant to get deterministic outputs which is useful when debugging a test.

Also very-far-side FYI, you can actually have cryptographically secure random data using `math/rand` given a cryptographically secure `rand.Source` implementation, and the linter would still complain about it 😕 I guess some chatgpt-aided-linter would be appreciated at some point 😄

dimartiro:
@kishansagathiya Can we close this PR?

PR #3119, opened 2023-02-16
fix(dot/state) : Retrieve the parent header in case of KeyNotFound

kishansagathiya:
## Changes

This fixes the recent problems of gossamer not being able to sync with kusama after 4k blocks.

## Tests

Run gossamer on kusama and see if we can sync beyond say 5k blocks in a single run.

## Primary Reviewer

@jimjbrettj


dot/state/epoch.go
```diff
@@ -14,6 +14,7 @@ import (
 	"github.com/ChainSafe/gossamer/dot/types"
 	"github.com/ChainSafe/gossamer/lib/blocktree"
 	"github.com/ChainSafe/gossamer/lib/common"
+	trieproof "github.com/ChainSafe/gossamer/lib/trie/proof"
```
qdm12, line 17: nit why not just without alias? 🤔

dot/state/epoch.go
```diff
@@ -399,7 +400,8 @@ func (nem nextEpochMap[T]) Retrieve(blockState *BlockState, epoch uint64, header
 		// sometimes while moving to the next epoch is possible the header
 		// is not fully imported by the blocktree, in this case we will use
 		// its parent header which migth be already imported.
-		if errors.Is(err, blocktree.ErrEndNodeNotFound) {
+		if errors.Is(err, blocktree.ErrEndNodeNotFound) || errors.Is(err, trieproof.ErrKeyNotFound) ||
+			errors.Is(err, blocktree.ErrKeyNotFound) || errors.Is(err, chaindb.ErrKeyNotFound) {
```
qdm12, line 404: Can the returned error be one of these 4 sentinel errors?? I don't think the `trieproof` code is called anywhere in `blockState.IsDescendantOf(hash, headerHash)` for example.

dot/sync/chain_processor.go
```diff
@@ -92,8 +92,8 @@ func (s *chainProcessor) processReadyBlocks() {
 
 		if err := s.processBlockData(*bd); err != nil {
 			// depending on the error, we might want to save this block for later
-			if !errors.Is(err, errFailedToGetParent) {
-				logger.Errorf("block data processing for block with hash %s failed: %s", bd.Hash, err)
+			if !errors.Is(err, errFailedToGetParent) && !errors.Is(err, blocktree.ErrParentNotFound) {
+				logger.Tracef("block data processing for block with hash %s failed: %s", bd.Hash, err)
```
qdm12, line 96: Change this back to `Errorf` since we should see less of them with the additional condition you added right?

```suggestion
				logger.Errorf("block data processing for block with hash %s failed: %s", bd.Hash, err)
```

lib/blocktree/leaves.go
```diff
@@ -41,7 +40,7 @@ func (lm *leafMap) store(key Hash, value *node) {
 func (lm *leafMap) load(key Hash) (*node, error) {
 	v, ok := lm.smap.Load(key)
 	if !ok {
-		return nil, errors.New("key not found")
+		return nil, ErrKeyNotFound
```
qdm12, line 43:
```suggestion
		return nil, fmt.Errorf("%w: %s", ErrKeyNotFound, key)
```

lib/blocktree/errors.go
```diff
@@ -20,6 +20,7 @@ var (
 	// ErrEndNodeNotFound is returned if the end of a subchain does not exist
 	ErrEndNodeNotFound = errors.New("end node does not exist")
 
+	ErrKeyNotFound = errors.New("key not found")
```
qdm12, line 23: nit should this be `ErrHashNotFound` instead? 🤔 It's a 'key' in the leaf map implementation details, but from a caller perspective, it's just a (block) hash right?

dot/sync/chain_processor.go
kishansagathiya: We have a whole lot of them. Look at kusama logs on staging.  It is polluting logs.

lib/blocktree/errors.go
kishansagathiya: There are multiple similar errors, there `node not found`, `end/start node not found`. I am happy with what exists right now. at least when it says key, i know that we could not find it from memory.

dot/state/epoch.go
kishansagathiya: I would rather have trieproof than proof

jimjbrettj (approved):
lgtm

dot/state/epoch.go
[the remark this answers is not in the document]
kishansagathiya: I would rather used the existing sentinel errors, that way these base errors will be part of the final error

EclesioMeloJunior (approved):

dot/state/epoch.go
kishansagathiya: resolved

dot/sync/chain_processor.go
kishansagathiya: I am going to revert this change and merge this. If I still see this error being spammed all over logs, I will just add it to one of the future PRs. It takes about a days to get till the logs that show such errors (if I remember right)

lib/blocktree/errors.go
kishansagathiya: Gonna skip this for now
- doesn't seem so important
- it has taken a lot of re-runs to get the CI green

Issue #3060, opened 2023-01-23
lib/babe: Rewrite skipped tests to correctly use westend


jimjbrettj:
## Issue summary

- Some of the integration tests in babe were not designed to be run on westend, so this PR is to rewrite and fix those tests
- General cleanup of babe integration tests if I see points of improvement
- Add comments to complicated tests to make them easier to follow
- Rename tests to be more descriptive when appropriate 
- There is a weird relationship between the passed-in config data and the VerificationManager that should be cleaned up if possible

Tests that need work are:
- [ ] TestService_HandleSlotWithSameSlot
- [ ] TestBuildBlock_ok
- [ ] TestBuildBlock_failing
- [ ] TestVerificationManager_VerifyBlock_Secondary
- [ ] TestVerificationManager_VerifyBlock_MultipleEpochs
- [ ] TestVerificationManager_VerifyBlock_InvalidBlockOverThreshold

## Other information and links

- Continues work from #3037

PR #3120, opened 2023-02-16
chore(lib/babe): fix test verification manager verfify block multiple epochs

jimjbrettj:
## Changes

- Rewrote the test for testing block verification over multiple epochs
- Inject babe config to test helper and use westend babe config for testing

## Tests

```sh
go test ./lib/babe --tags=integration 
```

## Issues

#3060 

## Primary Reviewer

@timwu20


lib/babe/verify_integration_test.go
```diff
@@ -153,121 +152,117 @@ func TestVerificationManager_OnDisabled_DuplicateDigest(t *testing.T) {
 	require.Equal(t, ErrAuthorityAlreadyDisabled, err)
 }
 
-func TestVerificationManager_VerifyBlock_Ok(t *testing.T) {
+func TestVerificationManager_VerifyBlock_CurrentEpoch(t *testing.T) {
```
jimjbrettj, line 155: rewrote the single epoch okay case

lib/babe/verify_integration_test.go
```diff
@@ -241,33 +211,105 @@ func TestVerificationManager_VerifyBlock_MultipleEpochs(t *testing.T) {
...
 
-	futureEpoch := uint64(5)
+	slot := getSlot(t, runtime, time.Unix(6, 0))
+	block := createTestBlockWithSlot(t, babeService, &genesisHeader, [][]byte{}, 0, epochData, slot)
+
+	err = vm.VerifyBlock(&block.Header)
+	require.NoError(t, err)
+}
+
+func TestVerificationManager_VerifyBlock_FutureEpoch(t *testing.T) {
```
jimjbrettj, line 221: Since we are testing a case of multiple epochs, I wanted to have an isolated test where we only test future epoch

lib/babe/verify_integration_test.go
```diff
@@ -49,7 +49,7 @@ func newTestVerificationManager(t *testing.T, genCfg *types.BabeConfiguration) *
 	require.NoError(t, err)
 
 	if genCfg == nil {
-		genCfg = genesisBABEConfig
+		genCfg = westendBABEConfig
```
jimjbrettj, line 52: My goal in this whole cleanup (not this PR though) is to remove this helper function and just inject the cfg into createTestService helper

qdm12 (approved):

lib/babe/verify_integration_test.go
```diff
@@ -241,33 +211,105 @@ func TestVerificationManager_VerifyBlock_MultipleEpochs(t *testing.T) {
...
+		EpochLength:        600,
+		C1:                 1,
+		C2:                 4,
+		GenesisAuthorities: []types.AuthorityRaw{*auth.ToRaw()},
+		Randomness:         [32]byte{},
+		SecondarySlots:     1,
+	}
+	genesis, genesisTrie, genesisHeader := newWestendDevGenesisWithTrieAndHeader(t)
+	babeService := createTestService(t, ServiceConfig{}, genesis, genesisTrie, genesisHeader, babeConfig, true)
+	vm := NewVerificationManager(babeService.blockState, babeService.epochState)
```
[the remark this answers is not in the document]
jimjbrettj: there already exists a verifier type so this might add confusion/is a shadow declaration, so I think best to leave as is for now. However, I do agree that renaming could be a good thing, so perhaps something to discuss in a different PR

lib/babe/helpers_test.go
```diff
@@ -224,7 +229,9 @@ func createTestService(t *testing.T, cfg ServiceConfig, genesis genesis.Genesis,
 	require.NoError(t, err)
 	cfg.BlockState.(*state.BlockState).StoreRuntime(cfg.BlockState.BestBlockHash(), runtime)
 
+	cfg.Authority = true
 	cfg.IsDev = true
+	cfg.Lead = isLead
```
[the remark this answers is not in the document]
jimjbrettj: oh yeah good catch!

lib/babe/helpers_test.go
```diff
@@ -43,7 +43,7 @@ var (
 		Digest: types.NewDigest(),
 	}
 
-	genesisBABEConfig = &types.BabeConfiguration{
+	westendBABEConfig = &types.BabeConfiguration{
```
EclesioMeloJunior, line 46: I agree with the rename but its values should match the westend's real values otherwise this name does not make much sense

EclesioMeloJunior (approved):
LGTM, just a small comment

lib/babe/verify_integration_test.go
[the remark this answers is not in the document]
qdm12: ok, then rename to `verificationManager`? 🤔

lib/babe/helpers_test.go
jimjbrettj: upon further investigation this isnt even a needed variable since it is only used once, so can just be defined locally. Plus, it is in a helper function that I hope to remove

PR #3133, opened 2023-02-27
fix(lib/babe): use current system time to yield a new slot

EclesioMeloJunior:
## Changes

- Implement a `slot handler` struct which tracks the latest issued slot
- Implement `waitForNextSlot` ([similar to substrate](https://github.com/paritytech/substrate/blob/fbddfbd76c60c6fda0024e8a44e82ad776033e4b/client/consensus/slots/src/slots.rs#L125)), this returns a `Slot` if there is enough time to produce a block and the slot is greater than the latest slot otherwise it will wait until the next slot start
- Replace the slice of timers to the `slot handler` struct and the `waitForNextSlot` function

## Tests

````
go test -timeout 10m github.com/ChainSafe/gossamer/lib/babe --tags=integration -v
````
## Issues

- Closes https://github.com/ChainSafe/gossamer/issues/3126

## Primary Reviewer

@jimjbrettj

### CODE REVIEW

**Comment by jimjbrettj** [comment]
Seems that TestEpochHandler_run is failing in babe

File: lib/babe/epoch_handler.go [hunk 1]
```diff
@@ -75,83 +75,27 @@ func (h *epochHandler) run(ctx context.Context, errCh chan<- error) {
...
-func getAuthoringSlots(slotToPreRuntimeDigest map[uint64]*types.PreRuntimeDigest) []uint64 {
-	authoringSlots := maps.Keys(slotToPreRuntimeDigest)
-	sort.Slice(authoringSlots, func(i, j int) bool {
-		return authoringSlots[i] < authoringSlots[j]
-	})
+		currentSlot := h.slotHandler.waitForNextSlot()
 
-	return authoringSlots
+		// check if the slot is an authoring slot otherwise wait for the next slot
+		digest, has := h.slotToPreRuntimeDigest[currentSlot.number]
```
jimjbrettj, line 91: nit: digest -> preRuntimeDigest

lib/babe/epoch_handler_test.go
```diff
@@ -45,54 +44,3 @@ func TestNewEpochHandler(t *testing.T) {
 	require.Equal(t, epochData, epochHandler.epochData)
 	require.NotNil(t, epochHandler.handleSlot)
 }
-
-func TestEpochHandler_run(t *testing.T) {
```
jimjbrettj, line 49: Nice work cleaning up this test!

lib/babe/slot.go
```diff
@@ -0,0 +1,55 @@
+package babe
+
+import (
+	"time"
+)
+
+// timeUntilNextSlotInNanos calculates, based on the current system time, the remainng
+// time to the next slot
```
jimjbrettj, line 8: perhaps helpful to link the substrate code here? Don't have too, just a possible suggestion if you think it could be helpful

lib/babe/slot.go
```diff
@@ -0,0 +1,55 @@
...
+	lastSlot     *Slot
+}
+
+func newSlotHandler(slotDuration time.Duration) *slotHandler {
+	return &slotHandler{
+		slotDuration: slotDuration,
+	}
+}
+
+func (s *slotHandler) waitForNextSlot() Slot {
```
jimjbrettj, line 30: perhaps helpful to link the substrate code here? Don't have too, just a possible suggestion if you think it could be helpful. At first I linked above function but I think if a link is anywhere this func is most appropriate

lib/babe/slot_test.go
```diff
@@ -0,0 +1,27 @@
+package babe
+
+import (
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestSlotHandeConstructor(t *testing.T) {
```
jimjbrettj, line 10:
```suggestion
func TestSlotHandlerConstructor(t *testing.T) {
```

lib/babe/epoch_handler.go
```diff
@@ -23,6 +20,7 @@ var (
 )
 
 type epochHandler struct {
+	slotHandler *slotHandler
```
qdm12, line 23: nit why is this a pointer? I think we can make it a value for now (until we would need to use an interface eventually later)

lib/babe/slot_test.go
```diff
@@ -0,0 +1,30 @@
...
+	expected := &slotHandler{
+		slotDuration: time.Duration(6000),
+	}
+
+	handler := newSlotHandler(time.Duration(6000))
+	require.Equal(t, expected, handler)
+}
+
+func TestSlotHandlerNextSlot(t *testing.T) {
+	slotDuration := 2 * time.Second
```
qdm12, line 23: supernit
```suggestion
	const slotDuration = 2 * time.Second
```

lib/babe/slot_test.go
```diff
@@ -0,0 +1,30 @@
...
+package babe
+
+import (
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestSlotHandlerConstructor(t *testing.T) {
```
qdm12, line 13:
```suggestion
func TestSlotHandlerConstructor(t *testing.T) {
	t.Parallel()
```

lib/babe/slot_test.go
```diff
@@ -0,0 +1,30 @@
...
+	expected := &slotHandler{
+		slotDuration: time.Duration(6000),
+	}
+
+	handler := newSlotHandler(time.Duration(6000))
+	require.Equal(t, expected, handler)
+}
+
+func TestSlotHandlerNextSlot(t *testing.T) {
```
qdm12, line 22:
```suggestion
func TestSlotHandlerNextSlot(t *testing.T) {
	t.Parallel()
```

lib/babe/epoch_handler.go
```diff
@@ -75,83 +75,27 @@ func (h *epochHandler) run(ctx context.Context, errCh chan<- error) {
...
-}
 
-// getAuthoringSlots returns an ordered slice of slot numbers where we can author blocks,
-// based on the given VRF output and proof map.
-func getAuthoringSlots(slotToPreRuntimeDigest map[uint64]*types.PreRuntimeDigest) []uint64 {
-	authoringSlots := maps.Keys(slotToPreRuntimeDigest)
-	sort.Slice(authoringSlots, func(i, j int) bool {
-		return authoringSlots[i] < authoringSlots[j]
-	})
+		currentSlot := h.slotHandler.waitForNextSlot()
```
timwu20, line 88: This is a blocking call, so the `epochHandler` will not be able to receive the cancellation signal.  I think we need to be able to abandon this call if the cancellation signal comes in.

lib/babe/epoch_handler.go
```diff
@@ -63,6 +62,7 @@ func newEpochHandler(epochNumber, firstSlot uint64, epochData *epochData, consta
 }
 
 func (h *epochHandler) run(ctx context.Context, errCh chan<- error) {
```
timwu20, line 64: Just gonna put a note on this line, and say that the `errCh` should be a return variable since it's the `epochHandler` that produces to this channel.

timwu20 (approved):

lib/babe/epoch_handler.go
```diff
@@ -75,83 +75,27 @@ func (h *epochHandler) run(ctx context.Context, errCh chan<- error) {
...
-	type slotWithTimer struct {
-		startTime time.Time
-		timer     *time.Timer
-		slotNum   uint64
-	}
+	for {
+		currentSlot, err := h.slotHandler.waitForNextSlot(ctx)
+		if errors.Is(err, context.DeadlineExceeded) {
+			errCh <- nil
+			return
```
timwu20, line 84: just to to clarify, you don't want to continue to call `handleSlot` if there is a timeout?

EclesioMeloJunior: mybad I changed it to be any error:

```
currentSlot, err := h.slotHandler.waitForNextSlot(ctx)
if err != nil {
  errCh <- err
  return
}
```

PR #3151, opened 2023-03-09
feat(lib/blocktree): reduce the entries in the runtimes mapping

EclesioMeloJunior:
## Changes

- Reduce the entries of `hashToRuntime` map since most of them were repeated values with different key hashes
- Introduces another pruning system where we keep any runtimes in the new canonical chain otherwise we just update the key hash of the existing one
- Relates to what Quentin is doing here https://github.com/ChainSafe/gossamer/pull/3071 however the difference is that we reduce the usage of stored runtimes and use another method to search for runtimes, also I introduce modification in the pruning implementation where we just use the `newCanonicalChain`.

## Tests

After some measurements I collected the following results:

Before my current implementation:

resource usage

With the current implementation:

resource usage

## Issues

N/A

## Primary Reviewer

@timwu20


qdm12:
Can you point it to `qdm12/hashtoruntime` to see additional commits/diffs perhaps? Thanks 🙏

EclesioMeloJunior:
> Can you point it to `qdm12/hashtoruntime` to see additional commits/diffs perhaps? Thanks 🙏

@qdm12 Done!

qdm12:
Isn't there a weird mix of commits? Maybe `git rebase qdm12/hashtoruntime`?

dot/state/block_test.go
```diff
@@ -1015,6 +1017,7 @@ func TestRange(t *testing.T) {
 			}
 
 			blockState := tt.newBlockState(t, ctrl, genesisHeader)
+			blockState.StoreRuntime(genesisHeader.Hash(), nil)
```
kishansagathiya, line 1020: Why store `nil` runtime?

kishansagathiya (commented):
Looks good! Ran it. Works well.
Haven't really looked at test files. Will take a look at those before approving.

docker/grafana/provisioning/dashboards/gossamer.json
```diff
@@ -155,65 +155,6 @@
...
-        {
-          "datasource": {
-            "type": "prometheus",
-            "uid": "prometheus_id"
-          },
-          "expr": "process_resident_memory_bytes{instance=~\".*gossamer.*\"}",
-          "refId": "A"
-        }
-      ],
-      "title": "Memory usage",
```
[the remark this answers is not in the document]
EclesioMeloJunior: yes, we still tracking the memory usage but I added more lines to the chart:
- go_memstats_heap_released_bytes
- go_memstats_heap_idle_bytes

kishansagathiya (approved):

qdm12 (commented):
Can you please rebase your branch on `qdm12/hashtoruntime`? It's hard to know if anything changed in there (which I would be curious), and there are comments from others on code I wrote. I just rebased `qdm12/hashtoruntime` on development, and then rebased your branch locally only and it works perfectly.

You can do:

```
git checkout qdm12/hashtoruntime
git reset --hard origin/qdm12/hashtoruntime
git checkout eclesio/runtimes-lookup
git rebase qdm12/hashtoruntime
git push -f
```

And then you can (if you want) point this PR to #3071 (then we can either merge this one in 3071, or merge 3071 in dev then merge this one in dev).

docker/prometheus/prometheus.yml
```diff
@@ -10,4 +10,6 @@ scrape_configs:
   - job_name: gossamer
     metrics_path: /metrics
     static_configs:
+      # for mac users, change to host.docker.internal:9876 if you want to
+      # comunicate prometheus container with gossamer running locally
```
timwu20, line 14:
```suggestion
      # communicate prometheus container with gossamer running locally
```

dot/state/block.go
```diff
@@ -892,14 +896,79 @@ func (bs *BlockState) HandleRuntimeChanges(newState *rtstorage.TrieState,
...
+	if errors.Is(err, blocktree.ErrNodeNotFound) {
+		// TODO: https://github.com/ChainSafe/gossamer/issues/3066
+		panic("TODO: while getting a runtime but node does not exists")
+	}
+
+	return runtimeInstance, err
+}
+
+func (bs *BlockState) closestAncestorWithInstance(blockHash common.Hash) (instance Runtime, err error) {
+	allHashesInMapping := bs.bt.GetInMemoryRuntimesBlockHashes()
```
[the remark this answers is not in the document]
timwu20: do the in-memory runtimes get loaded back in memory when the node resumes?

dot/state/block.go
```diff
@@ -892,14 +896,79 @@ func (bs *BlockState) HandleRuntimeChanges(newState *rtstorage.TrieState,
...
+
+		allHeaders[idx] = *header
+	}
+
+	if len(allHeaders) > 1 {
+		// sort the slice using the descending order
+		sort.Slice(allHeaders, func(i, j int) bool {
+			return allHeaders[i].Number > allHeaders[j].Number
+		})
+	}
```
timwu20, line 951: You should cache this, or just keep the sorted list in `BlockTree` if you're going to be doing this every time `GetRuntime()` is called.

lib/blocktree/blocktree.go
```diff
@@ -504,15 +526,54 @@ func (bt *BlockTree) DeepCopy() *BlockTree {
...
+	}
+
+	currentNode = currentNode.parent
+	for currentNode != nil {
+		runtimeInstance := bt.runtimes.get(currentNode.hash)
+		if runtimeInstance != nil {
+			return runtimeInstance, nil
+		}
+
+		currentNode = currentNode.parent
```
timwu20, line 559: Why are we walking the whole chain here and then do something different in `blockState.closestAncestorWithInstance`?

EclesioMeloJunior:
> Is this fine performance-wise to search in the blocktree every time? For example say we have no runtime upgrade for 1 million blocks, we would go through 1M nodes every time? I thought the runtimes mapping was there to have fast lookups originally 🤔

Well, this is an alternative approach to insert a new entry in the runtimes map for every new block imported in the trie. Every time finalization happens we update the latest runtime to the root of the block tree which should be similar to the current approach of wiping the whole runtime entries, I would consider a memory usage improvement to avoid populating a map (since deleting items from it does not reduce its capacity) and only insert valuable data there.

dot/state/block.go
```diff
@@ -892,14 +896,79 @@ func (bs *BlockState) HandleRuntimeChanges(newState *rtstorage.TrieState,
...
 	go bs.notifyRuntimeUpdated(newVersion)
 	return nil
 }
 
 // GetRuntime gets the runtime instance pointer for the block hash given.
 func (bs *BlockState) GetRuntime(blockHash common.Hash) (instance Runtime, err error) {
-	return bs.bt.GetBlockRuntime(blockHash)
+	// we search primarily in the blocktree so we ensure the
+	// fork aware property while searching for a runtime, however
+	// if there is no runtimes in that fork than we look for the
```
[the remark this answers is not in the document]
EclesioMeloJunior: Posted here: https://github.com/ChainSafe/gossamer/pull/3151#issuecomment-1570962796

lib/blocktree/blocktree.go
EclesioMeloJunior: Good catch! I just removed this method, I don't remember why I put it there

lib/blocktree/blocktree_test.go
```diff
@@ -381,39 +382,397 @@ func Test_BlockTree_LowestCommonAncestor_SameChain(t *testing.T) {
...
+		root:   rootNode,
+		leaves: newEmptyLeafMap(),
+		runtimes: &hashToRuntime{
+			mapping: map[common.Hash]Runtime{
+				{1}: rootRuntime,
+			},
+		},
+	}
+
+	// {1} -> {2}
```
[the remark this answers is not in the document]
EclesioMeloJunior: Thanks for the suggestion, I've created some helper functions that makes the test ease to read/understand

jimjbrettj (approved):
nice work, looks good to me!

docker/prometheus/prometheus.yml
```diff
@@ -10,4 +10,6 @@ scrape_configs:
   - job_name: gossamer
     metrics_path: /metrics
     static_configs:
+      # for mac users, change to host.docker.internal:9876 if you want to
```
dimartiro, line 13: Why do we need this if prometheus is running in the same docker-compose.yml ?

dimartiro (approved):
Very good changes Eclésio, I left only one question but is not a blocker. LGTM

docker/prometheus/prometheus.yml
EclesioMeloJunior: You don't need to change if you will run Prometheus + Grafana + Gossamer using docker, but if you're running a Gossamer instance locally (w/o docker) and want to run Prometheus + Grafana using docker you need to change the Prometheus config to find Gossamer using the host. This is the case for me when I was testing Gossamer changes and would like to see the metrics using Promtheus

dimartiro: Ok make sense

lib/blocktree/hashtoruntime.go
```diff
@@ -30,10 +32,76 @@ func (h *hashToRuntime) set(hash Hash, instance runtime.Instance) {
...
+
+// onFinalisation handles pruning and recording on block finalisation.
+// newCanonicalBlockHashes is the block hashes of the blocks newly finalised.
+// The last element is the finalised block hash.
+func (h *hashToRuntime) onFinalisation(newCanonicalBlockHashes []common.Hash) {
+	h.mutex.Lock()
+	defer func() {
+		totalInMemoryRuntimes := len(h.mapping)
+		inMemoryRuntimesGauge.Set(float64(totalInMemoryRuntimes))
+		h.mutex.Unlock()
```
timwu20, line 60: Can we update the gauge in an event driven way? We can update the gauge whenever we add/remove from the `h.mapping`.

timwu20 (changes requested):

lib/blocktree/hashtoruntime.go
EclesioMeloJunior: @timwu20 it happens in the methods `(h *hashToRuntime) set` and `(h *hashToRuntime) delete`, we execute this update here because we modify the mapping (remove forks)

timwu20: Maybe move this defer after line 67 then? There are no changes to the gauge if it returns early.

timwu20 (approved):

Issue #3157, opened 2023-03-14
lib(wasmer): Create wrapper around wasmer.Memory


jimjbrettj:
## Issue summary

- Create a thin wrapper around Wasmer memory to support the gossamer runtime.Memory interface
- This is needed for #3136

## Other information and links

- See this issue from old PR as reference: https://github.com/ChainSafe/gossamer/blob/fff8cb0897fb8279727fba2eb6075df92a0c59d1/lib/runtime/wasmer/memory.go#L29

PR #3160, opened 2023-03-14
feat(runtime/wasmer): create wrapper around wasmer.Memory 

jimjbrettj:
## Changes

- Create wrapper around wasmer.Memory type for new version of wasmer
- Created a new folder in runtime `newWasmer` to build out using newest wasmer version. Code does not compile when both versions are used in the same module so I thought this was a good way to work incrementally, however I am open to feedback on this.

## Tests

```sh
go test ./lib/newWasmer
```

## Issues

#3157

## Primary Reviewer

@timwu20


lib/runtime/newWasmer/memory_test.go
```diff
@@ -0,0 +1,130 @@
...
+		  (func $mem_size (type $mem_size_t) (result i32)
+		    (memory.size))
+		  (export "get_at" (func $get_at))
+		  (export "set_at" (func $set_at))
+		  (export "mem_size" (func $mem_size))
+		  (export "memory" (memory $mem)))
+	`)
+
+	wasmerConfig := wasmer.NewConfig()
+	wasmerConfig.UseSinglepassCompiler()
```
jimjbrettj, line 34: Can go over this in standup, but I think singlepass compiler is the way to go for us so thought I would use it here

lib/runtime/newWasmer/memory.go
```diff
@@ -0,0 +1,38 @@
+// Copyright 2023 ChainSafe Systems (ON)
+// SPDX-License-Identifier: LGPL-3.0-only
+
+package newWasmer
```
EclesioMeloJunior, line 4: I didn't get the point to create a new package to place the wrapper, is that a problem in placing it in the `runtime/wasmer` package?

EclesioMeloJunior (changes requested):

lib/runtime/newWasmer/memory.go
jimjbrettj: It is just a temporary folder to me to build out wasmer with the newest version. If I try to use both versions in the same file, it doesn't compile, so I wrote out newWasmer with the updated version and then used it to replace our old wasmer.

Can see here that there is no more newWasmer because I made my rewrite the default https://github.com/ChainSafe/gossamer/pull/3168

kanishkatn: That's a good approach! I'm doing the same with the cli

lib/runtime/newWasmer/memory.go
```diff
@@ -0,0 +1,38 @@
...
+}
+
+// Data returns the memory's data
+func (m Memory) Data() []byte {
+	return m.memory.Data()
+}
+
+// Length returns the memory's length
+func (m Memory) Length() uint32 {
+	return uint32(m.memory.DataSize())
```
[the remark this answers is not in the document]
jimjbrettj: I'm not set on this, and am open to feedback (going to tag @kishansagathiya here) but I think we are okay. In particular I am hesitant change because the wasmer grow function takes in a pages type, which contains a receiver func `ToUint32()` so since the type already seems to operate using uint32, my instinct is to leave as is.

But also like I said, I'm open to more feedback on this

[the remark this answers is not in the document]
EclesioMeloJunior: Good question, it does not raise an exception however it saturates the values to the `uint32` boundaries: https://go.dev/play/p/xyeyeO9CmWE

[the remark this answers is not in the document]
jimjbrettj: @EclesioMeloJunior do you think this is okay behavior for this situation?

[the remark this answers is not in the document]
dimartiro: If I'm not wrong, using uint32 will limit the memory up to 4Gb. Are we doing it intentionally?

lib/runtime/newWasmer/memory.go
```diff
@@ -0,0 +1,38 @@
+// Copyright 2023 ChainSafe Systems (ON)
```
qdm12, line 1: nit package names should be lowercase, maybe `latestwasmer`?

lib/runtime/newWasmer/memory.go
```diff
@@ -0,0 +1,38 @@
...
+// Length returns the memory's length
+func (m Memory) Length() uint32 {
+	return uint32(m.memory.DataSize())
+}
+
+// Grow grows the memory by the given number of pages
+func (m Memory) Grow(numPages uint32) error {
+	ok := m.memory.Grow(wasmgo.Pages(numPages))
+	if !ok {
+		return fmt.Errorf("%w", errGrowMemory)
```
qdm12, line 35: nit add the number of pages as context?

```suggestion
		return fmt.Errorf("%w: by %d pages", errGrowMemory, numPages)
```

lib/runtime/newWasmer/memory.go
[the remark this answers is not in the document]
qdm12: Maybe my memory is wrong (overflow?), but isn't wasmer limited to 32 bit anyway? 🤔

On top of that, I have never seen 64 bit pointers in our entire wasmer calls.

lib/runtime/newWasmer/memory_test.go
```diff
@@ -0,0 +1,130 @@
...
+	module, err := wasmer.NewModule(store, wasmBytes)
+	require.NoError(t, err)
+
+	importObject := wasmer.NewImportObject()
+
+	// Instantiate the Wasm module.
+	return wasmer.NewInstance(module, importObject)
+}
+
+func TestMemory_Length(t *testing.T) {
```
qdm12, line 49: nit could we run these few tests in parallel perhaps?

qdm12 (commented):
Nice, although quick question, shouldn't we take a stab at wazero instead of wasmer if we want to upgrade it? 🤔

lib/runtime/newWasmer/memory.go
```diff
@@ -0,0 +1,38 @@
+// Copyright 2023 ChainSafe Systems (ON)
+// SPDX-License-Identifier: LGPL-3.0-only
+
+package newWasmer
+
+import (
+	"errors"
+	"fmt"
+
+	wasmgo "github.com/wasmerio/wasmer-go/wasmer"
```
kishansagathiya, line 10:
```suggestion
	"github.com/wasmerio/wasmer-go/wasmer"
```
should be possible to just use wasmer

lib/runtime/newWasmer/memory.go
```diff
@@ -0,0 +1,38 @@
...
+)
+
+var errGrowMemory = errors.New("failed to grow memory")
```
kishansagathiya, line 13:
```suggestion
var errCantGrowMemory = errors.New("failed to grow memory")
```
?

lib/runtime/newWasmer/memory.go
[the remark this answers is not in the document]
EclesioMeloJunior: So I think it is not a problem to leave it as it is since, as @qdm12 says, currently wasm is limited to 32-bit aka `wasm32`, so by default in the current wasm model the limit is 4GB, more info: https://v8.dev/blog/4gb-wasm-memory

jimjbrettj:
> Nice, although quick question, shouldn't we take a stab at wazero instead of wasmer if we want to upgrade it? thinking

@qdm12 I definitely think it could be a good idea to look into waszero, however after talking with Tim before he left on vacation he thought it would be best to have wasmer be stable (we were experiencing issues with it) before migrating

lib/runtime/newWasmer/memory.go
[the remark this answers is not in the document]
jimjbrettj: Just to make sure I understand, to prevent unanticipated unsafe casts we will first cast to a uint64 and then do a bounds check on the value? This is desirable because we can have custom error handling in the case of an invalid cast and also we could realize if there is a change to wasm (e.i. move to uint64)

Is my understanding here correct @kishansagathiya?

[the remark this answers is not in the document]
jimjbrettj: @kishansagathiya I made a change so that we still use uint32 with respect to wasmer, but with the additions adding a check to make sure that the uint is withing the proper bounds. Let me know if further modifications are needed!

EclesioMeloJunior (approved):
Really cool! Great Work 🚀

kishansagathiya (approved):

dimartiro (approved):

Issue #3150, opened 2023-03-09
bug(sync/chain_processor): Failed to handle justification: `start node does not exist`


EclesioMeloJunior:
## Describe the bug

- While syncing on Westend locally, around 200k blocks synced I saw the following error:

```
block data processing for block with hash 0x54b1de84953457f10d821773c88d93320f4288f51446898b5ca846e2024c73a6 failed: handling justification: verifying block number 198940 justification: setting finalised hash: failed to set finalised subchain in db on finalisation: start node does not exist: 0x54b1de84953457f10d821773c88d93320f4288f51446898b5ca846e2024c73a6	chain_processor.go:L96	pkg=sync
```

and then it is not possible to see finalization logs like:
```
🔨 finalised block number 198785 with hash 0x62caf6a8c99d63744f7093bceead8fdf4c7d8ef74f16163ed58b1c1aec67bf18	chain_processor.go:L297	pkg=sync
``` 

- After a quick investigation I notice the error originates in the function `VerifyBlockJustification` where, among other things, is responsible for calling `blockState.SetFinalisedHash(...)` after all the checks are done. 
  - The method `SetFinalisedHash` receives the hash to finalize, the round and the setID. Then it gets the highest finalized hash at the moment and tries to retrieve the subchain between the `highest finalized hash` and the hash we passed as the argument, and here the error happens since the function `RangeInMemory(highestHash, hashToFinalize)` returns the error: `start node does not exists`, which means the `highestHash` was not found in the in-memory block tree.

- Here are my thoughts of the root cause:
  - Is known that the `root node` of the in-memory block tree is a finalized block, so this principle is broken or...
  - While getting the highest finalized block hash the information is being updated, since we call `GetHighestFinalisedHash` and it goes to the database to retrieve the highest round and set id known and with them retrieve the hash from database.

EclesioMeloJunior:
And when this problem happens I notice a increase in the number of tries in the memory as well as the increase of the heap in use bytes

![resource usage](https://user-images.githubusercontent.com/17255488/224149656-14190510-6ab2-4e7c-939b-e0361d84aa57.png)

![tries in memory](https://user-images.githubusercontent.com/17255488/224149587-30c792a5-1dfb-41ba-8c11-65331718e6bb.png)

EclesioMeloJunior:
Actually what is happening is that at block height 198k we are not saving the round and set id for the latest finalized block but we still mark it as a finalized one. As you can check in the log lines above:

[...]

PR #3167, opened 2023-03-21
fix(dot/babe): use `bs.latestFinalised` instead of using `round/set id`

EclesioMeloJunior:
## Changes

- This change fixes the problem where:
  - While syncing in the Westend chain Gossamer stores the finalized block hash bounding it to its respective round and set id, but in Westend happened updates to validators where they restarted the finalization in the same set id but reset the round counter. As described in https://github.com/ChainSafe/gossamer/issues/3150, we synced block 198,656 with round 106 and set 314 and then we synced block 198,785 with round *3* and set id 314
  - The fact that a higher block contains a lower round for the same set id was ignored by Gossamer, which means that Gossamer finalizes block 198,785 but did not bound the hash with the round and set id
  - When the next synced block arrives to be verified and finalized Gossamer understands that the latest finalized block is 198,656 and the `RangeInMemory` method fails to find that block since it is already pruned from memory.

- In `handleFinalisedBlock` use `bs.latestFinalised` instead of using `round/set id`
- Remove from `setHighestRoundAndSetID` the verification of the given round and set id compared to the current round and set id. 

## Tests

After running a Gossamer node for 10+ hours it is finalising blocks keeping the amount of tries in memory low

image

## Issues

- Closes https://github.com/ChainSafe/gossamer/issues/3150

## Primary Reviewer

@jimjbrettj


EclesioMeloJunior:
This strict validation was removed since in Westend there are a lot of weird cases that lead to rounds being messed up in the same set id, but I still wondering about adding an option called `validation-while-syncing` with 2 modes `relaxed` and `strict` than we can make this validation strict while syncing against Kusama/Polkadot and make it relaxed while syncing against Westend. @danforbes

kishansagathiya (approved):

dimartiro (approved):

jimjbrettj:
It seems that in state tests in block_finalization_test.go that TestHighestRoundAndSetID is failing

dot/state/block_finalisation.go
```diff
@@ -63,16 +63,6 @@ func (bs *BlockState) GetFinalisedHash(round, setID uint64) (common.Hash, error)
 }
 
 func (bs *BlockState) setHighestRoundAndSetID(round, setID uint64) error {
-	currRound, currSetID, err := bs.GetHighestRoundAndSetID()
-	if err != nil {
-		return err
-	}
-
-	// higher setID takes precedence over round
-	if setID < currSetID || setID == currSetID && round <= currRound {
```
jimjbrettj, line 72: I know that for relaxed mode we don't need to verify the round number, but do we also now skip verifying that the setId is greater?

jimjbrettj (changes requested):
Looks good, just need to fix failing tests

jimjbrettj (approved):
Nice work!

Issue #3136, opened 2023-02-28
lib(wasmer): Upgade wasmer to newest version


jimjbrettj:
## Issue summary

- Current version of wasmer is causing issues in our tests, specifically in TestVerifyAuthorshipRight_Equivocation in babe/verify_integration_test.go. Future issues could also arise from using the outdated version
- Could also be reason to investigate using wazero instead of wasmer
- See relevant links for more context

List of tasks for this epic:
- [x] #3157 
- [x] Remove usage of wasmer.InstanceContext as it is no longer supported
- [x] Create an imports_func file to create a *wasm.ImportObject (this will replace the current ImportsNodeRuntime() function)
- [x] Rewrite Instance.go to use current wasmer version
- [x] Refactor imports and exports go files to use new instance 

## Other information and links

- https://github.com/wasmerio/wasmer-go/issues/93
- https://github.com/ChainSafe/gossamer/pull/1484
- https://github.com/CosmWasm/cosmwasm/tree/main/packages/vm

jimjbrettj:
This is no longer relevant as we have replaced wasmer with wazero in our codebase

PR #3168, opened 2023-03-21
feat(runtime/wasmer): write wasmer using latest wasmer version

jimjbrettj:
## Changes

- Upgrade wasmer to latest version
- This PR removes the C dependencies in Gossamer, meaning that arm users no longer need to cross-compile to run our code. This means that they can now use delve to debug Gossamer code!

## Tests

I was experiencing inconsistent test results but seems to have been resolved. However, when reviewing I ask that you run tests multiple times to ensure consistency.

```sh
go test ./lib/runtime/wasmer 
```
Any other tests that you might think are important to ensure correctness.

## Issues

#3136

## Primary Reviewer

@timwu20


lib/runtime/wasmer/config.go
```diff
@@ -1,4 +1,4 @@
-// Copyright 2022 ChainSafe Systems (ON)
+// Copyright 2023 ChainSafe Systems (ON)
```
EclesioMeloJunior, line 1: I think we should keep the original date when the file was created

EclesioMeloJunior (changes requested):

lib/runtime/wasmer/config.go
jimjbrettj: Ah okay I wasn't sure, will update

lib/runtime/wasmer/imports.go
```diff
@@ -567,51 +500,57 @@ func ext_crypto_sr25519_generate_version_1(context unsafe.Pointer, keyTypeID C.i
...
 	}
 
 	ret, err := toWasmMemorySized(instanceContext, kp.Public().Encode())
 	if err != nil {
 		logger.Errorf("failed to allocate memory: %s", err)
-		return 0
+		return []wasmer.Value{wasmer.NewI32(0)}, nil
 	}
 
+	// TODO dont think this is safe
```
jimjbrettj, line 521: Looking for feedback on this type cast

lib/runtime/wasmer/imports.go
```diff
@@ -874,101 +827,111 @@ func ext_trie_blake2_256_ordered_root_version_1(context unsafe.Pointer, dataSpan
...
 	if err != nil {
 		logger.Errorf("failed computing trie Merkle root hash: %s", err)
-		return 0
+		return []wasmer.Value{wasmer.NewI32(0)}, nil
 	}
 
 	logger.Debugf("root hash is %s", hash)
 	copy(memory[ptr:ptr+32], hash[:])
-	return C.int32_t(ptr)
+	// TODO is this safe cast
```
jimjbrettj, line 849: Same with this cast

lib/runtime/wasmer/instance.go
```diff
@@ -196,96 +257,28 @@ func GetRuntimeVersion(code []byte) (version runtime.Version, err error) {
...
-	memory, err := wasm.NewMemory(23, 0)
-	if err != nil {
-		return instance, nil, fmt.Errorf("creating web assembly memory: %w", err)
-	}
-
-	_, err = imports.AppendMemory("memory", memory)
-	if err != nil {
-		return instance, nil, fmt.Errorf("appending memory to imports: %w", err)
+// UpdateRuntimeCode updates the runtime instance to run the given code
+func (in *Instance) UpdateRuntimeCode(code []byte) error {
```
jimjbrettj, line 261: This function appears to never be called. Delete?

lib/runtime/wasmer/imports_test.go
```diff
@@ -417,16 +423,16 @@ func Test_ext_storage_clear_prefix_version_2(t *testing.T) {
...
-	testkey2 := []byte("noot1")
+	testkey2 := []byte("jimboj1")
 	inst.ctx.Storage.Put(testkey2, []byte{1})
 
-	testkey3 := []byte("noot2")
+	testkey3 := []byte("jimboj2")
 	inst.ctx.Storage.Put(testkey3, []byte{1})
 
-	testkey4 := []byte("noot3")
+	testkey4 := []byte("jimboj3")
```
timwu20, line 435: update the comment on line 452?

lib/runtime/wasmer/instance.go
timwu20: We usually create a new instance for runtime upgrades?

jimjbrettj: Yeah exactly

lib/runtime/wasmer/imports.go
jimjbrettj: It seems that currently wasmer only allows type int to be passed in for I32 wasm type. I have opened an issue asking about this [here](https://github.com/wasmerio/wasmer-go/issues/382).

As discussed offline, I am going to add checks in all instances where we cast values to send to the runtime.

lib/runtime/wasmer/instance.go
```diff
@@ -69,26 +73,84 @@ func NewInstanceFromTrie(t *trie.Trie, cfg Config) (*Instance, error) {
...
+	return NewInstance(fileBytes, cfg)
 }
 
 // NewInstance instantiates a runtime from raw wasm bytecode
-func NewInstance(code []byte, cfg Config) (instance *Instance, err error) {
+func NewInstance(code []byte, cfg Config) (*Instance, error) {
+	return newInstance(code, cfg)
+}
+
+func newInstance(code []byte, cfg Config) (*Instance, error) {
```
kishansagathiya, line 89: why have these two newInstance funcs?

timwu20:
Make sure you run this on staging and ensure we're still syncing fine through runtime upgrades.

lib/runtime/wasmer/imports.go
```diff
@@ -244,54 +160,59 @@ func ext_crypto_ed25519_generate_version_1(context unsafe.Pointer, keyTypeID C.i
...
 	if err != nil {
 		logger.Warnf("failed to allocate memory: %s", err)
-		return 0
+		return []wasmer.Value{wasmer.NewI32(0)}, nil
+	}
+
+	castedRet, err := safeCastInt32(ret)
+	if err != nil {
+		logger.Errorf("failed to safely cast pointer: %s", err)
+		return []wasmer.Value{wasmer.NewI32(0)}, nil
```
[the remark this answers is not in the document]
jimjbrettj: The way our code currently operates with regard to runtime functions is that we respond to output variables and validate it on a per-use case basis. So rather than erroring out, we put the burden on the caller to check that the value received is correct.

Here is a conversation about this from another one of my PRs https://github.com/ChainSafe/gossamer/pull/3188#discussion_r1160610879 and I think Kishan's point is valid.

I do, however, wonder if we should migrate away from this and start returning errors from the runtime. I think that would be a pretty large effort and be outside of the scope of this codebase. 

One final note on this specific safe casting case, should we just panic? I know we typically avoid panics, but doing an invalid pointer cast on a pointer used for memory management feels like a potential good reason to me. Let me know your thoughts @timwu20

lib/runtime/wasmer/instance.go
jimjbrettj: Good call, updated

lib/runtime/wasmer/imports_test.go
```diff
@@ -943,21 +959,17 @@ func Test_ext_crypto_ecdsa_verify_version_2_Table(t *testing.T) {
 			key:      []byte{132, 2, 39, 0, 55, 134, 131, 142, 43, 100, 63, 134, 96, 14, 253, 15, 222, 119, 154, 110, 188, 20, 159, 62, 125, 42, 59, 127, 19, 16, 0, 161, 236, 109}, //nolint:lll
 			expected: []byte{0, 0, 0, 0},
 		},
-		"invalid_key": {
+		"invalid_key length": {
```
dimartiro, line 962:
```suggestion
		"invalid_key_length": {
```

lib/runtime/wasmer/imports_test.go
```diff
@@ -943,21 +959,17 @@ func Test_ext_crypto_ecdsa_verify_version_2_Table(t *testing.T) {
...
 			sig: []byte{5, 1, 187, 0, 0, 183, 46, 115, 242, 32, 9, 54, 141, 207, 44, 15, 238, 42, 217, 196, 111, 173, 239, 204, 128, 93, 49, 179, 137, 150, 162, 125, 226, 225, 28, 145, 122, 127, 15, 154, 185, 11, 3, 66, 27, 187, 204, 242, 107, 68, 26, 111, 245, 30, 115, 141, 85, 74, 158, 211, 161, 217, 43, 151, 120, 125, 1}, //nolint:lll
 			msg: []byte{48, 72, 101, 108, 108, 111, 32, 119, 111, 114, 108, 100, 33},
 			key: []byte{132, 2, 39, 55, 134, 131, 142, 43, 100, 63, 134, 96, 14, 253, 15, 222, 119, 154, 110, 188, 20, 159, 62, 125, 42, 59, 127, 19, 16, 0, 161, 236, 109}, //nolint:lll
-			err: wasmer.NewExportedFunctionError(
-				"rtm_ext_crypto_ecdsa_verify_version_2",
-				"running runtime function: Failed to call the `%s` exported function."),
+			err: fmt.Errorf("running runtime function: unreachable"),
 		},
-		"invalid_message": {
+		"invalid_message length": {
```
dimartiro, line 968:
```suggestion
		"invalid_message_length": {
```

dimartiro (changes requested):

dot/core/service_test.go
```diff
@@ -252,7 +252,7 @@ func Test_Service_handleCodeSubstitution(t *testing.T) {
 			},
 			blockHash:  common.Hash{0x01},
 			errWrapped: wasmer.ErrWASMDecompress,
-			errMessage: "creating new runtime instance: setting up VM: " +
+			errMessage: "creating new runtime instance: " +
 				"wasm decompression failed: unexpected EOF",
```
kanishkatn, line 256: nit
```suggestion
			errMessage: `creating new runtime instance: 
wasm decompression failed: unexpected EOF`,
```

lib/runtime/wasmer/imports.go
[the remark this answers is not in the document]
jimjbrettj: Yup, I would say I agree. I think propagating the errors back to handle is a good way to go.

I think in this case it would be good to pass the error out of the runtime context and then panic. I will open an issue addressing the rest of the runtime error-handling logic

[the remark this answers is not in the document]
jimjbrettj: Here is the issue: https://github.com/ChainSafe/gossamer/issues/3216

jimjbrettj:
@timwu20 Ran on staging, everything seems to be okay as far as I can tell. Talking with @EclesioMeloJunior he said that a runtime upgrade occurs around block 100k and I have made it well past that

lib/runtime/wasmer/instance.go
```diff
@@ -196,60 +253,74 @@ func GetRuntimeVersion(code []byte) (version runtime.Version, err error) {
...
 	if err != nil {
-		return instance, nil, fmt.Errorf("creating web assembly memory: %w", err)
+		panic(err)
 	}
 
-	_, err = imports.AppendMemory("memory", memory)
+	castedDataLength, err := safeCastInt32(dataLength)
 	if err != nil {
-		return instance, nil, fmt.Errorf("appending memory to imports: %w", err)
+		panic(err)
```
timwu20, line 289: so is this [issue](https://github.com/ChainSafe/gossamer/issues/3216) supposed to change this line to return an error?

lib/runtime/wasmer/instance.go
```diff
@@ -196,60 +253,74 @@ func GetRuntimeVersion(code []byte) (version runtime.Version, err error) {
...
 	}
 
-	// Instantiates the WebAssembly module.
-	instance, err = wasm.NewInstanceWithImports(code, imports)
+	wasmValue, err := runtimeFunc(castedInputPointer, castedDataLength)
 	if err != nil {
-		return instance, nil, fmt.Errorf("creating web assembly instance: %w", err)
+		if errors.Is(err, errMemoryValueOutOfBounds) {
+			panic(fmt.Errorf("executing runtime function: %v", err))
+		}
```
timwu20, line 296: so is this https://github.com/ChainSafe/gossamer/issues/3216 supposed to change this line to return the `errMemoryValueOutOfBounds` back to caller?

jimjbrettj: Yeah that is what I am thinking

lib/runtime/wasmer/instance.go
jimjbrettj: yes this is what I had in mind. Then whoever calls exec can handle different situations with respect to specific errors. Or even at the very least give helpful output

lib/runtime/wasmer/exports_test.go
```diff
@@ -22,10 +21,10 @@ import (
 	"github.com/ChainSafe/gossamer/lib/trie"
 	"github.com/ChainSafe/gossamer/lib/utils"
 	"github.com/ChainSafe/gossamer/pkg/scale"
-	"github.com/centrifuge/go-substrate-rpc-client/v4/signature"
-
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+
+	"github.com/centrifuge/go-substrate-rpc-client/v4/signature"
```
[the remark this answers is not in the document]
jimjbrettj: this should be updated but github isnt showing it as so

dimartiro (approved):

EclesioMeloJunior (approved):
LGTM! Just couple questions

lib/runtime/wasmer/exports_test.go
```diff
@@ -33,8 +31,6 @@ import (
 var testKeyOwnershipProof types.OpaqueKeyOwnershipProof = types.OpaqueKeyOwnershipProof([]byte{64, 138, 252, 29, 127, 102, 189, 129, 207, 47, 157, 60, 17, 138, 194, 121, 139, 92, 176, 175, 224, 16, 185, 93, 175, 251, 224, 81, 209, 61, 0, 71}) //nolint:lll
 
 func Test_Instance_Version(t *testing.T) {
-	t.Parallel()
```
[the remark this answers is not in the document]
jimjbrettj: After talking with @timwu20 we didn't think it made sense to execute the runtime tests in parallel

README.md
```diff
@@ -81,22 +81,11 @@ make install
...
-[Golang environment variables](https://pkg.go.dev/cmd/go#hdr-Environment_variables):
-
-```sh
-GOARCH="amd64"
-```
+**Note**: Apple Silicon users running aarch64 might run into issues 
+with our wasm interpreter since wasmer is still working on supporting this architecture.
+See their [README](https://github.com/wasmerio/wasmer-go) for me info on supported platforms.
+Currently, there are no known issues regarding this within the Gossamer repo, but if you run into one please open an issue
+on our GitHub.
```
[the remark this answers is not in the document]
jimjbrettj: My thought is this feels like a non-gossamer problem and has more to do with them having a proper go setup for their computer. I'm not opposed to the change but have some questions:

Is it just because we previously didn't support the arch so we want to include a reminder for devs? If not that, than why target aarch64 specifically?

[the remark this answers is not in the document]
jimjbrettj: This has been updated

timwu20 (approved):

Issue #3087, opened 2023-01-31
lib/runtime: remove outdated/unneeded runtimes in `constants.go`


EclesioMeloJunior:
## Description

Currently, in `lib/runtime` there is a file `constants.go` where we define some runtime names and URLs which is used for testing purposes. 

It is important to make our codebase concise and in order to address the epic #2895, we should remove all outdated and unneeded runtimes, for example:

- `NODE_RUNTIME` is a modified runtime generated using version 0.9 and placed in `github.com/noot/substrate/blob/noot/v0.9` which should not be used for testing since it is outdated

PR #3177, opened 2023-03-27
refactor(runtimes): remove deprecated node runtimes

dimartiro:
## Changes

- Remove / replace old node runtimes
- Use westend runtime when appropriate

## Tests

```sh
make test
```

## Issues

#3087 

## Primary Reviewer

@EclesioMeloJunior


dot/rpc/modules/state_integration_test.go
```diff
@@ -30,27 +30,31 @@ const (
...
 	// expected results based on responses from prior tests
 	expected := StateRuntimeVersionResponse{
-		SpecName:         "node",
-		ImplName:         "substrate-node",
-		AuthoringVersion: 10,
-		SpecVersion:      264,
+		SpecName:         "westend",
+		ImplName:         "parity-westend",
+		AuthoringVersion: 2,
+		SpecVersion:      9290,
```
jimjbrettj, line 36: Where did yu get all the info regarding versions? Does it all correspond to the current westend state?

jimjbrettj (commented):
Looks good, just have questions about how some of the data was retrieved

dot/rpc/modules/state_integration_test.go
dimartiro: @jimjbrettj thanks for the question, we are using westend v0.9.29 as defined [here](https://github.com/ChainSafe/gossamer/blob/fc1055dddce38cdcd61ec9afb593a10a30e2fde9/lib/runtime/constants.go#L54-L58).
So you can get the[ version data](https://github.com/paritytech/polkadot/blob/v0.9.29/runtime/westend/src/lib.rs#L105-L117) for that version looking into that release in substrate repo

jimjbrettj: Perfect!

jimjbrettj (approved):
Great work!

dot/rpc/modules/state_integration_test.go
qdm12: Should we comment in the link to the version data above `expected := StateRuntimeVersionResponse{`? It would be a nice little addition 😉

qdm12 (approved):

EclesioMeloJunior (approved):

PR #3178, opened 2023-03-27
fix(lib/babe): `TestCalculateThreshold` failing locally

EclesioMeloJunior:
## Changes

- `TestCalculateThreshold` failed locally due to a difference of 1 byte

## Tests

```sh
go test -timeout 10m -run ^TestCalculateThreshold$ github.com/ChainSafe/gossamer/lib/babe --tags=integration -v
```

## Issues

N/A

## Primary Reviewer

@kanishkatn @jimjbrettj @dimartiro

### DIFF
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -22,6 +22,10 @@ jobs:
     timeout-minutes: 60
     runs-on: buildjet-4vcpu-ubuntu-2204
     steps:
+      - name: uname -a
+        run: |
+          uname -a
+
       - name: Cancel Previous Runs
         uses: styfle/cancel-workflow-action@0.11.0
         with:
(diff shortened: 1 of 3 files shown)


dimartiro (approved):
All tests passed in my machine now! thanks!! Did you test it in a non ARM chip?

EclesioMeloJunior:
@kanishkatn yeah, I have an assumption that this 1-byte difference is because our CI is running on 32 bit OS runners

jimjbrettj:
This fails for me with the following output: 

image

jimjbrettj:
I and the CI both run on x86_64, so it seems there must be some difference in arm vs x86_64 causing this to fail

EclesioMeloJunior:
Jimmy's PR https://github.com/ChainSafe/gossamer/pull/3168 solves the problem by removing the C dependencies we have with the wasmer runtime so now Apple Silicon users can go back to `arm64` using: `go env -w GOARCH=arm64` and then the test will pass as a charm!! Nice work @jimjbrettj

Issue #3182, opened 2023-03-29
Implement RPC method state_queryStorageAt


edwardmack:
## Issue summary

- Polkadot.js/api uses RPC method `state_queryStorageAt` when making javascript calls using the `api.query` interface.  When making these calls to gossamer node we get error: 
```
RPC-CORE: queryStorageAt(keys: Vec<StorageKey>, at?: BlockHash): Vec<StorageChangeSet>:: -32000: rpc: can't find method "state.QueryStorageAt": null
```

## Other information and links

- see https://github.com/paritytech/substrate/pull/5362

PR #3185, opened 2023-03-30
feat(dot/rpc): implement RPC method state_queryStorageAt

edwardmack:
## Changes
- Implement code for RPC method `state_queryStorageAt` which returns storage values for given keys at best block or given block hash if `at` parameter is used.
- This fixes error `rpc: can't find method "state.QueryStorageAt"` that occurs when using polkadot.js/api methods in `api.query` package.

## Tests

```sh
go test github.com/ChainSafe/gossamer/dot/rpc/modules -run ^TestStateModuleQueryStorageAt$ -v
```

## Issues
closes: #3182 

## Primary Reviewer

@timwu20


dot/rpc/modules/state.go
```diff
@@ -76,6 +76,12 @@ type StateStorageQueryRangeRequest struct {
 	EndBlock   common.Hash `json:"block"`
 }
 
+// StateStorageQueryAtRequest holds json fields
+type StateStorageQueryAtRequest struct {
```
qdm12, line 80: I think `StateStorageQueryAtRequest` can be unexported

dot/rpc/modules/state.go
```diff
@@ -489,6 +495,41 @@ func (sm *StateModule) QueryStorage(
 	return nil
 }
 
+// QueryStorageAt queries historical storage entries (by key) at the best block if the given block hash is nil
```
qdm12, line 498:
```suggestion
// QueryStorageAt queries historical storage entries (by key) at the block hash given or 
// the best block if the given block hash is nil
```

dot/rpc/modules/state.go
```diff
@@ -489,6 +495,41 @@ func (sm *StateModule) QueryStorage(
...
+func (sm *StateModule) QueryStorageAt(
+	_ *http.Request, req *StateStorageQueryAtRequest, res *[]StorageChangeSetResponse) error {
```
qdm12, line 500: nit `req` -> `request` and `res` -> `response` (see comment below to avoid name conflict)

dot/rpc/modules/state.go
```diff
@@ -489,6 +495,41 @@ func (sm *StateModule) QueryStorage(
...
+	var atBlockHash common.Hash
+	if req.At.IsEmpty() {
+		atBlockHash = sm.blockAPI.BestBlockHash()
+	} else {
+		atBlockHash = req.At
+	}
```
qdm12, line 506: nit a tiny bit shorter

```suggestion
	atBlockHash := req.At
	if atBlockHash.IsEmpty() {
		atBlockHash = sm.blockAPI.BestBlockHash()
	}
```

dot/rpc/modules/state.go
```diff
@@ -489,6 +495,41 @@ func (sm *StateModule) QueryStorage(
...
+
+		changes = append(changes, [2]*string{stringPtr(key), hexValue})
+	}
+
+	response = append(response, StorageChangeSetResponse{
+		Block:   &atBlockHash,
+		Changes: changes,
+	})
+
+	*res = response
```
qdm12, line 529: nit change to directly be:

```go
	res = new([]StorageChangeSetResponse)
	*res = []StorageChangeSetResponse{{
		Block:   &atBlockHash,
		Changes: changes,
	}}
```

that way you can also remove the init `response := make([]StorageChangeSetResponse, 0, 1)` above and avoid name conflicts between `res` and `response`.

dot/rpc/modules/state.go
```diff
@@ -489,6 +495,41 @@ func (sm *StateModule) QueryStorage(
...
+		value, err := sm.storageAPI.GetStorageByBlockHash(&atBlockHash, common.MustHexToBytes(key))
+		if err != nil {
+			return fmt.Errorf("getting value by block hash: %w", err)
+		}
+		var hexValue *string
+		if len(value) > 0 {
+			hexValue = stringPtr(common.BytesToHex(value))
+		}
+
+		changes = append(changes, [2]*string{stringPtr(key), hexValue})
```
qdm12, line 521: nit since there is no `continue` clause, you can set it by index directly and create changes with

```go
changes := make([][2]*string, len(req.Keys))
```

and then have

```go
for i, key := range req.Keys {
    // ...
    changes[i] = [2]*string{stringPtr(key), hexValue})
}
```

dot/rpc/modules/state_test.go
```diff
@@ -1129,3 +1129,147 @@ func TestStateModuleQueryStorage(t *testing.T) {
...
+		fields    fields
+		args      args
+		errRegexp string
+		exp       []StorageChangeSetResponse
+	}{
+		"missing_start_block": {
+			fields: fields{
+				storageAPIBuilder: func(ctrl *gomock.Controller) StorageAPI {
+					mockStorageAPI := NewMockStorageAPI(ctrl)
+					mockStorageAPI.EXPECT().GetStorageByBlockHash(&common.Hash{2}, []byte{1, 2, 3}).Return([]byte{1, 1, 1}, nil)
```
qdm12, line 1155: nit split since it's a bit of a long line 😉 

```suggestion
					mockStorageAPI.EXPECT().GetStorageByBlockHash(&common.Hash{2}, []byte{1, 2, 3}).
						Return([]byte{1, 1, 1}, nil)
```

dot/rpc/modules/state_test.go
```diff
@@ -1129,3 +1129,147 @@ func TestStateModuleQueryStorage(t *testing.T) {
...
+func TestStateModuleQueryStorageAt(t *testing.T) {
+	t.Parallel()
+	errTest := errors.New("test error")
+
+	type fields struct {
+		storageAPIBuilder func(ctrl *gomock.Controller) StorageAPI
+		blockAPIBuilder   func(ctrl *gomock.Controller) BlockAPI
+	}
+	type args struct {
+		in0 *http.Request
```
qdm12, line 1142: Since `in0` is always ignored, just remove it and always pass nil in the subtest code?

dot/rpc/modules/state_test.go
```diff
@@ -1129,3 +1129,147 @@ func TestStateModuleQueryStorage(t *testing.T) {
...
 	}
 }
+
+func TestStateModuleQueryStorageAt(t *testing.T) {
+	t.Parallel()
+	errTest := errors.New("test error")
+
+	type fields struct {
+		storageAPIBuilder func(ctrl *gomock.Controller) StorageAPI
+		blockAPIBuilder   func(ctrl *gomock.Controller) BlockAPI
```
qdm12, line 1139: nit let's return mock types (better to not return interfaces generally)

```go
		storageAPIBuilder func(ctrl *gomock.Controller) *MockStorageAPI
		blockAPIBuilder   func(ctrl *gomock.Controller) *MockBlockAPI
```

dot/rpc/modules/state_test.go
```diff
@@ -1129,3 +1129,147 @@ func TestStateModuleQueryStorage(t *testing.T) {
...
+	}
+	type args struct {
+		in0 *http.Request
+		req *StateStorageQueryAtRequest
```
qdm12, line 1143: nit rename `req` -> `request`

dot/rpc/modules/state_test.go
```diff
@@ -1129,3 +1129,147 @@ func TestStateModuleQueryStorage(t *testing.T) {
...
+	}
+	tests := map[string]struct {
+		fields    fields
+		args      args
+		errRegexp string
```
qdm12, line 1148: no need for a regexp, the only error case is predictable (`getting value by block hash: test error`)

dot/rpc/modules/state_test.go
```diff
@@ -1129,3 +1129,147 @@ func TestStateModuleQueryStorage(t *testing.T) {
...
+		exp       []StorageChangeSetResponse
```
qdm12, line 1149: nit `exp` -> `response` or `expectedResponse` would be nicer 😉

dot/rpc/modules/state_test.go
```diff
@@ -1129,3 +1129,147 @@ func TestStateModuleQueryStorage(t *testing.T) {
...
+	for name, tt := range tests {
+		tt := tt
+		t.Run(name, func(t *testing.T) {
+			t.Parallel()
+			ctrl := gomock.NewController(t)
+			sm := &StateModule{
+				storageAPI: tt.fields.storageAPIBuilder(ctrl),
+				blockAPI:   tt.fields.blockAPIBuilder(ctrl),
+			}
+			res := []StorageChangeSetResponse{}
```
qdm12, line 1265: nit `res` -> `response` would be nicer 😉

dot/rpc/modules/state.go
```diff
@@ -489,6 +495,41 @@ func (sm *StateModule) QueryStorage(
...
+	response := make([]StorageChangeSetResponse, 0, 1)
+	changes := make([][2]*string, 0, len(req.Keys))
+
+	for _, key := range req.Keys {
+		value, err := sm.storageAPI.GetStorageByBlockHash(&atBlockHash, common.MustHexToBytes(key))
+		if err != nil {
+			return fmt.Errorf("getting value by block hash: %w", err)
+		}
+		var hexValue *string
+		if len(value) > 0 {
```
qdm12, line 517: should we differentiate `[]byte{}` and `[]byte(nil)`?? This comes from the trie and the empty byte slice means it's set. The `nil` byte slice means the value doesn't exist though. I would tend to think we should return `0x` in case the value returned is an empty-non-nil byte slice.

dot/rpc/modules/state.go
kishansagathiya: That would apply to its properties as well.

I don't see any variable of type StateStorageQueryAtRequest getting used by scale. Because scale requires structs and properties to be exported.

kishansagathiya (approved):

dot/rpc/modules/state.go
qdm12: I think they are exported for the encoding/json encoder to work right?

edwardmack: Yes, they are exported for encoding/json encoder to work.

dot/rpc/modules/state.go
edwardmack: Good point, I've updated this function and `state_QueryStorage` to handle `[]byte{}` and `[]byte(nil)` and added tests for both conditions.

dot/rpc/modules/state.go
edwardmack: Yes, we're using Gorilla web toolkit to expose functions through RPC, Gorilla exposes functions with a signature of containing exported structs for request and response.  I tried changing StateStorageQueryAtRequest to un-exported and the RPC function was no longer accessible.

dot/rpc/modules/state.go
```diff
@@ -489,6 +500,42 @@ func (sm *StateModule) QueryStorage(
...
+		} else {
+			if value != nil {
+				hexValue = stringPtr("0x")
+			}
+		}
+
+		changes[i] = [2]*string{stringPtr(key), hexValue}
+	}
+
+	*response = []StorageChangeSetResponse{{
```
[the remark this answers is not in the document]
edwardmack: I think this is because `response` is initialized since it is a function argument.

edwardmack:
This PR has been moved to PR #3191 which contains only code related to implementing state_queryStorageAt (not other merged code from other branches).  @qdm12 please let me know if the remaining comments on this can be resolved, I believe I've addressed all comments in the new PR.

cmd/gossamer/config_test.go
```diff
@@ -845,7 +845,7 @@ func TestGlobalNodeName_WhenNodeAlreadyHasStoredName(t *testing.T) {
 	cfg := newTestConfig(t, westendDevConfig)
 	cfg.Global.Name = globalName
 
-	runtimeFilePath, err := runtime.GetRuntime(context.Background(), runtime.NODE_RUNTIME)
+	runtimeFilePath, err := runtime.GetRuntime(context.Background(), runtime.WESTEND_RUNTIME_v0929)
```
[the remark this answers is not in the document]
edwardmack: This PR has been moved to now PR #3191 which fixes the rebase issue.

PR #3188, opened 2023-04-03
fix(runtime): initialize TransactionState when creating runtime instance

jimjbrettj:
## Changes

- Set transaction state as part of runtime context
- Make it so that the expected type of a KeyOwnershipProof is a scale encoded option for both babe and grandpa
See [babe docs](https://spec.polkadot.network/#sect-babeapi_generate_key_ownership_proof) and [grandpa docs](https://spec.polkadot.network/#sect-grandpaapi_generate_key_ownership_proof)

Note: The reason this change didn't work before is because we do not have proper nil checks on parameter values when creating a runtime instance. Perhaps this would be a good thing to discuss and add to avoid issues like this happening in the future.

## Tests

```sh
go test ./lib/babe --tags=integration 
go test ./lib/grandpa --tags=integration 
go test ./lib/runtime/wasmer
```

## Primary Reviewer

@kishansagathiya

### DIFF
--- a/lib/babe/verify.go
+++ b/lib/babe/verify.go
@@ -364,7 +364,7 @@ func (b *verifier) submitAndReportEquivocation(
 	if err != nil {
 		return fmt.Errorf("getting key ownership proof from runtime: %w", err)
 	} else if keyOwnershipProof == nil {
-		return fmt.Errorf("%w", errEmptyKeyOwnershipProof)
+		return errEmptyKeyOwnershipProof
 	}
 
 	equivocationProof := &types.BabeEquivocationProof{
--- a/lib/babe/verify.go
+++ b/lib/babe/verify.go
@@ -4,6 +4,7 @@
 package babe
 
 import (
+	"errors"
 	"fmt"
 	"sync"
 
@@ -12,6 +13,8 @@ import (
 	"github.com/ChainSafe/gossamer/lib/crypto/sr25519"
 	"github.com/ChainSafe/gossamer/pkg/scale"
 )
+
+var errEmptyKeyOwnershipProof = errors.New("key ownership proof is nil")
 
 // verifierInfo contains the information needed to verify blocks
 // it remains the same for an epoch
@@ -360,6 +363,8 @@ func (b *verifier) submitAndReportEquivocation(
 	keyOwnershipProof, err := runtimeInstance.BabeGenerateKeyOwnershipProof(slot, offenderPublicKey)
 	if err != nil {
 		return fmt.Errorf("getting key ownership proof from runtime: %w", err)
+	} else if keyOwnershipProof == nil {
+		return fmt.Errorf("%w", errEmptyKeyOwnershipProof)
 	}
 
 	equivocationProof := &types.BabeEquivocationProof{
--- a/lib/babe/verify_integration_test.go
+++ b/lib/babe/verify_integration_test.go
@@ -433,37 +433,17 @@ func TestVerifyAuthorshipRight(t *testing.T) {
 	require.NoError(t, err)
 }
 
-// TODO this test failing is related too issue #3136
 func TestVerifyAuthorshipRight_Equivocation(t *testing.T) {
-	t.Skip()
-	kp, err := sr25519.GenerateKeypair()
-	require.NoError(t, err)
-
-	cfg := ServiceConfig{
-		Keypair: kp,
-	}
-
 	genesis, genesisTrie, genesisHeader := newWestendDevGenesisWithTrieAndHeader(t)
-	babeService := createTestService(t, cfg, genesis, genesisTrie, genesisHeader, nil)
+	babeService := createTestService(t, ServiceConfig{}, genesis, genesisTrie, genesisHeader, nil)
+	verificationManager := NewVerificationManager(babeService.blockState, babeService.epochState)
+
 	epochData, err := babeService.initiateEpoch(testEpochIndex)
 	require.NoError(t, err)
 
 	bestBlockHash := babeService.blockState.BestBlockHash()
 	runtime, err := babeService.blockState.GetRuntime(bestBlockHash)
 	require.NoError(t, err)
-
-	epochData.threshold = maxThreshold
-	epochData.authorities = []types.Authority{
-		{
-			Key: kp.Public().(*sr25519.PublicKey),
-		},
-	}
-
-	verifier := newVerifier(babeService.blockState, testEpochIndex, &verifierInfo{
-		authorities: epochData.authorities,
-		threshold:   epochData.threshold,
-		randomness:  epochData.randomness,
-	})
 
 	// slots are 6 seconds on westend and using time.Now() allows us to create a block at any point in the slot.
 	// So we need to manually set time to produce consistent results. See here:
@@ -481,13 +461,13 @@ func TestVerifyAuthorshipRight_Equivocation(t *testing.T) {
 	err = babeService.blockState.AddBlock(block)
 	require.NoError(t, err)
 
-	err = verifier.verifyAuthorshipRight(&block.Header)
+	err = verificationManager.VerifyBlock(&block.Header)
 	require.NoError(t, err)
 
 	err = babeService.blockState.AddBlock(block2)
 	require.NoError(t, err)
 
-	err = verifier.verifyAuthorshipRight(&block2.Header)
+	err = verificationManager.VerifyBlock(&block2.Header)
 	require.ErrorIs(t, err, ErrProducerEquivocated)
 	require.EqualError(t, err, fmt.Sprintf("%s for block header %s", ErrProducerEquivocated, block2.Header.Hash()))
 }
--- a/lib/grandpa/vote_message.go
+++ b/lib/grandpa/vote_message.go
@@ -17,7 +17,10 @@ import (
 	"github.com/libp2p/go-libp2p/core/peer"
 )
 
-var errBeforeFinalizedBlock = errors.New("before latest finalized block")
+var (
+	errBeforeFinalizedBlock   = errors.New("before latest finalized block")
+	errEmptyKeyOwnershipProof = errors.New("key ownership proof is nil")
+)
 
 type networkVoteMessage struct {
 	from peer.ID
@@ -284,6 +287,8 @@ func (s *Service) reportEquivocation(stage Subround, existingVote *SignedVote, c
 	opaqueKeyOwnershipProof, err := runtime.GrandpaGenerateKeyOwnershipProof(setID, pubKey)
 	if err != nil {
 		return fmt.Errorf("getting key ownership proof: %w", err)
+	} else if opaqueKeyOwnershipProof == nil {
+		return fmt.Errorf("%w", errEmptyKeyOwnershipProof)
 	}
 
 	grandpaEquivocation := types.GrandpaEquivocation{
--- a/lib/grandpa/vote_message.go
+++ b/lib/grandpa/vote_message.go
@@ -288,7 +288,7 @@ func (s *Service) reportEquivocation(stage Subround, existingVote *SignedVote, c
 	if err != nil {
 		return fmt.Errorf("getting key ownership proof: %w", err)
 	} else if opaqueKeyOwnershipProof == nil {
-		return fmt.Errorf("%w", errEmptyKeyOwnershipProof)
+		return errEmptyKeyOwnershipProof
 	}
 
 	grandpaEquivocation := types.GrandpaEquivocation{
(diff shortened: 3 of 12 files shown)


lib/runtime/wasmer/exports.go
```diff
@@ -110,13 +113,18 @@ func (in *Instance) BabeGenerateKeyOwnershipProof(slot uint64, authorityID [32]b
...
 
-	keyOwnershipProof := types.OpaqueKeyOwnershipProof{}
+	var keyOwnershipProof *types.OpaqueKeyOwnershipProof
 	err = scale.Unmarshal(encodedKeyOwnershipProof, &keyOwnershipProof)
 	if err != nil {
 		return nil, fmt.Errorf("scale decoding key ownership proof: %w", err)
 	}
 
-	return keyOwnershipProof, nil
+	// TODO check with team if returning error is preferred or should we keep just returning empty value
```
EclesioMeloJunior, line 122: I tend to agree with returning an error since we expect that key to report an equivocation

EclesioMeloJunior (approved):

lib/runtime/wasmer/interfaces.go
```diff
@@ -58,6 +60,7 @@ type BasicNetwork interface {
 // TransactionState is the interface for the transaction state.
 type TransactionState interface {
 	AddToPool(vt *transaction.ValidTransaction) common.Hash
+	PopWithTimer(timerCh <-chan time.Time) (transaction *transaction.ValidTransaction)
```
qdm12, line 63: Why is this addition needed? It compiles fine with it commented 🤔

lib/runtime/interfaces.go
```diff
@@ -50,4 +52,5 @@ type BasicStorage interface {
 // TransactionState interface for adding transactions to pool
 type TransactionState interface {
 	AddToPool(vt *transaction.ValidTransaction) common.Hash
+	PopWithTimer(timerCh <-chan time.Time) (transaction *transaction.ValidTransaction)
```
qdm12, line 55: Why is this addition needed? It compiles fine with it commented 🤔

lib/babe/verify_integration_test.go
```diff
@@ -481,13 +461,13 @@ func TestVerifyAuthorshipRight_Equivocation(t *testing.T) {
...
 
-	err = verifier.verifyAuthorshipRight(&block.Header)
+	err = verificationManager.VerifyBlock(&block.Header)
 	require.NoError(t, err)
 
 	err = babeService.blockState.AddBlock(block2)
 	require.NoError(t, err)
 
-	err = verifier.verifyAuthorshipRight(&block2.Header)
+	err = verificationManager.VerifyBlock(&block2.Header)
```
kishansagathiya, line 470: This test is meant to test `verifyAuthorshipRight`, we can't call `VerifyBlock` directly. We must be calling `verifyAuthorshipRight`

lib/runtime/wasmer/exports.go
```diff
@@ -110,13 +113,18 @@ func (in *Instance) BabeGenerateKeyOwnershipProof(slot uint64, authorityID [32]b
 		return nil, fmt.Errorf("executing %s: %w", runtime.BabeAPIGenerateKeyOwnershipProof, err)
 	}
 
-	keyOwnershipProof := types.OpaqueKeyOwnershipProof{}
+	var keyOwnershipProof *types.OpaqueKeyOwnershipProof
 	err = scale.Unmarshal(encodedKeyOwnershipProof, &keyOwnershipProof)
```
kishansagathiya, line 117:
```suggestion
	err = scale.Unmarshal(encodedKeyOwnershipProof, keyOwnershipProof)
```

keyOwnershipProof is already a pointer. `&keyOwnershipProof` here becomes a pointer to pointer.

lib/runtime/wasmer/exports.go
kishansagathiya: I disagree. All these functions are only limited to just make a runtime call and return results as go structs.
If we get a nil key ownership proof, that still means that we called runtime function correctly and converted the go struct right way. This runtime func returns key ownership proof as scale encoded option. So, `keyOwnershipProof` could be nil.

I don't mind using `errEmptyKeyOwnershipProof`, but this is not the place where we use it.

lib/runtime/wasmer/exports.go
```diff
@@ -297,13 +306,17 @@ func (in *Instance) GrandpaGenerateKeyOwnershipProof(authSetID uint64, authority
 		return nil, err
 	}
 
-	keyOwnershipProof := types.GrandpaOpaqueKeyOwnershipProof{}
+	var keyOwnershipProof *types.GrandpaOpaqueKeyOwnershipProof
 	err = scale.Unmarshal(encodedOpaqueKeyOwnershipProof, &keyOwnershipProof)
```
kishansagathiya, line 310:
```suggestion
	err = scale.Unmarshal(encodedOpaqueKeyOwnershipProof, keyOwnershipProof)
```

kishansagathiya (changes requested):
```
        keyOwnershipProof := types.GrandpaOpaqueKeyOwnershipProof{}
	err = scale.Unmarshal(encodedOpaqueKeyOwnershipProof, &keyOwnershipProof)
```

if `encodedOpaqueKeyOwnershipProof` contains the value scaled encoded nil, would keyOwnershipProof be set to nil?

lib/runtime/wasmer/exports.go
jimjbrettj: I think this is a good point, I will update

lib/runtime/wasmer/exports.go
```diff
@@ -110,13 +113,18 @@ func (in *Instance) BabeGenerateKeyOwnershipProof(slot uint64, authorityID [32]b
...
 	err = scale.Unmarshal(encodedKeyOwnershipProof, &keyOwnershipProof)
 	if err != nil {
 		return nil, fmt.Errorf("scale decoding key ownership proof: %w", err)
 	}
 
-	return keyOwnershipProof, nil
+	// TODO check with team if returning error is preferred or should we keep just returning empty value
+	if keyOwnershipProof == nil {
+		return types.OpaqueKeyOwnershipProof{}, fmt.Errorf("%w", errEmptyKeyOwnershipProof)
+	}
```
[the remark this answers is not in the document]
jimjbrettj: Saw @kishansagathiya asked this too so answering here. Yes, because the expected type is an option and Gossamer handles options as pointers, where the 0 case is nil

lib/babe/verify_integration_test.go
jimjbrettj: So with how we use the VerificationManager in babe, the verifies is a private object so cannot retrieve it to call this receiver which is why I call VerifyBlock. Should I export or create a getter in order to call this specific function? Open to ideas on how to do this

lib/runtime/wasmer/exports.go
jimjbrettj: Yes that is the intended behavior now. Scale takes a pointer to the object, which in this case is a pointer, so a pointer to a pointer is intentional.

lib/runtime/wasmer/exports.go
jimjbrettj: Yes that is the intended behavior now. Scale takes a pointer to the object, which in this case is a pointer, so a pointer to a pointer is intentional.

lib/runtime/wasmer/exports.go
```diff
@@ -297,13 +306,17 @@ func (in *Instance) GrandpaGenerateKeyOwnershipProof(authSetID uint64, authority
...
-	keyOwnershipProof := types.GrandpaOpaqueKeyOwnershipProof{}
+	var keyOwnershipProof *types.GrandpaOpaqueKeyOwnershipProof
 	err = scale.Unmarshal(encodedOpaqueKeyOwnershipProof, &keyOwnershipProof)
 	if err != nil {
 		return nil, fmt.Errorf("scale decoding: %w", err)
 	}
 
-	return keyOwnershipProof, nil
+	if keyOwnershipProof == nil {
+		return types.GrandpaOpaqueKeyOwnershipProof{}, fmt.Errorf("%w", errEmptyKeyOwnershipProof)
```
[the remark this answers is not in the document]
jimjbrettj: Ended up removing this error message

lib/runtime/wasmer/exports.go
```diff
@@ -110,13 +113,18 @@ func (in *Instance) BabeGenerateKeyOwnershipProof(slot uint64, authorityID [32]b
...
+	var keyOwnershipProof *types.OpaqueKeyOwnershipProof
 	err = scale.Unmarshal(encodedKeyOwnershipProof, &keyOwnershipProof)
 	if err != nil {
 		return nil, fmt.Errorf("scale decoding key ownership proof: %w", err)
 	}
 
-	return keyOwnershipProof, nil
+	// TODO check with team if returning error is preferred or should we keep just returning empty value
+	if keyOwnershipProof == nil {
+		return types.OpaqueKeyOwnershipProof{}, fmt.Errorf("%w", errEmptyKeyOwnershipProof)
```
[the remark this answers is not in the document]
jimjbrettj: Ended up removing this error message

lib/runtime/wasmer/interfaces.go
jimjbrettj: good catch, not sure why that ended up being added, thank ya

lib/runtime/wasmer/exports.go
```diff
@@ -110,13 +110,18 @@ func (in *Instance) BabeGenerateKeyOwnershipProof(slot uint64, authorityID [32]b
...
 
-	keyOwnershipProof := types.OpaqueKeyOwnershipProof{}
+	var keyOwnershipProof *types.OpaqueKeyOwnershipProof
 	err = scale.Unmarshal(encodedKeyOwnershipProof, &keyOwnershipProof)
 	if err != nil {
 		return nil, fmt.Errorf("scale decoding key ownership proof: %w", err)
 	}
 
-	return keyOwnershipProof, nil
+	// TODO check with team if returning error is preferred or should we keep just returning empty value
```
EclesioMeloJunior, line 119:
```suggestion
```

EclesioMeloJunior (approved):
LGTM! Just a small adjustment

lib/grandpa/vote_message.go
```diff
@@ -284,6 +287,8 @@ func (s *Service) reportEquivocation(stage Subround, existingVote *SignedVote, c
 	opaqueKeyOwnershipProof, err := runtime.GrandpaGenerateKeyOwnershipProof(setID, pubKey)
 	if err != nil {
 		return fmt.Errorf("getting key ownership proof: %w", err)
+	} else if opaqueKeyOwnershipProof == nil {
+		return fmt.Errorf("%w", errEmptyKeyOwnershipProof)
```
kishansagathiya, line 291: why this error wrapping?

lib/runtime/wasmer/exports_test.go
```diff
@@ -1103,10 +1104,10 @@ func TestInstance_GrandpaGenerateKeyOwnershipProof(t *testing.T) {
 	identityPubKey, _ := ed25519.NewPublicKey(identity)
 	authorityID := identityPubKey.AsBytes()
 
-	encodedOpaqueKeyOwnershipProof, err := instance.GrandpaGenerateKeyOwnershipProof(uint64(0), authorityID)
-	require.NoError(t, err)
+	res, err := instance.GrandpaGenerateKeyOwnershipProof(uint64(0), authorityID)
 	// Since the input is not valid with respect to the instance, an empty proof is returned
-	require.Empty(t, encodedOpaqueKeyOwnershipProof)
+	require.NoError(t, err)
+	require.Nil(t, res)
```
kishansagathiya, line 1110: is this renaming and moving around necessary?

kishansagathiya (approved):
have small comments, but approving to avoid one more review-iteration.

lib/babe/verify.go
```diff
@@ -360,6 +363,8 @@ func (b *verifier) submitAndReportEquivocation(
 	keyOwnershipProof, err := runtimeInstance.BabeGenerateKeyOwnershipProof(slot, offenderPublicKey)
 	if err != nil {
 		return fmt.Errorf("getting key ownership proof from runtime: %w", err)
+	} else if keyOwnershipProof == nil {
+		return fmt.Errorf("%w", errEmptyKeyOwnershipProof)
```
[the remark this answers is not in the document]
jimjbrettj: I thought it would be good to do, but upon further review, I do not think it is needed. Will update in both cases.

lib/runtime/wasmer/exports_test.go
jimjbrettj: In this case I actually did think it was helpful, since to me it is misleading to call the variable encoded since it is unmarshalled. Since there is already a diff from adding the nil require check, I thought it was okay to rename. 

However res is not that descriptive so I will change to opaqueKeyOwnershipProof

Issue #2998, opened 2022-12-09
Updates logs to report when node has started to conform with ZombieNet's start check


edwardmack:
## Issue summary

- ZombieNet greps node log files looking for regular expression `'Listening for new connections|Running JSON-RPC'` to determine that a node has started.  Gossamer nodes do not use either of these terms when reporting it's started.
- Update gossamer node to output message to log `Listening for new connections` or `Running JSON-RPC` when it's started so that ZombieNet will be aware it's started.

## Other information and links

- https://github.com/paritytech/zombienet
- Example command to start ZombieNet with debug logging: `DEBUG=zombie* ./zombienet-linux-x64 spawn --provider native examples/0001-small-network-gossamer.toml`

Issue #2842, opened 2022-09-17
ZombieNet Support and CI Integration


danforbes:
## Task summary

- [ZombieNet](https://github.com/paritytech/zombienet) is a tool to easily spawn ephemeral Polkadot/Substrate networks and perform tests against them. This Epic will track the investigation into the capabilities that must be exposed by Gossamer in order to integrate with ZombieNet, including its [CLI](https://paritytech.github.io/zombienet/cli/index.html) and [DSL/domain-specific language](https://paritytech.github.io/zombienet/cli/test-dsl-definition-spec.html). Once Gossamer can integrate with ZombieNet, ZombieNet should be used to drive enhanced CI/CD testing piplelines for Gossamer in order to make it easier to ensure that Gossamer nodes behave as expected in a number of different network topologies.

qdm12:
Discussing this on https://github.com/paritytech/zombienet/discussions/623 (sorry I wasn't aware/didn't search for this issue)

PR #3172, opened 2023-03-23
feat(zombienet): add zombienet testing to github workflow

edwardmack:
## Changes
This PR adds configuration and test files to add Zombienet testing to github workflow to trigger Zombienet tests on pull requests.

## Tests
Create pull request to trigger workflow action, then review results of workflow to confirm zombienet has executed tests.

## Issues
closes: #2842 
closes: #2843 
closes: #2998 

## Primary Reviewer

@timwu20

PR #3192, opened 2023-04-05
feat(zombienet): add zombienet testing to github workflow

edwardmack:
## Changes
NOTE: This PR replaces PR #3172, which is being replaced because it contains code from other branches.  This PR contains only code relevant to zombienet testing.
 
This PR adds configuration and test files to add Zombienet testing to github workflow to trigger Zombienet tests on pull requests.

## Tests
Create pull request to trigger workflow action, then review results of workflow to confirm zombienet has executed tests.

## Issues

closes: #2842 
closes: #2843 
closes: #2998 

## Primary Reviewer

@timwu20


dot/network/service.go
```diff
@@ -352,6 +358,7 @@ func (s *Service) updateMetrics() {
 		case <-s.ctx.Done():
 			return
 		case <-ticker.C:
+			processStartTimeGauge.Set(float64(time.Now().Unix()))
```
dimartiro, line 361: Is there any way to send it only once? since it's called "process start time" we are not going to have the right value if we are sending it in every tick

dimartiro (changes requested):

zombienet_tests/functional/0001-basic-network.zndsl
```diff
@@ -0,0 +1,15 @@
+Description: Small Network test
```
timwu20, line 1: why the `functional` folder?

dot/network/service.go
timwu20: I agree, this gauge should be set on startup.

edwardmack: Good point, I've moved the processStartTimeGauge.Set to inside the service's start method.

zombienet_tests/functional/0001-basic-network.zndsl
edwardmack: I was replicating what was done on the polkadot repo.

dimartiro (approved):
Great job Ed! 🎉

dimartiro:
Could be great if you could change the makefile to add the zombienet test execution there too

dot/network/service.go
```diff
@@ -87,6 +87,13 @@ var (
 		Name:      "outbound_total",
 		Help:      "total number of outbound streams",
 	})
+	processStartTimeGauge = promauto.NewGauge(prometheus.GaugeOpts{
+		Namespace: "substrate", // Note: this is using substrate namespace because that is what zombienet uses
+		//  to confirm nodes have started TODO: consider other ways to handle this.
```
jimjbrettj, line 92: Is it worth opening an issue for this todo?

zombienet_tests/functional/0001-basic-network.toml
```diff
@@ -0,0 +1,22 @@
+[relaychain]
+chain_spec_path = "chain/westend-local/westend-local-spec-raw.json"
+
+chain = "westend-local"
+
+[[relaychain.nodes]]
+name = "alice"
+command = "gossamer"
+validator = true
+args = ["--key alice", "--babe-lead"]
```
jimjbrettj, line 10: Imagine that this flag will need to be removed with @dimartiro work right?

dimartiro: Yep, I tested the zombienet without this flag using my changes and I worked and I just merged.
@edwardmack could you rebase develop and remove this flag please?

dot/network/service.go
edwardmack: Yes, I've created issue #3205 to track this.

timwu20 (approved):

edwardmack:
> Would be great if you could change the makefile to add the zombienet test execution there too

Good idea, I've added it to Makefile.

Issue #3184, opened 2023-03-30
Remove babe-lead flag


dimartiro:
## Issue summary

- Investigate / discuss about the [babe-lead flag uses](https://github.com/ChainSafe/gossamer/blob/a19ec39d4a5e07b3534ee2e28e8218a13fdd1d13/cmd/gossamer/flags.go#L388-L391) 
- Remove it if it is not necessary

PR #3193, opened 2023-04-05
refactor(cli): Remove babe-lead flag

dimartiro:
## Changes

- Removes babe-lead flag that prevents a node to create blocks until it receive the first one

## Tests

```sh
make tests
```

- manual tests running nodes locally checking that all nodes are producing blocks using the same canonical chain
- also tested using zombienet

## Issues

#3184 

## Primary Reviewer

@timwu20

### DIFF
--- a/lib/babe/babe_integration_test.go
+++ b/lib/babe/babe_integration_test.go
@@ -41,7 +41,6 @@ func TestService_ProducesBlocks(t *testing.T) {
 		Return(nil).MinTimes(2)
 	cfg := ServiceConfig{
 		Authority:          true,
-		Lead:               true,
 		BlockImportHandler: blockImportHandler,
 	}
 
@@ -90,9 +89,7 @@ func TestService_GetAuthorityIndex(t *testing.T) {
 }
 
 func TestStartAndStop(t *testing.T) {
-	cfg := ServiceConfig{
-		Lead: true,
-	}
+	cfg := ServiceConfig{}
 	gen, genTrie, genHeader := newWestendLocalGenesisWithTrieAndHeader(t)
 	bs := createTestService(t, cfg, gen, genTrie, genHeader, nil)
 	err := bs.Start()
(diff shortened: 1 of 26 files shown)


noot:
the reason the --babe-lead flag exists is because the start slot of the network is based on the timestamp of block 1. if there are two conflicting block 1s, then the network will partition due to some nodes having a different start slot than others. I'd ensure this issue is resolved before merging this

dimartiro:
> the reason the --babe-lead flag exists is because the start slot of the network is based on the timestamp of block 1. if there are two conflicting block 1s, then the network will partition due to some nodes having a different start slot than others. I'd ensure this issue is resolved before merging this
> 
> the fork choice rule _should_ pick the same block 1 on every node, but I believe the start slot was being set on first validation which was the problem

Thanks for this comment Elizabeth.

We've made many changes since this flag was added so I think that's because it's solved now.
I did some tests locally and it seems to be working good selecting the right canonical chain but I gonna be doing more tests today to be sure.
I'm also working on use zombienet #3192 to automatize this tests 
What do you think is the best way to test this change? If you have any recommendation will be very helpful

noot:
okay awesome, I just wasn't sure if context was needed for why it was there in the first place lol. if it's working with multiple block producing nodes and multiple block 1s being produced then I think you're good 👍

lib/babe/babe_integration_test.go
```diff
@@ -90,9 +89,7 @@ func TestService_GetAuthorityIndex(t *testing.T) {
 }
 
 func TestStartAndStop(t *testing.T) {
-	cfg := ServiceConfig{
```
jimjbrettj, line 93: Very nice

dimartiro:
> image This is not a blocker I don't think, but do you know why two different blocks are built with the same number here?
> 
> To clarify, you did not introduce this, just an interesting thing I observed that probably should be looked into even if its not in this PR

Very good catch @jimjbrettj ! 👏  I executed the tests in development and it has the same behavior so it is not introduced by my change.

image

PR #3200, opened 2023-04-10
feat(zombienet): Add javascript tests to zombienet testing

edwardmack:
## Changes
This PR introduces code that conducts javascript testing of gossamer nodes using Zombienet testing framework.  The test files shows the following concepts:
- Test javascript code must contain export function called `run` that Zombienet calls as part of the test assertion: `async function run(nodeName, networkInfo, args)` 
- The test script gets access to the testing node via polkadot.js/api by connecting as follows:
  ```
    const {wsUri, userDefinedTypes} = networkInfo.nodesByName[nodeName];
    const api = await zombie.connect(wsUri, userDefinedTypes);
  ```
- The test script can also get reference to keyring function as follows:
  ```
  const keyring = new zombie.Keyring({ type: "sr25519" });
  ```
- Arguments can be passed from test dsl file to test javascript by using the `with` keyword:
   ```
   alice: js-script ./scripts/transfer-tokens.js with "Alice,Bob" return is equal to 1 within 200 seconds
   ```
   The values are acquired in the javascript via the `args` parameter of the `run` function:
   ```
   const FROM = keyring.addFromUri("//" + args[0]);
   ```
- Variables can be used with in the testing dsl file using the format: 
  - `{% set varName = "value" %}` to set values
  - `{{ varName }}` to read values
  
### NOTE: 
The test transfer contained in these tests fails because the gossamer nodes are not finalizing blocks and the test script checks to confirm transfers have been finalized.  The test transfer will run, then timeout and fail. So, these tests should NOT be added to CI workflow until that has been fixed.

## Tests

You can run `make install-zombienet` to download and install zombienet into your `GOPATH`.
```sh
 zombienet test -p native zombienet_tests/functional/0002-basic-network.zndsl 
```

## Primary Reviewer

@timwu20

### DIFF
--- a/dot/network/service.go
+++ b/dot/network/service.go
@@ -340,7 +340,6 @@ func (s *Service) Start() error {
 	logger.Info("started network service with supported protocols " + strings.Join(s.host.protocols(), ", "))
 
 	if s.Metrics.Publish {
-		// TODO(ed) remove before commit
 		processStartTimeGauge.Set(float64(time.Now().Unix()))
 		go s.updateMetrics()
 	}
--- a/dot/network/service.go
+++ b/dot/network/service.go
@@ -340,6 +340,7 @@ func (s *Service) Start() error {
 	logger.Info("started network service with supported protocols " + strings.Join(s.host.protocols(), ", "))
 
 	if s.Metrics.Publish {
+		// TODO(ed) remove before commit
 		processStartTimeGauge.Set(float64(time.Now().Unix()))
 		go s.updateMetrics()
 	}
--- a/dot/rpc/modules/state.go
+++ b/dot/rpc/modules/state.go
@@ -500,7 +500,6 @@ func (sm *StateModule) QueryStorage(
 
 // QueryStorageAt queries historical storage entries (by key) at the block hash given or
 // the best block if the given block hash is nil
-// TODO(ed): remove before merge into development after PR #3191 has been merged
 func (sm *StateModule) QueryStorageAt(
 	_ *http.Request, request *StateStorageQueryAtRequest, response *[]StorageChangeSetResponse) error {
 	atBlockHash := request.At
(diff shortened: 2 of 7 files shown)


dimartiro:
I think we have tu update the [makefile](https://github.com/ChainSafe/gossamer/blob/development/Makefile#L144) too?

dot/rpc/modules/state.go
```diff
@@ -500,6 +500,7 @@ func (sm *StateModule) QueryStorage(
 
 // QueryStorageAt queries historical storage entries (by key) at the block hash given or
 // the best block if the given block hash is nil
+// TODO(ed): remove before merge into development after PR #3191 has been merged
```
EclesioMeloJunior, line 503: was this already addressed or should we address it?

dot/network/service.go
```diff
@@ -340,6 +340,7 @@ func (s *Service) Start() error {
 	logger.Info("started network service with supported protocols " + strings.Join(s.host.protocols(), ", "))
 
 	if s.Metrics.Publish {
+		// TODO(ed) remove before commit
```
EclesioMeloJunior, line 343: does this TODO stills valid?

EclesioMeloJunior (approved):
Some small comments, but LGTM!

dimartiro (approved):
I approve it with one question, should we update the [makefile](https://github.com/ChainSafe/gossamer/blob/development/Makefile#L144) to add the new [test](https://github.com/ChainSafe/gossamer/pull/3200/files)?

jimjbrettj (approved):
nice work, lgtm!

edwardmack:
> I approve it with one question, should we update the [makefile](https://github.com/ChainSafe/gossamer/blob/development/Makefile#L144) to add the new [test](https://github.com/ChainSafe/gossamer/pull/3200/files)?

I don't want to add this test to makefile or CI yet, since it is testing a balance transfer, and currently blocks are not being finalized, so the test fails.

Issue #3202, opened 2023-04-11
refuse to import bad blocks


EclesioMeloJunior:
## Task summary

- Currently, in the chain spec we have a field called `badBlocks`, this field is an array of hashes and any time we face a block with that hash we should ignore it as well as report and disconnect from the peer who sent it to us.

## Other information and links

- https://github.com/paritytech/substrate/pull/4544

PR #3214, opened 2023-04-17
feat(sync): Add badBlocks support

dimartiro:
## Changes

Change sync process to reject bad blocks and disconnect from that peer

## Tests

```sh
make test
```

## Issues

https://github.com/ChainSafe/gossamer/issues/3202

## Primary Reviewer

@EclesioMeloJunior

### DIFF
--- a/dot/sync/chain_sync.go
+++ b/dot/sync/chain_sync.go
@@ -174,6 +174,7 @@ type chainSync struct {
 	finalityGadget     FinalityGadget
 	blockImportHandler BlockImportHandler
 	telemetry          Telemetry
+	badBlocks          []string
 }
 
 type chainSyncConfig struct {
@@ -189,6 +190,7 @@ type chainSyncConfig struct {
 	finalityGadget     FinalityGadget
 	blockImportHandler BlockImportHandler
 	telemetry          Telemetry
+	badBlocks          []string
 }
 
 func newChainSync(cfg chainSyncConfig) *chainSync {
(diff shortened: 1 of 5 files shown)


dot/sync/chain_sync.go
```diff
@@ -1014,6 +1021,13 @@ func (cs *chainSync) validateBlockData(req *network.BlockRequestMessage, bd *typ
 
 	requestedData := req.RequestedData
 
+	for _, badBlockHash := range cs.badBlocks {
+		if bd.Hash.String() == badBlockHash {
+			logger.Errorf("Rejecting known bad block Number: %d Hash: %s", bd.Number(), bd.Hash)
+			return errBadBlock
+		}
+	}
+
```
EclesioMeloJunior, line 1030:
```suggestion
if slices.Contains(cs.badBlocks, bd.Hash.String) {
    logger.Errorf("Rejecting known bad block Number: %d Hash: %s", bd.Number(), bd.Hash)
    return errBadBlock
}
```

EclesioMeloJunior (changes requested):

dot/sync/chain_sync.go
dimartiro: Done! Thks for the suggestion

dimartiro:
I'll close this PR cause I splitted it in two, [this is one](https://github.com/ChainSafe/gossamer/pull/3220) and the second one will be using Eclesio's branch with the new sync to use this validation there

Issue #3203, opened 2023-04-11
Failed to apply a scheduled change: `unfinalized ancestor`


EclesioMeloJunior:
## Describe the bug

- While syncing with Westend the sync process stops after the following error:

```
ERROR    failed to apply scheduled change: cannot get applicable scheduled change: failed while applying condition: unfinalized ancestor
```

The scheduled changes are changes that are placed in the blocks, whenever we see a scheduled change while **importing** a block we store it in the scheduled changes tree, whenever we finalize a block we verify if there is a scheduled change to be applied if there is one applicable change then we should ensure that there is no other change to be applied before it since those changes should be applied sequentially. And seems what is happening here

EclesioMeloJunior:
A more accurate log line: 
```
2023-04-16T21:46:21Z ERROR    failed to apply scheduled change: cannot get applicable scheduled change: failed while applying condition: unfinalized ancestor: 0x0b91df3d578e0a6ebe6510a280662b4ecb7f6048c534c721f45b8857d8f5e48c (1418075)	digest.go:L263	pkg=digest
```

EclesioMeloJunior:
On block number `1,491,033` https://polkadot.js.org/apps/#/explorer/query/1491033 happens a forced change scheduled to happen on block `1,491,043`, Gossamer is not handling this forced change

EclesioMeloJunior:
So after some investigations over the westend subscan (https://westend.subscan.io/event?address=&module=sudo&event=all&startDate=2020-07-22&endDate=2020-07-24&startBlock=&endBlock=&timeType=date&version=9401) I found the sudo call that originates a forced change, on block [#1490983](https://westend.subscan.io/block/1490983) we have a sudo call to the method `grandpa::note_stalled` and the arguments is: `delay: 5; best_finalized_block_number: 1416901`. 

* The block `#1416901` is the latest block finalized in the logs before the problem.
* function called by the sudo call `note_stalled`: https://github.com/paritytech/substrate/blob/785115b3a13901b0c708af8166430bcc9c71f28f/frame/grandpa/src/lib.rs#L246

EclesioMeloJunior:
After some investigation around the forced change dispatched by block `#1,491,033` I found that Gossamer handled the forced change but it was not pruning the stored scheduled changes. Every time a new forced change is applied we should ignore past scheduled changes https://spec.polkadot.network/#sect-finality-forced-changes

[...]

PR #3219, opened 2023-04-19
fix(dot/state): clean up scheduled changes once a forced change is applied

EclesioMeloJunior:
## Changes

After some investigation around the forced change dispatched by block `#1,491,033` I found that Gossamer handled the forced change but it was not pruning the stored scheduled changes. Every time a new forced change is applied we should ignore past scheduled changes or scheduled changes from forks.

- https://spec.polkadot.network/#sect-finality-forced-changes
- https://github.com/paritytech/substrate/blob/50de15d8740a129db9c18a6698fbd183b00326a2/client/consensus/grandpa/src/authorities.rs#L511

After the fix we could reach a higher number of synced blocks (1.79M)
image

## Tests

```sh
go test -timeout 10m -run ^TestApplyForcedChanges$ github.com/ChainSafe/gossamer/dot/state --tags=integration -v
```

## Issues

- Closes https://github.com/ChainSafe/gossamer/issues/3203

## Primary Reviewer

@timwu20


dimartiro (approved):

dot/state/grandpa.go
```diff
@@ -257,9 +257,10 @@ func (s *GrandpaState) ApplyForcedChanges(importedBlockHeader *types.Header) err
 		return fmt.Errorf("cannot set change set id at block")
 	}
 
-	logger.Debugf("Applying authority set forced change at block #%d",
-		forcedChange.announcingHeader.Number)
+	logger.Debugf("Applied authority set forced change: %s", forcedChange)
 
+	s.forcedChanges.pruneAll()
```
jimjbrettj, line 262: Can there be conflicts when applying a forced change and pruning other forced changes? Does it only prune changes that were requested before the current change?

jimjbrettj (approved):
Looks good! Just a few comments

dot/state/grandpa.go
EclesioMeloJunior: > Does it only prune changes that were requested before the current change

EclesioMeloJunior: >Does it only prune changes that were requested before the current change?

No, it prunes the entire set, substrate does it as well https://github.com/paritytech/substrate/blob/50de15d8740a129db9c18a6698fbd183b00326a2/client/consensus/grandpa/src/authorities.rs#L511. 
A forced change once applied should change the entire authority set for all forks https://spec.polkadot.network/#sect-finality-forced-changes

jimjbrettj: awesome thank you

PR #3223, opened 2023-04-20
fix(rpc-tests): Fix node port to execute tests on macOS

### LINKED ISSUE #3222: Fix rpc tests so they can run on macOS
## Issue summary

RPC tests are failing on mac because the port we are using to run our nodes is already binded with a OS proccess

image

Based on our tests this port is used by 

```
/System/Library/CoreServices/ControlCenter.app/Contents/MacOS/ControlCenter
```

We have to change the ports in the test to be able to run these tests in local

dimartiro:
## Changes

Modify node port in rpc-test to run on macOS

## Tests

```sh
make it-rpc
```

## Issues

Closes #3222

## Primary Reviewer

@timwu20

### DIFF
--- a/tests/utils/node/node.go
+++ b/tests/utils/node/node.go
@@ -83,7 +83,7 @@ func (n *Node) setDefaults(t *testing.T) {
 	}
 
 	if n.tomlConfig.Network.Port == 0 {
-		const basePort uint16 = 7000
+		const basePort uint16 = 7004
 		n.tomlConfig.Network.Port = basePort + uint16(*n.index)
 	}
 


EclesioMeloJunior (approved):

timwu20:
Were you able to verify that this change doesn't break interoperability with substrate clients?

dimartiro:
> Were you able to verify that this change doesn't break interoperability with substrate clients?

@timwu20 thanks for the question, this is only for testing purposes when we are starting nodes to run tests

kishansagathiya (approved):

kishansagathiya:
@timwu20 This just changes the port that we use to start a node in our testing setup (`tests/stress` etc). It would not change our interaction with substrate or other clients.

PR #3227, opened 2023-04-26
draft(dot/sync): fix bootstrap/tip sync

EclesioMeloJunior:
## Tests

```sh
```

## Primary Reviewer

@timwu20


chain/westend/defaults.go
```diff
@@ -29,6 +29,8 @@ func DefaultConfig() *cfg.Config {
 	config.Core.GrandpaAuthority = false
 	config.Core.Role = 1
 	config.Network.NoMDNS = false
+	config.Log.Digest = "trace"
```
kanishkatn, line 32: was this used for testing or do we really need the `trace` log level?

kanishkatn: I had missed out on the module log-level parsing. I've added it here #3285

EclesioMeloJunior: Yeah, I forget to remove this delta before committing, btw thanks for the new flags!

dot/network/discovery.go
```diff
@@ -212,7 +212,8 @@ func (d *discovery) findPeers() {
 				continue
 			}
 
-			logger.Tracef("found new peer %s via DHT", peer.ID)
+			//fmt.Printf("%v\n", peer.Addrs)
```
jimjbrettj, line 215: remove logs

dot/network/sync.go
```diff
@@ -13,9 +14,8 @@ import (
 	"github.com/libp2p/go-libp2p/core/peer"
 )
 
-var (
-	blockRequestTimeout = time.Second * 20
-)
+var blockRequestTimeout = time.Second * 20
+var ErrReceivedEmptyMessage = errors.New("received empty message")
```
jimjbrettj, line 18: should this be in errors file?

dot/sync/chain_sync.go
```diff
@@ -21,15 +21,10 @@ import (
...
 	"github.com/ChainSafe/gossamer/dot/network"
 	"github.com/ChainSafe/gossamer/dot/peerset"
+	"github.com/ChainSafe/gossamer/dot/telemetry"
 	"github.com/ChainSafe/gossamer/dot/types"
 	"github.com/ChainSafe/gossamer/lib/blocktree"
 	"github.com/ChainSafe/gossamer/lib/common"
-	"github.com/ChainSafe/gossamer/lib/common/variadic"
-)
-
-const (
```
jimjbrettj, line 30: Do we not have a max anymore?

dot/sync/chain_sync.go
```diff
@@ -53,693 +48,630 @@ func (s chainSyncState) String() string {
...
 
-	// bootstrap or tip (near-head)
-	state chainSyncState
-
-	// handler is set to either `bootstrapSyncer` or `tipSyncer`, depending on the current
-	// chain sync state
-	handler workHandler
-
-	benchmarker *syncBenchmarker
+	state atomic.Value
```
jimjbrettj, line 108: Why was atomic chosen to be used here?

dot/sync/chain_sync.go
```diff
@@ -53,693 +48,630 @@ func (s chainSyncState) String() string {
...
-			}
-		case fin := <-cs.finalisedCh:
-			// on finalised block, call pendingBlocks.removeLowerBlocks() to remove blocks on
-			// invalid forks from the pending blocks set
-			cs.pendingBlocks.removeLowerBlocks(fin.Header.Number)
-		case <-cs.ctx.Done():
-			return
+		gapLength := pendingBlock.number - highestFinalized.Number
+		if gapLength > 128 {
+			logger.Criticalf("GAP LENGHT: %d, GREATER THAN 128 block", gapLength)
```
jimjbrettj, line 427: LENGTH*

dot/sync/chain_sync.go
```diff
@@ -53,693 +48,630 @@ func (s chainSyncState) String() string {
...
+			gapLength = 128
```
jimjbrettj, line 428: Why is this critical and if it is why do we just auto reset to 128?

dot/sync/chain_sync.go
```diff
@@ -53,693 +48,630 @@ func (s chainSyncState) String() string {
...
-		return nil
-	case resultWorker.err.err.Error() == "protocol not supported":
-		cs.network.ReportPeer(peerset.ReputationChange{
-			Value:  peerset.BadProtocolValue,
-			Reason: peerset.BadProtocolReason,
-		}, resultWorker.err.who)
-		cs.ignorePeer(resultWorker.err.who)
-		return nil
-	}
+	// if availableWorkers > maxRequestsAllowed {
```
jimjbrettj, line 467: Is this needed?

dot/sync/chain_sync.go
```diff
@@ -53,693 +48,630 @@ func (s chainSyncState) String() string {
...
-		// if this worker was triggered due to a block in the pending blocks set,
-		// we want to remove it from the set, as we asked all our peers for it
-		// and none replied with the info we need.
-		if worker.pendingBlock != nil {
-			cs.pendingBlocks.removeBlock(worker.pendingBlock.hash)
-		}
-		return nil
+	if targetBlockNumber > realTarget {
+		// basically if our virtual target is beyond the real target
+		// that means we are few requests far from the tip, then we
```
jimjbrettj, line 482: we are only a few requests away*

dot/sync/chain_sync.go
```diff
@@ -53,693 +48,630 @@ func (s chainSyncState) String() string {
...
+			continue
 
-// dispatchWorker begins making requests to the network and attempts to receive responses up until the target
-// if it fails due to any reason, it sets the worker `err` and returns
-// this function always places the worker into the `resultCh` for result handling upon return
-func (cs *chainSync) dispatchWorker(w *worker) {
-	if w.targetNumber == nil || w.startNumber == nil {
-		return
-	}
+		// TODO: implement a case to stop
```
jimjbrettj, line 561: is this still a TODO?

dot/sync/chain_sync.go
```diff
@@ -53,693 +48,630 @@ func (s chainSyncState) String() string {
...
-		outcome := "success"
-		if w.err != nil {
-			outcome = "failure"
-		}
-		logger.Debugf(
-			"sync worker completed in %s with %s for worker id %d",
-			w.duration, outcome, w.id)
-		cs.resultQueue <- w
-	}()
+				cs.workerPool.submitRequest(taskResult.request, workersResults)
```
jimjbrettj, line 587: is this not the case where error is network.ErrReceivedEmptyMessage?

dot/sync/chain_sync_test.go
```diff
@@ -58,1587 +78,587 @@ func Test_chainSyncState_String(t *testing.T) {
...
-					hash:   someHash,
-					number: 2,
-				},
-			},
+			peerID:              somePeer,
+			blockAnnounceHeader: block2AnnounceHeader,
+			errWrapped:          errTest,
+			errMessage:          "while adding pending block header: test error",
 		},
+		//"announced_block_while_in_bootstrap_mode": {},
```
jimjbrettj, line 125: remove comment

dot/sync/requests_test.go
```diff
@@ -0,0 +1,131 @@
...
+					StartingBlock: *variadic.MustNewUint32OrHash(uint32(384)),
+					Direction:     network.Ascending,
+					Max:           &maxResponseSize,
+				},
+			},
+		},
+
+		"requesting_4_chunks_of_128_plus_3_blocks": {
+			startNumber:  0,
+			targetNumber: 512 + 3, // 128 * 4
```
jimjbrettj, line 87: (128 * 4) + 3

dot/sync/syncer.go
```diff
@@ -70,41 +61,104 @@ func NewService(cfg *Config) (*Service, error) {
...
+	// check if they are on a fork or not
+	bestBlockHeader, err := s.blockState.BestBlockHeader()
+	if err != nil {
+		return fmt.Errorf("best block header: %w", err)
+	}
+
+	if blockAnnounceHeader.Number <= bestBlockHeader.Number {
+		// check if our block hash for that number is the same, if so, do nothing
+		// as we already have that block
+		// TODO: check what happens when get hash by number retuns nothing or ErrNotExists
```
jimjbrettj, line 108: Is this still a TODO?

dot/sync/worker_pool.go
```diff
@@ -0,0 +1,270 @@
...
+			if peerSync.punishmentTime.Before(time.Now()) {
+				return peerID
+			}
+		case available:
+			return peerID
+		default:
+		}
+	}
+
+	//could not found an available peer to dispatch
```
jimjbrettj, line 189: could not find*

jimjbrettj (commented):
Nice! Just some comments/questions

dot/sync/chain_sync.go
```diff
@@ -53,693 +48,630 @@ func (s chainSyncState) String() string {
...
+			case errors.Is(err, errResponseIsNotChain):
+				logger.Criticalf("response invalid: %s", err)
+				cs.workerPool.punishPeer(taskResult.who)
+				cs.workerPool.submitRequest(taskResult.request, workersResults)
+				continue
+			case errors.Is(err, errEmptyBlockData):
+				cs.workerPool.punishPeer(taskResult.who)
+				cs.workerPool.submitRequest(taskResult.request, workersResults)
+				continue
+			case errors.Is(err, errUnknownParent):
```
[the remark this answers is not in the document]
EclesioMeloJunior: No, this error could happen since we are retrieving a chain of blocks concurrently, so it is usual for a part of the chain to arrive first then the beginning of the chain, so if the `validateResponse` returns `errUnknownParent` we just skip and include the chain it the `syncingChain` slice.

dot/sync/chain_sync.go
EclesioMeloJunior: Yeah, if we receive `network.ErrReceivedEmptyMessage` we don't punish the peer, IMO doesn't make sense to punish it for don't be able to send us a slice of the chain, so we just submit the same request for another peer

dot/sync/chain_sync.go
EclesioMeloJunior: Nop

dot/sync/chain_sync.go
EclesioMeloJunior: This still under development since it is part of the tip sync implementation, I will resolve it asap

dot/sync/chain_sync.go
```diff
@@ -53,693 +48,630 @@ func (s chainSyncState) String() string {
...
+		highestFinalizedHeader, err := cs.blockState.GetHighestFinalisedHeader()
 		if err != nil {
-			return fmt.Errorf("has header: %w", err)
+			return fmt.Errorf("getting highest finalized header")
 		}
 
-		// if so, do nothing, as we already have their fork
-		if has {
+		// ignore the block if it has the same or lower number
+		if announcedNumber <= highestFinalizedHeader.Number {
```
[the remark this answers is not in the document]
EclesioMeloJunior: Hmm, good question, I will append a TODO to investigate if we should punish the peer's reputation in such case!

dot/sync/chain_sync.go
EclesioMeloJunior: The state is a value that can be read among goroutines and can be written only by one of them. So I choose for using atomic since it provides the write and read safely, but it provides the `CompareAndSwap` which only replaces the value inside the atomic if it is different from what is been written, so basically I don't need to hold a lock only for this field and lock/unlock every time we need to check/change the value.

dot/sync/disjoint_block_set.go
```diff
@@ -26,7 +27,7 @@ var (
 // DisjointBlockSet represents a set of incomplete blocks, or blocks
 // with an unknown parent. it is implemented by *disjointBlockSet
 type DisjointBlockSet interface {
-	run(done <-chan struct{})
+	run(finalisedCh <-chan *types.FinalisationInfo, done <-chan struct{})
```
timwu20, line 30: `done` is usually reserved to signal that something has completed.
```suggestion
	run(finalisedCh <-chan *types.FinalisationInfo, stop <-chan struct{})
```

dot/sync/disjoint_block_set.go
```diff
@@ -286,32 +283,5 @@ func (s *disjointBlockSet) getBlocks() []*pendingBlock {
...
-			continue
-		}
-
-		// if the entire block's data is known, it's ready!
-		ready = append(ready, b.toBlockData())
-		ready = s.getReadyDescendants(c, ready)
-	}
-
-	return ready
+	return maps.Values(s.blocks)
```
timwu20, line 286: so experimental!

dot/sync/requests.go
```diff
@@ -0,0 +1,94 @@
...
+			RequestedData: requestedData,
+			StartingBlock: *start,
+			Direction:     network.Ascending,
+			Max:           &max,
+		}
+		startNumber += uint(max)
+	}
+
+	return reqs
+}
```
timwu20, line 84: I feel like these methods should belong in `network`. I also think they should follow the constructor naming conventions (ie. `NewSingleBlockRequestMessage`).

dot/sync/syncer.go
```diff
@@ -70,41 +61,104 @@ func NewService(cfg *Config) (*Service, error) {
...
+
+	// we assume that if a peer sends us a block announce for a certain block,
+	// that is also has the chain up until and including that block.
+	// this may not be a valid assumption, but perhaps we can assume that
+	// it is likely they will receive this block and its ancestors before us.
+	announcedBlock := announcedBlock{
+		who:    from,
+		header: blockAnnounceHeader,
+	}
+	return s.chainSync.onImportBlock(announcedBlock)
```
timwu20, line 161: nit:
```suggestion
	return s.chainSync.onImportBlock(announcedBlock{
		who:    from,
		header: blockAnnounceHeader,
	})
```

dot/sync/worker_pool.go
```diff
@@ -0,0 +1,270 @@
...
+	err      error
+}
+
+type peerSyncWorker struct {
+	status         byte
+	timesPunished  int
+	punishmentTime time.Time
+}
+
+type syncWorkerPool struct {
```
timwu20, line 44: no unit tests for this type?

dot/sync/disjoint_block_set.go
```diff
@@ -26,7 +27,7 @@ var (
 // DisjointBlockSet represents a set of incomplete blocks, or blocks
 // with an unknown parent. it is implemented by *disjointBlockSet
 type DisjointBlockSet interface {
```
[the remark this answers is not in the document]
EclesioMeloJunior: I think we don't need this interface anymore but just the implementation

dot/network/message.go
```diff
@@ -354,3 +360,79 @@ func (cm *ConsensusMessage) Hash() (common.Hash, error) {
...
 }
+
+// NewSingleBlockRequestMessage returns a request to retrieve a single block
+func NewSingleBlockRequestMessage(blockHash common.Hash, requestedData byte) *BlockRequestMessage {
+	one := uint32(1)
+	return &BlockRequestMessage{
+		RequestedData: requestedData,
+		StartingBlock: *variadic.MustNewUint32OrHash(blockHash),
+		Direction:     Descending,
+		Max:           &one,
```
dimartiro, line 371: Why are we using `*uint32` instead using `uint32`? is there any benefit?

dot/network/sync.go
```diff
@@ -13,9 +13,7 @@ import (
 	"github.com/libp2p/go-libp2p/core/peer"
 )
 
-var (
-	blockRequestTimeout = time.Second * 20
-)
+var blockRequestTimeout = time.Second * 20
```
dimartiro, line 16:
```suggestion
const blockRequestTimeout = time.Second * 20
```

dimartiro (dismissed):

dot/sync/syncer.go
```diff
@@ -70,41 +61,103 @@ func NewService(cfg *Config) (*Service, error) {
...
+	// check if they are on a fork or not
+	bestBlockHeader, err := s.blockState.BestBlockHeader()
+	if err != nil {
+		return fmt.Errorf("best block header: %w", err)
+	}
+
+	if blockAnnounceHeader.Number <= bestBlockHeader.Number {
+		// check if our block hash for that number is the same, if so, do nothing
+		// as we already have that block
+		// TODO: check what happens when get hash by number returns nothing or ErrNotExists
```
jimjbrettj, line 108: Is this still a todo?

dot/sync/worker_pool.go
```diff
@@ -0,0 +1,288 @@
...
+	if !has {
+		return
+	}
+
+	timesPunished := worker.timesPunished + 1
+	punishmentTime := time.Duration(timesPunished) * punishmentBaseTimeout
+	logger.Debugf("⏱️ punishement time for peer %s: %.2fs", who, punishmentTime.Seconds())
+
+	s.workers[who] = &peerSyncWorker{
+		status:         punished,
```
jimjbrettj, line 158: i still think we need a peer jail lol

dot/sync/worker_pool.go
```diff
@@ -0,0 +1,288 @@
...
+
+// getAvailablePeer returns the very first peer available, if there
+// is no peer avaible then the caller should wait for availablePeerCh
+func (s *syncWorkerPool) getAvailablePeer() peer.ID {
+	for peerID, peerSync := range s.workers {
+		switch peerSync.status {
+		case punished:
+			// if the punishedTime has passed then we can
+			// use it as an available peer
+			if peerSync.punishmentTime.Before(time.Now()) {
```
jimjbrettj, line 194: should the status be updated here?

dot/sync/worker_pool.go
```diff
@@ -0,0 +1,288 @@
...
+
+type peerSyncWorker struct {
+	status         byte
+	timesPunished  int
+	punishmentTime time.Time
+}
+
+type syncWorkerPool struct {
+	wg            sync.WaitGroup
+	mtx           sync.RWMutex
```
jimjbrettj, line 46: Do we need to have the mutex as a field since its already in the condvar?

jimjbrettj (commented):
Really awesome work, just a few questions

dot/sync/worker_pool.go
EclesioMeloJunior: The status will be updated by the caller, here we just return the first peer that can be used to retrieve blocks

dot/network/message.go
EclesioMeloJunior: I was trying to understand if this pointer was needed for protobuf encoding/decoding but no, it can be just a normal `uint32`, I can address this in another PR as it seems out of scope for this PR, sounds good?

dot/sync/worker_pool.go
EclesioMeloJunior: yes, the condvar requires a mutex and we still uses the mutex to avoid race conditions among the worker pool

dot/sync/syncer.go
EclesioMeloJunior: No, just removed!

dot/network/message.go
dimartiro: Ok np let's add a TODO to remember it

chain/westend/defaults.go
```diff
@@ -29,6 +29,14 @@ func DefaultConfig() *cfg.Config {
...
 	config.Network.NoMDNS = false
+	config.Network.MaxPeers = 128
+	config.PrometheusExternal = true
+	config.PrometheusPort = 9876
 
+	config.Log.Sync = "trace"
+	config.Log.Digest = "trace"
+
+	config.Pprof.Enabled = true
+	config.Pprof.ListeningAddress = "0.0.0.0:6060"
```
kanishkatn, line 40: a reminder to revert this

dot/network/discovery.go
```diff
@@ -28,7 +28,7 @@ var (
 	startDHTTimeout             = time.Second * 10
 	initialAdvertisementTimeout = time.Millisecond
 	tryAdvertiseTimeout         = time.Second * 30
-	connectToPeersTimeout       = time.Minute * 5
+	connectToPeersTimeout       = time.Minute
```
kanishkatn, line 31: What's the reason for such low timeout?

dot/network/message.go
```diff
@@ -17,6 +17,9 @@ import (
 	"github.com/ChainSafe/gossamer/pkg/scale"
 )
 
+// maxResponseSize is maximum number of block data a BlockResponse message can contain
```
kanishkatn, line 20:
```suggestion
// MaxBlocksInResponse is maximum number of block data a BlockResponse message can contain
```

dot/rpc/modules/gomock_reflect_2744729207/prog.go
```diff
@@ -0,0 +1,87 @@
+// Copyright 2023 ChainSafe Systems (ON)
+// SPDX-License-Identifier: LGPL-3.0-only
+
+
+package main
```
jimjbrettj, line 5: what is this file?

dot/sync/syncer.go
jimjbrettj: I think still needs to be removed

jimjbrettj (approved):
Few comments, but looks great awesome work

dot/network/message.go
```diff
@@ -354,3 +360,80 @@ func (cm *ConsensusMessage) Hash() (common.Hash, error) {
...
+	}
+}
+
+func NewAscedingBlockRequests(startNumber, targetNumber uint, requestedData byte) []*BlockRequestMessage {
+	if startNumber > targetNumber {
+		return []*BlockRequestMessage{}
+	}
+
+	diff := targetNumber - startNumber
+	fmt.Printf("target number: %d, start number: %d = %d\n", targetNumber, startNumber, diff)
```
kishansagathiya, line 391: use logger instead of fmt.Printf ? Also, let's keep it the first thing, before anything gets returned.

dot/network/message.go
```diff
@@ -354,3 +360,80 @@ func (cm *ConsensusMessage) Hash() (common.Hash, error) {
...
+	// start and end block are the same, just request 1 block
+	if diff == 0 {
+		one := uint32(1)
+		return []*BlockRequestMessage{
+			{
+				RequestedData: requestedData,
+				StartingBlock: *variadic.MustNewUint32OrHash(uint32(startNumber)),
+				Direction:     Ascending,
+				Max:           &one,
+			},
```
kishansagathiya, line 402: could just use a single function say `NewBlockRequest` instead of `NewDescendingBlockRequest` and `NewSingleBlockRequestMessage` with an extra direction parameter and reuse it here.

dot/network/message.go
```diff
@@ -354,3 +360,80 @@ func (cm *ConsensusMessage) Hash() (common.Hash, error) {
...
+
+		start := variadic.MustNewUint32OrHash(startNumber)
+
+		reqs[i] = &BlockRequestMessage{
+			RequestedData: requestedData,
+			StartingBlock: *start,
+			Direction:     Ascending,
+			Max:           &max,
+		}
+		startNumber += uint(max)
```
kishansagathiya, line 435: reuse funcs to create new block request?

dot/rpc/modules/gomock_reflect_2744729207/prog.go
kishansagathiya, line 5: This file looks auto-generated, looks like a result of running tests and closing them prematurely. I wonder if there are ways to create these files in temp directory instead of here.

dot/sync/chain_sync.go
```diff
@@ -53,692 +49,709 @@ func (s chainSyncState) String() string {
...
+			bestBlockHeader, err := cs.blockState.BestBlockHeader()
+			if err != nil {
+				logger.Criticalf("getting best block header: %s", err)
+				return
+			}
+
+			cs.workerPool.useConnectedPeers()
+			err = cs.requestMaxBlocksFrom(bestBlockHeader)
+			if err != nil {
+				logger.Errorf("while executing bootsrap sync: %s", err)
```
kishansagathiya, line 263:
```suggestion
				logger.Errorf("requesting max blocks from best block header: %s", err)
```
?

dot/sync/chain_sync.go
```diff
@@ -53,692 +49,709 @@ func (s chainSyncState) String() string {
...
+			}
+		} else {
+			// we are less than 128 blocks behind the target we can use tip sync
+			cs.state.Store(tip)
```
kishansagathiya, line 267: Can you put a comment on the `state` variable about what it is for? May be a comment for `chainSyncState`,`tip` and `bootstrap`

I am guessing that it represents whether we have synced to the tip or not. But I am not sure about it and it is not very apparent to me.

dot/sync/chain_sync.go
```diff
@@ -53,692 +49,709 @@ func (s chainSyncState) String() string {
...
-		worker.peersTried = make(map[peer.ID]struct{})
-	}
+	requests := network.NewAscedingBlockRequests(startRequestAt, targetBlockNumber,
+		network.BootstrapRequestData)
 
-	worker.peersTried[resultWorker.err.who] = struct{}{}
-	cs.tryDispatchWorker(worker)
-	return nil
-}
+	fmt.Printf("===> amount of requests: %d\n", len(requests))
```
kishansagathiya, line 537: use logger

dot/sync/chain_sync.go
```diff
@@ -53,692 +49,709 @@ func (s chainSyncState) String() string {
...
 
-// setMode stops all existing workers and clears the worker set and switches the `handler`
-// based on the new mode, if the mode is different than previous
-func (cs *chainSync) setMode(mode chainSyncState) {
-	if cs.state == mode {
-		return
+	var expectedAmountOfBlocks uint32
+	for _, request := range requests {
+		if request.Max != nil {
+			fmt.Printf("===> request max: %d\n", *request.Max)
```
kishansagathiya, line 542: use logger

dot/sync/chain_sync.go
```diff
@@ -53,692 +49,709 @@ func (s chainSyncState) String() string {
...
+			expectedAmountOfBlocks += *request.Max
+		}
 	}
 
-	// stop all current workers and clear set
-	cs.workerState.reset()
+	fmt.Printf("===> expected amount of blocks: %d\n", expectedAmountOfBlocks)
```
kishansagathiya, line 547: use logger

lib/trie/database.go
```diff
@@ -79,6 +79,10 @@ func (t *Trie) loadNode(db Getter, n *Node) error {
 		}
 
 		nodeHash := merkleValue
+		if len(nodeHash) < 1 {
+			fmt.Printf(">>>>>>>>>> trie loadNode, empty node hash\n")
+		}
+
```
kishansagathiya, line 85: Is this left over from debugging?

dot/rpc/modules/gomock_reflect_2744729207/prog.go
EclesioMeloJunior: Yeah, they are generated by the go:generate command, I don't think it is possible to generate them in another folder, I will delete them

dot/sync/worker_pool.go
```diff
@@ -0,0 +1,287 @@
...
+					s.workers[peerID] = peerSync
+
+					s.mtx.Unlock()
+
+					s.wg.Add(1)
+					go s.executeRequest(peerID, task)
+					break
+				}
+
+				s.availableCond.Wait()
```
timwu20, line 257: I'm confused why we need to use a `sync.Cond` in this case.  In what situation does this line get executed? We will wait on a specific peer to signal on this condition so that that the request can be assigned back to the peer? Maybe some more comments describing the logic here would be best.

timwu20 (changes requested):

dot/sync/chain_sync.go
```diff
@@ -53,692 +49,691 @@ func (s chainSyncState) String() string {
...
+		return cs.requestAnnouncedBlock(announced)
 	}
 
-	cs.workQueue <- ps
-	logger.Debugf("set peer %s head with block number %d and hash %s", p, number, hash)
+	// we are more than 128 blocks behind the head, switch to bootstrap
+	cs.syncMode.Store(bootstrap)
+	isSyncedGauge.Set(0)
+	logger.Debugf("switched sync mode to %d", bootstrap.String())
+	go cs.bootstrapSync()
```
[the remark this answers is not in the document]
EclesioMeloJunior: No, this is inside the `onBlockAnnounce` method, so every time we receive a block announce from the network, among other things, we check if the current sync mode is `bootstrap`, if it is then we just return (which means that the `bootstrapSync` goroutine is running yet, when it finishes then it changes the current sync mode to `tip`) otherwise we check how far we are from the tip of the chain, if needed we execute the `bootstrapSync` goroutine again but before when change the current mode to `bootstrap` to avoid spawning the goroutine more than once

dot/sync/worker_pool.go
EclesioMeloJunior: It makes it easy to handle requests waiting for peers to become available. Basically, when the peers in the pool are all busy then the next task will reach that line so it will wait for the shared `sync.Cond` to signalize when some peer becomes available.

Every worker that finishes handling a task signalizes using `sync.Cond.Signal()` and that is a non-blocking operation.

dot/sync/chain_sync.go
```diff
@@ -53,692 +49,691 @@ func (s chainSyncState) String() string {
...
-		n := len(cs.peerState)
-		cs.RUnlock()
-		if n >= cs.minPeers {
+		totalAvailable := cs.workerPool.totalWorkers()
+		if totalAvailable >= uint(cs.minPeers) {
 			break
 		}
-		time.Sleep(time.Millisecond * 100)
+
+		time.Sleep(time.Second)
```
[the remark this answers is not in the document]
EclesioMeloJunior: Created issue https://github.com/ChainSafe/gossamer/issues/3402

dot/sync/worker_pool.go
timwu20: My concern here is that there will be significant back pressure on the `s.taskQueue` in the event that a task is bound to a specific peer, where the `peerSyncWorker` record shows that it's busy.  Essentially no more tasks are pulled off the `taskQueue` channel in the case where line 257 is reached, even though other peers can execute other tasks in parallel.

I've tried to recreate your solution with `sync.Cond` in this [test](https://github.com/ChainSafe/gossamer/blob/6962ada3c3b2c3430f9173ad338a1966bf13b010/dot/sync/worker_pool_test.go#L390).  I've created an alternative solution using channels to represent a queue per peer with a global concurrency limit in this [test](https://github.com/ChainSafe/gossamer/blob/6962ada3c3b2c3430f9173ad338a1966bf13b010/dot/sync/worker_pool_test.go#L425). The second approach with the queue per peer, will execute transactions for all peers in parallel up to the concurrency limit.

dot/sync/worker.go
```diff
@@ -1,95 +1,106 @@
...
+func (w *worker) start() {
+	defer func() {
+		logger.Debugf("[STOPPED] worker %s", w.peerID)
+		close(w.doneCh)
+	}()
+
+	logger.Debugf("[STARTED] worker %s", w.peerID)
+	for {
+		select {
+		case <-w.stopCh:
```
timwu20, line 52: do we need this case? The dispatcher should be able to close `worker.queue` and this for loop should exit.

dot/sync/worker.go
```diff
@@ -1,95 +1,106 @@
...
-	sync.Mutex
-	nextWorker uint64
-	workers    map[uint64]*worker
+type worker struct {
+	status      byte
+	peerID      peer.ID
+	sharedGuard chan struct{}
+
+	punishment chan time.Duration
+	stopCh     chan struct{}
```
timwu20, line 23: not sure we need this `stopCh`. You can signal to stop the worker by closing `queue`.

dot/sync/worker_pool.go
```diff
@@ -0,0 +1,233 @@
...
+	if who != nil {
+		syncWorker := s.workers[*who]
+		syncWorker.processTask(task)
+		return
+	}
+
+	for syncWorkerPeerID, syncWorker := range s.workers {
+		if who != nil && *who == syncWorkerPeerID {
+			continue
+		}
```
timwu20, line 162: what is this if condition for?

dot/sync/worker_pool.go
```diff
@@ -0,0 +1,233 @@
...
+	// if the request is bounded to a specific peer then just
+	// request it and sent through its queue otherwise send
+	// the request in the general queue where all worker are
+	// listening on
+	s.mtx.RLock()
+	defer s.mtx.RUnlock()
+
+	if who != nil {
+		syncWorker := s.workers[*who]
+		syncWorker.processTask(task)
```
timwu20, line 155: what if it's not enqueued?

dot/sync/worker.go
```diff
@@ -1,95 +1,106 @@
...
-	s.Lock()
-	defer s.Unlock()
-	delete(s.workers, id)
+func (w *worker) processTask(task *syncTask) (enqueued bool) {
+	select {
+	case w.queue <- task:
+		return true
+	default:
+		return false
+	}
```
timwu20, line 66: it may be simpler to just block on this.  What happens if we a worker is block indefinitely and we just continue looping in `syncWorkerPool.submitRequest` trying to assign this task to this queue. We could detect if a worker was indefinitely blocked using this select write default technique.

timwu20 (commented):
mean to leave a comment instead of request for changes.

dot/network/message.go
```diff
@@ -354,3 +360,56 @@ func (cm *ConsensusMessage) Hash() (common.Hash, error) {
...
+		StartingBlock: startingBlock,
+		Direction:     direction,
+		Max:           &amount,
+	}
+}
+
+func NewAscendingBlockRequests(startNumber, targetNumber uint, requestedData byte) []*BlockRequestMessage {
+	if startNumber > targetNumber {
+		return []*BlockRequestMessage{}
+	}
```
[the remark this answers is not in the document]
EclesioMeloJunior: I cannot imagine a scenario where this could happen in production, but this guard is there to avoid weird results and issuing non-sense requests

dot/sync/chain_sync.go
```diff
@@ -53,692 +49,714 @@ func (s chainSyncState) String() string {
...
 
-func newChainSync(cfg chainSyncConfig, blockReqRes network.RequestMaker) *chainSync {
-	ctx, cancel := context.WithCancel(context.Background())
-	const syncSamplesToKeep = 30
-	const logSyncPeriod = 5 * time.Second
-	logSyncTicker := time.NewTicker(logSyncPeriod)
-
+func newChainSync(cfg chainSyncConfig) *chainSync {
+	atomicState := atomic.Value{}
+	atomicState.Store(tip)
```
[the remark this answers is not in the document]
EclesioMeloJunior: Ok, I've changed to only change to bootstrap mode if we need to, basically using the `onBlockAnnounceHandshake` method only

dot/node_integration_test.go
```diff
@@ -325,6 +324,7 @@ func TestStartStopNode(t *testing.T) {
 	config.ChainSpec = genFile
 	config.Core.GrandpaAuthority = false
 	config.Core.BabeAuthority = false
+	config.Network.MinPeers = 0
```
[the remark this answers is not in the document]
EclesioMeloJunior: no

dot/sync/worker_pool.go
EclesioMeloJunior: Not need at all, removed

dot/sync/chain_sync.go
```diff
@@ -53,692 +49,704 @@ func (s chainSyncState) String() string {
...
+	cs.wg.Add(1)
+	go cs.bootstrapSync()
 	return nil
 }
 
-func (cs *chainSync) logSyncSpeed() {
-	defer close(cs.logSyncDone)
-	defer cs.logSyncTicker.Stop()
+func (cs *chainSync) onBlockAnnounce(announced announcedBlock) error {
+	// TODO: create issue to be spec compliat with regard block annoucement
```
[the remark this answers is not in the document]
EclesioMeloJunior: Sure! https://github.com/ChainSafe/gossamer/issues/3432

timwu20 (approved):

dimartiro (approved):
Magnifique 👏

Issue #3217, opened 2023-04-19
Runtime calls are not working 


kanishkatn:
## Describe the bug

It looks like none of the runtime calls are working. I've tried it for all the chains on https://polkadot.js.org/apps by running the nodes locally. 

## Expected Behavior

Runtime calls shouldn't throw any errors.

## Current Behavior

Here's the response to some of the calls I tried
<img width="1725" alt="Screenshot 2023-04-19 at 1 25 05 PM" src="https://user-images.githubusercontent.com/4919645/233009325-7e2de15e-8880-456a-af35-53da79649bd8.png">

## To Reproduce
Steps to reproduce the behaviour:

1. Run a node `./bin/gossamer --chain westend --ws --ws-unsafe --ws-external --base-path ~/.gossamer/westend`
2. Go to https://polkadot.js.org/apps/#/runtime
3. Execute the runtime calls

## Specification

- go version: 1.20
- gossamer branch: `development`
- operating system: macOS

jimjbrettj:
![image](https://user-images.githubusercontent.com/40345116/233729672-9db5760c-f98c-4ff8-9266-5ed125227b5e.png)

From what I have seen, this seems to be because we have not implemented the [state_call](https://github.com/ChainSafe/gossamer/blob/c6b459b26f49f94f5ac53ce773c2dff8dffe8fae/dot/rpc/modules/state.go#L249) method. See above logs for the behavior that I was seeing.

When we perform a call to state_call(core_version, 0x) we fail with this decoding issue, but when we add an actual value to pass is i.e. state_call(core_version, 0x0) we no longer get this error. 

So it seems to me that implementing state_call is the right path forward to fix this

kanishkatn:
fixed in #3228

PR #3228, opened 2023-04-27
(feat/runtime-rpc): implement `state_call` RPC

kanishkatn:
## Changes

Implements `state_call` RPC

## Tests

```
MODE=rpc go test ./tests/rpc/... -timeout=15m -v -run TestStateRPCResponseValidation
```

## Issues

#3217 

## Primary Reviewer

@timwu20 @kishansagathiya @jimjbrettj


dimartiro:
Looks good to me, did you test it running a node and executing calls?

dot/rpc/modules/state.go
```diff
@@ -245,10 +245,24 @@ func (sm *StateModule) GetPairs(_ *http.Request, req *StatePairRequest, res *Sta
...
-// Call isn't implemented properly yet.
-func (sm *StateModule) Call(_ *http.Request, _ *StateCallRequest, _ *StateCallResponse) error {
-	_ = sm.networkAPI
-	_ = sm.storageAPI
+// Call makes a call to the runtime.
+func (sm *StateModule) Call(_ *http.Request, req *StateCallRequest, res *StateCallResponse) error {
+	rt, err := sm.blockAPI.GetRuntime(sm.blockAPI.BestBlockHash())
+	if err != nil {
+		return err
+	}
```
EclesioMeloJunior, line 253: The `state_call` RPC method has an optional field `Block`, so if the `StateCallRequest` contains the block hash information we should get a runtime based on it otherwise we use the best block hash 
```suggestion
        blockHash := sm.blockAPI.BestBlockHash()
        if req.Block != nil {
             blockHash = *req.Block
        }
	rt, err := sm.blockAPI.GetRuntime(blockHash)
	if err != nil {
		return err
	}
```

EclesioMeloJunior (changes requested):

dot/rpc/modules/state.go
kanishkatn: Nice catch, thank you!

kanishkatn:
> Looks good to me, did you test it running a node and executing calls?

Yes @dimartiro and the tests do that as well. Although it needs to cover more usecases. Let's do that in another PR.

EclesioMeloJunior (approved):

dimartiro (approved):

dot/rpc/modules/state.go
```diff
@@ -245,10 +245,29 @@ func (sm *StateModule) GetPairs(_ *http.Request, req *StatePairRequest, res *Sta
 	return nil
 }
 
-// Call isn't implemented properly yet.
-func (sm *StateModule) Call(_ *http.Request, _ *StateCallRequest, _ *StateCallResponse) error {
-	_ = sm.networkAPI
-	_ = sm.storageAPI
+// Call makes a call to the runtime.
+func (sm *StateModule) Call(_ *http.Request, req *StateCallRequest, res *StateCallResponse) error {
+	var blockHash = sm.blockAPI.BestBlockHash()
```
jimjbrettj, line 250: Would it be more effficient to do this in the other order to avoid calling BestBlockHash unless we need to. I don't see a reason to make the call if it is provided

dot/rpc/modules/state_test.go
```diff
@@ -280,15 +282,29 @@ func TestStateModuleGetKeysPaged(t *testing.T) {
...
 	ctrl := gomock.NewController(t)
+	testHash := common.NewHash([]byte{0x01, 0x02})
+	rt := wasmer.NewTestInstance(t, runtime.WESTEND_RUNTIME_v0929)
 
 	mockNetworkAPI := mocks.NewMockNetworkAPI(ctrl)
 	mockStorageAPI := mocks.NewMockStorageAPI(ctrl)
-	sm := NewStateModule(mockNetworkAPI, mockStorageAPI, nil, nil)
+	mockBlockAPI := mocks.NewMockBlockAPI(ctrl)
+	mockBlockAPI.EXPECT().BestBlockHash().Return(testHash)
+	mockBlockAPI.EXPECT().GetRuntime(gomock.Any()).Return(rt, nil)
```
jimjbrettj, line 297: nit: use actual type instead of gomock.Any

dot/rpc/modules/state_test.go
```diff
@@ -280,15 +282,29 @@ func TestStateModuleGetKeysPaged(t *testing.T) {
...
+	sm := NewStateModule(mockNetworkAPI, mockStorageAPI, nil, mockBlockAPI)
 
-	err := sm.Call(nil, nil, nil)
+	req := &StateCallRequest{
+		Method: "Core_version",
+		Params: "0x",
+		Block:  &testHash,
+	}
+	var res StateCallResponse
+	err := sm.Call(nil, req, &res)
```
jimjbrettj, line 307: Do we want to assert on res here?

tests/rpc/rpc_05-state_test.go
```diff
@@ -34,10 +36,17 @@ func TestStateRPCResponseValidation(t *testing.T) { //nolint:tparallel
...
+		rpcCtx, rpcCancel := context.WithTimeout(ctx, 10*time.Second)
+		defer rpcCancel()
+		endpoint := rpc.NewEndpoint(node.RPCPort())
+		data, err := rpc.Post(rpcCtx, endpoint, "state_call", params)
+		require.NoError(t, err)
+
+		err = rpc.DecodeScale(data, &response)
+		require.NoError(t, err)
 
 		// TODO assert stateCallResponse
```
jimjbrettj, line 51: what about this todo?

tests/rpc/rpc_05-state_test.go
```diff
@@ -34,10 +36,17 @@ func TestStateRPCResponseValidation(t *testing.T) { //nolint:tparallel
 	t.Run("state_call", func(t *testing.T) {
 		t.Parallel()
 
-		const params = `["", "","0x580d77a9136035a0bc3c3cd86286172f7f81291164c5914266073a30466fba21"]`
-		var response modules.StateCallResponse
+		const params = `["Core_version", "0x"]`
```
jimjbrettj, line 39: Out of curiosity, why is 0x the case we are testing with?

jimjbrettj:
This works on polkadot.js for me, nice stuff!

dot/rpc/modules/state_test.go
kanishkatn: Yes and it needs to cover all the usecases. It's being tracked here https://github.com/ChainSafe/gossamer/issues/3229

tests/rpc/rpc_05-state_test.go
kanishkatn: https://github.com/ChainSafe/gossamer/issues/3229

tests/rpc/rpc_05-state_test.go
kanishkatn: `Core_version` doesn't take any parameters

dot/rpc/modules/state.go
kanishkatn: It'd lead to a lot of pointer ops which I want to avoid. Do you know a better way to do this?

tests/utils/rpc/request.go
```diff
@@ -93,6 +94,50 @@ func Decode(body []byte, target interface{}) error {
...
+	}
+
+	var result string
+	err = json.Unmarshal(response.Result, &result)
+	if err != nil {
+		return fmt.Errorf("cannot decode response result: %s: %w",
+			string(response.Result), err)
+	}
+
+	scaleRawMessage, err := common.HexToBytes(result)
```
qdm12, line 126: nit `scaleRawMessage` -> `resultBytes`

tests/utils/rpc/request.go
```diff
@@ -93,6 +94,50 @@ func Decode(body []byte, target interface{}) error {
 	return nil
 }
 
+// DecodeScale decodes []body into the target interface.
+// It assumes that the response.Result is a SCALE-encoded value.
+func DecodeScale(body []byte, target interface{}) error {
```
qdm12, line 99: nit use `any` instead of `interface{}`

```suggestion
func DecodeScale(body []byte, target any) error {
```

dot/rpc/modules/state_test.go
```diff
@@ -280,15 +282,28 @@ func TestStateModuleGetKeysPaged(t *testing.T) {
...
 	mockStorageAPI := mocks.NewMockStorageAPI(ctrl)
-	sm := NewStateModule(mockNetworkAPI, mockStorageAPI, nil, nil)
+	mockBlockAPI := mocks.NewMockBlockAPI(ctrl)
+	mockBlockAPI.EXPECT().BestBlockHash().Return(testHash)
+	mockBlockAPI.EXPECT().GetRuntime(testHash).Return(rt, nil)
+
+	sm := NewStateModule(mockNetworkAPI, mockStorageAPI, nil, mockBlockAPI)
 
-	err := sm.Call(nil, nil, nil)
+	req := &StateCallRequest{
```
qdm12, line 301: nit `req` -> `request`

dot/rpc/modules/state_test.go
```diff
@@ -280,15 +282,28 @@ func TestStateModuleGetKeysPaged(t *testing.T) {
...
+		Method: "Core_version",
+		Params: "0x",
+	}
+	var res StateCallResponse
```
qdm12, line 305: nit `res` -> `response`

dot/rpc/modules/state_test.go
```diff
@@ -280,15 +282,28 @@ func TestStateModuleGetKeysPaged(t *testing.T) {
...
 }
 
-// Implement Tests once function is implemented
+// TestCall tests the state_call.
+// TODO: cover more cases in Zombienet tests
+// https://github.com/ChainSafe/gossamer/issues/3229
 func TestCall(t *testing.T) {
 	ctrl := gomock.NewController(t)
+	testHash := common.NewHash([]byte{0x01, 0x02})
+	rt := wasmer.NewTestInstance(t, runtime.WESTEND_RUNTIME_v0929)
```
qdm12, line 291: nit `rt` -> `runtime`

dot/rpc/modules/state_test.go
qdm12: Can we check at least for this single test case? Might be useful to copy paste expectations from here to zombienet tests 🤔

dot/rpc/modules/state.go
```diff
@@ -245,10 +245,31 @@ func (sm *StateModule) GetPairs(_ *http.Request, req *StatePairRequest, res *Sta
...
+	if err != nil {
+		return fmt.Errorf("cannot convert hex data %s to bytes: %w", req.Params, err)
+	}
+
+	response, err := rt.Exec(req.Method, request)
+	if err != nil {
+		return err
+	}
+
+	*res = StateCallResponse(common.BytesToHex(response))
```
qdm12, line 272: Should we check `res` is not a `nil` pointer? I guess it's unlikely, but checking it at the start of the method and returning a sentinel error `return fmt.Errorf("%w", ErrResponsePointerIsNil)` would be nice, if it's `nil`.

I also suppose we require this method signature, but if we don't, it would make more sense to return `StateCallResponse` (as value) instead of accepting a `*StateCallResponse` argument.

dot/rpc/modules/state.go
jimjbrettj: maybe something like:
`var blockHash common.Hash{}
if req.Block != nil {
    blockHash = *req.Block
} else {
      //bs.getBlockHash
}`

Just an idea though, I imagine you can leave as is so feel free to resolve this issue at your own discretion

tests/rpc/rpc_05-state_test.go
jimjbrettj: Could you add this issue number to the comment please

dot/rpc/modules/state_test.go
jimjbrettj: Even if we leave as is, if its being tracked in a different PR could we add a comment with link to the issue tracking it?

jimjbrettj (approved):
lgtm! I had comments about adding issue numbers to comments, but after that looks good to go!

tests/utils/rpc/request.go
```diff
@@ -93,6 +94,50 @@ func Decode(body []byte, target interface{}) error {
...
+		return fmt.Errorf("%w: %s", ErrResponseVersion, response.Version)
+	}
+
+	if response.Error != nil {
+		return fmt.Errorf("%w: %s (error code %d)",
+			ErrResponseError, response.Error.Message, response.Error.ErrorCode)
+	}
+
+	var result string
+	err = json.Unmarshal(response.Result, &result)
```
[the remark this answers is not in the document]
kanishkatn: I believe the intention behind that is to send back the request data to the client. Polkadot-js displays that along with the error message. 

There is no reason to include that in the node response since the client will already have that data. It's better to return just the error message. 

I feel we should just return `ErrResultIsNil`

dot/rpc/modules/state_test.go
kanishkatn: > Can we check at least for this single test case? Might be useful to copy paste expectations from here to zombienet tests 🤔
@qdm12 I'll be doing it the other way around

kanishkatn: > Even if we leave as is, if its being tracked in a different PR could we add a comment with link to the issue tracking it?

@jimjbrettj I've created another issue for this.

dot/rpc/modules/state.go
```diff
@@ -245,10 +245,31 @@ func (sm *StateModule) GetPairs(_ *http.Request, req *StatePairRequest, res *Sta
...
+	}
+
+	rt, err := sm.blockAPI.GetRuntime(blockHash)
+	if err != nil {
+		return err
+	}
+
+	request, err := common.HexToBytes(req.Params)
+	if err != nil {
+		return fmt.Errorf("cannot convert hex data %s to bytes: %w", req.Params, err)
```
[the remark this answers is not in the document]
kanishkatn: I see. Thank you!

tests/rpc/rpc_05-state_test.go
kanishkatn: done

tests/utils/rpc/request.go
```diff
@@ -57,6 +57,7 @@ var (
 )
 
 // Decode decodes []body into the target interface.
+// It assumes that the response.Result is a JSON-encoded value.
```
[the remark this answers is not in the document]
kanishkatn: Good catch! I've updated the comment

dot/rpc/modules/state.go
kanishkatn: I believe the module refactor is out of scope for this task. This is being worked on for `ParachainHost` runtime calls. Could you create an issue for this? I can check with @danforbes on how to prioritise this.

tests/rpc/rpc_05-state_test.go
```diff
@@ -31,15 +33,22 @@ func TestStateRPCResponseValidation(t *testing.T) { //nolint:tparallel
 	getBlockHashCancel()
 	require.NoError(t, err)
 
+	// TODO: Improve runtime tests
+	// https://github.com/ChainSafe/gossamer/issues/3234
```
kishansagathiya, line 37:
```suggestion
	// TODO: Improve runtime tests #3234
```


could do the similar thing at other places as well

tests/rpc/rpc_05-state_test.go
```diff
@@ -31,15 +33,22 @@ func TestStateRPCResponseValidation(t *testing.T) { //nolint:tparallel
...
 	t.Run("state_call", func(t *testing.T) {
 		t.Parallel()
 
-		const params = `["", "","0x580d77a9136035a0bc3c3cd86286172f7f81291164c5914266073a30466fba21"]`
-		var response modules.StateCallResponse
+		const params = `["Core_version", "0x"]`
+		var response runtime.Version
```
kishansagathiya, line 42: move this line right before
```
err = rpc.DecodeScale(data, &response)
```

tests/utils/rpc/request.go
```diff
@@ -93,6 +94,48 @@ func Decode(body []byte, target interface{}) error {
 	return nil
 }
 
+// DecodeScale decodes []body into the target interface.
+// It assumes that the response.Result is a SCALE-encoded value.
+func DecodeScale(body []byte, target any) error {
```
kishansagathiya, line 99: DecodeScale and Decode are quite similar. There might be a way to merge them or reuse code. For example I changed DecodeScale to below, which seems to pass the test where DecodeScale is getting used.

```
func DecodeScale(body []byte, target any) error {
	var result string
	err := Decode(body, &result)
	if err != nil {
		return err
	}

	resultBytes, err := common.HexToBytes(result)
	if err != nil {
		return fmt.Errorf("convert hex to bytes: %w", err)
	}

	err = scale.Unmarshal(resultBytes, target)
	if err != nil {
		return fmt.Errorf("decode response result: %w", err)
	}

	return nil
}
```

kishansagathiya (commented):
:tada: Looks pretty good. Just a few comments!

tests/rpc/rpc_05-state_test.go
kanishkatn: hmm, but why?

kishansagathiya (approved):

dot/rpc/modules/state_test.go
```diff
@@ -280,15 +282,28 @@ func TestStateModuleGetKeysPaged(t *testing.T) {
...
+
+	sm := NewStateModule(mockNetworkAPI, mockStorageAPI, nil, mockBlockAPI)
 
-	err := sm.Call(nil, nil, nil)
+	req := &StateCallRequest{
+		Method: "Core_version",
+		Params: "0x",
+	}
+	var res StateCallResponse
+	err := sm.Call(nil, req, &res)
```
timwu20, line 306: can we assert something in the response?  At least not zero length?

tests/utils/rpc/request.go
```diff
@@ -93,6 +94,28 @@ func Decode(body []byte, target interface{}) error {
...
```
timwu20, line 99:
```suggestion
func DecodeScaleUnmarshal(body []byte, target any) error {
```

kanishkatn: Done xD

dot/rpc/modules/state_test.go
kanishkatn: Good idea! I've added the check

timwu20 (approved):

PR #3233, opened 2023-05-02
(chore/runtime): don't use CoreVersion on creating an instance

kishansagathiya:
## Changes

CoreVersion runtime call is getting used while creating instance, which is not necessary.

Removing the use CoreVersion from NewInstance, let's me reuse NewInstance.

## Tests

```sh
```

## Primary Reviewer

@kanishkatn


lib/runtime/wasmer/instance.go
```diff
@@ -193,13 +193,7 @@ func NewInstance(code []byte, cfg Config) (*Instance, error) {
 	}
 
 	if cfg.testVersion != nil {
-		instance.ctx.Version = *cfg.testVersion
-	} else {
-		instance.ctx.Version, err = instance.version()
```
jimjbrettj, line 198: So to understand, we are avoiding setting the version here to push the installation to the first Version() call, where it can then be cached? Am I understanding this correct?

kishansagathiya: yes, that's correct

jimjbrettj (approved):
I really like this. We have talked about the overhead of instance calls before, and I feel this is a nice solution. Great work!

lib/runtime/wasmer/instance.go
kanishkatn: Does it get refreshed during the runtime upgrades?

kishansagathiya: an instance is an instance of a particular runtime. You in case of a runtime upgrade, we will be dealing with a new instance.

kanishkatn (approved):

PR #3238, opened 2023-05-04
task(runtime): use just one runtime instance interface

kishansagathiya:
## Changes
Right now there are four copies of runtime instance interface. We don't need so many copies. It just creates the extra work when we have to add new runtime call.

## Tests

```sh
```

## Primary Reviewer

@kanishkatn


kishansagathiya:
@EclesioMeloJunior That would create cyclical dependency. What is however possible is something like `lib/runtime/interfaces/instance.go`

kanishkatn:
Could we rename `runtimeinterface` to just `runtime`? It seems a bit long for me

kishansagathiya:
> Could we rename runtimeinterface to just runtime? It seems a bit long for me

It's there to avoid conflict with `lib/runtime` package.

kanishkatn:
> > Could we rename runtimeinterface to just runtime? It seems a bit long for me
> 
> It's there to avoid conflict with `lib/runtime` package.

I feel it's better to import it as ` import runtimeinterface "github.com/chainsafe/gossamer/lib/runtime/interfaces` than use the long package name.

kishansagathiya:
> I feel it's better to import it as  import runtimeinterface "github.com/chainsafe/gossamer/lib/runtime/interfaces than use the long package name.

it's something i am considering, https://github.com/ChainSafe/gossamer/pull/3238#issuecomment-1535719780

kishansagathiya:
> would be better to move this Instance interface to lib/runtime/interfaces.go?

@EclesioMeloJunior @kanishkatn Was able to do this without cyclical dependencies

EclesioMeloJunior (approved):

kanishkatn (approved):

Issue #3186, opened 2023-03-31
Implement ParachainHost runtime calls 


kishansagathiya:
## Issue summary

- Implement all the runtime call mentioned in https://spec.polkadot.network/#sect-anv-runtime-api

- [x] [ParachainHost_validators](https://spec.polkadot.network/#sect-rt-api-validators)
- [x] [ParachainHost_validator_groups](https://spec.polkadot.network/#sect-rt-api-validator-groups)
- [x] [ParachainHost_availability_cores](https://spec.polkadot.network/#sect-rt-api-availability-cores)
- [x] [ParachainHost_persisted_validation_data](https://spec.polkadot.network/#sect-rt-api-persisted-validation-data)
- [x] [ParachainHost_check_validation_outputs](https://spec.polkadot.network/#id-parachainhost_check_validation_outputs)
- [x] [ParachainHost_session_index_for_child](https://spec.polkadot.network/#id-parachainhost_session_index_for_child)
- [x] [ParachainHost_validation_code](https://spec.polkadot.network/#sect-rt-api-validation-code)
- [x] [ParachainHost_validation_code_by_hash](https://spec.polkadot.network/#sect-rt-api-validation-code-by-hash)
- [x] [ParachainHost_candidate_pending_availability](https://spec.polkadot.network/#id-parachainhost_candidate_pending_availability)
- [x] [ParachainHost_candidate_events](https://spec.polkadot.network/#id-parachainhost_candidate_events)
- [x] [ParachainHost_session_info](https://spec.polkadot.network/#sect-rt-api-session-info)
- [ ] [ParachainHost_dmq_contents](https://spec.polkadot.network/#id-parachainhost_dmq_contents)
- [ ] [ParachainHost_inbound_hrmp_channels_contents](https://spec.polkadot.network/#id-parachainhost_inbound_hrmp_channels_contents)

## Other information and links

-

kishansagathiya:
Only two runtime calls left here. I am going to close this in favour of a new issue for just those two issue. https://github.com/ChainSafe/gossamer/issues/3595

PR #3241, opened 2023-05-05
feat/runtime: Add few parachain runtime calls

kanishkatn:
## Changes

Added the ParachainHost runtime exports. It implements the following

- ParachainHost_validators
- ParachainHost_validator_groups
- ParachainHost_availability_cores
- ParachainHost_check_validation_outputs
- ParachainHost_session_index_for_child
- ParachainHost_candidate_pending_availability
- ParachainHost_candidate_events
- ParachainHost_session_info

## Tests

```
go test ./dot/types/
```

## Issues

#3186 

# Tasks

- [x] Fix SessionInfo 
      - #3253 
      - #3274
- [x] Find a way to write tests for `exports.go`

# Pending calls

- [ ] `ParachainHostDMQContents` related to #3296 
- [ ] `ParachainHostInboundHrmpChannelsContents` related to #3296 

## Primary Reviewer

@kishansagathiya


dot/types/parachain.go
```diff
@@ -0,0 +1,428 @@
...
+
+	"github.com/ChainSafe/gossamer/lib/common"
+	"github.com/ChainSafe/gossamer/lib/crypto"
+	"github.com/ChainSafe/gossamer/lib/crypto/sr25519"
+	"github.com/ChainSafe/gossamer/pkg/scale"
+	"github.com/centrifuge/go-substrate-rpc-client/v4/types"
+)
+
+// BlockNumber represents a block number
+type BlockNumber uint32
```
dimartiro, line 17: Can we move it to `dot/types/block.go` ?

dot/types/parachain.go
```diff
@@ -0,0 +1,428 @@
...
+
+// ValidatorID represents a validator ID
+type ValidatorID [sr25519.PublicKeyLength]byte
```
dimartiro, line 20: I think it should be in `dot/types/validator.go`

dimartiro (changes requested):
I think we could split some types from parachain.go into different files for each specific type

dot/types/parachain.go
kanishkatn: done

dot/types/parachain.go
```diff
@@ -0,0 +1,388 @@
...
+	"fmt"
+
+	"github.com/ChainSafe/gossamer/lib/common"
+	"github.com/ChainSafe/gossamer/lib/crypto"
+	"github.com/ChainSafe/gossamer/lib/crypto/sr25519"
+	"github.com/ChainSafe/gossamer/pkg/scale"
+	"github.com/centrifuge/go-substrate-rpc-client/v4/types"
+)
+
+// GroupRotationInfo represents the group rotation info
```
kishansagathiya, line 16: This is not informative enough. I know that probably rust also does not provide a great comment. 
It is unclear here what a group is made up of (validators?).

dot/types/parachain.go
```diff
@@ -0,0 +1,388 @@
...
+
+// ValidatorGroups represents the validator groups
+type ValidatorGroups struct {
+	// Validators is an array of validator set Ids
+	Validators [][]ValidatorIndex `scale:"1"`
+	// GroupRotationInfo is the group rotation info
+	GroupRotationInfo GroupRotationInfo `scale:"2"`
+}
+
+// ParaID The ID of a para scheduled.
```
kishansagathiya, line 34: What does `para scheduled` mean? 
This is basically ID for a parachain.

dot/types/parachain.go
```diff
@@ -0,0 +1,388 @@
...
+type ParaID uint32
+
+// GroupIndex The unique (during session) index of a validator group.
+type GroupIndex uint32
+
+// CollatorID represents a collator ID
```
kishansagathiya, line 40:
```suggestion
// CollatorID represents public key of a collator.
```

dot/types/parachain.go
kishansagathiya, line 1: I am against having parachain related types in `dot/types`

`dot/types` mostly contains types that are used across the project. Ideally, all types should be in their respective packages. By respective packages, I mean packages where they logically fit (this could be different than where they are used the most). Sometimes, it is challenging to do this because of cyclical dependencies, hence we have `dot/types`. But ideally `dot/types` should not exist.

In case of parachains, It would mean a new package say `lib/parachains`. One could also consider parachains as part of block building, so then `lib/babe` is also a good place to keep it.

kishansagathiya: Personally, I would prefer to keep `parachain_inherents.go` as it is.

Create, a new package, `lib/parachains/types` and put all new parachain related types there. 

If there are duplicated types with `parachain_inherents.go`, let them be. Remove duplicated types in a separate PR because it is going touch  a bunch of extra code, which is not relevant to runtime host functions.

dot/types/parachain_test.go
```diff
@@ -0,0 +1,271 @@
...
...37c2b49429a082633c8c7abbe502084c28c4a548ac75f97732521c7e66ea9917e704e5469572558c1e00000013000000003808000026662318f64c51f6ed779aa7e32b1b7489bbdb6118812d1b0a2902b7405bc43278b266c6c3ba60a6e648093bcd62419d4ea68eec772f22b7dfc78c6b255dac57bd149cd2b1b166d73be2ed9409a0bebe842c1901d65b1fd3c5f8d98d82ea2a5c3ae10acc8770276d737e6334c59e468be878e6264c050f48ac3e4a6524b76b37c399c4a820e1a50b2bcbc805192c2609e8372e8ddd8432127f6200427dcb2d72386429023e01fcf193d96cc55f1d4d85db5b2f0a9e3a5fb6855c0e07b891d6581ea371a39cb30eaa0ea2d0b9bbe5734e8793d4db7c366277619bdff9611dfb8d4bba89ec371746a5e689ece6f2a117d2b4e7894776b7a96209e7177446c076874bde474bbe3f7bb78669eb48648bf238110f690c144c097988314989bf9b3d4ef410cd043b32cf71a25a00655ab6a0558e383e546c3728412be664273b51908f4903de6835c05f9da3c6ca7fc65846438da7947f93a152f77bb75a85e3d05385feb47e071300b797adcbfd7d3a94ba4b2e0ab454bfded856f759d2b674839d4adcacb72d2c1361edc442862859f46d384db20db351390b2d261be35211db5232e585f985291a08066e6d6273804852f3c0a603d7da194fbff174c24700f3f89d86b462760033bdcb1663c76c60056e6d62730101e2f5f706646cbfc613a98564ff9b4a958acead02fa488e1f3bbcee114d13594d75e9b79d5f5be257e9da81d3ab97cefe36a620a2384f6f1f8719ec2b23532684280000001d000000" //nolint:all
```
kishansagathiya, line 256: Create a separate file `test_data.go` and put all such test data there.

kishansagathiya (commented):
Great work! 

I have put my initial round of comments. Will take one more look once these are addressed.

There some of the comments which are not useful. One made up example would be `CollatorID is ID of collator`.
See if you can add more information there. There are times when linter will complain if there isn't a comment, so you have to keep a comment. But if the comment is for a struct field, I don't think linter would complain. So, it does not add value, you can just remove it in those cases.

dot/types/parachain_test.go
kishansagathiya: @timwu20 @EclesioMeloJunior opinions, yay or nay?

kanishkatn: It just looks large in the Github's PR interface. In reality, it is just [one line](https://github.com/ChainSafe/gossamer/blob/feat/k/parachain-runtime/lib/parachain/types_test.go#L22) and doesn't affect the readability

dot/types/block.go
```diff
@@ -9,6 +9,9 @@ import (
 	"github.com/ChainSafe/gossamer/pkg/scale"
 )
 
+// BlockNumber represents a block number
+type BlockNumber uint32
```
jimjbrettj, line 13: Block numbers are actually type uint in our code, could this cause possible issues for encoding/decoding? 

Following up on this, what is the reason for this type being created? Is it just for readability? I'm not opposed to it, just want to make sure I understand its context

dot/types/validator.go
```diff
@@ -0,0 +1,46 @@
+// Copyright 2023 ChainSafe Systems (ON)
+// SPDX-License-Identifier: LGPL-3.0-only
+
+package types
```
jimjbrettj, line 4: should we have tests for this file?

lib/parachain/types.go
```diff
@@ -0,0 +1,388 @@
...
+)
+
+// GroupRotationInfo A helper data-type for tracking validator-group rotations.
+type GroupRotationInfo struct {
+	// SessionStartBlock is the block number at which the session started
+	SessionStartBlock uint32 `scale:"1"`
+	// GroupRotationFrequency indicates how often groups rotate. 0 means never.
+	GroupRotationFrequency uint32 `scale:"2"`
+	// Now indicates the current block number.
+	Now uint32 `scale:"3"`
```
jimjbrettj, line 24: I stated this above but gonna reiterate here, I'm a little concerned as to why we are representing it as a uint32 here. Is there a reason we are deviating from how we represent block number in the header? Worried about a situation where the block itself can process the number but other parts of the code cant

lib/parachain/types.go
```diff
@@ -0,0 +1,388 @@
...
+// Index returns the index
+func (Free) Index() uint {
+	return 2
+}
+
+// CoreState represents the state of a particular availability core.
+type CoreState scale.VaryingDataType
+
+// AvailabilityCores represents a slice of CoreState
+type AvailabilityCores scale.VaryingDataTypeSlice
```
jimjbrettj, line 146: Im wondering if for here and for CandidateEvents (and others if I missed them) if we should have a function to set the value when using it in exports.go.

Kinda like how we do [here](https://github.com/ChainSafe/gossamer/blob/85ebb2c9a4b60ab23006e80e31bd10ca05bb19c1/dot/types/digest.go#L24) but where we take a param.

Not sure if this is needed/preferred but putting it here as a thought

lib/parachain/types_test.go
```diff
@@ -0,0 +1,273 @@
...
+func Test_Validators(t *testing.T) {
+	t.Parallel()
+
+	resultHex := "0x444c8e878d7f558ce5086cc37ca0d5964bed54ddd6b15a6663a95fe42e368589360e101de266b8f5f05431dcaf63ecd936988cc348a271f42519bef19df1e9af7f4eb63e4aad805c06dc924e2f19b1dde7faf507e5bb3c1838d6a3cfc10e84fe72a262f83b46310770ae8d092147176b8b25e8855bcfbbe701d346b10db0c5385d94ef34321ca5d37a6e8953183406b76f8ebf6a4be5eefc3997d022ac6e0a050e804b9df571e2b744d65eca2d4c59eb8e4345286c00389d97bfc1d8d13aa6e57ecee3f65d78a239d7d199b100295e7a2d852ae898a6b81fd867b3471f25be7237ac837e8ca589521a83e7d9a7b307d1c41a5d9b940422488236f99646d21f3841ae40e895f46c8bfb3df63c119047d7faf21c3fe3e7a91994a3f00da6fa80f848e61d138eebd2069f1a76b3570f9de6a4b196289b198e33e6f0b59cef8837c511e2ac8f039eb02370a9577e49ffc6032e6b5bf5ff77783bdc676d1432d714fd53ce35fa64fe7a5a6fc456ed2830e64d5d1a5dba26e7a57ab458f8cedf1ec77016a0e038975cff34d01c62960828c23ec10a305fe9f5c3589c2ae40f51963e380a0825ba7677597ec9453ab5dbaa9e68bf89dc36694cb6e74cbd5a9a74b167e547b61cb85f7cf7616f9ef8f95010a51a68a4eae8afcdff715cc6a8d43da4a32a12382f17dae6b13a8ce5a7cc805056d9b592d918c8593f077db28cb14cf08a760c807fa54347a8957ff5ef6c28e2403c83947e5fad4aa805c914df0645a07aab5a" //nolint:all
```
jimjbrettj, line 22: where were the results of this retrieved from?

dot/types/parachain_test.go
jimjbrettj: It looks like there is quite a lot of result data no? Specifically in `TestSessionInfo`

jimjbrettj (commented):
Gave this a look through, but will wait to review further until tests are implemented

dot/types/block.go
kanishkatn: > Block numbers are actually type uint in our code, could this cause possible issues for encoding/decoding?

substrate uses `u32`. What's the reason for using `uint`?

> what is the reason for this type being created? Is it just for readability?

Yes, and usability

dot/types/validator.go
kanishkatn: Forgot to remove the methods in this. Thank you.

It doesn't need tests now.

lib/parachain/types_test.go
kanishkatn: through the runtime calls. Added comments for this.

lib/parachain/types.go
dimartiro: Can we create a `BlockNumber` type alias in `dot/types/block.go` and start using it in every place?

kanishkatn: Ah, I had missed it there. Good catch, thank you!

kanishkatn: Resolving this to keep the conversation in one place. Feel free to open it if something isn't addressed

dot/types/block.go
kishansagathiya: > Block numbers are actually type uint in our code

No, BlockNumbers are uint32 at most places in gossamer. `types.Header` uses uint though. `BlockAnnounceHandshake`, `ProveFinalityRequest` and many other places use uint32. I do remember a discussion about this and I think substrate+polkadot itself doesn't use the same type for block number across their codebase.

In this particular situation, we don't have a choice to choose between uint and uint32. We have to make our types identical to types used by polkadot. So, we have to use uint32.

>  could this cause possible issues for encoding/decoding?

If I rememeber well, uint and uint32 would produce different encoding.

lib/parachain/types.go
```diff
@@ -0,0 +1,403 @@
...
+}
+
+// IndexedValidatorGroup A validator group with its group index.
+type IndexedValidatorGroup struct {
+	GroupIndex []GroupIndex           `scale:"1"`
+	Validators []types.ValidatorIndex `scale:"2"`
+}
+
+// AuthorityDiscoveryID An authority discovery key.
+type AuthorityDiscoveryID [32]byte
```
kishansagathiya, line 244:
```suggestion
type AuthorityDiscoveryID [sr25519.PublicKeyLength]byte
```

lib/parachain/types.go
```diff
@@ -0,0 +1,403 @@
...
+type IndexedValidator struct {
+	Index     []types.ValidatorIndex `scale:"-"`
+	Validator []types.ValidatorID    `scale:"2"`
+}
+
+// IndexedValidatorGroup A validator group with its group index.
+type IndexedValidatorGroup struct {
+	GroupIndex []GroupIndex           `scale:"1"`
+	Validators []types.ValidatorIndex `scale:"2"`
+}
```
kishansagathiya, line 241: I can't find corresponding rust structs, can you point me to them?

lib/parachain/types.go
```diff
@@ -0,0 +1,403 @@
...
+	return 0
+}
+
+// CandidateIncluded This candidate receipt was included and became a parablock at the most recent block.
+// This includes the core index the candidate was occupying as well as the group responsible
+// for backing the candidate.
+type CandidateIncluded struct {
+	CandidateReceipt CandidateReceipt `scale:"1"`
+	HeadData         HeadData         `scale:"2"`
+	CoreIndex        uint32           `scale:"3"`
```
kishansagathiya, line 340:
```suggestion
	CoreIndex        CoreIndex           `scale:"3"`
```

EclesioMeloJunior (approved):
LGMT!

lib/runtime/wasmer/exports_test.go
```diff
@@ -1188,3 +1213,266 @@ func TestInstance_GrandpaSubmitReportEquivocationUnsignedExtrinsic(t *testing.T)
...
+			9,
+			6,
+			13,
+			15,
+			5,
+			0,
+			8,
+			10,
+			2,
+		},
```
kishansagathiya, line 1386: let's put this in one line.

lib/runtime/wasmer/exports.go
```diff
@@ -346,6 +347,174 @@ func (in *Instance) GrandpaSubmitReportEquivocationUnsignedExtrinsic(
 	return nil
 }
 
+// ParachainHostValidators Returns the validator set at the current state.
```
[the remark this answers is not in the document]
kanishkatn: Why? Technically, that's the beginning of the sentence, isn't it?

lib/runtime/wasmer/exports.go
```diff
@@ -346,6 +347,174 @@ func (in *Instance) GrandpaSubmitReportEquivocationUnsignedExtrinsic(
...
+func (in *Instance) ParachainHostValidators() ([]parachain.ValidatorID, error) {
+	encodedValidators, err := in.Exec(runtime.ParachainHostValidators, []byte{})
+	if err != nil {
+		return nil, fmt.Errorf("exec: %w", err)
+	}
+
+	var validatorIDs []parachain.ValidatorID
+	err = scale.Unmarshal(encodedValidators, &validatorIDs)
+	if err != nil {
+		return nil, fmt.Errorf("unmarshal: %w", err)
```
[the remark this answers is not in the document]
kanishkatn: hmm, interesting

lib/runtime/wasmer/exports.go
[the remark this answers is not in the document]
kishansagathiya: No. The sentence starts at `ParachainHostValidators`.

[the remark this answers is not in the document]
kanishkatn: No, that's the name of the function. The explanation sentence starts after that.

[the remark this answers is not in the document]
kanishkatn: Yeah, no you're right!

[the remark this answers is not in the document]
kanishkatn: My entire life has been a lie!

dimartiro (approved):
LGTM! Good work Kishan!

kanishkatn:
Thank you for adding the interface @kishansagathiya !

PR #3242, opened 2023-05-06
Diego/fix bootstrap sync

dimartiro:


dot/sync/worker_pool.go
```diff
@@ -114,19 +105,8 @@ func (s *syncWorkerPool) releaseWorker(who peer.ID) {
...
-		return
-	}
-
 	s.workers[who] = &peerSyncWorker{status: available}
-
-	if s.waiting {
-		s.waiting = false
-		s.availablePeerCh <- who
-	}
+	s.availablePeerCh <- who
```
EclesioMeloJunior, line 109: This could block the `handleWorkersResults` since the only way to release a peer is by calling `releaseWorker` method and the method will put the peer in the `s.availablePeerCh`, so if there are no requests waiting in the queue that means we never listen to this channel, so the `handleWorkersResults` will hang until someone listen to that channel to unblock and finish the `releaseWorker`

dimartiro: It's a good catch but I'm not sure, since it's a buffered channel, so you can put a max of maxRequestsAllowed in there upfront. It could block only if you have the channel fulled of available peers.

Issue #3248, opened 2023-05-08
Support `BitVec` in scale


kanishkatn:
## Issue summary

Extend `pkg/scale` to support [BitVec](https://github.com/paritytech/parity-scale-codec/blob/master/src/bit_vec.rs).  

## Tasks

- [x] A basic implementation for ParachainHost runtime exports #3253
- [ ] Support `BitOrder` https://github.com/ChainSafe/gossamer/pull/3253#discussion_r1192257419
- [ ] https://github.com/ChainSafe/gossamer/pull/3253#discussion_r1196935277

## Other information and links
-  #3241

axaysagathiya:
According to my observation and discussion with @EclesioMeloJunior, we need only the LSB0 ordering that we already have.
So, I guess this issue is not relevant anymore.

PR #3253, opened 2023-05-11
feat/scale: add `BitVec`

kanishkatn:
## Changes

Added support for `BitVec`.

This PR only implements the basic version of the `BitVec`. We'll be adding features to it when needed.

## Tests

```sh
go test pkg/scale
```

## Issues

#3248

## Primary Reviewer
@timwu20


pkg/scale/bitvec.go
```diff
@@ -0,0 +1,108 @@
...
+	var b []byte
+	for i := uint(0); i < bv.size; i += bitSize {
+		end := i + bitSize
+		if end > bv.size {
+			end = bv.size
+		}
+		chunk := bv.bits[i:end]
+		b = append(b, bitsToBytes(chunk)...)
+	}
+	return b
```
EclesioMeloJunior, line 60: I would say you don't need to create chunks since the function `bitsToBytes` already do that, so you can pass the entire bit slice to it
```suggestion
	return bitsToBytes(bv.bits)
```

EclesioMeloJunior (changes requested):

EclesioMeloJunior (approved):

dimartiro (approved):

pkg/scale/bitvec.go
```diff
@@ -0,0 +1,99 @@
...
+	// byteSize is the number of bits in a byte
+	byteSize = 8
+)
+
+// BitVec represents rust's `bitvec::BitVec` in SCALE
+// It is encoded as a compact u32 representing the number of bits in the vector
+// followed by the actual bits, rounded up to the nearest byte
+type BitVec interface {
+	// Bits returns the bits in the BitVec
+	Bits() []uint8
```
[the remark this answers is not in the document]
kanishkatn: We're keeping it basic for now since this is being worked on for ParachainHost runtime calls. I've updated the PR and the issue.

pkg/scale/bitvec.go
```diff
@@ -0,0 +1,99 @@
...
+	// Bytes returns the byte representation of the Bits
+	Bytes() []byte
+	// Size returns the number of bits in the BitVec
+	Size() uint
+}
+
+// bitVec implements BitVec
```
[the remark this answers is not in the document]
kanishkatn: That's okay. I'd like to keep it.

pkg/scale/bitvec.go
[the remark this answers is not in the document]
kanishkatn: For this PR, I've found it easy to work with `[]uint`. If it makes sense to change it bool in the future, we could do that.

pkg/scale/bitvec.go
```diff
@@ -0,0 +1,99 @@
...
+
+package scale
+
+const (
+	// maxLen equivalent of `ARCH32BIT_BITSLICE_MAX_BITS` in parity-scale-codec
+	maxLen = 268435455
+	// byteSize is the number of bits in a byte
+	byteSize = 8
+)
+
```
EclesioMeloJunior, line 12: I would suggest including this type assertion, we have those in the code to ensure in comptime if a type implements the interface
```suggestion
var _ BitVec = (*bitVec)(nil)
```

pkg/scale/bitvec.go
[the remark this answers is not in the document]
EclesioMeloJunior: I take a look at how substrate implements `BitVec` and they use a crate called `bitvec` ([scale-codec dependency](https://github.com/paritytech/parity-scale-codec/blob/aaea65ab7b7d0cded3e37f42a819ccedb5db645f/Cargo.toml#LL16C4-L16C4), [create repo](https://github.com/ferrilab/bitvec)), and in the README of this create they explain:

> bitvec provides a foundational API for bitfields in Rust. It specializes standard-library data structures (slices, arrays, and vectors of bool) to use one-bit-per-bool storage, similar to [std::bitset<N>](https://en.cppreference.com/w/cpp/utility/bitset) and [std::vector<bool>](https://en.cppreference.com/w/cpp/container/vector_bool) in C++.

So, I agree with @kishansagathiya that it would be more compatible to use a slice of `bool` as it is used in `parity-scale-codec`.

pkg/scale/bitvec.go
```diff
@@ -0,0 +1,99 @@
...
+	// Bits returns the bits in the BitVec
+	Bits() []uint8
+	// Bytes returns the byte representation of the Bits
+	Bytes() []byte
+	// Size returns the number of bits in the BitVec
+	Size() uint
+}
+
+// bitVec implements BitVec
+type bitVec struct {
```
EclesioMeloJunior, line 26: I notice that in the `parity-scale-codec` in the `src/bit_vec.rs` ([ref to the code line](https://github.com/paritytech/parity-scale-codec/blob/aaea65ab7b7d0cded3e37f42a819ccedb5db645f/src/bit_vec.rs#LL24C1-L24C1)) all implementations of `Encode` and `Decode` contains a generic called `BitOrder` which is related to least significant bit (Lsb) and Most significant bit (Msb), and in looking in the [test in the file](https://github.com/paritytech/parity-scale-codec/blob/aaea65ab7b7d0cded3e37f42a819ccedb5db645f/src/bit_vec.rs#L90) we can see that it tests using both order types. Should we make the bits ordering a required parameter for encoding/decoding in our implementation?

EclesioMeloJunior (changes requested):

pkg/scale/bitvec.go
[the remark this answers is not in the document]
kanishkatn: Thanks for looking that up @EclesioMeloJunior . I'll see if I can do it in this PR itself!

pkg/scale/bitvec.go
kanishkatn: Yes, it can be picked up later when needed. I've updated the issue for this.

kanishkatn: I don't know. I haven't looked into it since the task is to implement the basic version of `bitvec` which supports only `encode` and `decode` at the moment. 

@timoftime Let me know if this PR needs to be more exhaustive.

pkg/scale/bitvec.go
```diff
@@ -0,0 +1,91 @@
...
+
+const (
+	// maxLen equivalent of `ARCH32BIT_BITSLICE_MAX_BITS` in parity-scale-codec
+	maxLen = 268435455
+	// byteSize is the number of bits in a byte
+	byteSize = 8
+)
+
+// BitVec is the implementation of the bit vector
+type BitVec struct {
```
timwu20, line 14: I don't think this should belong in the `scale` package.  Given that we will probably need a `BitVec` at some other point, I think it makes sense to create a `bitvec` package to house this?

pkg/scale/bitvec.go
```diff
@@ -0,0 +1,91 @@
...
+	size uint   `scale:"1"`
+	bits []bool `scale:"2"`
```
timwu20, line 16: I would assume you're adding the scale struct tags given that you need to length encode the size of this.  I don't think is required.  After reading the `parity-scale-codec` encoding of `BitVec` ([link](https://github.com/paritytech/parity-scale-codec/blob/372d6177d8e66cdfe0968c200c8e29fda674bf14/src/bit_vec.rs)):

```
impl<O: BitOrder, T: BitStore + Encode> Encode for BitSlice<T, O> {
	fn encode_to<W: Output + ?Sized>(&self, dest: &mut W) {
		let bits = self.len();
		assert!(
			bits <= ARCH32BIT_BITSLICE_MAX_BITS,
			"Attempted to encode a BitSlice with too many bits.",
		);
		Compact(bits as u32).encode_to(dest);

		// Iterate over chunks
		for chunk in self.chunks(core::mem::size_of::<T>() * 8) {
			let mut element = T::ZERO;
			element.view_bits_mut::<O>()[..chunk.len()].copy_from_bitslice(chunk);
			element.encode_to(dest);
		}
	}
}

impl<O: BitOrder, T: BitStore + Encode> Encode for BitVec<T, O> {
	fn encode_to<W: Output + ?Sized>(&self, dest: &mut W) {
		self.as_bitslice().encode_to(dest)
	}
}
```

It looks like the encoding of a `BitVec` is essentially the encoding of the underlying `BitSlice`, where the encoding is essentially a length encoded (ie. `Compact(u32)`) slice of a certain generic type `T`.  This is directly analogous to the scale encoding of a slice of a given type `T`, so`[]T` in Go. 

So whether or not you actually need to support `BitVec` in our interpretation needs to be researched.  If the use case from the Parity devs is to just get the most memory efficient implementation of describing a `Vec<u32>` on their side, it doesn't mean we need to re-implement `BitVec` and `BitSlice` in Go to ensure we can interoperate with the runtime.  Maybe it's just easier to map it as `[]T` in our use cases knowing that it's not the most memory efficient representation at this time.

Feel free to reach out and we can sync up again if any of this is unclear.

pkg/scale/bitvec.go
kanishkatn: In that case, shouldn't we do that for `uint128` and `VaryingDataType` as well?

pkg/scale/bitvec.go
kanishkatn: I agree. Thank you!

pkg/scale/bitvec.go
EclesioMeloJunior: Ok, I don't think we should extend more this task, but I think we should choose LSB to encode/decode a slice of bits as it looks like Polkadot is using this as the main ordering, see the [query to see the usage of bitvec::order::Lsb0](https://github.com/search?q=repo%3Aparitytech%2Fpolkadot%20bitvec%3A%3Aorder%3A%3ALsb0&type=code), but I didn't found any usage of `Msb0` ordering. So you could highlight that the `bitsToBytes` function uses the `lsb` ordering. 

Just for exemplification here is an example of encoding using both ordering: https://go.dev/play/p/4rCqCZA8lma

pkg/scale/bitvec.go
timwu20: Yea.. I put those two in the `scale` package for other reasons:
- In rust `u128` is primitive type, and we don't have a native 128 bit unsigned integer implementation at this point.   I can argue the other side, where they should be their packages, but since the SCALE type system is essentially a subset of rust primitives, I thought it made sense to keep them in the same package.
- Similarly `VaryingDataType`/`enum` and `Result` are not supported in go natively, but are part of the supported types as defined in the [spec](https://spec.polkadot.network/id-cryptography-encoding#sect-scale-codec).

Looks like in the spec they do mention the use of `bitvec::BitVec` as [Hex encoding](https://spec.polkadot.network/id-cryptography-encoding#sect-scale-codec).  If we never use anything else from the `bitvec` other than `BitVec` I don't mind the go version of `BitVec` to live in `scale`.

kanishkatn: Makes sense. I had used the `pkg/scale` in one of my projects before and found it convenient just to import one package. 

Centrifuge mentions our package [here](https://github.com/centrifuge/go-substrate-rpc-client/tree/master/scale). I feel we should also consider the external usability.

pkg/scale/bitvec.go
kanishkatn: I've removed the `size`

EclesioMeloJunior (approved):

pkg/scale/bitvec.go
timwu20: > Centrifuge mentions our package here and it looks like the other scale codec lib isn't being maintained. I feel we should also consider the external usability.

I think this was before I did the scale rewrite for our codebase.

timwu20: Given the incomplete nature of this PR does this need to be merged into `development` now? Maybe this can be included on the release branch for the parachain work? I'd rather not introduce code into `development` that isn't used.

kanishkatn: This is a blocker for https://github.com/ChainSafe/gossamer/pull/3241. Are we looking to merge all the parachain work to this release branch?

timwu20: Yes I think that makes sense.  We are doing the same for the GRANDPA work.

kanishkatn: Ok.

Issue #2989, opened 2022-12-05
Populate more data in parachain inherents


kishansagathiya:
## Issue summary
- Parachain Inherents have four fields Parent Header, Bitfields, BackedCandidates and Disputes.
- However, we have populated only Parent Header, because that is our minimum requirement for a node to run (build blocks).
- We would like to populate other fields as well.

- The scope of this issue would be find where to and how to get those remaining information (Bitfields, BackedCandidates and Disputes).
- It would most like require a larger parachain support. In which case, figure out what extra feature we might need to populate these data and create other issues accordingly.

## Updates after research
This is a large task, one which requires building systems written below

- [ ] Collator Protocol
https://github.com/ChainSafe/gossamer/issues/3204
- [ ] Backing Subsystems
https://github.com/ChainSafe/gossamer/issues/3247
- [ ] Availability Subsystems
- [ ] Approval Subsystems
- [ ] Dispute Subsystems
- [ ] Some utility Subsystems
- [ ] Implement ParachainHost runtime calls
This can also be done independently https://github.com/ChainSafe/gossamer/issues/3186 

Details of how these subsystems work in polkadot is https://paritytech.github.io/polkadot/book/node/subsystems-and-jobs.html while spec is at https://spec.polkadot.network/#chapter-anv

Basically in order to populate parachain inherents we should be able to get data from collators, that data will make use of all the above mentioned subsystems and in the end we will get parachain inherents for us to put into new blocks.

I feel like the first systems for us to implement should be collator protocol. The Collator Protocol implements the network protocol by which collators and validators communicate. It is used by collators to distribute collations to validators and used by validators to accept collations by collators. 

The Collator Protocol is the first in the journey of a parachain block to become part of the relay chain. Thus, we can test this implementation independently. Whereas, all other systems interact with other not-yet-implemented systems, which would require us to mock behaviour.  

## Other information and links

- 
There is the [Node Implementer's Guide](https://paritytech.github.io/polkadot/book/node/index.html). Node implementer explains how parachain host is implemented in Polkadot (rust implementation). It is really helpful because 
- you can get the exact definitions of data types, 
- description is more detailed than spec
- you can learn an example of software architecture of parachain host
 

The is also the [spec](https://spec.polkadot.network/#chapter-anv). Personally, I find spec easy to understand. Spec is good to know just the requirement. With node implementer's guide one can get confused on what is the requirement and what is a implementation choice.

kishansagathiya:
[...]

PR #3254, opened 2023-05-11
feat(lib/runtime): add extra required runtime imports for parachain validation

kishansagathiya:
This PR adds following import calls:
- ext_default_child_storage_clear_prefix_version_2
- ext_default_child_storage_root_version_2
- ext_offchain_index_clear_version_1

## Tests

```sh
```

## Issues

#2989

## Primary Reviewer

@timwu20


kanishkatn (approved):
looks good!

kishansagathiya:
A corresping PR in polkadot spec repo for tests to work https://github.com/ChainSafe/polkadot-spec/pull/15

lib/runtime/wasmer/imports_test.go
```diff
@@ -1283,6 +1301,54 @@ func Test_ext_default_child_storage_clear_version_1(t *testing.T) {
 	require.Nil(t, val)
 }
 
+func Test_ext_default_child_storage_clear_prefix_version_2(t *testing.T) {
```
EclesioMeloJunior, line 1304: Could you include a test when the limit is not 0?

lib/runtime/wasmer/imports.go
```diff
@@ -1060,6 +1063,33 @@ func ext_default_child_storage_clear_prefix_version_1(context unsafe.Pointer, ch
...
+		logger.Errorf("failed to decode limit: %s", err)
+	}
+
+	err = storage.ClearPrefixInChildWithLimit(keyToChild, prefix, *limit)
+	if err != nil {
+		logger.Errorf("failed to clear prefix in child with limit: %s", err)
+	}
+
+	// TODO: Should this always be 0 or could this be something else as well?
+	return C.int64_t(0)
```
EclesioMeloJunior, line 1090: Following the spec the returned value should be a pointer-size to the following variant:

- Variant 0 indicates that all keys of the child storage have been removed, followed by the number of removed keys
- Variant 1 indicates that there are remaining keys, followed by the number of removed keys.

ref: https://spec.polkadot.network/chap-host-api#id-version-2-prototype-3

You can see how they implemented it in substrate:

```rs
#[derive(PassByCodec, Encode, Decode)]
pub enum KillStorageResult {
	/// All keys to remove were removed, return number of iterations performed during the
	/// operation.
	AllRemoved(u32),
	/// Not all key to remove were removed, return number of iterations performed during the
	/// operation.
	SomeRemaining(u32),
}

#[version(2)]
fn clear_prefix(&mut self, prefix: &[u8], limit: Option<u32>) -> KillStorageResult
```
- clear prefix v2: https://github.com/paritytech/substrate/blob/d58481569817f4a2074dbab014fa629dea7e0ec9/primitives/io/src/lib.rs#L186
- the returned variant: https://github.com/paritytech/substrate/blob/d58481569817f4a2074dbab014fa629dea7e0ec9/primitives/io/src/lib.rs#L96

EclesioMeloJunior (changes requested):

lib/runtime/storage/trie.go
```diff
@@ -259,6 +259,21 @@ func (s *TrieState) ClearPrefixInChild(keyToChild, prefix []byte) error {
...
+	s.lock.Lock()
+	defer s.lock.Unlock()
+
+	child, err := s.t.GetChild(keyToChild)
+	if err != nil {
+		return 0, false, err
+	}
+	if child == nil {
+		return 0, false, nil
+	}
```
EclesioMeloJunior, line 272:
```suggestion
	if err != nil || child == nil {
		return 0, false, err
	}
```

EclesioMeloJunior (approved):
LGTM, just a small suggestion

PR #3256, opened 2023-05-12
chore(logs): Remove unneeded logs

timwu20:
## Changes

- Remove unneeded logs introduced from a previous commit.

## Tests

```sh
```

## Primary Reviewer

@jimjbrettj


jimjbrettj (approved):
For context, the idea was to take out the log from the runtime code and log in the caller. However, if we don't feel they are needed removing them is fine with me

kishansagathiya (approved):

PR #3262, opened 2023-05-16
client(consensus/grandpa): initial authority implementation

jimjbrettj:
## Changes

- created a new package client/consensus/grandpa to house the grandpa standalone package integration 
- Initial logic for AuthoritySet management. This is a WIP but the current code is stable so I would like reviews on what I have so far
- Created an interface to mock out ForkTree while I work on the ForkTree implementation
- Based on [this file](https://github.com/paritytech/substrate/blob/master/client/consensus/grandpa/src/authorities.rs) in substrate

## Tests

```sh
go test ./client/consensus/grandpa -v
```

## Primary Reviewer

@timwu20


jimjbrettj:
> @jimjbrettj why do we need to reimplement the authority changes if we have an already stable implementation?
> 
>     * https://github.com/ChainSafe/gossamer/blob/development/dot/state/grandpa_changes.go
> 
>     * https://github.com/ChainSafe/gossamer/blob/development/dot/state/grandpa.go
> 
> 
> I think we don't need to reimplement it to simulate the substrate structures as it looks stable and works without problems

My understanding was that the goal of integrating the standalone grandpa package was to do basically a 1:1 mapping of how substrate does it. Since the existing code is not that and is not within even the current grandpa package, I thought that this would be included as part of the rewrite. However, if we decide it's better not to do a full rewrite for the integration code and use some of the existing code I can try to integrate with it. I do think however that since it isn't a 1:1 mapping it could cause added complexity.

What do you think @timwu20?

client/consensus/grandpa/authorities.go
```diff
@@ -0,0 +1,210 @@
...
+//
+// This will be applied when the announcing block is at some depth within
+// the finalized or unfinalized chain.
+type PendingChange struct {
+	nextAuthorities AuthorityList
+	delay           uint
+	canonHeight     uint
+	canonHash       common.Hash
+	delayKind       DelayKind
+}
```
timwu20, line 27: can we not use generics here?  Substrate defines a `H` type and `N` type to represent hash and number.  
```
pub struct PendingChange<H, N> {
	/// The new authorities and weights to apply.
	pub(crate) next_authorities: AuthorityList,
	/// How deep in the chain the announcing block must be
	/// before the change is applied.
	pub(crate) delay: N,
	/// The announcing block's height.
	pub(crate) canon_height: N,
	/// The announcing block's hash.
	pub(crate) canon_hash: H,
	/// The delay kind.
	pub(crate) delay_kind: DelayKind<N>,
}
```
You can use `constraints.Unsigned` for `H` and `comparable` for `N`.

client/consensus/grandpa/authorities.go
```diff
@@ -0,0 +1,210 @@
...
+
+// EffectiveNumber Returns the effective number this change will be applied at.
+func (pc *PendingChange) EffectiveNumber() uint {
+	return pc.canonHeight + pc.delay
+}
+
+// AuthoritySetChanges Tracks historical authority set changes. We store the block numbers for the last block
+// of each authority set, once they have been finalized. These blocks are guaranteed to
+// have a justification unless they were triggered by a forced change.
+type AuthoritySetChanges []struct {
```
timwu20, line 37: you should probably define an `AuthoritySetChange` and define this as `AuthoritySetChanges []AuthoritySetChange`.

Also please use the generic type `N` for block number.

client/consensus/grandpa/authorities.go
```diff
@@ -0,0 +1,210 @@
...
+	setId       uint64
+	blockNumber uint
```
timwu20, line 39: These attributes will need to be public.

client/consensus/grandpa/authorities.go
```diff
@@ -0,0 +1,210 @@
...
+}
```
timwu20, line 40: no receiver methods for this type?  It looks like there are `impl` blocks for this type in substrate ([link](https://github.com/paritytech/substrate/blob/7c8b89b3a6b603902709fc33fdb10a864b8b4204/client/consensus/grandpa/src/authorities.rs#L717))

client/consensus/grandpa/utils.go
```diff
@@ -0,0 +1,21 @@
+// Copyright 2023 ChainSafe Systems (ON)
+// SPDX-License-Identifier: LGPL-3.0-only
+
+package grandpa
+
+import "fmt"
+
+// SearchKey TODO for reviewer, this can be done by slices.BinarySearch however since it is a tuple being compared,
+// it's unclear to me what value to sort on
+func SearchKey(key Key, changes []PendingChange) (int, error) {
```
timwu20, line 10: this is wrong, it needs to return the index of where the key (in this case the block number) should be placed in the slice if it's not found.  I use it in grandpa [here](https://github.com/ChainSafe/gossamer/blob/332b301f0c86ae8fe53f33a8acb0cba868e2a356/pkg/finality-grandpa/dummy_chain_test.go#L104).  You need to ensure the slice is sorted.

timwu20 (changes requested):

client/consensus/grandpa/authorities.go
jimjbrettj: I actually do this already in my follow up PR.

jimjbrettj:
Closed in favor of #3283

PR #3267, opened 2023-05-18
0.9.42 runtimes

danforbes:
## Changes

0.9.42 runtimes

## Tests

Manual w/ westend-local

## Primary Reviewer

@timwu20


EclesioMeloJunior (approved):

EclesioMeloJunior:
While trying to fix the tests I face this WARN: 
```
2023-06-01T08:22:24-04:00 WARN     failed to handle slot 280937024: cannot build inherents: error applying inherent: transaction validity error: mandatory dispatch error	pkg=babe
```
So gossamer was not able to produce blocks since it fails to apply the inherents correctly. I was not able to dive deep into the problem yet but I am wondering if this could be related to the parachain inherents.

In the polkadot repository the parachain inherents is arguments for this pallet method `enter` `https://github.com/paritytech/polkadot/blob/9b1fc27cec47f01a2c229532ee7ab79cc5bb28ef/runtime/parachains/src/paras_inherent/mod.rs#L288`, and this methods has a weight definition:

```rs
#[pallet::weight((
	paras_inherent_total_weight::<T>(
		data.backed_candidates.as_slice(),
		data.bitfields.as_slice(),
		data.disputes.as_slice(),
	),
	DispatchClass::Mandatory,
))]
pub fn enter(
``` 

And currently we don't have any information other than just the parent header:

```go
parachainInherent := inherents.ParachainInherentData{
	ParentHeader: *parent,
}
```

So I **suppose** that the `weight` for that pallet method is `0` and this is causing the `mandatory dispatch error` 

@kanishkatn @timwu20

danforbes:
Repllaced by #3353

PR #3279, opened 2023-05-24
feat(lib/runtime): `wazero` implementation of `runtime.Instance`

### LINKED ISSUE #2585: Implement our runtime with github.com/tetratelabs/wazero
## Task summary

1. Implement the same functionality as wasmer with [wazero](https://github.com/tetratelabs/wazero) in `lib/runtime`
2. Remove the `lib/runtime/wasmer` implementation

This is likely at least one week of work since all the code lib/runtime has to be modified, and tests adapted to work with wazero without cgo.

## Outcomes

1. Faster debugging: no more CGO, so no more weird `unknown pc pointer` stack traces
2. Less dependencies: no need for glibc and libwasmer.so
4. Faster Go compilation
5. Lighter Docker images since we can now have alpine based or scratch based Docker images

- Performance wise, I do not think it would matter if it's slower. Our performance should be mostly hit by the scale codec code.

timwu20:
## Changes

- `wazero` implementation of `runtime.Runtime` interface
-  moved some `wasmer` interfaces to `runtime`.

## Tests

```sh
```

## Issues

closes #2585 

## Primary Reviewer

@EclesioMeloJunior


lib/runtime/wazero/instance.go
```diff
@@ -0,0 +1,828 @@
...
+
+	ctx := context.WithValue(context.Background(), runtimeContextKey, i.Context)
+
+	values, err := runtimeFunc.Call(ctx, api.EncodeI32(int32(inputPtr)), api.EncodeI32(int32(dataLength)))
+	if err != nil {
+		return nil, fmt.Errorf("running runtime function: %w", err)
+	}
+	if len(values) == 0 {
+		return nil, fmt.Errorf("no returned values from runtime function: %s", function)
+	}
```
EclesioMeloJunior, line 433: is expected for every runtime function to return a value?

EclesioMeloJunior (changes requested):
Looking forward to having these changes!! Just a small comment over the `Exec` method and I would suggest replacing `panic("wtf")` with something more detailed 😂

lib/runtime/wazero/instance.go
timwu20: In our case yes.  Any functions that we call exported by the Polkadot/Westend runtimes return something.

lib/runtime/wasmer/imports_test.go
```diff
@@ -1173,6 +1174,8 @@ func Test_ext_crypto_sr25519_sign_version_1(t *testing.T) {
 	require.NoError(t, err)
 	require.NotNil(t, val)
 
+	fmt.Println("val", val)
```
EclesioMeloJunior, line 1177:
```suggestion

```

EclesioMeloJunior (approved):
LGTM

dot/helpers_test.go
```diff
@@ -38,7 +38,7 @@ func newWestendDevGenesisWithTrieAndHeader(t *testing.T) (
 	require.NoError(t, err)
 	gen = *genPtr
 
-	genesisTrie, err = wasmer.NewTrieFromGenesis(gen)
+	genesisTrie, err = runtime.NewTrieFromGenesis(gen)
```
jimjbrettj, line 41: Why move this into runtime package?

lib/runtime/wazero/imports.go
```diff
@@ -0,0 +1,2247 @@
...
+	size := uint32(len(data))
+	pointer, err := allocator.Allocate(size)
+	if err != nil {
+		return 0, fmt.Errorf("allocating: %w", err)
+	}
+
+	ok := m.Memory().Write(pointer, data)
+	if !ok {
+		return 0, fmt.Errorf("out of range")
+	}
```
jimjbrettj, line 74: Do we not need to copy memory here any more?

We used to do:
`copy(memory[ptr:ptr+size], data)
pointerSize = toPointerSize(ptr, size)`
	
Why do we not need this copy any more?

lib/runtime/wazero/imports.go
```diff
@@ -0,0 +1,2247 @@
...
+func ext_crypto_ed25519_verify_version_1(ctx context.Context, m api.Module, sig uint32, msg uint64, key uint32) uint32 {
+	rtCtx := ctx.Value(runtimeContextKey).(*runtime.Context)
+	if rtCtx == nil {
+		panic("nil runtime context")
+	}
+
+	memory := m.Memory()
+	sigVerifier := rtCtx.SigVerifier
+
+	// signature := memory[sig : sig+64]
```
jimjbrettj, line 291: nit: remove

lib/runtime/wazero/imports.go
```diff
@@ -0,0 +1,2247 @@
...
+	)
+)
+
+const (
+	validateSignatureFail = "failed to validate signature"
+)
+
+// toPointerSize converts an uint32 pointer and uint32 size
+// to an int64 pointer size.
+func newPointerSize(ptr, size uint32) (pointerSize uint64) {
```
jimjbrettj, line 42: This is not properly casting to int64. If this is still the needed functionality should be 
`return int64(ptr) | (int64(size) << 32)`

lib/runtime/wazero/imports.go
```diff
@@ -0,0 +1,2247 @@
...
+	return uint64(ptr) | (uint64(size) << 32)
+}
+
+// splitPointerSize converts a 64bit pointer size to an
+// uint32 pointer and a uint32 size.
+func splitPointerSize(pointerSize uint64) (ptr, size uint32) {
```
jimjbrettj, line 48: previously this took in an int64 and cast to a pointer and size represented as uint32's, why did this functionality change?

Did the usage just change with using wazero? If this is intended behavior we should update the comments for these two functions

lib/runtime/wazero/imports.go
```diff
@@ -0,0 +1,2247 @@
...
+	}
+
+	logger.Debugf(
+		"recovered public key of length %d: 0x%x",
+		len(pub), pub)
+
+	var fixedSize [64]byte
+	copy(fixedSize[:], pub[1:])
+
+	err = res.Set(scale.OK, fixedSize)
```
jimjbrettj, line 369: Why is this fixed sized/do we need to account for case where data is empty (err result case)?

lib/runtime/wazero/imports.go
```diff
@@ -0,0 +1,2247 @@
...
+	if !ok {
+		panic("write overflow")
+	}
+
+	size := uint32(len(value[offset:]))
+	sizeBuf := make([]byte, 4)
+	binary.LittleEndian.PutUint32(sizeBuf, size)
+
+	// this is expected to be Option(Vec<u8>)
+	ret, err := write(m, rtCtx.Allocator, scale.MustMarshal(&sizeBuf))
```
jimjbrettj, line 1051: Note: previously we did not panic in this (and probably others in this file) case

lib/runtime/wazero/instance.go
```diff
@@ -0,0 +1,828 @@
...
+}
+
+// NewInstance instantiates a runtime from raw wasm bytecode
+func NewInstance(code []byte, cfg Config) (instance *Instance, err error) {
+	logger.Patch(log.SetLevel(cfg.LogLvl), log.SetCallerFunc(true))
+
+	ctx := context.Background()
+	rt := wazero.NewRuntime(ctx)
+
+	_, err = rt.NewHostModuleBuilder("env").
```
jimjbrettj, line 81: Maybe a separate function or file for these?

lib/runtime/wazero/instance.go
```diff
@@ -0,0 +1,828 @@
...
+	if !ok {
+		panic("write overflow")
+	}
+	return result, nil
+}
+
+// Version returns the instance version.
+// This is cheap to call since the instance version is cached.
+// Note the instance version is set at creation and on code update.
+func (in *Instance) Version() (runtime.Version, error) {
```
jimjbrettj, line 447: Why put exported function in this file?

jimjbrettj (commented):
looks nice to me, using wazero seems like the much better approach. Just a few comments and questions

lib/runtime/wazero/imports.go
timwu20: I'm representing a `pointerSize` as uint64.  Whether or not it actually will use the entire bounds of a `uint64` is another question, but technically by using `uint64` we have more possible values for the combined `pointerSize`.

lib/runtime/wazero/imports.go
timwu20: We are using `api.Memory` interface provided by `wazero`.  This isn't needed anymore since it's guarding against accessing the underlying memory which was pretty unsafe to begin with.

lib/runtime/wazero/imports.go
timwu20: I'm using `scale.Result` and on line 345, the  error case is empty.  I'm using a fixed size here, since that's what's expected from the runtime.  AKA, it shouldn't be length encoded.

lib/runtime/wazero/imports.go
timwu20: we shouldn't have any buffer overflow errors.  This would be a catastrophic error in this case, so I believe we should panic in this case.

lib/runtime/wazero/instance.go
timwu20: I'm going to leave it in for now.  It's a little long, but all the unimplemented exports are at the top, which are clear to see.

lib/runtime/wazero/instance.go
timwu20: These are the receiver functions that implement the interface.  Not sure why they were in an `exports.go` file in the first place for the wasmer implementation.

.github/workflows/unit-tests.yml
```diff
@@ -41,19 +41,19 @@ jobs:
 
       - uses: actions/checkout@v3
 
-      - name: Go build cache
-        uses: actions/cache@v3
-        with:
-          path: ${{ steps.go-cache-paths.outputs.go-build }}
-          key: ${{ runner.os }}-go-build-${{ hashFiles('**/go.sum') }}
-          restore-keys: ${{ runner.os }}-go-build
+      # - name: Go build cache
```
jimjbrettj, line 44: was this meant to be commented out?

lib/runtime/wazero/imports.go
jimjbrettj: ah okay its the comment that wasnt updated

lib/runtime/wazero/imports.go
jimjbrettj: yeah I think some of the new code was mixed with old comments which threw me

lib/runtime/wazero/imports.go
jimjbrettj: makes sense to me

jimjbrettj (approved):

.github/workflows/unit-tests.yml
timwu20: nope.. good catch

Issue #3275, opened 2023-05-23
feat(dot/parachain): Implement PoV Distributor


kishansagathiya:
## Issue summary

- Learn about erasure coding. 
Erasure coding is a data storage technique to protect against data loss.
We are going to break PoV bytes into fragments and save it accross peers. In future, when we need PoV, we will request peers for these fragments and build the full PoV from it. The benefit of using erasure coding is that even if we lose a few fragments, we will be able to reconstruct the full PoV based on rest of the fragments.
- See how erasure coding is implemented in polkadot to distribute PoV (Proof of Validity)

Considering there much of parachain support is not implemented, let's limit this to only two task,
- Given a PoV, create fragments that can be sent to peers
- Given fragments, create full PoV

Further task would be to send these fragments to peers, but that we can do as part of a different issue, once this is done.

## Other information and links

- Details of this is documented in availability subsystem in node implementors guide https://paritytech.github.io/polkadot/book/node/availability/index.html

haikoschol:
This issue is obsolete. Creation of erasure chunks happens in the [availability store](https://github.com/ChainSafe/gossamer/blob/7cdb52b2a74cf70c46aa3d37c46e3a029ef317b7/dot/parachain/availability-store/availability_store.go#L323) and recovery of the PoV from chunks is part of [availability recovery](https://github.com/ChainSafe/gossamer/issues/3604).

PR #3281, opened 2023-05-25
feat(erasure_coding): introduce erasure coding for PoV Distributor

edwardmack:
## Changes
This PR introduces functions to implement erasure coding for use by PoV Distributor for sharding PoV bytes.
- `ObtainChunks` splits the provided data into `validatorsQty` number of chunks plus additional parity chunks for reconsturction
- `Reconstruct` reconstructs decodable data from a set of chunks.

This uses [Reed-Solomon Erasure Coding go library](https://github.com/klauspost/reedsolomon) created by Backblaze, see [blog post](https://www.backblaze.com/blog/reed-solomon/) for details.

Currently these function work with []byte data, future functions will construct chunks from PoV structs.

## Tests

```sh
go test github.com/ChainSafe/gossamer/lib/erasure-coding -v
```

## Issues
Regarding issue #3275 

## Primary Reviewer

@kishansagathiya 
@timwu20


EclesioMeloJunior (approved):
LGTM

lib/erasure/erasure.go
```diff
@@ -0,0 +1,64 @@
...
+		return err
+	}
+	err = enc.Reconstruct(chunks)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func recoveryThreshold(validators int) (int, error) {
```
kishansagathiya, line 56: put a comment to explain recoveryThreshold. Something like

> recoveryThreshold gives the max number of shards/chunks that we can afford to lose and still construct the full initial data. Total number of chunks will be validatorQty+recoveryThreshold.

lib/erasure/erasure.go
```diff
@@ -0,0 +1,64 @@
...
+	if validators <= 1 {
+		return 0, ErrNotEnoughValidators
+	}
+
+	needed := (validators - 1) / 3
```
kishansagathiya, line 61: Reading through the blog that you shared, my understanding is that we have flexibility in picking recoveryThreshold (basically number of parity shards).

How did you land on `(validators - 1) / 3` number of parity shards? Is this what polkadot is doing?

lib/erasure/erasure_test.go
```diff
@@ -0,0 +1,127 @@
...
+	}
+	for name, tt := range tests {
+		t.Run(name, func(t *testing.T) {
+			got, err := ObtainChunks(tt.args.validatorsQty, tt.args.data)
+			if tt.expectedError != nil {
+				assert.EqualError(t, err, tt.expectedError.Error())
+			} else {
+				assert.NoError(t, err)
+			}
+			assert.Equal(t, tt.expectedValue, got)
```
kishansagathiya, line 69: before this check we can also test that length of got equals validatorQty+recoveryThreshold.

lib/erasure/erasure.go
```diff
@@ -0,0 +1,64 @@
...
+	err = enc.Encode(shards)
+	if err != nil {
+		return nil, err
+	}
+
+	return shards, nil
+}
+
+// Reconstruct decodable data from a set of chunks
+func Reconstruct(validatorsQty int, chunks [][]byte) error {
```
kishansagathiya, line 39: As part of Reconstruct we should get the original `data` back. It looks like we might be able to do that using `enc.Join`.
In the end return `[]byte, error`

lib/erasure/erasure.go
edwardmack: Yes, this was based on a function on polkadot: https://github.com/paritytech/polkadot/blob/478c6596f2f9137f2d66709665d96d01d5c189e1/erasure-coding/src/lib.rs#L86

lib/erasure/erasure_test.go
edwardmack: added.

kishansagathiya:
Now that there is a feature branch for parachains `feat/parachain`, we should target this PR to that branch.

lib/erasure/erasure.go
edwardmack: added return to get original data back.

dot/network/utils.go
```diff
@@ -186,17 +183,12 @@ func readLEB128ToUint64(r io.Reader, buf []byte) (uint64, int, error) {
 }
 
 // readStream reads from the stream into the given buffer, returning the number of bytes read
-func readStream(stream libp2pnetwork.Stream, bufPointer *[]byte, maxSize uint64) (int, error) {
+func readStream(stream libp2pnetwork.Stream, bufPointer *[]byte, maxSize uint64) (tot int, err error) {
```
kishansagathiya, line 186: opinion/preference: better to not use named returns to avoid returning unexpectedly by mistake

kishansagathiya (approved):
Epic!!

dot/network/utils.go
edwardmack: That makes sense, these changes were from @EclesioMeloJunior (I'm not sure why they're showing as changes on this PR, I think I may have done something wrong when rebasing this branch onto feat/parachain.)

PR #3283, opened 2023-05-25
client(consensus/grandpa): Implement ChangeTree and AuthoritySet logic

jimjbrettj:
## Changes

- Implement the ChangeTree and ForkTree Interface
- Implement almost all authority logic
- Implement all authority tests

Remaining tasks for authority set (to be done in separate PRs):
- [ ] add logs and telemetry
- [ ] implement shared authority set
- [ ] Move current Authority Type into client/consensus/grandpa package

## Tests

```sh
go test ./chain/consensus/grandpa
```

## Primary Reviewer

@timwu20


client/consensus/grandpa/changeTree.go
```diff
@@ -0,0 +1,296 @@
...
+	return ct.roots
+}
+
+// GetPreOrder does a preorder traversal of the ChangeTree to get all pending changes
+func (ct *ChangeTree) GetPreOrder() []PendingChange {
+	if len(ct.roots) == 0 {
+		return nil
+	}
+
+	changes := &[]PendingChange{}
```
EclesioMeloJunior, line 120: I would suggest to create a slice of `PendingChange` with a capacity equals the number of elements in the roots which is the minimal number of items that can exists in the `changes` slice, then you provide the ref of `changes` to the `getPreOrder` function

```go
{
        ...
	changes := make([]PendingChange, 0, len(ct.roots))
	for i := 0; i < len(ct.roots); i++ {
		getPreOrder(&changes, ct.roots[i])
	}
	
	return changes
}

func getPreOrder(changes *[]PendingChange, changeNode *pendingChangeNode) {
	if changeNode == nil {
		return
	}

        *changes = append(*changes, *changeNode.change)
	for i := 0; i < len(changeNode.children); i++ {
		getPreOrder(changes, changeNode.children[i])
	}
}
```

client/consensus/grandpa/changeTree.go
```diff
@@ -0,0 +1,296 @@
...
+		}
+	}
+
+	roots := ct.Roots()
+
+	// check if the given hash is equal or a descendent of any root, if we
+	// find a valid root that passes the predicate then we must ensure that
+	// we're not finalizing past any children node.
+	var position *uint
+	for i := 0; i < len(roots); i++ {
```
EclesioMeloJunior, line 178:
```suggestion
	for i, root := range roots {
```

client/consensus/grandpa/changeTree.go
```diff
@@ -0,0 +1,296 @@
...
+	if len(ct.roots) == 0 {
+		return nil
+	}
+
+	changes := &[]PendingChange{}
+
+	// this is basically a preorder search with rotating roots
+	for i := 0; i < len(ct.roots); i++ {
+		getPreOrder(changes, ct.roots[i])
+	}
```
EclesioMeloJunior, line 125:
```suggestion
	for _, root := range ct.roots {
		getPreOrder(changes, root)
	}
```

EclesioMeloJunior (changes requested):

client/consensus/grandpa/authorities.go
```diff
@@ -178,14 +209,115 @@ func (authSet *AuthoritySet) addStandardChange(pending PendingChange, isDescende
...
+		return status, err
+	}
+
+	if finalizationResult != nil {
+		// Changed Case
+		status.changed = true
+
+		// Flush pending forced changes to re add
+		pendingForcedChanges := authSet.pendingForcedChanges
+		authSet.pendingForcedChanges = []PendingChange{}
```
kanishkatn, line 282:
```suggestion
	if finalizationResult == nil {
	        return status, nil
	}
```

kanishkatn (dismissed):

EclesioMeloJunior (changes requested):

client/consensus/grandpa/authorities.go
```diff
@@ -78,79 +68,160 @@ func (authSet *AuthoritySet) InvalidAuthorityList(authorities AuthorityList) boo
...
-	if authSet.InvalidAuthorityList(pending.nextAuthorities) {
-		return errors.New("invalid authority set, either empty or with an authority weight set to 0")
+	authSet = &AuthoritySet{
+		currentAuthorities:     initial,
+		setId:                  0,
+		pendingStandardChanges: NewChangeTree(),
+		pendingForcedChanges:   nil,
+		authoritySetChanges:    nil,
 	}
+	return
```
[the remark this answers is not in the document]
jimjbrettj: yup sounds good to me, I will update

client/consensus/grandpa/authorities.go
```diff
@@ -78,79 +68,160 @@ func (authSet *AuthoritySet) InvalidAuthorityList(authorities AuthorityList) boo
...
+	}
+	return nil, nil
+}
+
+func (authSet *AuthoritySet) addStandardChange(pending PendingChange, isDescendentOf IsDescendentOf) error {
+	hash := pending.canonHash
+	number := pending.canonHeight
+
+	logger.Debugf(
+		"inserting potential standard set change signaled at block %d (delayed by %d blocks).",
```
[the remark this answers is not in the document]
jimjbrettj: Do we want to keep logs matching with substrate? If we arent to worried I can do this change

client/consensus/grandpa/authorities.go
```diff
@@ -160,51 +231,414 @@ func (authSet *AuthoritySet) addForcedChange(pending PendingChange, isDescendent
...
+					if standardChange.EffectiveNumber() <= medianLastFinalized && isDescStandard {
+						logger.Infof(
+							"Not applying authority set change forced at block %d, due to pending standard change at block %d",
+							change.canonHeight, standardChange.EffectiveNumber())
+						return nil, errForcedAuthoritySetChangeDependencyUnsatisfied
+					}
+				}
+
+				// apply this change: make the set canonical
+				logger.Infof("%v 👴 Applying authority set change forced at block %d", initialSync, change.canonHeight)
```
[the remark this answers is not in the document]
jimjbrettj: So actually I realize that it is used as a conditional variable to avoid printing logs during bootstrap. Do you think this would be nice to have? Or better to just either print it or not print it?

client/consensus/grandpa/authorities.go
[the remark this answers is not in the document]
EclesioMeloJunior: Ok, we can keep as it is!

client/consensus/grandpa/authorities.go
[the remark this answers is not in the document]
EclesioMeloJunior: I think this could be printed for Bootstrap sync as well! And I would suggest using the `#` when printing numbers so its easy to identify it in the log lines

client/consensus/grandpa/authorities.go
```diff
@@ -160,51 +231,414 @@ func (authSet *AuthoritySet) addForcedChange(pending PendingChange, isDescendent
...
+		logger.Infof("%v 👴 Applying authority set change forced at block %d", initialSync, *finalizationResult.value)
+
+		// TODO add telemetry
+
+		// Store the set_id together with the last block_number for the set
+		authoritySetChange := authorityChange{
+			setId:       authSet.setId,
+			blockNumber: finalizedNumber,
+		}
+		authSet.authoritySetChanges = append(authSet.authoritySetChanges, authoritySetChange)
```
EclesioMeloJunior, line 430: I notice you can use `authSet.authoritySetChanges.append(authSet.setId, finalizedNumber)`

client/consensus/grandpa/authorities.go
```diff
@@ -160,51 +231,414 @@ func (authSet *AuthoritySet) addForcedChange(pending PendingChange, isDescendent
...
+
+// EffectiveNumber Returns the effective number this change will be applied at.
+func (pc *PendingChange) EffectiveNumber() uint {
+	return pc.canonHeight + pc.delay
+}
+
+// AuthoritySetChanges Tracks historical authority set changes. We store the block numbers for the last block
+// of each authority set, once they have been finalized. These blocks are guaranteed to
+// have a justification unless they were triggered by a forced change.
+type AuthoritySetChanges []authorityChange
```
EclesioMeloJunior, line 488: Do we clear/prune these auth changes stored at some point? As I see we append and sometimes we iterate over them from the very first item if we don't clear/prune the complexity will grow linearly and in long operations the node can slow down waiting for the iteration to finish, right?

client/consensus/grandpa/authorities.go
```diff
@@ -160,51 +231,414 @@ func (authSet *AuthoritySet) addForcedChange(pending PendingChange, isDescendent
...
+
+func (authSetChanges *AuthoritySetChanges) insert(blockNumber uint) {
+	var idx int
+	if authSetChanges == nil {
+		idx = 0
+	} else {
+		idx, _ = searchSetChanges(blockNumber, *authSetChanges)
+	}
+
+	set := *authSetChanges
```
EclesioMeloJunior, line 536: if `authSetChanges == nil` and you deref it will panic and there is no way to initialize it from inside this metho since the initialization will not take effect outside ref: [nil-receiver-in-golang](https://medium.com/@reetas/nil-receiver-in-golang-9d61ed8fd230) so I would suggest to `if autSetChanges == nil` then `panic("authority set changes must be initialized")`

client/consensus/grandpa/authorities.go
```diff
@@ -160,51 +231,414 @@ func (authSet *AuthoritySet) addForcedChange(pending PendingChange, isDescendent
...
+
+	var setId uint64
+	if idx == 0 {
+		setId = 0
+	} else {
+		setId = set[idx-1].setId + 1
+	}
+
+	// TODO reviewer double check this logic
+	if !(idx == len(set) || set[idx].setId != setId) {
```
EclesioMeloJunior, line 546: In the current way is quite hard to understand the condition (I get lost 🤣 ), if you want you can apply the [De Morgan's Law](https://en.wikipedia.org/wiki/De_Morgan%27s_laws) and you will have:

```go
	if idx != len(set) && set[idx].setId == setId {
```

For example, if we have a set of 3 items and the idx is `4` it will panic since it will face an out-of-bounds error, I would suggest changing the first condition to be a guard `if idx < len(set) && set[idx].setId == setId` and change the idx to `uint` type then you don't have any problems with negative numbers.

So, it is not possible to have more than one change per set id, right? If so instead of panicking we can have an error `ErrChangeForSetIDAlreadyExists` and return it

EclesioMeloJunior (changes requested):

client/consensus/grandpa/authorities.go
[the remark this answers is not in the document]
jimjbrettj: updated!

client/consensus/grandpa/authorities.go
jimjbrettj: Good call!

client/consensus/grandpa/authorities.go
jimjbrettj: This is a good point. I think potentially beyond scope of this PR, since I was referencing substrate (unless I missed this functionality), but I think this is a good point and something we should think about. Will tag @timwu20 here to get his thoughts on this when he reviews, but my take is that we should do this but outside of this PR (and probably once we build out more critical grandpa components)

client/consensus/grandpa/authorities.go
jimjbrettj: removed the set index to 0 and replaced it with a panic, good call

client/consensus/grandpa/authorities.go
jimjbrettj: The rust code is `assert!(idx == self.0.len() || self.0[idx].0 != set_id);` which makes me think we should panic right? I think this is probably gets enforced that we cannot have more than one change per set ID right? My gut is to panic since they do, but also if you don't think its best I am more than happy to change it.

Your suggested change to use a `<` sign only applies if we error instead of panic right?

edwardmack (approved):
Looks great!

client/consensus/grandpa/authorities.go
```diff
@@ -78,79 +68,162 @@ func (authSet *AuthoritySet) InvalidAuthorityList(authorities AuthorityList) boo
...
+
+	_, err := authSet.pendingStandardChanges.Import(hash, number, pending, isDescendentOf)
+	if err != nil {
+		return err
+	}
+
+	logger.Debugf(
+		"There are now %d alternatives for the next pending standard change (roots), "+
+			"and a total of %d pending standard changes (across all forks)",
+		len(authSet.pendingStandardChanges.Roots()), authSet.pendingStandardChanges.GetPreOrder(),
```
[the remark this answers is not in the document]
jimjbrettj: good catch!

client/consensus/grandpa/authorities.go
```diff
@@ -184,27 +273,360 @@ func (authSet *AuthoritySet) addStandardChange(pending PendingChange, isDescende
...
+
+// Three states that can be returned: Latest, Set (tuple), Unknown
+// Latest => bool
+// Set => &AuthorityChange
+// Unknown => nil
+// TODO for reviewers, this can be a VDT but I'm not sure its needed
+func (authSetChanges *AuthoritySetChanges) getSetId(blockNumber uint) (latest bool, set *authorityChange, err error) {
+	if authSetChanges == nil {
+		return false, nil, fmt.Errorf("getSetId: authSetChanges is nil")
+	}
```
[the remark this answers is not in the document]
jimjbrettj: Since the next line is authSet := *asc, wouldn't we want a nil check to prevent a panic?

client/consensus/grandpa/lib.go
```diff
@@ -4,7 +4,12 @@
...
 import (
+	"github.com/ChainSafe/gossamer/dot/types"
 	"github.com/ChainSafe/gossamer/internal/log"
 )
 
 var logger = log.NewFromGlobal(log.AddContext("consensus", "grandpa"))
+
+// AuthorityList A list of Grandpa authorities with associated weights.
+// TODO migrate this type and associated functions to this package
+type AuthorityList []types.Authority
```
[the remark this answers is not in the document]
jimjbrettj: Removed, only usage now is the type common.Address which, pending permission, I will remove

client/consensus/grandpa/utils.go
```diff
@@ -3,19 +3,32 @@
...
+			return i
+		}
+	}
+
+	return len(changes)
+}
+
+// returns an index representing either the found element or the index to insert the given element, and a bool
+// indicating if the given element was found
+func searchSetChanges(number uint, changes AuthoritySetChanges) (int, bool) {
```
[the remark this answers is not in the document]
jimjbrettj: just to clarify, using slices.BinarySearchFunc will return the index the item belongs regardless of if it is found, then a bool indicating if it is found or not?

client/consensus/grandpa/authorities.go
```diff
@@ -184,27 +273,360 @@ func (authSet *AuthoritySet) addStandardChange(pending PendingChange, isDescende
...
+
+	iterChanges := authSet[idx:]
+	return &iterChanges, nil
+}
+
+////// Function Types //////
+
+// Predicate is a type to represent the function signature of a Predicate function
+type Predicate[T any] func(T) bool
```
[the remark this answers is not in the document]
jimjbrettj: updated, should I do this for the isDescendantOf func as well or is it used enough that its worth having a custom type for?

client/consensus/grandpa/authorities.go
```diff
@@ -0,0 +1,611 @@
...
+func (authSet *AuthoritySet[H, N]) applyForcedChanges(bestHash H,
+	bestNumber N,
+	isDescendentOf IsDescendentOf[H],
+	telemetry *telemetry.Client) (newSet *appliedChanges[H, N], err error) {
+
+	for _, change := range authSet.pendingForcedChanges {
+		effectiveNumber := change.EffectiveNumber()
+		if effectiveNumber > bestNumber {
+			continue
+		} else if effectiveNumber == bestNumber {
```
EclesioMeloJunior, line 316: You don't need the `else if effectiveNumber == bestNumber` because you might need to check effective numbers lower then the best number

```suggestion
		if effectiveNumber > bestNumber {
			continue
		}
		
		... the logic here
```

client/consensus/grandpa/authorities.go
```diff
@@ -0,0 +1,611 @@
...
+		if idx == 0 && authChange.setId != 0 {
+			return nil
+		}
+	}
+
+	iterChanges := authSet[idx:]
+	return &iterChanges
+}
+
+////// Additional types //////
```
EclesioMeloJunior, line 584: I would suggest to move those types definitions to the top of the file
```suggestion
```

EclesioMeloJunior (changes requested):

client/consensus/grandpa/authorities.go
jimjbrettj: Yes that makes sense to me. I had initially not done this to try and maintain substrates file structure as close as I could, but I think its cleaner to have on top

client/consensus/grandpa/authorities.go
jimjbrettj: It seems to me that we are actually not supposed to check when the effective number is lower than the best number. Substrate has test cases for this case and when I make this change I fail them

client/consensus/grandpa/authorities.go
```diff
@@ -184,27 +273,360 @@ func (authSet *AuthoritySet) addStandardChange(pending PendingChange, isDescende
...
+		blockNumber: blockNumber,
+	})
+}
+
+// Three states that can be returned: Latest, Set (tuple), Unknown
+// Latest => bool
+// Set => &AuthorityChange
+// Unknown => nil
+// TODO for reviewers, this can be a VDT but I'm not sure its needed
+func (authSetChanges *AuthoritySetChanges) getSetId(blockNumber uint) (latest bool, set *authorityChange, err error) {
```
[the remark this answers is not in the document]
jimjbrettj: It can go either way. Can use this if we don't want the overhead of a vdt type, but also would work well with vdt too. Can update if you think it is best

lib/crypto/keypair.go
```diff
@@ -31,7 +29,7 @@ type PublicKey interface {
 	Verify(msg, sig []byte) (bool, error)
 	Encode() []byte
 	Decode([]byte) error
-	Address() common.Address
+	Address() string
```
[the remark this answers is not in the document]
jimjbrettj: I was trying to reduce our usage of types outside of the package

client/consensus/grandpa/delayKind.go
```diff
@@ -0,0 +1,33 @@
...
+	return delayKind
+}
+
+// Finalized Depth in finalized chain.
+type Finalized struct{}
+
+// Best Depth in best chain. The median last finalized block is calculated at the time the
+// hashNumber was signaled.
+type Best struct {
+	medianLastFinalized uint
```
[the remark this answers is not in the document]
jimjbrettj: Im actually unsure how to do this is this case. Since DelayedKinds is itself used as a generic type, we have something like

```
func setDelayKind[T DelayedKinds](delayKind *DelayKind, val T) {
	delayKind.value = val
}
```
and it doesn't seem to be valid to add generic constraint to the generic constraint, something like 


```
func setDelayKind[T DelayedKinds[N or N any, both don't work](delayKind *DelayKind, val T) {
	delayKind.value = val
}
```

Do you know how I can do this? Perhaps can also just use a VDT instead

client/consensus/grandpa/authorities.go
```diff
@@ -0,0 +1,655 @@
...
+		}
+	}
+
+	key := key[N]{
+		pending.EffectiveNumber(),
+		pending.canonHeight,
+	}
+
+	// Search by effective key
+	idx := searchKey(key, authSet.pendingForcedChanges)
```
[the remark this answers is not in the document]
jimjbrettj: updated, let me know if the solution looks okay to you

client/consensus/grandpa/authorities.go
```diff
@@ -0,0 +1,655 @@
...
+		effectiveNumber := change.EffectiveNumber()
+		if effectiveNumber > bestNumber {
+			continue
+		} else if effectiveNumber == bestNumber {
+			// check if the given best block is in the same branch as
+			// the block that signaled the hashNumber.
+			isDesc, err := isDescendentOf(change.canonHash, bestHash)
+			if err != nil {
+				return nil, err
+			}
```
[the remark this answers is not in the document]
jimjbrettj: updated, let me know if the change looks okay

client/consensus/grandpa/authorities.go
```diff
@@ -0,0 +1,655 @@
...
+	})
+	if err != nil {
+		return status, err
+	}
+
+	if finalizationResult == nil {
+		return status, nil
+	}
+
+	// Changed Case
```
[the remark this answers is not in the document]
jimjbrettj: So my logic here was that I didn't think we needed the overHead of a VDT. Since all that the enum contains is changed and unchanged, where unchanged contains no data and prompts no action, I made it return a pointer where nil case represents unchanged and not nil represents change. That is why I do a check `if finalizationResult == nil {
		return status, nil
	}`
to break early in the unchanged case.

This was my reasoning, but if you think better still do just use a vdt I am more than happy to implement it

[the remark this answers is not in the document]
jimjbrettj: I also came up with a comperable alternative for a VDT here https://github.com/ChainSafe/gossamer/blob/93867f96ba82b3eb71a48c0d34a86c2f06dd390d/client/consensus/grandpa/authorities.go#L574, so I would appreciate you checking it out and letting me know if I should make this a vdt as well

client/consensus/grandpa/authorities.go
```diff
@@ -0,0 +1,814 @@
...
+// DelayKind struct to represent DelayedKinds
+type DelayKind struct {
+	value interface{}
+}
+
+func setDelayKind[N constraints.Unsigned, T DelayedKinds[N]](delayKind *DelayKind, val T) {
+	delayKind.value = val
+}
+
+func newDelayKind[N constraints.Unsigned, T DelayedKinds[N]](val T) DelayKind {
```
timwu20, line 43: if it was public, the set and new constructor should be public as well.

client/consensus/grandpa/authorities.go
```diff
@@ -0,0 +1,814 @@
...
+	// hashNumber) must be applied beforehand.
+	pendingForcedChanges []PendingChange[H, N]
+	// Track at which blocks the set id changed. This is useful when we need to prove finality for
+	// a given block since we can figure out what set the block belongs to and when the set
+	// started/ended.
+	authoritySetChanges AuthoritySetChanges[N]
+}
+
+// InvalidAuthorityList authority sets must be non-empty and all weights must be greater than 0
+func InvalidAuthorityList(authorities AuthorityList) bool {
```
timwu20, line 223: don't think we need to make this public.  
```suggestion
func invalidAuthorityList(authorities AuthorityList) bool {
```

client/consensus/grandpa/authorities.go
```diff
@@ -0,0 +1,814 @@
...
+	return changes
+}
+
+// CurrentLimit Get the earliest limit-block number, if any. If there are pending changes across
+// different forks, this method will return the earliest effective number (across the
+// different branches) that is higher or equal to the given min number.
+//
+// Only standard changes are taken into account for the current
+// limit, since any existing forced hashNumber should preclude the voter from voting.
+func (authSet *AuthoritySet[H, N]) CurrentLimit(min N) (limit *N) {
```
timwu20, line 458: This is `pub(crate)` in [substrate](https://github.com/paritytech/substrate/blob/7c8b89b3a6b603902709fc33fdb10a864b8b4204/client/consensus/grandpa/src/authorities.rs#L423).
```suggestion
func (authSet *AuthoritySet[H, N]) currentLimit(min N) (limit *N) {
```

client/consensus/grandpa/authorities.go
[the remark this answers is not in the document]
timwu20: I see that your version of `FinalizationResult` contains a pointer to a pending change.  And that's how you determine the `Changed` vs `Unchanged`.  If you're going to use the same name, I think it should be mimic the substrate enum. Or consider revising the forktree interfaces to return `*PendingChange` so the difference is clearer.

client/consensus/grandpa/interfaces.go
```diff
@@ -0,0 +1,38 @@
...
+	PendingChanges() []PendingChange[H, N]
+}
+
+// PublicKey interface
+type PublicKey interface {
+	Verify(msg, sig []byte) (bool, error)
+	Encode() []byte
+	Decode([]byte) error
+	Hex() string
+}
```
timwu20, line 35: why do we need this?

client/consensus/grandpa/lib.go
```diff
@@ -0,0 +1,25 @@
...
+	"github.com/ChainSafe/gossamer/internal/log"
+)
+
+var logger = log.NewFromGlobal(log.AddContext("consensus", "grandpa"))
+
+// Hash represents a grandpa hash
+type Hash [32]byte
+
+// Address ...
+type Address string
```
timwu20, line 16: unneeded.

client/consensus/grandpa/lib.go
```diff
@@ -0,0 +1,25 @@
...
+package grandpa
+
+import (
+	"github.com/ChainSafe/gossamer/internal/log"
+)
+
+var logger = log.NewFromGlobal(log.AddContext("consensus", "grandpa"))
+
+// Hash represents a grandpa hash
+type Hash [32]byte
```
timwu20, line 13: move this to test code.

client/consensus/grandpa/lib.go
```diff
@@ -0,0 +1,25 @@
...
+type Address string
+
+// Authority represents a grandpa authority
+type Authority struct {
+	Key    PublicKey
+	Weight uint64
+}
+
+// AuthorityList A list of Grandpa authorities with associated weights.
+type AuthorityList []Authority
```
timwu20, line 25: debatable if we need this.  It can be represented as `[]Authority` since there are no receiver functions.

client/consensus/grandpa/utils.go
```diff
@@ -0,0 +1,29 @@
+// Copyright 2023 ChainSafe Systems (ON)
+// SPDX-License-Identifier: LGPL-3.0-only
+
+package grandpa
+
+import (
+	"fmt"
+)
+
+func bytesToHash(b []byte) Hash {
```
timwu20, line 10: I don't think you need this.  You only use it for tests, to generate something that satisfies `comparable`.  You can just use the strings in the tests.

timwu20 (changes requested):

client/consensus/grandpa/interfaces.go
jimjbrettj: at first I was using the PublicKey interface, but in fact we do not need it. I have changed the type of public key and removed the interface in #3383

client/consensus/grandpa/utils.go
jimjbrettj: Really good call on this. Was able to remove whole file and clean up the code

client/consensus/grandpa/authorities.go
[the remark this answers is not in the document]
jimjbrettj: Actually in this case it is the outer struct that is determining if it is `Changed` vs `Unchanged` where the inner type `*PendingChange` is used to determine if there is data associated with the change. See [here for substrate code](https://github.com/paritytech/substrate/blob/28b1f79abedbad2ec97b49277b36e1f797ca9e5d/utils/fork-tree/src/lib.rs#L708).

That being said, do you think changing name of outer struct to be more clear is sufficient, or better to create VDT to mimic the enum?

client/consensus/grandpa/authorities.go
```diff
@@ -0,0 +1,814 @@
...
+// function `is_descendent_of` should return `true` if the second hash
+// (target) is a descendent of the first hash (base).
+func (sas *SharedAuthoritySet[H, N]) EnactsStandardChange(finalizedHash H, finalizedNumber N, isDescendentOf IsDescendentOf[H]) (*bool, error) {
+	sas.mtx.Lock()
+	defer sas.mtx.Unlock()
+	return sas.inner.EnactsStandardChange(finalizedHash, finalizedNumber, isDescendentOf)
+}
+
+// Status of the set after changes were applied.
+type Status[H comparable, N constraints.Unsigned] struct {
```
timwu20, line 189: Don't think this needs to be public.  It's defined as `pub(crate)` in the rust code.

client/consensus/grandpa/authorities.go
```diff
@@ -0,0 +1,814 @@
...
+	// block where that set changed.
+	newSetBlock *hashNumber[H, N]
+}
+
+// AuthoritySet A set of authorities.
+type AuthoritySet[H comparable, N constraints.Unsigned] struct {
+	// The current active authorities.
+	currentAuthorities []Authority
+	// The current set id.
+	setId uint64
```
timwu20, line 202:
```suggestion
	setID uint64
```

client/consensus/grandpa/authorities.go
```diff
@@ -0,0 +1,814 @@
...
+		}
+	}
+	return false
+}
+
+// NewGenesisAuthoritySet Get a genesis set with given authorities.
+func NewGenesisAuthoritySet[H comparable, N constraints.Unsigned](initial []Authority) (authSet *AuthoritySet[H, N]) {
+	if invalidAuthorityList(initial) {
+		return nil
+	}
```
timwu20, line 240: Seems odd that we return a pointer, but that it could be `nil` if it's invalid.  Maybe we should make `invalidAuthoritySet` return an error.

client/consensus/grandpa/authorities.go
```diff
@@ -0,0 +1,814 @@
...
+}
+
+// current Get the current set id and a reference to the current authority set.
+func (authSet *AuthoritySet[H, N]) current() (uint64, *[]Authority) {
+	return authSet.setId, &authSet.currentAuthorities
+}
+
+func (authSet *AuthoritySet[H, N]) revert() {
+	panic("AuthoritySet.revert not implemented yet")
```
timwu20, line 277: why wasn't this implemented?

client/consensus/grandpa/authorities.go
```diff
@@ -0,0 +1,814 @@
...
+type hashNumber[H, N any] struct {
+	hash   H
+	number N
+}
+
+// key is used to represent a tuple ordered first by effective number and then by signal-block number
+type key[N any] struct {
+	effectiveNumber   N
+	signalBlockNumber N
+}
```
timwu20, line 77: Since this is only used in `AuthoritySet.addForcedChange` move this inside that functiuon.

client/consensus/grandpa/authorities.go
```diff
@@ -0,0 +1,814 @@
...
+func (authSet *AuthoritySet[H, N]) applyForcedChanges(bestHash H,
+	bestNumber N,
+	isDescendentOf IsDescendentOf[H],
+	_ Telemetry) (newSet *appliedChanges[H, N], err error) {
+
+	for _, change := range authSet.pendingForcedChanges {
+		effectiveNumber := change.EffectiveNumber()
+		if effectiveNumber > bestNumber {
+			continue
+		} else if effectiveNumber == bestNumber {
```
timwu20, line 498: From the rust code:
```
		for change in self
			.pending_forced_changes
			.iter()
			.take_while(|c| c.effective_number() <= best_number) // to prevent iterating too far
			.filter(|c| c.effective_number() == best_number)
		{
```
I think it makes sense to just simplify this.
```suggestion
		effectiveNumber := change.EffectiveNumber()
		if effectiveNumber != bestNumber {
		         continue
		}
		... rest of logic for valid pending forced change
```

client/consensus/grandpa/authorities.go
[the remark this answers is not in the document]
timwu20: I think if we need to hold some generic type of data for the `FinalizationResult::Changed` case, I think it makes more sense to translate the enum 1:1.

client/consensus/grandpa/authorities.go
```diff
@@ -0,0 +1,814 @@
...
+		blockNumber: blockNumber,
+	})
+}
+
+// Three states that can be returned: Latest, Set (tuple), Unknown
+// Latest => bool
+// Set => &AuthorityChange
+// Unknown => nil
+// TODO for reviewers, this can be a VDT but I'm not sure its needed
+func (asc *AuthoritySetChanges[N]) getSetID(blockNumber N) (latest bool, set *setIDNumber[N], err error) {
```
timwu20, line 683: I think we should recreate the enum in this case, even though the `AuthoritySetChangeId` type in the rust code isn't actually exported outside of the crate, I think it makes it easier to cross reference.  I also think dealing with more than 1 non-error return value is kinda odd.

client/consensus/grandpa/interfaces.go
timwu20: can we just remove this here, makes the PR smaller.

client/consensus/grandpa/interfaces.go
```diff
@@ -0,0 +1,38 @@
...
+
+type Telemetry interface {
+}
```
timwu20, line 38: Add a TODO with accompanying issue to address this later.

client/consensus/grandpa/authorities.go
jimjbrettj: I had waited initially because it was never used in the tests, but I can implement and write my own tests

client/consensus/grandpa/authorities.go
```diff
@@ -0,0 +1,881 @@
...
+			switch {
+			case change.EffectiveNumber() == k.effectiveNumber && change.canonHeight == k.signalBlockNumber:
+				return 0
+			case change.EffectiveNumber() > k.effectiveNumber && change.canonHeight > k.signalBlockNumber:
+				return 1
+			case change.EffectiveNumber() < k.effectiveNumber && change.canonHeight < k.signalBlockNumber:
+				return -1
+			default:
+				panic("huh?")
+			}
```
EclesioMeloJunior, line 361: I don't think having the panic in this binary search makes sense because you can cover all the comparison cases, here is my suggestion:

Implement the methods `LessThan` and `GreaterThan` in the `PendingChange` struct

```go
func (pc *PendingChange[H, N]) GreaterThan(effectiveBlockNumber, signalBlockNumber N) bool {
	effectiveNumberGreaterThan := pc.EffectiveNumber() > effectiveBlockNumber
	cannonHeighGreaterThan := pc.EffectiveNumber() == effectiveBlockNumber &&
		pc.canonHeight > signalBlockNumber

	return effectiveNumberGreaterThan || cannonHeighGreaterThan
}

func (pc *PendingChange[H, N]) LessThan(effectiveBlockNumber, signalBlockNumber N) bool {
	effectiveNumberLessThan := pc.EffectiveNumber() < effectiveBlockNumber
	cannonHeighLessThan := pc.EffectiveNumber() == effectiveBlockNumber &&
		pc.canonHeight < signalBlockNumber

	return effectiveNumberLessThan || cannonHeighLessThan
}
```

And in the `slices.BinarySearchFunc`:

```go
idx, _ := slices.BinarySearchFunc(
	authSet.pendingForcedChanges,
	k,
	func(change PendingChange[H, N], k key[N]) int {
		switch {
		case change.LessThan(k.effectiveNumber, k.signalBlockNumber):
			return -1
		case change.GreaterThan(k.effectiveNumber, k.signalBlockNumber):
			return 1
		default:
			return 0
		}
	},
)
```

EclesioMeloJunior (changes requested):

client/consensus/grandpa/authorities.go
```diff
@@ -0,0 +1,881 @@
...
+	for _, authority := range authorities {
+		if authority.Weight == 0 {
+			return true
+		}
+	}
+	return false
+}
+
+// NewGenesisAuthoritySet Get a genesis set with given authorities.
+func NewGenesisAuthoritySet[H comparable, N constraints.Unsigned](initial []Authority) (authSet *AuthoritySet[H, N], err error) {
```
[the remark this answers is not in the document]
jimjbrettj: Good call

client/consensus/grandpa/authorities.go
timwu20: is this outdated now?

client/consensus/grandpa/authorities.go
jimjbrettj: This will be addressed in a different PR

go.sum
```diff
@@ -603,7 +598,6 @@ github.com/multiformats/go-multistream v0.4.1/go.mod h1:Mz5eykRVAjJWckE2U78c6xqd
 github.com/multiformats/go-varint v0.0.1/go.mod h1:3Ls8CIEsrijN6+B7PbrXRPxHRPuXSrVKRY101jdMZYE=
 github.com/multiformats/go-varint v0.0.7 h1:sWSGR+f/eu5ABZA2ZpYKBILXTTs9JWpdEM/nEGOHFS8=
 github.com/multiformats/go-varint v0.0.7/go.mod h1:r8PUYw/fD/SjBCiKOoDlGF6QawOELpZAu9eioSos/OU=
-github.com/mwitkow/go-conntrack v0.0.0-20190716064945-2f068394615f/go.mod h1:qRWi+5nqEBWmkhHvq77mSJWrCKwh8bxhgT7d/eI7P4U=
 github.com/nanobox-io/golang-scribble v0.0.0-20190309225732-aa3e7c118975 h1:zm/Rb2OsnLWCY88Njoqgo4X6yt/lx3oBNWhepX0AOMU=
 github.com/nanobox-io/golang-scribble v0.0.0-20190309225732-aa3e7c118975/go.mod h1:4Mct/lWCFf1jzQTTAaWtOI7sXqmG+wBeiBfT4CxoaJk=
 github.com/nats-io/jwt v0.3.0/go.mod h1:fRYCDE99xlTsqUzISS1Bi75UBJ6ljOJQOAAu5VglpSg=
```
timwu20, line 603: there shouldn't be a change to this file if there's no accompanying change to go.mod

pkg/finality-grandpa/voter_set_test.go
```diff
@@ -68,7 +68,6 @@ func TestVoterSet_Equality(t *testing.T) {
 			sum.Add(sum, new(big.Int).SetUint64(uint64(iw.Weight)))
 		}
 		return sum.Cmp(new(big.Int).SetUint64(uint64(math.MaxUint64))) > 0
-
```
timwu20, line 71: remove this change.

client/consensus/grandpa/changeTree.go
```diff
@@ -0,0 +1,453 @@
...
+
+import (
+	"errors"
+	"fmt"
+	"github.com/ChainSafe/gossamer/pkg/scale"
+	"golang.org/x/exp/constraints"
+)
+
+/*
+	TODO give a summary of how this works in context of grandpa
```
timwu20, line 14: can we add this description before merging

timwu20 (changes requested):

client/consensus/grandpa/authorities.go
```diff
@@ -0,0 +1,882 @@
...
+}
+
+// generic representation of hash and number tuple
+type hashNumber[H, N any] struct {
+	hash   H
+	number N
+}
+
+// appliedChanges represents the median and new set when a forced hashNumber has occured
+type appliedChanges[H comparable, N constraints.Unsigned] struct {
```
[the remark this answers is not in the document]
jimjbrettj: This represents the change caused by a ForcedChange, which was why I called it appliedChanges. However, I can still rename

client/consensus/grandpa/changeTree.go
jimjbrettj: I have given a small description, let me know if more is desired

client/consensus/grandpa/authorities.go
```diff
@@ -0,0 +1,908 @@
...
+			switch {
+			case a.blockNumber == b:
+				return 0
+			case a.blockNumber > b:
+				return 1
+			case a.blockNumber < b:
+				return -1
+			default:
+				panic("huh?")
+			}
```
EclesioMeloJunior, line 787: looks like all possible comparisons are covered so I think you can remove the `default` case

client/consensus/grandpa/authorities.go
```diff
@@ -0,0 +1,908 @@
...
+func (asc *authoritySetChangeID) Value() (val scale.VaryingDataTypeValue, err error) {
+	vdt := scale.VaryingDataType(*asc)
+	return vdt.Value()
+}
+
+func newAuthoritySetChangeID[N constraints.Unsigned]() authoritySetChangeID {
+	vdt, err := scale.NewVaryingDataType(latest{}, set[N]{}, unknown{})
+	if err != nil {
+		panic(err)
+	}
```
EclesioMeloJunior, line 734:
```suggestion
	vdt := scale.MustNewVaryingDataType(latest{}, set[N]{}, unknown{})
```

client/consensus/grandpa/authorities.go
```diff
@@ -0,0 +1,908 @@
...
+			func(a setIDNumber[N], b N) int {
+				switch {
+				case a.blockNumber == b:
+					return 0
+				case a.blockNumber > b:
+					return 1
+				case a.blockNumber < b:
+					return -1
+				default:
+					panic("huh?")
```
EclesioMeloJunior, line 834: looks like all possible comparisons are covered so I think you can remove the `default` case

client/consensus/grandpa/authorities.go
```diff
@@ -0,0 +1,908 @@
...
+		func(a setIDNumber[N], b N) int {
+			switch {
+			case a.blockNumber == b:
+				return 0
+			case a.blockNumber > b:
+				return 1
+			case a.blockNumber < b:
+				return -1
+			default:
+				panic("huh?")
```
EclesioMeloJunior, line 887: looks like all possible comparisons are covered so I think you can remove the `default` case
```suggestion
```

EclesioMeloJunior (changes requested):

client/consensus/grandpa/authorities.go
jimjbrettj: All cases of covered but a return is still needed since it's not literally every possible case. I have kept as is and modified the panic

client/consensus/grandpa/authorities.go
jimjbrettj: same here

client/consensus/grandpa/authorities.go
jimjbrettj: and here

client/consensus/grandpa/changeTree.go
```diff
@@ -0,0 +1,468 @@
...
+				if child.change.canonHeight <= number && (child.change.canonHash == *hash || isChildDescOf) {
+					return nil, errUnfinalisedAncestor
+				}
+			}
+
+			isEqual := false
+			for _, val := range roots {
+				if val.change.canonHash == root.change.canonHash {
+					isEqual = true
+					break
```
EclesioMeloJunior, line 169:
```suggestion
					isEqual = true
					return &isEqual, nil
```

EclesioMeloJunior (approved):

client/consensus/grandpa/authorities.go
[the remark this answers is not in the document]
timwu20: please do, I'd like to be able to reproduce the `AppliedChanges` enum with the same name.  We will need it for future work in this package.

timwu20 (commented):
can you fix the introduced deepsource issues as well. I'd rather not accumulate more deepsource issues when we merge this into development. A lot of it will be just ignoring the issues in the code, since grust is so unorthodox 😉

timwu20 (approved):

Issue #3286, opened 2023-05-29
Failed to read block response stream


EclesioMeloJunior:
## Describe the bug

- Gossamer is failing to read the block response message from other peers, after executing the request (log line below) 

```
2023-05-29T16:15:02Z DEBUG    [EXECUTING] worker 12D3KooWPVPzs42GvRBShdUMtFsk4SvnByrSdWqb6aeAAHvLMSLS: block request: BlockRequestMessage RequestedData=19 StartingBlock={6169473} Direction=0 Max=128	pkg=sync
```

it fails to read the full bytes

```
2023-05-29T16:15:03Z DEBUG    [FINISHED] worker 12D3KooWPVPzs42GvRBShdUMtFsk4SvnByrSdWqb6aeAAHvLMSLS: err: read stream error: failed to read entire message: expected 6517303 bytes, received 4177920 bytes	pkg=sync
```

PR #3287, opened 2023-05-29
fix(dot/network): remove `maxReads` limitation to read stream

EclesioMeloJunior:
## Changes

- Remove the `maxReads` limitation to read stream responses
- The `readStream` function has the information on the total size of the response so we only stop reading the stream once we reach the total length instead of using a `maxReads` that was leading to a wrong read of bigger stream responses

## Tests

```sh
```

## Issues

- Closes https://github.com/ChainSafe/gossamer/issues/3286

## Primary Reviewer

@kanishkatn

### DIFF
--- a/dot/network/utils.go
+++ b/dot/network/utils.go
@@ -217,7 +217,7 @@ func readStream(stream libp2pnetwork.Stream, bufPointer *[]byte, maxSize uint64)
 	}
 
 	tot = 0
-	for i := 0; i < maxReads; i++ {
+	for {
 		n, err := stream.Read(buf[tot:])
 		if err != nil {
 			return n + tot, err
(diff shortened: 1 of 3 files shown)


dot/network/utils.go
```diff
@@ -217,7 +217,7 @@ func readStream(stream libp2pnetwork.Stream, bufPointer *[]byte, maxSize uint64)
 	}
 
 	tot = 0
-	for i := 0; i < maxReads; i++ {
+	for {
```
dimartiro, line 220: What if we do something like:

```suggestion
	for tot < int(length) {
```

Letting the `if tot != int(length)` check and we can prevent creating a new error type

dimartiro (approved):
Approved with one suggestion

dimartiro:
Can you add at unit test for this function?

kishansagathiya (approved):
giphy-2735669673

timwu20 (approved):

timwu20 (approved):

timwu20 (approved):

kanishkatn (approved):

dimartiro (approved):
Thanks for the tests 😄

PR #3290, opened 2023-05-31
chore(lib/runtime): revise `lib/runtime` interfaces

timwu20:
## Changes

- remove local interfaces in `lib/runtime/wasmer`
- revise `runtime.Memory` interface to no longer pass mutable `[]byte`, and conforms to `api.Memory` from `wazero`
- update allocator tests

## Tests

```sh
```

## Primary Reviewer

@EclesioMeloJunior


lib/runtime/wasmer/instance.go
```diff
@@ -211,11 +211,50 @@ func setupVM(code []byte) (instance wasm.Instance,
 	// wasmer 0.3.x does not support this, but wasmer 1.0.0 does (#1268)
 	heapBase := runtime.DefaultHeapBase
 
-	allocator = runtime.NewAllocator(instance.Memory, heapBase)
+	allocator = runtime.NewAllocator(&memoryShim{instance.Memory}, heapBase)
 
 	return instance, allocator, nil
 }
 
+type memoryShim struct {
```
EclesioMeloJunior, line 219: What `Shim` means?

lib/runtime/allocator_test.go
```diff
@@ -264,8 +264,25 @@ func TestAllocator(t *testing.T) {
 		const size = 1 << 16
 		testobj := make([]byte, size)
 
-		memmock.EXPECT().Data().Return(testobj).AnyTimes()
-		memmock.EXPECT().Length().DoAndReturn(func() uint32 {
+		memmock.EXPECT().WriteByte(gomock.Any(), gomock.Any()).DoAndReturn(func(offset uint32, v byte) bool {
+			testobj[offset] = v
+			return true
+		}).AnyTimes()
```
EclesioMeloJunior, line 270: Given the @axaysagathiya PR to replace the all `AnyTimes()` occurrences, I think would be nice to change here as well to the right number using `Times(n)`

lib/runtime/allocator.go
```diff
@@ -56,29 +56,29 @@ func NewAllocator(mem Memory, ptrOffset uint32) *FreeingBumpHeapAllocator {
...
 	}
 
-	if mem.Length() <= ptrOffset {
-		err := mem.Grow(((ptrOffset - mem.Length()) / PageSize) + 1)
-		if err != nil {
-			panic(err)
+	if mem.Size() <= ptrOffset {
+		_, ok := mem.Grow(((ptrOffset - mem.Size()) / PageSize) + 1)
+		if !ok {
+			panic("wtf?")
```
EclesioMeloJunior, line 62: Can we panic with a more clear message? 😸

EclesioMeloJunior (changes requested):

lib/runtime/allocator.go
timwu20: 😂 yes I will address this.

lib/runtime/wasmer/instance.go
timwu20: > a washer or thin strip of material used to align parts, make them fit, or reduce wear.
It's definitely a North American term.

lib/runtime/allocator_test.go
timwu20: given the nature of the tests and the fact that it references `allTests`, I don't think it makes much sense in this case to arbitrarily ensure that the mocked methods are called at `n` times.  This will make it pretty difficult if we ever decided to add something to `allTests`.

lib/runtime/allocator.go
jimjbrettj: nit: wrap the error

jimjbrettj (approved):
lgtm

EclesioMeloJunior (approved):

PR #3291, opened 2023-05-31
Replace usages of .AnyTimes() with .Times(n)

### LINKED ISSUE #3038: mocks: replace usages of `.AnyTimes()` with `.Times(n)`
`.AnyTimes()` is too permissive for a good/deep test assertions, and should be replaced with `.Times(n)`.
Note `Times(1)` does not need to be specified.

Test helping functions returning a mock that can be called any number of times should contain the `AnyTimes` suffix in their name to avoid any confusion, and we should ideally reduce their usage/existence.

axaysagathiya:
closes #3038

## Changes
Changed AnyTime() with Times() from multiple tests. 

```sh
```

## Issues
Fixes #3038 

## Primary Reviewer

@timwu20


EclesioMeloJunior (changes requested):
Nice changes!! 🚀

dot/rpc/subscription/listeners_integration_test.go
```diff
@@ -189,8 +189,8 @@ func TestExtrinsicSubmitListener_Listen(t *testing.T) {
 	wsconn.BlockAPI = BlockAPI
 
 	TxStateAPI := NewMockTransactionStateAPI(ctrl)
-	TxStateAPI.EXPECT().FreeStatusNotifierChannel(gomock.Any()).AnyTimes()
-	TxStateAPI.EXPECT().GetStatusNotifierChannel(gomock.Any()).Return(make(chan transaction.Status)).AnyTimes()
+	TxStateAPI.EXPECT().FreeStatusNotifierChannel(gomock.Any()).Times(1)
```
EclesioMeloJunior, line 192: The calls the has `Times(1)` you can leave it without the `Times(1)`, like:
```suggestion
	TxStateAPI.EXPECT().FreeStatusNotifierChannel(gomock.Any())
```

EclesioMeloJunior (changes requested):

dimartiro (approved):

kishansagathiya (approved):

PR #3303, opened 2023-06-08
Use a single flag for log level in the CLI

### LINKED ISSUE #3210: Use a single flag for log-level in the CLI
## Issue summary

Currently, we have several flags to set the log-level of the modules.
- `--log` to set the global log-level
- Module specific flags (`--lcore`, `--lrpc` etc) to set the module log-level

We could use a single flag to accept all the log-levels. For example:
`--log global:info, core:debug, rpc:crit`

This was noticed during the CLI refactor (#3173) and was decided to pick it up at a later stage

Discussion around the CLI refactor: #3171 
CLI refactor main issue: #3163

axaysagathiya:
## Changes

added a `log` flag to set the log level of modules.

## Tests

```sh
```

## Issues

Fixes #3210 
## Primary Reviewer

@timwu20


cmd/gossamer/commands/root.go
```diff
@@ -162,8 +168,17 @@ func addRootFlags(cmd *cobra.Command) error {
 	}
 
 	// Log Config
-	if err := addLogFlags(cmd); err != nil {
-		return fmt.Errorf("failed to add log flags: %s", err)
+	if err := addStringFlagBindViper(
+		cmd, "logs", "",
```
kanishkatn, line 172: - Let's use `--log` and the shorthand `-l` since that's how it is defined in substrate
- Let's specify the available modules in the usage
- We don't need to bind it to viper since we are parsing it. Let's not use `addStringFlagBindViper`. Here's an [example](https://github.com/ChainSafe/gossamer/blob/development/cmd/gossamer/commands/root.go#L251)

cmd/gossamer/commands/root.go
```diff
@@ -260,78 +271,54 @@ Expected format is 'URL VERBOSITY', e.g. ''--telemetry-url wss://foo/bar:0, wss:
...
+		"wasmer":  cfg.DefaultLogLevel,
+	}
+
+	customLog := viper.GetString("logs")
+
+	if len(customLog) > 0 {
+		logConfigurations := strings.Split(customLog, ",")
+
+		for _, config := range logConfigurations {
+			parts := strings.SplitN(config, ":", 2)
```
kanishkatn, line 296: Instead of the format `module:loglvl`, let's use `module=loglvl` since that's how substrate does it

cmd/gossamer/commands/root.go
```diff
@@ -131,6 +133,10 @@ Usage:
 				}
 			}
 
+			if err := parseLogLevel(); err != nil {
```
kanishkatn, line 136: This must move this call before the viper configuration and validation since we're gonna use the result from this

kanishkatn (changes requested):
Good job! 

It needs few changes. I'll test it after that!

cmd/gossamer/commands/root.go
kanishkatn: Ah yes, it get's overridden because `parseLogLevel()` isn't setting the viper defaults. These should help you understand: https://github.com/ChainSafe/gossamer/blob/development/cmd/gossamer/commands/utils.go#L393, https://github.com/ChainSafe/gossamer/blob/development/cmd/gossamer/commands/utils.go#L421

cmd/gossamer/commands/root.go
```diff
@@ -260,78 +266,52 @@ Expected format is 'URL VERBOSITY', e.g. ''--telemetry-url wss://foo/bar:0, wss:
...
-	}
-
-	if err := addStringFlagBindViper(cmd,
-		"ldigest",
-		config.Log.Digest,
-		"Digest module log level",
-		"log.digest"); err != nil {
-		return fmt.Errorf("failed to add --ldigest flag: %s", err)
-	}
+func parseLogLevel() error {
```
kanishkatn, line 269: Let's move this to `utils.go` since that's where the rest of the parsers are!

cmd/gossamer/commands/root.go
```diff
@@ -162,9 +168,13 @@ func addRootFlags(cmd *cobra.Command) error {
...
-	if err := addLogFlags(cmd); err != nil {
-		return fmt.Errorf("failed to add log flags: %s", err)
-	}
+	cmd.Flags().StringVarP(&logLevel, "log", "l", "",
+		`Set a logging filter.
+	Syntax is a list of 'module=logLevel' (comma separated)
+	e.g. --log sync=debug,core=trace
+	Log levels (least to most verbose) are error, warn, info, debug, and trace.
+	By default, all modules log 'info'.
+	The global log level can be set with --log global=debug`)
```
kanishkatn, line 177: Let's add the list of available modules in the usage https://github.com/ChainSafe/gossamer/issues/3210#issuecomment-1511518468

kanishkatn (changes requested):

cmd/gossamer/commands/root.go
axaysagathiya: you mentioned the `module:loglvl` format in the description of this issue. 
pushed the changes for `module=loglvl` format.

kanishkatn: Thank you for making the change @axaysagathiya !

cmd/gossamer/commands/root.go
axaysagathiya: that way log-level can't be set for the `init` command. because `configureViper()` and `ParseConfig()` will be called only for the `gossamer` command.

kanishkatn: What way exactly? I don't understand.

axaysagathiya: there is a check [here](https://github.com/ChainSafe/gossamer/blob/1a34972b4cdfd49174bac566b3b26061fc23113f/cmd/gossamer/commands/root.go#L120) to check if it's gossamer command.
so, `configureViper()` and `ParseConfig()` will run only for gossamer command.

I think we should allow setting the log level for the `init` command also. The current code is working fine, you can test it.

kanishkatn (changes requested):

cmd/gossamer/commands/root.go
axaysagathiya: It updates in my system. attaching screenshots


before running the `gossamer init`:
Screenshot from 2023-06-10 21-09-43


after running the `gossamer init` command with   `--log core=trace`:
Screenshot from 2023-06-10 21-11-35


after running the `gossamer` command with `--log core=debug`:
Screenshot from 2023-06-10 21-10-17

kanishkatn: Glad you got it working! I'll take a look later this week

kanishkatn: Just to recap, here are the changes I suggested
1. Move `parseLogLevel()` before configuring the viper. Before this [line](https://github.com/ChainSafe/gossamer/blob/development/cmd/gossamer/commands/root.go#L120)
2. Set the parsed log config to viper

As you have already noticed, the parsers are setting the parsed values in both the `config` struct and the viper store. Here is the thought behind it
- `viper` store is always the source of the config since it manages the configuration 
- the variable `config` is just an easy way for us to use the configuration without having to use constants to fetch the values via `viper.Get()`

Because of these reasons, we need to handle # 2. See [this](https://github.com/ChainSafe/gossamer/pull/3303#discussion_r1224369725) for reference

As mentioned above, viper manages the configuration for us. Here's our flow:

1. Set default config 
2. override it with the values set in flags 
3. If it is `gossamer` command, merge overridden config and the config from `config.toml` 

The priority is given to the overwritten config. 

So, if we keep it after the prerun for `gossamer` command, we are basically overriding the parsed config again; which is a bad practise. Because of this reason, we need to handle # 1

Happy to hop on a call if any of these doesn't make sense.

cmd/gossamer/commands/utils.go
```diff
@@ -481,3 +482,55 @@ func setViperDefault(config *cfg.Config) {
...
+			moduleToLogLevel[module] = logLevel
+		}
+	}
+
+	// set global log level
+	config.LogLevel = moduleToLogLevel["global"]
+	viper.Set("log-level", config.LogLevel)
+
+	// set config.Log
+	jsonData, err := json.Marshal(moduleToLogLevel)
```
kanishkatn, line 525: nice!

cmd/gossamer/commands/root.go
```diff
@@ -117,6 +117,10 @@ Usage:
 				}
 			}
 
+			if err := parseLogLevel(); err != nil {
+				return fmt.Errorf("failed to configure log level: %s", err)
+			}
```
kanishkatn, line 122: nit
```suggestion
			if err := parseLogLevel(); err != nil {
				return fmt.Errorf("failed to parse log level: %s", err)
			}
```

cmd/gossamer/commands/utils.go
```diff
@@ -481,3 +482,55 @@ func setViperDefault(config *cfg.Config) {
...
+		"wasmer":  cfg.DefaultLogLevel,
+	}
+
+	if logLevel != "" {
+		logConfigurations := strings.Split(logLevel, ",")
+		for _, config := range logConfigurations {
+			parts := strings.SplitN(config, "=", 2)
+			if len(parts) != 2 {
+				return fmt.Errorf("Invalid log configuration: %s", config)
+			}
```
kanishkatn, line 508: nit: GoLand complains
```suggestion
			if len(parts) != 2 {
				return fmt.Errorf("invalid log configuration: %s", config)
			}
```

kanishkatn (commented):
Great job! Glad to have one more contributor to the CLI. Just few minor cleanup suggestions!

kanishkatn (approved):

EclesioMeloJunior (approved):
Lgmt!

kanishkatn (changes requested):
Let's also update the docs. I know this https://github.com/ChainSafe/gossamer/blob/development/docs/docs/usage/command-line.md, but check if we need to update anywhere else!

kanishkatn (approved):

Issue #3302, opened 2023-06-07
Implement dispute coordinator subsystem


kanishkatn:
## Task summary

Implement dispute coordinator subsystem. Here are the components
- Dispute queue
- Spam slots
- Database
- Chain scraping
- Metrics
- and the dispute coordinator module

## Specification reference
- [Host implementers guide](https://paritytech.github.io/polkadot/book/node/disputes/dispute-coordinator.html)
- [Parity's implementation](https://github.com/paritytech/polkadot/tree/master/node/core/dispute-coordinator)

## Other information
#2989

kishansagathiya:
@kanishkatn @P1sar 

Given that much of work has already happened, do you think creating smaller issues would add value here? I don't have a clear answer for this.

P1sar:
Why not? Despite how much work we put there, we still not sure how much work left, right? So it is nice to at least know about it at least approximately

PR #3314, opened 2023-06-13
parachain/disputes: add participation queue for dispute-coordinator

kanishkatn:
## Changes
- Implements the queues needed for dispute participation

read the note on `queue.go` for more details.

## Issues

#3302 

## Primary Reviewer

@kishansagathiya

### DIFF
--- a/dot/parachain/dispute/queue_test.go
+++ b/dot/parachain/dispute/queue_test.go
@@ -5,8 +5,8 @@ import (
 	"sync"
 	"testing"
 
+	parachain "github.com/ChainSafe/gossamer/dot/parachain/types"
 	"github.com/ChainSafe/gossamer/lib/common"
-	"github.com/ChainSafe/gossamer/lib/parachain"
 	"github.com/google/btree"
 	"github.com/stretchr/testify/require"
 )
--- a/dot/parachain/dispute/queue_test.go
+++ b/dot/parachain/dispute/queue_test.go
@@ -59,6 +59,7 @@ func runTests(t *testing.T, tests []test, queue Queue) {
 				err := queue.Queue(tt.comparator, tt.request, tt.priority)
 				if tt.mustError {
 					require.Error(t, err)
+					require.Equal(t, err, tt.expected)
 					return
 				}
 				require.NoError(t, err)
@@ -326,6 +327,7 @@ func TestQueue_OverflowPriority(t *testing.T) {
 			request:    dummyParticipationRequest(),
 			priority:   ParticipationPriorityHigh,
 			mustError:  true,
+			expected:   errorPriorityQueueFull,
 		},
 	}
 
@@ -372,6 +374,7 @@ func TestQueue_OverflowBestEffort(t *testing.T) {
 			request:    dummyParticipationRequest(),
 			priority:   ParticipationPriorityBestEffort,
 			mustError:  true,
+			expected:   errorBestEffortQueueFull,
 		},
 	}
 
--- a/dot/parachain/dispute/queue_test.go
+++ b/dot/parachain/dispute/queue_test.go
@@ -7,14 +7,14 @@ import (
 
 	parachain "github.com/ChainSafe/gossamer/dot/parachain/types"
 	"github.com/ChainSafe/gossamer/lib/common"
-	"github.com/google/btree"
 	"github.com/stretchr/testify/require"
+	"github.com/tidwall/btree"
 )
 
 func newTestQueue(size int) *QueueHandler {
 	return &QueueHandler{
-		bestEffort:        btree.New(size / 2),
-		priority:          btree.New(size / 2),
+		bestEffort:        btree.New(participationItemComparator),
+		priority:          btree.New(participationItemComparator),
 		bestEffortMaxSize: size,
 		priorityMaxSize:   size,
 	}
--- a/dot/parachain/dispute/queue_test.go
+++ b/dot/parachain/dispute/queue_test.go
@@ -1,0 +1,589 @@
+package dispute
+
+import (
+	"encoding/binary"
+	"sync"
+	"testing"
+
+	"github.com/ChainSafe/gossamer/lib/common"
+	"github.com/ChainSafe/gossamer/lib/parachain"
+	"github.com/google/btree"
+	"github.com/stretchr/testify/require"
+)
+
+func newTestQueue(size int) *QueueHandler {
+	return &QueueHandler{
+		bestEffort:        btree.New(size / 2),
+		priority:          btree.New(size / 2),
+		bestEffortMaxSize: size,
+		priorityMaxSize:   size,
+	}
+}
+
+func newComparator(blockNumber, order uint32) CandidateComparator {
+	candidateHash := make([]byte, 4)
+	binary.LittleEndian.PutUint32(candidateHash, order)
+
+	return CandidateComparator{
+		relayParentBlockNumber: &blockNumber,
+		candidateHash:          common.NewHash(candidateHash),
+	}
+}
+
+func dummyParticipationRequest() *ParticipationRequest {
+	return &ParticipationRequest{
+		candidateHash:    [32]byte{},
+		candidateReceipt: parachain.CandidateReceipt{},
+		session:          1,
+	}
+}
+
+type test struct {
+	name string
+	// operation one of "queue", "dequeue", "prioritise", "pop_priority", "pop_best_effort",
+	// "len_priority", "len_best_effort"
+	operation  string
+	comparator CandidateComparator
+	request    *ParticipationRequest
+	priority   ParticipationPriority
+	expected   any
+	mustError  bool
+}
+
+func runTests(t *testing.T, tests []test, queue Queue) {
+	for _, tt := range tests {
+		tt := tt
+		t.Run(tt.name, func(t *testing.T) {
+			switch tt.operation {
+			case "queue":
+				err := queue.Queue(tt.comparator, tt.request, tt.priority)
+				if tt.mustError {
+					require.Error(t, err)
+					return
+				}
+				require.NoError(t, err)
+			case "dequeue":
+				item := queue.Dequeue()
+				require.Equal(t, tt.expected, item)
+			case "prioritise":
+				err := queue.PrioritiseIfPresent(tt.comparator)
+				if tt.mustError {
+					require.Error(t, err)
+					return
+				}
+				require.NoError(t, err)
+			case "pop_priority":
+				item := queue.PopPriority()
+				require.Equal(t, tt.expected, item)
+			case "pop_best_effort":
+				item := queue.PopBestEffort()
+				require.Equal(t, tt.expected, item)
+			case "len_priority":
+				require.Equal(t, tt.expected, queue.Len(ParticipationPriorityHigh))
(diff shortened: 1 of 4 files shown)


dot/parachain/dispute/queue_test.go
```diff
@@ -0,0 +1,446 @@
...
+
+	runTests(t, testData, newTestQueue(10))
+}
+
+// TestQueue_EndToEnd tests the following:
+// - queueing 3 requests into priority queue
+// - queueing 3 requests into best effort queue
+// - popping the best effort queue
+// - popping the priority queue
+// - prioritising a request in the priority queue
```
kanishkatn, line 175:
```suggestion
// - prioritising a request in the best effort queue
```

dot/parachain/dispute/queue_test.go
```diff
@@ -0,0 +1,446 @@
...
+
+func dummyParticipationRequest() *ParticipationRequest {
+	return &ParticipationRequest{
+		candidateHash:    [32]byte{},
+		candidateReceipt: []byte{1, 2, 3},
+		session:          1,
+	}
+}
+
+type tests struct {
```
kanishkatn, line 39:
```suggestion
type test struct {
```

dot/parachain/dispute/queues.go
```diff
@@ -0,0 +1,230 @@
...
+
+func (q *QueueHandler) Len(queueType ParticipationPriority) int {
+	if queueType.IsPriority() {
+		return q.priority.Len()
+	}
+
+	return q.bestEffort.Len()
+}
+
+var _ Queue = &QueueHandler{}
```
EclesioMeloJunior, line 221: avoid to allocate memory, ref:  https://stackoverflow.com/a/60663003/8157685
```suggestion
var _ Queue = (*QueueHandler)(nil)
```

dot/parachain/dispute/queues.go
```diff
@@ -0,0 +1,230 @@
...
+	if item := q.priority.DeleteMin(); item != nil {
+		return item.(*ParticipationItem)
+	}
+
+	return nil
+}
+
+func (q *QueueHandler) Len(queueType ParticipationPriority) int {
+	if queueType.IsPriority() {
+		return q.priority.Len()
```
EclesioMeloJunior, line 215: should not lock here?

dot/parachain/dispute/queues.go
```diff
@@ -0,0 +1,230 @@
...
+	priorityQueueSize = 20000
+)
+
+func (q *QueueHandler) Queue(
+	comparator CandidateComparator,
+	request *ParticipationRequest,
+	priority ParticipationPriority,
+) error {
+	if priority.IsPriority() {
+		if q.priority.Len() >= q.priorityMaxSize {
```
EclesioMeloJunior, line 148: should not lock for reading?

EclesioMeloJunior (changes requested):

dot/parachain/dispute/queues.go
kanishkatn: Yeah, lemme change it to RWMutex

dot/parachain/dispute/queues.go
kanishkatn: It's done, thank you!

dot/parachain/dispute/queues.go
kanishkatn: Added it to `Len()`

kishansagathiya (commented):
- Lets move this from `dot/parachain` to `lib/parachain`, considering that we have kept everything so far.

- These branches needs to be updated. The diff is not right at the moment.

dot/parachain/dispute/queues.go
```diff
@@ -0,0 +1,234 @@
...
+	Len(queueType ParticipationPriority) int
+}
+
+// QueueHandler implements Queue
+// It uses two btree's to store the requests. One for best effort and one for priority.
+// The queues store participationItem's.
+// The btree is ordered by the CandidateComparator of participationItem.
+type QueueHandler struct {
+	bestEffort *btree.BTree
+	priority   *btree.BTree
```
timwu20, line 124: why aren't we using `btree.Map` from the `github.com/tidwall/btree` package? We are using it in the grandpa integration and the API is closer to the rust `BtreeMap`.

dot/parachain/dispute/queues.go
```diff
@@ -0,0 +1,234 @@
...
+		q.priorityLock.Unlock()
+	} else {
+		if q.Len(ParticipationPriorityBestEffort) >= q.bestEffortMaxSize {
+			return errorBestEffortQueueFull
+		}
+
+		q.bestEffortLock.Lock()
+		q.bestEffort.ReplaceOrInsert(newParticipationItem(comparator, request))
+		q.bestEffortLock.Unlock()
+	}
```
timwu20, line 163: I think you're missing some logic here.  Don't you need to remove something from `bestEffort` and then insert into `priority` btree map when `priority.IsPriority() == true`?
```
		if priority.is_priority() {
			if self.priority.len() >= PRIORITY_QUEUE_SIZE {
				return Err(QueueError::PriorityFull)
			}
			// Remove any best effort entry, using it to replace our new
			// request.
			if let Some(older_request) = self.best_effort.remove(&comparator) {
				req.discard_timer();
				req = older_request;
			}
			// Keeping old request if any.
			match self.priority.entry(comparator) {
				Entry::Occupied(_) => req.discard_timer(),
				Entry::Vacant(vac) => {
					vac.insert(req);
				},
			}
			self.metrics.report_priority_queue_size(self.priority.len() as u64);
			self.metrics.report_best_effort_queue_size(self.best_effort.len() as u64);
		} else {
			if self.priority.contains_key(&comparator) {
				// The candidate is already in priority queue - don't
				// add in in best effort too.
				return Ok(())
			}
			if self.best_effort.len() >= BEST_EFFORT_QUEUE_SIZE {
				return Err(QueueError::BestEffortFull)
			}
			// Keeping old request if any.
			match self.best_effort.entry(comparator) {
				Entry::Occupied(_) => req.discard_timer(),
				Entry::Vacant(vac) => {
					vac.insert(req);
				},
			}
			self.metrics.report_best_effort_queue_size(self.best_effort.len() as u64);
		}
```

kanishkatn: Yes, my bad. Good catch!

dot/parachain/dispute/queues.go
kanishkatn: You're right, it does seem better. I'll change it

kanishkatn: I took a look at the APIs and it looks like we can't use the `btree.Map` here since the key needs to be a primitive type whereas we're using `CandidateComparator` struct. I'll use `btree.BTree` instead. 

The google's btree seems API rich whereas `github.com/tidwall/btree` seems to have better performance. I'll change it

kanishkatn: It's done!

dot/parachain/dispute/queues.go
kanishkatn: I've added the missing checks.

dot/parachain/dispute/queue_test.go
```diff
@@ -0,0 +1,589 @@
...
+
+func runTests(t *testing.T, tests []test, queue Queue) {
+	for _, tt := range tests {
+		tt := tt
+		t.Run(tt.name, func(t *testing.T) {
+			switch tt.operation {
+			case "queue":
+				err := queue.Queue(tt.comparator, tt.request, tt.priority)
+				if tt.mustError {
+					require.Error(t, err)
```
jimjbrettj, line 61: Nit/perhaps assert against an expected error here

dot/parachain/dispute/queues.go
```diff
@@ -0,0 +1,242 @@
...
+}
+
+// ParticipationItem implements btree.Item
+type ParticipationItem struct {
+	comparator CandidateComparator
+	request    *ParticipationRequest
+}
+
+func participationItemComparator(a, b any) bool {
+	pi1, pi2 := a.(*ParticipationItem), b.(*ParticipationItem)
```
jimjbrettj, line 45: should we check that this cast is successful before using it?

dot/parachain/dispute/queues.go
```diff
@@ -0,0 +1,242 @@
...
+
+	if pi1.comparator.relayParentBlockNumber == nil && pi2.comparator.relayParentBlockNumber == nil {
```
jimjbrettj, line 47: do we need a nil check on the comparator itself?

dot/parachain/dispute/queues.go
```diff
@@ -0,0 +1,242 @@
...
+)
+
+// IsPriority returns true if the priority is high
+func (p ParticipationPriority) IsPriority() bool {
+	return p == ParticipationPriorityHigh
+}
+
+var (
+	// errorBestEffortQueueFull is returned when the best effort queue is full and the request could not be processed
+	errorBestEffortQueueFull = errors.New("best effort queue is full")
```
jimjbrettj, line 90: nit/errors at top of file

dot/parachain/dispute/queues.go
```diff
@@ -0,0 +1,242 @@
...
+
+// QueueHandler implements Queue
+// It uses two btrees to store the requests. One for best effort and one for priority.
+// The queues store participationItem's.
+// The btree is ordered by the CandidateComparator of participationItem.
+type QueueHandler struct {
+	bestEffort *btree.BTree
+	priority   *btree.BTree
+
+	bestEffortLock sync.RWMutex
```
jimjbrettj, line 124: should we have these locks within the same structure that they are associated with? example being make a struct with a bTree and the mutex for BestEffort

jimjbrettj (commented):
looks good, just a few comments

EclesioMeloJunior (approved):
LGTM

dot/parachain/dispute/queues.go
kanishkatn: it should always be successful since we can only insert a `ParticipationItem` to the btree

dot/parachain/dispute/queues.go
kanishkatn: comparator can never be nil. See https://github.com/ChainSafe/gossamer/blob/feat/k/participation-queue/dot/parachain/dispute/queues.go#L98

dot/parachain/dispute/queues.go
kanishkatn: That's an interesting thought. I'd rather not abstract it further though cause there is too much of it already. But, I am open to it if there is any added value to it.

dot/parachain/dispute/queue_test.go
kanishkatn: done

dot/parachain/dispute/queues.go
kanishkatn: I've made the change. This is really nice, I'll do it in rest of the modules as well. Thank you!

dot/parachain/dispute/queues.go
kanishkatn: I'm going to merge this PR to make it a bit easy for me to manage the branches. I'm happy to pick these conversations later if needed!

PR #3315, opened 2023-06-13
feat(pkg/scale): Use `New()` receiver function for construction of custom `VaryingDataType` 

timwu20:
## Changes

- When using a custom `VaryingDataType`, the `scale` decoder now looks for a `New()` method which will be used to construct the custom `VaryingDataType` and initialize the internal cache of the associated `VaryingDataTypeValue` types.

## Tests

```sh
go test -tags integration github.com/ChainSafe/gossamer/pkg/scale
```

## Primary Reviewer

@edwardmack


jimjbrettj (approved):
I think this is a nice addition

edwardmack (approved):
Excellent, this will solve my use case.

PR #3316, opened 2023-06-13
feat(parachain): Add StatementDistributionMessage varingDataType

### LINKED ISSUE #3289: Create a go struct for Statement Distribution Message
## Issue summary

Translate rust struct for Statement Distribution Message into go. 
Make sure that our go struct is same as rust using scale encoding.
- Make sure that both struct produces the same scale encoding.
- Make sure that scale encoding of go struct for `Statement Distribution Message ` can be decoded into rust struct for `Statement Distribution Message` and they have the same internal properties
- Make sure that scale encodoing of rust struct for `Statement Distribution Message ` can be decoded into go struct for `Statement Distribution Message` and they have the same internal properties

- Spec for Statement Distribution Message is at https://spec.polkadot.network/chapter-anv#net-msg-statement-distribution
- Rust code here https://github.com/paritytech/polkadot/blob/8eb5efcdaec745e16167043c75e2dd4da958cbc8/node/network/protocol/src/lib.rs#L421
- Our scale library https://github.com/ChainSafe/gossamer/tree/development/pkg/scale

axaysagathiya:
## Tests

```sh
```

## Issues

Fixes #3289 

## Primary Reviewer

@timwu20


parachain/statement_distribution_message.go
```diff
@@ -0,0 +1,73 @@
...
+
+type UncheckedSignedFullStatement struct {
+	Payload        Statement          `scale:"1"`
+	ValidatorIndex ValidatorIndex     `scale:"2"`
+	Signature      ValidatorSignature `scale:"3"`
+	// RealPayload    CompactStatement   `scale:"4"` // changes needed
+}
+
+type PhantomData struct{}
+
```
kishansagathiya, line 60:
```suggestion
```

parachain/statement_distribution_message.go
```diff
@@ -0,0 +1,73 @@
...
+func (l SecondedStatementWithLargePayload) Index() uint {
+	return 1
+}
+
+type UncheckedSignedFullStatement struct {
+	Payload        Statement          `scale:"1"`
+	ValidatorIndex ValidatorIndex     `scale:"2"`
+	Signature      ValidatorSignature `scale:"3"`
+	// RealPayload    CompactStatement   `scale:"4"` // changes needed
```
kishansagathiya, line 56:
```suggestion
```

kishansagathiya (commented):
:tada: Great work. 


Few suggestions to make this ready
- Add comment to fix linting
- Now that types PR got merged, see which structs can be reused.

kishansagathiya (approved):
epic

kanishkatn (changes requested):

kanishkatn:
Question, do we need so many files? I am not aware of how many such types are there but I feel we'll end up having a lot of files at this pace. 
I ended up putting all the types in a single file for runtime calls.

lib/parachain/statement_distribution_message_test.go
```diff
@@ -0,0 +1,199 @@
...
+		fmt.Printf("Error unmarshaling test data: %s\n", err)
+		return
+	}
+}
+
+func TestStatementDistributionMessage(t *testing.T) {
+	t.Parallel()
+
+	var collatorSignature CollatorSignature
+	tempSignature := common.MustHexToBytes("0xc67cb93bf0a36fcee3d29de8a6a69a759659680acf486475e0a2552a5fbed87e45adce5f290698d8596095722b33599227f7461f51af8617c8be74b894cf1b86") //nolint:lll
```
kanishkatn, line 31: Is there a reason for not moving this to yaml or you just missed it?

axaysagathiya: Moving this to YAML wouldn't be beneficial as it's not so long string.

kanishkatn: hmm, I see that it is using `//nolint:lll` to skip the linter and I believe the limit is set to 125 characters. The signature has a length of 130. I feel we should move this as well

kanishkatn:
> @kanishkatn Are you talking about just YAML files or do you think there are a lot of files in general?

The go files. I wasn't too bothered about the yaml files, but yeah it would help us to keep it cleaner as well!

axaysagathiya:
yup, I think we should maintain separate files.
otherwise, it would be so huge file if we put everything in a single file.

@kishansagathiya what's your take on this?

kishansagathiya:
@kanishkatn @axaysagathiya 
I feel like some of these things would form its own packages eventually, the way it is in polkadot. Like backing, validation, dispute, availability all with have its own packages. So, far I don't think we have that many files to make individual packages.

> Question, do we need so many files?

I am happy with number of file, but If we have just two files `statement.go` and `statement_test.go` that wouldn't be bad as well.

kanishkatn:
> @kanishkatn @axaysagathiya I feel like some of these things would form its own packages eventually, the way it is in polkadot. Like backing, validation, dispute, availability all with have its own packages. So, far I don't think we have that many files to make individual packages.

I wasn't suggesting to make individual packages. Let me rephrase

Many of these types will be used across packages, at least from what I've seen in the dispute-subsystem. So. it does make sense to keep the common types in `lib/parachain`. Naming the files based on the subsystems now itself ( since we're leaning that way ) will help us in the future refactor and we'll have less tech debts.

kanishkatn (approved):
Just a concern https://github.com/ChainSafe/gossamer/pull/3316#issuecomment-1597207002, but I guess we can cross that bridge when we get there!

looks good to me otherwise!

PR #3333, opened 2023-06-16
feat(dot/network): introduce libp2p resource manager + prometheus metrics

EclesioMeloJunior:
## Changes

- Introduce libp2p resource manager + Prometheus metrics to the `network` pkg

A little explanation of this library [go-libp2p/resource-manager](https://github.com/libp2p/go-libp2p/tree/master/p2p/host/resource-manager)
> The Resource Manager doesn't prioritize resource requests at all, it simply checks if the resource being requested is currently below the defined limits and returns an error if the limit is reached. It has no notion of honest vs bad peers.
## Tests

- Running gossamer with libp2p resource manager + exposing the metrics through Prometheus. It is possible to see the number of streams in use per protocol:

image

## Issues

- N/A

## Primary Reviewer

@timwu20

### DIFF
--- a/dot/network/host.go
+++ b/dot/network/host.go
@@ -26,7 +26,10 @@ import (
 	"github.com/libp2p/go-libp2p/core/peerstore"
 	"github.com/libp2p/go-libp2p/core/protocol"
 	"github.com/libp2p/go-libp2p/p2p/host/peerstore/pstoreds"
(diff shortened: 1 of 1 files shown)


kanishkatn (approved):
looks good!

dimartiro (approved):
LGTM

PR #3334, opened 2023-06-16
chore(dot/network): added common request-response protocol

kishansagathiya:
## Tests

```sh
```

## Primary Reviewer

@EclesioMeloJunior


dot/network/request_response.go
```diff
@@ -0,0 +1,92 @@
...
+	}()
+
+	if err = rrp.host.writeToStream(stream, req); err != nil {
+		return err
+	}
+
+	return rrp.ReceiveResponse(stream, res)
+}
+
+func (rrp *requestResponseProtocol) ReceiveResponse(stream libp2pnetwork.Stream, msg ResponseMessage) error {
```
EclesioMeloJunior, line 58: receive response could be a private method since it will be used only by the `DoRequest` method
```suggestion
func (rrp *requestResponseProtocol) ReceiveResponse(stream libp2pnetwork.Stream, msg ResponseMessage) error {
```

dot/network/service.go
```diff
@@ -576,6 +573,20 @@ func (s *Service) SendMessage(to peer.ID, msg NotificationsMessage) error {
 	return errors.New("message not supported by any notifications protocol")
 }
 
+func (s *Service) GetRequestResponseProtocol(subprotocol string, requestTimeout time.Duration,
+	maxResponseSize uint64) RequestResponseProtocol {
```
EclesioMeloJunior, line 577: Same here, we don't need to return an interface, we can return the actual implementation. ref: [accept-interfaces-return-structs-in-go](https://bryanftan.medium.com/accept-interfaces-return-structs-in-go-d4cab29a301b)

```suggestion
func (s *Service) GetRequestResponseProtocol(subprotocol string, requestTimeout time.Duration,
	maxResponseSize uint64) RequestResponseProtocol {
```

EclesioMeloJunior (changes requested):
Nice work so far!! Will provide more freedom to use the network layer, just few points

EclesioMeloJunior (approved):

dot/network/request_response.go
```diff
@@ -0,0 +1,92 @@
...
+		return err
+	}
+
+	return rrp.receiveResponse(stream, res)
+}
+
+func (rrp *RequestResponseProtocol) receiveResponse(stream libp2pnetwork.Stream, msg ResponseMessage) error {
+	// allocating a new (large) buffer every time slows down receiving response by a dramatic amount,
+	// as malloc is one of the most CPU intensive tasks.
+	// thus we should allocate buffers at startup and re-use them instead of allocating new ones each time.
```
jimjbrettj, line 61: nice this sounds like a good chance

jimjbrettj (approved):
looks very nice, great stuff!

PR #3336, opened 2023-06-19
parachain/disputes: participation module

kanishkatn:
## Changes

Implements the participation module for disputes-coordinator. Please find parity's implementation here: https://github.com/paritytech/polkadot/tree/master/node/core/dispute-coordinator/src/participation

It uses a queue #3314 

The `overseer` package is there only for testing purposes. No need to go through it.

## Issues
#3302

## Primary Reviewer

@kishansagathiya 
@EclesioMeloJunior if you have the time!


dot/parachain/dispute/overseer/message.go
```diff
@@ -0,0 +1,83 @@
...
+	ParentHead             []byte
+	RelayParentNumber      uint32
+	RelayParentStorageRoot common.Hash
+	MaxPOVSize             uint32
+}
+
+type AvailableData struct {
+	POV            []byte
+	ValidationData PersistedValidationData
+}
```
axaysagathiya, line 23: We already have [AvailableData](https://github.com/ChainSafe/gossamer/blob/f9b3281b3e1798510948499b6a42900334a97edc/dot/parachain/available_data_fetching.go#L52)

dot/parachain/dispute/overseer/message.go
```diff
@@ -0,0 +1,83 @@
...
+type PvfExecTimeoutKind uint32
+
+const (
+	PvfExecTimeoutKindBacking PvfExecTimeoutKind = iota
+	PvfExecTimeoutKindApproval
+)
+
+type ValidateFromChainState struct {
+	CandidateReceipt   parachainTypes.CandidateReceipt
+	PoV                []byte
```
axaysagathiya, line 64: we have [PoV](https://github.com/ChainSafe/gossamer/blob/f9b3281b3e1798510948499b6a42900334a97edc/dot/parachain/available_data_fetching.go#L62) type. Can you use it here?

dot/parachain/dispute/participation.go
```diff
@@ -0,0 +1,376 @@
...
+}
+
+// NewCandidateComparator creates a new CandidateComparator.
+func NewCandidateComparator(relayParentBlockNumber *uint32,
+	receipt parachainTypes.CandidateReceipt,
+) (CandidateComparator, error) {
+
+	candidateHash, err := receipt.Hash()
+	if err != nil {
+		return CandidateComparator{}, fmt.Errorf("hash candidate receipt: %w", err)
```
axaysagathiya, line 31: i don't see you using `CandidateComparator` instance in case of error. you can return nil.

dot/parachain/dispute/participation.go
```diff
@@ -0,0 +1,376 @@
...
+func (p *ParticipationHandler) numberOfWorkers() int {
+	return int(p.workers.Load())
+}
+
+func (p *ParticipationHandler) dequeueUntilCapacity(recentHead common.Hash) {
+	for {
+		if p.numberOfWorkers() >= MaxParallelParticipation {
+			break
+		}
+
```
axaysagathiya, line 201: I think you can write this in a single line
```suggestion
	for p.numberOfWorkers() < MaxParallelParticipation {
```

dot/parachain/dispute/overseer/message.go
kanishkatn: The types in the overseer is temporary. Please see the comment on top of the file.

dot/parachain/dispute/overseer/message.go
kanishkatn: The types in the overseer is temporary. Please see the comment on top of the file.

dot/parachain/dispute/participation.go
kanishkatn: The reason is because I don't want to return a pointer here. It's interesting, could you tell me why you'd prefer it the other way in this case?

dot/parachain/dispute/participation.go
kanishkatn: I agree, that's neater. Thank you!

dot/parachain/dispute/participation.go
```diff
@@ -0,0 +1,372 @@
...
+			return fmt.Errorf("validation failed: %s", result.Error)
+		}
+
+		if !result.IsValid {
+			sendResult(p.sender, request, types.ParticipationOutcomeInvalid)
+			return fmt.Errorf("validation failed: %s", result.Error)
+		} else {
+			sendResult(p.sender, request, types.ParticipationOutcomeValid)
+			return nil
+		}
```
EclesioMeloJunior, line 305:
```suggestion
		}
		
		sendResult(p.sender, request, types.ParticipationOutcomeValid)
		return nil
```

dot/parachain/dispute/participation_test.go
```diff
@@ -0,0 +1,633 @@
...
+					panic("invalid outcome")
+				}
+			}).Times(4)
+
+			time.Sleep(2 * time.Second)
+		}
+
+		go participationTest()
+		go requestHandler()
+		wg.Add(2)
```
EclesioMeloJunior, line 192:
```suggestion
		wg.Add(2)
		go participationTest()
		go requestHandler()
```

dot/parachain/mocks_generate_test.go
```diff
@@ -4,4 +4,4 @@
 package parachain
 
 //go:generate mockgen -destination=mocks_test.go -package=$GOPACKAGE . PoVRequestor
-//go:generate mockgen -destination=mocks_runtime_test.go -package $GOPACKAGE github.com/ChainSafe/gossamer/dot/parachain/runtime RuntimeInstance
+//go:generate mockgen -destination=runtime_mock.go -package $GOPACKAGE github.com/ChainSafe/gossamer/dot/parachain/runtime RuntimeInstance
```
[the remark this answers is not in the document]
kanishkatn: The reason for not using the `_test` suffix was so that It can be accessed from another package. It doesn't export them otherwise. Do you know any other way to use it?

dot/parachain/dispute/overseer/mocks_overseer.go
```diff
@@ -0,0 +1,61 @@
+// Code generated by MockGen. DO NOT EDIT.
```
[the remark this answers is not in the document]
kanishkatn: Resolving this to keep the conversation in one place.

dot/parachain/dispute/participation.go
```diff
@@ -0,0 +1,372 @@
...
+)
+
+// CandidateComparator comparator for ordering of disputes for candidate.
+type CandidateComparator struct {
+	relayParentBlockNumber *uint32
+	candidateHash          common.Hash
+}
+
+// NewCandidateComparator creates a new CandidateComparator.
+func NewCandidateComparator(relayParentBlockNumber *uint32,
```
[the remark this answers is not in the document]
kanishkatn: Could you elaborate on why you'd prefer it that way?

dot/parachain/dispute/participation.go
```diff
@@ -0,0 +1,372 @@
...
+	}
+
+	p.dequeueUntilCapacity(p.recentBlock.Hash)
+	return nil
+}
+
+func (p *ParticipationHandler) ProcessActiveLeavesUpdate(update overseer.ActiveLeavesUpdate) error {
+	if update.Activated == nil {
+		return nil
+	}
```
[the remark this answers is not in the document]
kanishkatn: Wouldn't that be a repetition? I believe this is being called from multiple places.

dot/parachain/dispute/participation.go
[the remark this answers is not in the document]
EclesioMeloJunior: I will rephrase myself hahah. So `New...` functions could return an error if an error was found while creating the instance/object, for example `os.Create` return an error if it could not create a file/folder in the OS. However in this case `NewCandidateComparator` receives the `CandidateReceipt` and executes the `Hash()` method which if fails return, but IMO this error is not intrinsic to the instance

Same for `getBlockNumber` function that receives the `CandidateReceipt` and do the same check but could receive only the hash as an argument 

```go
receiptHash, err :=  request.candidateReceipt.Hash()
if err != nil { ... }

blockNumber, err := getBlockNumber(ctx.Sender, receiptHash)
if err != nil {
	return fmt.Errorf("get block number: %w", err)
}

comparator :=  NewCandidateComparator(&blockNumber, receiptHash)
```

dot/parachain/dispute/participation.go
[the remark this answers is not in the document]
EclesioMeloJunior: I didn't find any usage other than one in a test file, could you explain to me when this found will be used

dot/parachain/dispute/participation.go
```diff
@@ -0,0 +1,372 @@
...
+
+	if p.recentBlock == nil {
+		panic("we never ever reset recentBlock to nil and we already received a result, so it must have been set before. qed")
+	}
+
+	p.dequeueUntilCapacity(p.recentBlock.Hash)
+	return nil
+}
+
+func (p *ParticipationHandler) ProcessActiveLeavesUpdate(update overseer.ActiveLeavesUpdate) error {
```
EclesioMeloJunior, line 135: could you remove the return since all returns are nil?

dot/parachain/dispute/participation.go
```diff
@@ -0,0 +1,372 @@
...
+		if update.Activated.Number > p.recentBlock.Number {
+			p.recentBlock.Number = update.Activated.Number
+			p.recentBlock.Hash = update.Activated.Hash
+		}
+	}
+
+	return nil
+}
+
+func (p *ParticipationHandler) BumpPriority(ctx overseer.Context, receipts []parachainTypes.CandidateReceipt) error {
```
EclesioMeloJunior, line 157: could you remove the return since all returns are nil?

EclesioMeloJunior (changes requested):

dot/parachain/mocks_generate_test.go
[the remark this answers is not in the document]
kanishkatn: Ah, that makes sense. Thank you!

dot/parachain/dispute/participation.go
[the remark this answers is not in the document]
kanishkatn: Yeah, it's a standalone module in this PR and it gets used by the coordinator. I think I'll leave a todo for this and come back once the integration is complete!

dot/parachain/dispute/participation.go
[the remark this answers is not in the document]
kanishkatn: I agree, thank you!

dot/parachain/types/types.go
```diff
@@ -26,6 +26,10 @@ type ValidatorID [sr25519.PublicKeyLength]byte
 // BlockNumber The block number type.
 type BlockNumber uint32
 
+func (b BlockNumber) Encode() ([]byte, error) {
```
jimjbrettj, line 29: why does block number need to be custom encoded as a pointer?

dot/parachain/types/types.go
```diff
@@ -323,6 +349,25 @@ type CandidateReceipt struct {
 	CommitmentsHash common.Hash `scale:"2"`
 }
 
+// Encode returns the SCALE encoding of the receipt.
+func (cr CandidateReceipt) Encode() ([]byte, error) {
```
jimjbrettj, line 353: Do we need custom encode methods here?

jimjbrettj (commented):
Looks good, just a few comments

dot/parachain/types/types.go
kanishkatn: I believe this is being used by the backend while storing the data. It does make the code simple if we abstract it out. Is there a reason not to use custom encode method for certain types?

dot/parachain/types/types.go
kanishkatn: Good catch! It was only being used by tests. I've removed it.

dot/parachain/dispute/participation.go
```diff
@@ -0,0 +1,372 @@
...
+	p.dequeueUntilCapacity(p.recentBlock.Hash)
+	return nil
+}
+
+func (p *ParticipationHandler) ProcessActiveLeavesUpdate(update overseer.ActiveLeavesUpdate) error {
+	if update.Activated == nil {
+		return nil
+	}
+
+	if p.recentBlock == nil {
```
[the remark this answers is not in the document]
kanishkatn: haha, I guess I wasn't thinking. Thank you!

dot/parachain/dispute/participation.go
kanishkatn: It's done!

dot/parachain/dispute/participation.go
kanishkatn: I've removed it!

dot/parachain/mocks_generate_test.go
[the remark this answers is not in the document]
kanishkatn: I've reverted the changes and have kept the mocks in the dispute package itself!

EclesioMeloJunior (approved):

dot/parachain/types/types.go
jimjbrettj: There are some cases where custom encoding is required. But unless it is actually required (data is not scale encoded, to meet an interface constraint, etc) there probably is not need.

For example, unless there is context that I am missing, this Encode function is just a wrapper around marshalling the data. IMO this is not necessary because on the caller side it is the exact same amount of complexity, and to me it makes more sense to just marshal directly rather than calling a separate encode method that simply just calls Marshal.

kanishkatn: I believe this is being used to hash the `CandidateReceipt` and we need to scale encode it to get the hash. 

Since this is being used only in the `Hash()` method of `CandidateReceipt`, I will remove it for now and maybe create the `Encode()` method if it is being used in multiple places in the future.

Thank you!

PR #3344, opened 2023-06-19
[WIP] feat: dispute coordinator backend

kanishkatn:
## Tests

```sh
```

## Primary Reviewer

@timwu20


kishansagathiya (commented):
lets move this from `dot/parachain` to `lib/parachain`, considering that we have kept everything so far

EclesioMeloJunior (commented):
Since we moved from chaindb/badger to pebble is it possible to change the `BadgerBackend` to be a wrapper over pebble db?

kanishkatn:
> Since we moved from chaindb/badger to pebble is it possible to change the `BadgerBackend` to be a wrapper over pebble db?

Yeah, `BadgerBackend` implements `DBBackend`. Has the badger been removed completely?

EclesioMeloJunior:
> > Since we moved from chaindb/badger to pebble is it possible to change the `BadgerBackend` to be a wrapper over pebble db?
> 
> Yeah, `BadgerBackend` implements `DBBackend`. Has the badger been removed completely?

Almost all the badger interaction was through a repository called chaindb however, there is some places that use badger like `offline_pruner`, `prune_state_test` and the discovery uses the `badger.Datastore`. But currently we use pebble db to store all important info like finalized blocks and babe epoch changes...

kanishkatn:
> > > Since we moved from chaindb/badger to pebble is it possible to change the `BadgerBackend` to be a wrapper over pebble db?
> > 
> > 
> > Yeah, `BadgerBackend` implements `DBBackend`. Has the badger been removed completely?
> 
> Almost all the badger interaction was through a repository called chaindb however, there is some places that use badger like `offline_pruner`, `prune_state_test` and the discovery uses the `badger.Datastore`. But currently we use pebble db to store all important info like finalized blocks and babe epoch changes...

I see. Thanks for sharing!

It looks like pebbleDb doesn't support transactions. It would require a bit of effort to implement the interface. Here's where it is mostly used https://github.com/paritytech/polkadot/blob/master/node/core/dispute-coordinator/src/db/v1.rs#L139

What do you think?

dot/parachain/dispute/backend.go
```diff
@@ -0,0 +1,207 @@
...
+	Backend
+
+	// Write writes the given data to the database.
+	Write(earliestSession *parachainTypes.SessionIndex,
+		recentDisputes *btree.BTree,
+		candidateVotes map[types.Comparator]*types.CandidateVotes) error
+}
+
+type syncedEarliestSession struct {
+	*sync.RWMutex
```
jimjbrettj, line 52: Why is the mutex a pointer?

dot/parachain/dispute/backend.go
```diff
@@ -0,0 +1,207 @@
...
+
+	"github.com/ChainSafe/gossamer/dot/parachain/dispute/types"
+	parachainTypes "github.com/ChainSafe/gossamer/dot/parachain/types"
+	"github.com/ChainSafe/gossamer/lib/common"
+	"github.com/dgraph-io/badger/v4"
+	"github.com/tidwall/btree"
+)
+
+// Backend is the backend for the dispute coordinator module.
+type Backend interface {
```
jimjbrettj, line 15: Would it be beneficial to have an interfaces file here? I feel like if we plan to use mocks for testing it can be nice to keep the interfaces together

dot/parachain/dispute/backend_test.go
```diff
@@ -0,0 +1,153 @@
...
+	"github.com/ChainSafe/gossamer/lib/common"
+	"github.com/dgraph-io/badger/v4"
+	"github.com/stretchr/testify/require"
+	"github.com/tidwall/btree"
+)
+
+func TestOverlayBackend_EarliestSession(t *testing.T) {
+	t.Parallel()
+
+	// with
```
jimjbrettj, line 19: What is this with, when, then?

dot/parachain/dispute/participation.go
```diff
@@ -340,6 +355,7 @@ func getBlockNumber(sender overseer.Sender, receipt parachainTypes.CandidateRece
 }
 
 func sendResult(sender overseer.Sender, request ParticipationRequest, outcome types.ParticipationOutcomeType) {
+	fmt.Println("send result")
```
jimjbrettj, line 358: nit/ remove

dot/parachain/dispute/test_helpers.go
```diff
@@ -0,0 +1,113 @@
...
+		commitmentsHash = *commitments
+	}
+
+	return parachainTypes.CandidateReceipt{
+		Descriptor:      dummyCandidateDescriptorBadSignature(relayParent),
+		CommitmentsHash: commitmentsHash,
+	}, nil
+}
+
+func dummyHash() common.Hash {
```
jimjbrettj, line 78: I see we do a mix of using this as well as just declaring a common.Hash{}. Is this helper needed?

dot/parachain/dispute/types/dispute.go
```diff
@@ -0,0 +1,45 @@
...
+	}
+
+	return &Dispute{
+		Comparator:    Comparator{},
+		DisputeStatus: disputeStatus,
+	}, nil
+}
+
+// DisputeComparator compares two disputes.
+func DisputeComparator(a, b any) bool {
```
jimjbrettj, line 37: if this is just for Dispute types, do inputs need to be generic?

dot/parachain/dispute/types/status.go
```diff
@@ -0,0 +1,231 @@
...
+			}
+		}
+	default:
+		return fmt.Errorf("invalid dispute status type %T", val)
+	}
+
+	return nil
+}
+
+var ErrDisputeNotConcluded = errors.New("dispute not concluded")
```
jimjbrettj, line 136: nit/suggestion: errors declared at top of file

dot/parachain/dispute/types/vote.go
```diff
@@ -0,0 +1,291 @@
...
+func (c *CandidateVoteState) IsConcludedAgainst() (bool, error) {
+	if c.DisputeStatus == nil {
+		return false, nil
+	}
+
+	return c.DisputeStatus.IsConcludedAgainst()
+}
+
+// NewCandidateVoteState creates a new CandidateVoteState
+// TODO: implement this later since nothing is using it yet
```
jimjbrettj, line 199: Should we have an issue for this?

dot/parachain/dispute/types/vote.go
```diff
@@ -0,0 +1,291 @@
...
+	return &CandidateVoteState{
+		Votes: votes,
+		Own:   ownVoteState,
+	}, nil
+}
+
+// CandidateVotes is a struct containing the votes for a candidate.
+type CandidateVotes struct {
+	CandidateReceipt parachainTypes.CandidateReceipt `scale:"1"`
+	// TODO: check if we need to use btree for this in the future
```
jimjbrettj, line 274: is this TODO resolved?

jimjbrettj (commented):
looks good, just a few comments!

dot/parachain/dispute/backend.go
kanishkatn: Hmm, I don't recall. It could have been cause of a Deepsource warning, I'm not sure.

dot/parachain/dispute/backend.go
kanishkatn: Yeah, maybe. Since it isn't using mocks at the moment, how about we revisit this if it changes?

dot/parachain/dispute/backend_test.go
kanishkatn: Just something I picked up from parity devs. `with` the give data, `when` we perform operations, `then` assert the responses!

dot/parachain/dispute/test_helpers.go
kanishkatn: Yeah, prolly not. These tests don't have any checks on the values.

dot/parachain/dispute/types/dispute.go
kanishkatn: This is being used as the `less` argument for btree. It expects this signature.

dot/parachain/dispute/types/vote.go
kanishkatn: I've already implemented it in `feat/k/disputes-coordinator`. Also, this is just going into another feature branch. I don't think we need an issue for this.

dot/parachain/dispute/types/vote.go
kanishkatn: No, not yet. I'm hoping to get a better visibility during the coordinator tests.

dot/parachain/dispute/participation.go
kanishkatn: This file is no longer part of this pr.

dot/parachain/dispute/backend.go
EclesioMeloJunior, line 52: I would suggest using a property name to hold those types instead of embedding them in the struct

```go
type syncedEarliestSession struct {
   mtx sync.RWMutex
   sessionIndex *parachainTypes.SessionIndex
}
```

dot/parachain/dispute/backend.go
```diff
@@ -0,0 +1,207 @@
...
+func newSyncedEarliestSession() syncedEarliestSession {
+	return syncedEarliestSession{
+		RWMutex: new(sync.RWMutex),
+	}
+}
+
+type syncedRecentDisputes struct {
+	*sync.RWMutex
+	*btree.BTree
+}
```
EclesioMeloJunior, line 65: I would suggest using a property name to hold those types instead of embedding them in the struct

```suggestion
type syncedRecentDisputes struct {
	mtx *sync.RWMutex
	btree *btree.BTree
}
```

dot/parachain/dispute/backend.go
```diff
@@ -0,0 +1,207 @@
...
+	return syncedRecentDisputes{
+		RWMutex: new(sync.RWMutex),
+		BTree:   btree.New(types.DisputeComparator),
+	}
+}
+
+type syncedCandidateVotes struct {
+	*sync.RWMutex
+	votes map[types.Comparator]*types.CandidateVotes
+}
```
EclesioMeloJunior, line 77:
```suggestion
type syncedCandidateVotes struct {
	mtx *sync.RWMutex
	votes map[types.Comparator]*types.CandidateVotes
}
```

EclesioMeloJunior (changes requested):

EclesioMeloJunior:
> It looks like pebbleDb doesn't support transactions. It would require a bit of effort to implement the interface. Here's where it is mostly used https://github.com/paritytech/polkadot/blob/master/node/core/dispute-coordinator/src/db/v1.rs#L139
Do you think it would be an easy switch or could we continue to use badger in the subsystem?

You're right, but AFAIK you can use the pebble's Batcher which could replace the current badger transactions, more exactly the IndexedBatcher (https://github.com/cockroachdb/pebble/blob/master/batch.go), basically you can aggregate operations and commit or reset them

dot/parachain/dispute/backend.go
kanishkatn: Could you elaborate on why you'd prefer it that way?

dot/parachain/dispute/backend.go
kanishkatn: Resolving this to keep the conversation in one place.

dot/parachain/dispute/backend.go
kanishkatn: Resolving this to keep the conversation in one place.

kanishkatn:
> > It looks like pebbleDb doesn't support transactions. It would require a bit of effort to implement the interface. Here's where it is mostly used https://github.com/paritytech/polkadot/blob/master/node/core/dispute-coordinator/src/db/v1.rs#L139
> > Do you think it would be an easy switch or could we continue to use badger in the subsystem?
> 
> You're right, but AFAIK you can use the pebble's Batcher which could replace the current badger transactions, more exactly the IndexedBatcher (https://github.com/cockroachdb/pebble/blob/master/batch.go), basically you can aggregate operations and commit or reset them

hmm, that's interesting. I had come across this old issue https://github.com/cockroachdb/pebble/issues/581. In their readme, they suggest rocksDB for transactions as well. I'm gonna have to go through it, but for now I'd like to continue with the integration tests. How about we revisit this at a later stage?

dot/parachain/dispute/backend.go
EclesioMeloJunior: After some research, I couldn't find any real constraint for using this implicit composition style. Looks like you want to have those structures concurrently safe since most of them use `sync.RWMutex` so I would suggest including the prefix/suffix `sync` or `shared` (similar to what they use in rust for structures that need to be protected against concurrent read/write). Also, I would suggest to use the `sync.RWMutext` value rather than the pointer to it

jimjbrettj: If its just for deepsource and the issue is embedding the mutex, perhaps it's better to have it as an attribute rather than embed?

dot/parachain/dispute/backend_test.go
jimjbrettj: hmmm interesting I have not seen this before! I'm somewhat skeptical of having to many different ways of formatting tests, but unless others disagree seems okay to me

dot/parachain/dispute/types/vote.go
jimjbrettj: sounds good to me!

dot/parachain/dispute/backend.go
jimjbrettj: I think this change would fix any deep source issues with embedding as well

jimjbrettj (approved):
some remaining comments, but lgtm

dot/parachain/dispute/backend.go
kanishkatn: No, the issue was something to do with not using the pointer. Could you help me understand what's wrong with the embed pointer?

dot/parachain/dispute/backend_test.go
kanishkatn: Yeah, it's mostly used in ink!. This format can't be used for all the tests. It just made sense here.

Are we following some sort of standards for tests?

dot/parachain/dispute/backend.go
kanishkatn: Sorry, been meaning to ping Eclesio on this; been a crazy couple of weeks. 

@EclesioMeloJunior I don't understand, doesn't it already include the prefix?

kanishkatn: I'll also check the deepsource and get more data on the warning!

EclesioMeloJunior: @kanishkatn nevermind I didn't pay attention it was already as the prefix 😅

jimjbrettj: I dont have any issue with it, I just thought that I recalled getting deepsource issues about embedding mutexs, but if this is not an issue sweet! Aside from that was just curious why a pointer was preferred over non-pointer, but I have no objection to its use

dot/parachain/dispute/backend_test.go
jimjbrettj: Well cool thank you for informing me! 

I would say we do not at this moment, as we do different things in different parts of the code, so I think if this makes sense here I am good with it

dot/parachain/dispute/backend.go
jimjbrettj: Yeah aside from deciding whether to embed or not, I am not sure of why its better to have a pointer to the mutex. 

My personal opinion is if I pointer is not needed then do not use one, though others may disagree here. Is there any benefit here to using a pointer?

kanishkatn: and deepsource doesn't complain about it anymore, haha. I've removed the pointer embed

kanishkatn: https://github.com/ChainSafe/gossamer/pull/3344#discussion_r1328508507

I'll resolve this!

kanishkatn:
We still need to do the pebbleDB implementation. I'll pick it up in the end.

PR #3348, opened 2023-06-21
parachain/disputes: spam slots for disputes-coordinator

kanishkatn:
## Changes

Implements spam slots for disputes-coordinator.

Parity's implementation: https://github.com/paritytech/polkadot/blob/8eb5efcdaec745e16167043c75e2dd4da958cbc8/node/core/dispute-coordinator/src/spam_slots.rs

## Issues
#3302

## Primary Reviewer

@kishansagathiya


dot/parachain/spam_slots.go
```diff
@@ -0,0 +1,200 @@
...
+
+	maxSpamVotes uint32
+}
+
+// byValidatorIndex is a comparator for ValidatorIndex
+func byValidatorIndex(a, b interface{}) int {
+	return int(a.(parachain.ValidatorIndex) - b.(parachain.ValidatorIndex))
+}
+
+func slotsKey(session parachain.SessionIndex, validator parachain.ValidatorIndex) slotKey {
```
EclesioMeloJunior, line 53: would suggest to use `new` to create instances of a struct, same for `unconfirmedDisputesKey`
```suggestion
func newSlotKey(session parachain.SessionIndex, validator parachain.ValidatorIndex) slotKey {
```

dot/parachain/spam_slots.go
```diff
@@ -0,0 +1,200 @@
...
+		s.incrementSpamCount(session, validator)
+	}
+
+	s.unconfirmed[unconfirmedDisputesKey(session, candidate)] = validators
+
+	return true
+}
+
+func (s *spamSlots) Clear(session parachain.SessionIndex, candidate common.Hash) {
+	if validators, ok := s.getValidators(session, candidate); ok {
```
EclesioMeloJunior, line 123: I would suggest to use an early return avoiding nested logic
```suggestion
     validators, ok := s.getValidators(session, candidate)
     if !ok { return }
```

dot/parachain/spam_slots.go
```diff
@@ -0,0 +1,200 @@
...
+		validatorSet := validators.Values()
+		s.clearValidators(session, candidate)
+
+		for _, validator := range validatorSet {
+			spamCount, ok := s.getSpamCount(session, validator.(parachain.ValidatorIndex))
```
EclesioMeloJunior, line 128: `getSpamCount` could return just the value 

```go
for _, validator := range validatorSet {
    spamCount := s.getSpamCount(session, validator.(parachain.ValidatorIndex))
    switch {
    case spamCount < 1:
    case spamCount == 1:
        s.clearSlots(session, validator.(parachain.ValidatorIndex))
    default:
        s.slotsLock.Lock()
        s.slots[slotsKey(session, validator.(parachain.ValidatorIndex))] = spamCount - 1
        s.slotsLock.Unlock()
    }
}
```

dot/parachain/spam_slots.go
```diff
@@ -0,0 +1,200 @@
...
+				s.slots[slotsKey(session, validator.(parachain.ValidatorIndex))] = spamCount - 1
+				s.slotsLock.Unlock()
+			}
+		}
+	}
+}
+
+func (s *spamSlots) PruneOld(oldestIndex parachain.SessionIndex) {
+	s.unconfirmedLock.Lock()
+	unconfirmedToDelete := make([]unconfirmedKey, 0)
```
EclesioMeloJunior, line 145: Since the max items `unconfirmedToDelete` can append is all items in the `s.unconfirmed` I would to suggest to pass it as the capacity of the slice
```suggestion
	unconfirmedToDelete := make([]unconfirmedKey, 0, len(s.unconfirmed))
```

dot/parachain/spam_slots.go
```diff
@@ -0,0 +1,200 @@
...
+	for k := range s.unconfirmed {
```
EclesioMeloJunior, line 146: instead of allocating a helper slice to delete items from a map you can use https://pkg.go.dev/golang.org/x/exp/maps#DeleteFunc, same for `s.slots` prunning

```go
maps.DeleteFunc(s.unconfirmed, func(k unconfirmedKey, _ *treeset.Set) bool {
    if k.session < oldestIndex {return true} else { return false }
})
```

dot/parachain/spam_slots.go
```diff
@@ -0,0 +1,200 @@
...
+	}
+
+	for _, k := range slotsToDelete {
+		delete(s.slots, k)
+	}
+	s.slotsLock.Unlock()
+}
+
+// NewSpamSlots returns a new SpamSlots instance
+func NewSpamSlots(maxSpamVotes uint32) SpamSlots {
```
EclesioMeloJunior, line 172: I would suggest returning the concrete type instead of the interface, and since Golang supports implicit interface you can still return a private struct that implements the interface methods and let the caller to receive the interface 

```go
// dot/parachain pkg
func NewSpamSlots(maxSpamVotes uint32) *spamSlots { ... }

// dot/sync pkg
func startSync() {
    sync(parachain.NewSpamSlots(10)) /* <- return the concrete type which implements the expected interface*/)
}
func sync(spamSlots parachain.SpamSlots /* <- interface */) {}

```

EclesioMeloJunior (changes requested):

dot/parachain/spam_slots.go
kanishkatn: Hmm, the reason why I return an interface is to make sure that the struct implements all the interface methods. If it doesn't, there will be a compile time error at `NewSpamSlots()`. 

If we return just the struct, I guess it will throw an error at `sync(parachain.NewSpamSlots(10))`. I'd rather handle this at the calling package rather than at the caller.

Is there like a convention that says we shouldn't return an interface?

dot/parachain/spam_slots.go
kanishkatn: This is nice, thank you!

dot/parachain/spam_slots.go
kanishkatn: I've changed this to use `maps.DeleteFunc` instead.

dot/parachain/spam_slots.go
kanishkatn: The reason was to identify if there are no spam slots for a set of session and validator index. 

In the sample code you shared, if there is no spam slot for a set of session and validator-index, it would set the spamCount to -1 since the default return would be 0. 

We could prolly add a case to check if it is 0 and do nothing, but we won't know if that key exists or not. 

I feel keeping it as is makes the `getSpamCount()` generic ( in case we wanna know that in the future ) and keeps the code clean and readable.

dot/parachain/spam_slots.go
kanishkatn: That's a good one, I've made the changes!

dot/parachain/spam_slots.go
EclesioMeloJunior: @kanishkatn in the sample code there is the case `case spamCount < 1` so if the `spamCount` is 0 that case take place and since there is nothing there then it will just jump to the next iteration, it only subtract one if the `spamCount` is greater than 1.

If you want to keep it as it is I would suggest an early continue
```go
spamCount, ok := s.getSpamCount(session, validator.(parachain.ValidatorIndex))
if !ok { continue }
...
```

kanishkatn: Yes, you're right and I'm blind, haha.

I'll keep it as is and do an early return. Thank you!

dot/parachain/spam_slots.go
EclesioMeloJunior: > Is there like a convention that says we shouldn't return an interface?

Mostly what I read is `accept interfaces return structs` but this is not a rule, the point is the consumer should be free to receive what interface it wants, for example, every time you want to introduce a new behavior for a certain struct you will need to update the interface too and all the consumers will be affected, 

Refs:
- https://mycodesmells.com/post/accept-interfaces-return-struct-in-go
- https://bryanftan.medium.com/accept-interfaces-return-structs-in-go-d4cab29a301b

> Hmm, the reason why I return an interface is to make sure that the struct implements all the interface methods. If it doesn't, there will be a compile-time error at NewSpamSlots().

In some parts of the code, you will see the following pattern to ensure the implementation at comptime

```go
// ensure at compile time the struct implements the interface
var _ Message = &RequestBlockMessage{} 

type Message interface {
    Encode() ([]byte, error)
}

type RequestBlockMessage struct {...}
func (r *RequestBlockMessage) Encode() ([]byte, error) {...}
```

> If we return just the struct, I guess it will throw an error at sync(parachain.NewSpamSlots(10)). I'd rather handle this at the calling package than at the caller.

No, there is no error. We expect a public interface as an argument and the returned struct to implement those methods

kanishkatn: > No, there is no error. We expect a public interface as an argument and the returned struct to implement those methods

I meant if the struct doesn't implement all the methods of an interface, it would throw an error at `sync(parachain.NewSpamSlots(10))`

There is always something new to learn everyday! Thank you for sharing @EclesioMeloJunior

dot/parachain/spam_slots.go
```diff
@@ -0,0 +1,191 @@
...
+func NewSpamSlotsFromState(unconfirmedDisputes map[unconfirmedKey]*treeset.Set, maxSpamVotes uint32) SpamSlots {
+	slots := make(map[slotKey]uint32)
+
+	for k, v := range unconfirmedDisputes {
+		for validator := range v.Values() {
+			// increment the spam count for this validator and session
+			key := newSlotsKey(k.session, parachain.ValidatorIndex(validator))
+			slots[key]++
+			if slots[key] > maxSpamVotes {
+				// TODO: log this
```
EclesioMeloJunior, line 181: this TODO will be addressed in this PR?

kanishkatn: I'm hoping to. I need to look into the loggers, dunno how it is configured in the project yet.

kanishkatn: Looks like I am going to skip this for now. This is just going into my feature branch, so I believe it's okay for now. Thank you for the review @EclesioMeloJunior !

dot/parachain/spam_slots.go
```diff
@@ -0,0 +1,202 @@
...
+	candidate common.Hash,
+	validator parachain.ValidatorIndex) bool {
+	if spamCount, _ := s.getSpamCount(session, validator); spamCount >= s.maxSpamVotes {
+		return false
+	}
+
+	s.unconfirmedLock.Lock()
+	defer s.unconfirmedLock.Unlock()
+
+	validators, ok := s.unconfirmed[newUnconfirmedDisputesKey(session, candidate)]
```
EclesioMeloJunior, line 112: I see that you're calling the method twice, you could use a variable and assign the return to it and use the variable in both places
```suggestion
        unconfirmerdDisputesKey := newUnconfirmedDisputesKey(session, candidate)
	validators, ok := s.unconfirmed[unconfirmerdDisputesKey]
	...
	s.unconfirmed[unconfirmerdDisputesKey] = validators
```

EclesioMeloJunior (changes requested):
LGTM!! 🎉 , few comments

dot/parachain/spam_slots.go
kanishkatn: Well, that's not good

dimartiro (approved):

kishansagathiya (commented):
A bunch of types are moved from `lib/parachain` to `lib/parachain/types` which causes `make lint` to fail

```
kishan@kishan-Inspiron-14-7425-2-in-1:~/code/gossamer$ make lint
go install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.53
golangci-lint run
dot/parachain/spam_slots.go:1: : # github.com/ChainSafe/gossamer/dot/parachain [github.com/ChainSafe/gossamer/dot/parachain.test]
dot/parachain/spam_slots.go:21:35: undefined: parachain.SessionIndex
dot/parachain/spam_slots.go:21:92: undefined: parachain.ValidatorIndex
dot/parachain/spam_slots.go:25:26: undefined: parachain.SessionIndex
dot/parachain/spam_slots.go:27:33: undefined: parachain.SessionIndex
dot/parachain/spam_slots.go:31:22: undefined: parachain.SessionIndex
dot/parachain/spam_slots.go:32:22: undefined: parachain.ValidatorIndex
dot/parachain/spam_slots.go:36:22: undefined: parachain.SessionIndex
dot/parachain/spam_slots.go:56:36: undefined: parachain.SessionIndex
dot/parachain/spam_slots.go:56:70: undefined: parachain.ValidatorIndex
dot/parachain/spam_slots.go:63:50: undefined: parachain.SessionIndex
dot/parachain/spam_slots.go:63:50: too many errors (typecheck)
package parachain
make: *** [Makefile:24: lint] Error 1
kishan@kishan-Inspiron-14-7425-2-in-1:~/code/gossamer$ 

```

kishansagathiya (commented):
since there is a lot of stuff in dispute subsystem, lets create `lib/parachain/disputes` and move these files there

kanishkatn:
> since there is a lot of stuff in dispute subsystem, lets create `lib/parachain/disputes` and move these files there

That's interesting. Why lib and not dot?

Issue #3311, opened 2023-06-12
feat(consensus/grandpa): implement SharedAuthoritySet for grandpa


jimjbrettj:
## Issue summary

- We have all of the AuthoritySet implemented, but need to implement the threadsafe SharedAuthoritySet wrapper around the AuthoritySet implementation 

## Other information and links

- https://github.com/paritytech/substrate/blob/85415fb3a452dba12ff564e6b093048eed4c5aad/client/consensus/grandpa/src/authorities.rs#L71

jimjbrettj:
Resolved by #3283

PR #3351, opened 2023-06-21
client(consensus/grandpa): implement shared data structure

jimjbrettj:
## Changes

- Create a generic shared data structure and use it for SharedDataSet
- Write tests for the data structure

Explanation

This is my take on substrates [SharedData](https://github.com/paritytech/substrate/blob/34134935dd894d8322c1de72e95cc5d37780ff32/client/consensus/common/src/shared_data.rs#L177). In Rust the [Parking Lot crate](https://amanieu.github.io/parking_lot/parking_lot/index.html), used by substrate for mutex and condvar implementations, does not implement 'Send' which means that it cannot be transferred across thread boundaries. 

Substrate created a workaround solution to allow for threadsafe access to shared data which respects rusts memory model. This is demonstrated by the interface used by shared data, see [shared_data()](https://github.com/paritytech/substrate/blob/34134935dd894d8322c1de72e95cc5d37780ff32/client/consensus/common/src/shared_data.rs#L205) and [shared_data_locked()](https://github.com/paritytech/substrate/blob/34134935dd894d8322c1de72e95cc5d37780ff32/client/consensus/common/src/shared_data.rs#L225) where these functions handle the transfer of ownership of the shared data between threads. Since golang uses a different memory model, I do not believe that we need to do this.

Rather than worrying about the transferring of ownership of the data like substrate does, I only worry about proper access to the data. So instead of copying the interface substrate uses for SharedData, I made a simple interface for threadsafe Reading and Writing of data to shared state, as well as the option to acquire/release locks to support batch operations and other possible use cases.

Note: I am actively hoping for feedback on if this makes sense and how to improve it, so any feedback is greatly appreciated

## Tests

```sh
go test ./client/consensus/grandpa -v
```

## Issues

#3311

## Primary Reviewer

@timwu20


client/consensus/grandpa/sharedAuthoritySet.go
```diff
@@ -0,0 +1,48 @@
...
+// SPDX-License-Identifier: LGPL-3.0-only
+
+package grandpa
+
+import (
+	"sync"
+)
+
+// SharedData Generic shared data structure
+type SharedData[T any] struct {
```
EclesioMeloJunior, line 11: I don't think we need to create another struct to safely read/write data, we can keep it simple by having the mutex as a property in the `SharedAuthoritySet`.

``` go
// SharedAuthoritySet A shared authority set
type SharedAuthoritySet struct {
        inner AuthoritySet
        mtx sync.Mutex <- mutex as a property
	// inner SharedData[AuthoritySet]
}
```

EclesioMeloJunior (changes requested):

client/consensus/grandpa/sharedAuthoritySet.go
jimjbrettj: Oh yeah we do not need it for safety, since substrate uses the SharedData type in other parts of the code I initially thought that it would be good to replicate, but given how much less overhead we have for this I think you are right on this.

Will ping @timwu20 just to make sure this approach is okay before removing the type

edwardmack (approved):

client/consensus/grandpa/sharedAuthoritySet.go
timwu20: Yea I don't think we need this.  We can just use a mutex in the `AuthoritySet` itself and make it thread safe.  Or we don't need the generic type of `SharedData`.

jimjbrettj: I updated the code to remove the SharedData Ttype, and replaced it with what @EclesioMeloJunior recommended.

EclesioMeloJunior (approved):

timwu20:
We should determine the receiver the methods of this type, and implement the locking.

client/consensus/grandpa/authorities.go
```diff
@@ -76,6 +82,109 @@ type appliedChanges[H comparable, N constraints.Unsigned] struct {
 	set    AuthoritySet[H, N]
 }
 
+// InvalidAuthorityList authority sets must be non-empty and all weights must be greater than 0
+func (authSet *SharedAuthoritySet) InvalidAuthorityList(authorities AuthorityList) bool {
```
timwu20, line 86: this doesn't build, since you need to bring in the generic params.

timwu20: nit:
```suggestion
func (sas *SharedAuthoritySet) InvalidAuthorityList(authorities AuthorityList) bool {
```

timwu20 (changes requested):

timwu20 (approved):

PR #3352, opened 2023-06-22
feat(lib/runtime): Update default `runtime.Instance` to `wazero`

### LINKED ISSUE #2585: Implement our runtime with github.com/tetratelabs/wazero

timwu20:
## Changes

- Updates default `runtime.Instance` implementation to the one backed by `wazero`.
- Add an import test that was missed on previous PR
- Fix `codeHash` panic
- Updated some tests, and migrated some test helpers to main package

## Tests

```sh
```

## Issues

closes #2585 

## Primary Reviewer

@EclesioMeloJunior


lib/runtime/wazero/test_helpers.go
```diff
@@ -0,0 +1,82 @@
+// Copyright 2023 ChainSafe Systems (ON)
```
EclesioMeloJunior, line 1: can we rename this file to `helpers_test.go` then they will be included only in the test build

EclesioMeloJunior (changes requested):
LGMT, just a small change

dot/rpc/modules/system_integration_test.go
```diff
@@ -354,7 +354,7 @@ func setupSystemModule(t *testing.T) *SystemModule {
 func newCoreService(t *testing.T, srvc *state.Service) *core.Service {
 	// setup service
 	tt := trie.NewEmptyTrie()
-	rt := wasmer.NewTestInstanceWithTrie(t, runtime.WESTEND_RUNTIME_v0929, tt)
+	rt := wazero_runtime.NewTestInstanceWithTrie(t, runtime.WESTEND_RUNTIME_v0929, tt)
```
jimjbrettj, line 357: why wazero_runtime instead of just wazero?

jimjbrettj (approved):
lgtm!

dot/rpc/modules/system_integration_test.go
timwu20: it's cause the package name is `wazero_runtime` so gopls just ends up using that.

lib/runtime/wazero/test_helpers.go
timwu20: I can't because the functions used in this file are littered throughout the codebase.  We'd have to move these functions into a separate runtime helpers package, but I still can't underscore it with test, since it's a separate package.

jimjbrettj (approved):
lgtm

dot/core/service_test.go
```diff
@@ -251,9 +251,7 @@ func Test_Service_handleCodeSubstitution(t *testing.T) {
 				}
 			},
 			blockHash:  common.Hash{0x01},
-			errWrapped: wasmer.ErrWASMDecompress,
```
EclesioMeloJunior, line 254: you don't need to remove this err, you can change to `io.ErrUnexpectedEOF`

dot/core/service_test.go
```diff
@@ -325,9 +323,11 @@ func Test_Service_handleCodeSubstitution(t *testing.T) {
 			service := testCase.serviceBuilder(ctrl)
 
 			err := service.handleCodeSubstitution(testCase.blockHash, testCase.trieState)
-			assert.ErrorIs(t, err, testCase.errWrapped)
 			if testCase.errWrapped != nil {
-				assert.EqualError(t, err, testCase.errMessage)
+				assert.ErrorIs(t, err, testCase.errWrapped)
```
EclesioMeloJunior, line 327: you can revert to `assert.ErrorIs(t, err, testCase.errWrapped)` without the if guard

EclesioMeloJunior (changes requested):

EclesioMeloJunior (approved):

PR #3363, opened 2023-06-29
fix(dot): use tempDir in tests as base path to avoid creating `dot/~`

kishansagathiya:
## Changes

When you run integration tests in for `dot` say `go test -tags integration --run github.com/ChainSafe/gossamer/dot/... -v`, a folder named `dot/~` gets created.

We shouldn't create a folder named `~` and use home.

## Tests

```sh
```

## Primary Reviewer

@EclesioMeloJunior


EclesioMeloJunior (approved):

jimjbrettj (commented):
Maybe I am not running the correct thing, but I don't seem to replicate this behavior. Where is the "dot" prefix coming from? I don't see it in the changes. Perhaps I'm missing something, but for me $HOME is the same as ~ so this PR doesn't appear to change much except for where we define the basepath

dimartiro:
> Maybe I am not running the correct thing, but I don't seem to replicate this behavior. Where is the "dot" prefix coming from? I don't see it in the changes. Perhaps I'm missing something, but for me $HOME is the same as ~ so this PR doesn't appear to change much except for where we define the basepath

Same here 
`~` actually means "home" in linux like systems (try `cd ~` for example) so I don't think if we really need this change

kishansagathiya:
> Maybe I am not running the correct thing, but I don't seem to replicate this behavior. Where is the "dot" prefix coming from? I don't see it in the changes. Perhaps I'm missing something, but for me $HOME is the same as ~ so this PR doesn't appear to change much except for where we define the basepath

@jimjbrettj @dimartiro 
Can none of you see a filder named `~` created in the `dot` folder on running the command below?

```
go clean -testcache
go test -tags integration github.com/ChainSafe/gossamer/dot
```

dimartiro:
@kishansagathiya you are right! I misunderstood the folder location.
Yes! I'm having the same issue, thanks for clarify

I ran the test and this folder was created.

image

Thanks for taking the issue and sorry for the misunderstood 🙏

dimartiro (approved):

kanishkatn (changes requested):
Correct me if I'm wrong, this is already being handled at https://github.com/ChainSafe/gossamer/blob/development/cmd/gossamer/commands/utils.go#L308. Not all integration tests create the folder in the project path.

I feel we need to identify which test is creating this and look into resolving it instead of changing the default config. Also, if there is an issue with the `base-path` resolution, it could affect the normal run as well since it accepts the same format.

kishansagathiya:
From what I understand `~` is a shell concept. Shell will understand the meaning of `~`, but go won't. https://go.dev/play/p/3Az7NSMMPFk

Sure, we can use `ExpandDir`, but that should happen first time that path is getting set. The approach of finding where a particular path is being used and using ExpandDir there, seems wrong to me. That way will have to use ExpandDir for all paths or trace where a path with `~` could be used and use ExpandDir there.

We could have something like this. 
```
defaultBasePath = "~/.gossamer/polkadot"
config.BasePath = ExpandDir(defaultBasePath)
```
But even in this `defaultBasePath` is an inappropriate path, since we would never want anyone to use it directly.

My preference is the code i submitted. The perfect use of `ExpandDir` would be to expand a user submitted path, like `--path=~/.gossamer`

kishansagathiya:
```
defaultBasePath = "~/.gossamer/polkadot"
config.BasePath = ExpandDir(defaultBasePath)
```
What's wrong with this ?


currently we have something like, 
```
defaultBasePath = "~/.gossamer/polkadot"
config.BasePath = defaultBasePath
```
base path with `~` in it is just wrong. Expanding it later, it fixing the mistake. 

> I feel we need to identify which test is creating this and look into resolving it

That would be a hack. The primary problem is that our base path in our default config is not usable.

kishansagathiya:
I don't think you are trying to understand what I am arguing. 

My main point is that there is no reason to have a path in our code base with `~` in it. And so far, you have not given me any reasoning  to keep it.

> he base-path parsing is already being handled in parseBasePath()

You don't need to parse `config.BasePath`, if you set the right path.

I get it that there is `parseBasePath` and bunch of other functions. But those functions are to re-populate config with with say flags, env, etc. It doesn't explain why we need to have `defaultBasePath = "~/.gossamer/kusama"`

> It's a feature, not a bug and it has been inherited from the previous cli. The feature being the ability to resolve the relative paths.

Have ability to use relative paths. But don't put a path with `~` if we can do without it.  I am arguing about removing `~` from `defaultBasePath = "~/.gossamer/kusama"`. That `~` has no feature.

kishansagathiya:
We have had a call. @kanishkatn pointed out that we shouldn't be using home directory anyway for tests and should be using temp dir. 

I am going to edit all the tests that use DefaultConfig() to use TempDir, instead of using the default base path.

kishansagathiya:
@EclesioMeloJunior @dimartiro Have changed this PR quite a bit. Feel free to re-review this.

EclesioMeloJunior (approved):

dimartiro (approved):

kanishkatn (approved):
I thought I approved it, my bad.

Issue #3317, opened 2023-06-13
Fix `types.GetSlotFromHeader` when finalization stalls


EclesioMeloJunior:
The following pprof output was extracted when Gossamer was in that region

![image](https://github.com/ChainSafe/gossamer/assets/17255488/85fea9e3-4b9c-4f0b-928e-f82780acdd70)

It is possible to notice that the methods `handleReadyBlock`, `handleWorkersResults`, `processBlockData` and `processBlockDataWithHeaderAndBody` are taking 0s in `flat` and 19.32s in `cum` which indicates that a function they call is hanging 99% of the time and we know that the method `VerifyBlock` is part of the sync flow

If we go deeper in the `processBlockDataWithHeaderAndBody` we can see that the `VerifyBlock` is taking 18.82s 
![image](https://github.com/ChainSafe/gossamer/assets/17255488/fcc57ec9-b8f2-4d22-b4d5-465f23552798)

To be brief the call chain is: 
```

VerifyBlock -> verifyAuthorshipRight -> verifyBlockEquivocation -> b.blockState.GetBlockHashesBySlot -> GetSlotForBlock -> types.GetSlotFromHeader
```

and in the end, the function `GetSlotFromHeader` calls the function `DecodeBabePreDigest` which uses the `scale.Unmarshal` to decode the varying data type and as we can notice it is quite expensive.

![image](https://github.com/ChainSafe/gossamer/assets/17255488/a5181f93-f36c-40cf-a345-553c270f7208)

This process (`DecodeBabePreDigest`) is made by every descendant of the highest finalized block, and we know that at this point 1.41M in the Westend chain the finalization stopped for some period of time.

Basically, each time we want to import a new block we need to verify it, and since we don't have any finalization happening the number of blocks since the latest finalized block only increases and this creates the bottleneck

EclesioMeloJunior:
When the finalization is working the method `VerifyBlocks` takes only 3.94s 
![image](https://github.com/ChainSafe/gossamer/assets/17255488/d11c0188-30b9-4bd1-b213-277c1bb9cdac)

timwu20:
Really good work @EclesioMeloJunior! Showing off those pprof chops 💪

PR #3364, opened 2023-06-29
fix: cache slot to header data while checking BABE equivocation

EclesioMeloJunior:
## Changes

- Refactor `verifyBlockEquivocation` by using a slot state, similar to how substrate works, https://github.com/paritytech/substrate/blob/c14ff62f0a2bca89f02f8e1cb7c1347f1d2a0baa/client/consensus/babe/src/lib.rs#L1221

## Tests

```sh
go test -timeout 10m -run ^Test_checkEquivocation$ github.com/ChainSafe/gossamer/dot/state --tags=integration -v
```

## Issues

- Closes https://github.com/ChainSafe/gossamer/issues/3317

## Primary Reviewer

@kishansagathiya


dot/state/service.go
```diff
@@ -34,6 +34,7 @@ type Service struct {
 	Transaction *TransactionState
 	Epoch       *EpochState
 	Grandpa     *GrandpaState
+	Slot        *SlotState
```
jimjbrettj, line 37: Should slot state be a part of babe state?

dot/state/slot_test.go
```diff
@@ -0,0 +1,134 @@
...
+	header3 := createHeader(t, 2) // @ slot 4
+	header4 := createHeader(t, 3) // @ slot MAX_SLOT_CAPACITY + 4
+	header5 := createHeader(t, 4) // @ slot MAX_SLOT_CAPACITY + 4
+	header6 := createHeader(t, 3) // @ slot 4
+
+	slotState := NewSlotState(inMemoryDB)
+
+	// It's ok to sign same headers.
+	equivProf, err := slotState.CheckEquivocation(2, 2, header1, aliceAuthorityID)
+	require.Nil(t, err)
```
jimjbrettj, line 84: nit: require.NoError() here and other usages

jimjbrettj (commented):
Looks great, just a few comments

dot/state/service.go
EclesioMeloJunior: the slot state is a dependency for the BABE verifier, which is part of the BABE service but I don't think it should be part of the BABE state as in substrate they have a module only for slots (`client/consensus/slots`)

dimartiro (approved):
Great work Eclésio! 👏  LGTM

jimjbrettj (approved):
great work!

PR #3382, opened 2023-07-12
feat(lib/parachain): Create and integrate Validation Protocol struct

### LINKED ISSUE #3318: Create a go struct for ValidationProtocol
## Issue summary

Translate rust struct for ValidationProtocol into go. 
Make sure that our go struct is same as rust using scale encoding.

- Make sure that both struct produces the same scale encoding.
- Make sure that scale encoding of go struct for `ValidationProtocol` can be decoded into rust struct for `ValidationProtocol` and they have the same internal properties
- Make sure that scale encodoing of rust struct for `ValidationProtocol` can be decoded into go struct for `ValidationProtocol` and they have the same internal properties

- ValidationProtocol contains three types of messages
  - Statement Distribution Message (https://github.com/ChainSafe/gossamer/issues/3289)
  - Approval Distribution Message (https://github.com/ChainSafe/gossamer/issues/3288)
  - Bitfield Distribution Message 

- This issue should be picked up after https://github.com/ChainSafe/gossamer/issues/3289 and https://github.com/ChainSafe/gossamer/issues/3288 as it requires them.

Code for ValidationProtocol here 
- https://github.com/paritytech/polkadot/blob/b6b74fdf546c14bf9ac93c03916abda9ee33a52f/node/network/protocol/src/lib.rs#L520
- https://paritytech.github.io/polkadot/book/types/network.html#validation-v1

edwardmack:
## Changes
- Create struct for ValidationProtocol based on rust code: https://github.com/paritytech/polkadot/blob/b6b74fdf546c14bf9ac93c03916abda9ee33a52f/node/network/protocol/src/lib.rs#L520
- Create tests to confirm scale encoding/decoding of ValidationProtocol sturct
- Replace empty ValidationProtocol struct with this struct in service.

## Tests

```sh
go test github.com/ChainSafe/gossamer/lib/parachain -v -run ^TestMarshalUnMarshalValidationProtocol$
```

## Issues
Closes #3318

## Primary Reviewer

@kishansagathiya 
@timwu20


lib/parachain/validation_protocol_test.go
```diff
@@ -4,11 +4,314 @@
...
+	if err != nil {
+		fmt.Printf("Error unmarshaling test data: %s\n", err)
+		return
+	}
+}
+
+func TestMarshalUnMarshalValidationProtocol(t *testing.T) {
+	t.Parallel()
+	/* ValidationProtocol with ApprovalDistribution with Assignments Rust code
+	fn try_msg_assignments_encode() {
```
dimartiro, line 35: I don't think if it is useful to have rust code here as explanation since we already have our own test in go and it's working ok. 
Can we add the permalink from github instead putting the code?

dimartiro (approved):
Approved with a little comment / opinion 
Thanks!

lib/parachain/validation_protocol_test.go
edwardmack: I'd like to keep the example rust code here so that reading can confirm how the encoding is generated.  I haven't committed the rust code back to repo, so I don't think I can create a github permalink.

dimartiro: Oh my bad, I thought it was from substrate 😅

lib/parachain/validation_protocol.go
```diff
@@ -8,23 +8,199 @@ import (
...
+	err = vdt.Set(val)
+	if err != nil {
+		return fmt.Errorf("setting value to varying data type: %w", err)
+	}
+
+	*ad = ApprovalDistribution(vdt)
+	return nil
+}
+
+type StatementDistribution StatementDistributionMessage
```
kishansagathiya, line 127: `StatementDistributionMessage` already has `New`, `Set`, `Value` and `NewStatementDistributionMessage`.

We can just add `Index` method for `StatementDistribution`, I think that should be enough

lib/parachain/validation_protocol.go
```diff
@@ -8,23 +8,199 @@ import (
...
+func (ad *ApprovalDistribution) Set(val scale.VaryingDataTypeValue) (err error) {
+	vdt := scale.VaryingDataType(*ad)
+	err = vdt.Set(val)
+	if err != nil {
+		return fmt.Errorf("setting value to varying data type: %w", err)
+	}
+
+	*ad = ApprovalDistribution(vdt)
+	return nil
+}
```
kishansagathiya, line 125: Can you check if just adding `Index` method here is sufficient? These creates a lot of repetitive methods.

lib/parachain/validation_protocol.go
```diff
@@ -8,23 +8,199 @@ import (
...
+func (bd *BitfieldDistribution) Set(val scale.VaryingDataTypeValue) (err error) {
+	vdt := scale.VaryingDataType(*bd)
+	err = vdt.Set(val)
+	if err != nil {
+		return fmt.Errorf("setting value to varying data type: %w", err)
+	}
+
+	*bd = BitfieldDistribution(vdt)
+	return nil
+}
```
kishansagathiya, line 93: Can you check if just adding `Index` method here is sufficient? These creates a lot of repetitive methods.

lib/parachain/validation_protocol.go
```diff
@@ -34,7 +210,7 @@ func (vp *ValidationProtocolV1) Hash() (common.Hash, error) {
 }
 
 // Encode a collator protocol message using scale encode
-func (vp *ValidationProtocolV1) Encode() ([]byte, error) {
+func (vp *ValidationProtocol) Encode() ([]byte, error) {
```
kishansagathiya, line 213: I think this would work even without making ValidationProtocol pointer

lib/parachain/validation_protocol.go
```diff
@@ -8,23 +8,199 @@ import (
...
 
 // Type returns ValidationMsgType
-func (*ValidationProtocolV1) Type() network.MessageType {
+func (ValidationProtocol) Type() network.MessageType {
 	return network.ValidationMsgType
 }
 
 // Hash returns the hash of the CollationProtocolV1
-func (vp *ValidationProtocolV1) Hash() (common.Hash, error) {
+func (vp *ValidationProtocol) Hash() (common.Hash, error) {
```
kishansagathiya, line 203: I think this would work even without making ValidationProtocol pointer

kishansagathiya (changes requested):

lib/parachain/service.go
```diff
@@ -123,7 +126,22 @@ func (s Service) run() {
 	collationMessage := CollationProtocolV1{}
 	s.Network.GossipMessage(&collationMessage)
 
-	validationMessage := ValidationProtocolV1{}
+	statementDistributionLargeStatement := NewStatementDistributionVDT()
+	err := statementDistributionLargeStatement.Set(SecondedStatementWithLargePayload{
+		RelayParent:   common.Hash{},
+		CandidateHash: CandidateHash{Value: common.Hash{}},
+		SignedBy:      5,
+		Signature:     ValidatorSignature{},
```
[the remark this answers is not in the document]
jimjbrettj: Is this just for testing right now? A bit confused why we would want to initialize a vdt with arbitrary values

lib/parachain/validation_protocol.go
jimjbrettj: I think our linter gets mad if we have a mix of pointer and non pointer receivers

lib/parachain/validation_protocol.go
```diff
@@ -8,23 +8,199 @@ import (
...
 )
 
 const MaxValidationMessageSize uint64 = 100 * 1024
 
-type ValidationProtocolV1 struct {
-	//	TODO: Implement this struct https://github.com/ChainSafe/gossamer/issues/3318
+type UncheckedSignedAvailabilityBitfield struct {
+	// The payload is part of the signed data. The rest is the signing context,
+	// which is known both at signing and at validation.
+	Payload scale.BitVec `scale:"1"`
```
jimjbrettj, line 21: just making sure, no 0 index here?

lib/runtime/wasmer/testdata/parachain.yaml
```diff
@@ -23,7 +23,7 @@ storage:
 
 # paras.currentCodeHash
   - key: "0xcd710b30bd2eab0352ddcc26417aa194e2d1c22ba0a888147714a3487bd51c63b6ff6f7d467b87a9e8030000"
-    value: "0xcafdc44448ed62f5f67e5c340b420c51fe3bee7c1577f9fa0e819383b5145337" 
+    value: "0xcafdc44448ed62f5f67e5c340b420c51fe3bee7c1577f9fa0e819383b5145337"
```
jimjbrettj, line 26: not: remove diff

lib/parachain/service.go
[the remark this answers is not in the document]
kishansagathiya: this is just for testing (to demonstrate that it works)

lib/runtime/wasmer/testdata/parachain.yaml
kishansagathiya: this is okay, it removes empty space

lib/parachain/validation_protocol.go
kishansagathiya: This is okay. It starts with 1. Search `scale:"0"` in the codebase, you won't find any entries.

lib/parachain/validation_protocol.go
kishansagathiya: > I think our linter gets mad if we have a mix of pointer and non pointer receivers

Yeah, that's true. I think I might have meant to change from pointer to non-pointer in all it's methods. But if linter gets mad, no need to address this comment. Anyway this comment ranges from super nitpicky to personal preference. :laughing:

lib/parachain/service.go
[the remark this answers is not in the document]
edwardmack: Added some values.

lib/parachain/validation_protocol.go
```diff
@@ -8,23 +8,199 @@ import (
...
+
+	// The index of the validator signing this statement.
+	ValidatorIndex parachaintypes.ValidatorIndex `scale:"2"`
+
+	// The signature by the validator of the signed payload.
+	Signature ValidatorSignature `scale:"3"`
+}
+
+type Bitfield struct {
```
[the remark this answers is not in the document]
edwardmack: added comment

lib/parachain/validation_protocol.go
edwardmack: Removed unneeded functions.

lib/parachain/validation_protocol.go
edwardmack: Removed unneeded functions.

lib/parachain/validation_protocol.go
edwardmack: Removed unneeded functions.

dot/parachain/validation_protocol.go
```diff
@@ -7,24 +7,143 @@ import (
...
+	if err != nil {
+		panic(err)
+	}
+	return ValidationProtocol(vdt)
+}
+
+// New returns new ApprovalDistributionMessage VDT
+func (ValidationProtocol) New() ValidationProtocol {
+	return NewValidationProtocolVDT()
+}
```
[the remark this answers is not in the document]
edwardmack: Comment updated.

New is a method if the struct because Varying Data Types need this to decode.

PR #3383, opened 2023-07-12
client(consensus/grandpa): implement grandpaDb and environment types

jimjbrettj:
## Changes

- add an interface to represent grandpa DB interactions. After reviewing their auxDB impl, I don't think we need to implement all of it right now and believe we are ok having simple db interactions. 
- add logic in auxDB that is relevant to our implementation as well as tests
- create all the environment types needed for our db impl and test them
- add ability to decode custom vdts within maps and custom vdts to scale package

## Tests

```sh
```

## Primary Reviewer

@EclesioMeloJunior


client/consensus/grandpa/interfaces.go
```diff
@@ -34,3 +34,18 @@ type PublicKey interface {
...
+// Provides access to an auxiliary database.
+//
+// This is a simple global database not aware of forks. Can be used for storing auxiliary
+// information like total block weight/difficulty for fork resolution purposes as a common use
+// case.
+type AuxStore interface {
+	// InsertAux Insert auxiliary data into key-Value store.
+	//
+	// Deletions occur after insertions.
+	InsertAux(insert map[string][]byte, deleted []string) error
```
EclesioMeloJunior, line 48:
```suggestion
	Insert(insert map[string][]byte, deleted []string) error
```

client/consensus/grandpa/interfaces.go
```diff
@@ -34,3 +34,18 @@ type PublicKey interface {
...
+	// GetAux Query auxiliary data from key-Value store.
+	GetAux(key []byte) *[]byte
```
EclesioMeloJunior, line 50:
```suggestion
	Get(key []byte) *[]byte
```

EclesioMeloJunior (changes requested):
The keys and values are always bytes, I see that you are receiving strings, what I would suggest is to:

```go
type databaseKeyValue [2][]byte

type AuxStore interface {
    Insert(toInsert []databaseKeyValue, toDelete [][]byte)
    Get(key []byte) []byte
}
```

EclesioMeloJunior (changes requested):

client/consensus/grandpa/interfaces.go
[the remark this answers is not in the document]
jimjbrettj: Will investigate! Currently, there is a difference between the 0 and empty value, so currently nil case represents "value not found for key" case. If you have a suggestion of a better way to do this let me know!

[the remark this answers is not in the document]
jimjbrettj: updated

client/consensus/grandpa/auxSchema.go
```diff
@@ -0,0 +1,233 @@
...
+
+type databaseKeyValue [2][]byte
+
+// SharedVoterSetState .
+type SharedVoterSetState[H comparable, N constraints.Unsigned] struct {
+	// TODO move and implement
+	inner VoterSetState[H, N]
+}
+
+type PersistentData[H comparable, N constraints.Unsigned] struct {
```
timwu20, line 28:
```suggestion
type persistentData[H comparable, N constraints.Unsigned] struct {
```

client/consensus/grandpa/auxSchema.go
```diff
@@ -0,0 +1,233 @@
...
+	authoritySet SharedAuthoritySet[H, N] // TODO this needs to be shared auth set
+	setState     SharedVoterSetState[H, N]
+}
+
+func loadPersistent[H comparable, N constraints.Unsigned](store AuxStore, genesisHash H, genesisNumber N, genesisAuthorities AuthorityList) (*PersistentData[H, N], error) {
+	genesis := finalityGrandpa.HashNumber[H, N]{Hash: genesisHash, Number: genesisNumber}
+	makeGenesisRound := finalityGrandpa.NewRoundState[H, N]
+
+	// No longer use version to tell if data is persisted, just use actual data
```
timwu20, line 37: revise this comment.  We never ported over different versions in the first place.

client/consensus/grandpa/interfaces.go
```diff
@@ -26,13 +26,19 @@ type ForkTree[H comparable, N constraints.Unsigned] interface {
...
+// This is a simple global database not aware of forks. Can be used for storing auxiliary
+// information like total block weight/difficulty for fork resolution purposes as a common use
+// case.
+type AuxStore interface {
+	// Insert auxiliary data into key-Value store.
+	//
+	// Deletions occur after insertions.
+	Insert(insert []databaseKeyValue, deleted [][]byte) error
+	// Get Query auxiliary data from key-Value store.
+	Get(key []byte) (*[]byte, error)
```
timwu20, line 43: Why not copy the interface method names?
```suggestion
	InsertAux(insert []databaseKeyValue, deleted [][]byte) error
	// Get Query auxiliary data from key-Value store.
	GetAux(key []byte) (*[]byte, error)
```

client/consensus/grandpa/interfaces.go
```diff
@@ -26,13 +26,19 @@ type ForkTree[H comparable, N constraints.Unsigned] interface {
...
+type Telemetry interface{}
 
-type Telemetry interface {
+// AuxStore is part of the substrate backend.
+// Provides access to an auxiliary database.
+//
+// This is a simple global database not aware of forks. Can be used for storing auxiliary
+// information like total block weight/difficulty for fork resolution purposes as a common use
+// case.
+type AuxStore interface {
```
timwu20, line 37: I don't think this belongs in here.  It's located in `client/api` in substrate, so we can assume this will be used elsewhere.

client/consensus/grandpa/interfaces.go
```diff
@@ -26,13 +26,19 @@ type ForkTree[H comparable, N constraints.Unsigned] interface {
...
+	// Insert auxiliary data into key-Value store.
+	//
+	// Deletions occur after insertions.
+	Insert(insert []databaseKeyValue, deleted [][]byte) error
```
timwu20, line 41: You shouldn't use private types in public interface methods.  It may be easier to read to create a `KeyValue` struct.

client/consensus/grandpa/lib.go
```diff
@@ -17,9 +19,17 @@ type Address string
...
 // AuthorityList A list of Grandpa authorities with associated weights.
 type AuthorityList []Authority
+
+// NewAuthoritySetStruct A new authority set along with the canonical block it changed at.
+type NewAuthoritySetStruct[H comparable, N constraints.Unsigned] struct {
+	CanonNumber N
+	CanonHash   H
+	SetId       N
+	Authorities AuthorityList
+}
```
timwu20, line 35: why do we need these changes?

pkg/scale/decode.go
```diff
@@ -344,7 +344,17 @@ func (ds *decodeState) decodeVaryingDataTypeSlice(dstv reflect.Value) (err error
 
 func (ds *decodeState) decodeCustomVaryingDataType(dstv reflect.Value) (err error) {
 	initialType := dstv.Type()
-	converted := dstv.Convert(reflect.TypeOf(VaryingDataType{}))
+	var converted reflect.Value
```
timwu20, line 347: these changes should be in a different PR if this isn't working.

timwu20: I just checked, you need to rebase.  These changes were added to scale a while ago.

timwu20 (changes requested):

client/consensus/grandpa/interfaces.go
jimjbrettj: I had in a previous instance but got feedback to change it. Can revert back if that is preferred

pkg/scale/decode.go
jimjbrettj: oh beautiful!

client/consensus/grandpa/lib.go
jimjbrettj: It is based on this structure in substrate https://github.com/paritytech/substrate/blob/19971bd3eafa6394d918030f4142f85ea54404c0/client/consensus/grandpa/src/lib.rs#L371

I added "struct" to the end because we already have a function called NewAuthoritySet

client/consensus/grandpa/authorities.go
```diff
@@ -33,11 +33,11 @@ type delayedKinds[N constraints.Unsigned] interface {
...
-	value interface{}
+	Value interface{}
 }
 
 func setDelayKind[N constraints.Unsigned, T delayedKinds[N]](delayKind *delayKind, val T) {
-	delayKind.value = val
+	delayKind.Value = val
 }
 
 func newDelayKind[N constraints.Unsigned, T delayedKinds[N]](val T) delayKind {
```
EclesioMeloJunior, line 43: Ok, I think I left this functions passes in some previous review, but now I'm wondering why we need those generics if the `delayKind` type holds an interface.

I would suggest to change to

```go

type delayKind[N constraints.Unsigned, T delayedKinds[N]] struct {
	Value T
}

func newDelayKind[N constraints.Unsigned, T delayedKinds[N]](val T) delayKind[N, T] {
	return delayKind[N, T]{
		Value: val,
	}
}
```

Then you can remove `setDelayKind`

EclesioMeloJunior (changes requested):

client/consensus/grandpa/authorities.go
EclesioMeloJunior: And then when you want to to switch over a type you can use:

```go
switch any(pending.DelayKind.Value).(type) {
	case Finalized:
		return authSet.addStandardChange(pending, isDescendentOf)
	case Best[N]:
		return authSet.addForcedChange(pending, isDescendentOf)
	default:
		panic("DelayKind is invalid type")
}
```

or even

```go
switch v := any(change.DelayKind.Value).(type) {
...
```

jimjbrettj: So while I like this suggestion, adding another generic constraint to the PendingChange would be a divergence from how substrate does it, so I think I would rather have the constraints match substrates in this case and leave as is.

However, if you still think it would be better to make the change I can.

client/consensus/grandpa/interfaces.go
timwu20: I remember Quentin doing that and a lot of people not liking it.  I think it makes sense to do it contextually, like say if we only need a portion of another package interface.

client/consensus/grandpa/authorities.go
timwu20: I think this conversation makes more sense in this PR #3283.  The code that introduces this type is there.

client/consensus/grandpa/environment.go
```diff
@@ -0,0 +1,466 @@
...
+	}
+
+	if len(compRounds.Rounds) > NumLastCompletedRounds {
+		compRounds.Rounds = compRounds.Rounds[:len(compRounds.Rounds)-1]
+	}
+}
+
+// CurrentRounds A map with voter status information for currently live rounds,
+// which votes have we cast and what are they.
+type CurrentRounds[H comparable, N constraints.Unsigned] map[uint64]HasVoted[H, N]
```
timwu20, line 111: Why not a `btree.Map`?

client/consensus/grandpa/environment.go
```diff
@@ -0,0 +1,466 @@
...
+			Rounds: make([]CompletedRound[H, N], 0, NumLastCompletedRounds),
+		},
+		CurrentRounds: make(map[uint64]HasVoted[H, N]), // init the map
+	}, Paused[H, N]{})
+	if err != nil {
+		panic(err)
+	}
+	vs := VoterSetState[H, N](vdt)
+	return vs
+}
```
timwu20, line 150: I think I'm going to make `scale.VaryingDataType` an interface.  That way it's not so cumbersome to create these enums.

client/consensus/grandpa/environment.go
```diff
@@ -0,0 +1,466 @@
...
+
+// Value will return the value from the underlying VaryingDataType
+func (tve *VoterSetState[H, N]) Value() (val scale.VaryingDataTypeValue, err error) { //skipcq: GO-W1029
+	vdt := scale.VaryingDataType(*tve)
+	return vdt.Value()
+}
+
+// New ...
+func (tve VoterSetState[H, N]) New() VoterSetState[H, N] {
```
timwu20, line 138: comment?

client/consensus/grandpa/environment.go
```diff
@@ -0,0 +1,466 @@
...
+// CurrentRounds A map with voter status information for currently live rounds,
+// which votes have we cast and what are they.
+type CurrentRounds[H comparable, N constraints.Unsigned] map[uint64]HasVoted[H, N]
+
+// VoterSetState The state of the current voter set, whether it is currently active or not
+// and information related to the previously completed rounds. Current round
+// voting status is used when restarting the voter, i.e. it will re-use the
+// previous votes for a given round if appropriate (same round and same local
+// key).
+type VoterSetState[H comparable, N constraints.Unsigned] scale.VaryingDataType
```
timwu20, line 118: don't think this needs to be public.

client/consensus/grandpa/environment.go
```diff
@@ -0,0 +1,466 @@
...
+		if contains {
+			return v.CompletedRounds, v.CurrentRounds, nil
+		}
+		return CompletedRounds[H, N]{}, CurrentRounds[H, N]{}, fmt.Errorf("voter acting on a live round we are not tracking")
+	case Paused[H, N]:
+		return CompletedRounds[H, N]{}, CurrentRounds[H, N]{}, fmt.Errorf("voter acting while in paused state")
+	default:
+		panic("CompletedRounds: invalid voter set state")
+	}
+}
```
timwu20, line 254: These methods should be private.  They are all `pub(crate)`

client/consensus/grandpa/environment.go
```diff
@@ -0,0 +1,466 @@
...
+
+// Live The voter is live, i.e. participating in rounds.
+type Live[H comparable, N constraints.Unsigned] struct {
```
timwu20, line 257: maybe more explicit to what this is a `VaryingDataTypeValue` for.
```suggestion
type voterSetStateLive[H comparable, N constraints.Unsigned] struct {
```

client/consensus/grandpa/environment.go
```diff
@@ -0,0 +1,466 @@
...
+type Paused[H comparable, N constraints.Unsigned] struct {
+	// The previously completed rounds
+	CompletedRounds CompletedRounds[H, N]
+}
+
+// Index returns VDT index
+func (Paused[H, N]) Index() uint { return 1 }
+
+// HasVoted Whether we've voted already during a prior run of the program
+type HasVoted[H comparable, N constraints.Unsigned] scale.VaryingDataType
```
timwu20, line 277: don't think this needs to be public

client/consensus/grandpa/environment.go
```diff
@@ -0,0 +1,466 @@
...
+// New is constructor for HasVoted
+func (hasVoted HasVoted[H, N]) New() HasVoted[H, N] {
+	vdt, _ := scale.NewVaryingDataType(No{}, Yes[H, N]{})
+
+	hv := HasVoted[H, N](vdt)
+	return hv
+}
+
+// No Has not voted already in this round
+type No struct{}
```
timwu20, line 305: don't think this needs to be public.

client/consensus/grandpa/environment.go
```diff
@@ -0,0 +1,466 @@
...
+
+// Index returns VDT index
+func (No) Index() uint { return 0 }
+
+// Yes Has voted in this round
+type Yes[H comparable, N constraints.Unsigned] struct {
```
timwu20, line 311: don't think this needs to be public.

client/consensus/grandpa/environment.go
```diff
@@ -0,0 +1,466 @@
...
+	return hasVoted.Prevote() == nil
+}
+
+// CanPrecommit Returns true if the voter can still precommit, false otherwise
+func (hasVoted *HasVoted[H, N]) CanPrecommit() bool {
+	return hasVoted.Precommit() == nil
+}
+
+// Vote Whether we've voted already during a prior run of the program
+type Vote[H comparable, N constraints.Unsigned] scale.VaryingDataType
```
timwu20, line 412: don't think this needs to be public.

client/consensus/grandpa/environment.go
```diff
@@ -0,0 +1,466 @@
...
+	vdt, err := scale.NewVaryingDataType(Propose[H, N]{}, Prevote[H, N]{}, Precommit[H, N]{})
+	if err != nil {
+		panic(err)
+	}
+	hv := Vote[H, N](vdt)
+	return hv
+}
+
+// Propose Has cast a proposal
+type Propose[H comparable, N constraints.Unsigned] struct {
```
timwu20, line 442: don't think this needs to be public.

client/consensus/grandpa/environment.go
```diff
@@ -0,0 +1,466 @@
...
+	PrimaryPropose grandpa.PrimaryPropose[H, N]
+}
+
+// Index returns VDT index
+func (Propose[H, N]) Index() uint { return 0 }
+
+// Prevote Has cast a prevote
+type Prevote[H comparable, N constraints.Unsigned] struct {
```
timwu20, line 450: don't think this needs to be public.

client/consensus/grandpa/environment.go
```diff
@@ -0,0 +1,466 @@
...
+	PrimaryPropose *grandpa.PrimaryPropose[H, N]
+	Vote           grandpa.Prevote[H, N]
+}
+
+// Index returns VDT index
+func (Prevote[H, N]) Index() uint { return 1 }
+
+// Precommit Has cast a precommit (implies prevote.)
+type Precommit[H comparable, N constraints.Unsigned] struct {
```
timwu20, line 459: don't think this needs to be public.

timwu20 (changes requested):

client/consensus/grandpa/environment.go
```diff
@@ -0,0 +1,466 @@
...
+		panic(err)
+	}
+	tve := VoterSetState[H, N](vdt)
+	return &tve
+}
+
+// Live Create a new live VoterSetState with round 0 as a completed round using
+// the given genesis state and the given authorities. Round 1 is added as a
+// current round (with state `HasVoted::No`).
+func (tve *VoterSetState[H, N]) Live(setId uint64, authSet AuthoritySet[H, N], genesisState grandpa.HashNumber[H, N]) (VoterSetState[H, N], error) {
```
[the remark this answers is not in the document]
jimjbrettj: good call

client/consensus/grandpa/aux_schema.go
```diff
@@ -0,0 +1,238 @@
...
+func loadPersistent[H comparable, N constraints.Unsigned](store AuxStore, genesisHash H, genesisNumber N, genesisAuthorities []Authority) (*persistentData[H, N], error) {
+	genesis := finalityGrandpa.HashNumber[H, N]{Hash: genesisHash, Number: genesisNumber}
+	makeGenesisRound := finalityGrandpa.NewRoundState[H, N]
+
+	set := AuthoritySet[H, N]{
+		PendingStandardChanges: NewChangeTree[H, N](),
+		PendingForcedChanges:   make([]PendingChange[H, N], 0),
+		AuthoritySetChanges:    make(AuthoritySetChanges[N], 0),
+	}
+	setState := *NewVoterSetState[H, N]()
```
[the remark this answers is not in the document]
jimjbrettj: I till need to define the ForkTree type, since it is an interface

client/consensus/grandpa/aux_schema.go
```diff
@@ -0,0 +1,238 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	// If there is no authoritySet data stored, then we are starting from genesis
+	if encodedAuthSet != nil {
+		err = scale.Unmarshal(*encodedAuthSet, &set)
+		if err != nil {
+			return nil, err
+		}
```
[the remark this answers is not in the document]
jimjbrettj: since a pointer to []byte is returned by db.Get(), we cannot unmarshall until we verify that the pointer is not nil

client/consensus/grandpa/aux_schema.go
```diff
@@ -0,0 +1,238 @@
...
+		}, nil
+	}
+}
+
+// UpdateAuthoritySet Update the authority set on disk after a change.
+//
+// If there has just been a handoff, pass a `new_set` parameter that describes the
+// handoff. `set` in all cases should reflect the current authority set, with all
+// changes and handoffs applied.
+func UpdateAuthoritySet[H comparable, N constraints.Unsigned](store AuxStore, set AuthoritySet[H, N], newSet *NewAuthoritySetStruct[H, N]) error {
```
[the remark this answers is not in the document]
jimjbrettj: Again, I just didnt think that this was necessary for out impl. Like the above comment though, if this would actually be beneficial I can add it

client/consensus/grandpa/aux_schema.go
```diff
@@ -0,0 +1,238 @@
...
+func UpdateBestJustification() {
+	// TODO impl when we have justification logic
+	panic("impl")
+}
+
+// BestJustification  Fetch the justification for the latest block finalized by GRANDPA, if any.
+func BestJustification() {
+	// TODO impl when we have justification logic
+	panic("impl")
+}
```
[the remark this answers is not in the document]
jimjbrettj: Yes, because none of the justification logic is implemented and IMO it is enough work to separate into a different PR. This has already been done and addressed in #3454

client/consensus/grandpa/environment.go
```diff
@@ -0,0 +1,466 @@
...
+type CompletedRounds[H comparable, N constraints.Unsigned] struct {
+	Rounds []CompletedRound[H, N]
+	SetId  uint64
+	Voters []ed25519.PublicKey
+}
+
+// NewCompletedRounds Create a new completed rounds tracker with NUM_LAST_COMPLETED_ROUNDS capacity.
+func NewCompletedRounds[H comparable, N constraints.Unsigned](genesis *CompletedRound[H, N], setId uint64, voters AuthoritySet[H, N]) CompletedRounds[H, N] {
+	rounds := make([]CompletedRound[H, N], 0, NumLastCompletedRounds)
+	if genesis != nil {
```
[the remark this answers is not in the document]
jimjbrettj: Could this end up appending an empty completedRound (with default empty values) to rounds though? Then we have a completed round with incorrect info

client/consensus/grandpa/environment.go
```diff
@@ -0,0 +1,466 @@
...
+		} else {
+			compRounds.Rounds = append(compRounds.Rounds[:idx+1], compRounds.Rounds[idx:]...)
+			compRounds.Rounds[idx] = completedRound
+		}
+	}
+
+	if len(compRounds.Rounds) > NumLastCompletedRounds {
+		compRounds.Rounds = compRounds.Rounds[:len(compRounds.Rounds)-1]
+	}
+}
```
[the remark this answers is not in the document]
jimjbrettj: set info is just a getter for information that is already public I do believe. Regarding iter, since as far as I can tell it is only used in one location and we don't have Iterator objects, I thought it better just to iterate in reverse in the one location where this is called. But can modify if that is not the ideal approach

client/consensus/grandpa/environment.go
```diff
@@ -0,0 +1,466 @@
...
+	if len(compRounds.Rounds) == 0 {
+		panic("inner is never empty; always contains at least genesis; qed")
+	}
+	return compRounds.Rounds[0]
+}
+
+// Push a new completed round, oldest round is evicted if number of rounds
+// is higher than `NUM_LAST_COMPLETED_ROUNDS`.
+func (compRounds *CompletedRounds[H, N]) Push(completedRound CompletedRound[H, N]) {
+	// TODO they use reverse, double check this
```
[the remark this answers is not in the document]
jimjbrettj: Yeah, I am a little confused by this section of code. Its not very clear to me how reversing impacts the binary search, and wanted to verify the expected behavior with reviewers

client/consensus/grandpa/environment.go
jimjbrettj: The fields are not public and cannot be scale encoded (as far as I am aware), which is necessary here

client/consensus/grandpa/authorities.go
timwu20: I'm not sure what the suggestion is? I thought we do have these type of constructors after you rebased.

jimjbrettj: This has been addressed in #3283

client/consensus/grandpa/aux_schema.go
```diff
@@ -0,0 +1,238 @@
...
+type KeyValue struct {
+	key   []byte
+	value []byte
+}
+
+// SharedVoterSetState .
+type SharedVoterSetState[H comparable, N constraints.Unsigned] struct {
+	// TODO move and implement
+	inner VoterSetState[H, N]
+}
```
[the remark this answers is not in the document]
jimjbrettj: updated

client/consensus/grandpa/aux_schema.go
```diff
@@ -0,0 +1,238 @@
...
+
+type persistentData[H comparable, N constraints.Unsigned] struct {
+	authoritySet SharedAuthoritySet[H, N] // TODO this needs to be shared auth set
+	setState     SharedVoterSetState[H, N]
+}
+
+func loadPersistent[H comparable, N constraints.Unsigned](store AuxStore, genesisHash H, genesisNumber N, genesisAuthorities []Authority) (*persistentData[H, N], error) {
```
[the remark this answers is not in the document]
jimjbrettj: I have implemented this

client/consensus/grandpa/aux_schema.go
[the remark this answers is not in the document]
jimjbrettj: I have implemented this

client/consensus/grandpa/environment.go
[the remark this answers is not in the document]
jimjbrettj: I was not able to see any reason that this should be reversed, so I will leave as is for now

client/consensus/grandpa/aux_schema.go
```diff
@@ -0,0 +1,349 @@
...
+
+	"github.com/ChainSafe/gossamer/client/api"
+	grandpa "github.com/ChainSafe/gossamer/pkg/finality-grandpa"
+)
+
+var (
+	SET_STATE_KEY      = []byte("grandpa_completed_round")
+	CONCLUDED_ROUNDS   = []byte("grandpa_concluded_rounds")
+	AUTHORITY_SET_KEY  = []byte("grandpa_voters")
+	BEST_JUSTIFICATION = []byte("grandpa_best_justification")
```
EclesioMeloJunior, line 19: For convention we don't use `snake_case` instead use `camelCase`

EclesioMeloJunior (changes requested):

client/consensus/grandpa/environment.go
```diff
@@ -0,0 +1,466 @@
...
+type CompletedRound[H comparable, N constraints.Unsigned] struct {
+	// The round number
+	Number uint64
+	// The round state (prevote ghost, estimate, finalized, etc.)
+	State grandpa.RoundState[H, N]
+	// The target block base used for voting in the round
+	Base grandpa.HashNumber[H, N]
+	// All the votes observed in the round
+	// I think this is signature type, double check
+	Votes []grandpa.SignedMessage[H, N, ed25519.SignatureBytes, ed25519.PublicKey]
```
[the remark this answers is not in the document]
timwu20: I'd suggest defining `AuthoritySignature` and `AuthorityID` as just type `any`.  Should make things simpler in t he tests and require one less import.  Actually calling ed25119 types should be dead simple at that point.  We just cast to `ed25519.SignatureBytes` or `ed25519.PublicKey` when we need to.

client/consensus/grandpa/environment.go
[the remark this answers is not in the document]
timwu20: it's not the job of this function to make sure we don't pass in a zero value `completedRound`.

client/consensus/grandpa/environment.go
```diff
@@ -0,0 +1,570 @@
...
+func (compRounds *completedRounds[H, N]) last() completedRound[H, N] {
+	if len(compRounds.Rounds) == 0 {
+		panic("inner is never empty; always contains at least genesis; qed")
+	}
+	return compRounds.Rounds[0]
+}
+
+// push a new completed round, oldest round is evicted if number of rounds
+// is higher than `NUM_LAST_COMPLETED_ROUNDS`.
+func (compRounds *completedRounds[H, N]) push(compRound completedRound[H, N]) {
```
timwu20, line 74: go best practice to use abbreviations
```suggestion
func (cr *completedRounds[H, N]) push(compRound completedRound[H, N]) {
```

client/consensus/grandpa/environment.go
```diff
@@ -0,0 +1,570 @@
...
+	Base grandpa.HashNumber[H, N]
+	// All the votes observed in the round
+	// I think this is signature type, double check
+	Votes []grandpa.SignedMessage[H, N, ed25519.SignatureBytes, ed25519.PublicKey]
+}
+
+// NumLastCompletedRounds NOTE: the current strategy for persisting completed rounds is very naive
+// (update everything) and we also rely on cloning to do atomic updates,
+// therefore this value should be kept small for now.
+const NumLastCompletedRounds = 2
```
timwu20, line 33: this shouldn't be public.

client/consensus/grandpa/environment_test.go
```diff
@@ -0,0 +1,687 @@
...
+	completedRounds := NewCompletedRounds[string, uint](&completedRound1, 1, authorities)
+	completedRounds.push(completedRound0)
+
+	lastCompletedRound := completedRounds.last()
+	require.Equal(t, completedRound0, lastCompletedRound)
+
+	completedRounds.push(completedRound2)
+	lastCompletedRound = completedRounds.last()
+	require.Equal(t, completedRound0, lastCompletedRound)
+}
```
timwu20, line 209: This test doesn't actually test any of the sorting logic.  `completedRound0` is the same as `completedRound1` and `completedRound2`.  I think you should write the same test with differing rounds in rust as well, and make sure we're doing the same thing as the rust type.

client/consensus/grandpa/environment.go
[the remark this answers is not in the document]
timwu20: Since they provide the `iter` method and explicitly reverse it we should probably replicate.  You can't expect users to know when or when to not reverse it.

client/consensus/grandpa/environment.go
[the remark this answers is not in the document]
timwu20: then remove the TODO?

client/consensus/grandpa/aux_schema.go
```diff
@@ -0,0 +1,349 @@
...
+		return err
+	}
+	return nil
+
+}
+
+func LoadAuthorities() {
+	// Not sure if this is really needed or just used in tests
+	panic("impl")
+}
```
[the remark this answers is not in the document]
jimjbrettj: removed for now since not used in current tests and will add when this is needed

client/consensus/grandpa/environment_test.go
jimjbrettj: This test has been updated, will verify with respect to substrate

client/consensus/grandpa/environment.go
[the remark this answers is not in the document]
jimjbrettj: updated

client/consensus/grandpa/environment_test.go
jimjbrettj: After spending more time reading the substrate code, I realized that I needed to change how these were ordered. I have updated the logic as well as the test.

timwu20 (approved):
great work!

PR #3386, opened 2023-07-13
feat(dot/state): keep latest state trie in memory

EclesioMeloJunior:
## Changes

Gossamer was taking too much time to load a state trie of the latest finalized block given that it was stored in the database and then pruned from memory. 

Is important for Gossamer to quickly load the state trie from the latest finalized block to keep the sync process at the normal speed, otherwise every finalization we will need to wait for Gossamer to load the entire trie from database which can be a bottleneck

## Tests

```sh
```

## Issues

N/A

## Primary Reviewer

@timwu20

### DIFF
--- a/dot/state/block_finalisation.go
+++ b/dot/state/block_finalisation.go
@@ -257,7 +257,8 @@ func (bs *BlockState) handleFinalisedBlock(currentFinalizedHash common.Hash) err
 			bs.tries.delete(blockHeader.StateRoot)
 		}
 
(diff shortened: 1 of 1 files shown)


jimjbrettj (approved):
Awesome work resolving this issue!

edwardmack (approved):
Looks great, I like the variable renames.

dimartiro (approved):
Awesome! as always

PR #3393, opened 2023-07-19
fix(lib/runtime): update `MaxPossibleAllocation` to `2^25`

EclesioMeloJunior:
## Changes

- This changes increases the amount of possible allocations from `2 ^ 24` to `2 ^ 25` as it is defined in substrate `primitives/core` 

```rs
pub const MAX_POSSIBLE_ALLOCATION: u32 = 33554432; // 2^25 bytes, 32 MiB
```
ref: https://github.com/paritytech/substrate/blob/2ca66197161b3b5c4e2aafdb7238f3c5af84a0e8/primitives/core/src/lib.rs#L400

- Also increase the `HeadsQty` to 23 as it is in substrate implementation

https://github.com/paritytech/substrate/blob/2ca66197161b3b5c4e2aafdb7238f3c5af84a0e8/client/allocator/src/freeing_bump.rs#L288C1-L288C1

This fixes the following error while syncing

image

## Tests

N/A

## Issues

N/A

## Primary Reviewer

@dimartiro

### DIFF
--- a/lib/runtime/allocator.go
+++ b/lib/runtime/allocator.go
@@ -20,7 +20,7 @@ const DefaultHeapBase = uint32(1469576)
 const alignment uint32 = 8
 
 // HeadsQty 22
-const HeadsQty = 22
+const HeadsQty = 23
 
 // MaxPossibleAllocation 2^25 bytes, 32 MiB
 const MaxPossibleAllocation = (1 << 25)
--- a/lib/runtime/allocator.go
+++ b/lib/runtime/allocator.go
@@ -22,7 +22,7 @@ const alignment uint32 = 8
 // HeadsQty 22
 const HeadsQty = 22
 
-// MaxPossibleAllocation 2^24 bytes
+// MaxPossibleAllocation 2^25 bytes, 32 MiB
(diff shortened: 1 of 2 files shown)


dimartiro (approved):
LGTM

dimartiro (changes requested):
There are some linting errors due arrays using 22 as max cap

dimartiro (changes requested):
I think you should fix [this test ](https://github.com/ChainSafe/gossamer/blob/4ea3361cb9d69300c74a053dd44d1108847b8068/lib/runtime/allocator_test.go#L486-L492) too

This could be a possible way to fix it:

```Go
func TestShouldGetMaxFromIndex(t *testing.T) {
	index := uint(HeadsQty - 1)
	itemSize := getItemSizeFromIndex(index)
	if itemSize != MaxPossibleAllocation {
		t.Errorf("item_size should be %d, got item_size: %d", MaxPossibleAllocation, itemSize)
	}
}
```

dimartiro (approved):

jimjbrettj (approved):

PR #3397, opened 2023-07-20
fixed decoding in Instance.ParachainHostPersistedValidationData

timwu20:
@kishansagathiya


kishansagathiya:
@timwu20 In this case, it is taking first two bytes from head data and considering them as length. That's why it is able to decode, but it decodes into wrong structs and first two bytes are missing from head data.

timwu20:
> @timwu20 In this case, it is taking first two bytes from head data and considering them as length. That's why it is able to decode, but it decodes into wrong structs and first two bytes are missing from head data.

But the rest of the struct works fine? That's what I'm assuming based on the asserts in the test.  I guess my questions is, how do you know the head data is incorrect? Have you tried taking the bytes that comes out from the runtime, and decoding into `Option<PersistedValidationData<H, N>>` in the polkadot repo and comparing?

kishansagathiya:
> how do you know the head data is incorrect

I have taken a head data value say `V` and used polkadot code to create encoding. So, when decoding with go-scale package, it should get the same head data value. Which it doesn't if we use pointer.


```
	let parent_head_string = "e902d91574d9e4897d88a7fb40130cf6c7900b5cb7238036726cd6c07a2255c8ed1c32a018010915879f32707df4a034c9a329ca83a80fab304d1a860690def304379ac236284091930e2b657bf56c4353bdca877b2c8a6bc33ba1611a5d79b2858b00bc707f08066175726120f4635e08000000000561757261010172b799cfe3e2ba2bd80349c7c92d1d84ff01ad6b3d491ff523ee2759e81dc22d58a94cd968ed300dbbc725144a04fa3622a11b2614255b802261d03c53af6f8e";
	let parent_head_bytes = FromHex::from_hex(parent_head_string).unwrap();
	let pvd1: PersistedValidationData<Hash, BlockNumber> = PersistedValidationData{
        parent_head: HeadData(parent_head_bytes),
        relay_parent_number: BlockNumber::from(15946390 as u32),
        relay_parent_storage_root: Hash::from_str("df650f4c6b9bfcc8f768c4d3037fafbd6831ff23473e090d443684fb5e305bd6").unwrap(),
        max_pov_size: 1024 * 1024 * 5,
    };
	
	// let encodedPVD = "f102e902d91574d9e4897d88a7fb40130cf6c7900b5cb7238036726cd6c07a2255c8ed1c32a018010915879f32707df4a034c9a329ca83a80fab304d1a860690def304379ac236284091930e2b657bf56c4353bdca877b2c8a6bc33ba1611a5d79b2858b00bc707f08066175726120f4635e08000000000561757261010172b799cfe3e2ba2bd80349c7c92d1d84ff01ad6b3d491ff523ee2759e81dc22d58a94cd968ed300dbbc725144a04fa3622a11b2614255b802261d03c53af6f8e9652f300f0b8960c208d6d10b4e0e9d9c046d241d00631205c4a4abc0a98ece9ce15b33f00005000";

	 let pvd1_to_hex = pvd1.encode()
	 .iter()
	 .map(|byte| format!("{:02x}", byte))
	 .collect::<Vec<String>>()
	 .join("");
	println!("{:?}",pvd1_to_hex);
```

kishansagathiya:
This was mostly to help me with a blocker in my work. don't think we need this anymore. Will leave it on @timwu20 to close this.

PR #3412, opened 2023-07-31
parachain/disputes: add scraping module for disputes-coordinator

kanishkatn:
## Changes

Implements scraping module for disputes-coordinator. From https://github.com/paritytech/polkadot/blob/master/node/core/dispute-coordinator/src/scraping/mod.rs

```
Scrapes unfinalized chain in order to collect information from blocks. Chain scraping
during disputes enables critical spam prevention. It does so by updating two important
criteria determining whether a vote sent during dispute distribution is potential
spam. Namely, whether the candidate being voted on is backed or included.

Concretely:

- Monitors for `CandidateIncluded` events to keep track of candidates that have been included on
   chains.
- Monitors for `CandidateBacked` events to keep track of all backed candidates.
- Calls `FetchOnChainVotes` for each block to gather potentially missed votes from chain.

 With this information it provides a `CandidateComparator` and as a return value of
 `process_active_leaves_update` any scraped votes.

Scraped candidates are available `DISPUTE_CANDIDATE_LIFETIME_AFTER_FINALIZATION` more blocks
after finalization as a precaution not to prune them prematurely. Besides the newly scraped
candidates `DISPUTE_CANDIDATE_LIFETIME_AFTER_FINALIZATION` finalized blocks are parsed as
another precaution to have their `CandidateReceipts` available in case a dispute is raised on
them
```

## Issues

#3302 

## Primary Reviewer

@EclesioMeloJunior


dot/parachain/dispute/scraping/inclusion.go
```diff
@@ -0,0 +1,90 @@
...
+	blocksIncluding, ok := i.inner[candidateHash]
+	if ok {
+		// Convert the map to a sorted slice for iteration
+		var sortedKeys []uint32
+		for h := range blocksIncluding {
+			sortedKeys = append(sortedKeys, h)
+		}
+		sort.Slice(sortedKeys, func(i, j int) bool {
+			return sortedKeys[i] < sortedKeys[j]
+		})
```
EclesioMeloJunior, line 68: As I noticed in the rust impl they don't use an ordered slice, is there any reason to use it here? https://github.com/paritytech/polkadot/blob/52209dcfe546ff39cc031b92d64e787e7e8264d4/node/core/dispute-coordinator/src/scraping/mod.rs#L132. Since `blocksIncluding` is a map from `uint32 -> []common.Hash{}` you can iterate over the values directly

```go
for height, blocksAtHeight := range blocksIncluding {
    for _, block := range blocksAtHeight {
	inclusionsAsSlice = append(inclusionsAsSlice, Inclusion{
		BlockNumber: height,
		BlockHash:   block,
	})
    }
}
```

dot/parachain/dispute/scraping/inclusion.go
```diff
@@ -0,0 +1,90 @@
...
+			}
+		}
+	}
+}
+
+// Get returns all inclusions for the given candidate hash.
+func (i *Inclusions) Get(candidateHash common.Hash) []Inclusion {
+	var inclusionsAsSlice []Inclusion
+	blocksIncluding, ok := i.inner[candidateHash]
+	if ok {
```
EclesioMeloJunior, line 60: Could you change to an early return if `!ok`?

EclesioMeloJunior (changes requested):

dot/parachain/dispute/scraping/inclusion.go
kanishkatn: Yes, it doesn't look like the order might not be needed. But, the rust implementation uses `BTreeMap` which will maintain the order of insertion. 

I'm sorting the block-height cause it made sense to process the ordered blocks rather than going about it in a random way. The tests expect it that way as well!

dot/parachain/dispute/test_helpers.go
```diff
@@ -108,3 +127,31 @@ func activateLeaf(
...
+}
+
+func DummyActivatedLeaf(blockNumber parachainTypes.BlockNumber) overseer.ActivatedLeaf {
+	return overseer.ActivatedLeaf{
+		Hash:   GetBlockNumberHash(blockNumber),
+		Number: uint32(blockNumber),
+	}
+}
+
+func NextLeaf(chain *[]common.Hash) overseer.ActivatedLeaf {
```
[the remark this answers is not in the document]
kanishkatn: Yeah, I've been meaning to revert this. Lemme see what I can do!

dot/parachain/dispute/scraping/inclusion.go
kanishkatn: The test: https://github.com/paritytech/polkadot/blob/52209dcfe546ff39cc031b92d64e787e7e8264d4/node/core/dispute-coordinator/src/scraping/tests.rs#L639

dot/parachain/dispute/test_helpers.go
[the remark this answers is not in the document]
kanishkatn: I remember now; this is needed for mocks. The `chain` undergoes runtime changes during the tests and those changes are needed by the mocks (overseer and runtime methods). 

We can't make it global since the tests run in parallel. Can you think of a better approach for this?

dot/parachain/dispute/scraping/scraping.go
```diff
@@ -0,0 +1,309 @@
...
+}
+
+// saturatingSub returns the result of a - b, saturating at 0.
+func saturatingSub(a, b uint32) uint32 {
+	result := int(a) - int(b)
+	if result < 0 {
+		return 0
+	}
+	return uint32(result)
+}
```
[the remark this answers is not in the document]
kanishkatn: I've modified the method. I'll keep the method here itself for now and remove it after the rebase!

dot/parachain/dispute/overseer/message.go
```diff
@@ -81,3 +81,23 @@ type ValidationResult struct {
...
+}
+
+type AncestorsResponse struct {
+	Ancestors []common.Hash
+	Error     error
+}
+
+type AncestorsRequest struct {
+	Hash            common.Hash
+	K               uint32
```
EclesioMeloJunior, line 101: What `K` means ?

dot/parachain/dispute/test_helpers.go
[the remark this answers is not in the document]
EclesioMeloJunior: I don't have anything specific, but since it is used for testing, I would suggest passing the `t *testing.T` and call the `t.Helper()`

dot/parachain/dispute/overseer/message.go
kanishkatn: The number of ancestors to request. 

I had worked on the `overseer` messages even before `feat/parachain` had the support for it so I did not document them. I'll be doing that in the final PR for disputes-coordinator.

Issue #3416, opened 2023-08-01
Construct a trie of chunks of erasure coded values


kishansagathiya:
## Issue summary

- To [make PoV available](https://github.com/paritytech/polkadot/blob/153543b0c8c582e73f520e5c08cbe33bddfb5f69/node/core/backing/src/lib.rs#L585), we need to validate erasure coded chunks against `expected_erasure_root`. We can do this by constructing a trie from erasure coded chunks. The root of that trie will be called `erasure_root`. We will check that the computed erasure root is the same as `expected_erasure_root`  from candidate descriptor.

- Currently in `lib/erasure`, we don't have ability to create a trie from erasure chunks.

## Other information and links

- [Use of this functionality in Polkadot's Candidate Backing](https://github.com/paritytech/polkadot/blob/153543b0c8c582e73f520e5c08cbe33bddfb5f69/node/core/backing/src/lib.rs#L601-L602)

- [Rust's implementation of this feature](https://github.com/paritytech/polkadot/blob/153543b0c8c582e73f520e5c08cbe33bddfb5f69/erasure-coding/src/lib.rs#L252-L274)

EclesioMeloJunior:
@axaysagathiya can we close this issue? tip: you can type `Closes #...` in the PR description then when it is merged the linked issue will be closed automatically!

PR #3417, opened 2023-08-01
feat(lib/erasure): construct a trie of chunks of erasure coded values

axaysagathiya:
## Tests

```sh
```

## Issues

#3416 
## Primary Reviewer

@kishansagathiya @edwardmack


lib/erasure/erasure.go
```diff
@@ -99,3 +103,24 @@ func Reconstruct(nValidators uint, chunks [][]byte) ([]byte, error) {
 
 	return res, nil
 }
+
+func ChunksToTrie(chunks [][]byte) (*trie.Trie, error) {
+	chunkTrie := trie.NewEmptyTrie()
+	for i, chunk := range chunks {
+		encodedI, err := scale.Marshal(uint32(i))
```
jimjbrettj, line 110: This specifically needs to be a uint32?

lib/erasure/erasure_test.go
```diff
@@ -12,100 +12,113 @@ import (
...
-	}{
-		{
-			name:              "1_validators",
-			validators:        1,
-			dataHex:           "0x04020000000000000000000000000000000000000000000000000000000000000000000000000000000000",
-			expectedChunksHex: []string{},
-			expectedError:     errors.New("expected at least 2 validators"),
+var testCases = []struct {
+	name          string
+	validators    uint
```
jimjbrettj, line 17: nit/should this be numValidators?

jimjbrettj (approved):
lgtm, just small comments

lib/erasure/erasure.go
axaysagathiya: for i = 6, 
`scale.Marshal(i)` returns [24] but,
`scale.Marshal(uint32(i))` returns [6, 0, 0, 0] which was the same encoding as the Polkadot code.

so, I think it should be uint32.

lib/erasure/erasure_test.go
axaysagathiya: Thank you for your suggestion! 
I've actually incorporated the use of `nValidators` consistently throughout the functions in this package. As a result, I've gone ahead and updated the field name from `validators` to `nValidators` to maintain uniformity.

PR #3428, opened 2023-08-08
feat(wazero): Add v2 runtime functions

dimartiro:
## Changes

This PR is a continuation of #2664 

### Changes
- [x] Implement runtime functions
  - [x] `ext_trie_blake2_256_root_version_2`
  - [x] `ext_trie_blake2_256_ordered_root_version_2`
  - [x] `ext_trie_blake2_256_verify_proof_version_2`
  - [x] `ext_default_child_storage_root_version_2`
  - [x] `ext_storage_root_version_2`
- [x] Add new test cases

Blocked by #3433 

## Tests

```sh
```

## Issues

#2418  

## Primary Reviewer

@timwu20


lib/runtime/wazero/imports_test.go
```diff
@@ -569,6 +571,35 @@ func Test_ext_trie_blake2_256_root_version_1(t *testing.T) {
 	require.Equal(t, expected[:], hash)
 }
 
+func Test_ext_trie_blake2_256_root_version_2(t *testing.T) {
+	inst := NewTestInstance(t, runtime.HOST_API_TEST_RUNTIME)
+
+	testinput := []string{"dimartiro", "was", "here", "??"}
+	encInput, err := scale.Marshal(testinput)
+	require.NoError(t, err)
+	encInput[0] = encInput[0] >> 1
```
EclesioMeloJunior, line 580: Why do you right shift the first index of `encInput`?

lib/trie/version.go
```diff
@@ -55,7 +55,17 @@ func (v Version) ShouldHashValue(value []byte) bool {
...
-func ParseVersion(s string) (version Version, err error) {
+func ParseVersion[T string | uint32](v T) (version Version, err error) {
+	var s string
+	switch value := any(v).(type) {
+	case string:
+		s = value
+	case uint32:
+		s = fmt.Sprintf("V%d", value)
+	default:
+		panic(fmt.Sprintf("unsupported type %T", s))
```
EclesioMeloJunior, line 66: This default case is useless since the type definitions in generic only defines `string | uint32` any other type will be denied at comptime

```suggestion
```

lib/trie/version_test.go
```diff
@@ -49,40 +50,63 @@ func Test_ParseVersion(t *testing.T) {
...
+			var version Version
+
+			var err error
+			switch typed := testCase.v.(type) {
+			case string:
+				version, err = ParseVersion(typed)
+			case uint32:
+				version, err = ParseVersion(typed)
+			default:
+				panic(fmt.Sprintf("unsupported type %T", testCase.v))
```
EclesioMeloJunior, line 108: instead of panic you could use `t.Fail()`

EclesioMeloJunior (changes requested):

lib/runtime/wazero/imports_test.go
dimartiro: Actually a very good question, I based this in the one Elizabeth made for v1. I asked Tim about this right shift a few days ago to see if he knows anything. Will investigate it more about it

lib/trie/version.go
dimartiro: True! thanks!!

lib/runtime/wazero/imports.go
```diff
@@ -2188,8 +2357,33 @@ func ext_storage_root_version_1(ctx context.Context, m api.Module) uint64 {
...
+	if rtCtx == nil {
+		panic("nil runtime context")
+	}
+	storage := rtCtx.Storage
+
+	stateVersionBytes, _ := m.Memory().Read(version, 4)
+	_, err := trie.ParseVersion(binary.LittleEndian.Uint32(stateVersionBytes))
+	if err != nil {
+		logger.Errorf("failed parsing state version: %s", err)
+		return 0
```
EclesioMeloJunior, line 2370: Since `ext_storage_root_version_2` caller expects a `Vec<u8>` I would suggest that in failure cases return a pointer to an empty `Vec<u8>` instead of 0

https://github.com/paritytech/substrate/blob/033d4e86cc7eff0066cd376b9375f815761d653c/primitives/io/src/lib.rs#L316C2-L316C2

lib/runtime/wazero/imports.go
```diff
@@ -1214,9 +1362,30 @@ func ext_default_child_storage_root_version_1(
...
+	stateVersion uint32) (ptrSize uint64) { //skipcq: RVV-B0012
+	rtCtx := ctx.Value(runtimeContextKey).(*runtime.Context)
+	if rtCtx == nil {
+		panic("nil runtime context")
+	}
+	storage := rtCtx.Storage
+	child, err := storage.GetChild(read(m, childStorageKey))
+	if err != nil {
+		logger.Errorf("failed to retrieve child: %s", err)
+		return 0
```
EclesioMeloJunior, line 1374: Since `ext_default_child_storage_root_version_2` caller expects a `Vec<u8>` I would suggest that in failure cases return a pointer to an empty `Vec<u8>` instead of 0

https://github.com/paritytech/substrate/blob/033d4e86cc7eff0066cd376b9375f815761d653c/primitives/io/src/lib.rs#L537

lib/runtime/wazero/imports.go
```diff
@@ -906,6 +1015,45 @@ func ext_trie_blake2_256_verify_proof_version_1(
...
+		logger.Errorf("failed parsing state version: %s", err)
+		return 0
+	}
+
+	toDecProofs := read(m, proofSpan)
+	var encodedProofNodes [][]byte
+	err = scale.Unmarshal(toDecProofs, &encodedProofNodes)
+	if err != nil {
+		logger.Errorf("failed scale decoding proof data: %s", err)
+		return uint32(0)
```
EclesioMeloJunior, line 1037: To keep the same as the other returns in the function
```suggestion
		return 0
```

EclesioMeloJunior (changes requested):

lib/runtime/wazero/imports.go
EclesioMeloJunior: Yes, I would suggest using the `mustWrite` function so you don't need to worry about errors and panics, and you can define a variable like `emptyByteVectorEncoded` and use it everywhere you need 

```go
// on top of imports.go

var (
 ...
 emptyByteVectorEncoded []byte = scale.MustMarshal([]byte{})
)

// in the ext_storage_root_version_2 function 
func ext_storage_root_version_2(...) ... {
   ...
   if err != nil {
       return mustWrite(m, rtCtx.Allocator, emptyByteVectorEncoded)
   }
   ...
}


```

dimartiro: Great! thanks!

PR #3433, opened 2023-08-10
chore(trie): Add support for trie V1

### LINKED ISSUE #2418: Implement storage trie version 1

dimartiro:
## Changes

Basically the thing that I changed is the way we encode the trie, when we are using v1 we can have hashed values for both branches and leaves.

What I did is to keep our trie as it is storing the value in the way we were doing it but I only change the encode / decode method to, for example, hash the values (greater than 32bytes) to be hashed in the encoded representation.
Since we are using the encoded node to calculate the hash I had to modify those methods too.

I also I added the `IsHashedValue` flag in `Node` structure for those cases where we are decoding nodes and they are encoded with the v1 format (usually used when verifying merkle proofs)

### Summary of changes
- [x]  Trie versions
    - [x] Calculate root hash based on state trie version
    - [x] Add default version 
- [x] Update runtime functions to use new trie versions when calculating hashes
- [x] Get trie version from runtime when needed
- [x] Added state-version arg in import state command 
- [x] Update tests
     - [x] Set V0 or DefaultVersion (V1) when needed
     - [x] Add test cases for both versions
- [x] Add new test cases

## Tests

```sh
make test
```

## Issues

close #2418 

## Primary Reviewer

@timwu20


lib/trie/trie.go
```diff
@@ -355,24 +361,40 @@ func findNextKeyChild(children []*Node, startIndex byte,
...
-	root, _, _, err := t.insert(t.root, nibblesKey, value, pendingDeltas)
+
+	isValueHashed := version.ShouldHashValue(value)
+	if isValueHashed {
+		hashedValue, err := common.Blake2bHash(value)
+		if err != nil {
+			return err
+		}
+
+		// Add original value in memory db
```
EclesioMeloJunior, line 389: here we don't know the database implementation exactly, so removing the `memory` makes more sense
```suggestion
		// Add original value in db and use the hash as node value
```

EclesioMeloJunior (changes requested):

dot/state/interfaces.go
```diff
@@ -29,10 +29,11 @@ type GetPutter interface {
 	Putter
 }
 
-// GetNewBatcher has methods to get values and create a
+// Batcher has methods to get values and create a
 // new batch.
-type GetNewBatcher interface {
+type Batcher interface {
```
kishansagathiya, line 34: `Batcher` highlights only one method of this interface (`NewBatch`).
I think it is common to name interface back on their methods, when we don't have other better names that would fit them.
So, names `GetterPutter` or `GetterPutterNewBatcher`, etc are acceptable.

cmd/gossamer/commands/import_state.go
```diff
@@ -62,7 +64,16 @@ func execImportState(cmd *cobra.Command) error {
 		return fmt.Errorf("header-file must be specified")
 	}
 
+	stateVersionFlag, err := cmd.Flags().GetString("state-version")
+	if err != nil {
+		return fmt.Errorf("failed to get state-version: %s", err)
+	}
+	stateVersion, err := trie.ParseVersion(stateVersionFlag)
+	if err != nil {
+		return fmt.Errorf("failed to parse state-version: %s", err)
```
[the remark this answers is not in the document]
dimartiro: Very good point! Thank you!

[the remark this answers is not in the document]
dimartiro: It's a valid point and I share it but I wanted to follow the convention we have in that file. We are returning `failed` in every error parsing a flag https://github.com/ChainSafe/gossamer/blob/development/cmd/gossamer/commands/import_state.go#L44-L60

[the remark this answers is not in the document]
kanishkatn: This is part of the `cmd` package and doesn't print the log with the `ERROR:` prefix like the node does. If we wanna use the same convention here, I believe it'll have to be a separate task.

EclesioMeloJunior (changes requested):

lib/runtime/genesis.go
```diff
@@ -26,7 +26,8 @@ func NewTrieFromGenesis(gen genesis.Genesis) (tr trie.Trie, err error) {
 			ErrGenesisTopNotFound, gen.Name)
 	}
 
-	tr, err = trie.LoadFromMap(keyValues)
+	//TODO: check if we could get the trie version from genesis
```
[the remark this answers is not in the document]
dimartiro: If I'm not wrong this is information is in the runtime, is there any way to get it in this function?

lib/trie/db/db.go
```diff
@@ -4,12 +4,20 @@ package db
...
 )
 
 type MemoryDB struct {
 	data map[common.Hash][]byte
+	l    sync.RWMutex
+}
+
+func NewEmptyMemoryDB() *MemoryDB {
+	return &MemoryDB{
+		data: make(map[common.Hash][]byte),
```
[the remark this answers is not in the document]
dimartiro: Ohh that's a good one! let me change it, thanks for the suggestion!

cmd/gossamer/commands/import_state.go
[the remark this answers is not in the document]
jimjbrettj: I tend to agree with @kishansagathiya to not include "failed" in the error message. However, since it is already used in this pkg, its my opinion that we leave as is for now, then can change in a separate task if/when we think it is appropriate to.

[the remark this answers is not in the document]
dimartiro: @kishansagathiya are you agree?

dot/import.go
```diff
@@ -62,7 +62,8 @@ func newTrieFromPairs(filename string) (*trie.Trie, error) {
 		entries[pairArr[0].(string)] = pairArr[1].(string)
 	}
 
-	tr, err := trie.LoadFromMap(entries)
+	//TODO: revisit this to use the right trie version
```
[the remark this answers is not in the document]
dimartiro: No, it's done, thanks!

lib/runtime/genesis.go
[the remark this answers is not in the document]
dimartiro: I was checking the code and we need to create the trie (execute this lines) before getting the runtime blob, so I'll change this line to use the default trie version instead V0

lib/trie/db/db.go
```diff
@@ -4,12 +4,20 @@ package db
...
 import (
 	"fmt"
+	"sync"
 
 	"github.com/ChainSafe/gossamer/lib/common"
 )
 
 type MemoryDB struct {
 	data map[common.Hash][]byte
+	l    sync.RWMutex
```
jimjbrettj, line 14: Why is this called 1? I think I would opt for a different name unless there is good reason for it

lib/trie/version.go
```diff
@@ -7,8 +7,12 @@ import (
 	"errors"
 	"fmt"
 	"strings"
+
+	"github.com/ChainSafe/gossamer/lib/common"
 )
 
+const V1MaxValueSize = common.HashLength
```
jimjbrettj, line 14: curious what others think, but im kind of in favor of avoiding using common types where we can

jimjbrettj (approved):
few comments, but overall looks great awesome work!

cmd/gossamer/commands/import_state.go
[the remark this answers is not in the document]
kishansagathiya: @dimartiro sorry, I had missed the conversation. This makes sense to me, resolving it.

lib/trie/version.go
jimjbrettj: But totally fine to keep here! Just airing my beef with the common pkg

lib/trie/trie.go
```diff
@@ -521,6 +518,7 @@ func (t *Trie) insertInBranch(parentBranch *Node, key, value []byte,
 			return nil, false, 0, fmt.Errorf("preparing branch for mutation: %w", err)
 		}
 		parentBranch.StorageValue = value
+		parentBranch.HashedValue = isValueHashed
```
timwu20, line 521: nit: maybe make it `node.IsHashedValue` to know that it's a boolean. 
```suggestion
		parentBranch.IsHashedValue = isValueHashed
```

lib/trie/trie.go
```diff
@@ -63,9 +64,15 @@ func (t *Trie) Snapshot() (newTrie *Trie) {
 		}
 	}
 
+	var dbCopy db.Database
+	if t.db != nil {
+		dbCopy = t.db.Copy()
+	}
```
timwu20, line 70: I don't think we should be copying the whole DB.  This isn't really possible for pebbledb.  I think we need to track the modifications for nodes with hashed values, and delete the old value from the DB if we end up pruning this trie.  Let me know if you wanna sync over this.

lib/trie/version.go
timwu20: I would just make this 
```suggestion
const V1MaxValueSize = 32
```

cmd/gossamer/commands/import_state_test.go
```diff
@@ -0,0 +1,83 @@
...
+	rootCmd.AddCommand(ImportStateCmd)
+
+	rootCmd.SetArgs([]string{ImportStateCmd.Name(),
+		"--state-version", "v0",
+		"--state-file", "test",
+		"--header-file", "test",
+	})
+	err = rootCmd.Execute()
+	assert.ErrorContains(t, err, "no such file or directory")
+}
```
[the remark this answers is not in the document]
dimartiro: Yep, I add these tests to increase the coverage. I think we can address that specific case in a different PR

lib/trie/trie.go
dimartiro: I solved this using the `deltas` and `pruner` we already have. Now every time we delete a hashed node that points to a value node we are deleting the value node too

lib/trie/trie.go
```diff
@@ -355,24 +353,37 @@ func findNextKeyChild(children []*Node, startIndex byte,
...
+	shouldHash := version.ShouldHashValue(value)
+	if shouldHash {
+		hashedValue := common.MustBlake2bHash(value)
+
+		// Add the original value as value node in db using the hashed value as key
+		err = t.db.Put(hashedValue.ToBytes(), value)
+		if err != nil {
+			return err
+		}
+		value = hashedValue.ToBytes()
```
timwu20, line 383:
```suggestion
		// Add the original value as value node in db using the hashed value as key
		value = hashedValue.ToBytes()
		err = t.db.Put(bytes, value)
		if err != nil {
			return err
		}
```

lib/trie/trie.go
```diff
@@ -383,22 +394,23 @@ func (t *Trie) insertKeyLE(keyLE, value []byte,
...
 	if parent == nil {
 		mutated = true
 		nodesCreated = 1
 		return &Node{
-			PartialKey:   key,
-			StorageValue: value,
-			Generation:   t.generation,
-			Dirty:        true,
+			PartialKey:    key,
+			StorageValue:  value,
```
timwu20, line 404: why are we storing the value in memory even if the value is hashed and stored in the DB?

lib/trie/trie.go
timwu20: Can you point me to the code that deletes the values from disk?

timwu20 (commented):
I think we need a test that validates that the merkle root in the event that we migrate a trie from v0 to v1.  I'm still unsure about the migration.  Say we read a node with a value greater than 32 bytes, are we expected to store the value and hash it and essentially update the merkle root even though we haven't changed the values stored?  I think a test that clearly shows the intended behaviour would be beneficial.

lib/runtime/wazero/imports.go
```diff
@@ -2198,8 +2254,33 @@ func ext_storage_root_version_1(ctx context.Context, m api.Module) uint64 {
...
+	}
+
+	logger.Debugf("root hash is: %s", root)
+
+	rootSpan, err := write(m, rtCtx.Allocator, root[:])
+	if err != nil {
+		logger.Errorf("failed to allocate: %s", err)
+		panic(err)
+	}
+	return rootSpan
```
timwu20, line 2283: from the comment found on substrate discussion ([link](https://github.com/paritytech/substrate/discussions/11607#discussioncomment-3102924)) I think a call to this host function is supposed to trigger a migration on the trie.   So all values > 32 bytes are supposed to be hashed from this point forward if the parsed version is v1.

dimartiro: I don't think this function should be responsible for that migration, they created a [pallet](https://github.com/paritytech/substrate/pull/10073) for that purpose. 

As far as I understand this function only calculates the root for the current state using 2 strategies depending on version parameter (v1 or v2). It does not change the values in the trie.

lib/trie/trie.go
dimartiro: I think we have to rethink the way we are retrieving this values. We should use lazy loading from the db but I have to check if other parts of the code are ready for that. I'm pretty sure the only part that loads the tries from DB is the [storage.go](https://github.com/ChainSafe/gossamer/blob/development/dot/state/storage.go#L103) but we are using standalone tries in different places.

Let me think more about it and find a generic solution

lib/trie/trie.go
dimartiro: Where is `bytes` defined in that version? are you thinking something like this?

lib/trie/version.go
```diff
@@ -18,27 +25,64 @@ const (
...
+	V0 Version = iota
+	V1
 )
 
+// DefaultStateVersion sets the state version we should use as default
+// See https://github.com/paritytech/substrate/blob/5e76587825b9a9d52d8cb02ba38828adf606157b/primitives/storage/src/lib.rs#L435-L439
+const DefaultStateVersion = V1
+
+type Entry struct{ Key, Value []byte }
+type Entries []Entry
```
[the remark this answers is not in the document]
dimartiro: Will add them

lib/runtime/wazero/imports_test.go
```diff
@@ -587,23 +618,47 @@ func Test_ext_trie_blake2_256_ordered_root_version_1(t *testing.T) {
 	require.Equal(t, expected[:], hash)
 }
 
+func Test_ext_trie_blake2_256_ordered_root_version_2(t *testing.T) {
+	inst := NewTestInstance(t, runtime.HOST_API_TEST_RUNTIME)
```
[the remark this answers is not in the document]
dimartiro: [This?](https://github.com/ChainSafe/polkadot-spec/pull/17)

lib/trie/trie.go
dimartiro: At the beginning I tought it was related with [this call](https://github.com/ChainSafe/gossamer/blob/53425ddf1db9619c43a7b19d8193ea52ecdd8c94/lib/trie/trie.go#L124 and [this part](https://github.com/ChainSafe/gossamer/blob/53425ddf1db9619c43a7b19d8193ea52ecdd8c94/dot/state/storage.go#L67-L72) in storage.go but I realized that pruner is [doing nothing](https://github.com/ChainSafe/gossamer/blob/53425ddf1db9619c43a7b19d8193ea52ecdd8c94/dot/state/pruner/pruner.go#L44-L47) so I'll be working on this too

dimartiro:
> I think we need a test that validates that the merkle root in the event that we migrate a trie from v0 to v1. I'm still unsure about the migration. Say we read a node with a value greater than 32 bytes, are we expected to store the value and hash it and essentially update the merkle root even though we haven't changed the values stored? I think a test that clearly shows the intended behaviour would be beneficial.

Thanks for all your comments and the time you spent in this review Tim.
After some more research I did, I really think we have to do a refactor in the way we are storing the state. Instead storing it as a trie we can store a key-value map and then use the trie functions only when we need to calculate the root / generate / verify proofs.

If you check the new trie functions for trie V1 there are basically read only:

`ext_trie_blake2_256_root_version_2`
`ext_trie_blake2_256_ordered_root_version_2`
`ext_trie_blake2_256_verify_proof_version_2`
`ext_default_child_storage_root_version_2`
`ext_storage_root_version_2`

There are not versioned Put / Set / Append functions, what makes me think that the values should be inserted in a generic way and then we should use the trie structure only when we need to calculate things like root hash, proofs, etc.
I'm really sure that thinking the state in a way that could be trie agnostic will simplify a lot of things.

I'm currently using my free time to experiment with this refactor in a different branch but I want to know your thought about it.

lib/trie/trie.go
dimartiro: I changed it this way

```suggestion
		valueKey := hashedValue.ToBytes()
		// Add the original value as value node in db using the hashed value as key
		err = t.db.Put(valueKey, value)
		if err != nil {
			return err
		}
		// Put the value key as this trie node value
		value = valueKey
```

lib/runtime/storage/trie_test.go
```diff
@@ -40,10 +40,56 @@ func TestTrieState_SetGet(t *testing.T) {
...
+		for _, tc := range testCases {
+			res, err := ts.GetChildStorage([]byte(tc), []byte(tc))
+			require.NoError(t, err)
+			require.Equal(t, []byte(tc), res)
+		}
+	}
+
+	ts := &TrieState{t: trie.NewEmptyTrie()}
+	testFunc(ts)
+}
```
axaysagathiya, line 63: What is the reason of having `testFunc`? Do we really need this extra layer of function?

lib/trie/db/db.go
```diff
@@ -4,12 +4,35 @@ package db
...
+}
+
+// DBPutter puts a value at the given key and returns an error.
+type DBPutter interface {
+	Put(key []byte, value []byte) error
+}
+
 type MemoryDB struct {
 	data map[common.Hash][]byte
+	l    sync.RWMutex
```
axaysagathiya, line 29: I suggest renaming this to mutex.
```suggestion
	mutex    sync.RWMutex
```

lib/trie/database_test.go
```diff
@@ -278,6 +287,17 @@ func Test_GetFromDB(t *testing.T) {
...
 
+func Test_GetFromDB_EmptyHash(t *testing.T) {
+	t.Parallel()
+
+	db := newTestDB(t)
+
+	value, err := GetFromDB(db, EmptyHash, []byte("test"))
+	assert.NoError(t, err)
+	assert.Nil(t, value)
+
```
axaysagathiya, line 298: remove this extra line

axaysagathiya (commented):
I found a few exported methods that don't have comments.
Other than that, LGTM.

lib/runtime/storage/trie_test.go
dimartiro: Agree, I removed it

lib/trie/version.go
```diff
@@ -18,27 +27,71 @@ const (
...
 }
 
-var ErrParseVersion = errors.New("parsing version failed")
+// ShouldHashValue returns true if the value should be hashed based on trie version
+func (v Version) ShouldHashValue(value []byte) bool {
+	return v == V1 && len(value) > V1MaxValueSize
+}
+
+// Root returns the root hash of the trie built using the given entries
+func (v Version) Root(entries Entries) (common.Hash, error) {
```
timwu20, line 65: I would have assumed this method would be on the `Trie` and would accept a version as a parameter, similar to how the host api functions are versioned.

dot/state/tries.go
```diff
@@ -51,6 +52,13 @@ func NewTries() (tries *Tries) {
 	}
 }
 
+// SetDB is to set the same db for every trie in the collection
+func (t *Tries) SetDB(db db.Database) {
```
timwu20, line 56: I thought we were going to get rid of the db?

lib/trie/version.go
dimartiro: Thats an option, but I prefer to use this one that is more like [substrate](https://github.com/paritytech/polkadot-sdk/blob/18257373b313fcfad4f7a2ab2cb02cb8bfc5b8f1/substrate/primitives/trie/src/lib.rs#L81)

lib/trie/db/db.go
```diff
@@ -30,16 +53,46 @@ func NewMemoryDBFromProof(encodedNodes [][]byte) (*MemoryDB, error) {
...
+	newDB.data = copyData
+	return newDB
+}
+
+func (mdb *MemoryDB) Get(key []byte) ([]byte, error) {
+	if len(key) != common.HashLength {
+		return nil, fmt.Errorf("expected %d bytes length key, given %d (%x)", common.HashLength, len(key), key)
 	}
 	var hash common.Hash
 	copy(hash[:], key)
```
EclesioMeloJunior, line 73: Is this copy needed? I would suggest using the key directly. We will not change the `key`, we just use it to get a value from a map

```
if value, found := mdb.data[key]; found {
	return value, nil
}
```

lib/trie/proof/generate.go
```diff
@@ -22,7 +23,7 @@ var (
 // Database defines a key value Get method used
 // for proof generation.
 type Database interface {
-	Get(key []byte) (value []byte, err error)
+	db.DBGetter
 }
```
EclesioMeloJunior, line 27: Since this interface is just a wrapper for `db.DBGetter` I would suggest to remove it and use `db.DBGetter` directly

EclesioMeloJunior (changes requested):
LGTM, just small observations

lib/runtime/wazero/imports.go
```diff
@@ -781,33 +782,35 @@ func ext_crypto_finish_batch_verify_version_1(ctx context.Context, m api.Module)
...
-
-	t := trie.NewEmptyTrie()
-
-	type kv struct {
-		Key, Value []byte
+	stateVersionBytes, _ := m.Memory().Read(version, 4)
+	stateVersion, err := trie.ParseVersion(binary.LittleEndian.Uint32(stateVersionBytes))
+	if err != nil {
+		logger.Errorf("failed parsing state version: %s", err)
+		return 0
```
EclesioMeloJunior, line 798: You don't need to read the version from memory, the state version is already given to you through the `version` argument, the version is either `0` or `1`.

https://spec.polkadot.network/chap-host-api#id-version-2-prototype-9

So you can just:
```go
stateVersion, err := trie.ParseVersion(version)
```

lib/runtime/wazero/imports.go
```diff
@@ -817,49 +820,48 @@ func ext_trie_blake2_256_root_version_1(ctx context.Context, m api.Module, dataS
...
 	data := read(m, dataSpan)
 
-	t := trie.NewEmptyTrie()
+	stateVersionBytes, _ := m.Memory().Read(version, 4)
+	stateVersion, err := trie.ParseVersion(binary.LittleEndian.Uint32(stateVersionBytes))
+	if err != nil {
+		logger.Errorf("failed parsing state version: %s", err)
+		return 0
+	}
+
```
EclesioMeloJunior, line 847: You don't need to read the version from memory, the state version is already given to you through the `version` argument, the version is either `0` or `1`.

https://spec.polkadot.network/chap-host-api#id-version-2-prototype-9

So you can just:

```suggestion
	stateVersion, err := trie.ParseVersion(version)
	if err != nil {
		logger.Errorf("failed parsing state version: %s", err)
		return 0
	}
```

lib/runtime/wazero/imports.go
```diff
@@ -898,7 +894,46 @@ func ext_trie_blake2_256_verify_proof_version_1(
...
+	if rtCtx == nil {
+		panic("nil runtime context")
+	}
+
+	stateVersionBytes, _ := m.Memory().Read(version, 4)
+	_, err := trie.ParseVersion(binary.LittleEndian.Uint32(stateVersionBytes))
+	if err != nil {
+		logger.Errorf("failed parsing state version: %s", err)
+		return 0
+	}
```
EclesioMeloJunior, line 929: Same here, you should use the version directly
```suggestion
	_, err := trie.ParseVersion(version)
	if err != nil {
		logger.Errorf("failed parsing state version: %s", err)
		return 0
	}
```

EclesioMeloJunior (changes requested):

internal/trie/node/header_test.go
```diff
@@ -293,7 +293,7 @@ func Test_encodeHeader_At_Maximum(t *testing.T) {
 		PartialKey: make([]byte, keyLength),
 	}
 
-	err := encodeHeader(node, buffer)
+	err := encodeHeader(node, math.MaxInt, buffer)
```
timwu20, line 296: use the `NoMaxInlineValueSize`?

lib/trie/version.go
```diff
@@ -18,27 +28,90 @@ const (
 	// V0 is the state trie version 0 where the values of the keys are
 	// inserted into the trie directly.
 	// TODO set to iota once CI passes
-	V0 Version = 1
+	V0 Version = iota
+	V1
```
timwu20, line 32: Do you think it makes sense to mirror the parity implementation and name this `Layout`? I think it makes more sense given that the underlying db isn't changed, so it's not like we're versioning what is stored.

lib/trie/db/db.go
dimartiro: The copy is to convert the `key` into a `common.Hash` since data is `data  map[common.Hash][]byte`

I think and alternative is to create the Hash using the key, like: `hashedKey := common.Hash(key)`

internal/trie/node/header.go
```diff
@@ -21,17 +21,19 @@ func encodeHeader(node *Node, writer io.Writer) (err error) {
 		panic(fmt.Sprintf("partial key length is too big: %d", partialKeyLength))
 	}
 
+	isHashedValue := len(node.StorageValue) > maxInlineValueSize
```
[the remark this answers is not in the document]
dimartiro: Since we are storing the "unhashed" value in our in memory trie, there is shouldn't be chance to have a `node.IsHashedValue` here

lib/trie/trie.go
```diff
@@ -181,84 +199,52 @@ func (t *Trie) MustHash() common.Hash {
 }
 
 // Hash returns the hashed root of the trie.
-func (t *Trie) Hash() (rootHash common.Hash, err error) {
+func (t *Trie) Hash(maxInlineValue int) (rootHash common.Hash, err error) {
```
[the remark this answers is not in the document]
dimartiro: I did it this way to match the substrate impl, where they have the [MAX_INLINE_VALUE](https://github.com/paritytech/trie/blob/6eb4b88e0781da2db7dda0021249a8fa4244bc5d/trie-db/src/lib.rs#L523) defined in the layout and then they use this value as [treshold](https://github.com/paritytech/polkadot-sdk/blob/2e001de9346257321a8e2447d22e86d7c71a917a/substrate/primitives/trie/src/lib.rs#L81-L88) when they are [hashing the node](https://github.com/paritytech/trie/blob/6eb4b88e0781da2db7dda0021249a8fa4244bc5d/trie-root/src/lib.rs#L171-L180)

lib/trie/version.go
```diff
@@ -6,7 +6,17 @@ package trie
...
 	"fmt"
+	"math"
 	"strings"
+
+	"github.com/ChainSafe/gossamer/lib/common"
+)
+
+const (
+	// NoMaxInlineValueSize is the numeric representation used to indicate that there is no max value size.
+	NoMaxInlineValueSize = math.MaxInt
```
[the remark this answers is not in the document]
dimartiro: I think `NoMaxInlineValueSize` is more generic and descriptive even if it is used by the V0 only

dot/core/service.go
```diff
@@ -116,14 +117,34 @@ func (s *Service) Stop() error {
 	return nil
 }
 
+func (s *Service) GetCurrentStateTrieVersion() (trie.Version, error) {
```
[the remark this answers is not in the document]
dimartiro: Sure! thks

lib/trie/version.go
dimartiro: Sure!

lib/runtime/wazero/imports.go
dimartiro: Very good point, since this is my first time changing the runtime I have some questions regarding the way we pass the parameters.

I'm doing this in the test:

```go
stateVersion := uint32(trie.V1)
stateVersionBytes := make([]byte, 4)
binary.LittleEndian.PutUint32(stateVersionBytes, stateVersion)
```

What is the right way to pass it as integer instead of bytes? or do I have to change anything else?

lib/runtime/wazero/imports.go
dimartiro: Yes, we don't need to use it with our current verify proof implementation

lib/runtime/wazero/imports.go
dimartiro: I maybe need to change the [changes](https://github.com/ChainSafe/polkadot-spec/blob/c6fa5767265e022f0fc5cd0c50a8c3e97b8a65bb/test/runtimes/hostapi/src/lib.rs#L621C41-L621C41) I made in our test runtime too?

EclesioMeloJunior: I think we don't need to change the testing, when you call a runtime function you need to parse everything to bytes.

EclesioMeloJunior (approved):

timwu20 (approved):

cmd/gossamer/commands/import_state.go
```diff
@@ -7,13 +7,15 @@ import (
...
 	"github.com/ChainSafe/gossamer/dot"
+	"github.com/ChainSafe/gossamer/lib/trie"
 	"github.com/ChainSafe/gossamer/lib/utils"
 	"github.com/spf13/cobra"
 )
 
 func init() {
 	ImportStateCmd.Flags().String("chain", "", "Chain id used to load default configuration for specified chain")
 	ImportStateCmd.Flags().String("state-file", "", "Path to JSON file consisting of key-value pairs")
+	ImportStateCmd.Flags().Uint32("state-version", uint32(trie.DefaultStateVersion), "State version to use when importing state") //nolint:lll
```
kanishkatn, line 18:
```suggestion
	ImportStateCmd.Flags().Uint32("state-version", 
	        uint32(trie.DefaultStateVersion), 
	        "State version to use when importing state",
	)
```

cmd/gossamer/commands/import_state.go
```diff
@@ -54,6 +57,15 @@ func execImportState(cmd *cobra.Command) error {
 		return fmt.Errorf("state-file must be specified")
 	}
 
+	stateVersion, err := cmd.Flags().GetUint32("state-version")
+	if err != nil {
+		return fmt.Errorf("failed to get state-file: %s", err)
```
kanishkatn, line 62:
```suggestion
		return fmt.Errorf("failed to get state-version: %s", err)
```

cmd/gossamer/commands/import_state.go
```diff
@@ -26,7 +28,8 @@ var ImportStateCmd = &cobra.Command{
 in the form of key-value pairs to be imported.
 Input can be generated by using the RPC function state_getPairs.
 Example: 
-	gossamer import-state --state-file state.json --header-file header.json --first-slot <first slot of network>`,
+	gossamer import-state --state-file state.json state-version 1 --header-file header.json 
+	--first-slot <first slot of network>`,
```
kanishkatn, line 32:
```suggestion
	gossamer import-state --state-file state.json --state-version 1 --header-file header.json 
	--first-slot <first slot of network>`,
```

kanishkatn (approved):
Great job!

cmd/gossamer/commands/import_state.go
dimartiro: Thanks!

Issue #3398, opened 2023-07-20
Gossamer panics with `runtime error: slice bounds out of range [:-8]`


EclesioMeloJunior:
## Describe the bug

After syncing around `6183757` blocks gossamer panics with `panic: runtime error: slice bounds out of range [:-8]` and this error comes from the badger db, here is the stack trace:

```
panic: runtime error: slice bounds out of range [:-8]

goroutine 10917 [running]:
github.com/dgraph-io/badger/v4/y.CompareKeys({0x0?, 0xc001baaf60?, 0x10000454c?}, {0xc000da6030?, 0x0?, 0xc001baaf70?})
	github.com/dgraph-io/badger/v4@v4.1.0/y/y.go:141 +0x109
github.com/dgraph-io/badger/v4/table.(*Iterator).seekFrom.func1(0xc0007308f0?)
	github.com/dgraph-io/badger/v4@v4.1.0/table/iterator.go:279 +0x6d
sort.Search(0x0?, 0xc001bab000)
	sort/search.go:65 +0x48
github.com/dgraph-io/badger/v4/table.(*Iterator).seekFrom(0xc0005ec460, {0xc000da6030, 0x30, 0x30}, 0x2?)
	github.com/dgraph-io/badger/v4@v4.1.0/table/iterator.go:276 +0x139
github.com/dgraph-io/badger/v4/table.(*Iterator).seek(...)
	github.com/dgraph-io/badger/v4@v4.1.0/table/iterator.go:310
github.com/dgraph-io/badger/v4/table.(*Iterator).Seek(0xc0001a62d0?, {0xc000da6030?, 0x2?, 0x1?})
	github.com/dgraph-io/badger/v4@v4.1.0/table/iterator.go:424 +0x39
github.com/dgraph-io/badger/v4.(*levelHandler).get(0xc000b0a780, {0xc000da6030, 0x30, 0x30})
	github.com/dgraph-io/badger/v4@v4.1.0/level_handler.go:293 +0x258
github.com/dgraph-io/badger/v4.(*levelsController).get(0x0?, {0xc000da6030, 0x30, 0x30}, {0x0, 0x0, 0x0, {0x0, 0x0, 0x0}, ...}, ...)
	github.com/dgraph-io/badger/v4@v4.1.0/levels.go:1595 +0x145
github.com/dgraph-io/badger/v4.(*DB).get(0xc000600d80, {0xc000da6030, 0x30, 0x30})
[...]

PR #3434, opened 2023-08-10
feat(internal/database): replace `chaindb/badgerdb` with `pebbledb`

EclesioMeloJunior:
## Changes

- Removes all occurrences of `chaindb` from Gossamer source code
- Introduces PebbleDB implementation to Gossamer under `internals/database`
- Uses the same interface as `chaindb`
- Changes `nodeBuilderIface.isNodeInitialised` to return `bool` instead of error, this method was responsible to check if the node was already initialized by checking the database folder, given the change in the database this method should be refactored and that is why I changed its signature
- Introduces `utils.ClearDatabase`, this function was a straightforward option to generate a fresh new database since PebbleDB does not have a method to clear the database, it is safe to set up a new database after the `ClearDatabase` was called

https://pkg.go.dev/github.com/cockroachdb/pebble
## Tests

```sh
go test -tags integration github.com/ChainSafe/gossamer/internal/database
```

## Issues

- Closes https://github.com/ChainSafe/gossamer/issues/3398

## Primary Reviewer

@timwu20


dot/node.go
```diff
@@ -78,26 +78,37 @@ type nodeBuilder struct{}
...
-	return err == nil
+	return nodeInstance.isNodeInitialised(basepath)
 }
 
 // isNodeInitialised returns nil if the node is successfully initialised
 // and an error otherwise.
-func (*nodeBuilder) isNodeInitialised(basepath string) error {
+func (*nodeBuilder) isNodeInitialised(basepath string) (bool, error) {
 	// check if key registry exists
-	registry := filepath.Join(basepath, utils.DefaultDatabaseDir, "KEYREGISTRY")
```
jimjbrettj, line 91: we dont need "KEYREGISTRY" anymore?

dot/state/initialize.go
```diff
@@ -37,11 +37,7 @@ func (s *Service) Initialise(gen *genesis.Genesis, header *types.Header, t *trie
 
 	s.db = db
 
-	if err = db.ClearAll(); err != nil {
```
jimjbrettj, line 40: Why not clear all anymore?

internal/database/pebble.go
```diff
@@ -0,0 +1,146 @@
...
+	return nil
+}
+
+func (p *pebbleDB) NewBatch() Batch {
+	return &pebbleBatch{
+		batch: p.db.NewBatch(),
+	}
+}
+
+func (p *pebbleDB) NewIterator() Iterator {
```
jimjbrettj, line 117: Public functions will need comments yes?

internal/database/table.go
```diff
@@ -0,0 +1,65 @@
...
+import (
+	"bytes"
+)
+
+type table struct {
+	db     Database
+	prefix []byte
+}
+
+var _ Table = (*table)(nil)
```
jimjbrettj, line 15: what is this doing?

jimjbrettj (commented):
Nice work! Just a few comments

dot/node.go
EclesioMeloJunior: No, pebbleDB don't create this file

dot/state/offline_pruner.go
```diff
@@ -172,33 +169,48 @@ func (p *OfflinePruner) Prune() error {
...
+		return nil // no upper-bound
+	}
+
+	prefixIterOptions := &pebble.IterOptions{
+		LowerBound: storagePrefixBytes,
+		UpperBound: keyUpperBound(storagePrefixBytes),
+	}
+
+	// Ignore non-storage keys
+	inputDBIter := inputDB.NewIter(prefixIterOptions)
```
kishansagathiya, line 191: possible to reuse `pebble.NewPrefixIterator` here ?

internal/database/pebble.go
```diff
@@ -0,0 +1,146 @@
...
+		return fmt.Errorf("writing 0x%x with value 0x%x to database: %w",
+			key, value, err)
+	}
+	return nil
+}
+
+func (p *pebbleDB) Get(key []byte) (value []byte, err error) {
+	value, closer, err := p.db.Get(key)
+	if err != nil {
+		return nil, err
```
kishansagathiya, line 60: wrap error?

internal/database/pebble.go
```diff
@@ -0,0 +1,146 @@
...
+		return false, fmt.Errorf("closing after get: %w", err)
+	}
+
+	return value != nil, err
+}
+
+func (p *pebbleDB) Del(key []byte) error {
+	err := p.db.Delete(key, &pebble.WriteOptions{})
+	if err != nil {
+		return err
```
kishansagathiya, line 92: wrap error?

internal/database/pebble.go
```diff
@@ -0,0 +1,146 @@
...
+
+		for i := len(end) - 1; i >= 0; i-- {
+			end[i] = end[i] + 1
+			if end[i] != 0 {
+				return end[:i+1]
+			}
+		}
+
+		return nil
+	}
```
kishansagathiya, line 136: dumb question: what does this do?

kishansagathiya (approved):

lib/trie/trie_endtoend_test.go
```diff
@@ -267,24 +267,15 @@ func TestDeleteOddKeyLengths(t *testing.T) {
...
 		err = db.Close()
 		require.NoError(t, err)
 	})
 
-	storageDB := chaindb.NewTable(db, "storage")
-	t.Cleanup(func() {
-		err = storageDB.Close()
-		require.NoError(t, err)
-	})
-
```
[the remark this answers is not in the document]
EclesioMeloJunior: No, `storageDB` is a table created using the `db` so closing the table will close the DB and there is another clean up in the same test that closes the `db` so closing it twice results in panic, I will remove the `Close` method from `Table` interface

internal/database/pebble.go
EclesioMeloJunior: I don't think wrapping the error here will be useful since this method is just a proxy to the `db.Get`, I mean, what kind of context we can add to the error message that is useful to the caller?

internal/database/pebble.go
EclesioMeloJunior: I don't think wrapping the error here will be useful since this method is just a proxy to the `db.Delete`, I mean, what kind of context we can add to the error message that is useful to the caller?

dot/state/offline_pruner.go
EclesioMeloJunior: Yes, done!

dot/state/initialize.go
jimjbrettj: seems good to me

jimjbrettj (approved):
awesome work!

cmd/gossamer/README.md
```diff
@@ -185,9 +185,8 @@ What follows is a list that describes the services and capabilities that inform
 
 #### State
 
-This service is a wrapper around an instance of [`chaindb`](https://github.com/ChainSafe/chaindb), a key-value database
-that is built on top of [BadgerDB](https://github.com/dgraph-io/badger) from [Dgraph](https://dgraph.io/). The state
-service provides storage capabilities for the other Gossamer services - each service is assigned a prefix that is added
+This service is a wrapper around an instance of [`pebble`](https://github.com/cockroachdb/pebble), a LevelDB/RocksDB inspired key-value.
```
danforbes, line 188:
```suggestion
This service is a wrapper around an instance of [`pebble`](https://github.com/cockroachdb/pebble), a LevelDB/RocksDB inspired key-value database. 
```

dimartiro (approved):
Good job! Now that we are not using chaindb anymore, are we planning to delete chaindb repo from chainsafe? or set it as unmaintained?

EclesioMeloJunior:
> Good job! Now that we are not using chaindb anymore, are we planning to delete chaindb repo from chainsafe? or set it as unmaintained?

Good point, IMO once we merge this PR we should mark the `ChainSafe/chaindb` repo as unmaintained since there is some external projects that use it. 

@danforbes @timwu20

lib/utils/utils.go
```diff
@@ -14,20 +14,22 @@ import (
...
-		InMemory: inMemory,
-	})
+func SetupDatabase(basepath string, inMemory bool) (database.Database, error) {
+	nodeDatabaseDir := filepath.Join(basepath, DefaultDatabaseDir)
+	return database.NewPebble(nodeDatabaseDir, inMemory)
+}
+
+func ClearDatabase(basepath string) error {
+	nodeDatabaseDir := filepath.Join(basepath, DefaultDatabaseDir)
+	return os.RemoveAll(nodeDatabaseDir)
```
timwu20, line 32: can you move this to the `database` package? Rather not continue to add to this `utils` package given this package doesn't conform to go package naming best practices.

lib/utils/utils.go
```diff
@@ -14,20 +14,22 @@ import (
...
 // DefaultDatabaseDir directory inside basepath where database contents are stored
 const DefaultDatabaseDir = "db"
 
 // SetupDatabase will return an instance of database based on basepath
-func SetupDatabase(basepath string, inMemory bool) (*chaindb.BadgerDB, error) {
-	return chaindb.NewBadgerDB(&chaindb.Config{
-		DataDir:  filepath.Join(basepath, DefaultDatabaseDir),
-		InMemory: inMemory,
-	})
+func SetupDatabase(basepath string, inMemory bool) (database.Database, error) {
```
timwu20, line 25:
```suggestion
func LoadDatabase(basepath string, inMemory bool) (database.Database, error) {
```

internal/database/pebble_batch.go
```diff
@@ -0,0 +1,52 @@
...
+)
+
+var _ Batch = (*pebbleBatch)(nil)
+
+type pebbleBatch struct {
+	batch *pebble.Batch
+}
+
+func (pb *pebbleBatch) Put(key, value []byte) error {
+	err := pb.batch.Set(key, value, &pebble.WriteOptions{})
```
timwu20, line 19: looks like this is ignored in `pebble.Batch.Set()`
```suggestion
	err := pb.batch.Set(key, value, nil)
```

timwu20 (changes requested):

internal/database/pebble.go
EclesioMeloJunior: They are required for public functions, I don't think public methods are required export comments, but I've added some

timwu20 (approved):

Issue #3409, opened 2023-07-27
Gossamer fails while syncing: failed to execute block `#7165794`


EclesioMeloJunior:
## Describe the bug

- failed to execute block number `7165794` with hash `0x03ff62c2f25209593b2fdf87b2206f6b09e5877f76026c7c7fa527b465453f61`

EclesioMeloJunior:
After a quick investigation over the stack trace and looking into the proxy pallet and the block extrinsic (https://westend.subscan.io/block/7165782), what was happening is that:

- The extrinsic executes a proxy pallet call that is directed to the crowdloan pallet, and the Crowdloan pallet uses storage version 2 https://github.com/paritytech/polkadot/blob/d73503cc2351ffec047a4912ecdb3e1eab5f46aa/runtime/common/src/crowdloan/mod.rs#L184
- This version uses 4 arguments to call the host storage and Gossamer host function only have 3, which is causing the runtime to panic

EclesioMeloJunior:
The problem is that the pallet is trying to retrieve a value from the host through the default child storage, as the value does not exist in the host the host should return `None` however Gossamer is returning the value 0 which is causing the problem since the runtime is expecting a `Option<Vec<u8>>`, to fix that we should return `None` or `[]byte{0x00}` if we don't find the value

EclesioMeloJunior:
Fixed with
- https://github.com/ChainSafe/gossamer/pull/3451
- https://github.com/ChainSafe/gossamer/pull/3459

PR #3451, opened 2023-08-23
fix(lib/runtime): return correct encoded value for `None`

EclesioMeloJunior:
## Changes

- Fixes the host functions that are returning `0` instead of returning `None`
- https://github.com/paritytech/substrate/blob/df383f224b4de179c68c64f533ff4957b881938c/frame/support/src/storage/child.rs#L34

## Tests

```sh
```

## Issues

- Closes https://github.com/ChainSafe/gossamer/issues/3457
- Related https://github.com/ChainSafe/gossamer/issues/3409

## Primary Reviewer

@kishansagathiya


kishansagathiya (approved):

jimjbrettj (approved):
lgtm

PR #3454, opened 2023-08-24
client(consensus/grandpa): implement grandpa justification logic

jimjbrettj:
## Changes

- Add justification logic to grandpa using generics
- Write unit tests for justification logic
- Added justification logic to grandpa aux db
- made finalityGrandpa Message value public for encoding/setting

## Tests

```sh
```

## Primary Reviewer

@EclesioMeloJunior


client/consensus/grandpa/aux_schema.go
```diff
@@ -179,15 +179,39 @@ func UpdateAuthoritySet[H comparable, N constraints.Unsigned](store AuxStore, se
 // We always keep around the justification for the best finalized block and overwrite it
 // as we finalize new blocks, this makes sure that we don't store useless justifications
 // but can always prove finality of the latest block.
-func UpdateBestJustification() {
-	// TODO impl when we have justification logic
-	panic("impl")
+func UpdateBestJustification[H HashI, N constraints.Unsigned, S comparable, ID constraints.Ordered, Header HeaderI[H, N]](store AuxStore, justification Justification[H, N, S, ID, Header]) error {
+	encodedJustificaiton, err := scale.Marshal(justification)
+	if err != nil {
+		return err
```
EclesioMeloJunior, line 185: wrap error
```suggestion
		return fmt.Errorf("marshaling: %w", err)
```

client/consensus/grandpa/aux_schema.go
```diff
@@ -179,15 +179,39 @@ func UpdateAuthoritySet[H comparable, N constraints.Unsigned](store AuxStore, se
...
+	}
+
+	insert := []KeyValue{
+		{BEST_JUSTIFICATION, encodedJustificaiton},
+	}
+	err = store.Insert(insert, nil)
+	if err != nil {
+		return err
```
EclesioMeloJunior, line 193: wrap error
```suggestion
		return fmt.Errorf("marshaling: %w", err)
```

client/consensus/grandpa/aux_schema.go
```diff
@@ -179,15 +179,39 @@ func UpdateAuthoritySet[H comparable, N constraints.Unsigned](store AuxStore, se
...
-	// TODO impl when we have justification logic
-	panic("impl")
+func UpdateBestJustification[H HashI, N constraints.Unsigned, S comparable, ID constraints.Ordered, Header HeaderI[H, N]](store AuxStore, justification Justification[H, N, S, ID, Header]) error {
+	encodedJustificaiton, err := scale.Marshal(justification)
+	if err != nil {
+		return err
+	}
+
+	insert := []KeyValue{
+		{BEST_JUSTIFICATION, encodedJustificaiton},
```
EclesioMeloJunior, line 189: For consistency with our code, I would suggest writing as camel case instead. So `bestJustification` for private or `BestJustification` if it needs to be public to other packages

client/consensus/grandpa/aux_schema.go
```diff
@@ -179,15 +179,39 @@ func UpdateAuthoritySet[H comparable, N constraints.Unsigned](store AuxStore, se
...
+	if encodedJustification != nil {
+		justification := Justification[H, N, S, ID, Header]{}
+		err = scale.Unmarshal(*encodedJustification, &justification)
+		if err != nil {
+			return nil, err
+		}
+		return &justification, nil
+	}
+	// Can return error here is we want to error out if nothing returned
+	return nil, nil
```
EclesioMeloJunior, line 214: I would suggest returning a sentinel error in case of not found, and also returning early 
```suggestion
        if encodedJustification == nil {
		return nil, ErrJustificationNotFound
        }
        
	justification := Justification[H, N, S, ID, Header]{}
	err = scale.Unmarshal(*encodedJustification, &justification)
	if err != nil {
		return nil, fmt.Errof("unmarshalling: %w", err)
	}
	return &justification, nil
```

client/consensus/grandpa/interfaces.go
```diff
@@ -42,3 +42,19 @@ type AuxStore interface {
 	// Get Query auxiliary data from key-Value store.
 	Get(key []byte) (*[]byte, error)
 }
+
+type HashI interface {
+	constraints.Ordered
```
EclesioMeloJunior, line 47: why do we need to hashes be ordered? since they are hashes how they can be ordered?

client/consensus/grandpa/interfaces.go
```diff
@@ -42,3 +42,19 @@ type AuxStore interface {
...
+	IsEmpty() bool
+}
+
+type HeaderI[H HashI, N constraints.Unsigned] interface {
+	ParentHash() H
+	Hash() H
+	Number() N
+}
+
+type HeaderBackend[H HashI, N constraints.Unsigned, Header HeaderI[H, N]] interface {
```
EclesioMeloJunior, line 57: why do we need to be generic over the header? why not use the concrete `types.Header` struct?

client/consensus/grandpa/lib.go
```diff
@@ -24,3 +26,37 @@ type NewAuthoritySetStruct[H comparable, N constraints.Unsigned] struct {
 	SetId       N
 	Authorities []Authority
 }
+
+type msgData[H comparable, N constraints.Unsigned] struct {
```
EclesioMeloJunior, line 30: I would say to avoid short names for struct since they can be confused with variable names
```suggestion
type messageData[H comparable, N constraints.Unsigned] struct {
```

client/consensus/grandpa/lib.go
```diff
@@ -24,3 +26,37 @@ type NewAuthoritySetStruct[H comparable, N constraints.Unsigned] struct {
...
+	Round   uint64
+	SetID   uint64
+	Message finalityGrandpa.Message[H, N]
+}
+
+// Check a message signature by encoding the message as a localized payload and
+// verifying the provided signature using the expected authority id.
+// The encoding necessary to verify the signature will be done using the given
+// buffer, the original content of the buffer will be cleared.
+func checkMessageSignature[H comparable, N constraints.Unsigned](
```
EclesioMeloJunior, line 40: this function should return an error since it might fail to marshal/verify

pkg/finality-grandpa/lib.go
```diff
@@ -80,18 +80,13 @@ func (m Message[H, N]) Target() HashNumber[H, N] {
...
-func (m Message[H, N]) Value() any {
-	return m.value
-}
-
 // Messages is the interface constraint for `Message`
 type Messages[Hash, Number any] interface {
 	Prevote[Hash, Number] | Precommit[Hash, Number] | PrimaryPropose[Hash, Number]
 }
 
 func setMessage[Hash, Number any, T Messages[Hash, Number]](m *Message[Hash, Number], val T) {
```
EclesioMeloJunior, line 88: I think I left this code pass in the previous reviews but I think this function is useless since you can just set the value directly in the `newMessage` func

```go 

func newMessage[Hash, Number any, T Messages[Hash, Number]](val T) (m Message[Hash, Number]) {
	return Message[Hash, Number]{
            Value: val
        }
}

```

EclesioMeloJunior (changes requested):

client/consensus/grandpa/interfaces.go
jimjbrettj: Its because in the finalityGrandpaPackage hashes are stored in the btree which requires an ordering property. It doesn't need to be ordered for our own purposes, but it is needed for btree storage

client/consensus/grandpa/interfaces.go
jimjbrettj: This is because we call the function 
`commitValidationResult, err := finalityGrandpa.ValidateCommit[H, N, S, ID](j.Commit, voters, ancestryChain)`
which requires ancestryChain (which uses the header) to be generic.

However, if there is a way around this let me know and I will update

pkg/finality-grandpa/lib.go
jimjbrettj: I plan to talk to Tim about this section of code in our sync, so I will bring this up as well

client/consensus/grandpa/justification.go
```diff
@@ -0,0 +1,289 @@
...
+	return fmt.Errorf("%w", errInvalidAuthoritiesSet)
+}
+
+// Validate the commit and the votes' ancestry proofs.
+func verifyWithVoterSet[H HashI, N constraints.Unsigned, S comparable, ID constraints.Ordered, Header HeaderI[H, N]](j Justification[H, N, S, ID, Header], setID uint64, voters finalityGrandpa.VoterSet[ID]) error {
+	ancestryChain := newAncestryChain[H, N](j.VotesAncestries)
+	commitValidationResult, err := finalityGrandpa.ValidateCommit[H, N, S, ID](j.Commit, voters, ancestryChain)
+	if err != nil || !commitValidationResult.Valid() {
+		return fmt.Errorf("%w: invalid commit in grandpa justification", errBadJustification)
+	}
```
timwu20, line 160:
```suggestion
	if err != nil {
		return fmt.Errorf("%w: invalid commit in grandpa justification", errBadJustification)
	}
	if !commitValidationResult.Valid() {
		return fmt.Errorf("%w: invalid commit in grandpa justification", errBadJustification)
	}
```

timwu20: nit: I find it easier to read.  I had to check in rust how/when the cases fall through. Not super intuitive.

client/consensus/grandpa/justification.go
```diff
@@ -0,0 +1,289 @@
...
+		return fmt.Errorf("%w: invalid precommit ancestries in grandpa justification with unused headers", errBadJustification)
+	}
+
+	// Check if maps are equal
+	for visitedHash, _ := range visitedHashes {
+		_, ok := ancestryHashes[visitedHash]
+		if !ok {
+			return fmt.Errorf("%w: invalid precommit ancestries in grandpa justification with unused headers", errBadJustification)
+		}
+	}
```
timwu20, line 222: might as well use `reflect.DeepEqual` in this case.

client/consensus/grandpa/aux_schema.go
```diff
@@ -182,15 +186,39 @@ func UpdateAuthoritySet[H comparable, N constraints.Unsigned](store AuxStore, se
 // We always keep around the justification for the best finalized block and overwrite it
 // as we finalize new blocks, this makes sure that we don't store useless justifications
 // but can always prove finality of the latest block.
-func UpdateBestJustification() {
-	// TODO impl when we have justification logic
-	panic("impl")
+func UpdateBestJustification[H HashI, N constraints.Unsigned, S comparable, ID constraints.Ordered, Header HeaderI[H, N]](store AuxStore, justification Justification[H, N, S, ID, Header]) error {
```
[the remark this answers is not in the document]
jimjbrettj: updated

client/consensus/grandpa/interfaces.go
```diff
@@ -43,3 +43,19 @@ type AuxStore interface {
 	// Get Query auxiliary data from key-Value store.
 	Get(key []byte) (*[]byte, error)
 }
+
+type HashI interface {
+	constraints.Ordered
+	IsEmpty() bool
+}
+
+type HeaderI[H HashI, N constraints.Unsigned] interface {
```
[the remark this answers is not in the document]
jimjbrettj: This makes me have to rename the current parameters that I use. For example:

`UpdateBestJustification[H HashI, N constraints.Unsigned, S comparable, ID constraints.Ordered, Header HeaderI[H, N]]`

would have to rename it to something like `Head Header[H, N]` which might be a little more confusing to read? I am fine with this update though whatever you think is best

client/consensus/grandpa/justification.go
```diff
@@ -0,0 +1,289 @@
...
+}
+
+// TODO get feedback on if I can avoid this, see below
+//type NewAuthority[ID constraints.Ordered] struct {
+//	Key    ID
+//	Weight uint64
+//}
+
+// Validate the commit and the votes' ancestry proofs.
+// func (j *Justification[H, N, S, ID]) verify(setID uint64, authorities []NewAuthority[ID]) error {
```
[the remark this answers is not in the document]
jimjbrettj: updated

client/consensus/grandpa/justification.go
```diff
@@ -0,0 +1,289 @@
...
+func verify[H HashI, N constraints.Unsigned, S comparable, ID constraints.Ordered, Header HeaderI[H, N]](j Justification[H, N, S, ID, Header], setID uint64, weights []finalityGrandpa.IDWeight[ID]) error {
+	// TODO Get reviewer feedback on this. In substrate they pass in data then convert to IDWeight, however for
```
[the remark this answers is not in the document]
jimjbrettj: Without modifying our types I don't believe we can do this. Currently the ID for `finalityGrandpa.IDWeight` is constrained to be ordered, however in practice, our authority key (our ID in this case) is a byte slice. 

Perhaps we can make a wrapper around the publicKey type if we wish

client/consensus/grandpa/justification.go
```diff
@@ -0,0 +1,315 @@
...
+	baseHash := minPrecommit.Precommit.TargetHash
+	visitedHashes := make(map[H]struct{})
+	for _, signed := range precommits {
+		/*
+				TODO signature is generic type any, but signature verification in Gossamer uses []byte
+			    TODO need access to message.value both for encoding and so I can set it
+		*/
+		//if !checkMessageSignature[H, N](signed.Precommit, signed.ID, signed.Signature, j.Round, setID) {
+		//	return fmt.Errorf("%w: invalid signature for precommit in grandpa justification", errBadJustification)
+		//}
```
timwu20, line 209: doesn't this need to be done?

client/consensus/grandpa/justification.go
```diff
@@ -0,0 +1,315 @@
...
+	if !reflect.DeepEqual(ancestryHashes, visitedHashes) {
+		return fmt.Errorf("%w: invalid precommit ancestries in grandpa justification with unused headers",
+			errBadJustification)
+	}
+
+	return nil
+}
+
+// The target block NumberField and HashField that this justifications proves finality for
+func (j *Justification[H, N, S, ID, Header]) target() hashNumber[H, N] { //nolint
```
timwu20, line 253: I think this needs to be public.

client/consensus/grandpa/justification.go
```diff
@@ -0,0 +1,289 @@
...
+			return nil, fmt.Errorf("%w", errBlockNotDescendentOfBase)
+		}
+		block = br.ParentHash()
+
+		if !block.IsEmpty() {
+			currentHash = block
+			route = append(route, currentHash)
+		} else {
+			return nil, fmt.Errorf("%w", errBlockNotDescendentOfBase)
+		}
```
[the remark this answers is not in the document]
timwu20: if the map doesn't contain `current_hash` it returns `Err(GrandpaError::NotDescendent)`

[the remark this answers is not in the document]
jimjbrettj: yup this makes sense, will update

client/consensus/grandpa/interfaces.go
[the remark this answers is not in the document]
jimjbrettj: Unless I am misunderstanding, this is not valid syntax. When I constrain HeaderBackend with `Header Header[H, N]` for example I get "Cannot use type parameter as constraint"

client/consensus/grandpa/justification.go
jimjbrettj: Yes, however, I am currently struggling because the justification ID is constrained to be constraints.Ordered and for this we need to be able to cast to an ed25519.PubKey type, which is an invalid conversion.

I am still brainstorming how to approach this and would be happy to sync to discuss this issue

client/consensus/grandpa/interfaces.go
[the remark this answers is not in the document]
jimjbrettj: This has been updated to name hash constraint `Hash` and header is now `H Header[Hash, N]`

client/consensus/grandpa/justification.go
jimjbrettj: This has been added and resolved now

client/consensus/grandpa/lib.go
```diff
@@ -20,10 +28,60 @@ type Authority[ID AuthorityID] struct {
...
+	SetID   uint64
+	Message finalityGrandpa.Message[H, N]
+}
+
+// Check a message signature by encoding the message as a localised payload and
+// verifying the provided signature using the expected authority id.
+// The encoding necessary to verify the signature will be done using the given
+// buffer, the original content of the buffer will be cleared.
+func checkMessageSignature[H comparable, N constraints.Unsigned, ID AuthorityID](
+	message any,
```
timwu20, line 52: why not just accept a `finalityGrandpa.Message` as the param instead of `any`?

jimjbrettj: updated

client/consensus/grandpa/justification.go
```diff
@@ -0,0 +1,313 @@
...
+	"golang.org/x/exp/constraints"
+)
+
+var (
+	errInvalidAuthoritiesSet    = errors.New("current state of blockchain has invalid authorities set")
+	errBadJustification         = errors.New("bad justification for header")
+	errBlockNotDescendentOfBase = errors.New("block not descendent of base")
+)
+
+// Justification A GRANDPA justification for block finality, it includes a commit message and
```
timwu20, line 22:
```suggestion
// Justification is a GRANDPA justification for block finality, it includes a commit message and
```

client/consensus/grandpa/justification.go
```diff
@@ -0,0 +1,313 @@
...
+// an ancestry proof including all headers routing all precommit target blocks
+// to the commit target block. Due to the current voting strategy the precommit
+// targets should be the same as the commit target, since honest voters don't
+// vote past authority set change blocks.
+//
+// This is meant to be stored in the db and passed around the network to other
+// nodes, and are used by syncing nodes to prove authority set handoffs.
+type Justification[
```
timwu20, line 30: Any reason why you didn't just make it name the same as substrate (`GrandpaJustification`)?

timwu20 (approved):

client/consensus/grandpa/justification.go
jimjbrettj: Yeah, the editor complains that the type name begins with the package name.

Issue #3418, opened 2023-08-02
Manage communication between different parachain subsystems


kishansagathiya:
## Issue summary
`lib/parachain` has a bunch of subsystems that need to run in parallel and there are more of these subsystems that will get added.

Many of them need to communicate with each other by say calling some function or fetching some data. 

It would not be a nice design to connect two parachain subsystems directly, because 
- It would require us to manage a lot of links/connections for each subsystems
- we might need to add extra logic to make these communication non-blocking 

It would be nicer to have say a separate subsystem to manage these communications (equivalent to polkadot's overseer subsystem). The idea is that
- each subsystem will only communicate with overseer subsystem only.
- each subsystem will send their request to a overseer and continue with rest of its task without waiting for a reply. overseer will send a reply whenever it is ready. 

Because overseer would mandate each subsystem to perform some common task, we could create an interface say `SubSystem` that could have tasks like
- receive requests for overseer
- process them and send a response.

Feel free to do these task as separate function or as part of the same function based on your best judgement.

## Other information and links

- Polkadot's implementation of overseer https://github.com/paritytech/polkadot/blob/153543b0c8c582e73f520e5c08cbe33bddfb5f69/node/overseer/src/lib.rs
- Explanation in node implementer's guide https://paritytech.github.io/polkadot/book/node/overseer.html

kishansagathiya:
@kanishkatn @edwardmack @axaysagathiya 

I would love to know if you have any alternate or additional design ideas for this.

kanishkatn:
I'm taking a look at the overseer and I have some concerns. 

In the Rust implementation, it supports three usecases
- SendMessage
Sends a message to the overseer and waits for the response
- Feed
Sends a message to the overseer and doesn't wait for any response
- SendUnBoundedMessage
A channel without any message limits. This ensures that the messages from the subsystem will always reach the overseer without wait.

In the current implementation, I've noticed that there is only one channel between the overseer and a subsystem. This may lead to channels getting blocked and lead to hard to debug issues. 

I suggest not to keep one single channel to send a response message from overseer to the subsystems and add a way for the subsystem to specify the receiver channel along with the message to the overseer.

I'm not sure if this has already been discussed; I'm open to suggestions as well!

kishansagathiya:
> I suggest not to keep one single channel to send a response message from overseer to the subsystems and add a way for the subsystem to specify the receiver channel along with the message to the overseer.

[...]

PR #3460, opened 2023-08-29
feat(dot/parachain): Implement overseer skeleton

edwardmack:
## Changes
The overseer is responsible for:
1. Setting up, monitoring, and handing failure for overseer subsystems.
2. Providing a "heartbeat" of which relay-parents subsystems should be working on.
3. Acting as a message bus between subsystems.

This PR starts to implement these by implementing how the overseer will start, send OverseerSignal::ActiveLeavesUpdate events to subsystems, and receive messages from subsystems.  This a base to build upon for further implementing functionality needed by the overseer.

## Tests

This test shows overseer starting subsystems, and send AciveLeaf and receiving messages.
```sh
go test github.com/ChainSafe/gossamer/dot/parachain/overseer -v
```

## Issues
Issue #3418 

## Primary Reviewer

@kishansagathiya 
@timwu20


dot/parachain/overseer/overseer.go
```diff
@@ -0,0 +1,74 @@
...
+package overseer
+
+import "fmt"
+
+type Overseer struct {
+	subsystems        map[Subsystem]*Context
+	subsystemMessages map[Subsystem]chan any
+}
+
+type ExamlpeSender struct {
```
jimjbrettj, line 13:
```suggestion
type ExampleSender struct {
```

jimjbrettj (approved):
lgtm

dot/parachain/overseer/overseer.go
```diff
@@ -0,0 +1,74 @@
...
+		subsystems:        make(map[Subsystem]*Context),
+		subsystemMessages: make(map[Subsystem]chan any),
+	}
+}
+
+func (o *Overseer) RegisterSubsystem(subsystem Subsystem) {
+	// TODO: determine best buffer size
+	comChan := make(chan any, 5)
+	o.subsystemMessages[subsystem] = comChan
+	receiverChan := make(chan any, 5)
```
kishansagathiya, line 33: we should add buffer size on top of the file as constant and use it here.

dot/parachain/overseer/overseer.go
```diff
@@ -0,0 +1,74 @@
...
+}
+
+func (o *Overseer) Start() {
+	// start subsystems
+	for subsystem, context := range o.subsystems {
+		go func(sub Subsystem, ctx *Context) {
+			err := sub.Run(ctx)
+			if err != nil {
+				// TODO: handle error
+				fmt.Printf("error running subsystem %v: %v\n", sub, err)
```
kishansagathiya, line 47: We should add a logger on top the file and use that here to log error, sorta like
```
var (
	logger = log.NewFromGlobal(log.AddContext("pkg", "parachain-overseer"))
)
```

dot/parachain/overseer/overseer.go
```diff
@@ -0,0 +1,74 @@
...
+		}(subsystem, context)
+	}
+
+	// wait for messages from subsystems
+	// TODO: this is a temporary solution, we will determine logic to handle different message types
+	for subsystem, recChan := range o.subsystemMessages {
+		go func(sub Subsystem, channel chan any) {
+			fmt.Printf("overseer waiting for messages from %v\n", sub)
+
+			for {
```
kishansagathiya, line 58: We should add a return or break in this for loop to end this goroutine. Currently, it is a forever goroutine.

dot/parachain/overseer/overseer.go
```diff
@@ -0,0 +1,74 @@
...
+
+func (o *Overseer) RegisterSubsystem(subsystem Subsystem) {
+	// TODO: determine best buffer size
+	comChan := make(chan any, 5)
+	o.subsystemMessages[subsystem] = comChan
+	receiverChan := make(chan any, 5)
+	o.subsystems[subsystem] = &Context{
+		Sender:   &ExamlpeSender{senderChan: comChan},
+		Receiver: receiverChan,
+	}
```
kishansagathiya, line 37: I think we would be able to just use a common channel on which all subsystems send and overseer receives. let's call this channel `overseerChan`. We can add this as a field in parachain service struct.

```
    // Declare a channel that can only receive values of type string
    var ch1 <-chan string
    // Declare a channel that can only send values of type string
    var ch2 chan<- string
```
Let's use directional channels to make sure that we can only receive on this channel from overseer and only send from other subsystems. Let's return this `overseerChan` in `NewOverseer`.

For subsystems to receive, we would need different channels for each subsystems. I feel like these channels should be created by different subsystems and passed to overseer on registerSubsystem.

dot/parachain/overseer/overseer.go
EclesioMeloJunior: Agree, we should pass to the `Start` method a `stop chan struct{}` argument that goroutines should listen for and whenever it triggers all goroutines should return. Also, I would suggest having a `sync.WaitGroup` in the `Overseer` struct and for each goroutine we increment 1 in the wait group

```go
type Overseer struct {
    wg sync.WaitGroup
    doneCh chan struct{}
    stopCh chan struct{}
}

// here stop channel is just a read-only channel which means that
func NewOverseer() *Overseer {
  return &Overseer{
    doneCh: make(chan struct{}),
    stopCh: make(chan struct{}),
  }
}

func (o *Overseer) Start() {
    for subsystem, context := range o.subsystems {
        o.wg.Add(1)
        go func(..., wg *sync.WaitGroup) {
            defer wg.Done()
            select {
            case <-o.stopCh:
               return // here you can shutdown gracefully the goroutine
            case ...
            }
        }
    }

    for subsystem, recChan := range o.subsystemMessages {
        o.wg.Add(1)
        go func(..., wg *sync.WaitGroup) {
            defer wg.Done()
            select {
            case <-o.stopCh:
               return // here you can shutdown gracefully the goroutine
            case ...
            }
        }
    }
}

func (o *Overseer) Stop() {
    if o.doneCh == nil {
        return nil // or return an error 
    }

    close(o.stopCh)
    // once you close the stop channel you should
    // wait for the goroutines to end until
    o.wg.Wait()

    // assign nil to stopCh prevents subsequent calls to
    // Stop method
    o.stopCh = nil  
}
```

EclesioMeloJunior (changes requested):

dot/parachain/overseer/overseer.go
```diff
@@ -0,0 +1,108 @@
...
+
+package overseer
+
+import (
+	"fmt"
+	"github.com/ChainSafe/gossamer/internal/log"
+	"sync"
+)
+
+const COMMS_BUFFER_SIZE = 5
```
EclesioMeloJunior, line 12: Following our project's style (also golang) you could rewrite it using camelCase `CommsBufferSize`

dot/parachain/overseer/overseer.go
```diff
@@ -0,0 +1,108 @@
...
+func (o *Overseer) Start() {
+	// start subsystems
+	for subsystem, context := range o.subsystems {
+		o.wg.Add(1)
+		go func(sub Subsystem, ctx *Context, wg sync.WaitGroup) {
+			defer wg.Done()
+			select {
+			case <-o.stopCh:
+				return
+			}
```
EclesioMeloJunior, line 65: you should put `default:` otherwise it will block until `stopCh` is triggered 
```suggestion
			select {
			case <-o.stopCh:
				return
			default:
			}
```

EclesioMeloJunior (changes requested):

dot/parachain/overseer/overseer.go
edwardmack: Good point, I also moved this logic to the go routine inside the subsystem, instead of in the go routine that starts the subsystems.

dot/parachain/overseer/overseer.go
edwardmack: I've updated this to share the overseerChan with subsystems, so the overseer can receive messages from subsystems.  I've added tests to simulate message passing, I think this is working the way that we intend, but I'm not sure I have used directional channels the way we envision.  Let me know your thoughts.

dot/parachain/overseer/types.go
```diff
@@ -0,0 +1,35 @@
...
+
+package overseer
+
+import (
+	"sync"
+
+	"github.com/ChainSafe/gossamer/lib/common"
+)
+
+type Context struct {
```
timwu20, line 12: does this need to be public?

dot/parachain/overseer/types.go
```diff
@@ -0,0 +1,35 @@
...
+}
+
+type ActivatedLeaf struct {
+	Hash   common.Hash
+	Number uint32
+}
+
+type ActiveLeavesUpdate struct {
+	Activated *ActivatedLeaf
+}
```
timwu20, line 30: Can you provide comments for exported types? Is this the same terminology they use in polkadot?

dot/parachain/overseer/overseer.go
```diff
@@ -0,0 +1,102 @@
...
+	// TODO: add logic to start listening for Block Imported events and Finalisation events
+}
+
+func (o *Overseer) Stop() {
+	if o.doneCh == nil {
+		return
+	}
+	close(o.stopCh)
+	o.wg.Wait()
+	o.stopCh = nil
```
timwu20, line 94: I think you need to add timeout functionality here.  What if the subsystem never stops?

dot/parachain/overseer/overseer.go
```diff
@@ -0,0 +1,102 @@
...
+			for { //nolint:gosimple
+				select {
+				case msg := <-channel:
+					fmt.Printf("overseer received message from %v: %v\n", sub, msg)
+				}
+			}
+		}(subsystem, recChan)
+	}
+
+	// TODO: add logic to start listening for Block Imported events and Finalisation events
```
timwu20, line 85: I think figuring out these parts may change how the overseer is currently implemented.

timwu20 (commented):
Take a look at the service architecture discussion https://github.com/ChainSafe/gossamer/discussions/2293.  I feel like this should be implemented using the service architecture, with the ability to send messages.  I do kinda like how the `Sender` is defined as an interface with a synchronous `SendMessage` function that returns an error.  So we don't have to manually determine if a connection is severed by monitoring whether a channel has closed.  But I feel like the `Subsystem` could determine if the connection to the `Overseer` is dropped by closing the `Receiver` channel which would allow you to remove `Context.stopCh`.

dot/parachain/overseer/types.go
edwardmack: no, updated.

dot/parachain/overseer/overseer.go
```diff
@@ -0,0 +1,102 @@
...
+}
+
+type ExampleSender struct {
+	senderChan chan any
+}
+
+func (s *ExampleSender) SendMessage(msg any) error {
+	s.senderChan <- msg
+	return nil
+}
```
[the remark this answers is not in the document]
edwardmack: updated.

dot/parachain/overseer/types.go
edwardmack: Yes, added comments, yes this is the terminology used in polkadot.

dot/parachain/overseer/overseer.go
edwardmack: Added.

dot/parachain/overseer/overseer.go
edwardmack: I agree, and I think (hope) that the interfaces defined will remain the same, while the functionality is finalized.

edwardmack:
I've updated this to use our service architecture, and added wait group timeout.

kanishkatn:
I've also come across a usecase while working on disputes-subsystem. In polkadot, they use two sender methods
- `SendMessage()` : Sends a message to the overseer and waits till it gets a response
- `Feed()` : Sends a message to the overseer and exits

I've found this to make the tests easier to write; but not sure how relevant it is for this PR.

edwardmack:
Updates: 
- Added use of content.ContextWithCancel to signal subsystems when overseer has canceled their work.
- Use SendMessage interface for subsystems to send messages to overseer.
- Use channels for sending messages from overseer to subsystem.
- Use sync.WaitGroup to monitor when subsystems have stopped.
- added waitTimeout function to end if subsystem waitgroup waits too long.

kishansagathiya:
I would like to merge this skeleton PR and do additions and improvements as extra smaller PRs.

kishansagathiya (approved):

dot/parachain/overseer/overseer.go
```diff
@@ -0,0 +1,115 @@
...
+	go func() {
+		defer close(c)
+		wg.Wait()
+	}()
+	select {
+	case <-c:
+		return false // completed normally
+	case <-time.After(timeout):
+		return true // timed out
+	}
```
EclesioMeloJunior, line 114: I would suggest using time.NewTimer, since we are avoiding using time.After, using NewTimer you can stop the timer once the channel is closed 
```suggestion
        timeoutTimer := time.NewTimer(timeout)
	select {
	case <-c:
	        if !timeoutTimer.Stop() { <-timeoutTimer.C }
		return false // completed normally
	case <-timeoutTimer.C:
		return true // timed out
	}
```

dot/parachain/overseer/overseer.go
```diff
@@ -0,0 +1,115 @@
...
+
+	return nil
+}
+
+// sendActiveLeavesUpdate sends an ActiveLeavesUpdate to the subsystem
+func (o *Overseer) sendActiveLeavesUpdate(update ActiveLeavesUpdate, subsystem Subsystem) {
+	o.subsystems[subsystem] <- update
+}
+
+func waitTimeout(wg *sync.WaitGroup, timeout time.Duration) bool {
```
EclesioMeloJunior, line 103:
```suggestion
func waitTimeout(wg *sync.WaitGroup, timeout time.Duration) (timeouted bool) {
```

EclesioMeloJunior (commented):
LGTM, just a small comment in the `waitTimeout` function

dot/parachain/overseer/overseer.go
```diff
@@ -0,0 +1,115 @@
...
+}
+
+func (o *Overseer) processMessages() {
+	for {
+		select {
+		case msg := <-o.SubsystemsToOverseer:
+			switch msg.(type) {
+			default:
+				logger.Error("unknown message type")
+			}
```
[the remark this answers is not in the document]
kishansagathiya: This is intentional. Actual message types haven't been added yet. Actual message types could be added along with subsystems. We can just copy message types from polkadot-sdk, but if we do it when we add the subsystem, the types would be more accurate since we would know what we need to send.
If we translate message types from polkadot-sdk, we might end up write a lot of varying data type code, unnecessarily.

EclesioMeloJunior (approved):

PR #3251, opened 2023-05-10
feat(infra): Use shared disk as state snapshot in staging instances

dimartiro:
## Changes

Use AWS EFS disks to store blocks state and share it between nodes for the same chain.
This EFS disks are configure to create backups daily acting as daily chain snapshots

Would be great if we update chainDB to use badgerV4 before merging this PR

## Tests

Tested on AWS

## Issues

#3252 

## Primary Reviewer

@timwu20

PR #3469, opened 2023-09-05
feat(dot/state): enable checkpoint creation every 1m blocks

EclesioMeloJunior:
## Changes

- This PR is a proposal to introduce `pebble.Checkpoints`. Basically, pebble allows us to programmatically create a point-in-time view of the DB state and output it into a destination directory which can be used to re-start the node. The current configurations are hardcoded to execute a checkpoint every 1 million blocks, also the destination dir is `${base-path}/snapshot/`.

- To avoid data duplication whenever a checkpoint is created it renames the current `snapshot` folder to `snapshot_old` and creates a new `snapshot` with the most up-to-date data and then removes the `snapshot_old` concurrently. This was chosen given that every new snapshot will be greater than the previous so first removing to then creating the snapshot can take a longer time, so renaming and asynchronously removing the old folder can be faster

## Tests

WIP

## Issues

N/A

## Primary Reviewer

@dimartiro


jimjbrettj (commented):
I like the addition, but I am wondering if it would be nice to have a flag that either enables or disables this feature. I think there could be cases where we don't want to have to deal with snapshots. What do you think?

internal/database/pebble.go
```diff
@@ -152,3 +153,42 @@ func (p *PebbleDB) NewPrefixIterator(prefix []byte) Iterator {
...
+		go func() {
+			err := os.RemoveAll(pathToRemove)
+			if err != nil {
+				logger.Errorf("failing to remove %s: %w", err)
+			}
+		}()
+	}
+
+	err = p.db.Checkpoint(finalDest)
+	if err != nil {
```
dimartiro, line 190: If I'm not wrong we can lose snapshots if we delete the "snapshot_old" folder before check that the new snapshot was created.

I'd suggest to execute the deletion only if the new checkpoint was created successful, and if it fails we can rename the "_old" folder back to the previous name to keep being in a consistent state.

What do you think?

dot/state/block.go
```diff
@@ -48,6 +50,10 @@ var (
 	})
 )
 
+type Checkpointer interface {
```
dimartiro, line 53: Are you using this? 🤔

dot/state/block.go
```diff
@@ -27,6 +27,8 @@ import (
 
 const (
 	blockPrefix = "block"
+	// should create a check point every 1m blocks
+	checkpointEvery = 1_000_000
```
dimartiro, line 31:
```suggestion
	checkpointFrequency = 1_000_000
```

dot/state/block.go
```diff
@@ -937,3 +946,18 @@ func (bs *BlockState) StoreRuntime(hash common.Hash, rt runtime.Instance) {
...
+
+func (bs *BlockState) Checkpoint(header *types.Header) {
+	if header.Number < bs.nextCheckpoint {
+		return
+	}
+
+	logger.Infof("creating a checkpoint, latest finalized block %s (#%d)",
+		header.Hash(), header.Number)
+
+	bs.nextCheckpoint = header.Number + checkpointEvery
```
dimartiro, line 958: Why not?

```suggestion
	bs.nextCheckpoint += checkpointEvery
```

dimartiro (changes requested):
I added a few questions and suggestions.

Are you going to create a different PR to start using it when we start a node? 
Based on that, we maybe will need to update [this PR](https://github.com/ChainSafe/gossamer/pull/3251) too.

internal/database/pebble.go
EclesioMeloJunior: Agreed, changed to address your suggestion!

dot/state/block.go
EclesioMeloJunior: This way we don't cover when we start from a snapshot, let's say we start using a snapshot from the block `#2 000 000`, since we started the node `bs.nextCheckpoint` is zero and by `bs.nextCheckpoint += checkpointEvery` will be `1 000 000` so every new block it will take snapshot since the condition to take the snapshot is `header.Number >= bs.nextCheckpoint`

dimartiro: Ok I got the point but I think we should solve that case when we are setting the first `nextCheckpoint` instead using the `header.Number` here

dot/state/block.go
```diff
@@ -27,6 +27,8 @@ import (
 
 const (
 	blockPrefix = "block"
+	// should create a check point every 1m blocks
+	checkpointFrequency = 10000
```
dimartiro, line 31: Some 0's missing? 🤔 

```suggestion
	checkpointFrequency = 1000000
```

dimartiro (changes requested):

EclesioMeloJunior:
> I like the addition, but I am wondering if it would be nice to have a flag that either enables or disables this feature. I think there could be cases where we don't want to have to deal with snapshots. What do you think?

Yes, I plan to do that in a next PR since it will touch other files, but the checkpoint feature and the frequency should be configurable through flags/toml file

dimartiro (approved):

timwu20 (commented):
Do we want this ability in the development branch? We use the development branch for release.  I would assume pebble checkpointing uses a significant amount of i/o.  I think we should test on staging so we can measure the performance hit at 1M block heights before this is merged.

EclesioMeloJunior:
> Do we want this ability in the development branch?

The checkpoint/snapshot feature is a must to have IMO, we tested the checkpoint creation in staging and it worked well, even though I didn't collected the time spent to take those snapshots but AFAIK it took 20 seconds to take a snapshot from block 7m, also we tested starting a Gossamer node from a snapshot and it worked well but it took more than 1 hour to start the node (I would assume that pebble db take that time since it needs to go through the data and perform some pre-checks, but this long time was happening with badger db too)


> I think we should test on staging so we can measure the performance hit at 1M block heights before this is merged

Yeah, also I would like to have checkpoint as something configurable so I raised https://github.com/ChainSafe/gossamer/pull/3479

PR #3493, opened 2023-09-25
refactor(trie): Use trieDB as entry point to get / store tries

dimartiro:
## Changes

- [x] Added trieDB structure
- [x] Modified state trie to use trieDB instead using their own table / in memory tries separately 
- [x] Modified blockstate to use trieDB instead using in memory tries only

This also help us to pave the way to add a trie cache

## Tests

```sh
make test
```

## Primary Reviewer

@timwu20


dot/state/block_finalisation.go
```diff
@@ -190,9 +193,17 @@ func (bs *BlockState) deleteFromTries(lastFinalised common.Hash) error {
...
-	stateRootTrie := bs.tries.get(lastFinalisedHeader.StateRoot)
+
+	stateRootTrie, err := bs.trieDB.Get(lastFinalisedHeader.StateRoot)
+	if err != nil {
+		return fmt.Errorf("unable to retrieve state root trie, hash: %s, err: %s", bs.lastFinalised, err)
+	}
+
 	if stateRootTrie != nil {
-		bs.tries.delete(lastFinalisedHeader.StateRoot)
+		err := bs.trieDB.Delete(lastFinalisedHeader.StateRoot)
```
EclesioMeloJunior, line 203: can just delete it directly without getting it from the database?

dot/state/block_finalisation.go
```diff
@@ -254,7 +265,9 @@ func (bs *BlockState) handleFinalisedBlock(currentFinalizedHash common.Hash) err
 		// prune all the subchain hashes state tries from memory
```
EclesioMeloJunior, line 265: I think we can update the comment to `prune all the subchain hashes state tries from database`

EclesioMeloJunior (changes requested):

dot/state/block_finalisation.go
dimartiro: Good catch, I don't know if there is a reason to check it before trying to delete it, I'm going to remove that check

q9f:
are you still working on this @dimartiro?

can we get a `db` label @P1sar

dimartiro:
We can close it in favor of https://github.com/ChainSafe/gossamer/pull/3658

Issue #3487, opened 2023-09-19
runtime panicked with `Transaction has a bad signature` at `#8077849`


EclesioMeloJunior:
## Describe the bug

- Gossamer sync stalls at block `#8077849` with runtime panicking as shown in the image

<img width="1167" alt="image" src="https://github.com/ChainSafe/gossamer/assets/17255488/0934b31b-45f2-4577-b9ca-b96ad5999af9">

PR #3494, opened 2023-09-25
fix(lib/runtime): prevents polkadot zero-address bug using `sr25519_verify` version 1

EclesioMeloJunior:
## Changes

- Gossamer was failing with `Transaction has a bad signature` since `go-schnorrkel` fails when the public key is all zeroes. A fix was introduced in the runtime to avoid these kinds of public keys to sign extrinsic but the fix is on a later block. To circumvent this problem Gossamer checks the public key bytes, and if they're all zeroes it uses `ext_crypto_sr25519_verify_version_1` to verify the signature 

## Tests

```sh
go test -timeout 10m -run ^TestInstance_ExecuteBlock_WestendRuntime_WestendBlock8077850$ github.com/ChainSafe/gossamer/lib/runtime/wazero --tags=integration -v
```

## Issues

- Closes https://github.com/ChainSafe/gossamer/issues/3487

## Primary Reviewer

@timwu20


dimartiro (approved):
There are still some checks to fix but I'll approve it based on what we discussed the day we found this issue.

.gitattributes
```diff
@@ -1,4 +1,4 @@
 *.go -text diff=golang
 
 lib/runtime/test_data/kusama/ filter=lfs diff=lfs merge=lfs -text
-lib/runtime/test_data/polkadot/ filter=lfs diff=lfs merge=lfs -text
+lib/runtime/test_data/polkadot/ filter=lfs diff=lfs merge=lfs -text
```
q9f, line 4: Please configure your IDE to trim whitespaces and add a new line at the end of each file (for consistency).

lib/runtime/constants.go
```diff
@@ -22,6 +22,12 @@ const (
 	WESTEND_RUNTIME_V0929_FP  = "westend_runtime-v929.compact.wasm"
 	WESTEND_RUNTIME_V0929_URL = "https://github.com/paritytech/polkadot/releases/download/v0.9." +
 		"29/westend_runtime-v9290.compact.compressed.wasm?raw=true"
+
+	// v0.9.12 westend used for zero-address bug
+	WESTEND_RUNTIME_v0912     = "westend_runtime-v9111"
+	WESTEND_RUNTIME_V0912_FP  = "westend_runtime-v9111.compact.wasm"
+	WESTEND_RUNTIME_V0912_URL = "https://github.com/paritytech/polkadot/releases/download/v0.9." +
+		"11/westend_runtime-v9111.compact.compressed.wasm?raw=true"
```
q9f, line 30: Is there maybe a better way to format this string? Concatenating in the middle of the version string seems suboptimal.

jimjbrettj (approved):
lgtm

Issue #3415, opened 2023-08-01
Implement availability store


kishansagathiya:
## Issue summary

- this would be go translation of `polkadot/node/core/av-store` polkadot package
- See the implementers' guide: https://paritytech.github.io/polkadot-sdk/book/node/utility/availability-store.html

## Sub Issues
- [x] #3630 
- [x] #3629 
- [x] #3561 
- [x] #3628 
- [x] #3741

### Original issue notes:
- It should be able to perform following tasks in `AvailabilityStoreMessage` 
https://github.com/paritytech/polkadot/blob/153543b0c8c582e73f520e5c08cbe33bddfb5f69/node/subsystem-types/src/messages.rs#L418
- look here to see how those tasks are implemented in rust, https://github.com/paritytech/polkadot/blob/153543b0c8c582e73f520e5c08cbe33bddfb5f69/node/core/av-store/src/lib.rs#L1022

-  candidate backing subsystem generates available data, we have to be storing them using av-store (https://github.com/paritytech/polkadot/blob/153543b0c8c582e73f520e5c08cbe33bddfb5f69/node/core/backing/src/lib.rs#L559)

- this can be picked up immediately as this does not really have any dependencies.

kishansagathiya:
I feel the best way to go about this PR would be to create a skeleton PR like https://github.com/ChainSafe/gossamer/pull/3497

Since, availability store is also a subsystem, there are a lot of ideas that we can borrow from that PR.

edwardmack:
Closed since all containing issues are closed.

PR #3521, opened 2023-10-06
feat(parachain): availability store skeleton

edwardmack:
## Changes
- Create types for messaging to availability store subsystem
- stub functions for processing availability store messages

## Issues
starts work on #3415 

## Primary Reviewer

@kishansagathiya 
@timwu20


dot/parachain/availability-store/messages.go
```diff
@@ -0,0 +1,101 @@
+// Copyright 2023 ChainSafe Systems (ON)
+// SPDX-License-Identifier: LGPL-3.0-only
+
+package availability_store
```
kishansagathiya, line 4: These types don't have to be varying data types. 

We need to make a type varying data type when we want to scale encode or decode it, which happens when 1) we use that type in a message sent to other node or 2) we use that type as an argument in a runtime call.

That is not the case here. The way to confirm if a type gets scale encoded or decoded is by checking the `#[derive]` attribute. If a type needs to be encoded or decoded attribute should be `#[derive(Encode, Decode)]`. 
If you look at this type, Encode and Decode traits are missing.
https://github.com/paritytech/polkadot-sdk/blob/cb944dc54882222453071fc99f2e5014c51823ef/polkadot/node/subsystem-types/src/messages.rs#L477

dot/parachain/availability-store/messages.go
```diff
@@ -0,0 +1,101 @@
...
+}
+
+// Index returns the index of varying data type
+func (QueryAvailableData) Index() uint {
+	return 0
+}
+
+type QueryDataAvailability struct {
+	CandidateHash common.Hash
+	// TDDO: add oneshot sender
```
kishansagathiya, line 27: we might use a channel here to get a response right here, based on whether we need to use response right then and there. But we can safely remove this comment. Same goes for other instances of `// TDDO: add oneshot sender`

dot/parachain/availability-store/messages.go
```diff
@@ -0,0 +1,101 @@
...
+func (StoreAvailableData) Index() uint {
+	return 7
+}
+
+// NewCollationFetchingResponse returns a new collation fetching response varying data type
+func NewAvailabilityStoreMessage() AvailabilityStoreMessage {
+	vdt := scale.MustNewVaryingDataType(QueryAvailableData{}, QueryDataAvailability{}, QueryChunk{},
+		QueryChunkSize{}, QueryAllChunks{}, QueryChunkAvailability{}, StoreChunk{}, StoreAvailableData{})
+	return AvailabilityStoreMessage(vdt)
+}
```
kishansagathiya, line 99: Won't be needed after we make it a normal type instead of VDT

dot/parachain/availability-store/messages.go
edwardmack: Stubbed in channels for these senders and removed TODOs.

dot/parachain/availability-store/messages.go
edwardmack: Removed.

dot/parachain/availability-store/availabilitystore.go
```diff
@@ -0,0 +1,84 @@
...
+func (av *AvailabilityStoreSubsystem) handleQueryChunkAvailability(msg QueryChunkAvailability) {
+	// TODO: handle query chunk availability
+}
+
+func (av *AvailabilityStoreSubsystem) handleStoreChunk(msg StoreChunk) {
+	// TODO: handle store chunk
+}
+
+func (av *AvailabilityStoreSubsystem) handleStoreAvailableData(msg StoreAvailableData) {
+	// TODO: handle store available data
```
kishansagathiya, line 83: I am going to avoid creating issues for each of these TODOs, since in most of them we just have to make some db queries.
It make sense to take all these TODOs in a single PR than separately. I don't think that PR would be come too large.
We can leave #3415 open and close it after we tackle these TODOs in next PR.

kishansagathiya (approved):

axaysagathiya (approved):
LGTM

dot/parachain/availability-store/availabilitystore.go
edwardmack: Yes I agree, I'll start working on these.

PR #3536, opened 2023-10-13
feat/btree: add BTree with scale

### LINKED ISSUE #3480: Add btree support to scale
## Issue summary

- Add ability to marshall and unmarshall btrees to scale package

kanishkatn:
## Changes

Adds BTree with scale encoding and decoding apis by creating a wrapper around the [tidwall's library](https://github.com/tidwall/btree).

## Tests

```sh
go test pkg/scale
```

## Issues

Closes #3480

## Primary Reviewer

@timwu20


pkg/scale/btree.go
```diff
@@ -0,0 +1,33 @@
...
+
+	"github.com/tidwall/btree"
+)
+
+// BTree is a wrapper around tidwall/btree.BTree that also stores the comparator function and the type of the items
+// stored in the BTree. This is needed during decoding because the BTree is a generic type, and we need to know the
+// type of the items stored in the BTree in order to decode them.
+type BTree struct {
+	*btree.BTree
+	Comparator func(a, b interface{}) bool
```
edwardmack, line 17: I'm curious why BTree is storing Comparator?  I see needing it to create the new BTree, but I'm not sure I understand why it's  in the BTree struct.

edwardmack (commented):
I'm concerned about adding dependency on "github.com/tidwall/btree" to scale package, I think they is a way to define scale encoding/decoding outside of the scale package.

pkg/scale/btree.go
kanishkatn: You're right. We need it to decode a BTree. 

Since we are encoding just the data in BTree and not the comparator, we'll need a reference to it during the decoding in order to create a new `btree.BTree`. 

The flow for decoding would be
- Create a new instance of `scale.BTree` by passing the type of the objects in the `btree.BTree` and the comparator method
- Call `scale.Unmarshal` with the initialised `BTree` struct

I couldn't think of any other way of handling this and I am open to suggestions!

timwu20 (commented):
I'm curious as to where we need the `btree.BTree`? We utilize `btree.Map` pretty extensively in `feat/grandpa`.

kanishkatn:
> I'm curious as to where we need the `btree.BTree`? We utilize `btree.Map` pretty extensively in `feat/grandpa`.

Have you seen #3314?

kanishkatn:
> I'm concerned about adding dependency on "github.com/tidwall/btree" to scale package, I think they is a way to define scale encoding/decoding outside of the scale package.

I looked into moving `BTree` outside the scale package; in order to do that, we'll have to export `encodeState`, `decodeState` and `fieldScaleIndicesCache`.

Could you elaborate on your concerns?

pkg/scale/decode.go
```diff
@@ -108,6 +111,20 @@ type decodeState struct {
 }
 
 func (ds *decodeState) unmarshal(dstv reflect.Value) (err error) {
+	// Handle BTreeMap type separately for the following reasons:
+	// 1. BTreeMap is a generic type, so we can't use the normal type switch
+	// 2. We cannot use BTreeCodec because we are comparing the type of the dstv.Interface() in the type switch
+	if isBTree(dstv.Type()) {
```
EclesioMeloJunior, line 117: `isBtree` also checks for `BTree` and `BTreeMap` right? so is the switch case on line `150` needed?

pkg/scale/btree.go
```diff
@@ -0,0 +1,187 @@
...
+
+	return &BTree{
+		BTree:      btree.New(comparator),
+		Comparator: comparator,
+		ItemType:   elementType,
+	}
+}
+
+// BTreeMap is a wrapper around tidwall/btree.Map
+type BTreeMap[K Ordered, V any] struct {
```
EclesioMeloJunior, line 106: I believe you can use `constraints.Ordered` (https://pkg.go.dev/golang.org/x/exp/constraints#Ordered)

EclesioMeloJunior (changes requested):

pkg/scale/decode.go
kanishkatn: Oh yeah, I meant to remove that. Thank you!

pkg/scale/btree.go
```diff
@@ -0,0 +1,181 @@
...
+		BTree:      bt.BTree.Copy(),
+		Comparator: bt.Comparator,
+		ItemType:   bt.ItemType,
+	}
+}
+
+// NewBTree creates a new BTree with the given comparator function.
+func NewBTree[T any](comparator func(a, b any) bool) *BTree {
+	// There's no instantiation overhead of the actual type T because we're only creating a slice type and
+	// getting the element type from it.
```
EclesioMeloJunior, line 89: I think the comment is not needed anymore, right?
```suggestion
```

kanishkatn: Yes 🤦. Thank you!

edwardmack:
> > I'm concerned about adding dependency on "github.com/tidwall/btree" to scale package, I think they is a way to define scale encoding/decoding outside of the scale package.
> 
> I looked into moving `BTree` outside the scale package; in order to do that, we'll have to export `encodeState`, `decodeState` and `fieldScaleIndicesCache`.
> 
> Could you elaborate on your concerns?
It's just that this is the first time (that I'm aware of) that we've added outside packages to the scale package, so that piqued my curiosity.  I was more curious than concerned and now I have a better understanding.

jimjbrettj:
> > I'm curious as to where we need the `btree.BTree`? We utilize `btree.Map` pretty extensively in `feat/grandpa`.
> 
> Have you seen #3314?

@kanishkatn Could you please elaborate more on this? While I see that you use the `btree.BTree` I do not see any scale interactions with respect to `btree.BTree` so I am still left with the question of where we need scale to have `btree.BTree` support.

pkg/scale/decode.go
```diff
@@ -768,3 +783,43 @@ func (ds *decodeState) decodeUint128(dstv reflect.Value) (err error) {
...
+	itemTypeField, hasItemType := t.FieldByName("ItemType")
+
+	if hasMap &&
+		hasDegree &&
+		mapField.Type.Kind() == reflect.Ptr &&
+		strings.HasPrefix(mapField.Type.String(), "*btree.Map[") {
+		return true
+	}
+
+	if hasBTree &&
```
jimjbrettj, line 809: nit/maybe one line these

kanishkatn:
> > > I'm curious as to where we need the `btree.BTree`? We utilize `btree.Map` pretty extensively in `feat/grandpa`.
> > 
> > 
> > Have you seen #3314?
> 
> @kanishkatn Could you please elaborate more on this? While I see that you use the `btree.BTree` I do not see any scale interactions with respect to `btree.BTree` so I am still left with the question of where we need scale to have `btree.BTree` support.

Here's another one: #3344. It has a custom encoding right now; I might look into changing it once I implement pebbleDB. 

I hope that helps!

pkg/scale/decode.go
```diff
@@ -108,6 +109,20 @@ type decodeState struct {
 }
 
 func (ds *decodeState) unmarshal(dstv reflect.Value) (err error) {
+	// Handle BTreeMap type separately for the following reasons:
+	// 1. BTreeMap is a generic type, so we can't use the normal type switch
+	// 2. We cannot use BTreeCodec because we are comparing the type of the dstv.Interface() in the type switch
+	if isBTree(dstv.Type()) {
```
jimjbrettj, line 115: Nit/Perhaps can rename to something the describes that it checks for both `btree.Btree` as well as `btree.Map`?

jimjbrettj, line 115: Just for my own clarification, once we determine that it is either `btree.Btree` or `btree.Map` we decode in the same fashion?

pkg/scale/btree.go
```diff
@@ -0,0 +1,183 @@
...
+}
+
+// NewBTreeMap creates a new BTreeMap with the given degree.
+func NewBTreeMap[K constraints.Ordered, V any](degree int) BTreeMap[K, V] {
+	return BTreeMap[K, V]{
+		Map: btree.NewMap[K, V](degree),
+	}
+}
+
+var _ BTreeCodec = (*BTreeMap[int, string])(nil)
```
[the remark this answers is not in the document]
jimjbrettj: or perhaps include it near similar line on 97? Maybe I am misunderstanding what they are doing

jimjbrettj (commented):
Looks good to me, just have some questions and nits

jimjbrettj:
I am going to change the primary reviewer of this PR to @timwu20. Given that Tim is the author of this package and has the most knowledge of it I would like to double-check with him before merging this in.

pkg/scale/decode.go
kanishkatn: hmm, both are based on the fundamental B-tree structure. I feel using `isBTree()` encapsulates the idea that it is checking for a type that is based on the B-tree structure, whether it's the tree itself or a map that uses it.

kanishkatn: Yes, since both are implementing the same interface, it just calls `Decode()`

pkg/scale/btree.go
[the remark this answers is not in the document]
kanishkatn: It is a compile time assertion to ensure they've implemented the interface. I don't know if we wanna move it on top; keeping it at the end of respective implementations, it follows the logical flow and acts as a conclusion. 

It's akin to saying, "Having seen all the methods and fields of this type, now note that it's intended to implement this interface."

pkg/scale/decode.go
jimjbrettj: Makes sense to me!

kanishkatn: I'll update the comments to capture this. Thank you!

timwu20 (changes requested):
I think we should take a different approach to supporting the `tidwall/btree` package.  I propose implementing custom unmarshalling/marshalling support in `pkg/scale` (#3558) first, and then creating a thin wrapper around the `tidwall/btree` package that implements the `Unmarshaller` and `Marshaller` interfaces for the types that we need.  We should make it clear that we're just doing this to support SCALE for the `btree` types. This should also reduce the changes required within `scale` to support the `btree` types.

kanishkatn:
I've updated this to use the changes from #3617

pkg/btree/btree.go
```diff
@@ -0,0 +1,187 @@
...
+	MarshalSCALE() ([]byte, error)
+	UnmarshalSCALE(reader io.Reader) error
+}
+
+// Tree is a wrapper around tidwall/btree.BTree that also stores the comparator function and the type of the items
+// stored in the BTree. This is needed during decoding because the Tree item is a generic type, and we need to know it
+// at the time of decoding.
+type Tree struct {
+	*btree.BTree
+	Comparator func(a, b interface{}) bool
```
timwu20, line 28: Not sure why you need this attribute given it's not used in any of the `MarshalSCALE` or `UnmarshalSCALE` methods.

timwu20 (changes requested):

pkg/btree/btree.go
kanishkatn: Please find the follow up questions below to get a better understanding.

Do you mean the comparator? Is the comment not descriptive enough?

pkg/btree/btree.go
```diff
@@ -0,0 +1,187 @@
...
+	}
+}
+
+var _ Codec = (*Tree)(nil)
+
+// Map is a wrapper around tidwall/btree.Map
+type Map[K constraints.Ordered, V any] struct {
+	*btree.Map[K, V]
+	Degree int
+}
```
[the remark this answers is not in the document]
kanishkatn: It is so that the user can set the degree beforehand and not worry about filtering the data after unmarshal.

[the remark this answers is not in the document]
kanishkatn: Let me rephrase. It gives the option for the user to set the Degree. One usecase that I see it supporting is, if there are 100 entries in the db and I care about only top 10, I can set that during the creation.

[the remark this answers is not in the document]
kanishkatn: But yeah, I guess I don't need to store it. 

I'm not familiar with your implementation style, feels like a lot of redundant code.

Issue #3509, opened 2023-10-03
Westend crashes with `wasm error: out of bounds memory access` at `#9412261`


EclesioMeloJunior:
## Describe the bug

```
September 28, 2023 at 17:28 (UTC-4:00) | 2023-09-28T21:28:27Z ERROR requesting max blocks from best block header: while handling workers results: while handling ready block: processing block data with header and body: handling block: failed to execute block 9412261: running runtime function: wasm error: out of bounds memory access 
September 28, 2023 at 17:28 (UTC-4:00) | wasm stack trace: 
September 28, 2023 at 17:28 (UTC-4:00) | .<memory_db::MemoryDB<H,KF,T,M> as core::default::Default>::default::h035979c933311a62(i32)
September 28, 2023 at 17:28 (UTC-4:00) | .pallet_session::historical::ProvingTrie<T>::generate_for::h1bf55405a66df38a(i32,i32) 
September 28, 2023 at 17:28 (UTC-4:00) | .<(TupleElement0,TupleElement1) as frame_support::traits::hooks::OnInitialize<BlockNumber>>::on_initialize::h8fe686fce2635624(i32) i64
September 28, 2023 at 17:28 (UTC-4:00) | .frame_executive::Executive<System,Block,Context,UnsignedValidator,AllPalletsWithSystem,COnRuntimeUpgrade>::initialize_block::h413dbea923e2629a(i32)
September 28, 2023 at 17:28 (UTC-4:00) | .Core_execute_block(i32,i32) i64 pkg=sync 
September 28, 2023 at 17:28 (UTC-4:00) | 2023-09-28T21:28:27Z INFO 🚣 currently syncing, 16 peers connected, 16 available workers, target block number 17655067, finalised block number 9412096 with hash 0x1e5b932b5aabb083b452b64a01ee7be1306b5a432a6b74bc9c5bdf8f074e3546 pkg=sync
September 28, 2023 at 17:28 (UTC-4:00) | 2023-09-28T21:28:20Z WARN failed to advertise in the DHT: failed to find any peer in table pkg=network 
September 28, 2023 at 17:28 (UTC-4:00) | 2023-09-28T21:28:04Z INFO 🔨 finalised block number 9412096 with hash 0x1e5b932b5aabb083b452b64a01ee7be1306b5a432a6b74bc9c5bdf8f074e3546 pkg=sync
[...]

PR #3570, opened 2023-11-08
feat(lib/allocator): Refactor `FreeingBumpHeapAllocator`

EclesioMeloJunior:
## Changes

- Refactor our current allocator to match the substrate allocator
- ref: https://github.com/paritytech/polkadot-sdk/blob/master/substrate/client/allocator/src/freeing_bump.rs

## Tests

```sh
go test -timeout 10m github.com/ChainSafe/gossamer/lib/runtime/allocator --tags=integration -v
```

## Issues

- Related to https://github.com/ChainSafe/gossamer/issues/3509

## Primary Reviewer

@timwu20


lib/runtime/allocator/freeing_bump.go
```diff
@@ -0,0 +1,604 @@
...
+
+	res := *bumper
+	*bumper += size
+	return res, nil
+}
+
+// pagesFromSize convert the given `size` in bytes into the number of pages.
+// The returned number of pages is ensured to be big enough to hold memory
+// with the given `size`.
+// Returns false if the number of pages do not fit into `u32`
```
kanishkatn, line 573:
```suggestion
// Returns false if the number of pages do not fit into `uint32`
```

lib/runtime/allocator/freeing_bump.go
```diff
@@ -0,0 +1,586 @@
...
+	addressSpaceUsedGague = promauto.NewGauge(prometheus.GaugeOpts{
+		Namespace: "gossamer_allocator",
+		Name:      "address_space_used",
+		Help: "the amount of address space (in bytes) used by the allocator this is calculated as " +
+			"the difference between the allocator's bumper and the heap base.",
+	})
+)
+
+var (
+	ErrInvalidOrder                 = errors.New("invalid order")
```
jimjbrettj, line 63: do these need to be exported?

lib/runtime/allocator/freeing_bump.go
```diff
@@ -0,0 +1,586 @@
...
+		if requiredPages > MaxWasmPages {
+			return 0, fmt.Errorf("%w: required pages %d greater than max wasm pages %d",
+				ErrAllocatorOutOfSpace, requiredPages, MaxWasmPages)
+		}
+
+		// ideally we want to double our current number of pages,
+		// as long as it's less than the double absolute max we can have
+		nextPages := min(currentPages*2, MaxWasmPages)
+		// ... but if even more pages are required then try to allocate that many
+		nextPages = max(nextPages, requiredPages)
```
jimjbrettj, line 533: could this hit a case where required pages > MaxWasmPages? Would this be addressed by the grow call below?

jimjbrettj (commented):
lgtm, just a few comments and will wait for tests to be fixed before approving

lib/runtime/allocator/freeing_bump.go
EclesioMeloJunior: I would say yes, since I moved the allocator from `lib/runtime` to `lib/allocator`, now the allocator methods like `Allocate` and `Deallocate` can return those errors and we can use them to assert on tests in the `runtime` pkg

lib/runtime/allocator/freeing_bump.go
EclesioMeloJunior: No, because on line 524 there is the following check

```go
if requiredPages > MaxWasmPages {
	return 0, fmt.Errorf("%w: required pages %d greater than max wasm pages %d",
		ErrAllocatorOutOfSpace, requiredPages, MaxWasmPages)
}
```

lib/runtime/allocator/freeing_bump.go
```diff
@@ -0,0 +1,586 @@
...
+	// because in that case we would have easier time to provide the
+	// guaranteed alignment of 8
+	//
+	// The maximum possible allocation size is set to 32Mib
+	//
+	// NumOrders represents the number of orders supported, this number
+	// corresponds to the number of powers between the minimum an maximum
+	// possible allocation (2^3 ... 2^25 both ends inclusive)
+	NumOrders              uint32 = 23
+	MinPossibleAllocations uint32 = 8
```
dimartiro, line 35: Can we change the way we define this to match with the doc and with the max?

```suggestion
	MinPossibleAllocations uint32 = (1 << 3)
```

EclesioMeloJunior: I prefer to keep it as it is, it is easier to understand values. For example, we don't like when we look at substrate docs/code and see `0x10000` to represent a `65536` (I mean, makes the reader life harder)

dimartiro: Agree, I just asked it to keep some consistency since we are setting `MaxPossibleAllocations` as `(1 << 25)` representing the `2^25` instead setting the final value `33554432` but we are setting the value in `MinPossibleAllocations` instead doing the same `2^3` but obviously it's not a must just to keep consistency 👍

EclesioMeloJunior: yeah, I changed `MaxPossibleAllocations` to a raw uint32 value instead of a bitwise op to keep consistency

jimjbrettj (approved):
Nice work!

dimartiro (approved):

Issue #3577, opened 2023-11-13
some of the documentation are outdated or no longer exist


1garo:
## Describe the bug

- `--base-path` command is outdated in the documentation.
- `gossamer account generate --ed25519` command don't exist anymore, it's all under `--scheme` now;

## Current Behavior

![image](https://github.com/ChainSafe/gossamer/assets/44412643/335aa05d-a47b-4dcd-a7fb-688778925989)

## Possible Solution

It's correct in the global flags, just need to adjust the documentation, if it's okay I could patch a PR for it.

## To Reproduce
Steps to reproduce the behaviour:

1. gossamer init --chain westend --key alice --basepath
2. gossamer account generate --ed25519

1garo:
There is a similar problem at: `gossamer account generate --ed25519`, I should open a separated PR for it?
![image](https://github.com/ChainSafe/gossamer/assets/44412643/6a550f9e-230c-4dec-be64-844a1dcc0c68)

kanishkatn:
> There is a similar problem at: `gossamer account generate --ed25519`, I should open a separated PR for it? ![image](https://user-images.githubusercontent.com/44412643/282525381-6a550f9e-230c-4dec-be64-844a1dcc0c68.png)
> 
> I think that the command don't exist anymore, it is under `--scheme` now.

You're right! I think it's better to include it in the same PR. Maybe just update the issue to capture both?

PR #3579, opened 2023-11-13
chore/cli: update outdated documentation

1garo:
## Changes

Closes https://github.com/ChainSafe/gossamer/issues/3577

- Adjust documentation to match CLI flag

## Issues

#3577 

@kanishkatn


cmd/gossamer/commands/account.go
```diff
@@ -24,15 +24,13 @@ var AccountCmd = &cobra.Command{
 	Use:   "account",
 	Short: "Create and manage node keystore accounts",
 	Long: `The account command is used to manage the gossamer keystore.
-Examples: 
+Examples:
 
-To generate a new sr25519 account: 
-	gossamer account generate --keystore-path=path/to/location --scheme=sr25519
-To generate a new ed25519 account: 
-	gossamer account generate --ed25519
```
1garo, line 32: I took the freedom to change the existing example to use `ed25519` instead of `sr25519` since it's the default one, but if it don't make sense I just rollback and remove the ed25519 command

cmd/gossamer/commands/account.go
```diff
@@ -24,15 +24,13 @@ var AccountCmd = &cobra.Command{
...
-To generate a new secp256k1 account: 
+To generate a new ed25519 account:
+	gossamer account generate --keystore-path=path/to/location --scheme=ed25519
```
1garo, line 30: I took the freedom to change the existing example to use ed25519 instead of sr25519 since the later it's the default one, but if it don't make sense I just rollback and remove the command.

EclesioMeloJunior (approved):

dimartiro (approved):
Thanks for your contribution @1garo

cmd/gossamer/commands/account.go
1garo: I think that `cobra` already does that, probably because `crypto.Sr25519Type` it's passed instead of a empty string, for example, in the other commands it's not and it does not output `defaults ...`.

image

1garo: I confirmed, the second argument of `Flags().String()` it's the default value.

kanishkatn (approved):

kanishkatn:
Thank you @1garo . It looks like you need to sign the CLA; we should be able to merge it after that!

1garo:
> Thank you @1garo . It looks like you need to sign the CLA; we should be able to merge it after that!

I signed the CLA, but I don't know if I need to do something else, and I notice that my user.email was wrong, so I solved that as well.

cmd/gossamer/commands/account.go
kanishkatn: Thank you for being thorough!

Issue #3255, opened 2023-05-11
cli: generate account key if not specified


kanishkatn:
## Issue summary

If the account key is not specified in the cli flags, we need to generate one for the user.

danforbes:
What is the current behavior?

1garo:
wanted to contribute to this issue, is it okay?

1garo:
> Of course @1garo ! I am off at the moment and can provide more context in a couple of days!

Sounds great, in the meantime I'm gonna take a look if I can have some progress.

kanishkatn:
Here's where the account is being parsed: https://github.com/ChainSafe/gossamer/blob/development/cmd/gossamer/commands/utils.go#L322

With the current [default config setup](https://github.com/ChainSafe/gossamer/blob/development/chain/westend/defaults.go), we won't have a scenario where the account would be empty. We want to remove the default values for account and generate one it if it is empty.

Happy to answer any questions you have. If you're on ChainSafe discord, say hi in #gossamer for easy collab!

1garo:
I was able to generate the `default` keypair, I added a "default" case to `LoadKeystore`, and used `ed25519.GenerateKeypair()` and `sr25519.GenerateKeypair()` to generate the pairs (`ed25519` for grandpa).

It passes but later on broke on: 

```
2023-11-16T13:56:37-03:00 CRITICAL failed to run block production engine: cannot handle epoch: cannot initiate and get epoch handler: failed to initiate epoch: cannot get epoch data and start slot: cannot get latest epoch data: cannot get authority index: key not in BABE authority data	pkg=babe
```

1garo:
Seems like that by generating this new pair, their public key is not understand as a Authority, because it fails in the `getAuthorityIndex` function in the following code:

```go
	for i, auth := range Authorities {
		if bytes.Equal(pub.Encode(), auth.Key.Encode()) {
			return uint32(i), nil
		}
	}
```

kanishkatn:
Hey @1garo , could you create a draft PR? We could move the conversation there!

1garo:
Of course, it's created: https://github.com/ChainSafe/gossamer/pull/3587

kanishkatn:
I believe we don't have the need to generate an account if not specified since we're still in development and only need development accounts. @EclesioMeloJunior could you confirm?

Further, If we want to support the use of auto-generated or external accounts, we'll have to
- Add a new flag dev to enable the clear separation for the development runs
[...]

PR #3587, opened 2023-11-17
feat(cli): generate account key if not specified

1garo:
## Changes

- If user don't specifcy account key in the cli flags, we generate.

## Tests

```sh
go run cmd/gossamer/main.go --chain westend-dev
```

## Issues

#3255 

## Primary Reviewer

@kanishkatn


lib/keystore/helpers.go
```diff
@@ -148,6 +148,22 @@ func LoadKeystore(key string, keyStore TyperInserter, keyRing KeyRing) (err erro
...
+
+			return keyStore.Insert(kp)
+		} else {
+			kp, err := sr25519.GenerateKeypair()
+			if err != nil {
+				return err
+			}
+
+			return keyStore.Insert(kp)
+		}
```
kanishkatn, line 166: We currently only have the test key flow in the CLI. Check https://github.com/1garo/gossamer/blob/feat/generate-account-if-not-specified/cmd/gossamer/commands/root.go#L577

I feel we should separate out the flow for the keystore setup in the `cmd` package instead of this change. The rust implementation has `--dev` flag; maybe we could add the same!

kanishkatn (commented):
I'm still looking into the error!

1garo:
> I'm still looking into the error!

I managed to make the error go away adding keyRing.Default() and reusing alice's grandpa priv key on `ed25519PrivateKeys`, so I feel that is something related to that, maybe we should be able to add this new key to babe authority, something like that

lib/keystore/helpers.go
1garo: you mean something like [subkey?](https://docs.substrate.io/reference/command-line-tools/subkey/)

1garo: Maybe `account generate` already do that, I don't know if I really understood what you meant

kanishkatn: I spent some more time going through the code. Thank you for your patience and time!

I believe we don't have the need to generate an account if not specified since we're still in development and only need development accounts.

If we want to support the use of auto-generated or external accounts, we'll have to
- Add a new flag `dev` to enable the clear separation for the development runs
- Clear separation for the development config
- Store the auto generated account in disk
- Ability to load the account from disk

I think this isn't a priority for us at the moment. I'll update the issue.

EclesioMeloJunior:
@kanishkatn regards this I think we don't need to introduce the `--dev` flag to separate development runs, actually it is possible to use development keys, such as `alice` or `bob` in any environment, so we should be able to use `auto-generated` in any environment also.

The following command should generate an account for me
```
./bin/gossamer --chain westend
```

But this following command should not generate an account but use alice test account instead
```
./bin/gossamer --chain westend --alice
```

I agree with these 2 specifications:
- Store the auto-generated account in the disk
- Ability to load the account from disk

kanishkatn:
@EclesioMeloJunior I agree. Thank you!

q9f:
@1garo do you want to continue working on this or shall we take this over?

1garo:
I definitely can continue working on it!

dimartiro:
Hey @1garo, how are the things going? Are you planning to continue working on this PR or can we close it as stale?

Let us know if you need any help

Thks!

1garo:
> Hey @1garo, how are the things going? Are you planning to continue working on this PR or can we close it as stale?
> 
> Let us know if you need any help
> 
> Thks!

Sorry for the delayed response, I will not be able to work on the PR.

Feel free to close or if someone wants to get the issue it's okay for me.

dimartiro:
> > Hey @1garo, how are the things going? Are you planning to continue working on this PR or can we close it as stale?
> > Let us know if you need any help
> > Thks!
> 
> Sorry for the delayed response, I will not be able to work on the PR.
> 
> Feel free to close or if someone wants to get the issue it's okay for me.

Thanks @1garo I'll close it for now.
Thanks for your contribution and I'm looking forward to see you around again.
Feel free to take any other issue or resume this one in the future 💪

PR #3588, opened 2023-11-17
fix(lib/runtime): Fix `wasm error: out of bounds memory access` at `#9412261`

EclesioMeloJunior:
## Changes

- Setup the wasm runtime pages to the max value (math.MaxUint32, as substrate does)
- Upgrading Wazero from 1.1.0 to 1.5.0

## Tests

Staging node was able to process/import block #9412261 (`wasm error: out of bounds memory access`) and keep syncing

image

## Issues

- Closes https://github.com/ChainSafe/gossamer/issues/3509

## Primary Reviewer

@timwu20


lib/runtime/wazero/instance.go
```diff
@@ -107,7 +107,7 @@ func NewInstance(code []byte, cfg Config) (instance *Instance, err error) {
 
 	_, err = rt.NewHostModuleBuilder("env").
 		// values from newer kusama/polkadot runtimes
-		ExportMemory("memory", 23).
+		ExportMemory("memory", allocator.MaxWasmPages-1).
```
dimartiro, line 110: Why the `-1` ?

dimartiro (approved):

lib/runtime/wazero/instance.go
EclesioMeloJunior: The substrate wasm runtime supports u64 but wazero don't. So the max wasm pages value is `65536` and each page has `65536` bytes if you multiply the total bytes will be `4294967296` which overflows the uint32 by one, so I setup the `minPages` to be `65535`

dimartiro: Great! thanks for the explanation!

jimjbrettj (approved):

timwu20 (approved):

Issue #3582, opened 2023-11-15
feat(client/grandpa): Implement finality proof logic


jimjbrettj:
## Issue summary

- Implement the logic to prove finality within the grandpa client
- Write unit tests 

## Other information and links

- https://github.com/paritytech/polkadot-sdk/blob/master/substrate/client/consensus/grandpa/src/finality_proof.rs

jimjbrettj:
Closed by: #3589

PR #3589, opened 2023-11-17
client(consensus/grandpa): implement finality proof logic

jimjbrettj:
## Changes

- Implement and test logic for creating and verifying grandpa finality proofs

## Tests

```sh
go test ./client/consensus/grandpa
```

## Issues

#3582

## Primary Reviewer

@timwu20


client/consensus/grandpa/blockOrigin.go
```diff
@@ -0,0 +1,70 @@
+package grandpa
```
EclesioMeloJunior, line 1: This file nomenclature pattern is different from others, I would suggest renaming to `block_origin.go`

client/consensus/grandpa/blockOrigin.go
```diff
@@ -0,0 +1,70 @@
...
+
+import "github.com/ChainSafe/gossamer/pkg/scale"
+
+/*
+	Following is from primitives/consensus/common
+*/
+
+// BlockOrigin Block data origin
+type BlockOrigin scale.VaryingDataType
```
EclesioMeloJunior, line 10: Does the block origin need to be encoded/decoded?

EclesioMeloJunior (changes requested):

client/consensus/grandpa/blockOrigin.go
jimjbrettj: Actually, with the changes I had made this type is no longer used, so I am removing it

timwu20 (changes requested):

client/consensus/grandpa/finality_proof.go
```diff
@@ -0,0 +1,245 @@
...
+
+const maxUnknownHeaders = 100_000
+
+// FinalityProofProvider Finality proof provider for serving network requests.
+type FinalityProofProvider[
+	BE Backend[Hash, N, H, B],
+	Hash constraints.Ordered,
+	N constraints.Unsigned,
+	AuthID AuthorityID,
+	S comparable,
```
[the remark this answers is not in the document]
jimjbrettj: S is the signature type

client/consensus/grandpa/finality_proof.go
```diff
@@ -0,0 +1,245 @@
...
+			encJustification, err = scale.Marshal(*justification)
+			if err != nil {
+				return nil, err
+			}
+			justBlock = justification.Target().number
+		} else {
+			logger.Trace("No justification found for the latest finalized block. Returning empty proof")
+			return nil, nil
+		}
+	case set[N]:
```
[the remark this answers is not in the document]
jimjbrettj: I like this. Good call

client/consensus/grandpa/interfaces.go
```diff
@@ -10,13 +10,134 @@ import (
...
+type Info[N constraints.Unsigned] struct {
+	FinalizedNumber N
+}
+
+// BlockchainBackend Blockchain database backend. Does not perform any validation.
+// pub trait Backend<Block: BlockT>:HeaderBackend<Block> + HeaderMetadata<Block, Error = Error
+// primitives/blockchains/src/backend
+type BlockchainBackend[Hash constraints.Ordered, N constraints.Unsigned, H Header[Hash, N]] interface {
+	HeaderBackend[Hash, N, H]
+	Justifications(hash Hash) Justifications
```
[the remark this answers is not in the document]
jimjbrettj: Good find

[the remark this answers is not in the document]
jimjbrettj: updated to return an option and an error to represent result

client/consensus/grandpa/justification.go
```diff
@@ -48,7 +48,7 @@ func NewJustificationFromCommit[
 	H Header[Hash, N]](
```
timwu20, line 48: You shouldn't need to have to use a generic parameter for the type `Header`. Header can be defined inside the struct as `Header[Hash, N]`.

timwu20 (changes requested):

client/consensus/grandpa/authorities.go
```diff
@@ -725,27 +725,28 @@ func (asc *authoritySetChangeID) Value() (val scale.VaryingDataTypeValue, err er
 }
 
 func newAuthoritySetChangeID[N constraints.Unsigned]() authoritySetChangeID {
-	vdt := scale.MustNewVaryingDataType(latest{}, set[N]{}, unknown{})
+	vdt := scale.MustNewVaryingDataType(authoritySetChangeIDLatest{}, authoritySetChangeIDSet[N]{},
```
EclesioMeloJunior, line 728: Since those types don't need to be scale encoded/decoded I would suggest that you use interfaces to represent enums, using interfaces brings advantages like, 1) not using reflection overhead, 2) is error prune and you will be bound since the concrete type will be bound to the interface

```go
type AuthoritySetChange interface {
    isAuthoritySetChange()
}


type Latest struct {}
func (Latest) isAuthoritySetChange() {}

type Set[N constraints.Unsigned] struct {
    inner setIdNumber[N]
}
func (Set) isAuthoritySetChange() {}

type Unknown struct {}
func (Unknown) isAuthoritySetChange() {}
```

and instead of having `authSetChangeID := newAuthoritySetChangeID[N]()` you just create a variable like this `var authSetChangeID AuthoritySetChange`, and instead of calling the `.Set(...)` and checking the error you can just `authSetChangeID = Latest{}`, `authSetChangeID = Set{inner: ...}` or `authSetChangeID = Unknown{}`

EclesioMeloJunior (changes requested):

client/consensus/grandpa/justification.go
jimjbrettj: To clarify, do you mean instead of using generics for  `Header ` I set the type of VoteAncestries to use `Header[Hash, N]`?

Since we do not have a concrete `Header` type in grandpa and we use an interface, this would mean setting the expected type to be an interface. I think this would be okay except that we need to be able to encode/decode the `GrandpaJustification` to the db, which from my experience does not work in this manner. 

However, if there is a way to do this that I am not aware of let me know and I can make this change.

timwu20: Encoding an interface should work.  SCALE uses `reflect.ValueOf` which gives the underlying the type ([playground ex.](https://go.dev/play/p/mklaCRumc_3)).

client/consensus/grandpa/authorities.go
```diff
@@ -705,63 +704,33 @@ func (asc *AuthoritySetChanges[N]) append(setID uint64, blockNumber N) {
...
 	}
-	authSetChangeID = newAuthoritySetChangeID[N]()
 	authSet := *asc
 	last := authSet[len(authSet)-1]
 	if last.BlockNumber < blockNumber {
-		err = authSetChangeID.Set(latest{})
-		if err != nil {
-			return authSetChangeID, err
-		}
+		authSetChangeID = authoritySetChangeIDLatest{}
```
EclesioMeloJunior, line 733:
```suggestion
		return authoritySetChangeIDLatest{}, nil
```

client/consensus/grandpa/authorities.go
```diff
@@ -786,25 +755,17 @@ func (asc *AuthoritySetChanges[N]) getSetID(blockNumber N) (authSetChangeID auth
 
 		// if this is the first index but not the first set id then we are missing data.
 		if idx == 0 && authChange.SetID != 0 {
-			err = authSetChangeID.Set(unknown{})
-			if err != nil {
-				return authSetChangeID, err
-			}
+			authSetChangeID = authoritySetChangeIDUnknown{}
```
EclesioMeloJunior, line 758:
```suggestion
			return authoritySetChangeIDUnknown{}, nil
```

client/consensus/grandpa/authorities.go
```diff
@@ -786,25 +755,17 @@ func (asc *AuthoritySetChanges[N]) getSetID(blockNumber N) (authSetChangeID auth
...
 			return authSetChangeID, nil
 		}
-		err = authSetChangeID.Set(set[N]{
+
+		authSetChangeID = authoritySetChangeIDSet[N]{
```
EclesioMeloJunior, line 762:
```suggestion
		return authoritySetChangeIDSet[N]{authChange}, nil
```

client/consensus/grandpa/authorities_test.go
```diff
@@ -1205,36 +1205,30 @@ func TestCleanUpStaleForcedChangesWhenApplyingStandardChangeAlternateCase(t *tes
...
-	require.NoError(t, err)
-	switch val := authSetVal.(type) {
-	case set[uint]:
+	switch val := authSetID.(type) {
+	case authoritySetChangeIDSet[uint]:
 		require.Equal(t, expected, val.inner)
 	default:
-		err = fmt.Errorf("invalid authSetID type")
+		err := fmt.Errorf("invalid authSetID type")
+		require.NoError(t, err)
```
kanishkatn, line 1213: nice!

jimjbrettj: This is kinda a funny one, but open to suggestions if there is a better way to do this!

jimjbrettj:
> Also, I noticed that the function `checkMessageSignature` can receive a signature that can be `[]byte` or `string`, but why not set the signature type to be `[]byte` always, and then let the caller to transform the `string` to `[]bytes`?

This is a good question. It is because in the code where this function is called the type is a generic signature type S. Perhaps we could still put logic of casting on the caller, but I think it is probably easier to handle it in the function itself.

client/consensus/grandpa/authorities.go
```diff
@@ -705,64 +704,33 @@ func (asc *AuthoritySetChanges[N]) append(setID uint64, blockNumber N) {
...
+type authoritySetChangeIDUnknown struct{}
 
-func (unknown) Index() uint {
-	return 2
-}
+func (authoritySetChangeIDUnknown) isAuthoritySetChangeID() {}
 
 // Three states that can be returned: Latest, Set (tuple), Unknown
-func (asc *AuthoritySetChanges[N]) getSetID(blockNumber N) (authSetChangeID authoritySetChangeID, err error) {
+func (asc *AuthoritySetChanges[N]) getSetID(blockNumber N) (authoritySetChangeID, error) {
```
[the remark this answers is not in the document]
jimjbrettj: There is still one case where an error is returned (if asc is nil)

client/consensus/grandpa/justification.go
jimjbrettj: With the work done in https://github.com/ChainSafe/gossamer/pull/3617 I was able to use the custom decoding helper function to make this work

client/consensus/grandpa/finality_proof.go
```diff
@@ -0,0 +1,248 @@
...
+// requested block until the block the justification refers to.
+func (provider FinalityProofProvider[BE, Hash, N, AuthID, S, ID, H, B]) proveFinalityProof(
+	block N,
+	collectUnknownHeaders bool) (*FinalityProof[Hash, N, H], error) {
+	if provider.sharedAuthoritySet == nil {
+		return nil, nil
+	}
+
+	// TODO address lock copy
+	authoritySetChanges := *provider.sharedAuthoritySet //nolint
```
timwu20, line 113: You need to address this.  We shouldn't be copying types that contain mutexes.

client/consensus/grandpa/finality_proof.go
```diff
@@ -0,0 +1,248 @@
...
+	N constraints.Unsigned,
+	S comparable,
+	ID AuthorityID,
+	H Header[Hash, N],
+	B BlockchainBackend[Hash, N, H],
+](
+	backend BE,
+	authSetChanges AuthoritySetChanges[N],
+	block N,
+	collectUnknownHeaders bool) (*FinalityProof[Hash, N, H], error) {
```
timwu20, line 151: nit:
```suggestion
	collectUnknownHeaders bool,
) (*FinalityProof[Hash, N, H], error) {
```

client/consensus/grandpa/finality_proof_test.go
```diff
@@ -0,0 +1,506 @@
...
+	Hash constraints.Ordered,
+	N constraints.Unsigned,
+	S comparable,
+	H Header[Hash, N],
+	ID AuthorityID,
+](
+	t *testing.T,
+	currentSetID uint64,
+	currentAuthorities AuthorityList[ID],
+	remoteProof []byte) (FinalityProof[Hash, N, H], error) {
```
timwu20, line 30: nit: easier to read when you add the trailing comma on the last param.

client/consensus/grandpa/finality_proof_test.go
```diff
@@ -0,0 +1,506 @@
...
+	t.Helper()
+	proof := FinalityProof[Hash, N, H]{}
+	err := scale.Unmarshal(remoteProof, &proof)
+	if err != nil {
+		return FinalityProof[Hash, N, H]{}, fmt.Errorf("failed to decode finality proof %s", err)
```
timwu20, line 35: I don't think you need to return an error from a helper function.  Probably makes more sense to just `t.Error()` or `t.Fail()`

client/consensus/grandpa/finality_proof_test.go
```diff
@@ -0,0 +1,506 @@
...
+	mockBackend.EXPECT().Get(Key("grandpa_best_justification")).Return(nil, nil).Times(1)
+
+	// No recent authority set change, so we are in the authoritySetChangeIDLatest set, and we will try to pickup
+	// the best stored justification, for which there is none in this case.
+	authoritySetChanges := AuthoritySetChanges[uint]{}
+	authoritySetChanges.append(0, 5)
+
+	/*
+		TODO For reviewer: When justification is not stored in db, substrate returns None(), but we return error.
+		I think I like error, and we can handle accordingly, but will note this for reviewer feedback
```
timwu20, line 418: I don't think we should deviate.  Using a pointer and returning `nil` is valid IMO.  I don't think it's an error if a db is missing a value.

client/consensus/grandpa/justification.go
```diff
@@ -27,15 +28,66 @@ var (
...
+// Type used for decoding grandpa justifications (can pass in generic Header type)
+type decodeGrandpaJustification[
+	Hash constraints.Ordered,
+	N constraints.Unsigned,
+	S comparable,
+	ID AuthorityID,
+	H Header[Hash, N],
+] GrandpaJustification[Hash, N, S, ID]
+
+func DecodeGrandpaJustification[Hash constraints.Ordered,
```
timwu20, line 51: this shouldn't be a public function.

client/consensus/grandpa/justification.go
```diff
@@ -27,15 +28,66 @@ var (
...
+	S comparable,
+	ID AuthorityID,
+	H Header[Hash, N],
+](encodedJustification []byte) (*GrandpaJustification[Hash, N, S, ID], error) {
+	newJustificaiton := decodeGrandpaJustification[Hash, N, S, ID, H]{}
+	err := scale.Unmarshal(encodedJustification, &newJustificaiton)
+	if err != nil {
+		return nil, err
+	}
+	return newJustificaiton.GrandpaJustification(), nil
```
timwu20, line 62: I thought you were going to put this in `GrandpaJustfication.UnmarshalSCALE()`.

timwu20 (changes requested):

client/consensus/grandpa/justification.go
jimjbrettj: I'm not really sure what you mean by this. The `decodeJustification` is just there to remove the need to interact with `decodeGrandpaJustification` from the caller since its only purpose is for decoding and I felt making the caller deal with it could lead to confusion.

I'm not sure how I would add `GrandpaJustification()` to `GrandpaJustfication.UnmarshalSCALE()` since this function has no context of an outer `GrandpaJustification` type.

I might be misunderstanding what you are suggesting here, so if I am could you elaborate more or suggest how to overcome what I said above?

client/consensus/grandpa/authorities.go
```diff
@@ -705,64 +704,33 @@ func (asc *AuthoritySetChanges[N]) append(setID uint64, blockNumber N) {
...
-	}
-	*asc = authoritySetChangeID(vdt)
-	return
-}
-
-// Value will return value from underying VaryingDataType
-func (asc *authoritySetChangeID) Value() (val scale.VaryingDataTypeValue, err error) {
-	vdt := scale.VaryingDataType(*asc)
-	return vdt.Value()
+type authoritySetChangeID interface {
```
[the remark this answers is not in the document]
jimjbrettj: It was per suggestion from this comment, https://github.com/ChainSafe/gossamer/pull/3589#discussion_r1408247589 and no I do not believe that it is encoded. Will try your suggestion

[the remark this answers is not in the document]
jimjbrettj: updated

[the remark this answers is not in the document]
EclesioMeloJunior: Changing `authoritySetChangeID` from an interface to any and introducing a constructor that returns `any` is not worth. It adds complexity but in the end, is the same as simply returning any. I don't get the point of a constructor that instead of construct type is returning any? So I would keep suggesting https://github.com/ChainSafe/gossamer/pull/3589#discussion_r1408247589, also if we want to implement enums that don't need to be scale encoded decoded we should use interfaces, take as example https://github.com/ChainSafe/gossamer/pull/3570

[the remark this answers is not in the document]
timwu20: `any` is the same as `interface{}`.  I don't see the need to create a superfluous interface to represent these 3 options when they can be represented by an interface constraint.  By creating a  constructor that uses this constraint as a generic parameter we can enforce what is stored as a `authoritySetChangeID` at runtime as well.

client/consensus/grandpa/justification.go
timwu20: I was thinking you could just `scale.Unmarshal` into a `GrandpaJustification` which would call the custom method.  However, I don't think that'll work given we need to introduce a new generic parameter.

client/consensus/grandpa/authorities.go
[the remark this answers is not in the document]
EclesioMeloJunior: @timwu20 @jimjbrettj  I wrote a simple post on hackmd with a cleaner example and explanation about why using the suggestion I made is not superfluous and has a better type-safety and less complex code and why have a constructor don't have those guarantees

https://hackmd.io/@-8f0KkDaSXGl-AgsJkF8Jg/B1X1WAxIT

client/consensus/grandpa/finality_proof_test.go
jimjbrettj: Yeah because the logic is repeated (but sometimes failing at different parts), but if this is confusing I can change it

client/consensus/grandpa/justification.go
```diff
@@ -19,23 +20,74 @@ var (
...
+	VotesAncestries []Header[Hash, N]
+}
+
+// Type used for decoding grandpa justifications (can pass in generic Header type)
+type decodeGrandpaJustification[
+	Hash constraints.Ordered,
+	N constraints.Unsigned,
+	S comparable,
+	ID AuthorityID,
+	H Header[Hash, N],
```
axaysagathiya, line 48: Do you need this `H` parameter? I think you are not using it.

client/consensus/grandpa/authorities.go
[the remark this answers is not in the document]
jimjbrettj: This approach makes sense to me, thank you for the write-up @EclesioMeloJunior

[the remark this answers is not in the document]
jimjbrettj: I have updated the code to reflect your suggestion @EclesioMeloJunior

client/consensus/grandpa/authorities_test.go
```diff
@@ -1205,36 +1204,29 @@ func TestCleanUpStaleForcedChangesWhenApplyingStandardChangeAlternateCase(t *tes
...
-	authSetVal, err := authSetID.Value()
-	require.NoError(t, err)
-	switch val := authSetVal.(type) {
-	case set[uint]:
+	switch val := authSetID.(type) {
+	case authoritySetChangeIDSet[uint]:
 		require.Equal(t, expected, val.inner)
 	default:
-		err = fmt.Errorf("invalid authSetID type")
+		t.FailNow()
```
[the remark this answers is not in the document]
jimjbrettj: As I am using eclesios suggestion I will leave as is

client/consensus/grandpa/finality_proof.go
```diff
@@ -0,0 +1,248 @@
...
+const maxUnknownHeaders = 100_000
+
+// FinalityProofProvider Finality proof provider for serving network requests.
+type FinalityProofProvider[
+	BE Backend[Hash, N, H, B],
+	Hash constraints.Ordered,
+	N constraints.Unsigned,
+	AuthID AuthorityID,
+	S comparable,
+	ID AuthorityID,
```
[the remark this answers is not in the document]
jimjbrettj: I did not need ID (good catch) as it was a duplicate, but I do need the signature because the constraint gets used to retrieve the best justification in `proveFinality`

client/consensus/grandpa/justification.go
jimjbrettj: Yes it is needed to properly decode a grandpa justification. This is a helper struct created to decode a `GrandpaJustification` because the `GrandpaJustification.VoterAncestries` is defined as an interface, which you cannot decode into. By passing in H as a generic constraint here we can define `VoterSncestries` to point to a concrete type to be decoded into

client/consensus/grandpa/authorities.go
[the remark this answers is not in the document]
timwu20: This approach is not necessary.  You can ensure type safety through any function by using the interface type constraint.  See this [play link](https://go.dev/play/p/6g58lCu2yir).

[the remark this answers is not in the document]
EclesioMeloJunior: Not in this case since to use interface type constraint we will need to use `any` as a "hack" to support the type constraint enum as a "shadowed" value, I will keep suggesting the approach described here https://github.com/ChainSafe/gossamer/pull/3589#discussion_r1408247589. ref: https://zephyrtronium.github.io/articles/generics-enums.html

client/consensus/grandpa/finality_proof.go
```diff
@@ -0,0 +1,246 @@
...
+		return &encodedProof, nil
+	}
+
+	return nil, nil
+}
+
+// Prove finality for the given block number by returning a Justification for the last block of
+// the authority set.
+//
+// If `collect_unknown_headers` is true, the finality proof will include all headers from the
```
timwu20, line 101:
```suggestion
// If `collectUnknownHeaders` is true, the finality proof will include all headers from the
```

client/consensus/grandpa/finality_proof_test.go
timwu20: well given that you never use the `t` parameter, it's not really a helper function.

client/consensus/grandpa/finality_proof.go
timwu20: do one last check on the comments to make it not so rust-y 🦀

client/consensus/grandpa/authorities.go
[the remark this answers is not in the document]
timwu20: Based on our conversations on slack, I think there's an acceptable tradeoff here to use an interface constraint here to constrain what we store in the shadowed `any` type.  Using an interface constraint makes it clear the intention here.  Hopefully they introduce something in future versions of Go to be able to replicate sum types (aka. rust enums).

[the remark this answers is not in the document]
jimjbrettj: This sounds good to me and I will keep this suggestion in my code and mark this as resolved.

EclesioMeloJunior (approved):

timwu20 (approved):

Issue #3597, opened 2023-11-20
tooling: export trie state through RPC endpoint


EclesioMeloJunior:
## Issue summary

- Exporting the trie state through an RPC endpoint is a useful tool for node tests, and also for sync debugging.
  - We should able to export a file or data with the respective trie state for a certain block (if no block is given then return for the best block)
  - If the block was already pruned then use database info to create the trie and export it through the RPC endpoint

## Other information and links

-

PR #3607, opened 2023-11-23
feat(dot/rpc): export block trie state entries for a block hash

EclesioMeloJunior:
## Changes

- Exports a new RPC method call `state_trie`, this method receives an optional hash `At *common.Hash`, if no hash is provided then it will use the best block hash, then it will search in the `storage trie` the trie entries for the block's state root, returning a list of `{Key []byte, Value []byte}`

## Tests

```sh
go test -timeout 10m -run ^TestStateTrie$ github.com/ChainSafe/gossamer/dot/rpc/modules --tags=integration -v
```

## Issues

- Closes https://github.com/ChainSafe/gossamer/issues/3597

## Primary Reviewer

@timwu20

### DIFF
--- a/dot/rpc/modules/state.go
+++ b/dot/rpc/modules/state.go
@@ -11,6 +11,7 @@ import (
 
 	"github.com/ChainSafe/gossamer/lib/common"
 	"github.com/ChainSafe/gossamer/lib/runtime"
+	"github.com/ChainSafe/gossamer/lib/trie"
 	"github.com/ChainSafe/gossamer/pkg/scale"
 )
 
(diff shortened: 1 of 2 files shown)


dimartiro (approved):

jimjbrettj (approved):

dimartiro (approved):
LGTM

dot/rpc/modules/state.go
```diff
@@ -245,6 +252,43 @@ func (sm *StateModule) GetPairs(_ *http.Request, req *StatePairRequest, res *Sta
 	return nil
 }
 
+// Trie RPC method returns a list of scale encoded trie.Entry{Key byte, Value byte} representing
+// all the entries in a trie for a block hash, if no block hash is given then it uses the best block hash
+func (sm *StateModule) Trie(_ *http.Request, req *StateTrieAtRequest, res *StateTrieResponse) error {
+	blockHash := sm.blockAPI.BestBlockHash()
```
P1sar, line 258: this calls looks pretty expensive and in current implementation we have to call it every time RPC method getting called. Maybe we can use if else to call it only when necessary?

PR #3617, opened 2023-12-01
feat(pkg/scale): Add `Marshaler` and `Unmarshaler` interfaces and functionality

### LINKED ISSUE #3558: feat(pkg/scale): Implement new `Marshaler` and `Unmarshaler` interfaces for SCALE
## Issue summary

- Adopt the same interface standard for [`Marshaler`](https://pkg.go.dev/encoding/json#Marshaler) and [`Unmarshaler`](https://pkg.go.dev/encoding/json#Unmarshaler) as `encoding/json`. 
- Provides SCALE encoding support for custom types moving forward.

- https://pkg.go.dev/encoding/json

timwu20:
## Changes
- Introduces `scale.Marshaler` and `scale.Unmarshaler` interface to enable custom encoding and decoding for provided types.
- Similar to `json/encoding` analogous interfaces, but `scale.Unmarshaler` takes in a `io.Reader` instead of `[]byte` due to the nature of SCALE encoded data and that the shape of the data is unknown until the destination type is known.

## Tests

```sh
```

## Issues

closes #3558 

## Primary Reviewer

@jimjbrettj

### DIFF
--- a/pkg/scale/decode_test.go
+++ b/pkg/scale/decode_test.go
@@ -5,6 +5,9 @@ package scale
 
 import (
 	"bytes"
+	"encoding/binary"
+	"fmt"
+	"io"
 	"math/big"
 	"reflect"
 	"testing"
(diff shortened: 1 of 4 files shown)


pkg/scale/encode.go
```diff
@@ -62,6 +67,21 @@ type encodeState struct {
...
+		values := methodVal.Call(nil)
+		if !values[1].IsNil() {
+			errIn := values[1].Interface()
+			err := errIn.(error)
+			return err
+		}
+		bytes := values[0].Interface().([]byte)
+		_, err = es.Write(bytes)
+		return
+	}
```
EclesioMeloJunior, line 83:
```suggestion
        _, ok := in.(Marshaler)
        if ok {
            inv := reflect.ValueOf(in)
            methodVal := inv.MethodByName("MarshalSCALE")
	    values := methodVal.Call(nil)
	    if !values[1].IsNil() {
		  errIn := values[1].Interface()
		  err := errIn.(error)
		  return err
	    }
	    bytes := values[0].Interface().([]byte)
	    _, err = es.Write(bytes)
	    return
        }
```

pkg/scale/decode_test.go
```diff
@@ -535,3 +538,85 @@ func Test_decodeState_decodeUint(t *testing.T) {
 		})
 	}
 }
+
+type myStruct struct {
```
jimjbrettj, line 542: Since one of the use cases is decoding into a struct with an interface as a field type, perhaps could be nice to have as a test case

EclesioMeloJunior (approved):
LGTM, just a small change to check using `in.(Marshal)` instead of reflection

jimjbrettj (approved):
lgtm

pkg/scale/encode.go
timwu20: Ended up not using reflect at all to call `Marshaler.MarshalSCALE`.

Issue #3623, opened 2023-12-05
(pkg/trie): Implement trie nodes / values cache


dimartiro:
In order to reduce the amount of memory used by our servers we have to implement a cache to prevent having 100% of the tries in memory.

This will be used with the lazy loading to have the most frequently used tries into the cached and improve the read times

### External resources:
[Substrate impl](https://github.com/paritytech/polkadot-sdk/blob/master/substrate/primitives/trie/src/cache/mod.rs#L18)

PR #3624, opened 2023-12-05
feat(trie): Implement LRU cache for in memory trie

dimartiro:
## Changes

We added an implementation for a LRU cache to be used in our in memories tries and prevent having the entire trie list in memory with the goal of reducing the memory usage

## Tests

```sh
make test
```

## Issues

#3623 

## Primary Reviewer

@timwu20


lib/utils/lru-cache/lru_cache.go
```diff
@@ -93,3 +119,27 @@ func (c *LRUCache[K, V]) Put(key K, value V) {
...
+	delete(c.cache, key)
+	return true
+}
+
+// Put adds a key-value pair to the cache.
+func (c *LRUCache[K, V]) Len() int {
+	c.Lock()
+	defer c.Unlock()
+
+	return len(c.cache)
```
P1sar, line 144: lruList has Len() attribute, it is 100% O(1) complexity while I am not sure about bult-in function

lib/utils/lru-cache/lru_cache.go
```diff
@@ -93,3 +119,27 @@ func (c *LRUCache[K, V]) Put(key K, value V) {
...
+		return false
+	}
+
+	c.lruList.Remove(val)
+
+	delete(c.cache, key)
+	return true
+}
+
+// Put adds a key-value pair to the cache.
```
P1sar, line 139: check godocs for LRU functions please

dot/state/tries.go
```diff
@@ -30,11 +29,12 @@ var (
 	})
 )
 
+const MaxInMemoryTries = 100
```
P1sar, line 32: How this was defined? Should we maybe add some description for this

lib/utils/lru-cache/lru_cache.go
dimartiro: Yes, it is O(1), [go keeps track of the items amount internally](https://github.com/golang/go/blob/master/src/runtime/map.go#L120)

lib/utils/lru-cache/lru_cache.go
dimartiro: Thanks! updated

dot/state/tries.go
dimartiro: This PR is still WIP, I want to use the same number used in substrate but I have to do a little research to find it. We don't use to review draft PRs because the code can change. Anyways, thanks for the comment.

dimartiro:
Closing it. We are implementing the same in https://github.com/ChainSafe/gossamer/pull/3658

Issue #3614, opened 2023-12-01
invalid ristretto encoding at `#10516963`


EclesioMeloJunior:
## Describe the bug
- Gossamer cannot decode a `next epoch data` consensus digest
- The problematic consensus digest exists in the block `#10516963` and is related to epoch `17866`

```
2023-12-01T03:35:11Z DEBUG    ⏩ next epoch data stored in memory: epoch 17866, hash: 0xa9e190ef90d38f4001294fe58cabdde23bbb6b8374fc25e8b80cff78e0564cb1	pkg=state
[...]

PR #3627, opened 2023-12-05
fix(dot/state): store raw authority keys and decode when verifying block signature

EclesioMeloJunior:
## Changes

- Gossamer should not parse the authority raw key bytes into sr25519 public keys as soon as it receives a `next epoch` babe consensus digest.
- The authorities' raw keys will be stored as raw bytes and only decoded to public keys when verifying some block signature, and in such cases if the key cannot be decoded the verification fails and the block is rejected

## Tests

```sh
WIP
```

## Issues

- Closes https://github.com/ChainSafe/gossamer/issues/3614

## Primary Reviewer

@jimjbrettj


jimjbrettj (approved):
Very nice work

EclesioMeloJunior:
> I feel like this shouldn't happen. Like why are we getting bad epoch digests in the first place? Does this authority set not change in this case, or do we receive another one in a future block?

This is not a bad epoch digest, this epoch digest is ok and it is from a canonical block that belongs to a finalized chain, the problem is that one authority public key bytes in that set cannot be decoded into a `sr25519` public key instance. so we are following the recommendations from substrate and smoldot, basically we only instantiate the public key when checking against the block signature

timwu20:
> This is not a bad epoch digest, this epoch digest is ok and it is from a canonical block that belongs to a finalized chain, the problem is that one authority public key bytes in that set cannot be decoded into a `sr25519` public key instance. As Andre (from parity) and Tomaka (from smoldot) recommended, they store the public keys as raw bytes and only decode when needed to verify the block's signature

Since we can't verify the signature, does the authority set change for the next epoch?

EclesioMeloJunior:
Yeah, every new epoch the authority set changes, sometimes the set has big changes sometimes the set keeps the same, the runtime is the one who chooses the new authorities, so let's say that Gossamer is producing blocks and it faces a similar case, it receives a next epoch digest which contains a authority key that cannot be decoded into `sr25519` Public Key, so whenever Gossamer receives a block from that authority and try to check the signature then it will fail because we cannot decode the public key raw bytes in first place and then the block will be rejected an AFAIK we should report this to the runtime. So, in this case, any blocks made by this invalid pub key will not be in the chain, the authority set will remain until the next epoch digest comes and changes the current set.

Now we can see the scenario where Gossamer or a node is syncing: the digest comes and we should keep the authorities even if in the set may exists one invalid authority (that will never be used to verify a canonical block through sync because it was not able to produce and persist a block in the chain)

timwu20 (approved):

PR #3532, opened 2023-10-11
feat(dot/parachain/backing): handle `statement` message coming from overseer

axaysagathiya:
## Tests

```sh
go test -tags integration github.com/ChainSafe/gossamer
```

## Issues
closes #3507 

## Primary Reviewer

@kishansagathiya

PR #3638, opened 2023-12-12
feat(dot/parachain): added overseer signals

### LINKED ISSUE #3495: feat(dot/parachain): add overseer signals
## Issue summary
- From time to time parachain subsystems need information about relay chain. We use overseer to get these information and pass them on to different subsystems as overseer signals.

Now that we have overseer merged, we would be able to add these overseer signals, which are as follows
- ActiveLeaveUpdate
- BlockFinalized
- Conclude (Research and confirm if we can live without adding the Conclude signal)

I would suggest to add `ActiveLeaveUpdate` and `BlockFinalized` first.
Conclude is used to tell that work of all subsystems and overseer is done. So, I feel that it's probably not required, or there could be other ways to give that signal. But just see where it is used and confirmed if it is not necessary, keeping in mind difference in functionalities of rust and go.

- Since all Subsystems will have to process these message, we could also add an extra method in the SubSystem interface say `ProcessOverseerSignals`

 

Link to OverseerSignal enum
- https://github.com/paritytech/polkadot-sdk/blob/7ca0d65f19497ac1c3c7ad6315f1a0acb2ca32f8/polkadot/node/subsystem-types/src/lib.rs#L151
- https://github.com/paritytech/polkadot-sdk/blob/7035034710ecb9c6a786284e5f771364c520598d/polkadot/node/overseer/src/lib.rs#L742

kishansagathiya:
## Changes
- Added ActiveLeavesUpdate and BlockFinalized overseer signals.
- Connected parachain service with a block state. Using this block state parachain service would be notified when a new relay chain block gets imported or finalized.
- On receiving these signals overseer would update its active leaves and broadcast an ActiveLeavesUpdate or/and BlockFinalized signal to all subsystems
- Added `ProcessOverseerSignals` method to the subsystem interface, all subsystems will have to implement this method based on their requirements to react to these signals

## Tests

```sh
```

## Issues
Fixes #3495

## Primary Reviewer

@kanishkatn @edwardmack


dot/parachain/overseer/overseer_test.go
```diff
@@ -45,12 +48,22 @@ func (s *TestSubsystem) Run(ctx context.Context, OverseerToSubSystem chan any, S
...
 }
 
+func (s *TestSubsystem) Stop() {}
+
 func TestStart2SubsytemsActivate1(t *testing.T) {
-	overseer := NewOverseer()
+	ctrl := gomock.NewController(t)
+
+	blockState := NewMockBlockState(ctrl)
+	overseer := NewOverseer(blockState)
```
edwardmack, line 65: This test is failing now, because of calls to new mock blockState:  
```
=== RUN   TestStart2SubsytemsActivate1
    overseer.go:80: Unexpected call to *overseer.MockBlockState.GetImportedBlockNotifierChannel([]) at /home/emack/projects/ChainSafe/gossamer/dot/parachain/overseer/overseer.go:80 because: there are no expected calls of the method "GetImportedBlockNotifierChannel" for that receiver
--- FAIL: TestStart2SubsytemsActivate1 (0.00s)
```

dot/parachain/overseer/overseer_test.go
```diff
@@ -141,7 +157,11 @@ func TestStart2SubsytemsActivate2Different(t *testing.T) {
 }
 
 func TestStart2SubsytemsActivate2Same(t *testing.T) {
-	overseer := NewOverseer()
+	ctrl := gomock.NewController(t)
+
+	blockState := NewMockBlockState(ctrl)
+	overseer := NewOverseer(blockState)
```
edwardmack, line 163: This test is failing now because of unexpected call to new mock block state:
```
=== RUN   TestStart2SubsytemsActivate2Different
    overseer.go:80: Unexpected call to *overseer.MockBlockState.GetImportedBlockNotifierChannel([]) at /home/emack/projects/ChainSafe/gossamer/dot/parachain/overseer/overseer.go:80 because: there are no expected calls of the method "GetImportedBlockNotifierChannel" for that receiver
--- FAIL: TestStart2SubsytemsActivate2Different (0.00s)
```

edwardmack (commented):
This looks good, just need to update some tests to handle the calls to mock block state.

axaysagathiya (commented):
LGTM, just address the failing unit tests.

kanishkatn (changes requested):
This is great, just an observation.

dot/parachain/availability-store/availability_store.go
```diff
@@ -170,7 +175,11 @@ func uint32ToBytes(value uint32) []byte {
 // Run runs the availability store subsystem
 func (av *AvailabilityStoreSubsystem) Run(ctx context.Context, OverseerToSubsystem chan any,
 	SubsystemToOverseer chan any) error {
-	av.processMessages()
+
+	av.wg.Add(2)
+	go av.processMessages()
+	go av.ProcessOverseerSignals()
```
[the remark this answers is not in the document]
kishansagathiya: You are right, I will fix this.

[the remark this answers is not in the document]
kishansagathiya: done

dot/parachain/availability-store/availability_store.go
```diff
@@ -170,7 +175,10 @@ func uint32ToBytes(value uint32) []byte {
 // Run runs the availability store subsystem
 func (av *AvailabilityStoreSubsystem) Run(ctx context.Context, OverseerToSubsystem chan any,
 	SubsystemToOverseer chan any) error {
-	av.processMessages()
+
+	av.wg.Add(2)
```
kanishkatn, line 179: I don't think there is a need for the waitGroup here.

kanishkatn (changes requested):
I propose we use generics for both subsystem messages and overseer signals. I've done the same in disputes-coordinator and since I am preparing it for merge; I believe this to be the right time to take a call on this.

Here's an example followed by benefits:

disputes-subsystem.go
```go
switch message := msg.(type) {
	case disputeTypes.Message[disputeTypes.ParticipationStatement]:
	case disputeTypes.Message[disputeTypes.ImportStatements]:
	case overseer.Signal[overseer.Conclude]:
        default: panic("unexpected type")
}
```
overseer.go
```go
switch message := msg.(type) {
	case disputeTypes.Message[any]: // send to disputes-subsystem
	case availabilityTypes.Message[any]: // send to availability subsystem
        default: panic("unexpected type")
 }
```

The type definitions:
disputes.go
```go
// Message messages to be handled in this subsystem.
type Message[data any] struct {
	Data            data
	ResponseChannel chan any
}
// ParticipationStatement is a statement as result of the validation process.
type ParticipationStatement struct {
	Session          parachainTypes.SessionIndex
	CandidateHash    common.Hash
	CandidateReceipt parachainTypes.CandidateReceipt
	Outcome          ParticipationOutcomeVDT
}
```
overseer.go
```go
// Signal represents a signal sent from overseer
type Signal[data any] struct {
	Data            data
	ResponseChannel chan<- any
}

// Conclude the work of overseer and all its subsystems.
type Conclude struct{}
```

Benefits:
- Decouples the response channel from the message and makes it part of the parent struct. It keeps the code clean that way.
- Makes the overseer message routing logic simple
- Code readability, reusability and maintainability 
- We can use wrappers for message communication and further keep the code more readable. Ref: https://github.com/ChainSafe/gossamer/blob/feat/k/disputes/coordinator/dot/parachain/dispute/comm/comm.go

axaysagathiya:
> // Message messages to be handled in this subsystem.
type Message[data any] struct {
	Data            data
	ResponseChannel chan any
}

IMO, This is not the right way to handle overseer messages.
This way, we can send anything. We need to send only disputeTypesMessage.

Implementing enums using an interface would be better.

kanishkatn:
> This way, we can send anything. We need to send only disputeTypesMessage.

Just addressing this concern. We can have checks for the types to ensure that. I've updated the example to capture it.

> Implementing enums using an interface would be better.

@axaysagathiya  That's an interesting approach. Would it be possible for you to share a code example?

axaysagathiya:
> Just addressing this concern. We can have checks for the types to ensure that. I've updated the example to capture it.

It is better to ensure we send the correct types instead of allowing incorrect ones and checking for the right type later.

> That's an interesting approach. Would it be possible for you to share a code example?

if u want example, you can check [overseer_message.go](https://github.com/ChainSafe/gossamer/pull/3532/files#diff-53e025e55da45bfd52d4a57aad47d509199664d8669967e1e78b499203527838) file in my PR to handle statement message.

if we implement overseer messages(of type enum) using interface, 
- there is no chance we send an incorrect message
- also, it would be easy to identify message types and divert messages to their respective subsystem. Instead of checking all the struct types in the switch case, we can check which message interface is implemented.

kanishkatn:
@axaysagathiya Thank you for sharing.

I don't think an incorrect message type is a concern though, since we can handle the error flows easily. 
 
That still got me thinking; How about this?
 
 ```go
 type DisputesMessage interface {
	IsDisputesMessage()
}

type Message[data DisputesMessage] struct {
	Data            data
	ResponseChannel chan any
}
 ```

kishansagathiya:
@axaysagathiya @kanishkatn Let's have this discussion separately. And make these changes to message type in a separate PR.
Both of you have edited message type in your respective PRs. It is not quite obvious to me that which approach is better.  Message definitely need to be structured with Data and Response both however. 

@kanishkatn It is a positive discussion that you have ignited, but it is not really a problem in this PR, but rather a request of enhancement to the current workings. I wouldn't like this PR to get blocked on this.

axaysagathiya:
> That still got me thinking: How about this?
> 
> type DisputesMessage interface {
>    IsDisputesMessage()
> }
> 
> type Message[data DisputesMessage] struct {
>    Data            data
>    ResponseChannel chan any
> }

We don't need a response to all the messages. So, I am not sure if we need this Message structure.

kishansagathiya:
> We don't need a response to all the messages. So, I am not sure if we need this Message structure.

Yes, all messages don't responses, but a decent number of them do.

kanishkatn:
> > That still got me thinking: How about this?
> > type DisputesMessage interface {
> > IsDisputesMessage()
> > }
> > type Message[data DisputesMessage] struct {
> > Data            data
> > ResponseChannel chan any
> > }
> 
> We don't need a response to all the messages. So, I am not sure if we need this Message structure.

Yes, you're right. It is optional, the handler won't use the channel in such cases. Optional fields are fine!

dot/parachain/availability-store/availability_store.go
kishansagathiya: a goroutine starts after this line. It is just a nice practice to make sure that the goroutine has ended when we run the Stop() function.

kanishkatn:
> @kanishkatn It is a positive discussion that you have ignited, but it is not really a problem in this PR, but rather a request of enhancement to the current workings. I wouldn't like this PR to get blocked on this.

As mentioned above, I am currently working on getting a PR to feat/parachain for the disputes-coordinator and this is a blocker for that. We could have a separate PR for the subsystem messages, but I feel we need to take a decision on the overseer signals to unblock everyone.

kishansagathiya:
@kanishkatn

> As mentioned above, I am currently working on getting a PR to feat/parachain for the disputes-coordinator and this is a blocker for that. We could have a separate PR for the subsystem messages, but I feel we need to take a decision on the overseer signals to unblock everyone.

> I also believe we should take a decision on the approach as a team at this time to avoid stacking up tech debts.

I liked both of your ideas. But importance and urgency that you are giving to this is way too much and unnecessary. It's not going to take even 30 min to restructure message type. And message type as it is, it is not so bad. Oversee signal is just a type of message, so far I have not seen a reason to have a separate type for that.

I will however push some changes to resolve this discussion.

> As mentioned above, I am currently working on getting a PR to feat/parachain for the disputes-coordinator and this is a blocker for that.

To avoid that problem, you should have kept your branch always updated with `feat/parachain`. Having gone through parts of that PR, I believe it is quite possible to ready that PR without edited message type. You can choose to keep your overseer changes in your branch, and those changes could be reviewed along with rest of the changes.

kishansagathiya:
Personally, I would like to not have a message struct at all. I have avoid using generics as I feel they un-necessarily complicate things. There is no right or wrong, it's just a choice.
In this case, we could create a generic like following
```
type Message[Data any, Response chan any] struct {
	Data     Data
	Response Response
}
```
but then, signals don't need response channels and there are many other messages that don't need response channels. Whenever we need a response we can directly put that into the struct of that particular message.
Sometimes, we wrap things with a struct or interface to hide some internal fields. But that is not the case here, whoever processes these messages need to know their type and values inside.

So, I don't see a need for this.
Look at `func (o *Overseer) processMessages() {` for more context. 
I have also created a branch with Message as a generic struct. Look at this commit https://github.com/ChainSafe/gossamer/commit/8804ef4cef31db2245f2bfe670274463a072944d. Don't like that change.

@kanishkatn Let me know if this reasoning makes sense to you.

kanishkatn:
@kishansagathiya  I still believe generics has a good usecase here. Please go ahead with what you deem fit at this moment. I'm withdrawing my change request.

kanishkatn (approved):

PR #3669, opened 2024-01-09
refactor(pkg/scale): revise `VaryingDataType` to an interface

### LINKED ISSUE #3721: Revise SCALE `VaryingDataType` to an interface
## Issue summary

- Given we would like to handle enums/sum types using different methodologies it makes sense that we should make `scale.VaryingDataType` an interface that implementors can use to implement decoding and encoding logic for the SCALE package.
- Update Gossamer references to the existing concrete type

- #1819

timwu20:
## Changes

- revises the `VaryingDataType` from a struct to an interface which the SCALE codec uses to encode/decode.
- removes `VaryingDataTypeSlice`
- updates SCALE docs
- updates all Gossamer interactions with `scale.VaryingDataType` and `scale.VaryingDataTypeSlice`
- needed for `feat/grandpa` work

## Tests

```sh
```

# Issues
- closes #3721

## Primary Reviewer

@jimjbrettj


pkg/scale/encode.go
```diff
@@ -78,6 +78,12 @@ func (es *encodeState) marshal(in interface{}) (err error) {
 		return
 	}
 
+	vdt, ok := in.(EncodeVaryingDataType)
+	if ok {
+		err = es.encodeVaryingDataType(vdt)
+		return
+	}
+
```
EclesioMeloJunior, line 86: for encoding a VDT it only needs to know the variant index, so I would suggest to

```go
type Indexable interface {
    Index() uint
}


vdt, ok := in.(Indexable)
if ok {
	err = es.encodeVaryingDataType(vdt)
	return
}

func (es *encodeState) encodeVaryingDataType(vdt Indexable) (err error) {
	_, err = es.Write([]byte{byte(vdt.Index())})
	if err != nil {
		return
	}
	err = es.marshal(vdt)
	return
}
```

Then you can encode a variant w/o wrapping it around another type, somewhat is done in rust

```rs
let v = RustEnum::Variant
v.encode()
```

in golang

```go
type Variant uint16
func(Variant) Index() uint {return 0}

v := Variant(15)
out, err := scale.Marshal(v)
```

instead of

```go
vdt := MyVaryingDataType{}
err := vdt.SetValue(Variant(15))
if err != nil {
	panic(err)
}

bytes, err := scale.Marshal(vdt)
if err != nil {
	panic(err)
}
```

timwu20: I think that makes sense for the encoding part.  I don't think It can work like that on the decoding side.  We will always need something that is able to provide the codec the types per index.  If there's a use case for just encoding types without wrapping it I think it makes sense to reintroduce the `VaryingDataTypeValue` interface and make the appropriate changes to the codec in a separate feature.

dot/types/babe_digest.go
```diff
@@ -11,13 +11,79 @@ import (
...
+		return *new(BabeSecondaryVRFPreDigest), nil
+
+	}
+	return nil, scale.ErrUnknownVaryingDataTypeValue
+}
+
 // NewBabeDigest returns a new VaryingDataType to represent a BabeDigest
-func NewBabeDigest() scale.VaryingDataType {
-	return scale.MustNewVaryingDataType(BabePrimaryPreDigest{}, BabeSecondaryPlainPreDigest{}, BabeSecondaryVRFPreDigest{})
+func NewBabeDigest() BabeDigest {
```
jimjbrettj, line 81: I like this a lot. Feels much more intuitive to interact with the types

dot/types/consensus_digest.go
```diff
@@ -147,26 +272,49 @@ func (d *NextConfigDataV1) ToConfigData() *ConfigData { //skipcq: GO-W1029
 }
 
 // VersionedNextConfigData represents the enum of next config data consensus digest messages
-type VersionedNextConfigData scale.VaryingDataType
+type VersionedNextConfigDataValues interface {
```
jimjbrettj, line 275: Why the rename here?

dot/types/digest.go
```diff
@@ -5,24 +5,123 @@ package types
...
+}
+
+func (mvdt DigestItem) Value() (value any, err error) {
+	_, value, err = mvdt.IndexValue()
+	return
+}
+
+func (mvdt DigestItem) ValueAt(index uint) (value any, err error) {
+	switch index {
+	case 6:
```
jimjbrettj, line 77: Nit/should we list these in order?

dot/types/digest_test.go
```diff
@@ -132,7 +133,7 @@ func TestDecodeDigest(t *testing.T) {
 	v := NewDigest()
 	err := scale.Unmarshal(d, &v)
 	require.NoError(t, err)
-	require.Equal(t, 3, len(v.Types))
```
jimjbrettj, line 135: Really really like getting rid of this

lib/babe/verify_integration_test.go
```diff
@@ -145,7 +145,9 @@ func TestVerificationManager_VerifyBlock_Secondary(t *testing.T) {
 	secondaryDigest := createSecondaryVRFPreDigest(t, keyring.Alice().(*sr25519.Keypair),
 		0, uint64(0), uint64(0), Randomness{})
 	babeDigest := types.NewBabeDigest()
-	err = babeDigest.Set(secondaryDigest)
+	// NOTE: I think this was get encoded incorrectly before the VDT interface change.
+	// *types.BabeSecondaryVRFPreDigest was being passed in and encoded later
+	err = babeDigest.SetValue(*secondaryDigest)
```
jimjbrettj, line 150: Does this PR change this behavior? I'm not seeing where

lib/babe/verify_integration_test.go
```diff
@@ -179,7 +181,7 @@ func TestVerificationManager_VerifyBlock_Secondary(t *testing.T) {
 		Body:   nil,
 	}
 	err = vm.VerifyBlock(&block.Header)
-	require.EqualError(t, err, "failed to verify pre-runtime digest: block producer is not in authority set")
+	require.EqualError(t, err, "invalid signature length")
```
jimjbrettj, line 184: Why change the expected error?

lib/grandpa/message.go
```diff
@@ -18,11 +18,97 @@ type GrandpaMessage interface {
...
+	return nil, scale.ErrUnknownVaryingDataTypeValue
+}
+
 // NewGrandpaMessage returns a new VaryingDataType to represent a GrandpaMessage
-func newGrandpaMessage() scale.VaryingDataType {
-	return scale.MustNewVaryingDataType(
-		VoteMessage{}, CommitMessage{}, newVersionedNeighbourPacket(),
-		CatchUpRequest{}, CatchUpResponse{})
+func newGrandpaMessage() grandpaMessage {
+	// return scale.MustNewVaryingDataType(
```
jimjbrettj, line 108: delete

lib/runtime/transaction_validity.go
```diff
@@ -16,25 +16,73 @@ var errInvalidTypeCast = errors.New("invalid type cast")
...
+	switch index {
+	case 0:
+		return InvalidTransaction{}, nil
+	case 1:
+		return UnknownTransaction{}, nil
+	}
+	return nil, scale.ErrUnknownVaryingDataTypeValue
 }
 
+// // Set will set a VaryingDataTypeValue using the underlying VaryingDataType
```
jimjbrettj, line 69: delete

pkg/scale/encode_test.go
```diff
@@ -942,7 +947,7 @@ var (
 	allTests = newTests(
 		fixedWidthIntegerTests, variableWidthIntegerTests, stringTests,
 		boolTests, structTests, sliceTests, arrayTests,
-		varyingDataTypeTests,
+		// varyingDataTypeTests,
```
jimjbrettj, line 950: should these be added back?

pkg/scale/varying_data_type.go
```diff
@@ -3,120 +3,16 @@
 
 package scale
 
-import (
```
jimjbrettj, line 6: So much cleaner, love it

pkg/scale/varying_data_type_example_test.go
```diff
@@ -1,12 +1,11 @@
-// Copyright 2021 ChainSafe Systems (ON)
+// Copyright 2024 ChainSafe Systems (ON)
```
jimjbrettj, line 1: nit/don't we keep the original copyright on edited files?

jimjbrettj (approved):
I am a really big fan of this refactor. Over the years since the previous scale refactor we have needed to add additional functionality to it to support new use cases as they arose in our development. While working, this added complexity not only to the scale package itself but to using the package as well. From my perspective, this PR is a cleaner take on VDT's to allow easier use of them as well as clean up the code within the scale package (I think the diff speaks for itself on that one), and thus I am a big fan of this change as I think it makes interacting with scale more explicit and straightforward.

One point to mention, should the docs (or somewhere) emphasize the importance of using the associated `Set` receivers or do we think that is implied in their existence?

Aside from a few comments this PR looks good to me and I look forward to having it merged

lib/babe/verify_integration_test.go
timwu20: After debugging this test I was getting a different error.  By passing in the non-pointer type on development, I got the same error, which leads me to believe that we shouldn't have been setting the VDT with pointer to secondary digest.

lib/babe/verify_integration_test.go
timwu20: see https://github.com/ChainSafe/gossamer/pull/3669#discussion_r1455861311

dot/types/digest.go
timwu20: I just kept this based on the order the types were put in the file.

jimjbrettj (dismissed):

timwu20:
> One point to mention, should the docs (or somewhere) emphasize the importance of using the associated `Set` receivers or do we think that is implied in their existence?

I think it's implied that users can also use the interface required for the SCALE codec.

jimjbrettj (approved):

EclesioMeloJunior (approved):
VDT ftw

Issue #3649, opened 2023-12-15
Failed to sync block `#11759946`: invalid memory address or nil pointer dereference (recovered by wazero)


EclesioMeloJunior:
## Describe the bug

- When Gossamer tries to import block `#11759946` it panics with `invalid memory address or nil pointer dereference`, but since this is a call managed by wazero the panic is recovered and returned as error

```
wasm stack trace: | 0273c3cd813a4ba7877099ce0c6f7420 | gossamer-westend
-- | -- | --
December 09, 2023 at 24:26 (UTC-4:00) | env.ext_storage_get_version_1(i64) i64 | 0273c3cd813a4ba7877099ce0c6f7420 | gossamer-westend
December 09, 2023 at 24:26 (UTC-4:00) | .sp_io::storage::extern_host_function_impls::get::h64f97d2167755354(i32,i32,i32) | 0273c3cd813a4ba7877099ce0c6f7420 | gossamer-westend
December 09, 2023 at 24:26 (UTC-4:00) | .frame_support::storage::unhashed::get_or_default::h6c7739c0a72a10b2(i32,i32) i32 | 0273c3cd813a4ba7877099ce0c6f7420 | gossamer-westend
December 09, 2023 at 24:26 (UTC-4:00) | .frame_support::storage::transactional::in_storage_layer::h890d5f5e450ca12e(i32,i32) | 0273c3cd813a4ba7877099ce0c6f7420 | gossamer-westend
December 09, 2023 at 24:26 (UTC-4:00) | .<pallet_nomination_pools::pallet::Call<T> as frame_support::traits::dispatch::UnfilteredDispatchable>::dispatch_bypass_filter::h21608f5b92d65a21(i32,i32,i32) | 0273c3cd813a4ba7877099ce0c6f7420 | gossamer-westend
December 09, 2023 at 24:26 (UTC-4:00) | .<westend_runtime::Call as frame_support::traits::dispatch::UnfilteredDispatchable>::dispatch_bypass_filter::h6d0f5d67d3fde819(i32,i32,i32) | 0273c3cd813a4ba7877099ce0c6f7420 | gossamer-westend
December 09, 2023 at 24:26 (UTC-4:00) | .<westend_runtime::Call as sp_runtime::traits::Dispatchable>::dispatch::h3d28764894ef20b5(i32,i32,i32) | 0273c3cd813a4ba7877099ce0c6f7420 | gossamer-westend
[...]

Issue #3508, opened 2023-10-02
trie: Add support for nested storage transactions


dimartiro:
## Task summary

With our current implementation we are not supporting nested transactions, we are only supporting 1 level.

Our implementation keeps track of one `oldTrie` to rollback the current transaction but that does not allow us to use nested transactions

https://github.com/ChainSafe/gossamer/blob/8b93d5e532d551d5284a14c202e85f4e850632f1/lib/runtime/storage/trie.go#L50-L55

## Specification reference

The [spec says](https://spec.polkadot.network/chap-host-api#sect-ext-storage-start-transaction) that the function `ext_storage_start_transaction` should create a new `nested` transaction 

## Other information and links

- [Substrate implementation ](https://github.com/paritytech/substrate/blob/master/frame/support/src/storage/transactional.rs)

- [Kagome implementation](https://github.com/qdrvm/kagome/blob/7159c117e08e5d0bd3c9ffb14add6bf7c430b593/core/runtime/common/trie_storage_provider_impl.cpp#L168-L204)

PR #3670, opened 2024-01-10
fix: support nested storage transactions

EclesioMeloJunior:
## Changes

- The following host functions can be called multiple times in a row
 - `ext_storage_start_transaction`
 - `ext_storage_rollback_transaction`
 - `ext_storage_commit_transaction`

Basically, these functions can be nested in multiple calls like:

```
- start_transaction
  - operations ...
  - start_transaction
     - operations ...
     - commit or rollback
  - commit or rollback
```

This PR aims to implement the following approach:

- First, to define a storage transaction:
  - A storage transaction contains the `current state` which is a copy/snapshot of the previous state and it will be modified by the operations under the transaction
  -  A storage transaction can rollback, in this case we basically discard the `current state` returning back to the previous and unmodified state
  - Commit a storage transaction means the `current state` will replace the previous state

- Gossamer holds a `transactions` stack of `*trie.Trie` (state), every time the runtime makes a call to `ext_storage_start_transaction` Gossamer creates a snapshot of the current state and insert in the `transactions` stack

- The function `getCurrentTrie()`
  - Returns the original trie if no storage transaction has started (the runtime can perform operations to the trie w/o a transaction)
  - If a transaction is started then this function returns the snapshot of that transaction, so any operations will be performed in a copy of the original trie that can be commited or discarded.

- Rollback method simply pops the top of the stack, discarding all the operations applied to that snapshot
- Commit method pops the top of stack and assign it to be the current state of a previous transaction OR be the original trie

## Tests

```sh
WIP
```

## Issues

- Closes https://github.com/ChainSafe/gossamer/issues/3649#issuecomment-1881861454


dimartiro:
I think it also closes #3508 right?

EclesioMeloJunior:
> I think it also closes #3508 right?

Yes! Thanks @dimartiro

dimartiro:
@EclesioMeloJunior since there are a lot of test failing, I think will be simpler if we divide this PR into 2, one keeping the original name `TrieState` but changing the behavior to support transactions and another one to change the name from `TrieState` to `TransactionalTrieState` and updating the tests, etc. 

What do you think?

EclesioMeloJunior:
> @EclesioMeloJunior since there are a lot of test failing, I think will be simpler if we divide this PR into 2, one keeping the original name `TrieState` but changing the behavior to support transactions and another one to change the name from `TrieState` to `TransactionalTrieState` and updating the tests, etc.
> 
> What do you think?

I should open as draft PR since I still working on tests. I think we can keep them in a single PR, the renaming makes the files changed quite big but is just renaming which means that it does not require a huge review effort

dimartiro:
> > @EclesioMeloJunior since there are a lot of test failing, I think will be simpler if we divide this PR into 2, one keeping the original name `TrieState` but changing the behavior to support transactions and another one to change the name from `TrieState` to `TransactionalTrieState` and updating the tests, etc.
> > What do you think?
> 
> I should open as draft PR since I still working on tests. I think we can keep them in a single PR, the renaming makes the files changed quite big but is just renaming which means that it does not require a huge review effort

Yeah I was not talking only about that file but also the other files you have to change due the renaming.
Since it is a fix + a refactor I think we can have it done (code + review) quicker if we divide it into two, one with the fix (the transaction support) and another one with the name refactor. But is only a suggestion 😅

EclesioMeloJunior:
@dimartiro I see your point, I will revert the renaming and then address it in a later point

lib/runtime/storage/trie.go
```diff
@@ -17,97 +17,129 @@ import (
...
-		t: t,
+		transactions: make([]*trie.Trie, 0),
+		currentTrie:  state,
 	}
 }
 
-// Trie returns the TrieState's underlying trie
-func (s *TrieState) Trie() *trie.Trie {
-	return s.t
+func (t *TrieState) getCurrentTrie() *trie.Trie {
```
dimartiro, line 32: We have a mixed behavior here, if we always hold 1 transaction we can be sure that the last item is the currentTrie and this method will be trivial

lib/runtime/storage/trie.go
```diff
@@ -17,97 +17,129 @@ import (
...
+	currentTrie  *trie.Trie
 }
 
-// NewTrieState returns a new TrieState with the given trie
-func NewTrieState(t *trie.Trie) *TrieState {
-	if t == nil {
-		t = trie.NewEmptyTrie()
-	}
-
+func NewTrieState(state *trie.Trie) *TrieState {
```
dimartiro, line 25: I think it could be more straightforward if we always hold at least 1 trie in the transactions list.

lib/runtime/storage/trie.go
```diff
@@ -17,97 +17,129 @@ import (
 // TrieState is a wrapper around a transient trie that is used during the course of executing some runtime call.
 // If the execution of the call is successful, the trie will be saved in the StorageState.
 type TrieState struct {
-	t       *trie.Trie
-	oldTrie *trie.Trie // this is the trie before BeginStorageTransaction is called. set to nil if it isn't called
-	lock    sync.RWMutex
+	mtx          sync.RWMutex
+	transactions []*trie.Trie
```
dimartiro, line 21: If we change the `transactions []*trie.Trie` to `transactions *list.List` the implementation could be simpler since we can use that list as a `stack` of transactions

lib/runtime/storage/trie.go
```diff
@@ -17,97 +17,129 @@ import (
 // TrieState is a wrapper around a transient trie that is used during the course of executing some runtime call.
 // If the execution of the call is successful, the trie will be saved in the StorageState.
 type TrieState struct {
```
dimartiro, line 19: Maybe, could be helpful to have a maximum depth of transactions to prevent the memory of growing up indefinitely

dimartiro (approved):
Approved with suggestions

lib/runtime/interfaces.go
```diff
@@ -30,9 +30,9 @@ type Storage interface {
 	ClearPrefix(prefix []byte) (err error)
 	ClearPrefixLimit(prefix []byte, limit uint32) (
 		deleted uint32, allDeleted bool, err error)
-	BeginStorageTransaction()
-	CommitStorageTransaction()
-	RollbackStorageTransaction()
+	StartTransaction()
+	Commit()
```
jimjbrettj, line 34: nit/ Maybe `Commit` and `Rollback` should be `CommitTransaction` and `RollbackTransaction`

lib/runtime/storage/trie.go
jimjbrettj: I would say if substrate/kagome bound in this way could be good to do, else probably okay to leave it as is. I'm not sure if we want an arbitrary limit here, but a non arbitrary one could be nice

lib/runtime/storage/trie.go
```diff
@@ -17,97 +18,111 @@ import (
...
+	t.transactions.PushBack(t.getCurrentTrie().Snapshot())
+}
+
+// Rollback rolls back all storage changes made since StartTransaction was called.
+func (t *TrieState) Rollback() {
+	t.mtx.Lock()
+	defer t.mtx.Unlock()
+
+	if t.transactions.Len() <= 1 {
+		panic("no transactions to rollback")
```
jimjbrettj, line 56: Is this a situation that we need to panic in?

lib/runtime/storage/trie.go
```diff
@@ -17,97 +18,111 @@ import (
...
-func (s *TrieState) RollbackStorageTransaction() {
-	s.lock.Lock()
-	defer s.lock.Unlock()
-	s.t = s.oldTrie
-	s.oldTrie = nil
+// Snapshot creates a new "version" of the trie. The trie before Snapshot is called
+// can no longer be modified, all further changes are on a new "version" of the trie.
+// It returns the new version of the trie.
+func (t *TrieState) Snapshot() *trie.Trie {
+	return t.getCurrentTrie().Snapshot()
```
jimjbrettj, line 83: nice

jimjbrettj (commented):
Looks good just a few comments. We also do not need .DS_STORE and the testdata files in this do we?

lib/runtime/storage/trie.go
EclesioMeloJunior: Yeah, I didn't find a limit value on spec or substrate code, so I will leave as it is.

lib/runtime/storage/trie.go
EclesioMeloJunior: Yeah, committing or rollback without an ongoing transaction will lead to a corrupt state

lib/runtime/storage/trie.go
dimartiro: Agree! thanks for the discussion, we can mark this conversation as resolved.

jimjbrettj (approved):
great work!

PR #3600, opened 2023-11-21
feat(dot/parachain/backing): handle CanSecond incoming overseer message

axaysagathiya:
## Changes
Implemented functionality to check whether it's allowed to second a given candidate.

Reference code in polkadot-sdk: https://github.com/paritytech/polkadot-sdk/blob/769bdd3ff33a291cbc70a800a3830638467e42a2/polkadot/node/core/backing/src/lib.rs#L756-L757
## Tests

```sh
go test -tags integration github.com/ChainSafe/gossamer
```

## Issues
closes #3505

## Note
need to merge only after #3532 gets merged

## Primary Reviewer

@kishansagathiya @jimjbrettj

PR #3682, opened 2024-01-11
feat(dot/parachain/backing): handle `second` incoming overseer message

### LINKED ISSUE #3506: feat(dot/parachain/backing): handle `Second` incoming overseer message
## Issue summary

- Can only be tackled after https://github.com/ChainSafe/gossamer/pull/3497 is merged
- Here in Gossamer 
https://github.com/ChainSafe/gossamer/blob/8f10724821e8d3a4ae502e0c5e8691bd98cdd12f/dot/parachain/backing/candidate_backing.go#L109-L110

Reference code in polkadot-sdk
https://github.com/paritytech/polkadot-sdk/blob/769bdd3ff33a291cbc70a800a3830638467e42a2/polkadot/node/core/backing/src/lib.rs#L748-L749

axaysagathiya:
## Note
need to merge ONLY AFTER #3600 gets merged

## Tests

```sh
```

## Issues

closes #3506 

## Primary Reviewer

@kishansagathiya @EclesioMeloJunior


dot/parachain/backing/candidate_backing.go
```diff
@@ -221,7 +221,16 @@ func (cb *CandidateBacking) processOverseerMessage(msg any, chRelayParentAndComm
...
-		cb.handleSecondMessage()
+		err := cb.handleSecondMessage(msg.CandidateReceipt, msg.PersistedValidationData, msg.PoV, chRelayParentAndCommand)
+
+		if errors.Is(err, errWrongPVDForSecondingCandidate) ||
+			errors.Is(err, errUnknownRelayParentForSecondingCandidate) ||
+			errors.Is(err, errParaOutsideAssignmentForSeconding) ||
+			errors.Is(err, errAlreadySignedValidStatement) {
+			logger.Error(err.Error())
+			return nil
+		}
```
kanishkatn, line 232: Could you tell me what this check is for? 

Also, can we move this to a function?

dot/parachain/backing/second.go
```diff
@@ -0,0 +1,81 @@
...
+)
+
+var (
+	errWrongPVDForSecondingCandidate = errors.New(
+		"candidate backing was asked to second candidate with wrong persisted validation data")
+	errUnknownRelayParentForSecondingCandidate = errors.New(
+		"we were asked to second a candidate outside of our view")
+	errParaOutsideAssignmentForSeconding = errors.New(
+		"subsystem asked to second for parachain outside of our assignment")
+	errAlreadySignedValidStatement = errors.New("we have already signed a valid statement for this candidate")
```
kanishkatn, line 22: nit:

Is this error message copied from polkadot?

dot/parachain/backing/candidate_backing.go
EclesioMeloJunior: Also you can use a switch:

```go
switch err {
case errWrongPVDForSecondingCandidate, 
  errUnknownRelayParentForSecondingCandidate,
  ...:
    logger.Error(err.Error())
    return nil
}
```

kanishkatn: We may not need a function if it's using switch!

kanishkatn (approved):
lgtm!

dot/parachain/backing/candidate_backing.go
axaysagathiya: > Could you tell me what this check is for?

In Rust, they are sometimes logging errors and returning the OK value instead of an error.
But I have returned an error from the function, and I am making sure we return nil error further for these particular errors.

dot/parachain/backing/second.go
axaysagathiya: Yes, the error messages are copied from Polkadot.

kanishkatn: I see. I kinda look at the node as a singular entity, so I'd suggest to remove "we". 

I'm just curious as to what the rest of the team thinks and feel free to resolve this before merging if you're in a hurry!

dot/parachain/backing/second.go
```diff
@@ -0,0 +1,81 @@
...
+		return fmt.Errorf("marshalling persisted validation data: %w", err)
+	}
+
+	pvdHash, err := common.Blake2bHash(pvdBytes)
+	if err != nil {
+		return fmt.Errorf("hashing persisted validation data: %w", err)
+	}
+
+	if candidateReceipt.Descriptor.PersistedValidationDataHash != pvdHash {
+		return errWrongPVDForSecondingCandidate
```
EclesioMeloJunior, line 42: I would say that given the error context can be useful for log informations, for example in this case if `A != B` then we just return `errWrongValue` but will be better if we return `errWrongValue: expected A got B`

```suggestion
		return fmt.Errorf("%w: mismatch between %v and %v", errWrongPVDForSecondingCandidate, 
		  candidateReceipt.Descriptor.PersistedValidationDataHash,
		  pvdHash)
```

EclesioMeloJunior (changes requested):

dot/parachain/backing/second.go
```diff
@@ -0,0 +1,94 @@
...
+
+	pvdBytes, err := scale.Marshal(pvd)
+	if err != nil {
+		return fmt.Errorf("marshalling persisted validation data: %w", err)
+	}
+
+	pvdHash, err := common.Blake2bHash(pvdBytes)
+	if err != nil {
+		return fmt.Errorf("hashing persisted validation data: %w", err)
+	}
```
kishansagathiya, line 45: I think we need to calculate Hash of PVD at a few places. We could add this logic as a method to `parachaintypes.PersistedValidationData` to be able to reuse this code.

dot/parachain/backing/second.go
```diff
@@ -0,0 +1,94 @@
...
+
+	if candidateReceipt.Descriptor.PersistedValidationDataHash != pvdHash {
+		return fmt.Errorf("%w: mismatch between %s and %s",
```
kishansagathiya, line 48: This will just print two random hashes, which doesn't mean much to us. We could make this more informative like 
```suggestion
		return fmt.Errorf("%w: mismatch between persisted validation hash in candidate decriptor: %s and calculated hash of given persisted validation data: %s",
```

kishansagathiya (approved):
Will wait for the other PR to merge first, but this looks good to me!!
:tada:

EclesioMeloJunior (approved):

Issue #3684, opened 2024-01-12
CLI: fix the node identity issue


kanishkatn:
## Issue summary

The node name and the id is being overridden by the default-config and the chain-spec.

dimartiro:
@kanishkatn I think you missed the `close` keyword in your PR and this issue is still open, can we close it now that your PR is merged? 😄

kanishkatn:
@dimartiro Aw, shoot. Thank you!

PR #3685, opened 2024-01-12
chore/cli: fix node identity

kanishkatn:
## Changes

Fixes an issue wherein, the name and id of the node was being overridden by default-config and chain-spec

## Issues

#3684

## Primary Reviewer

@EclesioMeloJunior

### DIFF
--- a/chain/kusama/defaults.go
+++ b/chain/kusama/defaults.go
@@ -8,10 +8,6 @@ import (
 )
 
 var (
-	// defaultName Default node name
-	defaultName = "Kusama"
-	// defaultID Default chain ID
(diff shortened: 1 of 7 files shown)


EclesioMeloJunior:
If those flags are not passed then what will be the values of `name` and `id`?

kanishkatn:
> If those flags are not passed then what will be the values of `name` and `id`?

It used to pick it up from the default config. I had missed to copy a change. 

I've added the defaults to the flags now.

EclesioMeloJunior (approved):

dimartiro (approved):

Issue #3622, opened 2023-12-04
Gossamer sometimes don't start syncing when starting


EclesioMeloJunior:
## Issue summary

Gossamer is failing to connect to bootnodes at startup, facing errors like:

```
failed to dial: failed to dial 12D3KooWAq2A7UNFS6725XFatD5QW7iYBezTLdAUx1SmRkxN79Ne: no good addresses pkg=network

failed to open connection for peer 12D3KooWByVpK92hMi9CzTjyFg9cPHDU5ariTM3EPMq9vdh5S5Po: failed to dial: context deadline exceeded pkg=network

[/ip4/34.141.212.103/tcp/30334/ws] dial tcp 34.141.212.103:30334: i/o timeout pkg=network

[/ip4/34.141.212.103/tcp/30333] dial tcp4 34.141.212.103:30333: connect: connection refused

2023-12-04T12:27:05Z WARN failed to open connection for peer 12D3KooWGi1tCpKXLMYED9y28QXLnwgD4neYb1Arqq4QpeV1Sv3K: failed to dial: failed to dial 12D3KooWGi1tCpKXLMYED9y28QXLnwgD4neYb1Arqq4QpeV1Sv3K: all dials failed
```

The following log stream shows the errors in ordered by time

```
December 04, 2023 at 08:28 (UTC-4:00) | 2023-12-04T12:28:51Z WARN failed to advertise in the DHT: failed to find any peer in table pkg=network | 08f43eb0b5f2465fa89f813d0846918f | gossamer-westend
-- | -- | -- | --
December 04, 2023 at 08:28 (UTC-4:00) | 2023-12-04T12:28:21Z WARN failed to advertise in the DHT: failed to find any peer in table pkg=network | 08f43eb0b5f2465fa89f813d0846918f | gossamer-westend
December 04, 2023 at 08:27 (UTC-4:00) | 2023-12-04T12:27:51Z WARN failed to advertise in the DHT: failed to find any peer in table pkg=network | 08f43eb0b5f2465fa89f813d0846918f | gossamer-westend
December 04, 2023 at 08:27 (UTC-4:00) | 2023-12-04T12:27:21Z WARN failed to advertise in the DHT: failed to find any peer in table pkg=network | 08f43eb0b5f2465fa89f813d0846918f | gossamer-westend
[...]

PR #1992, opened 2021-11-05
fix(dot/sync):outlier detection

omar391:
## Changes

- Implemented an initial version of the outlier detections algorithm using IQR methods. 
- It will now remove any `attacker induced` peer state number from the calculations and return the avg for the valid numbers in the `getTarget` method.

## Tests

```
go test -timeout 30s -run ^TestChainSync_getTarget$ github.com/ChainSafe/gossamer/dot/sync
go test -timeout 30s -run ^TestRemoveOutlier$ github.com/ChainSafe/gossamer/dot/sync
```

## Issues

- Closes https://github.com/ChainSafe/gossamer/issues/1861

## Primary Reviewer

- @noot

PR #3695, opened 2024-01-15
fix(dot/sync): execute p2p handshake when there is no target

EclesioMeloJunior:
## Changes

- Gossamer is not able to run bootstrap sync without a block target, the target is calculated given the peer's block announcements handshake (which contains the peer best block and best hash), so Gossamer waits for peers to send the data.

- This PR introduces a change in the `waitEnoughPeersAndTarget`, basically this function after X seconds will, pro-actively, execute a block announcement handshake with the connected bootnodes, given the responses Gossamer can calculates the target and start syncing 

## Tests

WIP

## Issues

- Closes https://github.com/ChainSafe/gossamer/issues/3622

## Primary Reviewer

@timwu20 @q9f


dot/sync/syncer.go
```diff
@@ -64,7 +64,7 @@ func NewService(cfg *Config) (*Service, error) {
 		telemetry:          cfg.Telemetry,
 		badBlocks:          cfg.BadBlocks,
 		requestMaker:       cfg.RequestMaker,
-		waitPeersDuration:  100 * time.Millisecond,
+		waitPeersDuration:  7 * time.Second,
```
dimartiro, line 67: Where does this number come from? substrate?

dimartiro:
Why do we have to wait instead of just start sync from bootnodes? I guess it is to prevent overload the bootnodes and being a "good ~samaritan~ client" but I want to be sure 😅

dot/network/light.go
```diff
@@ -36,7 +36,7 @@ func (s *Service) decodeLightMessage(in []byte, peer peer.ID, _ bool) (Message,
 func (s *Service) handleLightMsg(stream libp2pnetwork.Stream, msg Message) (err error) {
 	defer func() {
 		err := stream.Close()
-		if err != nil {
+		if err != nil && err.Error() != "stream reset" {
```
dimartiro, line 39: Can we use a constant to compare with?

dot/sync/chain_sync.go
```diff
@@ -99,6 +100,54 @@ type announcedBlock struct {
...
+
+	view, ok := p.view[peerID]
+	if !ok {
+		p.view[peerID] = newView
+		return
+	}
+
+	if view.number >= newView.number {
+		return
+	}
```
dimartiro, line 140: Why not just:

```suggestion
	if ok && view.number >= newView.number {
		return
	}
```

dot/sync/chain_sync.go
```diff
@@ -587,21 +640,19 @@ func (cs *chainSync) requestMaxBlocksFrom(bestBlockHeader *types.Header, origin
 // head block number, it would leave us in bootstrap mode forever
 // it would be better to have some sort of standard deviation calculation and discard any outliers (#1861)
```
dimartiro, line 641: I think this was solved https://github.com/ChainSafe/gossamer/pull/1992 so we can use this PR to remove this comment too 😄

dot/sync/chain_sync.go
```diff
@@ -587,21 +640,19 @@ func (cs *chainSync) requestMaxBlocksFrom(bestBlockHeader *types.Header, origin
...
 	// we are going to sort the data and remove the outliers then we will return the avg of all the valid elements
-	uintArr := make([]uint, 0, len(cs.peerView))
-	for _, ps := range cs.peerView {
-		uintArr = append(uintArr, ps.number)
+	for _, view := range cs.peerViewSet.values() {
+		numbers = append(numbers, view.number)
 	}
 
-	sum, count := nonOutliersSumCount(uintArr)
+	sum, count := nonOutliersSumCount(numbers)
```
dimartiro, line 655: We could move the outliers calculations to the new `peerViewSet` struct, what do you think?

dimartiro (dismissed):
Very good idea to have introduced the `peerViewSet` struct, I added a couple of suggestions to take advantage of that structure but they are not blockers

EclesioMeloJunior:
> Why do we have to wait instead of just start syncing from bootnodes? I guess it is to prevent overload the bootnodes and being a "good ~samaritan~ client" but I want to be sure 😅

Actually, this is a good option, we will not overload the network because this is just a handshake, which means we will not act as bad actors since this action will only take place at the start of the node. Will change to don't wait for peers/target but request those informations from peers and if there is no peer then we should wait

dot/network/service.go
```diff
@@ -742,3 +743,43 @@ func (s *Service) startProcessingMsg() {
 		}
 	}
 }
+
+func (s *Service) BlockAnnounceHandshake(header *types.Header) error {
+	peers := s.host.peers()
+	if len(peers) < 1 {
```
axaysagathiya, line 749: can you write it as `len(peers) == 0`?

dot/network/service.go
```diff
@@ -742,3 +743,43 @@ func (s *Service) startProcessingMsg() {
...
+	wg := sync.WaitGroup{}
+	wg.Add(len(peers))
+	for _, p := range peers {
+		protocol.peersData.setMutex(p)
+
+		go func(p peer.ID) {
+			defer wg.Done()
+			stream, err := s.sendHandshake(p, handshake, protocol)
+			if err != nil {
+				logger.Tracef("while sending block announce handshake: %v", err)
```
axaysagathiya, line 772:
```suggestion
				logger.Tracef("sending block announce handshake: %s", err)
```

dot/sync/chain_sync_test.go
```diff
@@ -80,58 +80,61 @@ func Test_chainSync_onBlockAnnounce(t *testing.T) {
...
+		// 			stopCh:        make(chan struct{}),
+		// 			pendingBlocks: pendingBlocks,
+		// 			syncMode:      state,
+		// 			peerViewSet:   newPeerViewSet(0),
+		// 			workerPool:    newSyncWorkerPool(NewMockNetwork(nil), NewMockRequestMaker(nil)),
+		// 		}
+		// 	},
+		// 	peerID:              somePeer,
+		// 	blockAnnounceHeader: block2AnnounceHeader,
+		// },
```
axaysagathiya, line 137: Uncomment them or remove them if you don't need them anymore.

dot/network/service.go
dimartiro: I think the preference for using `len(peers) < 1` over `len(peers) == 0` is because error tolerance. This is safety if there is an error somewhere in the code.

There is a concept: [Off-by-one error](https://en.wikipedia.org/wiki/Off-by-one_error) and using `<` could prevent it

dimartiro: On second thought, since the list can't have a negative len, it might make sense to compare with 0 😅

dimartiro (dismissed):
LGTM, agree with Axay's comment to remove / uncomment the test cases

dot/sync/chain_sync_test.go
EclesioMeloJunior: lol, sorry for that!

dimartiro (approved):

dot/sync/chain_sync.go
```diff
@@ -237,18 +238,14 @@ func (cs *chainSync) stop() error {
 	}
 }
 
-func (cs *chainSync) isBootstrap() (bestBlockHeader *types.Header, syncTarget uint,
+func (cs *chainSync) currentSyncInformations() (bestBlockHeader *types.Header, syncTarget uint,
```
jimjbrettj, line 241:
```suggestion
func (cs *chainSync) currentSyncInformation() (bestBlockHeader *types.Header, syncTarget uint,
```

EclesioMeloJunior: I will keep it as it is since this function is returning more than one data, so makes sense to keep it in plural

axaysagathiya (approved):
LGTM

PR #3698, opened 2024-01-16
chore(config): respect the xdg data home specification

### LINKED ISSUE #3678: Use $XDG_DATA_HOME not .gossamer
## Task summary

- config: `XDG_CONFIG_HOME` (linux: `$HOME/.config/gossamer`)
- data: `XDG_DATA_HOME` (linux: `$HOME/.local/share/gossamer`) 
- and so on ...
- don't use `.gossamer`

## Specification reference

- https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html

- https://github.com/adrg/xdg

q9f:
## Changes

uses XDG_DATA_HOME instead of dot-gossamer

it does not yet implement config home as this needs an additional PR

note this might break any staging setups we have @jimjbrettj @EclesioMeloJunior 

## Tests

```sh
```

## Issues

closes #3678 

## Primary Reviewer

@kishansagathiya @jimjbrettj


EclesioMeloJunior (dismissed):

dimartiro (dismissed):

jimjbrettj (approved):
lgtm!

dimartiro:
Just as a reminder, I think we have to retest https://github.com/ChainSafe/gossamer/pull/3251 after this change, we might need to update the paths there too

dimartiro (changes requested):
@q9f I think you have to update our [Dockerfile.staging](https://github.com/ChainSafe/gossamer/blob/942febbb86fb1a6df12ed8ce71edf06ad69072a3/Dockerfile.staging#L4) too

dimartiro (dismissed):

EclesioMeloJunior (approved):

chain/kusama/defaults.go
```diff
@@ -5,20 +5,22 @@ package kusama
...
 
 var (
-	// defaultBasePath Default node base directory path
-	defaultBasePath = "~/.gossamer/kusama"
-	// defaultChainSpec is the default chain-spec json path
-	defaultChainSpec = "./chain/kusama/chain-spec-raw.json"
+	// DefaultBasePath default node base directory path
+	DefaultBasePath = xdg.DataHome + "/gossamer/kusama"
+	// DefaultChainSpec is the default chain-spec json path
+	DefaultChainSpec = "./chain/kusama/chain-spec-raw.json"
```
kishansagathiya, line 16: Why export these vars? They never get used outside of this package!

kishansagathiya (commented):
Looks good! Just not sure why some vars were change to be exported!

chain/kusama/defaults.go
q9f: Good catch. I'm not very fluent in Go; this was on oversight on my end. I'll address this.

jimjbrettj (approved):

config/config.go
```diff
@@ -13,21 +13,20 @@ import (
 	"github.com/ChainSafe/gossamer/lib/genesis"
 	"github.com/ChainSafe/gossamer/lib/os"
 	wazero "github.com/ChainSafe/gossamer/lib/runtime/wazero"
+	"github.com/adrg/xdg"
 )
 
-const (
+var (
```
kishansagathiya: These values will never change. So, they should be kept as const.

dot/network/config.go
```diff
@@ -10,19 +10,20 @@ import (
...
 )
 
-const (
+var (
 	// DefaultKeyFile the default value for KeyFile
 	DefaultKeyFile = "node.key"
 
 	// DefaultBasePath the default value for Config.BasePath
-	DefaultBasePath = "~/.gossamer/gssmr"
+	DefaultBasePath = xdg.DataHome + "/gossamer"
```
kishansagathiya: We can change `DefaultBasePath` as var since xdg.DataHome is a var. But we can keep other values as const.

chain/paseo/defaults.go
```diff
@@ -5,15 +5,12 @@ package paseo
...
 import (
 	cfg "github.com/ChainSafe/gossamer/config"
+	"github.com/adrg/xdg"
 )
 
 var (
-	// defaultName Default node name
-	defaultName = "Paseo"
-	// defaultID Default chain ID
-	defaultID = "paseo"
```
kishansagathiya, line 14: why remove ID and Name?

chain/westend-local/default.go
```diff
@@ -5,18 +5,19 @@ package westendlocal
...
-	// DefaultBasePathBob is the default basepath for the westend local bob node
-	DefaultBasePathBob = "~/.gossamer/westend-local-bob"
-	// DefaultBasePathCharlie is the default basepath for the westend local charlie node
-	DefaultBasePathCharlie = "~/.gossamer/westend-local-charlie"
+	// defaultBasePathAlice is the default basepath for the westend local alice node
+	defaultBasePathAlice = xdg.DataHome + "/gossamer/westend-dev/alice"
+	// defaultBasePathBob is the default basepath for the westend local bob node
+	defaultBasePathBob = xdg.DataHome + "/gossamer/westend-dev/bob"
+	// defaultBasePathCharlie is the default basepath for the westend local charlie node
+	defaultBasePathCharlie = xdg.DataHome + "/gossamer/westend-dev/charlie"
```
kishansagathiya, line 20: Why does path in westend local have `dev` in it? Is this deliberate?

kishansagathiya (commented):
Few more comments. I have pushed a commit that addressed some of those comments (already marked as resolved)
Can approve and merge after this!!

chain/paseo/defaults.go
q9f: Because this is set through the config/cli interface. Name and ID should not be overridden by the chain specification. See #3685

chain/westend-local/default.go
q9f: No, another oversight 🤦🏼 
Thanks

kishansagathiya (approved):

EclesioMeloJunior (approved):

PR #3699, opened 2024-01-16
chore(ci): dont tag tim on every pr

q9f:
rather think about who you want to tag

## Tests

```sh
```

## Primary Reviewer

@P1sar

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -18,4 +18,4 @@ go test -tags integration github.com/ChainSafe/gossamer
 
 <!-- Tag a code owner to review your PR -->
 
-@timwu20
+@


EclesioMeloJunior (approved):

dimartiro (approved):
I think we have to rethink the concept of "primary reviewer"

I have some questions to start thinking about it:
- We are assigning the entire team as reviewers for every new PR, why isn't it enough to have 2 approves from them? 
- In case we need to assign a primary reviewer, why not use PR "assignes" field? that will help filtering too
- What is the criteria to choose the primary reviewer?

jimjbrettj (approved):

EclesioMeloJunior:
> We are assigning the entire team as reviewers for every new PR, why isn't it enough to have 2 approves from them?

The `Reviewers` are basically the users in the codeowners file. Github does that automatically 

> In case we need to assign a primary reviewer, why not use PR "assignes" field? that will help filtering too

Yeah, I would suggest to use `assignes` field

> * What is the criteria to choose the primary reviewer?

The primary reviewer should be someone that has a knowledge about the changes

EclesioMeloJunior:
Yeah, got your point but unless of have only the primaries in the reviewers section why not just merge the code once the primary reviewer approves?

I mean, we, as core contributors, are encouraged to provide good reviews so I don't think removing the team from `reviewers` is a good option because we can feel discouraged from reviewing some PR since it belongs to few "primary reviewers", so tagging the one we want to take a closer look could be better (as we do now) IMO and I believe for the majority of PRs we should not tag a primary reviewer but for specific work streams like `parachain-subsystems` makes sense to have as primary reviewer Kishan

dimartiro:
> Yeah, got your point but unless of have only the primaries in the reviewers section why not just merge the code once the primary reviewer approves?
> 
> I mean, we, as core contributors, are encouraged to provide good reviews so I don't think removing the team from `reviewers` is a good option because we can feel discouraged from reviewing some PR since it belongs to few "primary reviewers", so tagging the one we want to take a closer look could be better (as we do now) IMO and I believe for the majority of PRs we should not tag a primary reviewer but for specific work streams like `parachain-subsystems` makes sense to have as primary reviewer Kishan

I agree to not tag primary reviewers in some PRs.
But I still think we need to improve how we set the primary reviewer in general.

IMO we should take advantage of the GH interface to set it instead of using a tag in the PR description. This has many benefits, Eg: filtering, prioritization and for metrics.

We have 2 options in the GH interface `reviewers` or `assignees`.
Since, as you said, we currently use `reviewers` for a different purpose I propose to start using `assignees` to put the primary reviewers there.

This is also useful when an external dev wants to contribute and doesn't know what name to put in the Primary Reviewer section. We can easily search for unassigned PRs in the PR list and assign it to the right person.

If that's the case, we can use this PR to remove the entire `Primary Reviewer` section from the PRs template. And also update our [contributing guidelines](https://github.com/ChainSafe/gossamer/blob/development/.github/CONTRIBUTING.md) to describe how to use the assignee field
@timwu20 @P1sar thoughts?

kishansagathiya:
> I think we have to rethink the concept of "primary reviewer"

Totally agree. Asking everyone for review is same as asking no one for review. Automatically asking review from everyone hasn't really been practical. In most PR, it's only primary reviewer and one someone else reviewing. Nowadays, we have started tagging two people as primary reviewers to ensure that we get those two reviews. 
When system automatically asks everyone, we have had to find other ways to actually ask for reviews like tag them or message on slack. I say we might as well add only those two people in reviewers list and stop automatically asking for reviews.

P1sar:
> I think we have to rethink the concept of "primary reviewer"
@dimartiro @kishansagathiya Is there a concept of "primary reviewer" in GitHub? Moreover, we do not have `Require review from Code Owners` enabled as protection for the development branch, so having tim there is just creates additional notification for him, and does not bring any additional restrictions. 

If we are really expiriencing problem with everyone being automatically assigned, then we can update CODEOWNERS to include owners per module or remove codeowners at all by allowing PR author to select his own reviewers. But I also beleive this is a separate context that should not be discussed in this task. This task is about this "primary reviewer" tag concept whic is wrong and hence this PR should be merged

axaysagathiya (approved):

kishansagathiya (approved):

kishansagathiya:
> This task is about this "primary reviewer" tag concept whic is wrong and hence this PR should be merged

Sure!!

timwu20:
Agree with a lot of the thoughts expressed here.  Looking forward to the discussion in our next meeting.  I'm open to removing the primary reviewer section completely from the template.  I also like the idea of assigning PRs to certain individuals.  It probably opens up more automation possibilities when it comes to notifying assignees to review.

PR #3707, opened 2024-01-18
chore(dot/parachain/overseer): fix `TestHandleBlockEvents` the flaky test

### LINKED ISSUE #3717: chore(dot/parachain/overseer): fix `TestHandleBlockEvents` the flaky test
## Issue summary

- 

-

axaysagathiya:
## Changes
overseer

- If we close overseerToSub1, subsystem2 could still send a message for subsystem1 via subsystemToOverseer, and it will be panic.
- we can close overseerToSubsystem channels only if the subsystemToOverseer channel is closed.

IMO, we should not close either of them.
That is why I removed the code to close overseerToSubsystem channels.

Used a `waitGroup` in the test to wait till we received all the messages in mocked subsystems before we called the stop function of overseer. 

## Tests

```sh
```

## Issues

closes #3717
## Primary Reviewer

@kishansagathiya @EclesioMeloJunior

### DIFF
--- a/dot/parachain/overseer/overseer.go
+++ b/dot/parachain/overseer/overseer.go
@@ -232,10 +232,6 @@ func (o *Overseer) Stop() error {
 	// close the errorChan to unblock any listeners on the errChan
 	close(o.errChan)
 
-	for _, sub := range o.subsystems {
-		close(sub)
-	}
-
 	// wait for subsystems to stop
 	// TODO: determine reasonable timeout duration for production, currently this is just for testing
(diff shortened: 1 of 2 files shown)


kishansagathiya (approved):
Currently `TestHandleBlockEvents` fails a lot of times in CI. This PR shows a bigger problem of `how to gracefully shutdown overseer and subsystems`.

Even though, this solution is not ideal, since it keeps the channels open, it is necessary to unblock other PRs. Currently, we are having to run CI some 6-7 to pass it due to this test.
Hence approving this. Will tackle the problem to shutting down overseer and  subsystems separately.

Issue #3713, opened 2024-01-18
Create script to retrieve the trie state at a given block height


jimjbrettj:
## Issue summary

Currently, the process of getting the state from a particular block height means running a node until we get to that height and dumping the state once we get there. This means we have to restart a node every time and wait for syncing to occur, which is inefficient.

We have an RPC endpoint that allows us to get the state from a particular block height. We should write a script then that interacts with an already running node to retrieve the state from a given block as well as optionally assert the computed state root against the expected state root.

jimjbrettj:
Closed via #3714

PR #3714, opened 2024-01-18
feat(tests/scripts): create script to retrieve trie state via rpc

jimjbrettj:
## Changes

- Created a script to query a running node via rpc to retrieve the trie state at a given block and write it to a file
- I also added the ability to compute and verify the state root from the retrieved state

## Tests

Start up a gossamer node and execute the script against it (or any running gossamer node)

## Issues

#3713

## Primary Reviewer

@EclesioMeloJunior


tests/scripts/trieStateScript.go
```diff
@@ -0,0 +1,102 @@
...
+
+	newTrie, err := trie.LoadFromMap(entries)
+	if err != nil {
+		panic(fmt.Sprintf("loading trie from map %v", err))
+	}
+
+	trieHash := newTrie.MustHash(trie.V0.MaxInlineValue())
+	if expectedStateRoot != trieHash.String() {
+		panic("westendDevStateRoot does not match trieHash")
+	}
```
EclesioMeloJunior, line 101: Also this could be a function that is only called if the `expectedStateRoot` value is provided

EclesioMeloJunior: The trie version could be a CLI parameter

EclesioMeloJunior (changes requested):
Nice script! 🔥 🚀 🚀 

Few comments and suggestions to improve the script versatility, also could you write down some test cases?

dimartiro:
What if we move it to a `scripts` folder (at root level) instead of `tests/scripts`
Also we can create a `make` script to run it

tests/scripts/trieStateScript.go
```diff
@@ -0,0 +1,102 @@
...
+
+	var response modules.StateTrieResponse
+	fetchWithTimeout(ctx, "state_trie", params, &response)
+
+	encResponse, err := json.Marshal(response)
+	if err != nil {
+		panic(fmt.Sprintf("json marshalling response %v", err))
+	}
+
+	err = os.WriteFile("trie_state_data", encResponse, 0o600)
```
dimartiro, line 75: Can we take the `dst` path from `args` too to make it parametrizable when we run the script?

tests/scripts/trieStateScript.go
```diff
@@ -0,0 +1,102 @@
...
+		}
+		entries[common.BytesToHex(entry.Key)] = common.BytesToHex(entry.Value)
+	}
+
+	newTrie, err := trie.LoadFromMap(entries)
+	if err != nil {
+		panic(fmt.Sprintf("loading trie from map %v", err))
+	}
+
+	trieHash := newTrie.MustHash(trie.V0.MaxInlineValue())
```
dimartiro, line 98: I think this is more clear

```suggestion
	trieHash := trie.V0.MustHash(newTrie)
```

dimartiro (changes requested):

tests/scripts/trieStateScript.go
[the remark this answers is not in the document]
dimartiro: +1 and I think this could be more clear to know what are the permissions

```suggestion
        perm := os.ModeOwnerRead | os.ModeOwnerWrite // 0o600
	err = os.WriteFile("trie_state_data", encResponse, perm)
```

jimjbrettj:
Responded to all feedback except @dimartiro given how simple of a script this is I did not see the benefit of using a makefile

tests/scripts/trieStateScript.go
dimartiro: Also I think we should always use the `MustHash` from the version instead using the one from the trie, and make the one from the trie private. But that's a problem for a different PR 😄

EclesioMeloJunior (approved):

scripts/trieStateScript.go
```diff
@@ -0,0 +1,147 @@
...
+func fetchTrieState(ctx context.Context, blockHash common.Hash, destination string) modules.StateTrieResponse {
+	params := fmt.Sprintf(`["%s"]`, blockHash)
+	var response modules.StateTrieResponse
+	fetchWithTimeout(ctx, "state_trie", params, &response)
+
+	writeTrieState(response, destination)
+	return response
+}
+
+func compareStateRoots(response modules.StateTrieResponse, expectedStateRoot common.Hash, trieVersion int) {
```
dimartiro, line 70:
```suggestion
func compareStateRoots(response modules.StateTrieResponse, expectedStateRoot common.Hash, trieVersion trie.TrieLayout) {
```

scripts/trieStateScript.go
```diff
@@ -0,0 +1,147 @@
...
+	}
+
+	trieHash := common.Hash{} //nolint
+	if trieVersion == 0 {
+		trieHash = trie.V0.MustHash(newTrie)
+	} else if trieVersion == 1 {
+		trieHash = trie.V1.MustHash(newTrie)
+	} else {
+		panic("invalid trie version")
+	}
```
dimartiro, line 95:
```suggestion
	trieHash := trieVersion.MustHash(newTrie)
```

scripts/trieStateScript.go
```diff
@@ -0,0 +1,147 @@
...
+
+	blockHash, err := common.HexToHash(os.Args[1])
+	if err != nil {
+		panic("block hash must be in hex format")
+	}
+
+	destinationFile := os.Args[2]
+	expectedStateRoot := common.Hash{}
+	var trieVersion int
```
dimartiro, line 126:
```suggestion
	var trieVersion trie.TrieLayout
```

scripts/trieStateScript.go
```diff
@@ -0,0 +1,147 @@
...
+	if len(os.Args) == 5 {
+		expectedStateRoot, err = common.HexToHash(os.Args[3])
+		if err != nil {
+			panic("expected state root must be in hex format")
+		}
+
+		trieVersion, err = strconv.Atoi(os.Args[4])
```
dimartiro, line 133:
```suggestion
		trieVersion, err = trie.ParseVersion(os.Args[4])
```

dimartiro (changes requested):
I think using `trie.TrieLayout` could make it more scalable and safety.

Also, I think the file name should be `trie_state_script.go` and `trie_state_script_test.go`

dimartiro (approved):

PR #3719, opened 2024-01-22
Release 0.9.0

P1sar:
## Tests

```sh
```

## Primary Reviewer

@


q9f:
Release checklist: 
- Set version meta to something meaningful, e.g., `alpha` - https://github.com/ChainSafe/gossamer/blob/development/config/version.go#L16
- Create PR to set `development` version to 0.10.0

dimartiro:
I think we already have an [automation](https://github.com/ChainSafe/gossamer/blob/development/.github/workflows/release.yml) to set the release semver based on commits messages.

We could add more steps there to automatically set [major, minor and patch](https://github.com/ChainSafe/gossamer/blob/development/config/version.go#L13-L15) based on that generated version

q9f:
As I said in the planning, I don't think releases should be automated. Also, we haven't set a release version since 0.3.2 - so lets do it manually correctly first before considering automation.

EclesioMeloJunior (approved):

PR #3724, opened 2024-01-25
fixing some syntactic changes

kishansagathiya:
## Tests

```sh
```

## Primary Reviewer

@axaysagathiya @edwardmack

### DIFF
--- a/dot/parachain/availability-store/availability_store.go
+++ b/dot/parachain/availability-store/availability_store.go
@@ -382,9 +382,13 @@ func branchesFromChunks(chunks [][]byte) (branches, error) {
 			return branches{}, fmt.Errorf("putting chunk %d in trie: %w", i, err)
 		}
 	}
+	branchHash, err := trie.V1.Hash(tr)
+	if err != nil {
+		return branches{}, fmt.Errorf("hashing trie: %w", err)
+	}
 	b := branches{
 		trieStorage: tr,
-		root:        trie.V0.MustHash(*tr),
+		root:        branchHash,
 		chunks:      chunks,
 		currentPos:  0,
 	}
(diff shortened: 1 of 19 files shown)


EclesioMeloJunior (approved):

dot/parachain/availability-store/availability_store.go
```diff
@@ -384,7 +384,7 @@ func branchesFromChunks(chunks [][]byte) (branches, error) {
 	}
 	b := branches{
 		trieStorage: tr,
-		root:        tr.MustHash(),
+		root:        tr.MustHash(trie.NoMaxInlineValueSize),
```
dimartiro, line 387: I think is more clear to use the version directly 

```suggestion
		root:        trie.V0.MustHash(*tr),
```

And also I have a question, how do you know you should use the V0 to hash the trie?

dimartiro (changes requested):

dot/parachain/availability-store/availability_store.go
kishansagathiya: Looked more into this. It looks like the IDE thinks it is v1.
https://github.com/paritytech/polkadot-sdk/blob/7035034710ecb9c6a786284e5f771364c520598d/polkadot/erasure-coding/src/lib.rs#L278
Screenshot from 2024-02-12 12-39-14

LayoutV1 is for V1 trie layout.

dimartiro: Actually using V0 is the same that using `MustHash` with `NoMaxInlineValueSize` https://github.com/ChainSafe/gossamer/blob/f5b9c4c89557d6332396de9e9fa3a8cf782f7259/lib/trie/layout.go#L94-L96

Or we can wait until merge this [PR](https://github.com/ChainSafe/gossamer/pull/3739) from Eclesio and we can get rid of this parameter

dimartiro (approved):

PR #3733, opened 2024-02-01
fix: add a limit of number of bytes while scale decoding a slice

### LINKED ISSUE #3476: Panic “len out of range” while Unmarshaling Scale leading to a denial of service (DoS)
# Panic “len out of range” while Unmarshaling Scale leading to a denial of service (DoS)

We ([@FuzzingLabs](https://github.com/FuzzingLabs)) found a “len out of range” Panic issue while Unmarshaling Scale.

## Expected behavior

The function should check the length of the slice.

## Environment

- Ubuntu 22.04.1
- go version go1.21.0 linux/amd64

## Steps to reproduce

This test crashes.

```go
func TestScaleUnmarshal(t *testing.T) {
  block := NewBlock(*NewEmptyHeader(), Body{})
  err := scale.Unmarshal([]byte{48, 48, 48, 48, 48, 48, 48, 48, 48, 48, 48, 48, 48, 48, 48, 48, 48, 48, 48, 48, 48, 48, 48, 48, 48, 48, 48, 48, 48, 48, 48, 48, 48, 48, 48, 48, 48, 48, 48, 48, 48, 48, 48, 48, 48, 48, 48, 48, 48, 48, 48, 48, 48, 48, 48, 48, 48, 48, 48, 48, 48, 48, 48, 48, 48, 48, 48, 48, 48, 48, 48, 48, 48, 48, 48, 48, 48, 48, 48, 48, 48, 48, 48, 48, 48, 48, 48, 48, 48, 48, 48, 48, 48, 48, 48, 48, 48, 48, 4, 48, 48, 48, 48, 19, 48, 48, 48, 48, 48, 48, 48, 48}, &block)
  if err != nil {
    t.Fatal(err)
  }
}
```

## Root cause

The `decodeBytes` function tries to call `make` on the length read from the given bytes. By doing so it reads a very big value and fails to create a slice.

https://github.com/ChainSafe/gossamer/blob/d5c2c7fe3132b97fc6fa208bebcd254582ffb611/pkg/scale/decode.go#L610

```
panic: runtime error: makeslice: len out of range [recovered]
	panic: runtime error: makeslice: len out of range
```

## Detailed behavior

```
panic: runtime error: makeslice: len out of range [recovered]
	panic: runtime error: makeslice: len out of range

goroutine 8 [running]:
testing.tRunner.func1.2({0x954620, 0xab3230})
	/snap/go/10319/src/testing/testing.go:1545 +0x238
testing.tRunner.func1()
	/snap/go/10319/src/testing/testing.go:1548 +0x397
panic({0x954620?, 0xab3230?})
	/snap/go/10319/src/runtime/panic.go:914 +0x21f
github.com/ChainSafe/gossamer/pkg/scale.(*decodeState).decodeBytes(0xc00037c370, {0x91f1c0?, 0xc00007d0c8?, 0x20?})
	/home/tanguy/Documents/polkadot/gossamer/pkg/scale/decode.go:615 +0xca
github.com/ChainSafe/gossamer/pkg/scale.(*decodeState).unmarshal(0x91f1c0?, {0x91f1c0?, 0xc00007d0c8?, 0x4d3025?})
	/home/tanguy/Documents/polkadot/gossamer/pkg/scale/decode.go:126 +0x2a5
github.com/ChainSafe/gossamer/pkg/scale.(*decodeState).decodeStruct(0x991b40?, {0x991b40?, 0xc00007d040?, 0xac0e48?})
	/home/tanguy/Documents/polkadot/gossamer/pkg/scale/decode.go:490 +0x357
github.com/ChainSafe/gossamer/pkg/scale.(*decodeState).unmarshal(0x9577a0?, {0x991b40?, 0xc00007d040?, 0x991b40?})
	/home/tanguy/Documents/polkadot/gossamer/pkg/scale/decode.go:151 +0x63e
github.com/ChainSafe/gossamer/pkg/scale.(*decodeState).decodeVaryingDataType(0xc00037c370, {0x9781c0?, 0xc0000136b0?, 0x18?})
	/home/tanguy/Documents/polkadot/gossamer/pkg/scale/decode.go:394 +0x21b
github.com/ChainSafe/gossamer/pkg/scale.(*decodeState).unmarshal(0x9646c0?, {0x9781c0?, 0xc0000136b0?, 0x9781c0?})
	/home/tanguy/Documents/polkadot/gossamer/pkg/scale/decode.go:134 +0x105
github.com/ChainSafe/gossamer/pkg/scale.(*decodeState).decodeVaryingDataTypeSlice(0x9841a0?, {0x9841a0?, 0xc0000004e8?, 0x2?})
	/home/tanguy/Documents/polkadot/gossamer/pkg/scale/decode.go:339 +0x29f
github.com/ChainSafe/gossamer/pkg/scale.(*decodeState).unmarshal(0x9841a0?, {0x9841a0?, 0xc0000004e8?, 0x9841a0?})
	/home/tanguy/Documents/polkadot/gossamer/pkg/scale/decode.go:136 +0x20d
github.com/ChainSafe/gossamer/pkg/scale.(*decodeState).decodeStruct(0x9b1a20?, {0x9b1a20?, 0xc0002ccdd0?, 0xac0e48?})
	/home/tanguy/Documents/polkadot/gossamer/pkg/scale/decode.go:490 +0x357
github.com/ChainSafe/gossamer/pkg/scale.(*decodeState).unmarshal(0x9b1a20?, {0x9b1a20?, 0xc0002ccdd0?, 0x9b1a20?})
	/home/tanguy/Documents/polkadot/gossamer/pkg/scale/decode.go:151 +0x63e
github.com/ChainSafe/gossamer/pkg/scale.(*decodeState).decodeStruct(0x974de0?, {0x974de0?, 0xc0002cca90?, 0xac0e48?})
	/home/tanguy/Documents/polkadot/gossamer/pkg/scale/decode.go:490 +0x357
github.com/ChainSafe/gossamer/pkg/scale.(*decodeState).unmarshal(0x985b60?, {0x974de0?, 0xc0002cca90?, 0x1?})
	/home/tanguy/Documents/polkadot/gossamer/pkg/scale/decode.go:151 +0x63e
github.com/ChainSafe/gossamer/pkg/scale.Unmarshal({0xc000032620, 0x70, 0x70}, {0x985b60, 0xc0002cca90})
	/home/tanguy/Documents/polkadot/gossamer/pkg/scale/decode.go:74 +0x1ed
github.com/ChainSafe/gossamer/dot/types.FuzzUnmarshalBlock.func1(0xc0001f5520, {0xc000032620, 0x70, 0x70})
	/home/tanguy/Documents/polkadot/gossamer/dot/types/block_test.go:141 +0x27c
reflect.Value.call({0x932600?, 0xa29dd8?, 0x13?}, {0x9e247e, 0x4}, {0xc0002caae0, 0x2, 0x2?})
	/snap/go/10319/src/reflect/value.go:596 +0xce7
reflect.Value.Call({0x932600?, 0xa29dd8?, 0xe1cbe0?}, {0xc0002caae0?, 0x9e0fa0?, 0x527e8d?})
	/snap/go/10319/src/reflect/value.go:380 +0xb9
testing.(*F).Fuzz.func1.1(0x0?)
	/snap/go/10319/src/testing/fuzz.go:335 +0x347
testing.tRunner(0xc0001f5520, 0xc0000b4360)
	/snap/go/10319/src/testing/testing.go:1595 +0xff
created by testing.(*F).Fuzz.func1 in goroutine 7
	/snap/go/10319/src/testing/fuzz.go:322 +0x597
exit status 2
```

kishansagathiya:
While scale decoding we first read the length of bytes to decode and then we decode that many bytes.

Someone could ask us to decode some malicious bytes such that the read length is unreasonably big. In such case, we would have to create a byte slice as big as the length. So, we would either panic because of `makeslice: len out of range` or because the asked length would be more than the memory we have available in our machine.

Both of these are the situations we don't want. Hence, I have put a limit on decodable bytes (maxDecodableBytes), 16MB. I have kept it 16MB because for the usecase of polkadot specs, we would never need to decode a bigger set of bytes.

## Tests

```sh
```

## Issues

Fixes #3476 

## Primary Reviewer

@timwu20 @EclesioMeloJunior

### DIFF
--- a/pkg/scale/decode.go
+++ b/pkg/scale/decode.go
@@ -493,10 +493,12 @@ func (ds *decodeState) decodeUint(dstv reflect.Value) (err error) {
 	var value uint64
 	switch mode {
 	case 0:
-		// 0b00: single-byte mode; upper six bits are the LE encoding of the value (valid only for values of 0-63).
+		// 0b00: single-byte mode; upper six bits are the LE encoding of the value (valid only for
+		// values of 0-63).
 		value = uint64(prefix >> 2)
 	case 1:
-		// 0b01: two-byte mode: upper six bits and the following byte is the LE encoding of the value (valid only for values 64-(2**14-1))
+		// 0b01: two-byte mode: upper six bits and the following byte is the LE encoding of the
+		// value (valid only for values 64-(2**14-1))
 		buf, err := ds.ReadByte()
 		if err != nil {
 			return fmt.Errorf("reading byte: %w", err)
@@ -506,7 +508,8 @@ func (ds *decodeState) decodeUint(dstv reflect.Value) (err error) {
 			return fmt.Errorf("%w: %d (%b)", ErrU16OutOfRange, value, value)
 		}
 	case 2:
-		// 0b10: four-byte mode: upper six bits and the following three bytes are the LE encoding of the value (valid only for values (2**14)-(2**30-1)).
+		// 0b10: four-byte mode: upper six bits and the following three bytes are the LE encoding
+		// of the value (valid only for values (2**14)-(2**30-1)).
 		buf := make([]byte, 3)
 		_, err = ds.Read(buf)
 		if err != nil {
@@ -517,8 +520,9 @@ func (ds *decodeState) decodeUint(dstv reflect.Value) (err error) {
 			return fmt.Errorf("%w: %d (%b)", ErrU32OutOfRange, value, value)
 		}
 	case 3:
-		// 0b11: Big-integer mode: The upper six bits are the number of bytes following, plus four. The value is contained, LE encoded,
-		// in the bytes following. The final (most significant) byte must be non-zero. Valid only for values (2**30)-(2**536-1).
+		// 0b11: Big-integer mode: The upper six bits are the number of bytes following, plus four.
+		// The value is contained, LE encoded, in the bytes following. The final (most significant)
+		// byte must be non-zero. Valid only for values (2**30)-(2**536-1).
 		byteLen := (prefix >> 2) + 4
 		buf := make([]byte, byteLen)
 		_, err = ds.Read(buf)
--- a/pkg/scale/decode.go
+++ b/pkg/scale/decode.go
@@ -13,6 +13,8 @@ import (
 	"reflect"
 )
 
+var maxDecodableBytes = uint(1024 * 1024 * 16)
+
 // indirect walks down v allocating pointers as needed,
 // until it gets to a non-pointer.
 func indirect(dstv reflect.Value) (elem reflect.Value) {
@@ -557,7 +559,7 @@ func (ds *decodeState) decodeLength() (l uint, err error) {
 	dstv := reflect.New(reflect.TypeOf(l))
 	err = ds.decodeUint(dstv.Elem())
 	if err != nil {
-		return
+		return 0, fmt.Errorf("decoding uint: %w", err)
 	}
 	l = dstv.Elem().Interface().(uint)
 	return
@@ -568,6 +570,10 @@ func (ds *decodeState) decodeBytes(dstv reflect.Value) (err error) {
 	length, err := ds.decodeLength()
 	if err != nil {
 		return
+	}
+
+	if length > maxDecodableBytes {
+		return fmt.Errorf("byte array length %d exceeds maximum of %d", length, maxDecodableBytes)
 	}
 
 	b := make([]byte, length)
--- a/pkg/scale/decode.go
+++ b/pkg/scale/decode.go
@@ -14,11 +14,6 @@ import (
 	"reflect"
 )
 
-// var maxDecodableBytes = uint(1024 * 1024 * 16)
-var maxDecodableBytes = uint(math.MaxInt32)
-
-const maxPreAllocation = 4 * 1024
-
 // indirect walks down v allocating pointers as needed,
 // until it gets to a non-pointer.
 func indirect(dstv reflect.Value) (elem reflect.Value) {
@@ -582,12 +577,10 @@ func (ds *decodeState) decodeBytes(dstv reflect.Value) (err error) {
 		return
 	}
 
-	fmt.Println("length", length)
-	fmt.Println("maxDecodableBytes", maxDecodableBytes)
-	fmt.Println("math.MaxUint32", math.MaxUint32)
-	// if length > maxDecodableBytes {
-	// 	return fmt.Errorf("byte array length %d exceeds maximum of %d", length, maxDecodableBytes)
-	// }
+	// bytes length in encoded as Compact<u32>, so it can't be more than math.MaxUint32
+	if length > math.MaxUint32 {
+		return fmt.Errorf("byte array length %d exceeds max value of uint32", length)
+	}
 
 	b := make([]byte, length)
 
--- a/pkg/scale/decode.go
+++ b/pkg/scale/decode.go
@@ -9,11 +9,15 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"math"
 	"math/big"
 	"reflect"
 )
 
-var maxDecodableBytes = uint(1024 * 1024 * 16)
+// var maxDecodableBytes = uint(1024 * 1024 * 16)
+var maxDecodableBytes = uint(math.MaxInt32)
+
+const maxPreAllocation = 4 * 1024
 
 // indirect walks down v allocating pointers as needed,
 // until it gets to a non-pointer.
@@ -477,6 +481,7 @@ func (ds *decodeState) decodeBool(dstv reflect.Value) (err error) {
 	return
 }
 
+// TODO: Should this be renamed to decodeCompactInt?
 // decodeUint will decode unsigned integer
 func (ds *decodeState) decodeUint(dstv reflect.Value) (err error) {
 	const maxUint32 = ^uint32(0)
@@ -493,8 +498,10 @@ func (ds *decodeState) decodeUint(dstv reflect.Value) (err error) {
 	var value uint64
 	switch mode {
 	case 0:
+		// 0b00: single-byte mode; upper six bits are the LE encoding of the value (valid only for values of 0-63).
 		value = uint64(prefix >> 2)
 	case 1:
+		// 0b01: two-byte mode: upper six bits and the following byte is the LE encoding of the value (valid only for values 64-(2**14-1))
 		buf, err := ds.ReadByte()
 		if err != nil {
 			return fmt.Errorf("reading byte: %w", err)
@@ -504,6 +511,7 @@ func (ds *decodeState) decodeUint(dstv reflect.Value) (err error) {
 			return fmt.Errorf("%w: %d (%b)", ErrU16OutOfRange, value, value)
 		}
 	case 2:
+		// 0b10: four-byte mode: upper six bits and the following three bytes are the LE encoding of the value (valid only for values (2**14)-(2**30-1)).
 		buf := make([]byte, 3)
 		_, err = ds.Read(buf)
 		if err != nil {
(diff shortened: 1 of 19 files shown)


pkg/scale/decode.go
```diff
@@ -13,6 +13,8 @@ import (
 	"reflect"
 )
 
+var maxDecodableBytes = uint(1024 * 1024 * 16)
```
EclesioMeloJunior, line 16: Given that scale encoded slices are prepended with its size and given the fact that golang's  `make([]T, size)` only accepts the size as `int` I would suggest using `math.MaxInt`

EclesioMeloJunior (approved):
LGTM!

pkg/scale/decode.go
kishansagathiya: math.MaxInt is 9223372036854775807. Creating a byte slice of that length would require significant memory (8589934592 GB).

jimjbrettj: You said 16 was the largest response we have seen so far right? Is it a worthy trade off to increase it a bit in case we receiver larger messages in the future or is this not to much of a concern? I like the point you made about just worrying about polkadot support for now, just was wondering if we could receive bigger messages within that context

jimjbrettj (approved):

pkg/scale/decode.go
dimartiro: How is parity's scale crate handling this case?

kishansagathiya: > How is parity's scale crate handling this case?
It panics!

EclesioMeloJunior: @kishansagathiya I agree that this is a large value but since our `pkg/scale` is a stand-alone lib I don't think is a good approach to put 16kb as the `maxDecodableBytes` given that this is a limit specific for our use case/polkadot implementation. Other libraries that use our scale pkg may not have the same limit or any limit.

Another point is: `parity-scale-codec` has a limitation bound to `U32::MAX` for encoding and decoding
- encoding: see https://github.com/paritytech/parity-scale-codec/blob/8c3cc7f2311c8ea7522f4edc21d9cf1f7b661519/src/codec.rs#L1088
- decoding: see https://github.com/paritytech/parity-scale-codec/blob/8c3cc7f2311c8ea7522f4edc21d9cf1f7b661519/src/codec.rs#L1181

So I keep my suggestion to add a limitation in the stand-alone scale pkg lib to `math.MaxUint32`

EclesioMeloJunior (changes requested):

pkg/scale/decode.go
dimartiro: Isn't it a known issue? https://github.com/paritytech/parity-scale-codec?tab=readme-ov-file#known-issues

kishansagathiya: > 16kb as the maxDecodableBytes
16MB, not KB

> Another point is: parity-scale-codec has a limitation bound to U32::MAX for encoding and decoding

True. The problem is to prevent us from an attacker who misuses this and make us create a giant array for no good reason.
I can keep u32 max as the bound. I will have to use read bytes in a small buffers instead of declaring a huge array at the same time. That's what parity-scale-codec does. I will push with those changes.

tagging @timwu20 to confirm.  

> Isn't it a known issue? https://github.com/paritytech/parity-scale-codec?tab=readme-ov-file#known-issues

Yeah, I think they are saying that, even though we support such ridiculous size, chances are your machine memory can't take it so you will get stack overflow.

kishansagathiya: I think I have figured out why this is happening. 
I rust the error is `u32 out of range`, because block number there is u32. Inside header, we have kept block number uint. When we decode using `decodeUint`, the execution goes to https://github.com/ChainSafe/gossamer/blob/1c0c20347b4cbabcae68c2d733ff452055af6459/pkg/scale/decode.go#L539, which is for uint64.

I wonder if we should change block number to uint32 instead of just uint.

kishansagathiya: I think I have figured out why this is happening. 
I rust the error is `u32 out of range`, because block number there is u32. Inside header, we have kept block number uint. When we decode using `decodeUint`, the execution goes to https://github.com/ChainSafe/gossamer/blob/1c0c20347b4cbabcae68c2d733ff452055af6459/pkg/scale/decode.go#L539, which is for uint64.

I wonder if we should change block number to uint32 instead of just uint.

kishansagathiya: I think I know why this is happening. 
In rust, the error is `u32 out of range`. I think it decodes length as `Compact<u32>`. 

In `decodeUint`, we have combined logic to decode both `Compact<u32>` and `Compact<u64>`. The execution in gossamer goes to https://github.com/ChainSafe/gossamer/blob/2fcb906aec425e1658e0e559a92da5813309ffbc/pkg/scale/decode.go#L539, which is for uint64.

dimartiro: I believe they are using `Compact<u64>` based on this:

https://github.com/paritytech/scale-decode/blob/c5744c9e39e92a4f75e86c1b226fa162fed4e08b/scale-decode/src/visitor/types/sequence.rs#L33-L45

kishansagathiya: > I believe they are using Compact<u64> based on this:

Naah, Look at the error that I get while trying to decode this in rust. It says `out of range decoding Compact<u32>`, so, it is trying to decode Compact<u32>. 
Plus, I have been debugging with breakpoints, so I know execution is going to  https://github.com/paritytech/parity-scale-codec/blob/8c3cc7f2311c8ea7522f4edc21d9cf1f7b661519/src/compact.rs#L498 and fails at https://github.com/paritytech/parity-scale-codec/blob/8c3cc7f2311c8ea7522f4edc21d9cf1f7b661519/src/compact.rs#L531
```
running 1 test
thread 'generic::header::tests::ensure_format_is_unchanged' panicked at 'called `Result::unwrap()` on an `Err` value: Error { cause: Some(Error { cause: Some(Error { cause: Some(Error { cause: None, desc: "out of range decoding Compact<u32>" }), desc: "Could not decode `Digest::logs`" }), desc: "Could not decode `Header::digest`" }), desc: "Could not decode `Block::header`" }', substrate/primitives/runtime/src/generic/header.rs:214:39
stack backtrace:
   0: rust_begin_unwind
             at /rustc/8ede3aae28fe6e4d52b38157d7bfe0d3bceef225/library/std/src/panicking.rs:593:5
   1: core::panicking::panic_fmt
             at /rustc/8ede3aae28fe6e4d52b38157d7bfe0d3bceef225/library/core/src/panicking.rs:67:14
   2: core::result::unwrap_failed
             at /rustc/8ede3aae28fe6e4d52b38157d7bfe0d3bceef225/library/core/src/result.rs:1651:5
   3: core::result::Result<T,E>::unwrap
             at /rustc/8ede3aae28fe6e4d52b38157d7bfe0d3bceef225/library/core/src/result.rs:1076:23
   4: sp_runtime::generic::header::tests::ensure_format_is_unchanged
             at ./src/generic/header.rs:214:17
   5: sp_runtime::generic::header::tests::ensure_format_is_unchanged::{{closure}}
             at ./src/generic/header.rs:209:34
   6: core::ops::function::FnOnce::call_once
             at /rustc/8ede3aae28fe6e4d52b38157d7bfe0d3bceef225/library/core/src/ops/function.rs:250:5
   7: core::ops::function::FnOnce::call_once
             at /rustc/8ede3aae28fe6e4d52b38157d7bfe0d3bceef225/library/core/src/ops/function.rs:250:5
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
test generic::header::tests::ensure_format_is_unchanged ... FAILED
```

kishansagathiya: https://github.com/paritytech/parity-scale-codec/blob/8c3cc7f2311c8ea7522f4edc21d9cf1f7b661519/src/codec.rs#L1180-L1181

This shows that length is Compact<u32>

dimartiro: You are totally right Kishan! The one I sent was for sequences not Vec.

Good finding! 👏🏻

EclesioMeloJunior (approved):

dimartiro (approved):

PR #3736, opened 2024-02-02
fix: segfault on node restart

jimjbrettj:
## Changes

- We were running into segfaults when trying to restart the gossamer node, this PR resolves this problem
- The issue was when calling pebbles `Get()` function, we were calling `Close()` and then copying the got value. However, the slice is only valid until we call close so we needed to reverse this order.
- See: https://pkg.go.dev/github.com/cockroachdb/pebble@v1.0.0#DB.Get

## Tests

Start a gossamer node, stop it, then restart it
- ./bin/gossamer init --chain paseo
- ./bin/gossamer --chain paseo
- stop node
- ./bin/gossamer --chain paseo

Make sure there are no segfaults. Please stress-test whatever start/stop cases you can think of :) 

## Issues

#3709 

## Primary Reviewer

@EclesioMeloJunior

### DIFF
--- a/dot/state/service.go
+++ b/dot/state/service.go
@@ -238,6 +238,7 @@ func (s *Service) Rewind(toBlock uint) error {
 
 // Stop closes each state database
 func (s *Service) Stop() error {
+	fmt.Println("Stopping state service")
 	close(s.closeCh)
 
 	hash, err := s.Block.GetHighestFinalisedHash()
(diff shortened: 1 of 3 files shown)


EclesioMeloJunior (approved):
Really great finding!

dimartiro (approved):

PR #3737, opened 2024-02-02
fix: use last finalized block on startup

jimjbrettj:
## Changes

- When starting the syncing service we should be using the last finalized block that we have, not the best block that we have. This PR makes this change. 
- After startup we are okay to go back to using the best block

## Tests

```sh
```

## Primary Reviewer

@EclesioMeloJunior


EclesioMeloJunior (changes requested):
Instead of using this flag property (`isStartup`) I would suggest to: inside the `bootstrapSync` method inserting the call to `GetHighestFinalisedHeader` before starting the for loop, which means that the process will start using the latest finalized header.

Also, I've suggested to transform `currentSyncInformation` method into `isBootstrapSync` where it doesn't need to return the block but only a bool indicating if we keep on bootstrap or change

```go
func (cs *chainSync) bootstrapSync() {
  currentBlock, err := cs.blockState.GetHighestFinalisedHeader()
  if err != nil {
    panic("cannot find highest finalised header")
  }

  for {
    ...
    isBootstrap := cs.isBootstrapSync(currentBlock)
    
    if isBootstrap {
      cs.workerPool.useConnectedPeers()
      err = cs.requestMaxBlocksFrom(bestBlockHeader, networkInitialSync)
      if err != nil {
        logger.Errorf("requesting max blocks from best block header: %s", err)
      }

      currentBlock, err = cs.blockState.BestBlockHeader()
      if err != nil {
        return nil, false, fmt.Errorf("getting best block header: %w", err)
      }
    } else {
      ...
    }
  }
}

func (cs *chainSync) isBootstrapSync(syncHeader *types.Header) bool {
  syncTarget := cs.peerViewSet.getTarget()
  return syncHeader.Number+network.MaxBlocksInResponse < syncTarget
}
````

EclesioMeloJunior (approved):
LGTM!

dimartiro (approved):

Issue #3711, opened 2024-01-18
westend: Gossamer fails with `Storage root must match that calculated` at block `#14576856`


EclesioMeloJunior:
## Describe the bug

- After Gossamer detecting the runtime upgrade 

```
2024-01-18T17:30:43Z INFO 🔄 detected runtime code change, upgrading with block 0x0283b22f28a02825206b5cfd1668e117c4c1d3b49ec02a5b4731aff2c3ac1c40 from previous code hash 0xac5c0d2e187812ca20d6ea0c91fcc67d7eb4feb33edbfd4ddc0abf93bd4ac1b6 to new code hash 0x6eb47852c4f8f8b1ec2cb6084564f88c353f5d6d3a5edd1c85dcd744847652cb... pkg=state
```

It fails with:

```
January 18, 2024 at 13:30 (UTC-4:00) | 2024-01-18T17:30:45Z ERROR requesting max blocks from best block header: while handling workers results: while handling ready block: processing block data with header and body: handling block: failed to execute block 14576856: running runtime function: wasm error: unreachable | 22d1a73da4c241c39ca263c523451b1f | gossamer-westend
-- | -- | -- | --
January 18, 2024 at 13:30 (UTC-4:00) | wasm stack trace: | 22d1a73da4c241c39ca263c523451b1f | gossamer-westend
January 18, 2024 at 13:30 (UTC-4:00) | .rust_begin_unwind(i32) | 22d1a73da4c241c39ca263c523451b1f | gossamer-westend
January 18, 2024 at 13:30 (UTC-4:00) | ._ZN4core9panicking9panic_fmt17h32de9c76c9d5eb0cE(i32,i32) | 22d1a73da4c241c39ca263c523451b1f | gossamer-westend
January 18, 2024 at 13:30 (UTC-4:00) | ._ZN15frame_executive104Executive$LT$System$C$Block$C$Context$C$UnsignedValidator$C$AllPalletsWithSystem$C$COnRuntimeUpgrade$GT$13execute_block17h8d959aeed6ee6a68E(i32) | 22d1a73da4c241c39ca263c523451b1f | gossamer-westend
January 18, 2024 at 13:30 (UTC-4:00) | .Core_execute_block(i32,i32) i64 pkg=sync | 22d1a73da4c241c39ca263c523451b1f | gossamer-westend
[...]

PR #3739, opened 2024-02-06
fix(lib/trie): use `MustBeHashed` for V1 trie nodes with larger storage values

EclesioMeloJunior:
## Changes

- Uses `SetContextStorage` function to update the trie being set with the runtime's state version
- Any new node written in the trie that has a storage value greater than 32 bytes and the trie is V1 will contain the `MustBeHashed` property as true
- While hashing the root node of the trie we check if this property is true and then we encode only the hash, nodes that were written while trie in V0 will be not affected

## Tests

N/A

## Issues

- Closes https://github.com/ChainSafe/gossamer/issues/3711

## Primary Reviewer

@dimartiro


dot/import.go
```diff
@@ -62,7 +62,7 @@ func newTrieFromPairs(filename string) (*trie.Trie, error) {
 		entries[pairArr[0].(string)] = pairArr[1].(string)
 	}
 
-	tr, err := trie.LoadFromMap(entries)
+	tr, err := trie.LoadFromMap(entries, trie.V0)
```
jimjbrettj, line 65: why do we use V0 here?

dot/state/block.go
```diff
@@ -880,7 +880,7 @@ func (bs *BlockState) HandleRuntimeChanges(newState *rtstorage.TrieState,
 		return fmt.Errorf("failed to update code substituted block hash: %w", err)
 	}
 
-	newVersion, err := parentRuntimeInstance.Version()
+	newVersion, err := instance.Version()
```
jimjbrettj, line 883: Why this diff?

lib/runtime/genesis.go
```diff
@@ -26,7 +26,7 @@ func NewTrieFromGenesis(gen genesis.Genesis) (tr trie.Trie, err error) {
 			ErrGenesisTopNotFound, gen.Name)
 	}
 
-	tr, err = trie.LoadFromMap(keyValues)
+	tr, err = trie.LoadFromMap(keyValues, trie.V0)
```
jimjbrettj, line 29: Why V0 here?

lib/runtime/wazero/imports.go
```diff
@@ -2303,27 +2304,19 @@ func ext_storage_root_version_1(ctx context.Context, m api.Module) uint64 {
 	return rootSpan
 }
 
-func ext_storage_root_version_2(ctx context.Context, m api.Module, version uint32) uint64 { //skipcq: RVV-B0012
+func ext_storage_root_version_2(ctx context.Context, m api.Module, _ uint32) uint64 { //skipcq: RVV-B0012
```
jimjbrettj, line 2307: Is it weird that we have an unused argument for an imported func?

tests/sync/sync_test.go
```diff
@@ -1,4 +1,5 @@
 // Copyright 2021 ChainSafe Systems (ON)
+
```
jimjbrettj, line 2: nit/remove diff

jimjbrettj (approved):
A few questions/comments, but overall really nice work!

dot/core/service_test.go
```diff
@@ -217,99 +213,99 @@ func Test_Service_handleCodeSubstitution(t *testing.T) {
...
-						{0x01}: common.BytesToHex(validRuntimeCode),
-					},
-					codeSubstitutedState: codeSubstitutedState,
-				}
-			},
-			blockHash:  common.Hash{0x01},
-			errWrapped: errTest,
-			errMessage: "storing code substituted block hash: test error",
-		},
+		// "non_existent_block_hash_substitute": {
```
dimartiro, line 216: Are you planning to uncomment these tests?

lib/runtime/genesis.go
dimartiro: I believe this is because we are loading it from genesis and assuming that genesis is using V0 but if I'm not wrong paseo uses V1 on genesis right?

Maybe we should get the version from the genesis configuration

lib/runtime/storage/trie.go
```diff
@@ -71,15 +71,27 @@ func (t *TrieState) CommitTransaction() {
...
 
+func (t *TrieState) SetVersion(v trie.TrieLayout) {
+	t.mtx.Lock()
+	defer t.mtx.Unlock()
+	t.getCurrentTrie().SetVersion(v)
+}
+
 // Trie returns the TrieState's underlying trie
 func (t *TrieState) Trie() *trie.Trie {
+	t.mtx.RLock()
```
dimartiro, line 82: Good catch 👏

pkg/trie/database.go
```diff
@@ -39,10 +38,36 @@ func (t *Trie) Load(db db.DBGetter, rootHash common.Hash) error {
...
 	t.root = root
 	t.root.MerkleValue = rootHashBytes
 
-	return t.loadNode(db, t.root)
+	err = t.loadNode(db, t.root)
+	if err != nil {
+		return err
+	}
+
+	for _, key := range t.GetKeysWithPrefix(ChildStorageKeyPrefix) {
```
dimartiro, line 54: Weren't we loading child tries? 😱

EclesioMeloJunior: I just moved from loadNode() and bring to Load() method, since we don't need to load a child trie every time we load a node in the trie

lib/runtime/genesis.go
EclesioMeloJunior: Yeah, the `LoadFromMap` is not 100% accurate to V1 since we need the infos of the encoded trie and not only the keys and values to decode a trie from maps, that's why it is by default `trie.V0`. But for some **test cases** we can use the arg `trie.V1`.

@dimartiro paseo uses V0

dimartiro: Ok we can say that is safe to use V0 until we start using a genesis with V1 set, maybe we can add this as a "known issue"? 🤔

EclesioMeloJunior: @dimartiro yes!

dimartiro (approved):
This is a really great work @EclesioMeloJunior congrats! 👏  and thanks you for the time you spent fixing it and testing it.
I really like the way you get rid of the `max inline value` parameter, that makes the code much more clear. 

Good work!

lib/runtime/wazero/instance.go
```diff
@@ -855,6 +869,17 @@ func (in *Instance) Validator() bool {
 func (in *Instance) SetContextStorage(s runtime.Storage) {
 	in.Lock()
 	defer in.Unlock()
+
+	if in.Context.Version == nil {
```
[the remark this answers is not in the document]
dimartiro: I will mark this conversation as resolved since it is not a blocker

lib/runtime/wazero/imports.go
EclesioMeloJunior: Yeah, this argument can be ignored for this host func since it uses the same `state_version` value from the wasm `runtime_version` structure, and in Gossamer we cache the `runtime_version`. But the plan is to use the parameter as soon as trie refactor takes place.

Issue #3742, opened 2024-02-07
enhancement: clean up pubip logs


jimjbrettj:
## Issue summary

We used the package "github.com/chyeh/pubip" to retrieve our public IP address and there have been no updates to this code in 8 years, resulting in errors from outdated HTTP requests causing ugly logs. This package should be replaced and the logs cleaned up

![image](https://github.com/ChainSafe/gossamer/assets/40345116/661399a2-7403-4f37-bd69-929b0c3088c0)

## Other information and links

- https://github.com/chyeh/pubip

PR #3743, opened 2024-02-07
Fix IP error logs

jimjbrettj:
## Changes

- Given the minimal amount of code in https://github.com/chyeh/pubip, I got rid of the dependency entirely, removed unneeded code, and updated the APIURIs to remove the nonfunction ones.
- This allows us to get our public IP without the messy logs as well as gets rid of an unmaintained dependency  

## Tests

Start node and make sure the ugly logs are gone

## Issues

#3742 3742

## Primary Reviewer

@axaysagathiya

### DIFF
--- a/dot/network/host.go
+++ b/dot/network/host.go
@@ -15,7 +15,7 @@ import (
 	"time"
 
 	"github.com/ChainSafe/gossamer/dot/peerset"
-	"github.com/chyeh/pubip"
+	"github.com/ChainSafe/gossamer/internal/pubip"
 	"github.com/dgraph-io/ristretto"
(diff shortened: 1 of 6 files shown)


kishansagathiya (commented):
Does `pubip` have to be part of gossamer codebase?

I feel like it could become a separate repo, `chainsafe/pubip`. If we insist on keeping it part of gossamer repo, `pkg/pubip` would be a better place, since this is more of an independent package.

Let's also add readme file.

kishansagathiya (approved):

dimartiro (approved):

Issue #3716, opened 2024-01-22
Non deterministic panic when running TestStableNetworkRPC 


P1sar:
## Describe the bug

- While fixing bug #3616 another unexpected non deterministic behaviour was found. This happens much rarely than one mentioned before but never the less this might be a symptom of something wrong.
- Message of the panic mentions issue #3066

## Expected Behavior

- Test TestStableNetworkRPC should deterministically work

## Current Behavior

- Test TestStableNetworkRPC should un deterministically fail with next error;
```=== RUN   TestStableNetworkRPC
    system_integration_test.go:53: Node alice-0 is ready
    system_integration_test.go:53: Node bob-1 is ready
    system_integration_test.go:53: Node charlie-2 is ready
    system_integration_test.go:65: waiting for all nodes to be connected
    system_integration_test.go:71: starting node alice-0 with port http://localhost:8540
    system_integration_test.go:74: Reponse: {Peers:1 IsSyncing:false ShouldHavePeers:true}
    node.go:233: node charlie-2 crashed: charlie-2 encountered a runtime error: exit status 2
        command: /var/www/gossamer/bin/gossamer --base-path /var/folders/1w/hmk5hlnd4tj3hhlkm0g7kln80000gn/T/TestStableNetworkRPC1793814496/003 --chain /var/www/gossamer/chain/westend-dev/westend-dev-spec-raw.json --role full --no-telemetry

        /var/folders/1w/hmk5hlnd4tj3hhlkm0g7kln80000gn/T/TestStableNetworkRPC1793814496/003


        Logs:
        2024-01-19T16:01:03+01:00 INFO     🕸️ initialising node with name Gossamer, id gssmr, base path /var/folders/1w/hmk5hlnd4tj3hhlkm0g7kln80000gn/T/TestStableNetworkRPC1793814496/003 and chain-spec /var/folders/1w/hmk5hlnd4tj3hhlkm0g7kln80000gn/T/TestStableNetworkRPC1793814496/003/chain-spec-raw.json...	pkg=dot
        2024-01-19T16:01:04+01:00 INFO     block state hash genesis hash: 0x276bfa91f70859348285599321ea96afd3ae681f0be47d36196bac8075ea32e8	pkg=state
[...]

Issue #2993, opened 2022-12-06
Use MDNS implementation `github.com/libp2p/go-libp2p/p2p/discovery/mdns`


qdm12:
Our current implementation (before PR #2794) is based on https://github.com/libp2p/go-libp2p/tree/644dadc30bcc712ea13ad730f50b5e8ec1b949f2/p2p/discovery/mdns_legacy

Unfortunately, `mdns_legacy` got removed in https://github.com/libp2p/go-libp2p/commit/707100a521dace4a7a87fd17c04042ea483380b5 so all releases since v0.21.0 do not have it anymore. The only alternative since is https://github.com/libp2p/go-libp2p/blob/4c74181d0dada0b4d600885285b96afee74e223d/p2p/discovery/mdns/mdns.go

Unfortunately again, this new mdns implementation will not work with our RPC test `TestStableNetworkRPC` with 3 nodes.

```sh
MODE=rpc go test -run ^TestStableNetworkRPC$ github.com/ChainSafe/gossamer/tests/rpc -v
```

This issue may be related to https://github.com/libp2p/go-libp2p/issues/1333 (or not).

After #2794 is merged, we should try upgrading to the new mdns package https://github.com/libp2p/go-libp2p/blob/4c74181d0dada0b4d600885285b96afee74e223d/p2p/discovery/mdns/mdns.go and remove the code in `internal/mdns` (based on mdns_legacy, in order to be able to upgrade libp2p dependencies).

timwu20:
We should probably revisit this at some point since we're using a fork of the old `libp2p/go-libp2p/p2p/discovery/mdns_legacy`.

PR #3756, opened 2024-02-13
fix: fix non deterministic  panic during TestStableNetworkRPC integration test

P1sar:
## Changes

Generally speaking fix per se is just within the test itself and basically just configuration changes [here](https://github.com/ChainSafe/gossamer/pull/3756/files#diff-1a6640e6c830aec2534d94380ff3420348b7b0c1bc48b8c3ca5267b5015f3b71R37) 

Although this small change makes test to pass constantly it does not eliminate core of the problem that is briefly mentioned [here](https://github.com/ChainSafe/gossamer/pull/3756/files#diff-3aa928e48cf85dbc8dbdc07ed7656b60f922de626322f142614783ca508f0e89R719).

But in order to eliminate this error completely and debug it deeply require much more effort so I decided to create separate issue  #3758 that describes this error. 

Other changes are small chores, TODOs and comments that were done in order to make code more readable.

closes: #3716, #2993

## Tests

```sh
```

## Primary Reviewer

@


dot/network/discovery.go
```diff
@@ -92,6 +93,9 @@ func (d *discovery) waitForPeers() (peers []peer.AddrInfo, err error) {
 
 // start creates the DHT.
 func (d *discovery) start() error {
+	// this basically only works with enabled mDNS which is used only for local test setups. Without bootnodes kademilia
+	// would not bee able to connect to any peers and mDNS is used to find peers in local network.
+	// TODO: this is an ugly solution and needs to be refactored
```
EclesioMeloJunior, line 98: I believe that this refactor can be in this PR, given that we don't need to wait for peers to start the discovery

dot/network/discovery.go
```diff
@@ -92,6 +92,9 @@ func (d *discovery) waitForPeers() (peers []peer.AddrInfo, err error) {
 
 // start creates the DHT.
 func (d *discovery) start() error {
+	// this basically only works with enabled mDNS which is used only for local test setups. Without bootnodes kademilia
+	// would not bee able to connect to any peers and mDNS is used to find peers in local network.
+	// TODO: should be refactored because this if is basically used for local integration test purpose
```
EclesioMeloJunior, line 97: Instead of waiting for peers to connect to start kad we can upgrade the kad routing table on every connection, I think that using `d.dht.{LAN/WAN}.RoutingTable().UsefulNewPeer(peerID)` should be a good option

EclesioMeloJunior (approved):
LGTM, just a question, do you plan to create a PR for each `TODO` added? If so, would be nice to have issues for each TODO?

P1sar:
> LGTM, just a question, do you plan to create a PR for each `TODO` added? If so, would be nice to have issues for each TODO?

I consider most of this TODOs to be rather references for future possible refactoring, rather than list of small issues to fix

jimjbrettj (approved):
lgtm!

PR #3765, opened 2024-02-16
fix(sync): tip sync can follow chain and finalize blocks

### LINKED ISSUE #3705: Gossamer is not changing to `tip` mode properly
## Describe the bug

- Once Gossamer has a sync target (the block number which is the current tip of the chain) the value remains the same forever
- Once Gossamer reaches the tip it stales waiting for block announcements

## Expected Behavior

- The value should increase as we receive new block announces
- Once we reach the tip we can pro-actively start handshakes to check the chain height

jimjbrettj:
## Changes

- The gossamer node was built to only handle neighbor messages when we are an authority. However if we do not handle a neighbor message when we are not an authority then other nodes on the network will not send us messages. This meant that gossamer was not receiving commit messages and thus not tracking the finalized chain.
- Clean up some logs

## Tests

- Sync a paseo node and verify that we can transition to tip mode and continue processing the chain
- If you would like, spin up local devnet to test

```sh
go test -tags integration ./dot/sync
```

## Issues

Closes #3705

## Primary Reviewer

@EclesioMeloJunior

### DIFF
--- a/dot/state/block.go
+++ b/dot/state/block.go
@@ -51,17 +51,17 @@ var (
 // BlockState contains the historical block data of the blockchain, including block headers and bodies.
 // It wraps the blocktree (which contains unfinalised blocks) and the database (which contains finalised blocks).
 type BlockState struct {
-	bt                 *blocktree.BlockTree
-	baseState          *BaseState
-	dbPath             string
-	db                 BlockStateDatabase
-	lock               sync.RWMutex
-	genesisHash        common.Hash
-	lastFinalised      common.Hash
-	lastFinalisedRound uint64
-	lastFinalisedSetID uint64
-	unfinalisedBlocks  *hashToBlockMap
-	tries              *Tries
+	bt                *blocktree.BlockTree
+	baseState         *BaseState
+	dbPath            string
+	db                BlockStateDatabase
+	lock              sync.RWMutex
+	genesisHash       common.Hash
+	lastFinalised     common.Hash
+	lastRound         uint64
+	lastSetID         uint64
+	unfinalisedBlocks *hashToBlockMap
+	tries             *Tries
 
 	// block notifiers
 	imported                       map[chan *types.Block]struct{}
--- a/dot/state/block.go
+++ b/dot/state/block.go
@@ -51,15 +51,17 @@ var (
 // BlockState contains the historical block data of the blockchain, including block headers and bodies.
 // It wraps the blocktree (which contains unfinalised blocks) and the database (which contains finalised blocks).
 type BlockState struct {
-	bt                *blocktree.BlockTree
-	baseState         *BaseState
-	dbPath            string
-	db                BlockStateDatabase
-	lock              sync.RWMutex
-	genesisHash       common.Hash
-	lastFinalised     common.Hash
-	unfinalisedBlocks *hashToBlockMap
-	tries             *Tries
+	bt                 *blocktree.BlockTree
+	baseState          *BaseState
+	dbPath             string
+	db                 BlockStateDatabase
+	lock               sync.RWMutex
+	genesisHash        common.Hash
+	lastFinalised      common.Hash
+	lastFinalisedRound uint64
+	lastFinalisedSetID uint64
+	unfinalisedBlocks  *hashToBlockMap
+	tries              *Tries
 
 	// block notifiers
 	imported                       map[chan *types.Block]struct{}
--- a/dot/sync/worker_pool.go
+++ b/dot/sync/worker_pool.go
@@ -4,9 +4,8 @@
 package sync
 
 import (
-	"fmt"
-
 	"crypto/rand"
+	"fmt"
 	"math/big"
 	"sync"
 	"time"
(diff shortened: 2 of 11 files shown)


lib/grandpa/message_handler.go
```diff
@@ -80,27 +80,25 @@ func (h *MessageHandler) handleMessage(from peer.ID, m GrandpaMessage) (network.
...
-
-		logger.Debugf("sending neighbour message: %v", neighbourMessage)
-		h.grandpa.network.GossipMessage(cm)
+	// TODO(#2931): this is a simple hack to ensure that the neighbour messages
+	// sent by gossamer are being received by substrate nodes
+	// not intended to be production code
+	h.grandpa.roundLock.Lock()
+	neighbourMessage := &NeighbourPacketV1{
+		Round:  h.grandpa.state.round,
+		SetID:  h.grandpa.state.setID,
```
EclesioMeloJunior, line 89: Could you ensure that these grandpa state properties are being updated on every finalization?

EclesioMeloJunior (changes requested):

lib/grandpa/message_handler.go
jimjbrettj: good catch here, it wasnt being updated. Have updated the code so get this data from the blockstate now

dot/state/block.go
```diff
@@ -51,15 +51,17 @@ var (
...
-	tries             *Tries
+	bt                 *blocktree.BlockTree
+	baseState          *BaseState
+	dbPath             string
+	db                 BlockStateDatabase
+	lock               sync.RWMutex
+	genesisHash        common.Hash
+	lastFinalised      common.Hash
+	lastFinalisedRound uint64
+	lastFinalisedSetID uint64
```
EclesioMeloJunior, line 62: I would suggest using just `lastRound/lastSetID` since what was finalised was the block

```suggestion
	lastRound uint64
	lastSetID uint64
```

lib/grandpa/state.go
```diff
@@ -23,6 +23,8 @@ type BlockState interface {
 	LowestCommonAncestor(a, b common.Hash) (common.Hash, error)
 	HasFinalisedBlock(round, setID uint64) (bool, error)
 	GetFinalisedHeader(round, setID uint64) (*types.Header, error)
+	GetFinalisedRound() uint64
+	GetFinalisedSetID() uint64
```
EclesioMeloJunior, line 27:
```suggestion
	GetLastRound() uint64
	GetLastSetID() uint64
```

EclesioMeloJunior (approved):
Just a tiny suggestion regards the names

dot/sync/worker_pool.go
```diff
@@ -162,8 +162,10 @@ func (s *syncWorkerPool) submitRequest(request *network.BlockRequestMessage,
 
 	if who != nil {
 		syncWorker := s.workers[*who]
-		syncWorker.queue <- task
-		return
+		if syncWorker != nil {
```
timwu20: would there ever be a nil `*syncWorker` in this map?

lib/grandpa/message_handler.go
```diff
@@ -80,28 +80,26 @@ func (h *MessageHandler) handleMessage(from peer.ID, m GrandpaMessage) (network.
...
-			return fmt.Errorf("converting neighbour message to network message: %w", err)
-		}
-
-		logger.Debugf("sending neighbour message: %v", neighbourMessage)
-		h.grandpa.network.GossipMessage(cm)
+	// TODO(#2931): this is a simple hack to ensure that the neighbour messages
+	// sent by gossamer are being received by substrate nodes
+	// not intended to be production code
+	round := h.blockState.GetRound()
+	setID := h.blockState.GetSetID()
```
timwu20: should we check that these aren't zero values?

timwu20: we don't need to lock the mutex anymore?

dot/sync/worker_pool.go
jimjbrettj: Yeah it appears so. I was panicking because of this and thus added the check

lib/grandpa/message_handler.go
jimjbrettj: hmm I am not sure that we need to add checks. We were actually only using zero values before, and from my tests I know that the values do progress so I don't think we need to error out or anything.

As to locks, the locking still happens I moved it to internal locking instead of having the caller handle it, which I thought was cleaner.

dot/sync/worker_pool.go
timwu20: I would assume it's because we don't check that the map contains the key of `*who`.  It will default to zero value of `*syncWorker` if the key doesn't exist which would be `nil` in this case.  I think it makes more sense to do a key check instead of checking if `nil`.  A `nil` worker would probably lead me to believe it's a a programmer error, and I would panic.

lib/grandpa/message_handler.go
timwu20: I think you should return both round and setID in one call then, and do internal locking around that.  You have the risk of the round or set ID changing between these two functions calls.

jimjbrettj: good call, updated

dot/sync/worker_pool.go
```diff
@@ -161,9 +160,14 @@ func (s *syncWorkerPool) submitRequest(request *network.BlockRequestMessage,
...
-		syncWorker := s.workers[*who]
-		syncWorker.queue <- task
-		return
+		syncWorker, inMap := s.workers[*who]
+		if inMap {
+			if syncWorker != nil {
+				syncWorker.queue <- task
+				return
+			}
+			panic("sync worker should not be nil")
```
timwu20:
```suggestion
			if syncWorker == nil {
			        panic("sync worker should not be nil")
			}
			syncWorker.queue <- task
		        return
			
```

timwu20 (approved):
small change

PR #3768, opened 2024-02-22
refactor(pkg/trie): Create Trie interface to decouple our current implementation

### LINKED ISSUE #3760: (pkg/trie): Create Trie interface and rename our current implementation
Based on my [latest proposal](https://github.com/ChainSafe/gossamer/discussions/3646#discussioncomment-8465826) I created this issue to address the second part of Phase 0.

>  - Move all trie-related code to a single package in /pkg/trie.
> - Create a Trie interface with the minimum necessary methods for a trie. Such as: get, insert, remove, contains, root
Rename the current Trie struct to InMemoryTrie.

The goal is to have a Trie interface with the minimum required functions and use them as a way to add polymorphism to our trie. This will helpful later to replace our InMemory trie to a backed db trie

dimartiro:
## Changes

This PR aims to decouple our current trie impl and usage.

- [x] `Trie` interface creation
- [x] Rename previous `Trie` impl to `InMemoryTrie`
- [x] Use new `Trie` interface when necessary 
- [x] Rename `dot/state/tries.go` to `dot/state/inmemory_tries.go`
- [x] Rename `lib/runtime/storage/trie.go` to `lib/runtime/storage/inmemory/triestate.go`
- [x] Rename `dot/state/storage.go` to `dot/state/inmemory_storage.go`
- [x] Update tests

## Tests

```sh
make test
```
```
Also tested running a node locally until reach block 1M (with stop, restart too)
```

## Issues

closes #3760

## Note to reviewers

I know this PR is large but it will give us more flexibility in the future when we need to change from the in-memory trie implementation to the one with lazy loading.
Most of the changed files are part of the refactor to start using the interface and it didn't change the business logic.

## Primary Reviewer

@jimjbrettj


pkg/trie/child_storage.go
```diff
@@ -37,8 +37,21 @@ func (t *Trie) SetChild(keyToChild []byte, child *Trie) error {
...
+	copy(key[len(ChildStorageKeyPrefix):], keyToChild)
+
+	childHash := t.Get(key)
+	if childHash == nil {
+		return nil, fmt.Errorf("%w at key 0x%x%x", ErrChildTrieDoesNotExist, ChildStorageKeyPrefix, keyToChild)
+	}
+
+	return t.childTries[common.BytesToHash(childHash)], nil
+}
+
```
EclesioMeloJunior: seems that this method is a copy of the `GetChild` method.

pkg/trie/trie.go
```diff
@@ -1,1518 +1,81 @@
...
-
-	nextKey := findNextKey(t.root, prefix, key)
-	if nextKey == nil {
-		return nil
-	}
-
-	nextKeyLE = codec.NibblesToKeyLE(nextKey)
-	return nextKeyLE
+type Printable interface {
+	String() string
```
EclesioMeloJunior: I would suggest to use `fmt.Stringer `

pkg/trie/trie.go
```diff
@@ -1,1518 +1,81 @@
...
-	}
-
-	return trieCopy
+type ChildTrieManager interface {
+	GetChild(keyToChild []byte) (Trie, error)
+	GetFromChild(keyToChild, key []byte) ([]byte, error)
+	PutIntoChild(keyToChild, key, value []byte) error
+	DeleteChild(keyToChild []byte) (err error)
+	ClearFromChild(keyToChild, key []byte) error
+	GetChildTries() map[common.Hash]Trie
```
EclesioMeloJunior: I would suggest splitting those methods into small cohesive groups like: `ChildTrieGetter`, `ChildTrieSetter` and `ChildTrieDeleter`

pkg/trie/trie.go
```diff
@@ -1,1518 +1,81 @@
...
-	if parentBranch.StorageValue != nil {
-		return fullKey
-	}
-	const startChildIndex = 0
-	return findNextKeyChild(parentBranch.Children, startChildIndex,
-		fullKey, searchKey)
+type Hashable interface {
+	MustHash() common.Hash
+	Hash() (common.Hash, error)
+	GenesisBlock() (genesisHeader types.Header, err error)
```
EclesioMeloJunior, line 53: `GenesisBlock` does not seem to belong to this interface, why is the trie responsible for retrieving the genesis block? I think this responsibility is for the state manager who is aware of the chain

EclesioMeloJunior (changes requested):

pkg/trie/child_storage.go
dimartiro: Thanks! was an issue with the merge with dev, solved it

pkg/trie/trie.go
dimartiro: Great! I didn't know it, I'm going to change it

pkg/trie/trie.go
dimartiro: Agree, but we are using it now in [node.go](https://github.com/ChainSafe/gossamer/blob/development/dot/node.go#L168) I think I could leave it for a different refactor PR

pkg/trie/trie.go
dimartiro: Done!

pkg/trie/genesis_test.go
```diff
@@ -9,10 +9,12 @@ import (
 	"github.com/ChainSafe/gossamer/dot/types"
 	"github.com/ChainSafe/gossamer/lib/common"
 	"github.com/ChainSafe/gossamer/pkg/trie/node"
+
+	//"github.com/ChainSafe/gossamer/pkg/trie/node"
```
EclesioMeloJunior:
```suggestion
```

EclesioMeloJunior (approved):
LGTM

dot/state/initialize.go
```diff
@@ -40,8 +40,11 @@ func (s *Service) Initialise(gen *genesis.Genesis, header *types.Header, t *trie
 
 	s.db = db
 
-	if err = t.WriteDirty(database.NewTable(db, storagePrefix)); err != nil {
-		return fmt.Errorf("failed to write genesis trie to database: %w", err)
+	// TODO: all trie related db operations should be done in pkg/trie
```
jimjbrettj, line 43: Should we have an issue to represent this?

dot/state/initialize.go
```diff
@@ -134,10 +137,13 @@ func loadGrandpaAuthorities(t *trie.Trie) ([]types.GrandpaVoter, error) {
 }
 
 // storeInitialValues writes initial genesis values to the state database
-func (s *Service) storeInitialValues(data *genesis.Data, t *trie.Trie) error {
+func (s *Service) storeInitialValues(data *genesis.Data, t trie.Trie) error {
 	// write genesis trie to database
-	if err := t.WriteDirty(database.NewTable(s.db, storagePrefix)); err != nil {
-		return fmt.Errorf("failed to write trie to database: %s", err)
+	// TODO: all trie related db operations should be done in pkg/trie
```
jimjbrettj, line 142: same here

dot/state/inmemory_storage.go
```diff
@@ -76,9 +76,12 @@ func (s *StorageState) StoreTrie(ts *rtstorage.TrieState, header *types.Header)
 
 	logger.Tracef("cached trie in storage state: %s", root)
 
-	if err := ts.Trie().WriteDirty(s.db); err != nil {
-		logger.Warnf("failed to write trie with root %s to database: %s", root, err)
-		return err
+	// TODO: all trie related db operations should be done in pkg/trie
```
jimjbrettj, line 79: same here and for other instances

dot/state/initialize.go
dimartiro: That's a good question, I don't think so since we are going to replace it with lazy loading and that implementation encapsulates all the DB related operations in the same trie. It will execute the writes before calculating the root hash, so I don't think we need an issue for this, it's more an explanation for the type assertion 😅

dot/state/initialize.go
dimartiro: Yeah, same answer than [here](https://github.com/ChainSafe/gossamer/pull/3768#discussion_r1541503348)

dot/state/inmemory_storage.go
dimartiro: same answer than [here](https://github.com/ChainSafe/gossamer/pull/3768#discussion_r1541503348) it's more a comment to explain the usage of type assertion

pkg/trie/inmemory.go
```diff
@@ -0,0 +1,1518 @@
+// Copyright 2024 ChainSafe Systems (ON)
+// SPDX-License-Identifier: LGPL-3.0-only
+
+package trie
+
+import (
```
jimjbrettj: nit/should this file be in_memory.go?

dimartiro: Done in my last commit! thks!

jimjbrettj (approved):
great work, lgtm

PR #3777, opened 2024-02-27
refactor(lib/runtime/storage): Don't rely on trie snapshots for storage transactions

### LINKED ISSUE #3775: (lib/runtime/storage): Modify transactions support to not rely on trie snapshots
# Description

Continuing with the task of decoupling our current trie usage from our implementation, we need to modify our current support for [state storage transactions](https://github.com/ChainSafe/gossamer/blob/development/lib/runtime/storage/trie.go) so it does not rely on trie snapshots. Trie snapshots are a feature available in the `in-memory trie`, but they will not be available when we start using lazy loading. Therefore, we need to change our [TrieState](https://github.com/ChainSafe/gossamer/blob/development/lib/runtime/storage/trie.go) implementation to work without using snapshots. This will allow us to change the implementation in the near future.

We could use an [approach similar to smoldot](https://github.com/smol-dot/smoldot/blob/main/lib/src/executor/storage_diff.rs) where they are storing the diff between a trie and the next one and we can apply it when the last transaction commit is executed.

This will be a benefit not only to decouple this from our current implementation but also to save resources since the snapshot is an expensive operation

dimartiro:
## Changes

This PR aims to change the way we manage storage transactions to rely on a new `changeSet` structure instead of using trie snapshots.

Trie snapshots are a feature that makes sense for in-memory tries but makes our code too coupled and prevents us from starting to use lazy loading tries.

What I'm doing is storing the diff (trie changes) between different transactions and applying them to the trie when the last transaction is committed.

This is not only useful for decoupling this from our trie implementation but will also reduce the amount of memory we need to perform storage transactions.

## Tests

```sh
make test
```

## Issues

closes #3775 

## Primary Reviewer

@EclesioMeloJunior

### DIFF
--- a/dot/state/storage_test.go
+++ b/dot/state/storage_test.go
@@ -10,7 +10,6 @@ import (
 	"github.com/ChainSafe/gossamer/dot/telemetry"
 	"github.com/ChainSafe/gossamer/dot/types"
 	"github.com/ChainSafe/gossamer/internal/database"
-	"github.com/ChainSafe/gossamer/internal/trie/node"
 	"github.com/ChainSafe/gossamer/lib/common"
 	runtime "github.com/ChainSafe/gossamer/lib/runtime/storage"
 	"github.com/ChainSafe/gossamer/lib/trie"
@@ -31,23 +30,27 @@ func newTestStorageState(t *testing.T) *StorageState {
 }
 
 func TestStorage_StoreAndLoadTrie(t *testing.T) {
-	storage := newTestStorageState(t)
-	ts, err := storage.TrieState(&trie.EmptyHash)
-	require.NoError(t, err)
+	t.Skip()
 
-	root, err := ts.Root(trie.NoMaxInlineValueSize)
-	require.NoError(t, err)
-	err = storage.StoreTrie(ts, nil)
-	require.NoError(t, err)
+	/*
+		storage := newTestStorageState(t)
+		ts, err := storage.TrieState(&trie.EmptyHash)
+		require.NoError(t, err)
 
-	time.Sleep(time.Millisecond * 100)
+		root, err := ts.Root(trie.NoMaxInlineValueSize)
+		require.NoError(t, err)
+		err = storage.StoreTrie(ts, nil)
+		require.NoError(t, err)
 
-	trie, err := storage.LoadFromDB(root)
-	require.NoError(t, err)
-	ts2 := runtime.NewTrieState(trie)
-	newSnapshot := ts2.Snapshot()
+		time.Sleep(time.Millisecond * 100)
 
-	require.True(t, ts.Trie().Equal(newSnapshot))
+		trie, err := storage.LoadFromDB(root)
+		require.NoError(t, err)
+		ts2 := runtime.NewTrieState(trie)
+		newSnapshot := ts2.Snapshot()
+
+		require.True(t, ts.Trie().Equal(newSnapshot))
+	*/
 }
 
 func TestStorage_GetStorageByBlockHash(t *testing.T) {
@@ -59,7 +62,7 @@ func TestStorage_GetStorageByBlockHash(t *testing.T) {
 	value := []byte("testvalue")
 	ts.Put(key, value)
 
-	root, err := ts.Root(trie.NoMaxInlineValueSize)
+	root, err := ts.Root(trie.V0)
 	require.NoError(t, err)
 	err = storage.StoreTrie(ts, nil)
 	require.NoError(t, err)
@@ -91,7 +94,7 @@ func TestStorage_TrieState(t *testing.T) {
 	require.NoError(t, err)
 	ts.Put([]byte("noot"), []byte("washere"))
 
-	root, err := ts.Root(trie.NoMaxInlineValueSize)
+	root, err := ts.Root(trie.V0)
 	require.NoError(t, err)
 	err = storage.StoreTrie(ts, nil)
 	require.NoError(t, err)
@@ -125,7 +128,7 @@ func TestStorage_LoadFromDB(t *testing.T) {
 		ts.Put(kv.key, kv.value)
 	}
 
-	root, err := ts.Root(trie.NoMaxInlineValueSize)
+	root, err := ts.Root(trie.V0)
 	require.NoError(t, err)
 
 	// Write trie to disk.
@@ -183,16 +186,9 @@ func TestGetStorageChildAndGetStorageFromChild(t *testing.T) {
 	trieDB := NewMockDatabase(ctrl)
 	trieDB.EXPECT().Get(gomock.Any()).Times(0)
 
-	trieRoot := &node.Node{
-		PartialKey:   []byte{1, 2},
-		StorageValue: []byte{3, 4},
-		Dirty:        true,
-	}
-	testChildTrie := trie.NewTrie(trieRoot, trieDB)
+	genTrie.PutIntoChild([]byte("keyToChild"), []byte{1, 2}, []byte{3, 4})
+	genTrie.PutIntoChild([]byte("keyToChild"), []byte("keyInsidechild"), []byte("voila"))
 
-	testChildTrie.Put([]byte("keyInsidechild"), []byte("voila"))
-
-	err = genTrie.SetChild([]byte("keyToChild"), testChildTrie)
 	require.NoError(t, err)
 
 	tries := newTriesEmpty()
@@ -205,7 +201,7 @@ func TestGetStorageChildAndGetStorageFromChild(t *testing.T) {
 
 	trieState := runtime.NewTrieState(&genTrie)
 
-	header := types.NewHeader(blockState.GenesisHash(), trieState.MustRoot(trie.NoMaxInlineValueSize),
+	header := types.NewHeader(blockState.GenesisHash(), trieState.MustRoot(trie.V0),
 		common.Hash{}, 1, types.NewDigest())
 
 	err = storage.StoreTrie(trieState, header)
--- a/lib/runtime/interfaces.go
+++ b/lib/runtime/interfaces.go
@@ -11,28 +11,34 @@ import (
 
 // Storage runtime interface.
 type Storage interface {
+	// Main trie
+	Root(trie.TrieLayout) (common.Hash, error)
 	Put(key []byte, value []byte) (err error)
 	Get(key []byte) []byte
-	Root(maxInlineValueSize int) (common.Hash, error)
-	SetChild(keyToChild []byte, child *trie.Trie) error
+	Delete(key []byte) (err error)
+	NextKey([]byte) []byte
+	ClearPrefix(prefix []byte) (err error)
+	ClearPrefixLimit(prefix []byte, limit uint32) (
+		deleted uint32, allDeleted bool, err error)
+
+	// Child tries
+	GetChildRoot(keyToChild []byte, version trie.TrieLayout) (common.Hash, error)
 	SetChildStorage(keyToChild, key, value []byte) error
 	GetChildStorage(keyToChild, key []byte) ([]byte, error)
-	Delete(key []byte) (err error)
 	DeleteChild(keyToChild []byte) (err error)
 	DeleteChildLimit(keyToChild []byte, limit *[]byte) (
 		deleted uint32, allDeleted bool, err error)
 	ClearChildStorage(keyToChild, key []byte) error
-	NextKey([]byte) []byte
 	ClearPrefixInChild(keyToChild, prefix []byte) error
 	ClearPrefixInChildWithLimit(keyToChild, prefix []byte, limit uint32) (uint32, bool, error)
 	GetChildNextKey(keyToChild, key []byte) ([]byte, error)
-	GetChild(keyToChild []byte) (*trie.Trie, error)
-	ClearPrefix(prefix []byte) (err error)
-	ClearPrefixLimit(prefix []byte, limit uint32) (
-		deleted uint32, allDeleted bool, err error)
+
+	// Transactions
 	StartTransaction()
 	CommitTransaction()
 	RollbackTransaction()
+
+	// Runtime
 	LoadCode() []byte
 }
 
--- a/lib/runtime/storage/trie.go
+++ b/lib/runtime/storage/trie.go
@@ -19,23 +19,24 @@ import (
 // If the execution of the call is successful, the trie will be saved in the StorageState.
 type TrieState struct {
 	mtx          sync.RWMutex
+	state        *trie.Trie
 	transactions *list.List
 }
 
-func NewTrieState(state *trie.Trie) *TrieState {
+func NewTrieState(initialState *trie.Trie) *TrieState {
 	transactions := list.New()
-	transactions.PushBack(state)
 	return &TrieState{
 		transactions: transactions,
+		state:        initialState,
 	}
 }
 
-func (t *TrieState) getCurrentTrie() *trie.Trie {
-	return t.transactions.Back().Value.(*trie.Trie)
-}
-
-func (t *TrieState) updateCurrentTrie(new *trie.Trie) {
-	t.transactions.Back().Value = new
+func (t *TrieState) getCurrentTransaction() *changeSet {
+	innerTransaction := t.transactions.Back()
+	if innerTransaction == nil {
+		return nil
+	}
+	return innerTransaction.Value.(*changeSet)
 }
 
 // StartTransaction begins a new nested storage transaction
@@ -44,7 +45,12 @@ func (t *TrieState) StartTransaction() {
 	t.mtx.Lock()
 	defer t.mtx.Unlock()
 
-	t.transactions.PushBack(t.getCurrentTrie().Snapshot())
+	nextChangeSet := t.getCurrentTransaction()
+	if nextChangeSet == nil {
+		nextChangeSet = newChangeSet()
+	}
+
+	t.transactions.PushBack(nextChangeSet.snapshot())
 }
 
 // Rollback rolls back all storage changes made since StartTransaction was called.
@@ -52,7 +58,7 @@ func (t *TrieState) RollbackTransaction() {
 	t.mtx.Lock()
 	defer t.mtx.Unlock()
 
-	if t.transactions.Len() <= 1 {
+	if t.transactions.Len() < 1 {
 		panic("no transactions to rollback")
 	}
 
@@ -64,23 +70,22 @@ func (t *TrieState) CommitTransaction() {
 	t.mtx.Lock()
 	defer t.mtx.Unlock()
 
-	if t.transactions.Len() <= 1 {
+	if t.transactions.Len() == 0 {
 		panic("no transactions to commit")
 	}
 
-	t.transactions.Back().Prev().Value = t.transactions.Remove(t.transactions.Back())
+	if t.transactions.Len() > 1 {
+		// We merge this transaction with its parent transaction
+		t.transactions.Back().Prev().Value = t.transactions.Remove(t.transactions.Back())
+	} else {
+		// This is the last transaction so we apply all the changes to our state
+		t.transactions.Remove(t.transactions.Back()).(*changeSet).applyToTrie(t.state)
+	}
 }
 
 // Trie returns the TrieState's underlying trie
 func (t *TrieState) Trie() *trie.Trie {
-	return t.getCurrentTrie()
-}
-
-// Snapshot creates a new "version" of the trie. The trie before Snapshot is called
-// can no longer be modified, all further changes are on a new "version" of the trie.
-// It returns the new version of the trie.
-func (t *TrieState) Snapshot() *trie.Trie {
-	return t.getCurrentTrie().Snapshot()
+	return t.state
 }
 
 // Put puts a key-value pair in the trie
@@ -88,24 +93,46 @@ func (t *TrieState) Put(key, value []byte) (err error) {
 	t.mtx.Lock()
 	defer t.mtx.Unlock()
 
-	return t.getCurrentTrie().Put(key, value)
+	// If we have running transactions we apply the change there,
+	// if not, we apply the changes directly on our state trie
+	if t.getCurrentTransaction() != nil {
+		t.getCurrentTransaction().upsert(string(key), value)
+		return nil
+	} else {
+		return t.state.Put(key, value)
+	}
 }
 
 // Get gets a value from the trie
 func (t *TrieState) Get(key []byte) []byte {
 	t.mtx.RLock()
 	defer t.mtx.RUnlock()
-	return t.getCurrentTrie().Get(key)
+
+	// If we find the key or it is deleted return from latest transaction
+	val, deleted := t.getCurrentTransaction().get(string(key))
+	if val != nil || deleted {
+		return val
+	}
+
+	// If we didn't find the key in the latest transactions lookup from state
+	return t.state.Get(key)
 }
 
 // MustRoot returns the trie's root hash. It panics if it fails to compute the root.
-func (t *TrieState) MustRoot(maxInlineValue int) common.Hash {
-	return t.getCurrentTrie().MustHash(maxInlineValue)
+func (t *TrieState) MustRoot(version trie.TrieLayout) common.Hash {
+	hash, err := t.Root(version)
+	if err != nil {
+		panic(err)
+	}
+
+	return hash
 }
 
 // Root returns the trie's root hash
-func (t *TrieState) Root(maxInlineValue int) (common.Hash, error) {
-	return t.getCurrentTrie().Hash(maxInlineValue)
+func (t *TrieState) Root(version trie.TrieLayout) (common.Hash, error) {
+	entries := trie.NewEntriesFromMap(t.TrieEntries())
+
+	return version.Root(entries)
 }
 
 // Has returns whether or not a key exists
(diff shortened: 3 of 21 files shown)


dot/state/storage_test.go
```diff
@@ -31,23 +30,28 @@ func newTestStorageState(t *testing.T) *StorageState {
 }
 
 func TestStorage_StoreAndLoadTrie(t *testing.T) {
-	storage := newTestStorageState(t)
-	ts, err := storage.TrieState(&trie.EmptyHash)
-	require.NoError(t, err)
+	t.Skip()
+	// FIX ME
```
jimjbrettj: Will this be addressed in this PR? If not should probably link an issue representing the needed

lib/runtime/interfaces.go
```diff
@@ -11,28 +11,34 @@ import (
 
 // Storage runtime interface.
 type Storage interface {
+	// Main trie
```
jimjbrettj: Nice, I am a fan of labeling and sorting methods in large interfaces like this

lib/runtime/storage/storagediff.go
```diff
@@ -0,0 +1,230 @@
...
+	"maps"
+	"sync"
+
+	xmaps "golang.org/x/exp/maps"
+
+	"github.com/ChainSafe/gossamer/pkg/trie"
+	"github.com/ChainSafe/gossamer/pkg/trie/codec"
+)
+
+type storageDiff struct {
```
jimjbrettj, line 24: I think this would be a case where adding some nice docs in the form of comments would be useful. Maybe just have it be similar to the PR description, describing how this storage diff, batch processing type functionality works.

lib/runtime/storage/storagediff.go
```diff
@@ -0,0 +1,230 @@
...
+	upserts        map[string][]byte
```
jimjbrettj, line 25: Is upsert a typical term used in this context? I had never heard of it and was confused, but if its a typical term I am okay with it

lib/runtime/storage/storagediff.go
```diff
@@ -0,0 +1,230 @@
...
+	deletes        map[string]bool
+	childChangeSet map[string]*storageDiff
+	l              sync.RWMutex
```
jimjbrettj: nit/either embed or rename to something like "mtx" or "lock"

lib/runtime/storage/storagediff.go
```diff
@@ -0,0 +1,230 @@
...
+	cs.l.Lock()
+	defer cs.l.Unlock()
+
+	delete(cs.childChangeSet, key)
+	delete(cs.upserts, key)
+	cs.deletes[key] = true
+}
+
+func (cs *storageDiff) clearPrefix(prefix []byte, trieKeys []string, limit int) (deleted uint32, allDeleted bool) {
+	// TODO: consider actual keys in trie
```
jimjbrettj: Is this still a TODO?

lib/runtime/storage/storagediff.go
```diff
@@ -0,0 +1,230 @@
...
+	prefix = codec.KeyLEToNibbles(prefix)
+	prefix = bytes.TrimSuffix(prefix, []byte{0})
+	newKeys := xmaps.Keys(cs.upserts)
+	allKeys := append(newKeys, trieKeys...)
+	deleted = 0
+	for _, k := range allKeys {
+		if limit > -1 && deleted >= uint32(limit) {
```
jimjbrettj: I think you do this check outside of the for loop since we know deleted's max value

lib/runtime/storage/storagediff_test.go
```diff
@@ -0,0 +1,36 @@
...
+// SPDX-License-Identifier: LGPL-3.0-only
+
+package storage
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
```
jimjbrettj, line 12: Do you think it could be beneficial to have further tests for the child tree operations as well as the `applyToTrie` function?

lib/runtime/storage/trie.go
```diff
@@ -224,22 +321,23 @@ func (t *TrieState) DeleteChildLimit(key []byte, limit *[]byte) (
 	t.mtx.Lock()
 	defer t.mtx.Unlock()
 
-	trieSnapshot := t.getCurrentTrie().Snapshot()
+	if currentTx := t.getCurrentTransaction(); currentTx != nil {
+		panic("fix me")
+	}
 
-	tr, err := trieSnapshot.GetChild(key)
+	tr, err := t.state.GetChild(key)
```
jimjbrettj:
```suggestion
	child, err := t.state.GetChild(key)
```

lib/runtime/wazero/imports_test.go
```diff
@@ -931,10 +931,10 @@ func Test_ext_default_child_storage_read_version_1(t *testing.T) {
 			setupInstance: func(t *testing.T) *Instance {
 				inst := NewTestInstance(t, runtime.HOST_API_TEST_RUNTIME, TestWithVersion(DefaultVersion))
 
-				err := inst.Context.Storage.SetChild(testChildKey, trie.NewEmptyTrie())
-				require.NoError(t, err)
+				//err := inst.Context.Storage.SetChild(testChildKey, trie.NewEmptyTrie())
```
jimjbrettj: nit/delete or add back?

lib/runtime/wazero/imports_test.go
```diff
@@ -1071,10 +1071,10 @@ func Test_ext_default_child_storage_set_version_1(t *testing.T) {
 func Test_ext_default_child_storage_clear_version_1(t *testing.T) {
 	inst := NewTestInstance(t, runtime.HOST_API_TEST_RUNTIME, TestWithVersion(DefaultVersion))
 
-	err := inst.Context.Storage.SetChild(testChildKey, trie.NewEmptyTrie())
-	require.NoError(t, err)
+	//err := inst.Context.Storage.SetChild(testChildKey, trie.NewEmptyTrie())
```
jimjbrettj: Same here

lib/runtime/wazero/imports_test.go
```diff
@@ -1228,11 +1220,11 @@ func Test_ext_default_child_storage_next_key_version_1(t *testing.T) {
 			setupInstance: func(t *testing.T) *Instance {
 				inst := NewTestInstance(t, runtime.HOST_API_TEST_RUNTIME, TestWithVersion(DefaultVersion))
 
-				err := inst.Context.Storage.SetChild(testChildKey, trie.NewEmptyTrie())
-				require.NoError(t, err)
+				//err := inst.Context.Storage.SetChild(testChildKey, trie.NewEmptyTrie())
```
jimjbrettj: Same here

lib/runtime/wazero/imports_test.go
```diff
@@ -1250,11 +1242,11 @@ func Test_ext_default_child_storage_next_key_version_1(t *testing.T) {
...
```
jimjbrettj: Same here

dot/state/storage_test.go
dimartiro: Pushed in my last commit 😄

lib/runtime/wazero/imports_test.go
```diff
@@ -1411,13 +1394,18 @@ func Test_ext_default_child_storage_storage_kill_version_2_limit_all(t *testing.
...
+	require.Nil(t, value)
+
+	value, err = inst.Context.Storage.GetChildStorage(testChildKey, []byte(`key2`))
+	require.NoError(t, err)
+	require.Nil(t, value)
 }
 
 func Test_ext_default_child_storage_storage_kill_version_2_limit_1(t *testing.T) {
-	inst := NewTestInstance(t, runtime.HOST_API_TEST_RUNTIME, TestWithVersion(DefaultVersion))
+	t.Skip("FIX ME")
```
jimjbrettj: Is this still a todo?

jimjbrettj: Applys to below functions as well

jimjbrettj (commented):
A few comments, but a really nice change great work!

lib/runtime/interfaces.go
dimartiro: Yes, this is one approach. Another approach involves creating a sort of 'inheritance' between interfaces. We can create four interfaces, each for a different group, and then compose the `Storage` interface from those other interfaces.

lib/runtime/storage/storagediff.go
dimartiro: Sure! will do!

lib/runtime/storage/storagediff.go
dimartiro: I got it from databases context, when you want to mix "update" + "insert" https://www.cockroachlabs.com/blog/sql-upsert/#:~:text=The%20term%20upsert%20is%20a,value%20doesn't%20already%20exist.

lib/runtime/storage/storagediff.go
dimartiro: Fixed in my last commit

lib/runtime/storage/storagediff_test.go
dimartiro: Sure! I want to finish the implementation first and I'm working on the test coverage now

dimartiro:
> A few comments, but a really nice change great work!

Thanks! I'm working on fixing the tests right now! 😄

lib/runtime/wazero/imports_test.go
dimartiro: Fixed

lib/runtime/wazero/imports_test.go
dimartiro: Fixed

lib/runtime/wazero/imports_test.go
dimartiro: Fixed

lib/runtime/storage/storagediff.go
dimartiro: Done

lib/runtime/storage/storagediff.go
dimartiro: Done

lib/runtime/storage/trie.go
```diff
@@ -64,66 +79,71 @@ func (t *TrieState) CommitTransaction() {
...
 
 // Get gets a value from the trie
 func (t *TrieState) Get(key []byte) []byte {
 	t.mtx.RLock()
 	defer t.mtx.RUnlock()
-	return t.getCurrentTrie().Get(key)
+
+	// If we find the key or it is deleted return from latest transaction
+	val, deleted := t.getCurrentTransaction().get(string(key))
+	if val != nil || deleted {
```
EclesioMeloJunior: Need to check if `getCurrentTransaction` is not nil before getting a value

EclesioMeloJunior (changes requested):
I would suggest to `storageDiff` implement the same interface as `Trie` since both should like as a `Storage` but in different layers, for example the `NextKey` method in the `TrieState`, you check first in the `storageDiff` and if no storage diff then in the `state`.

If `storageDiff` has a method `NextKey(key []byte)` that does the logic that is in the `TrieState`, then you can have a pipe in the `TrieState` to execute a closure depend if there is a transaction or not

```
type CommonStorageInterface interface {
  NextKey(k []byte)
  ...
}

func (ts *TrieState) NextKey(k []byte) {
   storage = ts.ongoingState()
   return storage.NextKey(k)
}

func (ts *TrieState) ongoingState() CommonStorageInterface {
   if len(ts.transactions) == 0 {
      return ts.state
   }
   
   return ts.transactions.last().(*storageDiff)
}
```

dimartiro:
> I would suggest to `storageDiff` implement the same interface as `Trie` since both should like as a `Storage` but in different layers, for example the `NextKey` method in the `TrieState`, you check first in the `storageDiff` and if no storage diff then in the `state`.
> 
> If `storageDiff` has a method `NextKey(key []byte)` that does the logic that is in the `TrieState`, then you can have a pipe in the `TrieState` to execute a closure depend if there is a transaction or not
> 
> ```go
> type CommonStorageInterface interface {
>   NextKey(k []byte)
>   ...
> }
> 
> func (ts *TrieState) NextKey(k []byte) {
>    storage = ts.ongoingState()
>    return storage.NextKey(k)
> }
> 
> func (ts *TrieState) ongoingState() CommonStorageInterface {
>    // check if the current transaction is nil then return the current state
>    if len(ts.transactions) == 0 {
>       return ts.state
>    }
>    
>    // otherwise return the ongoing transaction
>    return ts.transactions.last().(*storageDiff)
> }
> ```

When I was implementing the `StorageDiff` I realized it start having the same method we have in a trie and yesterday I was thinking if we will have any benefit implementing the same interface. This is a very good point to do it I really like it and will do the changes.

Thanks!

jimjbrettj (approved):
nice work!

dimartiro:
> I would suggest to `storageDiff` implement the same interface as `Trie` since both should like as a `Storage` but in different layers, for example the `NextKey` method in the `TrieState`, you check first in the `storageDiff` and if no storage diff then in the `state`.
> 
> If `storageDiff` has a method `NextKey(key []byte)` that does the logic that is in the `TrieState`, then you can have a pipe in the `TrieState` to execute a closure depend if there is a transaction or not
> 
> ```go
> type CommonStorageInterface interface {
>   NextKey(k []byte)
>   ...
> }
> 
> func (ts *TrieState) NextKey(k []byte) {
>    storage = ts.ongoingState()
>    return storage.NextKey(k)
> }
> 
> func (ts *TrieState) ongoingState() CommonStorageInterface {
>    // check if the current transaction is nil then return the current state
>    if len(ts.transactions) == 0 {
>       return ts.state
>    }
>    
>    // otherwise return the ongoing transaction
>    return ts.transactions.last().(*storageDiff)
> }
> ```

@EclesioMeloJunior, since `storageDiff` is holding the diff between the previous state and the new one, this change isn't too straightforward. One way to solve it is to create a first `storageDiff` with the entries of the current state and keep the behavior similar to the current one (before this PR), always holding a storageDiff in our transaction list. However, if we do that, we're going to hold almost the same amount of memory we're currently using with in-memory trie and the operation to get all the entries is expensive in a lazy loading context; so I don't see any benefits in making this change this way 😞

dot/state/storage_test.go
```diff
@@ -31,6 +30,8 @@ func newTestStorageState(t *testing.T) *StorageState {
 }
 
 func TestStorage_StoreAndLoadTrie(t *testing.T) {
+	t.Skip()
```
EclesioMeloJunior: does this skip have a reason? can you introduce a `TODO` comment explaining why or pointing to an issue

lib/runtime/interfaces.go
EclesioMeloJunior: I would suggest to split them into specific interfaces that composes the storage like `Trie`, `ChildTrie`, `Transactional`, `Runtime`

lib/runtime/storage/storagediff.go
```diff
@@ -0,0 +1,308 @@
...
+// states of a trie, such as those occurring during the execution of a block.
+// It records updates (upserts), deletions, and changes to child tries.
+// This mechanism facilitates applying state transitions efficiently.
+// Changes accumulated in storageDiff can be applied to a trie using
+// the `applyToTrie` method
+type storageDiff struct {
+	upserts        map[string][]byte
+	deletes        map[string]bool
+	childChangeSet map[string]*storageDiff
+	mtx            sync.RWMutex
```
EclesioMeloJunior: is this structure used among different goroutines? from what I understand this structure should be used only when a block is being executed which means that this should be only accessible in the block context and once the execution finishes it is no longer needed, right?

lib/runtime/storage/storagediff.go
```diff
@@ -0,0 +1,308 @@
...
+}
+
+// newChangeSet initialises and returns a new storageDiff instance
+func newStorageDiff() *storageDiff {
+	return &storageDiff{
+		upserts:        make(map[string][]byte),
+		deletes:        make(map[string]bool),
```
EclesioMeloJunior, line 34: use a `map[string]struct{}` then it will behave as a hashset without allocating any size as value, it works with `map[string]interface{}` as well, suggesting since you only uses the map to track keys

lib/runtime/storage/storagediff.go
```diff
@@ -0,0 +1,308 @@
...
+	if cs == nil {
+		return
+	}
+
+	cs.mtx.Lock()
+	defer cs.mtx.Unlock()
+	// If we previously deleted this trie we have to undo that deletion
+	if cs.deletes[key] {
+		delete(cs.deletes, key)
+	}
```
EclesioMeloJunior, line 66: check if the key is present, or delete the key directly `delete(map, key)`
```suggestion
	// If we previously deleted this trie we have to undo that deletion
	if _, ok := cs.deletes[key]; ok {
		delete(cs.deletes, key)
	}
```

lib/runtime/storage/storagediff.go
```diff
@@ -0,0 +1,308 @@
...
+// if the limit is exceeded.
+// This method will do not consider keys created during actual block execution
+// https://spec.polkadot.network/chap-host-api#id-version-2-prototype-2
+func (cs *storageDiff) deleteChildLimit(keyToChild string,
+	currentChildKeys []string, limit int) (
+	deleted uint32, allDeleted bool) {
+	childChanges := cs.childChangeSet[keyToChild]
+	if childChanges == nil {
+		childChanges = newStorageDiff()
+	}
```
EclesioMeloJunior, line 95: in the moment of deletion a child and the key to child does not exists then means that there is no children with that key, so we can just `return 0, true // or false` without any other operation, right?

lib/runtime/storage/storagediff.go
```diff
@@ -0,0 +1,308 @@
...
+
+	if limit == -1 {
+		cs.delete(keyToChild)
+		deletedKeys := len(childChanges.upserts) + len(currentChildKeys)
+		return uint32(deletedKeys), true
```
EclesioMeloJunior, line 100: > // This method will do not consider keys created during actual block execution

So should this method no be part of the `storageDiff` struct since this struct contains keys created under actual block execution?

lib/runtime/storage/storagediff.go
```diff
@@ -0,0 +1,308 @@
...
+	}
+
+	newKeys := maps.Keys(childChanges.upserts)
+	allKeys := append(newKeys, currentChildKeys...)
+	sort.Strings(allKeys)
```
EclesioMeloJunior: > // This method will do not consider keys created during actual block execution

using `childChanges.upserts` is the same as using keys created under actual block exec, right?

lib/runtime/storage/storagediff.go
```diff
@@ -0,0 +1,308 @@
...
+	return deleted, deleted == uint32(len(allKeys))
+}
+
+// clearPrefixInChild clears keys with a specific prefix within a child trie.
+func (cs *storageDiff) clearPrefixInChild(keyToChild string, prefix []byte,
+	childKeys []string, limit int) (deleted uint32, allDeleted bool) {
+	childChanges := cs.childChangeSet[keyToChild]
+	if childChanges == nil {
+		childChanges = newStorageDiff()
+	}
```
EclesioMeloJunior, line 130: if the child does not exist then we can just return w/o any other operation, right?

lib/runtime/storage/storagediff.go
```diff
@@ -0,0 +1,308 @@
...
+	deleted, allDeleted = childChanges.clearPrefix(prefix, childKeys, limit)
+	cs.childChangeSet[keyToChild] = childChanges
+
+	return deleted, allDeleted
+}
+
+// clearPrefix removes all keys matching a specified prefix, within an
+// optional limit. It returns the number of keys deleted and a boolean
+// indicating if all keys with the prefix were removed.
+func (cs *storageDiff) clearPrefix(prefix []byte, trieKeys []string, limit int) (deleted uint32, allDeleted bool) {
```
EclesioMeloJunior, line 140: is the `trieKeys` argument necessary?

EclesioMeloJunior (changes requested):

dot/state/storage_test.go
dimartiro: Oh sorry! I almost missed it while I was working, thanks for detecting it!

dimartiro: Solved! thanks for the good catch!

lib/runtime/storage/storagediff.go
dimartiro: I added this mtx to do it "thread safe" but currently this is only used by `TrieState` and `TrieState` already has a mutex, I believe we can remove the mtx from `StorageDiff` if you don't see any issue with that

lib/runtime/storage/storagediff.go
dimartiro: Good tip! will do

lib/runtime/storage/storagediff.go
dimartiro: I might change the comment, they are not considered in the `limit` only

lib/runtime/storage/storagediff.go
dimartiro: Yes, because we need to know all the keys from the trie, `storageDiff` contains the new keys but `trieKeys` are the current keys

lib/runtime/storage/storagediff.go
dimartiro: I was doing the change but I believe in this case is more intuitive to keep using `map[string]bool`

This will simplify comparisons like:

```go
for _, k := range keys {
   if k > string(key) && !currentTx.deletes[k] {
      return allEntries[k]
   }
}
```

instead of doing:

```go
for _, k := range keys {
   if _, deleted := currentTx.deletes[k]; k > string(key) && !deleted {
      return allEntries[k]
   }
}
```
or 

```go
if cs.deletes[key] {
   delete(cs.deletes, key)
}
```

instead of doing:

```go
if _, deleted := cs.deletes[key]; deleted {
   delete(cs.deletes, key)
}
```

In terms of memory `boolean` takes only 1 byte so I think in this case I prefer to prioritize legibility than memory usage.

lib/runtime/storage/storagediff.go
dimartiro: Based on my previous [comment](https://github.com/ChainSafe/gossamer/pull/3777#discussion_r1534424586) I think is more intuitive to keep it with `map[string]bool` in this case the `if` will return `false` if the key does not exists

lib/runtime/storage/storagediff.go
dimartiro: Changed

lib/runtime/storage/storagediff.go
dimartiro: Yep, `upserts` are the keys-values created during the actual block exec

lib/runtime/storage/storagediff.go
dimartiro: In this scenario, the absence of the child in the diff indicates no updates were made to its keys/values during this block's execution. However, the child may still exist in the current trie. Therefore, we need to generate a `newStorageDiff` to clear any keys prefixed from the current state child

lib/runtime/storage/storagediff.go
dimartiro: Same than [here](https://github.com/ChainSafe/gossamer/pull/3777#discussion_r1534434353)
`childChanges` contains the changes to that child during this block's execution, but there could be keys in the original state that we want to change, thats why we created this `newStorageDiff` here

lib/runtime/interfaces.go
dimartiro: Done

lib/runtime/storage/storagediff.go
EclesioMeloJunior: Yeah, if we can lock on trie state than we can avoid locking here

lib/runtime/storage/storagediff.go
EclesioMeloJunior: yes, but my question is: if we should not consider the keys created under actual block exec then why are we appending the `childChanges.upserts` to the `currentChildKeys` and then iterating over `allKeys` and deleting them from the `childChanges`? 

I see that you check if the current iterated key is in not the `newKeys` if so you decrease one from the limit however the if the key belongs to the `newKeys` then it will be deleted

dimartiro: I updated the comment to `Keys created during the current block execution do not count toward the limit` as it said in the spec.

We are appending the keys because we have to delete them (even if they were created during the current block's exec) but we don't have to count them on the limit

EclesioMeloJunior (approved):

lib/runtime/storage/storagediff.go
dimartiro: Done! thks

lib/runtime/storage/storagediff_test.go
dimartiro: Done

dot/state/storage_test.go
EclesioMeloJunior: @dimartiro the skip should be removed or is it intentional? if so could you add a TODO here? plz 😅

lib/runtime/interfaces.go
```diff
@@ -9,34 +9,53 @@ import (
...
 	SetVersion(v trie.TrieLayout)
 }
 
+// Storage runtime interface.
+type Storage interface {
+	Trie
+	ChildTrie
+	Transactional
+	Runtime
+}
```
EclesioMeloJunior, line 57: So clean! love it ❤️

lib/runtime/storage/storagediff.go
```diff
@@ -0,0 +1,292 @@
...
+					panic("Error applying child trie keys deletion to trie")
+				}
+			}
+		}
+	}
+
+	// Apply trie deletions
+	for k := range cs.deletes {
+		key := []byte(k)
+		child, _ := t.GetChild(key)
```
EclesioMeloJunior, line 278: why ignore the error here?

dot/state/storage_test.go
dimartiro: Weird, seems that I didn't push it. And obviously the strategy of skipping tests to see that all are passing it's not the best one haha. I'll do TDD next time. thanks for catching it.

Solves now

lib/runtime/storage/storagediff.go
dimartiro: Usually the error we could have when we are getting a `Child` is that the child doesn't exists. I'm checking if `child != nil` for that so I think it's covered.
Another way to solve it is checking the error and ignoring it if it is `ErrChildTrieDoesNotExist` 

What do you think is the best way?

dimartiro: Hey, as we talked [I added a TODO](https://github.com/ChainSafe/gossamer/pull/3777/commits/fdca5bf66379d720b6e324626ce2010ee1e95e6f) to discuss it later

PR #3780, opened 2024-03-01
chore(ci): Turn off release success comment

dimartiro:
## Changes

Turn off releases comments to prevent notifications spam

## Tests

Next time we release


.releaserc
```diff
@@ -25,7 +25,8 @@
                     {
                         "path": "dist/**"
                     }
-                ]
+                ],
+                "successComment": false
```
jimjbrettj, line 29: I am very in favor of this change but I do have a naive question, without the notifications how will we tell if the release was cut successfully?

dimartiro: Good question Jimmy. I'm not sure, but one way is to check if the release process created the [changelog commit](https://github.com/ChainSafe/gossamer/commit/1a13e4cc3a66ac1f3573433353f23085109d4cc4)

P1sar (approved):

jimjbrettj (approved):

Issue #3758, opened 2024-02-15
Libp2p connect handshake error and outgoing slot permanent occupation


P1sar:
## Describe the bug

When node is running it writes all outgoing and incoming connection in to the slots that has limited capacity. But when libp2phost.connect happens with error there is no fallback route for us to free occupied connection slot inside PeerState. 
Error happens pereodically [here](https://github.com/ChainSafe/gossamer/blob/development/dot/network/service.go#L718).
Another problem is that apprently before libp2p ibrary version 32.2 there is a bug that does not bubbleups an error so this happens unnoticeably this statement also requires verification.  

Mentioned below log was done within this [branch](https://github.com/ChainSafe/gossamer/pull/3755) with additional info logs added 
## Expected Behavior
IF libp2p error happens we should free occupied perstate slot.
[here](https://github.com/ChainSafe/gossamer/blob/development/dot/network/service.go#L718)

-

## Current Behavior
PerrState slot is not getting released when libp2p.connect error hapens
```
 failed to negotiate security protocol: error reading handshake message: noise: message is too short
```

-

## Possible Solution
Basically handshake error is out of scope of this issue and generally we just need to add communication between network.service and peerstate so we can free occupied slot in case of error.

-

## To Reproduce
Steps to reproduce the behaviour:
Run integration test
1.  MODE=rpc go test ./tests/rpc/system_integration_test.go -v
2.  libp2p version should be 32.2 before that error happens silently without bubbling up
3. 

## Log output

<details>
  <summary>Log Output</summary>
  ```
   system_integration_test.go:54: Node alice-0 is ready
alice-0 2024-02-13T20:32:07+01:00 INFO     Allocating slots for 0	pkg=peerset
2024-02-13T20:32:07+01:00 INFO     hasFreeOutgoingSlot 0, numout 0  maxout 10	pkg=peerset
2024-02-13T20:32:07+01:00 INFO     hasFreeOutgoingSlot 0, numout 0  maxout 10	pkg=peerset
alice-0 2024-02-13T20:32:09+01:00 INFO     peer count 0	pkg=network
alice-0 2024-02-13T20:32:09+01:00 INFO     Allocating slots for 0	pkg=peerset
2024-02-13T20:32:09+01:00 INFO     hasFreeOutgoingSlot 0, numout 0  maxout 10	pkg=peerset
2024-02-13T20:32:09+01:00 INFO     hasFreeOutgoingSlot 0, numout 0  maxout 10	pkg=peerset
[...]

PR #3787, opened 2024-03-04
chore: add network integration test with kademilia instead mDNS

P1sar:
## Changes
While I was debugging network module I found out that our network tests using mDNS instead of Kademlia and so bypass or uses logic that is not actually used in Kademilia so I decided to implement integration test that covers producion code. 

## Tests
`MODE=network go test ./tests/network/netwrok_integration_test.go -v`

```sh
```

## Issues
Related: https://github.com/ChainSafe/gossamer/issues/3758

## Primary Reviewer

@


dot/network/discovery.go
```diff
@@ -233,6 +223,13 @@ func (d *discovery) findPeers() {
 	}
 }
 
-func (d *discovery) findPeer(peerID peer.ID) (peer.AddrInfo, error) {
-	return d.dht.FindPeer(d.ctx, peerID)
+func (d *discovery) stop() error {
```
jimjbrettj, line 226: was there a reason to move this function?

dot/network/service.go
```diff
@@ -292,14 +292,11 @@ func (s *Service) Start() error {
...
 	// connectHandler is a part of libp2p.Notifiee interface implementation and getting called in the very end
-	//after or Incoming or Outgoing node is connected
+	// after or Incoming or Outgoing node is connected.
 	s.host.cm.connectHandler = func(peerID peer.ID) {
 		for _, prtl := range s.notificationsProtocols {
 			prtl.peersData.setMutex(peerID)
 		}
-		// TODO: currently we only have one set so setID is 0, change this once we have more set in peerSet
-		const setID = 0
-		s.host.cm.peerSetHandler.Incoming(setID, peerID)
```
jimjbrettj, line 302: We do not need this code anymore?

jimjbrettj:
nit/rename file to ./tests/network/network_integration_test.go

tests/network/netwrok_integration_test.go
```diff
@@ -0,0 +1,107 @@
...
+	"github.com/adrg/xdg"
+	"github.com/stretchr/testify/require"
+
+	"context"
+	"os"
+	"testing"
+	"time"
+)
+
+func TestKadDHTNetworkDiscovery(t *testing.T) {
```
jimjbrettj: This test actually does not work locally for me and it times out. Based on my logs it seems that Alice progresses in the network but Bob and Charlie do not

dot/network/discovery.go
P1sar: I have removed discovery) findPeer justt to make  service smaller, there is no reason to keep this as a function when we can call it directly. 
I moved stop to the very end to give logic to method sort, I usually like to have stop in the very end and start in the begging of the file. But I guess it is pretty subjective

tests/network/netwrok_integration_test.go
P1sar: hmmm. how do you run it? it works fine for me, trying to undersrtand what I missed

dot/network/service.go
```diff
@@ -711,7 +708,7 @@ func (s *Service) processMessage(msg peerset.Message) {
 		addrInfo := s.host.p2pHost.Peerstore().PeerInfo(peerID)
 		if len(addrInfo.Addrs) == 0 {
 			var err error
-			addrInfo, err = s.host.discovery.findPeer(peerID)
+			addrInfo, err = s.host.discovery.dht.FindPeer(s.host.discovery.ctx, peerID)
```
EclesioMeloJunior: I would suggest to create a new context with timeout here then we have a limit while finding a peer
```suggestion
                         ctx, cancel := context.WithTimeout(s.host.discovery.ctx, _some_timeout_limit_)
                         defer cancel()
			addrInfo, err = s.host.discovery.dht.FindPeer(ctx, peerID)
```

tests/network/netwrok_integration_test.go
```diff
@@ -0,0 +1,107 @@
...
+	con.Core.Role = common.FullNodeRole
+	con.RPC.Modules = []string{"system", "author", "chain"}
+	con.Network.MinPeers = 1
+	con.Network.MaxPeers = 20
+	con.Network.NoMDNS = true // Turning off mDNS, purpose of this test is purly use only kadDHT discovery
+	con.Core.BabeAuthority = true
+	con.Log.Sync = "trace"
+	con.Network.Port = 7001
+	// ID: 12D3KooWMHixgmjFYM4VyQNDTKMvN9BPw47Tyyb6LPZ43EavV68m
+	con.BasePath = xdg.DataHome + "/gossamer/westend-local/alice"
```
EclesioMeloJunior: I would suggest to use `t.TempDir()` to create an ephemeral dir where we can test

EclesioMeloJunior (changes requested):

dot/network/service.go
P1sar: I was testing this part local and had some doubts about necessety of this part although eventually decided to return it back and rather remove it in separate PR with additional tests.

EclesioMeloJunior (approved):
LGTM

dot/network/discovery.go
jimjbrettj: I like it! I am good with this change was just curious

tests/network/netwrok_integration_test.go
jimjbrettj: retried and seems to be good now

jimjbrettj (approved):
lgtm

PR #3788, opened 2024-03-04
chore: add infra label

P1sar:
## Changes
Just add a new label

## Tests

```sh
```

## Primary Reviewer

@


jimjbrettj (approved):

dimartiro (approved):
I think we should also add the `dependencies` label to prevent[ these dependabot error messages](https://github.com/ChainSafe/gossamer/pull/3785#issuecomment-1975480431)

Issue #3791, opened 2024-03-05
fix(clear_prefix_version_2): decoding of the limit argument


radkomih:
## Issue summary

- `ext_storage_clear_prefix_version_2` is expected to receive the limit argument encoded as an Option of u32, but in the current implementation, it is decoded incorrectly, also it is encoded incorrectly in the test.

## Other information and links

- N/A

PR #3792, opened 2024-03-05
fix: decode limit as option of uint32

radkomih:
## Changes

* [x] fix decoding of the limit argument
* [x] fix encoding in the test 

## Tests

```sh
```

## Issues

#3791 

## Primary Reviewer

@

### DIFF
--- a/lib/runtime/wazero/imports.go
+++ b/lib/runtime/wazero/imports.go
@@ -2140,21 +2140,19 @@ func ext_storage_clear_prefix_version_2(ctx context.Context, m api.Module, prefi
 
 	limitBytes := read(m, lim)
 
-	var limitUint uint32
-
-	hasValue := limitBytes[0] == 1
-	if hasValue {
-		err := scale.Unmarshal(limitBytes[1:], &limitUint)
-		if err != nil {
-			logger.Warnf("failed scale decoding limit: %s", err)
-			panic(err)
-		}
-	} else {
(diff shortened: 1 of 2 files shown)


lib/runtime/wazero/imports.go
```diff
@@ -2139,20 +2140,20 @@ func ext_storage_clear_prefix_version_2(ctx context.Context, m api.Module, prefi
...
-		logger.Warnf("failed scale decoding limit: %s", err)
-		panic(err)
-	}
+	var limitUint uint32
 
-	if len(limit) == 0 {
-		// limit is None, set limit to max
-		limit = []byte{0xff, 0xff, 0xff, 0xff}
+	hasValue := limitBytes[0] == 1
+	if hasValue {
```
EclesioMeloJunior: For decoding Options using Gossamer scaled codec you can use `*uint32`:
```go
var limit *uint32

err := scale.Unmarshal(limitBytes, &limit)
if err != nil {
  panic("cannot unmarshal")
}

// if it is None
if limit == nil {
  *limit = math.MaxUint32
}

...
```

EclesioMeloJunior (changes requested):

EclesioMeloJunior (approved):
LGTM! Thanks for your contribution @radkomih

dimartiro (approved):
Thanks for the contribution @radkomih

PR #3796, opened 2024-03-08
feat(types/header): place slot get at `types.Header`

EclesioMeloJunior:
## Changes

- Move the slot number from `dot/epoch` to `types/header` since the slot number belongs to the block header, also makes things easy to understand

## Tests

```sh
N/A
```

## Issues

N/A

## Primary Reviewer

@


jimjbrettj (approved):
lgtm

dimartiro (approved):
LGTM

Issue #3740, opened 2024-02-06
Paseo: invalid signature error causing node not to sync on restart [Graceful shutdown] 


jimjbrettj:
## Issue summary

When stopping a running paseo node I get a log as follows:

![image](https://github.com/ChainSafe/gossamer/assets/40345116/9085ed9e-3fc0-4197-bfb5-2f216edb00f0)

Then after on restarting the node the sync workers are all deployed and retrieving blocks, however, finalization is not occurring and I get the error: 

![image](https://github.com/ChainSafe/gossamer/assets/40345116/ae2aed53-49b9-47ad-bee5-865983433a20)

PR #3797, opened 2024-03-12
fix(node): fix node restart

jimjbrettj:
## Changes

- Implement pause functionality on state and sync services. The issue with our shutdown is that our services have different shutdown times and our state gets corrupted as some services are running while others stopped. To get around this I build the functionality to pause key services on shutdown to avoid the state being corrupted.

## Tests

1) Start a node 
`./bin/gossamer init --chain paseo` then 
`./bin/gossamer --chain paseo`

2) Stop node with sigint
3) Restart node and see if it resumes syncing `./bin/gossamer --chain paseo`

## Issues

#3740

## Primary Reviewer

@EclesioMeloJunior

### DIFF
--- a/dot/state/block.go
+++ b/dot/state/block.go
@@ -63,8 +63,8 @@ type BlockState struct {
 	tries             *Tries
 
 	// State variables
-	mtx   sync.RWMutex
-	pause chan struct{}
+	pausedLock sync.RWMutex
+	pause      chan struct{}
 
 	// block notifiers
 	imported                       map[chan *types.Block]struct{}
@@ -161,8 +161,8 @@ func NewBlockStateFromGenesis(db database.Database, trs *Tries, header *types.He
 
 // Pause pauses the service ie. halts block production
 func (bs *BlockState) Pause() error {
-	bs.mtx.Lock()
-	defer bs.mtx.Unlock()
+	bs.pausedLock.Lock()
+	defer bs.pausedLock.Unlock()
 
 	if bs.IsPaused() {
 		return nil
--- a/dot/state/block.go
+++ b/dot/state/block.go
@@ -63,7 +63,7 @@ type BlockState struct {
 	tries             *Tries
 
 	// State variables
-	sync.RWMutex
+	mtx   sync.RWMutex
 	pause chan struct{}
 
 	// block notifiers
@@ -161,8 +161,8 @@ func NewBlockStateFromGenesis(db database.Database, trs *Tries, header *types.He
 
 // Pause pauses the service ie. halts block production
 func (bs *BlockState) Pause() error {
-	bs.Lock()
-	defer bs.Unlock()
+	bs.mtx.Lock()
+	defer bs.mtx.Unlock()
 
 	if bs.IsPaused() {
 		return nil
--- a/dot/state/block.go
+++ b/dot/state/block.go
@@ -62,6 +62,10 @@ type BlockState struct {
 	unfinalisedBlocks *hashToBlockMap
 	tries             *Tries
 
+	// State variables
+	sync.RWMutex
+	pause chan struct{}
+
 	// block notifiers
 	imported                       map[chan *types.Block]struct{}
 	finalised                      map[chan *types.FinalisationInfo]struct{}
@@ -85,6 +89,7 @@ func NewBlockState(db database.Database, trs *Tries, telemetry Telemetry) (*Bloc
 		finalised:                  make(map[chan *types.FinalisationInfo]struct{}),
 		runtimeUpdateSubscriptions: make(map[uint32]chan<- runtime.Version),
 		telemetry:                  telemetry,
+		pause:                      make(chan struct{}),
 	}
 
 	gh, err := bs.db.Get(headerHashKey(0))
@@ -120,6 +125,7 @@ func NewBlockStateFromGenesis(db database.Database, trs *Tries, header *types.He
 		genesisHash:                header.Hash(),
 		lastFinalised:              header.Hash(),
 		telemetry:                  telemetryMailer,
+		pause:                      make(chan struct{}),
 	}
 
 	if err := bs.setArrivalTime(header.Hash(), time.Now()); err != nil {
@@ -151,6 +157,29 @@ func NewBlockStateFromGenesis(db database.Database, trs *Tries, header *types.He
 	}
 
 	return bs, nil
+}
+
+// Pause pauses the service ie. halts block production
+func (bs *BlockState) Pause() error {
+	bs.Lock()
+	defer bs.Unlock()
+
+	if bs.IsPaused() {
+		return nil
+	}
+
+	close(bs.pause)
+	return nil
+}
+
+// IsPaused returns if the service is paused or not (ie. producing blocks)
+func (bs *BlockState) IsPaused() bool {
+	select {
+	case <-bs.pause:
+		return true
+	default:
+		return false
+	}
 }
 
 // encodeBlockNumber encodes a block number as big endian uint64
--- a/dot/state/block_finalisation.go
+++ b/dot/state/block_finalisation.go
@@ -127,10 +127,8 @@ func (bs *BlockState) SetFinalisedHash(hash common.Hash, round, setID uint64) er
 	bs.lock.Lock()
 	defer bs.lock.Unlock()
 
-	select {
-	case <-bs.pause:
+	if bs.IsPaused() {
 		return errors.New("blockstate service is paused")
-	default:
 	}
 
 	has, err := bs.HasHeader(hash)
--- a/dot/state/block_finalisation.go
+++ b/dot/state/block_finalisation.go
@@ -127,6 +127,13 @@ func (bs *BlockState) SetFinalisedHash(hash common.Hash, round, setID uint64) er
 	bs.lock.Lock()
 	defer bs.lock.Unlock()
 
+	select {
+	case <-bs.pause:
+		return errors.New("service paused")
+	default:
+
+	}
+
 	has, err := bs.HasHeader(hash)
 	if err != nil {
 		return fmt.Errorf("could not check header for hash %s: %w", hash, err)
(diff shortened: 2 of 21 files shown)


lib/services/services.go
```diff
@@ -52,9 +56,27 @@ func (s *ServiceRegistry) StartAll() {
 	s.logger.Debug("All services started.")
 }
 
+// PauseServices pauses key services before shutdown to allow a graceful shutdown
+func (s *ServiceRegistry) PauseServices() {
```
timwu20, line 61: Not sure why we need to introduce `Pause()` to the service interface if we just need to stop the sync and state services before we stop the others.  I would just modify `StopAll` to be less generic and stop the sync and state services first before stopping the rest.

timwu20 (changes requested):
I don't understand wj

lib/services/services.go
jimjbrettj: This is not possible as state service is a dependency of all the other services, so shutting it down anywhere but last causes a panic. This is what made this tricky in the first place. Introducing `Pause()` is a workaround to essentially stop the services without actually shutting it down

timwu20: Oh I see.  I would rather explicitly pause the sync and state services rather than introduce the `Pause` method for all services.  You can just access the `sync.Service` or `state.Service` via the map using `reflect.TypeOf` instead of ranging through each service.  You can then just cast specifically to `sync.Service` or `state.Service` which will have the `Pause` method.  That way you don't need to introduce it to every single `Service`.

timwu20 (dismissed):

lib/services/services.go
jimjbrettj: Oh nice, I like this idea. Will implement

jimjbrettj: updated

dot/sync/errors.go
```diff
@@ -8,11 +8,7 @@ import (
 )
 
 var (
-	// ErrServiceStopped is returned when the service has been stopped
-	ErrServiceStopped = errors.New("service has been stopped")
```
jimjbrettj, line 12: These errors were never used

dot/sync/chain_sync.go
```diff
@@ -536,15 +546,38 @@ func (cs *chainSync) requestMaxBlocksFrom(bestBlockHeader *types.Header, origin
 		}
 	}
 
-	resultsQueue := cs.workerPool.submitRequests(requests)
-	err := cs.handleWorkersResults(resultsQueue, origin, startRequestAt, expectedAmountOfBlocks)
+	resultsQueue, err := cs.submitRequests(requests)
+	if err != nil {
+		return err
```
EclesioMeloJunior, line 552: would suggest to add a context: `fmt.Errorf("submitting request: %w", err)`

dot/sync/chain_sync.go
```diff
@@ -582,12 +615,16 @@ func (cs *chainSync) handleWorkersResults(
 	workersResults chan *syncTaskResult, origin blockOrigin, startAtBlock uint, expectedSyncedBlocks uint32) error {
 	startTime := time.Now()
 
+	// TODO one idea is can refactor this to be a map (will be not as efficient tho, so prob not desired)
```
EclesioMeloJunior: I don't think using a map here can bring benefits, one of the reasons is that it should ensure the order of the synced blocks and maps does not hold this constraint.

dot/sync/chain_sync.go
```diff
@@ -582,12 +615,16 @@ func (cs *chainSync) handleWorkersResults(
...
 	syncingChain := make([]*types.BlockData, expectedSyncedBlocks)
 	// the total numbers of blocks is missing in the syncing chain
 	waitingBlocks := expectedSyncedBlocks
 
 taskResultLoop:
 	for waitingBlocks > 0 {
+		//if cs.blockState.IsPaused() {
+		//	return errors.New("blockstate is paused")
+		//}
```
EclesioMeloJunior: remove

lib/services/services.go
```diff
@@ -52,9 +55,39 @@ func (s *ServiceRegistry) StartAll() {
...
+
+	s.logger.Infof("Paused key services")
+}
+
 // StopAll calls `Service.Stop()` for all registered services
 func (s *ServiceRegistry) StopAll() {
 	s.logger.Infof("Stopping services: %v", s.serviceTypes)
+
+	s.PauseServices()
+
```
EclesioMeloJunior, line 83: Looks a little bit odd to me that we have a method called PauseServices that pauses only two specific services (sync and block state) and then we have a loop that goes over the services types (that includes sync and block state) and stop them again.

I think this is convoluted and one option would be to have some list that holds the order to stop services, then you can always ensure that sync and state services are going to be stopped first then any other service

**[edit]** Given that you should `Pause` services before stopping them I would suggest having an ordered list of services to pause, you can check where `s.serviceTypes` is populated then populate this ordered list, and when you want to pause these services just iterate over the ordered list and pause them

EclesioMeloJunior (changes requested):

lib/services/services.go
dimartiro: Agree, we discussed it with @jimjbrettj in a private chat last week, the best solution should be having a "dependency list / tree" to start and stop the services following that specific order. But I think this PR aims to be a hotfix instead of a final solution

EclesioMeloJunior: I believe that this modification can be included in this PR

dot/sync/chain_sync.go
jimjbrettj: yeah i actually meant to delete this for exactly this reason, will update

dot/sync/chain_sync.go
jimjbrettj: i am adding this to the helper functions so this will be applied on all calls to submit requests

EclesioMeloJunior (approved):

lib/services/services.go
jimjbrettj: updated!

dot/state/block_finalisation.go
```diff
@@ -127,6 +127,12 @@ func (bs *BlockState) SetFinalisedHash(hash common.Hash, round, setID uint64) er
 	bs.lock.Lock()
 	defer bs.lock.Unlock()
 
+	select {
+	case <-bs.pause:
+		return errors.New("blockstate service is paused")
+	default:
+	}
+
```
dimartiro: Why not just:

```suggestion
	if bs.IsPaused() {
		return errors.New("blockstate service is paused")
	}
```

dot/state/service.go
```diff
@@ -43,6 +43,14 @@ type Service struct {
 	BabeThresholdDenominator uint64
 }
 
+// Pause Pauses the state service
+func (s *Service) Pause() error {
+	if !s.Block.IsPaused() {
+		return s.Block.Pause()
+	}
+	return nil
```
dimartiro: Since `s.Block.Pause()` already has the `s.Block.IsPaused()` check we can simplify this logic to:

```suggestion
	return s.Block.Pause()
```

dot/sync/chain_sync.go
```diff
@@ -262,6 +262,9 @@ func (cs *chainSync) bootstrapSync() {
 			cs.workerPool.useConnectedPeers()
 			err = cs.requestMaxBlocksFrom(currentBlock, networkInitialSync)
 			if err != nil {
+				if errors.Is(err, errBlockStatePaused) {
+					return
```
dimartiro, line 267: We can maybe add a log here to know that we are trying to sync while the service is paused? 🤔

dot/sync/syncer.go
```diff
@@ -26,6 +26,14 @@ type Service struct {
 	network    Network
 }
 
+// Pause Pauses the sync service
+func (s *Service) Pause() error {
+	if !s.blockState.IsPaused() {
+		return s.blockState.Pause()
+	}
+	return nil
```
dimartiro: Why not just:

```suggestion
	return s.blockState.Pause()
```

dimartiro (changes requested):
I added some suggestions and I have some questions:

- Why do we have to use a channel to pause the service instead of using a simple `isPaused` boolean? is that to prevent blocking on reads?

- Another question is, what is the difference between `pause` and `stop`? are we planning to add a `resume` method in the future to resume a stopped service?

dot/state/block.go
```diff
@@ -62,6 +62,10 @@ type BlockState struct {
 	unfinalisedBlocks *hashToBlockMap
 	tries             *Tries
 
+	// State variables
+	sync.RWMutex
```
dimartiro, line 66: Do we need to expose this lock? can we do it private? Or even get rid of it?

dot/state/block.go
```diff
@@ -153,6 +159,29 @@ func NewBlockStateFromGenesis(db database.Database, trs *Tries, header *types.He
...
+func (bs *BlockState) Pause() error {
+	bs.Lock()
+	defer bs.Unlock()
+
+	if bs.IsPaused() {
+		return nil
+	}
+
+	close(bs.pause)
+	return nil
```
dimartiro, line 172: If we are using the lock only for this, why if we change this logic to:

```suggestion
        select {
	case <-pause:
	default:
		close(pause)
	}
	return nil
```

dot/state/block.go
dimartiro: Do we need to export this mtx? are we using it outside of `BlockState` ?

jimjbrettj:
> I added some suggestions and I have some questions:
> 
>     * Why do we have to use a channel to pause the service instead of using a simple `isPaused` boolean? is that to prevent blocking on reads?
> 
>     * Another question is, what is the difference between `pause` and `stop`? are we planning to add a `resume` method in the future to resume a stopped service?

@dimartiro This is a good idea about just using a boolean in both cases, initially I had just copied babe impl of pause but I don't think we need to do this. As far as the difference between pause and stop, stop shuts down the service whereas pause halts key functionality without actually shutting down the service. In this use case resume is not needed since we only use pause on shutdown

dot/sync/syncer.go
jimjbrettj: nice call!

dot/state/service.go
jimjbrettj: nice call!

dot/state/block.go
jimjbrettj: no we do not, updated

dimartiro (approved):
LGTM now! Good job Jimmy

Issue #3430, opened 2023-08-09
BABE: added reusing last epoch data in case of skipped epochs


EclesioMeloJunior:
## Issue summary

- Update to BABE spec to reuse last epoch data in case of skipped epochs

## Other information and links

- https://github.com/w3f/polkadot-spec/pull/685 (w3f spec update)
- https://github.com/w3f/polkadot-spec/issues/681 (issue describing the problem)
- https://github.com/paritytech/substrate/pull/11727 (substrate update)

## Tasks related

- [x] https://github.com/ChainSafe/gossamer/pull/3796
- [x] https://github.com/ChainSafe/gossamer/issues/3795
- [x] https://github.com/ChainSafe/gossamer/issues/3866
- [x] https://github.com/ChainSafe/gossamer/issues/3915
- [x] https://github.com/ChainSafe/gossamer/issues/3881

EclesioMeloJunior:
Openned PR https://github.com/ChainSafe/gossamer/pull/3841 that introduces the new approach to retrieve the very first slot

PR #3799, opened 2024-03-14
fix(lib/blocktree): improve `blocktree.GetHashesAtNumber`

EclesioMeloJunior:
## Changes

- Change the name from `GetBlocksAtNumber` to `GetHashesAtNumber` given the fact that this function returns a slice of hashes and not a slice of blocks
- Improve the function to run based on the actual number and not based on the parent hash of the desired number?
- The function first try to search the slice of hashes with the same number in the in-memory block tree first if results are found then we return the hashes, if nothing returns then try to retrieve the hash from database (we map the number -> hash in the db)

- These changes are needed to complete issue https://github.com/ChainSafe/gossamer/issues/3430. Gossamer should be aware of the very first slot number (slot number for the first non genesis block) and might exist a case where more than 1 block is the first non genesis block, then we should use `GetHashesAtNumber` to check if there is more than one block `#1` and in this case use the correct first slot
-  
## Tests

```sh
```

## Issues

- Related to https://github.com/ChainSafe/gossamer/issues/3430

## Primary Reviewer

@


lib/blocktree/blocktree.go
```diff
@@ -120,24 +120,24 @@ func (bt *BlockTree) AddBlock(header *types.Header, arrivalTime time.Time) (err
 	return nil
 }
 
-// GetAllBlocksAtNumber will return all blocks hashes with the number of the given hash plus one.
+// GetHashesAtNumber will return all blocks hashes that contains the number of the given hash plus one.
 // To find all blocks at a number matching a certain block, pass in that block's parent hash
-func (bt *BlockTree) GetAllBlocksAtNumber(hash common.Hash) (hashes []common.Hash) {
+func (bt *BlockTree) GetHashesAtNumber(number uint) (hashes []common.Hash) {
```
jimjbrettj, line 125: Nit/maybe this should be `GetBlockHashesAtNumber`? Up to you

lib/blocktree/node.go
```diff
@@ -59,20 +59,23 @@ func (n *node) getNode(h common.Hash) *node {
...
-		// number matches
-		if child.number == number {
-			hashes = append(hashes, child.hash)
-		}
-
-		// are deeper than desired number, return
-		if child.number > number {
-			return hashes
+// hashesAtNumber returns all nodes in the chain that contains the desired number
+func (n *node) hashesAtNumber(number uint, hashes []common.Hash) []common.Hash {
```
jimjbrettj, line 63: nit/same here, maybe `blockHashesAtNumber`

jimjbrettj (approved):
few comments, but lgtm

dot/state/grandpa_test.go
```diff
@@ -143,7 +143,7 @@ func testBlockState(t *testing.T, db database.Database) *BlockState {
 	return bs
 }
 
-func TestAddScheduledChangesKeepTheRightForkTree(t *testing.T) {
+func TestAddScheduledChangesKeepTheRightForkTree(t *testing.T) { //nolint:tparallel
```
timwu20, line 146: why did you have to introduce this `nolint`?

lib/blocktree/blocktree.go
```diff
@@ -120,24 +120,24 @@ func (bt *BlockTree) AddBlock(header *types.Header, arrivalTime time.Time) (err
...
+	if number < bt.root.number {
+		return []common.Hash{}
 	}
 
-	number := bt.getNode(hash).number + 1
-
-	if bt.root.number == number {
-		hashes = append(hashes, bt.root.hash)
-		return hashes
+	bestLeave := bt.leaves.bestBlock()
```
timwu20, line 133:
```suggestion
	bestLeaf := bt.leaves.bestBlock()
```

lib/blocktree/blocktree.go
```diff
@@ -120,24 +120,24 @@ func (bt *BlockTree) AddBlock(header *types.Header, arrivalTime time.Time) (err
...
 // To find all blocks at a number matching a certain block, pass in that block's parent hash
-func (bt *BlockTree) GetAllBlocksAtNumber(hash common.Hash) (hashes []common.Hash) {
+func (bt *BlockTree) GetHashesAtNumber(number uint) (hashes []common.Hash) {
 	bt.RLock()
 	defer bt.RUnlock()
 
-	if bt.getNode(hash) == nil {
-		return hashes
+	if number < bt.root.number {
+		return []common.Hash{}
```
timwu20, line 130:
```suggestion
		return nil
```

lib/blocktree/blocktree.go
```diff
@@ -120,24 +120,24 @@ func (bt *BlockTree) AddBlock(header *types.Header, arrivalTime time.Time) (err
...
 	}
 
-	number := bt.getNode(hash).number + 1
-
-	if bt.root.number == number {
-		hashes = append(hashes, bt.root.hash)
-		return hashes
+	bestLeave := bt.leaves.bestBlock()
+	if number > bestLeave.number {
+		return []common.Hash{}
```
timwu20, line 135:
```suggestion
		return nil
```

lib/blocktree/node.go
```diff
@@ -59,20 +59,23 @@ func (n *node) getNode(h common.Hash) *node {
...
+	if number == n.number {
+		hashes = append(hashes, n.hash)
+		return hashes
+	}
+
+	// if the number is greater than current node,
+	// then search among its children
+	if number > n.number {
+		for _, children := range n.children {
+			hashes = children.hashesAtNumber(number, hashes)
```
timwu20, line 75: why does this function return the same slice that is passed in?  Either it should return a slice and not accept the parameter, or just accept the slice as a param that it will append to.

timwu20 (changes requested):

dot/state/grandpa_test.go
EclesioMeloJunior: I've introduced this `nolint` to avoid using `t.Parallel()` under `t.Run` test cases, this test specifically creates a chain with forks and uses the same chain for each use case

lib/blocktree/node.go
EclesioMeloJunior: actually, the function appends to the slice passed as argument and returns an updated version of it

```go
func example(slice []common.Hash) {
   slice = append(slice, value) // the slice passed as argument is not updated
}

func example(slice *[]common.Hash) {
  *slice = append(*slice, value) // this updates the outer slice
}

func example(slice []common.Hash) []common.Hash {
   slice = append(*slice, value)
   return slice // return the updated slice has the same efficiency as passing a pointer
}
```

timwu20: Oh right I got mixed up with slices and maps.  Maybe it would make sense to pass in a pointer to a slice.  But I'm fine with leaving it as is.

timwu20 (approved):

PR #3803, opened 2024-03-21
refactor(pkg/trie): Encapsulate in-memory trie impl in its own package

### LINKED ISSUE #3804: (pkg/trie): Encapsulate inmemory trie logic in its own package
## Description

The goal is to encapsulate all the `in-memory trie` related code into the same folder within `pkg/trie`

### Benefits

- We are going to have a better organized codebase
- It is a dependency to start implementing the lazy loading trie (TrieDB) without mixing both codebases
   - The idea is to have independent packages for different trie implementations

dimartiro:
## Changes

This PR is to create an independent `inmemory` package within `pkg/trie` to encapsulate the `InmemoryTrie` implementation there.
This is useful to prevent mixing the new code for the lazy loading trie implementation (aka: TrieDB) with the old inmemory trie implementation

## Summary of changes

- [x] Move all files related with the in-memory trie implementation to `/pkg/trie/inmemory` folder
   - [x] The structures or methods that are shared between different trie implementation remains in `/pkg/trie/
   - [x] Update imports to import `inmemory_trie` as alias instead of using old `trie` package
   - [x] Add new `layout_test` as part of `/pkg/trie/inmemory` to test the layout for the in-memory trie implementation 
   - [x] Move `/pkg/trie/proof` to `/pkg/trie/inmemory/proof` since it's coupled with the trie implementation 
- [x] Add a `Trie` parameter to `layout.Root` function to make it generic based on `Trie` implementation

## Tests

```sh
make test
```
```
I also did run a node locally from genesis until block 1M (stopping it and restarting it a few times)
```

## Issues

closes #3804 

## Primary Reviewer

@jimjbrettj


pkg/trie/db/db.go
```diff
@@ -6,9 +6,22 @@ import (
...
 
+type Writer interface {
+	Put(key, value []byte) error
+	Del(key []byte) error
+	Flush() error
+}
+
+// NewBatcher creates a new database batch.
+type NewBatcher interface {
+	//TODO: to make this pkg standalone we need to get rid of thid database dep
```
jimjbrettj:
```suggestion
	//TODO: to make this pkg standalone we need to get rid of this database dep
```

pkg/trie/errors.go
```diff
@@ -0,0 +1,5 @@
+package trie
```
jimjbrettj, line 4: Think will need copyright here

dimartiro: True, is weird the license check didn't complain about it. Will add it

dimartiro: Done

jimjbrettj (approved):
lgtm

EclesioMeloJunior (approved):

Issue #3824, opened 2024-03-28
 Implement collator protocol's collator side 


kishansagathiya:
## Issue summary

This is not necessary to become an authoring node, but for becoming a collator node.
This might be a low priority, but having some of the things from this subsystem implemented helps us in the test when we want to spin up a collator node.

## Sub Issues

- [x] A skeleton PR, that introduces this subsystem, registers it, has processMessage, implements SubSystem interface
- [ ] handle collation seconded message
- [ ] handle distribute collation message
- [ ] handle network bridge update messages 
- [ ] handle active leaves update signal
- [ ] handle block finalized signal
- [ ] support v2 and v3 of the protocol

## Other information and links

- https://paritytech.github.io/polkadot-sdk/book/node/collators/index.html
- https://github.com/paritytech/polkadot-sdk/blob/aa68ea58f389c2aa4eefab4bf7bc7b787dd56580/polkadot/node/network/collator-protocol/src/collator_side/

PR #3825, opened 2024-03-28
Collator Protocol Collator Side Skeleton

kishansagathiya:
This commit introduces collator side of the collator protocol. This is just a skeleton that
- implements the SubSystem interface for the new subsystem
- registers it
- add common methods like processMessage

linked to issue #3824

## Tests

```sh
```

## Issues

#3824

## Primary Reviewer

@axaysagathiya


dot/parachain/collator-protocol/collator_side.go
```diff
@@ -0,0 +1,150 @@
...
+
+func (cpcs CollatorProtocolCollatorSide) ProcessBlockFinalizedSignal(signal parachaintypes.BlockFinalizedSignal) {
+	// NOTE: nothing to do here
+}
+
+func (cpcs CollatorProtocolCollatorSide) handleCollationMessage(
+	sender peer.ID, msg network.NotificationsMessage) (bool, error) {
+
+	// we don't propagate collation messages, so it will always be false
+	propagate := false
```
EclesioMeloJunior: use `const propagate = false`

dot/parachain/collator-protocol/collator_side.go
```diff
@@ -0,0 +1,150 @@
...
+		}, sender)
+	case 1: // AdvertiseCollation
+		logger.Errorf("unexpected collation advertise collation message from peer %s, decreasing its reputation", sender)
+		cpcs.net.ReportPeer(peerset.ReputationChange{
+			Value:  peerset.UnexpectedMessageValue,
+			Reason: peerset.UnexpectedMessageReason,
+		}, sender)
+	case 2: // CollationSeconded
+		// TODO: handle collation seconded message #3824
+	}
```
EclesioMeloJunior: use iota to each collator message type index

```go
type collatorProtocolMessage int
const (
   Declare collatorProtocolMessage = iota
   AdvertiseCollation
   CollationSeconded
)
```

dot/parachain/collator-protocol/collator_side.go
```diff
@@ -0,0 +1,150 @@
...
+			network.CollationMsgType,
+			getCollatorHandshake,
+			decodeCollatorHandshake,
+			validateCollatorHandshake,
+			decodeCollationMessage,
+			cpcs.handleCollationMessage,
+			nil,
+			MaxCollationMessageSize,
+		)
+
```
EclesioMeloJunior: This does not look useful for me, the unique error case of `RegisterNotificationsProtocol` is if the message type, in this case `CollationMsgType`, is already registered, which means that the legacy protocol will never be registered.

Would be nice if we could both of these protocols, but as I said `RegisterNotificationsProtocol` method can only hold one protocol per each message type, so I've created an issue to implement these missing part https://github.com/ChainSafe/gossamer/issues/3846

EclesioMeloJunior (changes requested):

dot/parachain/collator-protocol/collator_side.go
kishansagathiya: have removed legacy protocol registration.

kishansagathiya: there might be very few nodes using legacy protocols, so it might be safe to even not support it.

dot/parachain/collator-protocol/collator_side.go
kishansagathiya: done

EclesioMeloJunior (approved):

dot/parachain/collator-protocol/collator_side.go
```diff
@@ -0,0 +1,137 @@
...
+)
+
+// we don't propagate collation messages, so it will always be false
+const propagate = false
+
+var ErrNotExpectedOnCollatorSide = errors.New("message is not expected on the collator side of the protocol")
+
+type CollatorProtocolCollatorSide struct {
+	net         Network
+	collatingOn parachaintypes.ParaID
```
axaysagathiya: add nolint:unused
```suggestion
	collatingOn parachaintypes.ParaID //nolint:unused
```

axaysagathiya (approved):

dot/parachain/collator-protocol/collator_side.go
```diff
@@ -0,0 +1,137 @@
...
+	}
+
+	collatorProtocol, ok := msg.(*CollationProtocol)
+	if !ok {
+		return propagate, fmt.Errorf(
+			"failed to cast into collator protocol message, expected: *CollationProtocol, got: %T",
+			msg)
+	}
+
+	collatorProtocolV, err := collatorProtocol.Value()
```
timwu20:
```suggestion
	collatorProtocolVal, err := collatorProtocol.Value()
```

Issue #3515, opened 2023-10-04
feat(parachain/collator): handle network events for collator protocol


kishansagathiya:
## Issue summary

- This is part of collator protocol. 

We have to handle (or react to) certain network events like
- [x] if a peer disconnects, remove it from the list of collators if it is one.
- [x] handle our view change
- [ ] handle some other peer's view change

In polkadot-sdk, any kind of network interaction for collator protocol happens through overseer. We have not taken that approach. We are receiving and handling messages directly without overseer in between.

Implement it, keeping the differences in mind. We will use polkadot-sdk only to confirm the functionality.
 
- This message transfer is between two subsystems on a single node, so we have freedom on its structure.
- Should be done after https://github.com/ChainSafe/gossamer/pull/3512 is merged.

Let me know if you want to take this up.

## Other information and links
Code reference in polkadot sdk
https://github.com/paritytech/polkadot-sdk/blob/f4827dcd16fc5585a8d15c6a32bc27ecde00e7f9/polkadot/node/network/collator-protocol/src/validator_side/mod.rs#L1457

Make sure you follow the polkadot-sdk repo. The code has changed enough from the archived polkadot repo.

kishansagathiya:
Gonna close this 
since some changes for this were done in https://github.com/ChainSafe/gossamer/pull/3829 and https://github.com/ChainSafe/gossamer/pull/3827
One remaining feature of `handle some other peer's view change, I will track as part of https://github.com/ChainSafe/gossamer/issues/3864, all this code will move there.

PR #3827, opened 2024-03-28
Emit network events and handle them in collator protocol

kishansagathiya:
Subsystems needs to know about the peers in collation and validation protocol. This commit
- adds logic to network package for emitting network events when a new peer joins or a peer disconnects.
- introduces a method to get a channel that will give us network events.
- make collator protocol validator side listen to these events and react to it by editing its peer data

## Tests

```sh
```

## Issues
#3515

## Primary Reviewer

@axaysagathiya


dot/network/notifications.go
```diff
@@ -270,6 +270,7 @@ func (s *Service) sendData(peer peer.ID, hs Handshake, info *notificationsProtoc
 		return
 	}
 
+	info.peersData.setMutex(peer)
```
EclesioMeloJunior: The method `sendData` is been called by `broadcastExcluding` which also sets the mutex, I would suggest to set the mutex before calling `sendData`

EclesioMeloJunior (changes requested):

dot/network/notifications.go
timwu20: I don't understand this `setMutex` function.  It has a defer within it unlock the mutex, but it will just get unlocked as soon as this call is done.

dot/network/service.go
```diff
@@ -595,6 +600,24 @@ func (s *Service) GetRequestResponseProtocol(subprotocol string, requestTimeout
 	}
 }
 
+func (s *Service) GetNetworkEventsChannel() <-chan NetworkEventInfo {
```
timwu20: is this intended to be only called once? Sharing a channel doesn't guarantee delivery to all consumers of the channel.

dot/network/notifications.go
EclesioMeloJunior: From what I understand `setMutex` function should be placed when we connect to a peer and only there, setting a mutex we avoid multiple network calls be executed concurrently for the same protocol. So if we have peer A sending us 3 block announces and we still processing the first one then 2 block announces will be locked waiting. However I see that we're using `setMutex` in places where it should not exist, e.g before sending a message, we should not `setMutex` here but use the `info.peersData.getMutex(peer)` and with the mutex then lock the resource, @timwu20 what do you think?

kishansagathiya: From the codebase I see that
- Currently we are using this peer mutex only while sending a handshake. so, we can't send multiple handshakes of the same type to a particular peer simultaneously.
- we don't use this mutex while handling an incoming handshake.
- we don't use this mutex while writing to stream.

I feel like we should should use this mutex on handling incoming handshakes as well, to make sure that we don't try to create multiple connection with the same peer with the same protocol.

It should be safe to not use this mutex while writing to stream (since mutex will stop us from making same type of connections)


about where to put setMutex method, I think it should happen like this https://github.com/ChainSafe/gossamer/blob/972438112f4eddd8d281b358d1b1fae31fd70517/dot/network/service.go#L299-L316

This makes me realize that I should register these two parachain protocols such that  above linked code could run. That would be the ideal solution.

I bet we must be facing this setMutex related problems in grandpa service as well. 
So, if you go through how services start, network service starts first and then grandpa service starts. So this code of running connect handler and disconnect handler never runs for grandpa protocol.

https://github.com/ChainSafe/gossamer/blob/972438112f4eddd8d281b358d1b1fae31fd70517/lib/services/services.go#L43
https://github.com/ChainSafe/gossamer/blob/972438112f4eddd8d281b358d1b1fae31fd70517/lib/grandpa/network.go#L56

We should probably try to register all the network protocols at the same time through the same place. Would make these bugs go away. After we fix this, we would not have to set mutex anywhere else.

kishansagathiya: Let me know if this makes sense to both of you. I will create a separate issue for that.

dot/network/service.go
kishansagathiya: Right now it is being used by only one consumer. I am not aware of anything other code parts that could become a consumer in future.
Instead of passing around the channel directly, I am using a method, since that way I can call it with `Network` interface

dot/network/notifications.go
timwu20: yea I'm not really understanding the requirement to get and set mutexes based on peer id.  Like the `setMutex` function doesn't even check if a mutex is already in the map.  I feel like we should revisit the `peersData` use patterns to identify the correct solution in a new issue.

kishansagathiya: > Like the setMutex function doesn't even check if a mutex is already in the map.

Fixable

>  I feel like we should revisit the peersData use patterns to identify the correct solution in a new issue.

Yes, Sure.

kishansagathiya: > yea I'm not really understanding the requirement to get and set mutexes based on peer id.

I would think that this comment would explain it, but sure let's continue discussion in a separate issue and try to come up with a more appropriate solution.

https://github.com/ChainSafe/gossamer/pull/3827#discussion_r1557576846

kishansagathiya: I have created this issue https://github.com/ChainSafe/gossamer/issues/3890. Let's provide ideas and continue discussion there.

dot/network/service.go
```diff
@@ -595,6 +602,24 @@ func (s *Service) GetRequestResponseProtocol(subprotocol string, requestTimeout
...
```
EclesioMeloJunior: I would suggest to change to be an `Observer`-like pattern, the network emits events and a set of observes are warned about the event.

I would suggest instead of having a single channel that emits you can have a set of registered observers, in this case overseer is one observer, then whenever the `Network` emits an event you need to go over the set of observers sending the event for each one

```go
// the overseer should pass the channel here in order
// to receive an event
// like in the rust where you have a transmissor and a receiver
// let (tx, rx) = mspc::bound_channel(), so you pass the tx
// and keep the rx
func (n *Network) RegisterObserver(ch chan NetworkEvent) {
  n.observers = append(n.observers, ch)
}

...

for _, obs := range service.observers {
    obs <- NetworkEvent
}
```

EclesioMeloJunior (changes requested):

dot/network/service.go
kishansagathiya: It is a good and probably a prettier way to write compared to what I have now. But what's the point when there is going to be only one observer?
Can you think of any other observer?

EclesioMeloJunior: Since you're changing a package I would suggest to not change it thinking just in a single purpose since network is a widely shared package, for example sync package can use network events (connection/disconnection) to refresh the pool of available peers, I know that GRANDPA package also keep a set of peer views and using the network events it can also refresh these data, I think that using observers we can improve the network package to be more flexible and useful for other packages

kishansagathiya: made the changes as asked, in the last commit

EclesioMeloJunior (approved):

axaysagathiya (approved):
LGTM

dot/network/service.go
kishansagathiya: addressed. a consumer can now call a method and get a channel

PR #3828, opened 2024-03-28
chore(dot/node): remove `IsNodeInitialized` method from `nodeBuilder`

EclesioMeloJunior:
## Changes

- Notice that `IsNodeInitialized` can be its own function and thus reducing the amount of methods from the `nodeBuilder` interface

## Tests

```sh
N/A
```

## Issues

N/A

## Primary Reviewer

@dimartiro


jimjbrettj (approved):

dimartiro (approved):
LGTM just need to fix some mocks

PR #3829, opened 2024-03-28
Handle Active Leaves update in collator protocol

kishansagathiya:
Handling active leaves updates are quite important in collator protocol since
- It tells us which relay parents to remove or add
- which candidates to add or remove
- Most importantly it leads us to letting us know as a validator which parachain we are assigned to.

This commit add logic to react to active leaves update by
- updating view
- updating active leaves
- updating our parachain assignments

## Tests

```sh
```

## Issues
linked to issue #3515

## Primary Reviewer

@axaysagathiya


dot/parachain/collator-protocol/validator_side.go
```diff
@@ -115,12 +122,323 @@ func (cpvs CollatorProtocolValidatorSide) handleNetworkEvents(event network.Netw
...
+
+	newLiveHeads := []parachaintypes.ActivatedLeaf{}
+
+	for _, head := range liveHeads {
+		var sliceContains bool
+		for _, deactivated := range signal.Deactivated {
+			if head.Hash == deactivated {
+				sliceContains = true
+				break
+			}
```
EclesioMeloJunior: Use `slice.Contains`
```suggestion
                 if slice.Contains(signal.Deactivated, head.Hash) {
                     newLiveHeads = append(newLiveHeads, head)
                 }
```

dot/parachain/collator-protocol/validator_side.go
```diff
@@ -115,12 +122,323 @@ func (cpvs CollatorProtocolValidatorSide) handleNetworkEvents(event network.Netw
...
+	// Make everythin less messing.
+	// https://github.com/paritytech/polkadot-sdk/blob/1b5f4243d159fbb7cf7067241aca8a37f3dbf7ed/polkadot/node/network/bridge/src/rx/mod.rs#L798
+
+	// this active leaves are handled in bridge in rust, read the code of bridge properly
+
+	// TODO update cpvs.activeLeaves by adding new active leaves and removing deactivated ones
+
+	// majorSyncing means you are 5 blocks behind the tip of the chain and thus more aggressively
+	// download blocks etc to reach the tip of the chain faster.
+	var majorSyncing bool
```
EclesioMeloJunior, line 141: I think we can introduce a method in the sync package that returns the distance to the tip of the chain

dot/parachain/collator-protocol/validator_side.go
```diff
@@ -115,12 +122,323 @@ func (cpvs CollatorProtocolValidatorSide) handleNetworkEvents(event network.Netw
...
+	}
 }
 
-func (cpvs CollatorProtocolValidatorSide) ProcessBlockFinalizedSignal() {
-	// NOTE: nothing to do here
+func (cpvs *CollatorProtocolValidatorSide) updateOurView(liveHeads []parachaintypes.ActivatedLeaf) error {
+	headHashes := []common.Hash{}
+	for _, head := range cpvs.liveHeads {
+		headHashes = append(headHashes, head.Hash)
+	}
```
EclesioMeloJunior, line 173:
```suggestion
  headHashes := make([]common.Hash, len(cpvs.liveHeads))
  copy(headHashes[:], cpvs.liveHeads)
```

dot/parachain/collator-protocol/validator_side.go
```diff
@@ -115,12 +122,323 @@ func (cpvs CollatorProtocolValidatorSide) handleNetworkEvents(event network.Netw
...
+	newlyAdded := []common.Hash{}
+	for _, head := range view.heads {
+		if _, ok := activeLeaves[head]; !ok {
+			newlyAdded = append(newlyAdded, head)
+		}
+	}
+
+	// handled newly added leaves
+	for _, leaf := range newlyAdded {
+		mode := perspectiveParchainMode(leaf)
```
EclesioMeloJunior: I think it is `prospectiveParachainMode`, right?
```suggestion
		mode := prospectiveParachainMode(leaf)
```

dot/parachain/collator-protocol/validator_side.go
```diff
@@ -115,12 +122,323 @@ func (cpvs CollatorProtocolValidatorSide) handleNetworkEvents(event network.Netw
...
+	}
+
+	return nil
+}
+
+func (cpvs *CollatorProtocolValidatorSide) removeOngoing(perRelayParent PerRelayParent) {
+	if perRelayParent.assignment != nil {
+		entry := cpvs.currentAssignments[*perRelayParent.assignment]
+		entry--
+		if entry == 0 {
```
EclesioMeloJunior, line 296: are negative values allowed here?

dot/parachain/collator-protocol/validator_side.go
```diff
@@ -115,12 +122,323 @@ func (cpvs CollatorProtocolValidatorSide) handleNetworkEvents(event network.Netw
...
+
+	coreIndexNow := validatorGroups.GroupRotationInfo.CoreForGroup(groupIndex, uint8(len(availabilityCores.Types)))
+	coreNow, err := availabilityCores.Types[coreIndexNow.Index].Value()
+	if err != nil {
+		return fmt.Errorf("getting core now: %w", err)
+	}
+
+	paraNow := new(parachaintypes.ParaID)
+	switch coreNow.Index() {
+	case 0:
```
EclesioMeloJunior: use type and iota for each core now case

dot/parachain/collator-protocol/validator_side.go
```diff
@@ -115,12 +122,323 @@ func (cpvs CollatorProtocolValidatorSide) handleNetworkEvents(event network.Netw
...
+		if !ok {
+			return fmt.Errorf("couldn't cast VDT value now to scheduledCore")
+		}
+		*paraNow = scheduledCore.ParaID
+	case 2:
+		// Free
+
+	}
+
+	entry := cpvs.currentAssignments[*paraNow]
```
EclesioMeloJunior: in the case `2 (Free)` there is nothing being assigned to `paraNow`, could that be a problem?

dot/parachain/collator-protocol/validator_side.go
```diff
@@ -295,13 +614,61 @@ const (
 	Collating
 )
 
+// The maximum amount of heads a peer is allowed to have in their view at any time.
+// We use the same limit to compute the view sent to peers locally.
+const MAX_VIEW_HEADS uint8 = 5
```
EclesioMeloJunior:
```suggestion
const MaxViewHeads uint8 = 5
```

dot/parachain/collator-protocol/validator_side.go
```diff
@@ -295,13 +614,61 @@ const (
...
+	sort.Sort(SortableHeads(otherHeads))
+
+	return reflect.DeepEqual(localHeads, otherHeads)
+}
+
+func ConstructView(liveHeads map[common.Hash]struct{}, finalizedNumber uint32) View {
+	heads := make([]common.Hash, 0, len(liveHeads))
+	for head := range liveHeads {
+		heads = append(heads, head)
+	}
```
EclesioMeloJunior, line 653:
```suggestion
	heads := make([]common.Hash, len(liveHeads))
	copy(heads[:], liveHeads)
```

dot/parachain/collator-protocol/validator_side.go
```diff
@@ -295,13 +614,61 @@ const (
...
+
+	return View{
+		heads:           heads[:5],
```
EclesioMeloJunior: if heads does not have 5 or more elements it will panic with 
```
panic: runtime error: slice bounds out of range [:5] with capacity 2
```
I would suggest to 

```go
if len(heads) >= 5 {
  heads = heads[:5]
}

return View{
    heads: heads[:5],
    ...
}

```

dot/parachain/collator-protocol/validator_side.go
```diff
@@ -397,6 +778,43 @@ func (f fetchedCollationInfo) String() string {
 		f.relayParent.String(), f.paraID, f.candidateHash.Value.String(), f.collatorID)
 }
 
+func (f *fetchedCollationInfo) FromString(str string) error {
+	splits := strings.Split(str, ",")
+	if len(splits) != 4 {
+		return fmt.Errorf("invalid string format for fetched collation info")
```
EclesioMeloJunior: create a sentinel error and append the wrong string in the error context
```suggestion
		return fmt.Errorf("%w: %s", ErrInvalidStringFormat, str)
```

dot/parachain/collator-protocol/validator_side.go
```diff
@@ -397,6 +778,43 @@ func (f fetchedCollationInfo) String() string {
...
+	}
+	f.relayParent = relayParent
+
+	paraID, err := strconv.ParseUint(strings.TrimSpace(splits[1]), 10, 64)
+	if err != nil {
+		return fmt.Errorf("getting para id: %w", err)
+	}
+	f.paraID = parachaintypes.ParaID(paraID)
+
+	candidateHashBytes := common.MustHexToBytes(strings.TrimSpace(splits[2]))
```
EclesioMeloJunior: I would suggest using `HexToBytes` and return an error instead of `MustHexToBytes` that panics the node, also  in the `relayParent` you're using `HexToHash` instead of `MustHexToHash`

dot/parachain/collator-protocol/validator_side.go
```diff
@@ -397,6 +778,43 @@ func (f fetchedCollationInfo) String() string {
...
+
+	candidateHash := parachaintypes.CandidateHash{
+		Value: common.NewHash(candidateHashBytes),
+	}
+	if err != nil {
+		return fmt.Errorf("getting candidate hash: %w", err)
+	}
```
EclesioMeloJunior: This error check is alone 😢

dot/parachain/collator-protocol/validator_side.go
```diff
@@ -397,6 +778,43 @@ func (f fetchedCollationInfo) String() string {
...
+	f.candidateHash = candidateHash
+
+	var collatorID parachaintypes.CollatorID
+	collatorIDBytes := common.MustHexToBytes(strings.TrimSpace(splits[3]))
```
EclesioMeloJunior: I would suggest using HexToBytes and return an error instead of MustHexToBytes that panics the node

EclesioMeloJunior (changes requested):

dot/parachain/collator-protocol/validator_side.go
kishansagathiya: there is nothing to be done in case of free https://github.com/paritytech/polkadot-sdk/blob/aa68ea58f389c2aa4eefab4bf7bc7b787dd56580/polkadot/node/network/collator-protocol/src/validator_side/mod.rs#L505

dot/parachain/collator-protocol/validator_side.go
kishansagathiya: creating sentinel error. appending the string is just too much information. Anyways it's the same string in the argument.

dot/parachain/collator-protocol/validator_side.go
kishansagathiya: 🤦

dot/parachain/collator-protocol/validator_side.go
kishansagathiya: headHashes  and cpvs.liveHeads are not the same types

dot/parachain/collator-protocol/validator_side.go
kishansagathiya: can't, heads is a slice, liveHeads is a map

dot/parachain/collator-protocol/validator_side.go
EclesioMeloJunior: I see, so in this case it returns None to the `para_now` variable and then on line `523` they check if `para_now` contains a value, if `para_now` is None then they skip the entry update and only updates the `group_assignment`
```
if let Some(para_id) = para_now.as_ref() {}
```

dot/parachain/collator-protocol/validator_side.go
```diff
@@ -115,12 +124,307 @@ func (cpvs CollatorProtocolValidatorSide) handleNetworkEvents(event network.Netw
...
+}
+
+// signingKeyAndIndex finds the first key we can sign with from the given set of validators,
+// if any, and returns it along with the validator index.
+func signingKeyAndIndex(validators []parachaintypes.ValidatorID, ks keystore.Keystore) (*parachaintypes.ValidatorID, parachaintypes.ValidatorIndex) {
+	// TODO: I need a keystore that holds validator keys
+	// checkout here to figure out where do they find keystore from
+	//https://github.com/paritytech/polkadot-sdk/blob/aa68ea58f389c2aa4eefab4bf7bc7b787dd56580/polkadot/node/service/src/overseer.rs#L292
+	// It looks like they are using local keystore from here https://github.com/paritytech/polkadot-sdk/blob/aa68ea58f389c2aa4eefab4bf7bc7b787dd56580/polkadot/node/service/src/lib.rs#L779
+
```
axaysagathiya: It seems you have implemented this.
remove this comment

dot/parachain/collator-protocol/validator_side.go
kishansagathiya: changed

dot/parachain/collator-protocol/validator_side.go
kishansagathiya: no, entry should get removed on 0. edited this as well

dot/parachain/collator-protocol/validator_side.go
kishansagathiya: yes,  but so far this is hard coded. 

haven't implemented the logic of when major syncing is true. would be done with async backing.

dot/parachain/collator-protocol/validator_side.go
```diff
@@ -115,12 +124,306 @@ func (cpvs CollatorProtocolValidatorSide) handleNetworkEvents(event network.Netw
...
+	// this active leaves are handled in bridge in rust, read the code of bridge properly
+
+	// TODO update cpvs.activeLeaves by adding new active leaves and removing deactivated ones
+
+	// TODO: get the value for majorSyncing for syncing package
+	// majorSyncing means you are 5 blocks behind the tip of the chain and thus more aggressively
+	// download blocks etc to reach the tip of the chain faster.
+	var majorSyncing bool
+
+	liveHeads := cpvs.liveHeads[:]
```
timwu20: what are you trying to do here? deep copy the slice?

dot/parachain/collator-protocol/validator_side.go
```diff
@@ -397,6 +767,48 @@ func (f fetchedCollationInfo) String() string {
 		f.relayParent.String(), f.paraID, f.candidateHash.Value.String(), f.collatorID)
 }
 
+func (f *fetchedCollationInfo) FromString(str string) error {
```
timwu20: Kinda makes more sense for this to be a constructor.  Simplifies the calling code too.  Usually the `From` trait isn't doesn't reference `self`.

EclesioMeloJunior (approved):
LGTM

axaysagathiya (approved):

dot/parachain/collator-protocol/validator_side.go
kishansagathiya: looks like a mistake. I could have used cpvs.liveHeads directly. Push the change in latest commit

timwu20 (approved):

Issue #3630, opened 2023-12-05
Implement handling of overseer signals by availability store subsystem


edwardmack:
## Task summary

The availability store subsystem should handle the following signals from the overseer:
- [x] Conclude
- [x] ActiveLeaves
- [x] BlockFinalized

The handling of these messages will address the availability store subsystems functionality by implementing functions:
- [x] processBlockactivated
- [x] processBlockFinalized

## Specification reference

- See polkadot code as reference:  https://github.com/paritytech/polkadot-sdk/blob/15a34838819f09f3bcfb3c05c08fb53934d8efc4/polkadot/node/core/av-store/src/lib.rs#L594

## Other information and links

- See Implementers' guide: https://paritytech.github.io/polkadot-sdk/book/node/utility/availability-store.html#functionality

edwardmack:
Closed with PR #3837

Issue #3561, opened 2023-11-02
Implement availability data pruning in availability store


edwardmack:
## Issue summary

- The Availability Store subsystem keeps available data as well as pruning that data when its stale.  
- Follow guidance outline in the Polkadot Parachain Host Implementers' Guide for information on required functionality.   [https://paritytech.github.io/polkadot-sdk/book/node/utility/availability-store.html](https://paritytech.github.io/polkadot-sdk/book/node/utility/availability-store.html)
- Replicate functionality of Polkadot node, [https://github.com/paritytech/polkadot-sdk/blob/15a34838819f09f3bcfb3c05c08fb53934d8efc4/polkadot/node/core/av-store/src/lib.rs#L1340](https://github.com/paritytech/polkadot-sdk/blob/15a34838819f09f3bcfb3c05c08fb53934d8efc4/polkadot/node/core/av-store/src/lib.rs#L1340)

## Other information and links

-

edwardmack:
Closed with PR #3837.

PR #3837, opened 2024-03-28
feat(parachain/availability-store): Introduce Availability Store Subsystem

### LINKED ISSUE #3741: Create integration tests for availability store subsystem
## Task summary

Create tests to confirm that the availability store subsystem interacts as expected with the overseer.  This will involve creating a test harness that will generate messages and signals sent to the availability store subsystem, then verifying the state of data stored by the subsystem.  Based on the polkadot implementation, the following tests should be created:
- [x] testRuntimeApiErrorDoesNotStopTheSubsystem
- [x] testStoreChunkWorks
- [x] testStoreChunkDoesNothingIfNoEntryAlready
- [x] testQueryChunkChecksMeta
- [x] testStoreAvailableDataErasureMismatch
- [x] testStoreBlockWorks
- [x] testStorePovAndQueryChunkWorks
- [x] testQueryAllChunksWorks
- [x] testStoredButNotIncludedDataIsPruned
- [x] testStoredDataKeptUntilFinalized
- [x] testQueryChunkSizeWorks

And basic scenarios to test as described in the [Polkadot Parachain Host Implementers' Guide](https://paritytech.github.io/polkadot-sdk/book/node/utility/availability-store.html#basic-scenarios-to-test):
- [x] Stored data that is never included pruned in necessary timeout
- [x] Stored data is kept until we are certain it is finalized

## Specification reference

- [Polkadot Parachain Host Implementers' Guide](https://paritytech.github.io/polkadot-sdk/book/node/utility/availability-store.html#basic-scenarios-to-test)

- [polkadot test code](https://github.com/paritytech/polkadot-sdk/blob/15a34838819f09f3bcfb3c05c08fb53934d8efc4/polkadot/node/core/av-store/src/tests.rs#L218)

edwardmack:
## Changes

This PR introduces the base code for the Availability Store Subsystem for storing active leaves data for the parachain subsystems.

See [Host Implementers Guide](https://paritytech.github.io/polkadot-sdk/book/node/utility/availability-store.html) for more information regarding this subsystem.

## Tests

```sh
make test
```

## Issues

closes #3741 and #3561 and #3630
## Primary Reviewer
@kishansagathiya 

@


dot/parachain/availability-store/register.go
```diff
@@ -3,16 +3,44 @@
...
-import "github.com/ChainSafe/gossamer/dot/state"
+import (
+	"context"
 
-func Register(overseerChan chan<- any, st *state.Service) (*AvailabilityStoreSubsystem, error) {
-	availabilityStore := NewAvailabilityStore(st.DB())
+	"github.com/ChainSafe/gossamer/internal/database"
+)
+
+func CreateAndRegister(overseerChan chan<- any, db database.Database) (*AvailabilityStoreSubsystem, error) {
```
kishansagathiya: You could merge `CreateAndRegister` and `CreateAndRegisterPruning`. Accept PruningConfig as a pointer and use default config when PruningConfig  is nil.

Let just call the function `Register` (smaller and same as other places)

dot/parachain/types/statement_test.go
```diff
@@ -9,6 +9,7 @@ import (
 	"math"
 	"testing"
 
+	parachaintypes "github.com/ChainSafe/gossamer/dot/parachain/types"
```
kishansagathiya:
```suggestion
```

dot/parachain/types/testdata/statement_distribution_message.yaml
kishansagathiya, line 1: do we need this empty file?

dot/parachain/chainapi/chainapi.go
kishansagathiya, line 1: It might be nicer to merge this as a separate PR. I have created an issue https://github.com/ChainSafe/gossamer/issues/3840

kishansagathiya (commented):
will continue reviewing

dot/parachain/types/subsystems.go
```diff
@@ -14,6 +14,8 @@ const (
 	CandidateBacking  SubSystemName = "CandidateBacking"
 	CollationProtocol SubSystemName = "CollationProtocol"
 	AvailabilityStore SubSystemName = "AvailabilityStore"
+	ChainAPI          SubSystemName = "ChainAPI"
+	RuntimeAPI        SubSystemName = "RuntimeAPI"
```
kishansagathiya:
```suggestion
```
rust code has a RuntimeAPI subsytem because
- they want to cache responses from heavy runtime calls
- they need an ownerless wrapper (saves from rust's ownership hell)

I have just called runtime functions directly from instance. Caching could be useful, but I feel like we can't prioritize it right now.

dot/parachain/availability-store/messages.go
```diff
@@ -42,7 +41,7 @@ type ErasureChunk struct {
 // QueryChunk query an `ErasureChunk` from the AV store by candidate hash and validator index
 type QueryChunk struct {
 	CandidateHash  parachaintypes.CandidateHash
-	ValidatorIndex uint32
+	ValidatorIndex uint
```
kishansagathiya: why?

kishansagathiya: https://github.com/paritytech/polkadot-sdk/blob/aa68ea58f389c2aa4eefab4bf7bc7b787dd56580/polkadot/node/subsystem-types/src/messages.rs#L505

dot/parachain/availability-store/messages.go
```diff
@@ -79,9 +78,13 @@ type StoreChunk struct {
 // StoreAvailableData computes and checks the erasure root of `AvailableData`
 // before storing its chunks in the AV store.
 type StoreAvailableData struct {
-	CandidateHash       parachaintypes.CandidateHash
-	NumValidators       uint32
-	AvailableData       AvailableData
+	// A hash of the candidate this `ASMStoreAvailableData` belongs to.
+	CandidateHash parachaintypes.CandidateHash
+	// The number of validators in the session.
+	NumValidators uint
```
kishansagathiya: why?
https://github.com/paritytech/polkadot-sdk/blob/aa68ea58f389c2aa4eefab4bf7bc7b787dd56580/polkadot/node/subsystem-types/src/messages.rs#L536

dot/parachain/util/determine_new_blocks.go
```diff
@@ -0,0 +1,292 @@
...
+// /
+// / This is guaranteed to be a subset of the (inclusive) ancestry of `head` determined as all
+// / blocks above the lower bound or above the highest known block, whichever is higher.
+// / This is formatted in descending order by block height.
+// /
+// / An implication of this is that if `head` itself is known or not above the lower bound,
+// / then the returned list will be empty.
+// /
+// / This may be somewhat expensive when first recovering from major sync.
+// pub async fn determine_new_blocks<E, Sender>(
```
kishansagathiya: do we need this commented code?

dot/parachain/types/testdata/statement_distribution_message.yaml
edwardmack: No, I'm not sure where that came from, it's gone now.

dot/parachain/chainapi/chainapi.go
edwardmack: I've removed the ChainAPI stubs I created.

dot/parachain/availability-store/messages.go
edwardmack: Because this variable is used to call `erasure.ObtainChunks` which accepts a `uint`.  I've update these to `uint32`, and casting to `uint` when calling `erasure.ObtainChunks`

dot/parachain/util/determine_new_blocks.go
edwardmack: no. removed.

dot/parachain/availability-store/register.go
edwardmack: Good idea, I've updated this.

dot/parachain/availability-store/messages.go
edwardmack: Because this variable is used to call `erasure.ObtainChunks` which accepts a `uint`. I've update these to `uint32`, and casting to `uint` when calling `erasure.ObtainChunks`

dot/parachain/availability-store/availability_store.go
```diff
@@ -79,12 +85,38 @@ type AvailabilityStoreSubsystem struct {
...
+	return ok
+}
+
+func (kud *KnownUnfinalizedBlock) insert(hash common.Hash, number parachaintypes.BlockNumber) {
+	if kud.byHash == nil {
+		kud.byHash = make(map[common.Hash]struct{})
+	}
+	if kud.byNumber == nil {
+		kud.byNumber = make(map[BlockEntry]struct{})
+	}
```
timwu20: I think this code should be moved to a constructor for the `KnownUnfinalizedBlock` type which can be called in a constructor to construct `AvailabilityStoreSubsystem`.

dot/parachain/availability-store/availability_store.go
```diff
@@ -156,6 +188,91 @@ func (asb *availabilityStoreBatch) reset() {
 	asb.pruneByTime.Reset()
 }
 
+var (
+	TestChunk1 = ErasureChunk{
```
timwu20: These test fixtures shouldn't be here.  Move to a separate package if you want to share these fixtures between tests in different packages.

dot/parachain/availability-store/availability_store.go
```diff
@@ -275,7 +392,7 @@ func (as *availabilityStore) storeAvailableData(subsystem *AvailabilityStoreSubs
 	meta = &CandidateMeta{}
```
timwu20: why does this need to be a pointer?

dot/parachain/availability-store/availability_store.go
```diff
@@ -456,32 +566,281 @@ func (av *AvailabilityStoreSubsystem) processMessages() {
...
+				av.currentMessage = msg
 				av.ProcessBlockFinalizedSignal()
 
 			default:
-				logger.Error(parachaintypes.ErrUnknownOverseerMessage.Error())
+				// TODO(ed): handle unknown message type (if necessary)
+				if msg != nil {
+					logger.Infof("unknown message type %T", msg)
+					logger.Error(parachaintypes.ErrUnknownOverseerMessage.Error())
+				}
```
timwu20: maybe panic in this case until we can gracefully shut down?

dot/parachain/availability-store/availability_store.go
```diff
@@ -79,12 +85,38 @@ type AvailabilityStoreSubsystem struct {
 	SubSystemToOverseer chan<- any
 	OverseerToSubSystem <-chan any
 	availabilityStore   availabilityStore
-	pruningConfig       pruningConfig
+	currentMessage      any
+	finalizedNumber     parachaintypes.BlockNumber
+	knownBlocks         KnownUnfinalizedBlock
+	pruningConfig       PruningConfig
 	clock               subsystemClock
```
timwu20: not sure what the point of this attribute is given it's not stateful.  Couldn't you just reference a helper function to give you the `BETimestamp` of `time.Now()`?

dot/parachain/util/determine_new_blocks.go
```diff
@@ -0,0 +1,161 @@
+// Copyright 2024 ChainSafe Systems (ON)
+// SPDX-License-Identifier: LGPL-3.0-only
+
+package util
```
timwu20: maybe just call this file util.go?

dot/parachain/util/determine_new_blocks.go
```diff
@@ -0,0 +1,161 @@
...
+
+	ctx, cancel := context.WithTimeout(context.Background(), timeout)
+	defer cancel()
+
+	select {
+	case response := <-responseChan:
+		return response, nil
+	case <-ctx.Done():
+		return nil, ctx.Err()
+	}
```
timwu20: We shouldn't be using contexts just to implement timeouts.  Here's [an example](https://github.com/timwu20/service-example/blob/main/service/fan_out.go#L108) of just using channels.

dot/parachain/util/determine_new_blocks.go
```diff
@@ -0,0 +1,161 @@
...
+	// Send with timeout
+	ctx, cancel := context.WithTimeout(context.Background(), timeout)
+	defer cancel()
+
+	select {
+	case channel <- message:
+		return nil
+	case <-ctx.Done():
+		return ctx.Err()
+	}
```
timwu20: Same here.

dot/parachain/availability-store/availability_store.go
```diff
@@ -456,32 +566,281 @@ func (av *AvailabilityStoreSubsystem) processMessages() {
...
+	respChan := make(chan any)
+	message := util.ChainAPIMessage[util.BlockHeader]{
+		Message: util.BlockHeader{
+			Hash: activeLeave.Activated.Hash,
+		},
+		ResponseChannel: respChan,
+	}
+	response, err := util.Call(av.SubSystemToOverseer, message, message.ResponseChannel)
+	if err != nil {
+		logger.Errorf("sending message to get block header: %w", err)
```
timwu20: is there a reason why `ProcessActiveLeavesUpdateSignal` doesn't return an error? I don't think we should get in the practice of just logging errors that should actually be handled.

dot/parachain/availability-store/availability_store.go
```diff
@@ -456,32 +566,281 @@ func (av *AvailabilityStoreSubsystem) processMessages() {
...
+
+		av.processNewHead(tx, v.Hash, now, v.Header)
+
+		// add to known blocks
+		av.knownBlocks.insert(v.Hash, parachaintypes.BlockNumber(v.Header.Number))
+
+		// end db batch
+		err := tx.flush()
+		if err != nil {
+			logger.Errorf("failed to flush tx: %w", err)
```
timwu20: same here.  Should we handle this error?

dot/parachain/availability-store/availability_store.go
```diff
@@ -601,12 +960,214 @@ func (av *AvailabilityStoreSubsystem) handleStoreAvailableData(msg StoreAvailabl
...
+					err = meta.State.Set(Unfinalized{Timestamp: val.Timestamp, BlockNumberHash: val.BlockNumberHash})
+					if err != nil {
+						logger.Errorf("failed to set state to unfinalized: %w", err)
+					}
+				}
+
+			}
+
+			// write meta
+			// TODO(ed): write meta
```
timwu20: Do these TODOs need to be addressed?

dot/parachain/availability-store/availability_store.go
edwardmack: Good idea, I've moved this to AvailaibilyStoreSubsystem constructor.

dot/parachain/availability-store/availability_store.go
edwardmack: These have been moved to test availability store package.

dot/parachain/availability-store/availability_store.go
edwardmack: No reason, I've updated this.

dot/parachain/availability-store/availability_store.go
edwardmack: Good point, I've updated this.

dot/parachain/util/determine_new_blocks.go
edwardmack: updated.

dot/parachain/util/determine_new_blocks.go
edwardmack: updated.

dot/parachain/util/determine_new_blocks.go
edwardmack: updated.

dot/parachain/availability-store/availability_store.go
edwardmack: Added panic.

dot/parachain/availability-store/availability_store.go
edwardmack: Updated this to return error.

dot/parachain/availability-store/availability_store.go
edwardmack: Updated to return error.

dot/parachain/availability-store/availability_store.go
```diff
@@ -79,13 +80,52 @@ type AvailabilityStoreSubsystem struct {
...
+	pruningConfig       PruningConfig
 	//TODO: metrics       Metrics
 }
 
+// NewAvailabilityStoreSubsystem creates a new instance of AvailabilityStoreSubsystem
+func NewAvailabilityStoreSubsystem(db database.Database) *AvailabilityStoreSubsystem {
+	availabilityStore := NewAvailabilityStore(db)
+
+	ctx := context.Background()
+	ctx, cancel := context.WithCancel(ctx)
```
timwu20, line 94: I know this doesn't relate to to this PR and I should have mentioned it in the skeleton PR, but I feel pretty strongly that we should shouldn't be adding contexts as attributes of subsystems.  Read this [blog post ](https://dave.cheney.net/2017/08/20/context-isnt-for-cancellation) from Dave Cheney.  I think we should implement the service pattern to explicitly stop subsystems.  I can address this in a PR to the parachain branch.

dot/parachain/availability-store/messages.go
```diff
@@ -6,15 +6,14 @@ package availabilitystore
...
-	"time"
 
 	parachaintypes "github.com/ChainSafe/gossamer/dot/parachain/types"
 	"github.com/ChainSafe/gossamer/lib/common"
 	"github.com/ChainSafe/gossamer/lib/trie"
 	"github.com/ChainSafe/gossamer/pkg/scale"
 )
 
-var errInvalidErasureRoot = errors.New("Invalid erasure root")
+var ErrInvalidErasureRoot = errors.New("Invalid erasure root")
```
timwu20: does this need to be public? I don't see any references to it outside of this package.

dot/parachain/availability-store/availability_store.go
```diff
@@ -456,32 +479,250 @@ func (av *AvailabilityStoreSubsystem) processMessages() {
...
+			return fmt.Errorf("failed to flush tx: %w", err)
+		}
+	}
+	return nil
+}
+
+func (av *AvailabilityStoreSubsystem) processNewHead(tx *availabilityStoreBatch, hash common.Hash, now BETimestamp,
+	header types.Header) error {
+	logger.Infof("processNewHead hash %s, now %v, header %v\n", hash, now, header)
+	// TODO: call requestValidators to determine number of validators
```
timwu20: Is this done now?  If not should we create an issue?

dot/parachain/availability-store/availability_store.go
```diff
@@ -456,32 +479,250 @@ func (av *AvailabilityStoreSubsystem) processMessages() {
...
+		})
+		if err != nil {
+			return fmt.Errorf("failed to set state to unfinalized: %w", err)
+		}
+	case Finalized:
+		// This should never happen as a candidate would have to be included after
+		// finality.
+	}
+
+	//write unfinalized block contains
```
timwu20:
```suggestion
	// write unfinalized block contains
```

dot/parachain/availability-store/availability_store_test.go
```diff
@@ -788,3 +803,952 @@ func TestAvailabilityStore_storeAvailableData(t *testing.T) {
...
+	})
+
+	err := harness.overseer.Start()
+	require.NoError(t, err)
+
+	go harness.processMessages()
+
+	harness.triggerBroadcast()
+
+	time.Sleep(1000 * time.Millisecond)
```
timwu20: does it really have to be this long?

dot/parachain/overseer/overseer.go
```diff
@@ -97,16 +111,17 @@ func (o *Overseer) Start() error {
 	go o.processMessages()
 	go o.handleBlockEvents()
 
+	// TODO: add logic to start listening for Block Imported events and Finalisation events
```
timwu20: is there an issue for this?

dot/parachain/util/util.go
```diff
@@ -0,0 +1,147 @@
...
+	lastHeader := ancestry[len(ancestry)-1].Header
+	// This is always non-zero as determined by the loop invariant above.
+	ancestryStep := min(4, (lastHeader.Number - minBlockNeeded))
+
+	ancestors, err := GetBlockAncestors(subsystemToOverseer, head, uint32(ancestryStep))
+	if err != nil {
+		return nil, fmt.Errorf("getting block ancestors: %w", err)
+	}
+	fmt.Printf("ancestors: %v\n", ancestors)
+	// TODO(ed): finish this
```
timwu20: what else needs to be done here?

dot/parachain/availability-store/availability_store.go
```diff
@@ -79,13 +80,52 @@ type AvailabilityStoreSubsystem struct {
...
 	OverseerToSubSystem <-chan any
 	availabilityStore   availabilityStore
-	pruningConfig       pruningConfig
-	clock               subsystemClock
-	//TODO: pruningConfig PruningConfig
-	//TODO: clock         Clock
+	finalizedNumber     parachaintypes.BlockNumber
+	knownBlocks         KnownUnfinalizedBlock
+	pruningConfig       PruningConfig
 	//TODO: metrics       Metrics
```
axaysagathiya: Do we need this metrix field?

dot/parachain/types/types.go
```diff
@@ -654,3 +655,13 @@ type UncheckedSignedAvailabilityBitfield struct {
 	// The signature by the validator of the signed payload.
 	Signature ValidatorSignature `scale:"3"`
 }
+
+// Subsystem is an interface for subsystems to be registered with the overseer.
+type Subsystem interface {
```
kishansagathiya, line 660: There are two Subsystem interfaces, one here and one in https://github.com/ChainSafe/gossamer/blob/0dcd13fc6c7d6e26c8fbd5a3a7c238aefe13d139/dot/parachain/overseer/types.go#L13

Let's use just one of these two. Let try to use overseer.Subsystem, if possible

dot/parachain/util/util.go
```diff
@@ -25,3 +30,137 @@ func SigningKeyAndIndex(
 	}
 	return nil, 0
 }
+
```
kishansagathiya, line 50: let's create a package called chainapi and put all the chain api related code there.

dot/parachain/availability-store/availability_store.go
edwardmack: Good point, yes we should implement the service pattern.

dot/parachain/availability-store/messages.go
edwardmack: Updated.

dot/parachain/availability-store/availability_store.go
edwardmack: created issue.

dot/parachain/availability-store/availability_store_test.go
edwardmack: no, reduced to 50 milliseconds

dot/parachain/overseer/overseer.go
edwardmack: I removed this comment, it's being handled in handleBlockEvents rountine.

dot/parachain/availability-store/availability_store.go
```diff
@@ -79,13 +80,52 @@ type AvailabilityStoreSubsystem struct {
 	SubSystemToOverseer chan<- any
 	OverseerToSubSystem <-chan any
 	availabilityStore   availabilityStore
-	pruningConfig       pruningConfig
-	clock               subsystemClock
-	//TODO: pruningConfig PruningConfig
-	//TODO: clock         Clock
+	finalizedNumber     parachaintypes.BlockNumber
```
axaysagathiya: Can you rename from `finalizedNumber` to `finalizedBlockNumber`?

dot/parachain/availability-store/availability_store.go
```diff
@@ -79,13 +80,52 @@ type AvailabilityStoreSubsystem struct {
...
+	knownBlocks         KnownUnfinalizedBlock
```
axaysagathiya: rename from `knownBlocks` to `knownUnfinalizedBlocks`

dot/parachain/availability-store/availability_store.go
```diff
@@ -79,13 +80,52 @@ type AvailabilityStoreSubsystem struct {
...
+		knownBlocks:       *NewKnownUnfinalizedBlock(),
+	}
+
+	return availabilityStoreSubsystem
+}
+
+type KnownUnfinalizedBlock struct {
+	byHash   map[common.Hash]struct{}
+	byNumber map[BlockEntry]struct{}
+}
```
axaysagathiya: - `KnownUnfinalizedBlocks` struct name would make more sense as it stores multiple records.
- also do you need this type exported?

dot/parachain/util/util.go
edwardmack: I've created an issue for this.

dot/parachain/availability-store/availability_store.go
```diff
@@ -79,13 +80,52 @@ type AvailabilityStoreSubsystem struct {
...
+
+func NewKnownUnfinalizedBlock() *KnownUnfinalizedBlock {
```
axaysagathiya: Do you need this exported?

dot/parachain/util/util.go
```diff
@@ -25,3 +30,137 @@ func SigningKeyAndIndex(
...
+}
+
+type AncestorsResponse struct {
+	Ancestors []common.Hash
+	Error     error
+}
+
+type Ancestors struct {
+	Hash common.Hash
+	K    uint32
```
axaysagathiya: What is K? Can you give a meaningful name to this field?

dot/parachain/util/util.go
```diff
@@ -25,3 +30,137 @@ func SigningKeyAndIndex(
...
+}
+
+type BlockHeader struct {
+	Hash common.Hash
+}
+
+// DetermineNewBlocks determines the hashes of all new blocks we should track metadata for, given this head.
+// Given a new chain-head hash, this determines the hashes of all new blocks we should track
+// metadata for, given this head.
```
axaysagathiya: It seems to duplicate comments. Can you rewrite this?

dot/parachain/util/util.go
```diff
@@ -25,3 +30,137 @@ func SigningKeyAndIndex(
...
+// This may be somewhat expensive when first recovering from major sync.
+func DetermineNewBlocks(subsystemToOverseer chan<- any, isKnown func(hash common.Hash) bool, head common.Hash,
+	header types.Header,
+	lowerBoundNumber parachaintypes.BlockNumber) ([]HashHeader, error) {
+
+	minBlockNeeded := uint(lowerBoundNumber + 1)
+
+	// Early exit if the block is in the DB or too early.
+	alreadyKnown := isKnown(head)
+
```
axaysagathiya:
```suggestion

```

dot/parachain/util/util.go
```diff
@@ -25,3 +30,137 @@ func SigningKeyAndIndex(
...
+	beforeRelevant := header.Number < minBlockNeeded
+	if alreadyKnown || beforeRelevant {
+		return make([]HashHeader, 0), nil
```
axaysagathiya: Can you return nil, nil?

dot/parachain/util/util.go
```diff
@@ -25,3 +30,137 @@ func SigningKeyAndIndex(
...
+	ancestry = append(ancestry, HashHeader{Hash: head, Header: *headerClone})
+
+	// Early exit if the parent hash is in the DB or no further blocks are needed.
+	if isKnown(header.ParentHash) || header.Number == minBlockNeeded {
+		return ancestry, nil
+	}
+
+	lastHeader := ancestry[len(ancestry)-1].Header
+	// This is always non-zero as determined by the loop invariant above.
+	ancestryStep := min(4, (lastHeader.Number - minBlockNeeded))
```
axaysagathiya: What is ancestryStep.
also, what is 4. can you create a const for it?

dot/parachain/util/util.go
```diff
@@ -25,3 +30,137 @@ func SigningKeyAndIndex(
...
+	}
+
+	return response.Ancestors, nil
+}
+
+// Call sends the given message to the given channel and waits for a response with a timeout
+func Call(channel chan<- any, message any, responseChan chan any) (any, error) {
+	if err := SendMessage(channel, message); err != nil {
+		return nil, fmt.Errorf("send message: %w", err)
+	}
```
axaysagathiya: I don't think we need a function for sending message to the channel.
Also I don't think we need time out while sending message to the channel.
```suggestion
    channel <- message
```

dot/parachain/util/util.go
```diff
@@ -25,3 +30,137 @@ func SigningKeyAndIndex(
...
+
+	select {
+	case response := <-responseChan:
+		return response, nil
+	case <-time.After(timeout):
+		return nil, fmt.Errorf("timeout waiting for response")
```
axaysagathiya: use existing timeout variable
```suggestion
	case <-time.After(parachaintypes.SubsystemRequestTimeout):
		return nil, parachaintypes.ErrSubsystemRequestTimeout
```

dot/parachain/util/util.go
```diff
@@ -25,3 +30,137 @@ func SigningKeyAndIndex(
...
+func SendMessage(channel chan<- any, message any) error {
+	// Send with timeout
+
+	select {
+	case channel <- message:
+		return nil
+	case <-time.After(timeout):
+		return fmt.Errorf("timeout sending message")
+	}
+}
```
axaysagathiya: I don't think we need a timeout when sending a message to the overseers, so we don't need this function.

dot/parachain/types/types.go
edwardmack: Yes, there should only be one.  I created the one in `types` so that it's outside the `overseer` package since a subsystem can be implemented without the awareness of the overseer.  With it in types we can create a test overseer in the subsystem tests without creating a cycle import. 

So I'd like to just use  `types.Subsystem`

dot/parachain/availability-store/availability_store.go
edwardmack: no, not until we add metrics.  I've removed it.

dot/parachain/availability-store/availability_store.go
edwardmack: Yes, good idea, updated.

dot/parachain/availability-store/availability_store.go
edwardmack: updated.

dot/parachain/availability-store/availability_store.go
edwardmack: good points, I've updated these.

dot/parachain/availability-store/availability_store.go
edwardmack: no, updated.

dot/parachain/util/util.go
edwardmack: updated.

dot/parachain/util/util.go
edwardmack: i've updated, it's number of ancestors.

dot/parachain/util/util.go
edwardmack: updated.

dot/parachain/util/util.go
edwardmack: yes, I've updated this.

dot/parachain/util/util.go
edwardmack: I've updated this to numberOfAncestors.  4 is the number used in the polkadot code:  https://github.com/paritytech/polkadot-sdk/blob/aa68ea58f389c2aa4eefab4bf7bc7b787dd56580/polkadot/node/subsystem-util/src/determine_new_blocks.rs#L83

dot/parachain/util/util.go
edwardmack: removed SendMessage function.

dot/parachain/util/util.go
edwardmack: updated.

dot/parachain/util/util.go
edwardmack: removed this function.

dot/parachain/service.go
```diff
@@ -29,7 +29,7 @@ const (
 
 type Service struct {
 	Network  Network
-	overseer *overseer.Overseer
+	overseer overseer.OverseerSystem
```
axaysagathiya: Why?

dot/parachain/testdata/collation_protocol.yaml
axaysagathiya, line 1: Can you delete this empty file?

dot/parachain/availability-store/availability_store.go
```diff
@@ -459,32 +481,247 @@ func (av *AvailabilityStoreSubsystem) processMessages() {
 				if err != nil {
 					logger.Errorf("failed to process active leaves update signal: %w", err)
 				}
+
```
axaysagathiya:
```suggestion
```

dot/parachain/availability-store/availability_store.go
```diff
@@ -324,25 +357,21 @@ func (as *availabilityStore) storeAvailableData(subsystem *AvailabilityStoreSubs
...
+	for i := range candidateMeta.ChunksStored {
+		candidateMeta.ChunksStored[i] = true
 	}
 
-	dataBytes, err := scale.Marshal(meta)
-	if err != nil {
-		return false, fmt.Errorf("marshalling meta for candidate: %w", err)
-	}
-	err = batch.meta.Put(candidate.Value[:], dataBytes)
+	err = writeMeta(batch.meta, candidate, &candidateMeta)
```
axaysagathiya: Can you rename with some descriptive function name. May be `writeCandidateMetaToDB`

dot/parachain/availability-store/availability_store.go
```diff
@@ -459,32 +481,247 @@ func (av *AvailabilityStoreSubsystem) processMessages() {
...
+					logger.Errorf("failed to process block finalized signal: %w", err)
+				}
 
 			default:
-				logger.Error(parachaintypes.ErrUnknownOverseerMessage.Error())
+				if msg != nil {
+					logger.Infof("unknown message type %T", msg)
+					logger.Error(parachaintypes.ErrUnknownOverseerMessage.Error())
+					// this error shouldn't happen, so we'll panic to catch it during development
+					panic(parachaintypes.ErrUnknownOverseerMessage.Error())
```
axaysagathiya: Info, error log, and panic will print the same thing. Panic is enough. 

```suggestion
panic(fmt.Sprintf("%s: %T", parachaintypes.ErrUnknownOverseerMessage.Error(), msg))
```

dot/parachain/availability-store/availability_store.go
```diff
@@ -459,32 +481,247 @@ func (av *AvailabilityStoreSubsystem) processMessages() {
...
 }
 
-func (av *AvailabilityStoreSubsystem) ProcessActiveLeavesUpdateSignal(
-	update parachaintypes.ActiveLeavesUpdateSignal) error {
-	// TODO: #3630
+func (av *AvailabilityStoreSubsystem) processNewHead(tx *availabilityStoreBatch, hash common.Hash, now BETimestamp,
+	header types.Header) error {
+	logger.Infof("processNewHead hash %s, now %v, header %v\n", hash, now, header)
+	// TODO: call requestValidators to determine number of validators, see issue #3932
+	nValidators := uint(10)
```
axaysagathiya: Make an empty function in the util file with the TODO comment. And call that function here.
So We don't need to find all the places where this function should be called in the future.

dot/parachain/chainapi/chainapi.go
```diff
@@ -0,0 +1,130 @@
...
+
+// Ancestors is a message to get the ancestors of a block.
+type Ancestors struct {
+	Hash              common.Hash
+	numberOfAncestors uint32
+}
+
+type BlockHeader struct {
+	Hash common.Hash
+}
```
axaysagathiya:
```suggestion
type BlockHeader common.Hash
```

dot/parachain/chainapi/chainapi.go
```diff
@@ -0,0 +1,130 @@
...
+import (
+	"fmt"
+	"time"
+
+	parachaintypes "github.com/ChainSafe/gossamer/dot/parachain/types"
+	"github.com/ChainSafe/gossamer/dot/types"
+	"github.com/ChainSafe/gossamer/lib/common"
+)
+
+type HashHeader struct {
```
axaysagathiya: Do you need this exported?

dot/parachain/service.go
edwardmack: I thought it made sense to use an interface so that we could re-use that interface for mocking and testing.

dot/parachain/testdata/collation_protocol.yaml
edwardmack: yes.

dot/parachain/availability-store/availability_store.go
edwardmack: yes

dot/parachain/availability-store/availability_store.go
edwardmack: updated.

dot/parachain/chainapi/chainapi.go
edwardmack: Updated.

dot/parachain/availability-store/availability_store.go
edwardmack: good idea.

dot/parachain/chainapi/chainapi.go
edwardmack: no, I've updated this.

dot/parachain/chainapi/chainapi.go
```diff
@@ -0,0 +1,133 @@
...
+// Call sends the given message to the given channel and waits for a response with a timeout
+func Call(channel chan<- any, message any, responseChan chan any) (any, error) {
+	channel <- message
+	select {
+	case response := <-responseChan:
+		return response, nil
+	case <-time.After(timeout):
+		return nil, parachaintypes.ErrSubsystemRequestTimeout
+	}
+}
```
axaysagathiya: Great!
- We can use this in other packages as well. Please move this function to `util/util.go`.
- Please use descriptive function names. Maybe rename it to `sendOverseerMessage`?
- Can you use `parachaintypes.SubsystemRequestTimeout` instead of creating a new timeout variable?

dot/parachain/chainapi/chainapi.go
```diff
@@ -0,0 +1,133 @@
...
+//
+// This is guaranteed to be a subset of the (inclusive) ancestry of `head` determined as all
+// blocks above the lower bound or above the highest known block, whichever is higher.
+// This is formatted in descending order by block height.
+//
+// An implication of this is that if `head` itself is known or not above the lower bound,
+// then the returned list will be empty.
+//
+// This may be somewhat expensive when first recovering from major sync.
+func DetermineNewBlocks(subsystemToOverseer chan<- any, isKnown func(hash common.Hash) bool, head common.Hash,
```
axaysagathiya: - This function doesn't belong to the chain-api subsystem. Should move to util/util.go
image

- add a NOTE with the function comment saying this is an incomplete function and we have an issue implementing this. I know you have a comment inside the function, but still, having a NOTE would be better.

dot/parachain/chainapi/chainapi.go
```diff
@@ -0,0 +1,133 @@
...
+
+	parachaintypes "github.com/ChainSafe/gossamer/dot/parachain/types"
+	"github.com/ChainSafe/gossamer/dot/types"
+	"github.com/ChainSafe/gossamer/lib/common"
+)
+
+type hashHeader struct {
+	Hash   common.Hash
+	Header types.Header
+}
```
axaysagathiya: If I am not mistaken, you only use this struct in the `DetermineNewBlocks` function. So this struct should also be moved to `util/util.go` with `DetermineNewBlocks`.

dot/parachain/chainapi/chainapi.go
```diff
@@ -0,0 +1,133 @@
...
+	}
+
+	ancestry = append(ancestry, hashHeader{Hash: head, Header: *headerClone})
+
+	// Early exit if the parent hash is in the DB or no further blocks are needed.
+	if isKnown(header.ParentHash) || header.Number == minBlockNeeded {
+		return ancestry, nil
+	}
+
+	lastHeader := ancestry[len(ancestry)-1].Header
```
axaysagathiya: This can cause panic if the ancestry has only one entry.
(optional as this function is incomplete and you have an issue with the same; we can fix this later)


```suggestion
	if len(ancestry) == 1 {
		return nil, fmt.Errorf("ancestry has length 1 at initialization and is only added to.")
	}
	
	lastHeader := ancestry[len(ancestry)-1].Header
```

dot/parachain/chainapi/chainapi.go
```diff
@@ -0,0 +1,133 @@
...
+	message := ChainAPIMessage[Ancestors]{
+		Message: Ancestors{
+			Hash:              head,
+			numberOfAncestors: numAncestors,
+		},
+		ResponseChannel: respChan,
+	}
+	res, err := Call(overseerChannel, message, message.ResponseChannel)
+	if err != nil {
+		return nil, fmt.Errorf("sending message to get block ancestors: %w", err)
```
axaysagathiya: when you call this function, you are wrapping an error with the `"getting block ancestors"` message. There is no need to wrap errors return from the function again. You can return the error without wrapping it.
```suggestion
		return nil, err
```

dot/parachain/chainapi/chainapi.go
```diff
@@ -0,0 +1,133 @@
...
+	}
+
+	response, ok := res.(AncestorsResponse)
+	if !ok {
+		return nil, fmt.Errorf("getting block ancestors: got unexpected response type %T", res)
```
axaysagathiya:
```suggestion
		return nil, fmt.Errorf("got unexpected response type %T", res)
```

dot/parachain/chainapi/chainapi.go
```diff
@@ -0,0 +1,133 @@
...
+	}
+	if response.Error != nil {
+		return nil, fmt.Errorf("getting block ancestors: %w", response.Error)
```
axaysagathiya:
```suggestion
		return nil, response.Error
```

dot/parachain/chainapi/chainapi.go
edwardmack: Good ideas, updated.

dot/parachain/chainapi/chainapi.go
edwardmack: I added the note, and moved this to util/util.go

dot/parachain/chainapi/chainapi.go
edwardmack: updated.

dot/parachain/chainapi/chainapi.go
edwardmack: updated.

dot/parachain/chainapi/chainapi.go
edwardmack: updated.

dot/parachain/chainapi/chainapi.go
edwardmack: updated.

dot/parachain/chainapi/chainapi.go
edwardmack: updated.

dot/parachain/service.go
```diff
@@ -42,7 +42,7 @@ func NewService(net Network, forkID string, st *state.Service, ks keystore.Keyst
 	}
 	genesisHash := st.Block.GenesisHash()
 
-	availabilityStore, err := availability_store.Register(overseer.SubsystemsToOverseer, st)
+	availabilityStore, err := availability_store.Register(overseer.GetSubsystemToOverseerChannel(), st.DB(), nil)
```
timwu20, line 45: Seems odd to me that that this function assigns the `OverseerToSubsystem` channel on line 49.  I would've assumed that the `Register` function accepts this channel.

dot/parachain/service.go
timwu20: why not use the `overseer.Overseer` as the interface? You don't even have to export the type that implements it.

dot/parachain/availability-store/availability_store.go
```diff
@@ -76,14 +78,52 @@ type AvailabilityStoreSubsystem struct {
...
-	SubSystemToOverseer chan<- any
-	OverseerToSubSystem <-chan any
-	availabilityStore   availabilityStore
-	pruningConfig       pruningConfig
-	clock               subsystemClock
-	//TODO: pruningConfig PruningConfig
-	//TODO: clock         Clock
-	//TODO: metrics       Metrics
+	SubSystemToOverseer    chan<- any
+	OverseerToSubSystem    <-chan any
```
timwu20: I don't think these should be public.  They should be initialized in the constructor or explicitly set in some sort of `Register` function.

dot/parachain/availability-store/availability_store.go
```diff
@@ -272,31 +308,29 @@ func (as *availabilityStore) storeAvailableData(subsystem *AvailabilityStoreSubs
 		return true, nil // already stored
 	}
 
-	meta = &CandidateMeta{}
+	candidateMeta := CandidateMeta{}
```
timwu20: nit:
```suggestion
	var candidateMeta CandidateMeta
```

dot/parachain/availability-store/availability_store.go
```diff
@@ -272,31 +308,29 @@ func (as *availabilityStore) storeAvailableData(subsystem *AvailabilityStoreSubs
...
+	now := timeNow()
+	pruneAt := now + BETimestamp(subsystem.pruningConfig.KeepUnavailableFor.Seconds())
+	err = subsystem.writePruningKey(batch.pruneByTime, pruneAt, candidate)
 	if err != nil {
 		return false, fmt.Errorf("writing pruning key: %w", err)
 	}
 
-	meta.State = NewStateVDT()
-	err = meta.State.Set(Unavailable{Timestamp: now})
+	candidateMeta.State = NewStateVDT()
```
timwu20: Can you go through the messages file and ensure we're not exporting anything that's not needed.  I saw that `NewStateVDT` isn't called anywhere outside of this package.

dot/parachain/availability-store/availability_store.go
```diff
@@ -460,31 +483,241 @@ func (av *AvailabilityStoreSubsystem) processMessages() {
...
+		err = meta.State.Set(Unfinalized{
+			Timestamp:  val.Timestamp,
+			BlockEntry: append(val.BlockEntry, beBlock),
+		})
+		if err != nil {
+			return fmt.Errorf("failed to set state to unfinalized: %w", err)
+		}
+	case Finalized:
+		// This should never happen as a candidate would have to be included after
+		// finality.
```
timwu20, line 684: should we log or panic?

dot/parachain/availability-store/availability_store.go
```diff
@@ -604,12 +836,275 @@ func (av *AvailabilityStoreSubsystem) handleStoreAvailableData(msg StoreAvailabl
...
+			}
+
+			// write meta
+			err = writeCandidateMetaToDB(tx.meta, candidateHash, meta)
+			if err != nil {
+				logger.Errorf("failed to put meta: %w", err)
+			}
+		}
+	}
+}
```
timwu20, line 1111: There are a lot of errors not being handled here.  I understand that we want to continue processing candidates even if there are some errors, but are there no errors that should eventually cause this subsystem to shut down? I would assume if we're unable to write to the DB that could be a systemic error.  Just returning the error and panicking/logging on the caller side for now will make things easier when we implement graceful shutdown.

dot/parachain/overseer/overseer.go
```diff
@@ -60,14 +71,18 @@ func NewOverseer(blockState BlockState) *Overseer {
...
 		activeLeaves:         make(map[common.Hash]uint32),
 		SubsystemsToOverseer: make(chan any),
-		subsystems:           make(map[Subsystem]chan any),
-		nameToSubsystem:      make(map[parachaintypes.SubSystemName]Subsystem),
+		subsystems:           make(map[parachaintypes.Subsystem]chan any),
+		nameToSubsystem:      make(map[parachaintypes.SubSystemName]parachaintypes.Subsystem),
 	}
 }
 
+func (o *Overseer) GetSubsystemToOverseerChannel() chan any {
```
timwu20: I will change this when I propose the service architecture I've mentioned before.  The producer of the channel will close this channel to notify the consumer (overseer in this case) that the service has halted.

dot/parachain/util/util.go
```diff
@@ -25,3 +47,99 @@ func SigningKeyAndIndex(
...
+// This is guaranteed to be a subset of the (inclusive) ancestry of `head` determined as all
+// blocks above the lower bound or above the highest known block, whichever is higher.
+// This is formatted in descending order by block height.
+//
+// An implication of this is that if `head` itself is known or not above the lower bound,
+// then the returned list will be empty.
+//
+// This may be somewhat expensive when first recovering from major sync.
+//
+// NOTE: TOTO: this issue needs to be finished, see issue #3933
```
timwu20, line 62:
```suggestion
// NOTE: this issue needs to be finished, see issue #3933
```

timwu20: what else needs to be finished though? Can you clarify in the issue.

dot/parachain/chainapi/chainapi.go
```diff
@@ -0,0 +1,20 @@
...
+
+type ChainAPIMessage[message any] struct {
+	Message         message
+	ResponseChannel chan any
+}
+
+type BlockHeader common.Hash
+
+func GetNumberOfValidators() uint {
+	// TODO: implement this, currently it's just a stub that should be replaced, see issue #3932
```
kishansagathiya, line 18: +1

dot/parachain/collator-protocol/validator_side.go
```diff
@@ -422,12 +422,14 @@ func (s SortableActivatedLeaves) Swap(i, j int) {
 	s[i], s[j] = s[j], s[i]
 }
 
-func (cpvs *CollatorProtocolValidatorSide) ProcessBlockFinalizedSignal(signal parachaintypes.BlockFinalizedSignal) {
+func (cpvs *CollatorProtocolValidatorSide) ProcessBlockFinalizedSignal(signal parachaintypes.
+	BlockFinalizedSignal) error {
 	if cpvs.finalizedNumber >= signal.BlockNumber {
 		// error
-		return
+		return fmt.Errorf("finalized number is greater than or equal to the block number")
```
kishansagathiya: let's make this a sentinel error

dot/parachain/overseer/overseer.go
```diff
@@ -123,6 +138,18 @@ func (o *Overseer) processMessages() {
...
+			case chainapi.ChainAPIMessage[util.Ancestors], chainapi.ChainAPIMessage[chainapi.BlockHeader]:
+				subsystem = o.nameToSubsystem[parachaintypes.ChainAPI]
+
+			case parachain.RuntimeAPIMessage:
+				rt, err := o.blockState.GetRuntime(msg.Hash)
+				if err != nil {
+					logger.Errorf("failed to get runtime: %v", err)
+					continue
+				}
+				logger.Infof("runtime: %v", rt)
```
kishansagathiya:
```suggestion
```

dot/parachain/overseer/overseer.go
```diff
@@ -123,6 +138,18 @@ func (o *Overseer) processMessages() {
 
 				subsystem = o.nameToSubsystem[parachaintypes.AvailabilityStore]
 
+			case chainapi.ChainAPIMessage[util.Ancestors], chainapi.ChainAPIMessage[chainapi.BlockHeader]:
+				subsystem = o.nameToSubsystem[parachaintypes.ChainAPI]
+
+			case parachain.RuntimeAPIMessage:
+				rt, err := o.blockState.GetRuntime(msg.Hash)
```
kishansagathiya, line 146: Lets link this issue here #3940

dot/parachain/overseer/overseer.go
```diff
@@ -47,9 +57,10 @@ type BlockState interface {
 	FreeImportedBlockNotifierChannel(ch chan *types.Block)
 	GetFinalisedNotifierChannel() chan *types.FinalisationInfo
 	FreeFinalisedNotifierChannel(ch chan *types.FinalisationInfo)
+	GetRuntime(hash common.Hash) (runtime.Instance, error)
 }
 
-func NewOverseer(blockState BlockState) *Overseer {
+func NewOverseer(blockState BlockState) OverseerSystem {
```
kishansagathiya:
```suggestion
func NewOverseer(blockState BlockState) *Overseer {
```

A good practice is to `accept interfaces, return structs`

dot/parachain/backing/candidate_backing.go
```diff
@@ -250,20 +250,18 @@ func (cb *CandidateBacking) processMessage(msg any, chRelayParentAndCommand chan
 	case StatementMessage:
 		return cb.handleStatementMessage(msg.RelayParent, msg.SignedFullStatement, chRelayParentAndCommand)
 	case parachaintypes.ActiveLeavesUpdateSignal:
-		err := cb.ProcessActiveLeavesUpdateSignal(msg)
-		if err != nil {
-			return fmt.Errorf("processing active leaves update signal: %w", err)
-		}
+		return cb.ProcessActiveLeavesUpdateSignal(msg)
```
kishansagathiya: Let's keep the err as wrapped as before. I know we haven't wrapped all errors. but it is of course useful.

dot/parachain/availability-store/messages.go
```diff
@@ -141,8 +150,8 @@ func (Unavailable) Index() uint {
 // which case the same timestamp will be reused. Blocks are sorted ascending first by block
 // number and then hash. candidate data was first observed at the given time and is available in at least one block
 type Unfinalized struct {
-	Timestamp       time.Time
-	BlockNumberHash []BlockNumberHash
+	Timestamp  BETimestamp
```
kishansagathiya, line 153: Why are we using BETimestamp? cant we just use time.Time?

dot/parachain/availability-store/availability_store_test.go
```diff
@@ -27,48 +39,72 @@ var (
...
 	require.NoError(t, err)
-	meta := CandidateMeta{
+	meta := &CandidateMeta{
 		State:         metaState,
 		DataAvailable: false,
 		ChunksStored:  []bool{false, false, false},
 	}
 
-	dataBytes, err := scale.Marshal(meta)
+	dataBytes, err := scale.Marshal(*meta)
```
kishansagathiya: seems like an unnecessary change

dot/parachain/availability-store/availability_store.go
```diff
@@ -42,32 +46,30 @@ const (
...
 // The only purpose of this structure is to use different timing
 // configurations in production and in testing.
-type pruningConfig struct {
-	keepUnavailableFor time.Duration
-	keepFinalizedFor   time.Duration
-	pruningInterval    time.Duration
+type PruningConfig struct {
+	KeepUnavailableFor time.Duration
+	KeepFinalizedFor   time.Duration
+	PruningInterval    time.Duration
```
kishansagathiya: PruningConfig doesnt seem to get used outside this package. Why are we exporting this?

dot/parachain/availability-store/availability_store.go
edwardmack: updated.

dot/parachain/collator-protocol/validator_side.go
edwardmack: Updated.

dot/parachain/overseer/overseer.go
edwardmack: Removed.

dot/parachain/overseer/overseer.go
edwardmack: Updated.

dot/parachain/availability-store/availability_store.go
edwardmack: Updated, made unexported.

dot/parachain/availability-store/availability_store.go
edwardmack: Added log message.

dot/parachain/availability-store/availability_store.go
edwardmack: Added panics where systemic errors are reached.

dot/parachain/util/util.go
edwardmack: Added comment to issue.

dot/parachain/overseer/overseer.go
edwardmack: updated.

dot/parachain/backing/candidate_backing.go
edwardmack: updated.

dot/parachain/availability-store/availability_store.go
edwardmack: updated.

dot/parachain/availability-store/messages.go
edwardmack: This is done so that we can make and use receiver function `ToBigEndianBytes`

dot/parachain/availability-store/availability_store_test.go
edwardmack: reverted.

dot/parachain/service.go
edwardmack: We'll take a look at refactoring this when we consider refactoring the service architecture.

dot/parachain/service.go
edwardmack: I've renamed the interface to `overseer.Overseer` and changed the implementation name to `OverseerSystem`

dot/parachain/availability-store/availability_store.go
edwardmack: Updated SubSystemToOverseer, will update "OverseerToSubsystem' when 'Register" is refactored.

timwu20 (approved):

PR #3841, opened 2024-04-01
feat(babe, epoch): retrieve first slot number correctly

### LINKED ISSUE #3795: fix(lib/epoch): use correctly very first slot number
## Describe the bug

- Currently, the `VerificationManager.VerifyBlock` method changes the epoch state if the header passed as argument is 1 
- It does not make sense for the verification to change the epoch state
- Makes more sense to set the first slot at finalization as said here https://github.com/ChainSafe/gossamer/pull/1640

EclesioMeloJunior:
## Changes

- Introduced `retrieveFirstNonOriginBlockSlot` that retrieves the block `#1` slot number, if there is more than one them select the correct one based on ancestry
- Changed BABE initialize epoch to check if epoch was skipped
- Change node initialization `createStateService` method that now instantiates the genesis runtime and retrieve `BabeConfigurations` as it is required for the epoch state
- This PR will make the code extensible for a full node that does not keep the entire chain history

## Tests

```sh
```

## Issues

- Closes https://github.com/ChainSafe/gossamer/issues/3795
- Related to https://github.com/ChainSafe/gossamer/issues/3430

## Primary Reviewer

@kishansagathiya


dot/build_spec.go
```diff
@@ -98,10 +99,13 @@ func BuildFromDB(path string) (*BuildSpec, error) {
 		Path:      path,
 		LogLevel:  log.Info,
 		Telemetry: telemetry.NewNoopMailer(),
+		GenesisBABEConfig: &types.BabeConfiguration{
```
jimjbrettj, line 102: Will this change depending on chain? For example, is the epoch length the same on polkadot as it is on westend?

dot/import.go
```diff
@@ -21,7 +21,9 @@ import (
 )
 
 // ImportState imports the state in the given files to the database with the given path.
-func ImportState(basepath, stateFP, headerFP string, stateTrieVersion trie.TrieLayout, firstSlot uint64) error {
+func ImportState(
+	basepath, stateFP, headerFP string, stateTrieVersion trie.TrieLayout,
+	genesisBABEConfig *types.BabeConfiguration, firstSot uint64) error {
```
jimjbrettj:
```suggestion
	genesisBABEConfig *types.BabeConfiguration, firstSlot uint64) error {
```

dot/import.go
```diff
@@ -35,11 +37,12 @@ func ImportState(basepath, stateFP, headerFP string, stateTrieVersion trie.TrieL
...
 	config := state.Config{
-		Path:     basepath,
-		LogLevel: log.Info,
+		Path:              basepath,
+		LogLevel:          log.Info,
+		GenesisBABEConfig: genesisBABEConfig,
 	}
 	srv := state.NewService(config)
-	return srv.Import(header, tr, stateTrieVersion, firstSlot)
+	return srv.Import(header, tr, stateTrieVersion, firstSot)
```
jimjbrettj:
```suggestion
	return srv.Import(header, tr, stateTrieVersion, firstSlot)
```

dot/state/epoch.go
```diff
@@ -48,13 +48,19 @@ func configDataKey(epoch uint64) []byte {
 	return append(configDataPrefix, buf...)
 }
 
+type GenesisEpochDescriptor struct {
```
jimjbrettj, line 52: Nit/this should have comment since its exported

dot/state/epoch.go
```diff
@@ -191,21 +189,25 @@ func (s *EpochState) GetEpochForBlock(header *types.Header) (uint64, error) {
...
+		return 0, fmt.Errorf("retrieving very first slot number: %w", err)
 	}
 
 	slotNumber, err := header.SlotNumber()
 	if err != nil {
 		return 0, fmt.Errorf("getting slot number: %w", err)
 	}
 
-	return (slotNumber - firstSlot) / s.epochLength, nil
+	return (slotNumber - veryFirstSlotNumber) / s.epochLength, nil
```
jimjbrettj: nit/perhaps a better name here. Maybe generisBlockSlotNumber or something like that? I think having a first slot and very first slot is a bit confusing

dot/state/epoch.go
```diff
@@ -440,50 +450,106 @@ func (s *EpochState) GetLatestConfigData() (*types.ConfigData, error) {
 }
 
 // GetStartSlotForEpoch returns the first slot in the given epoch.
-// If 0 is passed as the epoch, it returns the start slot for the current epoch.
-func (s *EpochState) GetStartSlotForEpoch(epoch uint64) (uint64, error) {
-	firstSlot, err := s.baseState.loadFirstSlot()
-	if err != nil {
-		return 0, err
+func (s *EpochState) GetStartSlotForEpoch(epoch uint64, bestBlockHash common.Hash) (uint64, error) {
+	veryFirstSlotNumber, err := s.retrieveFirstNonOriginBlockSlot(bestBlockHash)
```
jimjbrettj: same here, as well as any other uses of veryFirstBlockNumber

lib/babe/epoch.go
```diff
@@ -34,131 +34,96 @@ func (b *Service) initiateEpoch(epoch uint64) (*epochData, error) {
...
-		startSlot, err = b.getFirstAuthoringSlot(epoch, epochData)
+	currentEpoch := epoch + diff
+	if skipped {
+		logger.Warnf("⏩ A total of %d epochs were skipped, from %d to %d",
+			diff, epoch, currentEpoch)
+	}
+
+	// if we're at genesis or epoch was skipped then we can estimate when the start
+	// slot of the epoch will be, the estimation is used to calculate the epoch end
+	// TODO: check how substrate deals with these estimation
```
jimjbrettj, line 64: Is this still an active TODO?

lib/babe/epoch.go
```diff
@@ -34,131 +34,96 @@ func (b *Service) initiateEpoch(epoch uint64) (*epochData, error) {
...
+		endSlot:   startSlot + b.constants.epochLength,
+	}, nil
 }
 
-func (b *Service) checkAndSetFirstSlot() error {
-	firstSlot, err := b.epochState.GetStartSlotForEpoch(0)
-	if err != nil {
-		return fmt.Errorf("cannot set first slot: %w", err)
-	}
+var ErrEpochLowerThanExpected = errors.New("epoch lower than expected")
```
jimjbrettj: nit/ should this be at top of file?

lib/babe/epoch_handler.go
```diff
@@ -86,13 +85,21 @@ func (h *epochHandler) run(ctx context.Context, errCh chan<- error) {
 			return
 		}
 
+		fmt.Printf("current slot issued: %s\n", currentSlot.String())
```
jimjbrettj: nit/remove

lib/babe/epoch_handler.go
```diff
@@ -86,13 +85,21 @@ func (h *epochHandler) run(ctx context.Context, errCh chan<- error) {
...
+
 		// check if the slot is an authoring slot otherwise wait for the next slot
 		preRuntimeDigest, has := h.slotToPreRuntimeDigest[currentSlot.number]
 		if !has {
 			continue
 		}
 
-		err = h.handleSlot(h.epochNumber, currentSlot, h.epochData.authorityIndex, preRuntimeDigest)
+		fmt.Printf("has primary digest: %s\n", preRuntimeDigest.String())
```
jimjbrettj: nit/remove

dot/state/epoch.go
EclesioMeloJunior: the genesis block does not have a slot number, I can change to `chainFirstSlotNumber`

lib/babe/epoch_integration_test.go
```diff
@@ -17,26 +17,27 @@ import (
 
 func TestInitiateEpoch_Epoch0(t *testing.T) {
 	genesis, genesisTrie, genesisHeader := newWestendLocalGenesisWithTrieAndHeader(t)
-	babeService := createTestService(t, ServiceConfig{}, genesis, genesisTrie, genesisHeader, nil)
+	babeService := createTestService(t, ServiceConfig{}, genesis, genesisTrie, genesisHeader, genesisBABEConfig)
 	babeService.constants.epochLength = 20
-	startSlot := uint64(1000)
+	//startSlot := uint64(1000)
```
jimjbrettj: Comments need to be removed or readded back into test

lib/blocktree/blocktree.go
```diff
@@ -126,6 +126,10 @@ func (bt *BlockTree) GetHashesAtNumber(number uint) (hashes []common.Hash) {
 	bt.RLock()
 	defer bt.RUnlock()
 
+	if bt.root == nil {
```
jimjbrettj, line 129: is this related to this PR?

jimjbrettj (changes requested):
Looks nice, just some comments

lib/blocktree/blocktree.go
EclesioMeloJunior: Yup! fixes one test

dot/build_spec.go
timwu20: seems odd to me that the `BabeConfiguration` is in a package called `types`.

dot/node_integration_test.go
```diff
@@ -75,6 +75,15 @@ func TestNewNode(t *testing.T) {
 	stateConfig := state.Config{
 		Path:     initConfig.BasePath,
 		LogLevel: logLevel,
+		GenesisBABEConfig: &types.BabeConfiguration{
```
timwu20: maybe just have some predefined test variable that represents this default test config.

dot/state/epoch.go
```diff
@@ -17,19 +17,19 @@ import (
...
-	errEpochNotInDatabase = errors.New("epoch data not found in the database")
-	errHashNotPersisted   = errors.New("hash with next epoch not found in database")
+	ErrEpochLengthCannotBeZero     = errors.New("epoch length cannot be zero")
+	ErrConfigNotFound              = errors.New("config data not found")
+	ErrEpochNotInMemory            = errors.New("epoch not found in memory map")
+	errHashNotInMemory             = errors.New("hash not found in memory map")
+	errEpochNotInDatabase          = errors.New("epoch data not found in the database")
+	errHashNotPersisted            = errors.New("hash with next epoch not found in database")
+	ErrNoFirstNonOriginBlock       = errors.New("no first non origin block")
+	ErrGivenTimeBeforeNetworkStart = errors.New("given time is before network start")
```
timwu20: I would only really add exportable errors to the package signature if they are useful for outside packages to determine the type.  From checking references, seems like only `ErrConfigNotFound` is referenced outside of this package.

dot/state/epoch.go
```diff
@@ -191,21 +191,25 @@ func (s *EpochState) GetEpochForBlock(header *types.Header) (uint64, error) {
...
 	}
 
-	return (slotNumber - firstSlot) / s.epochLength, nil
+	return (slotNumber - chainFirstSlotNumber) / s.epochLength, nil
 }
 
-// SetEpochDataRaw sets the epoch data raw for a given epoch
-func (s *EpochState) SetEpochDataRaw(epoch uint64, raw *types.EpochDataRaw) error {
+// StoreEpochDataRaw sets the epoch data raw for a given epoch
+func (s *EpochState) StoreEpochDataRaw(epoch uint64, raw *types.EpochDataRaw) error {
```
timwu20: nite: Get/Set syntax is still fine in this case.  I don't think it really matters to differentiate whether or not it is persisted or not in this context.

dot/state/epoch.go
```diff
@@ -440,50 +452,105 @@ func (s *EpochState) GetLatestConfigData() (*types.ConfigData, error) {
...
 	}
 
-	return s.epochLength*epoch + firstSlot, nil
+	if errors.Is(err, ErrNoFirstNonOriginBlock) && epoch != 0 {
+		return 0, fmt.Errorf(
+			"%w: first non origin block is needed for epoch %d",
+			ErrNoFirstNonOriginBlock,
+			epoch)
+	}
+	return s.epochLength*epoch + chainFirstSlotNumber, nil
```
timwu20, line 476: I think you can group the conditionals:
```
// GetStartSlotForEpoch returns the first slot in the given epoch.
func (s *EpochState) GetStartSlotForEpoch(epoch uint64, bestBlockHash common.Hash) (uint64, error) {
	chainFirstSlotNumber, err := s.retrieveFirstNonOriginBlockSlot(bestBlockHash)
	if err != nil {
		if errors.Is(err, ErrNoFirstNonOriginBlock) {
			if epoch == 0 {
				slotDuration, err := s.GetSlotDuration()
				if err != nil {
					return 0, fmt.Errorf("getting slot duration: %w", err)
				}
				return uint64(time.Now().UnixNano()) / uint64(slotDuration.Nanoseconds()), nil
			}
			return 0, fmt.Errorf(
				"%w: first non origin block is needed for epoch %d",
				ErrNoFirstNonOriginBlock,
				epoch)
		}
		return 0, fmt.Errorf("retrieving first non origin block slot: %w", err)
	}
	return s.epochLength*epoch + chainFirstSlotNumber, nil
}
```

dot/state/epoch_test.go
```diff
@@ -82,31 +82,44 @@ func TestEpochState_EpochData(t *testing.T) {
...
-	err = s.SetEpochDataRaw(3, info)
+	err = s.blockState.AddBlock(&types.Block{
+		Header: header1,
+		Body:   types.Body{},
+	})
 	require.NoError(t, err)
 
-	start, err := s.GetStartSlotForEpoch(0)
+	start, err := s.GetStartSlotForEpoch(0, header1.Hash())
+	fmt.Println(start)
```
timwu20: can we remove these.
```suggestion
```

lib/babe/epoch.go
```diff
@@ -11,154 +11,140 @@ import (
 	"github.com/ChainSafe/gossamer/lib/crypto/sr25519"
 )
 
+var ErrEpochLowerThanExpected = errors.New("epoch lower than expected")
```
timwu20:
```suggestion
var errEpochLowerThanExpected = errors.New("epoch lower than expected")
```

lib/babe/epoch.go
```diff
@@ -11,154 +11,140 @@ import (
...
+
+type EpochDescriptor struct {
```
timwu20: does this need to be exported? Doesn't look like it needs to be based on checking references.  Will need to add comment if you want it to be exported.

dot/state/epoch.go
```diff
@@ -63,80 +71,65 @@ type EpochState struct {
...
-	}
-
 	if err := s.baseState.storeSkipToEpoch(0); err != nil {
 		return nil, err
 	}
 
 	return s, nil
 }
 
+// TODO: remove NewEpochStateFromGenesis function
```
jimjbrettj: is this still TODO?

lib/babe/babe_integration_test.go
```diff
@@ -137,35 +166,46 @@ func TestService_HandleSlotWithLaggingSlot(t *testing.T) {
...
 	bestBlockHash := babeService.blockState.BestBlockHash()
 	rt, err := babeService.blockState.GetRuntime(bestBlockHash)
 	require.NoError(t, err)
 
-	bestBlockHeader, err := babeService.blockState.GetHeader(bestBlockHash)
+	epochDescriptor, err := babeService.initiateEpoch(testEpochIndex)
+	require.NoError(t, err)
+
+	startTime := getSlotStartTime(epochDescriptor.startSlot, babeService.constants.slotDuration)
+	fmt.Println(startTime)
```
jimjbrettj: Can we remove prints in tests?

lib/babe/epoch.go
```diff
@@ -11,154 +11,140 @@ import (
...
 
-	startSlot, err := b.epochState.GetStartSlotForEpoch(epoch)
+	epochToFindData := epoch
+	if skipped {
+		// subtract 1 since we consider 0 as a valid epoch, then skipping from epoch
+		// 0 to epoch 2 we actually skipped only one epoch (epoch 1), however when
+		// searching for a data we should subtract the diff
+		lastKnownEpoch := epoch - diff
+		epochsSkipped := epoch - (diff - 1)
+		logger.Warnf("⏩ A total of %d epochs were skipped, from %d to %d",
```
jimjbrettj, line 45: I like the use of symbals in our logs, nice

lib/babe/epoch.go
```diff
@@ -11,154 +11,140 @@ import (
...
+			epochsSkipped, lastKnownEpoch, epoch)
+
+		// we should use the epoch data already setup for the
```
jimjbrettj, line 48: This are nice descriptions to have

lib/babe/epoch_handler.go
```diff
@@ -21,41 +21,40 @@ var (
...
-		if errors.Is(err, errNotOurTurnToPropose) {
+		if errors.Is(err, errNotOurTurnToPropose) || errors.Is(err, errOverPrimarySlotThreshold) {
 			continue
 		}
 
 		return nil, fmt.Errorf("failed to create new epoch handler: %w", err)
 	}
 
 	return &epochHandler{
+		// TODO: we should keep the last slot number from the previous epoch
```
jimjbrettj: is this still active?

lib/babe/verify_integration_test.go
```diff
@@ -265,57 +294,74 @@ func TestVerificationManager_VerifyBlock_MultipleEpochs(t *testing.T) {
...
+		start:    getSlotStartTime(futureEpochDescriptor.startSlot, babeService.constants.slotDuration),
+		duration: babeService.constants.slotDuration,
+		number:   futureEpochDescriptor.startSlot,
+	}
+	blockNumber02 := createTestBlockWithSlot(t, babeService,
+		&blockNumber01.Header, [][]byte{}, futureEpochDescriptor, futureSlot)
+	err = verificationManager.VerifyBlock(&blockNumber02.Header)
 	require.NoError(t, err)
+
+	// TODO: include test to verify skipped epoch
```
jimjbrettj, line 359: is this still todo?

jimjbrettj (approved):
lgtm, very nice work

dot/state/epoch.go
```diff
@@ -440,50 +452,105 @@ func (s *EpochState) GetLatestConfigData() (*types.ConfigData, error) {
...
+	return s.epochLength*epoch + chainFirstSlotNumber, nil
 }
 
-// GetEpochFromTime returns the epoch for a given time
-func (s *EpochState) GetEpochFromTime(t time.Time) (uint64, error) {
-	slotDuration, err := s.GetSlotDuration()
+// retrieveFirstNonOriginBlockSlot returns the slot number of the very first non origin block
+// if there is more than one first non origin block then it uses the block hash to check ancestry
+// e.g to return the correct slot number for a specific fork
+func (s *EpochState) retrieveFirstNonOriginBlockSlot(blockHash common.Hash) (uint64, error) {
```
kishansagathiya, line 482: ✅

dot/state/epoch.go
```diff
@@ -440,50 +452,105 @@ func (s *EpochState) GetLatestConfigData() (*types.ConfigData, error) {
...
-	return (slot - firstSlot) / s.epochLength, nil
+	return chainFirstSlotNumber, nil
 }
 
-// SetFirstSlot sets the first slot number of the network
-func (s *EpochState) SetFirstSlot(slot uint64) error {
-	// check if block 1 was finalised already; if it has, don't set first slot again
-	header, err := s.blockState.GetHighestFinalisedHeader()
+// GetEpochFromTime returns the epoch for a given time
+func (s *EpochState) GetEpochFromTime(t time.Time, blockHash common.Hash) (uint64, error) {
```
kishansagathiya: - GetEpochFromTime doesn't get used anywhere.
- You could put a comment about why you have passed block hash in the argument. As in meaning of the block hash (kinda like best block hash, last finalized block hash etc)

dot/state/epoch.go
```diff
@@ -540,6 +607,7 @@ func (s *EpochState) FinalizeBABENextEpochData(finalizedHeader *types.Header) er
 	var nextEpoch uint64 = 1
 	if finalizedHeader.Number != 0 {
 		finalizedBlockEpoch, err := s.GetEpochForBlock(finalizedHeader)
+
```
kishansagathiya:
```suggestion
```

looks like a mistake

dot/state/epoch.go
```diff
@@ -191,21 +191,25 @@ func (s *EpochState) GetEpochForBlock(header *types.Header) (uint64, error) {
...
+		return 0, fmt.Errorf("retrieving very first slot number: %w", err)
 	}
 
 	slotNumber, err := header.SlotNumber()
 	if err != nil {
 		return 0, fmt.Errorf("getting slot number: %w", err)
 	}
 
-	return (slotNumber - firstSlot) / s.epochLength, nil
+	return (slotNumber - chainFirstSlotNumber) / s.epochLength, nil
```
kishansagathiya, line 209: I can understand `firstSlot`, but I am confused on what could `chainFirstSlotNumber` mean?

dot/state/epoch.go
```diff
@@ -191,21 +191,25 @@ func (s *EpochState) GetEpochForBlock(header *types.Header) (uint64, error) {
 		return 0, errors.New("header is nil")
 	}
 
-	firstSlot, err := s.baseState.loadFirstSlot()
+	if header.Number == 1 {
+		return 0, nil
```
kishansagathiya, line 196: Not a fan of returning silently. Can you put a comment on why we are returning here on block number 1.

dot/state/epoch.go
```diff
@@ -296,6 +304,10 @@ func (s *EpochState) setLatestConfigData(epoch uint64) error {
 // If the header params is nil then it will search only in the database.
 func (s *EpochState) GetConfigData(epoch uint64, header *types.Header) (configData *types.ConfigData, err error) {
 	for tryEpoch := int(epoch); tryEpoch >= 0; tryEpoch-- {
+		if tryEpoch == 0 {
```
kishansagathiya, line 308: were we not handling this case at all earlier?

dot/state/epoch.go
```diff
@@ -440,50 +452,105 @@ func (s *EpochState) GetLatestConfigData() (*types.ConfigData, error) {
...
+		return 0, fmt.Errorf("retrieving first non origin block slot: %w", err)
+	}
+
+	if errors.Is(err, ErrNoFirstNonOriginBlock) && epoch == 0 {
+		slotDuration, err := s.GetSlotDuration()
+		if err != nil {
+			return 0, fmt.Errorf("getting slot duration: %w", err)
+		}
+
+		return uint64(time.Now().UnixNano()) / uint64(slotDuration.Nanoseconds()), nil
```
kishansagathiya: might be useful to have a function to calculate slot from time. That could be reused here  and in GetEpochFromTime. 
Do we already have such a function though?

dot/import.go
```diff
@@ -21,7 +21,9 @@ import (
 )
 
 // ImportState imports the state in the given files to the database with the given path.
-func ImportState(basepath, stateFP, headerFP string, stateTrieVersion trie.TrieLayout, firstSlot uint64) error {
+func ImportState(
+	basepath, stateFP, headerFP string, stateTrieVersion trie.TrieLayout,
+	genesisBABEConfig *types.BabeConfiguration, firstSlot uint64) error {
```
kishansagathiya:
```suggestion
func ImportState(basepath, stateFP, headerFP string, stateTrieVersion trie.TrieLayout,
	genesisBABEConfig *types.BabeConfiguration, firstSlot uint64) error {
```

Common practice is sorta like this

dot/import_integration_test.go
```diff
@@ -93,21 +93,31 @@ func TestImportState_Integration(t *testing.T) {
...
-	err = ImportState(config.BasePath, stateFP, headerFP, trie.V0, firstSlot)
+	genesisBABEConfig := &types.BabeConfiguration{
+		SlotDuration:       1000,
+		EpochLength:        200,
+		C1:                 1,
+		C2:                 4,
+		GenesisAuthorities: []types.AuthorityRaw{},
+		Randomness:         [32]byte{},
+		SecondarySlots:     0,
+	}
```
kishansagathiya: this value repeats a lot throughout the codebase, we could create something like `testGenesisBabeConfig` and reuse that. 
I am not asking to reuse it everywhere. Some amount of redundancy is actually desirable in test data. So, we could try to reuse it when inside the same package.

dot/state/epoch_test.go
```diff
@@ -82,31 +82,44 @@ func TestEpochState_EpochData(t *testing.T) {
...
 	require.NoError(t, err)
 	require.Equal(t, uint64(1), start)
 
-	start, err = s.GetStartSlotForEpoch(1)
+	start, err = s.GetStartSlotForEpoch(1, header1.Hash())
+	fmt.Println(start)
```
kishansagathiya:
```suggestion
```

dot/state/epoch_test.go
```diff
@@ -82,31 +82,44 @@ func TestEpochState_EpochData(t *testing.T) {
...
 	require.NoError(t, err)
-	require.Equal(t, uint64(1)+s.epochLength, start)
+	require.Equal(t, uint64(201), start)
 
-	start, err = s.GetStartSlotForEpoch(2)
+	start, err = s.GetStartSlotForEpoch(2, header1.Hash())
+	fmt.Println(start)
```
kishansagathiya:
```suggestion
```

dot/state/service_integration_test.go
```diff
@@ -373,6 +427,15 @@ func TestService_Import(t *testing.T) {
...
 		Telemetry: telemetryMock,
+		GenesisBABEConfig: &types.BabeConfiguration{
+			SlotDuration:       1000,
+			EpochLength:        200,
+			C1:                 1,
+			C2:                 4,
+			GenesisAuthorities: []types.AuthorityRaw{},
+			Randomness:         [32]byte{},
+			SecondarySlots:     0,
+		},
```
kishansagathiya: this looks like a place where we can reuse testGenesisBABEConfig

dot/state/epoch.go
EclesioMeloJunior: @kishansagathiya `chainFirstSlotNumber` is the very first slot number that appear in the chain, it is also known as the slot number of block number `#1`, this slot number is important to calculate epochs correctly, I don't want to use `firstSlot` because we use this nomeclature to define the first slot of an epoch

dot/state/epoch.go
EclesioMeloJunior: This is not a silent return, actually the epoch number for block number `#1` is epoch 0, epochs start from 0 and are incremented sequentially `0...1...2`, so the block number `#1` belongs to epoch 0

kishansagathiya: Ah. Ok. That seems like interesting information that could be useful while reading code. would be nice to have that as a comment on line 195

dot/state/epoch.go
kishansagathiya: ok

dot/state/epoch.go
EclesioMeloJunior: Not really, my idea was to have a single constructor but will require more changes, will leave as it is

lib/babe/epoch_handler.go
EclesioMeloJunior: yes, I will resolve it in this PR

lib/babe/verify_integration_test.go
EclesioMeloJunior: it is already addressed here https://github.com/ChainSafe/gossamer/pull/3882

dot/state/epoch.go
EclesioMeloJunior: Removed the unneeded method `GetEpochFromTime`, the explanation regards the block hash is because we need to be fork aware when retrieving the first slot of the chain given that we might have many blocks `#1` and we should retrieve the correct one given the ancestry of the `blockHash`, this explanation is at `retrieveFirstNonOriginBlockSlot` method

dot/state/epoch.go
EclesioMeloJunior: Done!

dot/state/epoch.go
EclesioMeloJunior: actually not, before getting epoch data or config data we don't check the epoch, the check is inside both functions, and this check is under the loop bc the ConfigData is not present in every epoch, so if the runtime issues a config data in epoch 1 it will use this same config for many epochs, each time we want to get a config data we should use the latest we know for that chain and if the runtime does not issue any config we should use the genesis config data

dot/state/epoch.go
EclesioMeloJunior: is might be useful only for testing but for production use we never need to get the slot from real time

dot/import_integration_test.go
```diff
@@ -90,24 +91,25 @@ func TestNewHeaderFromFile(t *testing.T) {
 }
 
 func TestImportState_Integration(t *testing.T) {
-	config := DefaultTestWestendDevConfig(t)
+	defaultWndDevConfig := DefaultTestWestendDevConfig(t)
```
timwu20: supernit
```suggestion
	defaultWestendDevConfig := DefaultTestWestendDevConfig(t)
```

timwu20 (approved):
Can you look at the deepsource issues and ensure that you haven't introduced anything.

kishansagathiya (approved):

dot/state/epoch.go
```diff
@@ -17,19 +17,19 @@ import (
...
 
 var (
-	ErrConfigNotFound     = errors.New("config data not found")
-	ErrEpochNotInMemory   = errors.New("epoch not found in memory map")
-	errHashNotInMemory    = errors.New("hash not found in memory map")
-	errEpochNotInDatabase = errors.New("epoch data not found in the database")
-	errHashNotPersisted   = errors.New("hash with next epoch not found in database")
+	ErrConfigNotFound   = errors.New("config data not found")
+	ErrEpochNotInMemory = errors.New("epoch not found in memory map")
+
```
axaysagathiya:
```suggestion
```

lib/babe/verify_test.go
```diff
@@ -911,8 +910,6 @@ func Test_verifyBlockEquivocation(t *testing.T) {
 		tt := tt
 
 		t.Run(tname, func(t *testing.T) {
-			//t.Parallel()
```
axaysagathiya, line 914: Why?

EclesioMeloJunior: Removed it because it was commented, was not being used

axaysagathiya: You could uncomment them. Does it cause race condition?

EclesioMeloJunior: Okay! Done

axaysagathiya (approved):
Fix the lint problem. Otherwise, LGTM

edwardmack (approved):
Great work.

dimartiro (approved):
LGTM did you test it on staging?

PR #3844, opened 2024-04-01
refactor(pkg/trie): Remove GenesisBlock method from tries

### LINKED ISSUE #3845: (pkg/trie): Remove `GenesisBlock` method from trie pkg
# Description

Tries are not responsible to calculate the block header so we have to move the `GenesisBlock` method from `pkg/trie`, trie interface and implementations to the right package.

dimartiro:
## Changes

This PR is to remove the `GenesisBlock` method from tries, tries should not be responsible to calculate the genesis block / header, so I moved this method to `lib/runtime/genesis.go` instead

## Tests

```sh
make test
```

## Issues

closes #3845


EclesioMeloJunior (approved):

jimjbrettj (approved):
lgtm

PR #3855, opened 2024-04-03
feat(pkg/trie): Read-only lazy loading TrieDB

### LINKED ISSUE #3762: (pkg/trie): Use enums to define trie nodes
Based on my [latest proposal](https://github.com/ChainSafe/gossamer/discussions/3646#discussioncomment-8465826) I created this issue to address the first part of Phase 1.

> To improve node handling, I propose replacing our current Node structure with one oriented towards enums to more easily and structurally differentiate nodes.
> Along with this, we would replace the current StorageValue []byte and IsHashedValue bool fields with an enum NodeValue

The goal is to improve the way we handle the trie nodes, this will improve our code, understanding, checks and also makes the foundations to add more improvements (like lazy loading, caching) later

### LINKED ISSUE #3806: (pkg/trie): Implement read only TrieDB
Implement a `TrieDB` structure following this [proposal](https://hackmd.io/@dimartiro/trie#Step-2---Read-only-TrieDB-with-cache) 

This structure should be capable of lazy load nodes / values from the DB instead of retaining all the nodes in memory

dimartiro:
## Changes

This PR aims to add a new structure `TrieDB` to handle merkle tries with a lazy loading behavior.
This is the first PR of a serie of PRs to allow us to replace our in-memory trie to this one with lazy loading.

This PR only introduces `TrieDB` struct with support for `Get(key)` method only, the other methods will be addressed in a different PR to keep this one as clean as possible

## Summary of changes

- [x] Refactor trie interface to have read / write interfaces
- [x] Trie lookup method to lazy load trie nodes while traversing the trie
- [x] Get method using lazy loading
- [x] Decode branch nodes without decoding its children
   - [x] Use enums to decode nodes
   - [x] Tests
- [x] Test migration using `TrieDB` to read nodes created by `in memory` trie
   - [x] Happy path
   - [x] Error cases

`Entries()`, `NextKey()` and `GetKeysWithPrefix()` will be addressed in a different PR since we have to introduce new structures `TrieIterator` and `TrieKeyIterator` and I want to keep this PR as clean as possible

## Note for reviewers 

I copied some files from `/pkg/trie/node` to `/pkg/trie/triedb/codec` to reuse them without too much refactor. The plan is to remove the old `/pkg/trie/node` folder when we switch our `in-memory` implementation to the `TrieDB`

`lookup` method was made following [this algorithm](https://hackmd.io/@dimartiro/trie#Lookup)

## Tests

```sh
make test
```

## Issues

Closes #3762 
Related with #3806 

## Primary Reviewer

@EclesioMeloJunior


pkg/trie/inmemory/mocks_generate_test.go
```diff
@@ -0,0 +1,6 @@
+// Copyright 2022 ChainSafe Systems (ON)
```
jimjbrettj:
```suggestion
// Copyright 2024 ChainSafe Systems (ON)
```

pkg/trie/trie.go
```diff
@@ -13,17 +13,23 @@ import (
 // EmptyHash is the empty trie hash.
 var EmptyHash = common.MustBlake2bHash([]byte{0})
 
-type ChildTrieSupport interface {
+type ReadChildTries interface {
```
jimjbrettj: Nit/maybe can model like below with `PrefixTrieRead` and change to `ChildTriesRead` and `ChildTriesWrite`. This is just an idea, do whatever you think works best here

pkg/trie/inmemory/mocks_generate_test.go
dimartiro: Done thks!

pkg/trie/trie.go
```diff
@@ -54,13 +63,21 @@ type Hashable interface {
 	Hash() (common.Hash, error)
 }
 
-type Trie interface {
-	PrefixTrie
-	KVStore
+type ReadOnlyTrie interface {
```
jimjbrettj: I'm a little confused, why do we need separate read and write tries?

pkg/trie/triedb/codec/decode_test.go
```diff
@@ -0,0 +1,438 @@
...
+	"io"
+	"testing"
+
+	"github.com/ChainSafe/gossamer/lib/common"
+	"github.com/ChainSafe/gossamer/pkg/scale"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func scaleEncodeBytes(t *testing.T, b ...byte) (encoded []byte) {
```
jimjbrettj, line 17: Is it necessary to have both these helpers?

pkg/trie/triedb/codec/header.go
```diff
@@ -0,0 +1,111 @@
...
+		}
+
+		if buffer[0] < 255 {
+			// the end of the partial key length has been reached.
+			return nodeVariant, partialKeyLength, nil
+		}
+	}
+}
+
+var ErrVariantUnknown = errors.New("node variant is unknown")
```
jimjbrettj: nit/move this to top

pkg/trie/triedb/codec/header.go
```diff
@@ -0,0 +1,111 @@
...
+
+// variantsOrderedByBitMask is an array of all variants sorted
+// in ascending order by the number of LHS set bits each variant mask has.
+// See https://spec.polkadot.network/#defn-node-header
+// WARNING: DO NOT MUTATE.
+// This array is defined at global scope for performance
+// reasons only, instead of having it locally defined in
+// the decodeHeaderByte function below.
+// For 7 variants, the performance is improved by ~20%.
```
jimjbrettj, line 83: Nice details here!

pkg/trie/triedb/codec/decode.go
```diff
@@ -0,0 +1,162 @@
+// Copyright 2021 ChainSafe Systems (ON)
```
jimjbrettj:
```suggestion
// Copyright 2024 ChainSafe Systems (ON)
```

pkg/trie/triedb/codec/decode_test.go
```diff
@@ -0,0 +1,438 @@
+// Copyright 2021 ChainSafe Systems (ON)
```
jimjbrettj:
```suggestion
// Copyright 2024 ChainSafe Systems (ON)
```

pkg/trie/triedb/codec/header.go
```diff
@@ -0,0 +1,111 @@
+// Copyright 2021 ChainSafe Systems (ON)
```
jimjbrettj:
```suggestion
// Copyright 2024 ChainSafe Systems (ON)
```

pkg/trie/triedb/codec/header_test.go
```diff
@@ -0,0 +1,206 @@
+// Copyright 2021 ChainSafe Systems (ON)
```
jimjbrettj:
```suggestion
// Copyright 2024 ChainSafe Systems (ON)
```

pkg/trie/triedb/codec/key.go
```diff
@@ -0,0 +1,34 @@
+// Copyright 2021 ChainSafe Systems (ON)
```
jimjbrettj:
```suggestion
// Copyright 2024 ChainSafe Systems (ON)
```

pkg/trie/triedb/codec/key_test.go
```diff
@@ -0,0 +1,146 @@
+// Copyright 2021 ChainSafe Systems (ON)
```
jimjbrettj:
```suggestion
// Copyright 2024 ChainSafe Systems (ON)
```

pkg/trie/triedb/codec/key_test.go
```diff
@@ -0,0 +1,146 @@
...
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"go.uber.org/mock/gomock"
+)
+
+func repeatBytes(n int, b byte) (slice []byte) {
```
jimjbrettj: nit/maybe reverse order of args so number is last. Feels a bit more intuitive to me, but ultimately up to you here

pkg/trie/trie.go
dimartiro: We don't really need it but I created it to check that I'm just implementing the reads. We can rollback this change and go back to the previous interfaces

pkg/trie/triedb/codec/decode_test.go
dimartiro: We are using it for the `Decode` test

pkg/trie/triedb/codec/header.go
dimartiro: Oh this is a comment from Quentin 😄  I just copy the file to reuse it in the trieDB

pkg/trie/triedb/codec/mocks_generate_test.go
```diff
@@ -0,0 +1,6 @@
+// Copyright 2023 ChainSafe Systems (ON)
```
jimjbrettj:
```suggestion
// Copyright 2024 ChainSafe Systems (ON)
```

pkg/trie/triedb/codec/header.go
dimartiro: Done

pkg/trie/triedb/codec/decode_test.go
dimartiro: All copyrights years were updated 😄

pkg/trie/triedb/codec/node.go
```diff
@@ -0,0 +1,89 @@
...
+// SPDX-License-Identifier: LGPL-3.0-only
+
+package codec
+
+const ChildrenCapacity = 16
+
+// MerkleValue is a helper enum to differentiate between inline and hashed nodes
+// https://spec.polkadot.network/chap-state#defn-merkle-value
+type MerkleValue interface {
+	_isMerkleValue()
```
jimjbrettj: why the _ syntax? Im familiar with prefixing unexported global variables with _ as seen [here](https://github.com/uber-go/guide/blob/master/style.md#prefix-unexported-globals-with-_), does this apply for functions as well?

jimjbrettj: I personally dont love the syntax if it's not needed

pkg/trie/triedb/codec/key_test.go
dimartiro: Done! thks

pkg/trie/triedb/print.go
```diff
@@ -0,0 +1,18 @@
...
+
+package triedb
+
+import (
+	"fmt"
+
+	"github.com/ChainSafe/gossamer/lib/common"
+)
+
+func (t *TrieDB) String() string {
```
jimjbrettj, line 12: Why have this be its own file?

pkg/trie/triedb/codec/node.go
dimartiro: Taken from [here](https://zackoverflow.dev/writing/hacking-go-to-give-it-sumtypes#attempt-2-interfaces) but using single underscore instead of double

pkg/trie/triedb/print.go
dimartiro: Actually it doesn't need its own file, but I was trying to replicate the `inmemory` trie files structure https://github.com/ChainSafe/gossamer/blob/development/pkg/trie/inmemory/print.go

Where we are separating the trie methods into different files

pkg/trie/trie.go
```diff
@@ -54,13 +63,21 @@ type Hashable interface {
...
+	fmt.Stringer
+
+	KVSRead
 	Hashable
-	ChildTrieSupport
+	ChildTriesRead
+	PrefixTrieRead
 	TrieIterator
 	TrieDeltas
```
EclesioMeloJunior: I think we don't need to track deltas in a `ReadOnlyTrie`

pkg/trie/trie.go
EclesioMeloJunior: I like this interface separation, gives us a better way to compose the implementation also to understand the implementation's responsibilities

pkg/trie/triedb/codec/decode.go
```diff
@@ -0,0 +1,162 @@
...
+// decodeBranch reads from a reader and decodes to a node branch.
+// Note that we are not decoding the children nodes.
+func decodeBranch(reader io.Reader, variant variant, partialKeyLength uint16) (
+	node Branch, err error) {
+	node = Branch{}
+
+	node.PartialKey, err = decodeKey(reader, partialKeyLength)
+	if err != nil {
+		return Branch{}, fmt.Errorf("cannot decode key: %w", err)
+	}
```
EclesioMeloJunior: the `decodeKey` is common for `Branch` and `Leaf`, can we move this step to the `Decode` function and just pass the actual partial key to `decodeBranch/decodeKey`?

pkg/trie/triedb/codec/decode.go
```diff
@@ -0,0 +1,162 @@
...
+
+	switch variant {
+	case branchWithValueVariant:
+		valueBytes := make([]byte, 0)
+		err := sd.Decode(&valueBytes)
+
+		node.Value = NewInlineValue(valueBytes)
+		if err != nil {
+			return Branch{}, fmt.Errorf("%w: %s", ErrDecodeStorageValue, err)
+		}
```
EclesioMeloJunior, line 90: this error should not be right after `err := sd.Decode(&valueBytes)`?

pkg/trie/triedb/codec/decode.go
```diff
@@ -0,0 +1,162 @@
...
+		node.Value = NewHashedValue(hashedValue)
+	default:
+		// Do nothing, branch without value
+	}
+
+	for i := 0; i < ChildrenCapacity; i++ {
+		if (childrenBitmap[i/8]>>(i%8))&1 != 1 {
+			continue
+		}
```
EclesioMeloJunior: I think that this is a little bit hard to understand, I would like to suggest another approach:

```go 
// instead of dealing with a 2 items byte slice like `childrenBitmap := make([]byte, 2)` we deal with a // uint16 (which is also represented by a 2 byte slice)

var childrenBitmap uint16
err := binary.Read(reader, binary.LittleEndian, &bitmap) // will read 2 byte from the reader
if err != nil { ... }

// then you can go over the 16 positions checking each bit in the uint16 value
// but on every loop you shift childrenBitmap 1 to the left
for i := 0; i < ChildrenCapacity; i++{
   // this check is optional since we cover all bits (16) 
   // available in uint16 type
    if childrenBitmap == 0 {
        break;
     }
     
     // just check if we have a child in 
     // the current bit, if so, just decode
      if (childrenBitmap & 1) == 1 {
            err := decodeBranchChild()
            if err != nil {...}
       }
      
      // otherwise advance to check next bit in the bitmap
       childrenBitmap >>= 1
}

func decodeBranchChild() {
                 var hash []byte
		err := sd.Decode(&hash)
		if err != nil {
			return Branch{}, fmt.Errorf("%w: at index %d: %s",
				ErrDecodeChildHash, i, err)
		}

		if len(hash) < hashLength {
			node.Children[i] = NewInlineNode(hash)
		} else {
			node.Children[i] = NewHashedNode(hash)
		}
}
```

EclesioMeloJunior (changes requested):

pkg/trie/trie.go
dimartiro: True! Done

pkg/trie/triedb/codec/decode.go
dimartiro: Good catch! this will improve the nested error messages too!

pkg/trie/triedb/codec/decode.go
dimartiro: True! my bad! Fixed now!

pkg/trie/triedb/codec/decode.go
dimartiro: mmm, I agree that it's difficult to understand but I think we can solve it using a mix of both approaches.

1. We can parse the value into an `uint16` to not use an array of two elements
2. Since the uint16 is decoded using little endian can simplify the check to `if (childrenBitmap>>i)&1 != 1`

I think it's more understandable now with less changes

EclesioMeloJunior: Agreed!

pkg/trie/triedb/codec/decode_test.go
```diff
@@ -0,0 +1,408 @@
...
+func concatByteSlices(slices [][]byte) (concatenated []byte) {
+	length := 0
+	for i := range slices {
+		length += len(slices[i])
+	}
+	concatenated = make([]byte, 0, length)
+	for _, slice := range slices {
+		concatenated = append(concatenated, slice...)
+	}
+	return concatenated
```
EclesioMeloJunior: Use `bytes.Join` https://pkg.go.dev/bytes#Join
```suggestion
	return bytes.Join(slices, nil)
```

pkg/trie/triedb/codec/header.go
```diff
@@ -0,0 +1,110 @@
...
+	ErrVariantUnknown   = errors.New("node variant is unknown")
+)
+
+func decodeHeader(reader io.Reader) (nodeVariant variant,
+	partialKeyLength uint16, err error) {
+	buffer := make([]byte, 1)
+	_, err = reader.Read(buffer)
+	if err != nil {
+		return nodeVariant, 0, fmt.Errorf("reading header byte: %w", err)
+	}
```
EclesioMeloJunior, line 23: can this read be in the `decodeHeaderByte` and pass the reader as arg to the `decodeHeaderByte` method

```go 
decodeHeaderByte(reader)

func decodeHeaderByte(r io.Reader) ... {
        buffer := make([]byte, 1)
	_, err = reader.Read(buffer)
	if err != nil {
		return nodeVariant, 0, fmt.Errorf("reading header byte: %w", err)
	}
        ....
}
```

pkg/trie/triedb/codec/key.go
```diff
@@ -0,0 +1,34 @@
...
+	key := make([]byte, partialKeyLength/2+partialKeyLength%2)
+	n, err := reader.Read(key)
+	if err != nil {
+		return nil, fmt.Errorf("reading from reader: %w", err)
+	} else if n != len(key) {
+		return nil, fmt.Errorf("%w: read %d bytes instead of expected %d bytes",
+			ErrReaderMismatchCount, n, len(key))
+	}
+
+	return codec.KeyLEToNibbles(key)[partialKeyLength%2:], nil
```
EclesioMeloJunior, line 36: I think we can add a comment here explaining the motivation of this slice range

```
// if the partialKeyLength is an odd number means that when parsing the key to nibbles it will contains a useless 0 in the first index, otherwise we can use the entire nibbles
```

pkg/trie/triedb/codec/key_test.go
```diff
@@ -0,0 +1,146 @@
...
+	"github.com/stretchr/testify/assert"
+	"go.uber.org/mock/gomock"
+)
+
+func repeatBytes(b byte, n int) (slice []byte) {
+	slice = make([]byte, n)
+	for i := range slice {
+		slice[i] = b
+	}
+	return slice
```
EclesioMeloJunior: use `bytes.Repeat` https://pkg.go.dev/bytes#Repeat
```suggestion
	return bytes.Repeat([]byte{b}, n)
```

pkg/trie/triedb/codec/node.go
EclesioMeloJunior: Yeah, I think you can use just `is...`

pkg/trie/triedb/codec/node.go
```diff
@@ -0,0 +1,89 @@
...
+	IsHashed() bool
+}
```
EclesioMeloJunior: Instead of having a `MerkleValue` what about introducing a new item in to the Node variant called ref Ref

```go
type (
	// Empty node
	// Leaf always contains values
	// Branch could has or not has values
        // Ref is either a hash for a node or the node itself if Data is less than 32 bytes
	Ref struct {
		Data []byte
	}
)
```

So when you're decoding the branch children you always will have a 

```go
node.Children[i] = NewNodeRef(...)
```

which makes easy for looking up also, given that you will have a variant for node references, then you only need to check if the ref size and decode accordingly

pkg/trie/triedb/triedb.go
```diff
@@ -0,0 +1,173 @@
...
+
+// Get returns the value in the node of the trie
+// which matches its key with the key given.
+// Note the key argument is given in little Endian format.
+func (t *TrieDB) Get(key []byte) []byte {
+	val, err := t.lookup(key)
+	if err != nil {
+		// TODO: do we have to do anything else? maybe change the signature
+		// to return an error?
+		return nil
```
EclesioMeloJunior: This is up to you, I am ok with just returning the error

EclesioMeloJunior (changes requested):

pkg/trie/triedb/codec/node.go
dimartiro: But we are going to need to check the size to know if it is hashed or inlined... mmm I got your point but I'm not sure if this is an improvement.
I'd rather to keep `MerkleValue` that is mapped with the same concept in the [spec](https://spec.polkadot.network/chap-state#defn-merkle-value)

pkg/trie/triedb/triedb.go
dimartiro: We can address it in the future because I'll need to change the interface and the in-memory trie too

pkg/trie/triedb/codec/key_test.go
dimartiro: Good catch! very good opportunity to improve our tests! thks

pkg/trie/triedb/codec/decode_test.go
dimartiro: The future is here! thanks for the suggestion!

pkg/trie/triedb/codec/node.go
dimartiro: Changed! thks!

pkg/trie/triedb/codec/key.go
dimartiro: Yeah good idea! Added!

pkg/trie/triedb/codec/header.go
dimartiro: The problem with this change is that we need to use the `buffer` in some lines below this one so I would say that is better to keep it this way

pkg/trie/triedb/codec/node.go
EclesioMeloJunior: I just think that is easy to see them as references instead of merkle values, I would say that the `MerkleValue` and `NodeValue` seems confusing to me given that one refers to a byte slice that can become a node (`MerkleValue`) and the other is just the value contained by a node. So I still think that you can treat a byte slice that can be decoded into a node as a `Node::Ref`, and is responsibility of the decoder to check the size in order to decode directly or search in the database for the actual node bytes, if think that is better to have 2 variants for inline and hashed (`Node::InlineRef` and a `Node::HashedRef`) is up to you but I think these values should belong to the Node enum

pkg/trie/triedb/codec/header.go
EclesioMeloJunior: oh but this is not a problem, the `buffer` is just a variable that will be used to store a new `reader.Read` value, you can do like:

```go
buffer := make([]byte, 1)
for {
		_, err = reader.Read(buffer)
```

pkg/trie/trie.go
```diff
@@ -13,17 +13,23 @@ import (
...
+}
+
+type ChildTriesWrite interface {
 	PutIntoChild(keyToChild, key, value []byte) error
 	DeleteChild(keyToChild []byte) (err error)
 	ClearFromChild(keyToChild, key []byte) error
 }
 
-type KVStore interface {
+type KVSRead interface {
```
timwu20:
```suggestion
type KVStoreRead interface {
```

pkg/trie/trie.go
```diff
@@ -13,17 +13,23 @@ import (
...
 	Get(key []byte) []byte
+}
+
+type KVSWrite interface {
```
timwu20:
```suggestion
type KVStoreWrite interface {
```

pkg/trie/trie.go
timwu20:
```suggestion
type TrieRead interface {
```

pkg/trie/triedb/codec/node.go
```diff
@@ -0,0 +1,89 @@
...
+
+type (
+	// Value bytes as stored in a trie node
+	InlineNode struct {
+		Data []byte
+	}
+	// Containing a hash used to lookup in db for real value
+	HashedNode struct {
+		Data []byte
+	}
```
timwu20, line 23: Do these need to be exported?

pkg/trie/triedb/triedb.go
timwu20: Returning `nil` makes sense to me.

pkg/trie/triedb/codec/node.go
dimartiro: Actually no, I'll change it! thks

dimartiro: My bad, I was checking and yeah, since we are using them in `TrieDB` we need to export them

pkg/trie/triedb/codec/node.go
dimartiro: I'm still thinking that it is better to differentiate this "special nodes" using the `MerkleValue` enum.

- One reason is that it is the same concept in the [spec](https://spec.polkadot.network/chap-state#defn-merkle-value)
- Another reason is that if we change them, since they are "special nodes" we cannot get the `partial key` or the `value` from them since they are not decoded yet, so that will force us to add methods returning nil if we want to add more methods to the interface as I'm doing [here](https://github.com/ChainSafe/gossamer/pull/3884/files#diff-26147d32b2f68223030e0024afd41fc96e346f0b7fcdf14e065688df4d3c3248)

Don't be mad with me but I'll keep my vote on leaving them as they are now 😅

pkg/trie/triedb/codec/header.go
dimartiro: On it! I'll have to update the tests to mock the reader so it could take some time

pkg/trie/triedb/codec/node.go
EclesioMeloJunior: I will not be mad hahah! Okay, there is no problem at all and we can progress with your impl, but I would like to add some thoughts: 
- the first reason does not convince me too much since spec is more a general way of describing the idea and not much about how we should implement.
- For the second point: yeah, they are nodes but described in another form but they still belong to the nodes group, so for that reason we can fit them in the `Node` interface. 
> we cannot get the partial key or the value from them since they are not decoded yet

so it is like:
```go 
var a *int
fmt.Println(a)
```
we cannot retrieve the value that `a` holds, to that we need to deref the pointer, to do so we need access to the heap where the data is, so I think we can have the same to the `Ref` type, it is responsability to the `Ref` to retrieve the data when needed, like deref is resposability of the pointer in order to bring the actual value

```go
type Node interface {
   paritalKey()
   value()
}

type Ref struct {
  db DatabaseGetter
  encoded []byte
}

fn (r *Ref) partialKey() {
  _, keyLen := decodeHeader(r.encoded)
  return decodeKey(r.encoded, keyLen)
}

fn (r *Ref) value() {
 reader := bytes.NewReader(r.encoded)
 nodeKind, _ := decodeHeader(reader)
 // you might need the database if the value is just
 // a hash to a value store in database
 return decodeNodeValue(reader, nodeKind, r.db)
}
```

EclesioMeloJunior (approved):
LGTM! Looking forward to have that on our staging nodes! 🚀

pkg/trie/triedb/codec/node.go
dimartiro: I agree that it is feasible, but that's not the main blocker. 
Implementing this change would require us to modify the `Decode` function to add the `db` as part of the node. Currently, the `codec` is designed without any knowledge of the `db`. 
My goal with these new changes is to maintain the code's simplicity and avoid unnecessary complexity, in line with the [KISS](https://en.wikipedia.org/wiki/KISS_principle) principle. Therefore, although the proposed change might seem sensible, I believe we should defer it and possibly reconsider in the future if there's a compelling need.

Let me know if you are ok with this so we can resolve this conversation. 
Thank you again for the discussion ❤️

pkg/trie/triedb/codec/header.go
dimartiro: I was doing the change and I noticed that it will affect the decode byte [benchmarks](https://github.com/ChainSafe/gossamer/pull/3855/files#diff-00ff708a8b8c180cd4c0e9106030760c6b73962b4b148cbd375a6b07165c565aR195-R206) since after this change we are going to read from the reader and we are not going to be testing what we expect (just using the byte) so in my opinion we can skip for now

jimjbrettj (approved):
lgtm!

timwu20 (approved):

PR #3857, opened 2024-04-04
feat(lib): support `MinimumBackingVotes` runtime function call

### LINKED ISSUE #3856: add a runtime function to get minimum backing votes
## Issue summary

- export function to get minimum backing votes: `ParachainHost_minimum_backing_votes`

-

axaysagathiya:
## Changes

- added support to call runtime method to get minimum backing votes
- as the test was failing, I used runtime version 1.9.0 instead of 1.8.0
- set min memory pages to 2080

## Tests

```sh
```

## Issues
closes #3856 

## Primary Reviewer

@timwu20 @kishansagathiya


axaysagathiya:
- From Polkadot.js I get the minimum backing votes is 2 for block hash `0x104d86ba31c8835ca4ffc5b09804302711865242ed422a95250c86d63e6ba3f3`
- but in test I get 0, instead of 2

Screenshot from 2024-04-04 20-11-55

Screenshot from 2024-04-03 12-26-35

Screenshot from 2024-04-03 12-30-28

lib/runtime/wazero/instance.go
```diff
@@ -108,7 +108,7 @@ func NewInstance(code []byte, cfg Config) (instance *Instance, err error) {
 
 	_, err = rt.NewHostModuleBuilder("env").
 		// values from newer kusama/polkadot runtimes
-		ExportMemory("memory", 23).
+		ExportMemory("memory", 24).
```
axaysagathiya: Not sure about this value. 
I need help finding out what the correct value is.

I was getting an error, so I had to change it.
Screenshot from 2024-04-06 20-01-38

axaysagathiya: I used `2070` as the value of `memory pages` as I got the ans [here in Slack](https://chainsafeio.slack.com/archives/CEATMPULS/p1712414223681539)

kishansagathiya (approved):
Good work! LGTM! 🎉 

There is some confusion in the wording on memory pages. At places (in polkadot-sdk), it says min pages and at places it says max pages. Let's confirm which one is it.

EclesioMeloJunior:
@kishansagathiya the max pages in polkadot-sdk from what I know is `65536` and each page can have a maximum of `65536` bytes so the max bytes allocation is `4294967296`. Regards the `ExportMemory` method, it tells the wazero to initialize a memory buffer with size equals to the number of pages passed as argument, eg. we're passing `2070` pages which will initialize a buffer of `135659520` bytes (2070 * 65536)

EclesioMeloJunior (approved):
LGTM

lib/runtime/wazero/instance.go
```diff
@@ -108,7 +108,7 @@ func NewInstance(code []byte, cfg Config) (instance *Instance, err error) {
 
 	_, err = rt.NewHostModuleBuilder("env").
 		// values from newer kusama/polkadot runtimes
-		ExportMemory("memory", 23).
+		ExportMemory("memory", 2070).
```
P1sar: First of all, no magic variables please.  This should be const var with specific name. 
Secondly, if this value static it should be 65k. Please refer to our slack thread for more details.

P1sar (changes requested):

lib/runtime/wazero/instance.go
axaysagathiya: basti replied on element chat,
> Basti.await
> These values are implementation specific. They are just to optimize the memory usage
> If the instantiation would fail, we would just increase the values

So, we can use 24 now, and if we use more it will just waste the memory

P1sar (approved):

axaysagathiya:
Finally, as we decided on Slack, I'm setting the value of min memory pages to 2080.

PR #3865, opened 2024-04-04
fix(lib/runtime): allocator out of space and wasm out of bounds memory access

EclesioMeloJunior:
## Changes

- Detailed error description in issue: https://github.com/ChainSafe/gossamer/issues/3774
- This PR introduces:
  - Refactored the method `(*Instance) Exec` to create a wazero runtime instance with a new allocator to execute the runtime exported method
  - These changes are a fix for two problems `allocator out of space` and `wasm out of bounds memory access`
  - After checking out substrate `runtime/wasmtime` methods we notice they have a strategy called `RecreateInstance` that is responsible for instantiate a new `wasmtime` runtime instance and perform the method call using a fresh new allocator ([substrate call_impl](https://github.com/paritytech/polkadot-sdk/blob/0c6c837f689a287583508506e342ba07687e8d26/substrate/client/executor/wasmtime/src/runtime.rs#L170C39-L170C55) and [strategy](https://github.com/paritytech/polkadot-sdk/blob/0c6c837f689a287583508506e342ba07687e8d26/substrate/client/executor/wasmtime/src/runtime.rs#L143))
  - It is not possible to use a single runtime instance only refreshing the allocator since it leads to `wasm out of bounds memory access`.

## Tests

Before the changes we were facing a panic while executing block #2 on paseo network
image

After the changes we can execute entire chains
image

## Issues

- Closes https://github.com/ChainSafe/gossamer/issues/3774

## Primary Reviewer

@jimjbrettj @timwu20

Issue #3823, opened 2024-03-27
chore(sync): Sync speed metric


dimartiro:
## Task summary

Create prometheus metric for sync speed

We currently have logs like:

```
synced 7681 blocks, took: 46.96 seconds, bps: 163.58 blocks/second
```

But we don't have a way to graph this speed in datadog.
The goal is to send a metric and create a chart in datadog to measure our sync speed and help us to understand if a change affects this speed in a good or bad way

PR #3879, opened 2024-04-08
feat(lib/runtime/wazero): implement compiler cache

jimjbrettj:
## Changes

Use compilation caching provided by wazero to increase sync speeds

Speedup data compared to non cache solution:

| | |
|-----| ------|
| CACHED | ⛓️ synced 7681 blocks, took: 145.12 seconds, bps: 52.93 blocks/second |
| NON CACHED | ⛓️ synced 7681 blocks, took: 749.70 seconds, bps: 10.25 blocks/second	pkg=sync |

## Tests

- run paseo/westend nodes and makes sure it works and that sync speed increases

## Issues

 #3880

## Primary Reviewer

@EclesioMeloJunior


dimartiro:
Do you have any metric to check how much do we improve the sync speed with this change? I think it could be a good opportunity to revisit this issue https://github.com/ChainSafe/gossamer/issues/3823

dot/core/service.go
```diff
@@ -312,6 +312,12 @@ func (s *Service) handleCodeSubstitution(hash common.Hash,
 		cfg.Role = 4
 	}
 
+	err = rt.CleanCache()
+	if err != nil {
+		logger.Debugf("cleaning runtime compiler cache: %v", err)
```
EclesioMeloJunior: I think we don't need to log if we're about to return an error

dot/core/service.go
```diff
@@ -312,6 +312,12 @@ func (s *Service) handleCodeSubstitution(hash common.Hash,
...
+		return fmt.Errorf("cleaning runtime compiler cache: %w", err)
+	}
+
```
timwu20: Why are we cleaning the cache when handling substituted code? Shouldn't we persist this cache in the case we resume the node?

EclesioMeloJunior (changes requested):
Can we make the `CleanCache` a private function that belongs only to the `Instance` structure, that makes more sense given that the instance is the one who wraps the wazero instance, also call the `cleanCache` only in the `Stop` method, so we make sure that we clean the current instance cache only when the instance stop

jimjbrettj:
> Do you have any metric to check how much do we improve the sync speed with this change? I think it could be a good opportunity to revisit this issue #3823

@dimartiro I think this is a very nice thing to have, however, I do not thing that this is something worth doing right now considering our current needs. Eclesio's work fixes memory issues and gives us correctness and then this is a speedup on that. I think for now I have more pressing things to work on, but in the future this would be nice to have. I will however still add some basic metrics I collect to add a bit more context into the speed up

dot/core/service.go
jimjbrettj: We need to close the cache to prevent a memory leak, however I do not think we need to remove the dir so I will make this change

jimjbrettj: And you are right than in this case we do not need to clean the cache. This PR has been modified to only close the cache when the instance is pruned

dot/core/service.go
jimjbrettj: this has been removed

lib/runtime/wazero/instance.go
```diff
@@ -412,8 +422,19 @@ func NewInstance(code []byte, cfg Config) (instance *Instance, err error) {
 	logger.Info("instantiating a runtime!")
 	logger.Patch(log.SetLevel(cfg.LogLvl), log.SetCallerFunc(true))
 
+	// Prepare a cache directory.
+	cacheDir, err := os.MkdirTemp("", "wazeroCache")
```
EclesioMeloJunior: saves the `cacheDir` to delete it once we close the instance, looks like it does not remove the dir once it closes the cache https://github.com/tetratelabs/wazero/blob/253c0346e4c8d7bd604304447233c1a3ce0a71f7/cache.go#L70

jimjbrettj: yeah exactly, I talked with @timwu20 and we were thinking we can just keep it for if we restart the node

lib/runtime/wazero/instance.go
```diff
@@ -36,13 +37,19 @@ const runtimeContextKey = contextKey("runtime.Context")
...
+	config wazero.RuntimeConfig
+	cache  wazero.CompilationCache
+}
+
 // Instance backed by wazero.Runtime
 type Instance struct {
 	Runtime  wazero.Runtime
 	Module   api.Module
 	Context  *runtime.Context
+	code     []byte
```
axaysagathiya: Can you use a more meaningful field name here? 

If I am correct, it's a wasm bytecode. Maybe you can rename it to `wasmByteCode.`

lib/runtime/wazero/instance.go
```diff
@@ -100,15 +107,13 @@ func NewInstanceFromTrie(t trie.Trie, cfg Config) (*Instance, error) {
 	return NewInstance(code, cfg)
 }
 
-// NewInstance instantiates a runtime from raw wasm bytecode
```
axaysagathiya, line 103: You could have kept this comment. It will help to know what the `code` argument is.

lib/runtime/wazero/instance.go
timwu20: We should be putting this in `~/.gossamer` then if we plan to keep it.

lib/runtime/wazero/instance.go
```diff
@@ -396,35 +401,47 @@ func NewInstance(code []byte, cfg Config) (instance *Instance, err error) {
...
-		return nil, fmt.Errorf("wazero error: nil global for __heap_base")
-	}
+	return mod, rt, nil
+}
 
-	heapBase := api.DecodeU32(encodedHeapBase.Get())
-	// hb = runtime.DefaultHeapBase
+// NewInstance instantiates a runtime from raw wasm bytecode
+func NewInstance(code []byte, cfg Config) (instance *Instance, err error) {
+	logger.Info("instantiating a runtime!")
```
timwu20: maybe make this debug? Seems pretty obvious we are instantiating a runtime.

lib/runtime/wazero/instance.go
```diff
@@ -899,4 +937,9 @@ func (in *Instance) Stop() {
 	if err != nil {
 		log.Errorf("runtime failed to close: %v", err)
 	}
+
+	err = in.metadata.cache.Close(context.Background())
+	if err != nil {
+		log.Errorf("closing the wazero compilation cache: %v", err)
```
timwu20, line 943: why not return this error?

lib/runtime/wazero/instance_test.go
```diff
@@ -1096,7 +1095,7 @@ func newTrieFromPairs(t *testing.T, filename string) trie.Trie {
 		entries[pairArr[0].(string)] = pairArr[1].(string)
 	}
 
-	tr, err := inmemory.LoadFromMap(entries, trie.V0)
```
timwu20, line 1099: unnecessary diff

timwu20 (changes requested):

lib/runtime/wazero/instance.go
jimjbrettj: making this change, however, we have moved to using `xdg.DataHome` rather than `~/.gossamer` so using that instead

lib/runtime/wazero/instance.go
jimjbrettj: This comment is still there, just in a new place

lib/runtime/wazero/instance.go
jimjbrettj: The current stop function signature did not return an error, it just logged if there was an error closing so I just kept with what was being done before. I can update this if desired though

lib/runtime/wazero/instance_test.go
jimjbrettj: I see this comment on the previous PR (lost when we combined them) https://github.com/ChainSafe/gossamer/pull/3865#discussion_r1552423920

So maybe there is context to this change I am missing. @dimartiro could you add some context here? else I will delete this diff

dimartiro: This is just to remove the old import and use the new one with the alias

jimjbrettj: okay, i will keep then

jimjbrettj:
I did not see any noticeable improvement with the cache dir, and it was breaking all CI, and with @ec2 changes implementing more caching I do no think it is needed so I removed it

EclesioMeloJunior (approved):

chain/paseo/chain-spec-raw.json
```diff
@@ -6,7 +6,19 @@
     "/dns/paseo.bootnode.amforc.com/tcp/30333/wss/p2p/12D3KooWFD81HC9memUwuGMLvhDDEfmXjn6jC4n7zyNs3vToXapS",
     "/dns/paseo.bootnode.amforc.com/tcp/30344/p2p/12D3KooWFD81HC9memUwuGMLvhDDEfmXjn6jC4n7zyNs3vToXapS",
     "/dns/boot.stake.plus/tcp/43334/wss/p2p/12D3KooWNhgAC3hjZHxaT52EpPFZohkCL1AHFAijqcN8xB9Rwud2",
-    "/dns/boot.stake.plus/tcp/43333/p2p/12D3KooWNhgAC3hjZHxaT52EpPFZohkCL1AHFAijqcN8xB9Rwud2"
+    "/dns/boot.stake.plus/tcp/43333/p2p/12D3KooWNhgAC3hjZHxaT52EpPFZohkCL1AHFAijqcN8xB9Rwud2",
```
dimartiro, line 9: Are we adding these nodes because we are testing this change in paseo? I think this could be in a different PR

dimartiro (approved):
lgtm

chain/paseo/chain-spec-raw.json
jimjbrettj: Well yes, but often the node will not sync without this change. So I think okay to keep here since small and allows testing

lib/runtime/wazero/instance.go
```diff
@@ -100,15 +108,13 @@ func NewInstanceFromTrie(t trie.Trie, cfg Config) (*Instance, error) {
...
 }
 
-// NewInstance instantiates a runtime from raw wasm bytecode
-func NewInstance(code []byte, cfg Config) (instance *Instance, err error) {
-	logger.Info("instantiating a runtime!")
-	logger.Patch(log.SetLevel(cfg.LogLvl), log.SetCallerFunc(true))
-
-	ctx := context.Background()
-	rt := wazero.NewRuntime(ctx)
+func newRuntimeInstance(ctx context.Context,
```
timwu20: I feel like this function name should be changed given it doesn't return an `Instance` anymore.

lib/runtime/wazero/instance.go
```diff
@@ -36,13 +36,21 @@ const runtimeContextKey = contextKey("runtime.Context")
 
 var _ runtime.Instance = &Instance{}
 
+type cacheMetadata struct {
```
timwu20: maybe call this `wazeroMeta`? It's not just the cache anymore.

lib/runtime/wazero/instance.go
```diff
@@ -36,13 +36,21 @@ const runtimeContextKey = contextKey("runtime.Context")
...
+	config      wazero.RuntimeConfig
+	cache       wazero.CompilationCache
+	hostModule  wazero.CompiledModule
```
timwu20: why do we need to store this?

jimjbrettj: This is a nice find, i dont think we need to

lib/runtime/wazero/instance.go
jimjbrettj: updated: renamed to `newRuntime` since a runtime and associated data are still being created. Let me know if you prefer something else

EclesioMeloJunior: Agreed, maybe change to compiled modules

jimjbrettj: Hm but it returns one runtime and one module (since we don't cache host module anymore). Maybe something like `newRuntimeWithGuestModule`?

timwu20 (approved):

lib/runtime/wazero/instance.go
EclesioMeloJunior: yeah, we should only keep the guest compiled module since this is the piece that are going to be instantiated on every exec.

PR #3884, opened 2024-04-11
feat(pkg/trie): Create TrieDBIterator to easily traverse keys in TrieDB

dimartiro:
## Changes

This PR introduces the `TrieDBIterator` that is a helper structure to traverse Tries
This iterator is used in some `TrieDB` methods to simplify the implementation for those who need to traverse the trie and collect some information such as keys and values.

### Summary of changes

- [x] Create `TrieDBIterator`
- [x] Implement `NextKey(key)`, `Entries()`, `GetKeysWithPrefix(prefix)` using the iterator
- [x] Tests 

## Tests

```sh
make test
```

## Issues

#3806 

## Primary Reviewer

@EclesioMeloJunior


pkg/trie/triedb/iterator.go
```diff
@@ -3,14 +3,44 @@
...
 
 // NextKey returns the next key in the trie in lexicographic order.
 // It returns nil if no next key is found.
 func (t *TrieDB) NextKey(key []byte) []byte {
-	panic("not implemented yet")
+	iter := NewTrieDBIterator(t)
+
+	iter.Seek(key)
+	nextKey := iter.NextKey()
+	return nextKey
```
EclesioMeloJunior:
```suggestion
	return iter.NextKey()
```

pkg/trie/triedb/iterator.go
```diff
@@ -3,14 +3,44 @@
...
+// Endian formatted prefix in their key.
+func (t *TrieDB) GetKeysWithPrefix(prefix []byte) (keysLE [][]byte) {
+	iter := NewTrieDBIterator(t)
+	iter.Seek(prefix)
+
+	//Since seek consumes the prefix, we need to add it in the keys list
+	keys := [][]byte{prefix}
+
+	for key := iter.NextKey(); bytes.HasPrefix(key, prefix); key = iter.NextKey() {
+		keys = append(keys, key)
```
EclesioMeloJunior: is it possible that in the set of returned keys there is a key that has the prefix but is **after** the key w/o the prefix? I'm asking that because the for loop ends when the key does not have the prefix

pkg/trie/triedb/triedb_iterator.go
```diff
@@ -0,0 +1,118 @@
...
+func (i *TrieDBIterator) nextToVisit(parentKey []byte, node codec.Node) {
+	i.nodeStack = append(i.nodeStack, &iteratorState{
+		parentFullKey: parentKey,
+		node:          node,
+	})
+}
+
+// nextState pops the next node to visit from the stack
+func (i *TrieDBIterator) nextState() *iteratorState {
+	currentState := i.nodeStack[len(i.nodeStack)-1]
```
EclesioMeloJunior, line 79: can we check if `i.nodeStack` is not empty? or include a comment like `warn: this function does not check if the node stack is empty, this check is made at NextEntry method`

pkg/trie/triedb/triedb_iterator.go
```diff
@@ -0,0 +1,118 @@
...
+			key := currentState.fullKeyNibbles(nil)
+			value, err := i.db.loadValue(n.PartialKey, n.GetValue())
+			if err != nil {
+				panic("Error loading value")
+			}
+			return &Entry{key: key, value: value}
+		case codec.Branch:
+			// Reverse iterate over children because we are using a LIFO stack
+			// and we want to visit the leftmost child first
+			for idx := len(n.Children) - 1; idx >= 0; idx-- {
```
EclesioMeloJunior, line 100: is not the rightmost?
```
[0 <- left, ...., right -> 0]
```

pkg/trie/triedb/triedb_iterator.go
```diff
@@ -0,0 +1,118 @@
...
+	entry := i.NextEntry()
+	if entry != nil {
+		return entry.key
+	}
+	return nil
+}
+
+// Seek moves the iterator to the first key that is greater than the target key.
+func (i *TrieDBIterator) Seek(targetKey []byte) {
+	for key := i.NextKey(); bytes.Compare(key, targetKey) < 0; key = i.NextKey() {
```
EclesioMeloJunior, line 136: to check if `key` is greater than `targetKey` using `bytes.Compare` you should use `bytes.Compare(key, targetKey) > 0` https://pkg.go.dev/bytes#Compare
```suggestion
	for key := i.NextKey(); bytes.Compare(key, targetKey) > 0; key = i.NextKey() {
```

EclesioMeloJunior (changes requested):

pkg/trie/triedb/triedb_iterator.go
dimartiro: We are checking it [here](https://github.com/ChainSafe/gossamer/blob/d80fdae4f2e1b99f9a755e3a6469a37100abb756/pkg/trie/triedb/triedb_iterator.go#L65) that's the only place where the `nextState()` is call.
But I'll add the comment to let know that the caller should check it 👍

pkg/trie/triedb/triedb_iterator.go
dimartiro: nop, since we are using a stack, if we have children `[1,2,3]` we are going to stacking them as `[3,2,1]` so next time we pop from the stack the next node to visit will be `1` that is the leftmost

pkg/trie/triedb/triedb_iterator.go
dimartiro: But I want to advance the iterator until `key >= target` so I have to execute `NextKey()` while `key < targetKey`

pkg/trie/triedb/iterator.go
dimartiro: I don't think so, for example, if we have this trie:

image

And we want all the keys that share the prefix: `not`

We are first moving the iterator to the node with `key = not` using `iter.Seek(prefix)`:

image

- gray: nodes we already visited
- green: node we are currently visiting
- blue: pendings nodes to visit

We are going to visit "notable" and "notification" but we are going to stop visiting when the prefix is not shared anymore, for example for "test" node

dimartiro: One think I noticed now is that it could be expensive to skip a large amount of keys to seek the one we are looking for, we can change that with a smarter `Seek` method.
That will not only help us to speed up the prefix lookup process but also to create an iterator only for the keys we are interested in, and with that we can get rid of the `bytes.HasPrefix(key, prefix)` check too.

 I'm going to try to address it today.

pkg/trie/triedb/triedb_iterator.go
EclesioMeloJunior: I see but is confusing, since you still accessing the right most item in the array since you start from the last index hahah.

pkg/trie/triedb/triedb_iterator.go
EclesioMeloJunior: Ohh, right, yeah, that's correct you advance while key is lower than target, can you change the comment to 
```
Seek moves the iterator while key is lower than target key.
```

pkg/trie/triedb/triedb.go
```diff
@@ -17,6 +17,11 @@ import (
 
 var ErrIncompleteDB = errors.New("incomplete database")
 
+type Entry struct {
+	key   []byte
```
jimjbrettj, line 21: Do these need to be public for scale?

jimjbrettj (approved):

pkg/trie/triedb/triedb.go
dimartiro: Actually it doesn't need to be public, I changed it now! thks

pkg/trie/triedb/iterator.go
dimartiro: Improved! now we are traversing the trie going directly to the node with the prefix and creating an iterator from that node

pkg/trie/triedb/iterator.go
```diff
@@ -6,11 +6,38 @@ package triedb
...
+	return entries
 }
 
 // NextKey returns the next key in the trie in lexicographic order.
 // It returns nil if no next key is found.
 func (t *TrieDB) NextKey(key []byte) []byte {
-	panic("not implemented yet")
+	iter := NewTrieDBIterator(t)
+
+	// TODO: Seek will potentially skip a lot of keys, we need to find a way to
```
EclesioMeloJunior, line 24: Do you plan to address this TODO in this PR?

dimartiro: I'll create an issue and address it after having the writes working, I want to focus on that to finish the overall work about the tries and then focus on these improvements. Thanks for asking

pkg/trie/triedb/triedb_iterator.go
```diff
@@ -0,0 +1,137 @@
...
+)
+
+type iteratorState struct {
+	parentFullKey []byte     // key of the parent node of the actual node
+	node          codec.Node // actual node
+}
+
+// fullKeyNibbles return the full key of the node contained in this state
+// child is the child where the node is stored in the parent node
+func (s *iteratorState) fullKeyNibbles(child *int) []byte {
```
axaysagathiya: Add a full stop to differentiate both comments; when we hover over the function, it shows continuous text.
Screenshot from 2024-04-26 12-47-06
```suggestion
// fullKeyNibbles return the full key of the node contained in this state.
// child is the child where the node is stored in the parent node
func (s *iteratorState) fullKeyNibbles(child *int) []byte {
```

pkg/trie/triedb/triedb_iterator.go
```diff
@@ -0,0 +1,137 @@
...
+
+type TrieDBIterator struct {
+	db        *TrieDB          // trie to iterate over
+	nodeStack []*iteratorState // Pending nodes to visit
+}
+
+func NewTrieDBIterator(trie *TrieDB) *TrieDBIterator {
+	rootNode, err := trie.getRootNode()
+	if err != nil {
+		panic("trying to create trie iterator with incomplete trie DB")
```
axaysagathiya, line 38: `getRootNode()` has two possible errors that can be returned.
Why don't you panic with an actual error we get here?

pkg/trie/triedb/triedb_iterator.go
```diff
@@ -0,0 +1,137 @@
...
+	for len(i.nodeStack) > 0 {
+		currentState := i.nextState()
+		currentNode := currentState.node
+
+		switch n := currentNode.(type) {
+		case codec.Leaf:
+			key := currentState.fullKeyNibbles(nil)
+			value, err := i.db.loadValue(n.PartialKey, n.GetValue())
+			if err != nil {
+				panic("Error loading value")
```
axaysagathiya: Is it possible to add more context as to why it's failing?
Consider printing `err`.

pkg/trie/triedb/triedb_iterator.go
```diff
@@ -0,0 +1,137 @@
...
+						panic(err)
+					}
+					i.nextToVisit(currentState.fullKeyNibbles(&idx), childNode)
+				}
+			}
+			if n.GetValue() != nil {
+				key := currentState.fullKeyNibbles(nil)
+				value, err := i.db.loadValue(n.PartialKey, n.GetValue())
+				if err != nil {
+					panic("Error loading value")
```
axaysagathiya: Same as above,
Is it possible to add more context as to why it's failing?
Consider printing `err`.

pkg/trie/triedb/lookup.go
```diff
@@ -0,0 +1,131 @@
...
+	if value := node.GetValue(); value != nil {
+		return l.loadValue(node.GetPartialKey(), value)
+	}
+
+	return nil, nil
+}
+
+// loadValue gets the value from the node, if it is inlined we can return it
+// directly. But if it is hashed (V1) we have to look up for its value in the DB
+func (l *TrieLookup) loadValue(prefix []byte, value codec.NodeValue) ([]byte, error) {
```
axaysagathiya: By reading this function name, it seems you are loading(adding) some value.
But the comment says this function gets the value from the node. Also, I don't see any code to add value.

It's confusing for me. Can you explain why you chose this function name?

pkg/trie/triedb/triedb_iterator.go
kishansagathiya: I personally prefer to use panic only in extreme cases

jimjbrettj: I think if getting the rootNode fails and then corrupts our db or breaks storage, panic could be good. But I totally agree with @kishansagathiya that we need to make sure to use it only in extreme cases. For me, its pretty much if the correctness of the node is broken and we just need to shutdown.

However if we are going to panic, I think its a good question why not return the err as well in the panic?

pkg/trie/triedb/lookup.go
jimjbrettj: @axaysagathiya hmm I am a bit curious why you are thinking adding is involved here? For me a load function implies a lookup

axaysagathiya: @jimjbrettj yeah true,  but load is also used to add values. For example [load method of btree map](https://pkg.go.dev/github.com/kjx98/go-btree#Map.Load), it used to add values into map. 

considering this, to avoid confusion I would suggest using `getValue` or `fetchValue` method name

pkg/trie/triedb/triedb_iterator.go
dimartiro: Good point, I'll change it to return an error. Thanks for the comments!

dimartiro: This will make me modify all methods and the trie interface to start handling and returning the new error, I'll leave it for a future PR to prevent adding too many changes here

pkg/trie/triedb/lookup.go
dimartiro: I like the comment and to prevent ambiguity I'll change it to `fetchValue`, thanks Axay!

EclesioMeloJunior (approved):
LGTM

pkg/trie/triedb/triedb_iterator.go
kishansagathiya: > This will make me modify all methods and the trie interface to start handling and returning the new error, I'll leave it for a future PR to prevent adding too many changes here

If this is the case and if you have used a lot of panics that should otherwise just be errors, let's create an issue if we are going to do as a separate PR. It's okay to tackle this separately, but let's make this the next PR.

Btw, this panic thing is part of list of go proverbs. They seem like good rules to follow https://go-proverbs.github.io/

dimartiro: Totally agree! thanks, I'll address it in #3942

PR #3894, opened 2024-04-15
enhancement(network): Change log level of banned threshold error

jimjbrettj:
## Changes

- After investigating this log to see what the problem was, it does not seem to cause any problems that affect the node's behavior, so I change the log level from critical to debug

## Tests

```sh
```

## Primary Reviewer

@axaysagathiya

### DIFF
--- a/dot/peerset/constants.go
+++ b/dot/peerset/constants.go
@@ -54,7 +54,7 @@ const (
 	IncompleteHeaderReason = "Incomplete header"
 
 	// BannedThresholdValue used when we need to ban peer.
-	BannedThresholdValue Reputation = 82 * (math.MinInt32 / 100)
+	BannedThresholdValue Reputation = 71 * (math.MinInt32 / 100)
 	// BannedReason used when we need to ban peer.
 	BannedReason = "Banned"
 
--- a/dot/peerset/constants.go
+++ b/dot/peerset/constants.go
@@ -54,7 +54,7 @@ const (
 	IncompleteHeaderReason = "Incomplete header"
 
 	// BannedThresholdValue used when we need to ban peer.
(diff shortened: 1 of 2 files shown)


kishansagathiya (changes requested):
I have found this log quite useful while running gossamer network. When reputation is less than banned threshold, we remove that peer. If highest rated peer is below banned threshold, no one is connected to us. I would consider that as an important log to put as critical.

dimartiro:
> I have found this log quite useful while running gossamer network. When reputation is less than banned threshold, we remove that peer. If highest rated peer is below banned threshold, no one is connected to us. I would consider that as an important log to put as critical.

we can maybe add a `Critical` log for that case only, if no one is connected to us, and keep this as `Debug`

kishansagathiya:
> we can maybe add a Critical log for that case only, if no one is connected to us, and keep this as Debug

Seeing this message means we don't know any peers with enough reputation to be connected to us.

jimjbrettj:
@kishansagathiya it seems that we are hitting this case while we still do have peers connected, thus this does not seem critical to me. I think that critical logs should be reserved for cases where the node should shut down, and in this case, the node runs as I expect it to even with this error. I also do not see where we are removing the peer causing us to hit this case.

From my POV, I am getting spammed with a log that is not impacting the function of the node which feels not ideal. However, given the points you made above, I will do a little more investigating and see if I can modify how often this case gets hit, and perhaps change to error log instead. Will update this PR with my findings

jimjbrettj:
This PR does not fix the lower-level issue of why the log is occurring. Changing the log level therefore is not the best approach and thus I will continue to investigate the underlying issue here. I will close this PR since it is not needed and either open a new PR with the fix or create an issue to represent the results of my investigation.

